PR #1602, opened 2024-06-10
Git repo connector

LukeCarrier:
The existing GitHub and GitLab connectors seem to index only issues and pull requests. It would be nice for us to additionally index Git repo content.

This probably has a bunch of rough edges and needs some testing.

### DIFF
--- a/backend/danswer/configs/constants.py
+++ b/backend/danswer/configs/constants.py
@@ -75,6 +75,7 @@ class DocumentSource(str, Enum):
     GOOGLE_DRIVE = "google_drive"
     GMAIL = "gmail"
     REQUESTTRACKER = "requesttracker"
+    GIT = "git"
     GITHUB = "github"
     GITLAB = "gitlab"
     GURU = "guru"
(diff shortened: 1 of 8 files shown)


danielnaber:
The GitLab connector at least can index code, but I had to activate it in the source (`include_code_files: bool = True` instead of `include_code_files: bool = GITLAB_CONNECTOR_INCLUDE_CODE_FILES`) to make it work. There are other bugs, see e.g. [PR1767 ](https://github.com/danswer-ai/danswer/pull/1767)

yuhongsun96:
Hi @LukeCarrier, so sorry that I'm closing so many of your PRs, do you want to connect with us? Happy to give a bit of insight on the roadmap and areas to contribute if you're interested! You can always reach me at yuhong@danswer.ai

Code search can't be handled by just ingesting and chunking code files. It requires building a graph of the code base and other approaches that aren't directly keyword/vector chunk similarity. It's a completely separate flow which is why we haven't added the ability to ingest code via the GitHub Connector. It's DEFINITELY on the roadmap though

PR #1605, opened 2024-06-10
Add option to configure custom OpenAI-compatible model via env vars

sd109:
At the moment only 'well known' model providers can be configured automatically via env vars. To configure a custom provider, the admin account must set it up via the UI. This change makes it possible to configure a custom OpenAI compatible model provider (e.g. a HuggingFace model served via vLLM) entirely via the following env vars:
```
GEN_AI_DISPLAY_NAME: "My custom model name"
GEN_AI_MODEL_PROVIDER: custom
GEN_AI_LLM_PROVIDER_TYPE: openai
GEN_AI_MODEL_VERSION: model-org/model-name
GEN_AI_API_ENDPOINT: http://my-self-hosted-model.com
```

### DIFF
--- a/backend/danswer/configs/model_configs.py
+++ b/backend/danswer/configs/model_configs.py
@@ -61,6 +61,8 @@ SEARCH_DISTANCE_CUTOFF = 0
 GEN_AI_MODEL_PROVIDER = os.environ.get("GEN_AI_MODEL_PROVIDER") or "openai"
 # If using Azure, it's the engine name, for example: Danswer
 GEN_AI_MODEL_VERSION = os.environ.get("GEN_AI_MODEL_VERSION")
+# The fallback display name to use for default model when using a custom model provider
+GEN_AI_DISPLAY_NAME = os.environ.get("GEN_AI_DISPLAY_NAME") or "Custom LLM"
 
 # For secondary flows like extracting filters or deciding if a chunk is useful, we don't need
 # as powerful of a model as say GPT-4 so we can use an alternative that is faster and cheaper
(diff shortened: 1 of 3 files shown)


backend/danswer/llm/llm_initialization.py
```diff
@@ -28,49 +30,73 @@ def load_llm_providers(db_session: Session) -> None:
...
+            api_version=GEN_AI_API_VERSION,
+            custom_config={},
+            default_model_name=GEN_AI_MODEL_VERSION,
+            fast_default_model_name=FAST_GEN_AI_MODEL_VERSION,
+        )
 
-    if GEN_AI_MODEL_PROVIDER not in well_known_provider_name_to_provider:
-        logger.error(f"Cannot auto-transition LLM provider: {GEN_AI_MODEL_PROVIDER}")
-        return None
+    else:
```
sd109, line 54: Note, everything inside this `else` clause is identical to the previous code (GitHub diff makes this unclear) so the behaviour is only changed when `GEN_AI_MODEL_PROVIDER` is `custom`.

backend/danswer/configs/model_configs.py
```diff
@@ -61,6 +61,8 @@
 GEN_AI_MODEL_PROVIDER = os.environ.get("GEN_AI_MODEL_PROVIDER") or "openai"
 # If using Azure, it's the engine name, for example: Danswer
 GEN_AI_MODEL_VERSION = os.environ.get("GEN_AI_MODEL_VERSION")
+# The fallback display name to use for default model when using a custom model provider
+GEN_AI_DISPLAY_NAME = os.environ.get("GEN_AI_DISPLAY_NAME") or "Custom LLM"
```
clayrosenthal, line 65: Why not just use `os.environ.get("GEN_AI_DISPLAY_NAME", "Custom LLM")`?

sd109: I was trying to follow the pattern used everywhere else in this file 🙂

yuhongsun96:
We've added richer support for different LLMs via the UI. The backend env var based configuration is now only for developer ease of use purposes.

Thanks for the suggestion though!

PR #1606, opened 2024-06-10
Fix Helm Vespa resource limits

sd109:
The change made in https://github.com/danswer-ai/danswer/pull/1536 only changed the Vespa pod resource requests and not the corresponding limits. This means that Kubernetes refuses to created the Vespa pod because the requests must be less than or equal to the limits. This changes fixes the limits to match the requests by default.


yuhongsun96 (approved):
Looks good!

PR #1621, opened 2024-06-11
Add salesforce connector

hagen-danswer:


backend/danswer/connectors/salesforce/connector.py
```diff
@@ -0,0 +1,255 @@
...
+class SalesforceConnector(LoadConnector, PollConnector):
+    def __init__(
+        self,
+        batch_size: int = INDEX_BATCH_SIZE,
+        requested_objects: list[str] = [],
+    ) -> None:
+        self.batch_size = batch_size
+        self.sf_client: Salesforce | None = None
+        self.parent_object_list: list[str] = []
+        if requested_objects:
```
yuhongsun96, line 39: self.parent_object_list = [
    obj.capitalize() for obj in requested_objects
] if requested_objects else DEFAULT_PARENT_OBJECT_TYPES

backend/danswer/connectors/salesforce/connector.py
```diff
@@ -0,0 +1,255 @@
...
+            for requested_object in requested_objects:
+                if requested_object:  # ensure first letter is capitalized
+                    self.parent_object_list.append(
+                        requested_object[0].upper() + requested_object[1:]
+                    )
+        else:
+            self.parent_object_list = DEFAULT_PARENT_OBJECT_TYPES
+
+    def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None:
+        salesforce_username = credentials["sf_username"]
```
yuhongsun96, line 49: self.sf_client = Salesforce(
    username=credentials["sf_username"],
    password=credentials["sf_password"],
    security_token=credentials["sf_security_token"],
)

backend/danswer/connectors/salesforce/connector.py
```diff
@@ -0,0 +1,255 @@
...
+            password=salesforce_password,
+            security_token=salesforce_security_token,
+        )
+
+        return None
+
+    def _get_sf_type_object(self, type_name: str) -> SFType:
+        if self.sf_client is None:
+            raise ConnectorMissingCredentialError("Salesforce")
+        return SFType(type_name, self.sf_client.session_id, self.sf_client.sf_instance)
```
yuhongsun96, line 64: 1 liner functions aren't particularly useful what is the benefit of doing this?

backend/danswer/connectors/salesforce/connector.py
```diff
@@ -0,0 +1,255 @@
...
+        if self.sf_client is None:
+            raise ConnectorMissingCredentialError("Salesforce")
+        try:
+            user_object_info = self.sf_client.query(
+                f"SELECT Name FROM User WHERE Id = '{id}'"
+            )
+            name = user_object_info.get("Records", [{}])[0].get("Name", "Null User")
+            return name
+        except Exception:
+            return "Null User"
```
yuhongsun96, line 76: should log failures

backend/danswer/connectors/salesforce/connector.py
```diff
@@ -0,0 +1,255 @@
...
+        self.sf_client = Salesforce(
+            username=salesforce_username,
+            password=salesforce_password,
+            security_token=salesforce_security_token,
+        )
+
+        return None
+
+    def _get_sf_type_object(self, type_name: str) -> SFType:
+        if self.sf_client is None:
```
yuhongsun96, line 62: Not necessarily better but sometimes I do assert self.sf_client is not None instead when the way this code is called is fairly controlled and I know it won't be None

backend/danswer/connectors/salesforce/connector.py
```diff
@@ -0,0 +1,255 @@
...
+        except Exception:
+            return "Null User"
+
+    def _convert_object_instance_to_document(
+        self, object_dict: dict[str, Any]
+    ) -> Document:
+        if self.sf_client is None:
+            raise ConnectorMissingCredentialError("Salesforce")
+
+        extracted_id = object_dict["Id"]
```
yuhongsun96, line 84: it's too late for other connectors but let's try to prefix the IDs with the connector name to avoid collisions. Otherwise we have to depend on other connectors having good UUID systems. So like SALESFORCE_{}

yuhongsun96: In fact, we should probably have this as a central functionality and just make an exception for like "older connectors"

backend/danswer/connectors/salesforce/connector.py
```diff
@@ -0,0 +1,255 @@
...
+        extracted_link = f"https://{self.sf_client.sf_instance}/{extracted_id}"
+        extracted_doc_updated_at = time_str_to_utc(object_dict["LastModifiedDate"])
+        extracted_object_text = extract_dict_text(object_dict)
+        extracted_semantic_identifier = object_dict.get("Name", "Uknown Object Name")
```
yuhongsun96, line 88: typo, also probably "Unknown Object" is better

backend/danswer/connectors/salesforce/connector.py
```diff
@@ -0,0 +1,255 @@
...
+
+        fields = [
+            field.get("name")
+            for field in all_account_objects["fields"]
+            if field.get("type", "base64") != "base64"
+        ]
+
+        return fields
+
+    def _generate_query_per_parent_type(self, parent_sf_type: str) -> Iterator[str]:
```
yuhongsun96, line 171: Could you add a comment for how this might look? Helps reader understand what's going on

backend/danswer/connectors/salesforce/connector.py
```diff
@@ -0,0 +1,255 @@
...
+    def _fetch_from_salesforce(
+        self,
+        start: datetime | None = None,
+        end: datetime | None = None,
+    ) -> GenerateDocumentsOutput:
+        if self.sf_client is None:
+            raise ConnectorMissingCredentialError("Salesforce")
+
+        doc_batch: list[Document] = []
+        for parent_object_type in self.parent_object_list:
```
yuhongsun96, line 200: nit: log at the end of it how many of each type was processed

yuhongsun96 (approved):

yuhongsun96:
Just left some comments for your learning, changes are fine :)

backend/danswer/connectors/salesforce/connector.py
hagen-danswer: I will keep that in mind for the future but wont change it for now

backend/danswer/connectors/salesforce/connector.py
hagen-danswer: It performs 2 things now. I meant to do that earlier but i missed it

backend/danswer/connectors/salesforce/connector.py
hagen-danswer: i fixed it for this connector but wont be adding the central functionality just yet

PR #1629, opened 2024-06-12
Confluence Connector to Sync Child pages only

vggaur41522:
Update the confluence connector to sync only child pages of provided CONFLUENCE URL instead of entire space.

> Accept fully formed Wiki URL of format:
  - Space Level: https://danswer.atlassian.net/wiki/spaces/TDS/overview
  - Sub Page Level: https://danswer.atlassian.net/wiki/spaces/1234abcd/pages/5678efgh/overview
> Based on the shared URL (space level/Sub Page Level), the connector will scan in the following format:
  - Space Level: All the pages within the Space
  - Sub Page Level: Only provided parent page and its own child sub pages
> Also defaulted `break_on_unprocessable` to `False` for Confluence.


backend/danswer/connectors/confluence/connector.py
```diff
@@ -37,16 +37,17 @@
...
-# 3. Segment into Sections for more accurate linking, can split by headers but make sure no text/ordering is lost
+# 1. Include attachments, etc
+# 2. Segment into Sections for more accurate linking, can split by headers but make sure no text/ordering is lost
 
 
-def _extract_confluence_keys_from_cloud_url(wiki_url: str) -> tuple[str, str]:
+def _extract_confluence_keys_from_cloud_url(wiki_url: str) -> tuple[str, str, str]:
     """Sample
-    https://danswer.atlassian.net/wiki/spaces/1234abcd/overview
+    https://danswer.atlassian.net/wiki/spaces/1234abcd/pages/5678efgh/overview
```
Weves: nit: can we separate out these examples. E.g. `URL w/ page: https://danswer.atlassian.net/wiki/spaces/1234abcd/pages/5678efgh/overview` and `URL w/o page: https://danswer.atlassian.net/wiki/spaces/ASAM/overview`

backend/danswer/connectors/confluence/connector.py
```diff
@@ -76,26 +80,31 @@ def _extract_confluence_keys_from_datacenter_url(wiki_url: str) -> tuple[str, st
         + parsed_url.path.split(DISPLAY)[0]
     )
     space = DISPLAY.join(parsed_url.path.split(DISPLAY)[1:]).split("/")[0]
-    return wiki_base, space
+    page_id = parsed_url.path.split("/")[5]
```
Weves: I don't think this works in the general case, since afaik someone can put something like `https://danswer.ai/confluence/blah/blah/blah` as the wiki_base. In other words, I think we need to do the counting after `/display/`.

Also what happens if the page isn't included? This seems like it would break

Weves: Also how (if at all) did we verify that this works for data center? Is there a link to a guide that we can put at the start of the func that shows the standard URL structure of datacenter deployments?

That this works correctly is quite important since many of our customers are datacenter users

web/src/app/admin/connectors/confluence/page.tsx
```diff
@@ -207,13 +210,17 @@ const Main = () => {
...
-            entering{" "}
+            index either the entire page and its subpages OR the entire space.
+            For example, entering{" "}
             <i>
               https://danswer.atlassian.net/wiki/spaces/Engineering/overview
             </i>{" "}
             and clicking the Index button will index the whole{" "}
-            <i>Engineering</i> Confluence space.
+            <i>Engineering</i> Confluence space, but entering
+            https://pablosfsanchez.atlassian.net/wiki/spaces/Engineering/pages/164331/example+page
```
Weves: nit: change `pablosfsanchez` -> `danswer`

backend/danswer/connectors/confluence/connector.py
```diff
@@ -436,7 +541,6 @@ def _get_doc_batch(
                 page_text += attachment_text
                 comments_text = self._fetch_comments(self.confluence_client, page_id)
                 page_text += comments_text
-
                 doc_metadata: dict[str, str | list[str]] = {
```
Weves, line 632: nit: remove this (or use it)

web/src/app/admin/connectors/confluence/page.tsx
```diff
@@ -289,15 +295,22 @@ const Main = () => {
...
               validationSchema={Yup.object().shape({
                 wiki_page_url: Yup.string().required(
-                  "Please enter any link to your confluence e.g. https://danswer.atlassian.net/wiki/spaces/Engineering/overview"
+                  "Please enter any link to a Confluence space or Page e.g. https://danswer.atlassian.net/wiki/spaces/Engineering/overview"
                 ),
+                index_origin: Yup.boolean(),
               })}
               initialValues={{
                 wiki_page_url: "",
+                index_origin: false,
```
Weves: prefer default true, that feels like more of the intuitive, expected behavior (e.g. if you specify a page, that it would get indexed)

web/src/app/admin/connectors/confluence/page.tsx
```diff
@@ -207,13 +210,17 @@ const Main = () => {
...
+            index either the entire page and its subpages OR the entire space.
+            For example, entering{" "}
             <i>
               https://danswer.atlassian.net/wiki/spaces/Engineering/overview
             </i>{" "}
             and clicking the Index button will index the whole{" "}
-            <i>Engineering</i> Confluence space.
+            <i>Engineering</i> Confluence space, but entering
+            https://pablosfsanchez.atlassian.net/wiki/spaces/Engineering/pages/164331/example+page
+            will index that page&apos;s children (and optionally, its children).
```
Weves: I think `optionally, its children` should be something like `optionally, itself`?

backend/danswer/connectors/confluence/connector.py
```diff
@@ -286,8 +314,73 @@ def _fetch(start_ind: int, batch_size: int) -> Collection[dict[str, Any]]:
...
+        def _fetch_child_pages(
+            start_ind: int, batch_size: int
+        ) -> Collection[dict[str, Any]]:
+            get_page_by_id = make_confluence_call_handle_rate_limit(
+                confluence_client.get_page_by_id
+            )
+
+            child_pages: list[dict[str, Any]] = []
+
+            if self.index_origin:
```
Weves: since this is called on every batch, this will make the root page be included in every batch which is probably not what we want

backend/danswer/connectors/confluence/connector.py
```diff
@@ -286,8 +314,73 @@ def _fetch(start_ind: int, batch_size: int) -> Collection[dict[str, Any]]:
 
                 return view_pages
 
+        def _fetch_child_pages(
```
Weves: this seems like it only goes one level deep. If I imagine what someone wants to do when specifying a page, I feel like they'd generally want to index a whole "subspace" under a certain page. 

Maybe this could be pulled out into a separate `ConfluenceRecursivePageFetcher` which maintains its own state of possible child pages, etc. 🤔

backend/danswer/connectors/confluence/connector.py
```diff
@@ -200,6 +209,7 @@ class ConfluenceConnector(LoadConnector, PollConnector):
     def __init__(
         self,
         wiki_page_url: str,
+        index_origin: bool = False,
```
Weves: nit: prefer True here as well, that feels like more of the intuitive, expected behavior (e.g. if you specify a page, that it would get indexed)

Weves (changes requested):
Left a few comments. This is turning into a bigger change than expected 😅

backend/danswer/connectors/confluence/connector.py
```diff
@@ -76,26 +84,32 @@ def _extract_confluence_keys_from_datacenter_url(wiki_url: str) -> tuple[str, st
         + parsed_url.path.split(DISPLAY)[0]
     )
     space = DISPLAY.join(parsed_url.path.split(DISPLAY)[1:]).split("/")[0]
-    return wiki_base, space
+    page_id = parsed_url.path.split(DISPLAY)[1].split("/")[0]
```
Weves: Hmm, I'm not sure this one will work. If there's a URL without a page won't it give us a "page" but it'll actually be the space?

backend/danswer/connectors/confluence/connector.py
```diff
@@ -286,17 +442,42 @@ def _fetch(start_ind: int, batch_size: int) -> Collection[dict[str, Any]]:
 
                 return view_pages
 
+        def _fetch_page(batch_size: int) -> list[dict[str, Any]]:
+            indexer = RecursiveIndexer(
```
Weves: this will cause us to re-fetch all pages every batch, no? We should probably just do this once for a given run

Weves (changes requested):

backend/danswer/connectors/confluence/connector.py
pablonyx: Yep! Is/was only called once, but updated the naming to make this clearer

backend/danswer/connectors/confluence/connector.py
pablonyx: Believe this returns a null which should signal that it's a space, but I'll double check now to be sure!

backend/danswer/connectors/confluence/connector.py
Weves: Ah, I think for clarity / consistency it would be great if `_fetch_pages` always returned a list of pages of length `self.batch_size`. I see we don't do that here.

This also has some additional benefits, such as only fetching comments as needed, and then removing them from memory once we're done with the batch. LMK if that makes sense!

backend/danswer/connectors/confluence/connector.py
```diff
@@ -286,17 +447,39 @@ def _fetch(start_ind: int, batch_size: int) -> Collection[dict[str, Any]]:
 
                 return view_pages
 
+        def _fetch_page(start_ind: int, batch_size: int) -> list[dict[str, Any]]:
+            if self.recursive_indexer:
```
Weves: I would prefer to instantiate it here to (1) remove the duplication and (2) not instantiate it unless we are using the recursive approach

Weves (approved):

PR #1630, opened 2024-06-12
User Inviting and Disabling

yuhongsun96:


CONTRIBUTING.md
```diff
@@ -72,6 +72,10 @@ For convenience here's a command for it:
 python -m venv .venv
 source .venv/bin/activate
 ```
+
```
yuhongsun96, line 75: Why is this the case? I have it somewhere under my home dir ~/.virtual_envs and it's not related to the Danswer directory.

backend/danswer/dynamic_configs/users.py
```diff
@@ -0,0 +1,21 @@
+from typing import cast
```
yuhongsun96, line 1: Slight preference to have dynamic_configs only contain the interfaces/implementations of the storage option and not add specific feature functionality into here

backend/danswer/server/manage/administrative.py
```diff
@@ -231,3 +235,56 @@ def update_token_budget_settings(
     # Store the settings in the dynamic config store
     get_dynamic_config_store().store(TOKEN_BUDGET_SETTINGS, settings_json)
     return {"message": "Token budget settings updated successfully."}
+
+
+@router.put("/admin/users")
+def bulk_invite_users(
+    _: User | None = Depends(current_admin_user),
+    emails: list[str] = Body(..., embed=True),
```
yuhongsun96, line 243: super nit (doesn't matter): slight preference across the board to have the more important/variable arguments on top and the "less interesting" args under. So emails first, and the user dependency after

backend/danswer/server/manage/administrative.py
```diff
@@ -231,3 +235,56 @@ def update_token_budget_settings(
...
+    all_emails = list(set(emails) | set(get_invited_users()))
+    return write_invited_users(all_emails)
+
+
+@router.patch("/admin/remove-invited-user")
+def remove_invited_user(
+    user_email: UserByEmail,
+    _: User | None = Depends(current_admin_user),
+) -> int:
+    users = get_invited_users()
```
yuhongsun96, line 254: nit: for readability so reviewers don't need to check the function, users generally refer to User objects, in this case they're user_emails which are strings.

backend/danswer/server/manage/administrative.py
```diff
@@ -231,3 +235,56 @@ def update_token_budget_settings(
...
+@router.patch("/admin/block-user")
+def block_user(
+    user_email: UserByEmail,
+    _: User | None = Depends(current_admin_user),
+    db_session: Session = Depends(get_session),
+) -> None:
+    user_to_block = get_user_by_email(
+        email=user_email.user_email, db_session=db_session
+    )
+    if not user_to_block:
```
yuhongsun96, line 268: may want to log warning if user is already blocked

backend/danswer/server/manage/administrative.py
```diff
@@ -231,3 +235,56 @@ def update_token_budget_settings(
...
+@router.patch("/admin/unblock-user")
+def unblock_user(
+    user_email: UserByEmail,
+    _: User | None = Depends(current_admin_user),
+    db_session: Session = Depends(get_session),
+) -> None:
+    user_to_unblock = get_user_by_email(
+        email=user_email.user_email, db_session=db_session
+    )
+    if not user_to_unblock:
```
yuhongsun96, line 285: Same thing here with the logging

not a suggestion but I wonder if a user should not be returned to admin status immediately if they were blocked/disabled

backend/danswer/server/manage/users.py
```diff
@@ -67,13 +73,41 @@ async def demote_admin(
...
+) -> AllUsersResponse:
+    users = list_users(db_session, q=q)
+    accepted_emails = {user.email for user in users}
+    invited_emails = get_invited_users()
+    if q:
+        invited_emails = [
+            email for email in invited_emails if re.search(r"{}".format(q), email, re.I)
+        ]
+
+    return {
```
yuhongsun96, line 104: Nit: where possible use the Pydantic object instead of the dict

yuhongsun96 (commented):
Generally looks great! The functionality to actually block out the disabled users doesn't seem to exist yet though. We should definitely include that.

.gitignore
```diff
@@ -5,3 +5,5 @@
 .idea
 /deployment/data/nginx/app.conf
 .vscode/launch.json
+*.sw?
+package-lock.json
```
yuhongsun96, line 9: @Weves, is it standard to have the package-lock.json be ignored? Doesn't this enforce the version consistency for the frontend packages?

thoughtcodex: Ah, sorry, didn't see that it was already checked in, will back add in #1631.

CONTRIBUTING.md
thoughtcodex: Main point is that the virtual environment must not be inside the *danswer* project.  Will reword in #1631.

backend/danswer/server/manage/administrative.py
thoughtcodex: Better yet: users should probably not be allowed to block themselves ^_^

backend/danswer/server/manage/administrative.py
thoughtcodex: Can go either way, so maybe simpler is better?  FWIW, it's just as easy for the admin unblocking a user to adjust their privileges immediately afterwards.

PR #1647, opened 2024-06-17
Usage Report Generation

abdum99:
This PR adds a backend endpoint at `/api/admin/generate-usage-report` which fetches empty skeletons for ChatMessage and ChatSession.

The skeletons only store non-identifying message_id, session_id, user_id, and time stamps for correlation purposes.
The data is then written to CSV files which are stored in the postgres large object file store and sent back to the requesting admin as a single .zip file.

> [!Note]
> Currently the backend does _not_ respect the data range input. I'll be adding that in a separate PR that also adds pagination

image


backend/danswer/configs/constants.py
```diff
@@ -143,3 +143,11 @@ class FileOrigin(str, Enum):
     CHAT_IMAGE_GEN = "chat_image_gen"
     CONNECTOR = "connector"
     OTHER = "other"
+
+
+class ReportType(str, Enum):
```
yuhongsun96: Unused, can leave a #TODO if you want to keep it here when you extend it

backend/danswer/db/chat.py
```diff
@@ -155,6 +156,53 @@ def delete_chat_session(
     db_session.commit()
 
 
+class ChatSessionSkeleton:
```
yuhongsun96: We try to use pydantic classes throughout the code base. Let's keep it consistent

backend/danswer/db/chat.py
```diff
@@ -1,3 +1,4 @@
+import datetime
```
yuhongsun96: Slight preference for having this in a separate module like usage_export.py or something

backend/danswer/db/chat.py
```diff
@@ -202,6 +250,40 @@ def get_chat_messages_by_session(
...
+def get_empty_chat_messages_entries(
+    db_session: Session,
+) -> list[ChatMessageSkeleton]:
+    stmt = select(
+        ChatMessage.id, ChatMessage.chat_session_id, ChatMessage.time_sent
+    ).where(ChatMessage.message_type == MessageType.USER)
+
+    result = db_session.execute(stmt).all()
+
+    print(result)
```
yuhongsun96: We prefer not to use print in any of the main flows, we use the standard logger that we have throughout the code base.

yuhongsun96 (commented):
We need to port this to EE, let me give you access. We'll probably end up closing this PR

backend/danswer/db/chat.py
abdum99: oops good catch

PR #1650, opened 2024-06-17
Add support for indexing GitHub code files

artmatsak:
This PR adds indexing of GitHub code files, taking inspiration from https://github.com/danswer-ai/danswer/pull/1586. One notable difference from the GitLab PR is that we only index GitHub code files on initial load or complete re-indexing because there's currently no obvious way to define the "last updated on GitHub" date for a repository file. (We cannot use the last commit date because there can be an arbitrary delay between a commit being made and it actually being pushed to GitHub.)

This functionality is off by default and requires the `GITHUB_CONNECTOR_INCLUDE_CODE_FILES` env var to be set to `true`.


backend/danswer/configs/app_configs.py
```diff
@@ -188,6 +188,10 @@
 
 GITHUB_CONNECTOR_BASE_URL = os.environ.get("GITHUB_CONNECTOR_BASE_URL") or None
 
+GITHUB_CONNECTOR_INCLUDE_CODE_FILES = (
+    os.environ.get("GITHUB_CONNECTOR_INCLUDE_CODE_FILES", "").lower() == "true"
+)
+
```
os6sense, line 209: PR description would make a useful comment here.

yuhongsun96:
Hello! We're going to be building code search into the project in the feature (hopefully starting around end of year). The issue with ingesting code like with documentation or conversations is that code has dependencies across files that can't be captured well by a pure top n approach. We'll be building code ingestion through a different pipeline

PR #1652, opened 2024-06-17
Add connector document pruning task

hagen-danswer:


backend/alembic/versions/e209dc5a8156_added_prune_frequency.py
```diff
@@ -0,0 +1,28 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "e209dc5a8156"
+down_revision = "b85f02ec1308"
+branch_labels = None  # type: ignore
+depends_on = None  # type: ignore
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
```
yuhongsun96: let's clean this stuff :P

backend/danswer/background/celery/celery.py
```diff
@@ -173,42 +250,52 @@ def _sync_document_batch(document_ids: list[str], db_session: Session) -> None:
 # Periodic Tasks
 #####
 @celery_app.task(
-    name="check_for_document_sets_sync_task",
+    name="manage_all_tasks",
     soft_time_limit=JOB_TIMEOUT,
 )
-def check_for_document_sets_sync_task() -> None:
+def manage_all_tasks() -> None:
     """Runs periodically to check if any document sets are out of sync
```
yuhongsun96: this is no longer correct

backend/danswer/background/celery/celery_utils.py
```diff
@@ -21,3 +33,63 @@ def get_deletion_status(
         credential_id=credential_id,
         status=task_state.status,
     )
+
+
+def should_sync_doc_set(document_set: DocumentSet, db_session: Session) -> bool:
```
yuhongsun96, line 43: I like this, cleaner code. In this case, even though it's only used once, it still makes sense to pull it out

backend/danswer/background/celery/celery.py
```diff
@@ -173,42 +250,52 @@ def _sync_document_batch(document_ids: list[str], db_session: Session) -> None:
...
+                    )
+
 
 #####
 # Celery Beat (Periodic Tasks) Settings
 #####
 celery_app.conf.beat_schedule = {
-    "check-for-document-set-sync": {
-        "task": "check_for_document_sets_sync_task",
+    "manage-all-tasks": {
```
yuhongsun96: maybe call this like "run-tasks-scheduler" (below as well)

backend/danswer/background/celery/celery.py
```diff
@@ -173,42 +250,52 @@ def _sync_document_batch(document_ids: list[str], db_session: Session) -> None:
...
+            if should_sync_doc_set(document_set, db_session):
+                logger.info(f"Syncing the {document_set.name} document set")
                 sync_document_set_task.apply_async(
                     kwargs=dict(document_set_id=document_set.id),
                 )
 
+        # manage document pruning
+        all_connectors = fetch_connectors(db_session=db_session)
+
+        for connector in all_connectors:
```
yuhongsun96: Can we just get cc-pairs directly (which aren't disabled)

backend/danswer/background/task_utils.py
```diff
@@ -22,6 +22,10 @@ def name_document_set_sync_task(document_set_id: int) -> str:
     return f"sync_doc_set_{document_set_id}"
 
 
+def name_cc_pruning_task(connector_id: int, credential_id: int) -> str:
+    return f"pruning_connector_credential_pair_{connector_id}_{credential_id}"
```
yuhongsun96: super nit but being detail oriented is good, call this prune instead of pruning :|

backend/danswer/background/celery/celery.py
```diff
@@ -90,6 +98,75 @@ def cleanup_connector_credential_pair_task(
...
+    from the most recently pulled document ID list"""
+    with Session(get_sqlalchemy_engine()) as db_session:
+        try:
+            cc_pair = get_connector_credential_pair(
+                db_session=db_session,
+                connector_id=connector_id,
+                credential_id=credential_id,
+            )
+
+            if not cc_pair:
```
yuhongsun96, line 114: Should this log something? Maybe like "Skipping pruning of cc-pair with connector {id} and credential {id}" something like that

backend/danswer/background/celery/celery.py
```diff
@@ -90,6 +98,75 @@ def cleanup_connector_credential_pair_task(
...
+                return
+
+            runnable_connector, new_credential_json = instantiate_connector(
```
yuhongsun96: Since this logic is now duplicated, let's pull it out, place it somewhere reasonable, and reuse it

backend/danswer/background/celery/celery_utils.py
```diff
@@ -21,3 +33,63 @@ def get_deletion_status(
...
+        return False
+
+    if not last_pruning_task.start_time:
+        return False
+
+    time_since_last_pruning = current_db_time - last_pruning_task.start_time
+    return time_since_last_pruning.total_seconds() >= connector.prune_freq
+
+
+def extract_ids_from_runnable_connector(runnable_connector: BaseConnector) -> set[str]:
```
yuhongsun96, line 87: This logic doesn't cover all cases, if a connector is a Poll connector only, then it will never get pruned. If Poll connector (handle this last), then it should poll for docs from time 0 till now.

Have an else case that throws an exception. You want to put these like "traps" in the code so that if someone modifies connectors and say they create a new type, then this logic won't just be wrong. It will alert the dev by failing loudly.

So add an elif for case PollConnector and an else case which just raises an exception with some useful description

backend/danswer/connectors/interfaces.py
```diff
@@ -50,6 +50,12 @@ def poll_source(
         raise NotImplementedError
 
 
+class PruneConnector(BaseConnector):
+    @abc.abstractmethod
+    def prune_connector(self) -> set[str]:
```
yuhongsun96: Don't call it this. It's not actually pruning. This is just an efficient way of getting IDs

backend/danswer/background/celery/celery.py
```diff
@@ -90,6 +98,75 @@ def cleanup_connector_credential_pair_task(
...
+            if new_credential_json is not None:
+                backend_update_credential_json(
+                    cc_pair.credential, new_credential_json, db_session
+                )
+
+            all_connector_doc_ids: set[str] = extract_ids_from_runnable_connector(
+                runnable_connector
+            )
+
+            all_indexed_document_ids: set[str] = set()
```
yuhongsun96: all_indexed_document_ids = {doc.id for doc in get_documents_for_connector_credential_pair(
    db_session=db_session,
    connector_id=connector_id,
    credential_id=credential_id,
)}

backend/danswer/configs/app_configs.py
```diff
@@ -175,7 +175,8 @@
 
 # Avoid to get archived pages
 CONFLUENCE_CONNECTOR_INDEX_ONLY_ACTIVE_PAGES = (
-    os.environ.get("CONFLUENCE_CONNECTOR_INDEX_ONLY_ACTIVE_PAGES", "").lower() == "true"
+    os.environ.get("CONFLUENCE_CONNECTOR_INDEX_ONLY_ACTIVE_PAGES", "").lower()
```
yuhongsun96: why is this flipped now?

yuhongsun96: this type of logic always has to be "true" otherwise it makes no sense. otherwise if you set CONFLUENCE_CONNECTOR_INDEX_ONLY_ACTIVE_PAGES=True, which is not == "False" so CONFLUENCE_CONNECTOR_INDEX_ONLY_ACTIVE_PAGES becomes False when you set it to True... this is giga troll no?

yuhongsun96 (changes requested):
Nice work :D

backend/danswer/background/celery/celery.py
hagen-danswer: ):<

web/src/app/admin/connectors/google-sites/page.tsx
```diff
@@ -114,6 +114,7 @@ export default function GoogleSites() {
                           zip_path: filePaths[0],
                         },
                         refresh_freq: null,
+                        prune_freq: null,
```
Weves: should be 0?

web/src/app/admin/connectors/file/page.tsx
```diff
@@ -122,6 +122,7 @@ const Main = () => {
                         file_locations: filePaths,
                       },
                       refresh_freq: null,
+                      prune_freq: null,
```
Weves: should be 0?

web/src/components/admin/connectors/ConnectorForm.tsx
```diff
@@ -144,6 +146,7 @@ export function ConnectorForm<T extends Yup.AnyObject>({
             input_type: inputType,
             connector_specific_config: connectorConfig,
             refresh_freq: refreshFreq || 0,
+            prune_freq: pruneFreq || null,
```
Weves: if `pruneFreq` is 0, this will set it to `null` which is not intended I believe?

backend/danswer/background/celery/celery.py
```diff
@@ -173,42 +248,50 @@ def _sync_document_batch(document_ids: list[str], db_session: Session) -> None:
...
-
-                logger.info(f"Document set {document_set.id} syncing now!")
+            if should_sync_doc_set(document_set, db_session):
+                logger.info(f"Syncing the {document_set.name} document set")
                 sync_document_set_task.apply_async(
                     kwargs=dict(document_set_id=document_set.id),
                 )
 
+        # manage document pruning
+        all_cc_pairs = get_connector_credential_pairs(db_session)
```
Weves, line 279: prefer this to be a separate task whcih only checks for document pruning. The reason here is to try and have as much isolation as possible between these jobs—a failure in the document set sync check should not cause pruning to not run

Weves (changes requested):
Looks great! A few small comments

Weves (approved):

PR #1666, opened 2024-06-19
Internet Search

rashad-danswer:
Added an internet search tool supported by the Bing Search API

### Assistant Creation:
Screenshot 2024-06-19 at 9 26 17 AM

### Example Query:
Screenshot 2024-06-19 at 9 28 24 AM

### Assistant with both internal and internet search
image


backend/alembic/versions/141b02e819a6_merge_heads.py
```diff
@@ -0,0 +1,22 @@
+"""merge heads
```
yuhongsun96: Just to keep things simple for the reader, we have everything as a linear set of migrations with no branching. For PRs we just point the most recent change to whatever the HEAD is

backend/danswer/search/enums.py
```diff
@@ -23,6 +23,7 @@ class SearchType(str, Enum):
     KEYWORD = "keyword"
     SEMANTIC = "semantic"
     HYBRID = "hybrid"
+    INTERNET = "internet"
```
yuhongsun96: I think it's better if this is another Enum. This sources thing is for all the different documents that exist in the document index. We don't index the internet search results into the document index so this would create confusion

backend/danswer/configs/chat_configs.py
```diff
@@ -70,3 +70,6 @@
 
 # The backend logic for this being True isn't fully supported yet
 HARD_DELETE_CHATS = False
+
+# Internet Search
+BING_API_KEY = os.environ.get("BING_API_KEY") or None
```
yuhongsun96, line 83: If not set, I think this tool should not show up in the UI. Alternative is to have the tool but if they try to save the assistant then it should disallow it and give the error.

yuhongsun96: So you probably need an API endpoint to check this. At sommmmeeeee future point, we'll likely want to allow the admin to provide a key in the UI, but not now.

backend/danswer/tools/internet_search/internet_search_tool.py
```diff
@@ -0,0 +1,207 @@
...
+        source_type=DocumentSource.WEB,
+        metadata={},
+        updated_at=datetime.now(),
+        link=result.link,
+        source_links={0: result.link},
+    )
+
+
+class InternetSearchTool(Tool):
+    NAME = "run_internet_search"
```
yuhongsun96: Do we want to have a nice "display name" be standard for all the different tools and user specified tools? If not, I think that's fine too, this is a one off special tool anyhow

backend/danswer/tools/tool_selection.py
```diff
@@ -0,0 +1,65 @@
...
+    tools_and_args: list[tuple[Tool, dict[str, Any]]],
+    history: list[PreviousMessage],
+    query: str,
+    llm: LLM,
+) -> tuple[Tool, dict[str, Any]] | None:
+    if len(tools_and_args) == 1:
+        return tools_and_args[0]
+
+    tool_list_str = "\n".join(
+        f"""{ind}: {tool.name()} ({args}) - {tool.description()}"""
```
yuhongsun96: may want to do a lstrip on this. Also another potential improvement is to separate out the tools with ```, this seems to work pretty decent.
```
n: tool_name (args) - stuff
```

```
n+1: etc.
```

backend/danswer/tools/internet_search/internet_search_tool.py
```diff
@@ -0,0 +1,207 @@
...
+
+        logger.debug(
+            f"Evaluated if should use internet search: {use_internet_search_output}"
+        )
+
+        return (
+            YES_INTERNET_SEARCH.split()[0]
+        ).lower() in use_internet_search_output.lower()
+
+    def get_args_for_non_tool_calling_llm(
```
yuhongsun96, line 148: We may want a slightly different prompt for internet search. Currently it's the same as the Hybrid Index search rephrasing which probably isn't optimal. Currently looks something like the below:

Given the following conversation and a follow up input, rephrase the follow up into a SHORT, standalone query (which captures any relevant context from previous messages) for a vectorstore.
IMPORTANT: EDIT THE QUERY TO BE AS CONCISE AS POSSIBLE. Respond with a short, compressed phrase with mainly keywords instead of a complete sentence.
If there is a clear change in topic, disregard the previous messages.
Strip out any information that is not relevant for the retrieval task.
If the follow up message is an error or code snippet, repeat the same input back EXACTLY.

backend/danswer/tools/internet_search/internet_search_tool.py
```diff
@@ -0,0 +1,207 @@
...
+                },
+            },
+        }
+
+    def check_if_needs_internet_search(
+        self,
+        query: str,
+        history: list[PreviousMessage],
+        llm: LLM,
+        force_run: bool = False,
```
yuhongsun96: not needed right?

backend/danswer/tools/internet_search/internet_search_tool.py
```diff
@@ -0,0 +1,207 @@
...
+        metadata={},
+        updated_at=datetime.now(),
+        link=result.link,
+        source_links={0: result.link},
+    )
+
+
+class InternetSearchTool(Tool):
+    NAME = "run_internet_search"
+    DESCRIPTION = "Perform an internet search for up-to-date information."
```
yuhongsun96: can call these like _DESCRIPTION and make name and description properties

    @property
    def name(self) -> str:
        return self._NAME

    @property
    def description(self) -> str:
        return self._DESCRIPTION

backend/danswer/tools/internet_search/internet_search_tool.py
yuhongsun96: by the way, this comment isn't for this line specifically, I just don't have the place in the diff to add it :P

backend/danswer/tools/search/search_tool.py
```diff
@@ -57,6 +57,7 @@ class SearchResponseSummary(BaseModel):
 
 class SearchTool(Tool):
     NAME = "run_search"
+    DESCRIPTION = search_tool_description
```
yuhongsun96: Nit thing that was not even your change but since you're touching this, might as well touch it up. The module level var should be called SEARCH_TOOL_DESCRIPTION imo

backend/danswer/tools/tool_selection.py
```diff
@@ -0,0 +1,65 @@
...
+
+    history_str = combine_message_chain(
+        messages=history,
+        token_limit=GEN_AI_HISTORY_CUTOFF,
+    )
+    prompt = SINGLE_TOOL_SELECTION_PROMPT.format(
+        tool_list=tool_list_str, chat_history=history_str, query=query
+    )
+    output = message_to_string(llm.invoke(prompt))
+    try:
```
yuhongsun96, line 61: I think there should be some sort of regex match to try to extract the int or the tool names as some smaller/worse models may fail to do exactly as instructed. We can even just default to the first tool if it completely fails. We saw with Llama2 that it always tried to elaborate on its answers since it was trained to do this for better answer quality (kind of a forced Chain of Thought)

ie. "use tool run_internet_search which is tool 1 because danswer is not a common term and may be found using an internet search"

so we can try to extract the first number within the valid range, fall back to checking for any of the tool names, fall back to using the first tool

backend/danswer/chat/process_message.py
```diff
@@ -446,6 +486,13 @@ def stream_chat_message_objects(
                     tool_dict[db_tool_model.id] = [
                         ImageGenerationTool(api_key=dalle_key)
                     ]
+                elif tool_cls.__name__ == InternetSearchTool.__name__:
+                    bing_api_key = BING_API_KEY
```
yuhongsun96, line 520: Totally not needed but I wonder if it makes sense to have an internet search tool interface and have bing be an implementation of it. That way the bing specific implementation details isn't nested deeply into other flows

backend/danswer/chat/process_message.py
```diff
@@ -446,6 +486,13 @@ def stream_chat_message_objects(
...
+                    if not bing_api_key:
+                        raise ValueError("Internet search tool requires a Bing API key")
```
yuhongsun96: Can maybe add "please contact your Danswer admin to get it added!" We always want users to have some actionable way to get unblocked.

yuhongsun96 (approved):
Really great work! If you find some of my comments a bit too nit, that's just for now, just to share how I think about things :). In the future, I will be commenting less on details.

Another note - I don't review frontend code as it's not my area of expertise.

backend/danswer/tools/tool_selection.py
yuhongsun96: ah, Github formatted the markdown but you get the point :P

yuhongsun96:
Oh, one more general comment. Whenever you're putting new prompts and LLM interactions into the code, be sure to try some variation so we're not leaving easy improvements on the table. So just a little sprinkle of "prompt engineering" and try to verify that it works with gpt-3.5-turbo at least somewhat.

backend/danswer/tools/internet_search/internet_search_tool.py
rashad-danswer: This is a good idea. Currently the display names are hardcoded into the frontend, but I'll separate this out into the backend.

backend/danswer/chat/process_message.py
rashad-danswer: I did this in my other search project, but I made the interface once someone specifically requested another search provider. What do you think about this?

backend/danswer/search/enums.py
rashad-danswer: Got it, I removed this. For now, the `is_internet` flag is included in SavedSearchDoc, let me know if you had something else in mind though

FWangZil:
Great, hopefully this feature will be incorporated into the trunk branch soon

PR #1671, opened 2024-06-20
Transform HTML links to markdown behind config option

KevinShiCA:
adds an app config `ZENDESK_CONNECTOR_TRANSFORM_LINKS` that can be set to `strip` (default) or `markdown`. when set to strip links are turned into plain text with the link dropped, when set to markdown html links become markdown links.

unsure of docker setup, do we need to add `ZENDESK_CONNECTOR_TRANSFORM_LINKS` somewhere so that the environment variable can be passed through?

### DIFF
--- a/backend/danswer/file_processing/html_utils.py
+++ b/backend/danswer/file_processing/html_utils.py
@@ -39,7 +39,8 @@ def format_element_text(element_text: str, link_href: str | None) -> str:
 
     if (
         not link_href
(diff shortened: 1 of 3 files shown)


backend/danswer/file_processing/html_utils.py
```diff
@@ -98,7 +110,10 @@ def format_document_soup(
             elif in_table:
                 # don't handle other cases while in table
                 pass
-
+            elif e.name == "a":
+                link_href = e.get("href", None)
```
KevinShiCA: this was from chatgpt, is there a way to test this?

PR #1672, opened 2024-06-20
add slack groups to user response list

hagen-danswer:


backend/danswer/danswerbot/slack/handlers/handle_message.py
```diff
@@ -271,11 +273,17 @@ def handle_message(
         return False
 
     if respond_team_member_list:
-        send_to, _ = fetch_userids_from_emails(respond_team_member_list, client)
+        user_ids, _ = fetch_userids_from_emails(respond_team_member_list, client)
+        send_to = (send_to + user_ids) if send_to else user_ids
```
yuhongsun96: As discussed, can we remove the extra logic and keep the "set" throughout until we need it as a list.

yuhongsun96 (approved):

PR #1694, opened 2024-06-23
Feature/scroll

pablodanswer:
2 new basic hooks + redone streaming functionality
- Simplicity is key- no need to over-engineer the scroll!

Functionality + Tests
- Initial scroll: should scroll to input on message being submitted
	- [x] Normal case (message history on screen already)
	- [x] Input > screen size
	- [x] Scrolled all the way to the top
- Chat input bar resize: should scroll up / down if within locking distance and chat bar height resizes
	- [x] Normal case
	- [x] No text
	- [x] During streaming
- Streaming scroll: should smoothly scroll with streaming
	- [x] Normal text generation
	- [x] Input > screen size
	- [x] GPT 3.5 and "count to 100"
- "Scroll to bottom" button
	- [x] Normal case
- Chat session change: should scroll to the bottom of the page
	- [x] Normal case
	- [x] Short page
	- [x] Long


Weves (approved):
🎉

PR #1698, opened 2024-06-24
fixed confluence breaking on unknown filetypes

hagen-danswer:


backend/danswer/file_processing/extract_file_text.py
```diff
@@ -250,19 +294,30 @@ def epub_to_text(file: IO[Any]) -> str:
...
 def extract_file_text(
     file_name: str | None,
     file: IO[Any],
     break_on_unprocessable: bool = True,
 ) -> str:
+    if detect_encoding(file).lower() in TEXT_ENCODINGS:
+        return file_io_to_text(file, break_on_unprocessable)
     if not file_name:
-        return file_io_to_text(file)
+        logger.warning("No file_name and not known text encoding")
```
yuhongsun96: unknown is a word XD but doesn't matter

backend/danswer/file_processing/extract_file_text.py
```diff
@@ -251,47 +262,41 @@ def epub_to_text(file: IO[Any]) -> str:
 
 
 def file_io_to_text(file: IO[Any]) -> str:
-    encoding = detect_encoding(file)
-    file_content_raw, _ = read_text_file(file, encoding=encoding)
-    return file_content_raw
+    try:
```
yuhongsun96: This try should not be here, it can be a try except in the Confluence connector instead. It should still fail loudly in other flows. For example if someone uploads a wrong file type, it should not just be as if they uploaded nothing. It should fail

PR #1703, opened 2024-06-25
Ignore messages from Slack's official bot

hagen-danswer:


hagen-danswer:
I filtered out for the ID "USLACKBOT". i wasn't able to source the info from slack itself but it seems to be slack's official bot ID. It kinda makes sense because normally the "user" field has a UUID string instead of a name like "USLACKBOT".

sources:
[this stackoverflow post](https://stackoverflow.com/questions/40679819/slack-api-returns-is-bot-false-for-a-bot-how-else-to-check-if-an-user-is-bot)
[this github post](https://github.com/slackapi/python-slack-sdk/issues/1260)

yuhongsun96 (changes requested):
I have some fear about removing the existing check. If our users hit this again for the third time, it would be a very very very bad look. Let's keep in the existing check and also check for the official slack bot user id. Just to be extra safe, the check should also screen for:

If all of the following is true, also throw it out:
"first_name": "slackbot", "real_name": "Slackbot", "display_name": "Slackbot", "name": "slackbot"

I think specifically in this case, having a more safe "safety-net" is higher priority than code cleanliness

backend/danswer/danswerbot/slack/listener.py
```diff
@@ -93,15 +82,11 @@ def prefilter_requests(req: SocketModeRequest, client: SocketModeClient) -> bool
             return False
 
         if (
-            msg in _SLACK_GREETINGS_TO_IGNORE
-            or remove_danswer_bot_tag(msg, client=client.web_client)
-            in _SLACK_GREETINGS_TO_IGNORE
+            req.payload.setdefault("event", {}).get("user", "")
```
yuhongsun96, line 85: isn't this a user object? I dont think this gives back a string representing the ID

yuhongsun96 (approved):

PR #1704, opened 2024-06-25
Disallowed simultaneous pruning jobs

hagen-danswer:


backend/danswer/background/celery/celery_utils.py
```diff
@@ -84,25 +101,36 @@ def should_prune_cc_pair(
     return time_since_last_pruning.total_seconds() >= connector.prune_freq
 
 
+def rate_limit_doc_generation(doc_batch: list[Document]) -> set[str]:
```
Weves: this name doesn't feel accurate. It doesn't actually do any rate limiting. This is more like `document_batch_to_ids` and the rate limiting happens when wrapping this function

backend/danswer/configs/app_configs.py
```diff
@@ -213,6 +213,13 @@
 
 DEFAULT_PRUNING_FREQ = 60 * 60 * 24  # Once a day
 
+PREVENT_SIMULTANEOUS_PRUNING = (
+    os.environ.get("PREVENT_SIMULTANEOUS_PRUNING", "").lower() == "true"
+)
+
+# This is the maxiumum rate at which documents are queried for a pruning job. 0 disables the limitation.
+MAX_PRUNING_DOCUMENT_RETRIEVAL_PER_MINUTE = 0
```
Weves: this should be set via env variable?

backend/danswer/db/tasks.py
```diff
@@ -15,7 +15,7 @@ def get_latest_task(
 ) -> TaskQueueState | None:
     stmt = (
         select(TaskQueueState)
-        .where(TaskQueueState.task_name == task_name)
+        .where(TaskQueueState.task_name.like(f"%{task_name}%"))
```
Weves, line 35: why is this change needed? This feels like it should be a separate function (e.g. one for getting a task based on a specific name and another for getting the latest of a "family" of tasks)

backend/danswer/background/celery/celery_utils.py
```diff
@@ -73,6 +85,11 @@ def should_prune_cc_pair(
             return True
         return False
 
+    if PREVENT_SIMULTANEOUS_PRUNING and most_recent_generic_pruning_task:
+        if check_live_task_not_timed_out(most_recent_generic_pruning_task, db_session):
```
Weves: nit: while we're at it can we also make this name more descriptive. Right now it's not clear from the name what the behavior is if it's not live. E.g. something like `check_task_is_live_and_not_timed_out`

Weves (changes requested):

backend/danswer/background/celery/celery_utils.py
```diff
@@ -84,25 +103,36 @@ def should_prune_cc_pair(
...
         doc_batch_generator = runnable_connector.poll_source(start=start, end=end)
-        for doc_batch in doc_batch_generator:
-            all_connector_doc_ids.update(doc.id for doc in doc_batch)
     else:
         raise RuntimeError("Pruning job could not find a valid runnable_connector.")
 
+    if doc_batch_generator:
+        doc_generation_func = document_batch_to_ids
+        if MAX_PRUNING_DOCUMENT_RETRIEVAL_PER_MINUTE:
+            doc_generation_func = rate_limit_builder(
```
Weves: nit: should be more accurately called something like `doc_batch_processing_func`

Weves (approved):

PR #1705, opened 2024-06-25
Blob Storage

pablodanswer:
Credit for original base logic goes to `Matsak`

Adds most common blob storage methods
- AWS S3
- Oracle Storage
- Google Cloud Storage
- Cloudflare R2

Indexes all files within the blob storage buckets, but only analyzes the textual content / files.

Tested with 
- [x] AWS setup
- [x] Oracle setup
- [x] Google Cloud setup
- [x] Cloudflare setup
- [x] Invalid credentials
- [x] Invalid paths
- [x] Image files
- [x] Text content


backend/danswer/connectors/blob/connector.py
```diff
@@ -0,0 +1,275 @@
...
+
+        Raises ConnectorMissingCredentialError if required credentials are missing.
+        Raises ValueError for unsupported bucket types.
+        """
+
+        logger.info(
+            f"Loading credentials for {self.bucket_name} or type {self.bucket_type}"
+        )
+
+        if self.bucket_type == "R2":
```
Weves: nit: can we use a string Enum here? e.g. something like `self.bucket_type == BlobType.R2` and then `bucket_type: BlobType` for the class args?

backend/danswer/connectors/blob/connector.py
```diff
@@ -0,0 +1,275 @@
...
+
+        return None
+
+    def _download_object(self, key: str) -> bytes:
+        if self.s3_client is None:
+            raise ConnectorMissingCredentialError("Blob storage")
+        object = self.s3_client.get_object(Bucket=self.bucket_name, Key=key)
+        return object["Body"].read()
+
+    # def _get_presigned_url(self, key: str) -> str:
```
Weves, line 132: can we add a comment as to why we've left this in? E.g. it might be useful under X future circumstance

backend/danswer/connectors/blob/connector.py
```diff
@@ -0,0 +1,275 @@
...
+                        break_on_unprocessable=False,
+                    )
+                    batch.append(
+                        Document(
+                            id=f"{self.bucket_type}:{self.bucket_name}:{obj['Key']}",
+                            sections=[Section(link=link, text=text)],
+                            source=DocumentSource(self.bucket_type.lower()),
+                            semantic_identifier=name,
+                            doc_updated_at=last_modified,
+                            metadata={"type": "blob"},
```
Weves: this will show up in the UI / as a potential tag. I'm not sure this adds too much value, so I slightly prefer removing it

Weves (approved):
Overall looks great. Left a few small comments

backend/requirements/default.txt
```diff
@@ -72,3 +72,4 @@ zulip==0.8.2
 hubspot-api-client==8.1.0
 zenpy==2.0.41
 dropbox==11.36.2
+boto3-stubs[s3]==1.34.133
```
Weves: can we put this in the `dev.txt`

web/src/lib/types.ts
```diff
@@ -219,6 +223,32 @@ export interface ZendeskConfig {}
 
 export interface DropboxConfig {}
 
+export type BlobType = "S3" | "R2" | "GOOGLE_CLOUD_STORAGE" | "OCI_STORAGE";
```
Weves: these should be lowercase now?

backend/danswer/configs/constants.py
```diff
@@ -100,6 +100,17 @@ class DocumentSource(str, Enum):
     CLICKUP = "clickup"
     MEDIAWIKI = "mediawiki"
     WIKIPEDIA = "wikipedia"
+    S3 = "s3"
+    R2 = "r2"
+    GOOGLE_CLOUD_STORAGE = "google_cloud_storage"
+    OCI_STORAGE = "oci_storage"
+
+
+class BlobType(Enum):
```
Weves: I believe this needs to be `class BlobType(str, Enum):`

Otherwise,

```
from enum import Enum

class BlobType(Enum):
    R2 = "r2"
    S3 = "s3"
    GOOGLE_CLOUD_STORAGE = "google_cloud_storage"
    OCI_STORAGE = "oci_storage"

print(BlobType.R2 == "r2")
```

gives False 😓

web/src/lib/types.ts
pablonyx: Meant to delete that! Not used anywhere!

PR #1713, opened 2024-06-26
Added TTL to EE Celery tasks

hagen-danswer:


backend/danswer/db/pg_file_store.py
```diff
@@ -131,3 +138,25 @@ def delete_pgfilestore_by_file_name(
     db_session: Session,
 ) -> None:
     db_session.query(PGFileStore).filter_by(file_name=file_name).delete()
+
+
+def delete_chat_files_older_than(days_old: int, db_session: Session) -> None:
```
Weves: I would prefer to do this based on the `files` field in the `ChatMessage` table. This way we don't have to add any new columns + it feels more intuitive (e.g. we go through all old chat sessions, and then delete everything about each chat session including all messages, files, search docs, etc.)

backend/ee/danswer/background/celery/celery_app.py
```diff
@@ -23,30 +26,47 @@
 global_version.set_ee()
 
 
+@build_celery_task_wrapper(name_user_group_sync_task)
```
Weves, line 27: Did we check to see if user group syncing still works? 😅

backend/ee/danswer/background/celery/celery_app.py
```diff
@@ -103,5 +111,9 @@ def autogenerate_usage_report_task() -> None:
         "task": "autogenerate_usage_report_task",
         "schedule": timedelta(days=30),  # TODO: change this to config flag
     },
+    "check-ttl-management": {
+        "task": "check_ttl_management_task",
+        "schedule": timedelta(seconds=5),
```
Weves: I don't think we need to do this every 5 seconds. Maybe once every hour?

Weves (changes requested):

backend/ee/danswer/background/celery/celery_app.py
hagen-danswer: I agree, yuhong suggested 5 seconds but it doesn't seem like something that needs that level of fidelity (especially when simultaneous pruning jobs are disabled).

backend/alembic/versions/1639b454cdca_add_cascade.py
```diff
@@ -0,0 +1,72 @@
...
+    )
+    op.create_foreign_key(
+        "chat_message__search_doc_chat_message_id_fkey",
+        "chat_message__search_doc",
+        "chat_message",
+        ["chat_message_id"],
+        ["id"],
+        ondelete="CASCADE",
+    )
+
```
Weves: we probably also need another one for `chat_message__search_doc` -> `search_doc` to ensure that the `search_doc`s are deleted as well?

backend/danswer/db/pg_file_store.py
```diff
@@ -73,6 +92,22 @@ def delete_lobj_by_id(
     pg_conn.lobject(lobj_oid).unlink()
 
 
+def delete_lobj_by_name(
+    lobj_name: str,
+    db_session: Session,
+) -> None:
+    pgfilestore = db_session.query(PGFileStore).filter_by(file_name=lobj_name).first()
```
Weves: nit: can we use `get_pgfilestore_by_file_name` here?

backend/danswer/db/chat.py
```diff
@@ -139,25 +174,33 @@ def delete_chat_session(
...
-            ChatMessage.chat_session_id == chat_session_id
-        )
-        db_session.execute(stmt_messages)
+        delete_files_from_chat_session(chat_session_id, db_session)
 
-        stmt = delete(ChatSession).where(ChatSession.id == chat_session_id)
-        db_session.execute(stmt)
+        db_session.execute(delete(ChatSession).where(ChatSession.id == chat_session_id))
 
+        delete_orphaned_search_docs(db_session)
```
Weves: may not need this if we try the suggestion in the migration file

backend/danswer/db/chat.py
```diff
@@ -139,25 +174,33 @@ def delete_chat_session(
...
     if hard_delete:
-        stmt_messages = delete(ChatMessage).where(
-            ChatMessage.chat_session_id == chat_session_id
-        )
-        db_session.execute(stmt_messages)
+        delete_files_from_chat_session(chat_session_id, db_session)
 
-        stmt = delete(ChatSession).where(ChatSession.id == chat_session_id)
-        db_session.execute(stmt)
+        db_session.execute(delete(ChatSession).where(ChatSession.id == chat_session_id))
```
Weves, line 196: does this owrk for chat sessions with feedback? My hunch is that it won't 😓

web/src/app/admin/settings/SettingsForm.tsx
```diff
@@ -152,6 +234,34 @@ export function SettingsForm() {
             ]);
         }}
       />
+      <Title className="mb-4">Chat Settings</Title>
```
Weves: can you upload a video of what it's like to use this? 🙇

Weves (changes requested):

backend/danswer/db/pg_file_store.py
hagen-danswer: no

hagen-danswer:
here is a demo setting the ttl value:
https://github.com/danswer-ai/danswer/assets/171370825/3c3a4308-48d4-44be-89c7-aa67a217026e

hagen-danswer:
Screen-Recording-2024-06-27-at-4
here is demo setting TTL duration^

backend/danswer/db/models.py
```diff
@@ -808,14 +814,18 @@ class ChatMessageFeedback(Base):
     __tablename__ = "chat_feedback"
 
     id: Mapped[int] = mapped_column(Integer, primary_key=True)
-    chat_message_id: Mapped[int] = mapped_column(ForeignKey("chat_message.id"))
+    chat_message_id: Mapped[int] = mapped_column(
```
Weves: should be `Mapped[int | None]`?

backend/danswer/db/models.py
```diff
@@ -787,7 +789,9 @@ class DocumentRetrievalFeedback(Base):
     __tablename__ = "document_retrieval_feedback"
 
     id: Mapped[int] = mapped_column(primary_key=True)
-    chat_message_id: Mapped[int] = mapped_column(ForeignKey("chat_message.id"))
+    chat_message_id: Mapped[int] = mapped_column(
```
Weves: should be `Mapped[int | None]`?

web/src/app/admin/settings/SettingsForm.tsx
Weves: I think this section should only be visible if EE is enabled

web/src/app/admin/settings/SettingsForm.tsx
```diff
@@ -152,6 +234,34 @@ export function SettingsForm() {
...
+          if (numValue >= 1) {
+            setChatRetention(numValue.toString());
+          } else if (e.target.value === "") {
+            setChatRetention("");
+          }
+        }}
+        id="chatRetentionInput"
+        placeholder="Infinite Retention"
+      />
+      <button
```
Weves: can we use the `Tremor` buttons we use elsewhere?

backend/alembic/versions/23957775e5f5_remove_feedback_foreignkey_constraint.py
```diff
@@ -0,0 +1,58 @@
...
+revision = "23957775e5f5"
+down_revision = "bc9771dccadf"
+branch_labels = None  # type: ignore
+depends_on = None  # type: ignore
+
+
+def upgrade():
+    # Use batch mode for databases like SQLite
+    with op.batch_alter_table("chat_feedback", schema=None) as batch_op:
+        batch_op.drop_constraint("chat_feedback__chat_message_fk", type_="foreignkey")
```
Weves: how come this is handled differently from `document_retrieval_feedback`?

hagen-danswer:
image
this is what it looks like now^

backend/alembic/versions/23957775e5f5_remove_feedback_foreignkey_constraint.py
```diff
@@ -0,0 +1,79 @@
...
+
+def downgrade():
+    with op.batch_alter_table("chat_feedback", schema=None) as batch_op:
+        batch_op.drop_constraint("chat_feedback__chat_message_fk", type_="foreignkey")
+        batch_op.create_foreign_key(
+            "chat_feedback__chat_message_fk",
+            "chat_message",
+            ["chat_message_id"],
+            ["id"],
+            ondelete="CASCADE",
```
Weves: remove to maintain consistency with before

Weves (approved):

PR #1718, opened 2024-06-26
Fixes DAN-189 (safari bug in admin). Removed td/absolute positioning behavior which is unde…

rkuo-danswer:
…fined. Implemented row click by extending the TableRow component.

### DIFF
--- a/web/src/app/admin/indexing/status/CCPairIndexingStatusTable.tsx
+++ b/web/src/app/admin/indexing/status/CCPairIndexingStatusTable.tsx
@@ -15,7 +15,7 @@ import { timeAgo } from "@/lib/time";
 import { ConnectorIndexingStatus } from "@/lib/types";
 import { ConnectorTitle } from "@/components/admin/connectors/ConnectorTitle";
 import { getDocsProcessedPerMinute } from "@/lib/indexAttempt";
-import Link from "next/link";
+import { useRouter } from "next/navigation";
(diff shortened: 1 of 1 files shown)


web/src/app/admin/indexing/status/CCPairIndexingStatusTable.tsx
```diff
@@ -68,6 +68,28 @@ function CCPairIndexingStatusDisplay({
...
+  ...props
+}: {
+  url: string;
+  children: React.ReactNode;
+  [key: string]: any; // This allows for any additional props
+}) {
+  const router = useRouter();
+
+  const navigate = () => {
+    router.push(url);
```
Weves, line 87: When using `Link` from `next/link`, NextJS prefetches the data necessary to render the linked to page, which makes the navigation to the new page super fast. I don't believe this approach will do that 😓 It's not too big of a deal, but if possible it would be great to replicate that. My hunch is that there's a way to tell NextJS to prefetch a given url, which we could put in the `useEffect` of this component

Weves (approved):
Overall looks great, thanks for making Danswer usable on Safari 😅 

Let one comment that would be good to look into (but don't spend too much time, it's not that big of a deal)

web/src/app/admin/indexing/status/CCPairIndexingStatusTable.tsx
rkuo-danswer: Added router.prefetch inside useEffect.

PR #1721, opened 2024-06-27
Configuring Assistants for Most Recent Documents

luke-danswer:
- Editing create assistants page with new text field, 'Use Most Recent Documents'
- Consuming new text field **num_days**, populating into Persona model
- Editing DB's for new fields
- Consuming changes through a BaseFilters in search tool
Screenshot 2024-06-26 at 5 46 36 PM
Screenshot 2024-06-26 at 5 47 25 PM


backend/apicache/12fc95a3f40a1ed3a4bfb136a694f536d105b881759ae7ab2b67bf28a031fd2d
yuhongsun96, line 1: Is this needed for anything? I imagine not right?

backend/danswer/chat/load_yamls.py
```diff
@@ -85,6 +85,10 @@ def load_personas_from_yaml(
                 num_chunks=persona.get("num_chunks")
                 if persona.get("num_chunks") is not None
                 else default_chunks,
+                use_recent_documents=persona["use_recent_documents"]
```
yuhongsun96: This and above can be both cleaner

persona.get("user_recent_documents") or False or something like that

backend/alembic/versions/6dffd0cbb64f_add_persona_filtering_columns.py
```diff
@@ -0,0 +1,30 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "6dffd0cbb64f"
+down_revision = "bc9771dccadf"
+branch_labels: None = None
+depends_on: None = None
+
+
+def upgrade() -> None:
+    op.add_column(
```
yuhongsun96: I feel like this should just be combined into one field, if None then there's no default filter.

yuhongsun96: maybe something like "default_filter_days"

backend/alembic/versions/6dffd0cbb64f_add_persona_filtering_columns.py
```diff
@@ -0,0 +1,30 @@
...
+        "persona", sa.Column("use_recent_documents", sa.Boolean(), nullable=True)
+    )
+    op.execute("UPDATE persona SET use_recent_documents = false")
+    op.alter_column("persona", "use_recent_documents", nullable=False)
+    op.add_column("persona", sa.Column("num_days", sa.Integer(), nullable=True))
```
yuhongsun96: isn't this a float?

backend/danswer/llm/answering/answer.py
```diff
@@ -248,7 +248,20 @@ def _raw_output_for_explicit_tool_calling_llms(
             )
 
             if tool.name() == SearchTool.NAME:
-                self._update_prompt_builder_for_search_tool(prompt_builder, [])
+                final_context_documents = None
```
yuhongsun96: Why is this section needed?

yuhongsun96: If this is for a bug that was in recently, it was addressed in another PR already haha
my fault on the bug 😓

backend/package-lock.json
yuhongsun96, line 1: I think also not needed yes?

backend/danswer/search/models.py
```diff
@@ -32,6 +34,37 @@ class BaseFilters(BaseModel):
...
+        cls,
+        retrieval_filter: "BaseFilters | None",
+        persona_filter: "BaseFilters | None",
+    ) -> "BaseFilters | None":
+        if retrieval_filter is None and persona_filter is None:
+            return None
+        elif retrieval_filter is None:
+            return persona_filter
+        elif persona_filter is None:
+            return retrieval_filter
```
yuhongsun96: doesn't this run the risk of overwriting other filters passed in?

backend/requirements/default.txt
```diff
@@ -12,7 +12,7 @@ distributed==2023.8.1
 fastapi==0.109.2
 fastapi-users==12.1.3
 fastapi-users-db-sqlalchemy==5.0.0
-filelock==3.12.0
+filelock==3.12.2
```
yuhongsun96: I think we don't need to bump reqs right?

yuhongsun96 (commented):
Left some of my thoughts, looking forward to getting yours

backend/danswer/llm/answering/answer.py
luke-danswer: great, yes this sounds like a result of the bug you mentioned.  Was not sending documents with explicit response.  Can retest with latest

backend/package-lock.json
luke-danswer: not needed, should be in a git-ignore

backend/requirements/default.txt
luke-danswer: i'll try downgrading and retesting, maybe an issue with my local setup, but was not able to get it working originally with regular req's

backend/danswer/search/models.py
luke-danswer: yes, it may have a pass by reference/value issue.  Will search for the right way to address this.

backend/apicache/12fc95a3f40a1ed3a4bfb136a694f536d105b881759ae7ab2b67bf28a031fd2d
luke-danswer: i don't believe it's needed.  I'll add this all to a git-ignore when revising pr

backend/alembic/versions/6dffd0cbb64f_add_persona_filtering_columns.py
luke-danswer: that is certainly an option, it's easier to render the UI with this schema due to the way persona's are fetched and fed into the assistant editor page.   However, with some workarounds we would be able to combine this column into one.

backend/requirements/default.txt
luke-danswer: this one is an issue with using Node 20 instead of Node 18, so packages will need to be upgraded at that point.   Can be downgraded in the meantime.

backend/alembic/versions/6dffd0cbb64f_add_persona_filtering_columns.py
yuhongsun96: I'm not sure I follow, why can't the frontend check for or use the None value? Why does it need the 2 fields separate?

backend/danswer/chat/load_yamls.py
```diff
@@ -82,9 +82,9 @@ def load_personas_from_yaml(
                 persona_id=(-1 * p_id) if p_id is not None else None,
                 name=persona["name"],
                 description=persona["description"],
-                num_chunks=persona.get("num_chunks")
-                if persona.get("num_chunks") is not None
-                else default_chunks,
+                num_chunks=persona.get("num_chunks", default_chunks),
```
yuhongsun96: if persona.get("num_chunks") is None, it will just be None, it won't use default_chunks
better to use the "or" I think

backend/requirements/model_server.txt
yuhongsun96, line 1: Let's not bump these either

backend/alembic/versions/6dffd0cbb64f_add_persona_filtering_columns.py
Weves: Also pref using 1 field here 🧐

.gitignore
```diff
@@ -3,6 +3,8 @@
 .venv
 .mypy_cache
 .idea
+/backend/apicache/
+/backend/package-lock.json
```
Weves: hmm, why is this one needed? I believe the package-lock.json should only be in the `web` dir

backend/package-lock.json
Weves, line 1: this should probably be removed 😅

web/package-lock.json
Weves, line 1: This probably shouldn't be changed? I don't see any new dependencies added. We can probably just checkout the old version of this file to remove this from the PR

web/src/app/admin/assistants/lib.ts
```diff
@@ -7,6 +7,8 @@ interface PersonaCreationRequest {
   task_prompt: string;
   document_set_ids: number[];
   num_chunks: number | null;
+  use_recent_documents: boolean;
+  num_days: number | null;
```
Weves, line 10: nit: if possible, should be consistent across the board as opposed to `num_days: number | null` in some places and `num_days?: number` in others

Weves (commented):
Overall looks good!

backend/alembic/versions/6dffd0cbb64f_add_persona_filtering_columns.py
luke-danswer: sure, can try to rework in the limited time remaining.   More of a holdover from non-relational thinking, but get why it's important to change for your db.

PR #1726, opened 2024-06-27
Remove unused Helm chart config

sd109:
These config options are no longer used in the Helm chart since https://github.com/danswer-ai/danswer/pull/1552 so removing them will reduce confusion for end users (e.g. I just tried to modify the Vespa volume size on my own deployment using these values and then discovered that they have no effect).

### DIFF
--- a/deployment/helm/values.yaml
+++ b/deployment/helm/values.yaml
@@ -329,32 +329,6 @@ vespa:
   tolerations: []
   affinity: {}
 
-
-#ingress:
-#  enabled: false
-#  className: ""
-#  annotations: {}
-#    # kubernetes.io/ingress.class: nginx
-#    # kubernetes.io/tls-acme: "true"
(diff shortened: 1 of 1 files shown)


sd109:
In case it's useful for anyone else, the correct way to adjust the Vespa volume size in the current chart version is to pass the config through to the [correct place](https://github.com/unoplat/vespa-helm-charts/blob/e7db599f1272d3ff4752298dba1db14d05b70409/charts/vespa/values.yaml#L127-L135) in the underlying Vespa chart with
```
vespa:
  volumeClaimTemplates:
    - metadata:
        name: vespa-storage
      spec:
        accessModes:
          - ReadWriteOnce
        resources:
          requests:
            storage: <chosen-volume-size>Gi
```

yuhongsun96:
No longer applicable

PR #1730, opened 2024-06-27
add modern health check banner + expiration tracking

pablodanswer:
re-implemented from EE


deployment/docker_compose/docker-compose.dev.yml
```diff
@@ -231,6 +231,9 @@ services:
 
       # Enterprise Edition only
       - ENABLE_PAID_ENTERPRISE_EDITION_FEATURES=${ENABLE_PAID_ENTERPRISE_EDITION_FEATURES:-false}
+    tty: true
```
Weves: what do these do? can we add a comment here?

backend/danswer/server/manage/users.py
```diff
@@ -246,9 +250,33 @@ async def get_user_role(user: User = Depends(current_user)) -> UserRoleResponse:
...
+            return None, None
+
+    except Exception as e:
+        logger.error(f"Error fetching AccessToken: {e}")
+        return None, None
+
+
 @router.get("/me")
-def verify_user_logged_in(
+async def verify_user_logged_in(
```
Weves: hmm, why does this need to be async now?

backend/danswer/server/manage/users.py
```diff
@@ -246,9 +250,33 @@ async def get_user_role(user: User = Depends(current_user)) -> UserRoleResponse:
     return UserRoleResponse(role=user.role)
 
 
+def get_current_token_info(user: User | None, db_session: Session):
+    if not isinstance(user, User):
```
Weves: prefer `if user is None`

web/src/app/chat/ChatPage.tsx
```diff
@@ -1072,6 +1072,7 @@ export function ChatPage({
   const [editingRetrievalEnabled, setEditingRetrievalEnabled] = useState(false);
   const sidebarElementRef = useRef<HTMLDivElement>(null);
   const innerSidebarElementRef = useRef<HTMLDivElement>(null);
+  // const secondsUntilExpiration = getSecondsUntilExpiration(user);
```
Weves: nit: remove

web/src/app/chat/ChatPage.tsx
```diff
@@ -1088,6 +1089,9 @@ export function ChatPage({
     <>
       <HealthCheckBanner />
       <InstantSSRAutoRefresh />
+      {/* <div className="m-3">
```
Weves: nit: remove

web/src/lib/types.ts
```diff
@@ -19,6 +19,9 @@ export interface User {
   role: "basic" | "admin";
   preferences: UserPreferences;
   status: UserStatus;
+  current_token_created_at: Date;
```
Weves: all these can be `null`, right?

backend/danswer/db/models.py
```diff
@@ -138,7 +143,7 @@ class User(SQLAlchemyBaseUserTableUUID, Base):
 
 
 class AccessToken(SQLAlchemyBaseAccessTokenTableUUID, Base):
-    pass
+    expiry_length: Mapped[int] = mapped_column(Integer, nullable=False)
```
Weves: do we need to add this? Can we instead compute it on the fly in the `/me` endpoint?

Weves (changes requested):
Overall, looks great!

Small thing, I'm seeing this:

Screenshot 2024-07-10 at 6 37 11 PM

backend/danswer/db/models.py
pablonyx: Certainly can treat the expiry length time in seconds (environment variable) as a constant and use that instead

pablodanswer:
Should be set now!

PR #1731, opened 2024-06-27
add workflows that automate docker builds against merge group events

rkuo-danswer:
This adds GitHub workflows that will trigger docker builds (no pushes) on merge group events (which fire off when we enable merge queues for PR's).  The workflows are slimmed down versions of the build and push workflows that run on tag events.

So far, just doing backend and web, model server might be next.

This is following a pattern in Github's docs about this feature which checks against both pull requests on submission as well as when being integrated by the merge queue into main. Thoughts on if we want to limit this to the merge queue only?  Some googling suggest this is possible with a bit of magic syntax.

https://stackoverflow.com/questions/76655935/when-does-a-github-workflow-trigger-for-merge-group-and-is-it-restricted-by-bran

### DIFF
--- a/.github/workflows/docker-build-backend-container-on-merge-group.yml
+++ b/.github/workflows/docker-build-backend-container-on-merge-group.yml
@@ -11,7 +11,6 @@ env:
 
 jobs:
   build:
-    # for EE, run on special image-builders runners
     # TODO: make this a matrix build like the web containers
     runs-on: 
       group: amd64-image-builders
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-build-backend-container-on-merge-group.yml
```diff
@@ -0,0 +1,36 @@
...
+    branches: [ "main" ]
+  merge_group:
+    types: [checks_requested]
+
+env:
+  REGISTRY_IMAGE: danswer/danswer-backend
+
+jobs:
+  build:
+    # for EE, run on special image-builders runners
```
yuhongsun96: There's no longer a separate EE set of actions, let's just remove this comment here

yuhongsun96 (approved):
Looks good

PR #1742, opened 2024-06-28
added settings and persona seeding options

hagen-danswer:


backend/ee/danswer/server/seeding.py
```diff
@@ -43,11 +51,52 @@ def _seed_llms(
         return
 
     logger.info("Seeding LLMs")
-    seeded_providers = [
+    for llm_upsert_request in llm_upsert_requests:
```
Weves: @hagen-danswer why was this changed? This looks like it might break this 😓

Weves (changes requested):

Weves (approved):

PR #1747, opened 2024-06-30
Fix misc chat bugs

Weves:
Fixes:

- New chats causing a "refresh" which takes the user back to the top of the chat history
- "Loading bubbles" disappearing too early for search enabled
- When using any non-danswer assistant, it would show up as "Danswer" due to alternate assistant being cast to `0`,
- single "\`" chars were being treated as code blocks (triple ```) rather than the expected mild code highlighting (e.g. `<CODE>`)
- When switching between chats, it didn't scroll to the bottom immediately

### DIFF
--- a/web/src/app/chat/ChatPage.tsx
+++ b/web/src/app/chat/ChatPage.tsx
@@ -12,7 +12,6 @@ import {
   Message,
   RetrievalType,
   StreamingError,
-  ToolCallFinalResult,
   ToolCallMetadata,
 } from "./interfaces";
 import { ChatSidebar } from "./sessionSidebar/ChatSidebar";
@@ -35,7 +34,6 @@ import {
   removeMessage,
   sendMessage,
   setMessageAsLatest,
-  updateModelOverrideForChatSession,
   updateParentChildren,
   uploadFilesForChat,
   useScrollonStream,
@@ -61,11 +59,7 @@ import { AnswerPiecePacket, DanswerDocument } from "@/lib/search/interfaces";
 import { buildFilters } from "@/lib/search/utils";
 import { SettingsContext } from "@/components/settings/SettingsProvider";
 import Dropzone from "react-dropzone";
-import {
-  checkLLMSupportsImageInput,
-  getFinalLLM,
-  structureValue,
-} from "@/lib/llm/utils";
+import { checkLLMSupportsImageInput, getFinalLLM } from "@/lib/llm/utils";
 import { ChatInputBar } from "./input/ChatInputBar";
 import { ConfigurationModal } from "./modal/configuration/ConfigurationModal";
 import { useChatContext } from "@/components/context/ChatContext";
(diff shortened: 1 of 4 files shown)


web/src/app/chat/ChatPage.tsx
```diff
@@ -140,35 +134,22 @@ export function ChatPage({
...
-  const urlChatSessionId = useRef<number | null>();
   // this is triggered every time the user switches which chat
   // session they are using
   useEffect(() => {
-    if (
-      chatSessionId &&
-      !urlChatSessionId.current &&
-      llmOverrideManager.llmOverride
-    ) {
-      updateModelOverrideForChatSession(
```
Weves, line 153: I couldn't tell why this was needed, but @pablodanswer let me know if we should add this back in

web/src/app/chat/ChatPage.tsx
```diff
@@ -521,13 +531,9 @@ export function ChatPage({
     debounce,
   });
 
-  const [hasPerformedInitialScroll, setHasPerformedInitialScroll] =
-    useState(false);
-
-  // on new page
-  useEffect(() => {
```
Weves, line 528: note: moved this into the other "on switch" `useEffect` since we need to coordinate the timing of the scroll

web/src/app/chat/ChatPage.tsx
```diff
@@ -1221,12 +1230,14 @@ export function ChatPage({
...
+                          const messageMap = completeMessageDetail.messageMap;
+                          const messageReactComponentKey = `${i}-${completeMessageDetail.sessionId}`;
                           if (message.type === "user") {
                             const parentMessage = message.parentMessageId
-                              ? completeMessageMap.get(message.parentMessageId)
+                              ? messageMap.get(message.parentMessageId)
                               : null;
                             return (
-                              <div key={`${i}-${existingChatSessionId}`}>
+                              <div key={messageReactComponentKey}>
```
Weves, line 1240: refresh was caused by the keys changing, which is fixed now 🎉

web/src/app/chat/lib.tsx
```diff
@@ -385,7 +385,9 @@ export function processRawChatHistory(
       message: messageInfo.message,
       type: messageInfo.message_type as "user" | "assistant",
       files: messageInfo.files,
-      alternateAssistantID: Number(messageInfo.alternate_assistant_id),
+      alternateAssistantID: messageInfo.alternate_assistant_id
```
Weves, line 388: this was causing `alternateAssistantID` to be set to 0 when `messageInfo.alternate_assistant_id` was null 😓

web/src/app/chat/message/Messages.tsx
```diff
@@ -150,8 +150,7 @@ export const AIMessage = ({
   }
 
   const shouldShowLoader =
-    !toolCall ||
-    (toolCall.tool_name === SEARCH_TOOL_NAME && query === undefined);
+    !toolCall || (toolCall.tool_name === SEARCH_TOOL_NAME && !content);
```
Weves, line 153: fix for missing loading bubbles here

PR #1752, opened 2024-06-30
Multiple slack apps/bots

rkuo-danswer:
This supports multiple slack app / bots.

DB schema and backend was migrated to support multiple slack apps.
Frontend has been reworked to create and edit multiple apps/bots.
Configs now belong to an app.
Code support singleton tokens was removed.
Listener.py is multithreaded now and will activate/deactivate bots on the polling interval.


backend/alembic/versions/50fce29095ca_add_multiple_slack_apps.py
```diff
@@ -0,0 +1,118 @@
...
+        if not bot_token:
+            logger.info("bot_token not found")
+            raise ValueError("bot_token not found")
+
+        app_token = tokens.get("app_token")
+        if not app_token:
+            logger.info("app_token not found")
+            raise ValueError("app_token not found")
+
+        #         logger.info(f"{revision}: bot_token={bot_token} app_token={app_token}")
```
yuhongsun96: nit: unused

backend/alembic/versions/50fce29095ca_add_multiple_slack_apps.py
```diff
@@ -0,0 +1,118 @@
...
+branch_labels: None = None
+depends_on: None = None
+
+# Configure logging
+logger = logging.getLogger("alembic.runtime.migration")
+logger.setLevel(logging.INFO)
+
+
+def upgrade() -> None:
+    logger.info(f"{revision}: Starting upgrade")
```
yuhongsun96: This stuff is logged by alembic already, not particularly needed

backend/alembic/versions/50fce29095ca_add_multiple_slack_apps.py
```diff
@@ -0,0 +1,118 @@
...
+        # Delete the tokens in dynamic config
+        logger.info(f"{revision}: Removing old bot and app tokens.")
+        get_dynamic_config_store().delete(_SLACK_BOT_TOKENS_CONFIG_KEY)
+
+    except Exception as ex:
+        # Ignore if the dynamic config is not found
+        logger.info(f"{revision}: Exception: {ex}")
+        logger.info(f"{revision}: This is OK if there was not an existing Slack bot.")
+
+    # op.alter_column('slack_bot_config', 'app_id', existing_type=sa.Integer(), nullable=False)
```
yuhongsun96: nit: unused

backend/alembic/versions/50fce29095ca_add_multiple_slack_apps.py
```diff
@@ -0,0 +1,118 @@
...
+        sa.PrimaryKeyConstraint("id"),
+    )
+
+    logger.info(f"{revision}: add_column: slack_bot_config.app_id")
+    op.add_column(
+        "slack_bot_config",
+        sa.Column("app_id", sa.Integer(), nullable=True),
+    )
+
+    try:
```
yuhongsun96, line 50: preference to not use try for more scope than is needed, if it fails in the migration unrelated to the tokens being fetched, it should fail loudly and be rolled back instead of user losing tokens. Not big deal here but generally speaking.

backend/danswer/danswerbot/slack/config.py
```diff
@@ -26,13 +28,16 @@ def get_slack_bot_config_for_channel(
...
 def validate_channel_names(
+    app_id: int,
     channel_names: list[str],
     current_slack_bot_config_id: int | None,
     db_session: Session,
 ) -> list[str]:
     """Make sure that these channel_names don't exist in other slack bot configs.
     Returns a list of cleaned up channel names (e.g. '#' removed if present)"""
-    slack_bot_configs = fetch_slack_bot_configs(db_session=db_session)
+    slack_bot_configs = fetch_slack_bot_configs(
```
yuhongsun96, line 38: Should we disallow here having other Slack Apps also listening in the same channels?

backend/danswer/danswerbot/slack/listener.py
```diff
@@ -478,17 +505,125 @@ def _initialize_socket_client(socket_client: SocketModeClient) -> None:
...
+
+
+def main_worker(active_threads: SortedDict) -> None:
+    try:
+        with get_session_context_manager() as db_session:
+            apps = fetch_slack_apps(db_session=db_session)
+        if not apps:
+            return
+    except Exception as ex:
+        logger.info(f"Exception: {ex}")
```
yuhongsun96: There are different logging levels, might be good to use the right one throughout. For example, maybe it should be logger.exception here

backend/danswer/danswerbot/slack/listener.py
```diff
@@ -478,17 +505,125 @@ def _initialize_socket_client(socket_client: SocketModeClient) -> None:
...
+        return
+
+    m_apps = {}
```
yuhongsun96: prefer more explicit variable naming for ease of reading

backend/danswer/danswerbot/slack/listener.py
```diff
@@ -478,17 +505,125 @@ def _initialize_socket_client(socket_client: SocketModeClient) -> None:
...
+
+        active_threads[a_id] = tm
+
+    # Print final state if any state changed
+    if len(apps_to_deactivate) > 0 or len(apps_to_activate) > 0:
+        logger.info(f"Active bot count: {len(active_threads)}")
+        for k, v in active_threads.items():
+            logger.info(f"  ID: {k} Name: {v['name']}")
+
+    return
```
yuhongsun96: not needed

backend/danswer/db/models.py
```diff
@@ -1108,6 +1109,20 @@ class SlackBotConfig(Base):
...
 
+class SlackApp(Base):
+    __tablename__ = "slack_app"
+
+    id: Mapped[int] = mapped_column(primary_key=True)
+    name: Mapped[str] = mapped_column(String)
+    description: Mapped[str] = mapped_column(String)
+    enabled: Mapped[bool] = mapped_column(Boolean, default=True)
+
+    bot_token: Mapped[str] = mapped_column(String)
```
yuhongsun96: should be the encryptable type

backend/danswer/db/models.py
```diff
@@ -1108,6 +1109,20 @@ class SlackBotConfig(Base):
     persona: Mapped[Persona | None] = relationship("Persona")
 
 
+class SlackApp(Base):
```
yuhongsun96, line 1112: Do we want any uniqueness constraints here? Like if the bot and app tokens are the same across two apps, then it should not be allowed?

yuhongsun96 (approved):
Code looks great! A couple tiny nit things:
- On style/cleanliness: try not to leave commented out code that isn't actually useful for the reader and use more verbose variable naming
- On logging: to use the correct logging levels
- On design: try to cover edge cases such as user creating multiple copies of the same app, better to let things fail loudly so we catch mistakes in implementation (in this case by not wrapping more code in the "try" than is needed)

Note, I haven't tested it, leaving that to Chris and also I don't review frontend code so no feedback there

backend/danswer/danswerbot/slack/config.py
rkuo-danswer: I believe we should allow it. One good use case could be having multiple bots hanging out in a channel.  One responds to general chat but the others respond when directly tagged.

josh-at-straker:
This is such an amazing feature! Is there much left to have this implemented?

PR #1755, opened 2024-07-01
Notion date filter fix

danielnaber:
I think the date filter is wrong, it scans too many files as `compare_time <= end` is always true if you're just trying to get the most recent documents? This causes the connector to run for a long time on each run if your Notion workspace is large.


yuhongsun96 (approved):
Nice catch, that's entirely correct!

Thanks for the contribution!

PR #1758, opened 2024-07-02
allowed arbitrary types to handle the sqlalchemy datatype

hagen-danswer:


hagen-danswer:
Example of it working (and properly seeding an example default LLM):

Screen-Recording-2024-07-02-at-1

Weves (approved):

PR #1760, opened 2024-07-03
Fix citations + unit tests

pablodanswer:
Test cases in the testing file!


backend/scripts/test_citation_builder.py
```diff
@@ -0,0 +1,281 @@
...
+    while True:
+        unique_test_case = SimpleCitationExtractionTestCase(
+            test_name="Repeated citations",
+            input_text="""[1]Aasfasdasff [1] .""",
+            expected_text="""[[0]](https://0.com)Aasfasdasff [[0]](https://0.com).""",
+            expected_citations=["doc_0"],
+        ) 
+
+        if unique_test_case.run_test((mock_docs,mock_doc_mapping)):
+            print("SUCESS SO FAR")
```
Weves: nit: typo

backend/scripts/test_citation_builder.py
```diff
@@ -0,0 +1,281 @@
...
+
+import unittest
+import random
+from datetime import datetime
+from typing import List
+
+
+
+
+class TestCitationExtraction(unittest.TestCase):
```
Weves: Woohoo 🎉

backend/danswer/llm/answering/stream_processing/citation_processing.py
```diff
@@ -26,91 +26,161 @@ def extract_citations_from_stream(
...
-        # Special case of [1][ where ][ is a single token
-        # This is where the model attempts to do consecutive citations like [1][2]
-        if prepend_bracket:
-            curr_segment += "[" + curr_segment
-            prepend_bracket = False
-
         curr_segment += token
         llm_out += token
 
-        possible_citation_pattern = r"(\[\d*$)"  # [1, [, etc
```
Weves, line 60: if we're no longer looking for these possible citation patterns, we could now output partial citations? E.g.
if the stream is like `["[1", "]."]`, then the `[1` will flash in the UI before being converted into a citation. This is probably even more important once we switch over to the new chat UI with the special display for citations

backend/danswer/llm/answering/stream_processing/citation_processing.py
```diff
@@ -26,91 +26,161 @@ def extract_citations_from_stream(
...
-                        document_id=context_llm_doc.document_id,
+        citation_pattern = r"\[(\d+)\]"
+        print(f"FILTERING FOR CITATIONS {curr_segment}")
+        citations_found = list(re.finditer(citation_pattern, curr_segment))
+
+        if len(citations_found) == 0 and len(llm_out) - past_cite_count > 5:
+            current_citations = []
+
+        if citations_found and not in_code_block(llm_out):
+            # print(f"BEFORE: curr_segment: |{curr_segment}|")
```
Weves: nit: you probably already know this, but can we remove all the prints before we merge?

backend/danswer/llm/answering/stream_processing/citation_processing.py
```diff
@@ -26,91 +26,161 @@ def extract_citations_from_stream(
...
+    - Yields DanswerAnswerPiece objects for regular text.
+    - Yields CitationInfo objects for each unique citation encountered.
+
+    7. Context Awareness:
+    - Uses context_docs to access document information for citations.
+
+    This function effectively processes a stream of text, identifies and reformats citations,
+    and provides both the processed text and citation information as output.
+    """
+
```
Weves: when testing this locally in the chat UI, the text doesn't seem to stream as expected (nothing displays until we hit a citation) 🧐

Weves (commented):
The tests are awesome 🙇 Left a couple comments

backend/danswer/llm/answering/stream_processing/citation_processing.py
pablonyx: Haven't seen this behavior- might have to do with most recent update and `curr_segment`s- will investigate!

backend/danswer/llm/answering/stream_processing/citation_processing.py
```diff
@@ -26,91 +26,161 @@ def extract_citations_from_stream(
...
-                    rf"\[{numerical_value}\]", f"[{target_citation_num}]", curr_segment
-                )
-
-                if target_citation_num not in cited_inds:
-                    cited_inds.add(target_citation_num)
-                    yield CitationInfo(
-                        citation_num=target_citation_num,
-                        document_id=context_llm_doc.document_id,
+        citation_pattern = r"\[(\d+)\]"
+        print(f"FILTERING FOR CITATIONS {curr_segment}")
```
yuhongsun96: Here as well, I think we should just not log/print anything for this area of functionality

backend/danswer/llm/answering/stream_processing/citation_processing.py
```diff
@@ -26,91 +26,161 @@ def extract_citations_from_stream(
...
                 continue
-
             token = next_hold
             hold = ""
         else:
             token = raw_token
 
-        # Special case of [1][ where ][ is a single token
-        # This is where the model attempts to do consecutive citations like [1][2]
-        if prepend_bracket:
```
yuhongsun96, line 53: How is this case being handled now? Sorry if I'm slow brained, I just don't want to think about it so please explain <3

backend/danswer/llm/answering/stream_processing/citation_processing.py
```diff
@@ -26,91 +26,161 @@ def extract_citations_from_stream(
...
+
+        if len(citations_found) == 0 and len(llm_out) - past_cite_count > 5:
+            current_citations = []
+
+        if citations_found and not in_code_block(llm_out):
+            # print(f"BEFORE: curr_segment: |{curr_segment}|")
+            last_citation_end = 0
+            length_to_add = 0
+
+            print("CITATIONS")
```
yuhongsun96: Just general advice, since you're using Cursor and you have all of the deployment stuff set up, you should use breakpoints in the debugger instead of adding prints to the code

backend/scripts/test_citation_builder.py
yuhongsun96, line 1: Let's place this in the unit tests directory. Chris prefers pytest so let's port the existing ones all to pytest.

We don't need to have prints for tests, just have the standard asserts using the library. We have some actions for this automatically running as well so let's get it ported to pytest and update the Github actions as well:

image

backend/scripts/test_citation_builder.py
```diff
@@ -0,0 +1,281 @@
...
+from danswer.chat.models import CitationInfo
+from danswer.chat.models import DanswerAnswerPiece
+from danswer.chat.models import LlmDoc
+
+# pylint: enable=E402
+# flake8: noqa: E402
+import random
+
+
+def create_mock_tokens(text: str) -> list[str]:
```
yuhongsun96: lists are iterable so you can just mock specific test strings by passing lists of "tokens"

backend/scripts/test_citation_builder.py
```diff
@@ -0,0 +1,281 @@
...
+    tokens = []
+    i = 0
+    while i < len(text):
+        # Randomly choose to split after 2, 3, or 4 characters
+        split_length = random.choice([2, 3,4, 5, 6])
```
yuhongsun96: tests should not have any randomness to them, it's quite bad to have tests be nondeterministic

backend/scripts/test_citation_builder.py
```diff
@@ -0,0 +1,281 @@
...
+        end = min(i + split_length, len(text))
+        tokens.append(text[i:end])
+        i = end
+    return tokens
+
```
yuhongsun96: let's also keep the formatting enforced for the tests as well? I think unless there's some really good reason, the enforcement of best practices/styles across all modules

backend/scripts/test_citation_builder.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+def create_mock_doc_id_to_rank_map(num_docs: int) -> dict[str, int]:
```
yuhongsun96: unused, let's not include any dead code

backend/scripts/test_citation_builder.py
```diff
@@ -0,0 +1,281 @@
+import argparse
+
+# import unittest
+import os
+import sys
+from datetime import datetime
+from typing import List
+from typing import Union
+import pytest
```
yuhongsun96: pytest is not built in, should be added to dev requirements file

backend/scripts/test_citation_builder.py
```diff
@@ -0,0 +1,281 @@
...
+            input_text=""""Test! [1] [5].""",
+            expected_text=""""Test! [[0]](https://0.com) [[2]](https://2.com).""",
+            expected_citations=["doc_0", "doc_2"],
+        )
+        self.run_test_case(test_case)
+
+    def test_repeated_citations(self):
+        test_case = SimpleCitationExtractionTestCase(
+            test_name="Repeated citations",
+            input_text="""Test! [1][1][1]. And some more [1][2].""",
```
yuhongsun96: across the board, we should test by specifying tokens explicitly by passing in list of strings (tokens) rather than having the tokenizing function

backend/scripts/test_citation_builder.py
```diff
@@ -0,0 +1,281 @@
...
+        test_case_4 = SimpleCitationExtractionTestCase(
+            test_name="Hardest!",
+            input_text="""Multiple citations in one sentence [1][4][5]. """,
+            expected_text="""Multiple citations in one sentence [[0]](https://0.com)[[1]]()[[2]](https://2.com).""",
+            expected_citations=[("doc_0"), ("doc_1"), ("doc_2")],
+        )
+        test_case_4.run_test((mock_docs, mock_doc_mapping))
+
+
+if __name__ == '__main__':
```
yuhongsun96: Let's not run pytests using the __name__, pytests have a specific way of running certain classes etc. We should have the standard way and maybe just document it in a README

backend/scripts/test_citation_builder.py
```diff
@@ -0,0 +1,281 @@
...
+    citations = []
+    for piece in result:
+        if isinstance(piece, DanswerAnswerPiece):
+            final_answer_text += piece.answer_piece
+        elif isinstance(piece, CitationInfo):
+            citations.append(piece)
+    return final_answer_text, citations
+
+
+import argparse
```
yuhongsun96: Imports can be top of file but we don't want to have argparse for pytests generally

For typing List, let's just use the built in list

pablonyx: All great feedback! Sorry for leaving the unused logs / non-best practices in- was planning on cleaning this up before submitting as a PR (still using the non-deterministic tokenizing to thoroughly test some tricky edge cases). 

Completely agree with the comments and will make the modifications!

backend/tests/unit/test_citation_builder.py
```diff
@@ -0,0 +1,257 @@
...
+        ),
+        (
+            "Repeated citations with text",
+            ["[", "1", "]", "Aasf", "asda", "sff  ", "[", "1", "]", " ."],
+            "[[1]](https://0.com)Aasfasdasff  [[1]](https://0.com) .",
+            ["doc_0"],
+        ),
+        (
+            "Repeated citations with text",
+            ["[1][", "1", "]t", "[2]"],
```
yuhongsun96: Can we add a test here for a pattern like "[1][", "2", "]" where we do want to keep it?

backend/tests/unit/danswer/chat/test_chat_llm.py
```diff
@@ -1,38 +1,30 @@
...
-        links[3] = None
-        test_1 = "Something [2][4][5]"
-        res = "".join(list(extract_citations_from_stream(iter(test_1), links)))
-        self.assertEqual(res, "Something [[2]](link_2)[4][[5]](link_5)")
-        """
-
-
-if __name__ == "__main__":
-    unittest.main()
+def test_citation_extraction() -> None:
```
yuhongsun96: Can we port these test cases to the new one so it actually runs?

backend/throttle.ctrl
```diff
@@ -1 +1 @@
-f1f2 1 1718910083.03085 wikipedia:en
+f1f2 1 1718910083.03085 wikipedia:en
```
yuhongsun96, line 1: what is this?

backend/tests/unit/danswer/chat/test_chat_llm.py
yuhongsun96: or just delete it is fine

backend/tests/unit/danswer/llm/answering/stream_processing/test_citation_processing.py
```diff
@@ -0,0 +1,269 @@
...
+        ),
+        (
+            "Repeated citations with text",
+            ["[1][", "1", "]t", "[2]"],
+            "[[1]](https://0.com)t",
+            ["doc_0"],
+        ),
+        (
+            "Repeated citations with text",
+            ["[1][", "1", "]t]", "[2]"],
```
yuhongsun96, line 244: Just to be extra clear, I imagine it works fine but what if ][ is its own token, it may be different since
"[1][" as a token handles a complete citation, where as "[1" + "][" only becomes a complete citation after the second token. Anyway, I'm sure it's fine

backend/danswer/llm/answering/stream_processing/citation_processing.py
```diff
@@ -26,91 +26,152 @@ def extract_citations_from_stream(
...
-                curr_segment = re.sub(
-                    rf"\[{numerical_value}\]", f"[{target_citation_num}]", curr_segment
-                )
-
-                if target_citation_num not in cited_inds:
-                    cited_inds.add(target_citation_num)
-                    yield CitationInfo(
-                        citation_num=target_citation_num,
-                        document_id=context_llm_doc.document_id,
+        if len(citations_found) == 0 and len(llm_out) - past_cite_count > 5:
```
yuhongsun96, line 97: can we add two comments:
- what is past_cite_count
- why is there a 5 hardcoded here

yuhongsun96 (approved):
Looks good, final few questions

Issue #1761, opened 2024-07-03
Notion Connector - Block type `external_object_instance_page` is not supported via the API.


colachg:
This is related to issue #1053 (fixed in #1059) and #1230 (fixed in #1231).

![CleanShot 2024-07-03 at 09 48 54@2x](https://github.com/danswer-ai/danswer/assets/6825116/82c08320-cbe7-4831-b644-13f173b3fe40)

The connector consistently fails after indexing 7,552 documents, which is extremely frustrating.

![CleanShot 2024-07-03 at 09 51 17@2x](https://github.com/danswer-ai/danswer/assets/6825116/10044a5a-8298-45c4-9aea-2b6e6678e36c)

I'm not sure, this is a good way to fix.

```
                if result_type == "external_object_instance_page":
                    logger.warning(f"Skipping 'external_object_instance_page' ('{result_block_id}') for page '{page_block_id}': "
                                   f"Notion API does not currently support reading external blocks (as of 24/07/03) "
                                   f"(discussion: https://github.com/danswer-ai/danswer/issues/1761)")
                    continue
```
I believe people may encounter various types of block errors. Is there a universal method to catch all these errors? Currently investigating.

PR #1762, opened 2024-07-03
Notion Connector to skip reading external blocks in NotionConnector

colachg:
The commit skips reading 'external_object_instance_page' blocks in the NotionConnector due to the lack of support in the Notion API. This change is in response to the issue #1761.


yuhongsun96:
Nice thanks!

PR #1767, opened 2024-07-03
sort by date (like we do for MRs), so that the yield followed by return further down in the code makes sense

danielnaber:
Without this change, the issues come back in basically random order and we stop collecting as soon as the first one happens to be before our start date -> issues can get lost


danielnaber:
I just see that using a `return` there doesn't make sense anyway, does it? It would skip indexing the issues if it's reached when iterating over the MRs?

danielnaber:
Closing, will try to create a PR that fixes this issue and more.

PR #1774, opened 2024-07-03
Added search quality testing pipeline

hagen-danswer:


backend/danswer/configs/app_configs.py
```diff
@@ -106,8 +106,18 @@
 # NOTE: this is used if and only if the vespa config server is accessible via a
 # different host than the main vespa application
 VESPA_CONFIG_SERVER_HOST = os.environ.get("VESPA_CONFIG_SERVER_HOST") or VESPA_HOST
-VESPA_PORT = os.environ.get("VESPA_PORT") or "8081"
-VESPA_TENANT_PORT = os.environ.get("VESPA_TENANT_PORT") or "19071"
+
+
+def get_vespa_port() -> str:
```
yuhongsun96: Prefer not touching the config file directly. It's kind of nice that all of this gets run on import rather than relying on function calls. Let's keep it this way.

Things for testing that don't need to go in the core code base shouldn't be added in as it just causes bloat. The downside of course is that tests need to be maintained when there are breaking changes in the main code but I think it's better to avoid the bloat

backend/danswer/db/connector_credential_pair.py
```diff
@@ -152,7 +152,7 @@ def add_credential_to_connector(
     credential_id: int,
     cc_pair_name: str | None,
     is_public: bool,
-    user: User,
+    user: User | None,
```
yuhongsun96, line 155: This should arouse suspicion, why is this ok now, how else is this used. If you trace it to the endpoint, you'll realize that there was also a typing issue at the API level so we verify and fix that as well.

Luckily it's only a typing issue this time but it could potentially be something much worse

backend/danswer/db/engine.py
```diff
@@ -63,6 +63,13 @@ def get_sqlalchemy_engine() -> Engine:
     return _SYNC_ENGINE
 
 
+def get_sqlalchemy_engine_for_port_number(port: str) -> Engine:
```
yuhongsun96: Same as in the app_configs, let's not introduce bloat

backend/danswer/document_index/vespa/index.py
```diff
@@ -108,7 +128,7 @@ class _VespaUpdateRequest:
     update_request: dict[str, dict]
 
 
-@retry(tries=3, delay=1, backoff=2)
+@retry(tries=5, delay=10, backoff=2)
```
yuhongsun96: Can't just change this logic like that, we don't want search to hang for a long time, if it fails it should just fail

backend/danswer/configs/app_configs.py
yuhongsun96: Also I don't understand why this would need to change during runtime

backend/tests/regression/answer_quality/cli_utils.py
yuhongsun96, line 1: Since there is now so much functionality here, the user/reader should not have to go through all the code to understand what is available or how to use it. You should include a README here with some sample usages (like how to use the args etc). Point out requirements to use this (if any) etc.

backend/tests/regression/answer_quality/cli_utils.py
```diff
@@ -0,0 +1,291 @@
...
+        sys.exit(1)
+
+
+def check_docker_running() -> int:
+    print("Checking if Docker is running...")
+    returncode, _, _ = _run_command("docker info")
+    return returncode == 0
+
+
+def start_docker() -> None:
```
yuhongsun96: This is a nice touch but to be honest it's fine to just check if docker is running and just raise an exception and exit if it's not

backend/tests/regression/answer_quality/cli_utils.py
```diff
@@ -0,0 +1,291 @@
...
+def set_env_variables(
+    remote_server_ip: str,
+    remote_server_port: str,
+    use_cloud_gpu: bool,
+    openai_api_key: str,
+) -> None:
+    seed_json = {
+        "llms": [
+            {
+                "name": "Netflix NCP",
```
yuhongsun96: No, don't do this lmao

backend/tests/regression/answer_quality/cli_utils.py
```diff
@@ -0,0 +1,291 @@
...
+        ],
+    }
+    env_vars: dict = {}
+    env_vars["ENV_SEED_CONFIGURATION"] = json.dumps(seed_json)
+    env_vars["ENABLE_PAID_ENTERPRISE_EDITION_FEATURES"] = "true"
+    if use_cloud_gpu:
+        env_vars["MODEL_SERVER_HOST"] = remote_server_ip
+        env_vars["MODEL_SERVER_PORT"] = remote_server_port
+
+    for env_var, directory in env_vars.items():
```
yuhongsun96: why is this called directory? Also no point in gathering it all in a dict, just assign each to os.environ as you hit it

backend/danswer/configs/app_configs.py
yuhongsun96: Ok, I think it's been solved, see our Slack conversation

backend/danswer/db/engine.py
yuhongsun96: It's not needed so don't do this but you could just add a default argument to the function above this for the port which defaults to POSTGRES_PORT

backend/tests/regression/answer_quality/launch_search_test.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
yuhongsun96: mmm not a big fan of this type of approach, we can just have a set of exports to use and it can just be a command we share internally. If someone else wants to run this eval, they can set up their own model server. But also nobody else can really do this since they would need a dataset from Relari and also do all the cleanup and etc... at that point I'm sure they would have reached out to us

backend/tests/regression/answer_quality/launch_search_test.sh
```diff
@@ -0,0 +1,7 @@
...
+
+export PYTHONPATH="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../../.." && pwd):$PYTHONPATH"
+export MODEL_SERVER_HOST="18.218.165.17" 
+export MODEL_SERVER_PORT="80" 
+
+python3.11 backend/tests/regression/answer_quality/search_quality_test.py
```
yuhongsun96: I'm sure you would have cleaned it up prior to the final PR but let's always be super detail oriented, newline at end of file!

backend/tests/regression/answer_quality/launch_search_test.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+export PYTHONPATH="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../../.." && pwd):$PYTHONPATH"
+export MODEL_SERVER_HOST="18.218.165.17"
```
yuhongsun96: not as bad as hardcoding API keys and as discussed, the worst outcome is a denial of service on our testing which is basically harmless but... for good practice we should not hardcode IPs

backend/tests/regression/answer_quality/direct_file_uploader.py
```diff
@@ -0,0 +1,131 @@
...
+            in_progress_attemps = get_inprogress_index_attempts(conn_id, db_session)
+            if not in_progress_attemps:
+                return
+            wait_seconds = 15
+            print(f"still_indexing, waiting {wait_seconds} seconds")
+            time.sleep(wait_seconds)
+
+
+if __name__ == "__main__":
+    file_location = "/Users/danswer/testdocuments/20240628_source_documents.zip"
```
yuhongsun96: For stuff like this, better to have a relative path from working directory or base of repo

backend/tests/regression/answer_quality/relari.py
```diff
@@ -1,20 +1,53 @@
...
 from danswer.search.models import OptionalSearchSetting
 from danswer.search.models import RetrievalDetails
+from tests.regression.answer_quality.cli_utils import get_current_commit_sha
+from tests.regression.answer_quality.cli_utils import get_server_host_port
+
+
+def _get_answer_for_question(
+    query: str, db_session: Session, run_suffix: str
+) -> OneShotQAResponse:
+    while True:
```
yuhongsun96: Should probably be behind a flag (ie this function should take a bool option) whether or not to assume that there can't be any ongoing indexing. But this is really nit and unimportant

backend/tests/regression/answer_quality/search_test_config.yaml
```diff
@@ -0,0 +1,11 @@
+
```
yuhongsun96: Another good thing to put in the README is to update this file. The README should contain all of the info needed to run the tests without exploring the code.

backend/tests/regression/answer_quality/search_test_config.yaml
```diff
@@ -0,0 +1,11 @@
...
+branch: main
+model_server_ip: "18.218.165.17"
```
yuhongsun96: Same here, no hardcoded IPs

backend/tests/regression/answer_quality/search_test_config.yaml
```diff
@@ -0,0 +1,11 @@
...
+model_server_port: "80"
+zipped_documents_file: "/Users/danswer/testdocuments/samplepptx.pptx"
+questions_file: "/Users/danswer/danswer/backend/tests/regression/answer_quality/sample_questions.yaml"
```
yuhongsun96: I think local paths is best generally for this stuff, that way everything can be kept in the scope of the Danswer repo?

Under /home (~/) directory is often used as well, no strong preference

but let's not set a file that almost all people will not have like /Users/danswer

deployment/docker_compose/docker-compose.search-testing.yml
```diff
@@ -0,0 +1,406 @@
+version: '3'
```
yuhongsun96, line 1: I'm quite against having yet another docker compose yaml. Every time we add a new env based config we have to currently copy it to 3 docker compose yamls, kubernetes deployment, helm deployment. Let's see if we can just use the existing yamls

yuhongsun96 (commented):
Generally looks good, didn't scrutinize the scripts that much but seems reasonable.

By the way, there are a ton of errors in the automated checks. I know this is more of a [WIP] PR but let's definitely fix those up before the final PR

backend/danswer/db/connector_credential_pair.py
hagen-danswer: The intended behavior is if there is no user, the call defaults to admin access. The user is only consumed in the fetch_credential_by_id which also accepts a None so it should be fine to pass that default behavior up

backend/danswer/db/engine.py
```diff
@@ -55,10 +55,10 @@ def build_connection_string(
     return f"postgresql+{db_api}://{user}:{password}@{host}:{port}/{db}"
 
 
-def get_sqlalchemy_engine() -> Engine:
+def get_sqlalchemy_engine(port: str = POSTGRES_PORT) -> Engine:
```
yuhongsun96: Is this still needed? if every other container can use standard ports postgres should be able to as well no?

backend/tests/regression/answer_quality/README.md
```diff
@@ -0,0 +1,59 @@
+# Search Quality Test Script
```
yuhongsun96, line 1: great work with the README makes it much easier to understand/use

backend/tests/regression/answer_quality/cli_utils.py
```diff
@@ -0,0 +1,232 @@
...
+    command += " --pull always"
+    command += " --force-recreate"
+    if not launch_web_ui:
+        command += " --scale web_server=0"
+        command += " --scale nginx=0"
+    if use_cloud_gpu:
+        command += " --scale indexing_model_server=0"
+        command += " --scale inference_model_server=0"
+
+    # command = "docker compose -f docker-compose.dev.yml -p danswer-stack up -d"
```
yuhongsun96: cleanup stuff

backend/tests/regression/answer_quality/cli_utils.py
```diff
@@ -0,0 +1,232 @@
...
+    if returncode == 0:
+        print(
+            f"Successfully deleted {len(volumes_to_delete)} volumes for project: {project_name}"
+        )
+    else:
+        print(f"Failed to delete volumes. Error: {stderr}")
+
+
+@retry(tries=5, delay=5, backoff=2)
+def get_server_host_port(container_name: str, suffix: str, client_port: str) -> str:
```
yuhongsun96: I would leave a comment here on how it works. That way the next person doesn't need to have the context that you/Chris/I have. It's not common knowledge on how Docker maps ports between the internal and host networks

backend/tests/regression/answer_quality/file_uploader.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def _create_connector(run_suffix: str, file_paths: list[str]) -> int:
+    url = api_url_builder(run_suffix, "/manage/admin/connector")
+    headers = {
+        "Content-Type": "application/json",
+    }
+
+    connector = ConnectorBase(
+        name="ZipConnector_2",
```
yuhongsun96: Would give it a more descriptive name here in case user needs to debug what's going on. Maybe "AutomatedTestSingleRunFileZipConnector"

backend/tests/regression/answer_quality/file_uploader.py
```diff
@@ -0,0 +1,125 @@
...
+
+    conn_id = _create_connector(run_suffix, file_paths)
+    cred_id = _create_credential(run_suffix)
+
+    _create_cc_pair(run_suffix, conn_id, cred_id)
+    _run_cc_once(run_suffix, conn_id, cred_id)
+
+
+if __name__ == "__main__":
+    file_location = "/Users/danswer/testdocuments/20240628_source_documents.zip"
```
yuhongsun96: prefer local paths and also please clean up the commented one <3

backend/tests/regression/answer_quality/search_test_config.yaml
```diff
@@ -0,0 +1,16 @@
+zipped_documents_file: "~/sampledocs.zip"
```
yuhongsun96, line 5: Would add comments to all of these so it's easy for the user
Also we can follow common style of wrapping values that need to be changed in <WHATEVER>

deployment/docker_compose/docker-compose.search-testing.yml
yuhongsun96, line 1: Is this still needed? If so, can we make it very minimal so we don't need to keep maintaining the list of env vars etc?

I imagine for the test it's basically all using defaults ?

backend/danswer/document_index/vespa/index.py
```diff
@@ -118,7 +118,11 @@ def _does_document_exist(
     Specifically in this case, document refers to a vespa document which is equivalent to a Danswer
     chunk. This checks for whether the chunk exists already in the index"""
     doc_url = f"{DOCUMENT_ID_ENDPOINT.format(index_name=index_name)}/{doc_chunk_id}"
-    doc_fetch_response = http_client.get(doc_url)
+    try:
```
yuhongsun96: This does nothing, let's remove it

deployment/docker_compose/docker-compose.search-testing.yml
```diff
@@ -0,0 +1,406 @@
...
+      uvicorn danswer.main:app --host 0.0.0.0 --port 8080"
+    depends_on:
+      - relational_db
+      - index
+      # - inference_model_server
+    restart: always
+    ports:
+      - "8080"
+    environment:
+      # Auth Settings
```
yuhongsun96, line 20: this can rely on an .env file explicitly. The goal is to not have to duplicate the env vars so that when we introduce a new one it has to be changed in 20 places

deployment/docker_compose/docker-compose.search-testing.yml
```diff
@@ -0,0 +1,406 @@
...
+      o: bind
+      device: ${DANSWER_POSTGRES_DATA_DIR:-./postgres_data}
+  vespa_volume:
+    driver: local
+    driver_opts:
+      type: none
+      o: bind
+      device: ${DANSWER_VESPA_DATA_DIR:-./vespa_data}
+  model_cache_huggingface:
+  #   driver: local
```
yuhongsun96, line 370: general cleanup here (also other commented things around the PR). Be sure to give the PR a read on GitHub before the final review <3 thanks!

deployment/docker_compose/docker-compose.search-testing.yml
```diff
@@ -0,0 +1,406 @@
...
+#     driver_opts:
+#       type: none
+#       o: bind
+#       device: ${DANSWER_VESPA_DATA_DIR:-./vespa_data}
+#   model_cache_huggingface:
+#     driver: local
+#     driver_opts:
+#       type: none
+#       o: bind
+#       device: ${DANSWER_MODEL_CACHE_DIR:-./model_cache}
```
yuhongsun96, line 406: NEWLINE REEEE

deployment/docker_compose/docker-compose.search-testing.yml
```diff
@@ -0,0 +1,406 @@
...
+      # Query Options
+      - DOC_TIME_DECAY=${DOC_TIME_DECAY:-}  # Recency Bias for search results, decay at 1 / (1 + DOC_TIME_DECAY * x years)
+      - HYBRID_ALPHA=${HYBRID_ALPHA:-}  # Hybrid Search Alpha (0 for entirely keyword, 1 for entirely vector)
+      - EDIT_KEYWORD_QUERY=${EDIT_KEYWORD_QUERY:-}
+      - MULTILINGUAL_QUERY_EXPANSION=${MULTILINGUAL_QUERY_EXPANSION:-}
+      - LANGUAGE_HINT=${LANGUAGE_HINT:-}
+      - LANGUAGE_CHAT_NAMING_HINT=${LANGUAGE_CHAT_NAMING_HINT:-}
+      - QA_PROMPT_OVERRIDE=${QA_PROMPT_OVERRIDE:-}
+      # Other Services
+      - POSTGRES_HOST=relational_db
```
yuhongsun96, line 150: Some of the env things do still need to stay as it points to a service in the docker compose deployment which is not default

yuhongsun96 (approved):
Super amazing work! Thanks!

PR #1786, opened 2024-07-08
fix gpt-4o image issue

pablodanswer:
Fixes image generation issue with custom providers.

We were comparing compatible image models and providers (e.g. `openai`) against user-named "providers" (aka their keys).


Weves (approved):
lgtm!

PR #1789, opened 2024-07-09
add new standard answer test endpoint

pablodanswer:
# Add test-standard-answer Endpoint

Docs: 

This PR adds a new endpoint to test standard answers for given messages and categories.

## Changes

- Added `/chat/test-standard-answer` POST endpoint
- Uses `oneoff_standard_answers` for processing

## Usage

- Ensure you enable enterprise features

Test the endpoint with something like:

```bash
curl -X POST 'http://localhost:8080/chat/test-standard-answer' \
-H 'Content-Type: application/json' \
-d '{
  "message": "wifi password",
  "slack_bot_categories": ["General", "generaedge"]
}'
```

## Expected Response

```json
{
  "standard_answers": [
    {
      "id": 1,
      "keyword": "wifi",
      "answer": "The wifi password is \"XYZ\"",
      "categories": [
        {
          "id": 0,
          "name": "General"
        }
      ]
    }
  ],
  "error_msg": null
}
```

If no standard answers are found, the response will be:

```json
{
  "standard_answers": [],
  "error_msg": null
}
```


backend/ee/danswer/server/query_and_chat/chat_backend.py
```diff
@@ -51,6 +56,24 @@ def remove_answer_citations(answer: str) -> str:
     return re.sub(pattern, "", answer)
 
 
+@router.post("/test-standard-answer")
```
yuhongsun96: It's a valid use case, doesn't have to be for test right? I think we can just call it get-standard-answer or something

Technically to be correct about it, I think this is better as a GET since it doesn't change the state of anything, but I don't particularly care. I know we have it wrong in other places as well.

backend/ee/danswer/server/query_and_chat/chat_backend.py
```diff
@@ -51,6 +56,24 @@ def remove_answer_citations(answer: str) -> str:
...
+    try:
+        standard_answers = oneoff_standard_answers(
+            message=request.message,
+            slack_bot_categories=request.slack_bot_categories,
+            db_session=db_session,
+        )
+        return TestStandardAnswerResponse(standard_answers=standard_answers)
+
+    except Exception as e:
+        return TestStandardAnswerResponse(error_msg=f"An error has occured: {e}")
```
yuhongsun96: How do you think about this type of return vs just raising an HTTP exception?

backend/danswer/danswerbot/slack/handlers/handle_standard_answers.py
```diff
@@ -18,7 +18,42 @@
...
+        for standard_answer in category.standard_answers
+    }
+
+    matching_standard_answers = find_matching_standard_answers(
+        query=message,
+        id_in=[answer.id for answer in configured_standard_answers],
+        db_session=db_session,
+    )
+
+    server_tandard_answers = [
```
yuhongsun96: typo

yuhongsun96 (approved):

PR #1803, opened 2024-07-11
[1802] adjust the code to support a different db schemas

stianrincon:
ticket https://github.com/danswer-ai/danswer/issues/1802


yuhongsun96:
We're taking a pass at building multitenancy so we're introducing this logic already. Teammate is putting in some changes to allow multi-schema.

But the cleanup things are good! Thanks!

PR #1807, opened 2024-07-11
Fix slack group responses

hagen-danswer:


backend/danswer/configs/chat_configs.py
```diff
@@ -28,6 +28,10 @@
 FAVOR_RECENT_DECAY_MULTIPLIER = 2.0
 # Currently this next one is not configurable via env
 DISABLE_LLM_QUERY_ANSWERABILITY = QA_PROMPT_OVERRIDE == "weak"
+# For the highest matching base size chunk, how many chunks above and below do we pull in by default
```
Weves, line 31: why is this here?

backend/danswer/danswerbot/slack/utils.py
```diff
@@ -318,57 +318,65 @@ def fetch_userids_from_emails(
...
-            else:
-                failed_to_find.append(group_name)
-        except Exception as e:
-            logger.error(f"Error fetching user IDs for group {group_name}: {str(e)}")
-            failed_to_find.append(group_name)
+    try:
+        response = client.usergroups_list()
+        if not isinstance(response.data, dict):
+            logger.error("Error fetching user groups")
+            return user_ids, failed_to_find
```
Weves, line 330: shouldn't `failed_to_find` be all of the `given_names`

backend/danswer/danswerbot/slack/utils.py
```diff
@@ -318,57 +318,65 @@ def fetch_userids_from_emails(
...
+        name_id_map = {d["name"]: d["id"] for d in all_group_data}
+        handle_id_map = {d["handle"]: d["id"] for d in all_group_data}
+        for given_name in given_names:
+            group_id = name_id_map.get(given_name) or handle_id_map.get(
+                given_name.lstrip("@")
+            )
+            if not group_id:
+                failed_to_find.append(given_name)
+                continue
+            response = client.usergroups_users_list(usergroup=group_id)
```
Weves, line 342: if this call fails, we'll have a weird state, no? E.g. the union of `user_ids` and `failed_to_find` wouldn't have all the `given_names`

backend/danswer/danswerbot/slack/utils.py
```diff
@@ -318,57 +318,65 @@ def fetch_userids_from_emails(
...
-                    group_ids.add(name_id_map[group])
-                elif group in handle_id_map:
-                    group_ids.add(handle_id_map[group])
-                else:
-                    failed_to_find.append(group)
-        else:
-            # Most likely a Slack App scope issue
+        if not isinstance(response.data, dict):
             logger.error("Error fetching user groups")
+            return group_data, failed_to_find
```
Weves, line 362: same comment here as above (I know it doesn't have this consistency currently, but I think it would be good to add)

Weves (changes requested):
Overall looks good! Left a few comments + a couple additional notes:

- It would be a good practice to try (as much as possible) to have each PR address one issue / encompass one feature. This makes things a lot easier to review. So in this case, I would have split the Slack Connector credential changes into a separate PR. 
- Especially for things that touch a ton of files but are addressing a small, specific issue, a one or two sentence description describing what the problem ended up being (and maybe even pinpointing the key parts of the solution via inline comments) would probably save the reviewer ~10-30 minutes when reviewing but only take a minute or two. IMO this is generally a good practice (and Yuhong and I should be doing it too)

backend/danswer/configs/chat_configs.py
hagen-danswer: git mistake

backend/danswer/danswerbot/slack/utils.py
Weves: LGTM 🚢

PR #1816, opened 2024-07-12
Embeddings

pablodanswer:
Relevant Docs PR: https://github.com/danswer-ai/danswer-docs/pull/99

Adds cloud embeddings for:
- Google's Vertex AI
- OpenAI
- Anthropic
- Voyage AI

## Tests run: 

For each provider:

After deleting previously used model (should delete credentials from database):
1. Configure incorrect credentials (for cloud-based versions)
2. Configure correct credentials (for cloud-based versions)
3. Allow re-indexing of existing content to occur
4. Attempt to delete model / credentials (shouldn't be possible while set as the current embedding type)
5. Add a new connector
6. Test search
7. Test chat retrieval

Providers tested
- [x] Vertex
- [x] OpenAI
- [x] Voyage
- [x] Anthropic
- [x] Locally hosted embeddings


backend/danswer/server/manage/llm/api.py
yuhongsun96, line 1: Embedding models are generally not considered LLMs. Better to not conflate the two.

Also just to keep the modules more single-functionality, I think it would be better (regardless of the above) to break out the embedding models into a different module with a different router/prefix

backend/danswer/search/search_nlp_models.py
```diff
@@ -84,20 +84,24 @@ def build_model_server_url(
 class EmbeddingModel:
     def __init__(
         self,
-        model_name: str,
-        query_prefix: str | None,
-        passage_prefix: str | None,
-        normalize: bool,
         server_host: str,  # Changes depending on indexing or inference
         server_port: int,
+        model_name: str | None = None,
```
yuhongsun96: Preference to have these not as defaults because it makes it easier to make mistakes in usage in other places. I think it's better to be more intentional with these. If a function has a bunch of parameters, the caller should specify them as named arguments for clarity as well

backend/model_server/encoders.py
```diff
@@ -25,6 +35,100 @@
...
+        else:
+            raise ValueError(f"Unsupported provider: {self.provider}")
+
+    def encode(self, texts: list[str], model_name: str | None) -> list[list[float]]:
+        return [self.embed(text, model_name) for text in texts]
+
+    def embed(self, text: str, model: str | None = None) -> list[float]:
+        # logger.debug(f"Embedding text with provider: {self.provider}")
+        if self.provider == EmbeddingProvider.OPENAI:
+            return self._embed_openai(text, model or "text-embedding-ada-002")
```
yuhongsun96: can default to text-embedding-3-small? It's newer, less expensive, better benchmarks

backend/model_server/encoders.py
```diff
@@ -25,6 +35,100 @@
...
+            )
+
+        else:
+            raise ValueError(f"Unsupported provider: {self.provider}")
+
+    def encode(self, texts: list[str], model_name: str | None) -> list[list[float]]:
+        return [self.embed(text, model_name) for text in texts]
+
+    def embed(self, text: str, model: str | None = None) -> list[float]:
+        # logger.debug(f"Embedding text with provider: {self.provider}")
```
yuhongsun96: Can just keep this one in there (uncomment), might be useful

backend/model_server/encoders.py
```diff
@@ -25,6 +35,100 @@
...
+        elif self.provider == EmbeddingProvider.VOYAGE:
+            return voyageai.Client(api_key=self.api_key)
+        elif self.provider == EmbeddingProvider.VERTEX:
+            credentials = service_account.Credentials.from_service_account_info(
+                json.loads(self.api_key)
+            )
+            project_id = json.loads(self.api_key)["project_id"]
+            vertexai.init(project=project_id, credentials=credentials)
+            return TextEmbeddingModel.from_pretrained(
+                self.model or "text-embedding-004"
```
yuhongsun96, line 69: the defaults are a bit scattered everywhere, I would prefer having it cleaned up. Either as default args or somewhere else in a centralized module imported here

backend/model_server/encoders.py
```diff
@@ -25,6 +35,100 @@
...
+            return self._embed_voyage(text, model)
+        elif self.provider == EmbeddingProvider.VERTEX:
+            return self._embed_vertex(text, model)
+        else:
+            raise ValueError(f"Unsupported provider: {self.provider}")
+
+    def _embed_openai(
+        self, text: str, model: str = "text-embedding-ada-002"
+    ) -> list[float]:
+        logger.debug(f"Using OpenAI embedding with model: {model}")
```
yuhongsun96: Would just log everything in the wrapper function and not have a bunch of these

backend/model_server/encoders.py
```diff
@@ -25,6 +35,100 @@
...
+        response = self.client.embeddings.create(input=text, model=model)
+        return response.data[0].embedding
+
+    def _embed_cohere(self, text: str, model: str | None) -> list[float]:
+        if model is None:
+            model = "embed-english-v3.0"
+
+        logger.debug(f"Using Cohere embedding with model: {model}")
+        response = self.client.embed(
+            texts=[text], model=model, input_type="search_document"
```
yuhongsun96: A lot of these need to have the input type and prefix correctly, ask me if you have questions

backend/model_server/encoders.py
```diff
@@ -25,6 +35,100 @@
...
+        )
+        return response.embeddings[0]
+
+    def _embed_voyage(self, text: str, model: str | None) -> list[float]:
+        if model is None:
+            model = "voyage-01"
```
yuhongsun96: for many of the models, I think we are not using the latest, we should use latest

backend/model_server/encoders.py
yuhongsun96: Correct approach would be to have the different models use their defaults, a prefix can be provided as an override but generally we would just use the flag for document vs query.

yuhongsun96 (commented):
Some additional things to test:
- Try indexing some docs on the old version (main), then upgrading and selecting one of the cloud embedding models. Ideally do this in a containerized way and keep the old volumes during the upgrade
- Building the docker images and making sure those work

backend/alembic/versions/44f856ae2a4a_add_cloud_embedding_model.py
```diff
@@ -0,0 +1,65 @@
...
+Revision ID: 44f856ae2a4a
+Revises: bc9771dccadf
+Create Date: 2024-06-28 20:01:05.927647
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
+revision = "44f856ae2a4a"
```
yuhongsun96, line 12: make sure the SHAs point to the latest

PR #1819, opened 2024-07-13
3x indexing speed by removing multiple tokenizer initializations

hagen-danswer:


yuhongsun96:
Need to look 1 level deeper, this isn't a root cause fix, this is more of a bandaid fix

PR #1820, opened 2024-07-13
fix discourse connector rate limiting + topic fetching

pablodanswer:
- Update topic polling
- Implement rate limiting based on https://meta.discourse.org/t/available-settings-for-global-rate-limits-and-throttling/78612


yuhongsun96 (approved):
lgtm

PR #1822, opened 2024-07-14
COMPLETE USER EXPERIENCE OVERHAUL

pablodanswer:


backend/danswer/configs/chat_configs.py
```diff
@@ -75,6 +75,8 @@
     or "The name of the conversation must be in the same language as the user query."
 )
```
yuhongsun96, line 77: Let's add a comment for what this does and why someone would want to turn it on/off. Also let's include this across the board for all of the deployment options (you can refer to DISABLE_LLM_CHUNK_FILTER to see where env vars are set across deployments)

I would also prefer calling it DISABLE_AGENTIC_SEARCH as we likely will introduce other agent flows. We don't want to make users update their env setup things so renaming these down the line is more hassle.

yuhongsun96: Comment could be like:

Agentic search takes significantly more tokens and therefore has much higher cost. This configuration allows users to get a search-only experience with instant results and no involvement from the LLM.

backend/danswer/configs/model_configs.py
```diff
@@ -128,3 +128,6 @@
         logger.error(
             "Failed to parse LITELLM_PASS_THROUGH_HEADERS, must be a valid JSON object"
         )
+
+# Optinal maximum workers (if 0, disregard)
```
yuhongsun96: I don't think we need this, let's revert this. The other comment I would have here is, MAX_WORKERS is not very descriptive. They may get it confused with workers for indexing or workers for background async jobs.

backend/danswer/server/query_and_chat/query_backend.py
```diff
@@ -136,6 +142,110 @@ def query_validation(
     return QueryValidationResponse(reasoning=reasoning, answerable=answerable)
 
 
+# Search history
+@basic_router.get("/get-user-searches")
```
yuhongsun96: the get is redundant I feel

Also the it probably could just be a single query instead of two but this doesn't matter. No need to change it

backend/danswer/db/chat.py
```diff
@@ -81,17 +87,47 @@ def get_chat_sessions_by_slack_thread_id(
...
     user_id: UUID | None,
     deleted: bool | None,
     db_session: Session,
     include_one_shot: bool = False,
+    only_one_shot: bool = False,
 ) -> list[ChatSession]:
     stmt = select(ChatSession).where(ChatSession.user_id == user_id)
 
-    if not include_one_shot:
+    if not include_one_shot and not only_one_shot:
```
yuhongsun96: Generally it's not good to have one argument be contradictory to another and for one to just have precedence in the contradictory case. Since this would be just a programmer error and fairly unlikely to happen, you can just throw in an assert at the beginning of this and not handle this line as an "and" of both and instead do an if/else-if

backend/danswer/db/chat.py
```diff
@@ -219,6 +255,7 @@ def delete_chat_session(
     db_session.commit()
 
 
+# def get_chat
```
yuhongsun96: nit: cleanup

backend/danswer/db/chat.py
```diff
@@ -81,17 +87,47 @@ def get_chat_sessions_by_slack_thread_id(
...
 ) -> list[ChatSession]:
     stmt = select(ChatSession).where(ChatSession.user_id == user_id)
 
-    if not include_one_shot:
+    if not include_one_shot and not only_one_shot:
         stmt = stmt.where(ChatSession.one_shot.is_(False))
 
+    if only_one_shot:
+        stmt = stmt.where(ChatSession.one_shot.is_(True))
+        stmt = stmt.order_by(desc(ChatSession.id))
```
yuhongsun96: Why is this one ordered whereas the other use cases are not? If it makes sense to order then, we can just always do it, this should not increase latency in any non-trivial way.

Other thing is, technically IDs are strictly increasing but I feel it's better to use time_created instead of ID

backend/danswer/db/chat.py
```diff
@@ -275,6 +312,20 @@ def get_chat_messages_by_sessions(
     return db_session.execute(stmt).scalars().all()
 
 
+def get_search_docs_for_chat_message(
```
yuhongsun96, line 313: This works but I believe SQLAlchemy handles this for you if you just set the relationships correctly in the db models. Let's do that as a standard

backend/danswer/server/query_and_chat/query_backend.py
pablonyx: Definitely a lot of shared logic with get_chat_sessions, but included this as the naming is slightly different for searches (first user search vs name of chat session)- could unify into get_chat_session by just automatically updating search session names with first message's content

backend/danswer/configs/chat_configs.py
pablonyx: Sounds great! Renamed here and in the docs

backend/danswer/configs/model_configs.py
pablonyx: Sounds good- can always add if ends up being a real issue (perhaps for rate limiting reasons with provider apis?)

backend/danswer/db/chat.py
pablonyx: Brain fart! definitely cleaner that way

backend/danswer/db/chat.py
pablonyx: Another brain hiccup- thank you for pointing this one out

backend/danswer/one_shot_answer/answer_question.py
```diff
@@ -204,12 +213,17 @@ def stream_answer_objects(
         # tested quotes with tool calling too much yet
         skip_explicit_tool_calling=True,
         return_contexts=query_req.return_contexts,
+        skip_generation=True,
```
yuhongsun96: I'm pretty sure this breaks the other APIs that rely on this (and that do give back an answer)
Also this breaks the SlackBot no?

backend/danswer/llm/answering/answer.py
```diff
@@ -112,6 +115,7 @@ def __init__(
         skip_explicit_tool_calling: bool = False,
         # Returns the full document sections text from the search tool
         return_contexts: bool = False,
+        skip_generation: bool = False,
```
yuhongsun96: can be even more explicit here with the naming since generation could mean a few things, like we're using a generator for example. Maybe call it "skip_gen_ai_answer_generation"?

backend/danswer/one_shot_answer/answer_question.py
pablonyx: For some reason I was under the false pretension that one_shot's `answer_question` was only used by the search endpoint ad moved this flag from the request to a default. Tested the Slack bot but seems like I happened to have been on a different branch. Will move to the fetch request and include as an attribute to `DirectQARequest` like `evaluate_response` is currntly

backend/danswer/one_shot_answer/answer_question.py
```diff
@@ -204,12 +213,17 @@ def stream_answer_objects(
...
     )
-    # won't be any ImageGenerationDisplay responses since that tool is never passed in
+
+    # won't be any ImageGenerationDisplay responses since that tool is never passed in go
```
yuhongsun96: go?

pablonyx: Keyboard shortcuts betraying me (likely meant to use command bar space)

backend/danswer/one_shot_answer/answer_question.py
```diff
@@ -204,12 +213,17 @@ def stream_answer_objects(
...
     dropped_inds: list[int] = []
+
     for packet in cast(AnswerObjectIterator, answer.processed_streamed_output):
         # for one-shot flow, don't currently do anything with these
         if isinstance(packet, ToolResponse):
+            # add a handling of the response here to add the information regarding relevance to the search results.
```
yuhongsun96: What is this? There is no change here. Was this just a note to add the thing below?

backend/danswer/one_shot_answer/models.py
```diff
@@ -27,12 +27,14 @@ class DirectQARequest(ChunkContext):
...
     prompt_id: int | None
     persona_id: int
+    agentic: bool | None = None
     retrieval_options: RetrievalDetails = Field(default_factory=RetrievalDetails)
     # This is to forcibly skip (or run) the step, if None it uses the system defaults
     skip_rerank: bool | None = None
     skip_llm_chunk_filter: bool | None = None
     chain_of_thought: bool = False
     return_contexts: bool = False
+    evaluate_response: bool = False  # Analyze the results
```
yuhongsun96: Can we give a more descriptive comment?

backend/danswer/prompts/miscellaneous_prompts.py
```diff
@@ -24,6 +24,25 @@
...
+2. Useful Analysis:
+[ANALYSIS_START]
+State the most important point from the chain of thought.
+DO NOT refer to "the document" (describe it as "this")- ONLY state the core point in a description.
+[ANALYSIS_END]
+
+3. Relevance Determination:
+RESULT: True (if potentially relevant)
+RESULT: False (if not relevant)
+"""
```
yuhongsun96: Would add a strip() after the """

backend/danswer/search/pipeline.py
```diff
@@ -74,9 +86,82 @@ def __init__(
...
 
         # Generates reranked chunks and LLM selections
-        self._postprocessing_generator: Iterator[
-            list[InferenceSection] | list[int]
-        ] | None = None
+        self._postprocessing_generator: (
+            Iterator[list[InferenceSection] | list[int]] | None
+        ) = None
+
+    def extract_section(self, content: str, start_tag: str, end_tag: str) -> str:
```
yuhongsun96: not used, let's remove it

backend/danswer/search/pipeline.py
```diff
@@ -74,9 +86,73 @@ def __init__(
...
 
         # Generates reranked chunks and LLM selections
-        self._postprocessing_generator: Iterator[
-            list[InferenceSection] | list[int]
-        ] | None = None
+        self._postprocessing_generator: (
+            Iterator[list[InferenceSection] | list[int]] | None
+        ) = None
+
+    def evaluate(
```
yuhongsun96: This should be separated out into a different module. We want to not have things tightly coupled with classes like the SearchPipeline if the functionality can be separated out into a separate module. It makes things more reuseable, more testable, more extensible. So in this case, the bulk of the functionality should be in a new module in secondary_llm_flows (see the way chunk_usefulness is used for reference)

backend/danswer/server/query_and_chat/query_backend.py
yuhongsun96: as discussed, it was only for the route name

backend/danswer/configs/model_configs.py
yuhongsun96: Ya, that could be valid, we can add it as needed

backend/danswer/configs/chat_configs.py
```diff
@@ -75,6 +75,15 @@
...
 
+
+# Agentic search takes significantly more tokens and therefore has much higher cost.
+# This configuration allows users to get a search-only experience with instant results
+# and no involvement from the LLM.
+DISABLE_AGENTIC_SEARCH = (
```
yuhongsun96, line 84: Your comment reminded me of the other reason we added this, maybe we should make a note of it as well. Some LLM providers have fairly strict rate limits so this would run out their budget very quickly.

backend/danswer/server/query_and_chat/chat_backend.py
```diff
@@ -67,6 +67,8 @@
 from danswer.server.query_and_chat.token_limit import check_token_rate_limits
 from danswer.utils.logger import setup_logger
 
+# from danswer.db.models import SearchDoc
```
yuhongsun96: nit remove

backend/danswer/search/pipeline.py
yuhongsun96: Another nit comment on this one, since this function should never be directly called outside of the other functions/properties of this class, let's make it an underscored function if you still end up having it. But better yet just offload all this functionality as mentioned above and just have the property relevance_summaries call out to that instead.

backend/danswer/server/query_and_chat/models.py
```diff
@@ -171,11 +171,32 @@ def check_click_or_search_feedback(cls: BaseModel, values: dict) -> dict:
 
         if click is False and feedback is None:
             raise ValueError("Empty feedback received.")
-
         return values
 
 
 class ChatMessageDetail(BaseModel):
+    message_id: int
```
yuhongsun96: Since so many fields are shared, they should both subclass a parent class to inherit the shared stuff

backend/danswer/search/pipeline.py
```diff
@@ -74,9 +86,73 @@ def __init__(
...
+        document_id = document.center_chunk.document_id
+        chunk_id = document.center_chunk.chunk_id
+
+        prompt = f"""
+        Analyze the relevance of this document to the search query:
+        Title: {document_id.split("/")[-1]}
+        Blurb: {document.combined_content}
+        Query: {query}
+
+        {AGENTIC_SEARCH_EVALUATION_PROMPT}
```
yuhongsun96: slight preference to have the whole thing as a single string that allows for interpolation, we have some examples of it in the code

backend/danswer/search/pipeline.py
```diff
@@ -74,9 +86,73 @@ def __init__(
...
+        """
+
+        content = "".join(
```
yuhongsun96: there's an invoke option which is better for when streaming is not needed

backend/danswer/search/pipeline.py
```diff
@@ -74,9 +86,73 @@ def __init__(
...
+        if not analysis:
+            logger.warning(
+                f"Couldn't extract proper analysis for document {document_id}. Using full content."
+            )
+            analysis = content
+
+        relevance.content = analysis
+        relevance.relevant = relevant
+
+        results[f"{document_id}-{chunk_id}"] = relevance
```
yuhongsun96: chunks have a unique_id property which you can use for this instead right?

backend/danswer/db/chat.py
```diff
@@ -484,6 +531,27 @@ def get_doc_query_identifiers_from_model(
...
 
 
+def update_search_docs_table_with_relevance(
+    db_session: Session,
+    reference_db_search_docs: list[SearchDoc],
+    relevance_summary: LLMRelevanceSummaryResponse,
+) -> None:
+    for search_doc in reference_db_search_docs:
+        relevance_data = relevance_summary.relevance_summaries.get(
+            f"{search_doc.document_id}-{search_doc.chunk_ind}"
```
yuhongsun96, line 541: better to use the chunk.unique_id here too, better to have this stuff standardized instead of magically constructed strings

backend/danswer/search/pipeline.py
```diff
@@ -323,6 +399,20 @@ def relevant_section_indices(self) -> list[int]:
         )
         return self._relevant_section_indices
 
+    @property
+    def relevance_summaries(self) -> dict[str, RelevanceChunk]:
```
yuhongsun96, line 339: can check if DISABLE_AGENTIC_SEARCH if set here and raise an exception if it's disabled

backend/danswer/server/query_and_chat/query_backend.py
```diff
@@ -136,6 +142,109 @@ def query_validation(
     return QueryValidationResponse(reasoning=reasoning, answerable=answerable)
 
 
+@basic_router.get("/user-searches")
+def get_user_query_sessions(
```
yuhongsun96: would standardize naming wherever possible, like is query and search the same thing? I assume they are

backend/danswer/server/query_and_chat/query_backend.py
```diff
@@ -136,6 +142,109 @@ def query_validation(
...
+        search_session = get_chat_session_by_id(
+            chat_session_id=session_id,
+            user_id=user_id,
+            db_session=db_session,
+            is_shared=is_shared,
+        )
+    except ValueError:
+        raise ValueError("Search session does not exist or has been deleted")
+
+    if search_session.user_id is None and user_id is not None:
```
yuhongsun96: What is this one about?

backend/danswer/db/chat.py
pablonyx: Sent a ping about this one!

backend/danswer/search/pipeline.py
pablonyx: Sounds good! Updated

backend/danswer/server/query_and_chat/models.py
pablonyx: Absolutely

backend/danswer/server/query_and_chat/query_backend.py
pablonyx: Great point- don't want to create another assistants vs persona!

backend/danswer/search/pipeline.py
pablonyx: Makes sense and agree on all fronts

backend/danswer/configs/chat_configs.py
pablonyx: Great point! Adding now

backend/danswer/search/pipeline.py
pablonyx: Noted for future reference as well!

backend/danswer/server/query_and_chat/query_backend.py
pablonyx: Removing- Historical artifact! Was experimenting on a second branch with merging the chat/search flows and that check is needed for chats (due to seeding of first message)

PR #1858, opened 2024-07-18
Add config to skip zendesk article labels

KevinShiCA:
use case: some articles may not be draft articles, but also not appropriate/useful to be indexed. one example is template articles.

this PR adds an env config `ZENDESK_CONNECTOR_SKIP_ARTICLE_LABELS` which accepts a comma delimited list of label names to skip when indexing in the zendesk connector.


Weves (approved):
Looks great! Thanks for adding this 🙇

PR #1861, opened 2024-07-18
Generated images come inline as part of the Model response

emerzon:
This change allows image generation to work on air-gapped environments. The image is received as a base64 string along with the image model response, eliminating the need for a image retrieval step using http.


emerzon:
I noticed that this no longer works in current master - Apparently something changed in the message processing. I kindly ask you to review this part.

emerzon:
Closing as no longer working in current master

PR #1866, opened 2024-07-19
Add GPT-4o-Mini & fix a missing GPT-4o in Model Tests

versecafe:
This adds support for `gpt-4o-mini` at roughly 25 times a lower cost then GPT-4, note this doesn't set 4o-mini to the default but that should be a next step as the accuracy is near identical, max token input is the same, the model runs faster, and saves a massive amount when churning through large amounts of data.


Weves:
Thanks so much for this change, just put it in!

PR #1870, opened 2024-07-19
Added ability to control LLM access based on group

hagen-danswer:


backend/alembic/versions/795b20b85b4b_add_llm_group_permissions_control.py
```diff
@@ -0,0 +1,41 @@
...
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = "795b20b85b4b"
+down_revision = "05c07bf07c00"
+branch_labels = None
+depends_on = None
+
+
```
hagen-danswer, line 17: This stuff just supports the additions made in the db/models.py code

backend/danswer/db/llm.py
```diff
@@ -74,8 +78,29 @@ def fetch_existing_embedding_providers(
     return list(db_session.scalars(select(CloudEmbeddingProviderModel)).all())
```
hagen-danswer, line 100: this maintains identical functionality unless a user is passed in in which case it grabs all llm providers that:
- are public
- the user is in a usergroup that is in a group that has access

backend/danswer/db/models.py
```diff
@@ -1137,6 +1143,7 @@ class Persona(Base):
         viewonly=True,
     )
     # EE only
+    is_public: Mapped[bool] = mapped_column(Boolean, nullable=False, default=True)
```
hagen-danswer, line 1146: just moved this since it's ee only

backend/danswer/server/manage/llm/api.py
```diff
@@ -147,10 +147,10 @@ def set_provider_as_default(
 
 @basic_router.get("/provider")
 def list_llm_provider_basics(
```
hagen-danswer, line 149: we now pass in the user to make use of the new feature

web/src/app/admin/models/llm/CustomLLMProviderUpdateForm.tsx
```diff
@@ -307,121 +330,179 @@ export function CustomLLMProviderUpdateForm({
...
-                ) : existingLlmProvider ? (
-                  "Update"
-                ) : (
-                  "Enable"
+              label="[Optional] Fast Model"
+              placeholder="E.g. gpt-4"
+            />
+
+            <Divider />
+
```
hagen-danswer, line 404: This is the start of the actual changes I made.

hagen-danswer: besides a couple imports above to support this section

web/src/app/admin/models/llm/CustomLLMProviderUpdateForm.tsx
```diff
@@ -307,121 +330,179 @@ export function CustomLLMProviderUpdateForm({
...
-
-                    mutate(LLM_PROVIDERS_ADMIN_URL);
-                    onClose();
-                  }}
-                >
-                  Delete
-                </Button>
+              </>
+            )}
+
```
hagen-danswer, line 470: this is the end of my changes

web/src/app/admin/models/llm/LLMProviderUpdateForm.tsx
```diff
@@ -293,6 +308,61 @@ export function LLMProviderUpdateForm({
 
           <Divider />
 
+          {isPaidEnterpriseFeaturesEnabled && userGroups && (
```
hagen-danswer: did the same thing as the other form. let me know if i missed any forms but i think these are the only two?

There may have been a way to break this out into its own component but i figured it wasnt worth the effort atm since we are about to rework the admin stuff anyways

backend/danswer/db/llm.py
```diff
@@ -1,15 +1,41 @@
...
+from danswer.db.models import User
+from danswer.db.models import User__UserGroup
+from danswer.db.models import UserGroup
 from danswer.server.manage.embedding.models import CloudEmbeddingProvider
 from danswer.server.manage.embedding.models import CloudEmbeddingProviderCreationRequest
 from danswer.server.manage.llm.models import FullLLMProvider
 from danswer.server.manage.llm.models import LLMProviderUpsertRequest
 
 
+def update_group_llm_provider_relationships(
```
hagen-danswer, line 17: this removes all LLMProvider__UserGroup for the given LLMProvider.

then, if group_ids are given, adds group LLMProvider__UserGroup back to that table.

backend/danswer/db/llm.py
```diff
@@ -36,36 +62,35 @@ def upsert_llm_provider(
...
-        provider=llm_provider.provider,
-        api_key=llm_provider.api_key,
-        api_base=llm_provider.api_base,
-        api_version=llm_provider.api_version,
-        custom_config=llm_provider.custom_config,
-        default_model_name=llm_provider.default_model_name,
-        fast_default_model_name=llm_provider.fast_default_model_name,
-        model_names=llm_provider.model_names,
-        is_default_provider=None,
+
```
hagen-danswer, line 66: simplified this function and added in logic for the groups

backend/danswer/db/llm.py
```diff
@@ -36,36 +63,41 @@ def upsert_llm_provider(
...
+    # Make sure the relationship table stays up to date
+    update_group_llm_provider_relationships(
+        llm_provider_id=existing_llm_provider.id,
+        group_ids=llm_provider.groups,
+        db_session=db_session,
     )
-    db_session.add(llm_provider_model)
+
+    user_groups = (
```
Weves: I don't think is needed? This seems to do the same thing as `update_group_llm_provider_relationships`

Weves (approved):
Left one small comment

PR #1885, opened 2024-07-21
Remove visit api when unneded

pablodanswer:


backend/danswer/search/pipeline.py
```diff
@@ -240,32 +245,33 @@ def _get_sections(self) -> list[InferenceSection]:
...
         flat_ranges = [r for ranges in merged_ranges for r in ranges]
+        flattened_inference_chunks = []
+        parallel_functions_with_args = []
 
         for chunk_range in flat_ranges:
-            functions_with_args.append(
-                (
-                    # If Large Chunks are introduced, additional filters need to be added here
-                    self.document_index.id_based_retrieval,
+            if chunk_range.start == chunk_range.end:
```
Weves: can we add a comment here explaining why we can do this (e.g. no need to fetch anything additional from vespa if there is no range, so this improves latency)

backend/danswer/search/pipeline.py
```diff
@@ -169,36 +171,39 @@ def _get_sections(self) -> list[InferenceSection]:
...
-                            ),
-                        )
-                    )
-
-            list_inference_chunks = run_functions_tuples_in_parallel(
-                functions_with_args, allow_failures=False
-            )
+                    list_inference_chunks.append([chunk])
+
+                    # NOTE: left in for eventual refactor with Vespa API
```
Weves: can we leave this all the same as before? With this, it won't actually use full docs so the behavior will just be incorrect, right?

backend/danswer/search/pipeline.py
Weves: actually, I don't think we need to ever do any retrieval as long as `above` / `below` are both 0

backend/danswer/search/pipeline.py
```diff
@@ -240,32 +245,33 @@ def _get_sections(self) -> list[InferenceSection]:
         merged_ranges = [
             merge_chunk_intervals(ranges) for ranges in doc_chunk_ranges_map.values()
         ]
+
         flat_ranges = [r for ranges in merged_ranges for r in ranges]
+        flattened_inference_chunks = []
```
Weves: prefer if we could type these to make it more clear what's in them

backend/danswer/search/pipeline.py
```diff
@@ -154,6 +155,7 @@ def _get_chunks(self) -> list[InferenceChunk]:
 
         return cast(list[InferenceChunk], self._retrieved_chunks)
 
+    @log_function_time()
```
Weves: lets use `print_only` here so we don't send any telemetry

Weves (approved):
lgtm!

PR #1905, opened 2024-07-23
Implemented LLM disabling for api call

hagen-danswer:


yuhongsun96 (approved):
Nice! Well thought out solution!

PR #1911, opened 2024-07-24
Handle SSL error tracebacks in site indexing connector

ShukantPal:
Hey 👋, just checked out Danswer's code! I found an issue with indexing my website that was being mis-diagnosed by Danswer.

---

My website (https://shukantpal.com) uses Let's Encrypt certificates, which aren't accepted by the Python urllib certificate verifier for some reason. My website is set up correctly otherwise (https://www.sslshopper.com/ssl-checker.html#hostname=www.shukantpal.com)

This change adds a fix so the correct traceback is shown in Danswer, instead of a generic "unable to connect, check your Internet connection".

---

Screenshot 2024-07-23 at 10 49 51 PM


Weves (approved):
Awesome, this looks great!

Next up for us is to figure out why the Python urllib certificate verifier doesn't handle Let's Encrypt certificates 🧐

PR #1913, opened 2024-07-24
Add Missing Condition to If Condition for Tool Usage

mymro:
Currently an Error is thrown, if for example file upload is use, because no tool is selected. This is indicated by setting a flag on the force_use_tool object. This flag is however never checked, resulting in an error.


Weves:
This was fixed in https://github.com/danswer-ai/danswer/pull/1928 (basically the same thing).

Great catch!

PR #1914, opened 2024-07-24
fix pointers to various background tasks and scripts

rkuo-danswer:
fix paths to background job scripts


yuhongsun96:
looks good!

PR #1920, opened 2024-07-24
Fixed login issue

hagen-danswer:


backend/danswer/auth/users.py
```diff
@@ -92,7 +93,19 @@ def user_needs_to_be_verified() -> bool:
...
+        raise PermissionError("User not on allowed user whitelist")
+
+
+def verify_email_in_whitelist(
+    email: str,
+    db_session: Session = Depends(get_session),
+) -> None:
+    if get_user_by_email(email, db_session):
+        return
+
```
Weves: can we call `verify_email_is_invited` here to avoid duplicating?

Weves (approved):
left one small comment, otherwise lgtm

PR #1924, opened 2024-07-24
Added check for unknown tool names

hagen-danswer:


backend/danswer/llm/answering/answer.py
```diff
@@ -240,9 +240,15 @@ def _raw_output_for_explicit_tool_calling_llms(
         # if we have a tool call, we need to call the tool
         tool_call_requests = tool_call_chunk.tool_calls
         for tool_call_request in tool_call_requests:
-            tool = [
+            known_tools_by_name = [
                 tool for tool in self.tools if tool.name == tool_call_request["name"]
-            ][0]
+            ]
+
+            if not known_tools_by_name:
```
hagen-danswer, line 247: There is an argument to be made that this should raise an exception instead of moving on with a warning because it means that the user has indicated a tool with a name that doesnt match any names in self.tools.

Weves (approved):

PR #1934, opened 2024-07-25
 [FEATURE] Add gpu support to prod docker-compose file

CharlesBdg:
Add gpu support to docker-compose production files.


yuhongsun96:
We're trying to avoid adding too many duplicate compose files because it increase the risk of them falling out of sync. Perhaps it's better handled as a README that gives the steps to set up GPU across any of the deployment methods?

akhoury:
@yuhongsun96 hi, any update here? thank you

akhoury:
+1

PR #1942, opened 2024-07-25
Configurable models + updated assistants bar

pablodanswer:
Updated model logic
- Configurable list of user-accessible models
- Opinionated defaults for each provider

Updated Chat input bar tabs generally 

Screenshot 2024-07-25 at 4 30 11 PM
Screenshot 2024-07-25 at 4 30 23 PM
Screenshot 2024-07-25 at 4 30 30 PM
Screenshot 2024-07-25 at 4 31 37 PM
Screenshot 2024-07-25 at 4 31 44 PM


Weves:
Overall looks amazing! A few small notes:

- When I change model, I'd expect the popup to close (this will also fix this jumping around problem) 

https://github.com/user-attachments/assets/f86942ce-439d-4783-8194-9587e41344cd

- If we can apply the default "set of models" for existing LLM providers, that would be great

- When updating an assistant (to switch to gpt4-o mini) I get this error

https://github.com/user-attachments/assets/f1f51613-dd74-44ca-ba29-b7d37d497086

PR #1947, opened 2024-07-26
Add new admin page

pablodanswer:
Here's a draft PR description based on the context provided:

## Description
This PR introduces a significant rework of the admin page to modify the connection flow for all connectors. The main changes include:

1. Centralizing the flow for all connectors
2. Modifying Index Attempts to reference ConnectorCredentialPair instead of separate connector/credentials
3. Adding ability to swap credentials and persist/view credentials separately from connectors
4. Removing all previous connector-specific creation pages

These changes aim to streamline the connector management process and provide more flexibility in handling credentials.

#### Migration details

1. Adding a new `indexing_start` column to the `connector` table
2. Adding a `type` column to the `credentials` table
3. Associating index attempts with ConnectorCredentialPair instead of separate connector and credential entries

## How Has This Been Tested?
- Test each credential page + modal that is available

Credentials not tested
- Loopio
- Request Tracker
- ZenDesk
- Document360
- Axero 
- Hubspot
- BookStack
- Gong

## Checklist:


backend/alembic/versions/4ea2c93919c1_add_type_to_credentials.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    # Add the new 'source' column to the 'credential' table
+    op.add_column(
+        "credential",
+        sa.Column(
+            "source",
+            sa.Enum(
```
yuhongsun96: We don't want any enums in the DB explicitly, it's caused issues with upgrades and enums changing.

The enums are enforced in the code

backend/alembic/versions/4ea2c93919c1_add_type_to_credentials.py
```diff
@@ -0,0 +1,112 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "4ea2c93919c1"
+down_revision = "325975216eb3"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
```
yuhongsun96: Slight preference to have this stuff cleaned up

backend/alembic/versions/4ea2c93919c1_add_type_to_credentials.py
```diff
@@ -0,0 +1,112 @@
...
+    op.add_column(
+        "credential",
+        sa.Column(
+            "name",
+            sa.String(),
+            nullable=True,
+        ),
+    )
+
+    op.execute(
```
yuhongsun96, line 42: For stuff like this, better to leave some comments on why we're doing this and how it works, easier for future new members of the team to digest the code base

backend/danswer/db/credentials.py
```diff
@@ -74,6 +79,68 @@ def fetch_credential_by_id(
     return credential
 
 
+def fetch_credentials_by_source(
+    db_session: Session,
+    user: User | None,
```
yuhongsun96, line 84: this is unused but it probably should be. Should filter by the user having access or the credential being admin public. Everything is admin public right now but we may have personal connectors in the future

backend/danswer/db/credentials.py
```diff
@@ -74,6 +79,68 @@ def fetch_credential_by_id(
...
+    document_source: DocumentSource | None = None,
+) -> list[Credential]:
+    base_query = select(Credential).where(Credential.source == document_source)
+    credentials = db_session.execute(base_query).scalars().all()
+    return list(credentials)
+
+
+def swap_credentials_connector(
```
yuhongsun96, line 93: There can also be some permission check here just in case, we don't want any exploits even if for now it's fine (as I'm sure only admins can hit the endpoints that call these). Things like user owns the new credential being assigned

yuhongsun96 (commented):
Things to discuss (btw next time we can do a design review first):
- Should we make a ticket to port all cc-pair into to the connector, what are the benefits? We could even remove the relationship table, so connector just has a credential (or None). Does this complicate things when we have personal connectors (like Confluence personal spaces etc)? What should go in cc-pair vs connector?
- How does the default Credential work now?

PR #1951, opened 2024-07-26
Bugfix/persona access

rkuo-danswer:
fixes chat access to personas where the user is in the authorized list of users or groups


Weves (approved):

Weves:
actually, appears to be a mypy issue that would be great to address 🧐

PR #1957, opened 2024-07-27
Rework tokenizer

hagen-danswer:


hagen-danswer:
Tests Tried:
indexing a document with default, openai, cohere embedders and switching between them all

backend/danswer/natural_language_processing/utils.py
```diff
@@ -1,84 +1,128 @@
...
 transformer_logging.set_verbosity_error()
 os.environ["TOKENIZERS_PARALLELISM"] = "false"
 os.environ["HF_HUB_DISABLE_TELEMETRY"] = "1"
 os.environ["TRANSFORMERS_NO_ADVISORY_WARNINGS"] = "1"
 
 
-_TOKENIZER: tuple[Optional["AutoTokenizer"], str | None] = (None, None)
-_LLM_TOKENIZER: Any = None
-_LLM_TOKENIZER_ENCODE: Callable[[str], Any] | None = None
+def _set_encoder_from_provider(provider_type: str) -> Any:
```
yuhongsun96: This is a pure function, there is no "set" happening here and it should not be called this

backend/danswer/natural_language_processing/utils.py
```diff
@@ -1,84 +1,128 @@
...
+    import tiktoken
 
+    if provider_type.lower() == "OpenAI".lower():
+        return tiktoken.get_encoding("cl100k_base")
+    from tokenizers import Tokenizer  # type:ignore
```
yuhongsun96: If you're going to optimize for saving on the import, it should probably bo in the if statement to not use it unnecessarily.

However, I think the default should be previous HuggingFace BERT default tokenizer or whatever it was.

backend/danswer/natural_language_processing/search_nlp_models.py
```diff
@@ -119,7 +120,11 @@ def encode(
         # Batching for local embedding
         text_batches = batch_list(texts, batch_size)
         embeddings: list[list[float] | None] = []
+        logger.info(
```
yuhongsun96: This type of log is way too spammy, we don't need it. Maybe we can have it be debug logging but I would honestly just not have it preferably

backend/danswer/indexing/chunker.py
```diff
@@ -124,18 +124,18 @@ def chunk_document(
     tokenizer = get_default_tokenizer()
 
     blurb_splitter = SentenceSplitter(
-        tokenizer=tokenizer.tokenize, chunk_size=blurb_size, chunk_overlap=0
+        tokenizer=tokenizer.encode, chunk_size=blurb_size, chunk_overlap=0
```
yuhongsun96: typically tokenize doesn't include the CLS and other special tokens and just splits the text into subword tokens. This prevents us from overcounting with every string we concatenate.
So we would rather use tokenize than encode.

image

backend/danswer/natural_language_processing/utils.py
```diff
@@ -1,84 +1,128 @@
...
+
+    logger.warning(
+        f"Invalid Encoder Provider selected: {provider_type}\n"
+        "Supported providers: OpenAI, Cohere\n"
+        "Defaulting to OpenAI..."
+    )
+    return tiktoken.get_encoding("cl100k_base")
+
+
+class UnifiedTokenizer:
```
yuhongsun96: You should have an abstract CountTokenizer or something which just specifies the interfaces. Everywhere else it should be typed as the abstract class type so it knows what functions/attributes it should have access to.

You should then implement an OpenAITokenizer or TiktokenTokenizer and another one for HuggingFaceTokenizer. Up to you if you want to have another tokenizer for Claude or if it should just use the same Huggingface one with some arguments for the init.

yuhongsun96: It should have a tokenize function and that can be it honestly. For the one where it's encode, you can just comment in the implementation that it's giving back the ints. But then make sure to check the front and back for special tokens, and if it does prepend/append those, you should remove them.

backend/model_server/encoders.py
```diff
@@ -89,8 +89,18 @@ def _embed_openai(
         # OpenAI does not seem to provide truncation option, however
         # the context lengths used by Danswer currently are smaller than the max token length
         # for OpenAI embeddings so it's not a big deal
-        response = self.client.embeddings.create(input=texts, model=model)
-        return [embedding.embedding for embedding in response.data]
+        try:
+            response = self.client.embeddings.create(input=texts, model=model)
+            return [embedding.embedding for embedding in response.data]
+        except BaseException as e:
```
yuhongsun96: Please be thorough with your changes. It's very easy to get the full context of if this makes sense.

image

yuhongsun96 (changes requested):
The main issue is the miscounting. Would prefer to avoid any stacking "off by 2" errors when building up prompts.

Also, the whole point of using the classes and abstractions isn't really used here so would prefer a rework there

backend/danswer/background/update.py
```diff
@@ -349,8 +349,7 @@ def update_loop(
     if db_embedding_model.cloud_provider_id is None:
         logger.info("Running a first inference to warm up embedding model")
         warm_up_encoders(
```
hagen-danswer, line 357: just pass in the embedding model object instead of pulling out all the info and passing it 1 by 1

backend/danswer/chat/process_message.py
```diff
@@ -308,7 +308,13 @@ def stream_chat_message_objects(
         except GenAIDisabledException:
             raise RuntimeError("LLM is disabled. Can't use chat flow without LLM.")
 
-        llm_tokenizer = get_default_llm_tokenizer()
+        llm_provider = llm.config.model_provider
+        llm_model_name = llm.config.model_name
+
```
hagen-danswer, line 313: explicitly call correct tokenizer (if available)

backend/danswer/indexing/chunker.py
```diff
@@ -114,14 +117,18 @@ def _get_metadata_suffix_for_document_index(
 
 def chunk_document(
     document: Document,
+    embedder: IndexingEmbedder,
     chunk_tok_size: int = DOC_EMBEDDING_CONTEXT_SIZE,
     subsection_overlap: int = CHUNK_OVERLAP,
     blurb_size: int = BLURB_SIZE,  # Used for both title and content
     include_metadata: bool = not SKIP_METADATA_IN_CHUNK,
 ) -> list[DocAwareChunk]:
     from llama_index.text_splitter import SentenceSplitter
```
hagen-danswer, line 169: explicitly call correct tokenizer (if available)

backend/danswer/indexing/chunker.py
```diff
@@ -255,31 +262,71 @@ def chunk_document(
...
 
-def split_chunk_text_into_mini_chunks(
-    chunk_text: str, mini_chunk_size: int = MINI_CHUNK_SIZE
+def _split_chunk_text_into_mini_chunks(
+    chunk_text: str, embedder: IndexingEmbedder, mini_chunk_size: int = MINI_CHUNK_SIZE
 ) -> list[str]:
     """The minichunks won't all have the title prefix or metadata suffix
     It could be a significant percentage of every minichunk so better to not include it
     """
     from llama_index.text_splitter import SentenceSplitter
```
hagen-danswer, line 304: explicitly call correct tokenizer (if available)

backend/danswer/llm/answering/answer.py
```diff
@@ -139,7 +139,10 @@ def __init__(
         self.prompt_config = prompt_config
 
         self.llm = llm
```
hagen-danswer, line 141: explicitly call correct tokenizer (if available)

backend/danswer/llm/answering/prompts/build.py
```diff
@@ -66,7 +66,10 @@ def __init__(
         self.system_message_and_token_cnt: tuple[SystemMessage, int] | None = None
         self.user_message_and_token_cnt: tuple[HumanMessage, int] | None = None
```
hagen-danswer, line 68: explicitly call correct tokenizer (if available)

backend/danswer/llm/answering/prune_and_merge.py
```diff
@@ -135,8 +135,12 @@ def _apply_pruning(
     is_manually_selected_docs: bool,
     use_sections: bool,
     using_tool_message: bool,
+    llm_config: LLMConfig,
 ) -> list[InferenceSection]:
-    llm_tokenizer = get_default_llm_tokenizer()
+    llm_tokenizer = get_tokenizer(
```
hagen-danswer, line 140: explicitly call correct tokenizer (if available)

backend/danswer/llm/answering/prune_and_merge.py
```diff
@@ -165,14 +169,15 @@ def _apply_pruning(
             )
         )
```
hagen-danswer, line 171: more accurate name

backend/danswer/natural_language_processing/search_nlp_models.py
```diff
@@ -85,7 +86,10 @@ def encode(
                 tokenizer_trim_content(
                     content=text,
                     desired_length=self.max_seq_length,
```
hagen-danswer, line 88: explicitly call correct tokenizer (if available)

backend/danswer/natural_language_processing/search_nlp_models.py
```diff
@@ -192,29 +200,34 @@ def predict(
...
+    normalize = embedding_model.normalize
+    provider_type = embedding_model.provider_type
     warm_up_str = (
         "Danswer is amazing! Check out our easy deployment guide at "
         "https://docs.danswer.dev/quickstart"
     )
 
     # May not be the exact same tokenizer used for the indexing flow
-    get_default_tokenizer(model_name=model_name)(warm_up_str)
+    logger.info(f"Warming up encoder model: {model_name}")
```
hagen-danswer, line 216: explicitly call correct tokenizer (if available)

backend/danswer/one_shot_answer/answer_question.py
```diff
@@ -117,16 +117,22 @@ def stream_answer_objects(
         one_shot=True,
         danswerbot_flow=danswerbot_flow,
     )
+    llm, fast_llm = get_llms_for_persona(persona=chat_session.persona)
```
hagen-danswer, line 121: explicitly call correct tokenizer (if available)

backend/danswer/search/preprocessing/danswer_helper.py
```diff
@@ -74,7 +68,12 @@ def recommend_search_flow(
 
     # UNK tokens -> suggest Keyword (still may be valid QA)
     # TODO do a better job with the classifier model and retire the heuristics
```
hagen-danswer, line 70: explicitly call correct tokenizer (if available)

backend/danswer/server/documents/document.py
```diff
@@ -50,7 +50,10 @@ def get_document_info(
 
     # get actual document context used for LLM
     first_chunk = inference_chunks[0]
```
hagen-danswer, line 52: explicitly call correct tokenizer (if available)

backend/danswer/server/query_and_chat/chat_backend.py
```diff
@@ -442,6 +443,14 @@ def seed_chat(
         root_message = get_or_create_root_message(
             chat_session_id=new_chat_session.id, db_session=db_session
         )
+        llm, fast_llm = get_llms_for_persona(persona=new_chat_session.persona)
+
```
hagen-danswer, line 447: explicitly call correct tokenizer (if available)

backend/danswer/tools/message.py
```diff
@@ -24,9 +24,9 @@ class ToolCallSummary(BaseModel):
     tool_call_result: ToolMessage
```
hagen-danswer, line 26: you have to pass in a tokenizer now

backend/danswer/tools/utils.py
```diff
@@ -22,13 +20,9 @@ def explicit_tool_calling_supported(model_provider: str, model_name: str) -> boo
     return False
 
 
-def compute_tool_tokens(tool: Tool, llm_tokenizer: Encoding | None = None) -> int:
-    if not llm_tokenizer:
-        llm_tokenizer = get_default_llm_tokenizer()
+def compute_tool_tokens(tool: Tool, llm_tokenizer: BaseTokenizer) -> int:
     return len(llm_tokenizer.encode(json.dumps(tool.tool_definition())))
```
hagen-danswer, line 26: you have to pass in a tokenizer now

backend/model_server/encoders.py
```diff
@@ -89,8 +89,18 @@ def _embed_openai(
         # OpenAI does not seem to provide truncation option, however
         # the context lengths used by Danswer currently are smaller than the max token length
         # for OpenAI embeddings so it's not a big deal
```
hagen-danswer, line 91: better logging if we run into errors in the future

backend/danswer/one_shot_answer/qa_utils.py
```diff
@@ -18,16 +17,14 @@ def simulate_streaming_response(model_out: str) -> Generator[str, None, None]:
 def combine_message_thread(
     messages: list[ThreadMessage],
     max_tokens: int | None,
```
hagen-danswer, line 19: standardize passing around the tokenizer

backend/danswer/indexing/models.py
```diff
@@ -51,7 +51,12 @@ def to_short_descriptor(self) -> str:
         )
```
hagen-danswer: this is an intermediate chunk between chunks with embeddings and DocAwareChunks

backend/danswer/indexing/embedder.py
```diff
@@ -69,41 +76,27 @@ def __init__(
...
     def embed_chunks(
         self,
-        chunks: list[DocAwareChunk],
-        enable_mini_chunk: bool = ENABLE_MINI_CHUNK,
+        chunks_with_texts: list[TextChunk],
     ) -> list[IndexChunk]:
-        # Cache the Title embeddings to only have to do it once
-        title_embed_dict: dict[str, list[float] | None] = {}
-        embedded_chunks: list[IndexChunk] = []
-
```
hagen-danswer, line 78: minichunker moved to chunker.py.

its now an intermediate step between chunking and embedding

backend/danswer/indexing/chunker.py
```diff
@@ -255,31 +262,71 @@ def chunk_document(
...
+        provider_type=embedder.provider_type,
+    ).tokenize
     sentence_aware_splitter = SentenceSplitter(
         tokenizer=token_count_func, chunk_size=mini_chunk_size, chunk_overlap=0
     )
 
     return sentence_aware_splitter.split_text(chunk_text)
 
 
+def extract_chunk_texts_from_doc_aware_chunk(
```
hagen-danswer: this is now an intermediate step between chunking and embedding

this may be able to be done at the end of the chunking function itself (and therefor kept entirely out of the indexing_pipline)

backend/danswer/danswerbot/slack/listener.py
```diff
@@ -471,8 +471,7 @@ def _initialize_socket_client(socket_client: SocketModeClient) -> None:
                         embedding_model = get_current_db_embedding_model(db_session)
                         if embedding_model.cloud_provider_id is None:
                             warm_up_encoders(
```
hagen-danswer, line 473: just pass in the embedding model object instead of pulling out all the info and passing it 1 by 1

PR #1958, opened 2024-07-27
Allow removal of max_output_tokens by setting GEN_AI_MAX_OUTPUT_TOKENS=0

emerzon:
## Description
This allows the removal of the explicit max_output_tokens. By setting GEN_AI_MAX_OUTPUT_TOKENS=0 the actual max_output_tokens will be whatever the value supported by the model in use.

## How Has This Been Tested?
Change deployed and tested with different models with different output token values (Ie. gpt-4: 4096, gemini-1.5-pro: 8192)

## Accepted Risk
None known

## Related Issue(s)
None


yuhongsun96:
Looks good 👍

PR #1965, opened 2024-07-29
Fixed default indexing frequency

hagen-danswer:
## Description
Fixed the default indexing frequency

## How Has This Been Tested?
I created a connector and printed out the indexing frequency

### DIFF
--- a/web/src/app/admin/connectors/[connector]/AddConnectorPage.tsx
+++ b/web/src/app/admin/connectors/[connector]/AddConnectorPage.tsx
@@ -87,9 +87,10 @@ export default function AddConnector({
   );
 
   // Default to 10 minutes unless otherwise specified
-  const defaultRefresh = (configuration.overrideDefaultFreq || 10) / 60;
+  const defaultRefresh = configuration.overrideDefaultFreq || 10;
+  // default is 1 day (in minutes)
+  const defaultPrune = 24 * 60;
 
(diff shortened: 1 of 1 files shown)


web/src/app/admin/connectors/[connector]/AddConnectorPage.tsx
```diff
@@ -139,9 +140,9 @@ export default function AddConnector({
 
   const createConnector = async () => {
     const AdvancedConfig: AdvancedConfig = {
-      pruneFreq: pruneFreq || defaultRefresh * 60,
+      pruneFreq: (pruneFreq || defaultPrune) * 60,
       indexingStart,
-      refreshFreq: refreshFreq * 60,
+      refreshFreq: (refreshFreq || defaultRefresh) * 60,
```
Comment (pablonyx) [inline web/src/app/admin/connectors/[connector]/AddConnectorPage.tsx:145]: Nit: Probably fine for now, but worth considering that we may want 0 for a frequency to evaluate to "never refresh / prune" - here 0 would always evaluate to default time (no matter how the backend handles this.)

From my memory, we already handle this on the backend in the API, so a `??` may be more technically accurate.

Comment (yuhongsun96) [reply web/src/app/admin/connectors/[connector]/AddConnectorPage.tsx:145]: In the interest of speed, I'm putting this one in, Pablo if you have better suggestions feel free to open a new PR. Also, we'll want to support no refresh and no prune options sometime soon (this week)

PR #1974, opened 2024-07-30
Fix flow where oidc_expiry is different from token expiry

Weves:
## Description
If the OIDC token has expired, treat the user as not authenticated. Currently, the "needs to login" popup appears if you come back to the app after not using for a while, but we should just auto-log them in.

## How Has This Been Tested?
Tested OIDC via Okta locally.

## Accepted Risk
N/A

## Related Issue(s)
N/A

## Checklist:

### DIFF
--- a/backend/danswer/auth/users.py
+++ b/backend/danswer/auth/users.py
@@ -347,6 +347,12 @@ async def double_check_user(
             detail="Access denied. User is not verified.",
         )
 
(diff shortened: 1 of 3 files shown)


pablodanswer (approved):
LGTM- clean implementation of the redirection functionality we talked about putting in eventually

PR #1994, opened 2024-07-31
fix transaction already begun issue on doc set upsertZ

ThomaciousD:
## Description
This pull request fixes a bug that occurs when creating or updating a document set, which resulted in the error "Error: transaction is already begun". The issue was due to an unnecessary explicit transaction start (`db_session.begin()`) which conflicted with SQLAlchemy's automatic transaction management. The fix involves removing the explicit transaction start and adding proper error logging.

### Changes:
- Removed the explicit `db_session.begin()` call.
- Added error logging in the exception block for better debugging.

## How Has This Been Tested?
The fix has been tested directly on the web UI. The following steps were taken to verify the changes:
1. Created a new document set to ensure no errors occur.
2. Updated an existing document set to ensure the error does not reappear.
3. Repeated the above steps multiple times to confirm consistency and reliability.

## Accepted Risk
There are no known risks or failure modes associated with this fix. The changes are isolated to the document set creation and update process, and extensive testing has shown no adverse effects on other functionalities.

### DIFF
--- a/backend/danswer/db/document_set.py
+++ b/backend/danswer/db/document_set.py
@@ -16,8 +16,10 @@ from danswer.db.models import DocumentSet as DocumentSetDBModel
 from danswer.db.models import DocumentSet__ConnectorCredentialPair
 from danswer.server.features.document_set.models import DocumentSetCreationRequest
 from danswer.server.features.document_set.models import DocumentSetUpdateRequest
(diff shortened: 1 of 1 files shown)


yuhongsun96:
I believe this has already been addressed by latest, let me know if this issue is still coming up with recent versions!

PR #2004, opened 2024-07-31
Adding maximum attachment size for Confluence connector

emerzon:
## Description
Adds a limiter to the maximum size of the attachments to be fetch by the Confluence Connector, using a default size of 1MB.

## How Has This Been Tested?
Indexing a Confluence Space with Large files and check in the log that they have been properly skipped.


emerzon:
Closing for refactoring

PR #2011, opened 2024-08-01
Slack connector: add name to each contributor of a thread

danielnaber:
## Description

danswer misattributed facts from threads in Slack in my tests (i.e. claimed that X said something when it was actually someone else from the thread who said it). Adding the author's name seems to help with that.

I also changed the `__main__` part to be more useful for testing -> print all docs, not just the first one.

## How Has This Been Tested?
Manually tested results with a few cases. Indexing works with a large Slack instance with >50K docs indexed.

## Accepted Risk
I'm not aware of any risks (but I'm not very experienced with writing connectors).

## Checklist:

### DIFF
--- a/backend/danswer/connectors/slack/connector.py
+++ b/backend/danswer/connectors/slack/connector.py
@@ -182,17 +182,31 @@ def thread_to_doc(
 
     doc_sem_id = f"{initial_sender_name} in #{channel['name']}: {snippet}"
 
-    return Document(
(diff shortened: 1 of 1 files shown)


yuhongsun96:
The Slack connector is pretty prone to getting rate limited so this was a conscious choice to not add significantly more calls for each thread read. The value added is not high enough to cause a multiple in the indexing time :(

Thanks for the suggestion, hope you're enjoying Danswer!

PR #2020, opened 2024-08-01
Add answers to search

pablodanswer:
## Description

## How Has This Been Tested?

## Accepted Risk

## Related Issue(s)

## Checklist:


backend/danswer/chat/models.py
```diff
@@ -46,15 +46,18 @@ class LLMRelevanceFilterResponse(BaseModel):
...
-    # done at the section level currently rather than the chunk
-    relevant: bool | None = None
+class RelevanceAnalysis(BaseModel):
+    relevant: bool
     content: str | None = None
 
 
-class LLMRelevanceSummaryResponse(BaseModel):
-    relevance_summaries: dict[str, RelevanceChunk]
+class SectionRelevancePiece(RelevanceAnalysis):
```
yuhongsun96, line 54: Why is this Piece, is there some combination of this later on?

backend/danswer/chat/process_message.py
```diff
@@ -501,6 +503,7 @@ def stream_chat_message_objects(
                         chunks_above=new_msg_req.chunks_above,
                         chunks_below=new_msg_req.chunks_below,
                         full_doc=new_msg_req.full_doc,
+                        evaluation_type=LLMEvaluationType.BASIC,
```
yuhongsun96: My feeling on general best practices with this stuff, when the value is set, it should be set far up the chain at the right place where it makes sense, in this case this is the right place. It should be the right value set there, and then later during the evaluation it should check if that value is allowed. Ask me if this isn't making sense!

yuhongsun96: DISABLE_LLM_DOC_RELEVANCE

backend/danswer/chat/models.py
yuhongsun96: add a comment on why it's called piece

backend/danswer/chat/models.py
```diff
@@ -46,15 +46,18 @@ class LLMRelevanceFilterResponse(BaseModel):
...
+    document_id: str
+    chunk_id: int
+
+
+class DocumentRelevance(BaseModel):
```
yuhongsun96, line 61: would comment on this being for all of the documents of a search

backend/danswer/chat/process_message.py
yuhongsun96: - needs to disable both the agent and the non-agent relevance checks if set
- should this be unspecified type? (see below)

backend/danswer/one_shot_answer/answer_question.py
```diff
@@ -184,6 +185,11 @@ def stream_answer_objects(
     search_tool = SearchTool(
         db_session=db_session,
         user=user,
+        evaluation_type=LLMEvaluationType.SKIP
```
yuhongsun96, line 190: Prefer having symmetry across all the levels of abstraction. Let's make this:

class DirectQARequest(ChunkContext):
    messages: list[ThreadMessage]
    prompt_id: int | None
    persona_id: int
    agentic: bool | None = None
    
make the agentic boolean into the same enum of options, that way we don't have to translate 2 choices into 3. We can still have the DISABLE variable to always map it to the SKIP option here though

backend/danswer/one_shot_answer/answer_question.py
```diff
@@ -184,6 +185,11 @@ def stream_answer_objects(
...
+        if DISABLE_AGENTIC_SEARCH
+        else LLMEvaluationType.AGENTIC
+        if query_req.agentic
+        else LLMEvaluationType.BASIC,
```
yuhongsun96: Also since Chris and I both have a slight preference towards this point - let's not do nested conditionals in this style, it's easier for someone else to come through, misunderstand it, and make a mistake

backend/danswer/prompts/agentic_evaluation.py
```diff
@@ -28,7 +28,8 @@
 """
 
 AGENTIC_SEARCH_USER_PROMPT = """
-Document:
+
+Document Title: `{title}`
```
yuhongsun96: Thoughts on adding metadata here similar to how we do for generating the answer?

backend/danswer/one_shot_answer/models.py
```diff
@@ -31,7 +32,8 @@ class DirectQARequest(ChunkContext):
     retrieval_options: RetrievalDetails = Field(default_factory=RetrievalDetails)
     # This is to forcibly skip (or run) the step, if None it uses the system defaults
     skip_rerank: bool | None = None
-    skip_llm_chunk_filter: bool | None = None
+    evaluation_type: LLMEvaluationType = LLMEvaluationType.BASIC
```
yuhongsun96: prefer using the same name, it's also clearer with llm_evaluation_type. I think using the same thing instead of mutating it slightly here and there makes it generally easy to trace the impact of things across the code base.

backend/danswer/search/pipeline.py
```diff
@@ -332,44 +337,59 @@ def reranked_sections(self) -> list[InferenceSection]:
...
-                evaluate_inference_section, (section, self.search_query.query, self.llm)
-            )
-            for section in sections
-        ]
+        elif self.search_query.evaluation_type == LLMEvaluationType.AGENTIC:
+            if DISABLE_AGENTIC_SEARCH:
+                raise ValueError(
+                    "Agentic search operation called while DISABLE_AGENTIC_SEARCH is enabled."
+                )
+                # self._section_relevance = []
```
yuhongsun96: nit, also please very carefully scrutinize your own PRs before submitting for review!

backend/danswer/tools/search/search_tool.py
```diff
@@ -274,19 +282,29 @@ def run(self, **kwargs: str) -> Generator[ToolResponse, None, None]:
...
-            section_relevance_list=search_pipeline.section_relevance_list,
-            prompt_config=self.prompt_config,
-            llm_config=self.llm.config,
-            question=query,
-            document_pruning_config=self.pruning_config,
-        )
+        final_context_sections = search_pipeline.final_context_sections
+
+        if search_pipeline.search_query.evaluation_type == LLMEvaluationType.AGENTIC:
+            final_context_sections = [
```
yuhongsun96: Need to double check this logic, I'm worried there needs to be some pruning done here too

backend/danswer/tools/search/search_tool.py
```diff
@@ -274,19 +282,29 @@ def run(self, **kwargs: str) -> Generator[ToolResponse, None, None]:
             ),
         )
 
+        # if self.llm_doc_eval and not DISABLE_AGENTIC_SEARCH:
```
yuhongsun96: nit

backend/danswer/tools/search/search_tool.py
```diff
@@ -274,19 +282,29 @@ def run(self, **kwargs: str) -> Generator[ToolResponse, None, None]:
...
 
-        final_context_sections = prune_and_merge_sections(
-            sections=search_pipeline.reranked_sections,
-            section_relevance_list=search_pipeline.section_relevance_list,
-            prompt_config=self.prompt_config,
-            llm_config=self.llm.config,
-            question=query,
-            document_pruning_config=self.pruning_config,
-        )
+        final_context_sections = search_pipeline.final_context_sections
```
yuhongsun96: prefer just putting search_pipeline.final_context_sections on line 301 sections=search_pipeline.final_context_sections, instead of out here and have another variable named the same thing assigned to something slightly logically different (the pruned sections)

backend/danswer/search/pipeline.py
```diff
@@ -332,44 +337,59 @@ def reranked_sections(self) -> list[InferenceSection]:
...
     @property
     def section_relevance_list(self) -> list[bool]:
+        llm_indices = relevant_sections_to_indices(
+            relevance_chunks=self.section_relevance,
+            inference_sections=self.final_context_sections,
+        )
         return [
-            True if ind in self.relevant_section_indices else False
-            for ind in range(len(self.reranked_sections))
+            True if ind in llm_indices else False
```
yuhongsun96: not that you wrote this code but, it can just be ind in llm_indices

backend/danswer/chat/models.py
```diff
@@ -46,15 +46,18 @@ class LLMRelevanceFilterResponse(BaseModel):
...
+class RelevanceAnalysis(BaseModel):
+    relevant: bool
     content: str | None = None
 
 
-class LLMRelevanceSummaryResponse(BaseModel):
-    relevance_summaries: dict[str, RelevanceChunk]
+class SectionRelevancePiece(RelevanceAnalysis):
+    document_id: str
+    chunk_id: int
```
yuhongsun96: would specify as a comment here that this is the center chunk

backend/danswer/search/preprocessing/preprocessing.py
```diff
@@ -137,18 +138,22 @@ def retrieval_preprocessing(
         access_control_list=user_acl_filters,
     )
 
-    llm_chunk_filter = False
-    if search_request.skip_llm_chunk_filter is not None:
-        llm_chunk_filter = not search_request.skip_llm_chunk_filter
+    llm_evaluation_type = LLMEvaluationType.BASIC
+    if search_request.evaluation_type is not None:
```
yuhongsun96: This is never None anymore, I think we may need a fourth type for the LLMEvaluationType for UNSPECIFIED

backend/danswer/db/chat.py
```diff
@@ -541,11 +541,11 @@ def get_doc_query_identifiers_from_model(
 def update_search_docs_table_with_relevance(
     db_session: Session,
     reference_db_search_docs: list[SearchDoc],
-    relevance_summary: LLMRelevanceSummaryResponse,
+    relevance_summary: DocumentRelevance,
 ) -> None:
     for search_doc in reference_db_search_docs:
         relevance_data = relevance_summary.relevance_summaries.get(
-            f"{search_doc.document_id}-{search_doc.chunk_ind}"
+            search_doc.document_id
```
yuhongsun96, line 548: How come this only needs the document now? doesn't chat use the non-agent flow which uses non-merged sections?

backend/danswer/search/pipeline.py
```diff
@@ -332,44 +337,59 @@ def reranked_sections(self) -> list[InferenceSection]:
...
+            else:
+                sections = self.final_context_sections
+                functions = [
+                    FunctionCall(
+                        evaluate_inference_section,
+                        (section, self.search_query.query, self.llm),
+                    )
+                    for section in sections
+                ]
+                results = run_functions_in_parallel(function_calls=functions)
```
yuhongsun96: I think this should also run postprocessing (though not in parallel with reranking). The logic should be something like, pass in all the sections and optionally the merged sections into the postprocessing. Since the reranking and the merged sections no longer have a 1 to 1 matching, it can just be, for each reranked chunk, find the document id and that's the one the merged section matches with, if that makes sense

backend/danswer/search/postprocessing/postprocessing.py
```diff
@@ -49,7 +51,7 @@ def should_rerank(query: SearchQuery) -> bool:
 
 
 def should_apply_llm_based_relevance_filter(query: SearchQuery) -> bool:
```
yuhongsun96: let's just remove this function and put the logic where it's used

backend/danswer/search/preprocessing/preprocessing.py
```diff
@@ -137,18 +138,22 @@ def retrieval_preprocessing(
...
+        llm_evaluation_type = search_request.evaluation_type
     elif persona:
-        llm_chunk_filter = persona.llm_relevance_filter
+        llm_evaluation_type = (
+            LLMEvaluationType.Basic
+            if persona.llm_relevance_filter
+            else LLMEvaluationType.SKIP
+        )
 
     if disable_llm_chunk_filter:
```
yuhongsun96: this should be the shared env variable DISABLE_LLM_DOC_RELEVANCE

backend/danswer/one_shot_answer/models.py
yuhongsun96: more importantly, this should be type UNSPECIFIED I think

backend/danswer/search/models.py
```diff
@@ -78,7 +78,7 @@ class SearchRequest(ChunkContext):
     hybrid_alpha: float = HYBRID_ALPHA
     # This is to forcibly skip (or run) the step, if None it uses the system defaults
     skip_rerank: bool | None = None
-    skip_llm_chunk_filter: bool | None = None
+    evaluation_type: LLMEvaluationType = LLMEvaluationType.SKIP
```
yuhongsun96: Let's default this to UNSPECIFIED?

backend/ee/danswer/server/query_and_chat/query_backend.py
```diff
@@ -105,19 +110,26 @@ def handle_search_request(
     # Deduping happens at the last step to avoid harming quality by dropping content early on
     deduped_docs = top_docs
     dropped_inds = None
+
     if search_request.retrieval_options.dedupe_docs:
         deduped_docs, dropped_inds = dedupe_documents(top_docs)
 
+    llm_indices = relevant_documents_to_indices(
+        relevance_chunks=relevant_chunks,
```
yuhongsun96: let's call these sections and in the function as well to be more correct

backend/ee/danswer/server/query_and_chat/query_backend.py
```diff
@@ -76,7 +81,7 @@ def handle_search_request(
     )
     top_sections = search_pipeline.reranked_sections
     # If using surrounding context or full doc, this will be empty
```
yuhongsun96: this comment is no longer true,

yuhongsun96 (commented):
Things to consider:
- We need a UNSPECIFIED type to allow persona settings to get applied
- Slight preference to have the llm filter in the postprocessing function as well
- Need to handle the truncation for the agent flow somehow

backend/danswer/search/models.py
```diff
@@ -92,7 +92,7 @@ class SearchQuery(ChunkContext):
     offset: int = 0
     search_type: SearchType = SearchType.HYBRID
     skip_rerank: bool = not ENABLE_RERANKING_REAL_TIME_FLOW
-    skip_llm_chunk_filter: bool = DISABLE_LLM_CHUNK_FILTER
+    evaluation_type: LLMEvaluationType
```
yuhongsun96: non-default args on top, easier for reader to follow

PR #2039, opened 2024-08-04
Fix secondary index attempts showing up as the primary index status +…

Weves:
… scheduling while in-progress

## How Has This Been Tested?
Created a secondary index attempt with status failed. Verified that it previously would display the incorrect status and then that after the change it shows the correct status. 

Also verified that we would not schedule multiple index attempts for the secondary runs.

### DIFF
--- a/backend/danswer/server/documents/connector.py
+++ b/backend/danswer/server/documents/connector.py
@@ -422,7 +422,9 @@ def get_connector_indexing_status(
         )
 
         latest_finished_attempt = get_latest_finished_index_attempt_for_cc_pair(
-            connector_credential_pair_id=cc_pair.id, db_session=db_session
+            connector_credential_pair_id=cc_pair.id,
+            secondary_index=secondary_index,
+            db_session=db_session,
         )
 
         indexing_statuses.append(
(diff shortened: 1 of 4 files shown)


backend/danswer/server/documents/connector.py
```diff
@@ -433,24 +435,26 @@ def get_connector_indexing_status(
                 credential=CredentialSnapshot.from_credential_db_model(credential),
                 public_doc=cc_pair.is_public,
                 owner=credential.user.email if credential.user else "",
-                last_finished_status=latest_finished_attempt.status
-                if latest_finished_attempt
-                else None,
-                last_status=latest_index_attempt.status
-                if latest_index_attempt
-                else None,
+                last_finished_status=(
```
Weves, line 438: lines below are just better formatting

web/src/app/admin/indexing/status/CCPairIndexingStatusTable.tsx
```diff
@@ -130,7 +130,7 @@ function ConnectorRow({
   ccPairsIndexingStatus,
   invisible,
 }: {
-  ccPairsIndexingStatus: any;
+  ccPairsIndexingStatus: ConnectorIndexingStatus<any, any>;
```
Weves, line 133: better typing!! Prefer not using any as much as humanly possible

web/src/app/admin/indexing/status/CCPairIndexingStatusTable.tsx
```diff
@@ -354,12 +356,40 @@ export function CCPairIndexingStatusTable({
             name: "Sample File Connector",
             last_status: "success",
             connector: {
+              name: "Sample File Connector",
```
Weves, line 359: TODO: do this better

PR #2042, opened 2024-08-05
EE Connector Deletion Bugfix

nschwerd:
## Description
- Clean up user groups relationships when connectors are deleted
- Move disabled status to CCPair itself rather than the connector
- Add new `DELETING` state
- Keep trying to delete, never allow a connector to go back from `DELETING` state
- Don't include `DELETING` state connectors when applying document sets / user groups

## How Has This Been Tested?
- Manual testing performed, validated backing Vepsa documents and Postgres tables
- Created and ran automated tests

## Accepted Risk
N/A

## Checklist:


backend/danswer/db/document.py
```diff
@@ -152,6 +136,9 @@ def get_acccess_info_for_documents(
                 == ConnectorCredentialPair.credential_id,
             ),
         )
+        # don't include CC pairs that are being deleted
```
Weves, line 139: No longer consider `DELETING` cc pairs for the ACL

backend/danswer/db/document_set.py
```diff
@@ -459,6 +445,10 @@ def fetch_document_sets_for_documents(
             Document.id == DocumentByConnectorCredentialPair.id,
         )
         .where(Document.id.in_(document_ids))
+        # don't include CC pairs that are being deleted
```
Weves, line 448: No longer consider `DELETING` cc pairs for document set syncing

backend/danswer/background/connector_deletion.py
```diff
@@ -78,25 +74,37 @@ def delete_connector_credential_pair_batch(
...
-                ),
             )
+
+            # update Vespa
+            logger.debug(f"Updating documents: {document_ids_to_update}")
             update_requests = [
                 UpdateRequest(
                     document_ids=[document_id],
                     access=access,
+                    document_sets=new_doc_sets_for_documents[document_id],
```
Weves, line 105: Update ACL / document sets in real time as we delete batches rather than as a separate sync after the fact

backend/danswer/background/connector_deletion.py
```diff
@@ -108,48 +116,6 @@ def delete_connector_credential_pair_batch(
             db_session.commit()
 
 
-def cleanup_synced_entities(
```
Weves, line 111: No longer do any additional syncing after the fact—by the time we've gone through all the batches everything in vespa should be correct

PR #2043, opened 2024-08-05
Add user specific model defaults

pablodanswer:
## Description
Lets users set defualt models

## How Has This Been Tested?
- [x] Various combinations of assistant model defaults, model override, global model settings, etc.
- [x] Without auth (uses the dynamic config store)
- [x] With auth (basic + google)

## Expected behavior
- Persist globally set default models
- If no assistant specified model + no currently elected in chat bar model, use + select the globally set user-specific model!

## Related Issue(s)

## Checklist:


pablodanswer:
Screenshot 2024-08-04 at 10 17 10 PM
Screenshot 2024-08-04 at 10 17 16 PM
Screenshot 2024-08-04 at 10 18 08 PM

web/src/app/admin/assistants/AssistantEditor.tsx
```diff
@@ -531,13 +531,13 @@ export function AssistantEditor({
                     </TooltipProvider>
                   </div>
                   <p className="my-1 text-text-600">
-                    You assistant will use your system default (currently{" "}
-                    {defaultModelName}) unless otherwise specified below.
+                    Your assistant will use your user&apos;s set default unless
```
Weves: nit: `Your assistant will use your user's` -> `Your assistant will use the user's`. A lot of "your"s, feels a bit of confusing. Also would be nice to add something like "For you, this would be `X`". This is especially helpful when individuals (not admins) are creating assistants.

web/src/app/chat/ChatPage.tsx
```diff
@@ -116,7 +121,10 @@ export function ChatPage({
   const chatSessionIdRef = useRef<number | null>(existingChatSessionId);
 
   // LLM
-  const llmOverrideManager = useLlmOverride(selectedChatSession);
+  const llmOverrideManager = useLlmOverride(
+    user?.preferences.default_model!,
```
Weves: if user is always non-null, can we either adjust it in the `useChatContext` typing or handle it here (e.g. by setting it to null)?

web/src/app/chat/modal/SetDefaultModelModal.tsx
```diff
@@ -0,0 +1,171 @@
...
+          </h2>
+        </div>
+
+        <Text className="mb-4">
+          Select a default Large Language Model (Generative AI model) to power
+          this Assistant.
+          {defaultModel == null && "  No default model has been selected!"}
+        </Text>
+        <div className="w-full flex text-sm flex-col">
+          <div key={-1} className="w-full border-b hover:bg-gray-50">
```
Weves: nit: use named colors (here + in a few other places)

Weves (changes requested):

web/src/app/chat/modal/SetDefaultModelModal.tsx
pablonyx: need to add this to my local pre-commit to remind myself!

web/src/app/chat/modal/SetDefaultModelModal.tsx
```diff
@@ -0,0 +1,174 @@
...
+        {popup}
+        <div className="flex mb-4">
+          <h2 className="text-2xl text-emphasis font-bold flex my-auto">
+            Set Default Model
+          </h2>
+        </div>
+
+        <Text className="mb-4">
+          Select a default Large Language Model (Generative AI model) to power
+          this Assistant.
```
Weves: huh, I swear I added this comment but it seems I forgot to click confirm or something 😅 

This probably shouldn't be `this Assistant` since it applies to many assistants? Maybe something like "Select a default Large Language Model (Generative AI model) to power <Danswer | Whitelabeled Name>".

Additionally, it may be worth thinking about the best way to avoid confusion when the user selects a default model, but then they use an assistant with an overridden default 🤔

Weves (commented):
prettier isn't happy 😢 

Also we may need to rebase again due to https://github.com/danswer-ai/danswer/pull/2104/files

Weves (approved):

PR #2061, opened 2024-08-06
Don't persist error perennially

pablodanswer:
## Description
- Removes check for "error message in last index" before updating message history

This makes the entire error handling process much smoother as the chat message history doesn't break like it used to 
- Now uses a new ref that tracks the actual state of the chat messages + session loading to determine when to erase the previous errors
## How Has This Been Tested?

## Accepted Risk

## Related Issue(s)

## Checklist:

### DIFF
--- a/web/src/app/chat/ChatPage.tsx
+++ b/web/src/app/chat/ChatPage.tsx
@@ -266,7 +266,6 @@ export function ChatPage({
 
       const newMessageMap = processRawChatHistory(chatSession.messages);
       const newMessageHistory = buildLatestMessageChain(newMessageMap);
-      // if the last message is an error, don't overwrite it
 
       // Update message history except for edge where where
       // last message is an error and we're on a new chat.
--- a/web/src/app/chat/ChatPage.tsx
+++ b/web/src/app/chat/ChatPage.tsx
@@ -115,6 +115,7 @@ export function ChatPage({
   const chatSessionIdRef = useRef<number | null>(existingChatSessionId);
 
   // Only updates on message load (ie. rename / switching chat session)
+  // Useful for determining which session has been loaded (i.e. still on `new, empty session` or `previous session`)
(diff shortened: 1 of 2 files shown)


Weves (commented):
This happens without it: 

https://github.com/user-attachments/assets/be476b6a-71fc-4236-a8b2-a718c192e1d5

😓

pablodanswer:
https://github.com/user-attachments/assets/1b1e4dbd-2612-49da-88cf-976b641dd031

web/src/app/chat/ChatPage.tsx
```diff
@@ -244,7 +250,13 @@ export function ChatPage({
       const newMessageMap = processRawChatHistory(chatSession.messages);
       const newMessageHistory = buildLatestMessageChain(newMessageMap);
       // if the last message is an error, don't overwrite it
```
Weves: can we add a comment here about the situations where this may be relevant?

web/src/app/chat/ChatPage.tsx
```diff
@@ -244,7 +250,13 @@ export function ChatPage({
...
-      if (messageHistory[messageHistory.length - 1]?.type !== "error") {
+
+      if (
+        !(
```
Weves: nit: prefer to remove the `!` and just flip each condition internally. This feels pretty hard to understand currently

web/src/app/chat/ChatPage.tsx
```diff
@@ -244,7 +251,15 @@ export function ChatPage({
       const newMessageMap = processRawChatHistory(chatSession.messages);
       const newMessageHistory = buildLatestMessageChain(newMessageMap);
       // if the last message is an error, don't overwrite it
```
Weves: nit: we should remove this old message?

pablonyx: In fact we should!

Weves (approved):

PR #2065, opened 2024-08-06
Improve seeding (includes all enterprise features)

pablodanswer:
## Description
Seeding now supports (in addition to previous functioanlity)
- White labelling configuration
- Analytics script

## Doubts
- [ ] To enable backwards compatibility (we previously allowed for the "name" to be set in  a separate part of the `ENV_SEED` variable (despite it now being a part of the more encompassing `EnterpriseSettings` attribute - if a user has set the custom name but not the EnterpriseSettings attribute, we will continue seeding the enterprise name
  - We can remove this.
- [ ] Need an analytics script to test functionality

## Accepted Risk

## Related Issue(s)

## Checklist:

### DIFF
--- a/backend/ee/danswer/server/seeding.py
+++ b/backend/ee/danswer/server/seeding.py
@@ -15,7 +15,9 @@ from danswer.server.manage.llm.models import LLMProviderUpsertRequest
 from danswer.server.settings.models import Settings
 from danswer.server.settings.store import store_settings as store_base_settings
 from danswer.utils.logger import setup_logger
+from ee.danswer.server.enterprise_settings.models import AnalyticsScriptUpload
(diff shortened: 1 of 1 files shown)


backend/ee/danswer/server/seeding.py
```diff
@@ -103,12 +106,35 @@ def _seed_settings(settings: Settings) -> None:
...
+        except ValueError as e:
+            logger.error(f"Failed to seed analytics script: {str(e)}")
+
+
 def get_seed_config() -> SeedConfiguration | None:
     return _parse_env()
 
 
 def seed_db() -> None:
+    print("SEEEDING")
```
Weves: we should probably remove these

pablonyx: Agreed

PR #2094, opened 2024-08-09
Openai encoding temp hotfix

hagen-danswer:
Tested through the search eval pipeline and confirmed to work

Will need to be replaced with permanent fix later

### DIFF
--- a/backend/danswer/natural_language_processing/utils.py
+++ b/backend/danswer/natural_language_processing/utils.py
@@ -111,18 +111,12 @@ def _check_tokenizer_cache(tokenizer_name: str) -> BaseTokenizer:
     return _TOKENIZER_CACHE[tokenizer_name]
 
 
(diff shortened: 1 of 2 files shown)


backend/danswer/natural_language_processing/utils.py
```diff
@@ -111,18 +111,12 @@ def _check_tokenizer_cache(tokenizer_name: str) -> BaseTokenizer:
     return _TOKENIZER_CACHE[tokenizer_name]
```
hagen-danswer, line 113: This now just uses the default tokenizer no matter what.
Extra character handling is done elsewhere

yuhongsun96: there's some irony here and I love it lmao

yuhongsun96 (approved):

PR #2099, opened 2024-08-09
Speed up docker launch

hagen-danswer:


backend/model_server/main.py
```diff
@@ -32,22 +31,31 @@
 logger = setup_logger()
 
 
-async def manage_huggingface_cache() -> None:
-    temp_hf_cache = Path("/root/.cache/temp_huggingface")
-    hf_cache = Path("/root/.cache/huggingface")
-    if temp_hf_cache.is_dir() and any(temp_hf_cache.iterdir()):
-        hf_cache.mkdir(parents=True, exist_ok=True)
-        for item in temp_hf_cache.iterdir():
+def manage_huggingface_cache() -> None:
```
rkuo-danswer: A little wary of just hardcoding a bunch of directories into a function and just wrapping a recursive move function within that.  The source and target directories could be parameters to this function, for example, so that we aren't hiding a bunch of state or assumptions inside a nested function.

Also, the function itself could use a comment explaining in english the flow of what it's trying to do, and the issues you ran into mentioned in slack that you are trying to work around.

backend/model_server/main.py
```diff
@@ -32,22 +31,31 @@
...
-    else:
-        logger.info("Source directory is empty or does not exist. Skipping copy.")
+                target = dest / item.relative_to(source)
+                target.parent.mkdir(parents=True, exist_ok=True)
+                if target.exists():
+                    continue
+                shutil.move(str(item), str(target))
+
+    _move_files_recursively(temp_hf_cache, hf_cache)
+    shutil.rmtree(temp_hf_cache.parent, ignore_errors=True)
```
rkuo-danswer: It seems odd to be deleting the directory parent. without context, it seems like it could have dangerous side effects.  If this is truly OK, then perhaps passing in the parent directory might be more readable in terms of understanding that it might be side effected/deleted.

backend/model_server/main.py
```diff
@@ -32,22 +31,31 @@
...
-                    shutil.copytree, item, hf_cache / item.name, dirs_exist_ok=True
-                )
+                _move_files_recursively(item, dest / item.relative_to(source))
             else:
-                await asyncio.to_thread(shutil.copy2, item, hf_cache)
-        await asyncio.to_thread(shutil.rmtree, temp_hf_cache)
-        logger.info("Copied contents of temp_huggingface and deleted the directory.")
-    else:
-        logger.info("Source directory is empty or does not exist. Skipping copy.")
+                target = dest / item.relative_to(source)
```
rkuo-danswer: it seems like dest / item.relative_to(source) is used in both code paths. Can't this be lifted out?

backend/model_server/main.py
```diff
@@ -32,22 +31,31 @@
...
+                target.parent.mkdir(parents=True, exist_ok=True)
```
rkuo-danswer: I'm finding it odd to try to create the directory as we see the file. Shouldn't the directory exist already or shouldn't we create it ahead of time?

backend/model_server/main.py
hagen-danswer: Agreed!

backend/model_server/main.py
hagen-danswer: it sure can

backend/model_server/main.py
hagen-danswer: True, I changed what the base directory was and explicitly deleted that

backend/model_server/main.py
hagen-danswer: it doesnt necessarily already exist and creating it ahead of time makes this pretty complicated

rkuo-danswer (approved):

PR #2105, opened 2024-08-10
Updated search settings

pablodanswer:
## Description

## How Has This Been Tested?

## Accepted Risk

## Related Issue(s)

## Checklist:


backend/danswer/document_index/vespa/index.py
```diff
@@ -134,16 +160,17 @@ def ensure_indices_exist(
...
             DANSWER_CHUNK_REPLACEMENT_PAT, self.index_name
         ).replace(VESPA_DIM_REPLACEMENT_PAT, str(index_embedding_dim))
+        schema = add_ngrams_to_schema(schema) if needs_reindexing else schema
         zip_dict[f"schemas/{schema_names[0]}.sd"] = schema.encode("utf-8")
 
         if self.secondary_index_name:
             upcoming_schema = schema_template.replace(
                 DANSWER_CHUNK_REPLACEMENT_PAT, self.secondary_index_name
             ).replace(VESPA_DIM_REPLACEMENT_PAT, str(secondary_index_embedding_dim))
+            upcoming_schema = upcoming_schema
```
yuhongsun96: does this do anything?

backend/danswer/indexing/models.py
```diff
@@ -108,7 +108,9 @@ def from_model(
         embedding_model: "EmbeddingModel",
     ) -> "EmbeddingModelDetail":
         return cls(
-            model_name=embedding_model.model_name,
+            # When constructing EmbeddingModel Detail for user-facing flows, strip the
+            # unneeded additional data after the `_`s
+            model_name=embedding_model.model_name.split("__")[0],
```
yuhongsun96: I slightly prefer using something that is basically impossible to "get unlucky" with. Like suffixing with "__danswer_alt_index" instead of __alt and checking here just for __

yuhongsun96: __danswer_alt_index can be a globally set constant string as well just for consistency

backend/danswer/server/manage/search_settings.py
```diff
@@ -55,11 +55,10 @@ def set_new_embedding_model(
 
         embed_model_details.cloud_provider_id = cloud_id
 
+    # account for same model name being indexed with two different configurations
     if embed_model_details.model_name == current_model.model_name:
-        raise HTTPException(
-            status_code=status.HTTP_400_BAD_REQUEST,
-            detail="New embedding model is the same as the currently active one.",
-        )
+        if "__" not in current_model.model_name:
```
yuhongsun96: would use the more specific pattern here (see comment above) and check for it explicitly.

Also what if they

yuhongsun96 (approved):
Looks good, tiny request

backend/danswer/server/settings/api.py
```diff
@@ -99,21 +100,32 @@ def get_user_notifications(
...
-            user=user, notif_type=NotificationType.REINDEX, db_session=db_session
-        )
-        return [Notification.from_model(notif)]
-
-    if len(reindex_notifs) > 1:
-        logger.error("User has multiple reindex notifications")
-
-    reindex_notif = reindex_notifs[0]
+    try:
+        with db_session.begin_nested():  # Ensure we don't  overcount the reindex notifications
```
yuhongsun96: Would add a slightly more explicit comment here on how explicitly tying the read operations here to the transaction so that they can run extremely close in time to the write (as it's passed to the DB together rather than in 2 calls).

Also would make this just begin transaction instead of nested.

backend/danswer/document_index/vespa/index.py
pablodanswer: It does not!

pablonyx: It does not!

PR #2137, opened 2024-08-14
Model inference for connector classifier on queries

shukant-danswer:
## Description

The integrates the connector classifier model in the search preprocessing pipeline. Instead of using a LLM for auto detecting the sources, we use our own model instead if `FEATURE_FLAG_USE_CONNECTOR_CLASSIFIER` is set in the environment.

The model is warmed up by the API if the feature flag is set.

## How Has This Been Tested?
Hard coded `auto_detect_source_filter` and tested search queries in Danswer's UI.

## Checklist:


backend/danswer/main.py
```diff
@@ -344,6 +346,12 @@ async def lifespan(app: FastAPI) -> AsyncGenerator:
                 model_server_port=MODEL_SERVER_PORT,
             )
 
+        if os.getenv("FEATURE_FLAG_USE_CONNECTOR_CLASSIFIER"):
```
yuhongsun96: Would throw this one in app_configs.py or chat_configs.py for consistency

backend/danswer/secondary_llm_flows/source_filter.py
```diff
@@ -42,11 +44,36 @@ def _sample_document_sources(
         return random.sample(valid_sources, num_sample)
 
 
+def _sample_documents_using_custom_connector_classifier(
+    query: str,
+    valid_sources: list[DocumentSource],
+) -> list[DocumentSource] | None:
+    available_connectors = list(filter(
+        lambda conn: conn.lower() in query.lower(),
```
yuhongsun96: would also remove all whitespaces and punctuation just to be safe, in case some user types Git Hub or something

backend/shared_configs/configs.py
```diff
@@ -14,9 +14,12 @@
 )
 
 # Danswer custom Deep Learning Models
+CONNECTOR_CLASSIFIER_MODEL_REPO = "ShukantP/danswer-connector-classifier"
+CONNECTOR_CLASSIFIER_MODEL_TAG = "3fa7999c89ae72bc3892448dbd3a77c429e49153"
```
yuhongsun96: instead of using a SHA let's use SemVer standards or something approximately like that, can do v1.0.0

backend/shared_configs/model_server_models.py
```diff
@@ -7,6 +7,16 @@
 Embedding = list[float]
 
 
+class ConnectorClassificationRequest(BaseModel):
+    available_connectors: list[str]
+    query: str
+
+
+class ConnectorClassificationResponse(BaseModel):
```
yuhongsun96, line 15: I think this should abstract away the filter boolean and only return the connectors to filter by if any

backend/model_server/custom_models.py
```diff
@@ -59,6 +103,76 @@ def get_local_intent_model(
     return _INTENT_MODEL
 
 
+def tokenize_connector_classification_query(
+    connectors: list[str],
+    query: str,
+    tokenizer: PreTrainedTokenizer,
+    connector_token_end_id: int,
+):
```
yuhongsun96: Be sure to static type returns as well

backend/model_server/custom_models.py
```diff
@@ -59,6 +103,76 @@ def get_local_intent_model(
...
+    input_ids = torch.tensor([tokenizer.cls_token_id], dtype=torch.long)
+    attention_mask = torch.tensor([1], dtype=torch.long)
+
+    for connector in connectors:
```
yuhongsun96, line 128: where are you adding the prefix tokens?

backend/model_server/danswer_torch_model.py
```diff
@@ -66,3 +66,54 @@ def from_pretrained(cls, load_directory: str) -> "HybridClassifier":
...
+    def forward(
+        self,
+        input_ids: torch.Tensor,
+        attention_mask: torch.Tensor,
+    ):
+        hidden_states = self.distilbert(input_ids=input_ids, attention_mask=attention_mask).last_hidden_state
+
+        cls_position_id = input_ids == self.tokenizer.cls_token_id
+        cls_hidden_states = hidden_states[cls_position_id]
+        global_logits = self.connector_global_classifier(cls_hidden_states).view(-1)
```
yuhongsun96, line 105: I would have done this on just the leading CLS token specifically

backend/model_server/danswer_torch_model.py
```diff
@@ -66,3 +66,54 @@ def from_pretrained(cls, load_directory: str) -> "HybridClassifier":
...
+    def __init__(self, config: DistilBertConfig) -> None:
+        super().__init__()
+
+        self.config = config
+        self.distilbert = DistilBertModel(config)
+        self.connector_global_classifier = nn.Linear(self.distilbert.config.dim, 1)
+        self.connector_match_classifier = nn.Linear(self.distilbert.config.dim, 1)
+        self.tokenizer = DistilBertTokenizer.from_pretrained('distilbert-base-uncased')
+        self.dropout = nn.Dropout(self.distilbert.config.seq_classif_dropout)
+        self.connector_end_token_id = self.tokenizer.get_vocab()[self.config.connector_end_token]
```
yuhongsun96: would add a comment here on what this is

backend/model_server/custom_models.py
```diff
@@ -59,6 +103,76 @@ def get_local_intent_model(
...
+    connector_classifier_tokenizer = get_connector_classifier_tokenizer()
+    connector_classifier = get_local_connector_classifier()
+
+    input_ids, attention_mask = tokenize_connector_classification_query(
+        ["GitHub"],
+        "danswer classifier query google doc",
+        connector_classifier_tokenizer,
+        connector_classifier.connector_end_token_id,
+    )
+    input_ids = input_ids.to(next(connector_classifier.parameters()).device)
```
yuhongsun96: slightly prefer for the model to know it's own device rather than checking the layers, see other recent PR

backend/model_server/danswer_torch_model.py
```diff
@@ -66,3 +66,54 @@ def from_pretrained(cls, load_directory: str) -> "HybridClassifier":
...
+class ConnectorClassifier(nn.Module):
+    def __init__(self, config: DistilBertConfig) -> None:
+        super().__init__()
+
+        self.config = config
+        self.distilbert = DistilBertModel(config)
+        self.connector_global_classifier = nn.Linear(self.distilbert.config.dim, 1)
+        self.connector_match_classifier = nn.Linear(self.distilbert.config.dim, 1)
+        self.tokenizer = DistilBertTokenizer.from_pretrained('distilbert-base-uncased')
+        self.dropout = nn.Dropout(self.distilbert.config.seq_classif_dropout)
```
yuhongsun96: From my experience, when writing the inference version of models, people generally don't include this even as a layer definition

backend/model_server/danswer_torch_model.py
```diff
@@ -66,3 +66,54 @@ def from_pretrained(cls, load_directory: str) -> "HybridClassifier":
...
+        connector_end_position_ids = (input_ids == self.connector_end_token_id)
+        connector_end_hidden_states = hidden_states[connector_end_position_ids]
+        connector_end_hidden_states = self.dropout(connector_end_hidden_states)
+        classifier_output = self.connector_match_classifier(connector_end_hidden_states)
+        classifier_confidence = torch.nn.functional.sigmoid(classifier_output).view(-1)
+
+        return global_confidence, classifier_confidence
+
+    @classmethod
+    def from_pretrained(cls, repo_dir: str) -> "ConnectorClassifier":
```
yuhongsun96, line 116: I would lock this to inference mode and turn off all gradient tracking

backend/model_server/main.py
```diff
@@ -68,6 +69,7 @@ async def lifespan(app: FastAPI) -> AsyncGenerator:
     logger.info(f"Torch Threads: {torch.get_num_threads()}")
 
     if not INDEXING_ONLY:
+        # warm_up_connector_classifier_model()
```
yuhongsun96: nit: rm this and the import and the function definition, let's not have bloat in the code. We will never warm up from the model server because even if this is always used, it's only applicable for Slack so we don't want to warm up this model on server startup regardless if we remove the feature flag.

backend/model_server/danswer_torch_model.py
```diff
@@ -66,3 +66,54 @@ def from_pretrained(cls, load_directory: str) -> "HybridClassifier":
...
+            os.path.join(repo_dir, "pytorch_model.pt"),
+            map_location=device,
+            weights_only=True,
+        )
+
+        model = cls(config)
+        model.load_state_dict(state_dict)
+        model.to(device)
+
+        return model
```
yuhongsun96: nit: newline

backend/model_server/custom_models.py
```diff
@@ -59,6 +103,76 @@ def get_local_intent_model(
     return _INTENT_MODEL
 
 
+def tokenize_connector_classification_query(
```
yuhongsun96, line 113: generally with these types of functions, it's good to give a comment of what it might look like, write out the natural language representations of the tokens as a list or something along with the mask etc.

backend/model_server/danswer_torch_model.py
ghost: I typically like separating blocks of code with newlines; removed it here, but it's pretty similar in the `HybridClassifier`

backend/model_server/danswer_torch_model.py
ghost: Good pt. The `HybridClassifier` can also be cleaned up in this regard.

backend/danswer/secondary_llm_flows/source_filter.py
ghost: It depends if you want to lean towards filtering vs not filtering. It could backfire if the user asks "sending mail" and that becomes "gmail"

backend/model_server/custom_models.py
ghost: after the names of each connector, L131 now

backend/model_server/custom_models.py
ghost: I can see the appeal of that - updated.

backend/danswer/secondary_llm_flows/source_filter.py
yuhongsun96: ya... but the model should handle it! I see this step as more of a complexity reduction but we can lean towards allowing connectors to be passed to the model if it's a case like this

backend/model_server/danswer_torch_model.py
yuhongsun96: ah ya, feel free to toss that in!

backend/danswer/secondary_llm_flows/source_filter.py
ghost: true - updated!

backend/model_server/danswer_torch_model.py
ghost: done!

PR #2162, opened 2024-08-18
Performance testing stuff

Weves:
Makes our cloud FAST again.

The main changes:
- Changed ARRAY type to JSONB. For some reason this was causing >5sec queries (especially on initial page load)
- Reduced # of `router.refresh()` -> significantly fewer queries. 4x per endpoint previously, now just 2x on every page load.
- Improve the efficiency of a few endpoints that were particularly slow in the logs

Pushed this to our cloud instance, and it's now much, much faster


backend/alembic/versions/da4c21c69164_chosen_assistants_changed_to_jsonb.py
```diff
@@ -0,0 +1,65 @@
...
+    op.add_column(
+        "user",
+        sa.Column(
+            "chosen_assistants",
+            postgresql.JSONB(astext_type=sa.Text()),
+            nullable=True,
+        ),
+    )
+    for id, chosen_assistants in existing_ids_and_chosen_assistants:
+        conn.execute(
```
Weves, line 38: note: if any of these fail, the whole transaction rolls back and we don't lose the values in the columns

backend/danswer/db/models.py
```diff
@@ -120,7 +120,7 @@ class User(SQLAlchemyBaseUserTableUUID, Base):
     # if specified, controls the assistants that are shown to the user + their order
     # if not specified, all assistants are shown
     chosen_assistants: Mapped[list[int]] = mapped_column(
-        postgresql.ARRAY(Integer), nullable=True
+        postgresql.JSONB(), nullable=True
```
Weves, line 123: JSONB doesn't cause incredible slowness like ARRAY does for some reason 🤔

backend/danswer/db/persona.py
```diff
@@ -200,7 +202,16 @@ def get_personas(
     if not include_deleted:
         stmt = stmt.where(Persona.deleted.is_(False))
 
-    return db_session.scalars(stmt).all()
+    if joinedload_all:
```
Weves, line 205: most of the time we access all of these anyways, so just load them all in one query

backend/danswer/db/index_attempt.py
```diff
@@ -184,37 +182,19 @@ def get_last_attempt(
...
 
 def get_latest_index_attempts(
-    connector_credential_pair_identifiers: list[ConnectorCredentialPairIdentifier],
     secondary_index: bool,
     db_session: Session,
 ) -> Sequence[IndexAttempt]:
     ids_stmt = select(
         IndexAttempt.connector_credential_pair_id,
-        func.max(IndexAttempt.time_created).label("max_time_created"),
+        func.max(IndexAttempt.id).label("max_id"),
```
Weves, line 190: no strong reason to do this, but it might be more performance 🤷

backend/danswer/db/index_attempt.py
```diff
@@ -184,37 +182,19 @@ def get_last_attempt(
...
     ).join(EmbeddingModel, IndexAttempt.embedding_model_id == EmbeddingModel.id)
 
     if secondary_index:
         ids_stmt = ids_stmt.where(EmbeddingModel.status == IndexModelStatus.FUTURE)
     else:
         ids_stmt = ids_stmt.where(EmbeddingModel.status == IndexModelStatus.PRESENT)
 
-    where_stmts: list[ColumnElement] = []
-    for connector_credential_pair_identifier in connector_credential_pair_identifiers:
```
Weves, line 202: this was not necessary - we always fetch all of them anyways. Just added complexity / slowness

web/src/lib/chat/fetchSomeChatData.ts
```diff
@@ -0,0 +1,236 @@
...
+  | "assistants"
+  | "tags"
+  | "llmProviders"
+  | "folders"
+  | "userInputPrompts";
+
+/* 
+NOTE: currently unused, but leaving here for future use. 
+*/
+export async function fetchSomeChatData(
```
Weves, line 65: Not used, but in the future could be used for pages like AssistantGallery which don't need everything

backend/danswer/db/chat.py
```diff
@@ -130,6 +131,9 @@ def get_chat_sessions_by_user(
     if deleted is not None:
         stmt = stmt.where(ChatSession.deleted == deleted)
 
+    if limit:
```
Weves, line 134: getting all chat sessions got really slow for someone that had been using for a long time. 

In the future, we need to add pagination on the FE, but this is likely fine for now

hagen-danswer (approved):
Thanks for leaving comments, made it easy to follow what was important

backend/danswer/db/engine.py
```diff
@@ -65,6 +66,27 @@ def after_cursor_execute(  # type: ignore
             )
 
 
+if LOG_POSTGRES_CONN_COUNTS:
```
rkuo-danswer, line 69: Nit: it should be possible to get the checked in and checked out counts directly from sqlalchemy (probably within a single listener function for both events) and skip tracking this state.

PR #2166, opened 2024-08-19
Curator role

hagen-danswer:
## Description
This role is available with our EE version. This is a group specific admin which can edit objects associated with the group that they are curating


backend/alembic/versions/351faebd379d_add_curator_fields.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def upgrade() -> None:
+    # Add is_curator column to User__UserGroup table
+    op.add_column(
+        "user__user_group",
+        sa.Column("is_curator", sa.Boolean(), nullable=False, server_default="false"),
+    )
+
+    # Use batch mode to modify the enum type
```
hagen-danswer, line 25: Add Curator and Global Curator to userrole enum

backend/alembic/versions/351faebd379d_add_curator_fields.py
```diff
@@ -0,0 +1,90 @@
...
+                "GLOBAL_CURATOR",
+                name="userrole",
+                native_enum=False,
+            ),
+            existing_type=sa.Enum("BASIC", "ADMIN", name="userrole", native_enum=False),
+            existing_nullable=False,
+        )
+    # Create the association table
+    op.create_table(
+        "credential__user_group",
```
hagen-danswer, line 42: Associate a credential with a user group to facilitate curator permissions

backend/alembic/versions/351faebd379d_add_curator_fields.py
```diff
@@ -0,0 +1,90 @@
...
+            ["credential.id"],
+        ),
+        sa.ForeignKeyConstraint(
+            ["user_group_id"],
+            ["user_group.id"],
+        ),
+        sa.PrimaryKeyConstraint("credential_id", "user_group_id"),
+    )
+    op.add_column(
+        "credential",
```
hagen-danswer, line 56: added "curator_public" to control whether all curators can see a credential

backend/danswer/auth/users.py
```diff
@@ -389,6 +389,28 @@ async def current_user(
     return await double_check_user(user)
 
 
+async def current_curator_or_admin_user(
```
hagen-danswer, line 415: This is like the current_admin_user except it also allows curator/global curator

backend/danswer/danswerbot/slack/handlers/handle_regular_answer.py
hagen-danswer, line 1: For this (and everywhere there is a get_editable), there is a logic that curators can see/use some objects but cant edit. 

So requests need to indicate if they are for the purposes of editing or not (implying its for viewing/use)

backend/danswer/db/connector_credential_pair.py
hagen-danswer, line 1: In this file (and all others where there is an _add_user_filters() function) there are comments explaining what the filter conditions are.

bringing _add_user_filters()  out into a util file for all similar api calls was considered, however the filter conditions for each object call may differ slightly. There is an argument that maybe the function could just be the intersection of all the filters and the extra stuff could be added where the call is made. but then the id variables would also have to be made generic (confirming there is a object.user_id).

backend/danswer/db/credentials.py
```diff
@@ -153,13 +205,34 @@ def create_credential(
...
+        credential_user_groups.append(
+            Credential__UserGroup(
+                credential_id=credential.id,
+                user_group_id=group_id,
+            )
+        )
+    db_session.add_all(credential_user_groups)
     db_session.commit()
 
     return credential
```
hagen-danswer, line 255: here is Credential__UserGroup cleanup if credentials are deleted

backend/danswer/db/credentials.py
```diff
@@ -153,19 +221,38 @@ def create_credential(
         admin_public=credential_data.admin_public,
         source=credential_data.source,
         name=credential_data.name,
+        curator_public=credential_data.curator_public,
     )
     db_session.add(credential)
+    db_session.flush()  # This ensures the credential gets an ID
+
```
hagen-danswer, line 244: Credentials are related to groups on creation.
There is currently no way to add/edit/remove relationships after the credential is created.

backend/danswer/db/models.py
hagen-danswer, line 1: see  the alembic upgrade file

backend/danswer/db/users.py
```diff
@@ -1,21 +1,41 @@
...
 from sqlalchemy.orm import Session
-from sqlalchemy.schema import Column
 
 from danswer.db.models import User
+from danswer.db.models import User__UserGroup
+from danswer.db.models import UserRole
 
 
-def list_users(db_session: Session, q: str = "") -> Sequence[User]:
+def list_users(
```
hagen-danswer, line 12: These changes are for future extensibility but are not currently used because the users tab isnt accessible by curators

backend/danswer/server/documents/models.py
```diff
@@ -45,8 +45,9 @@ class ConnectorBase(BaseModel):
     indexing_start: datetime | None
```
hagen-danswer, line 47: this is an example of a well scoped refactor of something that didnt make any sense

backend/danswer/server/documents/credential.py
hagen-danswer, line 1: For this file (and other server files, the error message should adequately explain whats going on

backend/danswer/server/documents/connector.py
```diff
@@ -488,28 +501,63 @@ def _validate_connector_allowed(source: DocumentSource) -> None:
     )
```
hagen-danswer, line 504: This check would be annoying to perform correctly.

The only check that is missing is this doesnt check if the groups are curated by the user, however this check is already performed at the cc_pair creation. it shouldnt even be possible to do this incorrectly because the user is only presented group options from the groups they curate and level 

so performing it here would just prevent a connector with no cc pair from being created but we are gonna roll them together soon anyways so 🤷

backend/danswer/server/documents/models.py
hagen-danswer: also lower in the file as well

backend/ee/danswer/db/user_group.py
```diff
@@ -211,6 +236,64 @@ def _mark_user_group__cc_pair_relationships_outdated__no_commit(
         user_group__cc_pair_relationship.is_current = False
```
hagen-danswer, line 239: A curator is defined as a user that has 1+ User->UserGroup relationships where curator==True

backend/ee/danswer/db/token_limit.py
hagen-danswer, line 1: these are view only for curators

backend/ee/danswer/db/user_group.py
```diff
@@ -179,16 +188,32 @@ def insert_user_group(db_session: Session, user_group: UserGroupCreate) -> UserG
...
+    where_clause = User__UserGroup.user_group_id == user_group_id
+    if user_ids:
+        where_clause &= User__UserGroup.user_id.in_(user_ids)
+
     user__user_group_relationships = db_session.scalars(
-        select(User__UserGroup).where(User__UserGroup.user_group_id == user_group_id)
+        select(User__UserGroup).where(where_clause)
     ).all()
     for user__user_group_relationship in user__user_group_relationships:
         db_session.delete(user__user_group_relationship)
```
hagen-danswer, line 207: cleans up relationship if usergroup is deleted

web/src/components/modals/GenericConfirmModal.tsx
hagen-danswer, line 1: New generic modal that we should use elsewhere where stuff is deleted or other big changes are made

web/src/components/IsPublicGroupSelector.tsx
hagen-danswer, line 1: This component presents a checkbox to change if something is public or not. 
if not, a group selector is displayed
if the user is a curator, it only shows the group selector

web/src/app/ee/admin/groups/page.tsx
```diff
@@ -32,6 +34,24 @@ const Main = () => {
...
+          setCurrentUser(user);
+        } else {
+          console.error("Failed to fetch current user");
+        }
+      } catch (error) {
+        console.error("Error fetching current user:", error);
+      }
+    };
+    fetchCurrentUser();
+  }, []);
```
pablonyx, line 53: Seeing a lot of these throughout the code - would definitely abstract away into a hook unless you're explicitly using `setCurrentUser` - even then, could be passed by the hook. Checkout `lib/hooks.ts` for some inspiration!

backend/danswer/auth/schemas.py
```diff
@@ -7,6 +7,8 @@
 class UserRole(str, Enum):
     BASIC = "basic"
     ADMIN = "admin"
+    CURATOR = "curator"
```
pablonyx, line 20: Might be worth leaving a comment in the code explaining the distinction between the two here, since all future uses / people that try and use UserRole._ will refer back to here.

backend/danswer/auth/users.py
```diff
@@ -395,14 +395,36 @@ async def current_user(
...
+    user: User | None = Depends(current_user),
+) -> User | None:
+    if DISABLE_AUTH:
+        return None
+
+    is_global_curator = (
```
pablonyx: nit: would prefer a slightly cleaner interface here - something like 

```
    if not user or not hasattr(user, "role"):
        raise HTTPException(
            status_code=status.HTTP_403_FORBIDDEN,
            detail="Access denied. User is not authenticated or lacks role information.",
        )

    allowed_roles = {UserRole.GLOBAL_CURATOR, UserRole.CURATOR, UserRole.ADMIN}
    if user.role not in allowed_roles:
        raise HTTPException(
            status_code=status.HTTP_403_FORBIDDEN,
            detail="Access denied. User is not a curator or admin.",
        )
```

backend/danswer/db/users.py
```diff
@@ -1,21 +1,41 @@
...
+    db_session: Session, q: str = "", user: User | None = None
```
pablonyx: Nit: would either add a comment or make `q` something substantially more verbose

backend/danswer/db/users.py
```diff
@@ -1,21 +1,41 @@
...
+    return db_session.scalars(stmt).unique().all()
 
 
 def get_user_by_email(email: str, db_session: Session) -> User | None:
     user = db_session.query(User).filter(User.email == email).first()  # type: ignore
 
     return user
+
+
+def fetch_user_by_id(db_session: Session, user_id: UUID) -> User | None:
```
pablonyx, line 38: Nit: would prefer standardized naming for `(fetch / get) by`

web/src/app/admin/assistants/PersonaTable.tsx
```diff
@@ -116,28 +150,30 @@ export function PersonasTable({ personas }: { personas: Persona[] }) {
...
+                      }
+                    );
+                    if (response.ok) {
+                      router.refresh();
+                    } else {
+                      setPopup({
+                        type: "error",
+                        message: `Failed to update persona - ${await response.text()}`,
+                      });
                     }
```
pablonyx, line 181: Generally prefer to place API calls outside of the JSX
Looks like this was already being done in the component but a good chance to extract it.

web/src/app/admin/connectors/[connector]/pages/Create.tsx
```diff
@@ -292,18 +326,82 @@ const DynamicConnectionForm: React.FC<DynamicConnectionFormProps> = ({
...
+                                      key={userGroup.id}
+                                      className={`
+                                    px-3 
+                                    py-1
+                                    rounded-lg 
+                                    border
+                                    border-border 
+                                    w-fit 
+                                    flex 
+                                    cursor-pointer
```
Comment (pablonyx) [inline web/src/app/admin/connectors/[connector]/pages/Create.tsx]: Super nit: formatting

web/src/app/admin/documents/sets/page.tsx
```diff
@@ -79,15 +98,35 @@ interface DocumentFeedbackTableProps {
...
 }
 
 const DocumentSetTable = ({
   documentSets,
+  editableDocumentSets,
   refresh,
+  refreshEditable,
   setPopup,
-}: DocumentFeedbackTableProps) => {
+}: DocumentFeedbackTableProps & { editableDocumentSets: DocumentSet[] }) => {
```
pablonyx: Would place `editableDocumentSets: DocumentSet[]` inside the props definition

web/src/components/credentials/actions/CreateCredential.tsx
```diff
@@ -14,10 +14,49 @@ import {
...
+        <PlusCircleIcon size={16} className="text-indigo-100" />
+        Create
+      </div>
+    </Button>
+  </div>
+);
+
+type formType = IsPublicGroupSelectorFormType & {
+  name: string;
+  [key: string]: any; // For additional credential fields
```
pablonyx, line 58: `[key: string]: any` reduces type safety substantially - may be only way for this approach, but preferrably we don't let any additional keys with any values be set

backend/danswer/db/users.py
hagen-danswer: making a ticket for this :P our use of get/post/patch/etc. needs to be standardized as well

web/src/components/credentials/actions/ModifyCredential.tsx
```diff
@@ -66,6 +65,10 @@ const CredentialSelectionTable = ({
               const selected = currentCredentialId
                 ? credential.id == (selectedCredentialId || currentCredentialId)
                 : false;
+              const editable = editableCredentials.some(
+                (editableCredential) => editableCredential.id === credential.id
+              );
+              console.log(editable);
```
pablonyx: Probably should remove

Weves (approved):

PR #2172, opened 2024-08-19
add kombu message cleanup task

rkuo-danswer:
## Description
Fixes DAN-430. kombu_message table was expanding without bound, causing certain queries to be excessively slow.

## How Has This Been Tested?

## Accepted Risk

## Related Issue(s)

## Checklist:

### DIFF
--- a/backend/danswer/background/celery/celery_app.py
+++ b/backend/danswer/background/celery/celery_app.py
@@ -400,6 +400,12 @@ def kombu_message_cleanup_task_helper(ctx: dict, db_session: Session) -> bool:
             )
             if result.rowcount > 0:  # type: ignore
                 ctx["deleted"] += 1
+        else:
+            task_name = payload["headers"]["task"]
+            logger.warning(
+                f"Message found for task older than {ctx['cleanup_age']} days. "
+                f"id={task_id} name={task_name}"
+            )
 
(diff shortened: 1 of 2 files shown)


backend/danswer/background/celery/celery_app.py
```diff
@@ -291,6 +297,115 @@ def check_for_cc_pair_deletion_task() -> None:
...
+        task_id = payload["headers"]["id"]
+
+        # Check if task_id exists in celery_taskmeta
+        task_exists = db_session.execute(
+            text("SELECT 1 FROM celery_taskmeta WHERE task_id = :task_id"),
+            {"task_id": task_id},
+        ).fetchone()
+
+        # If task_id does not exist, delete the message
+        if not task_exists:
```
yuhongsun96, line 396: Would add a logger.warning statement if the task did exist. I don't think we anticipate tasks to run for over a week at least right now.

yuhongsun96 (approved):
lgtm

backend/danswer/background/celery/celery_app.py
rkuo-danswer: I'll add this. There's an edge case where if the user shuts down the app and starts it up again after 7 days or more, we might see a few of these task warnings in the logs.

PR #2182, opened 2024-08-20
Disable oidc_expiry by default

Weves:
Testing: 👀


pablodanswer:
Have not tested yet but implementation seems good - would add environment variables to the Docker files though

Weves:
@pablodanswer great callout, just added!

PR #2184, opened 2024-08-20
normalize emails on bulk invite, normalize/lowercase emails on invite…

rkuo-danswer:
… matching

## Description
Fixes DAN-432.  I did end up normalizing emails when inserting via bulk invite as the email_validator docs strongly suggest to follow this pattern.

## How Has This Been Tested?
Ran bulk invite, ran tests against upper and lowercase signup.

## Accepted Risk

## Related Issue(s)

## Checklist:

### DIFF
--- a/backend/danswer/auth/users.py
+++ b/backend/danswer/auth/users.py
@@ -8,6 +8,8 @@ from email.mime.text import MIMEText
 from typing import Optional
 from typing import Tuple
 
+from email_validator import EmailNotValidError
+from email_validator import validate_email
 from fastapi import APIRouter
 from fastapi import Depends
 from fastapi import HTTPException
@@ -61,7 +63,6 @@ from danswer.utils.telemetry import optional_telemetry
 from danswer.utils.telemetry import RecordType
 from danswer.utils.variable_functionality import fetch_versioned_implementation
 
-
 logger = setup_logger()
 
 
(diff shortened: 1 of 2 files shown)


backend/danswer/server/manage/users.py
```diff
@@ -159,12 +160,18 @@ def bulk_invite_users(
...
         raise HTTPException(
             status_code=400, detail="Auth is disabled, cannot invite users"
         )
 
-    all_emails = list(set(emails) | set(get_invited_users()))
+    normalized_emails = []
+    for email in emails:
+        email_info = validate_email(email)  # can raise EmailNotValidError
+        normalized_emails.append(email_info.normalized)  # type: ignore
+    all_emails = list(set(normalized_emails) | set(get_invited_users()))
```
yuhongsun96, line 174: any potential issue with having duplicates (like a normalized and a non normalized version) in the DB?

yuhongsun96 (approved):

backend/danswer/auth/users.py
```diff
@@ -8,6 +8,8 @@
 from typing import Optional
 from typing import Tuple
 
+from email_validator import EmailNotValidError
```
Weves, line 11: is there a reason why we use `from pydantic import validate_email` below but `from email_validator import validate_email` here?

rkuo-danswer: good catch, thanks!

backend/danswer/server/manage/users.py
rkuo-danswer: Before the code would have allowed anything into the database, so this is a strict improvement.

I don't see any issues with this on the read end of things.  At worst the email is in the database in more than one form, but we'll still only pass emails that match.

Moving forward we could add some code to renormalize existing emails, but I think we should get this change out first.

PR #2190, opened 2024-08-21
Account for edge case in indexing times with connectors

pablodanswer:
## Description

## How Has This Been Tested?

## Accepted Risk

## Related Issue(s)

## Checklist:

### DIFF
--- a/backend/danswer/db/connector_credential_pair.py
+++ b/backend/danswer/db/connector_credential_pair.py
@@ -159,6 +159,7 @@ def get_connector_credential_pair_from_id(
 def get_last_successful_attempt_time(
     connector_id: int,
     credential_id: int,
+    earliest_index: float,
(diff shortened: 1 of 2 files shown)


backend/danswer/db/connector_credential_pair.py
```diff
@@ -72,6 +72,7 @@ def get_last_successful_attempt_time(
     credential_id: int,
     embedding_model: EmbeddingModel,
     db_session: Session,
+    earliest_index: float,
```
yuhongsun96, line 162: Slight nit: the most useful/frequently changed args first, least frequently changed (or most predictable or unimportant for the reader) last. db_session is always the same so throw it at the end.

Caller should generally try to follow same arg order as the function def so it's less likely for someone to miss something if not args are required etc.

PR #2192, opened 2024-08-21
Zendesk tickets

jwjordan:
## Description
I updated the Zendesk connector to support a new required dropdown selection between "articles" and "tickets" with "articles" as the default for backward compatibility. If you already have a Zendesk connector set up, it will continue to work normally. If you set up a new one, you can choose "tickets" to have that connector sync all non-deleted Zendesk tickets. Therefore, you have the ability to create two separate Zendesk connectors using the same shared credentials, one for articles and one for tickets, which I felt would be useful in the creation of different Document Sets down the line (If you want both in the same Document Set, you can, but you can also easily separate knowledge of articles and tickets in the case where you want a Sales Assistant that only trains on articles, but a Support Assistant that trains on both articles and tickets).

## How Has This Been Tested?
I tested this against my own Zendesk account by setting up the original Zendesk connection for articles, making code changes, and verifying that I can sync tickets properly. I then queried against known ticket contents to verify that Chat and Search work, and also verified that the citation links properly open the Agent view of the Zendesk ticket in a new browser tab. Finally I tested deletion of the new "tickets" connector and verified it had no effect on the pre-existing "articles" connector.

## Accepted Risk
None that I'm aware of.

## Related Issue(s)
I don't believe this feature request was recorded as an issue.

## Checklist:

### DIFF
--- a/backend/danswer/connectors/zendesk/connector.py
+++ b/backend/danswer/connectors/zendesk/connector.py
@@ -3,7 +3,7 @@ from typing import Any
 import requests
 from retry import retry
 from zenpy import Zenpy  # type: ignore
-from zenpy.lib.api_objects import Ticket
+from zenpy.lib.api_objects import Ticket  # type: ignore
 from zenpy.lib.api_objects.help_centre_objects import Article  # type: ignore
 
 from danswer.configs.app_configs import INDEX_BATCH_SIZE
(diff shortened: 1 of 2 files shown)


jwjordan:
@yuhongsun96 Any chance you'd be able to review this PR? Thanks!

PR #2205, opened 2024-08-22
Update embedding interface

pablodanswer:
## Description
Modify embedding interface to be provider ID agnostic and use basic enums for identification instead (OpenAI, etc).

This involves migrating from a provider-based system, where each provider has
- `id`
- `default_model_id`
- `name` (better described as provider

To simply
- `provider_type`

## How Has This Been Tested?
On this branch
1. Creating various embedding providers
2. Switching between them
3. Embedding 
4. Using chat flows (both cloud + self-hosted)

Upgrade/downgrade tests
1. Creating embedding providers on main
2. Upgrading to this branch - testing + creating + switching
3. Downgrading to main - testing + creating + switching

### DIFF
--- a/backend/danswer/db/embedding_model.py
+++ b/backend/danswer/db/embedding_model.py
@@ -53,8 +53,8 @@ def create_embedding_model(
 def get_embedding_provider_from_provider(
     db_session: Session, cloud_provider_type: EmbeddingProvider
 ) -> CloudEmbeddingProvider | None:
-    query = select(CloudEmbeddingProvider).where(
-        CloudEmbeddingProvider.provider_type == cloud_provider_type
+    query = select(cloud_provider_type).where(
+        CloudEmbeddingProvider.cloud_provider_type == cloud_provider_type
     )
     provider = db_session.execute(query).scalars().first()
     return provider if provider else None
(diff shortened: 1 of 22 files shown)


backend/danswer/db/models.py
```diff
@@ -556,13 +556,14 @@ class EmbeddingModel(Base):
     index_name: Mapped[str] = mapped_column(String)
 
     # New field for cloud provider relationship
-    cloud_provider_id: Mapped[int | None] = mapped_column(
-        ForeignKey("embedding_provider.id")
+    provider_type: Mapped[EmbeddingProvider | None] = mapped_column(
+        ForeignKey("embedding_provider.provider_type")
```
yuhongsun96: should this be nullable?

backend/danswer/db/embedding_model.py
```diff
@@ -48,14 +50,14 @@ def create_embedding_model(
     return embedding_model
 
 
-def get_model_id_from_name(
-    db_session: Session, embedding_provider_name: str
-) -> int | None:
+def get_embedding_provider_from_provider(
```
yuhongsun96: mmm a bit confusing in terms of the name, makes sense when you read the code but ideally it can be clearer?

yuhongsun96 (approved):

backend/danswer/db/models.py
pablonyx: In my opinion, yes! Null for self-hosted models. Can make more explicit by adding a `nullable`

PR #2219, opened 2024-08-23
Add better logging for connectors

Weves:
Wraps all calls to connectors to catch the exception and log the state of all local vars at the time of the exception


backend/danswer/background/indexing/run_indexing.py
```diff
@@ -77,31 +74,9 @@ def _get_document_generator(
...
         raise e
 
-    if task == InputType.LOAD_STATE:
-        assert isinstance(runnable_connector, LoadConnector)
-        doc_batch_generator = runnable_connector.load_from_state()
-
-    elif task == InputType.POLL:
-        assert isinstance(runnable_connector, PollConnector)
-        if (
-            attempt.connector_credential_pair.connector_id is None
```
rkuo-danswer, line 87: Interestingly there's a bug in the old code here!

rkuo-danswer (approved):

PR #2226, opened 2024-08-23
first cut at redis

rkuo-danswer:
## Description
Fixes DAN-466.

This is a big change behind the scenes, but a no-op to end users other than making celery use redis.  There may be things we need to update customers and devs on in terms of custom deployments and local setup, etc and further iteration in those areas.

Internally we've observed that many celery features simply silently fail or don't work without redis as backend, forcing workarounds and hacks. In addition, being able to use redis for general caching and message queue like functionality will be useful.

## How Has This Been Tested?

## Accepted Risk

## Related Issue(s)

## Checklist:


deployment/docker_compose/docker-compose.dev.yml
```diff
@@ -328,9 +332,19 @@ services:
...
+    image: redis:7.4-alpine
+    restart: always
+    ports:
+      - '6379:6379'
+    command: redis-server
+    volumes: 
+      - cache:/data
+
 volumes:
+  cache:
```
Weves: prefer to name this `cache_volume` to match the other volumnes

Weves: and obviously same for all the other compose files

deployment/kubernetes/secrets.yaml
```diff
@@ -8,4 +8,6 @@ data:
   postgres_user: cG9zdGdyZXM= # "postgres" base64 encoded
   postgres_password: cGFzc3dvcmQ= # "password" base64 encoded
   google_oauth_client_id: ZXhhbXBsZS1jbGllbnQtaWQ= # "example-client-id" base64 encoded. You will need to provide this, use echo -n "your-client-id" | base64
-  google_oauth_client_secret: example_google_oauth_secret # "example-client-secret" base64 encoded. You will need to provide this, use echo -n "your-client-id" | base64
+  google_oauth_client_secret: ZXhhbXBsZV9nb29nbGVfb2F1dGhfc2VjcmV0 # "example-client-secret" base64 encoded. You will need to provide this, use echo -n "your-client-id" | base64
+  redis_password: cGFzc3dvcmQ= # "password" base64 encoded
```
Weves, line 12: I believe we need to add this as an env variable explicitly to the background / api server pods?

Weves (changes requested):
As discussed, need to add similar support for the Helm chart setup 👀

Weves (approved):

PR #2232, opened 2024-08-24
Cleaner + cleaner assistants creation flow etc.

pablodanswer:
## Description
Standardizes components across Danswer
- Assistant Creation Form Flow + Copy Updated
- Allow Assistant Editing / deleting in "My assistants page"
- Input fields + form updated across the board
- Input components unified for dynamic connection form and Formik


web/src/components/modals/DeleteEntityModal.tsx
```diff
@@ -2,24 +2,26 @@ import { FiTrash, FiX } from "react-icons/fi";
...
-export const DeleteChatModal = ({
-  chatSessionName,
+export const DeleteEntityModal = ({
   onClose,
   onSubmit,
+  entity,
+  entityName,
 }: {
-  chatSessionName: string;
+  entity: string;
```
rkuo-danswer: good idea to make this reusable, but what's the difference between entity and entity name?

web/src/components/admin/connectors/Field.tsx
```diff
@@ -353,26 +390,31 @@ export const BooleanFormField = ({
   subtext,
   onChange,
   noPadding,
+  verticalPadding,
```
rkuo-danswer: This looks like BooleanFormField is becoming highly overloaded.  Could it make sense to keep BooleanFormField as simple as possible and create a more complicated/feature rich subclass over it?

web/src/components/admin/connectors/Field.tsx
```diff
@@ -100,8 +103,11 @@ export function TextFormField({
   label,
   subtext,
   placeholder,
+  value,
```
rkuo-danswer, line 106: Might be worthwhile to consider consolidating all these parameters into a config object or create a subclass for a more highly customized version of this component.

web/src/app/assistants/mine/AssistantsList.tsx
```diff
@@ -158,6 +174,92 @@ function AssistantListItem({
...
+                          });
+                          return;
+                        }
+                        const success = await removeAssistantFromList(
+                          assistant.id,
+                          currentChosenAssistants || allAssistantIds
+                        );
+                        if (success) {
+                          setPopup({
+                            message: `"${assistant.name}" has been removed from your list.`,
```
rkuo-danswer, line 210: i can't quite see what the UI is doing here until we review, but it reads as if a popup appears when adding an assistant. Sounds like it could be a bit heavyweight messaging for this operation?

web/src/app/admin/embeddings/pages/AdvancedEmbeddingFormPage.tsx
```diff
@@ -131,54 +126,40 @@ const AdvancedEmbeddingFormPage = forwardRef<
...
-                    setRerankingDetails({ ...values, num_rerank: value });
-                    setFieldValue("num_rerank", value);
-                  }}
-                  setFieldValue={setFieldValue}
-                  type="number"
-                  label="Number of Results to Rerank"
-                  name="num_rerank"
-                />
-              </div>
+              <BooleanFormField
```
rkuo-danswer, line 129: It looks like a lot of this was there before, but for the sake of thoroughness, I'm a little unclear from a user interface perspective if the user is really supposed to understand what these settings do. It feels like a ton of explanation is required.

web/src/app/admin/connectors/[connector]/pages/ConnectorInput/NumberInput.tsx
```diff
@@ -0,0 +1,42 @@
+import { SubLabel } from "@/components/admin/connectors/Field";
+import { Field } from "formik";
+
+export default function NumberInput({
+  label,
+  value,
+  optional,
+  description,
+  name,
+  showNever,
```
Comment (rkuo-danswer) [inline web/src/app/admin/connectors/[connector]/pages/ConnectorInput/NumberInput.tsx]: I'm not entirely clear on why showNever is a part of a Number Input?  Also thinking generally if we want to reuse components, we should TSDoc / JSDoc them.

web/src/components/modals/DeleteEntityModal.tsx
pablonyx: Updated wording to make this clearer!

web/src/components/admin/connectors/Field.tsx
pablonyx: Agree- this is part of a broader effort to make our components more standardized / best practice when it comes to layout specifically.

web/src/app/assistants/mine/AssistantsList.tsx
pablonyx: Without another notification message type here, I think a popup (really it's a toast) should be suitable for now - but agree re: having more method of communicating events in-app.

web/src/app/admin/embeddings/pages/AdvancedEmbeddingFormPage.tsx
pablonyx: I agree- for BooleanFormField specifically the indent (in this case `noPadding`) is necessary for a few elements in the UI, but I've updated now to be a bit clearer

web/src/app/admin/connectors/[connector]/pages/ConnectorInput/NumberInput.tsx
Comment (pablonyx) [reply web/src/app/admin/connectors/[connector]/pages/ConnectorInput/NumberInput.tsx]: `showNever` is meant to show "never" if 0. Agree that this could be not verbose enough

rkuo-danswer (approved):

PR #2234, opened 2024-08-25
Combined the get document set endpoints

hagen-danswer:
## Description
Combined the admin and non admin endpoints for retrieving document sets and just applied the filters based on the users permissions

### DIFF
--- a/backend/danswer/server/features/document_set/api.py
+++ b/backend/danswer/server/features/document_set/api.py
@@ -9,16 +9,12 @@ from danswer.auth.users import current_user
 from danswer.auth.users import validate_curator_request
 from danswer.db.document_set import check_document_sets_are_public
 from danswer.db.document_set import fetch_all_document_sets_for_user
-from danswer.db.document_set import fetch_user_document_sets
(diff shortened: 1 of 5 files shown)


backend/danswer/server/features/document_set/api.py
```diff
@@ -88,7 +84,10 @@ def delete_document_set(
         raise HTTPException(status_code=400, detail=str(e))
 
 
-@router.get("/admin/document-set")
+"""Endpoints for non-admins"""
+
+
+@router.get("/document-set")
 def list_document_sets_admin(
```
Weves: this name should probably be changed

backend/danswer/server/features/document_set/api.py
```diff
@@ -88,7 +84,10 @@ def delete_document_set(
...
     user: User | None = Depends(current_curator_or_admin_user),
```
Weves: this would make it so we can't list document sets if you're a regular user? 🧐

Weves (changes requested):

Weves (approved):

PR #2235, opened 2024-08-25
Db search

pablodanswer:
## Description
Migrate from `Embedding Model` and `KV_STORE` to `Search Settings`.

New flows
1. Update search settings without re-indexing:  just update current Search Settings
2. Update search settings with value(s) that require re-indexing: update current Search Settings  (excl. multi-pass values + create new search settings configuration and begin embedding model swap)

Tested with
- [x] Upgrading from main with cloud + self-hosted models
- [x] Update search settings with values which require re-indexing
- [x] Update search stings with value that don't require re-indexing
- [x] Downgrade after swapping indexing


backend/danswer/background/update.py
```diff
@@ -60,7 +61,7 @@
 def _should_create_new_indexing(
     cc_pair: ConnectorCredentialPair,
     last_index: IndexAttempt | None,
-    model: EmbeddingModel,
+    model: SearchSettings,
```
rkuo-danswer: Should this param still be "model"?

backend/danswer/background/update.py
```diff
@@ -160,18 +161,18 @@ def create_indexing_jobs(existing_jobs: dict[int, Future | SimpleJob]) -> None:
...
-                    attempt.embedding_model_id,
+                    attempt.search_settings_id,
                 )
             )
 
-        embedding_models = [get_current_db_embedding_model(db_session)]
-        secondary_embedding_model = get_secondary_db_embedding_model(db_session)
-        if secondary_embedding_model is not None:
-            embedding_models.append(secondary_embedding_model)
+        search_settings = [get_current_search_settings(db_session)]
```
rkuo-danswer: Feels like get_search_settings with an enum passed in, instead of two separate functions, might be cleaner.

web/src/app/admin/embeddings/interfaces.ts
```diff
@@ -9,19 +12,32 @@ export enum RerankerProvider {
   COHERE = "cohere",
 }
 export interface AdvancedDetails {
-  multilingual_expansion: string[];
+  model_name: string;
```
rkuo-danswer, line 15: This class name "AdvancedDetails" seems too generic.  could mean anything.

rkuo-danswer (approved):

web/src/app/admin/embeddings/interfaces.ts
pablonyx: Definitely agree looking back! Updating across the board to be clearer

backend/danswer/background/update.py
pablonyx: Updated to be clearer! Should be clearer now

PR #2239, opened 2024-08-27
Fix freezing

Weves:
Large files (e.g. 10MB log files) can cause indexing to freeze. Adding a cap to the # of chars a confluence attachment can have to be considered.

Also added a test (which in the future can automatically run at an interval) to test the basics of the confluence connector.

### DIFF
--- a/backend/danswer/configs/app_configs.py
+++ b/backend/danswer/configs/app_configs.py
@@ -212,7 +212,12 @@ CONFLUENCE_CONNECTOR_SKIP_LABEL_INDEXING = (
 
 # Attachments exceeding this size will not be retrieved (in bytes)
 CONFLUENCE_CONNECTOR_ATTACHMENT_SIZE_THRESHOLD = int(
-    os.environ.get("CONFLUENCE_CONNECTOR_ATTACHMENT_SIZE_THRESHOLD", 50 * 1024 * 1024)
+    os.environ.get("CONFLUENCE_CONNECTOR_ATTACHMENT_SIZE_THRESHOLD", 10 * 1024 * 1024)
+)
+# Attachments with more chars than this will not be indexed. This is to prevent extremely
+# large files from freezing indexing. 200,000 is ~100 google doc pages.
+CONFLUENCE_CONNECTOR_ATTACHMENT_CHAR_COUNT_THRESHOLD = int(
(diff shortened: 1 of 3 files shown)


backend/danswer/configs/app_configs.py
```diff
@@ -212,7 +212,12 @@
 
 # Attachments exceeding this size will not be retrieved (in bytes)
 CONFLUENCE_CONNECTOR_ATTACHMENT_SIZE_THRESHOLD = int(
-    os.environ.get("CONFLUENCE_CONNECTOR_ATTACHMENT_SIZE_THRESHOLD", 50 * 1024 * 1024)
+    os.environ.get("CONFLUENCE_CONNECTOR_ATTACHMENT_SIZE_THRESHOLD", 10 * 1024 * 1024)
+)
+# Attachments with more chars than this will not be indexed. This is to prevent extremely
+# large files from freezing indexing. 200,000 is ~100 google doc pages.
+CONFLUENCE_CONNECTOR_ATTACHMENT_CHAR_COUNT_THRESHOLD = int(
+    os.environ.get("CONFLUENCE_CONNECTOR_ATTACHMENT_CHAR_COUNT_THRESHOLD", 200_000)
```
rkuo-danswer, line 220: default value looks like a typo

backend/danswer/connectors/confluence/connector.py
```diff
@@ -548,10 +551,64 @@ def _fetch_labels(self, confluence_client: Confluence, page_id: str) -> list[str
...
+            return None
+
+        response = confluence_client._session.get(download_link)
+        if response.status_code != 200:
+            logger.warning(
+                f"Failed to fetch {download_link} with invalid status code {response.status_code}"
+            )
+            return None
+
+        extracted_text = extract_file_text(
```
rkuo-danswer, line 595: how much overlap is there between the file size threshold and the character count threshold? Is it possible to accomplish the intent with just one of the settings?

rkuo-danswer (changes requested):

rkuo-danswer (approved):

backend/danswer/configs/app_configs.py
rkuo-danswer: TIL I learned underscores in numbers work for readability.

PR #2242, opened 2024-08-27
Detect GPU on startup for default multi-pass indexing value

pablodanswer:
## How Has This Been Tested?

## Accepted Risk

## Related Issue(s)

## Checklist:

### DIFF
--- a/backend/danswer/main.py
+++ b/backend/danswer/main.py
@@ -370,6 +370,7 @@ async def lifespan(app: FastAPI) -> AsyncGenerator:
 
         translate_saved_search_settings(db_session)
 
+        # update multipass indexing setting based on GPU availability
         update_default_multipass_indexing(db_session)
 
         # Does the user need to trigger a reindexing to bring the document index
--- a/backend/danswer/main.py
+++ b/backend/danswer/main.py
@@ -48,6 +48,7 @@ from danswer.db.engine import init_sqlalchemy_engine
 from danswer.db.engine import warm_up_connections
 from danswer.db.index_attempt import cancel_indexing_attempts_past_model
 from danswer.db.index_attempt import expire_index_attempts
+from danswer.db.models import Connector
 from danswer.db.persona import delete_old_default_personas
 from danswer.db.search_settings import get_current_search_settings
 from danswer.db.search_settings import get_secondary_search_settings
@@ -194,6 +195,7 @@ def setup_postgres(db_session: Session) -> None:
 
 def translate_saved_search_settings(db_session: Session) -> None:
     kv_store = get_dynamic_config_store()
+    db_session.query(Connector).count()
 
     try:
         search_settings_dict = kv_store.load(KV_SEARCH_SETTINGS)
--- a/backend/danswer/main.py
+++ b/backend/danswer/main.py
@@ -206,15 +206,15 @@ def update_default_multipass_indexing(db_session: Session) -> None:
         logger.info(f"GPU availability: {gpu_available}")
 
         current_settings = get_current_search_settings(db_session)
-        if current_settings:
-            logger.notice(f"Updating multipass indexing setting to: {gpu_available}")
-            updated_settings = SavedSearchSettings.from_db_model(current_settings)
-            updated_settings.multipass_indexing = gpu_available
-            update_current_search_settings(db_session, updated_settings)
-        else:
-            logger.warning(
-                "No current search settings found. Skipping multipass indexing update."
-            )
+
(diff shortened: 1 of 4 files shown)


backend/danswer/utils/gpu_utils.py
```diff
@@ -0,0 +1,28 @@
...
+
+from danswer.utils.logger import setup_logger
+from shared_configs.configs import MODEL_SERVER_HOST
+from shared_configs.configs import MODEL_SERVER_PORT
+
+logger = setup_logger()
+
+
+def gpu_status_request() -> bool:
+    model_server_url = f"{MODEL_SERVER_HOST}:{MODEL_SERVER_PORT}"
```
yuhongsun96: We will want this to have an argument for indexing: bool or something. We want to check the indexing model server for the multipass indexing.

backend/danswer/utils/gpu_utils.py
```diff
@@ -0,0 +1,28 @@
...
+    response = requests.get(f"{model_server_url}/api/gpu-status")
+
+    if response.status_code == 200:
+        gpu_status = response.json()
+        if gpu_status["gpu_available"]:
+            return True
+        else:
+            return False
+    else:
+        logger.warning(
```
yuhongsun96: probably this should just be a logger.error and not having error in the text 😅

backend/danswer/main.py
```diff
@@ -341,6 +370,9 @@ async def lifespan(app: FastAPI) -> AsyncGenerator:
 
         translate_saved_search_settings(db_session)
 
+        # update multipass indexing setting based on GPU availability
+        update_default_multipass_indexing(db_session)
```
yuhongsun96: I would place this later (after the embedding warm up setup) because that step has backoff and retries. Otherwise this call to check GPU could run before the model server is actually up

backend/danswer/main.py
```diff
@@ -192,6 +193,34 @@ def setup_postgres(db_session: Session) -> None:
...
+            "No existing docs or connectors found. Checking GPU availability for multipass indexing."
+        )
+        gpu_available = gpu_status_request()
+        logger.info(f"GPU availability: {gpu_available}")
+
+        current_settings = get_current_search_settings(db_session)
+        if current_settings:
+            logger.notice(f"Updating multipass indexing setting to: {gpu_available}")
+            updated_settings = SavedSearchSettings.from_db_model(current_settings)
+            updated_settings.multipass_indexing = gpu_available
```
yuhongsun96: should also apply if user has set a cloud embedding model

backend/danswer/main.py
```diff
@@ -192,6 +193,34 @@ def setup_postgres(db_session: Session) -> None:
...
+            update_current_search_settings(db_session, updated_settings)
+        else:
+            logger.warning(
```
yuhongsun96: Should this ever happen?

backend/danswer/main.py
```diff
@@ -192,6 +193,34 @@ def setup_postgres(db_session: Session) -> None:
...
+    logger.debug(f"Docs exist: {docs_exist}, Connectors exist: {connectors_exist}")
+
+    if not docs_exist and not connectors_exist:
+        logger.info(
+            "No existing docs or connectors found. Checking GPU availability for multipass indexing."
+        )
+        gpu_available = gpu_status_request()
+        logger.info(f"GPU availability: {gpu_available}")
+
+        current_settings = get_current_search_settings(db_session)
```
yuhongsun96, line 208: this returns SearchSettings, it shouldn't ever be None

yuhongsun96 (approved):

backend/danswer/main.py
yuhongsun96: note that the call later on checks for the inference server, not indexing

backend/danswer/utils/gpu_utils.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+@retry(tries=5, delay=5)
+def gpu_status_request(indexing: bool = True) -> bool:
+    if indexing:
+        model_server_url = f"{INDEXING_MODEL_SERVER_HOST}:{INDEXING_MODEL_SERVER_PORT}"
+    else:
+        model_server_url = f"{MODEL_SERVER_HOST}:{MODEL_SERVER_PORT}"
+
+    if "http" not in model_server_url:
```
yuhongsun96, line 20: No need to change but if this is now repeated functionality, we can pull it out into a function

PR #2248, opened 2024-08-27
Confluence improvements

Weves:
https://linear.app/danswer/issue/DAN-487/fix-confluence-indexing-permissions-issue

Tested with my 👀

### DIFF
--- a/backend/danswer/configs/app_configs.py
+++ b/backend/danswer/configs/app_configs.py
@@ -201,7 +201,8 @@ CONFLUENCE_CONNECTOR_LABELS_TO_SKIP = [
 
 # Avoid to get archived pages
 CONFLUENCE_CONNECTOR_INDEX_ONLY_ACTIVE_PAGES = (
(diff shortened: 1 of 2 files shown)


backend/danswer/configs/app_configs.py
```diff
@@ -201,7 +201,8 @@
 
 # Avoid to get archived pages
 CONFLUENCE_CONNECTOR_INDEX_ONLY_ACTIVE_PAGES = (
-    os.environ.get("CONFLUENCE_CONNECTOR_INDEX_ONLY_ACTIVE_PAGES", "").lower() == "true"
+    os.environ.get("CONFLUENCE_CONNECTOR_INDEX_ONLY_ACTIVE_PAGES", "true").lower()
```
Weves: Not indexing archived pages feels like the more reasonable default.

yuhongsun96: slightly weird approach no? If it's "" then it's False but if it's nothing, then it's true?

yuhongsun96 (approved):
one comment

PR #2254, opened 2024-08-28
Lowercase slack channels automatically

Weves:
Prevents people from accidentally using uppercase in their specified slack channels -> indexing nothing


pablodanswer:
Tested and works and logic is sound! Wonder if it's worth porting to `lowercase` instead but I can imagine a generalizable transform having more utility.

pablodanswer (approved):

PR #2256, opened 2024-08-28
Integration tests

hagen-danswer:
## How Has This Been Tested?
Yes


backend/danswer/auth/users.py
```diff
@@ -67,23 +67,6 @@
 logger = setup_logger()
```
hagen-danswer, line 69: moved to another file

backend/danswer/db/connector_credential_pair.py
```diff
@@ -334,9 +334,13 @@ def add_credential_to_connector(
...
     if credential is None:
+        error_msg = (
+            f"Credential {credential_id} does not exist or does not belong to user"
+        )
+        logger.error(error_msg)
         raise HTTPException(
             status_code=401,
-            detail="Credential does not exist or does not belong to user",
+            detail=error_msg,
         )
```
hagen-danswer, line 344: need to look into long term fix for bubbling up failed api calls

backend/danswer/server/documents/cc_pair.py
```diff
@@ -157,11 +153,12 @@ def associate_credential_to_connector(
     user: User | None = Depends(current_curator_or_admin_user),
     db_session: Session = Depends(get_session),
 ) -> StatusResponse[int]:
```
hagen-danswer, line 155: used a more robust existing function to check

backend/danswer/server/documents/connector.py
```diff
@@ -514,35 +514,6 @@ def _validate_connector_allowed(source: DocumentSource) -> None:
     )
 
 
-def _check_connector_permissions(
```
hagen-danswer, line 517: used a more robust existing function

backend/danswer/server/documents/models.py
```diff
@@ -48,9 +48,12 @@ class ConnectorBase(BaseModel):
 
 
 class ConnectorUpdateRequest(ConnectorBase):
```
hagen-danswer, line 50: should test this somehow maybe

backend/danswer/server/documents/models.py
```diff
@@ -261,6 +259,10 @@ class ConnectorCredentialPairMetadata(BaseModel):
     groups: list[int] = Field(default_factory=list)
```
hagen-danswer, line 261: ???

backend/tests/integration/tests/permissions/test_cc_pair_permissions.py
```diff
@@ -0,0 +1,165 @@
...
+        source=DocumentSource.CONFLUENCE,
+        groups=[user_group_2.id],
+        curator_public=False,
+        user_performing_action=admin_user,
+    )
+
+    # END OF HAPPY PATH
+
+    """
+    Curators should not be able to:
```
Weves: nit: prefer to leave the comment as to what each of the below cases is testing right next to the actual code e.g.:

```
    """Tests for things Curators should not be able to do"""

    # Curator shouldn't be able to create a public cc pair
    with pytest.raises(HTTPError):
        CCPairManager.create(
            connector_id=connector_1.id,
            credential_id=credential_1.id,
            name="invalid_cc_pair_1",
            groups=[user_group_1.id],
            is_public=True,
            user_performing_action=curator,
        )
```

This makes it so the reader doesn't have to look / scroll up to figure out what's supposed to be going on

backend/tests/integration/tests/permissions/test_cc_pair_permissions.py
```diff
@@ -0,0 +1,165 @@
...
+    - Create a cc pair for a user group they are not a curator of
+    - Create a cc pair for a user group that the connector does not belong to (NOT WORKING)
+    - Create a cc pair for a user group that the credential does not belong to
+    """
+    with pytest.raises(HTTPError):
+        CCPairManager.create(
+            connector_id=connector_1.id,
+            credential_id=credential_1.id,
+            name="invalid_cc_pair_1",
+            groups=[user_group_1.id],
```
Weves, line 94: could we remove this group here? This would make this case simpler remove other reasons why this would fail (e.g. if we later change this to make the user not a curator of `user_group_1` but forget to change this, it would no longer test what it is supposed to test)

backend/tests/integration/tests/permissions/test_cc_pair_permissions.py
Weves: Similar feelings in other places, although this is a bit more of a personal preference thing

backend/tests/integration/common_utils/managers/user.py
```diff
@@ -0,0 +1,146 @@
...
+        result_cookie = next(iter(response.cookies), None)
+
+        if not result_cookie:
+            raise Exception("Failed to login")
+
+        print(f"Logged in as {test_user.email}")
+        return f"{result_cookie.name}={result_cookie.value}"
+
+    @staticmethod
+    def add_api_key_to_user(
```
Weves: I don't believe API keys are really associated with users. There should probably be a separate "type" of user called an "API Key" user (or an APIKeyManager)?

Weves (commented):
👩‍🎨

backend/tests/integration/tests/permissions/test_cc_pair_permissions.py
hagen-danswer: no, then it would fail because curator is only a curator of group 1. if we dont designate user_group_1, it would be invalid because curators have to attach a cc_pair to a group to create a cc_pair

hagen-danswer: That being said i can add a test for explicity trying to create a cc pair without any groups

backend/tests/integration/common_utils/managers/user.py
hagen-danswer: done

Weves (approved):

PR #2260, opened 2024-08-28
Get accurate model output max

pablodanswer:
## Description
Adjusts maximum input size based on the actual context window

## How Has This Been Tested?
- [x] Various model providers, context lengths, documents in input window

### DIFF
--- a/backend/danswer/configs/model_configs.py
+++ b/backend/danswer/configs/model_configs.py
@@ -80,11 +80,13 @@ GEN_AI_API_VERSION = os.environ.get("GEN_AI_API_VERSION") or None
 GEN_AI_LLM_PROVIDER_TYPE = os.environ.get("GEN_AI_LLM_PROVIDER_TYPE") or None
 # Override the auto-detection of LLM max context length
 GEN_AI_MAX_TOKENS = int(os.environ.get("GEN_AI_MAX_TOKENS") or 0) or None
+
 # Set this to be enough for an answer + quotes. Also used for Chat
-GEN_AI_MAX_OUTPUT_TOKENS = int(os.environ.get("GEN_AI_MAX_OUTPUT_TOKENS") or 1024)
+GEN_AI_MAX_OUTPUT_TOKENS = int(os.environ.get("GEN_AI_MAX_OUTPUT_TOKENS")) or None
 
 # Typically, GenAI models nowadays are at least 4K tokens
-GEN_AI_MODEL_DEFAULT_MAX_TOKENS = 4096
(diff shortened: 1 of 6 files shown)


pablodanswer:
I can think of a few options here (where `max_context` is total max context, `input_context` is the maximum input we set aside and `max_output` is the actual maximum output for a model)
1. `input_context` = `max_context` - `max_output` (not feasible)
2. `input_context` = max(`max_context` - `max_output`, `minimum variable set by an environment variable`) - leaving aside at least some space for the output (could also be an environment variable)
3. `input_context` =`max_context` - `max_output` * `multiplier < 1 set by an environment variable`

backend/danswer/llm/chat_llm.py
```diff
@@ -228,6 +238,30 @@ def __init__(
     def log_model_configs(self) -> None:
         logger.debug(f"Config: {self.config}")
 
+    def _calculate_max_tokens(self, prompt: LanguageModelInput) -> int:
```
Weves: nit: renaming to `_calculate_max_output_tokens`. `max_tokens` is a bit ambiguous

backend/danswer/llm/chat_llm.py
```diff
@@ -209,7 +211,15 @@ def __init__(
         self._api_base = api_base
         self._api_version = api_version
         self._custom_llm_provider = custom_llm_provider
-        self._max_output_tokens = max_output_tokens
+        self._max_output_tokens = (
```
Weves: this is not used either, right? If so, can we comment it + add brief comment as to why

backend/danswer/configs/model_configs.py
```diff
@@ -80,11 +80,14 @@
 GEN_AI_LLM_PROVIDER_TYPE = os.environ.get("GEN_AI_LLM_PROVIDER_TYPE") or None
 # Override the auto-detection of LLM max context length
 GEN_AI_MAX_TOKENS = int(os.environ.get("GEN_AI_MAX_TOKENS") or 0) or None
+
 # Set this to be enough for an answer + quotes. Also used for Chat
+# This is the minimum token context we will leave for the LLM to generate an answer
 GEN_AI_MAX_OUTPUT_TOKENS = int(os.environ.get("GEN_AI_MAX_OUTPUT_TOKENS") or 1024)
```
Weves: This should be renamed to `GEN_AI_NUM_RESERVED_OUTPUT_TOKENS` or something similar?

Weves (approved):

PR #2274, opened 2024-08-30
Added frontend logical polish

hagen-danswer:


web/src/app/admin/documents/sets/DocumentSetCreationForm.tsx
```diff
@@ -125,14 +125,12 @@ export const DocumentSetCreationForm = ({
                 placeholder="Describe what the document set represents"
                 autoCompleteDisabled={true}
               />
```
hagen-danswer, line 127: should show the element regardless of situation

web/src/components/IsPublicGroupSelector.tsx
```diff
@@ -76,30 +78,27 @@ export const IsPublicGroupSelector = <T extends IsPublicGroupSelectorFormType>({
             disabled={!isAdmin}
             subtext={
               <span className="block mt-2 text-sm text-gray-500">
```
hagen-danswer, line 84: reworded string to make it more clear

web/src/components/IsPublicGroupSelector.tsx
```diff
@@ -76,30 +78,27 @@ export const IsPublicGroupSelector = <T extends IsPublicGroupSelectorFormType>({
...
-                Group) will have access.
+                If set, then this {objectName} will be usable by{" "}
+                <b>All {publicToWhom}</b>. Otherwise, only <b>{publicToWhom}</b>{" "}
+                who have explicitly been given access to this {objectName} (e.g.
+                via a User Group) and <b>All Admins</b> will have access.
               </span>
             }
           />
         </>
       )}
```
hagen-danswer, line 94: should not show group selector if there are no groups

web/src/components/UserDropdown.tsx
```diff
@@ -16,6 +16,33 @@ import {
 } from "./icons/icons";
 import { pageType } from "@/app/chat/sessionSidebar/types";
 
+interface DropdownOptionProps {
```
hagen-danswer, line 19: turned the dropdown options into a component

web/src/components/UserDropdown.tsx
```diff
@@ -100,44 +127,31 @@ export function UserDropdown({
...
-                >
-                  <LightSettingsIcon className="h-5 w-5 my-auto mr-2" />
-                  Curator Panel
-                </Link>
-              </>
+                  icon={<LightSettingsIcon className="h-5 w-5 my-auto mr-2" />}
+                  label="Curator Panel"
+                />
+              )
             )}
```
hagen-danswer, line 144: fixed logical issue that would cause the border to show when only logout was present 

also allows admins/curators to logout from admin/curator menu

yuhongsun96:
Make Documents Public
If set, then documents indexed by this connector will be visible to all users. If turned off, then only users who explicitly have been given access to the documents (e.g. through a User Group) will have access

Curator Accessible
If set, this Credential will be useable by all Curators. If turned off, only Admins or specific Curators (who been explicitly granted access via User Groups) can use this Credential.

PR #2284, opened 2024-08-31
Fix deletion in the overlapping connector case

Weves:
Wrote test, it failed. Fixed. It worked.

### DIFF
--- a/backend/danswer/background/update.py
+++ b/backend/danswer/background/update.py
@@ -17,6 +17,7 @@ from danswer.configs.app_configs import DASK_JOB_CLIENT_ENABLED
 from danswer.configs.app_configs import DISABLE_INDEX_UPDATE_ON_SWAP
 from danswer.configs.app_configs import NUM_INDEXING_WORKERS
 from danswer.configs.app_configs import NUM_SECONDARY_INDEXING_WORKERS
+from danswer.configs.constants import DocumentSource
 from danswer.configs.constants import POSTGRES_INDEXER_APP_NAME
 from danswer.db.connector import fetch_connectors
 from danswer.db.connector_credential_pair import fetch_connector_credential_pairs
@@ -66,6 +67,10 @@ def _should_create_new_indexing(
     db_session: Session,
 ) -> bool:
     connector = cc_pair.connector
+
+    # don't kick off indexing for `NOT_APPLICABLE`
+    if connector.source == DocumentSource.NOT_APPLICABLE:
+        return False
 
     # User can still manually create single indexing attempts via the UI for the
     # currently in use index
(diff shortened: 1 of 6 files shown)


backend/danswer/db/document_set.py
```diff
@@ -525,36 +525,52 @@ def fetch_document_sets_for_documents(
 ) -> Sequence[tuple[str, list[str]]]:
     """Gives back a list of (document_id, list[document_set_names]) tuples"""
     stmt = (
```
Weves, line 527: Previously, this just wouldn't return a row at all for document that didn't have any document sets associated with them.

backend/danswer/db/document_set.py
```diff
@@ -525,36 +525,52 @@ def fetch_document_sets_for_documents(
...
-            DocumentSet__ConnectorCredentialPair,
-            DocumentSetDBModel.id
-            == DocumentSet__ConnectorCredentialPair.document_set_id,
+        select(
+            Document.id,
+            func.coalesce(
+                func.array_remove(func.array_agg(DocumentSetDBModel.name), None), []
+            ).label("document_set_names"),
         )
-        .join(
```
Weves, line 534: joins are jsut re-ordered by Claude :(

backend/danswer/server/manage/administrative.py
```diff
@@ -166,10 +166,14 @@ def create_deletion_attempt_for_connector_id(
         get_editable=True,
     )
     if cc_pair is None:
+        error = (
```
Weves, line 169: adding more logging (previously would not give any details in the logs)

backend/danswer/background/update.py
```diff
@@ -67,6 +68,10 @@ def _should_create_new_indexing(
 ) -> bool:
     connector = cc_pair.connector
 
+    # don't kick off indexing for `NOT_APPLICABLE`
```
Weves, line 71: this was throwing off tests, should be a no-op in production

backend/danswer/db/document_set.py
hagen-danswer: outerjoin gang

backend/danswer/server/manage/administrative.py
hagen-danswer: I have been doing the same thing. 

However, this is just a consequence of the recent logging rework causing failed API calls to not show any error beyond whatever 400 code. We should probably look into solving that...

hagen-danswer (approved):
lgtm

I will merge so i can rebase/refactor

backend/tests/integration/tests/connector/test_deletion.py
```diff
@@ -188,3 +189,123 @@ def test_connector_deletion(reset: None, vespa_client: TestVespaClient) -> None:
...
+        if c1_details.connector_id not in [c["id"] for c in connectors]:
+            break
+
+        if time.time() - start > MAX_DELAY:
+            raise TimeoutError("Connector 1 was not deleted within the max delay")
+
+        time.sleep(2)
+
+    print("Connector 1 deleted")
+
```
hagen-danswer, line 292: check that only connector 2 is deleted?

PR #2292, opened 2024-09-02
Ordered indexing status

pablodanswer:
## Description
Prevent jumping around in indexing status table.

## How Has This Been Tested?

## Accepted Risk

## Related Issue(s)

## Checklist:


web/src/app/admin/configuration/search/UpgradingPage.tsx
```diff
@@ -49,6 +49,12 @@ export default function UpgradingPage({
     setIsCancelling(false);
   };
 
+  const sortedReindexingProgress = useMemo(() => {
```
yuhongsun96, line 59: Could we sort it by failed then in progress ones then completed? The failed one should should up on page 1 to give user the most important info IMO. It's also ok if the rows move a bit as they update but the other problem is that the table row alignment changes based on the contents, probably we want to freeze that

PR #2293, opened 2024-09-02
Connector deletion fix

hagen-danswer:


backend/tests/integration/tests/connector/test_deletion.py
```diff
@@ -274,32 +276,30 @@ def test_connector_deletion_for_overlapping_connectors(
         user_performing_action=admin_user,
     )
 
-    # EVERYTHING BELOW HERE IS CURRENTLY BROKEN AND NEEDS TO BE FIXED SERVER SIDE
-
     # wait for deletion to finish
```
hagen-danswer, line 279: Enabled this test, now that it should work properly

backend/danswer/background/connector_deletion.py
```diff
@@ -102,7 +102,7 @@ def delete_connector_credential_pair_batch(
                 UpdateRequest(
                     document_ids=[document_id],
                     access=access,
```
hagen-danswer: This is the fix. Basically, if the document didnt have any associated doc sets, it would not return any dict. 
Setting it to None would just not update the doc set relationship at all in vespa.
So we can assume that if we are updating it and the document id isnt in the dict, that it has no sets related to it

backend/danswer/db/document_set.py
```diff
@@ -524,38 +524,54 @@ def fetch_document_sets_for_documents(
     db_session: Session,
 ) -> Sequence[tuple[str, list[str]]]:
     """Gives back a list of (document_id, list[document_set_names]) tuples"""
```
hagen-danswer, line 526: Seperated out the stmt building and added a comment

PR #2299, opened 2024-09-02
Remove unused env variables

Weves:
Everything has been moved to the UI, these env variables just confuse people 😢


backend/danswer/llm/custom_llm.py
```diff
@@ -37,7 +36,7 @@ def __init__(
         # Not used here but you probably want a model server that isn't completely open
         api_key: str | None,
         timeout: int,
-        endpoint: str | None = GEN_AI_API_ENDPOINT,
```
Weves, line 40: NOTE: this class is unused currently, so this is a noop

hagen-danswer (approved):
negative LoC Classic value detraction

Looks good!

PR #2306, opened 2024-09-03
fixed doc set table

hagen-danswer:


hagen-danswer:
Here is an example of it working:

pablodanswer (approved):

PR #2310, opened 2024-09-03
Added instance domain to telemetry

hagen-danswer:
## Description
This pr adds a field called instance_domain to the anonymous_telemetry. This is only sent if EE features are enabled and telemetry is also enabled. the purpose is to help us track usage of our enterprise users per org (not per user)

### DIFF
--- a/.github/workflows/run-it.yml
+++ b/.github/workflows/run-it.yml
@@ -94,6 +94,7 @@ jobs:
           ENABLE_PAID_ENTERPRISE_EDITION_FEATURES=true \
           AUTH_TYPE=basic \
           REQUIRE_EMAIL_VERIFICATION=false \
+          DISABLE_TELEMETRY=true \
           IMAGE_TAG=it \
           docker compose -f docker-compose.dev.yml -p danswer-stack up -d --build
         id: start_docker
--- a/backend/danswer/configs/constants.py
+++ b/backend/danswer/configs/constants.py
@@ -57,6 +57,7 @@ KV_SLACK_BOT_TOKENS_CONFIG_KEY = "slack_bot_tokens_config_key"
 KV_GEN_AI_KEY_CHECK_TIME = "genai_api_key_last_check_time"
 KV_SETTINGS_KEY = "danswer_settings"
 KV_CUSTOMER_UUID_KEY = "customer_uuid"
+KV_INSTANCE_DOMAIN_KEY = "instance_domain"
 KV_ENTERPRISE_SETTINGS_KEY = "danswer_enterprise_settings"
 KV_CUSTOM_ANALYTICS_SCRIPT_KEY = "__custom_analytics_script__"
 
--- a/backend/danswer/main.py
+++ b/backend/danswer/main.py
@@ -325,6 +325,7 @@ async def lifespan(app: FastAPI) -> AsyncGenerator:
     # fill up Postgres connection pools
     await warm_up_connections()
 
+    # Generate a UUID for the instance so we can track usage
     get_or_generate_uuid()
 
     with Session(engine) as db_session:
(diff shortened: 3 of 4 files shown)


.github/workflows/run-it.yml
```diff
@@ -94,6 +94,7 @@ jobs:
           ENABLE_PAID_ENTERPRISE_EDITION_FEATURES=true \
           AUTH_TYPE=basic \
           REQUIRE_EMAIL_VERIFICATION=false \
```
hagen-danswer, line 96: this should disable telemetry spam from our integration tests

backend/danswer/configs/constants.py
```diff
@@ -57,6 +57,7 @@
 KV_GEN_AI_KEY_CHECK_TIME = "genai_api_key_last_check_time"
 KV_SETTINGS_KEY = "danswer_settings"
 KV_CUSTOMER_UUID_KEY = "customer_uuid"
```
hagen-danswer, line 59: the new value is stored in postgres as a kv so it stays consistent once generated

backend/danswer/utils/telemetry.py
```diff
@@ -21,14 +28,43 @@ class RecordType(str, Enum):
     FAILURE = "failure"
 
 
-def get_or_generate_uuid() -> str:
+def _get_or_generate_uuid() -> str:
+    global _CACHED_UUID
```
hagen-danswer, line 32: added caching for uuid generation

backend/danswer/utils/telemetry.py
```diff
@@ -21,14 +28,43 @@ class RecordType(str, Enum):
...
+
     try:
-        return cast(str, kv_store.load(KV_CUSTOMER_UUID_KEY))
+        _CACHED_INSTANCE_DOMAIN = cast(str, kv_store.load(KV_INSTANCE_DOMAIN_KEY))
     except ConfigNotFoundError:
-        customer_id = str(uuid.uuid4())
-        kv_store.store(KV_CUSTOMER_UUID_KEY, customer_id, encrypt=True)
-        return customer_id
+        with Session(get_sqlalchemy_engine()) as db_session:
+            first_user = db_session.query(User).first()
```
hagen-danswer, line 60: The logic to determine the "domain" of the danswer instance is to get the first user and use everything after the @ symbol of their email (E.g. danswer.ai)

backend/danswer/utils/telemetry.py
```diff
@@ -41,16 +77,19 @@ def optional_telemetry(
...
+                customer_uuid = _get_or_generate_uuid()
                 payload = {
                     "data": data,
                     "record": record_type,
                     # If None then it's a flow that doesn't include a user
                     # For cases where the User itself is None, a string is provided instead
                     "user_id": user_id,
-                    "customer_uuid": get_or_generate_uuid(),
+                    "customer_uuid": customer_uuid,
                 }
```
hagen-danswer, line 88: only run if EE is enabled (and Telemetry isnt disabled)

backend/danswer/utils/telemetry.py
```diff
@@ -21,14 +28,43 @@ class RecordType(str, Enum):
     FAILURE = "failure"
 
 
-def get_or_generate_uuid() -> str:
+def _get_or_generate_uuid() -> str:
```
Weves: can we generate this on startup?

backend/danswer/main.py
```diff
@@ -324,6 +325,8 @@ async def lifespan(app: FastAPI) -> AsyncGenerator:
     # fill up Postgres connection pools
     await warm_up_connections()
 
+    get_or_generate_uuid()
```
Weves, line 329: can we add a comment here about why we need this?

PR #2312, opened 2024-09-03
Add ingestion as a "Source" for the FE + improve typing

Weves:
Prevents page from crashing in the case that a connector with source `INGESTION_API` is setup.

It also fixes typing for `ValidSources` + fixes a small bug with the google sites connector.

### DIFF
--- a/web/src/app/chat/ChatIntro.tsx
+++ b/web/src/app/chat/ChatIntro.tsx
@@ -1,30 +1,9 @@
-import { getSourceMetadataForSources, listSourceMetadata } from "@/lib/sources";
+import { getSourceMetadataForSources } from "@/lib/sources";
 import { ValidSources } from "@/lib/types";
-import Image from "next/image";
 import { Persona } from "../admin/assistants/interfaces";
 import { Divider } from "@tremor/react";
-import { FiBookmark, FiCpu, FiInfo, FiX, FiZoomIn } from "react-icons/fi";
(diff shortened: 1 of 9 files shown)


web/src/app/chat/ChatIntro.tsx
```diff
@@ -1,30 +1,9 @@
-import { getSourceMetadataForSources, listSourceMetadata } from "@/lib/sources";
+import { getSourceMetadataForSources } from "@/lib/sources";
 import { ValidSources } from "@/lib/types";
-import Image from "next/image";
 import { Persona } from "../admin/assistants/interfaces";
 import { Divider } from "@tremor/react";
-import { FiBookmark, FiCpu, FiInfo, FiX, FiZoomIn } from "react-icons/fi";
```
Weves, line 6: This is just some cleanup—it's all unused

pablodanswer:
Looks good overall and definitely agree re: pushing back ambiguous types (ie. `any` / `string`)  and valid types in the same vein - would prevent issues like the ingestion api from popping up.

Nit: I think overall it might be worth making naming / delineation slightly clearer (afaict the Ingestion API still is "displayed" in the filters)  so maybe something like ConfigurableSources?

pablodanswer (approved):

Weves:
@pablodanswer, great suggestion! Changed to `ConfigurableSources` 🧐

PR #2317, opened 2024-09-03
Test stream + Update Copy

pablodanswer:
## Description
- Add most basic tests that still cover most cases for chat + stream endpoints
- (minor) Update model copy to be standardized

## How Has This Been Tested?

## Accepted Risk

## Related Issue(s)

## Checklist:


backend/tests/integration/common_utils/managers/chat.py
```diff
@@ -0,0 +1,144 @@
...
+from tests.integration.common_utils.constants import GENERAL_HEADERS
+from tests.integration.common_utils.test_models import StreamedResponse
+from tests.integration.common_utils.test_models import TestChatMessage
+from tests.integration.common_utils.test_models import TestChatSession
+from tests.integration.common_utils.test_models import TestUser
+from tests.integration.tests.streaming_endpoints.utils import analyze_response
+
+
+class ChatSessionManager:
+    @staticmethod
```
hagen-danswer, line 23: For object creation (and other overlapping functions), i think we should use the same name E.g. ChatSessionManager.create. The rest of the function names are good i think tho

backend/tests/integration/tests/streaming_endpoints/utils.py
```diff
@@ -0,0 +1,24 @@
+from typing import Any
+from typing import Dict
+from typing import List
+
+from tests.integration.common_utils.test_models import StreamedResponse
+
```
hagen-danswer: I think this should be a function in the chatmanager called like "translate_response".

So we can keep all the utilities for handling the api responses in the manager class. Feels a bit odd to have utils in 2 different places.

hagen-danswer: or maybe this could be a helper function in the same module as the chatmanager where the test writer isn't really expected to interact with it.

hagen-danswer (changes requested):
Couple comments, mostly nit. Looks good!

hagen-danswer (approved):

PR #2337, opened 2024-09-05
Feature/background deletion

rkuo-danswer:
## Description
Move connector deletion and subtask generation to celery + redis queue 

## How Has This Been Tested?

## Accepted Risk

## Related Issue(s)

## Checklist:


backend/danswer/background/celery/celery_app.py
```diff
@@ -411,26 +456,37 @@ def check_for_vespa_sync_task() -> None:
...
+    r = redis_pool.get_client()
 
-                cleanup_connector_credential_pair_task.apply_async(
-                    kwargs=dict(
-                        connector_id=cc_pair.connector.id,
-                        credential_id=cc_pair.credential.id,
-                    ),
+    lock_beat = r.lock(
+        DanswerRedisLocks.CHECK_CONNECTOR_DELETION_BEAT_LOCK,
+        timeout=CELERY_VESPA_SYNC_BEAT_LOCK_TIMEOUT,
```
Weves, line 473: is 60 seconds enough to create a job for a very large connector (e.g. 100k docs)?

backend/danswer/background/celery/celery_app.py
```diff
@@ -635,6 +691,89 @@ def vespa_metadata_sync_task(self: Task, document_id: str) -> bool:
...
+                delete_documents_complete__no_commit(
+                    db_session=db_session,
+                    document_ids=[document_id],
+                )
+            elif count > 1:
+                doc = get_document(document_id, db_session)
+                if not doc:
+                    return False
+
+                doc_access = get_access_for_document(
```
Weves, line 736: can we add a comment here about why these will be correct e.g. it will have the correct access since we don't consider deleting cc pairs in `get_access_for_document` / `fetch_document_sets_for_document`)?

I got confused myself for a second, and thought there was a bug here

backend/danswer/background/celery/celery_app.py
```diff
@@ -635,6 +691,89 @@ def vespa_metadata_sync_task(self: Task, document_id: str) -> bool:
...
+                    document_id=document_id, db_session=db_session
+                )
+
+                doc_sets = fetch_document_sets_for_document(document_id, db_session)
```
Weves, line 740: same comment here

backend/danswer/background/celery/celery_app.py
```diff
@@ -781,37 +931,136 @@ def monitor_usergroup_taskset(key_bytes: bytes, r: Redis, db_session: Session) -
...
+            db_session=db_session, user_group_id=usergroup_id
         )
+        if user_group:
+            if user_group.is_up_for_deletion:
+                delete_user_group = fetch_versioned_implementation_with_fallback(
+                    "danswer.db.user_group", "delete_user_group", noop_fallback
+                )
+
+                delete_user_group(db_session=db_session, user_group=user_group)
+                task_logger.info(f" Deleted usergroup. id='{usergroup_id}'")
```
Weves: nit: extra space

backend/danswer/background/celery/celery_app.py
```diff
@@ -735,28 +880,33 @@ def monitor_document_set_taskset(
...
+                task_logger.info(
+                    f"Successfully synced document set with ID: '{document_set_id}'!"
+                )
 
     r.delete(rds.taskset_key)
     r.delete(rds.fence_key)
 
 
-def monitor_usergroup_taskset(key_bytes: bytes, r: Redis, db_session: Session) -> None:
+def monitor_usergroup_taskset(key_bytes: bytes, r: Redis) -> None:
```
Weves: prefer having this whole function under the `/ee` directory. Moving this there would also remove the need to do 3 `fetch_versioned_implementation` calls below.

backend/danswer/document_index/vespa/index.py
```diff
@@ -377,6 +377,93 @@ def update(self, update_requests: list[UpdateRequest]) -> None:
...
 
+    def update_single(self, update_request: UpdateRequest) -> None:
+        """Note: if the document id does not exist, the update will be a no-op and the
+        function will complete with no errors or exceptions.
+        Handle other exceptions if you wish to implement retry behavior
+        """
+        if len(update_request.document_ids) != 1:
+            raise ValueError("update_request must contain a single document id")
+
+        # logger.debug(f"Updating {len(update_requests)} documents in Vespa")
```
Weves: nit: remove / uncomment

backend/danswer/document_index/vespa/index.py
```diff
@@ -377,6 +377,93 @@ def update(self, update_requests: list[UpdateRequest]) -> None:
             time.monotonic() - update_start,
         )
 
+    def update_single(self, update_request: UpdateRequest) -> None:
```
Weves, line 380: this is a decent amount of duplication with the standard `update`. Why do we need this?

Weves (changes requested):
A few small comments 🧐

backend/danswer/background/celery/celery_app.py
rkuo-danswer: 60 seconds is just the initial timeout ... the lock is actually passed down and reacquired frequently while iterating through rows to generate tasks, so this should be fine.

backend/danswer/document_index/vespa/index.py
rkuo-danswer: The single doc version is a great deal simpler to step through and debug, and I believe a great deal of the complexity in the update method and what it is doing with various threadpools, etc ... will be lifted out into celery's parallelism and appropriate usage of a single httpx connection pool dedicated to vespa.  Keeping the regular update function around for now, but it could be deprecated once this gets enough battle testing.

backend/danswer/background/celery/celery_app.py
rkuo-danswer: I'll look into this ... there were some circular dependency issues but i'll have to double check if that was an issue here.

Weves (approved):
lgtm 🙌

PR #2342, opened 2024-09-06
Add ability to delete users

pablodanswer:
## Description

Adds ability to delete users

Screenshot 2024-09-05 at 7 00 49 PM

Screenshot 2024-09-05 at 7 01 02 PM

### DIFF
--- a/backend/danswer/server/manage/users.py
+++ b/backend/danswer/server/manage/users.py
@@ -213,6 +213,50 @@ def deactivate_user(
     db_session.commit()
 
 
+@router.delete("/manage/admin/delete-user")
+async def delete_user(
+    user_email: UserByEmail,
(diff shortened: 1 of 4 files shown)


backend/danswer/server/manage/users.py
```diff
@@ -213,6 +213,50 @@ def deactivate_user(
...
+
+    try:
+        # Delete related OAuthAccounts first
+        for oauth_account in user_to_delete.oauth_accounts:
+            db_session.delete(oauth_account)
+
+        db_session.delete(user_to_delete)
+        db_session.commit()
+
+        user_emails = get_invited_users()
```
rkuo-danswer, line 249: Probably a race condition if two calls to this api are made at the same time?  Let's add a note to the code here and a fix it ticket.

PR #2344, opened 2024-09-06
[Feature] HighSpot Connector

0xRy4n:
## Description
Adds a HighSpot Connector. Syncs text-based items from all spots available to a user. Non-downloadable items are stored as a link to the item alongside it's description.

## How Has This Been Tested?
- Tested with Live HighSpot Instance

## Demo

HighSpot)

**After Indexing:**

Post-Index


rajivml:
Exactly what we are looking for. Thanks for raising this PR

PR #2357, opened 2024-09-07
Fix streaming auth locally

Weves:
Allows local dev with auth enabled + using streaming endpoints

### DIFF
--- a/web/src/app/api/[...path]/route.ts
+++ b/web/src/app/api/[...path]/route.ts
@@ -97,27 +97,10 @@ async function handleRequest(request: NextRequest, path: string[]) {
         headers: response.headers,
       });
     } else {
-      // If it's not a stream, handle it based on content type
-      const contentType =
-        response.headers.get("content-type") || "application/json";
-
-      if (contentType.startsWith("image/")) {
-        // For image types, return the response as-is
-        return new NextResponse(response.body, {
-          status: response.status,
-          headers: response.headers,
-        });
-      } else {
-        // For other types, handle as before
-        const data = await response.text();
-
-        return new NextResponse(data, {
-          status: response.status,
-          headers: {
-            "Content-Type": contentType,
-          },
-        });
-      }
+      return new NextResponse(response.body, {
+        status: response.status,
+        headers: response.headers,
+      });
     }
   } catch (error: unknown) {
     console.error("Proxy error:", error);
--- a/web/src/app/api/[...path]/route.ts
+++ b/web/src/app/api/[...path]/route.ts
@@ -1,5 +1,8 @@
 import { NextRequest, NextResponse } from "next/server";
 
+/* NextJS is annoying and makes use use a separate function for 
+each request type >:( */
+
 export async function GET(
   request: NextRequest,
   { params }: { params: { path: string[] } }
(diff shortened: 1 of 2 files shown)


web/src/app/api/[...path]/route.ts
```diff
@@ -0,0 +1,124 @@
...
+      { status: 404 }
+    );
+  }
+
+  try {
+    const backendUrl = new URL(`http://localhost:8080/${path.join("/")}`);
+    const requestBody = request.body
+      ? JSON.stringify(await request.json())
+      : null;
+    // console.log(requestBody)
```
Comment (pablonyx) [inline web/src/app/api/[...path]/route.ts]: Nit: should likely remove this

web/src/app/api/[...path]/route.ts
```diff
@@ -0,0 +1,124 @@
...
+      {
+        message:
+          "This API is only available in development mode. In production, something else (e.g. nginx) should handle this.",
+      },
+      { status: 404 }
+    );
+  }
+
+  try {
+    const backendUrl = new URL(`http://localhost:8080/${path.join("/")}`);
```
Comment (pablonyx) [inline web/src/app/api/[...path]/route.ts]: Nit: likely worth importing and using `INTERNAL_URL` in case API server is running on a different port

pablodanswer (changes requested):

web/src/app/api/[...path]/route.ts
```diff
@@ -0,0 +1,124 @@
...
+    const requestBody = request.body
+      ? JSON.stringify(await request.json())
```
Comment (pablonyx) [inline web/src/app/api/[...path]/route.ts]: Believe we need to add additional handling for different content types (specifically I know we're using formData, url params in a couple places)

web/src/app/api/[...path]/route.ts
```diff
@@ -0,0 +1,124 @@
...
+      const { readable, writable } = new TransformStream();
+      response.body?.pipeTo(writable);
+
+      return new NextResponse(readable, {
+        status: response.status,
+        headers: response.headers,
+      });
+    } else {
+      // If it's not a stream, handle it as before
+      const data = await response.text();
```
Comment (pablonyx) [inline web/src/app/api/[...path]/route.ts]: Don't think it's valid to assume we always want .text() - (e.g. for `204` or any other content types)

web/src/app/api/[...path]/route.ts
Comment (Weves) [reply web/src/app/api/[...path]/route.ts]: Good suggestion—added!

web/src/app/api/[...path]/route.ts
Comment (Weves) [reply web/src/app/api/[...path]/route.ts]: Added handling for formData! url params should be handled above 👀

web/src/app/api/[...path]/route.ts
Comment (Weves) [reply web/src/app/api/[...path]/route.ts]: good catch

web/src/app/api/[...path]/route.ts
Comment (Weves) [reply web/src/app/api/[...path]/route.ts]: Added handling for images. I believe 204 should be handled properly (.text() will just be empty string)

web/src/app/api/[...path]/route.ts
```diff
@@ -0,0 +1,133 @@
...
+        });
+      } else {
+        // For other types, handle as before
+        const data = await response.text();
+
+        return new NextResponse(data, {
+          status: response.status,
+          headers: {
+            "Content-Type": contentType,
+          },
```
Comment (pablonyx) [inline web/src/app/api/[...path]/route.ts]: Think we're not handling all the data types (e.g. 204 response code, json content type) - maybe we could have a stricter check for using response.text() and default to response.body more often?

Comment (Weves) [reply web/src/app/api/[...path]/route.ts]: I believe those should all be handled as expected (e.g. all our APIs return JSONs, but it all works as expected + 204 is "no content" which is handled fine here as well). Afaik most response types are just text / can be represented fully as just text

Comment (Weves) [reply web/src/app/api/[...path]/route.ts]: using body by default now

PR #2361, opened 2024-09-08
Clearer onboarding + Provider Updates

pablodanswer:
## Description
- Updates onboarding flow to be slightly clearer
- Update context to be more robust (should fix the chat history issue)

https://github.com/user-attachments/assets/28c5de41-ae36-4a6d-800b-f8e7b49d54b5

## How Has This Been Tested?

## Accepted Risk

## Related Issue(s)

## Checklist:


web/src/components/chat_search/NoCredentialText.tsx
```diff
@@ -0,0 +1,26 @@
+import { useProviderStatus } from "./ProviderContext";
+export default function NoCredentialText({
```
Weves: nit: probably should have a different name + some spacing above

web/src/components/chat_search/hooks.ts
```diff
@@ -1,4 +1,17 @@
-import { Dispatch, SetStateAction, useEffect, useRef, useState } from "react";
+import { useUser } from "../user/UserProvider";
```
Weves: nit: these aren't needed anymore?

web/src/components/context/SearchContext.tsx
```diff
@@ -0,0 +1,37 @@
+"use client";
+
+import React, { createContext, useContext } from "react";
+import { CCPairBasicInfo, DocumentSet, Tag } from "@/lib/types";
+import { Persona } from "@/app/admin/assistants/interfaces";
+
+interface SearchContextProps {
+  querySessions: any[]; // Replace 'any' with the correct type
```
Weves: can we add this type? 🧐

web/src/app/layout.tsx
```diff
@@ -70,11 +72,37 @@ export default async function RootLayout({
 
             <Card className="p-8 max-w-md">
               <h1 className="text-2xl font-bold mb-4 text-error">Error</h1>
-              <p className="text-text-500">
-                Your Danswer instance was not configured properly and your
-                settings could not be loaded. Please contact your admin to fix
-                this error.
-              </p>
+              {EE_ENABLED ? (
```
Weves: I don't think this is used anymore (we should just remove it). What's goal for this change anyways?

web/src/app/search/page.tsx
```diff
@@ -110,16 +112,16 @@ export default async function Home() {
     console.log(`Failed to fetch chat sessions - ${queryResponse?.text()}`);
   }
 
-  let assistants: Persona[] = initialAssistantsList;
+  let personas: Persona[] = initialAssistantsList;
```
Weves: nit: pref switching it back. We'll need to move everything over to assistants later, so this will just be one additional thing to transition 😓

pablonyx: Fair!

web/src/components/context/SearchContext.tsx
pablonyx: We can indeed

web/src/app/layout.tsx
pablonyx: Ideally not have self-hosted users confused by the "please contact your Danswer admin to get it added." Can make  more general copy

Weves: Oops, should have been more clear—`EE_ENABLED` isn't actually set anymore so this will always use the latter of the two options 😅

Also, EE_ENABLED isn't really the same as self-hosted vs cloud. Maybe we could just use the new message 100% of the time?

Weves (approved):

PR #2369, opened 2024-09-08
Add user when they interact outside of UI (e.g. Slack bot)

hj-danswer:
## Description
Add capability to add user when someone interacts with a Slack bot without having logged in via the UI. This invovles:
- Adding a `has_web_login` column to `user` table. This allows us to have logic where we allow a user to set their password with basic auth. Otherwise, a user would not know the placeholder password we set for them.

## How Has This Been Tested?
- [x] Create a new user with Slack only. Be able to then authenticate with web login (basic, oidc, google_oauth).
- [x] Authenticate with web first. Can continue to interact with Slack bot.

## Accepted Risk
More database calls / Slack APIs in Slack bot + login flows.

## Checklist:

### DIFF
--- a/backend/alembic/versions/f7e58d357687_add_has_web_column_to_user.py
+++ b/backend/alembic/versions/f7e58d357687_add_has_web_column_to_user.py
@@ -10,7 +10,7 @@ import sqlalchemy as sa
 
 # revision identifiers, used by Alembic.
 revision = "f7e58d357687"
-down_revision = "bceb1e139447"
+down_revision = "ba98eba0f66a"
 branch_labels = None
 depends_on = None
 
--- a/backend/alembic/versions/f7e58d357687_add_has_web_column_to_user.py
+++ b/backend/alembic/versions/f7e58d357687_add_has_web_column_to_user.py
@@ -1,0 +1,26 @@
+"""add has_web column to user
+
+Revision ID: f7e58d357687
+Revises: bceb1e139447
+Create Date: 2024-09-07 20:20:54.522620
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
+revision = "f7e58d357687"
+down_revision = "bceb1e139447"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "user",
+        sa.Column("has_web_login", sa.Boolean(), nullable=False, server_default="true"),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("user", "has_web_login")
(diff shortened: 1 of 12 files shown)


backend/alembic/versions/f7e58d357687_add_has_web_column_to_user.py
```diff
@@ -0,0 +1,26 @@
...
+revision = "f7e58d357687"
+down_revision = "bceb1e139447"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "user",
+        sa.Column("has_web_login", sa.Boolean(), nullable=False, server_default="true"),
```
yuhongsun96, line 21: We should test with another user as well (turn off email verification and use basic auth). Do a login with test@danswer.ai off of main and then upgrade the version. I think it will fail due to the NOT NULL constraint. We have to support graceful upgrades

backend/danswer/auth/schemas.py
```diff
@@ -1,5 +1,6 @@
 import uuid
 from enum import Enum
+from typing import Optional
```
yuhongsun96: Is there a reason we can't use the built in typing that came with the more recent python versions?

backend/danswer/danswerbot/slack/handlers/handle_message.py
```diff
@@ -47,6 +52,25 @@ def send_msg_ack_to_user(details: SlackMessageInfo, client: WebClient) -> None:
     )
 
 
+def add_user_if_not_exists(email: str, db_session: Session) -> User:
```
yuhongsun96: We like to throw these functions that interface with the DB models (and that may be reuseable) in the related file under backend/db. Makes organization/finding things easy and helps prevents circular imports.

backend/danswer/danswerbot/slack/handlers/handle_message.py
```diff
@@ -208,7 +232,14 @@ def handle_message(
     except SlackApiError as e:
         logger.error(f"Was not able to react to user message due to: {e}")
 
+    slack_user_info = expert_info_from_slack_id(
```
yuhongsun96: Since we've added this in a few places recently, is there a way we can do it at the top level and pass it through? That way we make less Slack calls. In this case the volume shouldn't be very high but I think we can probably make this small optimization. The case where it would matter is if for example they have a Slack connector that has been rate limited due to volume of requests and it's affecting this side as well (some folks don't create separate Slack apps for connector and SlackBot)

yuhongsun96 (commented):
I think this is a breaking migration, rest looks good 👍

yuhongsun96 (approved):
Whoohoo LGTM

PR #2373, opened 2024-09-09
ephemeral behavior for redis

rkuo-danswer:
## Description
Fixes DAN-579.

Turns out the default docker container behavior is somewhat persistent, despite nearly all articles on the internet stating otherwise.

## How Has This Been Tested?
Observed incorrect persistence behavior, then removed volume and forced command line and finally saw keys removed between restarts.

## Accepted Risk

## Related Issue(s)

## Checklist:

### DIFF
--- a/deployment/docker_compose/docker-compose.dev.yml
+++ b/deployment/docker_compose/docker-compose.dev.yml
@@ -341,12 +341,9 @@ services:
     restart: always
     ports:
       - '6379:6379'
-    command: redis-server
-    volumes: 
-      - cache_volume:/data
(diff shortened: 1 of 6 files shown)


Weves (commented):
@rkuo-danswer do we need to change the kubernetes / helm files as well

rkuo-danswer:
> @rkuo-danswer do we need to change the kubernetes / helm files as well

Yes, I'll add kubernetes now. Helm we're saving for later since there's quite a bit more going on there, will wait for when we do the full pass on that.

pablodanswer (approved):

PR #2377, opened 2024-09-09
Support regex in standard answers

hj-danswer:
## Description
Standard answers supports regex patterns. Checks if admin-provided regex phrase can be found in the user question. Maintains old behavior of keyword match as well.

## How Has This Been Tested?
Manually
Screenshot 2024-09-09 at 2 38 51 PM
Screenshot 2024-09-09 at 2 40 26 PM

## Checklist:


web/src/app/admin/standard-answer/StandardAnswerCreationForm.tsx
```diff
@@ -44,7 +44,7 @@ export const StandardAnswerCreationForm = ({
           }}
           validationSchema={Yup.object().shape({
             keyword: Yup.string()
-              .required("Keyword or phrase is required")
+              .required("Keywords or pattern is required")
```
yuhongsun96, line 51: How about "Keywords" or "Regex Pattern" depending -> with a checkbox underneath that says "enable regular expression"?

Personally I would also have a field in the DB to save whether or not it's a regex match and update the logic on the inference flow to only check either keyword or regex depending.

trial-danswer:
Taking over this PR. Went ahead and rebased to start.

web/src/app/admin/standard-answer/StandardAnswerCreationForm.tsx
trial-danswer: Sounds good. Should we also validate the regex upon submission? Might be a good user experience stopgap in lieu of a full regex composer or something similar.

web/src/app/admin/standard-answer/StandardAnswerCreationForm.tsx
```diff
@@ -86,12 +90,28 @@ export const StandardAnswerCreationForm = ({
...
+                  isCode
+                  tooltip="Triggers if the question contains this regex pattern (matches Python `re.search()`)"
+                  placeholder="danswer/.*?/admin"
+                />
+              ) : (
+                <TextFormField
+                  name="keyword"
+                  label="Keywords or pattern"
+                  tooltip="Triggers if all specified keywords are in the question or if regex pattern exists in the question"
+                  placeholder="wifi"
```
rkuo-danswer: 1. Does the placeholder "wifi" here make sense? Is there a more descriptive example we could use?
2. How should the user enter multiple keywords?
3. What's the difference between "keywords or pattern" and "regex pattern"?

backend/danswer/db/standard_answer.py
```diff
@@ -182,30 +187,39 @@ def find_matching_standard_answers(
     query: str,
     db_session: Session,
 ) -> list[StandardAnswer]:
+    """
+    Returns a list of tuples, where each tuple contains a StandardAnswer and a boolean
+    indicating if the query is a regex match.
```
rkuo-danswer: Doesn't this function do either a regex match or word matching?

backend/danswer/danswerbot/slack/handlers/handle_standard_answers.py
```diff
@@ -150,11 +150,13 @@ def handle_standard_answers(
...
         formatted_answers = []
         for standard_answer in matching_standard_answers:
-            block_quotified_answer = ">" + standard_answer.answer.replace("\n", "\n> ")
-            formatted_answer = (
-                f'Since you mentioned _"{standard_answer.keyword}"_, '
-                f"I thought this might be useful: \n\n{block_quotified_answer}"
+            since_you_mentioned_pretext = (
+                f'Since you mentioned _"{standard_answer.keyword}"_'
+                if not standard_answer.match_regex
+                else f"Since your question matches the regex pattern `{standard_answer.keyword}`"
```
rkuo-danswer: Putting my end user hat on, I feel like we shouldn't be showing regex's to end users ... was this was requested?  Seems much too technical.

rkuo-danswer (changes requested):

web/src/app/admin/standard-answer/StandardAnswerCreationForm.tsx
trial-danswer: Oh, this still has the copy from the original PR. Let me reword these.

This branch should be for exact keyword matches only.

backend/danswer/db/standard_answer.py
trial-danswer: Yep, this docstring is from the original PR made by the previous contributor. I'll edit this to match.

How does:

```Returns a list of StandardAnswer definitions matching the query. If `answer_instance.match_regex` is true, the definition is considered "matched" if the query matches the `answer_instance.keyword` using `re.search`. Otherwise, the definition is considered "matched" if each space-delimited token in `keyword` exists in `query`.```

Also happy to change the behavior in the exact match scenario, I kept the implementation from the original PR because there were no comments on it so I assumed that behavior was acceptable.

backend/danswer/danswerbot/slack/handlers/handle_standard_answers.py
trial-danswer: Gonna defer to @yuhongsun96 on this.

backend/danswer/db/standard_answer.py
trial-danswer: @yuhongsun96 are you okay with the original word-matching implementation? Happy to change that if desired.

web/src/app/admin/standard-answer/StandardAnswerCreationForm.tsx
trial-danswer: Fixed, take a look.

trial-danswer: To answer your other questions:

- the user provides keywords as a string, separated by spaces. If desired (probably in a separate PR?), we could replace this with some kind of multi-input tag system (like https://eui.elastic.co/#/forms/combo-box) ?
- open to suggestions on what might be a good replacement for `wifi`. maybe something that reflects the usefulness on matching all keywords? like `google groups`?

Starting to exceed the scope of the PR but we could also add a toggle for "any keywords" vs "all keywords"

web/src/app/admin/standard-answer/StandardAnswerCreationForm.tsx
yuhongsun96: I like that, makes sense!

backend/danswer/danswerbot/slack/handlers/handle_standard_answers.py
yuhongsun96: Was not requested, I think this probably will not add much value to the user but I don't hate it either. I think something more generic could work. We could grab the match group from the regex and show that instead. Like since your question contains "wifi password"... (in the regex case)

web/src/app/admin/standard-answer/StandardAnswerCreationForm.tsx
yuhongsun96: Ya I was thinking as well that any vs all keyword toggle would be great as well. Generally we try not to scope creep but this seems fairly minor and since it was an obvious one to you as well, let's do it.

How do you feel about the use case "IT ticket" and something like "If you need any help from the IT team, please email internalsupport@company.com"

yuhongsun96:
Generally looks good, please sync with Richard on his comments/suggestions. We can also do your suggestion on adding a match any vs match all. A few minor things on the copy here and there.

Also be sure to rebase on main (I think the migration head has moved)

web/src/app/admin/standard-answer/StandardAnswerCreationForm.tsx
trial-danswer: Cool, I added that to this PR using a Pydantic model validator.

web/src/app/admin/standard-answer/StandardAnswerCreationForm.tsx
trial-danswer: `IT Ticket` and the suggested markdown placeholder sounds good. How do you feel about scoping this PR specifically to the regex flag and then figuring out the priority of the any/all keywords feature in a separate PR?

backend/danswer/danswerbot/slack/handlers/handle_standard_answers.py
rkuo-danswer: Showing the matched result sounds more reasonable.

web/src/app/admin/standard-answer/StandardAnswerCreationForm.tsx
rkuo-danswer: Above suggestion sounds good.

trial-danswer: Talked with @yuhongsun96 and we'll work on that as a separate PR.

backend/danswer/danswerbot/slack/handlers/handle_standard_answers.py
trial-danswer: Alright I'll make that change.

trial-danswer:
We might also want to support case-sensitivity matching in the future, since certain terms (like `IT` vs `it`) can become ambiguous and noisy when doing case insensitive matching. See [this demo slack thread](https://danswerinternal.slack.com/archives/C060QGXF37F/p1726251893998139).

backend/danswer/danswerbot/slack/handlers/handle_standard_answers.py
trial-danswer: After talking with @yuhongsun96 I think it would be good to keep the matching logic in one place. Currently,  `find_matching_standard_answers` performs the matching, so I think extracting the matched string there makes sense.

In the interest of avoiding scope creep, I think it may make sense as a stopgap measure to reintroduce the `tuple[StandardAnswer, str]` response type, and then perform a more thorough refactor with a Pydantic type as a separate PR.

trial-danswer: I threw together a "fast but messy" implementation here to resolve this comment. We can decide how to prioritize the refactor separately.

rkuo-danswer (approved):

PR #2378, opened 2024-09-09
Migrate standard answers implementations to ee/

hj-danswer:
## Description
Migrate standard answers to `ee/`. Mostly involved copying files that are related to standard answers to `ee/` and using `fetch_versioned_implementation` to make the implementation toggle between the two versions.

## How Has This Been Tested?
Manually with EE turned on / off.

## Checklist:


backend/danswer/danswerbot/slack/blocks.py
```diff
@@ -360,22 +359,6 @@ def build_quotes_block(
     return [SectionBlock(text="*Relevant Snippets*\n" + "\n".join(quote_lines))]
 
 
-def build_standard_answer_blocks(
```
trial-danswer, line 363: I'm honestly not sure why this is removed; maybe stale code?

yuhongsun96: was moved to the ee equivalent module I believe

backend/danswer/danswerbot/slack/handlers/handle_standard_answers.py
```diff
@@ -66,153 +20,22 @@ def handle_standard_answers(
...
-            [
-                standard_answer.id
-                for chat_message in chat_messages
-                for standard_answer in chat_message.standard_answers
-            ]
-        )
-
-    usable_standard_answers = configured_standard_answers.difference(
-        used_standard_answer_ids
+    return versioned_handle_standard_answers(
```
yuhongsun96, line 29: Can we make these named arguments since there are so many of them? Generally better to use named args unless immediately obvious

backend/danswer/danswerbot/slack/handlers/handle_standard_answers.py
```diff
@@ -66,153 +20,22 @@ def handle_standard_answers(
...
+def _handle_standard_answers(
+    message_info: SlackMessageInfo,
+    receiver_ids: list[str] | None,
+    slack_bot_config: SlackBotConfig | None,
+    prompt: Prompt | None,
+    logger: DanswerLoggingAdapter,
+    client: WebClient,
+    db_session: Session,
+) -> bool:
+    return False
```
yuhongsun96, line 56: I know this was already the way this was handled but I think when functions return some cryptic type, it's good to give a short docstring just to say what the return is suppose to represent. We don't have this best practice implemented everywhere but it's generally good to do

backend/danswer/danswerbot/slack/handlers/handle_standard_answers.py
trial-danswer: Yeah can do. This is from before I took over the PR.

backend/ee/danswer/danswerbot/slack/blocks.py
```diff
@@ -0,0 +1,22 @@
+from slack_sdk.models.blocks import ActionsBlock
+from slack_sdk.models.blocks import Block
+from slack_sdk.models.blocks import ButtonElement
+from slack_sdk.models.blocks import SectionBlock
+
+from danswer.danswerbot.slack.constants import GENERATE_ANSWER_BUTTON_ACTION_ID
+
+
+def build_standard_answer_blocks(
```
yuhongsun96: honestly I would just throw it in the other module just for simplicity (handle_standard_answers.py under ee)

yuhongsun96 (approved):

backend/danswer/danswerbot/slack/handlers/handle_standard_answers.py
trial-danswer: This is done now.

backend/danswer/danswerbot/slack/handlers/handle_standard_answers.py
trial-danswer: Docstrings added.

backend/ee/danswer/danswerbot/slack/blocks.py
trial-danswer: Done

PR #2393, opened 2024-09-11
#2282 fixed

### LINKED ISSUE #2282: Github Pages Connector
Indexes all pages that are part of a [Github Pages](https://pages.github.com/)-based website. Should use the Github APIs directly, as we want to be able to index sites that are behind authentication / internal firewalls. Where possible, we should re-use functionality / common utilities present in the existing [Github](https://github.com/danswer-ai/danswer/blob/main/backend/danswer/connectors/github/connector.py) / [Web](https://github.com/danswer-ai/danswer/blob/main/backend/danswer/connectors/web/connector.py) connectors.

Checkout the connector creation README here for more details on the best way to add new connectors: https://github.com/danswer-ai/danswer/blob/main/backend/danswer/connectors/README.md.

techsouvik:
#2282 fixed

/claim #2282 

@Weves check if it was needed or a separate Github Pages Connector?

### DIFF
--- a/backend/alembic/versions/0ebb1d516877_add_ccpair_deletion_failure_message.py
+++ b/backend/alembic/versions/0ebb1d516877_add_ccpair_deletion_failure_message.py
@@ -1,0 +1,27 @@
+"""add ccpair deletion failure message
+
+Revision ID: 0ebb1d516877
+Revises: 52a219fb5233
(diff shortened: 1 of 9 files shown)


Weves:
@techsouvik we would like a separate github pages connector!

PR #2399, opened 2024-09-11
Made PR checks only run when relevant files are changed

hagen-danswer:
This should make prs that dont touch the backend not run any of the python specific checks


.github/workflows/check-backend-changes.yml
```diff
@@ -0,0 +1,23 @@
+name: Check Backend Changes
+
+on:
+  workflow_call:
+    outputs:
+      run-tests:
+        description: "Whether to run tests based on backend changes"
```
pablonyx, line 7: Might be slightly preferrable to have a name for each job/step in the workflow

pablodanswer (approved):

PR #2413, opened 2024-09-12
Display only failed re-index attempts

pablodanswer:
## Description
Make connector re-index failure page more intelligible for large customers. Instead of displaying those which need reindexing, dispaly those with failed reindex attempts.


hagen-danswer (approved):
seems good

PR #2419, opened 2024-09-12
simple apis now cited/context doc indices

hagen-danswer:


backend/danswer/chat/models.py
```diff
@@ -60,7 +60,11 @@ def model_dump(self, *args: list, **kwargs: dict[str, Any]) -> dict[str, Any]:
 
 
 class LLMRelevanceFilterResponse(BaseModel):
-    relevant_chunk_indices: list[int]
+    llm_selected_doc_indices: list[int]
```
hagen-danswer, line 63: renamed this everywhere

backend/danswer/server/utils.py
```diff
@@ -1,9 +1,28 @@
 import json
+from datetime import datetime
 from typing import Any
```
hagen-danswer, line 5: made this function able to encode datetime objects

backend/ee/danswer/server/query_and_chat/chat_backend.py
```diff
@@ -120,17 +144,26 @@ def handle_simplified_chat_message(
...
         if isinstance(packet, DanswerAnswerPiece) and packet.answer_piece:
             answer += packet.answer_piece
         elif isinstance(packet, QADocsResponse):
-            response.simple_search_docs = translate_doc_response_to_simple_doc(packet)
+            response.simple_search_docs = _translate_doc_response_to_simple_doc(packet)
         elif isinstance(packet, StreamingError):
             response.error_msg = packet.error
         elif isinstance(packet, ChatMessageDetail):
             response.message_id = packet.message_id
+        elif isinstance(packet, FinalUsedContextDocsResponse):
```
hagen-danswer, line 155: here we get the final context docs and get the int index of that doc in the list of docs we are already returning

backend/ee/danswer/server/query_and_chat/models.py
```diff
@@ -74,4 +74,6 @@ class ChatBasicResponse(BaseModel):
     simple_search_docs: list[SimpleDoc] | None = None
     error_msg: str | None = None
     message_id: int | None = None
```
hagen-danswer, line 76: these are the 2 new fields (and one renamed) that we are returning through the simple apis

backend/danswer/llm/answering/answer.py
```diff
@@ -525,8 +523,9 @@ def _process_stream(
                                 SearchResponseSummary, message.response
                             ).top_sections
                         ]
```
hagen-danswer, line 525: here we yield the final context docs

backend/danswer/server/utils.py
```diff
@@ -1,9 +1,28 @@
...
+class DateTimeEncoder(json.JSONEncoder):
+    """Custom JSON encoder that converts datetime objects to ISO format strings."""
+
+    def default(self, obj: Any) -> Any:
+        if isinstance(obj, datetime):
+            return obj.isoformat()
+        return super().default(obj)
+
+
+def get_json_line(json_dict: dict[str, Any]) -> str:
```
pablonyx: Super nit:  I'd prefer passing DateTimeEncoder as the default argument to an encoder prop

backend/ee/danswer/server/query_and_chat/chat_backend.py
```diff
@@ -60,6 +63,27 @@ def translate_doc_response_to_simple_doc(
...
+    simple_search_docs: list[SimpleDoc] | None,
+) -> list[int] | None:
+    """
+    this function returns a list of indices of the simple search docs
+    that were actually fed to the LLM.
+    """
+    if final_context_docs is None or simple_search_docs is None:
+        return None
+
+    final_context_indices: list[int] = []
```
pablonyx: Another super nit: set + list comprehension would be slightly faster (doesn't matter really since we'll only ever have a small amount of docs to check)
Something like 
```
    final_context_doc_ids = {all docs ids}
    return [i for i, doc in enumerate(simple_search_docs) if doc.id in final_context_doc_ids]
```

pablodanswer (approved):

PR #2464, opened 2024-09-16
Fixes : `It looks like you are using Playwright Sync API inside the asyncio loop. Please use the Async API instead.` AND Prevent by default SSRF Server-side Request Forgery

Youho99:
Fixes : `It looks like you are using Playwright Sync API inside the asyncio loop. Please use the Async API instead.` 

AND 

Prevent by default SSRF Server-side Request Forgery


Youho99:
@Weves @hagen-danswer @rkuo-danswer @pablodanswer

PR #2465, opened 2024-09-16
Fix match_regex upgrade

Weves:
Was running into not-null constraint issues when upgrading locally on latest main. This resolved it.


Weves:
Duplicate with https://github.com/danswer-ai/danswer/pull/2442

PR #2467, opened 2024-09-16
EE movement followup for Standard Answers

trial-danswer:
## Description
This PR continues moving Standard Answers to the enterprise edition licensed portion of the repository.

Since APIs are already gated behind the EE flag, this is primarily a movement for future code licensing.

## How Has This Been Tested?
Locally to confirm that functionality still works.

## Accepted Risk
Same as prior EE migration - non-EE usage of these APIs will break.

## Related Issue(s)
#2378

## Checklist:


backend/danswer/db/slack_bot_config.py
```diff
@@ -14,8 +15,12 @@
 from danswer.db.persona import get_default_prompt
 from danswer.db.persona import mark_persona_as_deleted
 from danswer.db.persona import upsert_persona
-from danswer.db.standard_answer import fetch_standard_answer_categories_by_ids
 from danswer.search.enums import RecencyBiasSetting
+from danswer.utils.errors import EERequiredError
+from danswer.utils.variable_functionality import (
+    fetch_versioned_implementation_with_fallback,
+)
+from ee.danswer.db.standard_answer import fetch_standard_answer_categories_by_ids
```
yuhongsun96: We never import anything from ee into MIT

trial-danswer: Ah yep there's a case that still pulled this in, fixed.

backend/ee/danswer/danswerbot/slack/handlers/handle_standard_answers.py
```diff
@@ -21,11 +21,11 @@
 from danswer.db.models import Prompt
 from danswer.db.models import SlackBotConfig
 from danswer.db.models import StandardAnswer as StandardAnswerModel
```
yuhongsun96, line 23: Often we'll call it DbStandardAnswer and the other one will be just StandardAnswer for the server model

trial-danswer: `StandardAnswerModel` is the original naming from this file before I touched it, do you want me to change it as part of this refactor?

yuhongsun96 (approved):
lgtm

backend/ee/danswer/danswerbot/slack/handlers/handle_standard_answers.py
yuhongsun96: Doesn't matter, nobody will get confused by this anyway

backend/danswer/db/slack_bot_config.py
yuhongsun96: awesome thanks

PR #2469, opened 2024-09-16
Add top_documents to APIs

Weves:
Need to have the search_doc_id to specify which docs to use in the next call


hagen-danswer (approved):
I see

PR #2490, opened 2024-09-17
Update default assistants to all visible

pablodanswer:
## Description
Updates so default assistants list (from `null`) is all those which are.

It's okay if includes visible assistants which aren't accessible to user, as upon first re-ordering will refresh to be those which user has access to.

Additionally, I think it's still okay to set server default to our assistants - users can add any additional assistants in the assistant gallery on first startup (and if none are available- we'll display all visible as was the previous behavior). Alternative would be to modify how we handle chosen assistants or set the list on user creation.

## How Has This Been Tested?

## Accepted Risk

## Related Issue(s)

## Checklist:

### DIFF
--- a/backend/alembic/versions/35e6853a51d5_server_default_chosen_assistants.py
+++ b/backend/alembic/versions/35e6853a51d5_server_default_chosen_assistants.py
@@ -41,7 +41,6 @@ def upgrade() -> None:
         OR chosen_assistants = 'null'
         OR jsonb_typeof(chosen_assistants) = 'null'
         OR (jsonb_typeof(chosen_assistants) = 'string' AND chosen_assistants = '"null"')
-        OR chosen_assistants = '[-2, -1, 0]'::jsonb
     """
     )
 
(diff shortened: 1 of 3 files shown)


rkuo-danswer (approved):

backend/alembic/versions/35e6853a51d5_server_default_chosen_assistants.py
```diff
@@ -20,14 +20,28 @@
...
+            FROM persona p
+            LEFT JOIN persona__user pu ON p.id = pu.persona_id AND pu.user_id = u.id
+            WHERE p.is_visible = true
+            AND (p.is_public = true OR pu.user_id IS NOT NULL)
+        )
         WHERE chosen_assistants IS NULL
         OR chosen_assistants = 'null'
         OR jsonb_typeof(chosen_assistants) = 'null'
         OR (jsonb_typeof(chosen_assistants) = 'string' AND chosen_assistants = '"null"')
+        OR chosen_assistants = '[-2, -1, 0]'::jsonb
```
Weves: I don't think we need this?

Weves: since this only runs once

PR #2491, opened 2024-09-18
Add back llm_chunks_indices

Weves:
Request from customer for backwards compatibility. Will remove the extra fields once they have switched over.


backend/ee/danswer/server/query_and_chat/chat_backend.py
```diff
@@ -81,6 +82,49 @@ def _get_final_context_doc_indices(
     ]
 
 
+def _convert_packet_stream_to_response(
```
Weves, line 85: logic was the same between the two APIs, so just centralizing it here

pablodanswer (approved):
Looks good- created ticket for backlog items

PR #2494, opened 2024-09-18
clarify ssl cert reqs

rkuo-danswer:
## Description
small clarification of ssl_cert_reqs

## How Has This Been Tested?

## Accepted Risk

## Related Issue(s)

## Checklist:

### DIFF
--- a/backend/danswer/configs/app_configs.py
+++ b/backend/danswer/configs/app_configs.py
@@ -164,7 +164,9 @@ REDIS_DB_NUMBER_CELERY_RESULT_BACKEND = int(
 )
 REDIS_DB_NUMBER_CELERY = int(os.environ.get("REDIS_DB_NUMBER_CELERY", 15))  # broker
 
-REDIS_SSL_CERT_REQS = os.getenv("REDIS_SSL_CERT_REQS", "CERT_NONE")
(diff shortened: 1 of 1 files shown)


backend/danswer/configs/app_configs.py
```diff
@@ -164,7 +164,9 @@
 )
 REDIS_DB_NUMBER_CELERY = int(os.environ.get("REDIS_DB_NUMBER_CELERY", 15))  # broker
 
-REDIS_SSL_CERT_REQS = os.getenv("REDIS_SSL_CERT_REQS", "CERT_NONE")
+# https://docs.celeryq.dev/en/stable/userguide/configuration.html#redis-backend-settings
+# should be one of "required", "optional", or "none"
+REDIS_SSL_CERT_REQS = os.getenv("REDIS_SSL_CERT_REQS", "none")
```
pablonyx, line 169: Might be worth an enum?

pablodanswer (approved):

backend/danswer/configs/app_configs.py
rkuo-danswer: Because this is intended to be a passthrough value to redis and celery url construction, I think it is slightly more direct/clear to have the strings represented directly so that they they map directly to the documentation for redis/celery.  If we have a Danswer defined enum for this, it would add a small layer of complexity.

PR #2501, opened 2024-09-18
Move code block default language creation to citation processing

pablodanswer:
## Description
Our existing implementation to handle code blocks without any languages is to dynamically insert a `plaintext` tag as we process the content.

This causes issues with the way React Markdown processes content, yet we still need some way of inserting a language for code blocks that were created (by the LLM) without any (so that React Markdown can process the code block accordingly. )

This adds processing at the citation processing layer to insert the plaintext marker at the lowest level possible, so React markdown is **always** fed the correct version of the data.

## How Has This Been Tested?

## Accepted Risk

## Related Issue(s)

## Checklist:

### DIFF
--- a/backend/danswer/llm/answering/stream_processing/citation_processing.py
+++ b/backend/danswer/llm/answering/stream_processing/citation_processing.py
@@ -85,6 +85,15 @@ def extract_citations_from_stream(
         curr_segment += token
         llm_out += token
 
+        # Handle code blocks without language tags
+        if "`" in curr_segment:
+            if curr_segment.endswith("`"):
+                continue
+            elif "```" in curr_segment:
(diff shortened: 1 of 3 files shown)


backend/danswer/llm/answering/stream_processing/citation_processing.py
```diff
@@ -85,6 +85,15 @@ def extract_citations_from_stream(
         curr_segment += token
         llm_out += token
 
+        # Handle code blocks without language tags
+        if "`" in curr_segment:
```
rkuo-danswer, line 89: is it possible to write a test for this?

rkuo-danswer (changes requested):

PR #2510, opened 2024-09-19
Citations prompt for slack now includes thread history

hagen-danswer:
This adds the chat history to the prompt for slack messages. Worth looking at what other flows this impacts before we merge it though...


pablodanswer (approved):
LGTM!

PR #2525, opened 2024-09-20
Wrote tests for the chat apis

hagen-danswer:


backend/tests/integration/tests/dev_apis/test_simple_chat_api.py
```diff
@@ -60,3 +60,88 @@ def test_send_message_simple_with_history(reset: None) -> None:
         )
         assert found_doc
         assert found_doc["metadata"]["document_id"] == doc.id
+
```
hagen-danswer, line 63: this is the flow that ramp has described

backend/tests/integration/tests/dev_apis/test_knowledge_chat.py
```diff
@@ -0,0 +1,161 @@
...
+    assert 2 in response_json["final_context_doc_indices"]
+
+    # check that the cited documents are correct
+    assert cc_pair_1.documents[0].id in response_json["cited_documents"].values()
+
+    # check that the top documents are correct
+    assert response_json["top_documents"][0]["document_id"] == cc_pair_1.documents[0].id
+    print("response 1/3 passed")
+
+    # TESTING RESPONSE FOR QUESTION 2
```
Weves, line 90: would remove the other two

Weves: or we could do a "follow up" case

Weves (approved):

PR #2530, opened 2024-09-20
Fix duplicate users from slack / web

Weves:
If a user first logs in via web, then sends a message on Slack (and their slack email is capitalized differently), then they will have two user rows associated with them.

This fixes that.


pablodanswer (approved):
Might prefer to enforce this as a rule (a check constraint) when creating users and have an alembic migration fix previous users- otherwise, looks good.

PR #2542, opened 2024-09-23
Support svg navigation items

Weves:
^

Screenshot 2024-09-23 at 12 57 13 PM


backend/danswer/db/llm.py
```diff
@@ -64,19 +64,12 @@ def upsert_cloud_embedding_provider(
 def upsert_llm_provider(
     llm_provider: LLMProviderUpsertRequest,
     db_session: Session,
-    is_creation: bool = True,
```
Weves, line 67: moved to API-layer

PR #2544, opened 2024-09-23
Fix misc tool call errors

Weves:
Take 2. People are still running into this, and this is the last place where we could have been sending ToolCallKickoffs to the stream handler.

Tested by running a chat message and seeing that it streamed correctly + a chat which uses a tool.

### DIFF
--- a/web/src/middleware.ts
+++ b/web/src/middleware.ts
@@ -2,24 +2,29 @@ import { NextResponse } from "next/server";
 import type { NextRequest } from "next/server";
 import { SERVER_SIDE_ONLY__PAID_ENTERPRISE_FEATURES_ENABLED } from "./lib/constants";
 
+// NOTE: have to have the "/:path*" here since NextJS doesn't allow any real JS to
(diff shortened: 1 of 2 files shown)


web/src/middleware.ts
```diff
@@ -2,24 +2,29 @@ import { NextResponse } from "next/server";
 import type { NextRequest } from "next/server";
 import { SERVER_SIDE_ONLY__PAID_ENTERPRISE_FEATURES_ENABLED } from "./lib/constants";
 
+// NOTE: have to have the "/:path*" here since NextJS doesn't allow any real JS to
+// be run before the config is defined e.g. if we try and do a .map it will complain
 const eePaths = [
```
Weves, line 7: These changes fix 

```
 ⚠ Next.js can't recognize the exported `config` field in route "/src/middleware":
Unknown identifier "eePathsForMatcher" at "config.matcher".
The default config will be used instead.
```

pablodanswer (approved):

PR #2548, opened 2024-09-24
Allow = in tag filter

Weves:
Screenshot 2024-09-23 at 7 30 10 PM

Previously this would have given 0 results.

### DIFF
--- a/backend/danswer/db/tag.py
+++ b/backend/danswer/db/tag.py
@@ -1,3 +1,4 @@
+from sqlalchemy import and_
 from sqlalchemy import delete
 from sqlalchemy import func
 from sqlalchemy import or_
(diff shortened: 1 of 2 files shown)


backend/danswer/db/tag.py
```diff
@@ -113,6 +114,8 @@ def get_tags_by_value_prefix_for_source_types(
     sources: list[DocumentSource] | None,
     limit: int | None,
     db_session: Session,
+    # if set, both tag_key_prefix and tag_value_prefix must be a match
```
rkuo-danswer, line 117: Does the name of this function make sense since it's checking key or value or both?

rkuo-danswer (approved):

PR #2559, opened 2024-09-25
Cleaned up foreign key cleanup for user group deletion

hagen-danswer:


backend/ee/danswer/db/user_group.py
```diff
@@ -33,6 +34,93 @@
 logger = setup_logger()
```
hagen-danswer, line 36: just moved these to the top of the file

backend/ee/danswer/db/user_group.py
```diff
@@ -476,21 +528,6 @@ def update_user_group(
...
-            TokenRateLimit__UserGroup.user_group_id == user_group_id
-        )
-    ).all()
-    for (
-        token_rate_limit__user_group_relationship
-    ) in token_rate_limit__user_group_relationships:
-        db_session.delete(token_rate_limit__user_group_relationship)
-
-
 def prepare_user_group_for_deletion(db_session: Session, user_group_id: int) -> None:
```
hagen-danswer, line 531: This function now cleans up all 7 fk relationships for usergroups:

- credential
- user
- llm_provider
- document_set
- token_rate_limit
- persona
- connector_credential_pair

backend/ee/danswer/db/user_group.py
```diff
@@ -517,31 +569,12 @@ def prepare_user_group_for_deletion(db_session: Session, user_group_id: int) ->
...
-def _cleanup_document_set__user_group_relationships__no_commit(
-    db_session: Session, user_group_id: int
-) -> None:
-    """NOTE: does not commit the transaction."""
-    db_session.execute(
-        delete(DocumentSet__UserGroup).where(
-            DocumentSet__UserGroup.user_group_id == user_group_id
-        )
-    )
+def delete_user_group(db_session: Session, user_group: UserGroup) -> None:
```
hagen-danswer, line 572: removed duplicate fk cleanup  so this function is now super simple (but it assumes fk cleanup has already been performed

backend/ee/danswer/db/user_group.py
```diff
@@ -553,29 +586,6 @@ def mark_user_group_as_synced(db_session: Session, user_group: UserGroup) -> Non
     db_session.commit()
 
 
-def delete_user_group(db_session: Session, user_group: UserGroup) -> None:
```
hagen-danswer, line 556: this function was moved up to be near the fk_cleanup function

backend/ee/danswer/db/user_group.py
```diff
@@ -517,31 +569,12 @@ def prepare_user_group_for_deletion(db_session: Session, user_group_id: int) ->
...
-    db_session.execute(
-        delete(DocumentSet__UserGroup).where(
-            DocumentSet__UserGroup.user_group_id == user_group_id
-        )
-    )
+def delete_user_group(db_session: Session, user_group: UserGroup) -> None:
+    """
+    This assumes that all the fk cleanup has already been done.
+    """
+    db_session.delete(user_group)
```
rkuo-danswer, line 576: The old function had an explicit comment about flushing, but that doesn't seem to be getting handled here.  Any reason why?

rkuo-danswer (changes requested):
Possible to add tests for these relationships?  As an example I just injected an entry in the DB for connector deletion ... maybe we could do something similar here.

rkuo-danswer (approved):

PR #2561, opened 2024-09-25
Added a script to autogenerate cleanup functions for relationships

hagen-danswer:
## Description
There is discussion to be had if this is useful. I feel like efforts to ensure that all foreignkey relationships are deleted before deleting objects is very error prone as well as duplicated for each object. this provides a centralized code location to perform these actions

There is an argument to be made that we should just switch to cascade. Thats fine, this pr took like 15 minutes

### DIFF
--- a/backend/danswer/db/relationship_cleanup.py
+++ b/backend/danswer/db/relationship_cleanup.py
@@ -1,0 +1,894 @@
+from sqlalchemy.orm import Session
+
+from danswer.db.models import ChatMessage__SearchDoc
+from danswer.db.models import ChatMessage__StandardAnswer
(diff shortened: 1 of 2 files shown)


backend/scripts/generate_relationship_cleanup_functions.py
hagen-danswer, line 1: This generates the functions based on the class names. It looks for double underscore ("__") table names and the functions to delete those

backend/danswer/db/relationship_cleanup.py
hagen-danswer, line 1: referencing this instead of building these cleanup functions for each deletion should overall reduce lines of code/complexity

PR #2593, opened 2024-09-28
User service provisioning

pablodanswer:
## Description

Main changes
- Provision users in addition to tenant infra when we receive request from tenant service in control plane.
- If using multi-tenancy, we must refer to proper schema when referencing users (so, must refer to public `tenant user mapping` table if not authenticated)
- When inviting a user, must also allocate their tenant user mapping
- Separate alembic base for the public schema when using multi tenancy

`alembic upgrade head` and similar commands will still defaults to the typical alembic directory, but now you can specify `-n schema_private` to use the alternative migrations for the public schemas.


backend/danswer/configs/app_configs.py
```diff
@@ -396,3 +396,5 @@
 
 DATA_PLANE_SECRET = os.environ.get("DATA_PLANE_SECRET", "")
 EXPECTED_API_KEY = os.environ.get("EXPECTED_API_KEY", "")
+
+ENABLE_EMAIL_INVITES = os.environ.get("ENABLE_EMAIL_INVITES", "").lower() == "true"
```
yuhongsun96, line 400: Why do we have this one and REQUIRE_EMAIL_VERIFICATION?

backend/danswer/db/engine.py
```diff
@@ -225,27 +230,120 @@ def get_sqlalchemy_async_engine() -> AsyncEngine:
     return _ASYNC_ENGINE
 
 
-def get_session_with_tenant(tenant_id: str | None = None) -> Session:
+# Dependency to get the current tenant ID and set the context variable
+def get_current_tenant_id(request: Request) -> str:
```
yuhongsun96, line 235: I generally feel like it's more clear to separate this out as a None vs whatever the tenant id is. What do you think? Maybe it just introduces more repeated logic that has to be spread out everywhere.

Also is there now a tenant_public? How do we differentiate between no tenant / public schema vs tenant_public?

backend/danswer/dynamic_configs/store.py
```diff
@@ -56,12 +60,19 @@ def delete(self, key: str) -> None:
...
+                if tenant_id == "public":
+                    raise HTTPException(
+                        status_code=401, detail="User must authenticate"
+                    )
+                if not is_valid_schema_name(tenant_id):
+                    raise HTTPException(status_code=400, detail="Invalid tenant ID")
+                # Set the search_path to the tenant's schema
+                session.execute(text(f'SET search_path = "{tenant_id}"'))
             yield session
-        finally:
```
yuhongsun96, line 63: Why do we remove this?

backend/ee/danswer/server/tenants/provisioning.py
```diff
@@ -61,3 +63,46 @@ def ensure_schema_exists(tenant_id: str) -> bool:
                 db_session.execute(stmt)
                 return True
             return False
+
+
+def user_owns_a_tenant(email: str) -> bool:
```
yuhongsun96, line 70: this seems to check that the user is part of a tenant, does own imply being the first admin or something else?

web/src/app/auth/create-account/page.tsx
```diff
@@ -0,0 +1,45 @@
+"use client";
```
yuhongsun96, line 1: Can we gate this so users don't visit this? If not, it doesn't really matter :P

yuhongsun96 (approved):

backend/danswer/dynamic_configs/store.py
pablonyx: Remember why I removed this - `with session` should handle this robustly for us

web/src/app/auth/create-account/page.tsx
pablonyx: Added comment!

PR #2596, opened 2024-09-29
Add tenant context

pablodanswer:
## Description
Initial addition of tenant context to background processing/tasks
- Upcoming PR updates provisioning to ensure all tenants get background processing tasks

### DIFF
--- a/backend/danswer/background/celery/celery_app.py
+++ b/backend/danswer/background/celery/celery_app.py
@@ -71,7 +71,6 @@ def celery_task_postrun(
         return
 
     task_logger.debug(f"Task {task.name} (ID: {task_id}) completed with state: {state}")
-    # logger.debug(f"Result: {retval}")
 
     if state not in READY_STATES:
         return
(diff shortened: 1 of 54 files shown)


backend/danswer/background/celery/celery_app.py
```diff
@@ -418,48 +419,65 @@ def stop(self, worker: Any) -> None:
...
+    celery_app.conf.beat_schedule = {}
+
+    for tenant_id in tenants:
+        celery_app.conf.beat_schedule.update(
+            {
+                "check-for-vespa-sync": {
+                    "task": "check_for_vespa_sync_task",
+                    "schedule": timedelta(seconds=5),
+                    "options": {"priority": DanswerCeleryPriority.HIGH},
+                    "args": (tenant_id,),
```
yuhongsun96: What is the upside/downside of doing this per tenant vs iterating over the tenants in the actual task?

backend/danswer/background/update.py
```diff
@@ -392,13 +405,35 @@ def kickoff_indexing_jobs(
     return existing_jobs_copy
 
 
+def get_all_tenant_ids() -> list[str] | list[None]:
```
yuhongsun96, line 415: Should we add some prefix for the tenants to double check they're danswer tenants? In case there is other stuff in the DB (which there probably shouldn't be it's better to be safe).

yuhongsun96 (approved):

PR #2600, opened 2024-09-29
Fix default assistant

Weves:
https://linear.app/danswer/issue/DAN-726/fix-bug-where-private-assistants-could-not-be-made-as-default

### DIFF
--- a/web/src/app/chat/ChatPage.tsx
+++ b/web/src/app/chat/ChatPage.tsx
@@ -162,6 +162,9 @@ export function ChatPage({
     user,
     availableAssistants
   );
+  const finalAssistants = user
+    ? orderAssistantsForUser(visibleAssistants, user)
+    : visibleAssistants;
 
   const existingChatSessionAssistantId = selectedChatSession?.persona_id;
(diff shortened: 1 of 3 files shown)


web/src/lib/chat/fetchChatData.ts
```diff
@@ -188,14 +188,6 @@ export async function fetchChatData(searchParams: {
     !hasAnyConnectors &&
     (!user || user.role === "admin");
 
-  const shouldDisplaySourcesIncompleteModal =
```
Weves, line 191: not used

web/src/app/chat/ChatPage.tsx
```diff
@@ -216,7 +219,7 @@ export function ChatPage({
   const liveAssistant =
     alternativeAssistant ||
     selectedAssistant ||
-    visibleAssistants[0] ||
+    finalAssistants[0] ||
```
Weves, line 222: liveAssistant wasn't taking into account the users specific ordering

pablodanswer (approved):

PR #2602, opened 2024-09-30
Added permissions syncing for slack

hagen-danswer:


backend/danswer/connectors/slack/connector.py
```diff
@@ -38,47 +36,18 @@
 # list of messages in a thread
 ThreadType = list[MessageType]
 
-basic_retry_wrapper = retry_builder()
-
-
```
hagen-danswer, line 43: moved this stuff to the utils file

backend/danswer/connectors/slack/connector.py
```diff
@@ -88,19 +57,38 @@ def _get_channels(
 def get_channels(
     client: WebClient,
     exclude_archived: bool = True,
```
hagen-danswer, line 59: changed this function to be able to request public or private channels separately 

added defaults to keep behavior identical

backend/danswer/connectors/slack/connector.py
```diff
@@ -328,7 +316,44 @@ def get_all_docs(
         )
```
hagen-danswer, line 318: Added this to make it an order of magnitude faster than running the whole connector each  time

backend/danswer/connectors/slack/connector.py
```diff
@@ -349,14 +374,24 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
         self.client = WebClient(token=bot_token)
         return None
 
+    def retrieve_all_source_ids(self) -> set[str]:
```
hagen-danswer, line 377: This means pruning will run an order of magnitude faster now as well

backend/ee/danswer/external_permissions/slack/group_sync.py
```diff
@@ -0,0 +1,89 @@
+"""
```
hagen-danswer, line 1: this comment is correct. Just keeping this here in casse we add support for group based stuff in the future. It's tested to work

backend/ee/danswer/external_permissions/slack/utils.py
```diff
@@ -0,0 +1,18 @@
+from slack_sdk import WebClient
+
+from danswer.connectors.slack.connector import make_paginated_slack_api_call_w_retries
+
+
```
hagen-danswer, line 5: this exists to create a map of all slack user ids to emails because the other calls return the slack user ids but we need the emails

backend/ee/danswer/external_permissions/slack/doc_sync.py
```diff
@@ -0,0 +1,187 @@
...
+        doc_id (str): The document ID string.
+
+    Returns:
+        str: The extracted channel ID.
+
+    Raises:
+        ValueError: If the doc_id doesn't contain the expected separator.
+    """
+    try:
+        channel_id, _ = doc_id.split("__", 1)
```
yuhongsun96, line 36: Try not to have this type of magical logic. Ideally if we can put the channel in the doc metadata and pull it from there and throw an exception if it's not found or something. That would be ideal (this can be a backoff check for messages that don't have a channel in the metadata)

hagen-danswer: I agree, that is why i commented thouroughly. The problem is the ID only connector. But maybe we shoudl have a version of the connector that retrieves id+metadata? probably worth a discussion. Regardless, the rework wouldnt be that difficult if we delay the decision

yuhongsun96 (approved):
didn't read too carefully, be sure to test!

PR #2603, opened 2024-09-30
Redis Cache for KV Store

yuhongsun96:
## Description

## How Has This Been Tested?

## Accepted Risk

## Related Issue(s)

## Checklist:

### DIFF
--- a/backend/danswer/background/celery/celery_app.py
+++ b/backend/danswer/background/celery/celery_app.py
@@ -30,7 +30,7 @@ from danswer.configs.constants import POSTGRES_CELERY_WORKER_HEAVY_APP_NAME
 from danswer.configs.constants import POSTGRES_CELERY_WORKER_LIGHT_APP_NAME
 from danswer.configs.constants import POSTGRES_CELERY_WORKER_PRIMARY_APP_NAME
 from danswer.db.engine import SqlEngine
-from danswer.redis.redis_pool import RedisPool
+from danswer.redis.redis_pool import get_redis_client
 from danswer.utils.logger import ColoredFormatter
 from danswer.utils.logger import PlainFormatter
 from danswer.utils.logger import setup_logger
@@ -39,8 +39,6 @@ logger = setup_logger()
 
 # use this within celery tasks to get celery task specific logging
 task_logger = get_task_logger(__name__)
-
-redis_pool = RedisPool()
 
 celery_app = Celery(__name__)
 celery_app.config_from_object(
@@ -79,13 +77,13 @@ def celery_task_postrun(
     if not task_id:
         return
 
+    r = get_redis_client()
+
     if task_id.startswith(RedisConnectorCredentialPair.PREFIX):
-        r = redis_pool.get_client()
         r.srem(RedisConnectorCredentialPair.get_taskset_key(), task_id)
         return
 
     if task_id.startswith(RedisDocumentSet.PREFIX):
-        r = redis_pool.get_client()
         document_set_id = RedisDocumentSet.get_id_from_task_id(task_id)
         if document_set_id is not None:
             rds = RedisDocumentSet(document_set_id)
@@ -93,7 +91,6 @@ def celery_task_postrun(
         return
 
     if task_id.startswith(RedisUserGroup.PREFIX):
-        r = redis_pool.get_client()
         usergroup_id = RedisUserGroup.get_id_from_task_id(task_id)
         if usergroup_id is not None:
             rug = RedisUserGroup(usergroup_id)
@@ -101,7 +98,6 @@ def celery_task_postrun(
         return
 
     if task_id.startswith(RedisConnectorDeletion.PREFIX):
-        r = redis_pool.get_client()
         cc_pair_id = RedisConnectorDeletion.get_id_from_task_id(task_id)
         if cc_pair_id is not None:
             rcd = RedisConnectorDeletion(cc_pair_id)
@@ -130,7 +126,7 @@ def on_worker_init(sender: Any, **kwargs: Any) -> None:
         SqlEngine.set_app_name(POSTGRES_CELERY_WORKER_PRIMARY_APP_NAME)
         SqlEngine.init_engine(pool_size=8, max_overflow=0)
 
-    r = redis_pool.get_client()
+    r = get_redis_client()
 
     WAIT_INTERVAL = 5
     WAIT_LIMIT = 60
@@ -190,7 +186,7 @@ def on_worker_init(sender: Any, **kwargs: Any) -> None:
 
     # This is singleton work that should be done on startup exactly once
     # by the primary worker
-    r = redis_pool.get_client()
+    r = get_redis_client()
 
     # For the moment, we're assuming that we are the only primary worker
     # that should be running.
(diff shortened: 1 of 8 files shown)


backend/danswer/background/celery/celery_app.py
```diff
@@ -40,8 +40,6 @@
 # use this within celery tasks to get celery task specific logging
 task_logger = get_task_logger(__name__)
 
-redis_pool = RedisPool()
```
yuhongsun96, line 43: Is there a benefit to instantiating it in the different modules? It's a singleton anyway right? Can we just do it from the one central place?

backend/danswer/dynamic_configs/store.py
```diff
@@ -64,8 +27,9 @@ def get_session(self) -> Iterator[Session]:
             session.close()
 
     def store(self, key: str, val: JSON_ro, encrypt: bool = False) -> None:
-        # The actual encryption/decryption is done in Postgres, we just need to choose
-        # which field to set
+        # Not encrypted in Redis, but encrypted in Postgres
```
yuhongsun96: should be safe right?

backend/danswer/dynamic_configs/store.py
```diff
@@ -82,19 +46,36 @@ def store(self, key: str, val: JSON_ro, encrypt: bool = False) -> None:
             session.commit()
 
     def load(self, key: str) -> JSON_ro:
+        redis_value = self.redis_client.get(key)
+        if redis_value:
+            if isinstance(redis_value, bytes):
+                return json.loads(redis_value.decode("utf-8"))
```
yuhongsun96: I imagine utf-8 should cover everything? Any concerns here?

backend/danswer/background/celery/celery_app.py
rkuo-danswer: No benefit, just how it started out. should be fine to make some sort of static or class method.

backend/danswer/redis/redis_pool.py
```diff
@@ -69,6 +69,13 @@ def create_pool(
         )
 
 
+redis_pool = RedisPool()
+
+
+def get_redis_client() -> Redis:
```
rkuo-danswer, line 75: This is OK for the moment, but I will probably make this a class method shortly.

backend/danswer/dynamic_configs/store.py
```diff
@@ -64,8 +27,9 @@ def get_session(self) -> Iterator[Session]:
             session.close()
 
     def store(self, key: str, val: JSON_ro, encrypt: bool = False) -> None:
-        # The actual encryption/decryption is done in Postgres, we just need to choose
-        # which field to set
+        # Not encrypted in Redis, but encrypted in Postgres
+        self.redis_client.set(key, json.dumps(val))
```
rkuo-danswer: By default strings will be encoded as UTF-8 so this should be OK paired with the decode utf-8 on load.

backend/danswer/dynamic_configs/store.py
```diff
@@ -1,59 +1,22 @@
...
-        if not file_path.exists():
-            raise ConfigNotFoundError
-        lock = _get_file_lock(file_path)
-        with lock.acquire(timeout=FILE_LOCK_TIMEOUT):
-            os.remove(file_path)
+from danswer.redis.redis_pool import get_redis_client
 
 
 class PostgresBackedDynamicConfigStore(DynamicConfigStore):
+    def __init__(self) -> None:
```
rkuo-danswer: Should we continue to name this PostgresBacked if it's Redis backed now?

backend/danswer/dynamic_configs/store.py
rkuo-danswer: What are we looking for in terms of "safe"?  It's possible to encrypt the connection to redis and set a password for the redis instance so that's arguably pretty "safe".  None of this is currently persisted to disk so that's also "safe".  Are there other considerations for us to look at?

backend/danswer/dynamic_configs/store.py
rkuo-danswer: Looks OK to me.

backend/danswer/dynamic_configs/store.py
```diff
@@ -82,19 +46,36 @@ def store(self, key: str, val: JSON_ro, encrypt: bool = False) -> None:
             session.commit()
 
     def load(self, key: str) -> JSON_ro:
+        redis_value = self.redis_client.get(key)
+        if redis_value:
+            if isinstance(redis_value, bytes):
+                return json.loads(redis_value.decode("utf-8"))
+            elif isinstance(redis_value, str):
```
rkuo-danswer: Doesn't redis_client.get always return type bytes?

rkuo-danswer (commented):
Added several comments, could use some context on what the original key value store was doing and what we're trying to accomplish with the switch.  its important to note that we're currently using redis as an ephemeral data store so that's a significant difference from storing things in postgres.

backend/danswer/dynamic_configs/store.py
yuhongsun96: Seems good enough, just wanted to double check

backend/danswer/dynamic_configs/store.py
yuhongsun96: Yup ok sounds good

backend/danswer/dynamic_configs/store.py
yuhongsun96: Haha, I'll rename it

backend/danswer/dynamic_configs/store.py
yuhongsun96: Ya, looked closer, seems good, I dont think it should ever be str

backend/danswer/key_value_store/store.py
```diff
@@ -0,0 +1,79 @@
...
+        factory = get_session_factory()
+        session: Session = factory()
+        try:
+            yield session
+        finally:
+            session.close()
+
+    def store(self, key: str, val: JSON_ro, encrypt: bool = False) -> None:
+        # Not encrypted in Redis, but encrypted in Postgres
+        self.redis_client.set(REDIS_KEY_PREFIX + key, json.dumps(val))
```
rkuo-danswer: we should handle a potential RedisError exception here.  Also note that without using a TTL the error handling gets significantly trickier as we run into potential consistency issues that won't resolve themselves if there is a spurious exception.

backend/danswer/key_value_store/store.py
```diff
@@ -0,0 +1,79 @@
...
+            else:
+                obj = KVStore(
+                    key=key, value=plain_val, encrypted_value=encrypted_val
+                )  # type: ignore
+                session.query(KVStore).filter_by(key=key).delete()  # just in case
+                session.add(obj)
+            session.commit()
+
+    def load(self, key: str) -> JSON_ro:
+        redis_value = self.redis_client.get(REDIS_KEY_PREFIX + key)
```
rkuo-danswer: handle exception here as well (this is easier than writing, just fall back to the DB)

backend/danswer/key_value_store/store.py
```diff
@@ -0,0 +1,79 @@
...
+                raise KvKeyNotFoundError
+
+            if obj.value is not None:
+                value = obj.value
+            elif obj.encrypted_value is not None:
+                value = obj.encrypted_value
+            else:
+                value = None
+
+            self.redis_client.set(REDIS_KEY_PREFIX + key, json.dumps(value))
```
rkuo-danswer: handle exception

rkuo-danswer (changes requested):

rkuo-danswer (approved):

PR #2658, opened 2024-10-02
Multi tenant vespa

pablodanswer:
## Description

## How Has This Been Tested?

## Accepted Risk

## Related Issue(s)

## Checklist:

### DIFF
--- a/backend/danswer/document_index/vespa/index.py
+++ b/backend/danswer/document_index/vespa/index.py
@@ -10,9 +10,10 @@ from datetime import timedelta
 from typing import BinaryIO
 from typing import cast
 
-import httpx
-import requests
+import httpx  # type: ignore
+import requests  # type: ignore
 
+from danswer.configs.app_configs import MULTI_TENANT
 from danswer.configs.chat_configs import DOC_TIME_DECAY
 from danswer.configs.chat_configs import NUM_RETURNED_HITS
 from danswer.configs.chat_configs import TITLE_CONTENT_RATIO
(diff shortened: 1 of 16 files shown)


backend/danswer/document_index/vespa/index.py
```diff
@@ -573,3 +667,158 @@ def admin_retrieval(
...
+
+                if not hits:
+                    break
+
+                for hit in hits:
+                    doc_id = hit.get("id")
+                    if doc_id:
+                        document_ids.append(doc_id)
+
+                offset += limit  # Move to the next page
```
yuhongsun96, line 755: Fine for now but we should add some performance logging and also a comment here about the risk. I'm not certain that the pagination in Vespa is handled ideally. It may be slower the large the offset (as in it still may need to process the hits before this "page").

backend/danswer/document_index/vespa/index.py
```diff
@@ -119,6 +125,12 @@ def ensure_indices_exist(
         index_embedding_dim: int,
         secondary_index_embedding_dim: int | None,
     ) -> None:
+        if MULTI_TENANT:
+            logger.info(
+                "Skipping Vespa index seup for multitenant (would wipe all indices)"
```
yuhongsun96, line 130: typo

yuhongsun96: Also add a TODO note here to add another layer of checks later on to ensure we don't easily wipe our MT cloud indices for everyone

backend/danswer/document_index/vespa/index.py
```diff
@@ -187,6 +199,88 @@ def ensure_indices_exist(
...
+        services = services_template.replace(DOCUMENT_REPLACEMENT_PAT, doc_lines)
+        services = services.replace(
+            SEARCH_THREAD_NUMBER_PAT, str(VESPA_SEARCHER_THREADS)
+        )
+
+        kv_store = get_dynamic_config_store()
+
+        needs_reindexing = False
+        try:
+            needs_reindexing = cast(bool, kv_store.load(KV_REINDEX_KEY))
```
yuhongsun96, line 236: This doesn't have tenant context here, how should it be handled? (also the n-gram thing below doesn't make sense here)

backend/danswer/document_index/vespa/index.py
```diff
@@ -187,6 +199,88 @@ def ensure_indices_exist(
                 f"Failed to prepare Vespa Danswer Index. Response: {response.text}"
             )
 
+    @staticmethod
+    def register_multitenant_indices(
```
yuhongsun96, line 203: If it's on kubernetes and it spins up another container for the api server, won't it recreate the indices, so if someone sets a new non-default model, it will get wiped? We don't need to support this for now but maybe we should make a note of it.

On the branch we deploy for the cloud, we might want to just choose a model for them and not let them specify their own.

yuhongsun96 (approved):
Small things to change but I don't think I need to take another pass at it after so approving :)

PR #2667, opened 2024-10-03
Billing

pablodanswer:
1. JWT-based authentication:
   - Uses `SECRET_JWT_KEY` for encrypting user/tenant context
   - Employs `JWT_ALGORITHM` (HS256) for token signing

2. Inter-service communication security:
   - `DATA_PLANE_SECRET` secures control plane and data plane communication
   - Short-lived tokens generated with specific scopes and expiration times

3. API access control:
   - `EXPECTED_API_KEY` required in headers for control plane API access
   - Dependency checks for valid API key and JWT in request headers

4. Stripe integration for secure billing:
   - Stripe API key used for subscription management
   - Tenant IDs stored in Stripe metadata for correlation


backend/Dockerfile.cloud
```diff
@@ -0,0 +1,109 @@
+FROM python:3.11.7-slim-bookworm
```
yuhongsun96, line 1: A lot of the comments and things are not correct. I would say this just should not be used by the community.

backend/Dockerfile.cloud
```diff
@@ -0,0 +1,109 @@
...
+
+LABEL com.danswer.maintainer="founders@danswer.ai"
+LABEL com.danswer.description="This image is the web/frontend container of Danswer which \
```
yuhongsun96: A lot of the comments and things are not correct. I would say this just should not be used by the community.

backend/danswer/server/auth_check.py
```diff
@@ -4,13 +4,13 @@
...
 from starlette.routing import BaseRoute
 
-from danswer.auth.users import control_plane_dep
 from danswer.auth.users import current_admin_user
 from danswer.auth.users import current_curator_or_admin_user
 from danswer.auth.users import current_user
 from danswer.auth.users import current_user_with_expired_token
 from danswer.configs.app_configs import APP_API_PREFIX
 from danswer.server.danswer_api.ingestion import api_key_dep
+from ee.danswer.server.tenants.access import control_plane_dep
```
yuhongsun96, line 13: We should not import ee modules from the MIT stuff. It confuses the separation between the versions.

yuhongsun96: There are other places this is done but we should address those.

backend/danswer/server/manage/users.py
```diff
@@ -60,6 +63,7 @@
 from ee.danswer.db.api_key import is_api_key_email_address
 from ee.danswer.db.external_perm import delete_user__ext_group_for_user__no_commit
 from ee.danswer.db.user_group import remove_curator_status__no_commit
+from ee.danswer.server.tenants.billing import register_tenant_users
```
yuhongsun96, line 66: same here and around this line, it should not import from ee

backend/ee/danswer/server/tenants/api.py
```diff
@@ -60,3 +74,53 @@ def create_tenant(
...
+@router.post("/product-gating")
+def gate_product(
+    product_gating_request: ProductGatingRequest, _: None = Depends(control_plane_dep)
+) -> None:
+    """
+    Gating the product means that the product is not available to the tenant.
+    They will be directed to the billing page.
+    We gate the product when
+    1) User has ended free trial without adding payment method
+    2) User's card has declined
```
yuhongsun96, line 88: We can let them use it and give a warning if the card declined. Hubspot does this and I appreciated that and paid them :P

backend/ee/danswer/server/tenants/api.py
```diff
@@ -60,3 +74,53 @@ def create_tenant(
...
+    """
+    if not MULTI_TENANT:
```
yuhongsun96: Is there a cleaner/easier way to check this at the relevant endpoints/modules than this? I think we would need to add this to all the MT stuff and it's 1. very easy to miss some, 2. poor code reuse

If there's not a better way then it's whatever

yuhongsun96: Also for this one, if we do single tenant cloud I guess it's not needed? I'm assuming with those we'd have an enterprise contract?

deployment/docker_compose/docker-compose.prod-cloud.yml
```diff
@@ -0,0 +1,243 @@
+services:
```
yuhongsun96, line 1: Why do we need this?

yuhongsun96 (approved):
If it works :P

backend/Dockerfile.cloud
yuhongsun96: let's remove this one

deployment/docker_compose/docker-compose.prod-cloud.yml
yuhongsun96: let's remove this

backend/ee/danswer/server/tenants/api.py
pablonyx: Think a better way is to conditionally add router to the main lifecycle– updated to reflect

PR #2671, opened 2024-10-03
update_single should be optimized for a single call now

rkuo-danswer:
## Description
Improves vespa update performance.
backports some small fixes for issues that were fixed in a larger pending PR 

## How Has This Been Tested?

## Accepted Risk

## Related Issue(s)

## Checklist:

### DIFF
--- a/backend/danswer/document_index/interfaces.py
+++ b/backend/danswer/document_index/interfaces.py
@@ -56,6 +56,21 @@ class DocumentMetadata:
 
 
 @dataclass
+class VespaDocumentFields:
+    """
+    Specifies fields in Vespa for a document.  Fields set to None will be ignored.
+    Perhaps we should name this in an implementation agnostic fashion, but it's more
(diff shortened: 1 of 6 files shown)


backend/danswer/document_index/interfaces.py
```diff
@@ -55,6 +55,21 @@ class DocumentMetadata:
     from_ingestion_api: bool = False
 
 
+@dataclass
+class VespaDocumentFields:
```
Weves, line 59: shoudl we also update `UpdateRequest` to take in one of these? That way there's only one source of truth for what can be updated

rkuo-danswer: This definitely makes sense! Let me add it as a TODO to one of my linear issues because when the pruning PR is in, I believe this can probably be refactored away completely.

Weves (approved):

PR #2686, opened 2024-10-04
docker deployment cors update

pablodanswer:
Prevent empty string from being CORS

### DIFF
--- a/backend/shared_configs/configs.py
+++ b/backend/shared_configs/configs.py
@@ -1,4 +1,5 @@
 import os
+from typing import List
 from urllib.parse import urlparse
 
 # Used for logging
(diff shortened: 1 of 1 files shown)


backend/shared_configs/configs.py
```diff
@@ -83,9 +83,15 @@ def validate_cors_origin(origin: str) -> None:
         raise ValueError(f"Invalid CORS origin: '{origin}'")
 
 
-CORS_ALLOWED_ORIGIN = os.environ.get("CORS_ALLOWED_ORIGIN", "*").split(",") or ["*"]
+CORS_ALLOWED_ORIGIN = os.environ.get("CORS_ALLOWED_ORIGIN", "*")
```
rkuo-danswer: Let's add some comments here about:

1. Examples of valid values for the env var.
2. What this code is trying to do ... it looks like it's converting the env string to a list and stripping whitespace and/or defaulting to "*"?

backend/shared_configs/configs.py
```diff
@@ -83,9 +83,15 @@ def validate_cors_origin(origin: str) -> None:
...
+if CORS_ALLOWED_ORIGIN:
+    CORS_ALLOWED_ORIGIN = [
```
rkuo-danswer, line 98: It seems like the subsequent validation loop should just go in here since if we default to "*" that should never need validation.

rkuo-danswer (changes requested):

backend/shared_configs/configs.py
pablonyx: Generally think that environment variables should be explained in the docs as much as possible rather than in the code that parses them, but I'll add a comment explaining the proper structure

rkuo-danswer (approved):

PR #2687, opened 2024-10-05
disabled llm when skip_gen_ai_answer_question set

evan-danswer:
## Description
the /search/answer-with-quote endpoint has long had a skip_gen_ai_answer_question flag that did not do anything; now it correctly skips only the code portions that query an LLM.

## How Has This Been Tested?
I ran my local version of the backend with `assert False` in the functions in interfaces.py that wrap all LLM invocations. Confirmed that the assertions are hit without my changes and that they are not hit with my changes. The run_qa scripts in backend/tests/regression/answer_quality completes much faster too: 65 seconds to retrieve docs for 664 questions.

Wrote a unit test for the backend's use of the `skip_gen_ai_answer_question` flag; confirmed that LLM `stream()` is called when it is off and not called when it is on.

## Accepted Risk

## Related Issue(s)

## Checklist:

### DIFF
--- a/backend/danswer/llm/answering/answer.py
+++ b/backend/danswer/llm/answering/answer.py
@@ -311,13 +311,13 @@ class Answer:
                     )
                 )
             yield tool_runner.tool_final_result()
+            if not self.skip_gen_ai_answer_generation:
+                prompt = prompt_builder.build(tool_call_summary=tool_call_summary)
 
-            prompt = prompt_builder.build(tool_call_summary=tool_call_summary)
-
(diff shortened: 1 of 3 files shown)


yuhongsun96 (approved):
lgtm, one thing to add to the search eval pipeline is that we should add the DISABLE_GENERATIVE_AI option (on by default) and the pipeline should still work.

The checks here are correct but there is still quite a bit of LLM related stuff that's still turned on during the flow. ie. we pass every document section to the LLM to determine if it's useful or not and this is quite token heavy.

PR #2689, opened 2024-10-05
add vespa + embedding timeout env variables

pablodanswer:
## Description

## How Has This Been Tested?

## Accepted Risk

## Related Issue(s)

## Checklist:

### DIFF
--- a/backend/danswer/configs/app_configs.py
+++ b/backend/danswer/configs/app_configs.py
@@ -401,6 +401,8 @@ CUSTOM_ANSWER_VALIDITY_CONDITIONS = json.loads(
     os.environ.get("CUSTOM_ANSWER_VALIDITY_CONDITIONS", "[]")
 )
 
+VESPA_REQUEST_TIMEOUT = int(os.environ.get("VESPA_REQUEST_TIMEOUT", 0)) or None
+
 
 #####
 # Enterprise Edition Configs
--- a/backend/danswer/configs/app_configs.py
+++ b/backend/danswer/configs/app_configs.py
@@ -401,7 +401,8 @@ CUSTOM_ANSWER_VALIDITY_CONDITIONS = json.loads(
     os.environ.get("CUSTOM_ANSWER_VALIDITY_CONDITIONS", "[]")
 )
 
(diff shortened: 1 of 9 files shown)


backend/danswer/main.py
```diff
@@ -353,6 +355,11 @@ def setup_vespa(
 
 @asynccontextmanager
 async def lifespan(app: FastAPI) -> AsyncGenerator:
+    # Set recursion limit
+    if SYSTEM_RECURSION_LIMIT is not None:
```
yuhongsun96, line 146: Why are we hitting these issues?

backend/danswer/configs/app_configs.py
```diff
@@ -396,6 +396,10 @@
 )
 
 
+VESPA_REQUEST_TIMEOUT = int(os.environ.get("VESPA_REQUEST_TIMEOUT") or "0") or None
```
yuhongsun96: Who is complaining about this? I mean this is fine to do but I am concerned that this is masking other issues

yuhongsun96: Would this also correctly be a no-op in the case where it's not set? I think the default is like 5 seconds or something right? Would this be correct if it were None?

deployment/docker_compose/docker-compose.dev.yml
```diff
@@ -74,6 +74,9 @@ services:
       - DISABLE_RERANK_FOR_STREAMING=${DISABLE_RERANK_FOR_STREAMING:-}
       - MODEL_SERVER_HOST=${MODEL_SERVER_HOST:-inference_model_server}
       - MODEL_SERVER_PORT=${MODEL_SERVER_PORT:-}
+      - VESPA_REQUEST_TIMEOUT=${VESPA_REQUEST_TIMEOUT:-}
```
yuhongsun96: If we're going to add these to the deployment files, we should add it to all of them and all of the different ways to deploy. I think in this case if it's a single customer, we can just tell them to add it themselves, should be ok hopefully

yuhongsun96 (approved):
Let's just remove it from the deployment files, the rest is good

PR #2700, opened 2024-10-06
Remove unnecessary hard limit on semantic id in vespa query

pablodanswer:
## Description

## How Has This Been Tested?

## Accepted Risk

## Related Issue(s)

## Checklist:

### DIFF
--- a/backend/danswer/document_index/vespa/chunk_retrieval.py
+++ b/backend/danswer/document_index/vespa/chunk_retrieval.py
@@ -107,7 +107,6 @@ def _vespa_hit_to_inference_chunk(
         dynamic_summary=hit["fields"].get(CONTENT_SUMMARY, hit["fields"][CONTENT]),
     )
 
-    # Remove semantic_identifier from fields
     semantic_identifier = fields.get(SEMANTIC_IDENTIFIER, "")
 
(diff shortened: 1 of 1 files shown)


backend/danswer/document_index/vespa/chunk_retrieval.py
```diff
@@ -130,7 +132,7 @@ def _vespa_hit_to_inference_chunk(
         document_id=fields[DOCUMENT_ID],
         source_type=fields[SOURCE_TYPE],
         title=fields.get(TITLE),
-        semantic_identifier=fields[SEMANTIC_IDENTIFIER],
+        semantic_identifier=fields.get(SEMANTIC_IDENTIFIER, ""),
```
yuhongsun96, line 135: Why would this ever be empty or missing? This is probably one of those cases we should fail loudly. If a doc has no semantic identifier, we have no way of showing it to the user in the UI. We wouldn't know what to call it. Better to actually understand what is causing this

pablonyx: Agreed– I don't think we should merge this PR, but needed a quick hotfix for user who was having a bad time with wonky Vespa instance

PR #2756, opened 2024-10-10
Fix slack responses

Weves:
https://linear.app/danswer/issue/DAN-800/fix-slack-bot-follow-up-questions


rkuo-danswer (approved):

andrewscarani:
Follow up questions to the slack bot seem to work in DMs with this fix, but not in threads in other channels...

PR #2790, opened 2024-10-14
fresh indexing feature branch

rkuo-danswer:
## Description

## How Has This Been Tested?

## Accepted Risk

## Related Issue(s)

## Checklist:


backend/danswer/background/celery/tasks/pruning/tasks.py
```diff
@@ -89,24 +91,29 @@ def ccpair_pruning_generator_task_creation_helper(
...
+    if cc_pair.status != ConnectorCredentialPairStatus.ACTIVE:
+        return False
 
     # skip pruning if the next scheduled prune time hasn't been reached yet
     last_pruned = cc_pair.last_pruned
     if not last_pruned:
-        # if never pruned, use the connector time created as the last_pruned time
-        last_pruned = cc_pair.connector.time_created
+        if not cc_pair.last_successful_index_time:
+            return False
```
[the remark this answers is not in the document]
rkuo-danswer: can't prune if nothing has been indexed.

backend/danswer/document_index/vespa/indexing_utils.py
```diff
@@ -200,6 +202,7 @@ def batch_index_vespa_chunks(
     chunks: list[DocMetadataAwareIndexChunk],
     index_name: str,
     http_client: httpx.Client,
+    multitenant: bool,
```
pablonyx, line 205: Is there reason we're passing multitenant down instead of using the environment variable directly in the couple of places where it's necessary?

backend/danswer/background/celery/celery_app.py
```diff
@@ -130,6 +168,31 @@ def on_worker_init(sender: Any, **kwargs: Any) -> None:
     elif hostname.startswith("heavy"):
         SqlEngine.set_app_name(POSTGRES_CELERY_WORKER_HEAVY_APP_NAME)
         SqlEngine.init_engine(pool_size=8, max_overflow=0)
+    elif hostname.startswith("indexing"):
+        SqlEngine.set_app_name(POSTGRES_CELERY_WORKER_INDEXING_APP_NAME)
+        SqlEngine.init_engine(pool_size=8, max_overflow=0)
+
+        # TODO: why is this necessary for the indexer to do?
+        engine = SqlEngine.get_engine()
```
pablonyx: How are we ensuring we pass tenant context here?

backend/danswer/document_index/vespa/indexing_utils.py
rkuo-danswer: Without writing a whole essay on the subject, global vars and state are widely considered bad practice, so this is just a tiny step in the right direction.  We actually need to start refactoring out the overuse of global config variables out of the codebase.

https://android.googlesource.com/platform/external/google-styleguide/+/refs/tags/android-s-beta-2/pyguide.md#2_5-global-variables

pablonyx: I understand that- what I meant was more along the lines of using the config where it's necessary for kicking off the initial tasks and letting the `tenant_id` value propagate through the various background tenant tasks (which should be `None` if multitenancy is not enabled), rather than passing an additional boolean to each task. 

But looking at the code again, it looks like it's only passed to Vespa for initialization and used for indexing, so no real complaints here.

backend/danswer/background/celery/celery_app.py
rkuo-danswer: think that was a bug, should be fixed now.

backend/supervisord.conf
```diff
@@ -46,7 +35,7 @@ stopasgroup=true
 [program:celery_worker_light]
 command=bash -c "celery -A danswer.background.celery.celery_run:celery_app worker \
     --pool=threads \
-    --concurrency=${CELERY_WORKER_LIGHT_CONCURRENCY:-24} \
+    --concurrency=${CELERY_WORKER_LIGHT_CONCURRENCY:-${NUM_INDEXING_WORKERS:-24}} \
```
Weves: this seems wrong, I think this should be replacing `CELERY_WORKER_INDEXING_CONCURRENCY` below, right?

Weves (changes requested):

backend/supervisord.conf
rkuo-danswer: that's right, will fix.

Weves (approved):

PR #2793, opened 2024-10-14
enable and run contextual rag (search untested)

evan-danswer:
## Description
implements Contextual RAG across chunk sizes with a UI flag or environment variable (in theory, although I think the env var code path doesn't work for either this or multipass indexing)

## How Has This Been Tested?
Indexing has been manually tested, search + retrieval has not. Also working on unit tests.

## Accepted Risk
The main risk is that this option adds significant cost for currently untested benefits
It could be fine if we call it a "beta test" feature and just make it available to get community feedback, but before that I need to test that it doesn't break the backend

## Related Issue(s)

## Checklist:


evan-danswer:
old version of contextual retrieval

PR #2801, opened 2024-10-14
ensure shared chats are shared

pablodanswer:
## Description

## How Has This Been Tested?

## Accepted Risk

## Related Issue(s)

## Checklist:

### DIFF
--- a/backend/danswer/server/features/persona/api.py
+++ b/backend/danswer/server/features/persona/api.py
@@ -239,11 +239,14 @@ def get_persona(
     persona_id: int,
     user: User | None = Depends(current_user),
     db_session: Session = Depends(get_session),
+    include_non_owned: bool = Query(
+        False, description="If true, return persona even if user doesn't own it."
+    ),
 ) -> PersonaSnapshot:
     return PersonaSnapshot.from_model(
         get_persona_by_id(
             persona_id=persona_id,
-            user=user,
+            user=user if not include_non_owned else None,
             db_session=db_session,
             is_for_edit=False,
         )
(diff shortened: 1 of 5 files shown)


backend/danswer/server/features/persona/api.py
```diff
@@ -239,11 +239,14 @@ def get_persona(
     persona_id: int,
     user: User | None = Depends(current_user),
     db_session: Session = Depends(get_session),
+    include_non_owned: bool = Query(
```
Weves: doesn't this give a way for any user to access any persona?

Weves: would slightly prefer to not offer that (e.g. just use a default icon if the user doesn't have access to the persona). Also, if we do keep this it should likely be something like `include_non_accessible`, since it already returns those that the user doesn't own

Weves (changes requested):

backend/danswer/server/features/persona/api.py
pablonyx: Agree on all fronts. Didn't mean to take this out of drafts. Was still testing to see if I could understand why the original issue was occurring.

web/src/app/chat/shared/[chatId]/page.tsx
```diff
@@ -53,18 +54,21 @@ export default async function Page({ params }: { params: { chatId: string } }) {
   if (user && !user.is_verified && authTypeMetadata?.requiresVerification) {
     return redirect("/auth/waiting-on-verification");
   }
+  const persona = chatSession?.persona_id
+    ? (availableAssistants?.find((p) => p.id === chatSession.persona_id) ??
+      availableAssistants?.[0] ??
+      null)
+    : (availableAssistants?.[0] ?? defaultPersona);
 
+  console.log(persona);
```
Comment (Weves) [inline web/src/app/chat/shared/[chatId]/page.tsx]: nit: remove console.log

PR #2844, opened 2024-10-17
Refactor + add global timeout env variable

Weves:
Also added a test for the litellm embedder.

Not really required, but it was something I did while looking into adding embedding batch limits (which turned out to not be necessary, since we already had a way to do it 😓)

### DIFF
--- a/backend/shared_configs/configs.py
+++ b/backend/shared_configs/configs.py
@@ -63,8 +63,15 @@ DEV_LOGGING_ENABLED = os.environ.get("DEV_LOGGING_ENABLED", "").lower() == "true
 # notset, debug, info, notice, warning, error, or critical
 LOG_LEVEL = os.environ.get("LOG_LEVEL", "notice")
 
+# Timeout for API-based embedding models
+# NOTE: does not apply for Google VertexAI, since the python client doesn't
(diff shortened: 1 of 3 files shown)


backend/shared_configs/configs.py
```diff
@@ -63,8 +63,25 @@
...
+# allow us to specify a custom timeout
+API_BASED_EMBEDDING_TIMEOUT = int(os.environ.get("API_BASED_EMBEDDING_TIMEOUT", "600"))
+
 # Only used for OpenAI
-OPENAI_EMBEDDING_TIMEOUT = int(os.environ.get("OPENAI_EMBEDDING_TIMEOUT", "600"))
+OPENAI_EMBEDDING_TIMEOUT = int(
+    os.environ.get("OPENAI_EMBEDDING_TIMEOUT", API_BASED_EMBEDDING_TIMEOUT)
+)
+
+_API_BASED_EMBEDDING_MAX_BATCH_SIZE_RAW = os.environ.get(
```
pablonyx: Is this used anywhere?

pablodanswer:
Like the new structure!  LGTM, assuming you sanity checked that 10 minutes is a rational default timeout for the API-based embeddings.

PR #2850, opened 2024-10-18
Minor tenant ID improvements

pablodanswer:
## Description
Minor improvements to tenant ID handling

- Ensure we set context in background processes when connectors / dynamic configs need it
- Move tenant ID retrieval logic to db engine


backend/danswer/danswerbot/slack/listener.py
```diff
@@ -357,51 +360,59 @@ def process_message(
         client=client.web_client, channel_id=channel
     )
 
-    with get_session_with_tenant(client.tenant_id) as db_session:
-        slack_bot_config = get_slack_bot_config_for_channel(
-            channel_name=channel_name, db_session=db_session
-        )
+    # Set the current tenant ID at the beginning for all DB calls within this thread
+    if client.tenant_id:
```
yuhongsun96, line 364: Not sure how you feel about these, but if we're in MT mode, we can also just double check here that if the env var is set that it must have a tenant_id and if it's not set that we never do, and throw an exception if the assumption is not true. The general concept being

yuhongsun96 (approved):
lgtm, can we add a comment on how the Slack clients work with threads which is how we know multiple requests from different Slack workspaces won't interfere with each other?

PR #2852, opened 2024-10-19
Account for malformed urls

pablodanswer:
## Description
User running into blocking issues with urls that use \ instead of / (malformed urls - not proper url syntax).


hagen-danswer (approved):
lgtm

PR #2858, opened 2024-10-19
Multitenant redis

pablodanswer:
## Description

Ensure that Redis client handles tenant id.

Previously
- Lock tasks based on keys that are global (not tenant specific)
- `postrun` and `worker init` would use tenant_id defined in initialization loop (ie. always last tenant id dynamically added to celery beat)

There is still work to be done to make sure 
1. We can horizontally scale with many nodes and many tenants 
2. We use a`TenantRedis` class that more explicitly handles the keys' cases

### DIFF
--- a/backend/danswer/background/celery/tasks/indexing/tasks.py
+++ b/backend/danswer/background/celery/tasks/indexing/tasks.py
@@ -51,10 +51,10 @@ logger = setup_logger()
     name="check_for_indexing",
     soft_time_limit=300,
 )
-def check_for_indexing(tenant_id: str | None) -> int | None:
+def check_for_indexing(*, tenant_id: str | None) -> int | None:
     tasks_created = 0
 
-    r = get_redis_client()
+    r = get_redis_client(tenant_id=tenant_id)
 
     lock_beat = r.lock(
         DanswerRedisLocks.CHECK_INDEXING_BEAT_LOCK,
@@ -64,7 +64,10 @@ def check_for_indexing(tenant_id: str | None) -> int | None:
     try:
         # these tasks should never overlap
         if not lock_beat.acquire(blocking=False):
+            task_logger.info(f"Lock acquired for tenant (Y): {tenant_id}")
             return None
+        else:
+            task_logger.info(f"Lock acquired for tenant (N): {tenant_id}")
 
         with get_session_with_tenant(tenant_id) as db_session:
             # Get the primary search settings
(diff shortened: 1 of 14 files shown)


backend/danswer/background/celery/tasks/indexing/tasks.py
```diff
@@ -64,7 +64,10 @@ def check_for_indexing(tenant_id: str | None) -> int | None:
     try:
         # these tasks should never overlap
         if not lock_beat.acquire(blocking=False):
+            task_logger.info(f"Lock acquired for tenant (Y): {tenant_id}")
```
rkuo-danswer, line 67: Are the Y and N reversed?  Also, seems easier to just write in english "Lock acquired" or "Lock not acquired".  Also the amount of logging for locks will be spammy, I'm not sure this logging should be INFO level.

backend/danswer/background/celery/tasks/indexing/tasks.py
```diff
@@ -64,7 +64,10 @@ def check_for_indexing(tenant_id: str | None) -> int | None:
...
             return None
+        else:
+            task_logger.info(f"Lock acquired for tenant (N): {tenant_id}")
```
rkuo-danswer, line 70: no else after return

backend/danswer/db/engine.py
```diff
@@ -295,30 +295,32 @@ async def get_async_session_with_tenant(
...
     tenant_id: str | None = None,
 ) -> Generator[Session, None, None]:
-    """Generate a database session with the appropriate tenant schema set."""
+    """Generate a database session bound to a connection with the appropriate tenant schema set."""
     engine = get_sqlalchemy_engine()
+
     if tenant_id is None:
         tenant_id = current_tenant_id.get()
+    else:
+        current_tenant_id.set(tenant_id)
```
rkuo-danswer, line 304: Can we name or organize "current_tenant_id" in such a way that the fact it's a context var is more obvious?

backend/danswer/key_value_store/store.py
```diff
@@ -28,7 +28,8 @@
 
 class PgRedisKVStore(KeyValueStore):
     def __init__(self) -> None:
-        self.redis_client = get_redis_client()
+        tenant_id = current_tenant_id.get()
```
rkuo-danswer, line 31: Is there any way to inject these global dependencies in instead of creating a hard dependency on them?

rkuo-danswer (commented):
Expect there will be merge conflicts ... happy to look again when resolved.

backend/danswer/background/celery/tasks/indexing/tasks.py
pablonyx: Good catch! Did not intend to leave these in- very spammy

backend/danswer/db/engine.py
pablonyx: Great point - I think renaming is a good idea– seems worth putting in this PR

PR #2869, opened 2024-10-22
Handle models like openai/bedrock/claude-3.5-...

Weves:
Fixes DAN-847

https://linear.app/danswer/issue/DAN-847/fix-issue-where-openaibedrockclaude-35-didnt-pick-up-proper-number-of


pablodanswer:
Good logic- only concern is if a model provider which doesn't exist in model map happens to have a different (shorter) context limit than the base model, but only way around that would be having users manually set limits, so LGTM

pablodanswer (approved):

PR #2874, opened 2024-10-22
Confluence polish

hagen-danswer:


backend/danswer/connectors/confluence/utils.py
```diff
@@ -18,7 +18,25 @@
 
 logger = setup_logger()
 
-_USER_NOT_FOUND = "Unknown User"
+
+_USER_EMAIL_CACHE: dict[str, str | None] = {}
+
+
```
pablonyx, line 24: Might consider a cache decorator instead

pablodanswer (approved):

backend/danswer/connectors/confluence/utils.py
hagen-danswer: unfortunately there are issues with caching complicated objects like confluence_connector. We believe this was causing the memory leak

PR #2876, opened 2024-10-22
upgraded claude model strings

hagen-danswer:


web/src/lib/llm/utils.ts
```diff
@@ -69,20 +69,20 @@ const MODEL_NAMES_SUPPORTING_IMAGE_INPUT = [
   "gpt-4-turbo",
   "gpt-4-1106-vision-preview",
   // standard claude names
-  "claude-3-5-sonnet-20240620",
+  "claude-3-5-sonnet-20241022",
   "claude-3-opus-20240229",
   "claude-3-sonnet-20240229",
   "claude-3-haiku-20240307",
```
pablonyx, line 76: Users that have previously set their default model should still have access to image input

backend/danswer/llm/llm_provider_options.py
```diff
@@ -100,8 +100,8 @@ def fetch_available_well_known_llms() -> list[WellKnownLLMProviderDescriptor]:
             api_version_required=False,
             custom_config_keys=[],
             llm_names=fetch_models_for_provider(ANTHROPIC_PROVIDER_NAME),
-            default_model="claude-3-5-sonnet-20240620",
-            default_fast_model="claude-3-5-sonnet-20240620",
+            default_model="claude-3-5-sonnet-20241022",
```
pablonyx, line 104: Watch out for context length issues - unless we bump our LiteLLM version (when they add "support" for this model), will use default max tokens (which is super low, something like `4096`)

pablodanswer (changes requested):
Looks good but

hagen-danswer:
Waiting for tagged litellm version:
[Repo](https://github.com/BerriAI/litellm/tags)

web/src/lib/hooks.ts
```diff
@@ -278,6 +278,7 @@ const MODEL_DISPLAY_NAMES: { [key: string]: string } = {
   "claude-2.0": "Claude 2.0",
   "claude-instant-1.2": "Claude Instant 1.2",
   "claude-3-5-sonnet-20240620": "Claude 3.5 Sonnet",
+  "claude-3-5-sonnet-20241022": "Claude 3.5 Sonnet (New)",
```
LostVector, line 281: seems possible we might need to rename (new) later but i suppose that's easily fixable if it happens

LostVector (approved):

pablodanswer (approved):

PR #2975, opened 2024-10-28
Prompting

pablodanswer:
## Description

Auto generate starter messages based on attached information.

May be worth moving functions to utility file, although logic is fairly starter message specific (no natural place other than right next to the functionality to store).

https://github.com/user-attachments/assets/899afcf0-75e3-42a5-837b-7360f9f6dc2f


Chillwind132:
Hi @yuhongsun96 @pablodanswer - Looks like an interesting feature, can we have this merged

yuhongsun96:
Couple items to address here:
- We removed the "Title" part of the prompt so let's do the same here.
- Let's not do the autogenerate actually, it runs too many times and is confusing to the user. Instead let's do a green "Generate" button or something like that?
- If the skeleton doesn't match the filled out form, let's not jump the top down, keep the top pinned and jump the bottom up if anything. Also there was this weird thing where the bottom has a white bar at times? Let's make sure that's patched
- Let's make the form skeleton a bit larger so the jump is less jarring
- Let's work on the prompting a bit and generate all 4 together with a single prompt. This would allow us to force them to be more different. It will be slower but whatever...
- Don't let the user create an assistant without at least generating prompts, or removing them explicitly (no accidental creating assistants without starter prompts).

PR #3025, opened 2024-11-01
Tool call per message

pablodanswer:
## Description
One tool call per message

First in a sequence of changes


backend/danswer/chat/process_message.py
```diff
@@ -779,6 +779,8 @@ def stream_chat_message_objects(
                     img_generation_response = cast(
                         list[ImageGenerationResponse], packet.response
                     )
+                    print(img_generation_response[0].__dict__)
```
rkuo-danswer: Did we mean to leave these in?

rkuo-danswer (approved):

PR #3027, opened 2024-11-02
Upgrade to latest NextJS + switch to turbopack

Weves:
Much faster page loading during dev!


web/src/middleware.ts
```diff
@@ -4,24 +4,26 @@ import { SERVER_SIDE_ONLY__PAID_ENTERPRISE_FEATURES_ENABLED } from "./lib/consta
...
+    "/admin/groups/:path*",
+    "/admin/api-key/:path*",
+    "/admin/performance/usage/:path*",
+    "/admin/performance/query-history/:path*",
+    "/admin/whitelabeling/:path*",
+    "/admin/performance/custom-analytics/:path*",
+    "/admin/standard-answer/:path*",
+
+    // Cloud only
+    "/admin/cloud-settings/:path*",
```
Weves, line 18: I don't see a reason to not just include this anyways

web/next.config.js
```diff
@@ -10,7 +10,6 @@ const { withSentryConfig } = require("@sentry/nextjs");
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   output: "standalone",
-  swcMinify: true,
```
Weves, line 13: not recognized anymore

web/package.json
```diff
@@ -71,5 +73,8 @@
     "eslint": "^8.48.0",
     "eslint-config-next": "^14.1.0",
     "prettier": "2.8.8"
+  },
+  "overrides": {
+    "react-is": "^19.0.0-rc-69d4b800-20241021"
```
Weves, line 78: needed to support shadcn charts with nextjs 15 

https://ui.shadcn.com/docs/react-19

Weves: a bit messy since we haven't switched to use react19, but it seems to work 🤷

Weves: without it, the charts are empty

pablodanswer (approved):

pablodanswer:
Seeing a coupel small issues - specifically when trying to import the base tailwind themes 

```
Module not found
   3 | const baseThemes = require("./tailwind-themes/tailwind.config.js");
   4 | const customThemes = process.env.NEXT_PUBLIC_THEME
>  5 |   ? require(
     |     ^^^^^^^^
>  6 |       `./tailwind-themes/custom/${process.env.NEXT_PUBLIC_THEME}/tailwind.config.js`
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>  7 |     )
     | ^^^^^^
   8 |   : null;
   9 |
  10 | /** @type {import('tailwindcss').Config} */

```

and  on the chat page- `Accessing element.ref was removed in React 19. ref is now a regular prop. It will be removed from the JSX Element type in a future release.`

PR #3042, opened 2024-11-04
update persona defaults

pablodanswer:
## Description

https://github.com/user-attachments/assets/a9b2775d-4899-449b-b1c0-b0ba86e315ee


yuhongsun96:
image
I prefer having just the 1 line like the "Give me an overview of what's here" and not show the other one (that one will be purely for prompting the LLM)

yuhongsun96:
image
I prefer these being smaller, like just the icon and name, maybe on hover it can drop down and show the other information

yuhongsun96:
image
Need to enable more scrolling so that on smaller screen it's not cut off (the dropdown thing we want to add)

yuhongsun96:
image
On smaller screens, the experience is kind of meh... (There is no option for any scrolling here either)

yuhongsun96:
I think in general, maybe a bit more spacing between the elements might be nice? Let's tweak it live together

yuhongsun96 (approved):

PR #3057, opened 2024-11-05
Google Drive Improvements

hagen-danswer:


backend/danswer/connectors/google_drive/connector.py
```diff
@@ -154,125 +189,167 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, str] | None
...
+
+    def _initialize_all_class_variables(self) -> None:
+        # Get all user emails
+        # Get admins first becuase they are more likely to have access to the most files
+        user_emails = [self.primary_admin_email]
+        for admins_only in [True, False]:
+            for email in self._get_all_user_emails(admins_only=admins_only):
+                if email not in user_emails:
+                    user_emails.append(email)
+        self._all_org_emails = user_emails
```
Weves, line 236: do not like this, let's avoid using these stateful instance variables as much as possible

Weves: let's pass this around to not create this implicit dependency

Weves (changes requested):

Weves (approved):

PR #3093, opened 2024-11-09
Remove ee

pablodanswer:
## Description
Fully remove ee imports from non ee modules (use fetch_versioned instead)

New fetch_versioned method for functionality that has no functional non-ee equivalent (and raises exception for missing modules in ee enabled environments

## How Has This Been Tested?
- Core EE flows tested (token rate limits, user groups, etc.)
- Authentication with multi tenancy enabled, disabled

N/A

N/A

## Mental Checklist:


Weves (approved):

steyer-mika:
@pablodanswer Are you sure that this code worked? If I try to register a user (im on the main branch without ee turned on) I get the following error:
`File "/app/danswer/auth/users.py", line 231, in create
    tenant_id = await fetch_ee_implementation_or_noop(
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: object function can't be used in 'await' expression`

It seems that when EE is disabled, the function fetch_ee_implementation_or_noop returns a no-op lambda that directly yields the async_return_default_schema function object instead of calling it. This leads to the error because await is applied to the function object itself rather than the result of calling the function.

To fix this, I believe the no-op lambda should actually invoke async_return_default_schema with the provided arguments, like so:
`return lambda *args, **kwargs: async_return_default_schema(*args, **kwargs)`

But I am not sure (I am no python dev).

PR #3094, opened 2024-11-09
Robustify cloud deployment + include initial KEDA configuration

pablodanswer:
## Description

- Ensure cloud tags point to private Dockerhub images to prevent confusion for open source users
- Initial KEDA configurations (limits are set very conservatively, but I am actively iterating on them- best to simply get the base configuration in the repo for maintainability)

### DIFF
--- a/.github/workflows/docker-build-push-backend-container-on-tag.yml
+++ b/.github/workflows/docker-build-push-backend-container-on-tag.yml
@@ -3,61 +3,61 @@ name: Build and Push Backend Image on Tag
 on:
   push:
     tags:
-      - '*'
+      - "*"
 
 env:
-  REGISTRY_IMAGE: danswer/danswer-backend
+  REGISTRY_IMAGE: ${{ contains(github.ref_name, 'cloud') && 'danswer/danswer-backend-cloud' || 'danswer/danswer-backend' }}
   LATEST_TAG: ${{ contains(github.ref_name, 'latest') }}
-  
+
 jobs:
   build-and-push:
-    # TODO: investigate a matrix build like the web container 
+    # TODO: investigate a matrix build like the web container
     # See https://runs-on.com/runners/linux/
-    runs-on: [runs-on,runner=8cpu-linux-x64,"run-id=${{ github.run_id }}"]
+    runs-on: [runs-on, runner=8cpu-linux-x64, "run-id=${{ github.run_id }}"]
 
     steps:
-    - name: Checkout code
-      uses: actions/checkout@v4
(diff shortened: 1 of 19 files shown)


backend/danswer/background/celery/apps/app_base.py
```diff
@@ -146,6 +146,8 @@ def on_celeryd_init(sender: Any = None, conf: Any = None, **kwargs: Any) -> None
 def wait_for_redis(sender: Any, **kwargs: Any) -> None:
     """Waits for redis to become ready subject to a hardcoded timeout.
     Will raise WorkerShutdown to kill the celery worker if the timeout is reached."""
+    if MULTI_TENANT:
```
rkuo-danswer: Understand we don't need the checks for multi-tenant/cloud, but conditionals like this should be as high in the call stack as possible, not embedded in tool functions.

.github/workflows/docker-build-push-backend-container-on-tag.yml
```diff
@@ -3,61 +3,61 @@ name: Build and Push Backend Image on Tag
 on:
   push:
     tags:
-      - '*'
+      - "*"
 
 env:
-  REGISTRY_IMAGE: danswer/danswer-backend
+  REGISTRY_IMAGE: ${{ contains(github.ref_name, 'cloud') && 'danswer/danswer-backend-cloud' || 'danswer/danswer-backend' }}
   LATEST_TAG: ${{ contains(github.ref_name, 'latest') }}
```
rkuo-danswer, line 10: is this going to potentially tag cloud images as latest?  should that be explicitly wired off if so?

rkuo-danswer (changes requested):

.github/workflows/docker-build-push-backend-container-on-tag.yml
pablonyx: Believe we do want this functionality within docker– the latest tag should apply separately to the docker repositories (ie open source users would never see the `latest` cloud tags, is my understanding)

backend/danswer/background/celery/apps/app_base.py
pablonyx: Completely agree! Moved all these checks up to the worker init

.github/workflows/docker-build-push-backend-container-on-tag.yml
rkuo-danswer: ok, completely separate docker images. sounds good!

rkuo-danswer (approved):

PR #3120, opened 2024-11-13
Updated google copy and added non admin oauth support

hagen-danswer:
- Updated copy
- changed it from textfield to textarea
- Fixed issue where incorrect options were being displayed to users using oauth
- updated backend logic to not break if non admin user in org is trying to access shared drives using oauth

here is new copy + textarea:

image

### DIFF
--- a/backend/danswer/connectors/google_drive/connector.py
+++ b/backend/danswer/connectors/google_drive/connector.py
@@ -192,23 +192,33 @@ class GoogleDriveConnector(LoadConnector, PollConnector, SlimConnector):
     def _update_traversed_parent_ids(self, folder_id: str) -> None:
         self._retrieved_ids.add(folder_id)
 
-    def _get_all_user_emails(self, admins_only: bool) -> list[str]:
+    def _get_all_user_emails(self) -> list[str]:
+        # Start with primary admin email
+        user_emails = [self.primary_admin_email]
+
+        # Only fetch additional users if using service account
+        if isinstance(self.creds, OAuthCredentials):
+            return user_emails
+
         admin_service = get_admin_service(
             creds=self.creds,
             user_email=self.primary_admin_email,
         )
-        query = "isAdmin=true" if admins_only else "isAdmin=false"
-        emails = []
-        for user in execute_paginated_retrieval(
-            retrieval_function=admin_service.users().list,
-            list_key="users",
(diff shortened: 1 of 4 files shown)


backend/danswer/connectors/google_drive/connector.py
```diff
@@ -219,40 +229,12 @@ def _get_all_drive_ids(self) -> set[str]:
         for drive in execute_paginated_retrieval(
             retrieval_function=primary_drive_service.drives().list,
             list_key="drives",
+            continue_on_404_or_403=isinstance(self.creds, OAuthCredentials),
```
Weves: add comment as to why

backend/danswer/connectors/google_drive/connector.py
```diff
@@ -219,40 +229,12 @@ def _get_all_drive_ids(self) -> set[str]:
...
-                f"Some shared drive IDs were not found. IDs: {invalid_drive_ids}"
-            )
-            logger.warning("Checking for folder access instead...")
-            self._requested_folder_ids.update(invalid_drive_ids)
-
-        if not self.include_shared_drives:
-            self._requested_shared_drive_ids = set()
-        elif not self._requested_shared_drive_ids:
-            self._requested_shared_drive_ids = self._all_drive_ids
+        return all_drive_ids or set(self._requested_shared_drive_ids)
```
Weves: prefer adding a comment why

backend/danswer/connectors/google_drive/connector.py
```diff
@@ -302,15 +291,34 @@ def _fetch_drive_items(
         start: SecondsSinceUnixEpoch | None = None,
         end: SecondsSinceUnixEpoch | None = None,
     ) -> Iterator[GoogleDriveFileType]:
-        self._initialize_all_class_variables()
+        all_org_emails: list[str] = self._get_all_user_emails()
+
+        all_drive_ids: set[str] = self._get_all_drive_ids()
+
+        # remove drive ids from the folder ids because they are queried differently
+        self._requested_folder_ids -= all_drive_ids
```
Weves: `requested` fields should not change, maybe create new ones?

Weves (changes requested):

Weves (approved):

PR #3136, opened 2024-11-14
Zendesk connector: add ability to skip tickets that include tags defined in env var, exclude deleted tickets server side

jwbrown77:
## Description

@yuhongsun96 - apologies if I am tagging the wrong person, but you seem to be the most active committer to the Zendesk connector.

I have a use case where I would like to ingest Zendesk tickets, however I need to be able to filter out certain tickets based on if they have a configured tag present.

There is a similar functionality with Zendesk articles that allows users to filter out articles based on certain labels.

This PR:

- Adds a new optional config: `ZENDESK_CONNECTOR_SKIP_TICKET_TAGS` - a comma separated list of tags that when any are present, the ingestion should skip.
- Add an if statement to the connector that performs this check, then continues to the next ticket iteration loop if the condition is met
- Unrelated, but according to Zendesk's Incremental Ticket API docs (https://developer.zendesk.com/api-reference/ticketing/ticket-management/incremental_exports/), an optional parameter called `exclude_deleted` can be set to `true` to have their API not return deleted tickets.  This should be more efficient than performing the check client side.

## How Has This Been Tested?

- I tested with no change.  It worked, but as designed, according to "Explorer", it pulled all tickets from the configured start date.
- I deleted what it ingested.
- I applied my patch and tested again, and the result was a lower number of tickets.  I spot checked several tickets and none of them had the tags I had configured in the environment variable.
- In addition, I manually tested the `if` statement in a simple script and verified it was behaving as desired:

```
ignore = ["ignore_this", "ignore_that"]

tags = ["test", "ignore_that"]

if any(tag in ignore for tag in tags):
    print("should ignore")
else:
    print("should not ignore")
```

N/A

N/A

## Mental Checklist:

### DIFF
--- a/backend/danswer/configs/app_configs.py
+++ b/backend/danswer/configs/app_configs.py
@@ -350,6 +350,10 @@ ZENDESK_CONNECTOR_SKIP_ARTICLE_LABELS = os.environ.get(
     "ZENDESK_CONNECTOR_SKIP_ARTICLE_LABELS", ""
 ).split(",")
 
+# comma delimited list of zendesk ticket tags to skip indexing for
+ZENDESK_CONNECTOR_SKIP_TICKET_TAGS = os.environ.get(
+    "ZENDESK_CONNECTOR_SKIP_TICKET_TAGS", ""
+).split(",")
 
 #####
 # Indexing Configs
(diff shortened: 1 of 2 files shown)


eric-zadara:
bump

jwbrown77:
bump

PR #3145, opened 2024-11-16
Add error message on JSON decode error in CustomTool

Weves:
^


backend/danswer/tools/tool_implementations/custom/custom_tool.py
```diff
@@ -241,6 +242,8 @@ def run(self, **kwargs: Any) -> Generator[ToolResponse, None, None]:
         )
         content_type = response.headers.get("Content-Type", "")
 
+        tool_result: Any
```
hagen-danswer, line 245: type any????

hagen-danswer (changes requested):

hagen-danswer (approved):
jk lgtm

PR #3196, opened 2024-11-21
make external permissioned user creation case insensitive

hagen-danswer:
## Description
^

## Mental Checklist:

### DIFF
--- a/backend/danswer/db/users.py
+++ b/backend/danswer/db/users.py
@@ -103,17 +103,6 @@ def list_users(
     return db_session.scalars(stmt).unique().all()
 
 
-def get_users_by_emails(
-    db_session: Session, emails: list[str]
-) -> tuple[list[User], list[str]]:
-    # Use distinct to avoid duplicates
-    stmt = select(User).filter(User.email.in_(emails))  # type: ignore
-    found_users = list(db_session.scalars(stmt).unique().all())  # Convert to list
-    found_users_emails = [user.email for user in found_users]
-    missing_user_emails = [email for email in emails if email not in found_users_emails]
-    return found_users, missing_user_emails
-
-
 def get_user_by_email(email: str, db_session: Session) -> User | None:
     user = (
         db_session.query(User)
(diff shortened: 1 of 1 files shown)


backend/danswer/db/users.py
```diff
@@ -103,17 +103,6 @@ def list_users(
     return db_session.scalars(stmt).unique().all()
```
hagen-danswer, line 105: moved this down to be closer to other logic

backend/danswer/db/users.py
```diff
@@ -128,7 +117,7 @@ def fetch_user_by_id(db_session: Session, user_id: UUID) -> User | None:
     return db_session.query(User).filter(User.id == user_id).first()  # type: ignore
 
 
-def _generate_non_web_slack_user(email: str) -> User:
+def _generate_slack_user(email: str) -> User:
     fastapi_users_pw_helper = PasswordHelper()
```
hagen-danswer, line 121: var name changes

backend/danswer/db/users.py
```diff
@@ -149,13 +138,29 @@ def add_slack_user_if_not_exists(db_session: Session, email: str) -> User:
...
     db_session.add(user)
     db_session.commit()
     return user
 
 
-def _generate_non_web_permissioned_user(email: str) -> User:
+def _get_users_by_emails(
+    db_session: Session, lower_emails: list[str]
+) -> tuple[list[User], list[str]]:
+    stmt = select(User).filter(func.lower(User.email).in_(lower_emails))  # type: ignore
```
hagen-danswer, line 150: this was main change: we now do a case insensitive user search and a case insensitive check for missing user emails.

I believe this is what caused admins to be set to externally permissioned users as this is the only place in the code where role is set to UserRole.EXT_PERM_USER. the failure case was there was a case mismatch for user upserting so now there are 2 users in the db with different roles. im looking into creating an alembic upgrade to fix it

hagen-danswer:
replaced by:

https://github.com/danswer-ai/danswer/pull/3257

PR #3294, opened 2024-12-01
Fix Citation Minor Bugs

yuhongsun96:
## Description
- Citations where the LLM produces double brackets do not work, for example [[, 4, ]]
- Citations that have some characters in the token after the end of the citation, for example [, 1, ].

## How Has This Been Tested?
Unit tests haven't broken and a new test was added to cover these two cases

That code is complicated and brittle, hopefully no new bugs

N/A

## Mental Checklist:

### DIFF
--- a/backend/danswer/llm/answering/stream_processing/citation_processing.py
+++ b/backend/danswer/llm/answering/stream_processing/citation_processing.py
@@ -67,9 +67,9 @@ class CitationProcessor:
                 if piece_that_comes_after == "\n" and in_code_block(self.llm_out):
                     self.curr_segment = self.curr_segment.replace("```", "```plaintext")
 
-        citation_pattern = r"\[(\d+)\]"
+        citation_pattern = r"\[(\d+)\]|\[\[(\d+)\]\]"
         citations_found = list(re.finditer(citation_pattern, self.curr_segment))
-        possible_citation_pattern = r"(\[\d*$)"  # [1, [, etc
+        possible_citation_pattern = r"(\[+\d*$)"
         possible_citation_found = re.search(
             possible_citation_pattern, self.curr_segment
         )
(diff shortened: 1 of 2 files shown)


backend/tests/unit/danswer/llm/answering/stream_processing/test_citation_processing.py
```diff
@@ -385,6 +385,16 @@ def process_text(
             "Here is some text[[1]](https://0.com). Some other text",
             ["doc_0"],
         ),
+        # ['To', ' set', ' up', ' D', 'answer', ',', ' if', ' you', ' are', ' running', ' it', ' yourself', ' and',
```
pablonyx, line 388: Nit

backend/danswer/llm/answering/stream_processing/citation_processing.py
pablonyx, line 1: Was removed previously, but maybe we could add back a comment explaining the logic [here](https://github.com/danswer-ai/danswer/blob/20184a12a55e5cf115a123ccdce7df21c22073a7/backend/danswer/llm/answering/stream_processing/citation_processing.py)?  

```
    """
    Key aspects:
    1. Stream Processing:
    - Processes tokens one by one, allowing for real-time handling of large texts.
    2. Citation Detection:
    - Uses regex to find citations in the format [number].
    - Example: [1], [2], etc.
    3. Citation Mapping:
    - Maps detected citation numbers to actual document ranks using doc_id_to_rank_map.
    - Example: [1] might become [3] if doc_id_to_rank_map maps it to 3.
    4. Citation Formatting:
    - Replaces citations with properly formatted versions.
    - Adds links if available: [[1]](https://example.com)
    - Handles cases where links are not available: [[1]]()
    5. Duplicate Handling:
    - Skips consecutive citations of the same document to avoid redundancy.
    6. Output Generation:
    - Yields DanswerAnswerPiece objects for regular text.
    - Yields CitationInfo objects for each unique citation encountered.
    7. Context Awareness:
    - Uses context_docs to access document information for citations.
    This function effectively processes a stream of text, identifies and reformats citations,
    and provides both the processed text and citation information as output.
    """
```

pablonyx: Previous comment (above one) wasn't very clear, but some steps of the logic would probably help keep this logic maintainable

pablodanswer (approved):

PR #3305, opened 2024-12-02
hard refresh on auth

pablodanswer:
## Description

## How Has This Been Tested?

N/A

N/A

## Mental Checklist:

### DIFF
--- a/web/src/app/auth/login/EmailPasswordForm.tsx
+++ b/web/src/app/auth/login/EmailPasswordForm.tsx
@@ -5,13 +5,10 @@ import { usePopup } from "@/components/admin/connectors/Popup";
 import { basicLogin, basicSignup } from "@/lib/user";
 import { Button } from "@/components/ui/button";
 import { Form, Formik } from "formik";
-import { useRouter } from "next/navigation";
 import * as Yup from "yup";
 import { requestEmailVerification } from "../lib";
 import { useState } from "react";
 import { Spinner } from "@/components/Spinner";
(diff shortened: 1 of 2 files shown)


web/src/app/auth/login/EmailPasswordForm.tsx
```diff
@@ -69,9 +66,9 @@ export function EmailPasswordForm({
           if (loginResponse.ok) {
             if (isSignup && shouldVerify) {
               await requestEmailVerification(values.email);
-              router.push("/auth/waiting-on-verification");
+              window.location.href = "/auth/waiting-on-verification";
```
Weves, line 72: can we add a comment as to why we're doing this?

Weves (changes requested):

web/src/app/auth/login/EmailPasswordForm.tsx
Weves: it feels "wrong" to do this, so would love to have the justification here to alleviate future confusion

Weves: also, would it be possible to just do a router.refresh() or similar?

pablonyx: Sadly router.refresh doesn't trigger server side refresh for providers. Could use specific <Link attributes to trigger full refresh, but we'd need to either move our providers client side or have a global refresh we use for this redirect (which feels less maintainable than just triggering global update here).

Weves (approved):

PR #3372, opened 2024-12-08
Prompting

pablonyx:
## Description
Auto generate starter messages based on assistant information.

https://github.com/user-attachments/assets/2f70e2e6-4fd1-4294-8bd7-dcc082731c6a

- Adds a dedicated endpoint (`/assistant-prompt-refresh`) to generate diverse starter messages for a persona.
- Uses random document chunk retrieval and configurable prompts to produce a set of unique conversation starters.
- Integrates logic for both category generation and individual starter message creation into the persona flow.

## How Has This Been Tested?

N/A

N/A

## Mental Checklist:

### DIFF
--- a/web/src/app/admin/assistants/StarterMessageList.tsx
+++ b/web/src/app/admin/assistants/StarterMessageList.tsx
@@ -1,0 +1,198 @@
+"use client";
+
+import { ArrayHelpers, ErrorMessage, Field, useFormikContext } from "formik";
+
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipProvider,
+  TooltipTrigger,
+} from "@radix-ui/react-tooltip";
+
+import { useEffect } from "react";
+import { FiInfo, FiTrash2, FiPlus } from "react-icons/fi";
+import { StarterMessage } from "./interfaces";
+import { Label } from "@/components/admin/connectors/Field";
+
+export default function StarterMessagesList({
+  values,
+  arrayHelpers,
+  isRefreshing,
+  touchStarterMessages,
+}: {
+  values: StarterMessage[];
+  arrayHelpers: ArrayHelpers;
+  isRefreshing: boolean;
+  touchStarterMessages: () => void;
+}) {
+  const { handleChange } = useFormikContext();
+
+  // Group starter messages into rows of 2 for display purposes
+  const rows = values.reduce((acc: StarterMessage[][], curr, i) => {
+    if (i % 2 === 0) acc.push([curr]);
+    else acc[acc.length - 1].push(curr);
+    return acc;
+  }, []);
+
+  const canAddMore = values.length <= 6;
+
+  return (
+    <div className="mt-4 flex flex-col gap-6">
+      {rows.map((row, rowIndex) => (
+        <div key={rowIndex} className="flex items-start gap-4">
+          <div className="grid grid-cols-2 gap-6 w-full xl:w-fit">
+            {row.map((starterMessage, colIndex) => (
+              <div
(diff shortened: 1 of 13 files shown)


backend/danswer/server/features/persona/api.py
```diff
@@ -363,3 +368,26 @@ def build_final_template_prompt(
             retrieval_disabled=retrieval_disabled,
         )
     )
+
+
+@basic_router.post("/assistant-prompt-refresh")
```
rkuo-danswer: Can we test this endpoint?  Is it even reliably testable due to the usage of an LLM?

backend/danswer/secondary_llm_flows/starter_message_creation.py
```diff
@@ -0,0 +1,270 @@
...
+
+
+def generate_starter_messages(
+    name: str,
+    description: str,
+    instructions: str,
+    document_set_ids: List[int],
+    db_session: Session,
+    user: User | None,
+) -> List[StarterMessage]:
```
rkuo-danswer: because we're calling an LLM this call is going to fail some percentage of the time ... is that OK?  Can we comment on what's expected and what happens in the failure case?

backend/danswer/document_index/vespa/index.py
```diff
@@ -899,6 +900,31 @@ def _delete_document(
 
         logger.info("Batch deletion completed")
 
+    def random_retrieval(
+        self,
+        filters: IndexFilters,
+        num_to_retrieve: int = 10,
+    ) -> list[InferenceChunkUncleaned]:
+        """Retrieve random chunks matching the filters using Vespa's random ranking"""
+        vespa_where_clauses = build_vespa_filters(filters)
```
rkuo-danswer: Seems odd that we would need to manually strip the "and" from a helper function's output ... is that something the helper function should be doing for us?

backend/danswer/document_index/vespa/index.py
```diff
@@ -899,6 +900,31 @@ def _delete_document(
 
         logger.info("Batch deletion completed")
 
+    def random_retrieval(
+        self,
+        filters: IndexFilters,
+        num_to_retrieve: int = 10,
+    ) -> list[InferenceChunkUncleaned]:
+        """Retrieve random chunks matching the filters using Vespa's random ranking"""
```
rkuo-danswer: could use a comment as to what the use case for this is. I think I understand it in this context but that will be quickly forgotten over time.

web/src/app/admin/assistants/StarterMessageList.tsx
```diff
@@ -0,0 +1,197 @@
...
+  touchStarterMessages,
+}: {
+  values: StarterMessage[];
+  arrayHelpers: ArrayHelpers;
+  isRefreshing: boolean;
+  touchStarterMessages: () => void;
+}) {
+  const { handleChange } = useFormikContext();
+
+  const rows = values.reduce((acc: StarterMessage[][], curr, i) => {
```
rkuo-danswer, line 31: could use a comment of what this is for

backend/danswer/server/features/persona/api.py
pablonyx: We certainly could add a test for all the secondary LLM flows, though would be flaky some percentage of the time.

backend/danswer/document_index/vespa/index.py
pablonyx: Updated for clarity- we don't need the and when we are using as the only part of the Vespa query. This is due to the nature of the random sampling.

rkuo-danswer (approved):

PR #3386, opened 2024-12-09
Model Server Async

yuhongsun96:
## Description
The model server does not make correct use of async so is blocking for network calls to external embedding models.

## How Has This Been Tested?
Additionally to unit tests:
(all following done with a web connector)
- Cohere embedding light
- Nomic
- Openai
- Cohere rerank
- Mixbread xsmall rerank

Haven't actually load tested this. It's assumed to be performance gains from an understanding of how FastAPI and Async are suppose to interact.

Google embeddings weren't tested
LiteLLM options were not tested
Azure was not tested
Voyage was not tested but not sure if anyone is using it anyway

N/A

## Mental Checklist:


hagen-danswer (commented):
i still think we should write at test to make sure the async embedding is actually improving performance E.g. by writing a test that actually kicks off a bunch of simultaneous embedding jobs and confirms that they are all running.

would also prevent regressions in the future if for some reason someone added something to the code that made it blocking for other async tasks, we wouldnt silently lose performance...

hagen-danswer (approved):
el gee tee em

PR #3409, opened 2024-12-10
fixed ephemeral slackbot messages

hagen-danswer:
## Description
/danswerbot command now responds privately to the user

tested by asking a /danswerbot question and ensuring only the asker could see the response by looking at the chat with another account

also tested the normal qa flow to make sure that was unaffected

also checkted @danswerbot flow to make sure that was unaffected


yuhongsun96:
How have you tested this?

yuhongsun96 (approved):
lgtm

PR #3410, opened 2024-12-10
made external group emails lowercase

hagen-danswer:
## Description
- This fixes an issue where when we create a user in the db, the email is converted to lowercase but we still use the upercase one to match


pablonyx (approved):
Would love to be able to assume that the values we're referring to are already normalized. This is likely something worth doing at the DB level if we're running into issues like these.

PR #3418, opened 2024-12-10
do not include slackbot sessions when fetching `chat sessions`

pablonyx:
## Description
Fixes https://linear.app/danswer/issue/DAN-1167/slackbot-threads-showing-in-chat-display

## How Has This Been Tested?

N/A

N/A

## Mental Checklist:

### DIFF
--- a/backend/onyx/db/chat.py
+++ b/backend/onyx/db/chat.py
@@ -141,13 +141,19 @@ def get_valid_messages_from_query_sessions(
     return {row.chat_session_id: row.message for row in first_messages}
 
 
+# Retrieves chat sessions by user
+# Chat sessions do not include onyxbot flows
 def get_chat_sessions_by_user(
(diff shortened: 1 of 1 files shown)


hagen-danswer (approved):
sounds good

backend/danswer/db/chat.py
```diff
@@ -141,6 +141,8 @@ def get_valid_messages_from_query_sessions(
     return {row.chat_session_id: row.message for row in first_messages}
 
 
+# Retrieves chat sessions by user
+# Chat sessions do not include danswerbot flows
 def get_chat_sessions_by_user(
     user_id: UUID | None,
     deleted: bool | None,
```
yuhongsun96: pref this being an argument. If we really want to have a default value we can set it to False. But may be good to not even have default value so that it enforces "thinking" about the logic upstream

PR #3419, opened 2024-12-10
Added ability to set timezone offset for Confluence connector in UI

hagen-danswer:
## Description
- we now allow the user to manually input their timezone through the UI if they are trying to use confluence server

### DIFF
--- a/backend/danswer/connectors/confluence/connector.py
+++ b/backend/danswer/connectors/confluence/connector.py
@@ -1,6 +1,5 @@
 from datetime import datetime
 from datetime import timedelta
-from datetime import timezone
 from typing import Any
 from urllib.parse import quote
 
@@ -108,7 +107,7 @@ class ConfluenceConnector(LoadConnector, PollConnector, SlimConnector):
             )
             self.cql_label_filter = f" and label not in ({comma_separated_labels})"
 
-        self.timezone: timezone = timezone(offset=timedelta(hours=timezone_offset))
+        self.timezone_offset = timezone_offset or CONFLUENCE_TIMEZONE_OFFSET
 
     @property
     def confluence_client(self) -> OnyxConfluence:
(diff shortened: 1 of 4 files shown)


backend/danswer/connectors/confluence/connector.py
```diff
@@ -108,7 +107,7 @@ def __init__(
             )
             self.cql_label_filter = f" and label not in ({comma_separated_labels})"
 
-        self.timezone: timezone = timezone(offset=timedelta(hours=timezone_offset))
+        self.timezone_offset = timezone_offset or CONFLUENCE_TIMEZONE_OFFSET
```
pablonyx, line 110: The argument `timezone_offset` is `timezone_offset: float = CONFLUENCE_TIMEZONE_OFFSET` - so we can probably remove this

backend/danswer/connectors/confluence/connector.py
```diff
@@ -249,13 +248,11 @@ def load_from_state(self) -> GenerateDocumentsOutput:
...
     def poll_source(self, start: float, end: float) -> GenerateDocumentsOutput:
-        # Add time filters
-        formatted_start_time = datetime.fromtimestamp(start, tz=self.timezone).strftime(
-            "%Y-%m-%d %H:%M"
-        )
-        formatted_end_time = datetime.fromtimestamp(end, tz=self.timezone).strftime(
-            "%Y-%m-%d %H:%M"
-        )
+        # Convert float timestamps to datetime, then adjust for timezone
+        start_dt = datetime.fromtimestamp(start) + timedelta(hours=self.timezone_offset)
```
pablonyx, line 252: We should specify timezone here

pablonyx (changes requested):

backend/danswer/connectors/confluence/connector.py
hagen-danswer: true!

etiennellipse:
About that; setting an offset will require us to update it when switching from/to Daylight Savings Time.

EST => EDT, EDT => EST

Setting the name of the timezone would be more permanent. 

Although... there should be a way to fix this without configuration! But I guess you are limited by the Confluence REST API somehow!?

PR #3459, opened 2024-12-14
Remove hardcoded root path for HF models

emerzon:
## Description
Remove hardcoded root path for HF models, using user home dir instead

## How Has This Been Tested?
Model server started and correctly loaded models
```
NOTICE:   12/13/2024 08:27:53 PM               custom_models.py   99: Loaded model from local cache: /home/egomes/.cache/huggingface/hub/models--danswer--hybrid-intent-token-classifier/snapshots/54ed7e946f84a252c5d4e5d7a0bc839f466e9c64
```

N/A

N/A


Weves (approved):
Awesome, thanks for this!

PR #3460, opened 2024-12-14
Proper user deletion / organization leaving

pablonyx:
## Description
Allow any admin to leave their organization (in multi tenant case only)
Note that this will prevent them from creating a new instance if they were the owner of a previous one (intentional, but perhaps not the flow we want).

## How Has This Been Tested?
- Create and try to leave as invited tenant in single-tenant + multi tenant case
- Should not be possible in single tenant case

Tested running multi tenant locally and
- Leaving as single admin user
- Leaving as admin user among admin users (other admin users should still persist)
- Leaving as admin user with basic users (users should create new organization next time they log in)

N/A

## Mental Checklist:


backend/onyx/db/users.py
```diff
@@ -185,3 +192,43 @@ def batch_add_ext_perm_user_if_not_exists(
     db_session.commit()
 
     return found_users + new_users
+
+
+def delete_user_from_db(
+    user_to_delete: User,
+    db_session: Session,
+) -> None:
+    for oauth_account in user_to_delete.oauth_accounts:
```
yuhongsun96, line 201: deleting users is often kind of tricky, we need to delete everything related to that user no? Like chat sessions, files in the blob store, etc?

yuhongsun96 (approved):

PR #3463, opened 2024-12-14
Auth specific rate limiting

pablonyx:
## Description
Rate limit auth endpoints

Screenshot 2024-12-16 at 7 06 47 PM

## How Has This Been Tested?
For auth type basic, google auth and cloud:
- Try logging in with and without rate limiting enabled
- Spam the register / signin with rate limiting enabled and verify the 429.

N/A

N/A

## Mental Checklist:

### DIFF
--- a/backend/onyx/configs/app_configs.py
+++ b/backend/onyx/configs/app_configs.py
@@ -179,6 +179,10 @@ REDIS_HOST = os.environ.get("REDIS_HOST") or "localhost"
 REDIS_PORT = int(os.environ.get("REDIS_PORT", 6379))
 REDIS_PASSWORD = os.environ.get("REDIS_PASSWORD") or ""
 
+# Rate limiting for auth endpoints
+RATE_LIMIT_WINDOW_SECONDS = 60
+RATE_LIMIT_MAX_REQUESTS = 10
+
 # Used for general redis things
 REDIS_DB_NUMBER = int(os.environ.get("REDIS_DB_NUMBER", 0))
 
--- a/backend/onyx/configs/app_configs.py
+++ b/backend/onyx/configs/app_configs.py
@@ -180,8 +180,15 @@ REDIS_PORT = int(os.environ.get("REDIS_PORT", 6379))
 REDIS_PASSWORD = os.environ.get("REDIS_PASSWORD") or ""
 
 # Rate limiting for auth endpoints
-RATE_LIMIT_WINDOW_SECONDS = 60
-RATE_LIMIT_MAX_REQUESTS = 10
+try:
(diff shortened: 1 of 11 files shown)


backend/onyx/server/middleware/rate_limiting.py
```diff
@@ -0,0 +1,46 @@
...
+
+from onyx.configs.app_configs import RATE_LIMIT_MAX_REQUESTS
+from onyx.configs.app_configs import RATE_LIMIT_WINDOW_SECONDS
+from onyx.configs.app_configs import REDIS_HOST
+from onyx.configs.app_configs import REDIS_PASSWORD
+from onyx.configs.app_configs import REDIS_PORT
+
+
+async def setup_limiter() -> None:
+    redis = await aioredis.from_url(
```
yuhongsun96: Should this be centralized somewhere related to redis?

backend/onyx/server/middleware/rate_limiting.py
```diff
@@ -0,0 +1,46 @@
...
+        f"redis://{REDIS_HOST}:{REDIS_PORT}", password=REDIS_PASSWORD
+    )
+    await FastAPILimiter.init(redis)
+
+
+async def close_limiter() -> None:
+    await FastAPILimiter.close()
```
yuhongsun96, line 22: Can we add a comment for what this does? It's not immediately obvious why it's needed

backend/onyx/server/middleware/rate_limiting.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def rate_limit_key(request: Request) -> str:
+    return (
+        request.client.host if request.client else "unknown"
```
yuhongsun96: In what cases is this unknown? Is there any worry that in majority of cases it may be unknown and we'll be blocking a ton of signups?

pablonyx: (Referring to updated logic) Usually, the IP is automatically passed to the server, but if a request goes through certain proxies or services that remove IP info, we fall back to “unknown.” Similarly, if a script or device doesn't send a User-Agent header (e.g., certain bots), we default to “none,” which can lead to collisions here

backend/onyx/configs/app_configs.py
```diff
@@ -179,6 +179,25 @@
 REDIS_PORT = int(os.environ.get("REDIS_PORT", 6379))
 REDIS_PASSWORD = os.environ.get("REDIS_PASSWORD") or ""
 
+# Rate limiting for auth endpoints
+
+
+RATE_LIMIT_WINDOW_SECONDS: int | None = None
```
yuhongsun96, line 185: Could be a shared util function to do something like
def _safe_int_env(env_var: str) -> int | None:

yuhongsun96 (approved):
lgtm, if it works then it's all good!

PR #3468, opened 2024-12-15
Proper display priority seeding

pablonyx:
## Description
- Default assistant created in alembic migrations has `null` display priority and would not be overridden by yaml loading
- Display priority needs to update for existing assistants (if we decide to migrate

## How Has This Been Tested?

N/A

N/A

## Mental Checklist:

### DIFF
--- a/backend/onyx/seeding/load_yamls.py
+++ b/backend/onyx/seeding/load_yamls.py
@@ -48,6 +48,7 @@ def load_personas_from_yaml(
         data = yaml.safe_load(file)
 
     all_personas = data.get("personas", [])
+
     for persona in all_personas:
         doc_set_names = persona["document_sets"]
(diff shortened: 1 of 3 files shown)


backend/onyx/seeding/load_yamls.py
```diff
@@ -127,6 +128,7 @@ def load_personas_from_yaml(
             display_priority=(
```
rkuo-danswer, line 128: what happens if both existing_persona.display_priority and persona.get("display_priority") are None. Is that OK?

pablonyx: This should never happen - we can raise an exception but I can't imagine a scenario in which this would occur

rkuo-danswer (approved):

PR #3494, opened 2024-12-17
Pagination Hook

skylares:
## Description
A hook to simplify receiving paginated data from the db.

usertable

## Backend Changes:

- Added an API endpoint and DB functions to return paginated user data and to allow filtering by user role and status.
- Added the necessary types for the paginated return of users and indexing attempts.
- Added tests for the users API to ensure pagination and filtering work as intended.
- Modified the indexing attempts API endpoint to use the updated return type.

## Frontend Changes:
- Added a hook for handling pagination
- Updated the users table to use the hook
- Updated the indexing attempts table to use the hook
- Added filtering by role and status for the users table


backend/onyx/db/users.py
```diff
@@ -110,13 +113,84 @@ def list_users(
...
+    # Init where clause and remove any users with API email domains
+    where_clause = [not_(User.email.endswith(DANSWER_API_KEY_DUMMY_EMAIL_DOMAIN))]  # type: ignore
+
+    # Exclude external permissioned users
+    if not include_external:
+        where_clause.append(User.role != UserRole.EXT_PERM_USER)
+
+    if email_filter_string:
+        where_clause.append(User.email.ilike(f"%{email_filter_string}%"))  # type: ignore
+
```
skylares, line 154: Used "type: ignore" here referencing what the preexisting function fetching users did

web/src/components/admin/users/buttons/InviteUserButton.tsx
```diff
@@ -35,17 +38,23 @@ export const InviteUserButton = ({
     {
       onSuccess: () => {
         setShowInviteModal(false);
-        mutate();
+        if (typeof mutate === "function") {
+          mutate();
+        } else {
+          mutate.forEach((fn) => fn());
+        }
```
skylares, line 45: invited users and accepted have different mutate functions, but needed to call both at once to update slack users invite status properly, so added the ability to pass in an array of mutate functions.

backend/onyx/db/users.py
```diff
@@ -110,13 +113,84 @@ def list_users(
     return db_session.scalars(stmt).unique().all()
 
 
+def _get_accepted_user_where_clause(
+    email_filter_string: str = "",
+    status_filter: UserStatus | None = None,
+    roles_filter: list[UserRole] = [],
+    include_external: bool = False,
+) -> list:
+    # Init where clause and remove any users with API email domains
```
yuhongsun96: Nit: better to comment these functions with some basic docstring explaining what it does and the uses. Some functions are clear what they do from scanning the code, this one requires readers to trace it up the stack to the callers etc to have a sense of why it's like this (like why does it have an email filter) so it should have a comment.

backend/onyx/db/users.py
```diff
@@ -110,13 +113,84 @@ def list_users(
     return db_session.scalars(stmt).unique().all()
 
 
+def _get_accepted_user_where_clause(
+    email_filter_string: str = "",
```
yuhongsun96: Better to differentiate between empties and real Nulls, so pref:
str | None = None
with corresponding handling in the function for the real Null case

backend/onyx/db/users.py
yuhongsun96: Pref only use type: ignore when absolutely necessary. Would fix it in the other spot you referenced too
image
LLM explains it pretty well here, better than I could, see image

backend/onyx/db/users.py
```diff
@@ -110,13 +113,84 @@ def list_users(
...
+    if email_filter_string:
+        where_clause.append(User.email.ilike(f"%{email_filter_string}%"))  # type: ignore
+
+    if roles_filter:
+        where_clause.append(User.role.in_(roles_filter))
+
+    # When status_filter = "live" the inner condition evaluates True, if status_filter is "deactivated" we get False.
+    # so if status_filter = "live" we select only active users, and if status_filter = "deactivated" we select only inactive users
+    if status_filter:
+        where_clause.append(User.is_active == (status_filter == UserStatus.LIVE))
```
yuhongsun96: Unintuitive logic here. If you pass in a UserStatus, it should filter by that one, not a one vs all type filter. If you want a one vs all type filter, it should be a boolean flag that is accordingly named.

backend/onyx/db/users.py
```diff
@@ -110,13 +113,84 @@ def list_users(
...
+
+    return where_clause
+
+
+def get_page_of_filtered_users(
+    db_session: Session,
+    page_size: int,
+    page: int,
```
yuhongsun96: Nit: with python naming, often types "page" references an object, better to have it as page_num or something. The typing makes it clear for the most part but this would be a tiny tiny improvement

backend/onyx/db/users.py
```diff
@@ -110,13 +113,84 @@ def list_users(
...
+    email_filter_string: str = "",
```
yuhongsun96: same as before (I will stop mentioning it elsewhere), nulls should be handled explicitly

backend/onyx/db/users.py
```diff
@@ -110,13 +113,84 @@ def list_users(
     return db_session.scalars(stmt).unique().all()
 
 
+def _get_accepted_user_where_clause(
+    email_filter_string: str = "",
+    status_filter: UserStatus | None = None,
+    roles_filter: list[UserRole] = [],
+    include_external: bool = False,
+) -> list:
```
yuhongsun96: would type it more strictly here, list of what?

backend/onyx/server/manage/users.py
```diff
@@ -112,6 +114,59 @@ def set_user_role(
...
 
 
+@router.get("/manage/users/accepted")
+def list_accepted_users(
+    q: str | None = Query(default=None),
+    page: int = Query(1, ge=1),
+    page_size: int = Query(10, ge=1, le=1000),
+    roles: list[UserRole] = Query(default=[]),
+    status: UserStatus | None = Query(default=None),
+    user: User | None = Depends(current_curator_or_admin_user),
```
yuhongsun96: Should curators be able to list users/statuses?

(also nit: unused variables should just be _ )

backend/onyx/server/documents/models.py
```diff
@@ -201,26 +205,19 @@ def from_db_model(cls, error: DbIndexAttemptError) -> "IndexAttemptError":
...
-        return cls(
-            index_attempts=[
-                IndexAttemptSnapshot.from_index_attempt_db_model(index_attempt_model)
-                for index_attempt_model in index_attempt_models
-            ],
-            page=page,
-            total_pages=total_pages,
-        )
+
+class PaginatedReturn(BaseModel, Generic[PaginatedType]):
```
yuhongsun96, line 218: Consider adding an index for where we are in the list of the total items. Good use of generic here, very correct :)

backend/onyx/server/manage/users.py
```diff
@@ -112,6 +114,59 @@ def set_user_role(
...
+    db_session: Session = Depends(get_session),
+) -> PaginatedReturn[FullUserSnapshot]:
+    if not q:
+        q = ""
```
yuhongsun96: it's good that it's an explicit null!

backend/onyx/server/manage/users.py
```diff
@@ -112,6 +114,59 @@ def set_user_role(
...
+        items=[
+            FullUserSnapshot.from_user_model(user) for user in filtered_accepted_users
+        ],
+        total_items=total_accepted_users_count,
+    )
+
+
+@router.get("/manage/users/invited")
+def list_invited_users(
+    user: User | None = Depends(current_curator_or_admin_user),
```
yuhongsun96: Same here and below, I think probably only admins should be able to get user information like this

backend/onyx/server/manage/users.py
```diff
@@ -411,7 +466,7 @@ def list_all_users_basic_info(
     _: User | None = Depends(current_user),
```
yuhongsun96, line 456: I think here as well? What is this used for actually? I tend to not read frontend code so plz double check that it is absolutely required that any user can access every user in the system's uuid and email

backend/tests/integration/common_utils/managers/user.py
```diff
@@ -89,7 +95,13 @@ def login_as_user(test_user: DATestUser) -> DATestUser:
         return test_user
 
     @staticmethod
-    def verify_role(user_to_verify: DATestUser, target_role: UserRole) -> bool:
+    def verify_role(
```
yuhongsun96: This doesn't really follow typical test patterns but the existing is already like this so whatever. The problems here are:

Separation of concerns: state-setup/mock functionality should not also be responsible for verification

Consider more "assert" style naming so that the caller can be something like:
assert UserManager.is_role(mock_user, UserRole.ADMIN) or something like that

The verify role accepting a None role also feels quite unexpected.

backend/onyx/server/manage/users.py
```diff
@@ -112,6 +114,59 @@ def set_user_role(
     db_session.commit()
 
 
+@router.get("/manage/users/accepted")
+def list_accepted_users(
+    q: str | None = Query(default=None),
+    page: int = Query(1, ge=1),
```
yuhongsun96: Page number should be decremented to 0 default ASAP. The rest of the code underneath the route should be 0 indexed. It shouldn't be the lowest level (right before DB) that does the decrementing, otherwise it can lead to intermediate functions being used wrong as the default in backend services is to use 0 indexing.

backend/tests/integration/common_utils/managers/user.py
```diff
@@ -101,41 +113,118 @@ def verify_role(user_to_verify: DATestUser, target_role: UserRole) -> bool:
     def set_role(
         user_to_set: DATestUser,
         target_role: UserRole,
-        user_to_perform_action: DATestUser | None = None,
-    ) -> None:
-        if user_to_perform_action is None:
-            user_to_perform_action = user_to_set
+        user_performing_action: DATestUser | None = None,
+    ) -> DATestUser:
```
yuhongsun96, line 125: In OOP context setter methods typically modify objects in place and do not return something separate. However, we are trying to be more functional. So prefer returning a separate object with the new state, but it's very important then to not also modify the state of the existing object. That becomes very unexpected for the caller.

backend/tests/integration/common_utils/managers/user.py
```diff
@@ -101,41 +113,118 @@ def verify_role(user_to_verify: DATestUser, target_role: UserRole) -> bool:
...
-            headers=user_to_perform_action.headers,
+            headers=user_performing_action.headers,
+        )
+        response.raise_for_status()
+        user_to_set.role = target_role
+        return user_to_set
+
+    @staticmethod
+    def verify_status(
+        user_to_verify: DATestUser, target_status: UserStatus | None = None
```
yuhongsun96: Same as earlier, verify_status that takes a null status is very unintuitive.

backend/tests/integration/common_utils/managers/user.py
```diff
@@ -101,41 +113,118 @@ def verify_role(user_to_verify: DATestUser, target_role: UserRole) -> bool:
...
+        response = requests.get(
+            url=f"{API_SERVER_URL}/me",
+            headers=user_to_verify.headers,
+        )
+        response.raise_for_status()
+
+        is_active = response.json().get("is_active", None)
+        if is_active is None:
+            raise KeyError("The 'is_active' field is not found in the user's info")
+        return target_status == UserStatus("live" if is_active else "deactivated")
```
yuhongsun96: Avoid this type of functionality where the function takes an Enum and only accepts 2 possible values when this is framed as a generic test.

Can consider updating the /me endpoint (without breaking functionality) or have special handling for the other cases so it doesn't just always fail if someone wants to check UserStatus invited

backend/tests/integration/common_utils/managers/user.py
```diff
@@ -101,41 +113,118 @@ def verify_role(user_to_verify: DATestUser, target_role: UserRole) -> bool:
...
+
+    @staticmethod
+    def set_status(
+        user_to_set: DATestUser,
+        target_status: UserStatus,
+        user_performing_action: DATestUser | None = None,
+    ) -> DATestUser:
+        if user_performing_action is None:
+            user_performing_action = user_to_set
+        if target_status == UserStatus.LIVE:
```
yuhongsun96: This mapping of 3 to 2 here is not ideal (same issue as above)

backend/tests/integration/common_utils/managers/user.py
```diff
@@ -101,41 +113,118 @@ def verify_role(user_to_verify: DATestUser, target_role: UserRole) -> bool:
...
+        users_list = []
+        for i in range(1, count + 1):
+            user = UserManager.create(name=f"{user_name_prefix}_{i}")
+            if role != UserRole.BASIC:
+                user = UserManager.set_role(user, role, user_performing_action)
+            if status:
+                user = UserManager.set_status(user, status, user_performing_action)
+
+            if status == UserStatus.DEACTIVATED:
+                with pytest.raises(HTTPError):
```
yuhongsun96: utilities should not be doing testing

backend/tests/integration/common_utils/managers/user.py
```diff
@@ -101,41 +113,118 @@ def verify_role(user_to_verify: DATestUser, target_role: UserRole) -> bool:
...
+        paginated_result = PaginatedReturn(
+            items=[FullUserSnapshot(**user) for user in data["items"]],
+            total_items=data["total_items"],
         )
-        for accepted_user in all_users.accepted:
-            if accepted_user.email == user.email and accepted_user.id == user.id:
-                return
-        raise ValueError(f"User {user.email} not found")
+        # Verify that the number of items in the page is equal to the page size
+        assert len(paginated_result.items) == page_size
```
yuhongsun96: utilities should not be doing tests

yuhongsun96 (commented):
Good work! I pointed out a lot of small things that aren't critical but I'm trying to be helpful to your learning so being extra nit picky.

Stopped at the test utilities file, ran out of time and have to move on to other work, Hagen will take a pass after this.

backend/onyx/server/manage/users.py
skylares: From what I can tell this is only used by the "Your Assistants" page found by clicking the Manage Assistants tab in the homepage sidebar. It's used to display users emails when you share a private assistant with others. The ID's are used to ensure you can't select any users who you've already selected or who have already had the assistant shared with them. 

Unsure if this is intended but I noticed if an assistant is shared with you, everyone will have the share button option and can see the list of users. Only other admins seem to be able to actually share with others in this case, but they can also remove people from the share list whether they own it or not.

web/src/components/admin/users/InvitedUserTable.tsx
```diff
@@ -6,29 +7,62 @@ import {
...
+  // Filter users based on the search query
+  const filteredUsers = q
+    ? users.filter((user) => user.email.includes(q))
+    : users;
+
+  // Get the current page of users
+  const currentPageOfUsers = filteredUsers.slice(
+    (currentPageNum - 1) * 10,
+    currentPageNum * 10
+  );
```
pablonyx: USERS_PER_PAGE for consistency

web/src/components/admin/users/SignedUpUserTable.tsx
```diff
@@ -67,56 +252,54 @@ const SignedUpUserTable = ({
...
+                .map((user) => (
+                  <TableRow key={user.id}>
+                    <TableCell>{user.email}</TableCell>
+                    <TableCell className="w-[180px]">
+                      {renderUserRoleDropdown(user)}
+                    </TableCell>
+                    <TableCell className="text-center w-[140px]">
+                      <i>{user.status === "live" ? "Active" : "Inactive"}</i>
+                    </TableCell>
+                    <TableCell className="text-right w-[200px]">
```
pablonyx: A little concerned about checking length and filtering separately

web/src/hooks/usePaginatedFetch.tsx
```diff
@@ -0,0 +1,237 @@
...
+  const currentPath = usePathname();
+
+  // State to initialize and hold the current page number
+  const [currentPage, setCurrentPage] = useState(() => {
+    if (typeof window !== "undefined") {
+      const urlParams = new URLSearchParams(window.location.search);
+      return parseInt(urlParams.get("page") || "1", 10);
+    }
+    return 1;
+  });
```
pablonyx, line 56: Is there a reason we're not using `useSearchParams`?

pablonyx: Would prefer using a util function here

web/src/hooks/usePaginatedFetch.tsx
```diff
@@ -0,0 +1,237 @@
...
+            if (Array.isArray(value)) {
+              value.forEach((str) => params.append(key, str));
+            } else {
+              params.set(key, value.toString());
+            }
+          }
+        }
+
+        const response = await fetch(`${endpoint}?${params.toString()}`);
+        if (!response.ok) throw new Error("Failed to fetch data");
```
pablonyx: Might want more verbose error handling here

PR #3522, opened 2024-12-20
associating credentials with connectors is not considered editing

hagen-danswer:
## Description
Changed the logic to apply filters such that relating a credential to a connector doesnt count as editing

## How Has This Been Tested?
Created a curator public credential as admin and used it to create a cc pair as a curator. before this change this would not work and after it would.

### DIFF
--- a/backend/onyx/db/credentials.py
+++ b/backend/onyx/db/credentials.py
@@ -152,10 +152,16 @@ def fetch_credential_by_id(
     user: User | None,
     db_session: Session,
     assume_admin: bool = False,
+    get_editable: bool = True,
 ) -> Credential | None:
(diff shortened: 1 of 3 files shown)


backend/onyx/db/credentials.py
```diff
@@ -152,10 +153,16 @@ def fetch_credential_by_id(
     user: User | None,
     db_session: Session,
     assume_admin: bool = False,
+    get_editable: bool = True,
```
hagen-danswer, line 156: this function now takes in get_editable and passes it forward (for use by cc_pair creation/deletion) instead of not indicating it and always defaulting to the less permissive option

yuhongsun96 (approved):

PR #3536, opened 2024-12-21
Curators can now update the curator relationship

hagen-danswer:
## Description
Curators can now promote basic users to curators of a group and demote them 
curators can also demote themselves but a modal pops up asking for confirmation:

Screen Recording 2024-12-21 at 9 01 51 AM

## How Has This Been Tested?
Watch the video. A in top right is admin. other 2 accounts are curator/basic

### DIFF
--- a/backend/ee/onyx/db/user_group.py
+++ b/backend/ee/onyx/db/user_group.py
@@ -2,7 +2,6 @@ from collections.abc import Sequence
 from operator import and_
 from uuid import UUID
 
-from fastapi import HTTPException
 from sqlalchemy import delete
 from sqlalchemy import func
 from sqlalchemy import Select
@@ -120,59 +119,6 @@ def _cleanup_document_set__user_group_relationships__no_commit(
             DocumentSet__UserGroup.user_group_id == user_group_id
         )
     )
-
-
-def validate_object_creation_for_user(
-    db_session: Session,
-    user: User | None,
-    target_group_ids: list[int] | None = None,
-    object_is_public: bool | None = None,
-    object_is_perm_sync: bool | None = None,
-) -> None:
-    """
-    All users can create/edit permission synced objects if they don't specify a group
-    All admin actions are allowed.
-    Prevents non-admins from creating/editing:
-    - public objects
-    - objects with no groups
-    - objects that belong to a group they don't curate
-    """
-    if object_is_perm_sync and not target_group_ids:
-        return
-
-    if not user or user.role == UserRole.ADMIN:
-        return
-
-    if object_is_public:
-        detail = "User does not have permission to create public credentials"
-        logger.error(detail)
-        raise HTTPException(
-            status_code=400,
-            detail=detail,
-        )
-    if not target_group_ids:
-        detail = "Curators must specify 1+ groups"
-        logger.error(detail)
-        raise HTTPException(
-            status_code=400,
-            detail=detail,
-        )
-
-    user_curated_groups = fetch_user_groups_for_user(
-        db_session=db_session,
-        user_id=user.id,
-        # Global curators can curate all groups they are member of
-        only_curator_groups=user.role != UserRole.GLOBAL_CURATOR,
-    )
-    user_curated_group_ids = set([group.id for group in user_curated_groups])
-    target_group_ids_set = set(target_group_ids)
-    if not target_group_ids_set.issubset(user_curated_group_ids):
-        detail = "Curators cannot control groups they don't curate"
-        logger.error(detail)
-        raise HTTPException(
-            status_code=400,
-            detail=detail,
-        )
 
 
 def fetch_user_group(db_session: Session, user_group_id: int) -> UserGroup | None:
--- a/backend/ee/onyx/db/user_group.py
+++ b/backend/ee/onyx/db/user_group.py
@@ -122,7 +122,7 @@ def _cleanup_document_set__user_group_relationships__no_commit(
     )
 
 
-def validate_user_creation_permissions(
+def validate_object_creation_for_user(
     db_session: Session,
     user: User | None,
     target_group_ids: list[int] | None = None,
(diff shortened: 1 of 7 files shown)


web/src/app/ee/admin/groups/[groupId]/GroupDisplay.tsx
```diff
@@ -95,30 +101,61 @@ const UserRoleDropdown = ({
...
-  } else {
-    return <div>{USER_ROLE_LABELS[localRole]}</div>;
-  }
+    user.role === UserRole.BASIC || user.role === UserRole.CURATOR;
+
+  return (
+    <>
+      {/* Confirmation modal - only shown when users try to demote themselves */}
+      {showDemoteConfirm && pendingRoleChange && (
+        <GenericConfirmModal
```
Comment (pablonyx) [inline web/src/app/ee/admin/groups/[groupId]/GroupDisplay.tsx:120]: Would refresh user here or an alternative- user should likely be redirected to chat page

pablonyx:
Looks good! Tested locally with admin + global curator/basic + curator/basic user

pablonyx (approved):

backend/ee/onyx/db/user_group.py
```diff
@@ -440,32 +440,108 @@ def remove_curator_status__no_commit(db_session: Session, user: User) -> None:
...
 ) -> None:
-    user = fetch_user_by_id(db_session, set_curator_request.user_id)
-    if not user:
-        raise ValueError(f"User with id '{set_curator_request.user_id}' not found")
+    """
+    This function validates that the user making the change has the necessary permissions
+    to update the curator relationship for the target user in the given user group.
+    """
+
+    if user_making_change is None or user_making_change.role == UserRole.ADMIN:
```
yuhongsun96, line 453: For avoidance of misuse of these functions, better to check that AUTH_TYPE is disabled rather than user is None. This allows more safer reuse. In this particular case, it is enforced correctly at the API level however, we don't want to set "traps" in the code.

backend/ee/onyx/db/user_group.py
```diff
@@ -440,32 +440,108 @@ def remove_curator_status__no_commit(db_session: Session, user: User) -> None:
...
+        return
 
-    if user.role == UserRole.ADMIN:
+    # check if the user making the change is a curator in the group they are changing the curator relationship for
+    user_making_change_curator_groups = fetch_user_groups_for_user(
+        db_session=db_session,
+        user_id=user_making_change.id,
+        # only check if the user making the change is a curator if they are a curator
```
yuhongsun96, line 460: like a group level curator? "only check if the user making the change is a curator if they are a curator" seems circular no?

backend/ee/onyx/db/user_group.py
```diff
@@ -440,32 +440,108 @@ def remove_curator_status__no_commit(db_session: Session, user: User) -> None:
...
+        only_curator_groups=user_making_change.role == UserRole.CURATOR,
+    )
+    requestor_curator_group_ids = [
+        group.id for group in user_making_change_curator_groups
+    ]
+    if user_group_id not in requestor_curator_group_ids:
         raise ValueError(
-            f"User '{user.email}' is an admin and therefore has all permissions "
+            f"user making change {user_making_change.email} is not a curator,"
+            f" admin, or global_curator for group '{user_group_id}'"
```
yuhongsun96, line 471: This also seems wrong, "global curator for group" is nonsensical no?

backend/ee/onyx/db/user_group.py
```diff
@@ -440,32 +440,108 @@ def remove_curator_status__no_commit(db_session: Session, user: User) -> None:
...
 
-    if user.role == UserRole.ADMIN:
+    # check if the user making the change is a curator in the group they are changing the curator relationship for
+    user_making_change_curator_groups = fetch_user_groups_for_user(
+        db_session=db_session,
+        user_id=user_making_change.id,
+        # only check if the user making the change is a curator if they are a curator
+        # otherwise, they are a global_curator and can update the curator relationship
+        # for any group they are a member of
+        only_curator_groups=user_making_change.role == UserRole.CURATOR,
```
yuhongsun96, line 463: Logic should be something like, if they are an admin or global curator, don't run this check at all. If they are a normal curator, run this check and this flag is true, if they are anything else, raise an exception

yuhongsun96: Technically with this being a local function, it's probably not going to cause any problems if we don't have this extremely clear logic but still better to do it this way no?

backend/ee/onyx/db/user_group.py
```diff
@@ -440,32 +440,108 @@ def remove_curator_status__no_commit(db_session: Session, user: User) -> None:
...
+    if target_user.role == UserRole.ADMIN:
+        raise ValueError(
+            f"User '{target_user.email}' is an admin and therefore has all permissions "
             "of a curator. If you'd like this user to only have curator permissions, "
             "you must update their role to BASIC then assign them to be CURATOR in the "
             "appropriate groups."
         )
+    elif target_user.role == UserRole.GLOBAL_CURATOR:
+        raise ValueError(
+            f"User '{target_user.email}' is a global_curator and therefore has all "
```
yuhongsun96, line 492: Is this propagated to the frontend? Best way typically is to raise a backend only error, catch it near the top near the API call and translate it to a user friendly error message.

We don't want to float up "dev facing" errors to the end user, even if it's admins/curator users. So nothing like "variable_underscore_something" in user facing flows.

backend/ee/onyx/db/user_group.py
```diff
@@ -440,32 +440,108 @@ def remove_curator_status__no_commit(db_session: Session, user: User) -> None:
...
+
+
+def update_user_curator_relationship(
+    db_session: Session,
+    user_group_id: int,
+    set_curator_request: SetCuratorRequest,
+    user_making_change: User | None = None,
+) -> None:
+    target_user = fetch_user_by_id(db_session, set_curator_request.user_id)
+    if not target_user:
```
yuhongsun96, line 525: Use explicit Nones where possible

backend/ee/onyx/db/user_group.py
```diff
@@ -440,32 +440,108 @@ def remove_curator_status__no_commit(db_session: Session, user: User) -> None:
     _validate_curator_status__no_commit(db_session, [user])
 
 
-def update_user_curator_relationship(
+def _validate_curator_relationship_update_requester(
     db_session: Session,
     user_group_id: int,
-    set_curator_request: SetCuratorRequest,
+    user_making_change: User | None = None,
```
yuhongsun96, line 446: pref not default this to None

backend/ee/onyx/db/user_group.py
hagen-danswer: they are not a global curator for this group is true

backend/ee/onyx/db/user_group.py
hagen-danswer: sure i guess it could be worded "only enforce the existance of a curator relationship if they are a curator, ..."

backend/ee/onyx/db/user_group.py
hagen-danswer: this is handled on line 453

backend/ee/onyx/db/user_group.py
hagen-danswer: unsure, the propagation of errors to the frontend is currently unstandardized. probably worth taking a cohesive pass if we would like to enforce standards

backend/ee/onyx/db/user_group.py
hagen-danswer: fair

backend/ee/onyx/db/user_group.py
hagen-danswer: fair

backend/ee/onyx/db/user_group.py
yuhongsun96: I see that but there is a different organization of code that I prefer, for example, this flag is always True, but if they are a normal user or some other role type, we throw an exception somewhere above.

backend/ee/onyx/db/user_group.py
yuhongsun96: Ok let me know if I'm interpreting this correctly, it should likely be something like "is not an admin, a global curator, or a group curator for the group <group id>"

backend/ee/onyx/db/user_group.py
yuhongsun96: Chatted with Chris, it's not worth the overhead to standardize this. The only comment here is to not have dev facing terms and user facing terms in the same message. Actually we should just never have dev facing terms in any user facing flows.

PR #3552, opened 2024-12-25
More efficient Vespa indexing

pablonyx:
## Description
Curent process:
- Check for existing docs, delete existing values, index into Vespa

New approach:
- Use`doc_updated_at` as a versioning scheme.
- Instead of checking for existing docs with visit API, we can just a) ensure consisted updated at time, index the documents, and then delete en masse documents with a different "version" (ie. `doc_updated_at` value)

Quick notes:
- We ensure all chunks for a given document share the same doc_updated_at, raise an error if there is a mismatch.  
- Can remove `fresh_index` flag as deletion of old chunks is much less intensive now
- We now use a quick `_does_doc_exist_in_vespa` check to maintain the functionality of determining which documents are wholly new

## How Has This Been Tested?
Various indexing + reindexing + pull updates for 4 connectors (file, web, confluence, Github)
- + Ensure that `already_existed` is accurate


backend/onyx/document_index/vespa/shared_utils/vespa_request_builders.py
```diff
@@ -106,3 +107,25 @@ def build_vespa_id_based_retrieval_yql(
 
     id_based_retrieval_yql_section += ")"
     return id_based_retrieval_yql_section
+
+
+def build_deletion_selection(doc_id: str, version_cutoff: int, doc_type: str) -> str:
```
rkuo-danswer: Any way to test this?  Or should we put this in a follow on PR?

backend/onyx/document_index/vespa/indexing_utils.py
```diff
@@ -248,3 +252,84 @@ def clean_chunk_id_copy(
...
+    url = (
+        f"{SEARCH_ENDPOINT}"
+        f'?yql=select+*+from+sources+{index_name}+where+document_id+contains+"{encoded_doc_id}"'
+        "&hits=0"
+    )
+
+    logger.debug(f"Checking existence for doc_id={doc_id} with URL={url}")
+    resp = http_client.get(url)
+
+    if resp.status_code == 200:
```
rkuo-danswer, line 279: The flow here is probably easier to follow if we treat 200 as the happy path and the other paths as exceptions/failures.

backend/onyx/document_index/vespa/shared_utils/vespa_request_builders.py
```diff
@@ -106,3 +107,25 @@ def build_vespa_id_based_retrieval_yql(
...
+    # Escape single quotes by doubling them for Vespa selection expressions
+    escaped_doc_id = doc_id.replace("'", "''")
+
+    return (
+        f"({doc_type}.document_id=='{escaped_doc_id}') and "
+        f"({doc_type}.doc_updated_at < {version_cutoff})"
+    )
+
+
+def build_selection_query_param(filter_str: str) -> str:
```
rkuo-danswer: Does this get used in more than one place?  Maybe it should just be moved inline with the calling function?

backend/onyx/document_index/vespa/index.py
```diff
@@ -307,47 +314,61 @@ def register_multitenant_indices(
...
+                    )
+                    chunk.source_document.doc_updated_at = chunk_val
+                doc_to_doc_updated_at[doc_id] = chunk_val
+            else:
+                # If the chunk's doc_updated_at is missing, fill it from the stored value
+                if chunk_val is None:
+                    chunk.source_document.doc_updated_at = doc_to_doc_updated_at[doc_id]
+                else:
+                    # Raise if doc_updated_at conflicts with what we've already stored for this doc_id
+                    assert chunk_val == doc_to_doc_updated_at[doc_id], (
```
rkuo-danswer: we probably mean to raise an exception here

backend/onyx/document_index/vespa/index.py
```diff
@@ -307,47 +314,61 @@ def register_multitenant_indices(
...
-        multiple chunk batches calling this function multiple times, otherwise only the last set of
-        chunks will be kept"""
+        """
+        Index a list of chunks into Vespa, optionally removing older chunk versions.
+        Raises an AssertionError if multiple chunks for the same doc_id have conflicting doc_updated_at values.
+        """
+        current_timestamp = int(datetime.utcnow().timestamp())
+
+        # Step 1: Enforce a single doc_updated_at for each doc_id across all chunks.
+        doc_to_doc_updated_at = {}
```
rkuo-danswer: This is an interesting use of doc_updated_at ... I'm unclear on if it was being used for anything important before this that might be affected?  What are the tradeoffs vs using a new field?

backend/onyx/document_index/vespa/shared_utils/vespa_request_builders.py
pablonyx: There certainly is! If you query vespa for a specific number of docs, pull in updates to a connector (which now has fewer chunks), you should see fewer results for the doc ID when you query. Tested this one manually but could probably use a test!

backend/onyx/document_index/vespa/indexing_utils.py
pablonyx: This was definitely the intention- is there a way we can make it clearer that 200 is the happy path in the existing code?

backend/onyx/document_index/vespa/shared_utils/vespa_request_builders.py
pablonyx: There certainly is! If you query vespa for a specific number of docs, pull in updates to a connector (which now has fewer chunks), you should see fewer results for the doc ID when you query. Tested this one manually but could probably use a test!

backend/onyx/document_index/vespa/index.py
pablonyx: Good catch

backend/onyx/document_index/vespa/indexing_utils.py
rkuo-danswer: Could we try:

if 404 ... return
if not 200 ... return

200 - happy path

Weves: +1 to this pattern!

backend/onyx/document_index/vespa/index.py
```diff
@@ -312,17 +312,18 @@ def index(
...
+        Index a list of chunks into Vespa. We rely on 'current_index_time'
         to keep track of when each chunk was added/updated in the index. We also raise a ValueError
-        if any chunk is missing a 'last_indexed_at' timestamp.
+        if any chunk is missing a 'current_index_time' timestamp.
         """
 
         # Clean chunks if needed (remove invalid chars, etc.)
         cleaned_chunks = [clean_chunk_id_copy(chunk) for chunk in chunks]
 
         #  We will store the set of doc_ids that previously existed in Vespa
```
rkuo-danswer, line 323: nit: extra space

Weves (commented):
Approving for now, will need to think more about whether the `current_index_time` should be part of the chunk obj

backend/onyx/document_index/vespa/index.py
```diff
@@ -347,9 +348,9 @@ def index(
                     executor=executor,
                 )
 
-            # c) Remove chunks with using versioning scheme 'last_indexed_at'
+            # c) Remove chunks with using versioning scheme 'current_index_time'
```
rkuo-danswer, line 351: can deleting old chunks after feeding new chunks cause any issues?  As in returning out of date chunks in results?

backend/onyx/document_index/vespa/shared_utils/vespa_request_builders.py
rkuo-danswer: ok, let's add a linear issue to follow up

backend/onyx/document_index/vespa/index.py
Weves: Yes I believe so, although the old approach had issues too (e.g. post deletion, pre re-insertion doc would not show up). Hard to say which is preferable.

PR #3559, opened 2024-12-29
Fix Null Value in PG

yuhongsun96:
## Description
Connectors may give some invalid values for Postgres, catching those

## How Has This Been Tested?
Tested with normal connector and also by injecting a \x00 manually

### DIFF
--- a/backend/onyx/background/indexing/run_indexing.py
+++ b/backend/onyx/background/indexing/run_indexing.py
@@ -1,6 +1,5 @@
 import time
 import traceback
-from copy import deepcopy
 from datetime import datetime
 from datetime import timedelta
 from datetime import timezone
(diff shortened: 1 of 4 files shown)


backend/onyx/background/indexing/run_indexing.py
```diff
@@ -239,6 +240,34 @@ def _run_indexing(
 
                 batch_description = []
                 for doc in doc_batch:
+                    if "\x00" in doc.id:
```
Weves: please break this out into a separate utility function 😃

Weves: and we are modifying the doc in place!!! Let's create a copy and use that

Weves (changes requested):

Weves (approved):

PR #3564, opened 2024-12-31
Airtable connector

Weves:
## Description
Adds Airtable connector.

## How Has This Been Tested?
"Unit" test + manually creating connector + running it locally.


web/src/lib/sources.ts
```diff
@@ -299,19 +300,25 @@ export const SOURCE_METADATA_MAP: SourceMap = {
...
+    icon: EgnyteIcon,
+    displayName: "Egnyte",
+    category: SourceCategory.Storage,
+    docs: "https://docs.onyx.app/connectors/egnyte",
+  },
+  airtable: {
+    icon: AirtableIcon,
+    displayName: "Airtable",
+    category: SourceCategory.Other,
+    docs: "https://docs.onyx.app/connectors/airtable",
```
rkuo-danswer, line 313: docs link doesn't exist yet

backend/onyx/connectors/airtable/airtable_connector.py
```diff
@@ -0,0 +1,263 @@
...
+
+        if isinstance(field_info, list):
+            return [str(item) for item in field_info]
+
+        return [str(field_info)]
+
+    def _should_be_metadata(self, field_type: str) -> bool:
+        """Determine if a field type should be treated as metadata."""
+        # NOTE: all are made lowercase to avoid case sensitivity issues
+        metadata_field_types = {
```
rkuo-danswer: maybe this should be a class var / constant?

backend/onyx/connectors/airtable/airtable_connector.py
```diff
@@ -0,0 +1,263 @@
...
+                    metadata.update(field_metadata)
+
+                semantic_id = (
+                    f"{table_name}: {primary_field_value}"
+                    if primary_field_value
+                    else table_name
+                )
+
+                record_document = Document(
+                    id=str(record_id),
```
rkuo-danswer: odds of a collision here are probably low due to how the id's are formatted, but a prefix on the document id has helped avoid them in other connectors.

Weves: Good point, added!

backend/onyx/connectors/airtable/airtable_connector.py
Weves: Done! I did this for closer proximity between definition and use (e.g. no need to scroll up to check if something is metadata), but I can see both ways being good (e.g. this allows for easier re-use of the metadata types)

rkuo-danswer (approved):
test needs updating with id changing format, otherwise looks good.

PR #3568, opened 2024-12-31
Small Egnyte tweaks

Weves:
## Description
Fixes cases where file paths contain "#" or other chars that need to be URL encoded + removes logging of access token.

## How Has This Been Tested?
:eyes:


rkuo-danswer (approved):
looks ok, only comment is would a single url builder helper make sense here?

PR #3571, opened 2025-01-01
Add support for OAuth connectors that require user input

Weves:
## Description
Supports a new paradigm where the user first fills in a form, then goes through the OAuth flow.

Without this, on our Cloud, everyone would be directed to OUR egnyte domain (which they obviously don't have an account on) -> they couldn't set it up.

https://github.com/user-attachments/assets/c7cc9965-88fb-4aa2-8b77-a4676bf48b8e

## How Has This Been Tested?
Ran Egnyte connector locally.


web/src/components/credentials/actions/CreateStdOAuthCredential.tsx
```diff
@@ -0,0 +1,89 @@
...
+          ...Object.fromEntries(additionalFields.map((field) => [field, ""])),
+        } as formType
+      }
+      validationSchema={Yup.object().shape({
+        ...Object.fromEntries(
+          additionalFields.map((field) => [field.name, Yup.string().required()])
+        ),
+      })}
+      onSubmit={(values, formikHelpers) => {
+        console.log("HI");
```
rkuo-danswer: nit: extraneous log line

backend/onyx/connectors/egnyte/connector.py
```diff
@@ -124,6 +125,15 @@ def _process_egnyte_file(
 
 
 class EgnyteConnector(LoadConnector, PollConnector, OAuthConnector):
+    class AdditionalOauthKwargs(OAuthConnector.AdditionalOauthKwargs):
+        egnyte_base_domain: str = Field(
+            title="Egnyte Base Domain",
```
rkuo-danswer: i think this would technically be a subdomain? Although not if Egnyte has some standard terminology around this.

rkuo-danswer (approved):

backend/onyx/connectors/egnyte/connector.py
Weves: kind of - they have things like `https://{Egnyte Domain}.egnyte.com/pubapi/` in their docs. So not quite Base Domain, but not subdomain either.

Weves: Can just mirror that

web/src/components/credentials/actions/CreateStdOAuthCredential.tsx
Weves: Good catch—fixed!

PR #3576, opened 2025-01-02
Add support for auto-refreshing available models based on an API call

Weves:
## Description
See title.

## How Has This Been Tested?
Tested with LiteLLM proxy locally (`LLM_MODEL_UPDATE_API_URL=http://localhost:4000/model/info`).


rkuo-danswer (commented):
Would there be any possibility of a race condition if this task were to overlap with itself?  I'm not entirely sure it's possible due to the expiration settings we have for tasks, but worth thinking about.

PR #3585, opened 2025-01-03
Latex for $10 and $100

pablonyx:
## Description
Fixes case where LLM outputs $ naturally and is incorrectly latex-formatted by plugin.
Now, as part of pre-processing text, we escape naturally occuring $s which occur outside of the context of latex equations
Screenshot 2025-01-03 at 11 47 50 AM

Screenshot 2025-01-03 at 11 47 57 AM

## How Has This Been Tested?
- Ask for various strings in response like `Respond with exactly $100 is more than $10" and various Latex equations- ensure all are formatted properly

### DIFF
--- a/web/src/app/chat/message/codeUtils.ts
+++ b/web/src/app/chat/message/codeUtils.ts
@@ -60,18 +60,25 @@ export function extractCodeText(
   return codeText || "";
 }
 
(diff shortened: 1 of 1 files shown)


rkuo-danswer (approved):
Look

PR #3594, opened 2025-01-04
Improved indexing

pablonyx:
## Description
For indexing:
- Document: new value `chunk_count`
- If no `chunk_count`, we can just assume that it has the old UUID system (including non tenancy):
	- If large chunks enabled, generate the UUID based on the old UUID logic + large chunk reference IDs
- If `chunk_count`, can infer UUIDs + exact values for deletion

## How Has This Been Tested?
In multi tenant and single-tenant scenario:
- Index various docs
- Increase size of doc
- Decrease size of doc
- Ensure proper doc chunk IDs are deleted in Vespa
- Do the above specifically when migrating from main to this branch

### DIFF
--- a/backend/onyx/db/document.py
+++ b/backend/onyx/db/document.py
@@ -416,6 +416,18 @@ def update_docs_last_modified__no_commit(
         doc.last_modified = now
 
 
+def update_docs_chunk_count__no_commit(
+    document_ids: list[str],
+    document_id_to_current_chunks_indexed: dict[str, int],
+    db_session: Session,
+) -> None:
+    documents_to_update = (
+        db_session.query(DbDocument).filter(DbDocument.id.in_(document_ids)).all()
+    )
+    for doc in documents_to_update:
+        doc.chunk_count = document_id_to_current_chunks_indexed[doc.id]
+
+
 def mark_document_as_modified(
     document_id: str,
     db_session: Session,
@@ -617,8 +629,20 @@ def get_document(
 def fetch_chunk_counts_for_documents(
     document_ids: list[str],
     db_session: Session,
-) -> dict[str, int]:
+) -> list[tuple[str, int | None]]:
+    """
+    Return a list of (document_id, chunk_count) tuples.
+    Note: chunk_count might be None if not set in DB,
+    so we declare it as Optional[int].
+    """
     stmt = select(DbDocument.id, DbDocument.chunk_count).where(
         DbDocument.id.in_(document_ids)
     )
-    return db_session.execute(stmt).all()
+
+    # results is a list of 'Row' objects, each containing two columns
+    results = db_session.execute(stmt).all()
+
+    # If DbDocument.id is guaranteed to be a string, you can just do row.id;
+    # otherwise cast to str if you need to be sure it's a string:
+    return [(str(row[0]), row[1]) for row in results]
+    # or row.id, row.chunk_count if they are named attributes in your ORM model
--- a/backend/onyx/db/document.py
+++ b/backend/onyx/db/document.py
@@ -418,14 +418,14 @@ def update_docs_last_modified__no_commit(
 
 def update_docs_chunk_count__no_commit(
     document_ids: list[str],
-    document_id_to_current_chunks_indexed: dict[str, int],
+    doc_id_to_chunk_count: dict[str, int],
     db_session: Session,
 ) -> None:
     documents_to_update = (
         db_session.query(DbDocument).filter(DbDocument.id.in_(document_ids)).all()
     )
     for doc in documents_to_update:
-        doc.chunk_count = document_id_to_current_chunks_indexed[doc.id]
+        doc.chunk_count = doc_id_to_chunk_count[doc.id]
 
 
 def mark_document_as_modified(
--- a/backend/onyx/db/document.py
+++ b/backend/onyx/db/document.py
@@ -612,3 +612,13 @@ def get_document(
     stmt = select(DbDocument).where(DbDocument.id == document_id)
     doc: DbDocument | None = db_session.execute(stmt).scalar_one_or_none()
     return doc
+
+
+def fetch_chunk_counts_for_documents(
+    document_ids: list[str],
+    db_session: Session,
+) -> dict[str, int]:
+    stmt = select(DbDocument.id, DbDocument.chunk_count).where(
+        DbDocument.id.in_(document_ids)
+    )
+    return db_session.execute(stmt).all()
--- a/backend/onyx/document_index/vespa/deletion.py
+++ b/backend/onyx/document_index/vespa/deletion.py
@@ -3,9 +3,6 @@ import concurrent.futures
 import httpx
 from retry import retry
 
-from onyx.document_index.vespa.chunk_retrieval import (
-    get_all_vespa_ids_for_document_id,
-)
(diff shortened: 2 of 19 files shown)


backend/onyx/db/document.py
```diff
@@ -416,6 +416,18 @@ def update_docs_last_modified__no_commit(
         doc.last_modified = now
 
 
+def update_docs_chunk_count__no_commit(
+    document_ids: list[str],
+    document_id_to_current_chunks_indexed: dict[str, int],
+    db_session: Session,
+) -> None:
+    documents_to_update = (
```
rkuo-danswer, line 424: are document_id's and document_id_to_current_chunks_indexed the same docs? what's the difference?

backend/onyx/document_index/vespa/indexing_utils.py
```diff
@@ -248,3 +252,31 @@ def clean_chunk_id_copy(
         }
     )
     return clean_chunk
+
+
+def _check_for_chunk_existence(vespa_chunk_id: uuid.UUID, index_name: str) -> bool:
+    vespa_url = f"{DOCUMENT_ID_ENDPOINT.format(index_name=index_name)}/{vespa_chunk_id}"
+    # vesp aurl would be  http://localhost:8081/document/v1/default/danswer_chunk_nomic_ai_nomic_embed_text_v1/docid/
```
rkuo-danswer: typo

backend/onyx/document_index/vespa/indexing_utils.py
```diff
@@ -248,3 +252,31 @@ def clean_chunk_id_copy(
...
+    try:
+        with get_vespa_http_client() as http_client:
```
rkuo-danswer: is this retried anywhere above this call? we generally need a retry strategy for anything hitting vespa

backend/onyx/document_index/vespa/index.py
```diff
@@ -595,15 +637,17 @@ def delete(self, doc_ids: list[str]) -> None:
 
         # NOTE: using `httpx` here since `requests` doesn't support HTTP2. This is beneficial for
         # indexing / updates / deletes since we have to make a large volume of requests.
-        with get_vespa_http_client() as http_client:
-            index_names = [self.index_name]
-            if self.secondary_index_name:
-                index_names.append(self.secondary_index_name)
+        # TODO: incorporate
```
rkuo-danswer: looks like WIP?

backend/onyx/document_index/vespa/index.py
```diff
@@ -324,30 +329,67 @@ def index(
...
+                    chunk_start_index=document_id_to_current_chunks_indexed.get(
+                        doc_id, 0
+                    ),
+                    chunk_end_index=final_chunk_index,
+                    old_version=old_version,
+                )
+                doc_infos.append(doc_chunk_info_with_index)
+
+            # Now, for each doc, we know exactly where to start and end our deletion
+            # So let's genrate the chunk IDs for each chunk to delete
```
rkuo-danswer: typo

backend/onyx/document_index/vespa/index.py
```diff
@@ -324,30 +336,54 @@ def index(
...
-                ]
-                for chunk_batch in batch_generator(first_chunks, BATCH_SIZE):
-                    existing_docs.update(
-                        get_existing_documents_from_chunks(
-                            chunks=chunk_batch,
-                            index_name=self.index_name,
-                            http_client=http_client,
-                            executor=executor,
-                        )
+            enriched_doc_infos: list[EnrichedDocumentIndexingInfo] = []
```
Weves, line 340: can we add a comment here about why we are enriching the docs (e.g. purpose of this step)?

backend/onyx/document_index/vespa/deletion.py
```diff
@@ -23,46 +23,42 @@ def _retryable_http_delete(http_client: httpx.Client, url: str) -> None:
 
 
 @retry(tries=3, delay=1, backoff=2)
```
Weves: we should probably remove this retry since the `_retryable_http_delete` already has retries built in

backend/onyx/indexing/indexing_pipeline.py
```diff
@@ -370,16 +373,35 @@ def index_doc_batch(
...
 
+        doc_id_to_previous_chunks_indexed: dict[str, int | None] = {
+            document_id: chunk_count
+            for document_id, chunk_count in fetch_chunk_counts_for_documents(
+                document_ids=updatable_ids,
+                db_session=db_session,
+            )
+        }
+
+        doc_id_to_current_chunks_indexed: dict[str, int] = {
```
Weves: nit: prefer `doc_id_to_new_chunk_cnt` since (1) `current` could very reasonably mean "what is in Vespa right now" (to me it's a >50% chance someone interprets it that way) and (2) this is really a count per document not the chunks themselves

backend/onyx/indexing/indexing_pipeline.py
```diff
@@ -370,16 +373,35 @@ def index_doc_batch(
...
         )
-        document_id_to_document_set = {
+        doc_id_to_document_set = {
             document_id: document_sets
             for document_id, document_sets in fetch_document_sets_for_documents(
                 document_ids=updatable_ids, db_session=db_session
             )
         }
 
+        doc_id_to_previous_chunks_indexed: dict[str, int | None] = {
```
Weves: nit: prefer `doc_id_to_previous_chunk_cnt` since this is really a count per document not the chunks themselves

backend/onyx/document_index/vespa/index.py
```diff
@@ -324,30 +336,54 @@ def index(
...
+            for document_id, _ in doc_id_to_previous_chunks_indexed.items():
+                last_indexed_chunk = doc_id_to_previous_chunks_indexed.get(
+                    document_id, None
+                )
+                is_old_version = False
+                if last_indexed_chunk is None:
```
Weves, line 346: similar, can we add a comment about what this step is for (I think it's primarily for migration purposes? or if something is wrong e.g. db is in a bad state and missing the chunk cnt?)

backend/onyx/document_index/vespa/deletion.py
```diff
@@ -23,46 +23,42 @@ def _retryable_http_delete(http_client: httpx.Client, url: str) -> None:
...
 
-def delete_vespa_docs(
-    document_ids: list[str],
+def delete_vespa_chunks(
+    doc_chunk_ids: list[UUID],
     index_name: str,
     http_client: httpx.Client,
     executor: concurrent.futures.ThreadPoolExecutor | None = None,
 ) -> None:
+    if not _does_doc_chunk_exist(doc_chunk_ids[0], index_name, http_client):
```
Weves: nit: can we add a comment as to why we check first rather than just always deleting?

Weves (commented):
Can't wait to get this one in 🙏

backend/onyx/document_index/vespa/deletion.py
pablonyx: This was more for my sake during testing and we shouldn't need it unless something goes wrong with the chunk UUID scheme (ie. we don't reach the logic where we update the database with chunk count, but do modify the number of chunks)

Weves (approved):
lgtm 🦺

PR #3601, opened 2025-01-05
Properly account for anonymous access in Confluence

hagen-danswer:
## Description
- Anonymous access is now properly handled for confluence server
- permissions now properly inherit across ancestors 

## How Has This Been Tested?
- tested on confluence server by changing permissions of a child/parent page for a non admin user
- tested by removing all group access from a space and still ensuring access is given to those where the space has anonymous access is enabled

### DIFF
--- a/backend/ee/onyx/external_permissions/confluence/doc_sync.py
+++ b/backend/ee/onyx/external_permissions/confluence/doc_sync.py
@@ -4,6 +4,8 @@ https://confluence.atlassian.com/conf85/check-who-can-view-a-page-1283360557.htm
 """
 from typing import Any
 
+from ee.onyx.configs.app_configs import CONFLUENCE_ANONYMOUS_ACCESS_IS_PUBLIC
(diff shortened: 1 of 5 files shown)


backend/ee/onyx/external_permissions/confluence/doc_sync.py
```diff
@@ -177,21 +193,57 @@ def _extract_read_access_restrictions(
         group["name"] for group in read_access_group_jsons if group.get("name")
     ]
 
+    return set(read_access_user_emails), set(read_access_group_names)
+
+
+def _get_all_page_restrictions(
```
Weves, line 199: nit: these functions are confusingly named. It's not clear what the difference between `_get_all_page_restrictions` and `_fetch_all_page_restrictions` is. 

One should probably named `_get_restrictons_for_page_batch` and the other `_get_restrictions_for_page` or something similar.

Let's address this in a follow up.

Weves (approved):

PR #3602, opened 2025-01-05
various lock diagnostics and timing adjustments

rkuo-danswer:
## Description

## How Has This Been Tested?

### DIFF
--- a/backend/onyx/configs/constants.py
+++ b/backend/onyx/configs/constants.py
@@ -76,7 +76,7 @@ KV_ENTERPRISE_SETTINGS_KEY = "onyx_enterprise_settings"
 KV_CUSTOM_ANALYTICS_SCRIPT_KEY = "__custom_analytics_script__"
 KV_DOCUMENTS_SEEDED_KEY = "documents_seeded"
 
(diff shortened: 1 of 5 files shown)


backend/onyx/configs/constants.py
```diff
@@ -76,7 +76,7 @@
 KV_CUSTOM_ANALYTICS_SCRIPT_KEY = "__custom_analytics_script__"
 KV_DOCUMENTS_SEEDED_KEY = "documents_seeded"
 
-CELERY_VESPA_SYNC_BEAT_LOCK_TIMEOUT = 60
+CELERY_VESPA_SYNC_BEAT_LOCK_TIMEOUT = 120
```
pablonyx, line 81: Since this is a pattern we follow pretty consistently for reacquiring locks, (i.e. reacquiring lock if past timeout / 4 since last reacquisition), may be worth noting here at the definition- for future timeouts we create

pablonyx:
Looks good to put in for diagnostics! I think we may need to think about the integration of these locks and how that relates to our timeouts, so we can set a standard we adhere to

pablonyx (approved):

PR #3618, opened 2025-01-07
Async Redis

pablonyx:
## Description
Ensure SSL context is propagated for async Redis

## How Has This Been Tested?

### DIFF
--- a/backend/onyx/server/middleware/rate_limiting.py
+++ b/backend/onyx/server/middleware/rate_limiting.py
@@ -45,8 +45,8 @@ def get_auth_rate_limiters() -> List[Callable]:
     return [
         Depends(
             RateLimiter(
-                times=RATE_LIMIT_MAX_REQUESTS,
-                seconds=RATE_LIMIT_WINDOW_SECONDS,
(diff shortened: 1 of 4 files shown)


backend/onyx/server/middleware/rate_limiting.py
```diff
@@ -6,18 +6,25 @@
...
 from fastapi_limiter.depends import RateLimiter
 
+from onyx.configs.app_configs import AUTH_RATE_LIMITING_ENABLED
 from onyx.configs.app_configs import RATE_LIMIT_MAX_REQUESTS
 from onyx.configs.app_configs import RATE_LIMIT_WINDOW_SECONDS
 from onyx.redis.redis_pool import get_async_redis_connection
 
 
-async def setup_limiter() -> None:
+async def setup_auth_limiter() -> None:
```
rkuo-danswer, line 15: generally speaking you want to use globals at the top level and pass down or switch on them as needed deeper in the logic.  would it make sense to simply switch on AUTH_RATE_LIMITING_ENABLED in lifespan and just not call this function if it isn't needed?

rkuo-danswer (approved):

PR #3661, opened 2025-01-12
Agent Search

yuhongsun96:
## Description
Not final, do not merge. Just giving comments.

## How Has This Been Tested?
NA


.vscode/env_template.txt
```diff
@@ -51,3 +51,9 @@ BING_API_KEY=<REPLACE THIS>
 # Enable the full set of Danswer Enterprise Edition features
 # NOTE: DO NOT ENABLE THIS UNLESS YOU HAVE A PAID ENTERPRISE LICENSE (or if you are using this for local testing/development)
 ENABLE_PAID_ENTERPRISE_EDITION_FEATURES=False
+
+# Agent Search configs  # TODO: Remove give proper namings
+AGENT_RETRIEVAL_STATS=False   # Note: This setting will incur substantial re-ranking effort
```
yuhongsun96, line 56: If there's a shared development branch, try to keep it as clean as possible. Pretend it's main but that everything will be squashed. Ok to have TODOs and such in there though.

Here - referring to the fact that AGENT_RETRIEVAL_STATS shows up twice like this. Concern if that we'll forget to clean things up as feature branches get more complex

backend/alembic/versions/1adf5ea20d2b_agent_doc_result_col.py
```diff
@@ -0,0 +1,29 @@
+"""agent_doc_result_col
```
yuhongsun96, line 1: Since these aren't making into any deployments directly, I prefer at the end to create a single migration. For the feature branch dev though, this is good so that other team members can just run upgrade.

backend/ee/onyx/db/analytics.py
```diff
@@ -345,7 +345,8 @@ def fetch_assistant_unique_users_total(
 def user_can_view_assistant_stats(
     db_session: Session, user: User | None, assistant_id: int
 ) -> bool:
-    # If user is None, assume the user is an admin or auth is disabled
+    # If user is None and auth is disabled, assume the user is an admin
```
yuhongsun96, line 348: Is there supposed to be a change here in the code?

backend/ee/onyx/db/token_limit.py
```diff
@@ -47,6 +48,12 @@ def _add_user_filters(
     that the user isn't a curator for
     - if we are not editing, we show all token_rate_limits in the groups the user curates
     """
+
+    # If user is None, this is an anonymous user and we should only show public token_rate_limits
```
yuhongsun96, line 52: Are these rebase issues? This seems out of scope no?

backend/ee/onyx/server/query_and_chat/chat_backend.py
```diff
@@ -179,6 +179,7 @@ def handle_simplified_chat_message(
         chunks_below=0,
         full_doc=chat_message_req.full_doc,
         structured_response_format=chat_message_req.structured_response_format,
+        use_pro_search=chat_message_req.use_pro_search,
```
yuhongsun96: try to use the right terminology in the code. Seems trivial but it becomes annoying for later team members. Like I've gotten the question "Is a Persona the same as an Assistant" like 5 times already and we're only 8 people

Since we're calling it agent search all the var should be agent too

backend/ee/onyx/server/query_and_chat/models.py
```diff
@@ -57,6 +57,9 @@ class BasicCreateChatMessageRequest(ChunkContext):
     # https://platform.openai.com/docs/guides/structured-outputs/introduction
     structured_response_format: dict | None = None
 
+    # If True, uses pro search instead of basic search
```
yuhongsun96: Let's use agent vs basic instead of pro. Though agent vs non_agent in the code base is probably fine.

backend/onyx/chat/answer.py
```diff
@@ -212,29 +191,63 @@ def _get_response(self, llm_calls: list[LLMCall]) -> AnswerStream:
...
+        # In the future, we'll want to handle citations and tool calls in the langgraph graph.
         response_handler_manager = LLMResponseHandlerManager(
             tool_call_handler, answer_handler, self.is_cancelled
         )
 
+        # In langgraph, whether we do the basic thing (call llm stream) or pro search
+        # is based on a flag in the pro search config
+
+        if self.pro_search_config:
+            if self.pro_search_config.search_request is None:
```
yuhongsun96, line 217: Prefer validation like this to happen at the model layer using Pydantic functionality. Then you can just assert here or just type cast it

backend/onyx/chat/answer.py
```diff
@@ -212,29 +191,63 @@ def _get_response(self, llm_calls: list[LLMCall]) -> AnswerStream:
...
+            if len(search_tools) == 0:
+                raise ValueError("No search tool found")
+            elif len(search_tools) > 1:
+                # TODO: handle multiple search tools
+                raise ValueError("Multiple search tools found")
+
+            search_tool = search_tools[0]
+            if self.db_session is None:
+                raise ValueError("db_session must be provided for pro search")
+            if self.fast_llm is None:
```
yuhongsun96, line 222: All of these checks can be broken out into a separate function for cleaner encapsulation.

backend/onyx/chat/answer.py
```diff
@@ -212,29 +191,63 @@ def _get_response(self, llm_calls: list[LLMCall]) -> AnswerStream:
...
+                raise ValueError("fast_llm must be provided for pro search")
+
+            stream = run_main_graph(
```
yuhongsun96, line 225: Nit: pref standardizing naming for reviewer. agent graph vs basic graph? idk

backend/onyx/chat/answer.py
```diff
@@ -212,29 +191,63 @@ def _get_response(self, llm_calls: list[LLMCall]) -> AnswerStream:
...
+                config=self.pro_search_config,
+                primary_llm=self.llm,
+                fast_llm=self.fast_llm,
+                search_tool=search_tool,
+                db_session=self.db_session,
+            )
+        else:
+            stream = run_basic_graph(
+                last_llm_call=current_llm_call,
+                primary_llm=self.llm,
```
yuhongsun96: Should this flow still have basic LLM filtering as an option?

backend/onyx/chat/answer.py
```diff
@@ -294,8 +307,9 @@ def processed_streamed_output(self) -> AnswerStream:
     def llm_answer(self) -> str:
         answer = ""
         for packet in self.processed_streamed_output:
-            if isinstance(packet, OnyxAnswerPiece) and packet.answer_piece:
+            if (isinstance(packet, OnyxAnswerPiece) and packet.answer_piece) or (isinstance(packet, AgentAnswerPiece) and packet.answer_piece and packet.answer_type == "agent_level_answer"):
```
yuhongsun96: Be sure to install Black formatter

backend/onyx/chat/answer.py
```diff
@@ -212,29 +191,63 @@ def _get_response(self, llm_calls: list[LLMCall]) -> AnswerStream:
...
+
+            stream = run_main_graph(
+                config=self.pro_search_config,
+                primary_llm=self.llm,
+                fast_llm=self.fast_llm,
+                search_tool=search_tool,
+                db_session=self.db_session,
+            )
+        else:
+            stream = run_basic_graph(
```
yuhongsun96, line 229: The basic and agent graph should share models as much as possible. Probably with basic having some input state pydantic model and agent having a superset (subclass with more fields)

backend/onyx/agent_search/basic/graph_builder.py
```diff
@@ -0,0 +1,95 @@
...
+)
+
+def basic_graph_builder() -> StateGraph:
+    graph = StateGraph(
+        state_schema=BasicState,
+        input=BasicInput,
+        output=BasicOutput,
+    )
+
+    ### Add nodes ###
```
yuhongsun96: Can we remove these comments and add comments describing each node and edge? Alternatively the node functions like get_response should be clearly commented so readers can have clarity on what is happening at this level of abstraction

backend/onyx/agent_search/basic/graph_builder.py
```diff
@@ -0,0 +1,95 @@
...
+    graph.add_node(
+        node="get_response",
+        action=get_response,
+    )
+
+    ### Add edges ###
+
+    graph.add_edge(start_key=START, end_key="get_response")
+
+    graph.add_conditional_edges("get_response", should_continue, ["get_response", END])
```
yuhongsun96: How does this conditional work? I would give this a fairly thorough description that mentions the tool calling and whatever else happens.

backend/onyx/agent_search/basic/graph_builder.py
```diff
@@ -0,0 +1,95 @@
...
+    graph.add_edge(
+        start_key="get_response",
+        end_key=END,
+    )
+
+    return graph
+
+
+def should_continue(state: BasicState) -> str:
```
yuhongsun96: This is hard for someone new to understand, it's hard to trace what is happening here since there is not a clear call stack

yuhongsun96: like in what case does last_llm_call become a None? This type of logic based on state mutation that's just keys of a dictionary is hard for a reader to trace so comments are very valuable.

PR #3666, opened 2025-01-13
fixed group sync to account for changes in drive permissions

hagen-danswer:
## Description
- this now attaches the drive id as a group to the doc
- the external group job now also resolves the drive->users relationship

## How Has This Been Tested?
- tested against the end to end permission sync test in pr 3635

### DIFF
--- a/backend/ee/onyx/external_permissions/google_drive/doc_sync.py
+++ b/backend/ee/onyx/external_permissions/google_drive/doc_sync.py
@@ -101,14 +101,14 @@ def _get_permissions_from_slim_doc(
 
     company_domain = google_drive_connector.google_domain
     user_emails: set[str] = set()
-    group_names: set[str] = set()
+    group_emails: set[str] = set()
     public = False
     for permission in permissions_list:
         permission_type = permission["type"]
         if permission_type == "user":
             user_emails.add(permission["emailAddress"])
         elif permission_type == "group":
-            group_names.add(permission["emailAddress"])
+            group_emails.add(permission["emailAddress"])
         elif permission_type == "domain" and company_domain:
             if permission.get("domain") == company_domain:
                 public = True
@@ -120,12 +120,12 @@ def _get_permissions_from_slim_doc(
         elif permission_type == "anyone":
             public = True
 
-    if parent_drive_id := permission_info.get("drive_id"):
-        group_names.add(parent_drive_id)
+    drive_id = permission_info.get("drive_id")
+    group_ids = group_emails | ({drive_id} if drive_id is not None else set())
 
     return ExternalAccess(
         external_user_emails=user_emails,
-        external_user_group_ids=group_names,
+        external_user_group_ids=group_ids,
         is_public=public,
     )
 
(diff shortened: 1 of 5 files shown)


backend/ee/onyx/external_permissions/google_drive/group_sync.py
```diff
@@ -1,52 +1,135 @@
...
-    google_drive_connector = GoogleDriveConnector(
-        **cc_pair.connector.connector_specific_config
-    )
-    google_drive_connector.load_credentials(cc_pair.credential.credential_json)
-    admin_service = get_admin_service(
-        google_drive_connector.creds, google_drive_connector.primary_admin_email
+def _get_drive_members(
+    google_drive_connector: GoogleDriveConnector,
+) -> dict[str, tuple[set[str], set[str]]]:
+    drive_ids = google_drive_connector.get_all_drive_ids()
```
Weves, line 23: can we add a comment here as to what this function does 😄 Especially for permission syncing, which is pretty complicated/requires an understanding of external services would prefer to have this a bit better commented.

backend/ee/onyx/external_permissions/google_drive/group_sync.py
```diff
@@ -1,52 +1,135 @@
...
+
+    drive_ids_to_members_map: dict[str, tuple[set[str], set[str]]] = {}
```
Weves: can we add a comment which says what each field in the dict is supposed to be?

Weves: nit: should probably be `drive_id_to_members_map` since it doens't map multiple drive IDs to members, it maps a single drive ID to it's members 🤓

backend/ee/onyx/external_permissions/google_drive/group_sync.py
```diff
@@ -1,52 +1,135 @@
...
-        if not group_member_emails:
-            continue
+def _build_onyx_groups(
+    drive_ids_to_members_map: dict[str, tuple[set[str], set[str]]],
+    group_to_members_map: dict[str, set[str]],
+) -> list[ExternalUserGroup]:
+    onyx_groups: list[ExternalUserGroup] = []
+
+    # Convert all drive member definitions to onyx groups
+    for drive_id, (group_emails, user_emails) in drive_ids_to_members_map.items():
```
Weves: can we add a comment as to why this assumption is safe (specifically, the fact that drive-level permissions are not revocable at the individual file level).

Weves (changes requested):

backend/ee/onyx/external_permissions/google_drive/group_sync.py
hagen-danswer: 🤓 ☝️

backend/ee/onyx/external_permissions/google_drive/doc_sync.py
```diff
@@ -101,14 +101,14 @@ def _get_permissions_from_slim_doc(
 
     company_domain = google_drive_connector.google_domain
     user_emails: set[str] = set()
-    group_emails: set[str] = set()
+    group_names: set[str] = set()
```
Weves: let's have separate this out into group_emails and drive_ids

Weves (approved):

PR #3671, opened 2025-01-14
Fix document counts

Weves:
## Description
Addresses two edge cases:

1. Permission syncing creating DocumentByConnectorCredentialPair objects before the indexing has occurred -> much larger count at the CC Pair level compared to the Index Attempt.
2. Docs that are filtered out were still counted as part of the Index Attempt cnt but not part of the CC Pair count.

## How Has This Been Tested?
Ran indexing locally, verified counts matched as expected.

### DIFF
--- a/backend/alembic/versions/c7bf5721733e_add_has_been_indexed_to_.py
+++ b/backend/alembic/versions/c7bf5721733e_add_has_been_indexed_to_.py
@@ -10,7 +10,7 @@ import sqlalchemy as sa
 
 # revision identifiers, used by Alembic.
 revision = "c7bf5721733e"
-down_revision = "0f7ff6d75b57"
+down_revision = "c5eae4a75a1b"
 branch_labels = None
 depends_on = None
 
--- a/backend/alembic/versions/c7bf5721733e_add_has_been_indexed_to_.py
+++ b/backend/alembic/versions/c7bf5721733e_add_has_been_indexed_to_.py
@@ -10,7 +10,7 @@ import sqlalchemy as sa
 
 # revision identifiers, used by Alembic.
 revision = "c7bf5721733e"
-down_revision = "c5eae4a75a1b"
+down_revision = "027381bce97c"
 branch_labels = None
 depends_on = None
 
--- a/backend/alembic/versions/c7bf5721733e_add_has_been_indexed_to_.py
+++ b/backend/alembic/versions/c7bf5721733e_add_has_been_indexed_to_.py
@@ -10,7 +10,7 @@ import sqlalchemy as sa
 
 # revision identifiers, used by Alembic.
 revision = "c7bf5721733e"
-down_revision = "fec3db967bf7"
+down_revision = "0f7ff6d75b57"
 branch_labels = None
 depends_on = None
 
--- a/backend/alembic/versions/c7bf5721733e_add_has_been_indexed_to_.py
+++ b/backend/alembic/versions/c7bf5721733e_add_has_been_indexed_to_.py
@@ -1,0 +1,48 @@
+"""Add has_been_indexed to DocumentByConnectorCredentialPair
+
+Revision ID: c7bf5721733e
+Revises: fec3db967bf7
+Create Date: 2025-01-13 12:39:05.831693
+
+"""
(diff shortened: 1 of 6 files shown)


backend/onyx/indexing/indexing_pipeline.py
```diff
@@ -263,21 +283,6 @@ def index_doc_batch_prepare(
 def filter_documents(document_batch: list[Document]) -> list[Document]:
     documents: list[Document] = []
     for document in document_batch:
-        # Remove any NUL characters from title/semantic_id
```
Weves, line 266: Removing this logic from this func and pushing it to `strip_null_characters` in `run_indexing.py`.

This isn't really a "filter" step, and the other function has this exact logic more-or-less, so it feels like it's a better fit there

backend/onyx/background/indexing/run_indexing.py
```diff
@@ -115,6 +122,12 @@ def strip_null_characters(doc_batch: list[Document]) -> list[Document]:
                 )
                 section.link = section.link.replace("\x00", "")
 
+            if section.text and "\x00" in section.text:
```
rkuo-danswer: how large can these sections get?  we're doing double work to scan and then replace.  this is technically wrong on the other fields too but those seem inherently smaller / less of a concern.

backend/onyx/background/indexing/run_indexing.py
```diff
@@ -289,27 +300,23 @@ def _run_indexing(
...
-                all_connector_doc_ids.update(doc.id for doc in doc_batch_cleaned)
+                net_doc_change += index_pipeline_result.new_docs
+                chunk_count += index_pipeline_result.total_chunks
+                document_count += index_pipeline_result.total_docs
 
                 # commit transaction so that the `update` below begins
                 # with a brand new transaction. Postgres uses the start
                 # of the transactions when computing `NOW()`, so if we have
-                # a long running transaction, the `time_updated` field will
+                # a long running transaction, the `time_updated`  field will
```
rkuo-danswer: nit: extra space

backend/onyx/indexing/indexing_pipeline.py
```diff
@@ -55,12 +56,20 @@ class DocumentBatchPrepareContext(BaseModel):
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
 
+class IndexingPipelineResult(BaseModel):
+    new_docs: int
```
rkuo-danswer, line 62: maybe commenting all the fields would be helpful here ... seems a little tricky

backend/alembic/versions/c7bf5721733e_add_has_been_indexed_to_.py
```diff
@@ -0,0 +1,48 @@
...
+    op.add_column(
+        "document_by_connector_credential_pair",
+        sa.Column("has_been_indexed", sa.Boolean(), nullable=True),
+    )
+    op.execute(
+        "UPDATE document_by_connector_credential_pair SET has_been_indexed = TRUE"
+    )
+    op.alter_column(
+        "document_by_connector_credential_pair",
+        "has_been_indexed",
```
rkuo-danswer, line 29: out of curiosity why does has been indexed belong on the cc pair? It feels like it belongs on the document table?  I've thought of the document by cc pair more as a ref count than the actual state of the document.  I do understand we're counting the docs in this table so perhaps the join is expensive or tricky.

Weves: the reason is that for "counting docs indexed by a connector" purposes we don't really care if the document has been indexed at all, we care if it has been indexed by the specific connector.

backend/onyx/indexing/indexing_pipeline.py
Weves: Added!

backend/onyx/background/indexing/run_indexing.py
Weves: they are max 2k tokens, which is ~6k chars. Removed duplicate scan just in case.

rkuo-danswer (approved):

PR #3684, opened 2025-01-15
Add is_cloud info to telemetry + get consistent customer_uuid's for a…

Weves:
… given tenant

## Description
Needed to separate out metrics from self-hosted vs cloud.

## How Has This Been Tested?
👀 

Verified telemetry could be sent from local.

### DIFF
--- a/backend/onyx/utils/telemetry.py
+++ b/backend/onyx/utils/telemetry.py
@@ -21,6 +21,7 @@ from onyx.utils.variable_functionality import (
 )
 from onyx.utils.variable_functionality import noop_fallback
 from shared_configs.configs import MULTI_TENANT
+from shared_configs.contextvars import CURRENT_TENANT_ID_CONTEXTVAR
 
 _DANSWER_TELEMETRY_ENDPOINT = "https://telemetry.onyx.app/anonymous_telemetry"
 _CACHED_UUID: str | None = None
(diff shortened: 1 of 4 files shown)


backend/onyx/utils/telemetry.py
```diff
@@ -36,7 +37,11 @@ class RecordType(str, Enum):
     METRIC = "metric"
 
 
-def get_or_generate_uuid() -> str:
+def _get_or_generate_customer_id_mt(tenant_id: str) -> str:
+    return str(uuid.uuid5(uuid.NAMESPACE_X500, tenant_id))
+
+
+def get_or_generate_uuid(tenant_id: str | None = None) -> str:
```
rkuo-danswer, line 44: is it possible we would want this id to be reused in other places? Something like this might belong one layer down ... aka it feels like the concept of a unique id is installation wide and this code should just be calling "get customer uuid", while an API below it handles generating and persisting the key if it doesn't exist.

backend/onyx/utils/telemetry.py
```diff
@@ -36,7 +37,11 @@ class RecordType(str, Enum):
...
     global _CACHED_UUID
 
     if _CACHED_UUID is not None:
```
rkuo-danswer, line 52: It seems like there might be a race condition here where multiple UUID's could be generated and stored by different services on startup ... then it would be ambiguous as to who wins.  probably unlikely?

rkuo-danswer (approved):
just needs some comments for tiny edge cases

Issue #3680, opened 2025-01-15
Curators can't edit Public Assistants


shutupderek:
## Scenario
We use the Curators group extensively for different departments to have and manage their own Assistants.

## Problem
1. Curators are unable to edit their Assistants' prompts once they're marked as Public.
2. Not only are Curators unable to edit their public Assistants' prompts, but any attempt to edit them results in the prompts being deleted.

## Steps to Reproduce
1. Assign an Assistant to a Curator group and mark it as Public.
2. Attempt to edit and update the prompt of the public Assistant as a Curator.
3. Observe that the prompt cannot be edited and any attempt to do so results in the prompt being deleted.
```json
Failed to create Assistant = {"detail":"Persona with ID ## does not exist or user is not authorized to access it"}
```
## Expected Behavior
Curators should be able to edit their public Assistants' prompts without any issues including data deletion.

## Actual Behavior
Curators are currently unable to edit their public Assistants' prompts. Any attempt to edit results in the prompts being deleted.

## Environment
- Onyx Version: 0.17.1

hagen-danswer:
https://github.com/onyx-dot-app/onyx/pull/3690

prompt deletion fixed in this pr^

PR #3690, opened 2025-01-16
Combined Persona and Prompt API

hagen-danswer:
## Description
- got rid of all prompt apis
- created a new file to perform all prompt db operations
- prompt manipulation happens exclusively through persona api calls
    - this has the benefit of not creating orphan prompts
    - it was 1:1 anyways so this is simpler to keep more logic on the backend
- bugfix for attaching personas to groups
- bugfix this https://github.com/onyx-dot-app/onyx/issues/3680
- deleted a lot of unused code for clarity

## How Has This Been Tested?
- tested with an admin
    - ability to manipulate all assistants
- tested with curator
    - ability to manipulate assistants they have created
    - ability to manipulate assistants that belong exclusively to groups they curate
- tested with basic user
    - ability to manipulate assistants they have created


backend/onyx/db/prompts.py
```diff
@@ -0,0 +1,132 @@
...
+
+@lru_cache()
+def get_default_prompt__read_only() -> Prompt:
+    """Due to the way lru_cache / SQLAlchemy works, this can cause issues
+    when trying to attach the returned `Prompt` object to a `Persona`. If you are
+    doing anything other than reading, you should use the `get_default_prompt`
+    method instead."""
+    with Session(get_sqlalchemy_engine()) as db_session:
+        return _get_default_prompt(db_session)
+
```
pablonyx: We need to make sure we refer to the proper schema here

backend/onyx/server/features/persona/models.py
```diff
@@ -27,32 +27,36 @@ class GenerateStarterMessageRequest(BaseModel):
...
+    include_citations: bool
     is_public: bool
-    llm_filter_extraction: bool
     recency_bias: RecencyBiasSetting
     prompt_ids: list[int]
-    document_set_ids: list[int]
-    # e.g. ID of SearchTool or ImageGenerationTool or <USER_DEFINED_TOOL>
-    tool_ids: list[int]
+    llm_filter_extraction: bool
+    llm_relevance_filter: bool
```
pablonyx: Would prefer a slighter cleaner relationship between `existing_prompt_id` and `prompts` if possible, otherwise a comment explaining the logic here 
Specifically, seems like `existing_prompt_id` and `prompts` are generally equivalent to the upsert request, but `existing_prompt_id` is always created in the API request and `prompts` is only used for the seeding flow

pablonyx (changes requested):

backend/onyx/server/features/persona/models.py
hagen-danswer: agreed, was not happy with this haha

backend/onyx/db/prompts.py
hagen-danswer: Not sure what this means

pablonyx: Discussed in person and has been addressed

pablonyx:
Looks good now! Tested creating, updating prompts on multi tenant and single tenant locally

pablonyx (approved):

PR #3696, opened 2025-01-17
Slackbot optimization

pablonyx:
## Description
Fixes https://linear.app/danswer/issue/DAN-1275/slack-bot-not-working-with-on-our-cloud

Fixes cloud slackbot "acquiring" many tenants

- Centralize removal logic
- Use self.tenant_ids to track genuinely active tenants - only consider tenants with slackbots "active"

## How Has This Been Tested?
- Single tenant + multiple slackbots
- Multi tenant + multiple Slackbots + delete slackbots

### DIFF
--- a/backend/onyx/onyxbot/slack/listener.py
+++ b/backend/onyx/onyxbot/slack/listener.py
@@ -352,7 +352,7 @@ class SlackbotHandler:
                         )
                         self._remove_tenant(tenant_id)
 
-                        # IMPORTANT: Release the lock here (in the same scope it was acquired)
+                        # NOTE: We release the lock here (in the same scope it was acquired)
                         if tenant_id in self.redis_locks and not DEV_MODE:
                             try:
                                 self.redis_locks[tenant_id].release()
--- a/backend/onyx/onyxbot/slack/listener.py
+++ b/backend/onyx/onyxbot/slack/listener.py
@@ -14,6 +14,7 @@ from typing import Set
 
 from prometheus_client import Gauge
 from prometheus_client import start_http_server
+from redis.lock import Lock
 from slack_sdk import WebClient
 from slack_sdk.socket_mode.request import SocketModeRequest
 from slack_sdk.socket_mode.response import SocketModeResponse
@@ -121,6 +122,9 @@ class SlackbotHandler:
         # The keys for these dictionaries are tuples of (tenant_id, slack_bot_id)
         self.socket_clients: Dict[tuple[str | None, int], TenantSocketModeClient] = {}
         self.slack_bot_tokens: Dict[tuple[str | None, int], SlackBotTokens] = {}
+
+        # Store Redis lock objects here so we can release them properly
+        self.redis_locks: Dict[str | None, Lock] = {}
 
         self.running = True
         self.pod_id = self.get_pod_id()
@@ -237,11 +241,10 @@ class SlackbotHandler:
 
     def acquire_tenants(self) -> None:
         """
-        - Attempt to acquire a lock for each tenant.
+        - Attempt to acquire a Redis lock for each tenant.
         - If acquired, check if that tenant actually has Slack bots.
         - If yes, store them in self.tenant_ids and manage the socket connections.
-        - If a tenant in self.tenant_ids no longer has Slack bots, remove it and release lock.
-        We do this to minimize the number of pods spun up to manage tenants.
+        - If a tenant in self.tenant_ids no longer has Slack bots, remove it and release the lock.
         """
         all_tenants = get_all_tenant_ids()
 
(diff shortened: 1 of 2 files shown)


backend/onyx/onyxbot/slack/listener.py
```diff
@@ -252,42 +268,111 @@ def acquire_tenants(self) -> None:
                 ex=TENANT_LOCK_EXPIRATION,
```
rkuo-danswer: Maybe we should add comments to this to help explain how the lock is being used.

For the above lock, why manually set the lock vs using the lock class? it's a lot more error prone to have to handle the lock manually.  Are we persisting the lock across runs?  When does it get deleted?

backend/onyx/onyxbot/slack/listener.py
```diff
@@ -252,42 +268,111 @@ def acquire_tenants(self) -> None:
...
+                        for bot in bots:
+                            self._manage_clients_per_tenant(
+                                db_session=db_session,
+                                tenant_id=tenant_id,
+                                bot=bot,
+                            )
             finally:
                 CURRENT_TENANT_ID_CONTEXTVAR.reset(token)
 
+    def _remove_tenant(self, tenant_id: str | None) -> None:
```
rkuo-danswer, line 376: Would suggest making the function more "pure" ... have the caller decide whether to remove the tenant and just have this function do what it is asked if called.

backend/onyx/onyxbot/slack/listener.py
```diff
@@ -252,42 +268,111 @@ def acquire_tenants(self) -> None:
...
+        """
+        Helper to remove a tenant from `self.tenant_ids`, close any socket clients,
+        and release the lock (if not in DEV_MODE).
+        """
+        if tenant_id not in self.tenant_ids:
+            return
+
+        # Close all socket clients for this tenant
+        for (tenant_id, slack_bot_id), client in list(self.socket_clients.items()):
+            if tenant_id == tenant_id:
```
rkuo-danswer: this seems like a bug

backend/onyx/onyxbot/slack/listener.py
```diff
@@ -252,42 +268,111 @@ def acquire_tenants(self) -> None:
...
+                del self.slack_bot_tokens[(tenant_id, slack_bot_id)]
+                logger.info(
+                    f"Stopped SocketModeClient for tenant: {tenant_id}, app: {slack_bot_id}"
+                )
+
+        # Remove from active set
+        self.tenant_ids.remove(tenant_id)
+
+        # Release the lock if not in dev mode
+        if not DEV_MODE:
```
rkuo-danswer: think we should just pass the redis_client into this function?

backend/onyx/onyxbot/slack/listener.py
```diff
@@ -217,29 +224,38 @@ def _manage_clients_per_tenant(
...
             if tenant_bot_pair in self.socket_clients:
                 asyncio.run(self.socket_clients[tenant_bot_pair].close())
 
             self.start_socket_client(bot.id, tenant_id, slack_bot_tokens)
 
     def acquire_tenants(self) -> None:
-        tenant_ids = get_all_tenant_ids()
-
-        for tenant_id in tenant_ids:
+        """
```
rkuo-danswer, line 243: would be good to comment on the practical scope of checking for active tenants ... aka better resource utilization per pod.

backend/onyx/onyxbot/slack/listener.py
```diff
@@ -252,42 +268,111 @@ def acquire_tenants(self) -> None:
...
+
+                    if not bots:
+                        logger.info(
+                            f"Tenant {tenant_id} no longer has Slack bots. Removing."
+                        )
+                        self._remove_tenant(tenant_id)
+                    else:
+                        # Manage or reconnect Slack bot sockets
+                        for bot in bots:
+                            self._manage_clients_per_tenant(
```
rkuo-danswer, line 368: would be helpful to docstring manage_clients_per_tenant as to what it does.

backend/onyx/onyxbot/slack/listener.py
```diff
@@ -252,42 +268,111 @@ def acquire_tenants(self) -> None:
...
+                    f"Stopped SocketModeClient for tenant: {tenant_id}, app: {slack_bot_id}"
+                )
+
+        # Remove from active set
+        self.tenant_ids.remove(tenant_id)
+
+        # Release the lock if not in dev mode
+        if not DEV_MODE:
+            redis_client = get_redis_client(tenant_id=tenant_id)
+            redis_client.delete(OnyxRedisLocks.SLACK_BOT_LOCK)
```
rkuo-danswer: actually i think if we are handling the lock, we should do it at the same level/scope as where it is acquired.  it's tricky / bug prone to remove locks in a nested function.

rkuo-danswer (commented):
General comment in glancing through the code ... would think about reducing the class member side effects of some of these methods ... if the functions can be made pure by adding just one or a few parameters and/or returning state, we should do that where possible to make it easier to reason through.

backend/onyx/onyxbot/slack/listener.py
pablonyx: Agree now that you bring it up- will bring locking operations up to the main

backend/onyx/onyxbot/slack/listener.py
pablonyx: Good point!

backend/onyx/onyxbot/slack/listener.py
pablonyx: Very fair- inherited from previous code approach

backend/onyx/onyxbot/slack/listener.py
pablonyx: Great point- inherited but 100% agree re lock stability / use

rkuo-danswer (approved):

PR #3721, opened 2025-01-20
Dropdown for slackbot channel names

ktjayamanna:
## Description
- Added a call back to SearchMultiSelectDropdown so parent can access the search term.
- Updated SearchMultiSelectDropdown's search term to reflect selected option instead of setting it to empty string.
- Created proxy to make a request to the backend from the frontend to get the channels.
- Created a function at the data layer to get the Slack credentials.
- Created a backend function code that utilizes the Slack credentials to get the Slack Channels, process it and send it back to the frontend.
- Replaced existing textbox input at SlackChannelConfigCreationForm.tsx with SearchMultiSelectDropdown.
- Stretch goal: Redirected user to channel configuration page after Slack bot creation.

## How Has This Been Tested?

Smoke tested!


yuhongsun96:
@pablonyx, did we accidentally rebuild this?

yuhongsun96:
Awesome work! We put this one in and there was a shoutout to this PR, closing as the feature is complete.

Thanks a ton for the contrib!

PR #3724, opened 2025-01-20
Infra and Deployment for ECS Fargate

trial-danswer:
## Description

## How Has This Been Tested?


deployment/data/nginx/app.conf.template
```diff
@@ -70,21 +70,3 @@ server {
         root /var/www/certbot;
     }
 }
-
-server {
```
Weves: we probably need to keep this around

deployment/aws_ecs_fargate/cloudformation/onyx_cluster_parameters.json
```diff
@@ -0,0 +1,5 @@
+[
+    {"ParameterKey": "Environment", "ParameterValue": "production"},
+    {"ParameterKey": "OnyxNamespace", "ParameterValue": "onyx2"},
```
Weves: maybe this should just be `onyx`

trial-danswer:
I've updated the PR with changes to include horizontal auto-scaling of tasks based on cpu and memory for api-backend, background, model-servers, and web-server.  Additionally, I've updated the templates to accept parameters for task memory, cpu, and desired count. Lastly, I set the default memory & cpu for tasks based on approx values from the Kubernetes helm-charts.  The ECS Fargate mem/cpu values are based on the following table: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-cpu-memory-error.html

deployment/data/nginx/app.conf.template.use-cloudtls
```diff
@@ -0,0 +1,69 @@
+# Log format to include request latency
```
Weves: what's the difference between this file and the `app.conf.template.dev`?

deployment/aws_ecs_fargate/cloudformation/services/onyx_services_parameters.json
```diff
@@ -0,0 +1,14 @@
+[
+{
```
Weves, line 2: can we add all the values that one might want to set in here? (e.g. HostedZoneId, DomainName, etc.)

deployment/aws_ecs_fargate/cloudformation/services/onyx_services_parameters.json
```diff
@@ -0,0 +1,14 @@
...
+	"ParameterKey": "Environment",
+	"ParameterValue": "production"
+},
+{
+	"ParameterKey": "SubnetIDs",
+	"ParameterValue": "subnet-0af372d676f65b287,subnet-06721b79e2e51d22c"
+},
+{
+	"ParameterKey": "VpcID",
```
Weves, line 11: some of these are duplicated across files, could we just centralize this into one large parameters json? I feel like that would be easier to track

Weves: similarly could there be a single place to configure resources across all the services?

trial-danswer: The parameters.json file passed to aws cloudformation deploy must match the parameters defined in the specific CloudFormation template being deployed. If a parameter is present in the JSON file but not defined in the template, CloudFormation will reject it. 

Possible to write a script that would parse out a single master parameter file...but that would become complex.

deployment/aws_ecs_fargate/cloudformation/services/onyx_services_parameters.json
trial-danswer: I had to separate the nginx parameters into a separate parameters.json file due to the fact that parameter file must match what is defined in the cloudformation template.  And the nginx cloudformation template has unqiue parameters defined unlike the other onyx services.

deployment/data/nginx/app.conf.template.use-cloudtls
trial-danswer: I merged it into app.conf.template.dev.  Prior to adding the env variables to run-nginx.sh I had to keep it separated..but now it can be a single conf.

PR #3726, opened 2025-01-20
Pass in tenant_id to kv_store in monitoring job

Weves:
## Description

We weren't passing in the right tenant ID for cloud-monitoring -> silent telemetry failures.

https://linear.app/danswer/issue/DAN-1296/fix-monitoring-telemetry-on-our-cloud

## How Has This Been Tested?

`kubectl cp`ed the files over to the monitoring pod and ran:

```
>>> from telemetry import optional_telemetry, RecordType
>>> data = {'metric_name': 'sync_start_latency_seconds', 'float_value': 0.0}
>>> optional_telemetry(record_type=RecordType.METRIC, data=data, user_id=None, tenant_id='<SOME_TENANT>')
```

Then verified we got the expected packet on the telemetry server.

### DIFF
--- a/backend/onyx/key_value_store/store.py
+++ b/backend/onyx/key_value_store/store.py
@@ -31,27 +31,27 @@ class PgRedisKVStore(KeyValueStore):
     def __init__(
         self, redis_client: Redis | None = None, tenant_id: str | None = None
     ) -> None:
+        self.tenant_id = tenant_id or CURRENT_TENANT_ID_CONTEXTVAR.get()
+
         # If no redis_client is provided, fall back to the context var
         if redis_client is not None:
             self.redis_client = redis_client
         else:
-            tenant_id = tenant_id or CURRENT_TENANT_ID_CONTEXTVAR.get()
-            self.redis_client = get_redis_client(tenant_id=tenant_id)
+            self.redis_client = get_redis_client(tenant_id=self.tenant_id)
 
     @contextmanager
-    def get_session(self) -> Iterator[Session]:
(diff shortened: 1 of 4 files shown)


backend/onyx/key_value_store/store.py
```diff
@@ -31,27 +31,27 @@ class PgRedisKVStore(KeyValueStore):
     def __init__(
         self, redis_client: Redis | None = None, tenant_id: str | None = None
     ) -> None:
+        self.tenant_id = tenant_id or CURRENT_TENANT_ID_CONTEXTVAR.get()
```
pablonyx, line 34: Slightly wary of default `None` to the current tenant ID. Elsewhere, we treat it as a valid tenant ID / as `POSTGRES_DEFAULT_SCHEMA`. Had a similar debate around `get_session_with_tenant` (default to a sentinel if nothing passed in?, etc.) and ended up deciding on different entrypoint (`get_session_with_default_tenant`)

pablonyx: https://github.com/onyx-dot-app/onyx/pull/3069

backend/onyx/utils/telemetry.py
```diff
@@ -94,16 +94,16 @@ def optional_telemetry(
     if DISABLE_TELEMETRY:
         return
 
+    tenant_id = tenant_id or CURRENT_TENANT_ID_CONTEXTVAR.get()
```
pablonyx, line 97: Similar comment here

pablonyx (approved):

PR #3730, opened 2025-01-21
Allow all LLMs for image generation assistants

pablonyx:
## Description

Fixes https://linear.app/danswer/issue/DAN-1297/never-hide-non-image-assistants

https://github.com/user-attachments/assets/9e6006de-61c5-4223-81ac-56ac6442f22c

## How Has This Been Tested?
- Multiple providers, create images, switch models with images in chat from user


Weves (commented):
Hmm, I'm running into as I tested again:

Screenshot 2025-01-21 at 4 43 46 PM

Maybe we could do some additional prompt tuning to prevent this? Or we could just hardcode a response based on the prompts sent over to dalle.

Also, would prefer to **not** have the warnings for a case where there are no images in the chat even if image generation is enabled. I don't think that represents much of a degradation in experience

Weves (approved):

PR #3735, opened 2025-01-21
Fix image wonkiness

pablonyx:
## Description

Fixes https://linear.app/danswer/issue/DAN-1302/images-appearing-wonky

## How Has This Been Tested?

- Use image in linear, ensure first one doesn't appear

### DIFF
--- a/backend/onyx/server/query_and_chat/chat_backend.py
+++ b/backend/onyx/server/query_and_chat/chat_backend.py
@@ -5,7 +5,6 @@ import os
 import uuid
 from collections.abc import Callable
 from collections.abc import Generator
-from typing import Tuple
 from uuid import UUID
 
 from fastapi import APIRouter
@@ -15,7 +14,6 @@ from fastapi import Request
 from fastapi import Response
 from fastapi import UploadFile
 from fastapi.responses import StreamingResponse
-from PIL import Image
 from pydantic import BaseModel
 from sqlalchemy.orm import Session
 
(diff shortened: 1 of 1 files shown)


backend/onyx/server/query_and_chat/chat_backend.py
```diff
@@ -758,5 +744,6 @@ def fetch_chat_file(
 
     media_type = file_record.file_type
     file_io = file_store.read_file(file_id, mode="b")
+    print("Meidate type: ", media_type)
```
rkuo-danswer: debugging code?

backend/onyx/server/query_and_chat/chat_backend.py
```diff
@@ -674,24 +660,24 @@ def upload_files_for_chat(
 
     file_store = get_default_file_store(db_session)
 
+    file_type_mapping = {
```
rkuo-danswer: should this function be defining its own acceptable content types or does that belong somewhere application wide?

backend/onyx/server/query_and_chat/chat_backend.py
```diff
@@ -674,24 +660,24 @@ def upload_files_for_chat(
...
-        elif file.content_type in document_content_types:
-            file_type = ChatFileType.DOC
-            file_content = io.BytesIO(file.file.read())
-            new_content_type = file.content_type or ""
+        file_type = file_type_mapping.get(
+            str(file.content_type), ChatFileType.PLAIN_TEXT
+        )
+
+        if file_type == ChatFileType.IMAGE:
+            file_content = file.file
```
rkuo-danswer, line 676: i would say if we don't want JPEG, we should leave a note here about why we wanted it in the first place and the reason for reverting it (which I assume is the transparency issue).  It's also possible to look for transparency in the image and decide based on that whether converting to JPEG makes sense.

rkuo-danswer (approved):

PR #3736, opened 2025-01-21
Add model updates

Weves:
## Description

https://linear.app/danswer/issue/DAN-1305/support-automatic-llm-updates-for-netflix-specific-api-format-backport

This is ported over from main + adds support for a different API format.

## How Has This Been Tested?

UT + testing with LiteLLM proxy locally

### DIFF
--- a/backend/danswer/background/celery/tasks/beat_schedule.py
+++ b/backend/danswer/background/celery/tasks/beat_schedule.py
@@ -1,6 +1,8 @@
 from datetime import timedelta
 from typing import Any
 
+from backend.danswer.configs.app_configs import LLM_MODEL_UPDATE_API_URL
+
 from danswer.configs.constants import DanswerCeleryPriority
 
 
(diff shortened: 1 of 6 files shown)


backend/danswer/background/celery/tasks/beat_schedule.py
```diff
@@ -55,6 +56,19 @@
...
 
+# Only add the LLM model update task if the API URL is configured
+if LLM_MODEL_UPDATE_API_URL:
+    tasks_to_schedule.append(
+        {
+            "name": "check-for-llm-model-update",
+            "task": "check_for_llm_model_update",
+            "schedule": timedelta(hours=1),  # Check every hour
+            "options": {
+                "priority": DanswerCeleryPriority.LOW,
```
pablonyx, line 67: Considering this is environment variable based, likely make sense to exclude this until we have a more built-out way of specifying tenant-specific configurations

pablonyx:
Looks good– I'm sure this has been thought through but it does assume the default provider remains the one that is compatible with `LLM_MODEL_UPDATE_API_URL` but presumably this is a safe assumption for the customer this is built for.

PR #3748, opened 2025-01-22
try using a redis replica in some areas

rkuo-danswer:
## Description

Fixes DAN-1321. https://linear.app/danswer/issue/DAN-1321/allow-usage-of-redis-replica

## How Has This Been Tested?

### DIFF
--- a/backend/onyx/redis/redis_pool.py
+++ b/backend/onyx/redis/redis_pool.py
@@ -21,6 +21,7 @@ from onyx.configs.app_configs import REDIS_HOST
 from onyx.configs.app_configs import REDIS_PASSWORD
 from onyx.configs.app_configs import REDIS_POOL_MAX_CONNECTIONS
 from onyx.configs.app_configs import REDIS_PORT
+from onyx.configs.app_configs import REDIS_REPLICA_HOST
 from onyx.configs.app_configs import REDIS_SSL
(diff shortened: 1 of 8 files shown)


backend/onyx/redis/redis_pool.py
```diff
@@ -132,23 +133,32 @@ class RedisPool:
...
     def get_client(self, tenant_id: str | None) -> Redis:
         if tenant_id is None:
             tenant_id = "public"
         return TenantRedis(tenant_id, connection_pool=self._pool)
 
+    def get_replica_client(self, tenant_id: str | None) -> Redis:
+        if tenant_id is None:
+            tenant_id = "public"
+        return TenantRedis(tenant_id, connection_pool=self._replica_pool)
+
```
pablonyx, line 161: Unrelated to this change but we should think about the benefits of using a constant/config for the default single-tenant (ie. none) here. Not sure if `public` makes sense

pablonyx (approved):
LGTM- though we should think about at what point it may make sense to support this for non-multi tenant customers (ie. include in deployment files)

backend/onyx/redis/redis_pool.py
rkuo-danswer: Agreed and i think "public" is misleading in the cloud.

PR #3749, opened 2025-01-23
Agent search feature

pablonyx:
## Description

https://linear.app/danswer/issue/DAN-1405/agentic-search

## How Has This Been Tested?

Unit tested and extensively tested in the UI. Integration tests should follow at some point.


backend/onyx/agents/agent_search/basic/utils.py
```diff
@@ -0,0 +1,69 @@
...
+def process_llm_stream(
+    stream: Iterator[BaseMessage],
+    should_stream_answer: bool,
+    final_search_results: list[LlmDoc] | None = None,
+    displayed_search_results: list[LlmDoc] | None = None,
+) -> AIMessageChunk:
+    tool_call_chunk = AIMessageChunk(content="")
+    # for response in response_handler_manager.handle_llm_response(stream):
+
+    if final_search_results and displayed_search_results:
```
evan-onyx, line 30: @yuhongsun96  we're using a CitationResponseHandler here, but using other code to handle agent search citations. A little unfortunate

backend/onyx/agents/agent_search/basic/utils.py
```diff
@@ -0,0 +1,69 @@
...
+        answer_handler = PassThroughAnswerResponseHandler()
+
+    full_answer = ""
+    # This stream will be the llm answer if no tool is chosen. When a tool is chosen,
+    # the stream will contain AIMessageChunks with tool call information.
+    for response in stream:
+        answer_piece = response.content
+        if not isinstance(answer_piece, str):
+            # TODO: handle non-string content
+            logger.warning(f"Received non-string content: {type(answer_piece)}")
```
evan-onyx: @yuhongsun96 in several places we just assume the output is a string, let us know if you want any specific handling of unexpected types

backend/onyx/agents/agent_search/deep_search_a/answer_initial_sub_question/nodes/answer_check.py
```diff
@@ -0,0 +1,59 @@
...
+
+    agent_searchch_config = cast(AgentSearchConfig, config["metadata"]["config"])
+    fast_llm = agent_searchch_config.fast_llm
+    response = list(
+        fast_llm.stream(
+            prompt=msg,
+        )
+    )
+
+    quality_str = merge_message_runs(response, chunk_separator="")[0].content
```
evan-onyx: @yuhongsun96 Is this best practice for merging a list of streamed messages?

backend/onyx/agents/agent_search/deep_search_a/answer_initial_sub_question/nodes/format_answer.py
```diff
@@ -0,0 +1,28 @@
...
+)
+
+
+def format_answer(state: AnswerQuestionState) -> AnswerQuestionOutput:
+    return AnswerQuestionOutput(
+        answer_results=[
+            QuestionAnswerResults(
+                question=state.question,
+                question_id=state.question_id,
+                quality=state.answer_quality
```
evan-onyx: probably should use bool here?

backend/onyx/agents/agent_search/deep_search_a/answer_initial_sub_question/states.py
```diff
@@ -0,0 +1,71 @@
...
+        ""  # 0_0 is original question, everything else is <level>_<question_num>.
+    )
+    # level 0 is original question and first decomposition, level 1 is follow up, etc
+    # question_num is a unique number per original question per level.
+
+
+## Graph State
+
+
+class AnswerQuestionState(
```
evan-onyx: @yuhongsun96 Let us know if you think up a better paradigm for handling langgraph state, at the moment there's lots of subclassing which can get a bit messy. I'm not convinced there's anything better, but felt like an important shout-out.

backend/onyx/agents/agent_search/deep_search_a/base_raw_search/nodes/ingest_initial_base_retrieval.py
```diff
@@ -0,0 +1,39 @@
...
+
+from onyx.agents.agent_search.deep_search_a.base_raw_search.states import (
+    BaseRawSearchOutput,
+)
+from onyx.agents.agent_search.deep_search_a.main.operations import logger
+from onyx.agents.agent_search.deep_search_a.main.states import ExpandedRetrievalUpdate
+from onyx.agents.agent_search.shared_graph_utils.models import AgentChunkStats
+
+
+def ingest_initial_base_retrieval(
```
evan-onyx: I'd like to do some name refactoring; to me "base" doesn't really mean much here

backend/onyx/agents/agent_search/deep_search_a/expanded_retrieval/edges.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def parallel_retrieval_edge(
+    state: ExpandedRetrievalState, config: RunnableConfig
+) -> list[Send | Hashable]:
+    agent_a_config = cast(AgentSearchConfig, config["metadata"]["config"])
+    question = state.question if state.question else agent_a_config.search_request.query
+
+    query_expansions = (
+        state.expanded_queries if state.expanded_queries else [] + [question]
```
evan-onyx: unclear precedence here; if we want [question] to always be part of query_expansions then the ternary statement needs parentheses. Otherwise, there's no need for the empty list, just `else [question]`. @joachim-danswer

backend/onyx/agents/agent_search/deep_search_a/expanded_retrieval/nodes/doc_reranking.py
```diff
@@ -0,0 +1,82 @@
...
+    verified_documents = state.verified_documents
+
+    # Rerank post retrieval and verification. First, create a search query
+    # then create the list of reranked sections
+
+    agent_a_config = cast(AgentSearchConfig, config["metadata"]["config"])
+    question = state.question if state.question else agent_a_config.search_request.query
+    if agent_a_config.search_tool is None:
+        raise ValueError("search_tool must be provided for agentic search")
+    with get_session_context_manager() as db_session:
```
evan-onyx: @yuhongsun96 is this approach to getting a db session fine?

backend/onyx/agents/agent_search/deep_search_a/expanded_retrieval/nodes/doc_reranking.py
```diff
@@ -0,0 +1,82 @@
...
+    ):
+        if len(verified_documents) > 1:
+            reranked_documents = rerank_sections(
+                _search_query,
+                verified_documents,
+            )
+        else:
+            num = "No" if len(verified_documents) == 0 else "One"
+            logger.warning(f"{num} verified document(s) found, skipping reranking")
+            reranked_documents = verified_documents
```
evan-onyx: @yuhongsun96 we have a PR for disallowing rerank when user has a local model with no gpu; joachim wanted to talk more with you about this

backend/onyx/agents/agent_search/deep_search_a/expanded_retrieval/nodes/doc_retrieval.py
```diff
@@ -0,0 +1,105 @@
...
+                    predicted_search=response.predicted_search,
+                    final_filters=response.final_filters,
+                    recency_bias_multiplier=response.recency_bias_multiplier,
+                )
+                break
+
+    retrieved_docs = retrieved_docs[:AGENT_MAX_QUERY_RETRIEVAL_RESULTS]
+    pre_rerank_docs = retrieved_docs
+    if search_tool.search_pipeline is not None:
+        pre_rerank_docs = (
```
evan-onyx: This is a hack to get stats, and adding state (search_pipeline) to the search tool means the stats are probably wrong when done in parallel.

backend/onyx/agents/agent_search/deep_search_a/expanded_retrieval/nodes/doc_verification.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+
+    question = state.question
+    doc_to_verify = state.doc_to_verify
+    document_content = doc_to_verify.combined_content
+
+    agent_a_config = cast(AgentSearchConfig, config["metadata"]["config"])
+    fast_llm = agent_a_config.fast_llm
+
+    document_content = trim_prompt_piece(
```
evan-onyx: @yuhongsun96 Definitely take a look at trim_prompt_piece, see if you think we should design something fancier

backend/onyx/agents/agent_search/deep_search_a/expanded_retrieval/nodes/expand_queries.py
```diff
@@ -0,0 +1,68 @@
...
+
+def expand_queries(
+    state: ExpandedRetrievalInput, config: RunnableConfig
+) -> QueryExpansionUpdate:
+    # Sometimes we want to expand the original question, sometimes we want to expand a sub-question.
+    # When we are running this node on the original question, no question is explictly passed in.
+    # Instead, we use the original question from the search request.
+    agent_a_config = cast(AgentSearchConfig, config["metadata"]["config"])
+    now_start = datetime.now()
+    question = (
```
evan-onyx: This conditional should be removed now that we're using BaseModels

backend/onyx/agents/agent_search/deep_search_a/expanded_retrieval/nodes/expand_queries.py
```diff
@@ -0,0 +1,68 @@
...
+    if chat_session_id is None:
+        raise ValueError("chat_session_id must be provided for agent search")
+
+    msg = [
+        HumanMessage(
+            content=REWRITE_PROMPT_MULTI_ORIGINAL.format(question=question),
+        )
+    ]
+
+    llm_response_list = dispatch_separated(
```
evan-onyx: @yuhongsun96 let us know if you're not a fan of our separated dispatch method

backend/onyx/agents/agent_search/deep_search_a/expanded_retrieval/nodes/verification_kickoff.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def verification_kickoff(
+    state: ExpandedRetrievalState,
+    config: RunnableConfig,
+) -> Command[Literal["doc_verification"]]:
+    documents = state.retrieved_documents
+    agent_a_config = cast(AgentSearchConfig, config["metadata"]["config"])
+    verification_question = (
+        state.question
```
evan-onyx: Should remove bc BaseModel

backend/onyx/agents/agent_search/deep_search_a/expanded_retrieval/operations.py
```diff
@@ -0,0 +1,97 @@
...
+    verified_documents: list[InferenceSection],
+    expanded_retrieval_results: list[QueryResult],
+) -> AgentChunkStats:
+    chunk_scores: dict[str, dict[str, list[int | float]]] = defaultdict(
+        lambda: defaultdict(list)
+    )
+
+    for expanded_retrieval_result in expanded_retrieval_results:
+        for doc in expanded_retrieval_result.search_results:
+            doc_chunk_id = f"{doc.center_chunk.document_id}_{doc.center_chunk.chunk_id}"
```
evan-onyx: should use the make_question_id helper function

backend/onyx/agents/agent_search/deep_search_a/expanded_retrieval/operations.py
```diff
@@ -0,0 +1,97 @@
...
+    ]
+    dismissed_doc_chunk_ids = []
+
+    raw_chunk_stats_counts: dict[str, int] = defaultdict(int)
+    raw_chunk_stats_scores: dict[str, float] = defaultdict(float)
+    for doc_chunk_id, chunk_data in chunk_scores.items():
+        if doc_chunk_id in verified_doc_chunk_ids:
+            raw_chunk_stats_counts["verified_count"] += 1
+
+            valid_chunk_scores = [
```
evan-onyx: duplicated code in the if/else cases; calculate mean chunk score before the conditional

backend/onyx/agents/agent_search/deep_search_a/initial_search_sq_subgraph/nodes/generate_initial_answer.py
```diff
@@ -0,0 +1,272 @@
...
+) -> InitialAnswerUpdate:
+    now_start = datetime.now()
+
+    logger.debug(f"--------{now_start}--------GENERATE INITIAL---")
+
+    agent_a_config = cast(AgentSearchConfig, config["metadata"]["config"])
+    question = agent_a_config.search_request.query
+    persona_prompt = get_persona_prompt(agent_a_config.search_request.persona)
+
+    history = build_history_prompt(agent_a_config.prompt_builder)
```
evan-onyx: @yuhongsun96 how we're building prompts (especially w.r.t. user configs) deserves scrutiny

backend/onyx/agents/agent_search/deep_search_a/initial_search_sq_subgraph/nodes/generate_initial_answer.py
```diff
@@ -0,0 +1,272 @@
...
+                "tool_response",
+                ExtendedToolResponse(
+                    id=tool_response.id,
+                    response=tool_response.response,
+                    level=0,
+                    level_question_nr=0,  # 0, 0 is the base question
+                ),
+            )
+
+        net_new_original_question_docs = []
```
evan-onyx: list comprehension

backend/onyx/agents/agent_search/deep_search_a/initial_search_sq_subgraph/nodes/generate_initial_answer.py
```diff
@@ -0,0 +1,272 @@
...
+
+        decomp_answer_results = state.decomp_answer_results
+
+        good_qa_list: list[str] = []
+
+        sub_question_nr = 1
+
+        for decomp_answer_result in decomp_answer_results:
+            decomp_questions.append(decomp_answer_result.question)
+            _, question_nr = parse_question_id(decomp_answer_result.question_id)
```
evan-onyx: somewhat concerned that we're reconstructing question numbers here

backend/onyx/agents/agent_search/deep_search_a/initial_search_sq_subgraph/nodes/generate_initial_answer.py
```diff
@@ -0,0 +1,272 @@
...
+        )
+
+        msg = [
+            HumanMessage(
+                content=base_prompt.format(
+                    question=question,
+                    answered_sub_questions=remove_document_citations(
+                        sub_question_answer_str
+                    ),
+                    relevant_docs=format_docs(relevant_docs),
```
evan-onyx: repeat call to format_docs, should pass in doc_context (also, this means the prompt is not trimmed correctly!)

backend/onyx/agents/agent_search/deep_search_a/initial_search_sq_subgraph/nodes/generate_initial_answer.py
```diff
@@ -0,0 +1,272 @@
...
+                    persona_specification=persona_specification,
+                    history=history,
+                    date_prompt=date_str,
+                )
+            )
+        ]
+
+        streamed_tokens: list[str | list[str | dict[str, Any]]] = [""]
+        for message in model.stream(msg):
+            # TODO: in principle, the answer here COULD contain images, but we don't support that yet
```
evan-onyx: @yuhongsun96 we do the stream + content is str + dispatch something pretty often, maybe warrants a helper where we pass in a callback to do the event dispatch?

backend/onyx/agents/agent_search/models.py
```diff
@@ -0,0 +1,93 @@
...
+    use_agentic_search: bool = False
+
+    # For persisting agent search data
+    chat_session_id: UUID | None = None
+
+    # The message ID of the user message that triggered the Pro Search
+    message_id: int | None = None
+
+    # Whether to persistence data for the Pro Search (turned off for testing)
+    use_persistence: bool = True
```
yuhongsun96: Should be cleaned up now

backend/onyx/agents/agent_search/models.py
```diff
@@ -0,0 +1,93 @@
...
+
+    # The database session for the Pro Search
+    db_session: Session | None = None
```
yuhongsun96: This should not be part of a config object

backend/onyx/agents/agent_search/models.py
```diff
@@ -0,0 +1,93 @@
...
+from onyx.chat.prompt_builder.answer_prompt_builder import AnswerPromptBuilder
+from onyx.context.search.models import SearchRequest
+from onyx.file_store.utils import InMemoryChatFile
+from onyx.llm.interfaces import LLM
+from onyx.tools.force import ForceUseTool
+from onyx.tools.tool import Tool
+from onyx.tools.tool_implementations.search.search_tool import SearchTool
+
+
+@dataclass
```
yuhongsun96: Don't use dataclass

backend/onyx/agents/agent_search/models.py
```diff
@@ -0,0 +1,93 @@
...
+    # message_history: list[PreviousMessage] | None = None
+    # single_message_history: str | None = None
+    prompt_builder: AnswerPromptBuilder
+
+    search_tool: SearchTool | None = None
+
+    use_agentic_search: bool = False
+
+    # For persisting agent search data
+    chat_session_id: UUID | None = None
```
yuhongsun96: These aren't configs. We can consider having a set of objects we pass in rather than a single amalgamate object

backend/onyx/chat/process_message.py
```diff
@@ -707,28 +826,41 @@ def stream_chat_message_objects(
...
-        tool_result = None
+        # reference_db_search_docs = None
+        # qa_docs_response = None
+        # # any files to associate with the AI message e.g. dall-e generated images
+        # ai_message_files = []
+        # dropped_indices = None
+        # tool_result = None
 
+        # TODO: different channels for stored info when it's coming from the agent flow
+        info_by_subq: dict[tuple[int, int], AnswerPostInfo] = defaultdict(
```
yuhongsun96: Don't use tuple[int, int], use a Pydantic object so we know what this means. At that model level, it would be great to explain what the level and level_question_nr is. Pref being explicit, what is nr?

backend/onyx/chat/process_message.py
```diff
@@ -688,6 +721,92 @@ def stream_chat_message_objects(
...
+                else None
+            ),
+        )
+        # TODO: Since we're deleting the current main path in Answer,
+        # we should construct this unconditionally inside Answer instead
+        # Leaving it here for the time being to avoid breaking changes
+        search_tools = [tool for tool in tools if isinstance(tool, SearchTool)]
+        search_tool: SearchTool | None = None
+
+        if len(search_tools) > 1:
```
yuhongsun96: Should raise here and fail loudly so we don't have undetected errors.

backend/onyx/configs/agent_configs.py
```diff
@@ -0,0 +1,94 @@
...
+    "AGENT_RETRIEVAL_STATS", False
+)
+
+AGENT_RETRIEVAL_STATS: bool = False
+if isinstance(agent_retrieval_stats_os, str) and agent_retrieval_stats_os == "True":
+    AGENT_RETRIEVAL_STATS = True
+elif isinstance(agent_retrieval_stats_os, bool) and agent_retrieval_stats_os:
+    AGENT_RETRIEVAL_STATS = True
+
+agent_max_query_retrieval_results_os: int | str = os.environ.get(
```
yuhongsun96: # Assuming we don't want to allow 0 results here which seems reasonable.
try:
    AGENT_MAX_QUERY_RETRIEVAL_RESULTS = int(
        os.environ.get("AGENT_MAX_QUERY_RETRIEVAL_RESULTS") or NUM_RETURNED_HITS
    )
except ValueError:
    raise ValueError(
        f"AGENT_MAX_QUERY_RETRIEVAL_RESULTS must be an integer, got {os.environ.get('AGENT_MAX_QUERY_RETRIEVAL_RESULTS')}"
    )

yuhongsun96: You can even just translate it to an int directly and if it fails so be it, it will just raise and it'll be clear to everyone anyway

backend/onyx/configs/agent_configs.py
```diff
@@ -0,0 +1,94 @@
...
+
+#####
+# Agent Configs
+#####
+
+agent_retrieval_stats_os: bool | str | None = os.environ.get(
+    "AGENT_RETRIEVAL_STATS", False
+)
+
+AGENT_RETRIEVAL_STATS: bool = False
```
yuhongsun96: AGENT_RETRIEVAL_STATS = str(agent_retrieval_stats_os).lower() == "true"

I'm going to lose my mind if you make me review more of this :'(

backend/onyx/configs/agent_configs.py
```diff
@@ -0,0 +1,94 @@
+import os
+
+from .chat_configs import NUM_RETURNED_HITS
```
yuhongsun96: no local imports

backend/onyx/db/chat.py
```diff
@@ -877,6 +939,121 @@ def translate_db_message_to_chat_message_detail(
...
+            level=level,
+            level_question_nr=level_question_nr,
+            sub_question=sub_question,
+            sub_answer=sub_answer,
+            sub_question_doc_results=sub_document_results,
+        )
+
+        db_session.add(sub_question_object)
+        db_session.commit()
+        # db_session.flush()
```
yuhongsun96: Remove dead code

yuhongsun96 (commented):
Another small batch set of reviews here

backend/onyx/agents/agent_search/deep_search_a/answer_initial_sub_question/nodes/answer_check.py
```diff
@@ -0,0 +1,59 @@
...
+
+    now_end = datetime.now()
+    return QACheckUpdate(
+        answer_quality=quality_str,
+        log_messages=[
+            f"""{now_start} -- Answer check SQ-{level}-{question_num} - Answer quality: {quality_str},
```
yuhongsun96: No reason to use a """ triple quoted string here

backend/onyx/agents/agent_search/deep_search_a/answer_initial_sub_question/nodes/answer_check.py
yuhongsun96: idk, I've never used this LangChain util. I mean it seems fine, you seem to be using it for the intended purpose.

backend/onyx/agents/agent_search/deep_search_a/answer_initial_sub_question/states.py
```diff
@@ -0,0 +1,71 @@
...
+from onyx.agents.agent_search.shared_graph_utils.operators import (
+    dedup_inference_sections,
+)
+from onyx.context.search.models import InferenceSection
+
+
+## Update States
+class QACheckUpdate(BaseModel):
+    answer_quality: str = ""
+    log_messages: list[str] = []
```
yuhongsun96: Is there a reason log messages need to be passed around? Why can't we just log it at the spot and be done with it?

backend/onyx/agents/agent_search/deep_search_a/answer_initial_sub_question/states.py
```diff
@@ -0,0 +1,71 @@
+from operator import add
```
yuhongsun96: keep it consistent with the rest of the code base. These "state container" files we tend to call models.py

backend/onyx/agents/agent_search/deep_search_a/answer_initial_sub_question/nodes/format_answer.py
yuhongsun96: Prefer explicit "None"s. It should be a nullable type and special handling upstream to handle the null case

backend/onyx/agents/agent_search/deep_search_a/answer_initial_sub_question/nodes/format_answer.py
```diff
@@ -0,0 +1,28 @@
...
+from onyx.agents.agent_search.shared_graph_utils.models import (
+    QuestionAnswerResults,
+)
+
+
+def format_answer(state: AnswerQuestionState) -> AnswerQuestionOutput:
+    return AnswerQuestionOutput(
+        answer_results=[
+            QuestionAnswerResults(
+                question=state.question,
```
yuhongsun96: Pref using the same names throughout and mapping it in a more concise way than typing each of these out like that.

backend/onyx/agents/agent_search/deep_search_a/base_raw_search/nodes/ingest_initial_base_retrieval.py
yuhongsun96: Yes, please refactor all the names to be as clear as possible

backend/onyx/agents/agent_search/deep_search_a/answer_initial_sub_question/states.py
evan-onyx: These are different from the rest of the codebase; rather than standard state containers they're langgraph-specific states and state updates

backend/onyx/agents/agent_search/deep_search_a/answer_initial_sub_question/nodes/format_answer.py
evan-onyx: Unfortunately this is largely a necessity for langgraph nodes

backend/onyx/agents/agent_search/deep_search_a/expanded_retrieval/edges.py
yuhongsun96: Depends on what you guys are looking for but my guess is that we do always want to include the question. In which case yes! add parens

backend/onyx/agents/agent_search/deep_search_a/expanded_retrieval/states.py
```diff
@@ -0,0 +1,91 @@
...
+    question: str = ""
+    base_search: bool = False
+    sub_question_id: str | None = None
+
+
+## Update/Return States
+
+
+class QueryExpansionUpdate(BaseModel):
+    expanded_queries: list[str] = ["aaa", "bbb"]
```
yuhongsun96: What are these default values, guys cmon

backend/onyx/agents/agent_search/deep_search_a/base_raw_search/nodes/ingest_initial_base_retrieval.py
yuhongsun96: sounds good, please refactor

backend/onyx/agents/agent_search/deep_search_a/expanded_retrieval/nodes/doc_reranking.py
yuhongsun96: I would test this to make sure, it looks ok but deploy it locally as multi-tenant (ask Pablo how) and set a breakpoint here to make sure you pick up the Tenant

backend/onyx/agents/agent_search/deep_search_a/expanded_retrieval/nodes/doc_reranking.py
yuhongsun96: Ok, sounds good, can sync tomorrow

backend/onyx/agents/agent_search/basic/utils.py
yuhongsun96: We should probably use the same style of objects at least, even if there is more than one of them for each case.

backend/onyx/agents/agent_search/basic/utils.py
yuhongsun96: If we're confident that in all our flows it's a string, you can just assert it

backend/onyx/agents/agent_search/deep_search_a/answer_initial_sub_question/states.py
yuhongsun96: Could have the other things as attributes of this class rather than inheriting like this. That way you don't get the diamond problem. In this case I think the "parent classes" are pretty logically cohesive and it makes sense to say the AnswerQuestionState (idk about this name btw), contains the input for it as well as other things as attributes

backend/onyx/agents/agent_search/deep_search_a/expanded_retrieval/nodes/doc_retrieval.py
yuhongsun96: There is some weirdness going on here this section. Why are we accessing private attributes. Also how could search_pipeline be None here? it becomes not None during thesearch_tool.run earlier no?

backend/onyx/agents/agent_search/deep_search_a/expanded_retrieval/nodes/doc_verification.py
yuhongsun96: What do you have in mind? I do not like using the tokenizer to truncate every doc that passes through. Does this not add latency?

yuhongsun96: Maybe it's not that slow, I'm not sure... double check though if we're keeping this

backend/onyx/agents/agent_search/deep_search_a/expanded_retrieval/nodes/expand_queries.py
yuhongsun96: First, what is agent_a_config, don't use this type of naming.

yuhongsun96: Also don't call things now_start

yuhongsun96: Yes, remove all redundant logic.

backend/onyx/agents/agent_search/deep_search_a/util__expanded_retrieval__subgraph/nodes/expand_queries.py
```diff
@@ -0,0 +1,74 @@
...
+
+def expand_queries(
+    state: ExpandedRetrievalInput, config: RunnableConfig
+) -> QueryExpansionUpdate:
+    # Sometimes we want to expand the original question, sometimes we want to expand a sub-question.
+    # When we are running this node on the original question, no question is explictly passed in.
+    # Instead, we use the original question from the search request.
+    agent_a_config = cast(AgentSearchConfig, config["metadata"]["config"])
+    now_start = datetime.now()
+    question = (
```
yuhongsun96: Remove dead code

backend/onyx/agents/agent_search/shared_graph_utils/utils.py
```diff
@@ -0,0 +1,367 @@
...
+
+
+# Post-processing
+def format_docs(docs: Sequence[InferenceSection]) -> str:
+    formatted_doc_list = []
+
+    for doc_nr, doc in enumerate(docs):
+        formatted_doc_list.append(f"Document D{doc_nr + 1}:\n{doc.combined_content}")
+
+    return "\n\n".join(formatted_doc_list)
```
yuhongsun96: pref module or global final variable for this stuff especially if it's not a single char. That way a single change can be propagated throughout without missing cases

But if you have \n\n in different places that logically correspond to different things, they should have different final variables

backend/onyx/agents/agent_search/shared_graph_utils/utils.py
```diff
@@ -0,0 +1,367 @@
...
+def dispatch_main_answer_stop_info(level: int) -> None:
+    stop_event = StreamStopInfo(
+        stop_reason=StreamStopReason.FINISHED,
+        stream_type="main_answer",
+        level=level,
+    )
+    dispatch_custom_event("stream_finished", stop_event)
+
+
+def get_today_prompt() -> str:
```
yuhongsun96: This exists in some other place, keep it close by or at least in some consistent organization

backend/onyx/agents/agent_search/shared_graph_utils/utils.py
```diff
@@ -0,0 +1,367 @@
...
+    content: str, dispatch_event: Callable[[str, int], None], num: int
+) -> None:
+    if content != "":
+        dispatch_event(content, num)
+
+
+def dispatch_separated(
+    token_itr: Iterator[BaseMessage],
+    dispatch_event: Callable[[str, int], None],
+    sep: str = "\n",
```
yuhongsun96: Should be some global final var

backend/onyx/agents/agent_search/shared_graph_utils/utils.py
```diff
@@ -0,0 +1,367 @@
...
+) -> list[str | list[str | dict[str, Any]]]:
```
yuhongsun96: this typing is too complex, it should be some object

backend/onyx/agents/agent_search/shared_graph_utils/utils.py
```diff
@@ -0,0 +1,367 @@
...
+    num = 1
+    streamed_tokens: list[str | list[str | dict[str, Any]]] = [""]
```
yuhongsun96: pref explicit Nones anywhere possible. We should not have empty strings representing null cases, empty list is fine

backend/onyx/agents/agent_search/shared_graph_utils/utils.py
```diff
@@ -0,0 +1,367 @@
...
+    for message in token_itr:
```
yuhongsun96: "for token in token_itr", or even just for "token in tokens". Don't make reader have to keep track of all of these mappings in their heads. Similar to how you can't just name variables "for a in b" and cycle through the letters of the alphabet.

backend/onyx/agents/agent_search/shared_graph_utils/utils.py
```diff
@@ -0,0 +1,367 @@
...
+    return f"{level}_{question_nr}"
+
+
+def parse_question_id(question_id: str) -> tuple[int, int]:
+    level, question_nr = question_id.split("_")
+    return int(level), int(question_nr)
+
+
+def _dispatch_nonempty(
+    content: str, dispatch_event: Callable[[str, int], None], num: int
```
yuhongsun96: pref more clear naming, num of what?

backend/onyx/agents/agent_search/shared_graph_utils/utils.py
```diff
@@ -0,0 +1,367 @@
...
+        force_use_tool=ForceUseTool(force_use=False, tool_name=""),
+        prompt_builder=AnswerPromptBuilder(
+            user_message=HumanMessage(content=search_request.query),
+            message_history=[],
+            llm_config=primary_llm.config,
+            raw_user_query=search_request.query,
+            raw_user_uploaded_files=[],
+        ),
+        # chat_session_id=UUID("123e4567-e89b-12d3-a456-426614174000"),
+        chat_session_id=UUID("edda10d5-6cef-45d8-acfb-39317552a1f4"),  # Joachim
```
yuhongsun96: Is this test code? This is not the right way to have maintainable tests

backend/onyx/agents/agent_search/deep_search_a/expanded_retrieval/nodes/expand_queries.py
yuhongsun96: I added some comments around it. The more unfortunate thing is I have no idea what dispatch is doing behind the scenes. I traced it like 4 calls deep into the LangGraph layer and just called it a day

backend/onyx/agents/agent_search/shared_graph_utils/utils.py
evan-onyx: This is a testing function, and the UUID was randomly generated and is tied to a local instance of Onyx. We can make a new one here if you prefer, but this makes running the main test function pretty annoying because we need to call the code that creates a new chat message chain, or test without persistence

backend/onyx/agents/agent_search/deep_search_a/expanded_retrieval/nodes/verification_kickoff.py
yuhongsun96: yes, please apply this type of change across the board

backend/onyx/agents/agent_search/deep_search_a/expanded_retrieval/operations.py
yuhongsun96: +1

backend/onyx/agents/agent_search/deep_search_a/expanded_retrieval/nodes/expand_queries.py
evan-onyx: It's how we stream stuff out from in the middle of nodes during graph execution; there's a handler behind the scenes that yields your custom event, so the event loop can then emit it at the place we call .astream_events()

backend/onyx/agents/agent_search/deep_search_a/initial_search_sq_subgraph/nodes/generate_initial_answer.py
yuhongsun96: Looks like this one is gone but probably this is a reasonable comment

backend/onyx/agents/agent_search/deep_search_a/main__graph/nodes/agent_search_start.py
```diff
@@ -0,0 +1,59 @@
...
+)
+from onyx.agents.agent_search.shared_graph_utils.utils import retrieve_search_docs
+from onyx.configs.agent_configs import AGENT_EXPLORATORY_SEARCH_RESULTS
+from onyx.context.search.models import InferenceSection
+
+
+def agent_search_start(
+    state: MainState, config: RunnableConfig
+) -> ExploratorySearchUpdate:
+    now_start = datetime.now()
```
yuhongsun96: Use function decorators (the ones we already have) for this type of timing

backend/onyx/agents/agent_search/deep_search_a/initial_search_sq_subgraph/nodes/generate_initial_answer.py
yuhongsun96: Yes, no repeat code

backend/onyx/agents/agent_search/models.py
```diff
@@ -0,0 +1,101 @@
...
+from onyx.chat.prompt_builder.answer_prompt_builder import AnswerPromptBuilder
+from onyx.context.search.models import SearchRequest
+from onyx.file_store.utils import InMemoryChatFile
+from onyx.llm.interfaces import LLM
+from onyx.tools.force import ForceUseTool
+from onyx.tools.tool import Tool
+from onyx.tools.tool_implementations.search.search_tool import SearchTool
+
+
+class AgentSearchConfig(BaseModel):
```
yuhongsun96: Would split this into maybe 2 logical state containers for like "MessageContext" and "AgentSearchConfig" and pass in db_session separately if needed

backend/onyx/agents/agent_search/models.py
```diff
@@ -0,0 +1,101 @@
...
+    use_agentic_search: bool = False
+
+    # For persisting agent search data
+    chat_session_id: UUID | None = None
+
+    # The message ID of the user message that triggered the Pro Search
+    message_id: int | None = None
+
+    # Whether to persistence data for Agentic Search (turned off for testing)
+    use_agentic_persistence: bool = True
```
yuhongsun96: Dev things like these, I prefer not propagating throughout the stack. This is bloat code that we want to avoid. I'm not sure why we would want to turn this off for testing at all, pref just not having this logic and persistence should always happen.

If you have to do this, just have a global final var that can depend on DEV_MODE (which is in app_configs.py) or just use DEV_MODE directly

backend/onyx/chat/models.py
```diff
@@ -61,11 +64,17 @@ def model_dump(self, *args: list, **kwargs: dict[str, Any]) -> dict[str, Any]:
 class StreamStopReason(Enum):
     CONTEXT_LENGTH = "context_length"
     CANCELLED = "cancelled"
+    FINISHED = "finished"
 
 
 class StreamStopInfo(BaseModel):
     stop_reason: StreamStopReason
 
+    stream_type: Literal["", "sub_questions", "sub_answer", "main_answer"] = ""
```
yuhongsun96: Dont use empty strings to represent null cases

backend/onyx/chat/models.py
```diff
@@ -61,11 +64,17 @@ def model_dump(self, *args: list, **kwargs: dict[str, Any]) -> dict[str, Any]:
...
+    # used to identify the stream that was stopped for agent search
+    level: int | None = None
```
yuhongsun96: As discussed, the different levels are distinct and require unique handling and not like a "list of the same things" that are all processed the same. Use an Enum here. Also consider renaming, but that's less critical.

If later we need iterations of each level, we can add an int for the iteration number but this falls under "don't build things we don't need yet"

backend/onyx/chat/models.py
```diff
@@ -108,6 +117,8 @@ class OnyxAnswerPiece(BaseModel):
 class CitationInfo(BaseModel):
     citation_num: int
     document_id: str
+    level: int | None = None
```
yuhongsun96: Same as above, since this is the same code, I would prefer this being a separate model that both of these use or inherit. That way consistency is enforced rather than implicitly remembered by devs

backend/onyx/chat/models.py
```diff
@@ -299,11 +310,56 @@ def from_model(
     model_config = ConfigDict(frozen=True)
 
 
+class SubQueryPiece(BaseModel):
+    sub_query: str
+    level: int
```
yuhongsun96: This level and level_question_nr is in way too many places to be copy pasta-ed like that, they should be explicitly associated together

backend/onyx/chat/answer.py
```diff
@@ -115,188 +86,134 @@ def __init__(
...
+                isinstance(packet, AgentAnswerPiece)
+                and packet.answer_piece
+                and packet.answer_type == "agent_level_answer"
+                and packet.level == 0
+            ):
                 answer += packet.answer_piece
 
         return answer
 
+    def llm_answer_by_level(self) -> dict[int, str]:
```
yuhongsun96, line 159: This and citations_by_subquestion can be properties as well. They're constructing something which logically acts as a property of the instance of Answer. You should also just follow the practice in this class of caching it. This function is fast so it doesn't really matter but... consistency is good

backend/onyx/agents/agent_search/deep_search_a/main__graph/nodes/entity_term_extraction_llm.py
```diff
@@ -0,0 +1,126 @@
...
+from onyx.agents.agent_search.shared_graph_utils.models import Relationship
+from onyx.agents.agent_search.shared_graph_utils.models import Term
+from onyx.agents.agent_search.shared_graph_utils.prompts import ENTITY_TERM_PROMPT
+from onyx.agents.agent_search.shared_graph_utils.utils import format_docs
+
+
+def entity_term_extraction_llm(
+    state: MainState, config: RunnableConfig
+) -> EntityTermExtractionUpdate:
+    now_start = datetime.now()
```
yuhongsun96: Use existing decorator utilities for this type of stuff

backend/onyx/agents/agent_search/deep_search_a/main__graph/nodes/entity_term_extraction_llm.py
```diff
@@ -0,0 +1,126 @@
...
+
+    logger.info(f"--------{now_start}--------GENERATE ENTITIES & TERMS---")
```
yuhongsun96: Keep info and above levels clean. Timings unless important tend to go into debug.

backend/onyx/agents/agent_search/deep_search_a/main__graph/nodes/entity_term_extraction_llm.py
```diff
@@ -0,0 +1,126 @@
...
+
+    agent_a_config = cast(AgentSearchConfig, config["metadata"]["config"])
+    if not agent_a_config.allow_refinement:
+        now_end = datetime.now()
+        return EntityTermExtractionUpdate(
```
yuhongsun96: The nesting here is extraneous, just use the inner model and have that inherit the logs class

backend/onyx/agents/agent_search/deep_search_a/main__graph/nodes/entity_term_extraction_llm.py
```diff
@@ -0,0 +1,126 @@
...
+            entity_relation_term_extractions=EntityRelationshipTermExtraction(
+                entities=[],
```
yuhongsun96, line 40: prefer explicit Nones

backend/onyx/agents/agent_search/deep_search_a/main__graph/nodes/entity_term_extraction_llm.py
```diff
@@ -0,0 +1,126 @@
...
+                terms=[],
+            ),
+            log_messages=[
+                f"{now_start} -- Main - ETR Extraction,  Time taken: {now_end - now_start}"
+            ],
+        )
+
+    # first four lines duplicates from generate_initial_answer
+    question = agent_a_config.search_request.query
+    initial_search_docs = state.exploratory_search_results[:15]
```
yuhongsun96: No magic numbers like 15, this should be some config. If it's only applicable in this module, we can keep it as a module level final var to avoid cluttering shared config files.

backend/onyx/agents/agent_search/deep_search_a/main__graph/nodes/entity_term_extraction_llm.py
```diff
@@ -0,0 +1,126 @@
...
+        )
+    ]
+    fast_llm = agent_a_config.fast_llm
+    # Grader
+    llm_response = fast_llm.invoke(
+        prompt=msg,
+    )
+
+    cleaned_response = re.sub(r"```json\n|\n```", "", str(llm_response.content))
+    parsed_response = json.loads(cleaned_response)
```
yuhongsun96: We should load this into an object rather than arbitrary dict. We should have handling for json exceptions and object mismatch exceptions

backend/onyx/agents/agent_search/deep_search_a/main__graph/nodes/entity_term_extraction_llm.py
```diff
@@ -0,0 +1,126 @@
...
+
+    entities = []
+    relationships = []
+    terms = []
+    for entity in parsed_response.get("retrieved_entities_relationships", {}).get(
+        "entities", {}
+    ):
+        entity_name = entity.get("entity_name", "")
```
yuhongsun96: explicit Nones, never use empty strings to represent a null case

backend/onyx/agents/agent_search/deep_search_a/main__graph/nodes/entity_term_extraction_llm.py
```diff
@@ -0,0 +1,126 @@
...
+        entity_type = entity.get("entity_type", "")
+        entities.append(Entity(entity_name=entity_name, entity_type=entity_type))
+
+    for relationship in parsed_response.get("retrieved_entities_relationships", {}).get(
+        "relationships", {}
+    ):
+        relationship_name = relationship.get("relationship_name", "")
```
yuhongsun96: If a variable is only assigned and used once, probably just have it where it's used

backend/onyx/agents/agent_search/deep_search_a/main__graph/nodes/entity_term_extraction_llm.py
```diff
@@ -0,0 +1,126 @@
...
+                term_name=term_name,
+                term_type=term_type,
+                term_similar_to=term_similar_to,
+            )
+        )
+
+    now_end = datetime.now()
+
+    logger.info(
+        f"{now_start} -- MAIN - Entity term extraction,  Time taken: {now_end - now_start}"
```
yuhongsun96: Avoid repeat of same/similar strings like these

backend/onyx/agents/agent_search/deep_search_a/main__graph/nodes/generate_refined_answer.py
```diff
@@ -0,0 +1,335 @@
...
+
+    consolidated_context_docs: list[InferenceSection] = sub_questions_cited_docs
+
+    counter = 0
+    for original_doc_number, original_doc in enumerate(all_original_question_documents):
+        if original_doc_number not in sub_questions_cited_docs:
+            if (
+                counter <= AGENT_MIN_ORIG_QUESTION_DOCS
+                or len(consolidated_context_docs)
+                < 1.5
```
yuhongsun96: why is this 1.5? feels kind of weird

backend/onyx/agents/agent_search/deep_search_a/main__graph/nodes/generate_refined_answer.py
```diff
@@ -0,0 +1,335 @@
...
+    query_info = get_query_info(state.original_question_retrieval_results)
+    if agent_a_config.search_tool is None:
+        raise ValueError("search_tool must be provided for agentic search")
+    # stream refined answer docs
+    for tool_response in yield_search_responses(
+        query=question,
+        reranked_sections=relevant_docs,
+        final_context_sections=relevant_docs,
+        search_query_info=query_info,
+        get_section_relevance=lambda: None,  # TODO: add relevance
```
yuhongsun96: does this need to get addressed?

backend/onyx/agents/agent_search/deep_search_a/main__graph/nodes/generate_refined_answer.py
```diff
@@ -0,0 +1,335 @@
...
+
+    if len(initial_documents) > 0:
+        revision_doc_effectiveness = len(relevant_docs) / len(initial_documents)
+    elif len(refined_documents) == 0:
+        revision_doc_effectiveness = 0.0
+    else:
+        revision_doc_effectiveness = 10.0
+
+    decomp_answer_results = state.decomp_answer_results
+    # revised_answer_results = state.refined_decomp_answer_results
```
yuhongsun96: nit: cleanup related stuff

backend/onyx/agents/agent_search/deep_search_a/main__graph/nodes/refined_answer_decision.py
```diff
@@ -0,0 +1,47 @@
...
+
+def refined_answer_decision(
+    state: MainState, config: RunnableConfig
+) -> RequireRefinedAnswerUpdate:
+    now_start = datetime.now()
+
+    logger.info(f"--------{now_start}--------REFINED ANSWER DECISION---")
+
+    agent_a_config = cast(AgentSearchConfig, config["metadata"]["config"])
+    if "?" in agent_a_config.search_request.query:
```
yuhongsun96: need to clean this stuff

backend/onyx/agents/agent_search/deep_search_a/main__graph/nodes/direct_llm_handling.py
```diff
@@ -0,0 +1,83 @@
...
+
+    agent_a_config = cast(AgentSearchConfig, config["metadata"]["config"])
+    question = agent_a_config.search_request.query
+    persona_contextualialized_prompt = get_persona_agent_prompt_expressions(
+        agent_a_config.search_request.persona
+    ).contextualized_prompt
+
+    logger.info(f"--------{now_start}--------LLM HANDLING START---")
+
+    model = agent_a_config.fast_llm
```
yuhongsun96: nit (doesn't matter), consistent naming can just be llm or fast_llm

backend/onyx/agents/agent_search/deep_search_a/main__graph/nodes/generate_initial_base_search_only_answer.py
```diff
@@ -0,0 +1,58 @@
...
+    logger.info(f"--------{now_start}--------GENERATE INITIAL BASE ANSWER---")
+
+    agent_a_config = cast(AgentSearchConfig, config["metadata"]["config"])
+    question = agent_a_config.search_request.query
+    original_question_docs = state.all_original_question_documents
+
+    model = agent_a_config.fast_llm
+
+    doc_context = format_docs(original_question_docs)
+    doc_context = trim_prompt_piece(
```
yuhongsun96: Is there any way we can know this stuff ahead of time? If not, no worries

backend/onyx/agents/agent_search/deep_search_a/main__graph/states.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+class MainInput(CoreState):
+    pass
+
+
+## Graph State
+
+
+class MainState(
```
yuhongsun96: Pref comments on what these are

backend/onyx/agents/agent_search/deep_search_a/util__expanded_retrieval__subgraph/nodes/doc_reranking.py
```diff
@@ -0,0 +1,96 @@
...
+        and _search_query.rerank_settings.num_rerank > 0
+        and len(verified_documents) > 0
+    ):
+        if len(verified_documents) > 1:
+            reranked_documents = rerank_sections(
+                _search_query,
+                verified_documents,
+            )
+        else:
+            num = "No" if len(verified_documents) == 0 else "One"
```
yuhongsun96: Extraneous logic, try to keep things minimal and just use len in the print

backend/onyx/agents/agent_search/shared_graph_utils/prompts.py
```diff
@@ -0,0 +1,1113 @@
+UNKNOWN_ANSWER = "I do not have enough information to answer this question."
```
yuhongsun96: Pref all prompts to be under the backend/prompts directory

backend/onyx/server/query_and_chat/chat_backend.py
```diff
@@ -145,6 +145,7 @@ def get_chat_session(
     user: User | None = Depends(current_chat_accesssible_user),
     db_session: Session = Depends(get_session),
 ) -> ChatSessionDetailResponse:
+    print("get_chat_session called")
```
yuhongsun96: no prints

backend/onyx/server/query_and_chat/chat_backend.py
```diff
@@ -386,6 +389,8 @@ def stream_generator() -> Generator[str, None, None]:
                 ),
                 is_connected=is_connected_func,
             ):
+                # with open('chat_packets.log', 'a') as log_file:
```
yuhongsun96: clean up

backend/onyx/server/query_and_chat/chat_backend.py
```diff
@@ -311,7 +314,7 @@ async def is_connected(request: Request) -> Callable[[], bool]:
     def is_connected_sync() -> bool:
         future = asyncio.run_coroutine_threadsafe(request.is_disconnected(), main_loop)
         try:
-            is_connected = not future.result(timeout=0.01)
+            is_connected = not future.result(timeout=0.05)
```
yuhongsun96, line 357: Why did this need to get bumped up?

backend/onyx/chat/process_message.py
```diff
@@ -282,10 +303,22 @@ def _get_force_search_settings(
     | MessageSpecificCitations
     | MessageResponseIDInfo
     | StreamStopInfo
+    | AgentSearchPacket
 )
 ChatPacketStream = Iterator[ChatPacket]
 
 
+# can't store a DbSearchDoc in a Pydantic BaseModel
+@dataclass
```
yuhongsun96: We typically just do a translation from DB models to Pydantic because the underlying types are all Pydantic compatible

backend/onyx/chat/process_message.py
```diff
@@ -502,11 +536,8 @@ def stream_chat_message_objects(
         files = load_all_chat_files(
             history_msgs, new_msg_req.file_descriptors, db_session
         )
-        latest_query_files = [
-            file
-            for file in files
-            if file.file_id in [f["id"] for f in new_msg_req.file_descriptors]
-        ]
+        req_file_ids = [f["id"] for f in new_msg_req.file_descriptors]
```
yuhongsun96, line 539: We can probably use the attribute here right? (though I know previously it uses this approach too)

backend/onyx/agents/agent_search/run_graph.py
```diff
@@ -0,0 +1,261 @@
...
+                pass
+            loop.close()
+
+    return _yield_async_to_sync()
+
+
+def run_graph(
+    compiled_graph: CompiledStateGraph,
+    config: AgentSearchConfig,
+    input: BasicInput | MainInput_a,
```
yuhongsun96, line 95: MainInput_a <- nonstandard class names. No underscores typically, words with capital first letters

backend/onyx/agents/agent_search/run_graph.py
```diff
@@ -0,0 +1,261 @@
...
+) -> AnswerStream:
+    graph = basic_graph_builder()
+    compiled_graph = graph.compile()
+    # TODO: unify basic input
+    input = BasicInput()
+    return run_graph(compiled_graph, config, input)
+
+
+if __name__ == "__main__":
+    from onyx.llm.factory import get_default_llms
```
yuhongsun96: If we want to keep this stuff for easy demo/execution - let's clean this up a bit

backend/onyx/agents/agent_search/shared_graph_utils/agent_prompt_ops.py
```diff
@@ -0,0 +1,134 @@
...
+
+    date_str = get_today_prompt()
+
+    docs_format_list = [
+        f"""Document Number: [D{doc_nr + 1}]\n
+                             Content: {doc.combined_content}\n\n"""
+        for doc_nr, doc in enumerate(docs)
+    ]
+
+    docs_str = "\n\n".join(docs_format_list)
```
yuhongsun96, line 49: Looks like we use \n\n in a lot of places, should it be standardized?

backend/onyx/agents/agent_search/run_graph.py
```diff
@@ -0,0 +1,261 @@
...
+
+    # Agent search is not a Tool per se, but this is helpful for the frontend
+    yield ToolCallKickoff(
+        tool_name="agent_search_0",
+        tool_args={"query": config.search_request.query},
+    )
+    yield from run_graph(compiled_graph, config, input)
+
+
+# TODO: unify input types, especially prosearchconfig
```
yuhongsun96: Has this been done? We should also probably only pass in the subset of things we need

backend/onyx/agents/agent_search/basic/states.py
```diff
@@ -0,0 +1,42 @@
...
+# Config is for values that are set at the start and never change.
+# If you are using a value from the config and realize it needs to change,
+# you should add it to the state and use/update the version in the state.
+
+## Graph Input State
+
+
+class BasicInput(BaseModel):
+    # TODO: subclass global log update state
+    logs: str = ""
```
yuhongsun96: Should be a list of string, maybe with a default value of empty list

backend/onyx/agents/agent_search/run_graph.py
```diff
@@ -0,0 +1,261 @@
...
+from onyx.db.engine import get_session_context_manager
+from onyx.tools.tool_runner import ToolCallKickoff
+from onyx.utils.logger import setup_logger
+
+logger = setup_logger()
+
+_COMPILED_GRAPH: CompiledStateGraph | None = None
+
+
+def _set_combined_token_value(
```
yuhongsun96: Seems unused

backend/onyx/agents/agent_search/run_graph.py
```diff
@@ -0,0 +1,261 @@
...
+    combined_token: str, parsed_object: AgentAnswerPiece
+) -> AgentAnswerPiece:
+    parsed_object.answer_piece = combined_token
+
+    return parsed_object
+
+
+def _parse_agent_event(
+    event: StreamEvent,
```
yuhongsun96, line 42: Most of this function doesn't do anything, we cast it to a specific type but the return type is AnswerPacket

backend/onyx/agents/agent_search/run_graph.py
```diff
@@ -0,0 +1,261 @@
...
+    """
+
+    event_type = event["event"]
+
+    # We always just yield the event data, but this piece is useful for two development reasons:
+    # 1. It's a list of the names of every place we dispatch a custom event
+    # 2. We maintain the intended types yielded by each event
+    if event_type == "on_custom_event":
+        # TODO: different AnswerStream types for different events
+        if event["name"] == "decomp_qs":
```
yuhongsun96, line 55: Don't use strings like these, very easy to introduce bugs, have them be under some constants file and always shared throughout the codebase so that typos are caught before runtime

backend/onyx/agents/agent_search/deep_search_a/main__graph/nodes/generate_refined_answer.py
joachim-danswer: Certainly arbitrary. Idea is that the refined answer has access to more docs.

backend/onyx/agents/agent_search/orchestration/nodes/llm_tool_choice.py
```diff
@@ -0,0 +1,144 @@
...
+        tool_name, tool_args = (
+            force_use_tool.tool_name,
+            force_use_tool.args,
+        )
+        tool = get_tool_by_name(tools, tool_name)
+
+    # special pre-logic for non-tool calling LLM case
+    elif not using_tool_calling_llm and tools:
+        chosen_tool_and_args = get_tool_call_for_non_tool_calling_llm_impl(
+            force_use_tool=force_use_tool,
```
yuhongsun96, line 63: Why does this need to handle the force use situation? Is this just for reuseability?

backend/onyx/agents/agent_search/deep_search_a/main__graph/nodes/generate_refined_answer.py
joachim-danswer: we are not using get_section_relevance. So None is probably ok?

backend/onyx/agents/agent_search/orchestration/nodes/llm_tool_choice.py
```diff
@@ -0,0 +1,144 @@
...
+        structured_response_format=structured_response_format,
+    )
+
+    tool_message = process_llm_stream(
+        stream, should_stream_answer and not agent_config.skip_gen_ai_answer_generation
+    )
+
+    # If no tool calls are emitted by the LLM, we should not choose a tool
+    if len(tool_message.tool_calls) == 0:
+        logger.info("No tool calls emitted by LLM")
```
yuhongsun96: probably a debug level is preferred here

backend/onyx/agents/agent_search/orchestration/nodes/tool_call.py
```diff
@@ -0,0 +1,65 @@
...
+from onyx.tools.message import ToolCallSummary
+from onyx.tools.tool_runner import ToolRunner
+from onyx.utils.logger import setup_logger
+
+
+logger = setup_logger()
+
+
+def emit_packet(packet: AnswerPacket) -> None:
+    dispatch_custom_event("basic_response", packet)
```
yuhongsun96: Tool call should seemingly not emit a basic_response type packet but this could just be a naming issue if every time we call a tool we want to send a packet back to the frontend.

backend/onyx/agents/agent_search/deep_search_a/expanded_retrieval/edges.py
joachim-danswer: Done

backend/onyx/agents/agent_search/deep_search_a/answer_initial_sub_question/states.py
evan-onyx: atm we're avoiding the diamond problem by having shared keys uniquely at the bottom of the diamond, but will talk to Joachim about this idea

backend/onyx/server/query_and_chat/chat_backend.py
evan-onyx: huge amount of asyncio timeout errors

backend/onyx/agents/agent_search/run_graph.py
evan-onyx: The point is to show future people the expected type for each event, but we can throw this out if you don't like that

evan-danswer (approved):
Agent Search Beta ready!

PR #3758, opened 2025-01-23
Vespa scripts + Redis script update

pablonyx:
## Description

Fixes https://linear.app/danswer/issue/DAN-1330/improved-redis-debugging

Fixes https://linear.app/danswer/issue/DAN-1325/improved-vespa-debugging

Most useful options:
- Delete (ie. logout) / get user token from Redis
- Fetch ACLs from Vespa (huge time saver already!)

## How Has This Been Tested?

- All options, multi tenant + non multi tenant

### DIFF
--- a/backend/scripts/onyx_redis.py
+++ b/backend/scripts/onyx_redis.py
@@ -184,30 +184,35 @@ def get_user_token_from_redis(r: Redis, user_email: str) -> str | None:
     # Scan for keys matching the auth key prefix
     auth_keys = r.scan_iter(f"{REDIS_AUTH_KEY_PREFIX}*", count=SCAN_ITER_COUNT)
 
+    matching_key = None
+
     for key in auth_keys:
         key_str = key.decode("utf-8")
         jwt_token = r.get(key_str)
 
-        if jwt_token:
-            try:
-                if isinstance(jwt_token, bytes):
-                    jwt_token = jwt_token.decode("utf-8")
+        if not jwt_token:
+            continue
 
-                if jwt_token.startswith("b'") and jwt_token.endswith("'"):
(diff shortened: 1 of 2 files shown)


backend/scripts/onyx_redis.py
```diff
@@ -134,6 +174,88 @@ def purge_by_match_and_type(
...
+                    jwt_token = jwt_token.decode("utf-8")
+
+                if jwt_token.startswith("b'") and jwt_token.endswith("'"):
+                    jwt_token = jwt_token[2:-1]  # Remove b'' wrapper
+
+                jwt_data = json.loads(jwt_token)
+                if jwt_data.get("tenant_id") == tenant_id and str(
+                    jwt_data.get("sub")
+                ) == str(user_id):
+                    return key_str[len(REDIS_AUTH_KEY_PREFIX) :]
```
rkuo-danswer: This nesting is a bit difficult to read ... would probably be cleaner if we exit the loop with the data and then return at the end based on whether we have it or not.

backend/scripts/onyx_redis.py
```diff
@@ -134,6 +174,88 @@ def purge_by_match_and_type(
...
+                jwt_data = json.loads(jwt_token)
+                if jwt_data.get("tenant_id") == tenant_id and str(
+                    jwt_data.get("sub")
+                ) == str(user_id):
+                    if dry_run:
+                        logger.info(f"(DRY-RUN) Would delete token key: {key_str}")
+                    else:
+                        r.delete(key_str)
+                        logger.info(f"Deleted token for user: {user_email}")
+                    return True
```
rkuo-danswer: same comment as above.

rkuo-danswer (approved):

PR #3760, opened 2025-01-24
Agent Search top level tool choice

evan-danswer:
## Description

Changed the early logic for agent search to allow tool selection + early exit if the query doesn't need a search. Also causes the Force Search button to work. Detailed breakdown:

- Agent Search and Basic Search are still different graphs, but share a set of tool choice and calling nodes
- force search now works, since the tool choice node uses the ForceUseTool object
- updated unit tests to work with the stripped-down Answer class
- removed unused _return_contexts from Answer
- change ToolResponseSummary to use a v2 BaseModel for consistency and to allow it to be nested in other BaseModels
- edge case handling in agent_logging, especially to handle the case of the LLM deciding not to do a search and just spit out an answer
- promptConfig now used correctly in basic search graph
- BaseModel v2 instead of TypedDict for basic search graph (i.e. the nodes in the orchestration directory).

## How Has This Been Tested?

ran in the UI and the few tests that work; seems good


backend/onyx/agents/agent_search/orchestration/prepare_tool_input.py
```diff
@@ -0,0 +1,16 @@
...
+from langchain_core.runnables.config import RunnableConfig
+
+from onyx.agents.agent_search.models import AgentSearchConfig
+from onyx.agents.agent_search.orchestration.states import ToolChoiceInput
+
+
+def prepare_tool_input(state: Any, config: RunnableConfig) -> ToolChoiceInput:
+    agent_config = cast(AgentSearchConfig, config["metadata"]["config"])
+    return ToolChoiceInput(
+        should_stream_answer=True,
```
joachim-danswer: Should this be configurable? If later, please add comment.

backend/onyx/agents/agent_search/orchestration/prepare_tool_input.py
```diff
@@ -0,0 +1,16 @@
+from typing import Any
```
joachim-danswer: Create nodes directory please.

joachim-danswer (approved):
For the big commit, please keep a bullet list ion Description of the things that are either i) big with a short logic explanation, or ii) have broader impact, like removing a var that you think isn't used (at least people should be aware of it.)

PR #3768, opened 2025-01-24
Input shortcut fix in multi tenant case

pablonyx:
## Description

Fixes https://linear.app/danswer/issue/DAN-1335/input-prompts-not-showing-up

## How Has This Been Tested?

### DIFF
--- a/backend/onyx/db/input_prompt.py
+++ b/backend/onyx/db/input_prompt.py
@@ -193,13 +193,13 @@ def fetch_input_prompts_by_user(
     """
     Returns all prompts belonging to the user or public prompts,
     excluding those the user has specifically disabled.
+    Also, if `user_id` is None and AUTH_TYPE is DISABLED, then all prompts are returned.
     """
 
-    # Start with a basic query for InputPrompt
     query = select(InputPrompt)
 
-    # If we have a user, left join to InputPrompt__User so we can check "disabled"
     if user_id is not None:
+        # If we have a user, left join to InputPrompt__User to check "disabled"
         IPU = aliased(InputPrompt__User)
         query = query.join(
             IPU,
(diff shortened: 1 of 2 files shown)


backend/onyx/db/input_prompt.py
```diff
@@ -208,29 +208,33 @@ def fetch_input_prompts_by_user(
         )
 
         # Exclude disabled prompts
-        # i.e. keep only those where (IPU.disabled is NULL or False)
         query = query.where(or_(IPU.disabled.is_(None), IPU.disabled.is_(False)))
 
         if include_public:
-            # user-owned or public
+            # Return either user-owned or public prompts
```
rkuo-danswer: i think we mean both, not either?

backend/onyx/db/input_prompt.py
```diff
@@ -208,29 +208,33 @@ def fetch_input_prompts_by_user(
...
-    elif include_public:
-        query = query.where(InputPrompt.is_public)
+    else:
+        # user_id is None
+        if AUTH_TYPE == AuthType.DISABLED:
+            # If auth is disabled, return all prompts
+            query = query.where(True)  # type: ignore
+        elif include_public:
+            # Anonymous usage
+            query = query.where(InputPrompt.is_public)
```
rkuo-danswer, line 232: for a hanging if/else, it's good practice to handle the case explicitly with a comment and a pass so that the reader knows the case has been thought about.

backend/onyx/db/input_prompt.py
```diff
@@ -208,29 +208,33 @@ def fetch_input_prompts_by_user(
...
 
     if active is not None:
         query = query.where(InputPrompt.active == active)
 
+    print(query)
```
rkuo-danswer: debugging or intentional?

backend/onyx/server/features/input_prompt/api.py
```diff
@@ -32,6 +32,7 @@ def list_input_prompts(
     include_public: bool = True,
     db_session: Session = Depends(get_session),
 ) -> list[InputPromptSnapshot]:
+    print("FETCHING INPUT PROMPTS")
```
rkuo-danswer: debugging?

rkuo-danswer (approved):

backend/onyx/db/input_prompt.py
pablonyx: Definitely see how this could be confusing

PR #3776, opened 2025-01-25
Update slack configs

pablonyx:
## Description

Fixes https://linear.app/danswer/issue/DAN-1337/slack-setup-improvementsafety

## How Has This Been Tested?

https://github.com/user-attachments/assets/31b8df35-83d5-42f0-8b81-953ef8ee0049

---

https://github.com/user-attachments/assets/c38c39cf-ac56-4dc0-adf9-609723fff8ad

---

https://github.com/user-attachments/assets/72a9e42a-275d-449d-8e3a-734a88a0e464


Weves (commented):
Could we use the same style as assistant creation for create/update?

Screenshot 2025-01-26 at 12 49 09 PM

After seeing this, prefer having text thats says `Some incompatible document sets are hidden. View all.` where clicking View all adds the non-selectable document sets (right after the `Select the document sets OnyxBot will use while answering questions in Slack.`). This way we don't show a bunch of irrelevant / unusable document sets which clutter things up and instead only if they are missing something do they get the option to appease their curiosity.

Screenshot 2025-01-26 at 1 04 56 PM

Ideally same for assistants.

For the on-hover of document sets with auto-synced, I think it would be good to explain why e.g. `Unable to use document set “X” because it contains a connector with auto-sync permissions. OnyxBot’s responses in “#channel-name” are visible to all Slack users, so mirroring the asker’s permissions could inadvertently expose private information.`

Can we make the private document set thing just a warning that appears next to the create/update button + has a thing that appears on hover? We already have a screen jump on selecting an option, so the double screen jump is a bit much imo.

For the different options, can we make the descriptions describe what why/when the user would want to use this option? I feel like it's currently a bit repetitive to the title. E.g. for `Specific Document Sets` the description could read `Control which documents to use for answering questions` and for `Specific Assistant` it could read `Control both the documents and the prompt to use for answering questions`. For `Public Knowledge` we could use the description to clarify that it's not public, public (e.g. web—people have been confused about this before) wiht something like `Let OnyxBot respond based on information from all public connectors` (e.g. tie it in with the connectors concept they are hopefully familiar with). 

It seems like the `Give a "Still need help?" button` option is broken for me. It always appears checked even if I un-check it.

I shouldn't be able to choose the document sets option without selecting a document set (e.g. an error should pop up when clicking create taht says I need to select at least one document set).

Ideally, we should tell the user which document set(s) are private in the error message w/ links to them. That way they don't have to open up 1-10 tabs to figure out which one has a private connector. Also prefer saying connector (hopefully they know what that means since they see it on the `Existing Connectors` page regularly, but they may not know what a "private document" is).

Weves (commented):
Almost there! A few notes:

I believe `Some incompatible document sets are visible. View all document sets` is reversed from the way it should be. 

The standard answers section seems to be extra squished now. Should be at least as wide as `Answer Type`:
Screenshot 2025-01-27 at 11 32 19 AM

For consistency, would prefer if the popup appeared on hover on the warning icon + if it was yellow (e.g. matching the behavior of the "LLM doesn't support image input" experience). It feels like these types of things are usually a hover rather than a click + nothing is "broken", it's just a warning thus the yellow.

Seems like auto-synced assistants are just hidden? Would like the same behavior as for document sets where assistants w/ auto-sync are called out explicitly (I can already see the complaints of "why is my assistant not here??").

PR #3783, opened 2025-01-26
basic tools no search

evan-danswer:
## Description

Allowed basic search to run when the user is using an assistant without a search tool

## How Has This Been Tested?

Ran in UI


joachim-danswer (approved):
Seems pretty clear. I also assume

PR #3801, opened 2025-01-27
Double check all chat accessible dependencies

pablonyx:
## Description

Fixes https://linear.app/danswer/issue/DAN-1343/cycle-42-goals
Refer to: https://github.com/onyx-dot-app/onyx/pull/3674/files

## How Has This Been Tested?

On main and this branch, create private LLMs and access chat as anonymous user

### DIFF
--- a/backend/onyx/db/llm.py
+++ b/backend/onyx/db/llm.py
@@ -146,8 +146,10 @@ def fetch_existing_llm_providers_for_user(
                 ).all()
             )
         else:
+            # Use `fetch_existing_llm_providers` to fetch all existing providers
+            # As opposed to passing `user=None`
             raise ValueError(
-                "User is required for this operation when authentication is enabled. Please provide a valid user object."
+                "A valid user is required for this operation when authentication is enabled."
             )
 
     stmt = select(LLMProviderModel).distinct()
--- a/backend/onyx/db/llm.py
+++ b/backend/onyx/db/llm.py
@@ -135,15 +135,21 @@ def fetch_existing_llm_providers_for_user(
     db_session: Session,
     user: User | None = None,
 ) -> list[LLMProviderModel]:
-    if not user and AUTH_TYPE == AuthType.DISABLED:
-        # User is anonymous
-        return list(
-            db_session.scalars(
-                select(LLMProviderModel).where(
-                    LLMProviderModel.is_public == True  # noqa: E712
-                )
-            ).all()
-        )
+    if not user:
+        if AUTH_TYPE == AuthType.DISABLED:
+            # User is anonymous
+            return list(
+                db_session.scalars(
+                    select(LLMProviderModel).where(
+                        LLMProviderModel.is_public == True  # noqa: E712
+                    )
+                ).all()
+            )
+        else:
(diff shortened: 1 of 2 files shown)


backend/onyx/db/llm.py
```diff
@@ -124,10 +124,24 @@ def fetch_existing_tools(db_session: Session, tool_ids: list[int]) -> list[ToolM
...
+    return list(db_session.scalars(stmt).all())
+
+
+def fetch_existing_llm_providers_for_user(
+    db_session: Session,
     user: User | None = None,
 ) -> list[LLMProviderModel]:
     if not user:
-        return list(db_session.scalars(select(LLMProviderModel)).all())
+        # User is anonymous
```
Weves: should this also check if AUTH_TYPE == AuthType.DISABLED?

pablonyx: Yes!

backend/onyx/db/llm.py
```diff
@@ -124,10 +126,24 @@ def fetch_existing_tools(db_session: Session, tool_ids: list[int]) -> list[ToolM
...
+        return list(
+            db_session.scalars(
+                select(LLMProviderModel).where(
+                    LLMProviderModel.is_public == True  # noqa: E712
+                )
+            ).all()
+        )
     stmt = select(LLMProviderModel).distinct()
     user_groups_select = select(User__UserGroup.user_group_id).where(
         User__UserGroup.user_id == user.id
```
Weves, line 154: I believe `user` could be `None` here still. I think we need something like:

```
if not user:
  if AUTH_TYPE == AuthType.DISABLED:  # give access to all
  else:  # give access to public only, anonymous user case
```

backend/onyx/db/llm.py
```diff
@@ -124,10 +126,32 @@ def fetch_existing_tools(db_session: Session, tool_ids: list[int]) -> list[ToolM
...
+
+
+def fetch_existing_llm_providers_for_user(
+    db_session: Session,
     user: User | None = None,
 ) -> list[LLMProviderModel]:
     if not user:
-        return list(db_session.scalars(select(LLMProviderModel)).all())
+        if AUTH_TYPE == AuthType.DISABLED:
+            # User is anonymous
```
Weves, line 140: hmm, this doesn't look quite right 🧐 If auth is **_not_** disabled, then this means they are an anonymous user. If auth is disabled, we always pass around a null user, so they should have access to everything?

backend/onyx/db/llm.py
```diff
@@ -124,10 +126,32 @@ def fetch_existing_tools(db_session: Session, tool_ids: list[int]) -> list[ToolM
...
+            return list(
+                db_session.scalars(
+                    select(LLMProviderModel).where(
+                        LLMProviderModel.is_public == True  # noqa: E712
+                    )
+                ).all()
+            )
+        else:
+            # Use `fetch_existing_llm_providers` to fetch all existing providers
```
Weves: Doesn't this break auth disabled? 😅

Weves (approved):

PR #3802, opened 2025-01-27
Disable Slow Rerank + fix asyncio timeout

evan-danswer:
## Description

- misc code cleanup
- disable reranking when user is using a local model with no gpu
- bump timeout to keep is_connected live

## How Has This Been Tested?

Ran branch in the UI, ran unit tests. Only tested the "reranking still works for nonlocal model or local with gpu" case, so would be good to check that reranking *doesn't* happen in the other case. Worst case is just that we still rerank though, so not a lot of risk.


evan-danswer:
Old branch getting re-done in https://github.com/onyx-dot-app/onyx/pull/4011

PR #3834, opened 2025-01-29
Fixed SharePoint connector polling

hagen-danswer:
## Description
https://linear.app/danswer/issue/DAN-1363/sharepoint-polling-doesnt-work
- graphql api doesnt support filtering by last modified time
- We now pull all files and then only pull the content+convert to document if it is within the polling time

## How Has This Been Tested?

- added a test for this

### DIFF
--- a/backend/onyx/connectors/sharepoint/connector.py
+++ b/backend/onyx/connectors/sharepoint/connector.py
@@ -127,13 +127,6 @@ class SharepointConnector(LoadConnector, PollConnector):
         start: datetime | None = None,
         end: datetime | None = None,
     ) -> list[tuple[DriveItem, str]]:
-        filter_str = ""
-        if start is not None and end is not None:
-            filter_str = (
-                f"last_modified_datetime ge {start.isoformat()} and "
-                f"last_modified_datetime le {end.isoformat()}"
-            )
-
         final_driveitems: list[tuple[DriveItem, str]] = []
         try:
             site = self.graph_client.sites.get_by_url(site_descriptor.url)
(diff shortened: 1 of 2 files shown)


backend/onyx/connectors/sharepoint/connector.py
```diff
@@ -196,19 +186,33 @@ def _fetch_driveitems(
...
+                            <= item.last_modified_datetime.replace(tzinfo=timezone.utc)
+                            <= end
+                        ]
+                        logger.debug(
+                            f"Found {len(driveitems)} items within time window in drive '{drive.name}'"
                         )
+
+                    # Now fetch content only for items that passed all filters
+                    for item in driveitems:
+                        item.get_content().execute_query()
```
Weves: why do we need to add this now?

hagen-danswer: good catch, fixed

backend/onyx/connectors/sharepoint/connector.py
```diff
@@ -196,19 +191,32 @@ def _fetch_driveitems(
...
+                            for item in driveitems
+                            if start
+                            <= item.last_modified_datetime.replace(tzinfo=timezone.utc)
+                            <= end
+                        ]
+                        logger.debug(
+                            f"Found {len(driveitems)} items within time window in drive '{drive.name}'"
                         )
+
+                    # Now fetch content only for items that passed all filters
```
Weves: nit: this comment feels a bit misleading. We aren't fetching the content now (generally when I see a "Now do X" I expect to see X immediately following)

Weves (approved):

PR #3837, opened 2025-01-29
Add more sync records + fix small bug in monitoring task causing dele…

Weves:
…tion metrics to never be emitted

https://linear.app/danswer/issue/DAN-1367/add-more-metrics-to-background-monitoring

## How Has This Been Tested?

Integration tests + testing each effected flow. 

### DIFF
--- a/backend/onyx/background/celery/tasks/doc_permission_syncing/tasks.py
+++ b/backend/onyx/background/celery/tasks/doc_permission_syncing/tasks.py
@@ -11,6 +11,7 @@ from celery import Task
 from celery.exceptions import SoftTimeLimitExceeded
 from redis import Redis
 from redis.lock import Lock as RedisLock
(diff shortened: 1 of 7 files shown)


backend/onyx/background/celery/tasks/doc_permission_syncing/tasks.py
```diff
@@ -185,6 +194,15 @@ def try_creating_permissions_sync_task(
             priority=OnyxCeleryPriority.HIGH,
         )
 
+        # create before setting fence to avoid race condition where the monitoring
```
rkuo-danswer, line 186: this is going to need to move once the PR to fix fence creation is in

rkuo-danswer (approved):
looks ok, one comment on moving one insert after the fence fix is in.

PR #3844, opened 2025-01-30
Ensure sidepanel defaults sidebar off

pablonyx:
## Description
Respect searchparam communicated by chrome extension

Fixes https://linear.app/danswer/issue/DAN-1372/sidebar-is-default-open-for-our-chrome-extension

## How Has This Been Tested?

- Locally with chrome extension https://github.com/onyx-dot-app/onyx-chrome-extension


web/src/app/chat/WrappedChat.tsx
```diff
@@ -5,14 +5,18 @@ import FunctionalWrapper from "./shared_chat_search/FunctionalWrapper";
...
+  // This is required for the chrome extension side panel
+  // we don't want to show the sidebar by default when the user opens the side panel
+  defaultSidebarOff?: boolean;
 }) {
   const { toggledSidebar } = useChatContext();
 
   return (
     <FunctionalWrapper
-      initiallyToggled={toggledSidebar}
+      initiallyToggled={toggledSidebar && !defaultSidebarOff}
```
rkuo-danswer: finding this a little hard to read ... i don't think toggled implies a particular state.  is it possible to use a word like "displayed"?

also, small thing, but naming a boolean as "off" adds an extra level of negation and can make things harder to reason through.

when possible, I prefer to avoid this.  But in user level settings where the off can be intuitive because there's an strong implied default, I would consider transforming the boolean internally to a non-"off" version.

rkuo-danswer (approved):

Issue #3851, opened 2025-01-30
Jira Server/Data Center authentication with token not working


julienfoenet:
With my team, we are testing the integration with Jira and the indexing of Jira issues from a specific Jira project. We have Jira hosted on our own (Jira Data Center, not Jira Cloud). Cloud Jira requires to authenticate using an email + password, while self-hosted Jira requires only a Personal Access Token. When we try to create the connector and indexing using our token, we see an error 401 Unauthorized, it seems that the wrong authentication method is used connect to Jira API. 
This is due to the fact that in the [connector code](https://github.com/onyx-dot-app/onyx/blob/74a472ece7f016b4ced8ce42befb43cf16a90ce5/backend/onyx/connectors/onyx_jira/connector.py#L279) when we load the credentials, we populate the creds dict with two keys "jira_user_email" and "jira_api_token", and right after in method [build_jira_client](https://github.com/onyx-dot-app/onyx/blob/787e25cd78b722f550a647c9f9d577682ed5ef42/backend/onyx/connectors/onyx_jira/utils.py#L71), we do a switch on the presence of the key "jira_user_email" (which is always there) to decide what authent method to use.
I think the switch should be made instead on whether user_email is provided by the user or not.

PR link:
https://github.com/onyx-dot-app/onyx/pull/3852

PR #3852, opened 2025-01-30
Fix switch authent method in build_jira_client

julienfoenet:
## Description

This PR fixes an issue in the authentication method when building Jira client

See issue [3851](https://github.com/onyx-dot-app/onyx/issues/3851)

## How Has This Been Tested?

I have run the stack locally with / without the code change to make sure this fixes the issue

### DIFF
--- a/backend/onyx/connectors/onyx_jira/utils.py
+++ b/backend/onyx/connectors/onyx_jira/utils.py
@@ -70,11 +70,11 @@ def build_jira_url(jira_client: JIRA, issue_key: str) -> str:
 
 def build_jira_client(credentials: dict[str, Any], jira_base: str) -> JIRA:
     api_token = credentials["jira_api_token"]
+    user_email = credentials["jira_user_email"]
     # if user provide an email we assume it's cloud
-    if "jira_user_email" in credentials:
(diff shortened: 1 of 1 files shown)


backend/onyx/connectors/onyx_jira/utils.py
pmontanari, line 1: If there is no jira_user_email in the credentials dict, it will simply fail and raise an exception.
user_email = credentials.get("jira_user_email") would receive None if not set

backend/onyx/connectors/onyx_jira/utils.py
```diff
@@ -70,11 +70,11 @@ def build_jira_url(jira_client: JIRA, issue_key: str) -> str:
 
 def build_jira_client(credentials: dict[str, Any], jira_base: str) -> JIRA:
     api_token = credentials["jira_api_token"]
+    user_email = credentials["jira_user_email"]
```
pmontanari, line 73: If there is no jira_user_email in the credentials dict, it will simply fail and raise an exception.
user_email = credentials.get("jira_user_email") would receive None if not set.
Anyway, this does not change from previous behaviour.

PR #3862, opened 2025-01-31
Add more airtable logging

Weves:
## Description

Tries to make Airtable faster

https://linear.app/danswer/issue/DAN-1384/airtable-indexing-extremely-slow

## How Has This Been Tested?

Ran locally, it was faster.

### DIFF
--- a/backend/onyx/connectors/connector_runner.py
+++ b/backend/onyx/connectors/connector_runner.py
@@ -1,4 +1,5 @@
 import sys
+import time
 from datetime import datetime
 
 from onyx.connectors.interfaces import BaseConnector
@@ -45,7 +46,17 @@ class ConnectorRunner:
     def run(self) -> GenerateDocumentsOutput:
         """Adds additional exception logging to the connector."""
         try:
-            yield from self.doc_batch_generator
+            start = time.monotonic()
+            for batch in self.doc_batch_generator:
+                # to know how long connector is taking
+                end = time.monotonic()
+                logger.debug(
+                    f"Connector tool in {end - start} seconds to build a batch."
+                )
+                start = end
+
+                yield batch
+
         except Exception:
             exc_type, _, exc_traceback = sys.exc_info()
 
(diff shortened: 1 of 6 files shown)


backend/onyx/natural_language_processing/search_nlp_models.py
```diff
@@ -256,7 +296,7 @@ def from_db_model(
         )
 
 
-class RerankingModel:
+class RerankingModel:  #
```
rkuo-danswer: empty comment?

backend/onyx/indexing/indexing_pipeline.py
```diff
@@ -380,6 +380,15 @@ def index_doc_batch(
...
 
+    doc_descriptors = [
+        {
+            "doc_id": doc.id,
+            "doc_length": doc.get_total_char_length(),
+        }
+        for doc in ctx.updatable_docs
+    ]
+    logger.debug(f"Starting indexing process for documents: {doc_descriptors}")
+
```
rkuo-danswer, line 391: Does this make the line below ... "starting chunking" ... irrelevant?

backend/onyx/connectors/connector_runner.py
```diff
@@ -45,7 +46,17 @@ def __init__(
     def run(self) -> GenerateDocumentsOutput:
         """Adds additional exception logging to the connector."""
         try:
-            yield from self.doc_batch_generator
+            start = time.monotonic()
```
rkuo-danswer, line 49: in the past i've seen that items yielded from generators in a loop are lazily evaluated (aka not processed until first accessed in the loop.  Just wanted to double check that the timings printed here are indeed working correctly.

rkuo-danswer (approved):
comments inline

backend/onyx/indexing/indexing_pipeline.py
Weves: imo its fine to keep it in (lets you know that chunking is the first step, although no strong opinion)

backend/onyx/natural_language_processing/search_nlp_models.py
Weves: removed!

backend/onyx/connectors/connector_runner.py
Weves: it does work, not sure exactly the standard behavior here though

PR #3867, opened 2025-01-31
Add user specific chat session temperature

pablonyx:
## Description

Fixes https://linear.app/danswer/issue/DAN-1387/user-specific-temperature-overrides

Logic:
- For existing chats: If user has overridden temperature for a given chat session, we persist that when a user comes back (including if this is set before sending first message in chat)
- For new chats: if search tool is available, default to 0, otherwise 0.5

https://github.com/user-attachments/assets/f5a82fc0-99a1-4d01-87c2-3efdcfe32069

Noticed during building: likely some centralization of the CRUD operations for the user settings that can occur. All essentially just flipping a bit (or reordering a list) for a non-auth (key-value store) or authed user

## How Has This Been Tested?
- Locally with auth and noauth

### DIFF
--- a/backend/onyx/llm/chat_llm.py
+++ b/backend/onyx/llm/chat_llm.py
@@ -387,10 +387,6 @@ class DefaultMultiLLM(LLM):
         self._record_call(processed_prompt)
 
         try:
-            logger.warning(
-                f"Model: {self.config.model_provider}/{self.config.deployment_name or self.config.model_name}"
-            )
-            logger.warning(f"Temperature: {self._temperature}")
             return litellm.completion(
                 mock_response=MOCK_LLM_RESPONSE,
                 # model choice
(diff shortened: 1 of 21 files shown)


backend/onyx/llm/chat_llm.py
```diff
@@ -386,6 +386,10 @@ def _completion(
         self._record_call(processed_prompt)
 
         try:
+            logger.warning(
+                f"Model: {self.config.model_provider}/{self.config.deployment_name or self.config.model_name}"
+            )
+            logger.warning(f"Temperature: {self._temperature}")
```
Weves: probably should remove

backend/onyx/natural_language_processing/search_nlp_models.py
```diff
@@ -175,7 +175,19 @@ def process_batch(
                 if self.callback.should_stop():
                     raise RuntimeError("_batch_encode_texts detected stop signal")
 
+            total_chars = sum(len(text) for text in text_batch)
```
Weves: is this supposed to be in here?

Weves (approved):
left a comment small comments, but otherwise lgtm!

PR #3869, opened 2025-01-31
E2e assistant tests

pablonyx:
## Description

Fixes https://linear.app/danswer/issue/DAN-1389/e2e-tests-for-ordering-llm-assistant-flows

Tests for 
- Assistant ordering
- Which assistant should be active?
- Which model should be selected?

Further details for each e2e test within the test code

## How Has This Been Tested?


backend/onyx/configs/app_configs.py
```diff
@@ -611,3 +611,8 @@
 DEV_MODE = os.environ.get("DEV_MODE", "").lower() == "true"
 
 TEST_ENV = os.environ.get("TEST_ENV", "").lower() == "true"
+
+# Set to true to mock LLM responses for testing purposes
+MOCK_LLM_RESPONSE = (
```
rkuo-danswer, line 616: i suppose this technically works but making the get twice seems weird

rkuo-danswer (approved):
one small comment, also, the tests themselves i feel could use just an extra level of detail describing the intended user / UI behavior in english.

PR #3871, opened 2025-02-01
Strict Tenant ID Enforcement

pablonyx:
## Description

fixes https://linear.app/danswer/issue/DAN-1390/clean-up-of-tenant-id-assumption

1. New Default behavior / assumptions
   - When a request comes in, the HTTP middleware sets `CURRENT_TENANT_ID_CONTEXTVAR` to the correct tenant ID.  
   - In Celery tasks, we override the Task class to be `TenantAwareTask` and you pass `tenant_id=...` in `kwargs`, the task automatically sets the contextvar before running.  
   - This means that, most of the time, you do not need to pass `tenant_id` around yourself. Just call:
     ```python
         ...
     ```
     or
     ```python
     r = get_redis_client()
     ```
     and it will pick up the ID from the contextvar.

2. Shared utils
   - For truly global logic not tied to any tenant, use these dedicated helpers:
     - `get_session_with_shared_schema()`: forces usage of the “public” schema in Postgres.  
     - `get_shared_redis_client()` / `get_shared_redis_replica_client()`: skip per-tenant prefixes in Redis.  
     - `get_shared_kv_store()`: global K/V store that ignores the per-tenant context.  

3. Overriding Tenant ID
   - If you’re doing multi-tenant “manager” work, you can explicitly pass a tenant ID:
     ```python
     with get_session_with_current_tenant("some_tenant") as db_session:
         ...
     ```
   - This directly sets the contextvar to `some_tenant` for the duration of that block.

4.  BG processes
   - Celery uses `TenantAwareTask`:  
     ```python
     @shared_task(base=TenantAwareTask)
     def my_task(*, tenant_id: str):
         # tenant_id is picked up and set into the contextvar automatically
         ...
     ```
   - When you call `my_task.delay(tenant_id="...")`, Celery sets that ID in the contextvar before running the task body.

## How Has This Been Tested?

### DIFF
--- a/backend/ee/onyx/background/celery/apps/primary.py
+++ b/backend/ee/onyx/background/celery/apps/primary.py
@@ -5,7 +5,7 @@ from onyx.background.celery.apps.primary import celery_app
 from onyx.background.task_utils import build_celery_task_wrapper
 from onyx.configs.app_configs import JOB_TIMEOUT
 from onyx.db.chat import delete_chat_sessions_older_than
-from onyx.db.engine import get_session_with_current_tenant
+from onyx.db.engine import get_session_with_tenant
 from onyx.server.settings.store import load_settings
 from onyx.utils.logger import setup_logger
 from shared_configs.configs import MULTI_TENANT
@@ -21,7 +21,7 @@ logger = setup_logger()
 def perform_ttl_management_task(
     retention_limit_days: int, *, tenant_id: str | None
 ) -> None:
-    with get_session_with_current_tenant(tenant_id) as db_session:
+    with get_session_with_tenant(tenant_id) as db_session:
         delete_chat_sessions_older_than(retention_limit_days, db_session)
 
 
@@ -44,7 +44,7 @@ def check_ttl_management_task(*, tenant_id: str | None) -> None:
 
     settings = load_settings()
     retention_limit_days = settings.maximum_chat_retention_days
-    with get_session_with_current_tenant(tenant_id) as db_session:
+    with get_session_with_tenant(tenant_id) as db_session:
         if should_perform_chat_ttl_check(retention_limit_days, db_session):
             perform_ttl_management_task.apply_async(
                 kwargs=dict(
@@ -62,7 +62,7 @@ def check_ttl_management_task(*, tenant_id: str | None) -> None:
 )
 def autogenerate_usage_report_task(*, tenant_id: str | None) -> None:
     """This generates usage report under the /admin/generate-usage/report endpoint"""
-    with get_session_with_current_tenant(tenant_id) as db_session:
+    with get_session_with_tenant(tenant_id) as db_session:
         create_new_usage_report(
             db_session=db_session,
             user_id=None,
--- a/backend/ee/onyx/background/celery/apps/primary.py
+++ b/backend/ee/onyx/background/celery/apps/primary.py
@@ -21,7 +21,7 @@ logger = setup_logger()
 def perform_ttl_management_task(
     retention_limit_days: int, *, tenant_id: str | None
 ) -> None:
-    with get_session_with_tenant(tenant_id) as db_session:
+    with get_session_with_tenant(tenant_id=tenant_id) as db_session:
         delete_chat_sessions_older_than(retention_limit_days, db_session)
 
 
@@ -44,7 +44,7 @@ def check_ttl_management_task(*, tenant_id: str | None) -> None:
 
     settings = load_settings()
     retention_limit_days = settings.maximum_chat_retention_days
-    with get_session_with_tenant(tenant_id) as db_session:
+    with get_session_with_tenant(tenant_id=tenant_id) as db_session:
         if should_perform_chat_ttl_check(retention_limit_days, db_session):
             perform_ttl_management_task.apply_async(
                 kwargs=dict(
@@ -62,7 +62,7 @@ def check_ttl_management_task(*, tenant_id: str | None) -> None:
 )
 def autogenerate_usage_report_task(*, tenant_id: str | None) -> None:
     """This generates usage report under the /admin/generate-usage/report endpoint"""
-    with get_session_with_tenant(tenant_id) as db_session:
+    with get_session_with_tenant(tenant_id=tenant_id) as db_session:
         create_new_usage_report(
             db_session=db_session,
             user_id=None,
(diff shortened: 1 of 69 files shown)


backend/shared_configs/contextvars.py
```diff
@@ -1,17 +1,21 @@
...
+] = contextvars.ContextVar(
+    "current_tenant_id", default=None if MULTI_TENANT else POSTGRES_DEFAULT_SCHEMA
 )
 
 
 """Utils related to contextvars"""
 
 
-def get_current_tenant_id() -> str:
+def current_tenant_id(strict: bool = True) -> str:
```
Weves: nit: prefer `get_current_tenant_id` for consistency. We usually have a `get` / `fetch` prefix (e.g. make them verbs) for functions. For example `get_llm` / `get_session` / `get_redis_client`.

backend/shared_configs/contextvars.py
```diff
@@ -1,17 +1,21 @@
...
 from shared_configs.configs import POSTGRES_DEFAULT_SCHEMA
 
+
 # Context variable for the current tenant id
-CURRENT_TENANT_ID_CONTEXTVAR = contextvars.ContextVar(
-    "current_tenant_id", default=POSTGRES_DEFAULT_SCHEMA
+CURRENT_TENANT_ID_CONTEXTVAR: contextvars.ContextVar[
+    str | None
+] = contextvars.ContextVar(
+    "current_tenant_id", default=None if MULTI_TENANT else POSTGRES_DEFAULT_SCHEMA
```
Weves, line 11: thoughts on this always being `None` and enforcing that it be set on both multi and single tenant to bring behavior closer together? E.g. middleware always sets this to `POSTGRES_DEFAULT_SCHEMA`.

Not sure if this is better, but just a thought.

backend/shared_configs/contextvars.py
Weves: also is there ever a reason to pass in `strict=False`?

Weves (commented):
As discussed, let's have a separate function for accessing a Postgres / Redis session with the public tenant to not overload the `get_session_with_tenant` etc. Let's also remove the `tenant_id` override option to enforce consistency.

backend/shared_configs/contextvars.py
pablonyx: Definitely agree in the abstract - didn't set this originally since we have a dependency called `get_current_tenant_id`, but with this refactory I think it makes sense to remove that function anyways.

backend/ee/onyx/background/celery/apps/primary.py
```diff
@@ -21,7 +21,7 @@
 def perform_ttl_management_task(
     retention_limit_days: int, *, tenant_id: str | None
 ) -> None:
-    with get_session_with_tenant(tenant_id) as db_session:
+    with get_session_with_current_tenant(tenant_id) as db_session:
```
rkuo-danswer: i don't believe saying "current tenant" has any semantic meaning to the function name when we're passing in a tenant id.  if the function takes a tenant id, it doesn't care where the tenant id came from.

backend/ee/onyx/server/middleware/tenant_tracking.py
```diff
@@ -33,7 +33,7 @@ async def set_tenant_id(
             return await call_next(request)
 
         except Exception as e:
-            logger.error(f"Error in tenant ID middleware: {str(e)}")
+            logger.exception(f"Error in tenant ID middleware: {str(e)}")
```
rkuo-danswer, line 36: str(e) should be redundant if using logger.exception.

backend/onyx/background/celery/apps/beat.py
```diff
@@ -19,6 +19,7 @@
 
 celery_app = Celery(__name__)
 celery_app.config_from_object("onyx.background.celery.configs.beat")
+celery_app.Task = app_base.TenantAwareTask  # type: ignore [misc]
```
rkuo-danswer: Monkey patching Task seems too invasive as there are already built in ways to subclass Task behavior.  Wouldn't something like https://docs.celeryq.dev/en/stable/userguide/tasks.html#task-inheritance accomplish this just fine?

backend/onyx/background/celery/tasks/indexing/tasks.py
```diff
@@ -68,12 +69,15 @@
 def check_for_indexing(self: Task, *, tenant_id: str | None) -> int | None:
     """a lightweight task used to kick off indexing tasks.
     Occcasionally does some validation of existing state to clear up error conditions"""
+    print(f"TENANT ID, {tenant_id}")
```
rkuo-danswer, line 72: debug logging?

backend/onyx/db/engine.py
```diff
@@ -405,82 +371,83 @@ async def get_async_session_with_tenant(
...
+                    text(
+                        f"SET SESSION idle_in_transaction_session_timeout = {POSTGRES_IDLE_SESSIONS_TIMEOUT}"
                     )
+                )
+        finally:
+            cursor.close()
+
+        with Session(bind=connection, expire_on_commit=False) as session:
+            try:
+                yield session
```
rkuo-danswer, line 449: i've seen some infailedsqltransaction errors involving these search_path operations and i'm wondering if we might need a rollback somewhere.  Don't have an exact direction to go on this but might be worth investigating.

backend/onyx/setup.py
```diff
@@ -251,8 +251,9 @@ def setup_vespa(
 
             logger.notice("Vespa setup complete.")
             return True
-        except Exception:
-            logger.notice(
+        except Exception as e:
+            logger.error(f"Vespa setup failed: {e}")
```
rkuo-danswer: seems redundant with the exception logging that follows.

backend/onyx/setup.py
```diff
@@ -364,7 +365,8 @@ def setup_vespa_multitenant(supported_indices: list[SupportedEmbeddingModel]) ->
 
             logger.notice("Vespa setup complete.")
             return True
-        except Exception:
+        except Exception as e:
+            logger.exception(f"Vespa setup failed: {e}")
```
rkuo-danswer: maybe follow the same pattern as the previous exception logging

backend/ee/onyx/background/celery/apps/primary.py
pablonyx: Agreed- replaced all

backend/onyx/background/celery/apps/beat.py
pablonyx: Certainly would but I believe we'd have to specify the app base for each task we define? Not sure if we want to have to remember this for each task we add.
Couldn't see any other way to override for all tasks, but I may be missing something.

rkuo-danswer: Yes, we would have to specify it for each task that is tenant aware ... that is fine. We should do this as it's idiomatic/standard to override the tasks in this way, and patching the task class is really not a very standard approach to things and is very hidden if you don't know it's happening.

rkuo-danswer (approved):

PR #3873, opened 2025-02-01
Fix Connector hover state goes off the page and file connector button…

jenny-onyx:
… issue

## Description

[[Provide a brief description of the changes in this PR]](https://linear.app/danswer/issue/DAN-1394/connector-hover-state-goes-off-the-page)

## How Has This Been Tested?

Tested in the UI


yuhongsun96:
image
Deletion of File connector would be blocked by this change as now I can't have it in Paused state anymore.

yuhongsun96:
Looks and works correctly, nice job!

Preference for logic here though that it should be like before (centered above the button) if it fits within the screen, and only change it if it goes off the screen to the right
image

Issue #3888, opened 2025-02-04
Can't upload images


sam-w:
See also https://github.com/onyx-dot-app/onyx/issues/3883 (although that manifests differently to my issue).

- I'm running `v0.21.0-beta.1`, self-hosted
- I can reproduce this issue with any tag which includes 2ad86aa9a68ccfe6cbd7d7bb6420bbf018b82382
- I can fix the issue by manually reverting the changes related to [this new file `.read()`](https://github.com/onyx-dot-app/onyx/commit/2ad86aa9a68ccfe6cbd7d7bb6420bbf018b82382#diff-5b691e234e5dfdba61a87ea8a76cbb842a51b7aaf8bc837f5a2d40bd376cdb99R675)
- Will raise a PR

<img width="418" alt="Image" src="https://github.com/user-attachments/assets/5d6f3e32-4468-45ad-9160-91d54b73b068" />
<img width="797" alt="Image" src="https://github.com/user-attachments/assets/71805741-cd58-467b-8f5e-745e2fc6c0a5" />

PR #3889, opened 2025-02-04
Fix chat image upload double read

sam-w:
## Description

Fixes https://github.com/onyx-dot-app/onyx/issues/3888

There is a bug which breaks image uploads in chats:

Screenshot 2025-02-04 at 11 21 20 am
Screenshot 2025-02-04 at 11 21 08 am

I believe this was caused by a change in 2ad86aa which causes an IO stream based on an image file to be read again without resetting the stream back to the beginning (first read in chat_backend.py, [upload_files_for_chat](https://github.com/onyx-dot-app/onyx/blob/715359c12060957bb37b7df545359b4b444d8ab6/backend/onyx/server/query_and_chat/chat_backend.py#L716), and then again in pg_file_store.py, [create_populate_lobj](https://github.com/onyx-dot-app/onyx/blob/715359c12060957bb37b7df545359b4b444d8ab6/backend/onyx/db/pg_file_store.py#L53)). Non image files are treated differently and thus are unaffected.

## How Has This Been Tested?

Tested both PDF and image uploading.

### DIFF
--- a/backend/onyx/server/query_and_chat/chat_backend.py
+++ b/backend/onyx/server/query_and_chat/chat_backend.py
@@ -713,8 +713,6 @@ def upload_files_for_chat(
             else ChatFileType.PLAIN_TEXT
         )
 
-        file_content = file.file.read()  # Read the file content
-
         if file_type == ChatFileType.IMAGE:
             file_content_io = file.file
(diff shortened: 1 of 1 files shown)


Weves (approved):

Weves:
Thanks for this fix!

PR #3892, opened 2025-02-04
Enforce Slack Channel Default Config

pablonyx:
## Description

https://github.com/user-attachments/assets/e34e8266-0e54-47e9-830e-e0bc83e5be3f

Each slack bot now has a default configuration that is applied. This accomplishes the following

a) Get rid of ambiguity in the implicit current Slack channel config
b) Ensure admins can configure the DM configuration

How it does this:
- Enforce a default Slack channel config for each slackbot
- Create default for existing Slackbots
- Add relevant UX components for modifying default Slack channel configuration (+ incorporate https://github.com/onyx-dot-app/onyx/pull/3721. thank you [ktjayamanna](https://github.com/ktjayamanna)) 

Fixes https://linear.app/danswer/issue/DAN-1406/default-slack-channel-configs
## How Has This Been Tested?
 
- [x] Upgrade from main with existing Slackbot + test default configuration for the channel  (in DMs + channels)
- [x] Create Slackbot and test / modify default + ensure correct behavior observed
- [x] Downgrade and test Slackbot behavior


yuhongsun96:
image
The default assistant is not a valid option here, it would not allow the use case of interest.

Also let's not use the blue pill thing for a header, we don't have this type of design pattern anywhere in the app so let's not introduce new patterns.

backend/onyx/db/constants.py
```diff
@@ -1 +1,2 @@
 SLACK_BOT_PERSONA_PREFIX = "__slack_bot_persona__"
+DEFAULT_PERSONA_SLACKCHANNEL_NAME = "DEFAULT"
```
rkuo-danswer: SLACK_CHANNEL maybe?

backend/onyx/db/slack_bot.py
```diff
@@ -74,3 +74,15 @@ def remove_slack_bot(
 
 def fetch_slack_bots(db_session: Session) -> Sequence[SlackBot]:
     return db_session.scalars(select(SlackBot)).all()
+
+
+def fetch_slack_bot_tokens(
+    db_session: Session, slack_bot_id: int
+) -> dict[str, str] | None:
+    slack_bot = db_session.scalar(select(SlackBot).where(SlackBot.id == slack_bot_id))
+    if slack_bot:
```
rkuo-danswer: prefer to return on errors first, then finally return the result/happy path at the end of the function

web/src/components/ui/badge.tsx
```diff
@@ -10,11 +10,13 @@ const badgeVariants = cva(
...
           "border-neutral-200 bg-neutral-100 text-neutral-600 hover:bg-neutral-200",
-        agent: "border-agent bg-agent text-white hover:bg-agent-hover",
+        agent:
+          "border-orange-200 bg-orange-50 text-orange-600 hover:bg-orange-75 dark:bg-orange-900 dark:text-neutral-50 dark:hover:bg-orange-850",
+
         canceled:
           "border-gray-200 bg-gray-50 text-gray-600 hover:bg-gray-75 dark:bg-gray-900 dark:text-neutral-50 dark:hover:bg-gray-850",
         orange:
-          "border-orange-200 bg-orange-50 text-orange-600 hover:bg-orange-75 dark:bg-orange-900 dark:text-neutral-50 dark:hover:bg-orange-850",
+          "border-orange-200 bg-orange-50 text-orange-600 h`over:bg-orange-75 dark:bg-orange-900 dark:text-neutral-50 dark:hover:bg-orange-850",
```
rkuo-danswer: typo here

backend/onyx/server/manage/slack_bot.py
```diff
@@ -315,3 +337,31 @@ def list_bot_configs(
...
+    if not tokens or "bot_token" not in tokens:
+        raise HTTPException(
+            status_code=404, detail="Bot token not found for the given bot ID"
+        )
+
+    bot_token = tokens["bot_token"]
+    client = WebClient(token=bot_token)
+
+    try:
+        response = client.conversations_list(limit=1000)
```
rkuo-danswer: seems like an odd way to get channels.  There isn't a better way to do this?

backend/onyx/server/manage/slack_bot.py
```diff
@@ -315,3 +337,31 @@ def list_bot_configs(
...
+        channels = {channel["name"]: channel["id"] for channel in response["channels"]}
+        return channels
```
rkuo-danswer, line 381: nit: would put the return outside the try clause ... can't really exception, looks more like a normal function, etc.

rkuo-danswer (approved):

backend/onyx/server/manage/slack_bot.py
pablonyx: Probably! Just threw the community contribution in here since it was reviewed previously. But taking a look at it now, agree with this and the below one as ewll

backend/onyx/server/manage/slack_bot.py
pablonyx: Same as above

PR #3895, opened 2025-02-04
Password reset tenant

pablonyx:
## Description

FastAPI users doesn't exactly play nice with our version of mult-tenancy, which means we need to occasionally employ some workarounds like here.

## How Has This Been Tested?

### DIFF
--- a/backend/onyx/configs/constants.py
+++ b/backend/onyx/configs/constants.py
@@ -15,6 +15,10 @@ ID_SEPARATOR = ":;:"
 DEFAULT_BOOST = 0
 SESSION_KEY = "session"
 
+# Cookies
+FASTAPI_USERS_AUTH_COOKIE_NAME = "fastapiusersauth"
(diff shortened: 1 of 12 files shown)


backend/onyx/configs/constants.py
```diff
@@ -15,6 +15,10 @@
 DEFAULT_BOOST = 0
 SESSION_KEY = "session"
 
+# Cookies
+FASTAPI_USERS_AUTH_COOKIE_NAME = "fastapiusersauth"
```
Weves: nit: can we pass this as the `cookie_name` in the `CookieTransport` initialization in users.py so that we can change this and things will still work

Weves (approved):

backend/onyx/configs/constants.py
pablonyx: Sounds good!

Issue #3921, opened 2025-02-06
Gmail connector fails to index on malformed date strings


sam-w:
Email is not great - it's pretty common to see malformed date strings of the form `<datetime> +<tz offset> (<invalid tz name>)` in the Date header field, as returned by the Gmail API.

Some concrete examples from our Gmail: 
- `Thu, 23 Jan 2025 11:04:48 +0300 (+03)` (DMARC report)
- `Sat, 25 Jan 2025 05:15:30 +1100 (AUSNSW)` (automated invoice from a partner's legacy system)
- `Mon, 10 Feb 2025 02:22:15 +1100 (AUS Eastern Daylight Time)` (automated notification from Australian Tax office)

These cause Gmail connector indexing to fail on [this line](https://github.com/onyx-dot-app/onyx/blob/5c7487e91f38a850bd386cf2b6769aa9ebb82951/backend/onyx/connectors/gmail/connector.py#L182). `dateutil.parse` doesn't know how to parse those strings because it's expecting the timezone name (in parentheses) to be < 5 characters and composed only of uppercase ASCII.

The gmail connector needs some way to handle these almost-ok-but-still-invalid date strings.

PR #3922, opened 2025-02-06
Add special case to `time_str_to_utc` to handle invalid tz names

sam-w:
## Description

See https://github.com/onyx-dot-app/onyx/issues/3921

tl;dr: malformed date strings are being returned from the Gmail API and being passed to the `time_str_to_utc` function here: https://github.com/onyx-dot-app/onyx/blob/5c7487e91f38a850bd386cf2b6769aa9ebb82951/backend/onyx/connectors/gmail/connector.py#L182

This is the least obtrusive solution I could come up with:
- if a datestring doesn't parse, see if it contains an offset followed by something in parens (`+1234 (SOMETHING)`)
- if the thing in parens is not a valid tz name, remove it parse the date with just the offset

Another option would be to use fuzzy parsing (i.e. `dt = parse(datetime_str, fuzzy=True)`) but that risks causing surprising behaviour elsewhere, given how widely this util is used.

## How Has This Been Tested?

Added unit tests.

### DIFF
--- a/backend/onyx/connectors/cross_connector_utils/miscellaneous_utils.py
+++ b/backend/onyx/connectors/cross_connector_utils/miscellaneous_utils.py
@@ -1,4 +1,5 @@
 import re
+import string
 from collections.abc import Callable
 from collections.abc import Iterator
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


yuhongsun96:
@evan-danswer is this something you've aware of?

PR #3929, opened 2025-02-06
small linear connector improvements

evan-danswer:
## Description

Some code cleanup and a tiny change to allow users to connect multiple linear workspaces through the OAuth flow.

https://github.com/user-attachments/assets/e64ba58a-8504-472c-83cd-37bd85b3d3dd

TBD: make the OAuth Flow more intuitive?

TBD: handle OAuth refresh for all new type OAuth connectors?

## How Has This Been Tested?

tested extensively in UI

### DIFF
--- a/backend/onyx/server/documents/standard_oauth.py
+++ b/backend/onyx/server/documents/standard_oauth.py
@@ -179,6 +179,7 @@ def oauth_callback(
         db_session=db_session,
     )
 
+    # TODO: use a library for url handling
     sep = "&" if "?" in desired_return_url else "?"
     return CallbackResponse(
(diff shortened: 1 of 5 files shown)


backend/onyx/server/documents/standard_oauth.py
```diff
@@ -179,12 +179,9 @@ def oauth_callback(
         db_session=db_session,
     )
 
+    sep = "&" if "?" in desired_return_url else "?"
```
rkuo-danswer, line 183: i think we should probably standardize on using a url class rather than string manipulation, but maybe that's out of scope for this ... add a TODO at least.

PR #3930, opened 2025-02-07
added context type for when internet search tool is used

evan-danswer:
## Description

Addresses
https://linear.app/danswer/issue/DAN-1427/internet-search-saying-cant-search-the-internet-after-searching-the

We tell the LLM that what it is seeing are internet search results so it knows not to erroneously tell the user that it can't search the internet

## How Has This Been Tested?

tested in UI with 4o and o1


yuhongsun96 (approved):
nit: pref not having multiple copies of strings in the code. Also better that default args like this are not literal strings but global final strings. Slightly more ok to have just strings within a function as it scopes it more but in cases like this, let's use a constant.

PR #3937, opened 2025-02-07
Agent search node timeout

pablonyx:
## Description

Timeouts for Agent Search. Goal is to i) avoid hanging because of a hanging validation step and ii) avoid the user not knowing whether the process is broken if the LLM hangs. 

Strategy:
 - added optional timeouts to all LLM calls (.invoke and .stream) in Agent Search
 - modified all calls accordingly down to litellm call
 - rely on litellm to use it properly with the LLM
 
 Behavior:
   - if validation LLM call times out, assume 'True' and continue
   - if answering of a subquestion times out, the subquestion will effectively be ignored downstream
   - if initial/refined answer generation step fails, show error message but keep the results thus far 

## How Has This Been Tested?
 - locally, by artificially setting timeouts to 0.2
 
 Still in works: behavior if question generation hangs.
 


backend/onyx/agents/agent_search/deep_search/initial/generate_individual_sub_answer/nodes/check_sub_answer.py
```diff
@@ -53,14 +58,46 @@ def check_sub_answer(
...
+    if agent_error:
+        answer_quality = True
+        log_result = agent_error.error_result
+
+    else:
+        if response:
+            quality_str: str = merge_message_runs(response, chunk_separator="")[
+                0
+            ].content
+            answer_quality = "yes" in quality_str.lower()
```
yuhongsun96: We should not be using string matches for logic like these, or wrap it in a utility function that is well encapsulated

backend/onyx/agents/agent_search/deep_search/initial/generate_individual_sub_answer/nodes/check_sub_answer.py
```diff
@@ -53,14 +58,46 @@ def check_sub_answer(
...
+            fast_llm.stream(
+                prompt=msg,
+                timeout_overwrite=AGENT_TIMEOUT_OVERWRITE_LLM_SUBANSWER_CHECK,
+            )
         )
-    )
 
-    quality_str: str = merge_message_runs(response, chunk_separator="")[0].content
-    answer_quality = "yes" in quality_str.lower()
+    except openai.APITimeoutError:
```
yuhongsun96: Since we support a lot of different LLMs and they may raise different errors, should there be a central place to catch these library specific timeout errors (like around the LLM layer) and raise our own general LLMTimeoutError?

That way the rest of the code can use the centralized exception.

backend/onyx/agents/agent_search/deep_search/initial/generate_individual_sub_answer/nodes/check_sub_answer.py
```diff
@@ -53,14 +58,46 @@ def check_sub_answer(
 
     graph_config = cast(GraphConfig, config["metadata"]["config"])
     fast_llm = graph_config.tooling.fast_llm
-    response = list(
-        fast_llm.stream(
-            prompt=msg,
+    agent_error: AgentError | None = None
+    response: list | None = None
+    try:
+        response = list(
```
yuhongsun96: If we don't need to capture individual tokens, there's an invoke function which gathers all of the tokens. Should be more suitable for this.

backend/onyx/agents/agent_search/deep_search/initial/generate_individual_sub_answer/nodes/check_sub_answer.py
```diff
@@ -1,6 +1,7 @@
 from datetime import datetime
 from typing import cast
 
+import openai
```
yuhongsun96: See comment below, it would allow us to not import openai here or LLM specifics in other modules when it can be correctly centralized.

backend/onyx/agents/agent_search/deep_search/initial/generate_individual_sub_answer/nodes/generate_sub_answer.py
```diff
@@ -79,41 +87,66 @@ def generate_sub_answer(
...
-    cited_documents = [
-        context_docs[id] for id in answer_citation_ids if id < len(context_docs)
-    ]
+        else:
+            answer_str = merge_message_runs(response, chunk_separator="")[0].content
+            answer_citation_ids = get_answer_citation_ids(answer_str)
+            cited_documents = [
+                context_docs[id] for id in answer_citation_ids if id < len(context_docs)
+            ]
+            log_results = ""
```
yuhongsun96: Probably some explicit null makes more sense here

backend/onyx/agents/agent_search/deep_search/initial/generate_individual_sub_answer/nodes/generate_sub_answer.py
```diff
@@ -79,41 +87,66 @@ def generate_sub_answer(
...
             )
-            end_stream_token = datetime.now()
-            dispatch_timings.append(
-                (end_stream_token - start_stream_token).microseconds
+
+        except Exception:
+            agent_error = AgentError(
+                error_type="LLM error",
+                error_message=AGENT_LLM_ERROR_MESSAGE,
+                error_result="LLM Error",
```
yuhongsun96: Reused strings should be a final module level variable, if it's only used here, we can just declare at top of file. If shared in other places, can place it in a constants.py file

backend/onyx/agents/agent_search/deep_search/initial/generate_initial_answer/nodes/validate_initial_answer.py
```diff
@@ -25,7 +25,7 @@ def validate_initial_answer(
         f"--------{node_start_time}--------Checking for base answer validity - for not set True/False manually"
     )
 
-    verdict = True
+    verdict = True  # not actually requitred as already streamed out. Refinement will do similar
```
yuhongsun96: nit: typo

backend/onyx/agents/agent_search/deep_search/initial/generate_sub_answers/nodes/decompose_orig_question.py
```diff
@@ -125,19 +157,24 @@ def decompose_orig_question(
...
-    deomposition_response = merge_content(*streamed_tokens)
+    if agent_error:
+        initial_sub_questions: list[str] = []
+        log_result = agent_error.error_result
+    else:
+        deomposition_response = merge_content(*streamed_tokens)
 
-    # this call should only return strings. Commenting out for efficiency
-    # assert [type(tok) == str for tok in streamed_tokens]
+        # this call should only return strings. Commenting out for efficiency
```
yuhongsun96: probably this commented stuff can just go

backend/onyx/agents/agent_search/deep_search/shared/expanded_retrieval/nodes/expand_queries.py
```diff
@@ -16,15 +18,23 @@
...
+from onyx.configs.agent_configs import (
+    AGENT_TIMEOUT_OVERWRITE_LLM_QUERY_REWRITING_GENERATION,
+)
+from onyx.prompts.agent_search import AGENT_LLM_ERROR_MESSAGE
+from onyx.prompts.agent_search import AGENT_LLM_TIMEOUT_MESSAGE
 from onyx.prompts.agent_search import (
     QUERY_REWRITING_PROMPT,
 )
 
+BaseMessage_Content = str | list[str | dict[str, Any]]
```
yuhongsun96: We should not declare this in more than one place. It's the same thing so should only be declared once and imported throughout. Also prefer below (on naming and annotating)

BaseMessageContent: TypeAlias = str | list[str | dict[str, Any]]

backend/onyx/agents/agent_search/deep_search/shared/expanded_retrieval/states.py
```diff
@@ -21,10 +21,13 @@
 
 
 class ExpandedRetrievalInput(SubgraphCoreState):
-    question: str = ""
-    base_search: bool = False
+    # exception from 'no default value' in inputs - used w/o explicit setting for original question
```
yuhongsun96: I don't understand this

backend/onyx/agents/agent_search/run_graph.py
```diff
@@ -123,9 +123,7 @@ def run_main_graph(
 ) -> AnswerStream:
     compiled_graph = load_compiled_graph()
 
-    input = MainInput_a(
-        base_question=config.inputs.search_request.query, log_messages=[]
-    )
+    input = MainInput_a(log_messages=[])
```
yuhongsun96: Class names should just follow the standards MainInputA(whatever)

backend/onyx/agents/agent_search/shared_graph_utils/models.py
```diff
@@ -56,6 +56,12 @@ class InitialAgentResultStats(BaseModel):
     agent_effectiveness: dict[str, float | int | None]
 
 
+class AgentError(BaseModel):
```
yuhongsun96: Is there some design where we could just use builtin python exceptions instead of this?

class AgentException(Exception):
    def __init__(self, message: str, error_type: str, error_result: str | None = None):
        self.message = message
        self.error_type = error_type
        self.error_result = error_result
        super().__init__(self.message)
        
        
        Something like this where the pattern is more like exception handling?

backend/onyx/agents/agent_search/shared_graph_utils/utils.py
```diff
@@ -372,7 +376,20 @@ def summarize_history(
...
 
-    history_response = llm.invoke(history_context_prompt)
+    try:
+        history_response = llm.invoke(
+            history_context_prompt,
+            timeout_overwrite=AGENT_TIMEOUT_OVERWRITE_LLM_HISTORY_SUMMARY_GENERATION,
+        )
+    except openai.APITimeoutError:
+        return (
+            history  # this is what is done at this point anyway, so wwe default to this
```
yuhongsun96: nit: typo here and below

backend/onyx/agents/agent_search/shared_graph_utils/utils.py
```diff
@@ -372,7 +376,20 @@ def summarize_history(
...
+        )
+
+    except Exception:
```
yuhongsun96: I wonder if it's better to fail loudly and propagate up the stack if not an exception we are explicitly handling. Here and in other places. It prevents silent failures from staying around for a long time

backend/onyx/configs/agent_configs.py
```diff
@@ -13,6 +13,20 @@
 AGENT_DEFAULT_MAX_ANSWER_CONTEXT_DOCS = 10
 AGENT_DEFAULT_MAX_STATIC_HISTORY_WORD_LENGTH = 2000
 
+AGENT_DEFAULT_TIMEOUT_OVERWRITE_LLM_GENERAL_GENERATION = 1  # in seconds
```
yuhongsun96: I think in these cases, it's actually better to just use the number in the code below directly 😰. But whatever, doesn't matter

backend/onyx/llm/chat_llm.py
```diff
@@ -444,14 +445,20 @@ def _invoke_implementation(
...
+        timeout_overwrite: int | None = None,
     ) -> BaseMessage:
         if LOG_DANSWER_MODEL_INTERACTIONS:
             self.log_model_configs()
 
         response = cast(
             litellm.ModelResponse,
             self._completion(
-                prompt, tools, tool_choice, False, structured_response_format
+                prompt,
```
yuhongsun96: Not that this matters and I'll stop pointing these out, but slight preference for named args in the case where there are many.

backend/onyx/prompts/agent_search.py
```diff
@@ -5,8 +5,9 @@
 NO_RECOVERED_DOCS = "No relevant information recovered"
 YES = "yes"
 NO = "no"
-
-
+AGENT_LLM_TIMEOUT_MESSAGE = "The agent timed out. Please try again."
```
yuhongsun96: These aren't prompts though, pref a constants.py file under agents directory

yuhongsun96 (commented):
Two main points, the rest are nits:
- We should have a single point where we are catching the timeouts instead of handling openai and other ones directly up the stack
- We should fail loudly in other exception cases to avoid hidden errors

backend/onyx/agents/agent_search/deep_search/initial/generate_individual_sub_answer/nodes/check_sub_answer.py
joachim-danswer: Agreed, created a function called binary_string_test(text, positive_value="yes") for that now.

backend/onyx/agents/agent_search/deep_search/initial/generate_individual_sub_answer/nodes/check_sub_answer.py
joachim-danswer: Done. LiteLLM errors caught in chatllm.py and then raised as our own errors LLMTimeoutError and LLMRateLimitError.

backend/onyx/agents/agent_search/deep_search/initial/generate_individual_sub_answer/nodes/check_sub_answer.py
joachim-danswer: Done

backend/onyx/agents/agent_search/deep_search/initial/generate_individual_sub_answer/nodes/check_sub_answer.py
joachim-danswer: done

backend/onyx/agents/agent_search/deep_search/initial/generate_individual_sub_answer/nodes/generate_sub_answer.py
joachim-danswer: Just for logging. But done.

backend/onyx/agents/agent_search/shared_graph_utils/utils.py
joachim-danswer: done

backend/onyx/agents/agent_search/shared_graph_utils/utils.py
joachim-danswer: In general, I agree. Here however it is only a nice to have.

backend/onyx/configs/agent_configs.py
joachim-danswer: I prefer as is to see the parameters in one go. In fact, I think we should do that in general IMO.

backend/onyx/llm/chat_llm.py
joachim-danswer: I was there beforehand. But done.

backend/onyx/prompts/agent_search.py
joachim-danswer: done

backend/onyx/agents/agent_search/deep_search/initial/generate_individual_sub_answer/nodes/check_sub_answer.py
```diff
@@ -53,14 +72,50 @@ def check_sub_answer(
...
     graph_config = cast(GraphConfig, config["metadata"]["config"])
     fast_llm = graph_config.tooling.fast_llm
-    response = list(
-        fast_llm.stream(
+    agent_error: AgentError | None = None
+    response: BaseMessage | None = None
+    try:
+        response = fast_llm.invoke(
             prompt=msg,
+            timeout_overwrite=AGENT_TIMEOUT_OVERWRITE_LLM_SUBANSWER_CHECK,
```
evan-onyx: prefer timeout_override

backend/onyx/agents/agent_search/deep_search/initial/generate_individual_sub_answer/nodes/check_sub_answer.py
```diff
@@ -53,14 +72,50 @@ def check_sub_answer(
...
         )
-    )
 
-    quality_str: str = merge_message_runs(response, chunk_separator="")[0].content
-    answer_quality = "yes" in quality_str.lower()
+    except LLMTimeoutError:
+        agent_error = AgentError(
+            error_type="timeout",
```
evan-onyx: enum for error types

backend/onyx/agents/agent_search/deep_search/initial/generate_individual_sub_answer/nodes/check_sub_answer.py
```diff
@@ -53,14 +72,50 @@ def check_sub_answer(
...
+            error_message=AGENT_LLM_ERROR_MESSAGE,
+            error_result="LLM Error",
+        )
+        logger.error("General LLM Error - check sub answer")
+    if agent_error:
+        answer_quality = True
+        log_result = agent_error.error_result
+
+    else:
+        if response:
```
evan-onyx, line 112: potentially unbound response? As far as I can see this is only set inside the try block

backend/onyx/agents/agent_search/deep_search/initial/generate_individual_sub_answer/nodes/check_sub_answer.py
```diff
@@ -53,14 +72,50 @@ def check_sub_answer(
...
+            quality_str: str = cast(str, response.content)
+            answer_quality = binary_string_test(text=quality_str, positive_value="yes")
```
evan-onyx: Probably should have "yes" defined as a string constant (if it isn't already)

backend/onyx/agents/agent_search/deep_search/initial/generate_initial_answer/nodes/generate_initial_answer.py
```diff
@@ -224,30 +238,91 @@ def generate_initial_answer(
...
+                agent_base_end_time=None,
+                agent_base_metrics=None,
+                log_messages=[
+                    get_langgraph_node_log_string(
+                        graph_component="initial - generate initial answer",
+                        node_name="generate initial answer",
+                        node_start_time=node_start_time,
+                        result=agent_error.error_result or "An LLM error occurred",
+                    )
+                ],
```
evan-onyx, line 322: Need to dispatch_main_answer_stop_info before returning

backend/onyx/agents/agent_search/deep_search/initial/generate_sub_answers/nodes/decompose_orig_question.py
```diff
@@ -125,19 +175,24 @@ def decompose_orig_question(
     )
     write_custom_event("stream_finished", stop_event, writer)
 
-    deomposition_response = merge_content(*streamed_tokens)
+    if agent_error:
+        initial_sub_questions: list[str] = []
+        log_result = agent_error.error_result
+    else:
+        deomposition_response = merge_content(*streamed_tokens)
```
evan-onyx, line 179: typo: decomposition

backend/onyx/agents/agent_search/deep_search/main/nodes/create_refined_sub_questions.py
```diff
@@ -96,29 +118,71 @@ def create_refined_sub_questions(
...
+            error_type="LLM error",
+            error_message=AGENT_LLM_ERROR_MESSAGE,
+            error_result="The LLM errored out, and the subquestions could not be generated.",
+        )
+        logger.error("General LLM Error - create refined sub questions")
+    if agent_error:
+        refined_sub_question_dict: dict[int, RefinementSubQuestion] = {}
+        log_result = agent_error.error_result
+        write_custom_event(
+            "refined_sub_question_creation_error",
```
evan-onyx, line 157: Need to add this event type to _parse_event

backend/onyx/agents/agent_search/deep_search/shared/expanded_retrieval/nodes/verify_documents.py
```diff
@@ -51,11 +68,49 @@ def verify_documents(
...
+        logger.error("General LLM Error - verify documents")
+    if agent_error or response is None:
+        verified_documents = [retrieved_document_to_verify]
 
-    verified_documents = []
-    if isinstance(response.content, str) and "yes" in response.content.lower():
-        verified_documents.append(retrieved_document_to_verify)
+    else:
+        verified_documents = []
+        if isinstance(response.content, str) and "yes" in response.content.lower():
```
evan-onyx: call helper function for "yes" in .... check

backend/onyx/agents/agent_search/shared_graph_utils/utils.py
evan-onyx: fwiw except Exception make me very sad :'(

backend/onyx/llm/chat_llm.py
```diff
@@ -424,6 +437,12 @@ def _completion(
         except Exception as e:
             self._record_error(processed_prompt, e)
             # for break pointing
+            if isinstance(e, litellm.Timeout):
```
evan-onyx, line 440: Prefer `except litellm.Timeout as e` and `except litellm.RateLimitError as e` to reduce the amount of times we except Exception

backend/onyx/llm/chat_llm.py
```diff
@@ -469,19 +494,31 @@ def _stream_implementation(
...
             self.log_model_configs()
 
         if DISABLE_LITELLM_STREAMING:
-            yield self.invoke(prompt, tools, tool_choice, structured_response_format)
+            yield self.invoke(
+                prompt,
+                tools,
+                tool_choice,
+                structured_response_format,
+                timeout_overwrite,
```
evan-onyx: override

web/src/app/chat/message/AgenticMessage.tsx
```diff
@@ -483,11 +485,28 @@ export const AgenticMessage = ({
...
+                            </p>
+                          )}
                         </div>
                       </div>
                     </>
-                  ) : isComplete ? null : (
-                    <></>
+                  ) : isComplete ? (
+                    error && (
+                      <p className="mt-2 mx-4 text-red-700 text-sm my-auto">
```
evan-onyx, line 498: maybe could have a cute lil error component here? Would make me happier by slightly reducing indentation :')

evan-onyx: @pablonyx

evan-danswer (changes requested):
A few small streaming things need changes

backend/onyx/llm/chat_llm.py
joachim-danswer: done

backend/onyx/agents/agent_search/deep_search/initial/generate_initial_answer/nodes/validate_initial_answer.py
joachim-danswer: done

backend/onyx/agents/agent_search/deep_search/initial/generate_individual_sub_answer/nodes/generate_sub_answer.py
joachim-danswer: Done

backend/onyx/agents/agent_search/deep_search/initial/generate_sub_answers/nodes/decompose_orig_question.py
joachim-danswer: done

backend/onyx/agents/agent_search/deep_search/shared/expanded_retrieval/nodes/expand_queries.py
joachim-danswer: done

backend/onyx/agents/agent_search/deep_search/shared/expanded_retrieval/states.py
joachim-danswer: expanded. It is a convention we chose for LangGraph states that we have an exception for here.

backend/onyx/agents/agent_search/run_graph.py
joachim-danswer: changed to MainInput

backend/onyx/agents/agent_search/shared_graph_utils/models.py
joachim-danswer: As discussed, this is for Logging purposes.

joachim-danswer: Will rename to make that clear.

backend/onyx/agents/agent_search/shared_graph_utils/utils.py
joachim-danswer: DONE!!

evan-danswer:
Rolled into https://github.com/onyx-dot-app/onyx/pull/3994

PR #3948, opened 2025-02-10
refactoring and update multiplier in real time

rkuo-danswer:
## Description

Fixes DAN-1434.

https://linear.app/danswer/issue/DAN-1434/update-cloud-beat-multiplier-from-redis

General refactoring and update the multiplier from redis, which allows us to respond to task throughput issues in near real time in prod.

## How Has This Been Tested?

### DIFF
--- a/backend/ee/onyx/background/celery/tasks/beat_schedule.py
+++ b/backend/ee/onyx/background/celery/tasks/beat_schedule.py
@@ -3,42 +3,44 @@ from typing import Any
 
 from onyx.background.celery.tasks.beat_schedule import BEAT_EXPIRES_DEFAULT
 from onyx.background.celery.tasks.beat_schedule import (
-    cloud_tasks_to_schedule as base_cloud_tasks_to_schedule,
+    beat_system_tasks as base_beat_system_tasks,
 )
 from onyx.background.celery.tasks.beat_schedule import (
-    tasks_to_schedule as base_tasks_to_schedule,
+    beat_task_templates as base_beat_task_templates,
 )
-from onyx.configs.constants import ONYX_CLOUD_CELERY_TASK_PREFIX
+from onyx.background.celery.tasks.beat_schedule import generate_cloud_tasks
+from onyx.background.celery.tasks.beat_schedule import (
+    get_tasks_to_schedule as base_get_tasks_to_schedule,
+)
 from onyx.configs.constants import OnyxCeleryPriority
 from onyx.configs.constants import OnyxCeleryTask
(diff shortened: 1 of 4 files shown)


backend/onyx/background/celery/apps/beat.py
```diff
@@ -121,44 +134,57 @@ def _generate_schedule(
...
+            break
 
-        if DynamicTenantScheduler._compare_schedules(current_schedule, new_schedule):
-            logger.info(
-                "_try_updating_schedule: Current schedule is up to date, no changes needed"
+        if not do_update:
+            # exit early if nothing changed
+            task_logger.info(
+                f"_try_updating_schedule - Schedule unchanged: "
+                f"tasks={len(new_schedule)} "
```
pablonyx, line 185: Believe we can get rid of the schedule comparison now? We used to have dynamic tasks scheduled by the beat since we relied on the beat to send out tasks by tenant, but I'm under the impression teh task list is now static (excluding the beat multiplier)

backend/ee/onyx/background/celery/tasks/beat_schedule.py
```diff
@@ -3,42 +3,44 @@
...
-        "kwargs": {
-            "task_name": OnyxCeleryTask.CHECK_TTL_MANAGEMENT_TASK,
+        {
+            "name": "check-ttl-management",
+            "task": OnyxCeleryTask.CHECK_TTL_MANAGEMENT_TASK,
+            "schedule": timedelta(hours=1),
+            "options": {
+                "priority": OnyxCeleryPriority.MEDIUM,
+                "expires": BEAT_EXPIRES_DEFAULT,
+            },
```
pablonyx, line 40: Just confirming this priority change is intentional?

pablonyx (approved):

backend/ee/onyx/background/celery/tasks/beat_schedule.py
rkuo-danswer: Yes intentional.  The function that converts the template to the beat task sets the generator tasks to a high/highest priority, and the priority here is now for the fanned out tasks created by the generators.

backend/onyx/background/celery/apps/beat.py
rkuo-danswer: i think we could in our current state, but i also think the general idea of comparing schedules and updating the beat based on that is likely to be useful in the future, so i'd like to leave this in even if it doesn't do much right now. I will add a comment tho!

PR #3952, opened 2025-02-10
Bugfix/dedupe ids

rkuo-danswer:
## Description

Fixes DAN-1433.

https://linear.app/danswer/issue/DAN-1433/assistant-editing-bug-for-private-assistant

## How Has This Been Tested?


backend/tests/integration/tests/permissions/test_persona_permissions.py
```diff
@@ -139,9 +140,13 @@ def test_persona_permissions(reset: None) -> None:
 
     """Test admin permissions"""
     # Admin can edit any persona
+
+    # the persona was created with the admin user, and this will add the admin user
```
Weves: nit: I think this comment could be made clearer. What's really happening is that the persona was shared with the admin user (`created with` is a little unclear imo).

backend/ee/onyx/db/persona.py
```diff
@@ -11,10 +11,13 @@
 
 def make_persona_private(
     persona_id: int,
-    user_ids: list[UUID] | None,
+    user_ids: list[str] | None,
```
Weves: just curious, what's the reason for change from UUID to str?

backend/tests/integration/common_utils/managers/persona.py
```diff
@@ -119,6 +119,11 @@ def edit(
     ) -> DATestPersona:
         system_prompt = system_prompt or f"System prompt for {persona.name}"
         task_prompt = task_prompt or f"Task prompt for {persona.name}"
+
+        upsert_users = copy.deepcopy(persona.users)
```
Weves: also curious here, why the change? Does this mean that you can't remove users from a persona anymore?

backend/ee/onyx/db/persona.py
rkuo-danswer: UUID is not a basic type so outside of this function the pydantic model for PersonaUpsertRequest.users had to be changed to str to serialize/deserialize properly.

Weves (approved):

PR #3954, opened 2025-02-10
Increase timeout for reasoning models + make o1 available by default

Weves:
## Description

https://linear.app/danswer/issue/DAN-1436/different-timeout-for-reasoning-models

## How Has This Been Tested?

Tested locally, previously timed out -> doesn't timeout anymore.

### DIFF
--- a/backend/onyx/llm/chat_llm.py
+++ b/backend/onyx/llm/chat_llm.py
@@ -27,6 +27,7 @@ from langchain_core.prompt_values import PromptValue
 
 from onyx.configs.app_configs import LOG_DANSWER_MODEL_INTERACTIONS
 from onyx.configs.app_configs import MOCK_LLM_RESPONSE
+from onyx.configs.chat_configs import QA_TIMEOUT
 from onyx.configs.model_configs import (
     DISABLE_LITELLM_STREAMING,
 )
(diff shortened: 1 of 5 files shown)


backend/onyx/llm/chat_llm.py
```diff
@@ -229,25 +231,32 @@ class DefaultMultiLLM(LLM):
     def __init__(
         self,
         api_key: str | None,
-        timeout: int,
         model_provider: str,
         model_name: str,
+        timeout: int | None = None,
```
Weves, line 236: making these `None` by default and setting the default within the init to allow for more complicated logic here

backend/onyx/llm/llm_provider_options.py
```diff
@@ -29,11 +29,11 @@ class WellKnownLLMProviderDescriptor(BaseModel):
 OPEN_AI_MODEL_NAMES = [
     "o3-mini",
     "o1-mini",
-    "o1-preview",
-    "o1-2024-12-17",
+    "o1",
```
Weves, line 32: o1 should be higher up than preview

web/src/lib/hooks.ts
```diff
@@ -647,11 +647,11 @@ export const useUserGroups = (): {
 
 const MODEL_DISPLAY_NAMES: { [key: string]: string } = {
   // OpenAI models
-  "o1-2025-12-17": "O1 (December 2025)",
-  "o3-mini": "O3 Mini",
-  "o1-mini": "O1 Mini",
-  "o1-preview": "O1 Preview",
-  o1: "O1",
+  "o1-2025-12-17": "o1 (December 2025)",
```
Weves, line 650: matching case of OpenAI's docs/chatgpt

pablonyx (approved):

PR #3970, opened 2025-02-12
Add delete all for tenants in Vespa

pablonyx:
## Description

Adds functionality for deleting documents by tenant (useful for inactive tenants we want to purge)

## How Has This Been Tested?

Tested with existing tenant 

### DIFF
--- a/backend/scripts/debugging/onyx_vespa.py
+++ b/backend/scripts/debugging/onyx_vespa.py
@@ -256,16 +256,28 @@ def get_documents_for_tenant_connector(
 
 
 def search_for_document(
-    index_name: str, document_id: str, max_hits: int | None = 10
+    index_name: str,
+    document_id: str | None = None,
+    tenant_id: str | None = None,
+    max_hits: int | None = 10,
 ) -> List[Dict[str, Any]]:
-    yql_query = (
-        f'select * from sources {index_name} where document_id contains "{document_id}"'
-    )
+    yql_query = f"select * from sources {index_name}"
+
+    conditions = []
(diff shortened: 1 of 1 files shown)


backend/scripts/debugging/onyx_vespa.py
```diff
@@ -630,7 +790,9 @@ def main() -> None:
     args = parser.parse_args()
     vespa_debug = VespaDebugging(args.tenant_id)
 
-    if args.action == "config":
+    if args.action == "delete-all-documents":
+        vespa_debug.delete_documents_for_tenant(args.count)
```
rkuo-danswer, line 794: VespaDebugging takes a tenant_id already, doesn't that make the method name "delete_documents_for_tenant" a bit redundant?  Could it just be delete_all_documents?

backend/scripts/debugging/onyx_vespa.py
```diff
@@ -600,6 +619,147 @@ def acls(self, cc_pair_id: int, n: int | None = 10) -> None:
...
+                response = client.delete(url, params=params)
+                if response.status_code == 200:
+                    logger.info(f"Successfully deleted doc_id={doc_id_value}")
+                    deleted_count += 1
+                else:
+                    logger.error(
+                        f"Failed to delete doc_id={doc_id_value}, "
+                        f"status={response.status_code}, response={response.text}"
+                    )
+                    print(
```
rkuo-danswer, line 751: Any reason to print if we are already using logger.error?

rkuo-danswer (approved):

PR #3971, opened 2025-02-12
Clean illegal chars in metadata

Weves:
## Description

https://linear.app/danswer/issue/DAN-1441/indexing-failure-due-to-illegal-code-point-in-metadata

## How Has This Been Tested?

👀 


rkuo-danswer (approved):
Approved, probably worth a comment in the code or follow up task to more broadly handle escaping in Vespa per https://stackoverflow.com/questions/54044575/should-ascii-control-characters-be-stripped-from-documents-before-sending-to-ves

PR #3976, opened 2025-02-12
Billing fixes

pablonyx:
#### Brief background

Should go in in tandem with https://github.com/onyx-dot-app/danswer-control-plane/pull/26

When it comes to billing, we have a few pieces of relevant information to store:
- Subscription status (trial / subscribed / etc.- all the other Stripe [subscription status states](https://docs.stripe.com/billing/subscriptions/overview#payment-status))
- Cancel at period end? (boolean)
- Application status (status in the Onyx application)
	- What the current status is for a given user (full access, payment reminder, no access)
- Time frames: trial start/end, bill period start/end

The ground truth is **always** Stripe, so we use the Stripe customer /subscription IDs to be able to reference the Stripe API as often as possible. That being said, we have two mechanisms for keeping subscription details up to date in the [control plane](https://github.com/onyx-dot-app/danswer-control-plane/pull/26):
- Background processes that regularly crawl for the current status from Stripe, update subscription and tenant objects as needed
- Web-hooks which should update details in real time

How do we make sure that the data plane always has accurate and up-to-date information on a given user's billing information (considering we don't want to add unnecessary complexity to the data plane)?
- For billing page specifically, ping control plane for billing details to show user 
- For application status (i.e. full / gated access),  transition "gating" notification endpoint to "application status update" endpoint which the control plane **must** receive a 200 from before updating the tenant's application status inthe control plane (ensuring the two are in sync).

#### What this PR does

- Remove extraneous payment reminder notification "gating type" and transition to "application status" which should be continuously kept in sync with the data plane / ground truth
- Clarify billing page with trial dates, "cancels at end of month?", and payment reminders.
- Enable un-subscribed + inactive tenants to re-subscribe from their gating page by creating new checkout session
- General code cleanup

## How Has This Been Tested?
Test various billing flows with local multi tenant setup,
with both and only one of each of background processes and webhooks enabled
- Cancel during trial with/without payment info
- Cancel after trial ends
- Cancel + immediately end trial
- Resubscribe from "gating" page after unsubscribing and expiring subscription
- Don't put in billing information  + get to <2 days of trial left 


web/src/app/ee/admin/billing/BillingAlerts.tsx
```diff
@@ -0,0 +1,72 @@
+import React from "react";
+import { Alert, AlertDescription, AlertTitle } from "@/components/ui/alert";
+import { CircleAlert, Info } from "lucide-react";
+
+interface BillingAlertsProps {
+  billingInformation: any;
+}
+
+export function BillingAlerts({ billingInformation }: BillingAlertsProps) {
+  const isTrialing = billingInformation.status === "trialing";
```
rkuo-danswer: maybe we can define constants for this somewhere?

rkuo-danswer (approved):
did my best to review this, seems OK.  This seems like a good candidate to add to Chris's testing bash document.

PR #3979, opened 2025-02-13
Fix Agent Slowness

yuhongsun96:
## Description
All parallel queries go through the intent model which is not needed. Only the initial user query should go through that one.

For Evan:
- Address all the TODOs
- Make sure the other flows do not use intent model
- Double check all Tools still work correctly
- Try to make the design maintainable (no string checks)

## How Has This Been Tested?

### DIFF
--- a/backend/ee/onyx/server/query_and_chat/query_backend.py
+++ b/backend/ee/onyx/server/query_and_chat/query_backend.py
@@ -83,6 +83,7 @@ def handle_search_request(
         user=user,
         llm=llm,
         fast_llm=fast_llm,
+        skip_query_analysis=False,
         db_session=db_session,
         bypass_acl=False,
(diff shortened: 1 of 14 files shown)


backend/onyx/agents/agent_search/deep_search/shared/expanded_retrieval/nodes/rerank_documents.py
```diff
@@ -57,10 +57,12 @@ def rerank_documents(
             persona=graph_config.inputs.search_request.persona,
             rerank_settings=graph_config.inputs.search_request.rerank_settings,
         )
+        # TODO clean up implementation here
```
yuhongsun96: Don't need to address this one

PR #3988, opened 2025-02-13
updated rerank function arguments

joachim-danswer:
## Description

Updated reranking function to not require a whole SearchQuery, but a more narrow set
(query_str and rerank_settings)

Changes have been propagated up, staring with  semantic_reranking at the bottom.

## How Has This Been Tested?

Tested locally with one query and both Basic Search and Agent Search 

### DIFF
--- a/backend/onyx/agents/agent_search/deep_search/shared/expanded_retrieval/nodes/rerank_documents.py
+++ b/backend/onyx/agents/agent_search/deep_search/shared/expanded_retrieval/nodes/rerank_documents.py
@@ -23,7 +23,7 @@ from onyx.configs.agent_configs import AGENT_RERANKING_STATS
 from onyx.context.search.models import InferenceSection
 from onyx.context.search.models import RerankingDetails
 from onyx.context.search.postprocessing.postprocessing import rerank_sections
-from onyx.context.search.postprocessing.postprocessing import reranking_is_runnable
+from onyx.context.search.postprocessing.postprocessing import should_rerank
 from onyx.db.engine import get_session_context_manager
 from onyx.db.search_settings import get_current_search_settings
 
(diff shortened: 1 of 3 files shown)


backend/onyx/agents/agent_search/deep_search/shared/expanded_retrieval/nodes/rerank_documents.py
```diff
@@ -47,35 +47,24 @@ def rerank_documents(
...
-        )
-        _search_query = retrieval_preprocessing(
-            search_request=search_request,
-            user=graph_config.tooling.search_tool.user,  # bit of a hack
-            llm=graph_config.tooling.fast_llm,
-            db_session=db_session,
-        )
 
-    # skip section filtering
+    search_request = SearchRequest(
```
yuhongsun96: This is also extraneous

backend/onyx/context/search/postprocessing/postprocessing.py
```diff
@@ -77,7 +78,8 @@ def _remove_metadata_suffix(chunk: InferenceChunkUncleaned) -> str:
 
 @log_function_time(print_only=True)
 def semantic_reranking(
-    query: SearchQuery,
+    query_str: str,
+    rerank_settings: RerankingDetails | None,
```
yuhongsun96: This should not allow a None

yuhongsun96 (approved):

backend/onyx/context/search/postprocessing/postprocessing.py
```diff
@@ -165,8 +165,20 @@ def semantic_reranking(
     return list(ranked_chunks), list(ranked_indices)
 
 
+def reranking_is_runnable(rerank_settings: RerankingDetails | None) -> bool:
```
evan-onyx: prefer calling this should_rerank

backend/onyx/context/search/postprocessing/postprocessing.py
```diff
@@ -260,16 +269,13 @@ def search_postprocessing(
...
-        and search_query.rerank_settings.num_rerank > 0
-    ):
+    if reranking_is_runnable(search_query.rerank_settings):
         post_processing_tasks.append(
             FunctionCall(
                 rerank_sections,
                 (
-                    search_query,
+                    search_query.query,
+                    search_query.rerank_settings,  # Cannot be None here
```
evan-onyx, line 278: probably don't need the comment since the function signature contains that info

evan-danswer (approved):
LGTM! Minor nits

backend/onyx/context/search/postprocessing/postprocessing.py
yuhongsun96: While true, unfortunately the static type checks do not work here due to the need for allowing arbitrary functions and inputs. You could pass in something that is completely the wrong type and it would only fail at runtime :(

Maybe there's a fix but... probably this is ok for now

backend/onyx/context/search/postprocessing/postprocessing.py
yuhongsun96: seems good, I did also think this was ugly

PR #3990, opened 2025-02-13
Pause cloud tasks for gated tenants

pablonyx:
## Description

## How Has This Been Tested?

### DIFF
--- a/backend/ee/onyx/server/tenants/product_gating.py
+++ b/backend/ee/onyx/server/tenants/product_gating.py
@@ -1,0 +1,48 @@
+from ee.onyx.configs.app_configs import GATED_TENANTS_KEY
+from onyx.configs.constants import ONYX_CLOUD_TENANT_ID
+from onyx.redis.redis_pool import get_redis_client
+from onyx.server.settings.models import ApplicationStatus
+from onyx.server.settings.store import load_settings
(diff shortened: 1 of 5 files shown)


backend/ee/onyx/server/tenants/product_gating.py
```diff
@@ -0,0 +1,51 @@
...
+
+        if token is not None:
+            CURRENT_TENANT_ID_CONTEXTVAR.reset(token)
+
+    except Exception:
+        logger.exception("Failed to gate product")
+        raise
+
+
+def get_gated_tenants() -> list[str]:
```
rkuo-danswer: I believe this lookup does not need to be strongly consistent, so perhaps we can use the replica client to pull this list?

backend/onyx/background/celery/tasks/shared/tasks.py
```diff
@@ -252,7 +253,11 @@ def cloud_beat_task_generator(
 
     try:
         tenant_ids = get_all_tenant_ids()
+        gated_tenants = get_gated_tenants()
```
rkuo-danswer, line 256: if we're doing a lookup for every tenant, let's return a set from get_gated_tenants instead of a list so that the lookup is not O(n).

rkuo-danswer (approved):

PR #4009, opened 2025-02-15
Seeding count issue

pablonyx:
## Description

Current issue:
- Chunk count not properly counted, not updated
- This leads to seeded docs not being deleted by our background processes as we assume their UUIDs were generated with the "old" system (we use presence of chunk counts to deleting the UUID process)

## How Has This Been Tested?


yuhongsun96 (approved):
Don't fully understand it but seems small, as long as you tested it :P

PR #4011, opened 2025-02-15
no reranking if local model w/o GPU for Agent Search

joachim-danswer:
## Description

If there is no GPU and a local reranking model is used, we will not rerank.

Implemented using:
 -  rerank_settings.rerank_type is None
 - gpu_status_request() used at the beginning of the  call. 
 - new variable graph_config.behavior.allow_agent_reranking
 as indicators

Linear: 
https://linear.app/danswer/issue/DAN-1454/disable-re-ranking-in-agent-search-if-local-model-wo-gpu-is-used

## How Has This Been Tested?

Tested locally on Mac w/ GPU and at scale.onyx w/o GPU. In the scale test, reranking was omitted.


evan-danswer (changes requested):
We should not be calling gpu_status_request() on every rerank; this should be added to the config in the GraphSearchConfig (the key is behavior in GraphConfig)

evan-danswer (approved):
LGTM

PR #4012, opened 2025-02-15
By default, use primary LLM for initial & refined answer

joachim-danswer:
Use of new env variable

## Description

Unless specified in an environment variable, the primary LLM should be used to generate initial and refined answer.

Addressed through new environment var AGENT_ANSWER_GENERATION_BY_FAST_LLM

Linear: https://linear.app/danswer/issue/DAN-1455/use-primary-llms-as-default-llm-for-initial-and-refined-answer

## How Has This Been Tested?

Locally.Run w/ and w/o this variable set.  


evan-danswer (approved):
LGTM

joachim-danswer:
I still think that we should also set an override parameter...

PR #4016, opened 2025-02-16
Connector validation

pablonyx:
## Description

Adds connector validation for ccpair creation / credential swapping.

New enum for ccpair status: `invalid`- marks ccpairs which have an invalid configuaiton

https://github.com/user-attachments/assets/50b80ccb-e2e8-4648-ae0c-6b27fea31edd

## How Has This Been Tested?

### DIFF
--- a/backend/onyx/background/indexing/run_indexing.py
+++ b/backend/onyx/background/indexing/run_indexing.py
@@ -21,6 +21,7 @@ from onyx.configs.constants import DocumentSource
 from onyx.configs.constants import MilestoneRecordType
 from onyx.connectors.connector_runner import ConnectorRunner
 from onyx.connectors.factory import instantiate_connector
+from onyx.connectors.interfaces import ConnectorValidationError
 from onyx.connectors.models import ConnectorCheckpoint
 from onyx.connectors.models import ConnectorFailure
 from onyx.connectors.models import Document
@@ -86,6 +87,11 @@ def _get_connector_runner(
             credential=attempt.connector_credential_pair.credential,
             tenant_id=tenant_id,
         )
+
+        # validate the connector settings
+
+        runnable_connector.validate_connector_settings()
+
     except Exception as e:
         logger.exception(f"Unable to instantiate connector due to {e}")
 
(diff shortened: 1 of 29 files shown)


backend/onyx/background/indexing/job_client.py
```diff
@@ -75,6 +76,11 @@ def _initializer(
         queue.put(error_msg)  # Send the exception to the parent process
 
         sys.exit(e.code)  # use the given exit code
+    except ConnectorValidationError as e:
+        error_msg = str(e)
+        queue.put(error_msg)  # Send the exception to the parent process
+
+        sys.exit(256)  # use 256 to indicate a connector validation error
```
rkuo-danswer: this ties the job client to connectors ... is there a way to separate this? Also the 256 is an issue as mentioned earlier.

backend/onyx/background/indexing/run_indexing.py
```diff
@@ -422,8 +427,30 @@ def _run_indexing(
             logger.exception(
                 f"Connector run exceptioned after elapsed time: {time.time() - start_time} seconds"
             )
+            logger.error(type(e))
```
rkuo-danswer: can we combine the error logging here with the exception log if that's intentional to keep in?

backend/onyx/connectors/factory.py
```diff
@@ -157,3 +161,39 @@ def instantiate_connector(
...
     return connector
+
+
+def validate_ccpair_for_user(
+    connector_id: int,
+    credential_id: int,
+    db_session: Session,
+    user: User | None,
+    tenant_id: str | None,
+) -> None:
```
rkuo-danswer, line 189: I'm unclear on what happens here if user is None or tenant_id is None.  Might be worth a comment?

web/src/app/admin/connector/[ccPairId]/page.tsx
```diff
@@ -283,6 +287,7 @@ function Main({ ccPairId }: { ccPairId: number }) {
         status={ccPair.last_index_attempt_status || "not_started"}
         disabled={ccPair.status === ConnectorCredentialPairStatus.PAUSED}
         isDeleting={isDeleting}
+        isInvalid={ccPair.status === ConnectorCredentialPairStatus.INVALID}
```
Comment (rkuo-danswer) [inline web/src/app/admin/connector/[ccPairId]/page.tsx]: Should isDeleting and isInvalid be part of a single enum?

web/src/app/admin/connector/[ccPairId]/ReIndexButton.tsx
```diff
@@ -125,15 +127,17 @@ export function ReIndexButton({
...
+        disabled={isDisabled || isDeleting || isInvalid}
         tooltip={
-          isDeleting
-            ? "Cannot index while connector is deleting"
-            : isIndexing
-              ? "Indexing is already in progress"
-              : isDisabled
-                ? "Connector must be re-enabled before indexing"
-                : undefined
+          isInvalid
```
Comment (rkuo-danswer) [inline web/src/app/admin/connector/[ccPairId]/ReIndexButton.tsx]: Do we like this pattern? I find it hard to read.

backend/onyx/connectors/factory.py
pablonyx: Comment added!

backend/onyx/background/indexing/run_indexing.py
pablonyx: Indeed!

web/src/app/admin/connector/[ccPairId]/page.tsx
Comment (pablonyx) [reply web/src/app/admin/connector/[ccPairId]/page.tsx]: Think this is fair generally- lots of re-use across the board

backend/onyx/background/indexing/job_client.py
pablonyx: Updated to be more in line with what we'd expect here!

rkuo-danswer (approved):

PR #4018, opened 2025-02-17
Additional connector validation classes

pablonyx:
## Description

Basic validation for:
- Github
- Notion
- Jira

## How Has This Been Tested?

### DIFF
--- a/backend/onyx/connectors/dropbox/connector.py
+++ b/backend/onyx/connectors/dropbox/connector.py
@@ -4,12 +4,17 @@ from typing import Any
 
 from dropbox import Dropbox  # type: ignore
 from dropbox.exceptions import ApiError  # type:ignore
+from dropbox.exceptions import AuthError  # type:ignore
+from dropbox.exceptions import HttpError  # type:ignore
 from dropbox.files import FileMetadata  # type:ignore
 from dropbox.files import FolderMetadata  # type:ignore
 
 from onyx.configs.app_configs import INDEX_BATCH_SIZE
 from onyx.configs.constants import DocumentSource
+from onyx.connectors.interfaces import ConnectorValidationError
+from onyx.connectors.interfaces import CredentialInvalidError
 from onyx.connectors.interfaces import GenerateDocumentsOutput
+from onyx.connectors.interfaces import InsufficientPermissionsError
 from onyx.connectors.interfaces import LoadConnector
 from onyx.connectors.interfaces import PollConnector
 from onyx.connectors.interfaces import SecondsSinceUnixEpoch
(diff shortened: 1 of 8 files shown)


backend/onyx/connectors/dropbox/connector.py
```diff
@@ -141,6 +146,31 @@ def poll_source(
...
         return None
 
+    def validate_connector_settings(self) -> None:
+        if self.dropbox_client is None:
+            raise ConnectorMissingCredentialError("Dropbox credentials not loaded.")
+
+        try:
+            self.dropbox_client.files_list_folder(path="", limit=1)
+        except AuthError as e:
+            logger.exception(f"Failed to validate Dropbox credentials: {e}")
```
rkuo-danswer: shouldn't be necessary to log {e} with logger.exception

backend/onyx/connectors/dropbox/connector.py
```diff
@@ -141,6 +146,31 @@ def poll_source(
...
+                e.error is not None
+                and "insufficient_permissions" in str(e.error).lower()
+            ):
+                raise InsufficientPermissionsError(
+                    "Your Dropbox token does not have sufficient permissions."
+                )
+            raise ConnectorValidationError(
+                f"Unexpected Dropbox error during validation: {e.user_message_text or e}"
+            )
+        except Exception as e:
```
rkuo-danswer, line 168: Do we need an UnexpectedError class? Or can we just reraise the exception?

backend/onyx/connectors/notion/connector.py
```diff
@@ -616,6 +622,64 @@ def poll_source(
             else:
                 break
 
+    def validate_connector_settings(self) -> None:
+        if "Authorization" not in self.headers or not self.headers["Authorization"]:
```
rkuo-danswer: This seems redundant ... wouldn't "if not self.headers.get("Authorization"):" work for this?

rkuo-danswer (approved):

backend/onyx/connectors/dropbox/connector.py
pablonyx: Not sure why I keep leaving that in!

backend/onyx/connectors/notion/connector.py
pablonyx: Sure would! Legacy from previous check– definitely superfluous here

backend/onyx/connectors/dropbox/connector.py
pablonyx: The thought was to eventually have some additional handling around ephemeral errors, but agree– in the current iteration, pretty superfluous

PR #4019, opened 2025-02-17
added timeouts for agent llm calls

evan-danswer:
## Description

https://linear.app/danswer/issue/DAN-1456/llm-call-timeouts-for-agent-search

- new utility for running a function for a specified number of seconds
- used the new function to put time limits on agent search llm calls. We believe the old timeouts only trigger if the connection is lost for that amount of time; the new ones are "absolute", i.e. if the call runs longer than the specified amount of time, the function is exited and an error is raised.

## How Has This Been Tested?

Passed unit tests, added unit tests for new timeout utility, ran in UI and confirmed that a) fast things ran with no problem and b) things exceeding the timeout were ended early


backend/onyx/agents/agent_search/deep_search/main/nodes/extract_entities_terms.py
```diff
@@ -84,9 +88,11 @@ def extract_entities_terms(
     ]
     fast_llm = graph_config.tooling.fast_llm
     # Grader
-    llm_response = fast_llm.invoke(
+    llm_response = run_with_timeout(
```
yuhongsun96: Do we think it's feasible to continue with this step timing out? I imagine it's ok if this portion fails right? We can hopefully still complete the flow?

evan-onyx: valid, just addressed this

backend/onyx/agents/agent_search/deep_search/main/nodes/create_refined_sub_questions.py
```diff
@@ -134,15 +138,17 @@ def create_refined_sub_questions(
     agent_error: AgentErrorLog | None = None
     streamed_tokens: list[BaseMessage_Content] = []
     try:
-        streamed_tokens = dispatch_separated(
+        streamed_tokens = run_with_timeout(
+            AGENT_TIMEOUT_LLM_REFINED_SUBQUESTION_GENERATION,
+            dispatch_separated,
             model.stream(
```
yuhongsun96, line 144: this doesn't need to be a stream right? We don't seem to stream it back anyway

evan-onyx: Not anymore, but imo it's better to keep it this way because it's the same way the initial subquestions are done, so no need to write extra code. Also makes it easy to go back to showing refined stuff streaming if we ever want to do that/ api consumers want it

yuhongsun96: seems good

yuhongsun96 (approved):
lgtm, well done

PR #4024, opened 2025-02-17
Force use tool overrides

evan-danswer:
## Description

When the user has a query override provided through the api, agent search should respect it

## How Has This Been Tested?

in UI


backend/onyx/agents/agent_search/basic/graph_builder.py
```diff
@@ -33,13 +33,13 @@ def basic_graph_builder() -> StateGraph:
...
-        action=llm_tool_choice,
+        node="choose_tool",
+        action=choose_tool,
     )
 
     graph.add_node(
-        node="tool_call",
-        action=tool_call,
+        node="call_tool",
+        action=call_tool,
```
joachim-danswer, line 42: I like that for consistency.

backend/onyx/tools/tool_implementations/search/search_tool.py
```diff
@@ -58,6 +58,7 @@
 SEARCH_DOC_CONTENT_ID = "search_doc_content"
 SECTION_RELEVANCE_LIST_ID = "section_relevance_list"
 SEARCH_EVALUATION_ID = "llm_doc_eval"
+QUERY_FIELD = "query"
```
joachim-danswer, line 61: Am I missing something here? (Could well be...).
Here, QUERY_FIELD is a string, but below it is a dict?

joachim-danswer (approved):
Looks good.

backend/onyx/tools/tool_implementations/search/search_tool.py
evan-onyx, line 61: its a str!

PR #4029, opened 2025-02-17
Contextual Retrieval

evan-danswer:
## Description

Addresses https://linear.app/danswer/issue/DAN-1504/context-rag

- Contextual Retrieval Search setting in UI
- Adds chunk and document summaries to each chunk in vespa
- uses the above information for retrieval only
- Model selection and cost estimation per 1M tokens
  - TODO: cost estimation based on size of current vespa index

## How Has This Been Tested?

tested in UI, unit tests


backend/onyx/indexing/chunker.py
```diff
@@ -309,7 +350,7 @@ def _chunk_document_with_sections(
                 continue
 
             # CASE 2: Normal text section
-            section_token_count = len(self.tokenizer.tokenize(section_text))
+            section_token_count = len(self.tokenizer.encode(section_text))
```
yuhongsun96, line 334: Why do we want to use encode? Are the added tokens valuable or more correct to include?

backend/onyx/indexing/embedder.py
```diff
@@ -121,7 +121,7 @@ def embed_chunks(
             if chunk.large_chunk_reference_ids:
                 large_chunks_present = True
             chunk_text = (
-                f"{chunk.title_prefix}{chunk.content}{chunk.metadata_suffix_semantic}"
+                f"{chunk.title_prefix}{chunk.doc_summary}{chunk.content}{chunk.chunk_context}{chunk.metadata_suffix_semantic}"
```
yuhongsun96, line 124: Is it safer to pass around a None rather than an empty string for this?

backend/alembic/versions/8e1ac4f39a9f_enable_contextual_retrieval.py
```diff
@@ -0,0 +1,50 @@
...
+Create Date: 2024-12-20 13:29:09.918661
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "8e1ac4f39a9f"
+down_revision = "3781a5eb12cb"
```
yuhongsun96, line 14: likely needs an update

web/src/app/admin/embeddings/pages/AdvancedEmbeddingFormPage.tsx
```diff
@@ -190,6 +258,56 @@ const AdvancedEmbeddingFormPage = forwardRef<
...
+                                )
+                              : "Cost information not available"}
+                          </span>
+                        </>
+                      ) : (
+                        "Loading cost information..."
+                      )}
+                    </div>
+                  )}
+              </div>
```
yuhongsun96, line 338: Fix reindex button (make it invalid until user selects a model)

yuhongsun96 (approved):
- How does it handle mini chunks and large chunks, make sure this is thought about and tested
- Contextualizing step in the pipeline instead of having it as part of chunking (refer to our discussion together to make best decision)

PR #4031, opened 2025-02-18
Non-SMTP password reset

pablonyx:
## Description

Adds abilitiy for:
- Users to change their password if they remember their current one + are using basic auth loginc
- Admins to reset a user's password to a random one if they are using basic auth login.

https://github.com/user-attachments/assets/5e146781-5ee2-48a9-bf1e-0e65c1a03f46

## How Has This Been Tested?

Locally with multi tenant + single tenant


backend/onyx/server/features/password/api.py
```diff
@@ -0,0 +1,56 @@
...
+        await user_manager.change_password_if_old_matches(
+            user=current_user,
+            old_password=form_data.old_password,
+            new_password=form_data.new_password,
+        )
+    except InvalidPasswordException as e:
+        raise HTTPException(status_code=400, detail=str(e.reason))
+    except Exception as e:
+        raise HTTPException(
+            status_code=500, detail=f"An unexpected error occurred: {str(e)}"
```
rkuo-danswer, line 40: using f"" and str() seems redundant.

backend/onyx/server/features/password/api.py
```diff
@@ -0,0 +1,56 @@
...
+        )
+
+
+@router.post("/reset_password")
+async def admin_reset_user_password(
+    user_reset_request: UserResetRequest,
+    user_manager: UserManager = Depends(get_user_manager),
+    db_session: Session = Depends(get_session),
+    _: User = Depends(current_admin_user),
+) -> UserResetResponse:
```
rkuo-danswer, line 50: let's try to document these endpoints more fully via docstrings as moving forward using tools like openapi will probably be a means of self documenting/supporting our api's.

rkuo-danswer (approved):

rkuo-danswer (approved):

PR #4032, opened 2025-02-18
Admin default

pablonyx:
## Description

For auto scroll and default temperature enabled workspace settings:
- If user has no specified option, default to global workspace settings
- Otherwise, use user's specified option

## How Has This Been Tested?


rkuo-danswer (approved):
Found the inability to revert back to or know what the system default is unintuitive. For temperature setting, we allow the user to set this but then it's really quite undiscoverable ... better help text?  Maybe some comments on notes on this will be helpful.

PR #4035, opened 2025-02-18
Content of .xlsl are not properly read during indexing.

ktjayamanna:
## Description
Problem:
Based on the Slack thread: https://onyx-dot-app.slack.com/archives/C056265VB1N/p1739853716830199?thread_ts=1733379700.624259&cid=C056265VB1N
Spreadsheets that are uploaded to Google Drive (which get a MIME of `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet`) end up having empty content, resulting inaccurate search results.
Solution:
Download the file and use pandas to extract the data and apply the same transforms as if it is a `application/vnd.google-apps.spreadsheet`.

## How Has This Been Tested?
Smoke tested the following and all passed.

- Created a Google Drive connector for a specific folder with a mix of `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet` and  `application/vnd.google-apps.spreadsheet` files which was then connected to a bot and it correctly performed search.
- Updated the test `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet` file and verified the search results were uptodate when `Run Update` as well as `Run complete re-indexing` were performed.

### DIFF
--- a/backend/onyx/connectors/google_drive/doc_conversion.py
+++ b/backend/onyx/connectors/google_drive/doc_conversion.py
@@ -4,7 +4,7 @@ from datetime import datetime
 from datetime import timezone
 from tempfile import NamedTemporaryFile
 
-import pandas as pd
+import openpyxl
 from googleapiclient.discovery import build  # type: ignore
 from googleapiclient.errors import HttpError  # type: ignore
 
@@ -131,21 +131,28 @@ def _extract_sections_basic(
                     tmp.write(response)
                     tmp_path = tmp.name
 
-                xls = pd.ExcelFile(tmp_path)
-                sections = []
-                for sheet_name in xls.sheet_names:
-                    df = pd.read_excel(xls, sheet_name=sheet_name, header=None)
-                    df.dropna(how="all")
-                    values = [df.columns.tolist()] + df.fillna("").astype(
-                        str
-                    ).values.tolist()
+                TEXT_SECTION_SEPARATOR = "\n\n"
 
-                    if values:
-                        text = f"Sheet: {sheet_name}\n"
-                        for row in values:
-                            text += "\t".join(str(cell) for cell in row) + "\n"
-                        sections.append(Section(link=link, text=text))
+                workbook = openpyxl.load_workbook(tmp_path, read_only=True)
+
+                sections = []
+
+                for sheet in workbook.worksheets:
+                    sheet_string = TEXT_SECTION_SEPARATOR.join(
+                        ",".join(map(str, row))
+                        for row in sheet.iter_rows(min_row=1, values_only=True)
+                    )
+
+                    if sheet_string.strip():
+                        sections.append(
+                            Section(
+                                link=link,
+                                text=f"Sheet: {sheet.title}\n\n{sheet_string}",
+                            )
+                        )
+
                 os.remove(tmp_path)
+
                 return sections
 
             except Exception as e:
--- a/backend/onyx/connectors/google_drive/doc_conversion.py
+++ b/backend/onyx/connectors/google_drive/doc_conversion.py
@@ -1,7 +1,10 @@
 import io
+import os
 from datetime import datetime
 from datetime import timezone
+from tempfile import NamedTemporaryFile
 
+import pandas as pd
 from googleapiclient.discovery import build  # type: ignore
 from googleapiclient.errors import HttpError  # type: ignore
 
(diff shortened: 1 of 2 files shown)


backend/onyx/connectors/google_drive/doc_conversion.py
```diff
@@ -131,21 +131,28 @@ def _extract_sections_basic(
                     tmp.write(response)
                     tmp_path = tmp.name
 
-                xls = pd.ExcelFile(tmp_path)
+                TEXT_SECTION_SEPARATOR = "\n\n"
+
+                workbook = openpyxl.load_workbook(tmp_path, read_only=True)
```
ktjayamanna: Emulated  file connector excel file read implementation [here](https://github.com/onyx-dot-app/onyx/blob/1f2af373e1236f15058ebb5f6dd10e6551cca10e/backend/onyx/file_processing/extract_file_text.py#L304).
notes:
I did not just import this function because the ramifications of cramming all the sheet tabs into one Section instance was ambiguous. Wanted to stick to the original return type of the function so it's less likely that the code would break.

backend/onyx/connectors/google_drive/doc_conversion.py
```diff
@@ -131,21 +131,28 @@ def _extract_sections_basic(
...
+
                 sections = []
-                for sheet_name in xls.sheet_names:
-                    df = pd.read_excel(xls, sheet_name=sheet_name, header=None)
-                    df.dropna(how="all")
```
ktjayamanna: Because openpyxl does not natively support dropping null values, we will have to do this with good old slow loops which is expensive compared to an optimized lib like Pandas or Fireducks. What are you thoughts?
Current solution is do nothing but it makes the search results look bad. Attached a screenshot as well.  But we could do something like,
`  sheet_string = TEXT_SECTION_SEPARATOR.join(
                        ",".join(str(cell).strip() for cell in row if cell is not None and str(cell).strip())
                        for row in sheet.iter_rows(min_row=1, values_only=True)
                        if any(cell is not None and str(cell).strip() for cell in row)  # Skip empty rows
                    )` 
                    to filter null values but less optimal than Pandas. Thoughts?
ugly-search-reults

backend/onyx/connectors/google_drive/doc_conversion.py
```diff
@@ -43,12 +46,22 @@ def _extract_sections_basic(
 ) -> list[Section]:
     mime_type = file["mimeType"]
     link = file["webViewLink"]
+    supported_file_types = set(item.value for item in GDriveMimeType)
```
yuhongsun96, line 48: Prefer to have data models contain the full exhaustive list of the enum values

backend/onyx/connectors/google_drive/doc_conversion.py
```diff
@@ -109,7 +122,49 @@ def _extract_sections_basic(
                     f"Ran into exception '{e}' when pulling data from Google Sheet '{file['name']}'."
                     " Falling back to basic extraction."
                 )
+        # ---------------------------
+        # Microsoft Excel (.xlsx or .xls) extraction branch
+        elif mime_type in excel_file_types:
+            try:
+                response = service.files().get_media(fileId=file["id"]).execute()
+                with NamedTemporaryFile(suffix=".xlsx", delete=False) as tmp:
```
yuhongsun96: this can wrap more so that in cases of error, it gets cleaned up

backend/onyx/connectors/google_drive/doc_conversion.py
```diff
@@ -1,7 +1,10 @@
 import io
+import os
 from datetime import datetime
 from datetime import timezone
+from tempfile import NamedTemporaryFile
 
+import openpyxl
```
yuhongsun96: Recommend installing mypy for type checking as you develop. Here's the dev env setup we use internally: https://docs.google.com/document/d/1Zg5XjdAm4551-XvUot6MtzUCxQ9LUhd60nu5Ubex9CI/edit?usp=sharing

backend/onyx/connectors/google_drive/doc_conversion.py
ktjayamanna: Got it! will ensure that the temporary file is always deleted, even if an error occurs while processing the workbook.

backend/onyx/connectors/google_drive/doc_conversion.py
ktjayamanna: Can you tell me more about this? Are you thinking about #49 to be something like
```
supported_file_types = {
    "application/vnd.google-apps.document",
    "application/vnd.google-apps.spreadsheet",
    # ... all other specific mime types listed explicitly
}
```
just FYI, OG code was something like [this](https://github.com/onyx-dot-app/onyx/blob/ce2b4de24585fae2cb4bf9a303815663982f7ef3/backend/onyx/connectors/google_drive/doc_conversion.py#L47)

yuhongsun96 (approved):
Nice!

PR #4052, opened 2025-02-19
Agent search history displayed answer

evan-danswer:
## Description

https://linear.app/danswer/issue/DAN-1464/use-refined-answer-if-improvement-in-history-in-agent-search

Use the refined answer as message history when it is better, otherwise stick with the original answer

## How Has This Been Tested?

to be tested by joachim


joachim-danswer (approved):
checked it

PR #4054, opened 2025-02-19
Pin xmlsec version + improve SAML flow

Weves:
## Description

^

## How Has This Been Tested?

Tested SAML locally + at `https://saml.danswer.dev`

### DIFF
--- a/deployment/data/nginx/app.conf.template
+++ b/deployment/data/nginx/app.conf.template
@@ -4,6 +4,24 @@ log_format custom_main '$remote_addr - $remote_user [$time_local] "$request" '
                 '"$http_user_agent" "$http_x_forwarded_for" '
                 'rt=$request_time';
 
+# Map X-Forwarded-Proto or fallback to $scheme
+map $http_x_forwarded_proto $forwarded_proto {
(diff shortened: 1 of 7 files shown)


deployment/data/nginx/app.conf.template
```diff
@@ -4,6 +4,24 @@ log_format custom_main '$remote_addr - $remote_user [$time_local] "$request" '
                 '"$http_user_agent" "$http_x_forwarded_for" '
                 'rt=$request_time';
 
+# Map X-Forwarded-Proto or fallback to $scheme
```
Weves, line 7: Needed to pass these through from the ssl handler all the way through to the final call to the web_server / api_server

deployment/data/nginx/app.conf.template
```diff
@@ -21,8 +39,7 @@ upstream web_server {
 }
 
 server {
-    listen 80;
-    server_name ${DOMAIN};
+    listen 80 default_server;
```
Weves, line 42: DOMAIN isn't actually necessary - no need to add this additional config step

rkuo-danswer (approved):

PR #4074, opened 2025-02-20
thread utils respect contextvars

evan-danswer:
## Description

https://linear.app/danswer/issue/DAN-1471/parallelism-should-respect-contextvars

Our utils for running functions with thread-level parallelism didn't pass contextvars through, i.e. tenant id. Now they do! Should be safe to use with i.e. acquiring a db session, although in most cases there isn't a need to acquire sessions in parallel.

## How Has This Been Tested?

New unit tests; they fail without the code changes

### DIFF
--- a/backend/onyx/utils/threadpool_concurrency.py
+++ b/backend/onyx/utils/threadpool_concurrency.py
@@ -1,3 +1,4 @@
+import contextvars
 import threading
 import uuid
 from collections.abc import Callable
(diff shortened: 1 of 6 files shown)


backend/onyx/utils/threadpool_concurrency.py
```diff
@@ -46,7 +43,7 @@ def run_functions_tuples_in_parallel(
     results = []
     with ThreadPoolExecutor(max_workers=workers) as executor:
         future_to_index = {
-            executor.submit(func, *args): i
+            executor.submit(contextvars.copy_context().run, func, *args): i
```
pablonyx, line 49: Seems worth adding a comment here explaining why we need the contextvars to be persisted. Also perhaps a cautionary note on performance impact if warranted (though probably not needed)

pablonyx (approved):

PR #4081, opened 2025-02-21
faster indexing status at scale plus minor cleanups

evan-danswer:
## Description

https://linear.app/danswer/issue/DAN-1477/fix-slowness-of-manageadminconnectorindexing-status

Improve the performance of the /manage/admin/connector/indexing-status endpoint.

Main optimizations:
 - eager loading of relationships
 - letting sql handle loops (i.e. moving db queries out of for loops)
 - thread level parallelism

## How Has This Been Tested?

tested in UI: ~2000 connectors and 10mil index attempts yields 3.5-5.5s endpoint runtimes. roughly split by 1/3 first batch of parallel queries, 1/3 second batch of parallel queries, 1/3 remaining code.

### DIFF
--- a/backend/onyx/db/connector_credential_pair.py
+++ b/backend/onyx/db/connector_credential_pair.py
@@ -9,6 +9,7 @@ from sqlalchemy import Select
 from sqlalchemy import select
 from sqlalchemy.orm import aliased
 from sqlalchemy.orm import joinedload
+from sqlalchemy.orm import selectinload
 from sqlalchemy.orm import Session
 
 from onyx.configs.app_configs import DISABLE_AUTH
@@ -106,13 +107,17 @@ def get_connector_credential_pairs_for_user(
     eager_load_credential: bool = False,
     eager_load_user: bool = False,
 ) -> list[ConnectorCredentialPair]:
+    if eager_load_user:
+        assert (
+            eager_load_credential
+        ), "eager_load_credential must be True if eager_load_user is True"
     stmt = select(ConnectorCredentialPair).distinct()
 
     if eager_load_connector:
-        stmt = stmt.options(joinedload(ConnectorCredentialPair.connector))
+        stmt = stmt.options(selectinload(ConnectorCredentialPair.connector))
 
     if eager_load_credential:
-        load_opts = joinedload(ConnectorCredentialPair.credential)
+        load_opts = selectinload(ConnectorCredentialPair.credential)
         if eager_load_user:
             load_opts = load_opts.joinedload(Credential.user)
         stmt = stmt.options(load_opts)
@@ -187,6 +192,9 @@ def get_cc_pair_groups_for_ids(
     return list(db_session.scalars(stmt).all())
 
 
+# For use with our thread-level parallelism utils. Note that any relationships
+# you wish to use MUST be eagerly loaded, as the session will not be available
+# after this function to allow lazy loading.
 def get_cc_pair_groups_for_ids_parallel(
     cc_pair_ids: list[int],
 ) -> list[UserGroup__ConnectorCredentialPair]:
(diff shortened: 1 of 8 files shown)


backend/onyx/db/connector_credential_pair.py
```diff
@@ -151,6 +187,13 @@ def get_cc_pair_groups_for_ids(
     return list(db_session.scalars(stmt).all())
 
 
+def get_cc_pair_groups_for_ids_parallel(
```
Weves, line 198: nit: would be good to add a similar comment to as above.

backend/onyx/db/connector_credential_pair.py
```diff
@@ -98,17 +103,48 @@ def get_connector_credential_pairs_for_user(
     get_editable: bool = True,
     ids: list[int] | None = None,
     eager_load_connector: bool = False,
+    eager_load_credential: bool = False,
+    eager_load_user: bool = False,
```
Weves, line 108: this arg only seems valid if `eager_load_credential` is set. We should probably fail loudly if the user of this function passes in an invalid set of args

backend/onyx/db/index_attempt.py
```diff
@@ -395,7 +414,50 @@ def get_latest_index_attempts(
...
+    if eager_load_cc_pair:
+        stmt = stmt.options(
+            joinedload(IndexAttempt.connector_credential_pair),
+            joinedload(IndexAttempt.error_rows),
+        )
+
+    return db_session.execute(stmt).scalars().unique().all()
+
+
+def get_latest_index_attempts_parallel(
```
Weves, line 432: nit : add same comment as for other parallel calls

backend/onyx/db/document.py
```diff
@@ -229,12 +230,12 @@ def get_document_connector_counts(
 
 
 def get_document_counts_for_cc_pairs(
-    db_session: Session, cc_pair_identifiers: list[ConnectorCredentialPairIdentifier]
+    db_session: Session, cc_pairs: list[ConnectorCredentialPair]
```
Weves: nit: prefer to follow the principle of "pass in the minimal required set of information to make a function easier to call". In other words, leaving it as is 😅

backend/onyx/server/documents/connector.py
```diff
@@ -632,23 +640,32 @@ def get_connector_indexing_status(
...
+    # see https://stackoverflow.com/questions/75758327/
+    # sqlalchemy-method-connection-for-bind-is-already-in-progress
+    # for why we can't pass in the current db_session to these functions
+    (
+        cc_pairs,
+        latest_index_attempts,
+        latest_finished_index_attempts,
+    ) = run_functions_tuples_in_parallel(
+        [
+            (
```
Weves, line 653: can we add a comment as to what the goals of each of these? Since we don't have arg names its a bit hard to tell what the goal of each of these calls is (especially the last two).

backend/onyx/server/documents/models.py
```diff
@@ -92,9 +94,8 @@ def from_connector_db_model(cls, connector: Connector) -> "ConnectorSnapshot":
             connector_specific_config=connector.connector_specific_config,
             refresh_freq=connector.refresh_freq,
             prune_freq=connector.prune_freq,
-            credential_ids=[
```
Weves, line 95: nit: prefer the following, feels easier to read

```
credential_ids=(
    credential_ids
    or [association.credential.id for association in connector.credentials]
)
```

backend/onyx/server/documents/connector.py
```diff
@@ -613,6 +621,9 @@ def get_connector_indexing_status(
         False, description="If true, return editable document sets"
     ),
 ) -> list[ConnectorIndexingStatus]:
+    import time
```
Weves: should probably remove this timing?

backend/onyx/server/documents/connector.py
```diff
@@ -751,6 +802,8 @@ def get_connector_indexing_status(
         properties=None,
         db_session=db_session,
     )
+    end_time = time.time()
```
Weves: same here

Weves (approved):

PR #4084, opened 2025-02-21
coerce useLanggraph to boolean

evan-danswer:
## Description

minor fix for use_agentic_search param type

## How Has This Been Tested?

n/a

### DIFF
--- a/web/src/app/chat/lib.tsx
+++ b/web/src/app/chat/lib.tsx
@@ -236,7 +236,7 @@ export async function* sendMessage({
           }
         : null,
     use_existing_user_message: useExistingUserMessage,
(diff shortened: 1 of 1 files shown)


web/src/app/chat/lib.tsx
```diff
@@ -236,7 +236,7 @@ export async function* sendMessage({
           }
         : null,
     use_existing_user_message: useExistingUserMessage,
-    use_agentic_search: useLanggraph,
+    use_agentic_search: Boolean(useLanggraph),
```
pablonyx: Would prefer ?? false unless we have a good reason to use `Boolean`

pablonyx (approved):

PR #4089, opened 2025-02-21
Fix gpt o-series code block formatting

evan-danswer:
## Description

Addresses https://linear.app/danswer/issue/DAN-1479/gpt-o-series-code-blocks

Added to the default system prompt construction to tell the model to use markdown code blocks. To get o3 mini to work, I had to explicitly specify the use of backticks

## How Has This Been Tested?

Tested in UI; without the changes Our code block parser often gets some of it right but misses parts that lead to unreadable code and comments that are interpreted as markdown headings.

### DIFF
--- a/backend/onyx/chat/prompt_builder/answer_prompt_builder.py
+++ b/backend/onyx/chat/prompt_builder/answer_prompt_builder.py
@@ -41,7 +41,7 @@ def default_build_system_message(
         and llm_config.model_provider == OPENAI_PROVIDER_NAME
         and llm_config.model_name.startswith("o")
     ):
-        system_prompt += "\n" + CODE_BLOCK_MARKDOWN
+        system_prompt = CODE_BLOCK_MARKDOWN + system_prompt
     tag_handled_prompt = handle_onyx_date_awareness(
         system_prompt,
         prompt_config,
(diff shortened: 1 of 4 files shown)


backend/onyx/chat/prompt_builder/answer_prompt_builder.py
```diff
@@ -31,8 +33,15 @@
 
 def default_build_system_message(
     prompt_config: PromptConfig,
+    llm_config: LLMConfig | None = None,
```
Weves: is there a place where we wouldn't want to pass this in?

evan-onyx: no longer defaulted o7

backend/onyx/chat/prompt_builder/answer_prompt_builder.py
```diff
@@ -31,8 +33,14 @@
...
 def default_build_system_message(
     prompt_config: PromptConfig,
+    llm_config: LLMConfig,
 ) -> SystemMessage | None:
     system_prompt = prompt_config.system_prompt.strip()
+    if (
+        llm_config.model_provider == OPENAI_PROVIDER_NAME
+        and llm_config.model_name.startswith("o")
+    ):
+        system_prompt = CODE_BLOCK_MARKDOWN + system_prompt
```
Weves, line 45: can we add a comment w/ a link to the relevant article/announcement?

Weves (approved):

PR #4094, opened 2025-02-22
ratelimited slack client.conversations_list poorman solution

nzin-alayacare:
## Description

When configuring a Slack Bot, Onyx try to list all public/private slack channels, to let you propose on which slack channel(s) you want to configure Onyx.
The problem is that on big Slack workspace (in my case there are > 4000 public/private slack channels), the `client.conversations_list` returns very quickly the first ~ 1000 slack channels, but face a rate limiting issue after that.

The proposed solution is not a prodution-ready solution, but it is a "poor man" solution: 
- if the time to get back the list of all channels is taking more than 10s we failed (we returns a 202)
-  BUT we start a background thread, to continue a "cold start" job to collect all channels
- when we have finished to collect all channels, we will be able to serve the saved-for-later list, next time someone ask for all channels

## How Has This Been Tested?

It was tested on our company workspace, which contains > 4000 slack channels. It "works" but the cold start takes ~ 30 minutes to get the list of all channels.

### DIFF
--- a/backend/onyx/server/manage/slack_bot.py
+++ b/backend/onyx/server/manage/slack_bot.py
@@ -6,6 +6,8 @@ from fastapi import HTTPException
 from slack_sdk import WebClient
 from slack_sdk.errors import SlackApiError
 from sqlalchemy.orm import Session
+import time
+import threading
 
 from onyx.auth.users import current_admin_user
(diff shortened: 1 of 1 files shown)


nzin-alayacare:
cf https://github.com/onyx-dot-app/onyx/issues/4095

PR #4101, opened 2025-02-23
Fix starter message editing

pablonyx:
## Description

We expose name / message as a single value for end-users (built in starter messages can modify both). Currently, updated message doesn't update name– should both have same value.

## How Has This Been Tested?


evan-danswer (approved):
LGTM as long as you tested, I don't have context on redirectType

Issue #4095, opened 2025-02-22
ratelimited Slack bot issue


nzin-alayacare:
When trying to install the slack bot on a given Slack channel, the "Select a Slack Channel" fails and behind the scene receive a "ratelimited" error message from the API server

<img width="902" alt="Image" src="https://github.com/user-attachments/assets/fce21708-64bb-454a-aa24-99855adf24de" />

The issue is that our Slack workspace contains > 4000 public/private slack channels, and the slack `conversations_list` API is really slow after the first ~ 1000 channels.

A first poor man solution is proposed in https://github.com/onyx-dot-app/onyx/pull/4094 , but I think more work is needed like
- having a cold start, that is 1/ started when Onyx is started, 2/ saved in a DB to avoid to have this cold start again when Onyx is started
- or dont try to list all slack channels (if there are more than 1000 slack channels?), but ask the customer to enter manually the slack channel?

pablonyx:
That is quite a few Slack channels! We'll take a look at your PR and incorporate a solution from there. We appreciate you bringing this to our attention.

PR #4102, opened 2025-02-23
Fix slackbot rate limit on selection of channels

pablonyx:
## Description

Fixes https://github.com/onyx-dot-app/onyx/issues/4095

Very slow API from Slack for listing channels- if >200 channels, let users input their channel name

## How Has This Been Tested?

With slack bot configuration
- Mock having more than 200 channels


evan-danswer (approved):
Why can't we index an arbitrary number of channels? Might include a comment explaining that, otherwise LGTM

PR #4122, opened 2025-02-25
Update tenant logic

pablonyx:
## Description

Previously, we relied on the `tenant_id` being `None` in order to determine whether or not to proceed with various tenancy-specific processes. This led to some confusion as the "tenant" in the single-tenant case is `POSTGRES_DEFAULT_SCHEMA`, at least in the sense that that's their search path. This confusion was compounded by the recent change to the codebase which ensured we always assume the `CURRENT_TENANT_ID_CONTEXTVAR` is set to a non-null value when we grab a db / redis session.

This change mitigates that confusion by ensuring tenant_id is typed as a string, and introduce a new standard where we rely directly on the `MULTI_TENANT` config for determining if we are in the multi-tenant case or not

## How Has This Been Tested?

### DIFF
--- a/backend/onyx/auth/api_key.py
+++ b/backend/onyx/auth/api_key.py
@@ -10,6 +10,7 @@ from pydantic import BaseModel
 
 from onyx.auth.schemas import UserRole
 from onyx.configs.app_configs import API_KEY_HASH_ROUNDS
+from shared_configs.configs import MULTI_TENANT
 
 
 _API_KEY_HEADER_NAME = "Authorization"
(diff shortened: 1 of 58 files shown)


backend/onyx/auth/api_key.py
```diff
@@ -35,8 +36,7 @@ class ApiKeyDescriptor(BaseModel):
 
 
 def generate_api_key(tenant_id: str | None = None) -> str:
-    # For backwards compatibility, if no tenant_id, generate old style key
-    if not tenant_id:
+    if not MULTI_TENANT or not tenant_id:
```
evan-onyx, line 39: isn't it bad when MULTI_TENANT is True and tenant_id is None?

backend/onyx/db/engine.py
```diff
@@ -360,7 +360,7 @@ def _set_search_path(
     """Every time a new transaction is started,
     set the search_path from the session's info."""
     tenant_id = session.info.get("tenant_id")
-    if tenant_id:
+    if tenant_id and MULTI_TENANT:
```
evan-onyx: also seems like a weird state if MULTI_TENANT is False and tenant_id is set? Or maybe now tenant_id is Always set so we should assert it?

evan-danswer (approved):
LGTM with minor qs, what a fun PR

Issue #3164, opened 2024-11-19
Google Drive connector : do not follow shortcuts anymore


mathieu-coingt:
Since v0.12.0 release and the "new" Google Drive form, the option to follow shortcut has disappeared, and so at the end the crawling do not follow shortcut.

It's very inconvenient since documents needed could come from differents users or shared drive into the same Google Workspace, and centralized thanks to shortcuts into dedicated folders, segmented by scope (and so at the end multiple connectors and datasets).

Tested also on v0.13.0-beta.1

bharath-muthineni:
Danswer team, I think shortcuts was a great feature. 
Would like to know why this was removed in recent releases. 
Would love to see the shortcuts feature back.

mikeacook:
I also found this feature really useful... and removing it basically made Onyx totally impossible to use for my company.

mikeacook:
I just pushed up a PR containing the workaround I've implemented while we wait for official support from the maintainers.

bharath-muthineni:
Hey @mikeacook , would you mind closing the PR that you have raised. 😅

evan-onyx:
We added this in #12176

PR #4124, opened 2025-02-25
feat: implements support for following Google Drive shoftcuts by default

mikeacook:
## Description

This is an incomplete change, in that it implements _always_ trying to follow Google Drive shortcuts by default. I'm currently using this as a workaround for https://github.com/onyx-dot-app/onyx/issues/3164, and sharing to help unblock others frustrated by the same issue.

## How Has This Been Tested?

Ran on our local self-host

### DIFF
--- a/backend/onyx/connectors/google_utils/google_utils.py
+++ b/backend/onyx/connectors/google_utils/google_utils.py
@@ -96,6 +96,14 @@ def execute_paginated_retrieval(
         if next_page_token:
             request_kwargs["pageToken"] = next_page_token
 
+        # Add fields for shortcut detection if not already specified
+        if list_key == "files":
+            if "fields" in request_kwargs:
+                if "shortcutDetails" not in request_kwargs["fields"]:
+                    request_kwargs["fields"] += ", files/shortcutDetails"
+                else:
+                    request_kwargs["fields"] = "nextPageToken, files/shortcutDetails, *"
+
         try:
             results = retrieval_function(**request_kwargs).execute()
         except HttpError as e:
@@ -120,6 +128,25 @@ def execute_paginated_retrieval(
         next_page_token = results.get("nextPageToken")
         if list_key:
             for item in results.get(list_key, []):
-                yield item
+                if (item.get("shortcutDetails") and
+                    item.get("mimeType") == "application/vnd.google-apps.shortcut"):
+                    # If this is a shortcut, fetch the target file
+                    try:
+                        target_file = retrieval_function.__self__.get(
+                            fileId=item["shortcutDetails"]["targetId"],
+                            fields="id, name, mimeType, parents",
+                        ).execute()
+                        # Preserve the original file's location metadata
+                        target_file["parents"] = item.get("parents", [])
+                        target_file["webViewLink"] = item.get("webViewLink", "")
+                        target_file["modifiedTime"] = item.get("modifiedTime", "")
+                        yield target_file
+                    except HttpError as e:
+                        if continue_on_404_or_403 and e.resp.status in (403, 404):
+                            logger.debug(f"Error resolving shortcut target: {e}")
+                            continue
+                        raise
+                else:
+                    yield item
         else:
             yield results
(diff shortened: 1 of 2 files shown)


backend/onyx/connectors/google_utils/google_utils.py
```diff
@@ -96,6 +96,14 @@ def execute_paginated_retrieval(
         if next_page_token:
             request_kwargs["pageToken"] = next_page_token
 
+        # Add fields for shortcut detection if not already specified
+        if list_key == "files":
+            if "fields" in request_kwargs:
```
evan-onyx, line 101: Ideally I'd prefer not to have this if-block at all; files(..., shortcutDetails) is already passed in above via `FILE_FIELDS` and now `SLIM_FILE_FIELDS` as far as I can tell. I've never seen this `files/shortcutDetails` as part of the fields param, if that's necessary could you link me to the documentation?

backend/onyx/connectors/google_utils/google_utils.py
```diff
@@ -120,6 +128,25 @@ def execute_paginated_retrieval(
         next_page_token = results.get("nextPageToken")
         if list_key:
             for item in results.get(list_key, []):
-                yield item
+                if (item.get("shortcutDetails") and
```
evan-onyx, line 131: This shouldn't be in execute_paginated_retrieval, as it is google drive specific. I'd recommend writing a wrapper in file_retrieval.py that intercepts each `file` yielded by `execute_paginated_retrieval` and does effectively what you're doing here (it should also be a bit nicer since at that level you can directly use the drive resource to make the requests rather than needing `__self__` 🙂)

backend/onyx/connectors/google_utils/google_utils.py
```diff
@@ -120,6 +128,25 @@ def execute_paginated_retrieval(
...
+                    item.get("mimeType") == "application/vnd.google-apps.shortcut"):
+                    # If this is a shortcut, fetch the target file
+                    try:
+                        target_file = retrieval_function.__self__.get(
+                            fileId=item["shortcutDetails"]["targetId"],
+                            fields="id, name, mimeType, parents",
+                        ).execute()
+                        # Preserve the original file's location metadata
+                        target_file["parents"] = item.get("parents", [])
```
evan-onyx, line 140: If you expect that these fields are always present (parents, webviewLink, modifiedTime) then we prefer using hard brackets and thus failing loudly when they don't exist.

backend/onyx/connectors/google_utils/google_utils.py
evan-onyx: Maybe something like paginated_retrieval_follow_shortcuts ?

backend/onyx/connectors/google_utils/google_utils.py
```diff
@@ -120,6 +128,25 @@ def execute_paginated_retrieval(
...
         if list_key:
             for item in results.get(list_key, []):
-                yield item
+                if (item.get("shortcutDetails") and
+                    item.get("mimeType") == "application/vnd.google-apps.shortcut"):
+                    # If this is a shortcut, fetch the target file
+                    try:
+                        target_file = retrieval_function.__self__.get(
+                            fileId=item["shortcutDetails"]["targetId"],
+                            fields="id, name, mimeType, parents",
```
evan-onyx, line 137: Unless there's a really good reason for it, I'd strongly prefer getting the same FILE_FIELDS that we get in file_retrieval.py (this should be easier to do after the above change is made too)

evan-danswer (changes requested):
Thanks so much for the contribution! A few structural changes, but broadly looks good and we appreciate your effort on this!

bharath-muthineni:
@evan-danswer any possibility of merging this PR?

bharath-muthineni:
Sure, and thanks @mikeacook

bharath-muthineni:
> @evan-onyx @bharath-muthineni Sorry for the delay here, I finally have some bandwidth to incorporate this feedback and finish the task.
> 
> I saw on the linked issue that @bharath-muthineni asked me to close this PR. To clarify, has "follow shortcuts" support been implemented by someone else already or is this still needed? Assuming the latter (Google Drive sync still isn't working for us with shortcuts, even on the latest build), I'm still planning to take this over the line.

Thanks a ton @mikeacook

PR #4127, opened 2025-02-26
slight improvements to user group endpoints

evan-onyx:
## Description

Customer was having long load times from user-group and persona endpoints; these changes improve performance by about ~30% in some cases (worst I saw was equivalent performance to the non-eager loading version). 

## How Has This Been Tested?

This was tested with about 10,000 users, 100 user groups, and 500 personas. 

### DIFF
--- a/backend/ee/onyx/db/user_group.py
+++ b/backend/ee/onyx/db/user_group.py
@@ -182,8 +182,10 @@ def validate_object_creation_for_user(
 
 def eager_usergroup_options(stmt: Select[tuple[UserGroup]]) -> Select[tuple[UserGroup]]:
     return stmt.options(
+        # Which users are in this group
         selectinload(UserGroup.users),
         selectinload(UserGroup.user_group_relationships),
+        # Which CC pairs this group has access to
         selectinload(UserGroup.cc_pair_relationships)
         .selectinload(UserGroup__ConnectorCredentialPair.cc_pair)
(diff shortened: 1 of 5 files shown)


backend/ee/onyx/db/user_group.py
```diff
@@ -175,6 +180,42 @@ def validate_object_creation_for_user(
         )
 
 
+def eager_usergroup_options(stmt: Select[tuple[UserGroup]]) -> Select[tuple[UserGroup]]:
+    return stmt.options(
+        selectinload(UserGroup.users),
```
Weves, line 186: can we roughly group them into sections w/ comments

backend/ee/onyx/db/user_group.py
```diff
@@ -175,6 +180,42 @@ def validate_object_creation_for_user(
...
+        .selectinload(DocumentSet.connector_credential_pairs)
+        .joinedload(ConnectorCredentialPair.connector)
+        .contains_eager(Connector.credentials),
+        selectinload(UserGroup.personas).selectinload(Persona.users),
+        selectinload(UserGroup.personas).selectinload(Persona.groups),
+        selectinload(UserGroup.personas).selectinload(Persona.labels),
+    )
+
+
 def fetch_user_group(db_session: Session, user_group_id: int) -> UserGroup | None:
```
Weves, line 225: eager load should probably be an option

Weves (approved):

PR #4129, opened 2025-02-26
Fix confluence permission syncing at scale

Weves:
## Description

Pagination wasn't working for either data center or cloud -> when you had more than 200 users you would only get a partial group sync. This fixes that.

## How Has This Been Tested?

Tested with data center with 502 users + cloud with 8 users but manually setting page size to 1.

### DIFF
--- a/backend/onyx/connectors/confluence/onyx_confluence.py
+++ b/backend/onyx/connectors/confluence/onyx_confluence.py
@@ -5,9 +5,7 @@ from collections.abc import Iterator
 from typing import Any
 from typing import cast
 from typing import TypeVar
-from urllib.parse import parse_qs
 from urllib.parse import quote
-from urllib.parse import urlparse
 
 from atlassian import Confluence  # type:ignore
 from pydantic import BaseModel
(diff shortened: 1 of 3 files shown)


backend/onyx/connectors/confluence/onyx_confluence.py
```diff
@@ -236,9 +238,41 @@ def _paginate_url(
...
+            # the configured server, it will artificially limit the amount of
+            # results returned BUT will not apply this to the start parameter.
+            # This will cause us to miss results.
+            if url_suffix and "start" in url_suffix:
+                new_start = get_start_param_from_url(url_suffix)
+                previous_start = get_start_param_from_url(old_url_suffix)
+                if new_start - previous_start > len(results):
+                    logger.warning(
+                        f"Start was updated by more than the amount of results "
+                        f"retrieved. This is a bug. Start: {new_start}, "
```
pablonyx: Might be worth clarifying here it's a bug with Confluence (i.e. not a bug in our code)

pablonyx (approved):

backend/onyx/connectors/confluence/onyx_confluence.py
Weves: Makes sense, done!

PR #4137, opened 2025-02-26
Improved chat search

pablonyx:
## Description

# PR Description: Chat Search Optimization

-  Replaced the slow LIKE-based search with PostgreSQL full-text search using tsvector columns and GIN indexes

- N.B. **COMMIT** statements in the migration are necessary because PostgreSQL requires CREATE INDEX CONCURRENTLY to run outside a transaction. Without them, the migration would fail with "CREATE INDEX CONCURRENTLY cannot run inside a transaction block"

## How Has This Been Tested?

### DIFF
--- a/backend/alembic/versions/3bd4c84fe72f_improved_index.py
+++ b/backend/alembic/versions/3bd4c84fe72f_improved_index.py
@@ -15,6 +15,18 @@ branch_labels = None
 depends_on = None
 
 
+# NOTE:
+# This migration addresses issues with the previous migration (8f43500ee275) which caused
+# an outage by creating an index without using CONCURRENTLY. This migration:
+#
+# 1. Creates more efficient full-text search capabilities using tsvector columns and GIN indexes
+# 2. Uses CONCURRENTLY for all index creation to prevent table locking
+# 3. Explicitly manages transactions with COMMIT statements to allow CONCURRENTLY to work
+# (see: https://www.postgresql.org/docs/9.4/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY)
+# (see: https://github.com/sqlalchemy/alembic/issues/277)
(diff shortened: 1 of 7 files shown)


backend/onyx/db/chat_search.py
```diff
@@ -26,127 +24,87 @@ def search_chat_sessions(
...
+        sessions = result.scalars().all()
 
-        has_more = len(chat_sessions) > page_size
+        has_more = len(sessions) > page_size
         if has_more:
-            chat_sessions = chat_sessions[:page_size]
+            sessions = sessions[:page_size]
 
-        return list(chat_sessions), has_more
+        return sessions, has_more
```
rkuo-danswer: see what list[] was removed for?

backend/alembic/versions/3bd4c84fe72f_improved_index.py
```diff
@@ -0,0 +1,72 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "3bd4c84fe72f"
+down_revision = "8f43500ee275"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # Create a GIN index for full-text search on chat_message.message
```
rkuo-danswer, line 31: add comments about why we needed this index  and the improvement (per netflix context)

rkuo-danswer (approved):

backend/onyx/db/chat_search.py
pablonyx: Looks like it was causing some typing issues!

backend/alembic/versions/3bd4c84fe72f_improved_index.py
pablonyx: Good call!

PR #4138, opened 2025-02-26
Address Auth Edge Case

pablonyx:
## How Has This Been Tested?


backend/onyx/auth/users.py
```diff
@@ -420,7 +420,7 @@ async def oauth_callback(
             except exceptions.UserNotExists:
                 try:
                     # Attempt to get user by email
-                    user = await self.get_by_email(account_email)
+                    user = cast(User, await self.user_db.get_by_email(account_email))
```
evan-onyx, line 423: Slightly prefer asserting non-None rather than casting here

evan-danswer (approved):
LGTM with small suggestion

PR #4142, opened 2025-02-27
Ephemeral slack messages as responses to user questions

joachim-danswer:
## Description

We want to enable our customers to configure an Onyx bot for a channel such that the bot is able to respond as an ephemeral message to the user who is asking the question. That provides us with the ability to leverage documents for the search that a user has private access to, not just documents that are public to everyone.  

See details in the current design doc.

Ticket: https://linear.app/danswer/issue/DAN-1489/enable-ephemeral-replies-by-answer-bot-in-group-channels

## How Has This Been Tested?

Locally:
 - set up new bot in test mode
 - configured two channels, one with the new is_ephemeral parameter=True, the other one set to false
 - for each, I sent a request to the bot from
     - the channel
     - a thread
 - in the channel with the ephemeral setting I check for both question sources (channel vs thread) the button clicks 'Shae with Channel', and 'Keep for Yourself.
 
 Key test cases:
  - are the documents properly transferred
  - are the new locations correct 

### DIFF
--- a/backend/onyx/connectors/slack/utils.py
+++ b/backend/onyx/connectors/slack/utils.py
@@ -72,6 +72,11 @@ def make_slack_api_rate_limited(
     @wraps(call)
     def rate_limited_call(**kwargs: Any) -> SlackResponse:
         last_exception = None
+
+        if "thread_ts" in kwargs:
+            if kwargs["thread_ts"] is None:
+                kwargs.pop("thread_ts", None)
+
         for _ in range(max_retries):
             try:
                 # Make the API call
--- a/backend/onyx/connectors/slack/utils.py
+++ b/backend/onyx/connectors/slack/utils.py
@@ -73,10 +73,6 @@ def make_slack_api_rate_limited(
     def rate_limited_call(**kwargs: Any) -> SlackResponse:
         last_exception = None
 
-        if "thread_ts" in kwargs:
-            if kwargs["thread_ts"] is None:
-                kwargs.pop("thread_ts", None)
-
         for _ in range(max_retries):
             try:
                 # Make the API call
--- a/backend/onyx/onyxbot/slack/blocks.py
+++ b/backend/onyx/onyxbot/slack/blocks.py
@@ -154,14 +154,26 @@ def _build_ephemeral_publication_block(
             "thread_to_respond": respond_ts,
         },
         "channel_conf": {
-            "channel_name": channel_conf.get("channel_name"),
-            "respond_tag_only": channel_conf.get("respond_tag_only"),
-            "respond_to_bots": channel_conf.get("respond_to_bots"),
-            "is_ephemeral": channel_conf.get("is_ephemeral"),
-            "respond_member_group_list": channel_conf.get("respond_member_group_list"),
-            "answer_filters": channel_conf.get("answer_filters"),
-            "follow_up_tags": channel_conf.get("follow_up_tags"),
-            "show_continue_in_web_ui": channel_conf.get("show_continue_in_web_ui"),
+            "channel_name": channel_conf.get("channel_name") if channel_conf else None,
+            "respond_tag_only": channel_conf.get("respond_tag_only")
+            if channel_conf
+            else None,
+            "respond_to_bots": channel_conf.get("respond_to_bots")
+            if channel_conf
+            else None,
+            "is_ephemeral": channel_conf.get("is_ephemeral") if channel_conf else None,
+            "respond_member_group_list": channel_conf.get("respond_member_group_list")
+            if channel_conf
+            else None,
+            "answer_filters": channel_conf.get("answer_filters")
+            if channel_conf
+            else None,
+            "follow_up_tags": channel_conf.get("follow_up_tags")
+            if channel_conf
+            else None,
+            "show_continue_in_web_ui": channel_conf.get("show_continue_in_web_ui")
+            if channel_conf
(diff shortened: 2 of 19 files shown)


backend/scripts/document_seeding_prep.py
```diff
@@ -161,27 +161,33 @@ class SeedPresaveDocument(BaseModel):
     url="https://docs.onyx.app/more/use_cases/overview",
     title=overview_title,
     content=overview,
-    title_embedding=model.encode(f"search_document: {overview_title}"),
-    content_embedding=model.encode(f"search_document: {overview_title}\n{overview}"),
+    title_embedding=list(model.encode(f"search_document: {overview_title}")),
```
Weves: are these just typing changes?

backend/onyx/connectors/slack/utils.py
```diff
@@ -72,6 +72,11 @@ def make_slack_api_rate_limited(
     @wraps(call)
     def rate_limited_call(**kwargs: Any) -> SlackResponse:
         last_exception = None
+
+        if "thread_ts" in kwargs:
```
Weves: can we add a comment as to why we're doing this? Also, are we sure this should be the desired logic for all slack calls?

backend/onyx/onyxbot/slack/blocks.py
```diff
@@ -105,6 +111,88 @@ def _build_qa_feedback_block(
...
 
+def _build_ephemeral_publication_block(
+    channel_id: str,
+    original_question_ts: str | None = None,
+    chat_message_id: int | None = None,
+    message_info: SlackMessageInfo | None = None,
+    channel_conf: ChannelConfig | None = None,
+    feedback_reminder_id: str | None = None,
+) -> Block:
+    if not chat_message_id:
```
Weves: we shouldn't need to have these checks here—we can just rely on mypy / typing to handle this. 

In other words, we could just have the typing as:

`chat_message_id: int`

backend/onyx/onyxbot/slack/blocks.py
```diff
@@ -105,6 +111,88 @@ def _build_qa_feedback_block(
...
+        raise ValueError("Chat message id is required to change the ephemeral message")
+
+    if message_info is None:
```
Weves: same as the ^ for these 2

backend/onyx/onyxbot/slack/utils.py
```diff
@@ -296,6 +298,14 @@ def build_feedback_id(
     return unique_prefix + ID_SEPARATOR + feedback_id
 
 
+def build_publish_ephemeral_message_id(
+    original_question_ts: str | None = None,
+) -> str:
+    if original_question_ts is None:
```
Weves: same comment here

backend/onyx/connectors/slack/utils.py
joachim-danswer: Was actually not necessary. Sending thread_ts as None is equivalent to removing that key.

Code snippet removed.

backend/onyx/onyxbot/slack/blocks.py
joachim-danswer: Agreed. Done.

backend/onyx/onyxbot/slack/blocks.py
joachim-danswer: Done

backend/onyx/onyxbot/slack/utils.py
joachim-danswer: done

backend/scripts/document_seeding_prep.py
joachim-danswer: Yes. For anopther PR I need to upgrade HF version, which apparently changes the type here, so 'list' was needed to make it work with my local version.

As that is another PR though, I will remove.

backend/onyx/onyxbot/slack/handlers/handle_buttons.py
```diff
@@ -143,6 +188,180 @@ def handle_generate_answer_button(
...
+            ),
+            llm_selected_doc_indices=None,
+            error_msg=None,
+        )
+
+    # Note: we need to use the webhook and the respond_url to update/delete ephemeral messages
+    if action_id == SHOW_EVERYONE_ACTION_ID:
+        # Convert to non-ephemeral message in thread
+        try:
+            webhook.send(
```
Weves, line 272: what does this one specifically do? Would be nice to add a comment here

backend/onyx/onyxbot/slack/handlers/handle_buttons.py
```diff
@@ -143,6 +188,180 @@ def handle_generate_answer_button(
...
+    ephemeral_message_ts = req.payload["container"]["message_ts"]
+
+    slack_sender_id = req.payload["user"]["id"]
+    response_url = req.payload["response_url"]
+    webhook = WebhookClient(url=response_url)
+
+    # The additional data required that was added to buttons.
+    # Specifically, this contains the message_info, channel_conf information
+    # and some additional attributes.
+    value_dict = json.loads(req.payload["actions"][0]["value"])
```
Weves, line 209: for future reference / a small enhancement: would prefer to get things into the "typed world" as soon as possible. In this case, that would mean converting this into a pydantic model / a typed dict.

Weves (approved):

PR #4157, opened 2025-02-28
Add option to not re-index

Weves:
## Description

https://linear.app/danswer/issue/DAN-1492/add-options-to-address-vespa-memory-issues-at-scale

Allows for:
1. Instant switch. Switches search settings without re-indexing in the background. 
2. Switching the vector floating point precision. Supports either bfloat16 or float (default).
3. For OpenAI embedding models, allows for cutting the dimensionality.

This is hidden for multi-tenant for now, as we should probably just provide a default / not expose this.

Also includes some misc modal cleanup (standardizing button layout in the bottom right) + label standardization.

## How Has This Been Tested?

Tested all locally (switching between all different options).


backend/onyx/background/celery/tasks/indexing/tasks.py
```diff
@@ -456,22 +457,17 @@ def check_for_indexing(self: Task, *, tenant_id: str) -> int | None:
...
                         search_settings_instance=search_settings_instance,
-                        search_settings_primary=search_settings_primary,
                         secondary_index_building=len(search_settings_list) > 1,
                         db_session=db_session,
                     ):
                         continue
 
                     reindex = False
-                    if search_settings_instance.id == search_settings_list[0].id:
+                    if search_settings_instance.status == IndexModelStatus.PRESENT:
```
Weves: @rkuo-danswer any specific reason we did it based on ordering rather than checking the status?

pablonyx (approved):

PR #4158, opened 2025-02-28
Fix curator default persona editing

pablonyx:
## Description

Curators should be able to edit featured assistants if they own them. They should NOT be able to modify the "featured" status

## How Has This Been Tested?

### DIFF
--- a/backend/onyx/db/persona.py
+++ b/backend/onyx/db/persona.py
@@ -214,8 +214,15 @@ def create_update_persona(
             if not create_persona_request.is_public:
                 raise ValueError("Cannot make a default persona non public")
 
-            if user and user.role != UserRole.ADMIN:
-                raise ValueError("Only admins can make a default persona")
+            if user:
+                # Curators can edit default personas, but not make them
+                if (
+                    user.role == UserRole.CURATOR
+                    or user.role == UserRole.GLOBAL_CURATOR
+                ):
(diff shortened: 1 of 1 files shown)


backend/onyx/db/persona.py
```diff
@@ -209,13 +209,21 @@ def create_update_persona(
...
 
-            if user and user.role != UserRole.ADMIN:
-                raise ValueError("Only admins can make a default persona")
+            if user:
+                # Curators can edit default personas, but not make them
+                if (
+                    user.role == UserRole.CURATOR
+                    or user.role == UserRole.GLOBAL_CURATOR
+                ):
+                    is_default_persona = None
```
rkuo-danswer, line 224: it feels like we should have a curator class with behavior like this stored inside instead of trying to reason out the behavior in a bunch of disparate functions.

backend/onyx/db/persona.py
```diff
@@ -428,7 +436,7 @@ def upsert_persona(
     remove_image: bool | None = None,
     search_start_date: datetime | None = None,
     builtin_persona: bool = False,
-    is_default_persona: bool = False,
+    is_default_persona: bool | None = None,
```
rkuo-danswer, line 439: having persona in the parameter name seems redundant but i suppose this is propagating from other variable names?

rkuo-danswer (approved):

PR #4159, opened 2025-02-28
Fix user group edge case

pablonyx:
## Description

We should not update roles of non-curators on relationship changes (edge case where user used to be curator and now is admin)

## How Has This Been Tested?

Locally, edge case before and after


backend/ee/onyx/server/tenants/billing.py
```diff
@@ -75,6 +75,7 @@ def register_tenant_users(tenant_id: str, number_of_users: int) -> stripe.Subscr
     Send a request to the control service to register the number of users for a tenant.
     """
 
+    return None
```
rkuo-danswer: Some dead code after the return ... likely a bug?

rkuo-danswer (approved):

PR #4166, opened 2025-03-01
decrease log spam model server

pablonyx:
## Description

## How Has This Been Tested?

### DIFF
--- a/backend/model_server/encoders.py
+++ b/backend/model_server/encoders.py
@@ -60,6 +60,16 @@ _OPENAI_MAX_INPUT_LEN = 2048
 _COHERE_MAX_INPUT_LEN = 96
 
 
+# Custom exception for authentication errors
+class AuthenticationError(Exception):
+    """Raised when authentication fails with a provider."""
+
+    def __init__(self, provider: str, message: str = "API key is invalid or expired"):
+        self.provider = provider
+        self.message = message
+        super().__init__(f"{provider} authentication failed: {message}")
+
+
 class CloudEmbedding:
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


backend/model_server/encoders.py
```diff
@@ -203,17 +287,43 @@ async def _embed_litellm_proxy(
...
+                headers=headers,
+            )
+            response.raise_for_status()
+            result = response.json()
+            return [embedding["embedding"] for embedding in result["data"]]
+        except httpx.HTTPStatusError as e:
+            if e.response.status_code == 401:
+                raise AuthenticationError(provider="LiteLLM Proxy")
+
+            error_string = (
```
evan-onyx: Feels to me like the construction of these error strings is pretty formulaic, would like to see that abstracted to a function to reduce code clutter

backend/model_server/encoders.py
```diff
@@ -142,53 +170,109 @@ async def _embed_voyage(
...
+                api_base=self.api_url,
+                api_version=self.api_version,
+            )
+            embeddings = [embedding["embedding"] for embedding in response.data]
+            return embeddings
+        except Exception as e:
+            if (
+                "401" in str(e)
+                or "unauthorized" in str(e).lower()
+                or "invalid api key" in str(e).lower()
```
evan-onyx: This logic is repeated a few times, I'd say it's probably reasonable to put it in a function to reduce code clutter. Also strongly prefer string constants for "unauthorized" and "invalid api key"

evan-danswer (commented):
Two relatively easy to do structural things

backend/model_server/encoders.py
```diff
@@ -59,6 +59,66 @@
...
+    error: Exception,
+    service_name: str,
+    model: str | None,
+    provider: EmbeddingProvider,
+    status_code: int | None = None,
+) -> str:
+    """
+    Format a standardized error string for embedding errors.
+    """
+    if status_code:
```
evan-onyx: slightly prefer something like
```
detail = f'Status {status_code}' if status_code else f'{type(error)}'
```
and then only need a single source of truth for the returned string :D

evan-danswer (approved):
LGTM with a teeeeny change

PR #4186, opened 2025-03-05
improved basic search latency

evan-danswer:
## Description

Before this PR, if a user did basic search with
- old "Agentic" search (LLM-based document filtering) enabled
- a reranking model enabled
The time-to-first-docs was pretty high because we were running filtering and reranking before sending docs. 

We also decided to look for other opportunities to reduce latency. In particular, the first LLM call we make to decide what tool and args to use (if any) often takes as long as a second. In that time, we now run 
- our intent model (classifying the query as keyword or non-keyword)
- query embedding (for use in vector db retrieval)
in parallel with tool choice.

All told, there used to be 1-2 seconds of latency after choosing a tool that is down to 0.1-0.2s with these changes, with the caveat that at worst the newly introduced parallelism makes the tool choice node take worst case 0.5 seconds longer.

NOTE: Technically we could pass the TimeoutThread objects around until their value is absolutely necessary, but there just isn't enough going on between the time I decided to join() and the place where those values are needed, so I opted for code cleanliness.

## How Has This Been Tested?

Tested + Benchmarked in UI

### DIFF
--- a/backend/onyx/agents/agent_search/orchestration/nodes/choose_tool.py
+++ b/backend/onyx/agents/agent_search/orchestration/nodes/choose_tool.py
@@ -10,9 +10,6 @@ from onyx.agents.agent_search.models import GraphConfig
 from onyx.agents.agent_search.orchestration.states import ToolChoice
 from onyx.agents.agent_search.orchestration.states import ToolChoiceState
 from onyx.agents.agent_search.orchestration.states import ToolChoiceUpdate
-from onyx.agents.agent_search.shared_graph_utils.constants import EMBEDDING_KEY
-from onyx.agents.agent_search.shared_graph_utils.constants import IS_KEYWORD_KEY
-from onyx.agents.agent_search.shared_graph_utils.constants import KEYWORDS_KEY
 from onyx.chat.prompt_builder.answer_prompt_builder import AnswerPromptBuilder
 from onyx.chat.tool_handling.tool_response_handler import get_tool_by_name
 from onyx.chat.tool_handling.tool_response_handler import (
@@ -20,6 +17,7 @@ from onyx.chat.tool_handling.tool_response_handler import (
 )
 from onyx.context.search.preprocessing.preprocessing import query_analysis
 from onyx.context.search.retrieval.search_runner import get_query_embedding
+from onyx.tools.models import SearchToolOverrideKwargs
 from onyx.tools.tool import Tool
 from onyx.tools.tool_implementations.search.search_tool import SearchTool
 from onyx.utils.logger import setup_logger
@@ -54,6 +52,7 @@ def choose_tool(
 
     embedding_thread: TimeoutThread[Embedding] | None = None
     keyword_thread: TimeoutThread[tuple[bool, list[str]]] | None = None
+    override_kwargs: SearchToolOverrideKwargs | None = None
     if (
         not agent_config.behavior.use_agentic_search
         and agent_config.tooling.search_tool is not None
@@ -61,6 +60,7 @@ def choose_tool(
             not force_use_tool.force_use or force_use_tool.tool_name == SearchTool.name
         )
     ):
+        override_kwargs = SearchToolOverrideKwargs()
         # Run in a background thread to avoid blocking the main thread
         embedding_thread = run_in_background(
             get_query_embedding,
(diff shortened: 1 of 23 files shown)


backend/onyx/agents/agent_search/orchestration/nodes/choose_tool.py
```diff
@@ -37,6 +49,29 @@ def choose_tool(
...
+
+    force_use_tool = agent_config.tooling.force_use_tool
+
+    embedding_thread: TimeoutThread[Embedding] | None = None
+    keyword_thread: TimeoutThread[tuple[bool, list[str]]] | None = None
+    if (
+        not agent_config.behavior.use_agentic_search
+        and agent_config.tooling.search_tool is not None
+        and (
+            not force_use_tool.force_use or force_use_tool.tool_name == SearchTool.name
```
Weves, line 60: why do we skip this if we're force using the search tool?

backend/onyx/chat/stream_processing/citation_processing.py
```diff
@@ -90,97 +90,97 @@ def process_token(
                     next(group for group in citation.groups() if group is not None)
                 )
 
-                if 1 <= numerical_value <= self.max_citation_num:
-                    context_llm_doc = self.context_docs[numerical_value - 1]
-                    final_citation_num = self.final_order_mapping[
-                        context_llm_doc.document_id
-                    ]
+                if not (1 <= numerical_value <= self.max_citation_num):
```
Weves, line 93: what's going on with these changes?

backend/onyx/context/search/preprocessing/preprocessing.py
```diff
@@ -118,7 +118,9 @@ def retrieval_preprocessing(
     )
 
     run_query_analysis = (
```
Weves, line 122: nit: prefer comment saying something like `Sometimes this is pre-computed in parallel with other heavy tasks to improve latency, and in that case we don't need to run the model again"

backend/onyx/tools/tool_implementations/search/search_tool.py
```diff
@@ -281,6 +287,11 @@ def run(
         self, override_kwargs: SearchToolOverrideKwargs | None = None, **llm_kwargs: Any
     ) -> Generator[ToolResponse, None, None]:
         query = cast(str, llm_kwargs[QUERY_FIELD])
+        precomputed_query_embedding = cast(
```
Weves: these should probably be override_kwargs?

backend/onyx/tools/tool_implementations/search/search_tool.py
```diff
@@ -385,8 +399,9 @@ def build_next_prompt(
 # the retrieved docs (reranking, deduping, etc.) after the SearchTool has run.
 def yield_search_responses(
     query: str,
-    reranked_sections: list[InferenceSection],
-    final_context_sections: list[InferenceSection],
+    get_retrieved_sections: Callable[[], list[InferenceSection]],
```
Weves, line 407: why are these made callables? I don't see anywhere where we are taking advantage of this

backend/onyx/agents/agent_search/orchestration/nodes/choose_tool.py
evan-onyx: The third condition is "if we are force using a tool, then it is the search tool"

backend/onyx/chat/stream_processing/citation_processing.py
evan-onyx: I can take them out if you want, but I needed to read through this code when fixing citations and I thought this improved readability

evan-onyx: It's just a simple transformation of 
```
for <stuff>
    if <stuff>
        <loop body>
```

to

```
for <stuff>
    if not <stuff>
        continue
    <loop body>
```

backend/onyx/tools/tool_implementations/search/search_tool.py
evan-onyx: yeah unfortunately it makes the code kind of ugly :/ v reasonable though, will do

backend/onyx/tools/tool_implementations/search/search_tool.py
evan-onyx: These 'properties' are actually functions defined with `@property` -_- so passing them into a function causes the underlying function to get called (yuck). Technically at the moment we don't *need* `get_retrieved_sections` to be delayed, but I thought I would save someone in the future a headache if the call gets moved around. Can put a comment describing this

Weves (approved):

PR #4187, opened 2025-03-05
Attempt to address tool happy claude

Weves:
## Description

Ref https://linear.app/danswer/issue/DAN-1519/address-image-happy-claude

## How Has This Been Tested?

Ran a query and verified it could still generate images / could skip images.


evan-danswer (approved):
This is technically fine, potential unintended consequence though of Claude refusing to use our non-explicit-tool-calling framework and therefore never calling tools...

PR #4190, opened 2025-03-05
* Replaces Amazon and Anthropic Icons with version better suitable fo…

pablonyx:
…r both Dark and  Light modes;

* Adds icon for DeepSeek;
* Simplify logic on icon selection;
* Adds entries for Phi-4, Claude 3.7, Ministral and Gemini 2.0 models

This is @emerzon's PR

## Description

## How Has This Been Tested?


web/src/lib/hooks.ts
```diff
@@ -730,14 +736,17 @@ const MODEL_DISPLAY_NAMES: { [key: string]: string } = {
   "gemini-1.5-flash-001": "Gemini 1.5 Flash",
   "gemini-1.5-pro-002": "Gemini 1.5 Pro (v2)",
   "gemini-1.5-flash-002": "Gemini 1.5 Flash (v2)",
+  "gemini-2.0-flash-exp": "Gemini 2.0 Flash (Experimental)",
   "gemini-2.0-flash-001": "Gemini 2.0 Flash",
+  "gemini-2.0-flash-lite-preview-02-05": "Gemini 2.0 Flash Lite (Prv)",
+  "gemini-2.0-flash-thinking-exp-01-02": "Gemini 2.0 Flash Thinking (Exp)",
```
rkuo-danswer, line 742: The descriptions for the "gemini-2.0-flash-thinking-exp-01-02" and "gemini-2.0-flash-thinking-exp-01-21" don't seem to distinguish them well.  Or maybe we should just replace the older one with the newer one?

web/src/lib/llm/utils.ts
```diff
@@ -90,6 +90,8 @@ const MODEL_NAMES_SUPPORTING_IMAGE_INPUT = [
   "anthropic.claude-3-5-sonnet-20240620-v1:0",
   "anthropic.claude-3-5-sonnet-20241022-v2:0",
   "anthropic.claude-3-7-sonnet-20250219-v1:0",
+  "claude-3.7-sonnet@202502019",
```
rkuo-danswer, line 93: is it normal to have two names for the same model in this form? We don't seem to be doing this for the other models?

web/src/lib/hooks.ts
emerzon: One concern here is that long descriptions are cropped in the menu

web/src/lib/llm/utils.ts
emerzon: Its normal. In this case we have both the naming used by AWS (with `-`) and GCP (with `@`)

PR #4218, opened 2025-03-06
Tenants on standby

pablonyx:
## Description

Maintain a list of ready tenants on standby to provision new tenants / users into. Mitigates the 9+ second latency on first-time registration.

---

Fixes https://linear.app/danswer/issue/DAN-1212/separate-page-for-users-during-creation-of-onyx-some-indication-that

## How Has This Been Tested?

### DIFF
--- a/backend/ee/onyx/server/tenants/provisioning.py
+++ b/backend/ee/onyx/server/tenants/provisioning.py
@@ -455,6 +455,8 @@ def get_tenant_by_domain_from_control_plane(
     except Exception as e:
         logger.error(f"Error fetching tenant by domain: {str(e)}")
         return None
+
+
 async def get_available_tenant() -> str | None:
     """
     Get an available pre-provisioned tenant from the NewAvailableTenant table.
@@ -540,8 +542,6 @@ async def assign_tenant_to_user(
     """
     # First, add the user to the tenant in a transaction
 
-
-    
     try:
         add_users_to_tenant([email], tenant_id)
 
@@ -563,5 +563,3 @@ async def assign_tenant_to_user(
     # Notify control plane with retry logic
     if not DEV_MODE:
         await notify_control_plane(tenant_id, email, referral_source)
-
-
--- a/backend/ee/onyx/server/tenants/provisioning.py
+++ b/backend/ee/onyx/server/tenants/provisioning.py
@@ -560,3 +560,5 @@ async def assign_tenant_to_user(
     # Notify control plane with retry logic
     if not DEV_MODE:
         await notify_control_plane(tenant_id, email, referral_source)
+
+
--- a/backend/ee/onyx/server/tenants/provisioning.py
+++ b/backend/ee/onyx/server/tenants/provisioning.py
@@ -28,11 +28,12 @@ from onyx.auth.users import exceptions
 from onyx.configs.app_configs import CONTROL_PLANE_API_BASE_URL
 from onyx.configs.app_configs import DEV_MODE
 from onyx.configs.constants import MilestoneRecordType
+from onyx.db.engine import get_session_with_shared_schema
 from onyx.db.engine import get_session_with_tenant
-from onyx.db.engine import get_sqlalchemy_engine
 from onyx.db.llm import update_default_provider
 from onyx.db.llm import upsert_cloud_embedding_provider
 from onyx.db.llm import upsert_llm_provider
+from onyx.db.models import AvailableTenant
 from onyx.db.models import IndexModelStatus
 from onyx.db.models import SearchSettings
 from onyx.db.models import UserTenantMapping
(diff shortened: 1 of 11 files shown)


backend/ee/onyx/server/tenants/provisioning.py
```diff
@@ -353,3 +347,97 @@ async def delete_user_from_control_plane(tenant_id: str, email: str) -> None:
...
+
+async def get_available_tenant() -> str | None:
+    """
+    Get an available pre-provisioned tenant from the NewAvailableTenant table.
+    Returns the tenant_id if one is available, None otherwise.
+    """
+    if not MULTI_TENANT:
+        return None
+
+    try:
```
rkuo-danswer: really looks like there are one or more race conditions here.  We need to be able to query and assign a tenant atomically, or retry an operation until everything is in a verified state.

backend/ee/onyx/server/tenants/provisioning.py
```diff
@@ -353,3 +347,97 @@ async def delete_user_from_control_plane(tenant_id: str, email: str) -> None:
...
+    tenant_id: str, email: str, referral_source: str | None = None
+) -> None:
+    """
+    Assign a tenant to a user and perform necessary operations.
+    """
+    # Add the user to the tenant
+    add_users_to_tenant([email], tenant_id)
+
+    # Create milestone record
+    with get_session_with_tenant(tenant_id=tenant_id) as db_session:
```
rkuo-danswer: writing to the db is error prone ... what happens if we crash and don't notify the control plane?

backend/onyx/background/celery/tasks/tenant_provisioning/tasks.py
```diff
@@ -0,0 +1,177 @@
...
+
+        task_logger.info(
+            f"Available tenants: {available_tenants_count}, "
+            f"Target: {target_available_tenants}, "
+            f"To provision: {tenants_to_provision}"
+        )
+
+        # Trigger pre-provisioning tasks for each tenant needed
+        for _ in range(tenants_to_provision):
+            pre_provision_tenant.apply_async(
```
rkuo-danswer: I think we've moved to send task nearly everywhere due to circular dependency issues with directly applying tasks.  Would prefer to stay away from apply async unless there's a good reason for it.

backend/onyx/background/celery/tasks/tenant_provisioning/tasks.py
```diff
@@ -0,0 +1,177 @@
...
+                priority=OnyxCeleryPriority.LOW,
+            )
+
+    except Exception as e:
+        task_logger.exception(f"Error in check_available_tenants task: {e}")
```
rkuo-danswer: {e} is redundant with logger.exception

backend/onyx/background/celery/tasks/tenant_provisioning/tasks.py
```diff
@@ -0,0 +1,177 @@
...
+    bind=True,
+)
+def pre_provision_tenant(self: Task) -> None:
+    """
+    Pre-provision a new tenant and store it in the NewAvailableTenant table.
+    This function fully sets up the tenant with all necessary configurations,
+    so it's ready to be assigned to a user immediately.
+    """
+    task_logger.info("STARTING PRE_PROVISION_TENANT")
+    if not MULTI_TENANT:
```
rkuo-danswer: need to stop gating functions with globals.  usually we can gate outside them by simply not calling them.

backend/onyx/background/celery/tasks/tenant_provisioning/tasks.py
```diff
@@ -0,0 +1,177 @@
...
+    )
+
+    # Allow multiple pre-provisioning tasks to run, but ensure they don't overlap
+    if not lock_provision.acquire(blocking=False):
+        task_logger.info(
+            "Skipping pre_provision_tenant task because it is already running"
+        )
+        return
+
+    try:
```
rkuo-danswer, line 138: it definitely looks problematic to be running this series of operations not in a single transaction or lock.  Have we thought about this yet?

backend/onyx/background/celery/tasks/tenant_provisioning/tasks.py
```diff
@@ -0,0 +1,177 @@
...
+                alembic_version=alembic_version,
+                date_created=datetime.datetime.now(),
+            )
+            db_session.add(new_tenant)
+            db_session.commit()
+
+        task_logger.info(f"Successfully pre-provisioned tenant {tenant_id}")
+
+    except Exception as e:
+        task_logger.exception(f"Error in pre_provision_tenant task: {e}")
```
rkuo-danswer: {e} redundant.

backend/onyx/background/celery/tasks/tenant_provisioning/tasks.py
pablonyx: Yeah, did not mean to leave this in- was purely for testing

backend/onyx/background/celery/tasks/tenant_provisioning/tasks.py
rkuo-danswer: I still don't see a transaction or rollback around the entire operation.  Did we handle that?

backend/ee/onyx/server/tenants/provisioning.py
```diff
@@ -60,42 +61,72 @@ async def get_or_provision_tenant(
...
+        if not DEV_MODE and referral_source:
             await notify_control_plane(tenant_id, email, referral_source)
+
     except Exception as e:
-        logger.error(f"Tenant provisioning failed: {e}")
-        await rollback_tenant_provisioning(tenant_id)
+        logger.exception(f"Tenant provisioning failed: {str(e)}")
+        # Attempt to rollback the tenant provisioning
+        try:
+            await rollback_tenant_provisioning(tenant_id)
```
rkuo-danswer, line 127: i was expecting more of a session.rollback ... is the manual approach here better?

pablonyx: For this case, I believe it may be– since the provisioning process has various stages, it's not as if we are modifying a few columns; instead, we're creating the schema, updating a public table, then doing various operations (post commit) within the new schema (and we want to handle failure cases in any part of this flow

backend/onyx/background/celery/tasks/tenant_provisioning/tasks.py
pablonyx: Great point- will re-use above function

rkuo-danswer (approved):

PR #4224, opened 2025-03-06
use max_tokens to do better rate limit handling

evan-onyx:
## Description

Fixes https://linear.app/danswer/issue/DAN-1518/use-max-tokens-in-agent-search-where-possible

- Some timeout bumps to make o1 happy
- use max_tokens to make (especially the smaller) Claude models not give rate limit errors. 

From https://docs.anthropic.com/en/api/rate-limits

"OTPM rate limits are estimated based on max_tokens at the beginning of each request, and the estimate is adjusted at the end of the request to reflect the actual number of output tokens used. If you’re hitting OTPM limits earlier than expected, try reducing max_tokens to better approximate the size of your completions."

## How Has This Been Tested?

Tested in UI


joachim-danswer (approved):
LGTM

PR #4228, opened 2025-03-07
Separate out indexing-time image analysis into new phase

pablonyx:
## Description

Refactor so that we can keep connectors more focused. Instead of doing image analysis within connectors and having connector-specific handling there, we simply yield back ImageSections / TextSections that are then converted to Sections (as part of the indexing pipeline)

Also some misc Drive improvements for downloading media

N.B. I can see an argument for doing this conversion earlier in the pipeline– open to thoughts

Fixes https://linear.app/danswer/issue/DAN-1536/image-search-enhancements
## How Has This Been Tested?

Various image formats in:

- Drive
- File
- Confluence


backend/onyx/indexing/indexing_pipeline.py
```diff
@@ -362,19 +493,25 @@ def index_doc_batch(
...
             "doc_id": doc.id,
             "doc_length": doc.get_total_char_length(),
         }
-        for doc in ctx.updatable_docs
+        for doc in ctx.indexable_docs
     ]
     logger.debug(f"Starting indexing process for documents: {doc_descriptors}")
 
+    # Process any ImageSection objects before chunking
+    logger.debug("Processing image sections")
```
rkuo-danswer: logging statement outdated?

backend/onyx/connectors/blob/connector.py
```diff
@@ -341,7 +341,12 @@ def validate_connector_settings(self) -> None:
                 print("Sections:")
                 for section in doc.sections:
                     print(f"  - Link: {section.link}")
-                    print(f"  - Text: {section.text[:100]}...")
+                    if isinstance(section, TextSection) and section.text is not None:
+                        print(f"  - Text: {section.text[:100]}...")
+                    elif (
+                        hasattr(section, "image_file_name") and section.image_file_name
+                    ):
+                        print(f"  - Image: {section.image_file_name}")
```
rkuo-danswer, line 349: explicitly handle else case ... exception or warning, etc.

backend/onyx/connectors/bookstack/connector.py
```diff
@@ -167,7 +167,7 @@ def _page_to_document(
         time.sleep(0.1)
```
rkuo-danswer, line 167: Agree with greptile here ... what is this sleep doing? needs to be commented at a minimum

backend/onyx/connectors/confluence/connector.py
```diff
@@ -264,18 +264,23 @@ def _convert_page_to_document(self, page: dict[str, Any]) -> Document | None:
...
                         sections.append(attachment_section)
-                    elif result.error:
+                    elif result and result.file_name:
+                        # Create an ImageSection for image attachments
+                        image_section = ImageSection(
+                            link=f"{page_url}#attachment-{attachment['id']}",
+                            image_file_name=result.file_name,
+                        )
+                        sections.append(image_section)
+                    elif result and result.error:
```
rkuo-danswer: handle else case with exception/warning/pass etc.

backend/onyx/connectors/file/connector.py
```diff
@@ -26,11 +26,13 @@
 from onyx.file_processing.extract_file_text import load_files_from_zip
 from onyx.file_processing.image_utils import store_image_and_create_section
 from onyx.file_store.file_store import get_default_file_store
-from onyx.llm.interfaces import LLM
 from onyx.utils.logger import setup_logger
 
 logger = setup_logger()
 
+# Constants
+IMAGE_EXTENSIONS = {".jpg", ".jpeg", ".png", ".webp", ".gif"}
```
rkuo-danswer: can this go in the connector base class?

backend/onyx/file_processing/image_utils.py
```diff
@@ -18,25 +15,24 @@ def store_image_and_create_section(
     image_data: bytes,
     file_name: str,
     display_name: str,
+    link: str | None = None,
```
rkuo-danswer, line 18: Is the media_type supposed to be a mime_type? I don't think we should be sending mime types around that aren't real types.

backend/onyx/connectors/google_drive/doc_conversion.py
```diff
@@ -327,44 +215,53 @@ def convert_drive_item_to_document(
...
 
-        doc_id = file["webViewLink"]
-        updated_time = datetime.fromisoformat(file["modifiedTime"]).astimezone(
-            timezone.utc
-        )
-
+        # Create the document
         return Document(
-            id=doc_id,
+            id=f"gdrive-{file.get('id')}",
```
rkuo-danswer: if we change this id, does that mean docs will be duplicated or reindexed?

backend/onyx/connectors/file/connector.py
```diff
@@ -170,58 +187,75 @@ def _process_file(
...
+    # Extract text and images from the file
     text_content, embedded_images = extract_text_and_images(
         file=file,
         file_name=file_name,
         pdf_pass=pdf_pass,
     )
 
     # Build sections: first the text as a single Section
-    sections = []
+    sections: list[TextSection | ImageSection] = []
```
rkuo-danswer, line 233: this sort of union typing could get unwieldy ... I think it's better to just have the list be a base class of Section and then switch on the subclass type on the other end of things.

backend/onyx/connectors/bookstack/connector.py
pablonyx: Not sure why this was added– looks like it was added 20 months ago and we have no valid Bookstack credentials to test this with, unfortunately.

backend/onyx/connectors/file/connector.py
pablonyx: Definitely agree re: union typing getting unwieldy with more section types, but I think it may be worth keeping as is, considering we need to distinguish between TextSection and ImageSection during processing before converting to the fully populated Section type (so this gives us some more type safety)

backend/onyx/file_processing/image_utils.py
pablonyx: Fair point!

backend/onyx/connectors/file/connector.py
pablonyx: Yes!

backend/onyx/connectors/file/connector.py
rkuo-danswer: ok we can revisit this at some point

PR #4231, opened 2025-03-07
Feature/agentic buffered

rkuo-danswer:
## Description

Fixes DAN-1499.
https://linear.app/danswer/issue/DAN-1499/agent-search-non-streaming-endpoints

## How Has This Been Tested?


backend/onyx/main.py
```diff
@@ -233,6 +233,8 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
 
     yield
 
+    SqlEngine.reset_engine()
```
evan-onyx, line 236: Was this an accidental addition?

backend/onyx/chat/models.py
```diff
@@ -44,9 +47,41 @@ class LlmDoc(BaseModel):
...
+    """None represents references to objects in the original flow. To our understanding,
+    these will not be None in the packets returned from agent search.
+    """
+
     level: int | None = None
     level_question_num: int | None = None
 
+    @staticmethod
+    def make_dict_by_level(
+        original_dict: Mapping[tuple[int | None, int | None], "SubQuestionIdentifier"]
```
evan-onyx: I'm in favor of removing all the | None's and asserting that values aren't None for typing; probably better to find out that way if something breaks

evan-onyx: It might be out of scope for this PR, but I'm now leaning that we should have made the fields for SubQuestionIdentifier non-Nullable and added it as an attribute instead of subclassing... oh well

backend/onyx/chat/models.py
```diff
@@ -44,9 +47,41 @@ class LlmDoc(BaseModel):
...
+    ) -> dict[int | None, list["SubQuestionIdentifier"]]:
+        """returns a dict of level to object list (sorted by level_question_num)
```
evan-onyx, line 61: Maybe we can make this return a list once we purge the Nones?

backend/ee/onyx/server/query_and_chat/models.py
```diff
@@ -88,6 +91,66 @@ class SimpleDoc(BaseModel):
...
+
+        # sort each query_id list and question_index
+        for key1, obj1 in level_question_dict.items():
+            for key2, value2 in obj1.items():
+                # sort the query_id list of each question_index
+                level_question_dict[key1][key2] = sorted(
+                    value2, key=lambda o: o.query_id
+                )
+            # sort the question_index dict of level
+            level_question_dict[key1] = OrderedDict(
```
evan-onyx, line 141: In favor of getting rid of None's explicitly via assert and using lists

backend/ee/onyx/server/query_and_chat/chat_backend.py
```diff
@@ -113,11 +137,98 @@ def _convert_packet_stream_to_response(
...
+            logger.warning(
+                "_convert_packet_stream_to_response: Unexpected chat packet type ExtendedToolResponse!"
+            )
+        elif isinstance(packet, RefinedAnswerImprovement):
+            response.agent_refined_answer_improvement = (
+                packet.refined_answer_improvement
+            )
+        else:
+            logger.warning(
+                "_convert_packet_stream_to_response: Unrecognized chat packet type!"
```
evan-onyx: could add the type to the logger warning

evan-danswer (approved):
LGTM with some minor edits!

backend/onyx/main.py
rkuo-danswer: this was in fact intentional.  We should clean up the engine when shutting down.

backend/onyx/chat/models.py
rkuo-danswer: there's still levels and questions per level, so i think it's better the way it is.

evan-danswer (approved):
LGTM thanks Richard!

PR #4234, opened 2025-03-07
Add basic memory logging

Weves:
## Description

Fixes https://linear.app/danswer/issue/DAN-1544/add-basic-memory-loggin

## How Has This Been Tested?

Ran docker compose locally, verified I saw the logs as expected in the dir I expected + verified that on container restart the logs were persisted.

### DIFF
--- a/backend/onyx/background/celery/memory_monitoring.py
+++ b/backend/onyx/background/celery/memory_monitoring.py
@@ -56,5 +56,5 @@ def emit_process_memory(
             f"vms_mb={memory_info.vms / (1024 * 1024):.2f} "
             f"cpu={cpu_percent:.2f}{metadata_str}"
         )
-    except Exception as e:
-        logger.error(f"Error monitoring worker memory: {e}")
+    except Exception:
+        logger.exception("Error monitoring process memory.")
--- a/backend/onyx/background/celery/memory_monitoring.py
+++ b/backend/onyx/background/celery/memory_monitoring.py
@@ -1,0 +1,60 @@
+# backend/onyx/background/celery/memory_monitoring.py
+import logging
+import os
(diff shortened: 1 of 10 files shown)


backend/onyx/background/celery/tasks/indexing/tasks.py
```diff
@@ -1024,6 +1028,23 @@ def connector_indexing_proxy_task(
                     job.release()
                     break
 
+            # log the memory usage for tracking down memory leaks / connector-specific memory issues
+            pid = job.process.pid
+            if pid is not None:
+                # Only emit memory info once per minute (60 seconds)
+                current_time = time.time()
```
rkuo-danswer: time.monotonic preferable in most cases

backend/onyx/background/celery/memory_monitoring.py
```diff
@@ -0,0 +1,60 @@
...
+from logging.handlers import RotatingFileHandler
+
+import psutil
+
+from onyx.utils.logger import setup_logger
+
+# Regular application logger
+logger = setup_logger()
+
+# Set up a dedicated memory monitoring logger
```
rkuo-danswer, line 13: Will there be sufficient information in the log line to distinguish between various indexing connector sources so we can zero in on what failed?  Is there any intention to use this from sources other than indexing where we would need more metadata to distinguish those from indexing metrics?

rkuo-danswer (approved):

backend/onyx/background/celery/memory_monitoring.py
Weves: it should already be possible via the process_name param (+ additional_metadata). I'm not sure if that's what you're referring to?

PR #4258, opened 2025-03-11
Improved logout flow

pablonyx:
## Description

Fixes https://linear.app/danswer/issue/DAN-1567/better-logged-out-flow

## How Has This Been Tested?


backend/onyx/redis/redis_pool.py
```diff
@@ -160,6 +160,20 @@ def get_client(self, tenant_id: str) -> Redis:
...
+        """
+        return redis.Redis(connection_pool=self._pool)
+
+    def get_raw_replica_client(self) -> Redis:
+        """
+        Returns a Redis client with direct access to the replica connection pool,
+        without tenant prefixing.
+        """
+        return redis.Redis(connection_pool=self._replica_pool)
+
```
rkuo-danswer, line 176: there's already a "get_shared_redis_client" set of functions ... is that usable for this use case or do we want the raw versions?

backend/onyx/server/manage/users.py
```diff
@@ -506,6 +512,42 @@ def get_current_token_expiration_jwt(
...
+
+        # Get the TTL of the token
+        ttl = cast(int, redis.ttl(redis_key))
+        if ttl <= 0:
+            logger.error("Token has expired or doesn't exist in Redis")
+            return None
+
+        # Calculate the creation time based on TTL and session expiry
+        # Current time minus (total session length minus remaining TTL)
+        current_time = datetime.now()
```
rkuo-danswer: docstring the form of time being returned here, and ensure datetime.now is what we want (utc is usually the right choice in most cases).

web/src/components/health/healthcheck.tsx
```diff
@@ -3,52 +3,105 @@
...
       const attemptTokenRefresh = async () => {
         let retryCount = 0;
         const maxRetries = 3;
 
         while (retryCount < maxRetries) {
           try {
-            // NOTE: This is a mocked refresh token for testing purposes.
-            // const refreshTokenData = mockedRefreshToken();
-
+            console.debug("Attempting token refresh");
```
rkuo-danswer: extra logging?

web/src/components/health/healthcheck.tsx
```diff
@@ -93,64 +161,59 @@ export const HealthCheckBanner = () => {
...
       </Modal>
     );
+  }
+
+  if (!error && !expired) {
+    return null;
+  }
+
+  console.debug(
+    `Rendering HealthCheckBanner. Error: ${error}, Expired: ${expired}`
```
rkuo-danswer: extra logging?

rkuo-danswer (commented):
tried with a short session timeout ... the model appears, but the dialog and in fact the entire page are unclickable.

backend/onyx/redis/redis_pool.py
pablonyx: Unfortunately no, since FastAPI Users directly uses the Redis connection (no prefixes whatsoever) and we need to interface with the token it sets

rkuo-danswer: Makes sense!  Maybe let's docstring each client getter with the intent (raw for use with consumers that have inflexible requirements, shared otherwise for using a single shared namespaced keyspace to better organize data, and per tenant for tenant keyspacing.

rkuo-danswer (approved):

backend/onyx/redis/redis_pool.py
pablonyx: Sounds great!

PR #4260, opened 2025-03-11
Feature/email whitelabeling

rkuo-danswer:
## Description

Fixes DAN-1498.
https://linear.app/danswer/issue/DAN-1498/email-whitelabeling

## How Has This Been Tested?


backend/ee/onyx/server/enterprise_settings/api.py
```diff
@@ -131,31 +131,64 @@ def put_logo(
     upload_logo(file=file, db_session=db_session, is_logotype=is_logotype)
 
 
-def fetch_logo_or_logotype(is_logotype: bool, db_session: Session) -> Response:
+# def fetch_logo_or_logotype(is_logotype: bool, db_session: Session) -> Response:
```
Weves: nit: remove commented-out function

backend/onyx/utils/variable_functionality.py
```diff
@@ -108,6 +108,25 @@ def fetch_versioned_implementation_with_fallback(
         return fallback
 
 
+def fetch_ee_implementation_or_none(module: str, attribute: str) -> T | None:
```
Weves: is there a reason why we need this when we have `fetch_ee_implementation_or_noop`?

backend/onyx/server/runtime/onyx_runtime.py
```diff
@@ -0,0 +1,95 @@
...
+        if get_logo_filename_fn:
+            db_filename = get_logo_filename_fn()
+
+        return OnyxRuntime._get_with_static_fallback(db_filename, STATIC_FILENAME)
+
+    @staticmethod
+    def get_emailable_logo() -> OnyxFile:
+        STATIC_FILENAME = "static/images/logo.png"
+
+        db_filename: str | None = None
```
Weves: nit: agree with comment below. Should we call `get_logo` here?

backend/onyx/server/runtime/onyx_runtime.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class OnyxRuntime:
+    @staticmethod
+    def _get_with_static_fallback(
+        db_filename: str | None, static_filename: str
+    ) -> OnyxFile:
+        onyx_file: OnyxFile | None = None
+
+        while True:
```
Weves: what's the need for the `while` loop here?

backend/onyx/server/runtime/onyx_runtime.py
```diff
@@ -0,0 +1,95 @@
...
+from onyx.db.engine import get_session_with_shared_schema
+from onyx.file_store.onyx_file_store import OnyxFileStore
+from onyx.utils.file import OnyxFile
+from onyx.utils.file import OnyxStaticFileManager
+from onyx.utils.variable_functionality import (
+    fetch_ee_implementation_or_none,
+)
+
+
+class OnyxRuntime:
```
Weves, line 15: nit: what is intended with the name `OnyxRuntime`? What other functionality would be added in the future? As is now, would prefer to not have the class and just have a `logo.py` file with these two functions (maybe in the `file_store` dir).

backend/onyx/utils/file.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class OnyxFile(BaseModel):
+    data: bytes
+    mime_type: str
+
+
+class OnyxStaticFileManager:
+    @staticmethod
+    def get_static(filename: str) -> OnyxFile | None:
```
Weves: nit: pref a short description here like `meant for fetching files bundled with the Onyx Docker image. For example, files within "static/images"`

backend/onyx/file_store/onyx_file_store.py
```diff
@@ -0,0 +1,20 @@
+from typing import cast
+
+import puremagic
+
+from onyx.file_store.file_store import PostgresBackedFileStore
+from onyx.utils.file import OnyxFile
+
+
+class OnyxFileStore(PostgresBackedFileStore):
```
Weves: Two quick thoughts:

1. In general, we've had a fairly functional leaning. One of the things that's a part of that is trying to avoid implementation inheritance as much as possible. For consistency, would prefer to keep that going (e.g. have this take in a `FileStore` object and use that).
2. Prefer more descriptive naming here. E.g. something like `def get_file_with_mime_type(filename: str, file_store: FileStore) -> FileWithMimeType`. Currently, I'm not sure based on the naming, why/when I would use this over the `PostgresBackedFileStore` directly.

backend/onyx/server/runtime/onyx_runtime.py
rkuo-danswer: Basically we need a layer that presents directly the value to be used by the application at runtime.  Settings often don't exist or go through layers of translation before the application knows what it's supposed to do, aka, check db, check env, check cache, use default ... logo's are a case in point where we default to the onyx logo if there isn't a db entry, but we use a custom value if there is one.  However, the caller mostly doesn't care and just wants the logo.

I can add a docstring here to clarify this but I think we really should design with this moving forward so that we don't scatter configuration gathering logic all over the place.

backend/onyx/server/runtime/onyx_runtime.py
rkuo-danswer: this is a python equivalent of a do while break, but in this case it's probably not helping to linearize the logic as there ended up being not many exit conditions, so i'll remove it.

backend/onyx/utils/variable_functionality.py
rkuo-danswer: oops! it looks that can be done with the noop function, will remove.

backend/onyx/server/runtime/onyx_runtime.py
rkuo-danswer: looks reusable, will do that

backend/onyx/file_store/onyx_file_store.py
rkuo-danswer: I refactored the logo methods out of this class which sort of removes the rationale for this class since the remaining logic can be application agnostic.  I'll move this method to the pgfilestore

Weves (approved):

PR #4262, opened 2025-03-12
Gdrive checkpointed connector

evan-danswer:
## Description

Addresses https://linear.app/danswer/issue/DAN-1505/google-drive-checkpointed-connector

- Refactored CheckpointConnector to have better typing
- Implemented Checkpointing for google drive using OAuth and service account credentials
  - Checkpointing is done in stages based on what the user specified they wanted retrieved
  - within each stage, retrieval is now ordered by modified time to allow returning to the stage at the stored time
  - When retrieval reaches the folder crawling stage, checkpointing is based on a set of previously seen parent folder ids. This means we will re-traverse the folder structure to find new docs to index. potential future work could include making this more efficient, but it likely isn't too many extra calls since we exit early from already-seen folders.
  - Slim retrieval modifies a dummy checkpoint in some cases, but this shouldn't cause any issues.
- fixed a parallelism issue for service account retrieval (was submitting a sequence of generators to as_completed)
- added some parallel utilities to support the above with associated tests

## How Has This Been Tested?

So far, a full run of OAuth drive retrieval has worked, with a small number of overlapped docs (the last doc of one indexing batch become the first one retrieved by the next run of that checkpoint). 

Testing checklist (X = done):
X full run with OAuth
- simulated error(s) with OAuth
- full exit halfway through and continue from checkpoint later with OAuth
- a full run with service account
- simulated error(s) with service account
- full exit halfway through and continue from checkpoint later with service account

TODO automated tests:
- unit test for serializing/deserializing drive checkpoint
- unit tests for restarting from various points in the checkpointing process
  - include boundary cases (ended on last doc in a section)

### DIFF
--- a/backend/onyx/agents/agent_search/deep_search/main/nodes/extract_entities_terms.py
+++ b/backend/onyx/agents/agent_search/deep_search/main/nodes/extract_entities_terms.py
@@ -106,7 +106,6 @@ def extract_entities_terms(
         first_bracket = cleaned_response.find("{")
         last_bracket = cleaned_response.rfind("}")
         cleaned_response = cleaned_response[first_bracket : last_bracket + 1]
-        cleaned_response += "g8o723dg87db"
 
         try:
             entity_extraction_result = EntityExtractionResult.model_validate_json(
--- a/backend/onyx/agents/agent_search/deep_search/main/nodes/extract_entities_terms.py
+++ b/backend/onyx/agents/agent_search/deep_search/main/nodes/extract_entities_terms.py
@@ -106,6 +106,7 @@ def extract_entities_terms(
         first_bracket = cleaned_response.find("{")
         last_bracket = cleaned_response.rfind("}")
         cleaned_response = cleaned_response[first_bracket : last_bracket + 1]
+        cleaned_response += "g8o723dg87db"
 
         try:
             entity_extraction_result = EntityExtractionResult.model_validate_json(
--- a/backend/onyx/connectors/google_drive/connector.py
+++ b/backend/onyx/connectors/google_drive/connector.py
@@ -15,6 +15,7 @@ from googleapiclient.errors import HttpError  # type: ignore
 from typing_extensions import override
 
 from onyx.configs.app_configs import INDEX_BATCH_SIZE
+from onyx.configs.app_configs import MAX_DRIVE_WORKERS
 from onyx.configs.constants import DocumentSource
 from onyx.connectors.exceptions import ConnectorValidationError
 from onyx.connectors.exceptions import CredentialExpiredError
@@ -436,7 +437,7 @@ class GoogleDriveConnector(SlimConnector, CheckpointConnector[GoogleDriveCheckpo
             )
             for email in all_org_emails
         ]
-        yield from parallel_yield(user_retrieval_gens, max_workers=10)
+        yield from parallel_yield(user_retrieval_gens, max_workers=MAX_DRIVE_WORKERS)
 
         remaining_folders = (
             drive_ids_to_retrieve | folder_ids_to_retrieve
--- a/backend/onyx/connectors/google_drive/connector.py
+++ b/backend/onyx/connectors/google_drive/connector.py
@@ -154,21 +154,6 @@ class DriveIdStatus(Enum):
     FINISHED = "finished"
 
 
-class DriveIdServer:
-    """
-    Returns an iterator over drive_ids.
-    """
-
-    def __init__(self, drive_ids: list[str], retrieved_ids: set[str]):
-        self.original_ids = set(drive_ids)
-
-        # drives available for processing
-        self.available_ids = copy.copy(drive_ids)
-
-        # use a reference to allow outside updates to affect this
-        self.finished_ids = retrieved_ids
-
-
 class GoogleDriveConnector(SlimConnector, CheckpointConnector[GoogleDriveCheckpoint]):
     def __init__(
         self,
@@ -362,7 +347,6 @@ class GoogleDriveConnector(SlimConnector, CheckpointConnector[GoogleDriveCheckpo
             processed_ids: set[str], thread_id: str
         ) -> tuple[str | None, bool]:
             found_future_work = False
-            print(f"{thread_id} drive_id_status: {drive_id_status}")
             for drive_id, status in drive_id_status.items():
                 if drive_id in self._retrieved_ids:
                     drive_id_status[drive_id] = DriveIdStatus.FINISHED
@@ -430,7 +414,6 @@ class GoogleDriveConnector(SlimConnector, CheckpointConnector[GoogleDriveCheckpo
                 stage=DriveRetrievalStage.MY_DRIVE_FILES,
                 completed_until=start or 0,
             )
-            checkpoint.completion_map[user_email] = curr_stage
             resuming = False
         drive_service = get_drive_service(self.creds, user_email)
 
--- a/backend/onyx/connectors/google_drive/connector.py
+++ b/backend/onyx/connectors/google_drive/connector.py
@@ -512,9 +512,9 @@ class GoogleDriveConnector(SlimConnector, CheckpointConnector[GoogleDriveCheckpo
                 folder_id: str, folder_start: SecondsSinceUnixEpoch | None
             ) -> Iterator[RetrievedDriveFile]:
                 yield from crawl_folders_for_files(
-                    is_slim=is_slim,
                     service=drive_service,
                     parent_id=folder_id,
+                    is_slim=is_slim,
                     user_email=user_email,
                     traversed_parent_ids=self._retrieved_ids,
                     update_traversed_ids_func=self._update_traversed_parent_ids,
@@ -741,9 +741,9 @@ class GoogleDriveConnector(SlimConnector, CheckpointConnector[GoogleDriveCheckpo
             folder_id: str, folder_start: SecondsSinceUnixEpoch | None
         ) -> Iterator[RetrievedDriveFile]:
             yield from crawl_folders_for_files(
-                is_slim=is_slim,
                 service=drive_service,
                 parent_id=folder_id,
+                is_slim=is_slim,
                 user_email=self.primary_admin_email,
                 traversed_parent_ids=self._retrieved_ids,
                 update_traversed_ids_func=self._update_traversed_parent_ids,
@@ -1149,3 +1149,7 @@ class GoogleDriveConnector(SlimConnector, CheckpointConnector[GoogleDriveCheckpo
             completion_map=ThreadSafeDict(),
             has_more=True,
         )
+
+    @override
+    def validate_checkpoint_json(self, checkpoint_json: str) -> GoogleDriveCheckpoint:
+        return GoogleDriveCheckpoint.model_validate_json(checkpoint_json)
--- a/backend/onyx/connectors/google_drive/connector.py
+++ b/backend/onyx/connectors/google_drive/connector.py
@@ -798,7 +798,6 @@ class GoogleDriveConnector(SlimConnector, CheckpointConnector[GoogleDriveCheckpo
         if is_slim:
             return drive_files
 
-        all_indexed: list[RetrievedDriveFile] = []
         for file in drive_files:
             if file.error is not None:
                 checkpoint.completion_map[file.user_email].update(
@@ -807,15 +806,6 @@ class GoogleDriveConnector(SlimConnector, CheckpointConnector[GoogleDriveCheckpo
                     completed_until_parent_id=file.parent_id,
                 )
             yield file
-            all_indexed.append(file)
-        for file in all_indexed:
-            print(
-                file.user_email,
-                file.drive_file["name"],
-                file.drive_file["modifiedTime"],
-            )
-
-        print("done")
 
     def _manage_oauth_retrieval(
         self,
@@ -1010,7 +1000,20 @@ class GoogleDriveConnector(SlimConnector, CheckpointConnector[GoogleDriveCheckpo
                             if doc is not None:
                                 documents.append(doc)
                         except Exception as e:
-                            logger.error(f"Error converting file: {e}")
+                            error_str = f"Error converting file: {e}"
+                            logger.error(error_str)
+                            yield [
+                                ConnectorFailure(
+                                    failed_document=DocumentFailure(
+                                        document_id=retrieved_file.drive_file["id"],
+                                        document_link=retrieved_file.drive_file[
+                                            "webViewLink"
+                                        ],
+                                    ),
+                                    failure_message=error_str,
+                                    exception=e,
+                                )
+                            ]
 
                     if documents:
                         yield documents
--- a/backend/onyx/connectors/google_drive/connector.py
+++ b/backend/onyx/connectors/google_drive/connector.py
@@ -8,8 +8,8 @@ from concurrent.futures import ThreadPoolExecutor
 from enum import Enum
 from functools import partial
 from typing import Any
-from urllib.parse import urlparse
 from typing import Protocol
+from urllib.parse import urlparse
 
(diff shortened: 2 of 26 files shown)


backend/onyx/connectors/google_drive/connector.py
```diff
@@ -350,15 +375,24 @@ def _impersonate_user_for_retrieval(
...
                 end=end,
+                key=lambda check: user_email
+                + "@"
+                + drive_id,  # completion map keyed by drive id
             )
 
+        # I believe there may be some duplication here,
+        # i.e. if two users have access to the same folder
+        # and are retrieving in parallel.
         remaining_folders = filtered_folder_ids - self._retrieved_ids
```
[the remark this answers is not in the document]
evan-onyx: Duplicate retrievals isn't that bad relative to what it would take to fix this (synchronized retrieved_ids across threads)

backend/onyx/connectors/interfaces.py
```diff
@@ -184,14 +190,21 @@ def handle_event(self, event: Any) -> GenerateDocumentsOutput:
         raise NotImplementedError
 
 
-class CheckpointConnector(BaseConnector):
+CT = TypeVar("CT", bound=ConnectorCheckpoint)
+# TODO: find a reasonable way to parameterize the return type of the generator
+CheckpointOutput: TypeAlias = Generator[
+    Document | ConnectorFailure, None, ConnectorCheckpoint
+]
```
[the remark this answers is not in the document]
evan-onyx: Yeah, I also wish Type Aliases and Generics worked like that. They don't, though.

backend/onyx/connectors/interfaces.py
```diff
@@ -212,3 +225,10 @@ def load_from_checkpoint(
         ```
         """
         raise NotImplementedError
+
+    # Ideally return typeshould be CT, but that's not possible if
+    # we want to override build_dummy_checkpoint and have BaseConnector
+    # return a base ConnectorCheckpoint
```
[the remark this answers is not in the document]
evan-onyx: 🐐

backend/onyx/connectors/models.py
```diff
@@ -198,18 +197,13 @@ class IndexAttemptMetadata(BaseModel):
...
-
     def __str__(self) -> str:
         """String representation of the checkpoint, with truncation for large checkpoint content."""
         MAX_CHECKPOINT_CONTENT_CHARS = 1000
 
-        content_str = json.dumps(self.checkpoint_content)
+        content_str = self.model_dump_json()
         if len(content_str) > MAX_CHECKPOINT_CONTENT_CHARS:
             content_str = content_str[: MAX_CHECKPOINT_CONTENT_CHARS - 3] + "..."
         return f"ConnectorCheckpoint(checkpoint_content={content_str}, has_more={self.has_more})"
```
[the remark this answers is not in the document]
evan-onyx: Good point, I kinda autopiloted here

backend/onyx/connectors/google_drive/models.py
```diff
@@ -20,3 +24,75 @@ class GDriveMimeType(str, Enum):
...
+    completion_map: ThreadSafeDict[str, SecondsSinceUnixEpoch] = ThreadSafeDict()
+
+    # cached version of the drive and folder ids to retrieve
+    drive_ids_to_retrieve: list[str] | None = None
+    folder_ids_to_retrieve: list[str] | None = None
+
+    # cached user emails
+    user_emails: list[str] | None = None
+
+    # @field_serializer("completion_map")
```
Weves: nit: remove if unused

backend/onyx/connectors/interfaces.py
```diff
@@ -186,14 +192,21 @@ def handle_event(self, event: Any) -> GenerateDocumentsOutput:
         raise NotImplementedError
 
 
-class CheckpointConnector(BaseConnector):
+CT = TypeVar("CT", bound=ConnectorCheckpoint)
+# TODO: find a reasonable way to parameterize the return type of the generator
+CheckpointOutput: TypeAlias = Generator[
```
Weves, line 197: should we remove/comment out if not usable due to generic annoying-ness?

backend/onyx/utils/threadpool_concurrency.py
```diff
@@ -190,3 +308,30 @@ def wait_on_background(task: TimeoutThread[R]) -> R:
...
+        future_to_index: dict[Future[tuple[int, R | None]], int] = {
+            executor.submit(_next_or_none, i, g): i for i, g in enumerate(gens)
+        }
+
+        next_ind = len(gens)
+        while future_to_index:
+            done, _ = wait(future_to_index, return_when=FIRST_COMPLETED)
+            for future in done:
+                ind, result = future.result()
+                print(ind, result)
```
Weves: nit: remove `print`

backend/tests/daily/connectors/google_drive/consts_and_utils.py
```diff
@@ -202,3 +204,13 @@ def assert_retrieved_docs_match_expected(
         retrieved=valid_retrieved_texts,
     )
     assert expected_file_texts == valid_retrieved_texts
+
+
+def load_all_docs(connector: GoogleDriveConnector) -> list[Document]:
```
Weves, line 218: this only loads all docs from the first checkpoint if I understand the function correctly? If so, should probably be re-named

backend/onyx/connectors/google_drive/connector.py
```diff
@@ -361,25 +395,27 @@ def _impersonate_user_for_retrieval(
...
-        elif self.include_shared_drives:
-            drive_ids_to_retrieve = all_drive_ids
+        if checkpoint.completion_stage == DriveRetrievalStage.START:
+            checkpoint.completion_stage = DriveRetrievalStage.USER_EMAILS
+
+        if checkpoint.completion_stage == DriveRetrievalStage.USER_EMAILS:
+            all_org_emails: list[str] = self._get_all_user_emails()
+            if not is_slim:
+                checkpoint.user_emails = all_org_emails
+            checkpoint.completion_stage = DriveRetrievalStage.DRIVE_IDS
```
Weves, line 532: we should `return` checkpoint here to save our progress

backend/onyx/connectors/google_drive/file_retrieval.py
```diff
@@ -243,16 +287,22 @@ def get_all_files_for_oauth(
     yield from execute_paginated_retrieval(
```
Weves, line 290: to properly checkpoint this, we should return back a new checkpoint every X calls?

Weves: similar comment for most of the other flows

Weves: nvm on ^, looking into the short-circuited approach we discussed

backend/onyx/connectors/google_drive/connector.py
```diff
@@ -499,46 +579,163 @@ def _manage_oauth_retrieval(
...
+                            executor.submit(convert_func, file) for file in files_batch
+                        ]
+                        documents = []
+                        for future in as_completed(futures):
+                            try:
+                                doc = future.result()
+                                if doc is not None:
+                                    documents.append(doc)
+                            except Exception as e:
+                                logger.error(f"Error converting file: {e}")
```
Weves: we should probably yield a failure here? E.g. if a single document fails to be downloaded

backend/onyx/connectors/google_drive/connector.py
```diff
@@ -499,46 +579,163 @@ def _manage_oauth_retrieval(
...
+                            yield documents
+                            batches_complete += 1
+                        files_batch = []
+
+                        if batches_complete > BATCHES_PER_CHECKPOINT:
+                            checkpoint.retrieved_ids = list(self._retrieved_ids)
+                            return  # create a new checkpoint
+
+                # Process any remaining files
+                if files_batch:
```
Weves, line 976: same note on yielding a failure below

backend/onyx/connectors/google_drive/connector.py
```diff
@@ -553,40 +750,52 @@ def _extract_docs_from_google_drive(
...
-                        logger.error(f"Error converting file: {e}")
-
-                if documents:
-                    yield documents
-
-    def load_from_state(self) -> GenerateDocumentsOutput:
-        try:
-            yield from self._extract_docs_from_google_drive()
         except Exception as e:
+            logger.exception(f"Error extracting documents from Google Drive: {e}")
```
Weves, line 1005: if we reach here, it's likely a terminal / massive failure, right? The current implementation, would could mark the index attempt as a "success with failures" despite only completing a few docs out of millions?

If so, we should probably just raise the exception regardless, and error out loudly

backend/tests/daily/connectors/google_drive/consts_and_utils.py
evan-onyx: It's a testing function and it doesn't take a checkpoint as a parameter so I feel like it's pretty self-explanatory? Would you want something like load_all_docs_from_empty or _from_first_checkpoint?

backend/onyx/connectors/interfaces.py
evan-onyx: CT is still being used, I figure this is worth leaving in for when python typing libraries get better

backend/onyx/connectors/google_drive/connector.py
```diff
@@ -238,6 +241,25 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, str] | None
 
         return new_creds_dict
 
+    def _checkpoint_yield(
+        self,
+        drive_files: Iterator[GoogleDriveFileType],
+        checkpoint: GoogleDriveCheckpoint,
+        key: Callable[
```
Weves: prefer a more descriptive name here (`key` is not clear if it's a function or an actual key + what exactly it does)

backend/onyx/connectors/google_drive/connector.py
```diff
@@ -330,8 +353,11 @@ def _impersonate_user_for_retrieval(
                 service=drive_service,
                 update_traversed_ids_func=self._update_traversed_parent_ids,
                 is_slim=is_slim,
+                checkpoint=checkpoint,
                 start=start,
                 end=end,
+                key=lambda check: user_email
```
Weves: is there a realistic reason why someone would call this function with a different key? If so, this flexibility feels unnecessary and probably makes it harder to understand what's going on in `get_all_files_in_my_drive`. E.g. the reader of `get_all_files_in_my_drive` has to look at where it was called to understand the expected behavior

backend/onyx/connectors/google_drive/connector.py
```diff
@@ -341,15 +367,23 @@ def _impersonate_user_for_retrieval(
                 service=drive_service,
                 drive_id=drive_id,
                 is_slim=is_slim,
+                checkpoint=checkpoint,
                 update_traversed_ids_func=self._update_traversed_parent_ids,
                 start=start,
                 end=end,
+                key=lambda check: user_email
```
Weves: same comment here as above

backend/onyx/connectors/google_drive/connector.py
Weves: as a general rule, I think it would be good to return after each stage. It's slightly safer (e.g. we don't lose the progress of fetching all user emails if the first call of the next step fails)

Weves (commented):
As a general rule, we prefer to not have functions modify objects in place and have the caller's functionality depend on these "hidden" modifications. It's a much clearer contract (and as a result, hopefully less bugs) if a function's behavior is clearly defined by its input and output (I'm sure you're already familiar with all the theory/rationale behind it, but just re-iterating the key point imo). 

All of these functions appear to modify the checkpoint in place in a way that effects the caller. Instead, would it be possible to yield/return back statuses which allow the parent function to modify the deep copied checkpoint -> none of the dependence on in-place modifications of objects passed in.

backend/tests/daily/connectors/google_drive/consts_and_utils.py
Weves: well it doesn't actually load all docs, right? It loads from "Zero Checkpoint" to "Checkpoint 1". So it could be a tiny subset of all docs to load

backend/onyx/agents/agent_search/deep_search/main/nodes/extract_entities_terms.py
```diff
@@ -106,6 +106,7 @@ def extract_entities_terms(
         first_bracket = cleaned_response.find("{")
         last_bracket = cleaned_response.rfind("}")
         cleaned_response = cleaned_response[first_bracket : last_bracket + 1]
+        cleaned_response += "g8o723dg87db"
```
Weves: what's going on here 🧐

backend/onyx/connectors/google_drive/connector.py
```diff
@@ -287,22 +342,96 @@ def get_all_drive_ids(self) -> set[str]:
...
+            if drive_id in self._retrieved_ids
+            else DriveIdStatus.AVAILABLE
+            for drive_id in drive_ids
+        }
+
+        def _get_available_drive_id(
+            processed_ids: set[str], thread_id: str
+        ) -> tuple[str | None, bool]:
+            found_future_work = False
+            print(f"{thread_id} drive_id_status: {drive_id_status}")
```
Weves: nit: remove print

backend/onyx/connectors/google_drive/connector.py
```diff
@@ -113,7 +122,54 @@ def _clean_requested_drive_ids(
...
+
+class DriveIdStatus(Enum):
+    AVAILABLE = "available"
+    IN_PROGRESS = "in_progress"
+    FINISHED = "finished"
+
+
+class DriveIdServer:
+    """
+    Returns an iterator over drive_ids.
```
Weves: How does this actually work? I don't see an `__iter__` implementation

Weves: seems unused? If so, should remove

backend/onyx/connectors/google_drive/connector.py
```diff
@@ -287,22 +342,96 @@ def get_all_drive_ids(self) -> set[str]:
...
         logger.info(f"Impersonating user {user_email}")
-
+        curr_stage = checkpoint.completion_map.get(user_email, None)
+        resuming = True
+        if curr_stage is None:
+            curr_stage = StageCompletion(
+                stage=DriveRetrievalStage.MY_DRIVE_FILES,
+                completed_until=start or 0,
+            )
+            checkpoint.completion_map[user_email] = curr_stage
```
Weves: this looks like we still have some of the behavior from before? E.g. the correct functioning of this connector depends on the lower level functions modifying the object in place?

Weves (commented):
We may want to spend some time cleaning this up. This single class is a bit massive with a very deep chain with somewhat redundant sounding functions (e.g. `load_from_checkpoint` -> `_extract_docs_from_google_drive` -> `_fetch_drive_items` -> `_checkpointed_retrieval` -> `_manage_service_account_retrieval` -> `_impersonate_user_for_retrieval` -> `get_all_files_in_my_drive`.

Maybe some of the functions can be flatted and/or the service account retrieval/oauth retrieval can be split into separate classes (or functions) or maybe different files.

Not for now since we do want to get this out ASAP, but something to think about in the future.

backend/onyx/connectors/google_drive/connector.py
evan-onyx: whoops yeah I switched to a different approach, this was just left over

backend/onyx/connectors/google_drive/connector.py
evan-onyx: Correct functionality doesn't depend on this at all, I'll just delete this line

Weves (approved):

PR #4263, opened 2025-03-12
Fix slash mystery

pablonyx:
## Description

Fix unintentional backslash for each dollar sign (below is a before and after)

Screenshot 2025-03-11 at 6 46 21 PM

Screenshot 2025-03-11 at 6 46 28 PM

Fixes https://linear.app/danswer/issue/DAN-1563/fix-weird-formatting-with-in-llm-output

## How Has This Been Tested?

- Many latex equations, many dollars signs in scripts


evan-danswer (commented):
Why was most of the markdown parsing deleted? This seems concerning to me 😅

evan-danswer (approved):
Got confirmation irl that the deleted code wasn't being used, LGTM

PR #4280, opened 2025-03-13
possible fix for confluence query filter

rkuo-danswer:
## Description

Fixes DAN-1570. (We hope)
https://linear.app/danswer/issue/DAN-1570/search-issues-in-older-versions-of-confluence

## How Has This Been Tested?


evan-danswer (approved):
ideally we could confirm that i.e. gifs were being retrieved with the current query, but up to you whether it's worth the time in this case!

PR #4304, opened 2025-03-19
Address invalid connector state

Weves:
## Description

Fixes https://linear.app/danswer/issue/DAN-1615/connectors-going-into-invalid-state-all-the-time

## How Has This Been Tested?

Verified indexing still works locally + 👀 


backend/onyx/background/indexing/run_indexing.py
```diff
@@ -596,16 +598,44 @@ def _run_indexing(
...
+                            index_attempt
+                            for index_attempt in recent_index_attempts
+                            if index_attempt.error_msg
+                            and index_attempt.error_msg.startswith(
+                                CONNECTOR_VALIDATION_ERROR_MESSAGE_PREFIX
+                            )
+                        ]
+                    )
+
+                    if num_validation_errors >= VALIDATION_ERROR_THRESHOLD:
```
rkuo-danswer, line 628: not in scope for this but i feel like this is reinforcing that we want to get this kind of state and db updating out from the indexing process itself.

backend/onyx/background/celery/tasks/external_group_syncing/tasks.py
```diff
@@ -426,12 +420,8 @@ def connector_external_group_sync_generator_task(
...
             except ConnectorValidationError as e:
                 msg = f"Error syncing external groups for {source_type} for cc_pair: {cc_pair_id} {e}"
-                update_connector_credential_pair(
-                    db_session=db_session,
-                    connector_id=cc_pair.connector.id,
-                    credential_id=cc_pair.credential.id,
-                    status=ConnectorCredentialPairStatus.INVALID,
-                )
+                # TODO: add some notification to the admins here
+                logger.exception(msg)
```
rkuo-danswer: if we're not doing anything else with msg, could just move the f-string into the logger.exception call.

rkuo-danswer (approved):

PR #4306, opened 2025-03-19
Notion improvement

Weves:
## Description

Fixes https://linear.app/danswer/issue/DAN-1546/notion-indexing-issue.

Turns on recursive indexing by default + makes it so we don't skip empty pages w/ titles.

## How Has This Been Tested?

Ran locally + UT


backend/onyx/connectors/notion/connector.py
```diff
@@ -38,8 +39,7 @@
 # TODO: Tables need to be ingested, Pages need to have their metadata ingested
 
 
-@dataclass
```
Weves, line 41: standardize to pydantic to match rest of the codebase

PR #4307, opened 2025-03-19
Checkpointed GitHub connector

evan-danswer:
## Description

addresses https://linear.app/danswer/issue/DAN-1617/checkpointed-github-connector
Checkpointing for the Github Connector

## How Has This Been Tested?

not tested yet; planning to write integration test and test in UI


backend/onyx/connectors/github/connector.py
```diff
@@ -234,68 +239,131 @@ def _fetch_from_github(
...
-            if self.include_issues:
+                    # if we went past the start date during the loop or there are no more
+                    # prs to get, we move on to issues
+                    if done_with_prs or len(pr_batch) == 0:
+                        checkpoint.stage = GithubConnectorStage.ISSUES
+                        checkpoint.curr_page = 0
+                        break
+
+                    # if we have enough PRs, yield them and return the checkpoint
+                    if len(doc_batch) >= self.batch_size:
```
Weves: we probably don't need to respect the batch_size anymore (can just remove it completely). Ideally, we would have each checkpoint be a single "page" for maximum efficiency. 

What is controlled by `INDEX_BATCH_SIZE` is now handled downstream in a connector-independent way.

backend/onyx/connectors/github/connector.py
```diff
@@ -234,68 +239,131 @@ def _fetch_from_github(
             # All repositories
             repos = self._get_all_repos(self.github_client)
 
+        # sorting allows saving less state
+        repos = sorted(repos, key=lambda x: x.id)
```
Weves: could we just find the specific repo we're on and then have the logic below not be within a `for` loop? Feels simpler that way

evan-onyx: you mean do a literal .find() with a repository object? Imo not worth the call to create the object? I could take this out into a helper function so it doesn't clutter this main function up though

evan-onyx: Agree it feels like there should be a better way though, am I missing a super-pythonic way to do this?

evan-onyx: Actually I found a binary search library, makes it a bit cleaner

backend/onyx/connectors/github/connector.py
```diff
@@ -122,19 +115,44 @@ def _convert_issue_to_document(issue: Issue) -> Document:
...
+    headers: dict[str, str | int]
+    raw_data: dict[str, Any]
+
+    def to_Repository(self, requester: Requester) -> Repository.Repository:
+        return Repository.Repository(
+            requester, self.headers, self.raw_data, completed=True
+        )
+
+
+# Repository.Repository(self.github_client._Github__requester, repos[0]._headers, repos[0].raw_data, completed=True)
```
Weves: nit: remove

backend/onyx/connectors/github/connector.py
```diff
@@ -217,85 +235,197 @@ def _get_all_repos(
...
+                return checkpoint
+
+            checkpoint.cached_repo_ids = sorted([repo.id for repo in repos])
+            checkpoint.cached_repo = SerializedRepository(
+                id=checkpoint.cached_repo_ids[0],
+                headers=repos[0].raw_headers,
+                raw_data=repos[0].raw_data,
+            )
+            checkpoint.stage = GithubConnectorStage.PRS
+            checkpoint.curr_page = 0
```
Weves, line 274: imo for proper checkpointing we should `return` here to save the progress of having successfully fetched all repos. Practically, this probably doesn't matter but does set a good standard

backend/onyx/connectors/github/connector.py
```diff
@@ -217,85 +235,197 @@ def _get_all_repos(
...
+                        and pr.updated_at
+                        and pr.updated_at.replace(tzinfo=timezone.utc) > end
+                    ):
+                        continue
+                    try:
                         doc_batch.append(_convert_pr_to_document(cast(PullRequest, pr)))
-                    yield doc_batch
+                    except Exception as e:
+                        error_msg = f"Error converting PR to document: {e}"
+                        logger.error(error_msg)
```
Weves: nit: prefer `logger.exception("Error converting PR to document")`. 

`logger.exception` automatically displays full stack trace + exception nicely

backend/onyx/connectors/github/connector.py
```diff
@@ -217,85 +235,197 @@ def _get_all_repos(
...
+            issues = repo.get_issues(
+                state=self.state_filter, sort="updated", direction="desc"
+            )
 
-            if self.include_issues:
-                logger.info(f"Fetching issues for repo: {repo.name}")
-                issues = repo.get_issues(
-                    state=self.state_filter, sort="updated", direction="desc"
+            doc_batch = []
+            while True:
```
Weves: same comment here

backend/onyx/connectors/github/connector.py
```diff
@@ -217,85 +235,197 @@ def _get_all_repos(
...
-                            continue
-                        if issue.pull_request is not None:
-                            # PRs are handled separately
-                            continue
+                    try:
                         doc_batch.append(_convert_issue_to_document(issue))
-                    yield doc_batch
+                    except Exception as e:
+                        error_msg = f"Error converting issue to document: {e}"
+                        logger.error(error_msg)
```
Weves: same comment here

backend/onyx/connectors/github/connector.py
```diff
@@ -217,85 +235,197 @@ def _get_all_repos(
...
+        repo = checkpoint.cached_repo.to_Repository(self.github_client.requester)
+
+        if self.include_prs and checkpoint.stage == GithubConnectorStage.PRS:
+            logger.info(f"Fetching PRs for repo: {repo.name}")
+            pull_requests = repo.get_pulls(
+                state=self.state_filter, sort="updated", direction="desc"
+            )
+
+            doc_batch: list[Document] = []
+            while True:
```
Weves: do we even need this while loop? Can't we just reduce indenting by one here since after every batch we always return

evan-onyx: We return if any documents were found; this just ensures progress is made with each checkpoint

Weves (approved):

PR #4311, opened 2025-03-20
zendesk checkpointed connector

evan-danswer:
## Description

Addresses https://linear.app/danswer/issue/DAN-1616/checkpointed-zendesk-connector
Adds checkpointing to the Zendesk connector

## How Has This Been Tested?

Tested in UI and unit tests

### DIFF
--- a/backend/onyx/connectors/zendesk/connector.py
+++ b/backend/onyx/connectors/zendesk/connector.py
@@ -384,7 +384,7 @@ class ZendeskConnector(SlimConnector, CheckpointConnector[ZendeskConnectorCheckp
         if self.client is None:
             raise ZendeskCredentialsNotSetUpError()
 
-        if not checkpoint.cached_content_tags:
+        if checkpoint.cached_content_tags is None:
             checkpoint.cached_content_tags = _get_content_tag_mapping(self.client)
             return checkpoint  # save the content tags to the checkpoint
         self.content_tags = checkpoint.cached_content_tags
--- a/backend/onyx/connectors/zendesk/connector.py
+++ b/backend/onyx/connectors/zendesk/connector.py
@@ -343,6 +343,7 @@ class ZendeskConnectorCheckpoint(ConnectorCheckpoint):
     after_cursor: str | None
     next_start_time: int | None
     cached_author_map: dict[str, BasicExpertInfo] | None
+    cached_content_tags: dict[str, str] | None
 
 
 class ZendeskConnector(SlimConnector, CheckpointConnector[ZendeskConnectorCheckpoint]):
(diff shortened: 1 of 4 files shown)


backend/onyx/connectors/zendesk/connector.py
```diff
@@ -278,13 +321,17 @@ def _ticket_to_document(
     )
 
 
-class ZendeskConnector(LoadConnector, PollConnector, SlimConnector):
+class ZendeskConnectorCheckpoint(ConnectorCheckpoint):
+    after_cursor: str | None
```
Weves: do you know if zendesk cursors expire? E.g. if we pause onyx and return after a few days, will this be able to seamlessly pick up where we left off?

backend/onyx/connectors/zendesk/connector.py
```diff
@@ -304,104 +370,161 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+        end: SecondsSinceUnixEpoch,
+        checkpoint: ZendeskConnectorCheckpoint,
+    ) -> CheckpointOutput[ZendeskConnectorCheckpoint]:
         if self.client is None:
             raise ZendeskCredentialsNotSetUpError()
 
-        self.content_tags = _get_content_tag_mapping(self.client)
+        if not checkpoint.cached_content_tags:
+            checkpoint.cached_content_tags = _get_content_tag_mapping(self.client)
+        self.content_tags = checkpoint.cached_content_tags
```
Weves, line 390: prefer returning here to save progress

backend/onyx/connectors/zendesk/connector.py
Weves: also can we clarify that `after_cursor` is only for articles and `next_start_time` is only for tickets? Prefer if we can put that in the name of the field

backend/onyx/connectors/zendesk/connector.py
```diff
@@ -304,104 +370,161 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
-                continue
-
-            new_author_map, documents = _article_to_document(
-                article, self.content_tags, author_map, self.client
+        author_map: dict[str, BasicExpertInfo] = checkpoint.cached_author_map or {}
+        after_cursor = checkpoint.after_cursor
+        doc_batch: list[Document] = []
+
+        # ensure at least one document is retrieved (connector makes progress)
+        while len(doc_batch) == 0:
```
Weves: similar to github, I don't think we need to ensure one document is retrieved, as long as we make progress (e.g. move forward the after_cursor/next_start_time) that should be okay.

Weves: we run the risk of the "not respecting start time", but if that doesn't work we have a big problem anyways

backend/onyx/connectors/zendesk/connector.py
evan-onyx: Pretty sure it just persists, see https://developer.zendesk.com/documentation/api-basics/pagination/paginating-through-lists-using-cursor-pagination/#python-example-with-the-after-cursor

"After you reach the end of the records, you can save the value of next or after_cursor to retrieve new records in the future" and yeah will leave a comment

evan-onyx: (and change var names)

Weves (approved):

PR #4317, opened 2025-03-21
Handle move errors

Weves:
## Description

Fixes https://linear.app/danswer/issue/DAN-1627/fix-ramp-startup-issue-with-huggingface-dir

## How Has This Been Tested?

👁️ 


rkuo-danswer (approved):
If not critical, maybe a logger.warning would be more appropriate? depends on the context.

PR #4345, opened 2025-03-25
Daylight savings time handling

evan-danswer:
## Description

Error handling for Daylight Savings and improvements to default timezone guess

## How Has This Been Tested?

Will rely on automated tests for this because it's difficult to test. 

### DIFF
--- a/backend/onyx/configs/app_configs.py
+++ b/backend/onyx/configs/app_configs.py
@@ -1,6 +1,8 @@
 import json
 import os
 import urllib.parse
+from datetime import datetime
+from datetime import timezone
 from typing import cast
 
 from onyx.auth.schemas import AuthBackend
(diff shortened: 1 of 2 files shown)


backend/onyx/configs/app_configs.py
```diff
@@ -383,10 +385,23 @@
 # https://community.developer.atlassian.com/t/confluence-cloud-time-zone-get-via-rest-api/35954/16
 # https://jira.atlassian.com/browse/CONFCLOUD-69670
 
+
+def get_current_tz_offset() -> int:
+    # datetime now() gets local time, datetime.now(timezone.utc) gets UTC time.
+    # remove tzinfo to compare non-timezone-aware objects.
+    time_diff = datetime.now() - datetime.now(timezone.utc).replace(tzinfo=None)
```
Weves, line 392: out of curiosity, is there a way to do this like `datetime.now(timezone.local) - datetime.now(timezone.utc)`? That feels easier to understand

Weves (approved):

backend/onyx/configs/app_configs.py
evan-onyx: I tried a version that looked more like that, unfortunately it always evaluates to 0 since both datetimes represent the same time.

PR #4356, opened 2025-03-26
fix issue with drive connector service account indexing

evan-danswer:
## Description

We were overwriting the checkpoint map for service account indexing due to a last-minute change; this causes us to not overwrite those entries.

## How Has This Been Tested?

Tested in UI

### DIFF
--- a/backend/onyx/connectors/google_drive/connector.py
+++ b/backend/onyx/connectors/google_drive/connector.py
@@ -2,6 +2,9 @@ import copy
 import threading
 from collections.abc import Callable
 from collections.abc import Iterator
+from concurrent.futures import as_completed
+from concurrent.futures import ThreadPoolExecutor
+from datetime import datetime
 from enum import Enum
 from functools import partial
 from typing import Any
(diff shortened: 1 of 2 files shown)


backend/onyx/connectors/google_drive/connector.py
```diff
@@ -562,6 +564,8 @@ def _manage_service_account_retrieval(
         )
 
         for email in all_org_emails:
+            if email in checkpoint.completion_map:
```
Weves: Should this just happen ~line 547? E.g. this is all part of the "initial preparation phase)

backend/onyx/connectors/google_drive/connector.py
```diff
@@ -954,30 +960,15 @@ def _extract_docs_from_google_drive(
...
-                            )
-                        ]
-                    elif result is not None:
-                        documents.append(result)
-
-                if documents:
-                    yield documents
+                docs_and_failures = [result for result in results if result is not None]
+
+                if docs_and_failures:
```
Weves: Should this function just yield individual docs/failures rather than lists of them?

backend/onyx/connectors/google_drive/connector.py
```diff
@@ -988,30 +979,15 @@ def _extract_docs_from_google_drive(
...
+                    yield docs_and_failures
```
Weves: same comment here

Weves (approved):

PR #4368, opened 2025-03-27
Save all logs + add log persistence to most Onyx-owned containers

Weves:
## Description

Ref https://linear.app/danswer/issue/DAN-1670/resolve-alation-instance-crashing

## How Has This Been Tested?

Tested compose file locally + verified logs were persisted.


rkuo-danswer (approved):
Agree with greptile that it does seem a little bug prone to mount /var/log where lots of different logs could be going instead of just our logs?  But if it works it works.

PR #4369, opened 2025-03-27
refactor file extension checking and add test for blob s3

rkuo-danswer:
## Description

## How Has This Been Tested?

Fixes DAN-1671.
https://linear.app/danswer/issue/DAN-1671/blob-connector-shouldnt-index-images-fully


backend/onyx/file_processing/extract_file_text.py
```diff
@@ -49,27 +51,51 @@
...
+    + [
+        ".pdf",
+        ".docx",
+        ".pptx",
+        ".xlsx",
+        ".eml",
+        ".epub",
+        ".html",
+    ]
+)
```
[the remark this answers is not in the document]
evan-onyx: I agree with greptile here unless there's a good reason not to

backend/onyx/file_processing/extract_file_text.py
```diff
@@ -83,8 +109,18 @@ def is_valid_media_type(media_type: str) -> bool:
     return media_type in IMAGE_MEDIA_TYPES
 
 
-def is_valid_file_ext(ext: str) -> bool:
-    return ext in VALID_FILE_EXTENSIONS
+def is_accepted_file_ext(ext: str, ext_type: OnyxExtensionType) -> bool:
```
evan-onyx, line 104: would appreciate a comment here explaining why we need the consecutive set intersections; is there a reason not to do this with a dict mapping type -> extensions list instead?

backend/tests/daily/connectors/blob/test_blob_connector.py
```diff
@@ -0,0 +1,77 @@
...
+        if file_extension in DOCUMENT_FILE_EXTENSIONS:
+            assert len(section.text) > 0
+            continue
+
+        if file_extension in IMAGE_FILE_EXTENSIONS:
+            assert len(section.text) == 0
+            continue
+
+        # unknown extension
+        assert len(section.text) == 0
```
evan-onyx, line 77: Hooray Tests! 😄

evan-danswer (approved):
LGTM with a few nits!

backend/onyx/file_processing/extract_file_text.py
rkuo-danswer: yeah this makes no sense to build the list, we can just check the extension lists directly. fixing.

PR #4377, opened 2025-03-27
Bugfix/acl prefix

rkuo-danswer:
## Description

Fixes DAN-1682.
https://linear.app/danswer/issue/DAN-1682/mit-version-adding-improperly-formatted-acls

MIT version was setting ACL as user_email:user_email:email@anonymous.invalid
EE version didn't because it was constructing DocumentAccess directly.

Now forcing DocumentAccess.build for all flows and fixed how it works.

## How Has This Been Tested?

### DIFF
--- a/backend/onyx/access/access.py
+++ b/backend/onyx/access/access.py
@@ -18,13 +18,15 @@ def _get_access_for_document(
         document_id=document_id,
     )
 
-    return DocumentAccess.build(
+    doc_access = DocumentAccess.build(
         user_emails=info[1] if info and info[1] else [],
         user_groups=[],
         external_user_emails=[],
         external_user_group_ids=[],
         is_public=info[2] if info else False,
     )
+
+    return doc_access
 
 
 def get_access_for_document(
@@ -67,7 +69,7 @@ def _get_access_for_documents(
         for document_id, user_emails, is_public in document_access_info
     }
 
-    # Sometimes the document has not be indexed by the indexing job yet, in those cases
+    # Sometimes the document has not been indexed by the indexing job yet, in those cases
     # the document does not exist and so we use least permissive. Specifically the EE version
     # checks the MIT version permissions and creates a superset. This ensures that this flow
     # does not fail even if the Document has not yet been indexed.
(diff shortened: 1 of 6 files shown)


backend/onyx/access/access.py
```diff
@@ -18,14 +18,16 @@ def _get_access_for_document(
         document_id=document_id,
     )
 
-    return DocumentAccess.build(
+    doc_access = DocumentAccess.build(
```
evan-onyx, line 21: nit: prefer returning directly

backend/onyx/access/models.py
```diff
@@ -93,29 +105,32 @@ def build(
...
+            obj,
+            "external_user_emails",
+            {external_email for external_email in external_user_emails},
         )
+        object.__setattr__(
+            obj,
+            "external_user_group_ids",
+            {external_group_id for external_group_id in external_user_group_ids},
+        )
+        object.__setattr__(obj, "is_public", is_public)
```
evan-onyx, line 125: Is there a reason we have to use __setattr__ instead of setting attributes normally? If its a mypy thing could we instead cast the obj to i.e. ExternalAccess after creating it with object.__new__?

backend/tests/integration/tests/connector/test_connector_deletion.py
```diff
@@ -168,11 +174,15 @@ def test_connector_deletion(reset: None, vespa_client: vespa_fixture) -> None:
         verify_deleted=True,
     )
 
+    cc_pair_2_group_name_expected = []
+    if bool(os.getenv("ENABLE_PAID_ENTERPRISE_EDITION_FEATURES")):
```
evan-onyx: Is this intended to be false if the environment variable is set to the string "false"? I believe as is bool('false') will return True

evan-danswer (approved):
LGTM but left a few nits + questions!

backend/onyx/access/access.py
rkuo-danswer: I actually do not like doing this as it is difficult to see in a debugger.

backend/onyx/access/models.py
rkuo-danswer: apparently this is necessary because we used dataclass frozen!

backend/tests/integration/tests/connector/test_connector_deletion.py
rkuo-danswer: always get this falsiness stuff mixed up.  good catch.

PR #4384, opened 2025-03-28
DELETE endpoint for ingestion docs

trial-danswer:
## Description

Added a DELETE endpoint to the ingestion docs on the MIT version of Onyx that fully deletes a document and associated images from the vector index and the Postgres database.

## How Has This Been Tested?

Locally tested with documents with text-only sections, image-only sections, and combined sections. Verified that data was deleted in database tables, vector index, and file store.

### DIFF
--- a/backend/onyx/db/document.py
+++ b/backend/onyx/db/document.py
@@ -587,6 +587,7 @@ def delete_documents_complete__no_commit(
     """This completely deletes the documents from the db, including all foreign key relationships"""
 
     # Start by deleting the chunk stats for the documents
+    # BUG: Is there a reason this is called twice?
     delete_chunk_stats_by_connector_credential_pair__no_commit(
         db_session=db_session,
         document_ids=document_ids,
--- a/backend/onyx/server/onyx_api/ingestion.py
+++ b/backend/onyx/server/onyx_api/ingestion.py
@@ -202,6 +202,9 @@ def delete_ingestion_doc(
         filters=IndexFilters(access_control_list=None, tenant_id=tenant_id),
     )
 
+    # TODO: add error recovery so that if any of these operations fail,
+    #       we can revert back to the original state with all documents
+    #       and chunks intact.
     file_store = get_default_file_store(db_session)
     for chunk in chunks:
         if chunk.image_file_name:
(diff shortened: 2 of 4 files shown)


backend/onyx/db/document.py
```diff
@@ -587,6 +587,7 @@ def delete_documents_complete__no_commit(
     """This completely deletes the documents from the db, including all foreign key relationships"""
 
     # Start by deleting the chunk stats for the documents
+    # BUG: Is there a reason this is called twice?
     delete_chunk_stats_by_connector_credential_pair__no_commit(
```
yuhongsun96, line 591: Yes, this is a bug... good catch

backend/onyx/server/onyx_api/ingestion.py
```diff
@@ -163,3 +170,57 @@ def upsert_ingestion_doc(
...
+
+
+@router.delete("/ingestion/{document_id}")
+def delete_ingestion_doc(
+    document_id: str,
+    user: User | None = Depends(api_key_dep),
+    db_session: Session = Depends(get_session),
+) -> DeleteIngestionResult:
+    tenant_id = get_current_tenant_id()
+    if user is None or user.role != UserRole.ADMIN:
```
yuhongsun96, line 182: If the auth type setting is "disabled" then there is no sense of users but this functionality should be enabled. Basically there is no login and the 1 account that exists is like a superuser.

Be careful with this one but it should be easy when you understand it

backend/onyx/server/onyx_api/ingestion.py
```diff
@@ -163,3 +170,57 @@ def upsert_ingestion_doc(
...
+    file_store = get_default_file_store(db_session)
+    for chunk in chunks:
+        if chunk.image_file_name:
+            file_store.delete_file(chunk.image_file_name)
+
+    # Completely delete the document from the vector index.
+    doc_index.delete_single(
+        document.id,
+        tenant_id=tenant_id,
+        chunk_count=document.chunk_count,
```
yuhongsun96, line 217: What is the difference between passing in the actual count and none here?

backend/onyx/server/query_and_chat/chat_backend.py
```diff
@@ -683,6 +683,8 @@ def upload_files_for_chat(
             raise HTTPException(status_code=400, detail="File content type is required")
 
         if file.content_type not in allowed_content_types:
+            # BUG: The error detail code here seems incorrect: it seems like it only ever generates
```
yuhongsun96, line 686: True, probably some logic changed around it 🤷 , I would never write bugs of course 😅

We should fix this up

PR #4386, opened 2025-03-28
Bugfix/slack rate limiting

rkuo-danswer:
## Description

Fixes DAN-1619.
https://linear.app/danswer/issue/DAN-1619/address-rate-limiting-for-slack

## How Has This Been Tested?


backend/onyx/connectors/slack/onyx_retry_handler.py
```diff
@@ -0,0 +1,156 @@
...
+        error: Optional[Exception] = None,
+    ) -> None:
+        """it seems this function is responsible for the wait to retry ... aka we
+        actually sleep in this function."""
+        retry_after_value: list[str] | None = None
+        retry_after_header_name: Optional[str] = None
+        duration_s: float = 1.0  # seconds
+
+        if response is None:
+            if error:
```
evan-onyx, line 85: If only raising the error when response is None is actually the logic we want, could you put a comment explaining why it's fine to continue despite an error if the response exists? Otherwise I'd say check for and raise the error outside the check for response

backend/onyx/connectors/slack/onyx_retry_handler.py
```diff
@@ -0,0 +1,156 @@
...
+                raise error
+
+            return
+
+        state.next_attempt_requested = True  # this signals the caller to retry
+
+        # calculate wait duration based on retry-after + some jitter
+        for k in response.headers.keys():
```
evan-onyx, line 93: could we iterate over k,v in response.headers.values() and set the `retry_after_value` as well if the condition is met? Point being I don't think we need to store `retry_after_header_name` since it's only used to get the value from the dictionary anyways

backend/onyx/connectors/slack/onyx_retry_handler.py
```diff
@@ -0,0 +1,156 @@
...
+                raise ValueError(
+                    "OnyxRedisSlackRetryHandler.prepare_for_next_attempt: retry-after header value is None"
+                )
+
+            retry_after_value_int = int(
+                retry_after_value[0]
+            )  # will raise if somehow we can't convert to int
+            jitter = retry_after_value_int * 0.25 * random.random()
+            duration_s = math.ceil(retry_after_value_int + jitter)
+        except Exception:
```
evan-onyx: This feels a bit too broad to me; IMO better to `except ValueError` since that's what we expect to be raised by int parsing (and is also raised by your other two cases)

evan-danswer (approved):
Just a few nits, looks good!

backend/onyx/connectors/slack/onyx_retry_handler.py
rkuo-danswer: TBH i don't quite know and am just following the RateLimitErrorRetryHandler logic in the slack SDK that this was based on.  I can leave a comment to that effect.

backend/onyx/connectors/slack/onyx_retry_handler.py
rkuo-danswer: prefer the current way since it unnests some of the logic

backend/onyx/connectors/slack/onyx_retry_handler.py
rkuo-danswer: I'll change it to ValueError ... was being a bit conservative since this is being run by the slack sdk but probably ok.

PR #4390, opened 2025-03-29
Fix image input processing without LLMs

pablonyx:
## Description

Fixes https://linear.app/danswer/issue/DAN-1687/fix-image-input-for-google-drive

## How Has This Been Tested?


backend/onyx/indexing/chunker.py
```diff
@@ -264,7 +243,7 @@ def _chunk_document_with_sections(
 
         for section_idx, section in enumerate(sections):
             # Get section text and other attributes
-            section_text = clean_text(section.text or "")
+            section_text = clean_text(str(section.text or ""))
```
evan-onyx, line 246: It might be good to add error logs if section.text isn't a string or None saying what type it is to help us catch the root cause in the future

evan-danswer (approved):
LGTM, suggested adding error logs at the earliest point an unexpected type is found

PR #4413, opened 2025-04-01
Enable Support for LiteLLM Customer feature

emerzon:
## Description

This patch enables the usage of LiteLLM Customer feature (via an OpenAI provider) by sending out the user id as part of the LLM call.

## How Has This Been Tested?

When enabled, the user email should be received by LiteLLM and show up as a customer entry.

### DIFF
--- a/backend/onyx/agents/agent_search/models.py
+++ b/backend/onyx/agents/agent_search/models.py
@@ -36,6 +36,7 @@ class GraphTooling(BaseModel):
     # force tool args IF the tool is used
     force_use_tool: ForceUseTool
     using_tool_calling_llm: bool = False
+    user_id: str | None = None
 
     class Config:
         arbitrary_types_allowed = True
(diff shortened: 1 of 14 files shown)


backend/onyx/agents/agent_search/models.py
```diff
@@ -36,6 +36,7 @@ class GraphTooling(BaseModel):
     # force tool args IF the tool is used
     force_use_tool: ForceUseTool
     using_tool_calling_llm: bool = False
+    user_id: str | None = None
```
evan-onyx, line 39: would prefer to call this litellm_user_id to avoid confusion with user ids from other contexts

backend/onyx/llm/chat_llm.py
```diff
@@ -403,59 +404,61 @@ def _completion(
         structured_response_format: dict | None = None,
         timeout_override: int | None = None,
         max_tokens: int | None = None,
+        user_id: str | None = None,
```
evan-onyx, line 407: Please add this to the signatures for custom LLMs to maintain parity

evan-danswer (commented):
Looking good! Just a few nits from me

backend/onyx/llm/chat_llm.py
emerzon: @evan-danswer I am not sure to follow exactly whats needed here, could you please provide a few more details?

emerzon:
Bumping

PR #4414, opened 2025-04-01
Prune First, then Merge (Simpler Approach)

joachim-danswer:
In the Search pipeline, we need to first prune, and then merge. This desired behavior was not adhered to anymore and has been re-established.

 prompt_config and contextual_pruning _config both need to be set for this to work.

This fix is also related to https://linear.app/danswer/issue/DAN-1626/dalberg-projects-not-being-foundcited-as-expected
## Description

Key changes:
 - passed contextual_pruning_config to Search pipeline
 - used prune_and_merge to create final chunks
 - if prompt_config not done or contextual_pruning_config not available, default to earlier behavior of calling _merge_sections

## How Has This Been Tested?

Locally. Tested with default Search, and then also an Assistant that does not do AI filtering

### DIFF
--- a/backend/onyx/context/search/pipeline.py
+++ b/backend/onyx/context/search/pipeline.py
@@ -5,11 +5,13 @@ from typing import cast
 
 from sqlalchemy.orm import Session
 
+from onyx.chat.models import ContextualPruningConfig
 from onyx.chat.models import PromptConfig
 from onyx.chat.models import SectionRelevancePiece
 from onyx.chat.prune_and_merge import _merge_sections
 from onyx.chat.prune_and_merge import ChunkRange
 from onyx.chat.prune_and_merge import merge_chunk_intervals
+from onyx.chat.prune_and_merge import prune_and_merge_sections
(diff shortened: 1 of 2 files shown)


backend/onyx/context/search/pipeline.py
```diff
@@ -420,7 +426,26 @@ def final_context_sections(self) -> list[InferenceSection]:
...
+            self.contextual_pruning_config is not None
+            and self.prompt_config is not None
+        ):
+            self._final_context_sections = prune_and_merge_sections(
+                sections=self.reranked_sections,
+                section_relevance_list=None,
+                prompt_config=self.prompt_config,
+                llm_config=self.llm.config,
+                question=self.search_query.query,
+                contextual_pruning_config=self.contextual_pruning_config,
```
evan-onyx, line 439: Would be nice to have a test to make sure that the prompt config and contextual pruning config are passed through to here when a search tool is ran, but that's probably a backlog item

evan-danswer (approved):
looks good!

PR #4439, opened 2025-04-03
Fix weird re-index state

Weves:
## Description

Fixes https://linear.app/danswer/issue/DAN-1746/fix-index-swap-weirdness

## How Has This Been Tested?

Tested many index swaps locally

### DIFF
--- a/backend/onyx/db/connector_credential_pair.py
+++ b/backend/onyx/db/connector_credential_pair.py
@@ -613,6 +613,7 @@ def fetch_connector_credential_pairs(
 
 def resync_cc_pair(
     cc_pair: ConnectorCredentialPair,
+    search_settings_id: int,
     db_session: Session,
 ) -> None:
     def find_latest_index_attempt(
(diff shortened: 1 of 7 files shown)


backend/onyx/db/index_attempt.py
```diff
@@ -710,6 +710,25 @@ def cancel_indexing_attempts_past_model(
     )
 
 
+def cancel_indexing_attempts_for_search_settings(
+    search_settings_id: int,
+    db_session: Session,
+) -> None:
+    """Stops all indexing attempts that are in progress or not started for
+    any embedding model that not present/future"""
```
rkuo-danswer: that "is" not

backend/onyx/db/connector_credential_pair.py
```diff
@@ -613,6 +613,7 @@ def fetch_connector_credential_pairs(
 
 def resync_cc_pair(
     cc_pair: ConnectorCredentialPair,
+    search_settings_id: int,
```
rkuo-danswer, line 616: would help to docstring the intent here. it looks like we're changing so that we can filter index attempts by a specific search settings id?

rkuo-danswer (approved):

backend/onyx/db/index_attempt.py
Weves: done!

backend/onyx/db/connector_credential_pair.py
Weves: Makes sense, added

PR #4458, opened 2025-04-04
feat: Vertex AI support

raunakab:
## Description

This PR adds support for the Gemini/VertexAI LLM Provider.
Addresses https://linear.app/danswer/issue/DAN-1704/built-in-gemini-vertex-ai-support.

## How Has This Been Tested?

Primarily UI based changes; tested manually to make sure registration/deletion works.


hagen6835 (approved):
lgtm

backend/onyx/llm/llm_provider_options.py
```diff
@@ -1,12 +1,26 @@
...
 
 
+class CustomConfigKeyType(Enum):
+    # used for configuration values that require manual input
+    # i.e., textual API keys (e.g., "abcd1234")
+    TEXT_INPUT = "text_input"
+
+    # used for configuration values that require a file to be select/drag-and-dropped
+    # i.e., file based credentials (e.g., "/path/to/credentials/file.json")
+    DRAG_AND_DROP = "drag_and_drop"
```
Weves: nit: prefer `FILE_INPUT` feels more descriptive of what is actually different about this config type. Also, you technically can click and select rather than drag and drop 😅

backend/onyx/llm/llm_provider_options.py
```diff
@@ -77,17 +91,25 @@ class WellKnownLLMProviderDescriptor(BaseModel):
 AZURE_PROVIDER_NAME = "azure"
 
 
+VERTEXAI_PROVIDER_NAME = "vertex_ai"
+VERTEXAI_DEFAULT_MODEL = "gemini-1.5-pro-002"
+VERTEXAI_MODEL_NAMES = [
+    VERTEXAI_DEFAULT_MODEL,
```
Weves, line 105: I think there are other vertex models as well, no? E.g. the new gemini 2.5

web/src/app/admin/configuration/llm/LLMProviderUpdateForm.tsx
```diff
@@ -226,19 +229,16 @@ export function LLMProviderUpdateForm({
...
-          {!hideAdvanced && (
-            <TextFormField
-              name="name"
-              label="Display Name"
-              subtext="A name which you can use to identify this provider when selecting it in the UI."
-              placeholder="Display Name"
-              disabled={existingLlmProvider ? true : false}
-            />
-          )}
+          <TextFormField
```
Weves: I def agree that `hideAdvanced` shouldn't control this, but I think we still need a way to hide this. Basically, when a user first comes to the app, we don't want to ask them to put in a name, it slightly increases friction (see ApiKeyForm)

web/src/app/admin/configuration/llm/LLMProviderUpdateForm.tsx
```diff
@@ -257,98 +256,108 @@ export function LLMProviderUpdateForm({
...
+                  // like us.anthropic.claude-3-7-sonnet-20250219-v1:0 and anthropic.claude-3-7-sonnet-20250219-v1:0
+                  name: name,
+                  value: name,
+                }))}
+                includeDefault
+                maxHeight="max-h-56"
+              />
+            ) : (
               <TextFormField
-                small={hideAdvanced}
```
Weves, line 270: where is this used / why was it there? Want to make sure we're not impacting an existing flow 😅

backend/onyx/llm/llm_provider_options.py
raunakab: Good point! Will update in the coming commit.

backend/onyx/llm/llm_provider_options.py
raunakab: Yes, there are. I added a singular one early during dev but didn't come around to updating it later. Will add the 2.5 models as well!

backend/onyx/llm/llm_provider_options.py
raunakab: Updated!

web/src/app/admin/configuration/llm/LLMProviderUpdateForm.tsx
```diff
@@ -257,98 +256,108 @@ export function LLMProviderUpdateForm({
...
+                <FileUploadFormField
+                  key={customConfigKey.name}
+                  name={`custom_config.${customConfigKey.name}`}
+                  label={customConfigKey.display_name}
+                  subtext={customConfigKey.description || undefined}
+                />
+              );
+            } else {
+              throw "Unreachable; there should only exist 2 options";
+            }
```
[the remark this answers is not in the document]
raunakab: Updated.

web/src/app/admin/configuration/llm/LLMProviderUpdateForm.tsx
raunakab: Ah, you're right. I think I was a little bullish with the removal of this field. I've added it back (and renamed it to be a little bit more aligned with its purpose).

I've renamed it to `firstTimeConfiguration`. Do you think that's an appropriate name here? It's only used in `ApiKeyForm`, which I believe is only shown during first time configurations?

web/src/app/admin/configuration/llm/LLMProviderUpdateForm.tsx
raunakab: Added this back!

web/src/components/admin/connectors/Field.tsx
```diff
@@ -319,6 +356,39 @@ export function TextFormField({
...
+  label,
+  subtext,
+}: {
+  name: string;
+  label: string;
+  subtext?: string | JSX.Element;
+}) {
+  // We create a *temporary* field inside of `Formik` to throw the `File` object into.
+  // The actual *contents* of the file will be thrown into the field called `name`.
+  const fileName = "temporary.credentialsFile";
```
Weves: does this mean that only one `FileUploadFormField` can be used for a specific form? We could probably add the name to the fileName in some way to get around this if I'm understanding things correctly.

web/src/app/admin/configuration/llm/LLMProviderUpdateForm.tsx
Weves: Yep, that sounds great

backend/onyx/llm/llm_provider_options.py
raunakab: Added more models!

backend/onyx/llm/chat_llm.py
```diff
@@ -469,6 +476,12 @@ def _completion(
 
     @property
     def config(self) -> LLMConfig:
+        credentials_file: str | None = (
+            self._custom_config.get("CREDENTIALS_FILE", None)
```
Weves: nit: prefer for this to be a constant since it's used in two places

backend/onyx/llm/llm_provider_options.py
```diff
@@ -143,6 +192,26 @@ def fetch_available_well_known_llms() -> list[WellKnownLLMProviderDescriptor]:
             default_model="anthropic.claude-3-5-sonnet-20241022-v2:0",
             default_fast_model="anthropic.claude-3-5-sonnet-20241022-v2:0",
         ),
+        WellKnownLLMProviderDescriptor(
+            name=VERTEXAI_PROVIDER_NAME,
+            display_name="Gemini",
```
Weves: nit: Prefer calling it "GCP Vertex AI". It supports more models than just Gemini!

Weves (commented):
On startup, I'm seeing this. Previously there would not be the `Default Model` section. Would like to keep it that way to keep the initial set up as simple as possible.

Screenshot 2025-04-07 at 1 02 51 PM

backend/onyx/llm/llm_provider_options.py
raunakab: Updated!

backend/onyx/llm/chat_llm.py
raunakab: Updated!

web/src/components/admin/connectors/Field.tsx
raunakab: Good point, yes we can do that. I'll update the logic right now.

raunakab: Updated!

Weves (commented):
Screenshot 2025-04-07 at 3 58 30 PM

When editing, this is the view I see. I believe it should have all the same options as when creating a new one (e.g. be able to change the default model, display models, etc.)

Weves (approved):
lgtm 🍾

PR #4468, opened 2025-04-07
Id not set in checkpoint2

evan-danswer:
## Description

https://linear.app/danswer/issue/DAN-1762/more-drive-id-not-set-fixes

More drive connector improvements to cover cases where "folder id not set in checkpoint" might occur. One such common case is that a user doesn't have access to the first several folders being indexed, so despite that user's completion stage being set to "folders", they don't yield any documents from the first few folders, leading to no information being set in the checkpoint.

Also added a mypy change and associated fixes to prevent broken equality checks for the future

## How Has This Been Tested?

tested in UI

### DIFF
--- a/backend/onyx/agents/agent_search/deep_search/initial/generate_initial_answer/nodes/generate_initial_answer.py
+++ b/backend/onyx/agents/agent_search/deep_search/initial/generate_initial_answer/nodes/generate_initial_answer.py
@@ -117,16 +117,16 @@ def generate_initial_answer(
 
     consolidated_context_docs = structured_subquestion_docs.cited_documents
     counter = 0
-    for original_doc_number, original_doc in enumerate(
-        orig_question_retrieval_documents
-    ):
-        if original_doc_number not in structured_subquestion_docs.cited_documents:
-            if (
-                counter <= AGENT_MIN_ORIG_QUESTION_DOCS
-                or len(consolidated_context_docs) < AGENT_MAX_ANSWER_CONTEXT_DOCS
-            ):
-                consolidated_context_docs.append(original_doc)
-                counter += 1
+    for original_doc in orig_question_retrieval_documents:
+        if original_doc in structured_subquestion_docs.cited_documents:
+            continue
+
+        if (
+            counter <= AGENT_MIN_ORIG_QUESTION_DOCS
+            or len(consolidated_context_docs) < AGENT_MAX_ANSWER_CONTEXT_DOCS
+        ):
+            consolidated_context_docs.append(original_doc)
+            counter += 1
 
     # sort docs by their scores - though the scores refer to different questions
     relevant_docs = dedup_inference_section_list(consolidated_context_docs)
--- a/backend/onyx/background/celery/tasks/doc_permission_syncing/tasks.py
+++ b/backend/onyx/background/celery/tasks/doc_permission_syncing/tasks.py
@@ -98,9 +98,6 @@ def _is_external_doc_permissions_sync_due(cc_pair: ConnectorCredentialPair) -> b
     if cc_pair.status != ConnectorCredentialPairStatus.ACTIVE:
         return False
 
-    if cc_pair.status == ConnectorCredentialPairStatus.DELETING:
-        return False
-
     # If the last sync is None, it has never been run so we run the sync
     last_perm_sync = cc_pair.last_time_perm_sync
     if last_perm_sync is None:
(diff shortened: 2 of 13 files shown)


backend/onyx/agents/agent_search/deep_search/initial/generate_initial_answer/nodes/generate_initial_answer.py
```diff
@@ -117,16 +117,16 @@ def generate_initial_answer(
...
-    ):
-        if original_doc_number not in structured_subquestion_docs.cited_documents:
-            if (
-                counter <= AGENT_MIN_ORIG_QUESTION_DOCS
-                or len(consolidated_context_docs) < AGENT_MAX_ANSWER_CONTEXT_DOCS
-            ):
-                consolidated_context_docs.append(original_doc)
-                counter += 1
+    for original_doc in orig_question_retrieval_documents:
+        if original_doc in structured_subquestion_docs.cited_documents:
```
Weves, line 121: what's going on here?

backend/onyx/background/celery/tasks/doc_permission_syncing/tasks.py
```diff
@@ -98,9 +98,6 @@ def _is_external_doc_permissions_sync_due(cc_pair: ConnectorCredentialPair) -> b
     if cc_pair.status != ConnectorCredentialPairStatus.ACTIVE:
         return False
 
-    if cc_pair.status == ConnectorCredentialPairStatus.DELETING:
```
Weves, line 101: why remove?

backend/onyx/connectors/google_drive/file_retrieval.py
```diff
@@ -122,14 +122,16 @@ def crawl_folders_for_files(
...
-                found_files = True
                 logger.info(f"Found file: {file['name']}, user email: {user_email}")
+                found_files = True
                 yield RetrievedDriveFile(
                     drive_file=file,
                     user_email=user_email,
                     parent_id=parent_id,
                     completion_stage=DriveRetrievalStage.FOLDER_FILES,
                 )
+            if found_files:
```
Weves, line 134: why the move? If it's important, would prefer to add a comment as to why. If not / purely stylistic, ignore

backend/onyx/connectors/google_drive/connector.py
```diff
@@ -526,16 +527,28 @@ def _yield_from_folder_crawl(
...
+            last_processed_folder = None
             if resuming:
                 folder_id = curr_stage.completed_until_parent_id
                 assert folder_id is not None, "folder id not set in checkpoint"
                 resume_start = curr_stage.completed_until
                 yield from _yield_from_folder_crawl(folder_id, resume_start)
+                last_processed_folder = folder_id
+
+            skipping_seen_folders = last_processed_folder is not None
+            for folder_id in sorted(filtered_folder_ids):
```
Weves: could we merge the `if` statement and this `for` loop into one?

backend/onyx/agents/agent_search/deep_search/initial/generate_initial_answer/nodes/generate_initial_answer.py
evan-onyx: typing fixes

backend/onyx/background/celery/tasks/doc_permission_syncing/tasks.py
evan-onyx: Was never True, see lines above

backend/onyx/connectors/google_drive/file_retrieval.py
evan-onyx: previously we were marking folders as traversed if at least one file from the folder was retrieved without an error; now it will only be marked as done if ALL files from it are retrieved. With the new system for tracking folder completion (sorting and continuing from the last SEEN folder rather than last retrieved), this shouldn't cause us to get stuck and should let us handle pathological cases like a bunch of different users having individual access to files in a "shared folder" that isn't actually fully shared due to permission revoking.

backend/onyx/connectors/google_drive/connector.py
evan-onyx: It's probably doable, but imo it's better/easier to read to have separate logic that handles resuming from a checkpoint

Weves (approved):

PR #4493, opened 2025-04-10
fix claude bug

evan-danswer:
## Description

Fix issue where task prompt was included as part of the user query when selecting tool call args

## How Has This Been Tested?

tested in UI, code tests coming

### DIFF
--- a/backend/onyx/chat/tool_handling/tool_response_handler.py
+++ b/backend/onyx/chat/tool_handling/tool_response_handler.py
@@ -165,7 +165,6 @@ def get_tool_call_for_non_tool_calling_llm_impl(
     user_query = prompt_builder.raw_user_query
     history = prompt_builder.raw_message_history
     if isinstance(prompt_builder, AnswerPromptBuilder):
-        user_query = prompt_builder.get_user_message_content()
         history = prompt_builder.get_message_history()
 
     if force_use_tool.force_use:


pablonyx (approved):

backend/onyx/chat/tool_handling/tool_response_handler.py
```diff
@@ -165,7 +165,6 @@ def get_tool_call_for_non_tool_calling_llm_impl(
     user_query = prompt_builder.raw_user_query
     history = prompt_builder.raw_message_history
     if isinstance(prompt_builder, AnswerPromptBuilder):
-        user_query = prompt_builder.get_user_message_content()
```
raunakab, line 168: Could I get some context here, please? What's the point of deleting this line?

evan-onyx: get_user_message_content() adds the task prompt (a prompt piece specified during assistant/persona creation; the search assistant has one that we wrote) to the user query. This is Good during the final prompt to the LLM asking for an answer, but the function that gets tool call args for non-tool-calling LLMs expects only the raw user query to be passed in. This was actually a holdover from a previous approach to a fix in which having the task prompt attached to the user query was actually desired behavior, but a later different approach (adding system prompt to the history, actually) made that unnecessary and I forgot to remove this part 😭

PR #4505, opened 2025-04-11
fix max 10 drives issue

evan-danswer:
## Description

Most likely addresses https://linear.app/danswer/issue/DAN-1547/gdrive-missing-some-shared-drives-during-indexing

We weren't asking google for a nextPageToken in the drives list request, causing us to index at most 10 shared drives (the default pageSize)

## How Has This Been Tested?

tested in UI

### DIFF
--- a/backend/onyx/indexing/chunker.py
+++ b/backend/onyx/indexing/chunker.py
@@ -1,3 +1,5 @@
+from llama_index.core.node_parser import SentenceSplitter
+
 from onyx.configs.app_configs import AVERAGE_SUMMARY_EMBEDDINGS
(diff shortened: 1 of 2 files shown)


rkuo-danswer (approved):

backend/onyx/indexing/chunker.py
```diff
@@ -1,3 +1,5 @@
+from llama_index.core.node_parser import SentenceSplitter
```
Weves, line 1: why the change?

evan-onyx: imports in functions should be avoided imo

PR #4531, opened 2025-04-15
Fix foreign key

pablonyx:
## Description

## How Has This Been Tested?

### DIFF
--- a/backend/onyx/db/search_settings.py
+++ b/backend/onyx/db/search_settings.py
@@ -18,6 +18,7 @@ from onyx.db.enums import EmbeddingPrecision
 from onyx.db.llm import fetch_embedding_provider
 from onyx.db.models import CloudEmbeddingProvider
 from onyx.db.models import IndexAttempt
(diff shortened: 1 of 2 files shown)


backend/onyx/db/search_settings.py
```diff
@@ -118,13 +119,27 @@ def delete_search_settings(db_session: Session, search_settings_id: int) -> None
...
+    # First, identify the index attempts to be deleted
+    index_attempts = (
+        db_session.query(IndexAttempt)
+        .filter(IndexAttempt.search_settings_id == search_settings_id)
+        .all()
+    )
+
+    # Delete associated error records first
+    for attempt in index_attempts:
+        error_records_query = delete(IndexAttemptError).where(
```
evan-onyx, line 132: could we use `IN` here instead of python looping? or is that worse in some way

evan-danswer (approved):
qq but looks good! Kind of wondering whether this could/should be handled with a cascading delete instead

PR #4533, opened 2025-04-15
fix: Fix duplicate kwarg issue when calling `litellm.main.completion`

raunakab:
## Description

This PR fixes the error in which duplicate `vertex_credentials` kwargs would be passed into the call to `litellm.main.completion`. This resulted in an error being thrown while registering a custom LLM provider.

Addresses: https://linear.app/danswer/issue/DAN-1819/fix-duplicate-vertex-credentials-kwarg-error-while-creating-a-vertexai.

## How Has This Been Tested?

Ran through Custom LLM Provider registration process while also providing a "vertex_credentials" custom config key.

### DIFF
--- a/backend/onyx/llm/chat_llm.py
+++ b/backend/onyx/llm/chat_llm.py
@@ -457,14 +457,14 @@ class DefaultMultiLLM(LLM):
                     if structured_response_format
                     else {}
                 ),
-                **(
-                    {
-                        "vertex_credentials": self.config.credentials_file,
-                    }
-                    if self.config.model_provider == "vertex_ai"
-                    else {}
-                ),
-                **self._model_kwargs,
+                **{
+                    "vertex_credentials": (
+                        self.config.credentials_file
+                        if self.config.model_provider == "vertex_ai"
+                        else None
+                    ),
+                    **self._model_kwargs,
+                },
             )
         except Exception as e:
             self._record_error(processed_prompt, e)
--- a/backend/onyx/llm/chat_llm.py
+++ b/backend/onyx/llm/chat_llm.py
@@ -412,6 +412,9 @@ class DefaultMultiLLM(LLM):
         processed_prompt = _prompt_to_dict(prompt)
         self._record_call(processed_prompt)
 
+        if "vertex_credentials" not in self._model_kwargs:
+            self._model_kwargs["vertex_credentials"] = self.config.credentials_file
+
         try:
             return litellm.completion(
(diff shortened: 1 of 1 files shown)


backend/onyx/llm/chat_llm.py
```diff
@@ -412,6 +412,9 @@ def _completion(
         processed_prompt = _prompt_to_dict(prompt)
         self._record_call(processed_prompt)
 
+        if "vertex_credentials" not in self._model_kwargs:
+            self._model_kwargs["vertex_credentials"] = self.config.credentials_file
```
Weves: let's not modify this in place. E.g. something like:

```
final_model_kwargs = {**self._model_kwargs}
if "vertex_credentials" not in final_model_kwargs:
            final_model_kwargs["vertex_credentials"] = self.config.credentials_file
            
...
```

Weves (changes requested):

backend/onyx/llm/chat_llm.py
raunakab: Ah okay, will update!

raunakab: Updated.

backend/onyx/llm/chat_llm.py
```diff
@@ -412,6 +412,10 @@ def _completion(
         processed_prompt = _prompt_to_dict(prompt)
         self._record_call(processed_prompt)
 
+        final_model_kwargs = {**self._model_kwargs}
+        if "vertex_credentials" not in final_model_kwargs:
```
Weves: nit: prefer to make this a constant at the top of the file, since it's used multiple times now

Weves (approved):

backend/onyx/llm/chat_llm.py
raunakab: Hmm, could you please elaborate on this? Do you mean that `final_model_kwargs` should be a global that is empty, and then we populate it inside of this function? Or `final_model_kwargs` should be a member of this class?

Weves: sorry, to clarify: that "vertex_credentials" should be a constant!

raunakab: Ahh I see. Yes, let me update it to be a constant prior to merging this in.

raunakab: Sorry, forgot to update this. Suggestion has been implemented.

PR #4535, opened 2025-04-16
Skip image files for S3

pablonyx:
## Description

Fixes https://linear.app/danswer/issue/DAN-1821/images-for-s3

## How Has This Been Tested?

### DIFF
--- a/backend/onyx/connectors/blob/connector.py
+++ b/backend/onyx/connectors/blob/connector.py
@@ -52,6 +52,11 @@ class BlobStorageConnector(LoadConnector, PollConnector):
         self.prefix = prefix if not prefix or prefix.endswith("/") else prefix + "/"
         self.batch_size = batch_size
         self.s3_client: Optional[S3Client] = None
+        self._allow_images = None
+
+    def set_allow_images(self, allow_images: bool) -> None:
+        """Set whether to process images in this connector."""
+        self._allow_images = allow_images
 
     def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None:
         """Checks for boto3 credentials based on the bucket type.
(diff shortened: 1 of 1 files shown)


backend/onyx/connectors/blob/connector.py
```diff
@@ -178,6 +185,9 @@ def _yield_blob_objects(
         if self.s3_client is None:
             raise ConnectorMissingCredentialError("Blob storage")
 
+        # Check if image extraction and analysis is enabled
+        process_images = get_image_extraction_and_analysis_enabled()
```
rkuo-danswer: the connector can implement set_allow_images as is done in the confluence connector for example.

backend/onyx/connectors/blob/connector.py
```diff
@@ -195,22 +205,65 @@ def _yield_blob_objects(
...
+                        logger.debug(
+                            f"Skipping image file: {key} (image processing not enabled)"
+                        )
+                        continue
+
+                    # Process the image file
+                    try:
+                        downloaded_file = self._download_object(key)
+
+                        with get_session_with_current_tenant() as db_session:
```
rkuo-danswer, line 229: the connector shouldn't be accessing the db directly and the task of syncing data should probably be tenant unaware, but maybe that's something that needs clean up everywhere?

backend/onyx/connectors/blob/connector.py
```diff
@@ -220,9 +273,7 @@ def _yield_blob_objects(
                         batch = []
 
                 except Exception as e:
-                    logger.exception(
-                        f"Error decoding object {obj['Key']} as UTF-8: {e}"
-                    )
+                    logger.exception(f"Error decoding object {key} as UTF-8: {e}")
```
rkuo-danswer: logger.exception already logs the exception

backend/onyx/connectors/blob/connector.py
pablonyx: Unfortunately, would require broader refactoring- creating a relevant ticket

backend/onyx/connectors/blob/connector.py
```diff
@@ -178,6 +191,13 @@ def _yield_blob_objects(
         if self.s3_client is None:
             raise ConnectorMissingCredentialError("Blob storage")
 
+        # Check if image extraction and analysis is enabled
+        process_images = (
+            self._allow_images
+            if self._allow_images is not None
+            else get_image_extraction_and_analysis_enabled()
```
rkuo-danswer: i don't think we should be calling this if we have allow_images available.

rkuo-danswer (approved):

PR #4542, opened 2025-04-16
refactor a mega function for readability and make sure to increment r…

rkuo-danswer:
…etry_count on exception so that we don't infinitely loop

## Description

Fixes https://linear.app/danswer/issue/DAN-1829/web-connector-can-infinitely-loop

## How Has This Been Tested?

### DIFF
--- a/backend/onyx/connectors/web/connector.py
+++ b/backend/onyx/connectors/web/connector.py
@@ -18,6 +18,7 @@ from oauthlib.oauth2 import BackendApplicationClient
 from playwright.sync_api import BrowserContext
 from playwright.sync_api import Playwright
 from playwright.sync_api import sync_playwright
+from pydantic import BaseModel
 from requests_oauthlib import OAuth2Session  # type:ignore
 from urllib3.exceptions import MaxRetryError
 
@@ -42,6 +43,27 @@ from onyx.utils.sitemap import list_pages_for_site
 from shared_configs.configs import MULTI_TENANT
 
 logger = setup_logger()
+
+
+class ScrapeContext(BaseModel):
+    base_url: str
+    to_visit: list[str]
+    playwright: Playwright | None = None
+    playwright_context: BrowserContext | None = None
+
+    retry_count: int = 0
+    retry_success: bool = False
+
+    visited_links: set[str] = set()
+    content_hashes: set[int] = set()
+
+    doc_batch: list[Document] = []
+
+    restart_playwright: bool = False
+
+    at_least_one_doc: bool = False
+    last_error: str | None = None
+
 
 WEB_CONNECTOR_MAX_SCROLL_ATTEMPTS = 20
 # Threshold for determining when to replace vs append iframe content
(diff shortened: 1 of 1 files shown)


backend/onyx/connectors/web/connector.py
```diff
@@ -444,259 +478,250 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+        True if the page scraped normally"""
+
+        if session_ctx.playwright is None:
+            raise RuntimeError("scrape_context.playwright is None")
+
+        if session_ctx.playwright_context is None:
+            raise RuntimeError("scrape_context.playwright_context is None")
+
+        if session_ctx.restart_playwright:
+            session_ctx.playwright, session_ctx.playwright_context = start_playwright()
```
Weves: prefer not having a function modify a passed in object in place (avoidable side effects outside of the direct input/output of the method). 

Could we copy and return that (as an example)?

backend/onyx/connectors/web/connector.py
```diff
@@ -444,259 +478,250 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+        # Sometimes pages with #! will serve duplicate content
+        # There are also just other ways this can happen
+        hashed_text = hash((parsed_html.title, parsed_html.cleaned_text))
+        if hashed_text in session_ctx.content_hashes:
+            logger.info(
+                f"{index}: Skipping duplicate title + content for {initial_url}"
+            )
+            return False
+
+        session_ctx.content_hashes.add(hashed_text)
```
Weves: same comments here

backend/onyx/connectors/web/connector.py
rkuo-danswer: would rather just modify it outside the function actually i think, but can copy and return in the next comment

backend/onyx/connectors/web/connector.py
```diff
@@ -444,261 +484,248 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
             logger.warning("Unexpected credentials provided for Web Connector")
         return None
 
+    def _do_scrape(
+        self,
+        index: int,
+        initial_url: str,
+        session_ctx_in: ScrapeSessionContext,
+    ) -> ScrapeSessionContext:
+        """Returns False if something happened and the page scrape can be retried.
```
Weves: comment seems out of date?

backend/onyx/connectors/web/connector.py
```diff
@@ -444,261 +484,248 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
             logger.warning("Unexpected credentials provided for Web Connector")
         return None
 
+    def _do_scrape(
+        self,
+        index: int,
+        initial_url: str,
+        session_ctx_in: ScrapeSessionContext,
+    ) -> ScrapeSessionContext:
```
Weves: as discussed we can just return the retrieved document id

Weves (approved):

PR #4543, opened 2025-04-16
small improvement to checking for image attachments

rkuo-danswer:
## Description

Fixes https://linear.app/danswer/issue/DAN-1816/confluence-image-attachments-being-downloaded-unexpectedly

## How Has This Been Tested?

### DIFF
--- a/backend/onyx/connectors/confluence/utils.py
+++ b/backend/onyx/connectors/confluence/utils.py
@@ -59,15 +59,16 @@ def validate_attachment_filetype(
     """
     Validates if the attachment is a supported file type.
     """
-    attachment.get("metadata", {})
     media_type = attachment.get("metadata", {}).get("mediaType", "")
-
     if media_type.startswith("image/"):
         return is_valid_image_type(media_type)
 
     # For non-image files, check if we support the extension
     title = attachment.get("title", "")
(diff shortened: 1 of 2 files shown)


backend/onyx/connectors/confluence/utils.py
```diff
@@ -59,16 +63,17 @@ def validate_attachment_filetype(
...
     if media_type.startswith("image/"):
         return is_valid_image_type(media_type)
 
     # For non-image files, check if we support the extension
     title = attachment.get("title", "")
     extension = Path(title).suffix.lstrip(".").lower() if "." in title else ""
-    return extension in ["pdf", "doc", "docx", "txt", "md", "rtf"]
+
+    return is_accepted_file_ext(
+        "." + extension, OnyxExtensionType.Plain | OnyxExtensionType.Document
```
evan-onyx, line 75: This does incorporate quite a few extensions over the old code, but as long as you think it's fine I don't see an issue

evan-danswer (approved):
minor note but lgtm

backend/onyx/connectors/confluence/utils.py
rkuo-danswer: discussed with pablo, we think this is correct for now.

PR #4552, opened 2025-04-17
v1 file connector with metadata

evan-danswer:
## Description

Fixes https://linear.app/danswer/issue/DAN-1836/fix-file-connector-metadata

## How Has This Been Tested?

Tested in UI, daily connector test in the works

### DIFF
--- a/backend/onyx/connectors/file/connector.py
+++ b/backend/onyx/connectors/file/connector.py
@@ -1,5 +1,4 @@
 import os
-from collections.abc import Iterator
 from datetime import datetime
 from datetime import timezone
 from pathlib import Path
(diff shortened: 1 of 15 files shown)


backend/onyx/connectors/file/connector.py
```diff
@@ -32,30 +30,22 @@
 logger = setup_logger()
 
 
-def _read_files_and_metadata(
+def _read_file_from_postgres(
```
Weves, line 33: nit: prefer not naming it this, since it might not be postgres soon 🙏 

E.g. `_read_file_from_filestore`

Weves (approved):

PR #4559, opened 2025-04-18
Small improvements to connector UI

Weves:
## Description

Fixes https://linear.app/danswer/issue/DAN-1841/improve-connector-display-part-2

## How Has This Been Tested?

### DIFF
--- a/backend/onyx/server/documents/models.py
+++ b/backend/onyx/server/documents/models.py
@@ -293,7 +293,13 @@ class CCPairFullInfo(BaseModel):
             ),
             last_pruned=cc_pair_model.last_pruned,
             last_permission_sync=(
-                last_index_attempt.time_started if last_index_attempt else None
+                min(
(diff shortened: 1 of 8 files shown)


backend/onyx/server/documents/models.py
```diff
@@ -293,7 +293,13 @@ def from_models(
             ),
             last_pruned=cc_pair_model.last_pruned,
             last_permission_sync=(
-                last_index_attempt.time_started if last_index_attempt else None
+                min(
```
Weves: 😅

evan-onyx: any issue with some connectors not running either type of sync?

evan-danswer (approved):

backend/onyx/server/documents/models.py
Weves: good catch, addressed

PR #4571, opened 2025-04-21
fix: LaTeX processing issue in #4570

### LINKED ISSUE #4570: LaTeX delimiters incorrectly escaped causing "$2" to appear instead of "$" in rendered math expressions
## Description
There is a bug in the message rendering component where LaTeX expressions that use dollar sign delimiters are being incorrectly escaped, causing the rendered output to show `\$2x = 1 \$2` instead of the proper `$ x = 1 $`.

- Onyx version: [v0.27.0-beta.1](https://github.com/onyx-dot-app/onyx/releases/tag/v0.27.0-beta.1) 
- OS: macOS Sequoia 15.4.1 arm64 and Ubuntu 24.04.2 LTS aarch64

### Steps to Reproduce
Have tested on both my MacBook and remoted Ubuntu server, and I can reproduce on both machines.
1. Have an AI response that includes LaTeX expressions in the format `\( x = 1 \)`
2. Observe that after rendering, the expression appears as `\$2x = 1 \$2` instead of `$ x = 1 $`

Image

### Logs/Console Output

The following codes are involved in processing the LaTeX expressions of LLM messages.

https://github.com/onyx-dot-app/onyx/blob/b97628070e90329f323d3c9355a1c4efe2b3571f/web/src/app/chat/message/Messages.tsx#L343-L346

I added console.log statements to track the transformation of content through the processing pipeline. The logs clearly show that the issue occurs during the dollar sign escaping phase:

```
6894-92a79cacd8924a89.js:1 Initial content: To output \( x = 1 \) in an inline math equation, you can write it as \( x = 1 \).
6894-92a79cacd8924a89.js:1 Before preprocessLaTeX: To output \( x = 1 \) in an inline math equation, you can write it as \( x = 1 \).
6894-92a79cacd8924a89.js:1 After preprocessLaTeX: To output $ x = 1 $ in an inline math equation, you can write it as $ x = 1 $.
6894-92a79cacd8924a89.js:1 Before escaping dollar signs: To output $ x = 1 $ in an inline math equation, you can write it as $ x = 1 $.
6894-92a79cacd8924a89.js:1 After escaping dollar signs: To output \$2x = 1 \$2in an inline math equation, you can write it as \$2x = 1 \$2
6894-92a79cacd8924a89.js:1 Final result: To output \$2x = 1 \$2in an inline math equation, you can write it as \$2x = 1 \$2
```

## Root Cause
The issue is in the `processContent` function where dollar signs are being escaped:

```javascript
const escapedDollarSigns = processed.replace(/([\s])\$([^\$])/g, "$1\\$$2");
```

The replacement pattern `"$1\\$$2"` is being interpreted incorrectly in the production environment. The `\\$` part (meant to be an escaped dollar sign) is followed by `$2` (a reference to the second capture group), but this is causing the rendered output to include a literal "2" character after the dollar sign.

Also, in PR #4259, the processing functions for Dollar Signs have been extensively tested; hence, `escapedDollarSigns` is not needed here.

## Proposed Fix

Remove these two lines:
https://github.com/onyx-dot-app/onyx/blob/b97628070e90329f323d3c9355a1c4efe2b3571f/web/src/app/chat/message/Messages.tsx#L345-L346

ZhipengHe:
## Description

fix #4570: simplify LaTeX processing by removing unnecessary dollar sign escaping

## How Has This Been Tested?

Has been tested on local machine. The issue has been fixed.

image

### DIFF
--- a/web/src/app/chat/message/Messages.tsx
+++ b/web/src/app/chat/message/Messages.tsx
@@ -342,12 +342,7 @@ export const AIMessage = ({
     }
     const processed = preprocessLaTeX(content);
 
-    // Escape $ that are preceded by a space and followed by a non-$ character
-    const escapedDollarSigns = processed.replace(/([\s])\$([^\$])/g, "$1\\$$2");
-
-    return (
-      escapedDollarSigns + (!isComplete && !toolCallGenerating ? " [*]() " : "")
-    );
+    return processed + (!isComplete && !toolCallGenerating ? " [*]() " : "");
   };
 
   const finalContentProcessed = processContent(finalContent as string);


ZhipengHe:
The issue has been fixed in PR #4519 
I can now close this PR

PR #4575, opened 2025-04-22
fix: Return default value instead of throwing error

raunakab:
## Description

The `/admin/llm/test` API tried to fetch the model configurations from an existing LLM Provider, but would fail if it wasn't available. However, this is used during LLM Provider creation, so no existing LLM Provider should be available in the first place.

This PR fixes that bug by using the default `max_input_tokens` instead.

Addresses: https://linear.app/danswer/issue/DAN-1858/fix-failure-on-no-llm-provider-present-during-llm-provider-creation.

## How Has This Been Tested?

Tested that registration still succeeds.

### DIFF
--- a/backend/onyx/db/llm.py
+++ b/backend/onyx/db/llm.py
@@ -17,6 +17,7 @@ from onyx.db.models import SearchSettings
 from onyx.db.models import Tool as ToolModel
 from onyx.db.models import User
 from onyx.db.models import User__UserGroup
+from onyx.llm.utils import get_max_input_tokens
 from onyx.llm.utils import get_max_input_tokens_from_llm_provider
 from onyx.llm.utils import model_supports_image_input
 from onyx.server.manage.embedding.models import CloudEmbeddingProvider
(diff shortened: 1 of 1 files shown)


backend/onyx/server/manage/llm/api.py
```diff
@@ -62,23 +62,33 @@ def test_llm_configuration(
...
+            name=test_llm_request.name, db_session=db_session
         )
         if existing_provider:
             test_api_key = existing_provider.api_key
 
-    max_input_tokens = fetch_max_input_tokens(
-        db_session=db_session,
-        provider_name=test_llm_request.provider,
-        model_name=test_llm_request.name or test_llm_request.default_model_name,
+    model_name = (
```
Weves: why the change here compared to what it was before?

raunakab: Ah you're right, I can pull it back out into the helper function to make the code more reusable.

raunakab: Will update in the coming commit.

raunakab: The previous helper function was calling `fetch_existing_llm_provider` again, which would perform another DB call.

raunakab: Updated to use a dummy-value instead (i.e., `-1`).

PR #4579, opened 2025-04-22
no more duplicate files during folder indexing

evan-onyx:
## Description

Addresses https://linear.app/danswer/issue/DAN-1851/duplicate-docs-in-google-drive-connector
previously, multiple threads could work on a shared folder at once and yield back the same doc multiple times. Now, we only yield docs that haven't been yielded previously.

## How Has This Been Tested?

tested in UI + added a connector test

### DIFF
--- a/backend/onyx/connectors/google_drive/connector.py
+++ b/backend/onyx/connectors/google_drive/connector.py
@@ -89,7 +89,7 @@ def _clean_requested_drive_ids(
     requested_drive_ids: set[str],
     requested_folder_ids: set[str],
     all_drive_ids_available: set[str],
-) -> tuple[set[str], set[str]]:
+) -> tuple[list[str], list[str]]:
     invalid_requested_drive_ids = requested_drive_ids - all_drive_ids_available
     filtered_folder_ids = requested_folder_ids - all_drive_ids_available
     if invalid_requested_drive_ids:
@@ -100,7 +100,7 @@ def _clean_requested_drive_ids(
         filtered_folder_ids.update(invalid_requested_drive_ids)
 
     valid_requested_drive_ids = requested_drive_ids - invalid_requested_drive_ids
-    return valid_requested_drive_ids, filtered_folder_ids
+    return sorted(valid_requested_drive_ids), sorted(filtered_folder_ids)
 
 
 class CredentialedRetrievalMethod(Protocol):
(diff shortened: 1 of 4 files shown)


backend/onyx/connectors/google_drive/connector.py
```diff
@@ -85,11 +85,19 @@ def _extract_ids_from_urls(urls: list[str]) -> list[str]:
     return [urlparse(url).path.strip("/").split("/")[-1] for url in urls]
 
 
+def _cleanup_folder_file_ids(
```
Weves: what's the point of this? Can we add a comment describing it?

backend/onyx/connectors/google_drive/connector.py
```diff
@@ -485,7 +494,7 @@ def _yield_from_drive(
             def _yield_from_folder_crawl(
```
Weves, line 501: for the future — thoughts on breaking things out to separate `folder_based_retrieval.py`, `shared_drive_based_retrieval.py`, etc. files? A little hard to tell what's folder, what's shared drive, etc.

backend/onyx/connectors/google_drive/connector.py
Weves: Also, the fact that we have to do this, feels a little suspect? And why do files need to be different than regular files, can't we have multiple users try and retrieve the same file at the same time when just selecting all drives?

evan-onyx: We don't have to do this; it's an optimization to keep the checkpoint small. And yeah we could do this to some extent for all drives, but in practice that'll lead to enormous checkpoints. Part of the implicit assumption here is that when people specify a specific folder to index, they a) really want that specific folder indexed quickly and correctly and b) the scale isn't so massive enough to make the checkpoint unfathomably large. (b) Probably isn't met by larger orgs when indexing their whole drive, but we could do this for the rest of the drive files too if we really want no duplicate docs in any part of the pipeline (although for now it would really only affect Shared with Me, since shared drives are indexed by at most one user)

backend/onyx/connectors/google_drive/connector.py
evan-onyx: That's one way to do it; another option I've been thinking about is that there isn't actually a lot different between the different retrieval methods. There might be a reasonable way to unify a lot of our logic so it isn't so gross

PR #4589, opened 2025-04-23
Confluence server api time fix

evan-onyx:
## Description

Addresses https://linear.app/danswer/issue/DAN-1899/confluence-documents-indexed-not-matching-total-docs
(and a series of other confluence customer issues)

There's an issue with the Confluence Server API; there doesn't appear to be a consistent field in page responses that mirrors the "lastmodified" key used during the query. This means that sometimes the field in response pages that we use for lastmodified can be outside the specified range. To fix this, we're switching to page-start-based checkpointing for confluence server.

Confluence Cloud  ALSO returns slightly wrong lastmodified values (though in testing they were less off than server), and uses cursor-based pagination. We currently believe that confluence uses a stateless model for cursor pagination, so the cursors never expire.

So, the current approach is to only store the "next page url" in the checkpoint. 

## How Has This Been Tested?

Ran in UI + unit tests

### DIFF
--- a/backend/onyx/connectors/confluence/connector.py
+++ b/backend/onyx/connectors/confluence/connector.py
@@ -70,6 +70,8 @@ _RESTRICTIONS_EXPANSION_FIELDS = [
 
 _SLIM_DOC_BATCH_SIZE = 5000
 
+MAX_CACHED_IDS = 100
+
 ONE_HOUR = 3600
 
 
@@ -469,7 +471,9 @@ class ConfluenceConnector(
         end = (end or time.time()) + 3600 * 24
         checkpoint = copy.deepcopy(checkpoint)
         prev_doc_ids = checkpoint.last_seen_doc_ids
-        checkpoint.last_seen_doc_ids = []
+
+        seen_cutoff = max(0, len(prev_doc_ids) - MAX_CACHED_IDS)
+        checkpoint.last_seen_doc_ids = checkpoint.last_seen_doc_ids[seen_cutoff:]
         # use "start" when last_updated is 0
         start_ts = checkpoint.last_updated or start
         page_query = self._construct_page_query(start_ts, end)
@@ -483,26 +487,29 @@ class ConfluenceConnector(
             limit=2 * self.batch_size,
         ):
             # create checkpoint after enough documents have been processed
-            # if yield_count >= self.batch_size:
-            #     return checkpoint
+            if yield_count >= self.batch_size:
+                return checkpoint
 
             if page["id"] in prev_doc_ids:
-                # There are a few seconds of fuzziness in the request,
-                # so we skip if we saw this page on the last run
+                # There is a discrepancy between the lastmodified field confluence uses
+                # for the query and the actual modification time files they return.
+                # We skip the page if it's within the last MAX_CACHED_IDS pages
                 continue
 
             ts = datetime_from_string(page["version"]["when"]).timestamp()
 
-            # if ts < checkpoint.last_updated:
-            #     logger.warning(
-            #         f"Confluence Returned results out of order. Request start time: {start_ts}, "
-            #         f"current item time: {ts}, checkpoint.last_updated: {checkpoint.last_updated}"
-            #     )
-            #     continue
+            if ts < checkpoint.last_updated:
+                # NOTE: this is a known issue on confluence server deployments. We mitigate by
+                # not counting
+                logger.debug(
+                    f"Confluence Returned results out of order. Request start time: {start_ts}, "
+                    f"current item time: {ts}, checkpoint.last_updated: {checkpoint.last_updated}"
+                )
+            else:
+                yield_count += 1
 
             # Build doc from page
             doc_or_failure = self._convert_page_to_document(page)
-            yield_count += 1
 
             if isinstance(doc_or_failure, ConnectorFailure):
                 yield doc_or_failure
--- a/backend/onyx/connectors/confluence/connector.py
+++ b/backend/onyx/connectors/confluence/connector.py
@@ -483,8 +483,8 @@ class ConfluenceConnector(
             limit=2 * self.batch_size,
         ):
             # create checkpoint after enough documents have been processed
-            if yield_count >= self.batch_size:
-                return checkpoint
+            # if yield_count >= self.batch_size:
+            #     return checkpoint
 
             if page["id"] in prev_doc_ids:
                 # There are a few seconds of fuzziness in the request,
--- a/backend/onyx/connectors/confluence/connector.py
+++ b/backend/onyx/connectors/confluence/connector.py
@@ -19,6 +19,7 @@ from onyx.connectors.confluence.utils import build_confluence_document_id
 from onyx.connectors.confluence.utils import convert_attachment_to_content
 from onyx.connectors.confluence.utils import datetime_from_string
 from onyx.connectors.confluence.utils import process_attachment
+from onyx.connectors.confluence.utils import update_param_in_path
 from onyx.connectors.confluence.utils import validate_attachment_filetype
 from onyx.connectors.exceptions import ConnectorValidationError
 from onyx.connectors.exceptions import CredentialExpiredError
@@ -81,12 +82,8 @@ def _should_propagate_error(e: Exception) -> bool:
 
 
 class ConfluenceCheckpoint(ConnectorCheckpoint):
-    # Used for Confluence Cloud checkpointing
-    last_updated: SecondsSinceUnixEpoch
-    last_seen_doc_ids: list[str]  # avoid duplicate page processing
 
-    # used for Confluence Server checkpointing
-    next_start: int
+    next_page_url: str | None
 
 
 class ConfluenceConnector(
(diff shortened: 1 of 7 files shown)


backend/onyx/connectors/confluence/onyx_confluence.py
```diff
@@ -387,6 +387,7 @@ def wrapped_call(*args: list[Any], **kwargs: Any) -> Any:
 
     def __getattr__(self, name: str) -> Any:
         """Dynamically intercept attribute/method access."""
+        print("hello")
```
Weves: greetings are great, but should probably remove

backend/onyx/connectors/confluence/onyx_confluence.py
```diff
@@ -562,6 +554,7 @@ def cql_paginate_all_expansions(
         The limit only applies to the top level query.
         All expansion paginations use default pagination limit (defined by Atlassian).
         """
+        print(f"Paginating through expansions for {cql}")
```
Weves: remove prints

backend/tests/unit/onyx/connectors/confluence/test_confluence_checkpointing.py
```diff
@@ -366,35 +417,35 @@ def test_checkpoint_progress(
         confluence_connector, 0, end_time
     )
 
-    assert len(outputs) == 1
+    for output in outputs:
```
Weves: nit: remove

backend/onyx/connectors/confluence/connector.py
```diff
@@ -551,6 +540,7 @@ def retrieve_all_slim_documents(
         Return 'slim' docs (IDs + minimal permission data).
         Does not fetch actual text. Used primarily for incremental permission sync.
         """
+        print("Retrieving all slim documents")
```
Weves: nit: remove

backend/onyx/connectors/confluence/onyx_confluence.py
```diff
@@ -407,22 +408,23 @@ def __getattr__(self, name: str) -> Any:
...
 
-        return wrapped_method
+        return rate_limited_method
 
     def _paginate_url(
-        self, url_suffix: str, limit: int | None = None, auto_paginate: bool = False
+        self,
+        url_suffix: str,
+        limit: int | None = None,
+        next_page_callback: Callable[[int], None] | None = None,
```
Weves: could we add a comment which says what this gets called with

backend/onyx/connectors/confluence/onyx_confluence.py
```diff
@@ -541,13 +528,18 @@ def paginated_cql_retrieval(
         cql: str,
         expand: str | None = None,
         limit: int | None = None,
+        start: int = 0,
+        next_page_callback: Callable[[int], None] | None = None,
```
Weves: same comment here as above

backend/onyx/connectors/confluence/connector.py
```diff
@@ -70,15 +70,23 @@
...
+TIME_OFFSET = 120
+MAX_CACHED_IDS = 100
 
 
 def _should_propagate_error(e: Exception) -> bool:
     return "field 'updated' is invalid" in str(e)
 
 
 class ConfluenceCheckpoint(ConnectorCheckpoint):
+    # Used for Confluence Cloud checkpointing
```
Weves: how do we handle the fact that the schema has changed?

evan-onyx: handled in connector runner logic

backend/onyx/connectors/confluence/connector.py
```diff
@@ -461,66 +466,65 @@ def _fetch_document_batches(
...
+        page_query_url = checkpoint.next_page_url or self._build_page_retrieval_url(
+            start_ts, end, self.batch_size
+        )
+        logger.debug(f"page_query_url: {page_query_url}")
+
+        # store the next page start for confluence server, cursor for confluence cloud
+        def store_next_page_url(next_page_url: str) -> None:
+            print("STORING NEXT PAGE URL")
+            print(next_page_url)
+            checkpoint.next_page_url = next_page_url
```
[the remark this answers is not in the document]
Weves: +1

backend/onyx/connectors/confluence/onyx_confluence.py
```diff
@@ -494,10 +500,29 @@ def _paginate_url(
...
             results = cast(list[dict[str, Any]], next_response.get("results", []))
-            yield from results
-
+            # make sure we don't update the start by more than the amount
+            # of results we were able to retrieve. The Confluence API has a
+            # weird behavior where if you pass in a limit that is too large for
+            # the configured server, it will artificially limit the amount of
+            # results returned BUT will not apply this to the start parameter.
+            # This will cause us to miss results.
+            # print(f"next_response: {next_response}")
```
Weves: nit: remove

backend/tests/unit/onyx/connectors/confluence/test_confluence_checkpointing.py
```diff
@@ -165,6 +176,11 @@ def test_load_from_checkpoint_happy_path(
     )
 
     # Check that the documents were returned
+    for output in outputs:
+        print("OUTPUT")
+        print(output)
+        print()
+        print()
```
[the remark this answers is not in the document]
Weves: +1

backend/tests/unit/onyx/connectors/confluence/test_confluence_checkpointing.py
```diff
@@ -404,12 +461,8 @@ def test_checkpoint_progress(
     )
 
     # Verify only the new page was processed since the others were in last_seen_doc_ids
-    assert len(outputs_with_checkpoint) == 1
+    assert len(outputs_with_checkpoint) == 1 or len(outputs_with_checkpoint) == 2
```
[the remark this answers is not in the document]
Weves: +1

backend/tests/unit/onyx/connectors/confluence/test_confluence_checkpointing.py
```diff
@@ -268,12 +304,24 @@ def test_retrieve_all_slim_documents(
...
+    confluence_client.get = get_mock  # type: ignore
+    get_mock.side_effect = [
+        # First page response
+        MagicMock(
+            json=lambda: {
+                "results": [mock_page1, mock_page2],
+                "_links": {"next": "rest/api/content/search?cql=type=page&start=2"},
+            }
+        ),
+        MagicMock(json=lambda: {"results": []}),
```
Weves, line 316: can we add quick comments for all of these cases as well? 😁

backend/tests/unit/onyx/connectors/confluence/test_confluence_checkpointing.py
```diff
@@ -146,16 +148,25 @@ def test_load_from_checkpoint_happy_path(
...
```
Weves, line 162: same here

backend/tests/unit/onyx/connectors/confluence/test_confluence_checkpointing.py
```diff
@@ -41,21 +41,22 @@ def space_key() -> str:
...
     """Create a mock Confluence client with proper typing"""
-    mock = MagicMock(spec=OnyxConfluence)
-    # Initialize with empty results for common methods
-    mock.paginated_cql_retrieval.return_value = []
-    mock.get_all_spaces = MagicMock()
-    mock.get_all_spaces.return_value = {"results": []}
-    return mock
+    # Server mode just Also updates the start value
+    return OnyxConfluence(
+        is_cloud=False, url="test", credentials_provider=MagicMock(), timeout=None
```
Weves, line 48: why is this one false, but the below is `is_cloud=True` 🧐

backend/tests/daily/connectors/confluence/test_confluence_basic.py
```diff
@@ -13,6 +13,14 @@
 from onyx.connectors.models import Document
 from tests.daily.connectors.utils import load_all_docs_from_checkpoint_connector
 
+EXPIRED_CURSOR = (
```
Weves: nit: remove since unused

backend/onyx/connectors/confluence/connector.py
```diff
@@ -70,15 +71,19 @@
 _SLIM_DOC_BATCH_SIZE = 5000
 
 ONE_HOUR = 3600
+ONE_DAY = ONE_HOUR * 24
+
+TIME_OFFSET = 120
```
Weves: nit: remove since these are unused

backend/onyx/connectors/confluence/connector.py
```diff
@@ -461,66 +466,65 @@ def _fetch_document_batches(
...
             yield doc_or_failure
 
+            # Create checkpoint once a full page of results is returned
+            if checkpoint.next_page_url and checkpoint.next_page_url != page_query_url:
+                return checkpoint
+
         checkpoint.has_more = False
         return checkpoint
 
+    def _build_page_retrieval_url(
```
Weves, line 511: [optional] prefer merging this function w/ `_construct_page_query`. They seem very similar, not sure why'd we'd call one over the other.

Alternatively, name them more distinctly / descriptively,. since just by the names I'm not sure which to use for what.

backend/tests/unit/onyx/connectors/confluence/test_confluence_checkpointing.py
evan-onyx: It used to be the case that this let us test both cases, but now it shouldn't matter! will change

backend/onyx/connectors/confluence/connector.py
evan-onyx: ONE_DAY is still used to handle time zone weirdness

backend/onyx/connectors/confluence/connector.py
evan-onyx: _construct_page_query is used to build the cql string, _build_page_retrieval_url gives the full url that we use to hit the API. I'll add cql to the name and leave a comment

Weves (approved):

PR #4594, opened 2025-04-23
fix: Fix migration issue in which display-model-names were not being appropriately set

raunakab:
## Description

Migration did not address the case in which:

- `model_names = []` and `display_model_names = []`
- `default_model_name = "some-model"` and `fast_default_model_name = "some-fast-model"`

I.e., the case in which the default and fast-default models were specified, but they *did not* exist inside of the `model_names` or `display_model_names` lists.

The reason why this bug was pushed was because of incorrect assumptions; I assumed that `model_names` and `display_model_names` would *always* contain the `default_model_name` and `fast_default_model_name`. I.e., I assumed that:

$`(default\_model\_name \in model\_names)`$
and
$`(default\_model\_name \in display\_model\_names)`$
and
$`(fast\_default\_model\_name \in model\_name)`$
and
$`(fast\_default\_model\_name \in display\_model\_names)`$

However, this was apparently not an invariant that was necessary to enforce (not sure why, though).

As such, this push towards data normalization resulted in a lossy migration in which the `display_model_names` data was lost.

This PR aims to fix that error. It appropriately appends `default_model_name` and `fast_default_model_name` to the `model_names` and `display_model_names` sets prior to iteratively inserting into the new database table.

Addresses: https://linear.app/danswer/issue/DAN-1874/fix-migration-script-to-properly-transform-display-model-namesto-the.

## How Has This Been Tested?

TODO!

### DIFF
--- a/backend/alembic/versions/7a70b7664e37_add_model_configuration_table.py
+++ b/backend/alembic/versions/7a70b7664e37_add_model_configuration_table.py
@@ -10,11 +10,56 @@ from alembic import op
 import sqlalchemy as sa
 from sqlalchemy.dialects import postgresql
 
+from onyx.llm.llm_provider_options import PROVIDER_TO_MODELS_MAP
+
 # revision identifiers, used by Alembic.
 revision = "7a70b7664e37"
 down_revision = "d961aca62eb3"
 branch_labels = None
 depends_on = None
+
+
+def get(provider_name: str) -> set[str] | None:
+    model_names: list[str] | None = PROVIDER_TO_MODELS_MAP.get(provider_name)
+    return set(model_names) if model_names else None
+
+
+def resolve(
+    provider_name: str,
+    model_names: list[str] | None,
+    display_model_names: list[str] | None,
+    default_model_name: str,
+    fast_default_model_name: str,
+) -> set[tuple[str, bool]]:
+    models = set(model_names) if model_names else None
+    display_models = set(display_model_names) if display_model_names else None
+
+    # if both are defined, we need to make sure that `model_names` is a superset of `display_model_names`
+    if models and display_models:
+        if not display_models.issubset(models):
+            models = display_models.union(models)
+
+    # if only the model-names are defined,
+    elif models and not display_models:
+        new = get(provider_name)
+        display_models = models.union(new) if new else set(models)
+
+    # if only the display-model-names are defined, then
+    elif not models and display_models:
+        new = get(provider_name)
+        models = display_models.union(new) if new else set(display_models)
+
+    else:
+        new = get(provider_name)
+        models = set(new) if new else set()
+        display_models = set(new) if new else set()
+
+    models.add(default_model_name)
+    models.add(fast_default_model_name)
+    display_models.add(default_model_name)
+    display_models.add(fast_default_model_name)
+
+    return set([(model, model in display_models) for model in models])
 
 
 def upgrade() -> None:
@@ -36,6 +81,7 @@ def upgrade() -> None:
     llm_provider_table = sa.sql.table(
         "llm_provider",
         sa.column("id", sa.Integer),
+        sa.column("provider", sa.Integer),
         sa.column("model_names", postgresql.ARRAY(sa.String)),
         sa.column("display_model_names", postgresql.ARRAY(sa.String)),
         sa.column("default_model_name", sa.String),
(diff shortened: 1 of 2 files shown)


backend/alembic/versions/7a70b7664e37_add_model_configuration_table.py
```diff
@@ -60,6 +62,13 @@ def upgrade() -> None:
         provider_id = llm_provider[0]
         model_names_set = set(llm_provider[1] or [])
         display_names_set = set(llm_provider[2] or [])
+        default_model_name = llm_provider[3]
```
Weves: I think actually we want to set the display model names to the "default for the provider" in the case that no display models exist.

raunakab: Updated in the latest commit.

backend/onyx/llm/llm_provider_options.py
```diff
@@ -128,7 +128,7 @@ class WellKnownLLMProviderDescriptor(BaseModel):
 ]
 
 
-_PROVIDER_TO_MODELS_MAP = {
```
Weves, line 131: can we just use `fetch_models_for_provider` within the migration rather than making this non-private?

backend/alembic/versions/7a70b7664e37_add_model_configuration_table.py
```diff
@@ -58,14 +106,21 @@ def upgrade() -> None:
...
     for llm_provider in llm_providers:
         provider_id = llm_provider[0]
-        model_names_set = set(llm_provider[1] or [])
-        display_names_set = set(llm_provider[2] or [])
-
-        # Insert all models from model_names
-        for model_name in model_names_set:
-            # If model is in display_model_names, set is_visible to True
-            is_visible = model_name in display_names_set
+        provider_name = llm_provider[1]
```
Weves, line 148: this seems broken, right? We'd need to also add the new values to the select on line 100?

backend/alembic/versions/7a70b7664e37_add_model_configuration_table.py
```diff
@@ -10,13 +10,58 @@
...
+from onyx.llm.llm_provider_options import PROVIDER_TO_MODELS_MAP
+
 # revision identifiers, used by Alembic.
 revision = "7a70b7664e37"
 down_revision = "d961aca62eb3"
 branch_labels = None
 depends_on = None
 
 
+def get(provider_name: str) -> set[str] | None:
```
Weves: these should probably be private methods — they aren't meant to be used outside of this file

Weves: also prefer more descriptive names (e.g. `get_model_set_for_provider`).

backend/alembic/versions/7a70b7664e37_add_model_configuration_table.py
```diff
@@ -10,13 +10,58 @@
...
+    display_models = set(display_model_names) if display_model_names else None
+
+    # if both are defined, we need to make sure that `model_names` is a superset of `display_model_names`
+    if models and display_models:
+        if not display_models.issubset(models):
+            models = display_models.union(models)
+
+    # if only the model-names are defined,
+    elif models and not display_models:
+        new = get(provider_name)
```
Weves: nit: also prefer more descriptive name instead of new

backend/onyx/llm/llm_provider_options.py
raunakab: `fetch_models_for_provider` returns an empty list if the `provider_name` doesn't exist, which is categorically different than `None`.

Maybe I could move my helper function in the migration file to this file?

backend/alembic/versions/7a70b7664e37_add_model_configuration_table.py
raunakab: Yup, they should be private! Will update the names to be private and more descriptive.

backend/alembic/versions/7a70b7664e37_add_model_configuration_table.py
raunakab: Will update in the coming commit.

raunakab: Updated.

backend/alembic/versions/7a70b7664e37_add_model_configuration_table.py
raunakab: Updated.

backend/onyx/llm/llm_provider_options.py
raunakab: Moved the helper to `llm_provider_options.py` instead. Reverted back to `_PROVIDER_TO_MODELS_MAP` (private instead of public).

backend/alembic/versions/7a70b7664e37_add_model_configuration_table.py
raunakab: Yup, sorry forgot to update that line.

raunakab: Updated!d

backend/alembic/versions/7a70b7664e37_add_model_configuration_table.py
```diff
@@ -10,13 +10,84 @@
...
+    #
+    # This will also preserve the invariant that `display_models` is a subset of `models`.
+    else:
+        default_models_for_provider = fetch_model_names_for_provider_as_set(
+            provider_name
+        )
+        models = (
+            set(default_models_for_provider) if default_models_for_provider else set()
+        )
+        display_models = (
```
Weves: is this correct? Won't this make everything display when we really just want a sub-set of them?

raunakab: Hmm, you're right. I can update this to only set `is_visible` for the models which are visible in `llm_provider_options.py`.

I guess I should also fix the above case (i.e., `models and not display_models`).

raunakab: Updated!

Weves (approved):

PR #4595, opened 2025-04-23
ensure drive id set in checkpoint

evan-onyx:
## Description

https://linear.app/danswer/issue/DAN-1873/drive-id-not-set-in-checkpoint

Fix the remaining "id not set in checkpoint" issues; the last fix only addressed "folder not set in checkpoint" while this addresses "drive id not set in checkpoint". Also changes asserts -> conditional error throwing to align with our new code practices.

## How Has This Been Tested?

n/a, this is a tiny one

### DIFF
--- a/backend/onyx/connectors/google_drive/connector.py
+++ b/backend/onyx/connectors/google_drive/connector.py
@@ -476,6 +476,8 @@ class GoogleDriveConnector(SlimConnector, CheckpointedConnector[GoogleDriveCheck
                 logger.info(
                     f"Getting files in shared drive '{drive_id}' as '{user_email}'"
                 )
+                curr_stage.completed_until = 0
+                curr_stage.completed_until_parent_id = drive_id
                 yield from _yield_from_drive(drive_id, start)
             curr_stage.stage = DriveRetrievalStage.FOLDER_FILES
             resuming = False  # we are starting the next stage for the first time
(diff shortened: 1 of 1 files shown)


backend/onyx/connectors/google_drive/connector.py
```diff
@@ -613,10 +618,11 @@ def _manage_service_account_retrieval(
...
                 f"Some folders/drives were not retrieved. IDs: {remaining_folders}"
             )
-        assert all(
-            checkpoint.completion_map[user_email].stage == DriveRetrievalStage.DONE
+        if any(
+            checkpoint.completion_map[user_email].stage != DriveRetrievalStage.DONE
             for user_email in all_org_emails
-        ), "some users did not complete retrieval"
+        ):
+            raise ValueError("some users did not complete retrieval")
```
Weves: nit: prefer RuntimeError here. Feels like an overall invalid state rather than a bad passed in value.

backend/onyx/connectors/google_drive/connector.py
```diff
@@ -646,12 +652,10 @@ def _determine_retrieval_ids(
                 checkpoint.folder_ids_to_retrieve = list(folder_ids_to_retrieve)
             checkpoint.completion_stage = next_stage
         else:
-            assert (
-                checkpoint.drive_ids_to_retrieve is not None
-            ), "drive ids to retrieve not set"
-            assert (
-                checkpoint.folder_ids_to_retrieve is not None
-            ), "folder ids to retrieve not set"
+            if checkpoint.drive_ids_to_retrieve is None:
```
Weves, line 655: same here, prefer runtime

Weves (approved):

PR #4596, opened 2025-04-24
Fix tool choice

Weves:
## Description

Fixes https://linear.app/danswer/issue/DAN-1875/fix-tool-choice

## How Has This Been Tested?

Tested an assistant with a tool before (would always search). Now it doesn't. 

### DIFF
--- a/web/src/app/chat/ChatPage.tsx
+++ b/web/src/app/chat/ChatPage.tsx
@@ -45,6 +45,7 @@ import {
   processRawChatHistory,
   removeMessage,
   sendMessage,
+  SendMessageParams,
   setMessageAsLatest,
   updateLlmOverrideForChatSession,
   updateParentChildren,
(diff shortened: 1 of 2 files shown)


web/src/app/chat/ChatPage.tsx
```diff
@@ -1080,7 +1081,7 @@ export function ChatPage({
 
   async function updateCurrentMessageFIFO(
     stack: CurrentMessageFIFO,
-    params: any
```
Weves, line 1083: remove the evil `any` !

web/src/app/chat/lib.tsx
```diff
@@ -168,7 +191,6 @@ export async function* sendMessage({
   userFolderIds,
   parentMessageId,
   chatSessionId,
-  promptId,
```
Weves, line 171: wasn't actually used

PR #4600, opened 2025-04-24
fix: Create migration to re-introduce `display_model_names`

raunakab:
## Description

A previous migration erroneously converted the `llm_providers.model_names` to a new `model_configurations` database table, but did not properly account for edge cases in which model_names or display_model_names could be `null`.

As a result, certain configurations describing which models were visible or not were lost (*for native LLM Providers only*).

This PR addresses that issue by looking through the existing state, and if the `model_configurations` for a given `llm_provider` are *all* `false`, or there are *no model-configurations*, then the default `model_names` and default visibilities are used.

Addresses: https://linear.app/danswer/issue/DAN-1874/fix-migration-script-to-properly-transform-display-model-namesto-the.

## How Has This Been Tested?

TODO!

### DIFF
--- a/backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
+++ b/backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
@@ -1,0 +1,82 @@
+"""Fix invalid model-configurations state
+
+Revision ID: 47a07e1a38f1
+Revises: 7a70b7664e37
+Create Date: 2025-04-23 15:39:43.159504
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+from onyx.llm.llm_provider_options import fetch_model_names_for_provider_as_set
+
+
+# revision identifiers, used by Alembic.
+revision = "47a07e1a38f1"
+down_revision = "7a70b7664e37"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    llm_provider_table = sa.sql.table(
+        "llm_provider",
+        sa.column("id", sa.Integer),
+        sa.column("provider", sa.Integer),
+        sa.column("model_names", postgresql.ARRAY(sa.String)),
+        sa.column("display_model_names", postgresql.ARRAY(sa.String)),
+        sa.column("default_model_name", sa.String),
+        sa.column("fast_default_model_name", sa.String),
+    )
+    model_configuration_table = sa.sql.table(
+        "model_configuration",
+        sa.column("id", sa.Integer),
+        sa.column("llm_provider_id", sa.Integer),
+        sa.column("name", sa.String),
+        sa.column("is_visible", sa.Boolean),
+        sa.column("max_input_tokens", sa.Integer),
+    )
+
+    connection = op.get_bind()
+
+    llm_providers = connection.execute(
+        sa.select(
+            llm_provider_table.c.id,
+            llm_provider_table.c.provider,
+        )
+    ).fetchall()
+
+    for llm_provider in llm_providers:
+        llm_provider_id, provider_name = llm_provider
+
+        models = fetch_model_names_for_provider_as_set(provider_name)
+
+        # if `fetch_model_names_for_provider_as_set` returns `None`, then
+        # that means that `provider_name` is not a well-known llm provider.
+        if not models:
+            continue
+
+        _model_configurations = connection.execute(
+            sa.select(
+                model_configuration_table.c.id,
+                model_configuration_table.c.llm_provider_id,
+                model_configuration_table.c.name,
+                model_configuration_table.c.is_visible,
+            ).where(model_configuration_table.c.llm_provider_id == llm_provider_id)
+        ).fetchall()
+
+        all_are_private = True
+
+        # If `not all_are_private`, then there must exist at least one
+        # model which is public. Therefore, this is a valid state;
+        # don't touch this state and move onto the next one.
+        if not all_are_private:
+            continue
+
+        ...
+
+
+def downgrade() -> None:
+    pass
--- a/backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
+++ b/backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
@@ -10,7 +10,17 @@ from alembic import op
 import sqlalchemy as sa
 from sqlalchemy.dialects import postgresql
 
-from onyx.llm.llm_provider_options import fetch_model_names_for_provider_as_set
+from onyx.llm.llm_provider_options import (
+    ANTHROPIC_PROVIDER_NAME,
+    ANTHROPIC_VISIBLE_MODEL_NAMES,
+    BEDROCK_PROVIDER_NAME,
+    OPEN_AI_VISIBLE_MODEL_NAMES,
+    OPENAI_PROVIDER_NAME,
+    VERTEXAI_DEFAULT_FAST_MODEL,
+    VERTEXAI_DEFAULT_MODEL,
+    VERTEXAI_PROVIDER_NAME,
+    fetch_model_names_for_provider_as_set,
+)
 
 
 # revision identifiers, used by Alembic.
@@ -18,6 +28,14 @@ revision = "47a07e1a38f1"
 down_revision = "7a70b7664e37"
 branch_labels = None
 depends_on = None
+
+
+_PROVIDER_TO_VISIBLE_MODELS_MAP = {
+    OPENAI_PROVIDER_NAME: set(OPEN_AI_VISIBLE_MODEL_NAMES),
+    BEDROCK_PROVIDER_NAME: set(),  # TODO!
+    ANTHROPIC_PROVIDER_NAME: set(ANTHROPIC_VISIBLE_MODEL_NAMES),
+    VERTEXAI_PROVIDER_NAME: set([VERTEXAI_DEFAULT_MODEL, VERTEXAI_DEFAULT_FAST_MODEL]),
+}
 
 
 def upgrade() -> None:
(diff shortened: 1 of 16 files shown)


backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
```diff
@@ -0,0 +1,137 @@
...
+# revision identifiers, used by Alembic.
+revision = "47a07e1a38f1"
+down_revision = "7a70b7664e37"
+branch_labels = None
+depends_on = None
+
+
+_PROVIDER_TO_VISIBLE_MODELS_MAP = {
+    OPENAI_PROVIDER_NAME: set(OPEN_AI_VISIBLE_MODEL_NAMES),
+    BEDROCK_PROVIDER_NAME: set(),  # TODO!
```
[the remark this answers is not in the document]
raunakab: @Weves Couldn't find the `BEDROCK_VISIBLE_MODEL_NAMES`. Do you have a suggestion for what they should be?

backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
```diff
@@ -0,0 +1,137 @@
...
+    ANTHROPIC_PROVIDER_NAME: set(ANTHROPIC_VISIBLE_MODEL_NAMES),
+    VERTEXAI_PROVIDER_NAME: set([VERTEXAI_DEFAULT_MODEL, VERTEXAI_DEFAULT_FAST_MODEL]),
+}
+
+
+def upgrade() -> None:
+    llm_provider_table = sa.sql.table(
+        "llm_provider",
+        sa.column("id", sa.Integer),
+        sa.column("provider", sa.Integer),
```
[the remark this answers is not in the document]
raunakab: Updated.

backend/alembic/versions/7a70b7664e37_add_model_configuration_table.py
```diff
@@ -51,21 +125,31 @@ def upgrade() -> None:
...
-
-        # Insert all models from model_names
-        for model_name in model_names_set:
-            # If model is in display_model_names, set is_visible to True
-            is_visible = model_name in display_names_set
+        provider_name = llm_provider[1]
+        model_names = llm_provider[2]
+        display_model_names = llm_provider[3]
+        default_model_name = llm_provider[4]
+        fast_default_model_name = llm_provider[5]
```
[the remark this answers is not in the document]
raunakab: Not relevant to this PR anymore.

backend/onyx/llm/llm_provider_options.py
```diff
@@ -228,3 +228,8 @@ def fetch_available_well_known_llms() -> list[WellKnownLLMProviderDescriptor]:
 
 def fetch_models_for_provider(provider_name: str) -> list[str]:
     return _PROVIDER_TO_MODELS_MAP.get(provider_name, [])
+
+
+def fetch_model_names_for_provider_as_set(provider_name: str) -> set[str] | None:
+    model_names: list[str] | None = _PROVIDER_TO_MODELS_MAP.get(provider_name)
+    return set(model_names) if model_names else None
```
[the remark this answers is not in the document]
raunakab: This function returns:
- `None` if `model_names` is `None`
- `None` if `model_names` is the empty list
- A non-empty set if `model_names` is a non-empty list

backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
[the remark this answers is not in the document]
Weves: @raunakab just the default ones is probably fine

backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
```diff
@@ -0,0 +1,135 @@
...
+            connection.execute(
+                sa.select(
+                    model_configuration_table.c.id,
+                    model_configuration_table.c.llm_provider_id,
+                    model_configuration_table.c.name,
+                ).where(model_configuration_table.c.llm_provider_id == llm_provider_id)
+            ).fetchall()
+        )
+
+        display_model_names: set[str] = _PROVIDER_TO_VISIBLE_MODELS_MAP[provider_name]
```
Weves: we should probably use the utility from the other PR, right?

backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
```diff
@@ -0,0 +1,135 @@
...
+
+        if model_configurations:
+            at_least_one_is_public = any(
+                [model_configuration[3] for model_configuration in model_configurations]
```
Weves: this is wrong, right? It looks like `model_configuration[3]` is the name?

backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
```diff
@@ -0,0 +1,135 @@
...
+            )
+
+            # If there is at least one model which is public, this is a valid state.
+            # Therefore, don't touch it and move on to the next one.
+            if at_least_one_is_public:
+                continue
+
+            existing_visible_model_names: set[str] = set(
```
Weves, line 110: same here

backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
```diff
@@ -0,0 +1,135 @@
...
+
+        model_configurations = list(
+            connection.execute(
+                sa.select(
+                    model_configuration_table.c.id,
+                    model_configuration_table.c.llm_provider_id,
+                    model_configuration_table.c.name,
+                ).where(model_configuration_table.c.llm_provider_id == llm_provider_id)
+            ).fetchall()
+        )
```
Weves: can we convert these into pydantic models (or similar) to make the rest of the code less error prone/easier to read?

backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
```diff
@@ -0,0 +1,135 @@
...
+                ]
+            )
+
+            for model_name in display_model_names.difference(
+                existing_visible_model_names
+            ):
+                if not model_name:
+                    continue
+
+                connection.execute(
```
Weves, line 131: does this handle updates? E.g. that model exists but isn't visible?

backend/onyx/llm/llm_provider_options.py
```diff
@@ -228,3 +228,8 @@ def fetch_available_well_known_llms() -> list[WellKnownLLMProviderDescriptor]:
 
 def fetch_models_for_provider(provider_name: str) -> list[str]:
     return _PROVIDER_TO_MODELS_MAP.get(provider_name, [])
+
+
+def fetch_model_names_for_provider_as_set(provider_name: str) -> set[str] | None:
+    model_names: list[str] | None = _PROVIDER_TO_MODELS_MAP.get(provider_name)
```
Weves: we should probably use the `fetch_models_for_provider` function from above

backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
raunakab: Yup, sorry that other one was being iterated on in parallel to this one, so forgot to pull those changes in.

raunakab: Updated.

backend/onyx/llm/llm_provider_options.py
raunakab: Updated!

backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
raunakab: Good catch! I updated it to be a `on-conflict-do-nothing` statement.

backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
raunakab: Updated to using `pydantic` instead!

backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
raunakab: Updated to using `pydantic` instead!

backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
raunakab: Updated!

backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
```diff
@@ -0,0 +1,148 @@
...
+                    model_configuration_table.c.llm_provider_id,
+                    model_configuration_table.c.name,
+                    model_configuration_table.c.is_visible,
+                    model_configuration_table.c.max_input_tokens,
+                ).where(model_configuration_table.c.llm_provider_id == llm_provider_id)
+            ).fetchall()
+        ]
+
+        if model_configurations:
+            at_least_one_is_public = any(
```
Weves: this should probably be `at_least_one_is_visible` as discussed?

backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
```diff
@@ -0,0 +1,148 @@
...
+        )
+
+        # if `fetch_model_names_for_provider_as_set` returns `None`, then
+        # that means that `provider_name` is not a well-known llm provider.
+        if not default_models:
+            continue
+
+        # If `default_models` is non-None, `display_model_names` must be non-None too.
+        if not display_models:
+            raise RuntimeError
```
Weves: nit: generally have descriptive messages with exceptions

Weves (commented):
Tested it locally and I seemed to only get gpt-4o and gpt-4o-mini enabled

raunakab:
> Tested it locally and I seemed to only get gpt-4o and gpt-4o-mini enabled

@Weves was it maybe because you had those two models inly enabled before?

backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
raunakab: Yes, sorry forgot to update.

backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
raunakab: Sure, will update.

raunakab: Updated.

backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
raunakab: Updated.

backend/alembic/versions/47a07e1a38f1_fix_invalid_model_configurations_state.py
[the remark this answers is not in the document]
raunakab: Updated.

Weves (approved):

PR #4601, opened 2025-04-24
perf: Implement checkpointing for Teams Connector.

raunakab:
## Description

This PR implements checkpointing for the Teams connector.

Addresses: https://linear.app/danswer/issue/DAN-1835/teams-checkpointing-connector.

## How Has This Been Tested?

Wrote some automated tests inside of `backend/tests/daily/connectors/teams`.

### DIFF
--- a/backend/onyx/connectors/slack/connector.py
+++ b/backend/onyx/connectors/slack/connector.py
@@ -868,11 +868,19 @@ if __name__ == "__main__":
         channels=[slack_channel] if slack_channel else None,
     )
 
+    slack_bot_token = os.environ.get(
+        "SLACK_BOT_TOKEN", os.environ.get("DANSWER_BOT_SLACK_BOT_TOKEN")
+    )
+    if not slack_bot_token:
+        raise RuntimeError(
+            "`SLACK_BOT_TOKEN` or `DANSWER_BOT_SLACK_BOT_TOKEN` not found"
+        )
+
     provider = OnyxStaticCredentialsProvider(
         tenant_id=get_current_tenant_id(),
         connector_name="slack",
         credential_json={
-            "slack_bot_token": os.environ["SLACK_BOT_TOKEN"],
+            "slack_bot_token": slack_bot_token,
         },
     )
     connector.set_credentials_provider(provider)
--- a/backend/onyx/connectors/slack/connector.py
+++ b/backend/onyx/connectors/slack/connector.py
@@ -868,19 +868,11 @@ if __name__ == "__main__":
         channels=[slack_channel] if slack_channel else None,
     )
 
-    slack_bot_token = os.environ.get(
-        "SLACK_BOT_TOKEN", os.environ.get("DANSWER_BOT_SLACK_BOT_TOKEN")
-    )
-    if not slack_bot_token:
-        raise RuntimeError(
-            "`SLACK_BOT_TOKEN` or `DANSWER_BOT_SLACK_BOT_TOKEN` not found"
-        )
-
     provider = OnyxStaticCredentialsProvider(
         tenant_id=get_current_tenant_id(),
         connector_name="slack",
         credential_json={
-            "slack_bot_token": slack_bot_token,
+            "slack_bot_token": os.environ["SLACK_BOT_TOKEN"],
         },
     )
     connector.set_credentials_provider(provider)
--- a/backend/onyx/connectors/teams/connector.py
+++ b/backend/onyx/connectors/teams/connector.py
@@ -22,6 +22,7 @@ from onyx.connectors.interfaces import LoadConnector
 from onyx.connectors.interfaces import PollConnector
 from onyx.connectors.interfaces import SecondsSinceUnixEpoch
 from onyx.connectors.models import BasicExpertInfo
+from onyx.connectors.models import ConnectorCheckpoint
 from onyx.connectors.models import ConnectorMissingCredentialError
 from onyx.connectors.models import Document
 from onyx.connectors.models import TextSection
@@ -187,6 +188,9 @@ def _convert_thread_to_document(
         metadata={},
     )
     return doc
+
+
+class TeamsCheckpoint(ConnectorCheckpoint): ...
 
 
 class TeamsConnector(LoadConnector, PollConnector):
--- a/backend/onyx/connectors/teams/connector.py
+++ b/backend/onyx/connectors/teams/connector.py
@@ -229,7 +229,10 @@ def _get_created_datetime(chat_message: ChatMessage) -> datetime:
 
 def _extract_channel_members(channel: Channel) -> list[BasicExpertInfo]:
     channel_members_list: list[BasicExpertInfo] = []
-    members = channel.members.get_all(page_loaded=lambda _: None).execute_query_retry()
+    members = channel.members.get_all(
+        # explicitly needed because of incorrect type definitions provided by the `office365` library
+        page_loaded=lambda _: None
+    ).execute_query_retry()
     for member in members:
         channel_members_list.append(BasicExpertInfo(display_name=member.display_name))
     return channel_members_list
@@ -354,7 +357,10 @@ def _collect_all_team_ids(
         if filter:
             query = graph_client.teams.get().filter(filter)
         else:
-            query = graph_client.teams.get_all(page_loaded=lambda _: None)
+            query = graph_client.teams.get_all(
+                # explicitly needed because of incorrect type definitions provided by the `office365` library
+                page_loaded=lambda _: None
+            )
 
         if next_url:
             url = next_url
@@ -443,7 +449,10 @@ def _collect_all_channels_for_team_id(
     next_url = None
 
     while True:
-        query = team.channels.get_all(page_loaded=lambda _: None)
+        query = team.channels.get_all(
+            # explicitly needed because of incorrect type definitions provided by the `office365` library
+            page_loaded=lambda _: None
+        )
         if next_url:
             url = next_url
             query = query.before_execute(
@@ -475,6 +484,7 @@ def _collect_document_for_channel_id(
     channel = channel_collection[0]
 
     message_collection = channel.messages.get_all(
+        # explicitly needed because of incorrect type definitions provided by the `office365` library
         page_loaded=lambda _: None
     ).execute_query()
 
--- a/backend/onyx/connectors/teams/connector.py
+++ b/backend/onyx/connectors/teams/connector.py
@@ -17,9 +17,9 @@ from onyx.connectors.exceptions import ConnectorValidationError
 from onyx.connectors.exceptions import CredentialExpiredError
 from onyx.connectors.exceptions import InsufficientPermissionsError
 from onyx.connectors.exceptions import UnexpectedValidationError
+from onyx.connectors.interfaces import CheckpointedConnector
+from onyx.connectors.interfaces import CheckpointOutput
 from onyx.connectors.interfaces import GenerateDocumentsOutput
-from onyx.connectors.interfaces import LoadConnector
-from onyx.connectors.interfaces import PollConnector
 from onyx.connectors.interfaces import SecondsSinceUnixEpoch
 from onyx.connectors.models import BasicExpertInfo
 from onyx.connectors.models import ConnectorCheckpoint
@@ -190,10 +190,13 @@ def _convert_thread_to_document(
     return doc
 
 
-class TeamsCheckpoint(ConnectorCheckpoint): ...
+class TeamsCheckpoint(ConnectorCheckpoint):
+    channel_ids: list[str] | None
 
 
-class TeamsConnector(LoadConnector, PollConnector):
+class TeamsConnector(
+    CheckpointedConnector[TeamsCheckpoint],
+):
     MAX_CHANNELS_TO_LOG = 50
 
     def __init__(
@@ -207,6 +210,8 @@ class TeamsConnector(LoadConnector, PollConnector):
         self.graph_client: GraphClient | None = None
         self.requested_team_list: list[str] = teams
         self.msal_app: msal.ConfidentialClientApplication | None = None
+
+    # impls for BaseConnector
 
     def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None:
         teams_client_id = credentials["teams_client_id"]
@@ -230,10 +235,85 @@ class TeamsConnector(LoadConnector, PollConnector):
             token = self.msal_app.acquire_token_for_client(
                 scopes=["https://graph.microsoft.com/.default"]
             )
+
+            if not isinstance(token, dict):
+                raise RuntimeError("...")
+                ...
+
             return token
 
         self.graph_client = GraphClient(_acquire_token_func)
         return None
+
+    def validate_connector_settings(self) -> None:
+        if self.graph_client is None:
+            raise ConnectorMissingCredentialError("Teams credentials not loaded.")
+
+        try:
+            # Minimal call to confirm we can retrieve Teams
+            # make sure it doesn't take forever, since this is a syncronous call
+            found_teams = run_with_timeout(10, self._get_all_teams)
+
+        except ClientRequestException as e:
+            status_code = e.response.status_code
+            if status_code == 401:
+                raise CredentialExpiredError(
+                    "Invalid or expired Microsoft Teams credentials (401 Unauthorized)."
+                )
+            elif status_code == 403:
+                raise InsufficientPermissionsError(
+                    "Your app lacks sufficient permissions to read Teams (403 Forbidden)."
+                )
+            raise UnexpectedValidationError(f"Unexpected error retrieving teams: {e}")
+
+        except Exception as e:
+            error_str = str(e).lower()
+            if (
+                "unauthorized" in error_str
+                or "401" in error_str
+                or "invalid_grant" in error_str
+            ):
+                raise CredentialExpiredError(
+                    "Invalid or expired Microsoft Teams credentials."
+                )
+            elif "forbidden" in error_str or "403" in error_str:
+                raise InsufficientPermissionsError(
+                    "App lacks required permissions to read from Microsoft Teams."
+                )
+            raise ConnectorValidationError(
+                f"Unexpected error during Teams validation: {e}"
+            )
+
+        if not found_teams:
+            raise ConnectorValidationError(
+                "No Teams found for the given credentials. "
+                "Either there are no Teams in this tenant, or your app does not have permission to view them."
+            )
+
+    # impls for CheckpointedConnector
+
+    def build_dummy_checkpoint(self) -> TeamsCheckpoint:
+        return TeamsCheckpoint(
+            channel_ids=None,
+            has_more=True,
+        )
+
+    def validate_checkpoint_json(self, checkpoint_json: str) -> TeamsCheckpoint:
+        return TeamsCheckpoint.model_validate_json(checkpoint_json)
+
+    def load_from_checkpoint(
+        self,
+        start: SecondsSinceUnixEpoch,
+        end: SecondsSinceUnixEpoch,
+        checkpoint: TeamsCheckpoint,
+    ) -> CheckpointOutput[TeamsCheckpoint]:
+        if self.graph_client is None:
+            ...
+
+        # TODO!
+        ...
+
+    # private helpers
 
     def _get_all_teams(self) -> list[Team]:
         if self.graph_client is None:
@@ -341,71 +421,42 @@ class TeamsConnector(LoadConnector, PollConnector):
                 doc_batch = []
         yield doc_batch
 
-    def load_from_state(self) -> GenerateDocumentsOutput:
-        return self._fetch_from_teams()
-
-    def poll_source(
-        self, start: SecondsSinceUnixEpoch, end: SecondsSinceUnixEpoch
-    ) -> GenerateDocumentsOutput:
-        start_datetime = datetime.fromtimestamp(start, timezone.utc)
-        end_datetime = datetime.fromtimestamp(end, timezone.utc)
-        return self._fetch_from_teams(start=start_datetime, end=end_datetime)
-
-    def validate_connector_settings(self) -> None:
-        if self.graph_client is None:
-            raise ConnectorMissingCredentialError("Teams credentials not loaded.")
-
-        try:
-            # Minimal call to confirm we can retrieve Teams
-            # make sure it doesn't take forever, since this is a syncronous call
-            found_teams = run_with_timeout(10, self._get_all_teams)
-
-        except ClientRequestException as e:
-            status_code = e.response.status_code
-            if status_code == 401:
-                raise CredentialExpiredError(
-                    "Invalid or expired Microsoft Teams credentials (401 Unauthorized)."
-                )
-            elif status_code == 403:
-                raise InsufficientPermissionsError(
-                    "Your app lacks sufficient permissions to read Teams (403 Forbidden)."
-                )
-            raise UnexpectedValidationError(f"Unexpected error retrieving teams: {e}")
-
-        except Exception as e:
-            error_str = str(e).lower()
-            if (
-                "unauthorized" in error_str
-                or "401" in error_str
-                or "invalid_grant" in error_str
-            ):
-                raise CredentialExpiredError(
-                    "Invalid or expired Microsoft Teams credentials."
-                )
-            elif "forbidden" in error_str or "403" in error_str:
-                raise InsufficientPermissionsError(
-                    "App lacks required permissions to read from Microsoft Teams."
-                )
-            raise ConnectorValidationError(
-                f"Unexpected error during Teams validation: {e}"
-            )
-
-        if not found_teams:
-            raise ConnectorValidationError(
-                "No Teams found for the given credentials. "
-                "Either there are no Teams in this tenant, or your app does not have permission to view them."
-            )
-
 
 if __name__ == "__main__":
-    connector = TeamsConnector(teams=os.environ["TEAMS"].split(","))
+    import time
+
+    app_id = os.environ["TEAMS_APPLICATION_ID"]
+    dir_id = os.environ["TEAMS_DIRECTORY_ID"]
+    secret = os.environ["TEAMS_SECRET"]
+    teams = os.environ.get("TEAMS", "").split(",")
+
+    connector = TeamsConnector(teams=teams)
 
     connector.load_credentials(
         {
-            "teams_client_id": os.environ["TEAMS_CLIENT_ID"],
-            "teams_client_secret": os.environ["TEAMS_CLIENT_SECRET"],
-            "teams_directory_id": os.environ["TEAMS_CLIENT_DIRECTORY_ID"],
+            "teams_client_id": app_id,
+            "teams_directory_id": dir_id,
+            "teams_client_secret": secret,
         }
     )
-    document_batches = connector.load_from_state()
-    print(next(document_batches))
+
+    start = 0.0
+    end = time.time()
+    checkpoint = connector.build_dummy_checkpoint()
+
+    gen = connector.load_from_checkpoint(
+        start=start,
+        end=end,
+        checkpoint=checkpoint,
+    )
+
+    try:
+        for doc_or_failure in gen:
+            if isinstance(doc_or_failure, Document):
+                print(doc_or_failure)
+            elif isinstance(doc_or_failure, Document):
+                print(doc_or_failure)
+    except StopIteration as e:
+        checkpoint = e.value
+
+    print(checkpoint)
--- a/backend/onyx/connectors/teams/connector.py
+++ b/backend/onyx/connectors/teams/connector.py
@@ -187,7 +187,7 @@ class TeamsConnector(
                 has_more=False,
             )
 
-        todo_team_id = todos[-1]
+        todo_team_id = todos.pop()
         team = _get_team_by_id(
             graph_client=self.graph_client,
             team_id=todo_team_id,
@@ -196,7 +196,6 @@ class TeamsConnector(
             graph_client=self.graph_client,
             team=team,
         )
-        todos.pop()
 
         with ThreadPoolExecutor(max_workers=self.max_workers) as executor:
             futures: list[Future[Document | None]] = []
--- a/backend/onyx/connectors/teams/connector.py
+++ b/backend/onyx/connectors/teams/connector.py
@@ -425,7 +425,7 @@ def _get_team_by_id(
     )
 
     if not team_collection:
-        raise RuntimeError(f"No team with {team_id=} was found")
+        raise ValueError(f"No team with {team_id=} was found")
     elif team_collection.has_next:
         # shouldn't happen, but catching it regardless
         raise RuntimeError(f"Multiple teams with {team_id=} were found")
--- a/backend/onyx/connectors/teams/connector.py
+++ b/backend/onyx/connectors/teams/connector.py
@@ -32,7 +32,6 @@ from onyx.connectors.models import Document
 from onyx.connectors.models import TextSection
 from onyx.file_processing.html_utils import parse_html_page_basic
 from onyx.utils.logger import setup_logger
-from tests.daily.connectors.utils import load_everything_from_checkpoint_connector
 
 logger = setup_logger()
 
@@ -436,6 +435,7 @@ def _collect_document_for_channel_id(
 
 if __name__ == "__main__":
     import time
+    from tests.daily.connectors.utils import load_everything_from_checkpoint_connector
 
     app_id = os.environ["TEAMS_APPLICATION_ID"]
     dir_id = os.environ["TEAMS_DIRECTORY_ID"]
--- a/backend/onyx/connectors/teams/connector.py
+++ b/backend/onyx/connectors/teams/connector.py
@@ -371,7 +371,7 @@ def _collect_all_channels_for_team_id(
         # shouldn't happen, but catching it regardless
         raise RuntimeError(f"Multiple teams with {team_id=} were found")
 
-    [team] = team_collection
+    team = team_collection[0]
 
     todo_channels: list[TodoTeam | TodoChannel] = []
     next_url = None
@@ -407,12 +407,12 @@ def _collect_document_for_channel_id(
     team_collection = (
         graph_client.teams.get().filter(f"id eq '{team_id}'").top(1).execute_query()
     )
-    [team] = team_collection
+    team = team_collection[0]
 
     channel_collection = (
         team.channels.get().filter(f"id eq '{channel_id}'").execute_query()
     )
-    [channel] = channel_collection
+    channel = channel_collection[0]
 
     message_collection = channel.messages.get_all(
         page_loaded=lambda _: None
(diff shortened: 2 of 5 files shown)


backend/onyx/connectors/slack/connector.py
```diff
@@ -868,11 +868,19 @@ def validate_checkpoint_json(self, checkpoint_json: str) -> SlackCheckpoint:
         channels=[slack_channel] if slack_channel else None,
     )
 
+    slack_bot_token = os.environ.get(
+        "SLACK_BOT_TOKEN", os.environ.get("DANSWER_BOT_SLACK_BOT_TOKEN")
```
evan-onyx: Since we're changing this anyways, might as well do ONYX_SLACK_BOT_TOKEN

backend/onyx/connectors/teams/connector.py
```diff
@@ -1,97 +1,241 @@
...
+                            )
+                        )
 
-        # start a list containing the base message and its replies
-        thread: list[ChatMessage] = [base_message]
-        thread.extend(replies)
+                    for future in as_completed(futures):
+                        doc = future.result()
+                        if doc:
+                            print(doc)
```
evan-onyx: yoink that print

backend/onyx/connectors/teams/connector.py
```diff
@@ -1,97 +1,241 @@
...
+from typing import cast
 
 import msal  # type: ignore
 from office365.graph_client import GraphClient  # type: ignore
 from office365.runtime.client_request_exception import ClientRequestException  # type: ignore
+from office365.runtime.http.request_options import RequestOptions  # type: ignore[import-untyped]
 from office365.teams.channels.channel import Channel  # type: ignore
 from office365.teams.chats.messages.message import ChatMessage  # type: ignore
-from office365.teams.team import Team  # type: ignore
+from pydantic import BaseModel  # type: ignore
```
evan-onyx: We really shouldn't have to type: ignore this one

backend/onyx/connectors/teams/connector.py
```diff
@@ -1,97 +1,241 @@
...
+        if checkpoint.todos is None:
+            root_todos = _collect_all_teams(self.graph_client)
+            return TeamsCheckpoint(
+                todos=root_todos,
+                has_more=True if root_todos else False,
+            )
+
+        todos = checkpoint.todos
+
+        if not todos:
```
evan-onyx: shouldn't need this if, it's handled by the while loop

backend/onyx/connectors/teams/connector.py
```diff
@@ -1,97 +1,241 @@
...
+        if self.graph_client is None:
+            raise ConnectorMissingCredentialError("Teams")
+
+        checkpoint = cast(TeamsCheckpoint, copy.deepcopy(checkpoint))
+
+        if checkpoint.todos is None:
+            root_todos = _collect_all_teams(self.graph_client)
+            return TeamsCheckpoint(
+                todos=root_todos,
+                has_more=True if root_todos else False,
```
evan-onyx: `has_more = bool(root_todos)`

backend/onyx/connectors/teams/connector.py
```diff
@@ -189,219 +333,130 @@ def _convert_thread_to_document(
...
+    next_url = None
 
-        authority_url = f"https://login.microsoftonline.com/{teams_directory_id}"
-        self.msal_app = msal.ConfidentialClientApplication(
-            authority=authority_url,
-            client_id=teams_client_id,
-            client_credential=teams_client_secret,
-        )
+    while True:
+        query = graph_client.teams.get_all(page_loaded=lambda _: None)
```
evan-onyx: there were some useful comments in the old version I think would be good to keep around! Such as the fact that get_all() handles pagination for you under the hood, which I think is fairly non-obvious

backend/onyx/connectors/teams/connector.py
```diff
@@ -1,97 +1,241 @@
...
+        self,
+        # TODO: (chris) move from "Display Names" to IDs, since display names
+        # are NOT guaranteed to be unique
+        teams: list[str] = [],
+        max_workers: int = MAX_WORKERS,
+    ) -> None:
+        self.graph_client: GraphClient | None = None
+        self.msal_app: msal.ConfidentialClientApplication | None = None
+        self.max_workers = max_workers
+        self.requested_team_list: list[str] = teams
```
evan-onyx, line 59: `requested_team_list` is completely ignored now, is that new behavior that we wanted? If so Would appreciate having this separate from the other instance variables along with a "deprecated" comment. But we definitely need to confirm that we don't need this anymore first.

backend/onyx/connectors/teams/connector.py
```diff
@@ -189,219 +333,130 @@ def _convert_thread_to_document(
...
+        raise RuntimeError(f"No team with {team_id=} was found")
+    elif team_collection.has_next:
+        raise RuntimeError(f"Multiple teams with {team_id=} were found; this is weird")
 
-    def _fetch_from_teams(
-        self, start: datetime | None = None, end: datetime | None = None
-    ) -> GenerateDocumentsOutput:
-        if self.graph_client is None:
-            raise ConnectorMissingCredentialError("Teams")
+    [team] = team_collection
```
evan-onyx: Although this is super cool, we index in with 0 all over the codebase so I'd prefer that for consistency

backend/onyx/connectors/teams/connector.py
```diff
@@ -189,219 +333,130 @@ def _convert_thread_to_document(
...
+) -> list[TodoTeam | TodoChannel]:
+    team_collection = (
+        graph_client.teams.get().filter(f"id eq '{team_id}'").top(1).execute_query()
+    )
 
-        return teams
+    if not team_collection:
+        raise RuntimeError(f"No team with {team_id=} was found")
+    elif team_collection.has_next:
+        raise RuntimeError(f"Multiple teams with {team_id=} were found; this is weird")
```
evan-onyx: probably should remove "this is weird" from the error message, feel free to leave a comment saying basically the same thing though

backend/onyx/connectors/teams/connector.py
```diff
@@ -1,97 +1,241 @@
...
-        thread: list[ChatMessage] = [base_message]
-        thread.extend(replies)
+                    for future in as_completed(futures):
+                        doc = future.result()
+                        if doc:
+                            print(doc)
+                            yield doc
 
-        threads.append(thread)
+        return TeamsCheckpoint(
```
evan-onyx, line 215: We need a strategy for returning a checkpoint part of the way through execution; to my understanding the current approach only checkpoints once after retrieving all teams. To me it seems reasonable to return a checkpoint after each team is finished. An info log saying that we completed a Team would be good too!

backend/onyx/connectors/teams/connector.py
```diff
@@ -1,97 +1,241 @@
...
+                todos.pop()
+                todos.extend(todo_channels)
+            elif isinstance(todo, TodoChannel):
+                tc: list[TodoChannel] = []
+                while todos and isinstance(todos[-1], TodoChannel):
+                    todo = todos.pop()
+                    assert isinstance(todo, TodoChannel)
+                    tc.append(todo)
+
+                with ThreadPoolExecutor(max_workers=self.max_workers) as executor:
```
evan-onyx: Would like to see this code abstracted into a utility in threadpool_concurrency, likely something like `yield_as_completed`! Alternatively you could just switch to using `parallel_yield` if you make your `_collect_document_for_channel_id` a generator.

evan-onyx: If you haven't already, it's worth looking into Teams API rate limits and extrapolating a bit from our testing env to determine whether the default number of workers will get us throttled. Leave your findings in a comment!

backend/onyx/connectors/teams/connector.py
```diff
@@ -189,219 +333,130 @@ def _convert_thread_to_document(
...
-                "Either there are no Teams in this tenant, or your app does not have permission to view them."
-            )
+    channel_collection = (
+        team.channels.get().filter(f"id eq '{channel_id}'").execute_query()
+    )
+    [channel] = channel_collection
+
+    message_collection = channel.messages.get_all(
+        page_loaded=lambda _: None
+    ).execute_query()
```
evan-onyx, line 484: Is there no way to only get recently modified content? Ideally we don't want to run a complete re-indexing every half an hour (or whatever the connector frequency is). It would be good to make use of start and end (from load_checkpoint); maybe we can modify this query to filter based on time like is done https://learn.microsoft.com/en-us/graph/filter-query-parameter?tabs=http
Although, I had a hard time finding the documentation for doing that with the API we're using. Should be doable though!

backend/onyx/connectors/teams/connector.py
```diff
@@ -189,219 +333,130 @@ def _convert_thread_to_document(
...
+
+    chat_messages = [message for message in message_collection]
```
evan-onyx: `chat_messages = list(message_collection)`

evan-onyx: or even just `list(message_collection)` directly in the object you're returning

backend/onyx/connectors/teams/connector.py
```diff
@@ -189,219 +333,130 @@ def _convert_thread_to_document(
...
-            raise ConnectorValidationError(
-                f"Unexpected error during Teams validation: {e}"
-            )
+def _collect_document_for_channel_id(
+    graph_client: GraphClient,
+    team_id: str,
+    channel_id: str,
+) -> Document | None:
+    team_collection = (
+        graph_client.teams.get().filter(f"id eq '{team_id}'").top(1).execute_query()
```
evan-onyx: We shouldn't need to make this query per-document right? I would say pull it out to before this function and pass the result in as an argument

backend/onyx/connectors/teams/connector.py
```diff
@@ -1,97 +1,241 @@
...
-        replies = reply_query.execute_query_retry()
+        while todos:
+            todo = todos[-1]
+            if isinstance(todo, TodoTeam):
+                todo_channels = _collect_all_channels_for_team_id(
+                    graph_client=self.graph_client, team_id=todo.team_id
+                )
+                todos.pop()
+                todos.extend(todo_channels)
+            elif isinstance(todo, TodoChannel):
```
evan-onyx: Rather than having todos be a list of TodoTeam and TodoChannel, here's an alternative I strongly believe is better:

- todos is just a list of TodoTeam
- no more if/elif on type of todo
- you already have todo_channels from the original `if` case, just use that as `tc` in your original `elif` case

The only reason I could see to want the current structure is if you're checkpointing in the middle of processing a team, which isn't currently the case. If you think that's how we should be doing it (below I suggest a checkpointing by team approach, but feel free to push back on that), I would still recommend having a separate list in the checkpoint for handling channels to be processed

backend/onyx/connectors/teams/connector.py
```diff
@@ -189,219 +333,130 @@ def _convert_thread_to_document(
     return doc
```
evan-onyx, line 329: Sorry but could you fix Hagen's code by moving the `if not thread_text` before the `if not post_members_list`? There's no reason to make more API calls if we're just going to return None anyways haha

backend/onyx/connectors/teams/connector.py
```diff
@@ -1,97 +1,241 @@
...
+        return None
+
+    def validate_connector_settings(self) -> None:
+        if self.graph_client is None:
+            raise ConnectorMissingCredentialError("Teams credentials not loaded.")
+
+        try:
+            # Minimal call to confirm we can retrieve Teams
+            # make sure it doesn't take forever, since this is a syncronous call
+            # found_teams = run_with_timeout(10, self._get_all_teams)
```
evan-onyx: Any reason we're removing the run with timeout?

backend/onyx/connectors/teams/connector.py
```diff
@@ -189,219 +333,130 @@ def _convert_thread_to_document(
...
-                    "Team.ReadBasic.All, TeamMember.Read.All, "
-                    "Channel.ReadBasic.All, ChannelMessage.Read.All, "
-                    "Group.Read.All, TeamSettings.ReadWrite.All, "
-                    "ChannelMember.Read.All, ChannelSettings.ReadWrite.All"
-                )
-            raise
+def _collect_all_channels_for_team_id(
+    graph_client: GraphClient,
+    team_id: str,
+) -> list[TodoTeam | TodoChannel]:
```
evan-onyx: We should chat with Chris about this one; it kind of bothers me that we do one big paginated call at the start to get all the teams, then need to do this one-by-one kind of call to retrieve that team a second time once we get around to processing its channels. Another approach would be to only get one page of teams at a time (when we're ready for them) and store a checkpoint after those teams are processed. In that case we'd be doing `team_batch_size` teams per checkpoint unlike the 1 I suggested somewhere else, and the checkpoint would store the offset instead of the full list of teams to be processed

evan-onyx (commented):
Definitely some stuff to chat about/fix!

backend/onyx/connectors/teams/connector.py
raunakab: Good point, not sure how that got added by my editor... Removed!

backend/onyx/connectors/teams/connector.py

backend/onyx/connectors/teams/connector.py
raunakab: Removed!

backend/onyx/connectors/teams/connector.py
raunakab: Yoinked!

backend/onyx/connectors/teams/connector.py
raunakab: Ah whoops, forgot to clean up my weird error messages.

raunakab: Removed!

backend/onyx/connectors/teams/connector.py

backend/onyx/connectors/teams/connector.py

backend/onyx/connectors/teams/connector.py
raunakab: Removed it during testing / development. Added it back now!

backend/onyx/connectors/teams/connector.py
raunakab: Hmm, which old version are you referring to?

backend/onyx/connectors/teams/connector.py
raunakab: Ya, to be honest, I'm not a big fan of it either.

I tried looking into whether or not the Teams API supports querying for a channel directly, but I don't believe it does, unfortunately. I may be wrong, however...

backend/onyx/connectors/slack/connector.py
raunakab: Hmm, these are slightly unrelated changes to this PR, so for simplicity, I'd prefer to remove them.

Updating the env-variable name would also mean updating the `.env.template` file that we have. I don't mind doing that, but this should ideally be in a separate PR!

backend/onyx/connectors/teams/connector.py
raunakab: This is not the new feature. I just forgot to implement it. Oops! Will add support for it back.

backend/onyx/connectors/teams/connector.py
raunakab: Good catch; updated.

backend/onyx/connectors/teams/connector.py
raunakab: Ah, good catch. Will pull it out!

raunakab: Actually, the `todos` should contain a list of `TodoTeam` that have unique `team_id`s, so this should not perform duplicated calls to get the same team. However, I've still updated it for cleanliness's sake.

raunakab: Pulled out of for-loop.

backend/onyx/connectors/teams/connector.py
raunakab: Ah good catch.

In the returned `TeamCollection` object, there exists a value inside of `team.properties["createdDateTime"]` which I can use to filter out channels based off of creation time.

I've updated the code to filter out all of the teams which fall outside the `start - end` datetime range.

raunakab: However, this is a client-side filter only... Will look into performing a server-side filter.

(By client-side, I mean *our* backend, and by server-side, I mean *`MS`'s* backend).

backend/onyx/connectors/teams/connector.py
evan-onyx: the teams connector that's currently on main

backend/onyx/connectors/teams/connector.py
evan-onyx: hopefully that's a property graphql lets you filter by!

backend/onyx/connectors/teams/connector.py
raunakab: I've updated to remove `TodoChannel` and simplified the `TeamCheckpoint` in general, but the only concern that I have now is how to checkpoint when we're in the middle of indexing teams. I guess that is not as critical of an edge-case, however...

backend/onyx/connectors/teams/connector.py
raunakab: Looked through the MS documentation and tried filtering by `createdDateTime` on the objects. Looks like that is not something that can be filtered by on the server-side. Defaulting to doing it on the client-side only for now.

Maybe I can chat with @Weves about this.

raunakab: Closing for now.

backend/onyx/connectors/teams/connector.py
raunakab: Updated!

backend/onyx/connectors/teams/connector.py
raunakab: Added back now! 😄

backend/onyx/connectors/teams/connector.py
raunakab: Add support for cursor based pagination instead.

backend/onyx/connectors/teams/connector.py
```diff
@@ -189,219 +331,181 @@ def _convert_thread_to_document(
...
+            for team_id in [
+                _filter_team_id(team, start, end, requested) for team in team_collection
+            ]
+            if team_id
+        ]
+
+        team_ids.extend(filtered_team_ids)
+
+        if team_collection.has_next:
+            next_url = cast(str, team_collection._next_request_url)
```
evan-onyx: slightly prefer raising ValueError if _next_request_url isn't a string

backend/onyx/connectors/teams/connector.py
```diff
@@ -189,219 +331,181 @@ def _convert_thread_to_document(
...
+    if requested:
+        filter = " or ".join(f"displayName eq '{team_name}'" for team_name in requested)
 
-            token = self.msal_app.acquire_token_for_client(
-                scopes=["https://graph.microsoft.com/.default"]
+    while True:
+        if filter:
+            query = graph_client.teams.get().filter(filter)
+        else:
+            query = graph_client.teams.get_all(page_loaded=lambda _: None)
```
evan-onyx: Why do we need page_loaded? delete or leave a comment if there's a reason for it

backend/onyx/connectors/teams/connector.py
```diff
@@ -189,219 +331,181 @@ def _convert_thread_to_document(
...
-                while True:
-                    # Get a page of teams with the filter
-                    teams_page = (
-                        self.graph_client.teams.get()
-                        .filter(filter_conditions)
-                        .top(page_size)
-                        .skip(skip)
-                        .execute_query()
-                    )
+    if created_at := props.get("createdDateTime"):
```
evan-onyx: If we filter teams by created time, new messages in channels in those teams will be ignored by incremental update connector runs. We need to filter messages by their time of "update" (creation or modification) ideally, but if that isn't possible then filtering by message creation time is probably good enough here (we'll just miss edge cases like a message gets modified after the indexing attempt that originally picked it up)

backend/onyx/connectors/teams/connector.py
```diff
@@ -189,219 +331,181 @@ def _convert_thread_to_document(
...
+    return team.id
 
-        return teams
 
-    def _fetch_from_teams(
-        self, start: datetime | None = None, end: datetime | None = None
-    ) -> GenerateDocumentsOutput:
-        if self.graph_client is None:
-            raise ConnectorMissingCredentialError("Teams")
+def _get_team_with_id(
```
evan-onyx: nit: kind of prefer `_get_team_by_id` I think that's more in line with the rest of the codebase

backend/onyx/connectors/teams/connector.py
```diff
@@ -189,219 +331,181 @@ def _convert_thread_to_document(
...
+    )
 
-        teams = self._get_all_teams()
-        logger.debug(f"Found available teams: {[str(t) for t in teams]}")
-        if not teams:
-            msg = "No teams found."
-            logger.error(msg)
-            raise ValueError(msg)
+    if not team_collection:
+        raise RuntimeError(f"No team with {team_id=} was found")
```
evan-onyx: nit: I think this should probably be ValueError but not a big deal

backend/onyx/connectors/teams/connector.py
```diff
@@ -189,219 +331,181 @@ def _convert_thread_to_document(
...
+        raise RuntimeError(f"The {team=} has an empty `id` field")
 
-        logger.debug(
-            f"Found available channels (max {TeamsConnector.MAX_CHANNELS_TO_LOG} shown): "
-            f"{[c.id for c in channels[:TeamsConnector.MAX_CHANNELS_TO_LOG]]}"
+    team_and_channel_id_pairs: list[tuple[str, str]] = []
+    next_url = None
+
+    while True:
+        query = team.channels.get_all(page_loaded=lambda _: None)
```
evan-onyx: leave comment about page_loaded or delete

evan-onyx: Just give this a thumbs up if you already checked for server side modified time filtering by channel, otherwise give it a look please!

backend/onyx/connectors/teams/connector.py
evan-onyx: correct me if I'm wrong, but I don't think it's possible to reach this `if` statement with a `None` or empty list for `todos`. If that's true, I'd still say we should remove this `if`!

evan-onyx: And just leave a comment instead explaining that fact

backend/onyx/connectors/teams/connector.py
```diff
@@ -32,66 +38,202 @@
...
+
+        if not todos:
+            return TeamsCheckpoint(
+                todo_team_ids=[],
+                has_more=False,
+            )
 
-        reply_query = base_message.replies.get_all()
-        replies = reply_query.execute_query_retry()
+        todo_team_id = todos[-1]
```
evan-onyx: nit: `todo_team_id = todos.pop()` and then you can remove the later `todos.pop()`

backend/onyx/connectors/teams/connector.py
evan-onyx: Please add the info log!

backend/onyx/connectors/teams/connector.py
```diff
@@ -32,66 +38,202 @@
...
+            graph_client=self.graph_client,
+            team_id=todo_team_id,
+        )
+        team_and_channel_id_pairs = _collect_all_channels_for_team_id(
+            graph_client=self.graph_client,
+            team=team,
+        )
+        todos.pop()
+
+        with ThreadPoolExecutor(max_workers=self.max_workers) as executor:
```
evan-onyx: not sure how my comment on this got lost; I think this pattern (yielding values as they are completed from a list of function tuples to run in parallel) really should be either a utility in threadpool_concurrency.py or you can re-use parallel_yield by turning `_collect_document_for_channel_id` into a generator

backend/onyx/connectors/teams/connector.py
evan-onyx: Just wanted to reiterate the above point; filtering teams by created time doesn't work, so make sure to filter the messages here instead!

evan-onyx (commented):
Make sure to add automated tests too!

backend/onyx/connectors/teams/connector.py
raunakab: Updated.

backend/onyx/connectors/teams/connector.py
raunakab: The reason why it's needed is because of the incorrect type-annotations that the `office365` library has.

It defines `page_loaded: Callable[.., None]`, but default sets it to `None`. Their type declaration is incorrect and that causes my editor to say that declaring `page_loaded=None` (or not defining it at all) to be an error.

I'll jot down a comment.

backend/onyx/connectors/teams/connector.py

backend/onyx/connectors/teams/connector.py

backend/onyx/connectors/teams/connector.py

backend/onyx/connectors/teams/connector.py
raunakab: Comment added!

backend/onyx/connectors/teams/connector.py
raunakab: Left a comment about `page_loaded`.

I'll give the server-side channel-modified-datetime another look.

raunakab: No luck with the server-side `modifiedDateTime` field for teams or for channels... Resolving this thread for now.

backend/onyx/connectors/teams/connector.py
raunakab: I was able to test that edge case, and yes, we do miss updates to previously sent messages.

@Weves do you have any suggestions on this? Hagen's previous implementation did not tackle this; not sure if this is something that the MS Graph API supports in the first place.

backend/onyx/connectors/teams/connector.py
raunakab: Nice suggestion. Using `parallel_yield` here instead.

backend/onyx/connectors/teams/connector.py
raunakab: Updated to using `parallel_yield`!

backend/onyx/connectors/teams/connector.py
raunakab: Filtering messages is done a couple lines down. I filter them based on whether or not `_filter_message` returns `True` or `False`.

backend/onyx/connectors/teams/connector.py
raunakab: Updated to filtering (client-side 😭) on a `"lastModifiedDatetime"` field that exists. Although it's not perfect (server-side filtering would've made it ideal), it's the best we can do to avoid this edge-case.

raunakab: Resolving for now.

backend/onyx/connectors/teams/connector.py
```diff
@@ -1,97 +1,238 @@
...
-    threads: list[list[ChatMessage]] = []
-    for base_message in base_messages:
-        message_datetime = time_str_to_utc(
-            base_message.properties["lastModifiedDateTime"]
+    def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None:
+        teams_client_id = credentials["teams_client_id"]
+        teams_client_secret = credentials["teams_client_secret"]
+        teams_directory_id = credentials["teams_directory_id"]
+
+        authority_url = f"https://login.microsoftonline.com/{teams_directory_id}"
```
rkuo-danswer: This constant prefix might be better as a class var.

backend/onyx/connectors/teams/connector.py
```diff
@@ -189,219 +330,212 @@ def _convert_thread_to_document(
...
+        if _filter_message(message=message, start=start, end=end)
+    ]
+
+    yield _convert_thread_to_document(
+        channel=channel,
+        thread=thread,
+    )
+
+
+def _filter_message(
```
rkuo-danswer, line 498: docstring for this.  what does returning True or False mean?

backend/onyx/connectors/teams/connector.py
```diff
@@ -189,219 +330,212 @@ def _convert_thread_to_document(
...
+        channel_collection = query.execute_query()
+        channels.extend(channel for channel in channel_collection if channel.id)
+
+        if not channel_collection.has_next:
+            break
+
+    return channels
+
+
+def _collect_document_for_channel_id(
```
rkuo-danswer, line 461: document or document(s)?

backend/onyx/connectors/teams/connector.py
```diff
@@ -189,219 +330,212 @@ def _convert_thread_to_document(
...
+        raise ValueError(f"No team with {team_id=} was found")
+    elif team_collection.has_next:
+        # shouldn't happen, but catching it regardless
+        raise RuntimeError(f"Multiple teams with {team_id=} were found")
+
+    return team_collection[0]
+
+
+def _collect_all_channels_from_team(
+    graph_client: GraphClient,
```
rkuo-danswer, line 432: any reason all these functions are standalone instead of class methods?

backend/onyx/connectors/teams/connector.py
```diff
@@ -189,219 +330,212 @@ def _convert_thread_to_document(
...
-        teams = self._get_all_teams()
-        logger.debug(f"Found available teams: {[str(t) for t in teams]}")
-        if not teams:
-            msg = "No teams found."
-            logger.error(msg)
-            raise ValueError(msg)
+def _filter_team_id(
+    team: Team,
+    requested: list[str] | None = None,
+) -> str | None:
```
rkuo-danswer, line 390: is this a filter?  can we docstring the intent of this?

backend/onyx/connectors/teams/connector.py
```diff
@@ -1,97 +1,238 @@
...
+        team = _get_team_by_id(
+            graph_client=self.graph_client,
+            team_id=todo_team_id,
+        )
+        channels = _collect_all_channels_from_team(
+            graph_client=self.graph_client,
+            team=team,
+        )
+
+        gens = [
```
rkuo-danswer: what are "gens"?

backend/onyx/connectors/teams/connector.py
```diff
@@ -1,97 +1,238 @@
...
+        ]
+
+        for doc in parallel_yield(
+            gens=gens,
+            max_workers=self.max_workers,
+        ):
+            if doc:
+                yield doc
+
+        logger.info(f"Processed team with id {todo_team_id}")
```
rkuo-danswer: would be more useful with the len of remaining known team_id's

backend/onyx/connectors/teams/connector.py
```diff
@@ -1,97 +1,238 @@
...
-    teams: list[Team],
-) -> list[Channel]:
-    channels_list: list[Channel] = []
-    for team in teams:
-        query = team.channels.get_all()
-        channels = query.execute_query_retry()
-        channels_list.extend(channels)
+        except ClientRequestException as e:
+            if not e.response:
+                raise RuntimeError("TODO!")
```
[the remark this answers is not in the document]
raunakab: Updated.

backend/onyx/connectors/teams/connector.py

backend/onyx/connectors/teams/connector.py
raunakab: Short for generator. I used them since that was the parameter name inside of `parallel_yield`.

I updated it to be `docs`, since that would be more appropriate.

backend/onyx/connectors/teams/connector.py
raunakab: Added docstring!

backend/onyx/connectors/teams/connector.py
raunakab: Added docstring!

backend/onyx/connectors/teams/connector.py
raunakab: This actually just yields one singular `Document`. The reason why this method is an `Iterator` is because that's what `parallel_yield` expects.

backend/onyx/connectors/teams/connector.py
raunakab: This is the pattern that I saw in the Slack connector. I tried to adhere to that style.

backend/onyx/connectors/teams/connector.py

backend/onyx/connectors/teams/connector.py
```diff
@@ -189,219 +333,226 @@ def _convert_thread_to_document(
...
-        teams_client_id = credentials["teams_client_id"]
-        teams_client_secret = credentials["teams_client_secret"]
-        teams_directory_id = credentials["teams_directory_id"]
+def _collect_all_team_ids(
+    graph_client: GraphClient,
+    start: SecondsSinceUnixEpoch,
+    end: SecondsSinceUnixEpoch,
+    requested: list[str] | None = None,
+) -> list[str]:
+    # The geniuses over at Microsoft decided that the Graph API does not
```
evan-onyx: loool as much as I love this, probably should not merge comments that explicitly flame microsoft 😅 Can just remove everything until "The Graph API does not..."

evan-onyx: start/end shouldn't be params of this function anymore right? and The comment should probably be moved to the place where we retrieve messages

backend/onyx/connectors/teams/connector.py
evan-onyx: Should add this context to the docstring!

backend/onyx/connectors/teams/connector.py
```diff
@@ -189,219 +333,226 @@ def _convert_thread_to_document(
...
+        return start <= dt_ts and dt_ts < end
+
+    if modified_at := props.get("lastModifiedDateTime"):
+        if isinstance(modified_at, str):
+            return compare(modified_at)
+
+    if created_at := props.get("createdDateTime"):
+        if isinstance(created_at, str):
+            return compare(created_at)
+
```
evan-onyx, line 524: could we add a warning log here saying no created or modified field found

backend/tests/daily/connectors/teams/test_teams_connector.py
```diff
@@ -0,0 +1,75 @@
...
+            ...
+        elif isinstance(teams, list):
+            for name in teams:
+                assert isinstance(name, str)
+        else:
+            raise ValueError(
+                f"`request.param` must either be `None` or of type `list[str]`; instead got {type(teams)}"
+            )
+
+    teams_connector = TeamsConnector(teams=teams if teams else [])
```
evan-onyx: `teams or []`

evan-onyx (approved):
Looking good! Make sure to address the last few nits and put teams credentials in our github secrets if you haven't already so the connector tests pass

backend/tests/daily/connectors/teams/test_teams_connector.py
raunakab: Updated!

backend/onyx/connectors/teams/connector.py
raunakab: Added!

backend/onyx/connectors/teams/connector.py
raunakab: Added!

backend/onyx/connectors/teams/connector.py

PR #4602, opened 2025-04-24
perf: Optimize query history exporting process

raunakab:
## Description

For sufficiently small query-histories, the CSV exporting process succeeds. For larger query-histories, the exporting process hangs and dies.

This PR optimizes that process.

Addresses: https://linear.app/danswer/issue/DAN-1520/export-query-csv-times-out.

## How Has This Been Tested?

TODO!


backend/ee/onyx/background/celery/apps/primary.py
```diff
@@ -79,3 +96,77 @@ def autogenerate_usage_report_task(*, tenant_id: str) -> None:
...
+
+
+@celery_app.task(
+    name=OnyxCeleryTask.EXPORT_QUERY_HISTORY_TASK,
+    ignore_result=True,
+    soft_time_limit=JOB_TIMEOUT,
+    bind=True,
+)
+def export_query_history_task(self: Task, *, start: datetime, end: datetime) -> None:
+    # Importing here because importing in the global namespace causes a circular dependency issue.
```
evan-onyx: Is this common for bg tasks?

backend/ee/onyx/background/celery/apps/primary.py
```diff
@@ -79,3 +96,77 @@ def autogenerate_usage_report_task(*, tenant_id: str) -> None:
...
+    from ee.onyx.server.query_history.api import fetch_and_process_chat_session_history
+    from ee.onyx.server.query_history.api import ONYX_ANONYMIZED_EMAIL
+
+    if not self.request.id:
+        raise RuntimeError("No task id defined for this task; cannot identify it")
+
+    task_id = self.request.id
+
+    with get_session_with_current_tenant() as db_session:
```
evan-onyx: I thought tenant needed to be passed into background jobs explicitly? Would be very happy to be wrong about this, though!

backend/ee/onyx/server/query_history/api.py
```diff
@@ -238,52 +241,97 @@ def get_chat_session_admin(
...
+    request_id: str,
+    _: User | None = Depends(current_admin_user),
+    db_session: Session = Depends(get_session),
+) -> dict[str, str]:
+    assert_query_history_is_enabled()
+
+    task = get_task_with_id(db_session=db_session, task_id=request_id)
+
+    if task:
+        status = task.status
```
evan-onyx: optional nit: I personally prefer returning task.status here so you can remove the `else:` and de-indent. But if you like the current structure it's nbd

backend/ee/onyx/server/query_history/api.py
```diff
@@ -238,52 +241,97 @@ def get_chat_session_admin(
...
+    if not task:
+        raise HTTPException(
+            404,
+            f"No task with {request_id=} was found",
+        )
+
+    # Maybe we should handle each specific TaskStatus separately?
+    # I.e., if it's a `TaskStatus.PENDING` return a different error code vs if it's a `TaskStatus.FAILURE`.
+    if task.status != TaskStatus.SUCCESS:
+        raise HTTPException(500, f"Task with {request_id=} has not finished yet")
```
evan-onyx: Should this be a 400? Imo 500 is mostly used when stuff actually crashes

evan-onyx: And yeah I'm a fan of having a different error message based on task status!

web/src/app/ee/admin/performance/query-history/DownloadAsCSV.tsx
```diff
@@ -1,13 +1,92 @@
...
+const CHECK_QUERY_HISTORY_EXPORT_STATUS_URL =
+  "/api/admin/query-history/export-status";
+const DOWNLOAD_QUERY_HISTORY_URL = "/api/admin/query-history/download";
+
+type StartQueryHistoryExportResponse = { request_id: string };
+type CheckQueryHistoryExportStatusResponse = {
+  status: "PENDING" | "STARTED" | "SUCCESS" | "FAILURE";
+};
+
+type Status = "NULL" | "PENDING" | "SUCCESS" | "FAILURE";
```
evan-onyx: what kind of status is this? Would like a comment or small name change considering the overlap with `CheckQueryHistoryExportStatusResponse`

web/src/app/ee/admin/performance/query-history/DownloadAsCSV.tsx
```diff
@@ -1,13 +1,92 @@
...
+
+const withRequestId = (url: string, requestId: string): string =>
+  `${url}?request_id=${requestId}`;
```
evan-onyx: do we really not have utils for this? If so that's pretty cooked

evan-onyx (commented):
I don't fully understand how the frontend is supposed to work, but broadly looks good! Left some comments on things we should confirm about the backend.

backend/ee/onyx/server/query_history/api.py
raunakab: Oh that's a good point, will de-indent!

backend/ee/onyx/server/query_history/api.py
raunakab: Will update in the coming commit.

backend/ee/onyx/server/query_history/api.py
raunakab: Updated.

backend/ee/onyx/server/query_history/api.py
raunakab: Updated.

web/src/app/ee/admin/performance/query-history/DownloadAsCSV.tsx
raunakab: Haha, not sure if we do! Ping @Weves: do you know if we have some?

web/src/app/ee/admin/performance/query-history/DownloadAsCSV.tsx
raunakab: Updated the name of this type. I agree it was too vague before.

backend/ee/onyx/background/celery/apps/primary.py
raunakab: This is how I've seen other instance of background tasks being implemented. Was essentially pattern-matching on the background task launching mechanics...

raunakab: Ping @LostVector and @Weves; do you guys think this is how background tasks should be launched?

backend/ee/onyx/background/celery/apps/primary.py
raunakab: Another ping for @Weves and @LostVector. Do you guys know if the tenant needs to be passed in explicitly?

raunakab: @evan-onyx The function should fetch it (assuming the name of it is correct). Looked into the function body, and it gets the current tenant and returns the according database session.

backend/ee/onyx/server/query_history/api.py
```diff
@@ -41,6 +44,16 @@
 ONYX_ANONYMIZED_EMAIL = "anonymous@anonymous.invalid"
 
 
+def assert_query_history_is_enabled(
+    disallowed: list[QueryHistoryType] = [QueryHistoryType.DISABLED],
```
Weves: nit: prefer no default here. Would rather have the user always pass them in.

backend/ee/onyx/background/celery/apps/primary.py
Weves: I think this is not ideal but fine. Can keep it in for now

web/src/app/ee/admin/performance/query-history/DownloadAsCSV.tsx
```diff
@@ -1,13 +1,137 @@
...
+    if (timerIdRef.current) {
+      clearInterval(timerIdRef.current);
+      timerIdRef.current = null;
+    }
+    retryCount.current = 0;
+
+    if (failure) {
+      // Throw up an alert here.
+      //
+      // TODO:
```
Weves: we should use `setPopup`

web/src/app/ee/admin/performance/query-history/DownloadAsCSV.tsx
Weves: we do not have a util for this, although I don't think this is a standard format (the `request_id=` part). 

Unless you're talking about a util to add query params? If so, I think there's a standard lib for that

Weves: I think this is fine

raunakab: Sounds good. Just needed a quick, ad-hoc way of generating the URL.

backend/ee/onyx/background/celery/apps/primary.py
raunakab: Sounds good. Keeping it in for now.

backend/ee/onyx/server/query_history/api.py
raunakab: Updated.

web/src/app/ee/admin/performance/query-history/DownloadAsCSV.tsx
raunakab: Updated.

backend/ee/onyx/server/query_history/api.py
```diff
@@ -238,52 +241,98 @@ def get_chat_session_admin(
...
+    if not task:
+        raise HTTPException(
+            404,
+            f"No task with {request_id=} was found",
+        )
+
+    # Maybe we should handle each specific TaskStatus separately?
+    # I.e., if it's a `TaskStatus.PENDING` return a different error code vs if it's a `TaskStatus.FAILURE`.
+    if task.status in [TaskStatus.STARTED, TaskStatus.PENDING]:
+        raise HTTPException(102, f"Task with {request_id=} is still being worked on")
```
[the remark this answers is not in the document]
raunakab: Updated.

web/src/app/ee/admin/performance/query-history/DownloadAsCSV.tsx
```diff
@@ -1,13 +1,141 @@
...
+// Otherwise, the spinning animation should be shown.
+type SpinnerStatus = "static" | "spinning";
+
+const withRequestId = (url: string, requestId: string): string =>
+  `${url}?request_id=${requestId}`;
+
 export function DownloadAsCSV() {
+  const timerIdRef = useRef<null | number>(null);
+  const retryCount = useRef<number>(0);
+  const [, rerender] = useState();
```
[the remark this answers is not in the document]
raunakab: Updated.

web/src/app/ee/admin/performance/query-history/DownloadAsCSV.tsx
```diff
@@ -1,13 +1,141 @@
...
+      () => checkStatus(request_id),
+      RETRY_COOLDOWN_MILLISECONDS
+    ) as unknown as number;
+    timerIdRef.current = timer;
+    rerender(undefined);
+  };
+
+  const checkStatus = async (requestId: string) => {
+    rerender(undefined);
+    if (retryCount.current >= MAX_RETRIES) {
```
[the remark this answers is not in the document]
raunakab: Updated.

backend/ee/onyx/background/celery/apps/primary.py
```diff
@@ -79,3 +96,83 @@ def autogenerate_usage_report_task(*, tenant_id: str) -> None:
...
         )
+
+
+@celery_app.task(
+    name=OnyxCeleryTask.EXPORT_QUERY_HISTORY_TASK,
+    ignore_result=True,
+    soft_time_limit=JOB_TIMEOUT,
+    bind=True,
+)
+def export_query_history_task(self: Task, *, start: datetime, end: datetime) -> None:
```
rkuo-danswer: it might be a good time to look into switching everything here to the @shared_task annotation that we use almost everywhere else to get rid of the strong coupling that results in circular dependencies between tasks and code.

backend/ee/onyx/background/celery/apps/primary.py
```diff
@@ -79,3 +96,83 @@ def autogenerate_usage_report_task(*, tenant_id: str) -> None:
...
+    # Importing here because importing in the global namespace causes a circular dependency issue.
+    from ee.onyx.server.query_history.api import fetch_and_process_chat_session_history
+    from ee.onyx.server.query_history.api import ONYX_ANONYMIZED_EMAIL
+
+    if not self.request.id:
+        raise RuntimeError("No task id defined for this task; cannot identify it")
+
+    task_id = self.request.id
+
+    with get_session_with_current_tenant() as db_session:
```
rkuo-danswer: This session looks like it could be long running.  We should look at shortening the session in opportune places ... where would those be?

backend/ee/onyx/server/query_history/api.py
```diff
@@ -41,6 +44,16 @@
 ONYX_ANONYMIZED_EMAIL = "anonymous@anonymous.invalid"
 
 
+def assert_query_history_is_enabled(
```
rkuo-danswer: assert has a very specific meaning, we shouldn't call this that

backend/ee/onyx/server/query_history/api.py
```diff
@@ -238,52 +241,100 @@ def get_chat_session_admin(
...
+    assert_query_history_is_enabled(disallowed=[QueryHistoryType.DISABLED])
+
+    start = start or datetime.fromtimestamp(0, tz=timezone.utc)
+    end = end or datetime.now(tz=timezone.utc)
+
+    if start >= end:
         raise HTTPException(
-            status_code=HTTPStatus.FORBIDDEN,
-            detail="Query history has been disabled by the administrator.",
+            400,
```
rkuo-danswer: i like using the HTTPStatus enum, but that's just me!

backend/ee/onyx/server/query_history/api.py
```diff
@@ -238,52 +241,100 @@ def get_chat_session_admin(
...
+            f"Start time must come before end time, but instead got the start time coming after; {start=} {end=}",
```
rkuo-danswer, line 295: wow i did not know this {var=} notation. awesome

backend/ee/onyx/server/query_history/api.py
```diff
@@ -238,52 +241,100 @@ def get_chat_session_admin(
...
 
-    # this call is very expensive and is timing out via endpoint
-    # TODO: optimize call and/or generate via background task
-    complete_chat_session_history = fetch_and_process_chat_session_history(
-        db_session=db_session,
-        start=start or datetime.fromtimestamp(0, tz=timezone.utc),
-        end=end or datetime.now(tz=timezone.utc),
-        feedback_type=None,
-        limit=None,
+    task = export_query_history_task.apply_async(
```
rkuo-danswer: Again let's not use apply_async, we've switched to send_task everywhere i think.

backend/ee/onyx/background/celery/apps/primary.py
```diff
@@ -79,3 +96,83 @@ def autogenerate_usage_report_task(*, tenant_id: str) -> None:
...
+                content=stream,
+                display_name=report_name,
+                file_origin=FileOrigin.GENERATED_REPORT,
+                file_type="text/csv",
+            )
+        except Exception:
+            logger.exception(f"Failed to save query history export file {report_name}")
+            raise
+
+        mark_task_as_finished_with_id(
```
rkuo-danswer: we need a cleanup strategy for tasks that die in the middle of processing.

backend/ee/onyx/server/query_history/api.py
raunakab: It's quite convenient!

backend/ee/onyx/background/celery/apps/primary.py
raunakab: Refactored the code and shortened it up into 2 smaller pieces.

backend/ee/onyx/server/query_history/api.py
raunakab: Changed it to `ensure_query_history_is_enabled` instead.

backend/ee/onyx/server/query_history/api.py
raunakab: Oh good point. Updated to use that enum instead.

backend/ee/onyx/server/query_history/api.py
raunakab: Switched over to using `shared_task` and calling `client_app.send_task` instead.

backend/ee/onyx/background/celery/apps/primary.py
raunakab: Switched over to using `shared_task` instead. However, I would still like a way of jotting down tasks as they come up (i.e., for very long-running tasks). For those types of situations, I like registering a task.

backend/ee/onyx/background/celery/apps/primary.py
raunakab: Yes, that's true... Looking into this right now.

raunakab: Added try-excepts around the logic; I then mark the tasks as failed if an exception was raised.

backend/ee/onyx/server/query_history/api.py
```diff
@@ -238,52 +246,126 @@ def get_chat_session_admin(
...
 
-    # this call is very expensive and is timing out via endpoint
-    # TODO: optimize call and/or generate via background task
-    complete_chat_session_history = fetch_and_process_chat_session_history(
-        db_session=db_session,
-        start=start or datetime.fromtimestamp(0, tz=timezone.utc),
-        end=end or datetime.now(tz=timezone.utc),
-        feedback_type=None,
-        limit=None,
+    task = client_app.send_task(
```
rkuo-danswer, line 298: you still want to make sure this goes to a proper queue and worker ... possibly the heavy queue.

backend/ee/onyx/background/celery/apps/primary.py
rkuo-danswer: what if the process is terminated without a chance for the exception handler to run?

backend/ee/onyx/server/query_history/api.py
raunakab: Updated to throw onto a new queue.

backend/ee/onyx/background/celery/apps/heavy.py
```diff
@@ -0,0 +1,137 @@
...
+
+@shared_task(
+    name=OnyxCeleryTask.EXPORT_QUERY_HISTORY_TASK,
+    ignore_result=True,
+    soft_time_limit=JOB_TIMEOUT,
+    bind=True,
+    trail=False,
+)
+def export_query_history_task(self: Task, *, start: datetime, end: datetime) -> None:
+    # Importing here because importing in the global namespace causes a circular dependency issue.
```
rkuo-danswer: do we still need this after switching to shared task?

backend/ee/onyx/background/celery/apps/heavy.py
```diff
@@ -0,0 +1,137 @@
...
+            chat_session_snapshot
+        )
+
+    qa_pairs: list[QuestionAnswerPairSnapshot] = [
+        qa_pair
+        for chat_session_snapshot in complete_chat_session_history
+        for qa_pair in to_qa_pair(chat_session_snapshot)
+    ]
+
+    report_name = construct_query_history_report_name(task_id)
```
rkuo-danswer: perhaps we can call this closer to where it's being used

backend/ee/onyx/background/celery/apps/heavy.py
```diff
@@ -0,0 +1,137 @@
...
+                db_session=db_session,
+                task_id=task_id,
+                success=False,
+            )
+            raise
+
+    def to_qa_pair(
+        chat_session_snapshot: ChatSessionSnapshot,
+    ) -> list[QuestionAnswerPairSnapshot]:
+        if ONYX_QUERY_HISTORY_TYPE == QueryHistoryType.ANONYMIZED:
```
rkuo-danswer: Some possibilities to make this function more pure?

backend/ee/onyx/background/celery/tasks/cleanup/tasks.py
```diff
@@ -0,0 +1,46 @@
...
+                    delete_task_with_id(db_session=db_session, task_id=task.task_id)
+                    log_error_message(task)
+                    continue
+
+                deadline = task.start_time + timedelta(hours=24)
+                now = datetime.now()
+
+                if now >= deadline:
+                    log_error_message(task)
+                    delete_task_with_id(db_session=db_session, task_id=task.task_id)
```
rkuo-danswer: usually prefer not to have dangling if's.  We should explicitly pass the else: case if there's truly nothing to do

backend/ee/onyx/background/task_name_builders.py
```diff
@@ -1,2 +1,14 @@
...
+
+
+QUERY_HISTORY_TASK_NAME_PREFIX = f"{OnyxCeleryTask.EXPORT_QUERY_HISTORY_TASK}_"
+
+
 def name_chat_ttl_task(retention_limit_days: int, tenant_id: str | None = None) -> str:
     return f"chat_ttl_{retention_limit_days}_days"
+
+
+def query_history_task_name(start: datetime, end: datetime) -> str:
```
rkuo-danswer, line 13: did we want an underscore before start?

backend/ee/onyx/server/query_history/api.py
```diff
@@ -238,52 +249,151 @@ def get_chat_session_admin(
...
+
+    if has_file:
+        try:
+            csv_stream = file_store.read_file(report_name)
+        except Exception as e:
+            raise HTTPException(
+                HTTPStatus.INTERNAL_SERVER_ERROR,
+                f"Failed to read query history file: {str(e)}",
+            )
+        csv_stream.seek(0)
```
rkuo-danswer, line 369: do we actually need to seek to 0 if we just got the stream?

backend/onyx/background/task_utils.py
```diff
@@ -114,3 +117,13 @@ def wrap_task(task: Task) -> Task:
         return task
 
     return wrap_task
+
+
+def construct_query_history_report_name(
+    task_id: str,
+) -> str:
+    return f"{QUERY_REPORT_NAME_PREFIX}{task_id}.csv"
```
rkuo-danswer: I would slightly prefer not to have the prefix contain its own delimiter, but not hugely important.

backend/onyx/db/tasks.py
```diff
@@ -57,6 +67,60 @@ def register_task(
...
+
+
+def mark_task_as_finished_with_id(
+    db_session: Session,
+    task_id: str,
+    success: bool = True,
+) -> None:
+    task = get_task_with_id(db_session=db_session, task_id=task_id)
+    if not task:
+        raise RuntimeError(f"A task with the task-id {task_id} does not exist")
```
rkuo-danswer: could use that fancy format= here?

backend/onyx/db/tasks.py
```diff
@@ -57,6 +67,60 @@ def register_task(
...
+    db_session.execute(delete(TaskQueueState).where(TaskQueueState.task_id == task_id))
+    db_session.commit()
+
+
+def get_all_query_history_export_tasks(
+    db_session: Session,
+) -> list[TaskQueueState]:
+    return list(
+        db_session.scalars(
+            select(TaskQueueState).where(
```
rkuo-danswer, line 92: do we need an index on any columns or are we good?

backend/onyx/background/task_utils.py
```diff
@@ -114,3 +117,13 @@ def wrap_task(task: Task) -> Task:
...
+
+
+def deconstruct_query_history_report_name(name: str) -> str:
```
rkuo-danswer: what is this doing?   extracting the task id?  Comments might help

backend/onyx/db/tasks.py
raunakab: Updated!

backend/onyx/background/task_utils.py
raunakab: That's a fair critique. I don't feel too strongly either, so I'll go with your suggestion. Updated!

backend/ee/onyx/server/query_history/api.py
raunakab: I've seen @Weves do this in other examples, so I left it in there for parity's sake.

backend/onyx/background/task_utils.py
raunakab: The function name was a bit obtuse; I changed it to be more descriptive. It is now `extract_task_id_from_query_history_report_name`, which I think should fit the bill here.

backend/ee/onyx/background/task_name_builders.py
raunakab: This was the same issue as the report-name: the prefix contained the delimiter, and therefore I intentionally left the underscore before start out.

However, placing the delimiter in front of the prefix doesn't seem as clean, so thus I removed it.

backend/ee/onyx/background/celery/apps/heavy.py
raunakab: Moved.

backend/ee/onyx/background/celery/apps/heavy.py
raunakab: Good point; I've moved them to the top of the file now!

backend/ee/onyx/background/celery/tasks/cleanup/tasks.py
raunakab: I updated the flow just a bit. Now we don't have a dangling if-statement.

backend/onyx/db/tasks.py
raunakab: I don't think so. We only recognize "query-history-export" tasks via their name, so selecting based off of TaskQueueState.task_name.list(..) should be sufficient for now.

backend/ee/onyx/background/celery/apps/heavy.py
raunakab: Updated!

backend/ee/onyx/server/query_history/api.py
raunakab: I guess it's not strictly needed, but I think @Weves likes it to be explicit? Pinging @Weves here.

rkuo-danswer (approved):

PR #4604, opened 2025-04-24
fix provisioning and don't spawn tasks which could result in a race c…

rkuo-danswer:
…ondition

## Description

Task was not being discovered correctly.
Also had a design issue where it would continue to dispatch new tenants if the queue was backed up.  Performing the check and the provisioning together fixes this.

Fixes https://linear.app/danswer/issue/DAN-1876/tenant-pre-provisioning-is-broken

## How Has This Been Tested?

### DIFF
--- a/backend/onyx/background/celery/tasks/tenant_provisioning/tasks.py
+++ b/backend/onyx/background/celery/tasks/tenant_provisioning/tasks.py
@@ -14,9 +14,8 @@ from ee.onyx.server.tenants.provisioning import setup_tenant
 from ee.onyx.server.tenants.schema_management import create_schema_if_not_exists
 from ee.onyx.server.tenants.schema_management import get_current_alembic_version
 from onyx.background.celery.apps.app_base import task_logger
-from onyx.configs.app_configs import JOB_TIMEOUT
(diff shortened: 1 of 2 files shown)


backend/onyx/background/celery/tasks/tenant_provisioning/tasks.py
```diff
@@ -71,32 +71,28 @@ def check_available_tenants(self: Task) -> None:
...
-        target_available_tenants = getattr(
+        num_minimum_available_tenants = getattr(
             TARGET_AVAILABLE_TENANTS, "value", DEFAULT_TARGET_AVAILABLE_TENANTS
         )
 
         # Calculate how many new tenants we need to provision
-        tenants_to_provision = max(
-            0, target_available_tenants - available_tenants_count
-        )
+        if num_available_tenants < num_minimum_available_tenants:
```
evan-onyx, line 82: nit: probably easier to keep the max() and change the variable names?

evan-onyx (approved):

backend/onyx/background/celery/tasks/tenant_provisioning/tasks.py
rkuo-danswer: i don't find the max intuitive

Issue #4605, opened 2025-04-24
Feature Request:  Jira connector indexes ticket numbers


bsmithwick:
Currently the Jira connector doesn't index ticket numbers, which means that you can only refer to tickets by their titles, NOT their numbers, when chatting with Onyx.  Users typically want to refer to tickets by their numbers, so this limits the utility of the connector.

We have prepared a fix for this issue, and will create a PR shortly.

PR #4606, opened 2025-04-24
Add ticket number indexing to Jira connector

bsmithwick:
## Description

Currently the Jira connector doesn't index ticket numbers, which means that you can only refer to tickets by their titles, NOT their numbers, when chatting with Onyx. Users typically want to refer to tickets by their numbers, so this limits the utility of the connector.  See issue #4605.

## How Has This Been Tested?

Tested a project with a Jira connector both with and without the code in this PR.  Without this code, we could not refer to tickets by their numbers in an Onyx search (e.g. "Tell me about ST-1234").  We were able to refer to tickets by their numbers with this code in place, *after* reindexing all documents for that connector.


bsmithwick:
cc @jwjordan

Orbital-Web:
This (should) be fixed in the latest version. Please let me know otherwise. Thank you!

PR #4623, opened 2025-04-27
Fix LLM API key

Weves:
## Description

Fixes a few things:

1. Model configurations were getting wiped on every update (set to empty list).
2. You could not edit an LLM provider easily. If you tried to re-submit, it would try and use the "masked key" and fail.
3. We ran the "test" call on every update, even if nothing had changed.

Fixes https://linear.app/danswer/issue/DAN-1912/fix-misc-model-configuration-stuff

## How Has This Been Tested?

Tested locally. In the future, might want to add e2e tests for these.

### DIFF
--- a/backend/onyx/server/manage/llm/models.py
+++ b/backend/onyx/server/manage/llm/models.py
@@ -28,7 +28,10 @@ class TestLLMRequest(BaseModel):
     fast_default_model_name: str | None = None
     deployment_name: str | None = None
 
-    model_configurations: list["ModelConfigurationUpsertRequest"] = []
+    model_configurations: list["ModelConfigurationUpsertRequest"]
+
+    # if try and use the existing API key
+    api_key_changed: bool
 
 
 class LLMProviderDescriptor(BaseModel):
--- a/web/src/app/admin/configuration/llm/CustomLLMProviderUpdateForm.tsx
+++ b/web/src/app/admin/configuration/llm/CustomLLMProviderUpdateForm.tsx
@@ -20,7 +20,7 @@ import {
 } from "@/components/admin/connectors/Field";
 import { useState } from "react";
 import { useSWRConfig } from "swr";
-import { LLMProviderView, ModelConfiguration } from "./interfaces";
+import { LLMProviderView } from "./interfaces";
 import { PopupSpec } from "@/components/admin/connectors/Popup";
 import * as Yup from "yup";
 import isEqual from "lodash/isEqual";
--- a/web/src/app/admin/configuration/llm/CustomLLMProviderUpdateForm.tsx
+++ b/web/src/app/admin/configuration/llm/CustomLLMProviderUpdateForm.tsx
@@ -20,7 +20,7 @@ import {
 } from "@/components/admin/connectors/Field";
 import { useState } from "react";
 import { useSWRConfig } from "swr";
-import { LLMProviderView } from "./interfaces";
+import { LLMProviderView, ModelConfiguration } from "./interfaces";
 import { PopupSpec } from "@/components/admin/connectors/Popup";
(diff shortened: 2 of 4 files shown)


backend/onyx/server/manage/llm/models.py
```diff
@@ -28,7 +28,10 @@ class TestLLMRequest(BaseModel):
     fast_default_model_name: str | None = None
     deployment_name: str | None = None
 
-    model_configurations: list["ModelConfigurationUpsertRequest"] = []
+    model_configurations: list["ModelConfigurationUpsertRequest"]
```
Weves, line 31: should never assign to `[]` or `{}` (or any object) in a class variable / function args. If the object is modified, then it iwll be modified for all future initializations as well.

web/src/app/admin/configuration/llm/CustomLLMProviderUpdateForm.tsx
```diff
@@ -114,15 +113,20 @@ export function CustomLLMProviderUpdateForm({
       onSubmit={async (values, { setSubmitting }) => {
         setSubmitting(true);
 
-        values.model_configurations.forEach((modelConfiguration) => {
```
Weves, line 117: prefer to not have functions modify external objects in place (unexpected side effects)

web/src/app/admin/configuration/llm/LLMProviderUpdateForm.tsx
```diff
@@ -75,7 +75,7 @@ export function LLMProviderUpdateForm({
       ),
     is_public: existingLlmProvider?.is_public ?? true,
     groups: existingLlmProvider?.groups ?? [],
-    model_configurations: [] as ModelConfiguration[],
```
Weves, line 78: model configs being set to empty list on update

backend/onyx/server/manage/llm/models.py
raunakab: Oh, good point.

Should we set it to `model_configurations: list[..] | None`?

web/src/app/admin/configuration/llm/LLMProviderUpdateForm.tsx
```diff
@@ -146,19 +145,19 @@ export function LLMProviderUpdateForm({
...
               max_input_tokens: null,
             }) as ModelConfiguration
         );
-
-        delete values.selected_model_names;
+        delete finalValues.selected_model_names;
+        finalValues.api_key_changed = values.api_key !== initialValues.api_key;
 
         // test the configuration
         if (!isEqual(values, initialValues)) {
```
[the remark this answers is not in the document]
raunakab: @Weves How about this comment? Seems valid.

raunakab (commented):
I think there's one small Greptile change that I think is important. Other than that, seems good to me.

Seems as if some of the misc mistakes in here are due to weird default initialization values that I set.

backend/onyx/server/manage/llm/models.py
Weves: mhmm, is there a reason why someone would pass in None?

raunakab (approved):

backend/onyx/server/manage/llm/models.py
raunakab: Potentially in APIs in which no changes to the model-configurations have been requested. Could be a way of signaling that no updates to the model-configurations table needs to be performed.

raunakab: Resolving for now, since this can be opened up in a new PR / discussion for later.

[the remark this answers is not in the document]
raunakab: Not fixing.

PR #4630, opened 2025-04-29
Bugfix/chat images 2

rkuo-danswer:
## Description
image

Fixes https://linear.app/danswer/issue/DAN-1923/fix-binary-data-in-filesourcecard

## How Has This Been Tested?

### DIFF
--- a/backend/onyx/chat/prune_and_merge.py
+++ b/backend/onyx/chat/prune_and_merge.py
@@ -155,6 +155,7 @@ def _apply_pruning(
 
     section_idx_token_count: dict[int, int] = {}
 
+    ind = 0
     final_section_ind = None
     total_tokens = 0
     for ind, section in enumerate(sections):
--- a/backend/onyx/db/chat.py
+++ b/backend/onyx/db/chat.py
@@ -870,7 +870,7 @@ def create_search_doc_from_user_file(
             content_sample = associated_chat_file.content[:100]
             # Remove null bytes which can cause SQL errors
             content_sample = content_sample.replace(b"\x00", b"")
-            blurb = content_sample.decode("utf-8", errors="replace")
+            blurb = content_sample.decode("utf-8", errors="strict")
         except Exception:
             # If decoding fails completely, provide a generic description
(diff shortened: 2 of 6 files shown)


backend/onyx/db/chat.py
```diff
@@ -870,7 +870,7 @@ def create_search_doc_from_user_file(
             content_sample = associated_chat_file.content[:100]
             # Remove null bytes which can cause SQL errors
             content_sample = content_sample.replace(b"\x00", b"")
-            blurb = content_sample.decode("utf-8", errors="replace")
+            blurb = content_sample.decode("utf-8", errors="strict")
```
rkuo-danswer, line 876: Main bug here.

backend/onyx/chat/prune_and_merge.py
```diff
@@ -155,6 +155,7 @@ def _apply_pruning(
 
     section_idx_token_count: dict[int, int] = {}
 
+    ind = 0
```
raunakab, line 158: Is this necessary?

backend/onyx/db/chat.py
raunakab: Nice catch.

raunakab (approved):
Just a small comment on why `ind = 0` was necessary in `prune_and_merge.py`.

Other than the main bug, looks like the changes are primarily the addition of docs, moving some variables around here and there, and some formatting changes.

backend/onyx/db/chat.py
Weves: @rkuo-danswer what was the problem with the before? Could we add a comment on why this is necessary 🧐

backend/onyx/chat/prune_and_merge.py
rkuo-danswer: yes, it's possible to not even enter the loop and for ind to be unbound in subsequent logging.

backend/onyx/db/chat.py
rkuo-danswer: this would not actually fail on binary content unless changed to strict. I can add a comment.

PR #4634, opened 2025-04-29
Small improvements to checkpoint pickup logic

Weves:
## Description

Fixes https://linear.app/danswer/issue/DAN-1916/better-checkpoint-handling-on-repeated-failures

## How Has This Been Tested?

👀 

### DIFF
--- a/backend/onyx/background/indexing/checkpointing_utils.py
+++ b/backend/onyx/background/indexing/checkpointing_utils.py
@@ -20,8 +20,7 @@ from onyx.utils.object_size_check import deep_getsizeof
 
 logger = setup_logger()
 
-_NUM_RECENT_ATTEMPTS_TO_CONSIDER = 20
-_NUM_DOCS_INDEXED_TO_BE_VALID_CHECKPOINT = 100
+_NUM_RECENT_ATTEMPTS_TO_CONSIDER = 50
 
 
 def _build_checkpoint_pointer(index_attempt_id: int) -> str:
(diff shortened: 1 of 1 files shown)


backend/onyx/background/indexing/checkpointing_utils.py
```diff
@@ -80,34 +79,56 @@ def get_latest_valid_checkpoint(
...
+                candidate.total_docs_indexed is not None
+                and candidate.total_docs_indexed > 0
+            ) or candidate.status.is_successful():
+                had_any_progress = True
+                break
+
+        if not had_any_progress:
+            logger.warning(
+                f"{_NUM_RECENT_ATTEMPTS_TO_CONSIDER} consecutive failed attempts without progress "
+                f"found for cc_pair={cc_pair_id}. Ignoring checkpoint to let the run start "
```
evan-onyx, line 99: personally I'd like putting more info here like the connector source type. Might save us 1 minute * the number of times we look for these logs in the future

evan-onyx (approved):

backend/onyx/background/indexing/checkpointing_utils.py
Weves: source type is not accessible w/o another db call or by passing more stuff into the function 😢

I generally agree with the sentiment, but I think this one is a pretty rare one that doesn't warrant an extra DB call / passing more stuff in

PR #4635, opened 2025-04-29
Fix race condition with archived channels

Weves:
## Description

Fixes https://linear.app/danswer/issue/DAN-1913/archived-channel-failures

## How Has This Been Tested?

Created channel, set breakpoint at `client.conversations_join` in `_get_messages`, then archived it. Verified the index attempt was a success.

### DIFF
--- a/backend/alembic/versions/7a70b7664e37_add_model_configuration_table.py
+++ b/backend/alembic/versions/7a70b7664e37_add_model_configuration_table.py
@@ -27,7 +27,7 @@ def _resolve(
     model_names: list[str] | None,
     display_model_names: list[str] | None,
     default_model_name: str,
-    fast_default_model_name: str,
+    fast_default_model_name: str | None,
 ) -> set[tuple[str, bool]]:
     models = set(model_names) if model_names else None
     display_models = set(display_model_names) if display_model_names else None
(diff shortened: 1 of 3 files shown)


backend/alembic/versions/7a70b7664e37_add_model_configuration_table.py
```diff
@@ -94,9 +94,11 @@ def _resolve(
     # It is possible that `default_model_name` is not in `models` and is not in `display_models`.
     # It is also possible that `fast_default_model_name` is not in `models` and is not in `display_models`.
     models.add(default_model_name)
-    models.add(fast_default_model_name)
```
Weves, line 97: this was causing migrations to fail for me for providers with null for fast default

backend/onyx/connectors/slack/utils.py
```diff
@@ -16,7 +16,8 @@
 
 logger = setup_logger()
 
-basic_retry_wrapper = retry_builder()
+# retry after 0.1, 1.2, 3.4, 7.8, 16.6, 34.2 seconds
+basic_retry_wrapper = retry_builder(tries=7)
```
Weves, line 20: old number of retries (20) is a bit excessive in a checkpointed world

evan-onyx (approved):

PR #4642, opened 2025-04-30
github cursor pagination

evan-onyx:
## Description

Fixes https://linear.app/danswer/issue/DAN-1915/github-pagination-errors
(hopefully)

We've been seeing weird cases of 
```
github.GithubException.GithubException: 422 {"message": "Pagination with the page parameter is not supported for large datasets, please use cursor based pagination (after/before)" ...
```

This is intended to fix the issue using pagination as described here: https://github.com/orgs/community/discussions/69826#cursor-based-pagination

The intended approach is:
Each indexing attempt will first (if we don't have stored cursor) try to use page-based, then if it fails (or a prev cursor stored) will try to use that cursor until an error occurs. If the cursor is expired we hit the nasty case of "just need to run through everything", but otherwise we return the next checkpoint after the first failure or index completion.

## How Has This Been Tested?

Unfortunately I couldn't reproduce the issue in local testing, so the best we have for now is automated tests TBD. Tested in the UI that the standard case (offset-based pagination) still works.

### DIFF
--- a/backend/onyx/connectors/github/connector.py
+++ b/backend/onyx/connectors/github/connector.py
@@ -1,5 +1,6 @@
 import copy
 import time
+from collections.abc import Callable
 from collections.abc import Generator
 from datetime import datetime
 from datetime import timedelta
@@ -33,12 +34,14 @@ from onyx.connectors.interfaces import SecondsSinceUnixEpoch
 from onyx.connectors.models import ConnectorMissingCredentialError
 from onyx.connectors.models import Document
 from onyx.connectors.models import DocumentFailure
+from onyx.connectors.models import EntityFailure
 from onyx.connectors.models import TextSection
 from onyx.utils.logger import setup_logger
 
 logger = setup_logger()
 
 ITEMS_PER_PAGE = 100
+CURSOR_LOG_FREQUENCY = 100
 
 _MAX_NUM_RATE_LIMIT_RETRIES = 5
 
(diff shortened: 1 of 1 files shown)


backend/onyx/connectors/github/connector.py
```diff
@@ -229,7 +357,10 @@ def _get_all_repos(
             # Try to get organization first
             try:
                 org = github_client.get_organization(self.repo_owner)
-                return list(org.get_repos())
+                # return list(org.get_repos())
+                return [x for x in list(org.get_repos()) if x.name == "ffe"]
```
Weves: should probably remove this one

backend/onyx/connectors/github/connector.py
```diff
@@ -52,26 +55,139 @@ def _sleep_after_rate_limit_exception(github_client: Github) -> None:
...
+
+
+# git_objs should produce a fresh PaginatedList each time it's called.
+# We don't want to use the same object for cursor-based pagination
+# after a partial offset-based pagination call. We expect the cursor-based approach
+# to repeat at most one page of work from the offset-based approach.
 def _get_batch_rate_limited(
-    git_objs: PaginatedList, page_num: int, github_client: Github, attempt_num: int = 0
-) -> list[PullRequest | Issue]:
+    git_objs: Callable[[], PaginatedList],
```
Weves, line 134: nit: add a comment as to why we pass in a func here?

backend/onyx/connectors/github/connector.py
```diff
@@ -342,37 +492,63 @@ def _fetch_from_github(
...
+            if num_prs > 0 and not done_with_prs and not cursor_mode_done_with_prs:
                 return checkpoint
 
             # if we went past the start date during the loop or there are no more
             # prs to get, we move on to issues
             checkpoint.stage = GithubConnectorStage.ISSUES
-            checkpoint.curr_page = 0
+            checkpoint.reset()
+
+            if cursor_mode_done_with_prs:
```
Weves: why do we only do this in this case?

backend/onyx/connectors/github/connector.py
```diff
@@ -52,26 +55,139 @@ def _sleep_after_rate_limit_exception(github_client: Github) -> None:
...
+
+            if num_objs % CURSOR_LOG_FREQUENCY == 0:
+                logger.info(
+                    f"Retrieved {num_objs} objects with current cursor url: {pag_list.__nextUrl}"
+                )
+
+    except Exception as e:
+        logger.exception(f"Error during cursor-based pagination: {e}")
+        if num_objs > 0:
+            yield ConnectorFailure(
```
Weves: this is the case where we fail midway through pagination, right? To be consistent, I think we should just throw an un-handled exception and end the attempt in an error. Currently, we'd just mark it as a success despite making it only some % of the way through (potentially very small %)?

backend/onyx/connectors/github/connector.py
evan-onyx: The way offset based works is we save the checkpoint whenever we're not done, then rely on the fact that github gives us an empty page to know when to move on to the next phase. if we returned the checkpoint here in that case as well, we would never make it past PR mode

Weves (approved):

PR #4653, opened 2025-05-02
Preliminary internet search using Exa

trial-danswer:
## Description

Internet search with Bing deprecated - changed to use Exa AI
Currently sends full webpages for summarization and can exceed token limit.

## How Has This Been Tested?


trial-danswer:
Latest change attempts to chunk, find most relevant, and prune results to not exceed context window. However, the tool responses are not being handled correctly and the entire internet search result is being sent to the LLM. This should work if the selected model is large enough, if the search results are not very long, or if num_results are set to a smaller value (defualt 10).

PR #4655, opened 2025-05-02
Fix non-default schema in KV store

Weves:
## Description

Previously, if you set the `POSTGRES_DEFAULT_SCHEMA` env variable, it wouldn't be respected in `PgRedisKVStore`, which causes issues.

Fixes https://linear.app/danswer/issue/DAN-1938/use-correct-schema-in-kv-store

## How Has This Been Tested?

Tested different schema locally.

### DIFF
--- a/backend/onyx/db/engine.py
+++ b/backend/onyx/db/engine.py
@@ -10,7 +10,7 @@ from contextlib import asynccontextmanager
 from contextlib import contextmanager
 from datetime import datetime
 from typing import Any
-from typing import ContextManager
+from typing import AsyncContextManager
 
 import asyncpg  # type: ignore
 import boto3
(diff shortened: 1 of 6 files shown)


rkuo-danswer (approved):

backend/onyx/db/engine.py
```diff
@@ -189,45 +190,6 @@ class SqlEngine:
     _lock: threading.Lock = threading.Lock()
     _app_name: str = POSTGRES_UNKNOWN_APP_NAME
 
-    # NOTE(rkuo) - this appears to be unused, clean it up?
```
Weves, line 192: not used, was moved into the `init_engine` function

backend/onyx/db/engine.py
```diff
@@ -430,33 +380,6 @@ def _set_search_path(
 )
 
 
-@asynccontextmanager
-async def get_async_session_with_tenant(
```
Weves, line 434: replaced with `get_async_session_context_manager`

PR #4656, opened 2025-05-02
id not set in checkpoint FINAL

evan-onyx:
## Description

It is not impossible for indexing to error because drive id is not set in checkpoint. I am convinced that these errors occur due to rare states where a user Just Finished their My Drive but has not started the next stage, meaning it is correct to simply continue on to the next stage when this happens.

## How Has This Been Tested?

n/a, this should only reduce errors.

### DIFF
--- a/backend/ee/onyx/external_permissions/google_drive/group_sync.py
+++ b/backend/ee/onyx/external_permissions/google_drive/group_sync.py
@@ -107,7 +107,7 @@ def _map_group_email_to_member_emails(
             admin_service.members().list,
             list_key="members",
             groupKey=group_email,
-            fields="members(email)",
+            fields="members(email),nextPageToken",
         ):
             group_member_emails.add(member["email"])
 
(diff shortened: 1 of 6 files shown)


backend/ee/onyx/external_permissions/google_drive/group_sync.py
```diff
@@ -51,7 +51,7 @@ def _get_drive_members(
                 drive_service.permissions().list,
                 list_key="permissions",
                 fileId=drive_id,
-                fields="permissions(emailAddress, type)",
+                fields="permissions(emailAddress, type),nextPageToken",
```
Weves, line 54: should we throw an exception if this isn't specified for a `execute_paginated_retrieval`?

Weves (approved):

backend/ee/onyx/external_permissions/google_drive/group_sync.py
evan-onyx: probably yeah

PR #4664, opened 2025-05-05
fix: Fix error in which channel names would not have the leading "#" removed

raunakab:
## Description

The prefixing "#" from channel names would not be removed prior to indexing channels. This would lead to an error in the check for whether or not that channel-name exists.

This PR removes all prefixed "#" during instantiation of the `SlackConnector` class.

Addresses: https://linear.app/danswer/issue/DAN-1941/melio-slack-issue.

## How Has This Been Tested?

Ran through manual indexing of a Slack channel with and without the leading "#".


backend/tests/daily/connectors/slack/test_slack_connector.py
```diff
@@ -0,0 +1,106 @@
...
+        ["doesnt-exist", "#doesnt-exist"],
+    ],
+)
+def test_indexing_channels_that_dont_exist(
+    slack_connector: SlackConnector,
+    channels: list[str],
+) -> None:
+    slack_connector.channels = channels
+    slim_docs_generator = slack_connector.retrieve_all_slim_documents()
+    while True:
```
Weves: this should be caught during the validation, so should we tset that instead? This isn't really a valid state/one that we should run into.

Also, for this type of test see https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions

backend/tests/daily/connectors/slack/test_slack_connector.py
```diff
@@ -0,0 +1,106 @@
...
+    ],
+)
+def test_indexing_channel(
+    slack_connector: SlackConnector,
+    channels: list[str],
+) -> None:
+    slack_connector.channels = channels
+    slim_docs_generator = slack_connector.retrieve_all_slim_documents()
+    for slim_docs in slim_docs_generator:
+        # just test to make sure that the generator steps through all slim-docs appropriately
```
Weves: can we verify correctness here in some way? E.g. a correct number of messages

backend/onyx/connectors/slack/connector.py
```diff
@@ -23,6 +23,7 @@
 from onyx.configs.app_configs import INDEX_BATCH_SIZE
 from onyx.configs.app_configs import SLACK_NUM_THREADS
 from onyx.configs.constants import DocumentSource
+from onyx.connectors.credentials_provider import OnyxStaticCredentialsProvider
```
Weves: if only used with `__main__` lets add the import there

backend/onyx/connectors/slack/connector.py
```diff
@@ -52,6 +53,7 @@
 from onyx.indexing.indexing_heartbeat import IndexingHeartbeatInterface
 from onyx.redis.redis_pool import get_redis_client
 from onyx.utils.logger import setup_logger
+from shared_configs.contextvars import get_current_tenant_id
```
Weves: same comment here

backend/tests/daily/connectors/slack/test_slack_connector.py
```diff
@@ -0,0 +1,148 @@
...
+                [
+                    "Hello, world!",
+                    "",
+                    "Testing again...",
+                ]
+            ),
+        ),
+    ],
+)
+def test_indexing_channels_with_message_count(
```
Weves, line 96: why are we testing `get_channels` rather than the actual connector?

backend/tests/daily/connectors/slack/test_slack_connector.py
```diff
@@ -0,0 +1,148 @@
...
+    ],
+)
+def test_indexing_channels_that_dont_exist(
+    slack_connector: SlackConnector,
+    channel_name: str,
+) -> None:
+    if not slack_connector.client:
+        raise RuntimeError("Web client must be defined")
+
+    slack_connector.channels = [channel_name]
```
Weves: same here

backend/tests/daily/connectors/slack/test_slack_connector.py
raunakab: Hmm, this was more so testing that the channel-name sanitization (i.e., removing the "#" prefix) works as intended.

backend/onyx/connectors/slack/connector.py
raunakab: Updated!

backend/onyx/connectors/slack/connector.py
raunakab: Updated!

backend/tests/daily/connectors/slack/test_slack_connector.py
raunakab: Updated to testing the retrieval of messages in the given channel.

backend/tests/daily/connectors/slack/test_slack_connector.py
raunakab: Maybe the location of this test isn't appropriate? Should it be placed in some other "slack-utils-testing" file?

raunakab: Updated to test connectors instead!

backend/tests/daily/connectors/slack/test_slack_connector.py
raunakab: Updated to test connectors instead!

backend/tests/daily/connectors/slack/test_slack_connector.py
```diff
@@ -0,0 +1,147 @@
...
+        "doesnt-exist",
+        # w/ hashtag
+        "#doesnt-exist",
+    ],
+)
+def test_indexing_channels_that_dont_exist(
+    slack_connector: SlackConnector,
+    channel_name: str,
+) -> None:
+    slack_connector.channels = [channel_name]
```
Weves: rather than setting it here, can we just pass it in to the init func? It's best to as closely as possible emulate the actual usage of the object

backend/tests/daily/connectors/slack/test_slack_connector.py
```diff
@@ -0,0 +1,147 @@
...
+            ),
+        ),
+    ],
+)
+def test_indexing_channels_with_message_count(
+    slack_connector: SlackConnector,
+    channel_name: str,
+    expected_messages: set[str],
+) -> None:
+    slack_connector.channels = [channel_name]
```
Weves: same comment as below

Weves (approved):

backend/tests/daily/connectors/slack/test_slack_connector.py
raunakab: Updated!

backend/tests/daily/connectors/slack/test_slack_connector.py
raunakab: Updated!

PR #4670, opened 2025-05-07
fixes foreign key violation

evan-onyx:
## Description

Fixes https://linear.app/danswer/issue/DAN-1946/foreign-key-violation-when-deleting-connector

We were seeing a foreign key violation during attempts at connector deletion When there were still user files associated with the connector.

## How Has This Been Tested?

n/a, seems pretty basic


backend/onyx/background/celery/tasks/connector_deletion/tasks.py
```diff
@@ -468,6 +471,12 @@ def monitor_connector_deletion_taskset(
             # related to the deleted DocumentByConnectorCredentialPair during commit
             db_session.expire(cc_pair)
 
+            # delete the userfile for the cc_pair
```
rkuo-danswer: maybe pluralize as userfiles?

rkuo-danswer (approved):

PR #4671, opened 2025-05-07
make 404s skippable

evan-onyx:
## Description

Fixes https://linear.app/danswer/issue/DAN-1947/drive-404s-on-file-retrieval
In some cases for read only files, retrievers get a 404 when trying to download. This PR lets us retry as other possible retrievers when this happens.

## How Has This Been Tested?

n/a, pretty basic. Ideally we could reproduce the 404s in our testing environment and add a connector test.


rkuo-danswer (approved):
sounds like a comment might be helpful here but looks good

PR #4677, opened 2025-05-08
Fix user count

Weves:
## Description

Fixes https://linear.app/danswer/issue/DAN-1953/fix-user-count-for-billing

## How Has This Been Tested?

👀 

### DIFF
--- a/backend/onyx/db/auth.py
+++ b/backend/onyx/db/auth.py
@@ -26,7 +26,7 @@ from onyx.utils.variable_functionality import (
     fetch_versioned_implementation_with_fallback,
 )
 
-T = TypeVar("T")
+T = TypeVar("T", bound=tuple[Any, ...])
 
 
 def get_default_admin_user_emails() -> list[str]:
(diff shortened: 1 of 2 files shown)


backend/onyx/db/auth.py
```diff
@@ -37,31 +40,44 @@ def get_default_admin_user_emails() -> list[str]:
...
+    users who should not be included in the live user count.
     """
-    user_count = (
-        db_session.query(User)
-        .filter(
-            ~User.email.endswith(get_api_key_email_pattern()),  # type: ignore
+    select_stmt = select_stmt.where(~User.email.endswith(get_api_key_email_pattern()))  # type: ignore
+    if only_admin_users:
+        return select_stmt.where(User.role == UserRole.ADMIN)
+    else:
```
rkuo-danswer: avoid else after return

rkuo-danswer (approved):

backend/onyx/db/auth.py
Weves: fixed!

PR #4679, opened 2025-05-08
Feature/helm separate workers

rkuo-danswer:
## Description

Fixes https://linear.app/danswer/issue/DAN-1956/update-helm-charts-to-separate-workers

## How Has This Been Tested?


deployment/helm/charts/onyx/values.yaml
```diff
@@ -296,58 +296,218 @@ api:
...
-    app: background
+    app: celery-beat
   podSecurityContext:
     {}
-    # fsGroup: 2000
   securityContext:
     privileged: true
     runAsUser: 0
-  enableMiniChunk: "true"
   resources: {}
```
Weves, line 376: is it best to leave all the resources as blank? Or do we have an idea of a reasonable resource limit/request we can put in based on our cloud?

Weves (approved):

deployment/helm/charts/onyx/values.yaml
rkuo-danswer: We could put something in. Was just defaulting to what was there.

PR #4701, opened 2025-05-13
Fix heavy import

Weves:
## Description

^

## How Has This Been Tested?

👀 


raunakab (approved):
Looks like another Hagen created bug...

PR #4703, opened 2025-05-14
Bugfix/usage report UUID

Orbital-Web:
## Description
[DAN-1747](https://linear.app/danswer/issue/DAN-1747/dont-use-random-uids-for-ui)

- Changed `create_new_usage_report` and `get_all_usage_reports` to return the username of the person who made the report instead of their user_id in the requestor field.
- Changed frontend to disable and gray out arrows when they shouldn't be active (e.g., left arrow disabled on 1st page, right arrow disabled on last page)

image

## How Has This Been Tested?

- Tested UI with 0, 1, 2, and 3 pages of reports
- Tested system generated and user generated reports

### DIFF
--- a/backend/ee/onyx/db/usage_export.py
+++ b/backend/ee/onyx/db/usage_export.py
@@ -5,6 +5,8 @@ from typing import IO
 from typing import Optional
 
 from fastapi_users_db_sqlalchemy import UUID_ID
+from sqlalchemy import cast
+from sqlalchemy.dialects.postgresql import UUID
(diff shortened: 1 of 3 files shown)


Orbital-Web:
Some extra things we can do:
Knowing the name of the reports is probably pretty important for calling the API get request, so I can't just replace it with the report.id. Though, maybe we can add a new report_id field to the UsageReportMetadata and display that instead in the usage report page table

Weves (approved):

PR #4719, opened 2025-05-15
Highlight active link in AdminSidebar based on current pathname

Subash-Mohan:
This pull request enhances the `AdminSidebar` component by introducing active link highlighting based on the current pathname. 

## How Has This Been Tested?
By looking into the UI.

Before:
Screenshot 2025-05-15 at 7 30 28 PM

After:
Screenshot 2025-05-15 at 7 29 12 PM


evan-onyx (approved):
Can you try to debug why the playwright tests are failing? Might be transient (just need to run them again) but would prefer that those pass before we merge this!

Subash-Mohan:
I looked into the Playwright test failure — it wasn’t transient. I made a small fix in the code, and now the tests are passing. Adding it to the merge queue

PR #4723, opened 2025-05-16
Fix OAuth w/ ext_perm_user for multi-tenant

Weves:
## Description

Fixes https://linear.app/danswer/issue/DAN-1981/fix-oauth-with-ext-perm-user-on-cloud

## How Has This Been Tested?

Tested w/ multi-tenant set up locally.

### DIFF
--- a/backend/onyx/auth/users.py
+++ b/backend/onyx/auth/users.py
@@ -463,15 +463,8 @@ class UserManager(UUIDIDMixin, BaseUserManager[User, uuid.UUID]):
                     }
 
                     user = await self.user_db.create(user_dict)
-
-                    # Add OAuth account only if user creation was successful
-                    if user is not None:
-                        await self.user_db.add_oauth_account(user, oauth_account_dict)
-                        await self.on_after_register(user, request)
-                    else:
-                        raise HTTPException(
-                            status_code=500, detail="Failed to create user account"
-                        )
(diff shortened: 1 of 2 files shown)


backend/onyx/auth/users.py
```diff
@@ -463,15 +463,8 @@ async def oauth_callback(
                     }
 
                     user = await self.user_db.create(user_dict)
-
-                    # Add OAuth account only if user creation was successful
-                    if user is not None:
```
Weves, line 468: user is never None

backend/onyx/auth/users.py
```diff
@@ -489,12 +482,6 @@ async def oauth_callback(
                                 oauth_account_dict,
                             )
 
-            # Ensure user is not None before proceeding
-            if user is None:
```
Weves, line 493: user is never None

backend/onyx/db/engine.py
```diff
@@ -498,6 +498,7 @@ async def get_async_session(
 
     async with AsyncSession(engine, expire_on_commit=False) as async_session:
         # set the search path on sync session as well to be extra safe
+        async_session.info["tenant_id"] = tenant_id
         event.listen(
```
Weves, line 503: seemingly the search_path gets reset after every commit on the async session?

Weves: not 100% sure, but that's what I was observing

PR #4729, opened 2025-05-18
perf: Change query-exporting to use generators instead of expanding fully into memory

raunakab:
## Description

Query history exporting fetches the entire history and populates it into memory. This causes significant pressure on memory, and can lead to OOMs for very large datasets.

This PR updates the logic to incrementally fetch pages of data, transform them, and write them to file, instead of fetching the entire dataset at once, transforming each row, and then writing it to file.

I.e., we don't fully collect / materialize the entire dataset, we do it in pages instead.

This PR also parallelizes the chat-session fetching logic (given that chat-session reading does not have any cross-thread dependencies [I think?]).

Addresses: https://linear.app/danswer/issue/DAN-1984/improve-performance-of-query-history-exporting.

## How Has This Been Tested?

This is performance change, not a feature-addition. We don't have much (if any) performance testing / benchmarking in our code. Therefore, this is not tested.

The feature still outputs the same output for the same input.

### DIFF
--- a/backend/ee/onyx/background/celery/apps/heavy.py
+++ b/backend/ee/onyx/background/celery/apps/heavy.py
@@ -9,7 +9,6 @@ from celery import Task
 from ee.onyx.background.task_name_builders import query_history_task_name
 from ee.onyx.server.query_history.api import fetch_and_process_chat_session_history
 from ee.onyx.server.query_history.api import ONYX_ANONYMIZED_EMAIL
-from ee.onyx.server.query_history.models import ChatSessionSnapshot
 from ee.onyx.server.query_history.models import QuestionAnswerPairSnapshot
 from onyx.background.celery.apps.heavy import celery_app
 from onyx.background.task_utils import construct_query_history_report_name
@@ -45,6 +44,13 @@ def export_query_history_task(self: Task, *, start: datetime, end: datetime) -> 
     task_id = self.request.id
     start_time = datetime.now(tz=timezone.utc)
 
+    stream = io.StringIO()
+    writer = csv.DictWriter(
+        stream,
+        fieldnames=list(QuestionAnswerPairSnapshot.model_fields.keys()),
+    )
+    writer.writeheader()
+
     with get_session_with_current_tenant() as db_session:
         try:
             register_task(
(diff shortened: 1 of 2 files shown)


backend/ee/onyx/server/query_history/api.py
```diff
@@ -61,41 +62,52 @@ def ensure_query_history_is_enabled(
...
-            )
-        ]
+        for snapshot in paged_snapshots:
+            if snapshot:
+                yield snapshot
+
+        # If we've fetched *less* than a `PAGE_SIZE` worth
+        # of data, we have reached the end of the
+        # pagination sequence; break.
+        if len(paged_chat_sessions) < PAGE_SIZE:
```
rkuo-danswer, line 110: what happens if we get exactly a PAGE_SIZE worth of sessions and nothing is left to paginate?

backend/ee/onyx/background/celery/apps/heavy.py
```diff
@@ -45,6 +44,13 @@ def export_query_history_task(self: Task, *, start: datetime, end: datetime) ->
     task_id = self.request.id
     start_time = datetime.now(tz=timezone.utc)
 
+    stream = io.StringIO()
```
rkuo-danswer, line 47: any tests around this task?

backend/ee/onyx/server/query_history/api.py
raunakab: We should re-enter the loop and perform another invocation of `get_page_of_chat_sessions`. That should return an empty list, which should lead to `paged_snapshots` being empty, which should lead back to this condition, but this time it will be evaluated to `False`.

raunakab: I actually just added an early, conditional `break` after the invocation of `get_page_of_chat_sessions`, which should break us out if `paged_chat_sessions` is empty. This will avoid us having to call `parallel_yield` (which then avoids having to unnecessarily allocate a `ThreadPoolExecutor` instance).

backend/ee/onyx/background/celery/apps/heavy.py
raunakab: We have tests in `backend/tests/integration/tests/query_history`.

rkuo-danswer (approved):

PR #4731, opened 2025-05-19
fix: use ImplementationSpecific for pathType on ingress-api

ciaransweet:
## Description

When using the Helm Charts, I observed:

```
Error: 1 error occurred:
        * admission webhook "validate.nginx.ingress.kubernetes.io" denied the request: ingress contains invalid paths: path /api(/|$)(.*) cannot be used with pathType Prefix
```

Which caused Helm to fail. Changing `pathType` from `Prefix` to `ImplementationSpecific` fixed this in `ingress-api.yml`

## How Has This Been Tested?

Applied to my helm deployment and it worked.

### DIFF
--- a/deployment/helm/charts/onyx/templates/ingress-api.yaml
+++ b/deployment/helm/charts/onyx/templates/ingress-api.yaml
@@ -14,7 +14,7 @@ spec:
       http:
         paths:
           - path: /api(/|$)(.*)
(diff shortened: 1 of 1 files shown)


ciaransweet:
@rkuo-danswer maybe you're interested?

@wenxi-onyx maybe? 😇

PR #4744, opened 2025-05-20
fix: Add back Teams' replies processing

raunakab:
## Description

Previous PR left out the processing of replies to a message. This PR adds it back.

Addresses: https://linear.app/danswer/issue/DAN-1997/teams-replies-not-being-parsed-and-processed-in-document.

## How Has This Been Tested?

Tests were written in the previous PR. For this one, since replies are included, they have been updated to include the replies in the expected output.

### DIFF
--- a/backend/onyx/connectors/teams/connector.py
+++ b/backend/onyx/connectors/teams/connector.py
@@ -485,6 +485,9 @@ def _collect_documents_for_channel(
     ).execute_query()
 
     for message in message_collection:
+        if not message.id:
+            continue
+
         if not _should_process_message(message=message, start=start, end=end):
             continue
 
@@ -504,7 +507,7 @@ def _collect_documents_for_channel(
         except Exception as e:
             yield ConnectorFailure(
                 failed_entity=EntityFailure(
-                    entity_id="messages",
+                    entity_id=message.id,
                 ),
                 failure_message=f"Retrieval of message and its replies failed; {channel.id=} {message.id}",
                 exception=e,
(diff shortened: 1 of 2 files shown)


backend/tests/daily/connectors/teams/test_teams_connector.py
```diff
@@ -48,14 +48,24 @@ def teams_connector(
 @pytest.mark.parametrize(
     "teams_connector,expected_messages",
     [
-        [["Onyx-Testing"], set(["This is the first message in Onyx-Testing ..."])],
+        [
+            ["Onyx-Testing"],
+            set(
+                [
```
evan-onyx, line 54: Could you add/change a test that checks whether replies are picked up?

raunakab: Some of those tests do test for replies. The replies are embedded in the document (after the "root" message).

backend/onyx/connectors/teams/connector.py
```diff
@@ -483,19 +484,34 @@ def _collect_document_for_channel_id(
...
+            # I.e., we don't convert each individual message and each individual reply into their own individual `Document`s.
+            if doc := _convert_thread_to_document(
+                channel=channel,
+                thread=thread,
+            ):
+                yield doc
+        except Exception as e:
+            yield ConnectorFailure(
+                failed_entity=EntityFailure(
+                    entity_id="messages",
```
Weves: this should be the ID of the root message ideally

Weves: see slack connector for an example

raunakab: Ah okay, sounds good.

raunakab: Updated.

rkuo-danswer (approved):

PR #4767, opened 2025-05-23
Fix/add back search with files

Weves:
## Description

Fixes https://linear.app/danswer/issue/DAN-2000/add-back-upload-file-search

## How Has This Been Tested?

Tested running searches both with files dragged into the chat UI + when uploading via the FilePicker.

### DIFF
--- a/web/src/app/chat/input/ChatInputBar.tsx
+++ b/web/src/app/chat/input/ChatInputBar.tsx
@@ -444,6 +444,45 @@ export function ChatInputBar({
     }
   };
 
+  // Combine selectedFiles and currentMessageFiles for unified rendering
+  const allFiles = useMemo(() => {
+    const combined: Array<{
+      id: string;
+      name: string;
+      chatFileType: ChatFileType;
+      isUploading?: boolean;
+      source: "selected" | "current";
+      originalFile: any;
+    }> = [];
+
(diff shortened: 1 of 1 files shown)


web/src/app/chat/lib.tsx
```diff
@@ -202,7 +201,6 @@ export async function* sendMessage({
   useExistingUserMessage,
   alternateAssistantId,
   signal,
-  forceUserFileSearch,
```
Weves, line 205: not used

web/src/app/chat/ChatPage.tsx
```diff
@@ -1956,10 +1965,7 @@ export function ChatPage({
     }
   };
 
-  const handleImageUpload = async (
-    acceptedFiles: File[],
-    intent: UploadIntent
```
Weves, line 1961: we weren't using, so removed for now

backend/onyx/tools/tool_implementations/search_like_tool_utils.py
```diff
@@ -45,12 +42,8 @@ def build_next_prompt_for_search_like_tool(
         build_citations_user_message(
             # make sure to use the original user query here in order to avoid duplication
             # of the task prompt
-            message=HumanMessage(
```
Weves, line 48: prompt was a bit weird in the case where we had user files

PR #4771, opened 2025-05-24
connector filter bug fix

Subash-Mohan:
This pull request has updated logic for Status column to display the scheduled badge and changes for connector filtering.

## How Has This Been Tested?
Tested in UI

### DIFF
--- a/web/src/app/admin/indexing/status/CCPairIndexingStatusTable.tsx
+++ b/web/src/app/admin/indexing/status/CCPairIndexingStatusTable.tsx
@@ -156,7 +156,13 @@ border border-border dark:border-neutral-700
       </TableCell>
       <TableCell>
         <CCPairStatus
-          ccPairStatus={ccPairsIndexingStatus.cc_pair_status}
+          ccPairStatus={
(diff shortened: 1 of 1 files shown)


web/src/app/admin/indexing/status/CCPairIndexingStatusTable.tsx
```diff
@@ -322,11 +328,7 @@ export function CCPairIndexingStatusTable({
 
         // Filter by last status
         if (filterOptions.lastStatus && filterOptions.lastStatus.length > 0) {
-          if (
-            !filterOptions.lastStatus.includes(
-              status.last_finished_status as any
-            )
-          ) {
+          if (!filterOptions.lastStatus.includes(status.last_status as any)) {
```
Weves: why do we need to do `status.last_status as any`? that feels wrong

Subash-Mohan: There is no specific reason for that, now i changed the type.

Subash-Mohan:
Thanks @evan-onyx

PR #4780, opened 2025-05-28
fix(discourse-connector): handle redirect issue 

suvodhoy:
## Description

When accessing a Discourse category via `<baseUrl>/c/<categoryId>.json?page={pageNumber}&sys=latest`, Discourse issues a redirect to `<baseUrl>/c/<categorySlug>/<categoryId>.json?....`. During this redirect, any numeric occurrence matching the `categoryId` in the query string is also rewritten - including the page parameter - causing incorrect behavior (e.g., `page=5` becoming `page=community/5`).

This PR updates the connector logic to use the correct category url to avoid the redirection ensuring that the query parameters remain intact.

## How to reproduce?

Try opening this link https://meta.discourse.org/c/10.json?page=10&sys=latest in the browser.  It will throw a 400 Bad Request error. When we inspect the url, we will notice the page number has been replaced by the `slug/categoryId`.

## How Has This Been Tested?

The issue was found when I was trying to sync my own discourse setup. I tested the changes by building the `background` service locally and running the sync.

### DIFF
--- a/backend/onyx/connectors/discourse/connector.py
+++ b/backend/onyx/connectors/discourse/connector.py
@@ -60,7 +60,7 @@ class DiscourseConnector(PollConnector):
         self.base_url = base_url
 
         self.categories = [c.lower() for c in categories] if categories else []
-        self.category_id_map: dict[int, str] = {}
+        self.category_id_map: dict[int, dict] = {}
 
         self.batch_size = batch_size
         self.permissions: DiscoursePerms | None = None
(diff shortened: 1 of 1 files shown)


suvodhoy:
Raised a [fix](https://github.com/discourse/discourse/pull/33009) in the `discourse` repo as well.

suvodhoy:
Hi @Weves, can you review this PR?

wenxi-onyx:
@suvodhoy Good catch on this one! Since this is fixed in Discourse, is this PR still applicable for Onyx?

suvodhoy:
Hi @wenxi-onyx, although this issue will be fixed in the latest versions of Discourse, some users may still be on older versions where the bug exists. As a result, they could continue facing issues during data sync. Hence I'll recommend handling this in Onyx as well to ensure broader compatibility.

PR #4783, opened 2025-05-28
fix: Remove "Refining Answer" popup

raunakab:
## Description

The "Refining Answer" would always pop up for previous *already answered* queries. This PR removes that code path altogether.

Addresses: https://linear.app/danswer/issue/DAN-1886/refining-answer-disappears-then-comes-back-before-finally-disappearing.

## How Has This Been Tested?

This is a pure UI change; no tests were written.


web/src/app/chat/message/AgenticMessage.tsx
```diff
@@ -214,18 +204,15 @@ export const AgenticMessage = ({
     }
   }, [hasThinkingTokens]);
 
-  const [isViewingInitialAnswer, setIsViewingInitialAnswer] = useState(true);
+  const isViewingInitialAnswer = true;
```
Orbital-Web, line 207: I think we can remove this entirely if we are hardcoding it to true, alongside places where isViewingInitialAnswer is used.

Orbital-Web (approved):
lgtm, added a few nits

web/src/app/chat/message/AgenticMessage.tsx
raunakab: Good point! Will update.

PR #4788, opened 2025-05-29
drive perm sync logs + misc deployment improvements

evan-onyx:
## Description

Fixes https://linear.app/danswer/issue/DAN-2027/drive-perm-sync-logging-and-db-shm-size
A bunch of small logging and fixes for google drive connector/permission syncing

Also some memory bumps and a vespa version upgrade that helps with queries containing many common words

## How Has This Been Tested?

most of it has been tested in UI, although the helm changes haven't been tested yet

### DIFF
--- a/backend/onyx/connectors/google_drive/doc_conversion.py
+++ b/backend/onyx/connectors/google_drive/doc_conversion.py
@@ -62,6 +62,10 @@ GOOGLE_MIME_TYPES = {
 }
 
 
+def onyx_document_id_from_drive_file(file: GoogleDriveFileType) -> str:
+    return file[WEB_VIEW_LINK_KEY]
+
+
 def _summarize_drive_image(
     image_data: bytes, image_name: str, image_analysis_llm: LLM | None
 ) -> str:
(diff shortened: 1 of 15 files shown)


backend/onyx/file_processing/extract_file_text.py
```diff
@@ -310,7 +310,12 @@ def docx_to_text_and_images(
     paragraphs = []
     embedded_images: list[tuple[bytes, str]] = []
 
-    doc = docx.Document(file)
+    try:
+        doc = docx.Document(file)
+    except BadZipFile as e:
+        error_str = f"Failed to extract text from {file_name or 'docx file'}: {e}"
```
rkuo-danswer: what's the point of separating out the error str here?

backend/onyx/connectors/google_drive/doc_conversion.py
```diff
@@ -62,6 +62,10 @@
 }
 
 
+def onyx_document_id_from_drive_file(file: GoogleDriveFileType) -> str:
```
rkuo-danswer, line 65: this seems to do too little for a function

rkuo-danswer (approved):

backend/onyx/connectors/google_drive/doc_conversion.py
evan-onyx: the point was to ensure that this is consistent everywhere we get the document id from the drive file; I forgot to replace the other places we were doing that in doc_conversion. Thanks!

PR #4790, opened 2025-05-29
feat: Knowledge graph full-stack implementation

raunakab:
## Description

This PR contains the frontend + backend code for the knowledge graph.

Addresses: https://linear.app/danswer/issue/DAN-2017/knowledge-graph-full-stack-implementation.

## How Has This Been Tested?

Frontend: Todo!
Backend: Tests have been added to `backend/tests/integration/tests/kg/test_kg.py`.

## Screenshots

This is the main configuration page (when KG is enabled):

image

This is the main KG configuration modal:

image


backend/onyx/context/search/models.py
```diff
@@ -121,6 +121,11 @@ class BaseFilters(BaseModel):
 class UserFileFilters(BaseModel):
     user_file_ids: list[int] | None = None
     user_folder_ids: list[int] | None = None
+    kg_entities: list[str] | None = None
```
Orbital-Web: You can remove these lines. This is an error on git's side

backend/onyx/db/kg_config.py
```diff
@@ -141,3 +150,153 @@ def get_kg_processing_in_progress_status(
         return False
 
     return config.kg_variable_values[0] == "true"
+
+
+# server API helpers
+
+
+VALID_ENTITY_TYPE_NAMES = set(
```
Orbital-Web: I think this should be a db query, instead of a hardcoded list. 
Or you can pull from KGDefaultPrimaryGroundedEntityDefinitions for the time beings as the entity types will be set by us

backend/onyx/db/kg_config.py
```diff
@@ -141,3 +150,153 @@ def get_kg_processing_in_progress_status(
...
+
+
+def enable_kg(
+    db_session: Session,
+    enable_req: EnableKGConfigRequest,
+) -> None:
+    # cannot be empty string
+    if not enable_req.vendor:
+        raise ValueError(
+            f"KG vendor must be specified; instead got {enable_req.vendor=}"
```
Orbital-Web: nit: might be nice to wrap this in '' so we get `instead got ''` rather than `instead got`

backend/onyx/db/kg_config.py
```diff
@@ -141,3 +150,153 @@ def get_kg_processing_in_progress_status(
...
+        )
+
+    # cannot be empty list
+    if not enable_req.vendor_domains:
```
Orbital-Web: nit: same here

backend/onyx/document_index/vespa/shared_utils/vespa_request_builders.py
```diff
@@ -113,7 +113,9 @@ def _build_time_filter(
 
     # TODO: add error condition if MULTI_TENANT and no tenant_id filter is set
```
Orbital-Web: nit: we can remove this comment

web/src/app/admin/kg/page.tsx
```diff
@@ -0,0 +1,420 @@
...
+      header: "Name",
+    },
+    {
+      accessorKey: "description",
+      header: "Description",
+      cell: ({ row }) => (
+        <div className="h-20 w-[800px]">
+          <TextAreaField
+            name={`${row.original.name.toLowerCase()}.description`}
+            className="resize-none"
```
Orbital-Web: Would prefer `className="resize-none border rounded-md bg-background text-text focus:ring-2 focus:ring-blue-500 transition duration-200"` to keep consistent with other pages

image

backend/onyx/db/kg_config.py
```diff
@@ -150,3 +188,222 @@ def get_kg_processing_in_progress_status(
...
+            Connector.source.in_(
+                [
+                    source
+                    for source in DocumentSource
+                    if source.value.lower() in active_entity_type_sources
+                ]
+            )
+        )
+        .where(~Connector.kg_processing_enabled)
+        .values(kg_processing_enabled=True)
```
Orbital-Web: Let's also set coverage days of the connector to a higher number here (18000)

backend/onyx/db/kg_config.py
```diff
@@ -150,3 +188,222 @@ def get_kg_processing_in_progress_status(
...
+            kg_variable_values=enable_req.vendor_domains,
+        ),
+        KGConfig(
+            kg_variable_name=KGConfigVars.KG_IGNORE_EMAIL_DOMAINS,
+            kg_variable_values=enable_req.ignore_domains,
+        ),
+        KGConfig(
+            kg_variable_name=KGConfigVars.KG_COVERAGE_START,
+            kg_variable_values=[enable_req.coverage_start.strftime("%Y-%m-%d")],
+        ),
```
Orbital-Web, line 233: For now, can you also set the max coverage days to a high number (say, 18000 days). That way, the coverage start day will always be used (if the max coverage day is shorter, it could miss older documents)

Orbital-Web (commented):
more stuff

web/src/app/admin/kg/page.tsx
```diff
@@ -0,0 +1,495 @@
...
+  refreshKGEntityTypes,
+}: {
+  kgEntityTypes: EntityTypeValues;
+  sortedKGEntityTypes: EntityType[];
+  setPopup?: (spec: PopupSpec | null) => void;
+  refreshKGEntityTypes?: () => void;
+}) {
+  const columns: ColumnDef<EntityType>[] = [
+    {
+      accessorKey: "name",
```
Orbital-Web, line 243: nit but can we make the name column slightly wider

backend/onyx/db/kg_config.py
Orbital-Web: In fact, having this as a query is necessary as otherwise the types are wrong (gets failed to configure entity table)

backend/onyx/db/kg_config.py
```diff
@@ -150,3 +188,222 @@ def get_kg_processing_in_progress_status(
...
+    ]
+
```
Orbital-Web, line 239: Also, I guess if we can make a "KG Dev" assistant if it doesn't exist already, that would be great. 
Probably don't have to bother with removing it on disable

backend/onyx/db/kg_config.py
```diff
@@ -1,34 +1,106 @@
...
-def get_kg_enablement(db_session: Session) -> bool:
-    check = (
+def _get_connector_sources(db_session: Session) -> set[str]:
+    return {
+        source.value.lower()
+        for source in fetch_unique_document_sources(db_session=db_session)
+    }
+
+
+def _get_boolean_value(
```
evan-onyx: this name should be more kg specific like `_get_kg_config_boolean`

backend/onyx/db/kg_config.py
```diff
@@ -80,6 +152,14 @@ def get_kg_config_settings(db_session: Session) -> KGConfigSettings:
             )
         elif result.kg_variable_name == KGConfigVars.KG_EXPOSED:
             kg_config_settings.KG_EXPOSED = result.kg_variable_values[0] == "true"
+        elif result.kg_variable_name == KGConfigVars.KG_BETA_PERSONA_ID:
+            if result.kg_variable_values and result.kg_variable_values[0]:
+                try:
+                    kg_config_settings.KG_BETA_PERSONA_ID = int(
+                        result.kg_variable_values[0]
+                    )
+                except ValueError:
```
evan-onyx: leave a comment explaining why we fail silently here

backend/onyx/db/kg_config.py
```diff
@@ -150,3 +230,194 @@ def get_kg_processing_in_progress_status(
...
+            KG_VENDOR_DOMAINS=enable_req.vendor_domains,
+            KG_IGNORE_EMAIL_DOMAINS=enable_req.ignore_domains,
+            KG_COVERAGE_START=enable_req.coverage_start,
+        )
+    )
+
+    vars = [
+        KGConfig(
+            kg_variable_name=KGConfigVars.KG_ENABLED,
+            kg_variable_values=[bool_to_string(True)],
```
evan-onyx: nit: just use "true" directly, or better yet define something like KG_VARIABLE_TRUE = 'true' and use that

backend/onyx/db/kg_config.py
```diff
@@ -150,3 +230,194 @@ def get_kg_processing_in_progress_status(
...
+
+
+def disable_kg(db_session: Session) -> None:
+    var = (
+        db_session.query(KGConfig)
+        .filter(KGConfig.kg_variable_name == KGConfigVars.KG_ENABLED)
+        .first()
+    )
+
+    values = [bool_to_string(False)]
```
evan-onyx: same as above; people will think we're weird for having a function that computes `str(b).lower()` that's only used once for each case haha

backend/onyx/db/kg_config.py
```diff
@@ -150,3 +230,194 @@ def get_kg_processing_in_progress_status(
...
+            EntityType.from_model(kg_entity_type)
+            for kg_entity_type in populate_default_entity_types(
+                db_session=db_session,
+            )
+        ]
+
+    configured_connector_sources = _get_connector_sources(db_session=db_session)
+
+    available_entity_types = [
+        et
```
evan-onyx: et tu, brutus?

backend/onyx/kg/setup/kg_default_entity_definitions.py
```diff
@@ -330,6 +276,10 @@ def populate_default_entity_types(
     Returns the *entire* list of Entity Types.
     """
 
+    # Imported here to get rid of circular import.
+    from onyx.db.kg_config import get_kg_config_settings
+    from onyx.db.kg_config import validate_kg_settings
```
evan-onyx: why are these in the function? If there's a circular import otherwise, that indicates a more fundamental problem

backend/onyx/db/kg_config.py
```diff
@@ -150,3 +230,194 @@ def get_kg_processing_in_progress_status(
...
+        for et in existing_entity_types
+        if et.grounded_source_name in configured_connector_sources
+    ]
+    return available_entity_types
+
+
+def update_kg_entity_types(
```
evan-onyx: I believe the convention for db functions is to add __commit or __no_commit to the end based on their behavior; would recommend following this for all the db functions in the PR

backend/onyx/db/kg_config.py
```diff
@@ -150,3 +230,194 @@ def get_kg_processing_in_progress_status(
...
+        db_session.execute(
+            update(KGEntityType)
+            .where(KGEntityType.id_name == upd.name)
+            .values(
+                description=upd.description,
+                active=upd.active,
+            )
+        )
+
+    db_session.commit()
```
evan-onyx: this should be a flush()

backend/onyx/db/kg_config.py
```diff
@@ -150,3 +230,194 @@ def get_kg_processing_in_progress_status(
...
+        )
+        .where(Connector.kg_processing_enabled)
+        .values(kg_processing_enabled=False)
+    )
+
+    db_session.commit()
+
+
+def bool_to_string(b: bool) -> str:
+    return "true" if b else "false"
```
evan-onyx: remove bool_to_string

backend/onyx/server/kg/api.py
```diff
@@ -0,0 +1,206 @@
...
+# global resets
+
+
+@admin_router.put("/reset")
+def reset_kg(
+    _: User | None = Depends(current_admin_user),
+    db_session: Session = Depends(get_session),
+) -> list[EntityType]:
+    # reset all entity types to inactive
+    # TODO: before merging, convert to celery task function in other PR
```
evan-onyx: this todo sounds good!

backend/onyx/server/kg/api.py
```diff
@@ -0,0 +1,206 @@
...
+    if isinstance(req, EnableKGConfigRequest):
+        enable_req = req if req.enabled else None
+    elif isinstance(req, DisableKGConfigRequest):
+        if req.enabled:
+            raise ValueError("Cannot update KG Config with only `enabled: true`")
+        enable_req = None
+    else:
+        raise ValueError("Invalid request body")
+
+    if enable_req:
```
evan-onyx:
```
if not enable_req:
   <current body of else>
   return

<the current body of this if>
```

backend/onyx/server/kg/api.py
```diff
@@ -0,0 +1,206 @@
...
+        kg_config.enable_kg(db_session=db_session, enable_req=enable_req)
+        # Create or restore KG Beta persona
+        # Create prompt for KG Beta
+        prompt = upsert_prompt(
+            db_session=db_session,
+            user=None,
+            name=build_prompt_name_from_persona_name("KG Beta"),
+            system_prompt=(
+                "You are a knowledge graph assistant that helps users explore and "
```
evan-onyx: I think these should go in a prompts file

backend/onyx/server/kg/api.py
```diff
@@ -0,0 +1,206 @@
...
+                    )
+                    if persona.deleted:
+                        mark_persona_as_not_deleted(
+                            persona_id=persona_id,
+                            user=None,
+                            db_session=db_session,
+                        )
+                    return
+                except ValueError:
+                    # If persona doesn't exist or can't be restored, create a new one
```
evan-onyx: If you're wrapping the whole thing with a try: except Exception (which I am still Not A Fan Of), then no reason to try and except ValueError. Even better would be to keep the inner try-except and remove the outer one, but Definitely should not have both unless you really want to add some logging or something like that specifically for ValueError. Even in that case, the two "except"s should be attached to a single try with ValueError coming first

backend/onyx/server/kg/api.py
```diff
@@ -0,0 +1,206 @@
...
+            name="KG Beta",
+            description=(
+                "The KG Beta assistant uses the Onyx Knowledge Graph (beta) structure "
+                "to answer questions"
+            ),
+            system_prompt="Use the Onyx Knowledge Graph (beta) to answer questions.",
+            task_prompt="",
+            datetime_aware=False,
+            include_citations=True,
+            num_chunks=25,
```
evan-onyx: should this be a constant somewhere?

backend/onyx/server/kg/models.py
```diff
@@ -0,0 +1,67 @@
...
+        )
+
+
+class EnableKGConfigRequest(BaseModel):
+    enabled: bool
+    vendor: str
+    vendor_domains: list[str]
+    ignore_domains: list[str] = []
+    coverage_start: datetime
+
```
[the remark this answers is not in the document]
evan-onyx: I agree with greptile

backend/scripts/dev_run_background_jobs.py
```diff
@@ -137,6 +150,13 @@ def run_jobs() -> None:
         text=True,
     )
 
+    worker_kg_processing_process = subprocess.Popen(
```
evan-onyx, line 153: I know we use another SimpleJob class to handle process spawning elsewhere; we should probably use that instead unless there's a good reason not to. If there is a good reason, comment it here

backend/onyx/server/kg/api.py
joachim-danswer: This has been done in a PR that is merged. Will remove comment.

backend/onyx/server/kg/api.py
joachim-danswer: Done

backend/onyx/db/kg_config.py
joachim-danswer: Done. We probably should remove though as I believe not used. @raunakab ?

backend/onyx/db/kg_config.py
joachim-danswer: Done. Rewritten w/o try/except. 
value = result.kg_variable_values[0] if result.kg_variable_values else None
kg_config_settings.KG_BETA_PERSONA_ID = int(value) if value and str(value).isdigit() else None

Var is int or None in model. Stored as str (in list) in db.

backend/onyx/db/kg_config.py
joachim-danswer: @raunakab - to add, a function with this name appears to be defined twice, once in api.py and also in kg_config.py

backend/onyx/server/kg/api.py
joachim-danswer: Double-layer has been removed. Just one try/except. That is needed because the persona may not be returned. And if it is not, the function raises a ValueError.

backend/onyx/server/kg/api.py
joachim-danswer: Done

backend/onyx/server/kg/models.py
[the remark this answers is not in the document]
joachim-danswer: @raunakab - this one is critical.

backend/onyx/db/kg_config.py
raunakab: Seems that we're not using it anymore and has been removed.

backend/onyx/kg/setup/kg_default_entity_definitions.py
raunakab: `kg_config.py` defines `get_kg_config_settings` and `validate_kg_settings` which are used by `populate_default_entity_types`, and `populate_default_entity_types` is used by *another* function in `kg_config.py`.

It's a pretty classic Python (+ dynamic-importing language) problem, unfortunately.

We could pull out the shared functions into a new file, but that would be another file to juggle contextually, which I'm not the biggest fan of, to be honest.

backend/onyx/db/kg_config.py
raunakab: The way I was calling the one in `kg_config.py` was by importing `kg_config` and calling the symbol inside of it via the dot-operator. I.e.:
```py
import onyx.db.kg_config

kg_config.update_kg_entity_types
```

Wasn't aware of the convention that Evan mentioned, but I'm all for it.

raunakab: Believe @Orbital-Web was able to make this change.

backend/onyx/server/kg/api.py
```diff
@@ -0,0 +1,189 @@
...
+
+    # Create KG Beta persona
+    persona_request = PersonaUpsertRequest(
+        name="KG Beta",
+        description=_KG_BETA_ASSISTANT_DESCRIPTION,
+        system_prompt=KG_BETA_ASSISTANT_SYSTEM_PROMPT,
+        task_prompt=KG_BETA_ASSISTANT_TASK_PROMPT,
+        datetime_aware=False,
+        include_citations=True,
+        num_chunks=25,
```
evan-onyx, line 135: This should be a constant

evan-onyx (approved):
one nit, otherwise looks good

Issue #4795, opened 2025-05-30
OAuth Error: invalid redirect_uri returned by Linear before user authorization


pavelt72:
### Description

When attempting to add a new **Linear connector**, the OAuth flow fails **before the user is able to authorize Onyx in Linear**. The following error is returned immediately: invalid redirect_uri

### 🔁 Steps to Reproduce

1. Navigate to the **Connectors** page.
2. Click **“Add Linear Connector”**.
3. Attempt to authorize Onyx in the Linear OAuth window.
4. Observe the error.

### ✅ Expected Behavior

The user should be redirected to Linear's authorization screen and be able to authorize the app successfully.

### ❌ Actual Behavior

The user is immediately shown an `invalid redirect_uri` error from Linear, which halts the OAuth process before authorization.

### 💡 Possible Cause

Linear may be rejecting the `redirect_uri` parameter due to:

- A mismatch with the URI registered in the Linear developer app (must match exactly).
- The redirect URI being incorrectly encoded or missing expected parameters.

### 🧪 Environment

- **Onyx version**: nightly-latest-20250527

PR #4796, opened 2025-05-30
fix: update OAuth callback URI encoding in Linear connector. Issue: https://github.com/onyx-dot-app/onyx/issues/4795

pavelt72:
- Add URL encoding option to get_oauth_callback_uri utility function
- Update Linear connector to use non-encoded callback URI for token exchange

## Description

Onyx fails to add Linear connector with invalid_redirect_uri error. This is because callback uri isn't URL encoded and, since Linear expects it to be exactly like it is specified in Linear application, redirect fails.

Added URL encoding when generating a redirect_uri. In addition Linear connector shouldn't use URL encoded URI for token exchanges.

## How Has This Been Tested?

We build our own version with this fix and were able to successfully add Linear connector.

### DIFF
--- a/backend/onyx/connectors/cross_connector_utils/miscellaneous_utils.py
+++ b/backend/onyx/connectors/cross_connector_utils/miscellaneous_utils.py
@@ -4,6 +4,7 @@ from collections.abc import Iterator
 from datetime import datetime
 from datetime import timezone
 from typing import TypeVar
+from urllib.parse import quote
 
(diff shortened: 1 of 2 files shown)


evan-onyx (commented):
This functionality change also affects the Egnyte connector; we'll need to verify that Egnyte still works before approving. If you're willing to demonstrate that as well I'm happy to approve!

PR #4811, opened 2025-06-03
drive oom investigation logs

evan-onyx:
## Description

Some logs that will help us figure out what is causing OOMs from the drive indexing worker

## How Has This Been Tested?

tested in UI


evan-onyx:
not needed

PR #4822, opened 2025-06-04
Fixed indexing when no sites are specified

wenxi-onyx:
## Description

Sharepoint connector didn't correctly list sites in organization

Fixes https://linear.app/danswer/issue/DAN-2044/sharepoint-indexing-bug

## How Has This Been Tested?

Created Sharepoint connector with no sites specified


wenxi-onyx:
@Weves Added test - note that it only checks if any docs are retrieved. Does not assert "expected sites" because number of sites in our tenant may change at any time.

PR #4832, opened 2025-06-05
KG prompt overhaul

Orbital-Web:
## Description

- Made prompt more concise
- Grouped instructions
- Added more guidance for smaller models
- Fixed issue where random attributes will get created
- Fixed issue where it would look at the wrong entity (target/source) for smaller models
- Fixed isse where it was filtering out columns when it should really be returning null (i.e., no info found)

## How Has This Been Tested?

Tried these prompts 3 times on 4.1 and 4o (couldn't use smaller models because earlier kg stages failed)
- How many accounts are there
- What are the 5 most recently completed linear tickets
- What are the jira tickets under EN 13?
- Summarize the second most recently merged GitHub PR
- Which linear tickets did Joachim work on	
- When did Joachim complete linear ticket DAN 2043	
- Which linear tickeets did Rei complete
- When was linear ticket DAN 1747 completed


Orbital-Web:
Should be tested more before merging

PR #4851, opened 2025-06-09
feat: default kg entity types

Orbital-Web:
## Description

Added a bunch of default entity entity types

## How Has This Been Tested?

Locally


backend/onyx/kg/models.py
```diff
@@ -218,7 +219,7 @@ class KGDefaultEntityDefinition(BaseModel):
     description: str
     grounding: KGGroundingType
     active: bool = False
-    grounded_source_name: str | None
+    grounded_source_name: DocumentSource | None
```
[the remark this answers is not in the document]
Orbital-Web: Only used here

backend/onyx/db/kg_config.py
```diff
@@ -84,6 +84,15 @@ def get_kg_config_settings(db_session: Session) -> KGConfigSettings:
     return kg_config_settings
 
 
+def validate_kg_settings(kg_config_settings: KGConfigSettings) -> None:
```
Orbital-Web, line 87: was previously in kg/configuration, which only had this function + the default entity type thing. Moved it into kg_configs for consistency

backend/onyx/db/entity_type.py
```diff
@@ -104,106 +99,6 @@ def get_entity_types(
         )
 
 
-def populate_default_primary_grounded_entity_type_information(
```
Orbital-Web, line 107: Moved into kg/setup/kg_default_entity_definitions and combined the two very similar functions into one `populate_default_entity_types`

backend/onyx/kg/configuration.py
Orbital-Web, line 1: as per previous comment, moved all functions in here to where it makes more sense

backend/onyx/kg/kg_default_entity_definitions.py
Orbital-Web, line 1: moved to setup folder

Orbital-Web, line 1: moved to setup folder

backend/onyx/kg/configuration.py
Orbital-Web, line 1: as per previous comment, deleted and moved its functions to where it makes more sense

backend/onyx/chat/process_message.py
```diff
@@ -98,14 +98,13 @@
 from onyx.file_store.utils import load_all_chat_files
 from onyx.file_store.utils import save_files
 from onyx.kg.clustering.clustering import kg_clustering
-from onyx.kg.configuration import populate_default_account_employee_definitions
-from onyx.kg.configuration import populate_default_grounded_entity_types
```
raunakab, line 102: Are these functions being used elsewhere? If not, then this could be a prime opportunity for some good, ol'fashioned code-deleting.

backend/onyx/db/entity_type.py
```diff
@@ -104,106 +99,6 @@ def get_entity_types(
...
-        )
-
-        # Add to session
-        db_session.add(new_entity_type)
-
-    # Commit changes
-    db_session.flush()
-
-
 def get_grounded_entity_types_with_null_grounded_source(
```
raunakab: Also, is this function still used? Maybe we should try to get ride of any unnecessary utility functions that we have in this file?

backend/onyx/kg/configuration.py
raunakab, line 1: Deleted code always makes me happy.

backend/onyx/db/kg_config.py
```diff
@@ -84,6 +84,15 @@ def get_kg_config_settings(db_session: Session) -> KGConfigSettings:
...
+    if not kg_config_settings.KG_ENABLED:
+        raise ValueError("KG is not enabled")
+    if not kg_config_settings.KG_VENDOR:
+        raise ValueError("KG_VENDOR is not set")
+    if not kg_config_settings.KG_VENDOR_DOMAINS:
+        raise ValueError("KG_VENDOR_DOMAINS is not set")
```
raunakab, line 93: These checks, ideally, should be at the source level (i.e., in the actual schema of the database).

This is not a nit for this PR, so feel free to resolve after reading this (+ adding your opinions, if needed), but this is something that we all should be more cognizant of.

backend/onyx/kg/setup/kg_default_entity_definitions.py
```diff
@@ -0,0 +1,377 @@
...
+
+        # Create an instance of the default definitions
+        default_definitions = [
+            KGDefaultPrimaryGroundedEntityDefinitions(),
+            KGDefaultAccountEmployeeDefinitions(),
+        ]
+
+        # Iterate over all attributes in the default definitions
+        for default_definition in default_definitions:
+            for id_name, definition in default_definition.model_dump().items():
```
raunakab: Hmm, instead of iterating through `KGDefaultPrimaryGroundedEntityDefinitions` and `KGDefaultAccountEmployeeDefinitions`, maybe we could have all the fields as entries in a list instead? Then we could simply iterate over those elements. Something like:

```py
ENTITIES = {
  "Linear": KGDefaultEntityDefinition(
        description="A formal Linear ticket about a product issue or improvement request.",
        attributes={
            "metadata_attributes": {
                "team": "",
                "state": "",
                "priority": "",
                "created_at": "",
                "completed_at": "",
            },
            "entity_filter_attributes": {},
            "classification_attributes": {},
        },
        grounding=KGGroundingType.GROUNDED,
        grounded_source_name=DocumentSource.LINEAR,
    ),
  # etc.
}
```

backend/onyx/db/kg_config.py
Orbital-Web: I agree. For now, I think this whole kg config business is temporary so I will mentally note it down as a todo when the inevitable refactoring of the config comes.

Will probably make these fields non optional and differentiate kg_enabled vs disabled based on whether the config object as a whole exists or not.

backend/onyx/chat/process_message.py
Orbital-Web: I believe these are deleted (merged into one, rather)

backend/onyx/db/entity_type.py
Orbital-Web: Removed a couple unused code

backend/onyx/kg/setup/kg_default_entity_definitions.py
Orbital-Web: agreed, will do

raunakab (approved):
Looks good!

raunakab:
The failing MS Teams test is an unrelated error. Will force merge this in, since everything else seems to be working!

PR #4854, opened 2025-06-09
fix: Add rate-limiting to Teams API request

raunakab:
## Description

This PR adds rate-limiting to API requests made to MS Teams.

Addresses: https://linear.app/danswer/issue/DAN-2063/add-rate-limiting-to-ms-teams-connector-apis.

## How Has This Been Tested?

The only way to test this would be through scale-testing, which we don't have just yet.

### DIFF
--- a/backend/onyx/connectors/teams/connector.py
+++ b/backend/onyx/connectors/teams/connector.py
@@ -16,6 +16,7 @@ from office365.teams.team import Team  # type: ignore
 
 from onyx.configs.constants import DocumentSource
 from onyx.connectors.cross_connector_utils.miscellaneous_utils import time_str_to_utc
+from onyx.connectors.cross_connector_utils.rate_limit_wrapper import rate_limit_builder
 from onyx.connectors.exceptions import ConnectorValidationError
 from onyx.connectors.exceptions import CredentialExpiredError
 from onyx.connectors.exceptions import InsufficientPermissionsError
@@ -492,7 +493,12 @@ def _collect_documents_for_channel(
             continue
 
         try:
-            replies = list(message.replies.get_all().execute_query())
+
+            @rate_limit_builder(max_calls=50, period=60)
+            def fetch_replies() -> list[ChatMessage]:
+                return list(message.replies.get_all().execute_query())
+
+            replies = fetch_replies()
             thread = [message]
             thread.extend(replies[::-1])
 


backend/onyx/connectors/teams/connector.py
```diff
@@ -492,7 +493,12 @@ def _collect_documents_for_channel(
             continue
 
         try:
-            replies = list(message.replies.get_all().execute_query())
+
+            @rate_limit_builder(max_calls=50, period=60)
```
Weves: how did we get these numbers?

Weves: can we add a comment about that

raunakab: Copy-pasta from another invocation to `rate_limit_builder`. I believe the reference was from `backend/onyx/connectors/discourse/connector.py:40`.

I didn't think too much about the numbers, just wanted to throw up some changes and get some quick feedback! I'll try hammering the APIs and see what the sweet spot is.

backend/onyx/connectors/teams/connector.py
```diff
@@ -492,7 +493,54 @@ def _collect_documents_for_channel(
...
+                        continue
+                    if cre.response.status_code != 429:
+                        continue
+
+                    retry_after = int(cre.response.headers.get("Retry-After", 10))
+                    time.sleep(retry_after)
+                    retries += 1
+
+            if cre:
+                failure_message = f"Retrieval of message and its replies failed; {channel.id=} {message.id}"
```
evan-onyx: maybe `message.id=` ?

backend/onyx/connectors/teams/connector.py
```diff
@@ -492,7 +493,54 @@ def _collect_documents_for_channel(
...
+                )
+
+            if not replies:
+                yield ConnectorFailure(
+                    failed_entity=EntityFailure(
+                        entity_id=message.id,
+                    ),
+                    failure_message=f"Retrieval of message and its replies failed; {channel.id=} {message.id}",
+                    exception=cre,
+                )
```
evan-onyx, line 529: This whole ConnectorFailure block should probably be replaced with the following
```
failure_message = f"Retrieval of message and its replies failed; {channel.id=} {message.id=}"
if cre or not replies:
    if cre.response:
        failure_message = f"{failure_message}; {cre.response.status_code=}"
    yield ConnectorFailure(
        failed_entity=EntityFailure(
            entity_id=message.id,
        ),
        failure_message=failure_message,
        exception=cre,
    )
```

evan-onyx (approved):
left some nits

backend/onyx/connectors/teams/connector.py
raunakab: This is why you're the goat.

raunakab: Updated!

backend/onyx/connectors/teams/connector.py

backend/onyx/connectors/teams/connector.py
raunakab: Outdated comment; resolving!

PR #4855, opened 2025-06-09
fix(discourse): Remove early break that was limiting topics to indexing batch size

suvodhoy:
## Description

The Discourse connector was incorrectly `break`ing after processing the first batch of topics (16) even though the API returns 30 topics per page. This caused the connector to miss processing the remaining topics from each page.

## How Has This Been Tested?

Noticed a discrepancy in the document count indexed when trying to sync my own discourse setup. I tested by making the changes and building the `background` service locally and running the sync again. 


Weves (approved):

Weves:
Thanks for this fix!

Issue #4775, opened 2025-05-26
Make password validation configurable


Django149:
Right now, when creating a new user, the `validate_password` function is always called and enforces hardcoded rules for length, uppercase/lowercase letters, digits, and special characters.

It would be great to make this configurable, so people can adjust the password policy to fit their needs. This could be done in a simple way, like an `ENABLE_PASSWORD_VALIDATION` flag, or in a more flexible way, like a `PASSWORD_POLICY` config.

SupeRoy100:
+1

PR #4869, opened 2025-06-10
Make password validation configurable

Django149:
## Description

Make password validation configurable
Related Issue: https://github.com/onyx-dot-app/onyx/issues/4775

## How Has This Been Tested?

Locally.


wenxi-onyx:
Hey @Django149 - thanks for the contribution! Is there a use case where someone would need this much control over password? The only time this has come up for me is testing features that require auth during dev.

Django149:
Hi @wenxi-onyx ! The goal is to support different deployments with varying password policies—some organization require specific rules, so this adds flexibility while keeping secure defaults.

PR #4871, opened 2025-06-11
Feat: Better subtype support for Knowledge Graph

Orbital-Web:
## Description

- Partially removed kg_terms as it messes up the entity extraction. Will get fully removed in following PR
- Added source & target entity class columns to kg temp relationship view so `a3_sgenerate_simple_sql` works with subtypes
- Made changes to `get_allowed_relationship_type` and `get_allowed_relationship_type_pairs` to work with subtypes
- Made changes to `normalize_entities` and `normalize_relationships` to work with subtypes
- dropped `normalize_terms` and `NormalizedTerms` model for same reason
- added the entity class in the entity type list string that gets inserted into prompts so the entity class can get extracted as a valid entity type
- added utility tool to get entity class and subclass from an entity type (if any)
- modified entity and relationship extraction prompt to be aware of subtypes. Entity extraction prompt remains unchanged as we know exactly what the subtype is
- TODO: modify sql generation prompt to be aware of subtypes

## How Has This Been Tested?

### DIFF
--- a/backend/onyx/agents/agent_search/kb_search/models.py
+++ b/backend/onyx/agents/agent_search/kb_search/models.py
@@ -8,7 +8,6 @@ from onyx.agents.agent_search.kb_search.states import YesNoEnum
 
 class KGQuestionEntityExtractionResult(BaseModel):
     entities: list[str]
-    terms: list[str]
     time_filter: str | None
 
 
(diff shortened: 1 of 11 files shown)


Orbital-Web:
Reworking subtypes entirely

PR #4876, opened 2025-06-11
cleanup: Edit logic for default entity-types population

raunakab:
## Description

This PR updates how the `populate_default_entity_types` function works internally, and what it returns. With this update, it returns all the entity-types (including the fresh, templatized defaults) which have been inserted into the DB.

Addresses: https://linear.app/danswer/issue/DAN-2080/edit-how-populate-default-entity-types-works-and-what-it-returns.

## How Has This Been Tested?

Not tested.

### DIFF
--- a/backend/onyx/kg/setup/kg_default_entity_definitions.py
+++ b/backend/onyx/kg/setup/kg_default_entity_definitions.py
@@ -355,7 +355,6 @@ def populate_default_entity_types(
 
     db_session.commit()
 
-    for existing_entity_type in existing_entity_types.values():
-        entity_types.append(existing_entity_type)
+    entity_types.extend(existing_entity_types.values())
 
     return entity_types
--- a/backend/onyx/kg/setup/kg_default_entity_definitions.py
+++ b/backend/onyx/kg/setup/kg_default_entity_definitions.py
@@ -10,352 +10,315 @@ from onyx.kg.models import KGDefaultEntityDefinition
 from onyx.kg.models import KGGroundingType
 
 
-KG_DEFAULT_PRIMARY_GROUNDED_ENTITIES = {
-    "LINEAR": KGDefaultEntityDefinition(
-        description="A formal Linear ticket about a product issue or improvement request.",
-        attributes={
-            "metadata_attributes": {
-                "team": "",
-                "state": "",
-                "priority": "",
-                "created_at": "",
-                "completed_at": "",
-            },
-            "entity_filter_attributes": {},
-            "classification_attributes": {},
-        },
-        grounding=KGGroundingType.GROUNDED,
-        grounded_source_name=DocumentSource.LINEAR,
-    ),
-    "JIRA-EPIC": KGDefaultEntityDefinition(
-        description=(
-            "A formal Jira ticket describing large bodies of work that can be broken down into "
-            "a number of smaller Jira Tasks, Stories, or Bugs."
-        ),
-        attributes={
-            "metadata_attributes": {
-                "status": "",
-                "priority": "",
-                "reporter": "",
-                "project_name": "",
-                "created": "",
-                "updated": "",
-            },
-            "entity_filter_attributes": {"issuetype": "Epic"},
-            "classification_attributes": {},
-        },
-        grounding=KGGroundingType.GROUNDED,
-        grounded_source_name=DocumentSource.JIRA,
-    ),
-    "JIRA-STORY": KGDefaultEntityDefinition(
-        description=(
-            "Also called 'user stories', these are Jira tickets describing short requirements or requests "
-            "written from the perspective of the end user."
-        ),
-        attributes={
-            "metadata_attributes": {
-                "status": "",
-                "priority": "",
-                "reporter": "",
-                "project_name": "",
-                "created": "",
-                "updated": "",
-            },
-            "entity_filter_attributes": {"issuetype": "Story"},
-            "classification_attributes": {},
-        },
-        grounding=KGGroundingType.GROUNDED,
-        grounded_source_name=DocumentSource.JIRA,
-    ),
-    "JIRA-BUG": KGDefaultEntityDefinition(
-        description=("A Jira ticket describing a bug."),
-        attributes={
-            "metadata_attributes": {
-                "status": "",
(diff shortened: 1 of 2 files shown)


backend/onyx/kg/setup/kg_default_entity_definitions.py
```diff
@@ -309,48 +310,109 @@
...
+    )
+
+
+def generate_non_existing_entity_types(
+    existing_entity_types: dict[str, KGEntityType],
+    vendor_name: str,
+) -> Generator[KGEntityType]:
+    default_definitions = {
+        **KG_DEFAULT_PRIMARY_GROUNDED_ENTITIES,
+        **KG_DEFAULT_ACCOUNT_EMPLOYEE_ENTITIES,
```
Orbital-Web: I believe this makes it so that if there are two values with the same key, the one that comes later is picked. This means the ACCOUNT entity type gets overwritten, which is a problem. I honestly think we should just combine everything into a single default list, (or deduplicate in the default lists)

backend/onyx/kg/setup/kg_default_entity_definitions.py
```diff
@@ -309,48 +310,109 @@
 }
 
 
-def populate_default_entity_types() -> None:
-    with get_session_with_current_tenant() as db_session:
-        kg_config_settings = get_kg_config_settings(db_session)
-        validate_kg_settings(kg_config_settings)
+TEMPLATE = "---vendor_name---"
```
Orbital-Web: I think if we're doing this, we should update the default entity description to use this variable too so it is consistent. Otherwise, we're just pulling out a constant for no reason.

backend/onyx/kg/setup/kg_default_entity_definitions.py
```diff
@@ -309,48 +310,109 @@
 }
 
 
-def populate_default_entity_types() -> None:
-    with get_session_with_current_tenant() as db_session:
-        kg_config_settings = get_kg_config_settings(db_session)
-        validate_kg_settings(kg_config_settings)
```
Orbital-Web, line 315: I think I prefer this over having a separate function to grab just the vendor. Yes, we are grabbing other stuff that is not really needed, but I think the plan is to refactor kg config anyways so we have a single jsonb type column with all the configs, and we won't be able to grab individual fields in the kgconfig like this. It also feels like a slight duplicated logic.

Orbital-Web (commented):
few comments

backend/onyx/kg/setup/kg_default_entity_definitions.py
raunakab: I'll combine into one list. I think that solution is the best for us.

raunakab: But good catch! I didn't look into the actual contents of the map; I was more so copying your code (and mainly the logic) over.

raunakab: Updated!

backend/onyx/kg/setup/kg_default_entity_definitions.py
raunakab: I actually updated this to use functions to generate the verbiage instead of search+replace. A function seems a bit more appropriate here!

backend/onyx/kg/setup/kg_default_entity_definitions.py
raunakab: I think the reason why I opted for this is because I was getting "circular import errors" when using this logic.

raunakab: Whoops, looks like I was getting ahead of myself.

The KG-FS PR which will consume the changes in this PR will utilize the `populate_default_entity_types` function, which will (in the future) cause a circular dependency issue.

Sorry about the confusion!

raunakab: I've edited the code right now to use your implementation. But we'll need to come up with a solution for this soon!

backend/onyx/kg/setup/kg_default_entity_definitions.py
```diff
@@ -1,356 +1,360 @@
...
-        grounding=KGGroundingType.GROUNDED,
-        active=False,
-        grounded_source_name=None,
-    ),
-}
+    }
+
+
+def _generate_non_existing_entity_types(
+    existing_entity_types: dict[str, KGEntityType],
```
Orbital-Web, line 303: Should be set

backend/onyx/kg/setup/kg_default_entity_definitions.py
```diff
@@ -1,356 +1,360 @@
...
-                )
-                new_entity_type = KGEntityType(
-                    id_name=id_name,
-                    description=description,
-                    attributes=definition.attributes,
-                    grounding=definition.grounding,
-                    grounded_source_name=grounded_source_name,
-                    active=False,
-                )
+    for existing_entity_type in existing_entity_types.values():
```
Orbital-Web: existing_entity_types should be set, so we can simply do entity_types.extend(existing_entity_types)

Orbital-Web (commented):
a few comments, otherwise good

backend/onyx/kg/setup/kg_default_entity_definitions.py
raunakab: Actually, I edited the line to be:

```py
existing_entity_types = {
    et.id_name: et for et in db_session.query(KGEntityType).all()
}
```

raunakab: This creates a `dict`.

backend/onyx/kg/setup/kg_default_entity_definitions.py
raunakab: It's a `dict`, but you are right in that this is unnecessarily verbose.

I can instead perform:

```py
entity_types.extend(existing_entity_types.values())
```

raunakab: Updated!

Orbital-Web (approved):
looks good 👍

PR #4877, opened 2025-06-11
fix: answer streaming

Orbital-Web:
## Description

Fixed answer streaming, particularly for models like claude and made it standardized with the other streamings.

Before
image

After
image
(the bullet point not being right is a claude issue)

## How Has This Been Tested?

Locally, with openai and claude

### DIFF
--- a/backend/onyx/agents/agent_search/kb_search/nodes/d1_generate_answer.py
+++ b/backend/onyx/agents/agent_search/kb_search/nodes/d1_generate_answer.py
@@ -7,13 +7,7 @@ from langgraph.types import StreamWriter
 
 from onyx.access.access import get_acl_for_user
 from onyx.agents.agent_search.kb_search.graph_utils import rename_entities_in_answer
-from onyx.agents.agent_search.kb_search.graph_utils import (
-    stream_write_close_main_answer,
-)
 from onyx.agents.agent_search.kb_search.graph_utils import stream_write_close_steps
-from onyx.agents.agent_search.kb_search.graph_utils import (
-    stream_write_main_answer_token,
-)
 from onyx.agents.agent_search.kb_search.ops import research
 from onyx.agents.agent_search.kb_search.states import MainOutput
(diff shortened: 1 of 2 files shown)


backend/onyx/agents/agent_search/kb_search/nodes/d1_generate_answer.py
```diff
@@ -221,70 +203,24 @@ def generate_answer(
...
+            lambda: stream_llm_answer(
+                llm=graph_config.tooling.fast_llm,
+                prompt=msg,
+                event_name="initial_agent_answer",
+                writer=writer,
+                agent_answer_level=0,
+                agent_answer_question_num=0,
+                agent_answer_type="agent_level_answer",
+                timeout_override=30,
+                max_tokens=1000,
```
raunakab: Should these be globals?

backend/onyx/agents/agent_search/kb_search/nodes/d1_generate_answer.py
```diff
@@ -221,70 +203,24 @@ def generate_answer(
...
-                )
-                response.append(token)
-        return response
-
     try:
-        response = run_with_timeout(
+        run_with_timeout(
             KG_ANSWER_GENERATION_TIMEOUT,
-            stream_answer,
+            lambda: stream_llm_answer(
```
raunakab, line 211: What's the difference that this change makes? Why are we now running `stream_llm_answer` as opposed to `stream_answer`?

Orbital-Web: stream answer does the same thing as stream_llm_answer, with some extra stuff that are not needed anymore. Thus, it both fixes the issue and removes duplicate code.

raunakab: Deleting duplicated code is the greatest thing ever invented.

raunakab (approved):
Looks good!

PR #4879, opened 2025-06-11
GCS metadata processing

wenxi-onyx:
## Description

Text files from blob can have metadata just like file connector

Fixes https://linear.app/danswer/issue/DAN-2078/gcs-metadata
## How Has This Been Tested?

### DIFF
--- a/backend/onyx/connectors/blob/connector.py
+++ b/backend/onyx/connectors/blob/connector.py
@@ -372,7 +372,11 @@ class BlobStorageConnector(LoadConnector, PollConnector):
                     batch.append(
                         Document(
                             id=f"{self.bucket_type}:{self.bucket_name}:{key}",
-                            sections=sections,
+                            sections=(
+                                sections
+                                if sections
+                                else [TextSection(link=final_link, text="")]
+                            ),
                             source=DocumentSource(self.bucket_type.value),
                             semantic_identifier=file_display_name,
                             doc_updated_at=final_time_updated,
--- a/backend/onyx/connectors/blob/connector.py
+++ b/backend/onyx/connectors/blob/connector.py
@@ -315,13 +315,11 @@ class BlobStorageConnector(LoadConnector, PollConnector):
                     onyx_metadata, custom_tags = process_onyx_metadata(
                         extraction_result.metadata
                     )
-                    file_display_name = (
-                        onyx_metadata.get("file_display_name") or file_name
-                    )
-                    time_updated = onyx_metadata.get("doc_updated_at") or last_modified
-                    link = onyx_metadata.get("link") or link
-                    primary_owners = onyx_metadata.get("primary_owners")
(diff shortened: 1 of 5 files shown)


backend/onyx/connectors/blob/connector.py
```diff
@@ -304,19 +307,82 @@ def _yield_blob_objects(
...
+                    dt_str = extraction_result.metadata.get("doc_updated_at")
+                    final_time_updated = (
+                        time_str_to_utc(dt_str) if dt_str else last_modified
+                    )
+
+                    file_display_name = (
+                        extraction_result.metadata.get("file_display_name") or file_name
+                    )
+
+                    metadata_tags = {
```
Weves: some of this is repeated logic. Can we move it into `cross_connector_utils`?

Weves (changes requested):

backend/onyx/connectors/blob/connector.py
wenxi-onyx: Moved to utils --> also merged #4866 as that also uses this logic

backend/onyx/connectors/cross_connector_utils/miscellaneous_utils.py
```diff
@@ -81,6 +82,60 @@ def get_metadata_keys_to_ignore() -> list[str]:
     return [IGNORE_FOR_QA]
 
 
+def process_onyx_metadata(
+    metadata: dict[str, Any],
+) -> tuple[dict[str, Any], dict[str, Any]]:
```
Weves: would love to add some typing here (e.g. a Pydantic model or TypedDict which describes the return type)

Weves: this would make sure that any changes to the names of fields here would get caught by static type checking in anything that consumes this function

backend/onyx/connectors/cross_connector_utils/miscellaneous_utils.py
```diff
@@ -81,6 +82,60 @@ def get_metadata_keys_to_ignore() -> list[str]:
...
+            "primary_owners": p_owners,
+            "secondary_owners": s_owners,
+            "time_updated": time_updated,
+        },
+        {
+            k: v
+            for k, v in metadata.items()
+            if k
+            not in [
+                "document_id",
```
Weves, line 123: can we remove this? I don't think we should allow folks to override `document_id` specifically (there's actually some problems here that I don't need to get into)

backend/onyx/connectors/cross_connector_utils/miscellaneous_utils.py
wenxi-onyx: Added OnyxMetadata typeddict

Weves (approved):

PR #4892, opened 2025-06-13
Kg Subtype Rework

Orbital-Web:
## Description

- Refactored entity types to ensure type safety
- Refactored metadata extraction to ensure type safety
- Removed entity class and subtype columns
- Reworked handling of attributes after query entity extractions for more robustness
- Removed terms extraction as it was messing up the entity extraction once in a while for some reason
- Made it so normalization respects subtypes to narrow down filter

## How Has This Been Tested?

Tested on these questions
- What are the available labels for github issues
- How many open jira tickets are there
- Which open tasks are under JIRA epic EN 13
- Which epic has the most open tasks


backend/alembic/versions/65bc6e0f8500_remove_kg_subtype_from_db.py
```diff
@@ -0,0 +1,41 @@
...
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "65bc6e0f8500"
+down_revision = "cec7ec36c505"
```
evan-onyx, line 15: make sure these are correct by rebasing before merging

backend/alembic/versions/65bc6e0f8500_remove_kg_subtype_from_db.py
```diff
@@ -0,0 +1,41 @@
...
+
+def upgrade() -> None:
+    op.drop_column("kg_entity", "entity_class")
+    op.drop_column("kg_entity", "entity_subtype")
+    op.drop_column("kg_entity_extraction_staging", "entity_class")
+    op.drop_column("kg_entity_extraction_staging", "entity_subtype")
+
+
+def downgrade() -> None:
+    op.add_column(
```
evan-onyx, line 28: typically should do these operations in reverse order of the upgrade

backend/onyx/db/models.py
```diff
@@ -647,14 +649,24 @@ class KGEntityType(Base):
...
         comment="Filtering based on document attribute",
     )
 
+    @property
+    def parsed_attributes(self) -> KGEntityTypeAttributes:
+        if self.attributes is None:
+            return KGEntityTypeAttributes()
+
+        try:
+            return KGEntityTypeAttributes(**self.attributes)
```
evan-onyx, line 666: does it make sense to try adding attributes one by one and throwing out invalid ones?

backend/onyx/document_index/vespa/index.py
```diff
@@ -139,6 +129,34 @@ class KGUDocumentUpdateRequest(BaseModel):
     terms: set[str]
 
 
+def generate_kg_update_request(
+    kg_update_request: KGUChunkUpdateRequest,
+) -> dict[str, dict]:
+    kg_update_dict: dict[str, dict] = {"fields": {}}
```
evan-onyx: just create and return the inner dictionary in this function; the caller should be responsible for wrapping the output in {"fields": <the dict>}

backend/onyx/document_index/vespa/shared_utils/vespa_request_builders.py
```diff
@@ -69,20 +70,42 @@ def _build_kg_filter(
...
                 filter_parts.append(
-                    " and ".join(f'({filter_type} contains "{val}") ' for val in values)
+                    "(kg_relationships contains sameElement("
+                    f"source contains {_build_kge(source)},"
+                    f'rel_type contains "{rel_type}",'
+                    f"target contains {_build_kge(target)}))"
                 )
+            combined_filter_parts.append(f"{' and '.join(filter_parts)}")
+
+        # TODO: remove kg terms entirely from prompts and codebase
```
evan-onyx, line 105: ideally we could finish that in this PR but if it's a huge lift it's fine to leave this in

evan-onyx (approved):
some nits, looks good!

backend/onyx/db/models.py
Orbital-Web: Actually, some of the fields have the same type as before, but the values that goes in has been changed. Probably best to keep it this way so we don't get a weird mix of the old format and this new format. 

Shouldn't be invalid anyways if starting from a valid state.

PR #4902, opened 2025-06-16
Kg config refactor

Orbital-Web:
## Description

- Replaced kg config with key value store
- Used locks for managing mutual exclusion of kg processing/reset tasks on a per-tenant basis. Fixes issues where if a process dies in the middle of processing, the lock stays locked forever
- kg_p and other commands now only runs from the KG Beta assistant

Benefits
- faster access because cached
- considerably simpler code for accessing and modifying the config

Downsides-ish
- may need to migrate if we add/remove stuff to the kg config unless we're fine with filling missing values with null (which technically doesn't make it any different from the current setup, so not really a downside)

## How Has This Been Tested?

Single tenant:
- [x] Processing starts if there are docs to be processed every minute or so
- [x] kg_p while processing/resetting is ongoing will not do anything
- [x] kg_rs_source while processing/resetting is ongoing will not do anything
- [x] the periodic processing does not start while kg_p or kg_rs_source is running
- [x] Killing the kg processing instance while processing is ongoing should not leave the lock locked

Multi tenant:
- [x] Processing starts if there are docs to be processed every minute or so, for that tenant
- [x] kg_p while processing/resetting is ongoing for that tenant will not do anything
- [x] kg_rs_source while processing/resetting is ongoing for that tenant will not do anything
- [x] the periodic processing does not start while kg_p or kg_rs_source is running for that tenant
- [x] multiple processing can occur simultaneously if they are for different tenants

### DIFF
--- a/backend/alembic/versions/03bf8be6b53a_rework_kg_config.py
+++ b/backend/alembic/versions/03bf8be6b53a_rework_kg_config.py
@@ -12,7 +12,7 @@ import json
 from datetime import datetime
 from datetime import timedelta
 from sqlalchemy.dialects import postgresql
-
+from sqlalchemy import text
 from alembic import op
 import sqlalchemy as sa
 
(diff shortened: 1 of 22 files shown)


backend/onyx/db/kg_config.py
```diff
@@ -1,131 +1,36 @@
...
-        .values(
-            kg_variable_name=KGConfigVars.KG_BETA_PERSONA_ID,
-            kg_variable_values=value,
-        )
-        .on_conflict_do_update(
-            index_elements=["kg_variable_name"],
-            set_=dict(kg_variable_values=value),
+        stored_config = kv_store.load(KV_KG_CONFIG_KEY)
+        return (
+            KGConfigSettings.model_validate(stored_config)
```
evan-onyx: `KGConfigSettings.model_validate(stored_config or {})` instead of the if/else

backend/alembic/versions/03bf8be6b53a_rework_kg_config.py
```diff
@@ -0,0 +1,116 @@
...
+                "KG_IGNORE_EMAIL_DOMAINS", []
+            ),
+            "KG_COVERAGE_START": current_config_dict.get(
+                "KG_COVERAGE_START", default_coverage_start
+            ),
+            "KG_MAX_COVERAGE_DAYS": current_config_dict.get("KG_MAX_COVERAGE_DAYS", 90),
+            "KG_MAX_PARENT_RECURSION_DEPTH": current_config_dict.get(
+                "KG_MAX_PARENT_RECURSION_DEPTH", 2
+            ),
+            "KG_BETA_PERSONA_ID": current_config_dict.get("KG_BETA_PERSONA_ID", None),
```
evan-onyx, line 63: any reason why this isn't in the downgrade?

backend/onyx/background/celery/tasks/kg_processing/kg_indexing.py
```diff
@@ -39,32 +31,29 @@ def try_creating_kg_processing_task(
         )
 
         if not result:
-            raise RuntimeError("send_task for kg processing failed.")
-
+            task_logger.error("send_task for kg processing failed.")
+            return False
```
evan-onyx: nit: if you `return bool(result)` de-indented here then you can remove the return True at the end of the function

backend/onyx/background/celery/tasks/kg_processing/kg_indexing.py
```diff
@@ -79,11 +68,13 @@ def try_creating_kg_source_reset_task(
         )
 
         if not result:
-            raise RuntimeError("send_task for kg source reset failed.")
+            task_logger.error("send_task for kg source reset failed.")
+            return False
```
evan-onyx: similar to above

backend/onyx/chat/process_message.py
```diff
@@ -632,6 +602,56 @@ def stream_chat_message_objects(
...
+            kg_config_settings
+        ) and persona.name.startswith("KG Beta"):
+            # get Vespa index
+            search_settings = get_current_search_settings(db_session)
+            index_str = search_settings.index_name
+
+            if new_msg_req.message == "kg_p":
+                success = try_creating_kg_processing_task(tenant_id)
+                if success:
+                    raise NotImplementedError("KG processing scheduled")
```
evan-onyx: This probably should be a different exception type, maybe make a new KGException?

evan-onyx: bc NotImplemented could in theory be raised somewhere unexpected in library code

backend/onyx/chat/process_message.py
```diff
@@ -632,6 +602,56 @@ def stream_chat_message_objects(
             default_persona=chat_session.persona,
         )
 
+        # Temporarily, until we have a draft UI for the KG Operations/Management
```
evan-onyx: pls put this code block in a function that lives somewhere else bc process_message is already a monstrosity haha

backend/onyx/key_value_store/store.py
```diff
@@ -52,14 +52,17 @@ def store(self, key: str, val: JSON_ro, encrypt: bool = False) -> None:
...
-                assert isinstance(redis_value, bytes)
-                return json.loads(redis_value.decode("utf-8"))
-        except Exception as e:
-            logger.error(f"Failed to get value from Redis for key '{key}': {str(e)}")
+    def load(self, key: str, refresh_cache: bool = False) -> JSON_ro:
+        if not refresh_cache:
+            try:
+                redis_value = self.redis_client.get(REDIS_KEY_PREFIX + key)
+                if redis_value:
+                    assert isinstance(redis_value, bytes)
```
evan-onyx: raise a ValueError instead of asserting bc Chris wants to move away from asserts

evan-onyx (approved):
Left a few nits, looking good though!

backend/alembic/versions/03bf8be6b53a_rework_kg_config.py
Orbital-Web: got lazy, will do...

PR #4906, opened 2025-06-17
fix: Implement time-filtering for MS Teams document fetching

raunakab:
## Description

This PR enables performing *source-side* filtering for fetching new messages.

Addresses: https://linear.app/danswer/issue/DAN-2100/teams-source-side-delta-document-filtering.

## Notes

The filtering which is implemented only filters for messages *past* a given time, *NOT WITHIN A TIME FRAME*. I.e., it does *not* support fetching messages from `time-a` to `time-b`. It only performs fetching messages from `time-a` onwards.

This is not my error; this is because the MS Teams API does not support filtering based off of a time-window...

## How Has This Been Tested?

Existing tests already sufficient.

### DIFF
--- a/backend/onyx/connectors/teams/utils.py
+++ b/backend/onyx/connectors/teams/utils.py
@@ -5,7 +5,6 @@ from datetime import timezone
 from http import HTTPStatus
 
 from office365.graph_client import GraphClient  # type: ignore
-from pydantic import Json
 
 from onyx.connectors.interfaces import SecondsSinceUnixEpoch
 from onyx.connectors.teams.models import Message
@@ -14,14 +13,18 @@ from onyx.connectors.teams.models import Message
 def retry(
     graph_client: GraphClient,
     request_url: str,
-) -> Json:
+) -> dict:
     MAX_RETRIES = 10
     retry_number = 0
 
     while retry_number < MAX_RETRIES:
         response = graph_client.execute_request_direct(request_url)
         if response.ok:
-            return response.json()
+            json = response.json()
+            if not isinstance(json, dict):
+                raise RuntimeError(f"Expected a JSON object, instead got {json=}")
+
+            return json
 
         if response.status_code == int(HTTPStatus.TOO_MANY_REQUESTS):
             retry_number += 1
@@ -38,6 +41,23 @@ def retry(
     )
 
 
+def get_next_url(
+    graph_client: GraphClient,
+    json_response: dict,
+) -> str | None:
+    next_url = json_response.get("@odata.nextLink")
+
+    if not next_url:
+        return None
+
+    if not isinstance(next_url, str):
+        raise RuntimeError(
+            f"Expected a string for the `@odata.nextUrl`, instead got {next_url=}"
+        )
+
+    return next_url.removeprefix(graph_client.service_root_url()).removeprefix("/")
+
+
 def fetch_messages(
     graph_client: GraphClient,
     team_id: str,
@@ -48,15 +68,22 @@ def fetch_messages(
         "%Y-%m-%dT%H:%M:%SZ"
     )
 
-    request_url = (
+    initial_request_url = (
         f"teams/{team_id}/channels/{channel_id}/messages/delta"
         f"?$filter=lastModifiedDateTime gt {startfmt}"
     )
 
-    json_response = retry(graph_client=graph_client, request_url=request_url)
+    request_url: str | None = initial_request_url
 
-    for value in json_response.get("value", []):
-        yield Message(**value)
+    while request_url:
+        json_response = retry(graph_client=graph_client, request_url=request_url)
+
+        for value in json_response.get("value", []):
+            yield Message(**value)
+
+        request_url = get_next_url(
+            graph_client=graph_client, json_response=json_response
+        )
 
 
 def fetch_replies(
@@ -65,11 +92,18 @@ def fetch_replies(
     channel_id: str,
     root_message_id: str,
 ) -> Generator[Message]:
-    request_url = (
+    initial_request_url = (
         f"teams/{team_id}/channels/{channel_id}/messages/{root_message_id}/replies"
     )
 
-    response_json = retry(graph_client=graph_client, request_url=request_url)
+    request_url: str | None = initial_request_url
 
-    for value in response_json.get("value", []):
-        yield Message(**value)
+    while request_url:
+        json_response = retry(graph_client=graph_client, request_url=request_url)
+
+        for value in json_response.get("value", []):
+            yield Message(**value)
+
+        request_url = get_next_url(
+            graph_client=graph_client, json_response=json_response
+        )


backend/onyx/connectors/teams/connector.py
```diff
@@ -159,7 +158,7 @@ def validate_checkpoint_json(self, checkpoint_json: str) -> TeamsCheckpoint:
     def load_from_checkpoint(
         self,
         start: SecondsSinceUnixEpoch,
-        end: SecondsSinceUnixEpoch,
+        _end: SecondsSinceUnixEpoch,
```
evan-onyx: this is an overridden function, so don't change the function signature

backend/onyx/connectors/teams/connector.py
```diff
@@ -223,34 +220,23 @@ def load_from_checkpoint(
...
 
 
-def _get_created_datetime(chat_message: ChatMessage) -> datetime:
-    # Extract the 'createdDateTime' value from the 'properties' dictionary and convert it to a datetime object
-    return time_str_to_utc(chat_message.properties["createdDateTime"])
-
-
-def _extract_channel_members(channel: Channel) -> list[BasicExpertInfo]:
-    channel_members_list: list[BasicExpertInfo] = []
+def _extract_channel_members(channel: Channel) -> dict[str, BasicExpertInfo]:
```
evan-onyx: I'd say this should just return a set of strings; later when you want to create a document you can convert into a list of BasicExpertInfo

backend/onyx/connectors/teams/connector.py
```diff
@@ -223,34 +220,23 @@ def load_from_checkpoint(
...
 
+    return {
+        member.display_name: BasicExpertInfo(display_name=member.display_name)
+        for member in members
+    }
 
-def _construct_semantic_identifier(channel: Channel, top_message: ChatMessage) -> str:
-    # NOTE: needs to be done this weird way because sometime we get back `None` for
-    # the fields which causes things to explode
-    top_message_from = top_message.properties.get("from") or {}
```
evan-onyx, line 245: imo these variables add clarity

backend/onyx/connectors/teams/connector.py
```diff
@@ -293,42 +280,34 @@ def _convert_thread_to_document(
...
+            )
 
     if not thread_text:
         return None
 
     # if there are no found post members, grab the members from the parent channel
-    if not post_members_list:
-        post_members_list = _extract_channel_members(channel)
+    if not posters:
+        _extract_channel_members(channel)
```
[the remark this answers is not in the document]
evan-onyx: agree

backend/onyx/connectors/teams/connector.py
```diff
@@ -468,67 +446,28 @@ def _collect_documents_for_channel(
...
     """
     This function yields an iterator of `Document`s, where each `Document` corresponds to a "thread".
 
     A "thread" is the conjunction of the "root" message and all of its replies.
     """
 
-    # Server-side filter conditions are not supported on the chat-messages API.
-    # Therefore, we have to do this client-side, which is quite a bit more inefficient.
-    #
-    # Not allowed:
```
evan-onyx, line 482: should probably leave this comment in?

backend/onyx/connectors/teams/models.py
```diff
@@ -0,0 +1,52 @@
...
+    content_type: str
+    content: str | None
+
+    model_config = ConfigDict(
+        alias_generator=to_camel,
+        populate_by_name=True,
+    )
+
+
+class From(BaseModel):
```
evan-onyx, line 28: unused as far as I can see, delete

backend/onyx/connectors/teams/models.py
```diff
@@ -0,0 +1,52 @@
...
+from pydantic import Field
+from pydantic.alias_generators import to_camel
+
+
+class Message(BaseModel):
+    id: str
+    replyToId: str | None
+    subject: str | None
+    from_: "From | None" = Field(alias="from")
+    body: "Body"
```
evan-onyx: I think you can just define them in the right order to avoid having to use strings

backend/onyx/connectors/teams/utils.py
```diff
@@ -0,0 +1,75 @@
...
+    start: SecondsSinceUnixEpoch,
+) -> Generator[Message]:
+    startfmt = datetime.fromtimestamp(start, tz=timezone.utc).strftime(
+        "%Y-%m-%dT%H:%M:%SZ"
+    )
+
+    request_url = (
+        f"teams/{team_id}/channels/{channel_id}/messages/delta"
+        f"?$filter=lastModifiedDateTime gt {startfmt}"
+    )
```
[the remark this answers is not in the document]
evan-onyx: investigate whether this is valid

backend/onyx/connectors/teams/connector.py
raunakab: Ah right, updated.

backend/onyx/connectors/teams/connector.py
raunakab: Added back!

backend/onyx/connectors/teams/connector.py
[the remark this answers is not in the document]
raunakab: This is why greptile is the goat and is going to steal all jobs.

backend/onyx/connectors/teams/connector.py
raunakab: Hmm, source-side filtering is now implemented (by hitting a different API [not exposed by the `office365` library]).

I can update the comment!

backend/onyx/connectors/teams/models.py
raunakab: Haha, yes that's true. In an ideal world, Python would not be weird. But alas, we do not live in such a world.

raunakab: Updated!

backend/onyx/connectors/teams/utils.py
```diff
@@ -0,0 +1,75 @@
...
+
+def fetch_replies(
+    graph_client: GraphClient,
+    team_id: str,
+    channel_id: str,
+    root_message_id: str,
+) -> Generator[Message]:
+    request_url = (
+        f"teams/{team_id}/channels/{channel_id}" f"/messages/{root_message_id}/replies"
+    )
```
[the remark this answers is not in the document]
raunakab: Updated!

backend/onyx/connectors/teams/models.py
raunakab: Hmm, it's used inside of `onyx.connectors.teams.connector._construct_semantic_identifier`.

backend/onyx/connectors/teams/connector.py

backend/onyx/connectors/teams/utils.py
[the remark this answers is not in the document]
raunakab: Was valid! Implemented the change.

backend/onyx/connectors/teams/models.py
evan-onyx: Wait where? I've expanded the full function and don't see it?

evan-onyx: ah mb, I guess it's used for validation? seems fine

evan-onyx (approved):

PR #4924, opened 2025-06-21
Update container scripts for new S3 changes

Orbital-Web:
## Description

Updated migrations and the container scripts. Now, there is an option to restart the containers without clearing them.

Here is what you need to do to properly do the S3 migrations, if you are using the launch.json to start up Onyx.
1. Copy the new `launch.template.jsonc` into `launch.json`.
In particular, replace the old `Clear and Restart External Volumes and Containers` with the new `Clear and Restart External Volumes and Containers` and `Restart External Volumes and Containers`.
2. Run `Restart External Volumes and Containers`. This will add the new MinIO container and run the migration, without clearing the containers.

## How Has This Been Tested?

Locally

### DIFF
--- a/.vscode/launch.template.jsonc
+++ b/.vscode/launch.template.jsonc
@@ -371,6 +371,22 @@
         "request": "launch",
         "runtimeExecutable": "bash",
         "runtimeArgs": [
+          "-c",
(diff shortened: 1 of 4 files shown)


Orbital-Web:
something feels wrong... idk what

PR #4925, opened 2025-06-21
KG Updates

joachim-danswer:
## Description

 Here are the highlights:
 
 - no classification if deep extraction is False
 - separate names for views in LLM generation
 - better prompts
 - any relationship type provided to LLM that relates to identified entities

Here is the Linear ticket: https://linear.app/danswer/issue/DAN-2109/small-kg-improvements

## How Has This Been Tested?

Locally, multi-tenant

### DIFF
--- a/backend/onyx/db/relationships.py
+++ b/backend/onyx/db/relationships.py
@@ -490,8 +490,9 @@ def get_allowed_relationship_type_pairs(
         entities: List of entity type ID names to filter by
 
     Returns:
-        List of id_names from KGRelationshipType where both source and target entity types
-        are in the provided entities list
+        List of id_names from KGRelationshipType where source or target entity types
+        are in the provided entities list. We also filter out for now the catch-all
+        relationship types 'VENDOR__<relationship>__<target entity type>'
     """
 
     entity_types = list({get_entity_type(entity) for entity in entities})
--- a/backend/onyx/db/relationships.py
+++ b/backend/onyx/db/relationships.py
@@ -493,14 +493,20 @@ def get_allowed_relationship_type_pairs(
         List of id_names from KGRelationshipType where both source and target entity types
         are in the provided entities list
     """
+
     entity_types = list({get_entity_type(entity) for entity in entities})
 
     return [
         row[0]
         for row in (
             db_session.query(KGRelationshipType.id_name)
-            .filter(KGRelationshipType.source_entity_type_id_name.in_(entity_types))
-            .filter(KGRelationshipType.target_entity_type_id_name.in_(entity_types))
+            .filter(
+                or_(
+                    KGRelationshipType.source_entity_type_id_name.in_(entity_types),
+                    KGRelationshipType.target_entity_type_id_name.in_(entity_types),
+                )
+            )
+            .filter(~KGRelationshipType.source_entity_type_id_name.like("VENDOR::%"))
             .distinct()
             .all()
         )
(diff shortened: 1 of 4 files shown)


backend/onyx/prompts/kg_prompts.py
```diff
@@ -374,7 +374,9 @@
 
 Note that the notation for the entities is <ENTITY_TYPE>::<ENTITY_NAME>.
 
-Here are the options for the relationship types(!) between the entities you have identified earlier:
+Here are the options for the relationship types(!) between the entities you have identified earlier \
```
Orbital-Web, line 377: what's the (!) for? lol

backend/onyx/prompts/kg_prompts.py
```diff
@@ -893,6 +902,10 @@
 - If you do joins consider the possibility that the second entity does not exist for all examples. \
 Therefore joins should generally be LEFT joins (or RIGHT joins) as appropriate. Think about which \
 entities you are interested in, and which ones provides attributes.
+Another important note:
```
Orbital-Web, line 905: I'm guessing you put this because the new get_allowed_relationship_type_pairs shows the VENDOR relates to relationships (which will break the entity only path)? 

Either way, might be worth considering if we should keep the VENDOR entity type in the future

Orbital-Web (approved):
looks good!

backend/onyx/prompts/kg_prompts.py
joachim-danswer: emphasis for model, making it very clear that those are not relationships

backend/onyx/prompts/kg_prompts.py
joachim-danswer: I agree. I don't think we need it anymore. I anticipate that going away over the next few days.

backend/onyx/db/relationships.py
```diff
@@ -493,14 +493,20 @@ def get_allowed_relationship_type_pairs(
...
             db_session.query(KGRelationshipType.id_name)
-            .filter(KGRelationshipType.source_entity_type_id_name.in_(entity_types))
-            .filter(KGRelationshipType.target_entity_type_id_name.in_(entity_types))
+            .filter(
+                or_(
+                    KGRelationshipType.source_entity_type_id_name.in_(entity_types),
+                    KGRelationshipType.target_entity_type_id_name.in_(entity_types),
+                )
+            )
+            .filter(~KGRelationshipType.source_entity_type_id_name.like("VENDOR::%"))
```
Weves, line 510: what's this about?

backend/onyx/db/relationships.py
```diff
@@ -493,14 +493,20 @@ def get_allowed_relationship_type_pairs(
         List of id_names from KGRelationshipType where both source and target entity types
```
Weves: this comment is wrong now?

backend/onyx/db/relationships.py
joachim-danswer: We have currently a catch-all relationship type VENDOR__relates_to__<any entity> which we needed while always retrieved information exclusively through the relationship view. We have changed that so that we can also - if no relationships in question - directly retrieve from a new entity view, which is substantially faster and simpler. So this type of relationship is not needed anymore. Will fully remove from code this week.

backend/onyx/db/relationships.py
joachim-danswer: Correct. I missed to update. This was fitting earlier. Will do and re-queue.

PR #4927, opened 2025-06-21
Hubspot connector enhancements

Weves:
## Description

Add support for Companies, Deals, and Contacts

Fixes https://linear.app/danswer/issue/DAN-2111/enhance-hubspot-connector

Screenshot 2025-06-21 at 3 42 15 PM

## How Has This Been Tested?

Tests + manual testing.

### DIFF
--- a/backend/onyx/connectors/hubspot/connector.py
+++ b/backend/onyx/connectors/hubspot/connector.py
@@ -36,8 +36,8 @@ class HubSpotConnector(LoadConnector, PollConnector):
         object_types: list[str] | None = None,
     ) -> None:
         self.batch_size = batch_size
-        self.access_token = access_token
-        self.portal_id: str | None = None
+        self._access_token = access_token
+        self._portal_id: str | None = None
 
         # Set object types to fetch, default to all available types
         if object_types is None:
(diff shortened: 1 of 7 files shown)


backend/onyx/connectors/hubspot/connector.py
```diff
@@ -42,78 +85,734 @@ def get_portal_id(self) -> str:
...
         if self.access_token:
             self.portal_id = self.get_portal_id()
-            self.ticket_base_url = f"{HUBSPOT_BASE_URL}{self.portal_id}/ticket/"
 
         return None
 
+    def _get_object_url(self, object_type: str, object_id: str) -> str:
+        """Generate HubSpot URL for different object types"""
+        if object_type == "tickets":
+            return f"{HUBSPOT_BASE_URL}/contacts/{self.portal_id}/ticket/{object_id}"
```
joachim-danswer: The /contacts/ is not a typo here?

backend/onyx/connectors/hubspot/connector.py
```diff
@@ -42,78 +85,734 @@ def get_portal_id(self) -> str:
...
+        elif object_type == "companies":
+            return f"{HUBSPOT_BASE_URL}/contacts/{self.portal_id}/company/{object_id}"
```
joachim-danswer: As {HUBSPOT_BASE_URL}/contacts/{self.portal_id} is repeated many times, better to create a variable at start?

backend/onyx/connectors/hubspot/connector.py
```diff
@@ -42,78 +85,734 @@ def get_portal_id(self) -> str:
...
+                    for obj_id in object_ids:
+                        try:
+                            obj = api_client.crm.deals.basic_api.get_by_id(
+                                deal_id=obj_id,
+                                properties=[
+                                    "dealname",
+                                    "amount",
+                                    "dealstage",
+                                    "closedate",
+                                    "pipeline",
```
joachim-danswer, line 208: How does the 'deal' know which company it belongs to?

Nvm - I see that it comes later through associated objects. Leaving this as a note for myself here for now.

backend/onyx/connectors/hubspot/connector.py
```diff
@@ -42,78 +85,734 @@ def get_portal_id(self) -> str:
...
+
+            # Add association IDs to metadata
+            if associated_contact_ids:
+                metadata["associated_contact_ids"] = associated_contact_ids
+            if associated_deal_ids:
+                metadata["associated_deal_ids"] = associated_deal_ids
+            if associated_ticket_ids:
+                metadata["associated_ticket_ids"] = associated_ticket_ids
+
+            doc_batch.append(
```
joachim-danswer, line 563: On the knowledge graph we currently require that the child know about its parent, but not the other way around. Reason is that there could be a lot of opportunities per account, etc. While each opportunity would at best have very few accounts.
But we can leave it this way if we don't think this gets too big. 

So mostly just something to consider.

backend/onyx/connectors/hubspot/connector.py
Weves: it isn't, but for some of the objects the links were wrong. Good catch!

PR #4928, opened 2025-06-22
KG process email fix

Orbital-Web:
## Description

Accounts extracted from GitHub attributes were showing up weird because of the way emails are shown in the metadata. This now extracts the email portion before passing it into the process_email function, rather than passing the whole string and treating everything before the @ as the name and after as the domain. 

Now, even if the attribute looks like {'name' John, 'email': johndoe@gmail.com}, the correct name and domain will be extracted. Same goes with any arbitrary string that contains an email. 

## How Has This Been Tested?

Without this PR:
Accounts extracted from GitHub have a weird } at the end. You'd have to test this on a repo with users that display their email. 

With this PR:
The name and account should look correct.


Orbital-Web:
Will be part of antoher pr

PR #4933, opened 2025-06-23
KG Attribute Overhaul + Processing Tests

Orbital-Web:
## Description

**Bugfixes and general refactoring:**
- Renamed several variables for clarity
- Updated comment when running "kg_p" while processing is blocked or there are no docs to process
- Wrote test cases for extraction and clustering
- Caught and fixed several bugs
    -  ingestion api does not set `doc_updated_at` (confirmed with Evan that this is not intended)
    - entity and parent key was not being properly updated in `merge_entities` and `transfer_entity`
    - attributes were not being properly updated in `merge_entities` 
    - classification path was referring to wrong `document_id` which resulted in mypy variable may be undefined error
    - typing of `succeeded_chunk_extraction` in `_kg_chunk_batch_extraction` was wrong
    - `process_email` assumed passed `email` argument was just the email, but other functions were passing in the entire string containing the email as the argument, resulting in weird account extraction
    - implicit entities could get extracted even if they were not active
    - http client was being closed by the `_kg_update_chunk function`, causing client already closed error on vespa retry
    - metadata tracking only stopped tracking an attribute if the number of possible values exceeded `KG_METADATA_TRACKING_THRESHOLD` in the LAST iteration
- Removed duplicate/no longer needed logic (e.g., generalizing entities in extraction, extracting metadata from documents in several places)

**Features:**
- Frontend now lets you enable/disable implied entity types. E.g., if you have GitHub connector, employee and accounts will also show up in the frontend (as it can be created as an implied entity from GitHub)
- Greatly reworked how metadata is handled. Now, you can specify which relationships to extract, how you want them extracted, which implied entities to extract, etc., plus which attributes to keep at the end. Now you know exactly which implied relationships will get created (and implied entities)
- As a side benefit of the above, there can now be multiple parent child relationships for up to depth 1, and parents can get extracted after the child gets put in the entities table (unlike with the key parent approach) 
E.g., for Hubspot, you would have:
type=`CONTACT`, `metadata_attribute_conversion`:
    "accounts" -> `name`="accounts", `implied_entity_type`="ACCOUNT", `implied_relationship_name`="has_contact"
 and that will automatically create an ACCOUNT entity and the has_contact relationship (for each item in `accounts` field of the CONTACT document metadata)

## How Has This Been Tested?

Locally, plus with the new tests cases. 


backend/onyx/kg/models.py
```diff
@@ -44,18 +44,36 @@ class KGAttributeTrackInfo(BaseModel):
     values: set[str] | None
 
 
+class KGAttributeEntityOption(str, Enum):
+    FROM_EMAIL = "from_email"  # use email to determine type (ACCOUNT or EMPLOYEE)
```
evan-onyx, line 48: just make this a string constant since there's only one option :)

backend/onyx/kg/utils/extraction_utils.py
```diff
@@ -27,90 +25,84 @@
...
+
+    # Quantity needed for call processing - participants from vendor
+    company_participant_emails: set[str] = set()
+    # Quantity needed for call processing - external participants
+    account_participant_emails: set[str] = set()
 
     # Chunk treatment variables
 
-    document_is_from_call = document_entity_type.lower() in [
+    document_is_from_call = document_entity_type.lower() in {
```
evan-onyx: If you make this a generator expression it's more efficient :D

backend/onyx/kg/utils/extraction_utils.py
```diff
@@ -280,99 +251,46 @@ def _prepare_llm_document_content_call(
...
+        return None
+
+    if kg_person.employee and "EMPLOYEE" in active_entity_types:
+        person_entity = make_entity_id("EMPLOYEE", kg_person.name)
         return (
-            implied_entities,
-            implied_relationships,
-            company_participant_emails,
-            account_participant_emails,
+            person_entity,
```
evan-onyx: the returns are the same in the two cases, imo nicer to combine the two outside the if/elif inside an `if person_entity`

backend/onyx/kg/utils/formatting_utils.py
```diff
@@ -158,6 +157,15 @@ def aggregate_kg_extractions(
     return aggregated_kg_extractions
 
 
+def extract_email(email: str) -> str | None:
+    """
+    Extract an email from an arbitrary string (if any).
+    Only the first email is returned.
+    """
+    match = re.search(r"([A-Za-z0-9._+-]+@[A-Za-z0-9-]+(?:\.[A-Za-z0-9-]+)+)", email)
```
evan-onyx, line 165: hmmmm I think there's a canonical email regex out there that we should be able to import and use; It's gross iirc but we should probably still use it

evan-onyx (approved):
some nits from me

backend/onyx/kg/models.py
Orbital-Web: could potentially have more in the future. I think... If I can't come up with a use case that we would reasonably do, will convert into a const

backend/onyx/kg/utils/formatting_utils.py
Orbital-Web: Looked online, no canonical one from what I can find, just a lot of variations people came up with. Though, a lot of them aligns very closely with the above.

backend/onyx/kg/models.py
Orbital-Web: leaving it to have room for polymorphic relationships in metadata

backend/onyx/kg/extractions/extraction_processing.py
```diff
@@ -730,135 +532,82 @@ def kg_extraction(
...
-                        logger.error(
-                            f"Invalid relationship {relationship} in aggregated_kg_extractions.relationships"
-                        )
-                        continue
+                        if potential_document_id:
+                            event_time = get_document_updated_at(
+                                potential_document_id, db_session
+                            )
+                        else:
+                            event_time = None
```
evan-onyx: nit: just initialize this before the if and then no need for else

PR #4959, opened 2025-06-27
fix: Fix bug in which emails would be fetched during initial indexing

raunakab:
## Description

During initial sync, the `ExternalAccess.external_user_emails` field of a `Document` will always be populated with a set of actual user-emails, regardless of whether that channel is public or not.

During permission-sync, the `ExternalAccess.external_user_emails` will only be populated with the actual emails if that channel is private. Otherwise (if it's public), it'll set that field to the empty set.

This is asymmetric behaviour; we want both initial-indexing + permission-sync to produce identical `ExternalAccess` instances (given that the underlying document's permissions have not changed).

This PR fixes that issue.

Addresses: https://linear.app/danswer/issue/DAN-2150/fix-asymmetric-bug-in-externalaccess-generation-for-perm-sync-vs.

## How Has This Been Tested?

Existing tests are sufficient.

### DIFF
--- a/backend/onyx/connectors/teams/utils.py
+++ b/backend/onyx/connectors/teams/utils.py
@@ -8,12 +8,16 @@ from office365.graph_client import GraphClient  # type: ignore
 from office365.teams.channels.channel import Channel  # type: ignore
 from office365.teams.channels.channel import ConversationMember  # type: ignore
 
+from onyx.access.models import ExternalAccess
 from onyx.connectors.interfaces import SecondsSinceUnixEpoch
 from onyx.connectors.models import BasicExpertInfo
 from onyx.connectors.teams.models import Message
 from onyx.utils.logger import setup_logger
 
 logger = setup_logger()
+
+
+_PUBLIC_MEMBERSHIP_TYPE = "standard"  # public teams channel
 
 
 def _retry(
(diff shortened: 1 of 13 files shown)


backend/onyx/connectors/teams/utils.py
```diff
@@ -64,6 +68,34 @@ def _get_next_url(
...
+    if not isinstance(email, str):
+        logger.warn(f"Expected email to be of type str, instead got {email=}")
+        return None
+
+    return email
+
+
+def _is_channel_public(channel: Channel) -> bool:
+    return (
+        channel.membership_type and channel.membership_type == _PUBLIC_MEMBERSHIP_TYPE
```
Orbital-Web, line 95: why not just channel.membership_type == _PUBLIC_MEMBERSHIP_TYPE?

raunakab: It's possible that `channel.membership_type` is `None`. The way that Teams library works (especially with the number of type-errors that they have written), I wanted to add this check for safety's sake!

backend/onyx/connectors/teams/utils.py
```diff
@@ -164,3 +174,27 @@ def fetch_expert_infos(
...
+    channel: Channel,
+    expert_infos: list[BasicExpertInfo] | None = None,
+) -> ExternalAccess:
+    is_public = _is_channel_public(channel=channel)
+
+    if is_public:
+        return ExternalAccess.public()
+
+    expert_infos = (
+        expert_infos
```
evan-onyx, line 190: `expert_infos or fetch_expert_infos(graph_client=graph_client, channel=channel)`

evan-onyx (approved):
1 nit

backend/onyx/connectors/teams/utils.py
raunakab: You're the goat! 🐐🐐🐐

This is a logical bug. I wanted to originally specify `expert_infos if expert_infos is not None else ..` to explicitly avoid the truthiness check. However, I must have been typing too fast (💨) and forgot in the rush!

Will update in the coming commit!

raunakab: Updated!

PR #4968, opened 2025-06-29
Onyx Metadata Header for File Connector

yuhongsun96:
## Description
Now supports additional override fields such as title, source_type, doc_id

## How Has This Been Tested?
Locally with file connector


Weves (approved):
remove doc_id and we're good!

PR #4989, opened 2025-07-04
Re-adding .epub file support

emerzon:
## Description
.epub files apparently were forgotten and were not allowed for upload in the frontend, even though the backend can process them.

## How Has This Been Tested?
Upload a .epub file


Orbital-Web:
seems good, thanks for the catch!

PR #4996, opened 2025-07-06
Feature/GitHub permission sync

Subash-Mohan:
## Description
This pull request consists of GitHub doc sync, group sync and unit tests.
Added new connector type **CheckpointedSlimConnector** which helps in handling connectors with multiple repositories. The GithubConnector implements this interface and provides the `checkpointed_retrieve_all_slim_documents() `method that allows me to individually check whether each repository has any permission changes one by one. I also created SlimCheckpointOutputWrapper to handle the output of the new connector type's method, which yields tuples of `(SlimDocument | None, ConnectorFailure | None, GithubConnectorCheckpoint | None) `to make the checkpointed output more digestible for the calling code.

This architecture enables efficient processing of multiple repositories by allowing the system to pause and resume at repository boundaries, making it possible to check permission changes for each repository individually without losing progress.

## How Has This Been Tested?
By creating connector and syncing it

### DIFF
--- a/backend/ee/onyx/external_permissions/github/doc_sync.py
+++ b/backend/ee/onyx/external_permissions/github/doc_sync.py
@@ -30,6 +30,8 @@ from onyx.utils.logger import setup_logger
 
 logger = setup_logger()
 
+GITHUB_DOC_SYNC_LABEL = "github_doc_sync"
+
 
 def github_doc_sync(
     cc_pair: ConnectorCredentialPair,
@@ -79,6 +81,11 @@ def github_doc_sync(
 
     # Process documents in batches using checkpointing
     while checkpoint.has_more:
+        # Check for stop signal before processing each batch
+        if callback:
+            if callback.should_stop():
+                raise RuntimeError(f"{GITHUB_DOC_SYNC_LABEL}: Stop signal detected")
+
         logger.info(f"Processing checkpoint batch, has_more: {checkpoint.has_more}")
         slim_doc_generator = github_connector.checkpointed_retrieve_all_slim_documents(
             start=start_time,
@@ -89,6 +96,11 @@ def github_doc_sync(
         for slim_doc, failure, new_checkpoint in SlimCheckpointOutputWrapper[
             GithubConnectorCheckpoint
         ]()(slim_doc_generator):
+            # Check for stop signal before processing each document
+            if callback:
+                if callback.should_stop():
+                    raise RuntimeError(f"{GITHUB_DOC_SYNC_LABEL}: Stop signal detected")
+
             # New checkpoint means we've moved to a different repository
             if new_checkpoint:
                 logger.info(f"Received new checkpoint: {new_checkpoint}")
@@ -108,6 +120,7 @@ def github_doc_sync(
                         github_client=github_connector.github_client,
                         fetch_all_existing_docs_fn=fetch_all_existing_docs_fn,
                         checkpoint=checkpoint,
+                        callback=callback,
                     )
                     continue
                 else:
@@ -117,6 +130,10 @@ def github_doc_sync(
             # Yield document external access records
             if slim_doc and slim_doc.external_access:
                 if isinstance(slim_doc, SlimDocument):
+                    # Report progress for each document processed
+                    if callback:
+                        callback.progress(GITHUB_DOC_SYNC_LABEL, 1)
+
                     yield DocExternalAccess(
                         doc_id=slim_doc.id,
                         external_access=slim_doc.external_access,
@@ -128,6 +145,7 @@ def _check_repo_permission_changes(
     github_client: Github,
     fetch_all_existing_docs_fn: FetchAllDocumentsFunction,
     checkpoint: GithubConnectorCheckpoint,
+    callback: IndexingHeartbeatInterface | None = None,
 ) -> GithubConnectorCheckpoint:
     """
     Check if repository has any permission changes (visibility or team updates).
@@ -135,6 +153,11 @@ def _check_repo_permission_changes(
     Returns:
         Updated checkpoint with next repository to process
     """
+    # Check for stop signal before processing repository
+    if callback:
+        if callback.should_stop():
+            raise RuntimeError(f"{GITHUB_DOC_SYNC_LABEL}: Stop signal detected")
+
     # Check for repository visibility changes, if it's changed,
     # we need to re-sync the repository permissions for all documents in the repository
     if is_repo_visibility_changed(
@@ -174,6 +197,7 @@ def _check_repo_permission_changes(
             github_client=github_client,
             fetch_all_existing_docs_fn=fetch_all_existing_docs_fn,
             checkpoint=checkpoint,
+            callback=callback,
         )
 
     # If no more repositories to process, return the checkpoint
--- a/backend/ee/onyx/external_permissions/github/doc_sync.py
+++ b/backend/ee/onyx/external_permissions/github/doc_sync.py
@@ -16,7 +16,7 @@ from ee.onyx.external_permissions.perm_sync_types import FetchAllDocumentsFuncti
 from onyx.access.models import DocExternalAccess
 from onyx.access.utils import build_ext_group_name_for_onyx
 from onyx.configs.constants import DocumentSource
-from onyx.connectors.connector_runner import CheckpointOutputWrapper
+from onyx.connectors.connector_runner import SlimCheckpointOutputWrapper
 from onyx.connectors.github.connector import GithubConnector
 from onyx.connectors.github.connector import GithubConnectorCheckpoint
 from onyx.connectors.github.models import SerializedRepository
@@ -86,7 +86,7 @@ def github_doc_sync(
             checkpoint=checkpoint,
         )
 
-        for slim_doc, failure, new_checkpoint in CheckpointOutputWrapper[
+        for slim_doc, failure, new_checkpoint in SlimCheckpointOutputWrapper[
             GithubConnectorCheckpoint
         ]()(slim_doc_generator):
             # New checkpoint means we've moved to a different repository
@@ -309,7 +309,6 @@ def is_repo_visibility_changed(
                 [onyx_organization_group_id]
             ),
         )
-        logger.info(f"Organization group ID: {onyx_organization_group_id}  ")
         logger.info(f"Found {len(existing_docs)} existing docs with organization group")
 
         if existing_docs:
--- a/backend/ee/onyx/external_permissions/github/doc_sync.py
+++ b/backend/ee/onyx/external_permissions/github/doc_sync.py
@@ -1,19 +1,30 @@
-import os
 from collections.abc import Generator
 from datetime import datetime
 from datetime import timezone
 
 from github import Github
+from github.Repository import Repository
 
+from ee.onyx.external_permissions.github.utils import fetch_repository_team_slugs
+from ee.onyx.external_permissions.github.utils import form_collaborators_group_id
+from ee.onyx.external_permissions.github.utils import form_organization_group_id
+from ee.onyx.external_permissions.github.utils import (
+    form_outside_collaborators_group_id,
+)
+from ee.onyx.external_permissions.github.utils import get_repository_visibility
 from ee.onyx.external_permissions.perm_sync_types import FetchAllDocumentsFunction
 from onyx.access.models import DocExternalAccess
+from onyx.access.utils import build_ext_group_name_for_onyx
+from onyx.configs.constants import DocumentSource
 from onyx.connectors.connector_runner import CheckpointOutputWrapper
 from onyx.connectors.github.connector import GithubConnector
 from onyx.connectors.github.connector import GithubConnectorCheckpoint
-from onyx.connectors.github.connector import SerializedRepository
+from onyx.connectors.github.models import SerializedRepository
+from onyx.connectors.github.utils import deserialize_repository
 from onyx.connectors.interfaces import CheckpointedSlimConnector
 from onyx.connectors.models import SlimDocument
 from onyx.db.models import ConnectorCredentialPair
+from onyx.db.models import Document
 from onyx.indexing.indexing_heartbeat import IndexingHeartbeatInterface
 from onyx.utils.logger import setup_logger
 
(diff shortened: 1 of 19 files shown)


backend/ee/onyx/external_permissions/github/doc_sync.py
```diff
@@ -0,0 +1,303 @@
...
+            new_external_access = get_external_access_permission(
+                repo, github_connector.github_client
+            )
+
+            repo_documents = fetch_all_existing_docs_fn(
+                columns=["id"],
+                document_filter=DocumentFilter(
+                    field="id",
+                    operation=FilterOperation.LIKE,
+                    value=f"%{repo.full_name}%",
```
Weves: this feels a bit dangerous/fragile. For example, what if the repo name is `pull`, then all pull requests will match (since their URL is like `https://github.com/onyx-dot-app/onyx/pull/5018`). 

Should we add a new metadata JSON column where we can add stuff like this?

backend/ee/onyx/external_permissions/github/doc_sync.py
```diff
@@ -0,0 +1,303 @@
...
+        True if repository has changes that require document permission updates
+    """
+    logger.info(f"Checking repository {repo.id} ({repo.name}) for changes")
+
+    docs = fetch_all_existing_docs_fn(
+        columns=["id", "external_user_group_ids"],
+        document_filter=DocumentFilter(
+            field="id",
+            operation=FilterOperation.LIKE,
+            value=f"%{repo.full_name}%",
```
Weves: same here

backend/ee/onyx/external_permissions/github/doc_sync.py
```diff
@@ -0,0 +1,303 @@
...
+        limit=1,
+    )
+
+    # If no documents exist for this connector/credential pair, no changes to process
+    if not docs:
+        logger.info(f"No documents found for repository {repo.id} ({repo.name})")
+        return False
+
+    doc = docs[0]
+    current_external_group_ids = doc.get("external_user_group_ids", []) or []
```
Weves: is it possible to make this typed? Would love if changes to the model (for example) would get caught by static type checking

backend/onyx/db/document.py
```diff
@@ -210,6 +214,77 @@ def get_document_ids_for_connector_credential_pair(
...
+            DocumentByConnectorCredentialPair.credential_id == credential_id,
+        )
+    )
+
+    where_clause = None
+
+    if document_filter:
+        where_clause = build_where_clause_from_filter(document_filter)
+
+    # If limit is 1, order by last_modified ASC to get the oldest record
```
Weves: this feels like hidden behavior e.g. as the caller you wouldn't expect setting the limit to a certain number to cause sorting to happen

backend/ee/onyx/external_permissions/github/group_sync.py
```diff
@@ -0,0 +1,43 @@
...
+    tenant_id: str,
+    cc_pair: ConnectorCredentialPair,
+) -> Generator[ExternalUserGroup, None, None]:
+    github_connector: GithubConnector = GithubConnector(
+        **cc_pair.connector.connector_specific_config
+    )
+    github_connector.load_credentials(cc_pair.credential.credential_json)
+    if not github_connector.github_client:
+        raise ValueError("github_client is required")
+    github_connector.load_credentials(cc_pair.credential.credential_json)
```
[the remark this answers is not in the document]
Weves: +1

backend/ee/onyx/external_permissions/github/group_sync.py
```diff
@@ -0,0 +1,43 @@
...
+from ee.onyx.external_permissions.github.utils import get_external_user_group
+from onyx.connectors.github.connector import GithubConnector
+from onyx.db.models import ConnectorCredentialPair
+from onyx.utils.logger import setup_logger
+
+logger = setup_logger()
+
+
+def github_group_sync(
+    tenant_id: str,
```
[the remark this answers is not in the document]
Weves: +1

backend/ee/onyx/external_permissions/github/utils.py
```diff
@@ -0,0 +1,482 @@
...
+            raise RuntimeError(error_msg)
+    except GithubException as e:
+        logger.warning(f"GitHub API error during {description}: {e}")
+        return None
+    except Exception as e:
+        logger.exception(f"Unexpected error during {description}: {e}")
+        return None
+
+
+@dataclass
```
Weves: for consistency, we prefer to use pydantic models everywhere

backend/ee/onyx/external_permissions/github/utils.py
```diff
@@ -0,0 +1,482 @@
...
+    org_members: List[UserInfo] = []
+    logger.info(f"Fetching organization members for {org_name}")
+
+    org = _run_with_retry(
+        lambda: github_client.get_organization(org_name),
+        f"get organization {org_name}",
+        github_client,
+    )
+    if not org:
+        logger.warning(f"Failed to fetch organization {org_name}")
```
Weves: should this throw an exception rather than failing silently?

backend/ee/onyx/external_permissions/github/utils.py
```diff
@@ -0,0 +1,447 @@
...
+        if repo.organization:
+            org: Organization | None = _run_with_retry(
+                lambda: github_client.get_organization(repo.organization.login),
+                f"get organization {repo.organization.login}",
+                github_client,
+            )
+
+            if org is not None:
+                org_obj = org  # type: ignore
+                membership = _run_with_retry(
```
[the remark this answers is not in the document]
Weves: Yea, why do we have the `#type: ignore` here?

backend/ee/onyx/external_permissions/github/utils.py
```diff
@@ -0,0 +1,482 @@
...
+    logger.info(
+        f"Generating ExternalAccess for {repo.full_name}: visibility={repo_visibility.value}"
+    )
+
+    if repo_visibility == GitHubVisibility.PUBLIC:
+        logger.info(
+            f"Repository {repo.full_name} is public - allowing access to all users"
+        )
+        return ExternalAccess(
+            external_user_emails=user_emails,
```
Weves: nit: prefer just setting these to `set()` to make it clearer what the value is

backend/ee/onyx/external_permissions/github/utils.py
```diff
@@ -0,0 +1,482 @@
...
+                f"Unknown visibility '{visibility}' for repo {repo.full_name}, defaulting to private"
+            )
+            return GitHubVisibility.PRIVATE
+
+    logger.info(f"Repository {repo.full_name} is private")
+    return GitHubVisibility.PRIVATE
+
+
+def get_external_access_permission(
+    repo: Repository, github_client: Github, add_prefix: bool = False
```
Weves, line 298: why would I set add_prefix or not? That's not clear to me 🧐

Weves: Would be great to add a comment / remove it if not necesssary

backend/ee/onyx/external_permissions/github/utils.py
```diff
@@ -0,0 +1,482 @@
...
+            collaborators_group_id = build_ext_group_name_for_onyx(
+                source=DocumentSource.GITHUB,
+                ext_group_name=collaborators_group_id,
+            )
+            outside_collaborators_group_id = build_ext_group_name_for_onyx(
+                source=DocumentSource.GITHUB,
+                ext_group_name=outside_collaborators_group_id,
+            )
+
+        group_ids.add(collaborators_group_id)
```
Weves: could we remove the initialization at the top and just have:

```
group_ids = {collaborators_group_id, outside_collaborators_group_id}
```

backend/ee/onyx/external_permissions/github/utils.py
```diff
@@ -0,0 +1,482 @@
...
+        logger.info(
+            f"Repository {repo.full_name} is internal - accessible to org members"
+        )
+        org_group_id = form_organization_group_id(repo.organization.id)
+        if add_prefix:
+            org_group_id = build_ext_group_name_for_onyx(
+                source=DocumentSource.GITHUB,
+                ext_group_name=org_group_id,
+            )
+        group_ids.add(org_group_id)
```
Weves: same comment here

backend/ee/onyx/external_permissions/github/utils.py
```diff
@@ -0,0 +1,482 @@
...
+                id=form_collaborators_group_id(repo.id),
+                user_emails=list(user_emails),
+            )
+            external_user_groups.append(collaborators_group)
+            logger.info(f"Created collaborators group with {len(user_emails)} emails")
+
+        # Create group for outside collaborators
+        user_emails = set()
+        for collab in outside_collaborators:
+            if collab.email:
```
Weves, line 424: we should log an error if we can't get their email

backend/ee/onyx/external_permissions/github/utils.py
```diff
@@ -0,0 +1,482 @@
...
+
+        collaborators, outside_collaborators = (
+            _get_collaborators_and_outside_collaborators(github_client, repo)
+        )
+        teams = _fetch_repository_teams_detailed(repo, github_client)
+        external_user_groups = []
+
+        user_emails = set()
+        for collab in collaborators:
+            if collab.email:
```
Weves, line 408: we should log an error if we can't get their email

backend/ee/onyx/external_permissions/github/utils.py
```diff
@@ -0,0 +1,482 @@
...
+            external_user_groups.append(outside_collaborators_group)
+            logger.info(
+                f"Created outside collaborators group with {len(user_emails)} emails"
+            )
+
+        # Create groups for teams
+        for team in teams:
+            user_emails = set()
+            for member in team.members:
+                if member.email:
```
Weves, line 443: same here

backend/ee/onyx/external_permissions/github/utils.py
```diff
@@ -0,0 +1,482 @@
...
+        logger.info(f"Processing internal repository {repo.full_name}")
+
+        org_group_id = form_organization_group_id(repo.organization.id)
+        org_members = _fetch_organization_members(
+            github_client, repo.organization.login
+        )
+
+        user_emails = set()
+        for member in org_members:
+            if member.email:
```
Weves, line 473: same here

backend/ee/onyx/external_permissions/utils.py
```diff
@@ -62,7 +63,10 @@ def generic_doc_sync(
             )
 
     logger.info(f"Querying existing document IDs for CC Pair ID: {cc_pair.id=}")
-    existing_doc_ids = set(fetch_all_existing_docs_fn())
+    existing_docs: list[dict[str, Any]] = fetch_all_existing_docs_fn(
```
Weves: typing would be great here

Weves (changes requested):

backend/ee/onyx/external_permissions/github/doc_sync.py
Subash-Mohan: I initially considered creating a new field, and using an ID seemed like the better choice. But you're right — your point makes sense. We can go ahead and create a new field.

backend/ee/onyx/external_permissions/github/utils.py
Subash-Mohan: #type: ingore is not needed, just removed it and also change now the public repo returns empty set directly instead of assigning a variable.

backend/ee/onyx/external_permissions/github/utils.py
Subash-Mohan: Changed to pydantic model and removed the extra load_credentials method call.

backend/ee/onyx/external_permissions/github/utils.py
Subash-Mohan: When this method is called during the initial permission sync via the connector, the group ID isn't prefixed with the source while inserting the document record. So in that case, I'll set `add_prefix` to `true`, allowing the method itself to handle prefixing. However, when the same method is invoked from `doc_sync`, our system already adds the prefix to the group ID while processing the `externalaccess` object.

backend/onyx/db/document.py
Subash-Mohan: Added a new parameter called `sort_order`, which allows us to specify whether to sort the records in ascending or descending order. If no value is provided, the records won't be sorted.

backend/ee/onyx/external_permissions/github/doc_sync.py
Subash-Mohan: I don't think we can enforce static typing for every column name change in the `Document`. However, I've now added a return type to `fetch_all_existing_docs_fn` as `DocumentColumns`. So, if we update any field name and reflect that change in the `DocumentColumns` type, the static type checker will catch all the places where those fields are used in doc permission sync.

backend/ee/onyx/external_permissions/utils.py
Subash-Mohan: Changed the type as list[dict[DocumentColumns, Any]]

backend/ee/onyx/external_permissions/github/utils.py
Subash-Mohan: Now we are raising exception here

backend/ee/onyx/external_permissions/github/doc_sync.py
Weves: in that case, would prefer to make this less flexible but strictly typed (don't like the `Any`). E.g. we could just provide `fetch_all_existing_docs_fn` + `fetch_all_existing_doc_ids_fn`, and the first would fetch a well-specified number of fields and the second would just fetch IDs (lightweight).

Especially in permission sync, where the logic is quite complicated and errors are hard to catch, I would prefer to  have maximum static-time checks.

backend/ee/onyx/external_permissions/github/group_sync.py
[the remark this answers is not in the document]
Weves: @subash-onyx 🧐

backend/ee/onyx/external_permissions/github/utils.py
Weves: Got it, thanks. Can we add a backlog item in Linear to clean this up? It feels like we should auto-prefix in both cases

backend/onyx/connectors/github/connector.py
```diff
@@ -893,27 +909,33 @@ def build_dummy_checkpoint(self) -> GithubConnectorCheckpoint:
...
-    # Create a time range from epoch to now
-    end_time = datetime.now(timezone.utc)
-    start_time = datetime.fromtimestamp(0, tz=timezone.utc)
-    time_range = (start_time, end_time)
-
-    # Initialize the runner with a batch size of 10
-    runner: ConnectorRunner[GithubConnectorCheckpoint] = ConnectorRunner(
-        connector, batch_size=10, include_permissions=False, time_range=time_range
-    )
+    if connector.github_client:
```
Weves, line 918: should probably remove this / have it beneath the below, since that standard indexing is the more standard/critical functionality

backend/onyx/connectors/github/rate_limit_utils.py
```diff
@@ -0,0 +1,25 @@
...
+
+    Args:
+        github_client: The GitHub client that hit the rate limit
+    """
+    sleep_time = github_client.get_rate_limit().core.reset.replace(
+        tzinfo=timezone.utc
+    ) - datetime.now(tz=timezone.utc)
+    sleep_time += timedelta(minutes=1)  # add an extra minute just to be safe
+    logger.notice(f"Ran into Github rate-limit. Sleeping {sleep_time.seconds} seconds.")
+    time.sleep(sleep_time.seconds)
```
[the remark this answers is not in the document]
Weves: @subash-onyx  +1

backend/ee/onyx/external_permissions/github/group_sync.py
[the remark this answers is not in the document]
Subash-Mohan: Its an required parameter for group sync function.

backend/ee/onyx/external_permissions/github/doc_sync.py
Subash-Mohan: Now I added two parameters as we discussed. And a pydantic class as return type. Also I created another pydantic class for doc metadata as well.

backend/ee/onyx/external_permissions/github/utils.py
Subash-Mohan: task created - https://linear.app/danswer/issue/DAN-2222/auto-prefix-group-id-in-document-record-insertion

backend/alembic/versions/73a4e89857e9_merge_heads.py
```diff
@@ -0,0 +1,21 @@
+"""Merge heads
```
Weves: @subash-onyx rather than having this extra revision, can we just rebase the old revision on top of the latest revision?

Subash-Mohan: The change has been done.

PR #4998, opened 2025-07-07
Bugfix/drive doc ids3

evan-onyx:
## Description

Finalized the migration that fixes duplicate drive ids
- cursor-based pagination for the vespa visit API
- added deletes of missing rows that had foreign keys on the deleted docs

## How Has This Been Tested?

tested in UI

### DIFF
--- a/backend/alembic/versions/12635f6655b7_drive_canonical_ids.py
+++ b/backend/alembic/versions/12635f6655b7_drive_canonical_ids.py
@@ -18,8 +18,6 @@ from onyx.document_index.vespa.shared_utils.utils import (
     replace_invalid_doc_id_characters,
 )
 from onyx.document_index.vespa_constants import DOCUMENT_ID_ENDPOINT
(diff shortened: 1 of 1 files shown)


backend/alembic/versions/12635f6655b7_drive_canonical_ids.py
```diff
@@ -123,9 +126,9 @@ def update_document_id_in_database(old_doc_id: str, new_doc_id: str) -> None:
     )
     row = result.fetchone()
     if row and row[0] > 0:
-        raise RuntimeError(
-            f"Document with ID {new_doc_id} already exists, cannot create duplicate"
-        )
+        # print(f"Document with ID {new_doc_id} already exists, deleting old one")
```
Weves, line 129: nit: add back prints (or remove them)?

evan-onyx: they're pretty spammy, but I'm anticipating having to debug a migration or two in the future :/

PR #5021, opened 2025-07-13
Test matrix

Weves:
## Description

## How Has This Been Tested?

### DIFF
--- a/.github/workflows/pr-mit-integration-tests.yml
+++ b/.github/workflows/pr-mit-integration-tests.yml
@@ -20,8 +20,38 @@ env:
   JIRA_USER_EMAIL: ${{ secrets.JIRA_USER_EMAIL }}
   JIRA_API_TOKEN: ${{ secrets.JIRA_API_TOKEN }}
   PLATFORM_PAIR: linux-amd64
+
 jobs:
+  discover-test-dirs:
+    runs-on: ubuntu-latest
+    outputs:
+      test-dirs: ${{ steps.set-matrix.outputs.test-dirs }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      
+      - name: Discover test directories
+        id: set-matrix
+        run: |
+          # Find all leaf-level directories in both test directories
+          tests_dirs=$(find backend/tests/integration/tests -mindepth 1 -maxdepth 1 -type d ! -name "__pycache__" -exec basename {} \; | sort)
+          connector_dirs=$(find backend/tests/integration/connector_job_tests -mindepth 1 -maxdepth 1 -type d ! -name "__pycache__" -exec basename {} \; | sort)
+          
+          # Create JSON array with directory info
+          all_dirs=""
+          for dir in $tests_dirs; do
+            all_dirs="$all_dirs{\"path\":\"tests/$dir\",\"name\":\"tests-$dir\"},"
+          done
+          for dir in $connector_dirs; do
+            all_dirs="$all_dirs{\"path\":\"connector_job_tests/$dir\",\"name\":\"connector-$dir\"},"
+          done
+          
+          # Remove trailing comma and wrap in array
+          all_dirs="[${all_dirs%,}]"
+          echo "test-dirs=$all_dirs" >> $GITHUB_OUTPUT
+
   integration-tests-mit:
+    needs: discover-test-dirs
     # See https://runs-on.com/runners/linux/
     runs-on:
       [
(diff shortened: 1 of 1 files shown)


justin-tahara:
@greptileai

.github/workflows/pr-mit-integration-tests.yml
```diff
@@ -11,6 +11,13 @@ on:
       - "release/**"
 
 env:
+  # AWS ECR Configuration. Use us-west-2 since blacksmith is located there.
```
justin-tahara: Maybe clarify that they themselves are not located their but their EC2 instances are in that region

.github/workflows/pr-mit-integration-tests.yml
```diff
@@ -23,21 +30,42 @@ env:
...
-      [
-        runs-on,
-        runner=32cpu-linux-x64,
-        disk=large,
-        "run-id=${{ github.run_id }}",
-      ]
+  discover-test-dirs:
+    runs-on: blacksmith-2vcpu-ubuntu-2404
+    outputs:
+      test-dirs: ${{ steps.set-matrix.outputs.test-dirs }}
```
justin-tahara: This is so simple to convert

.github/workflows/pr-mit-integration-tests.yml
```diff
@@ -70,71 +100,160 @@ jobs:
...
         with:
-          username: ${{ secrets.DOCKER_USERNAME }}
-          password: ${{ secrets.DOCKER_TOKEN }}
+          name: openapi-artifacts
+          path: backend/generated/
 
-      # tag every docker image with "test" so that we can spin up the correct set
-      # of images during testing
+  build-backend-image:
+    runs-on: blacksmith-16vcpu-ubuntu-2404
```
justin-tahara: Are we saying we need a larger instance because of how much CPU cores are needed for building the image?

.github/workflows/pr-mit-integration-tests.yml
```diff
@@ -70,71 +100,160 @@ jobs:
...
           platforms: linux/amd64
-          tags: onyxdotapp/onyx-backend:test
-          push: false
-          load: true
-          cache-from: type=s3,prefix=cache/${{ github.repository }}/mit-integration-tests/backend-${{ env.PLATFORM_PAIR }}/,region=${{ env.RUNS_ON_AWS_REGION }},bucket=${{ env.RUNS_ON_S3_BUCKET_CACHE }}
-          cache-to: type=s3,prefix=cache/${{ github.repository }}/mit-integration-tests/backend-${{ env.PLATFORM_PAIR }}/,region=${{ env.RUNS_ON_AWS_REGION }},bucket=${{ env.RUNS_ON_S3_BUCKET_CACHE }},mode=max
-
-      - name: Build Model Server Docker image
-        uses: ./.github/actions/custom-build-and-push
+          tags: ${{ env.ECR_REGISTRY }}/integration-test-onyx-backend:test-${{ github.run_id }}
```
justin-tahara, line 137: How does this ECR Registry get updated with our backend images? Do we need to do this for all images and move them over to ECR?

.github/workflows/pr-mit-integration-tests.yml
```diff
@@ -70,71 +100,160 @@ jobs:
...
+
+  integration-tests-mit:
+    needs:
+      [
+        discover-test-dirs,
+        build-backend-image,
+        build-model-server-image,
+        build-integration-image,
+      ]
+    # See https://runs-on.com/runners/linux/
```
justin-tahara: Can we update this comment?

Weves: Deon!

.github/workflows/pr-mit-integration-tests.yml
Weves: Updated!

.github/workflows/pr-mit-integration-tests.yml
Weves: we also build them + push them in this action. For our public images, we can still use dockerhub.

It's nicer imo to just use one container repo across the board, but dockerhub seems to be throttling us for these high bandwidth use cases :(

.github/workflows/pr-mit-integration-tests.yml
Weves: I think we used to use `32cpu-linux-x64`, so this is smaller. Not sure if we actually need this large of an instance though 🤔

PR #5028, opened 2025-07-14
[DAN-2028] Fix search time filter - No End Time Set

trial-danswer:
## Description
https://linear.app/danswer/issue/DAN-2028/fix-search-time-filter-no-end-time-set

## How Has This Been Tested?

Uploading Screen Recording 2025-07-14 at 4.19.18 PM.mov…

1. Tested locally
2. Added a unit test for vespa filters

---

## Summary by cubic
Added support for an end date filter (time_cutoff_end) in search filters to allow users to specify a date range for document retrieval. This enables more precise filtering of search results by both start and end dates.

### DIFF
--- a/backend/onyx/context/search/models.py
+++ b/backend/onyx/context/search/models.py
@@ -109,7 +109,7 @@ class Tag(BaseModel):
 class BaseFilters(BaseModel):
     source_type: list[DocumentSource] | None = None
     document_set: list[str] | None = None
-    time_cutoff_start: datetime | None = None
+    time_cutoff: datetime | None = None
     time_cutoff_end: datetime | None = None
     tags: list[Tag] | None = None
     kg_entities: list[str] | None = None
@@ -123,14 +123,12 @@ class BaseFilters(BaseModel):
     def validate_time_range(
         cls, time_cutoff_end: datetime | None, info: Any
     ) -> datetime | None:
-        """Validate that time_cutoff_end is after time_cutoff_start when both are provided."""
-        time_cutoff_start = (
-            info.data.get("time_cutoff_start") if hasattr(info, "data") else None
-        )
+        """Validate that time_cutoff_end is after time_cutoff when both are provided."""
+        time_cutoff = info.data.get("time_cutoff") if hasattr(info, "data") else None
 
-        if time_cutoff_start is not None and time_cutoff_end is not None:
-            if time_cutoff_end < time_cutoff_start:
-                raise ValueError("time_cutoff_end must be after time_cutoff_start")
+        if time_cutoff is not None and time_cutoff_end is not None:
+            if time_cutoff_end < time_cutoff:
+                raise ValueError("time_cutoff_end must be after time_cutoff")
 
         return time_cutoff_end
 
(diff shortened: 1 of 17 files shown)


web/src/lib/search/interfaces.ts
```diff
@@ -195,6 +195,7 @@ export interface Filters {
   source_type: string[] | null;
   document_set: string[] | null;
   time_cutoff: Date | null;
+  time_cutoff_end: Date | null;
```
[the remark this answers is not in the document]
trial-danswer: not changing in this PR due to backwards compatibility

web/src/lib/search/utils.ts
```diff
@@ -15,6 +15,7 @@ export const buildFilters = (
       sources.length > 0 ? sources.map((source) => source.internalName) : null,
     document_set: documentSets.length > 0 ? documentSets : null,
     time_cutoff: timeRange?.from ? timeRange.from : null,
+    time_cutoff_end: timeRange?.to ? timeRange.to : null,
```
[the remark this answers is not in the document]
trial-danswer: not changing in this PR due to backwards compatibility

backend/onyx/context/search/models.py
```diff
@@ -110,13 +110,28 @@ class BaseFilters(BaseModel):
...
     kg_sources: list[str] | None = None
     kg_chunk_id_zero_only: bool | None = False
 
+    @field_validator("time_cutoff_end")
+    @classmethod
+    def validate_time_range(
+        cls, time_cutoff_end: datetime | None, info: Any
+    ) -> datetime | None:
+        """Validate that time_cutoff_end is after time_cutoff when both are provided."""
+        time_cutoff = info.data.get("time_cutoff") if hasattr(info, "data") else None
```
evan-onyx: probably should be a model validator since this is verifying properties across two fields of the model

backend/onyx/context/search/preprocessing/preprocessing.py
```diff
@@ -71,9 +71,11 @@ def retrieval_preprocessing(
             document_set.name for document_set in persona.document_sets
         ]
 
-    time_filter = preset_filters.time_cutoff
-    if time_filter is None and persona:
-        time_filter = persona.search_start_date
+    time_filter_start = preset_filters.time_cutoff
+    if time_filter_start is None and persona:
+        time_filter_start = persona.search_start_date
```
evan-onyx, line 76: probably should pick the max of the two times if the persona has a time set

backend/onyx/context/search/preprocessing/preprocessing.py
```diff
@@ -91,7 +93,7 @@ def retrieval_preprocessing(
         logger.debug("Auto detect filters enabled")
 
     if (
-        time_filter is not None
```
evan-onyx, line 94: could you add a TODO to extract a time filter end in the extract source filter function?

evan-onyx (approved):
left some nits, but mostly looking good!

backend/onyx/context/search/preprocessing/preprocessing.py
trial-danswer: keeping the same for now because this was the functionality before (it looks like +/- just because I renamed time_filter to time_filter_start)

PR #5035, opened 2025-07-16
fix: Fix Confluence pagination

raunakab:
## Description

This PR uses index-based pagination to retrieve all the paginated queries that are available. https://developer.atlassian.com/cloud/confluence/rest/v1/api-group-search/#api-wiki-rest-api-search-user-get.

Addresses: https://linear.app/danswer/issue/DAN-2170/investigate-confluence-user-pagination.

## How Has This Been Tested?

Wrote a test for group-sync. All existing Confluence tests continue to pass.

### DIFF
--- a/backend/onyx/connectors/confluence/onyx_confluence.py
+++ b/backend/onyx/connectors/confluence/onyx_confluence.py
@@ -46,13 +46,11 @@ _REPLACEMENT_EXPANSIONS = "body.view.value"
 _USER_NOT_FOUND = "Unknown Confluence User"
 _USER_ID_TO_DISPLAY_NAME_CACHE: dict[str, str | None] = {}
 _USER_EMAIL_CACHE: dict[str, str | None] = {}
+_DEFAULT_PAGINATION_LIMIT = 6
 
 
 class ConfluenceRateLimitError(Exception):
     pass
-
-
-_DEFAULT_PAGINATION_LIMIT = 1000
 
 
 class OnyxConfluence:
@@ -463,6 +461,7 @@ class OnyxConfluence:
         limit: int | None = None,
         # Called with the next url to use to get the next page
         next_page_callback: Callable[[str], None] | None = None,
+        force_offset_pagination: bool = False,
     ) -> Iterator[dict[str, Any]]:
         """
         This will paginate through the top level query.
@@ -571,6 +570,12 @@ class OnyxConfluence:
                         )
                     # notify the caller of the new url
                     next_page_callback(url_suffix)
+
+                elif force_offset_pagination and i == len(results) - 1:
+                    url_suffix = update_param_in_path(
+                        old_url_suffix, "start", str(updated_start)
+                    )
+
                 yield result
 
             # we've observed that Confluence sometimes returns a next link despite giving
@@ -668,7 +673,9 @@ class OnyxConfluence:
             url = "rest/api/search/user"
             expand_string = f"&expand={expand}" if expand else ""
             url += f"?cql={cql}{expand_string}"
-            for user_result in self._paginate_url(url, limit):
+            for user_result in self._paginate_url(
+                url, limit, force_offset_pagination=True
+            ):
                 # Example response:
                 # {
                 #     'user': {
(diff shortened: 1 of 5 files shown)


backend/onyx/connectors/confluence/onyx_confluence.py
```diff
@@ -46,15 +60,13 @@
 _USER_NOT_FOUND = "Unknown Confluence User"
 _USER_ID_TO_DISPLAY_NAME_CACHE: dict[str, str | None] = {}
 _USER_EMAIL_CACHE: dict[str, str | None] = {}
+_DEFAULT_PAGINATION_LIMIT = 6
```
evan-onyx: probably should go back to 1000

backend/onyx/connectors/confluence/onyx_confluence.py
```diff
@@ -571,6 +584,12 @@ def _paginate_url(
                         )
                     # notify the caller of the new url
                     next_page_callback(url_suffix)
+
+                elif force_offset_pagination and i == len(results) - 1:
```
evan-onyx, line 606: leave a comment explaining why we cant just yield from results since this tripped us up

backend/tests/daily/connectors/confluence/test_confluence_group_sync.py
```diff
@@ -0,0 +1,102 @@
...
+        gives_anyone_access=False,
+    ),
+]
+
+
+def test_confluence_group_sync(
+    confluence_connector_config: dict[str, Any],
+    confluence_credential_json: dict[str, Any],
+) -> None:
+    SqlEngine.init_engine(pool_size=10, max_overflow=10)
```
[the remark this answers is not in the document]
evan-onyx: agree with greptile, we should have code for this already

backend/tests/daily/connectors/confluence/test_confluence_permissions_basic.py
```diff
@@ -151,48 +149,10 @@ def test_confluence_connector_restriction_handling(
...
+    root_page = next(d for d in doc_access_list if d.doc_id.endswith("Root+Page+2"))
     assert (
         root_page.external_access.external_user_emails == restricted_emails
     ), "Root page emails do not match expected values"
     assert (
         root_page.external_access.external_user_group_ids == restricted_user_groups
     ), "Root page groups do not match expected values"
-
-    # check child page has restriction propagated
-    child_page = next(d for d in doc_access_list if d.doc_id.endswith("Child+Page"))
```
evan-onyx: why are all these checks gone?

evan-onyx (approved):
some things to address but if it's fine, it's fine

backend/tests/daily/connectors/confluence/test_confluence_permissions_basic.py
raunakab: I was using the incorrect username/app-token, and all the tests were failing on my side. I was only getting 2 pages back instead of 7. I didn't assume permissions privileges (or the lack thereof) because I was using the `admin@onyx-test.com` credentials, and assumed they would have access to everything.

I assumed that someone had removed the test pages but had forgotten to update the tests, so I removed these extra checks. Then I later found out the credentials that I was using were not able to access those pages, so I was permission-restricted to 2 pages only. The credentials in Github Actions have the full permissions, so that's why the tests succeed there.

I've added these checks back!

backend/onyx/connectors/confluence/onyx_confluence.py
raunakab: Reverted. I made a note to do this, but looks like I forgot!

backend/onyx/connectors/confluence/onyx_confluence.py
raunakab: Added.

backend/tests/daily/connectors/confluence/test_confluence_group_sync.py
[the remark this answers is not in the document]
raunakab: Updated.

PR #5042, opened 2025-07-17
fix: improve check for indexing status

Weves:
## Description

Fixes https://linear.app/danswer/issue/DAN-2215/improve-performance-of-check-for-indexing-check-for-vespa-sync-task

## How Has This Been Tested?

Tested locally that indexing + doc sync works

---

## Summary by cubic
Improved the performance and reliability of indexing and Vespa sync checks by simplifying document sync tracking and removing legacy Redis classes.

- **Refactors**
  - Replaced RedisConnectorCredentialPair and related classes with a new document sync tracking module.
  - Updated indexing and sync tasks to use simpler, global document sync logic.
  - Reduced database queries and removed unused code for connector credential pairs.

### DIFF
--- a/backend/onyx/background/celery/tasks/indexing/tasks.py
+++ b/backend/onyx/background/celery/tasks/indexing/tasks.py
@@ -54,6 +54,7 @@ from onyx.configs.constants import OnyxRedisLocks
 from onyx.configs.constants import OnyxRedisSignals
 from onyx.connectors.exceptions import ConnectorValidationError
 from onyx.db.connector import mark_ccpair_with_indexing_trigger
+from onyx.db.connector_credential_pair import ConnectorType
 from onyx.db.connector_credential_pair import (
     fetch_indexable_connector_credential_pair_ids,
 )
@@ -87,6 +88,8 @@ from shared_configs.configs import MULTI_TENANT
 from shared_configs.configs import SENTRY_DSN
 
 logger = setup_logger()
+
+USER_FILE_INDEXING_LIMIT = 100
 
 
 def _get_fence_validation_block_expiration() -> int:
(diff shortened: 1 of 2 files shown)


backend/onyx/configs/app_configs.py
```diff
@@ -329,7 +329,7 @@
 )
 
 # The maximum number of tasks that can be queued up to sync to Vespa in a single pass
-VESPA_SYNC_MAX_TASKS = 1024
+VESPA_SYNC_MAX_TASKS = 8192
```
Weves, line 332: increased since we're not going cc-pair by cc-pair anymore

backend/onyx/redis/redis_utils.py
```diff
@@ -11,8 +8,6 @@
 
 def is_fence(key_bytes: bytes) -> bool:
     key_str = key_bytes.decode("utf-8")
-    if key_str == RedisGlobalConnectorCredentialPair.FENCE_KEY:
```
evan-onyx, line 14: is this fence fully gone now? no need to check for DOCUMENT_SYNC_FENCE_KEY?

backend/onyx/background/celery/tasks/indexing/tasks.py
```diff
@@ -480,20 +482,26 @@ def check_for_indexing(self: Task, *, tenant_id: str) -> int | None:
...
+        # gather cc_pair_ids + current search settings
         lock_beat.reacquire()
-        cc_pair_ids: list[int] = []
         with get_session_with_current_tenant() as db_session:
-            cc_pairs = fetch_connector_credential_pairs(
-                db_session, include_user_files=True
-            )
-            for cc_pair_entry in cc_pairs:
-                cc_pair_ids.append(cc_pair_entry.id)
+            cc_pair_ids = fetch_indexable_connector_credential_pair_ids(db_session)
```
evan-onyx: we discussed restricting the number of user files we're looking over, any reason you decided against that?

evan-onyx (approved):
some qs

backend/onyx/redis/redis_utils.py
Weves: yep

backend/onyx/background/celery/tasks/indexing/tasks.py
Weves: added

Weves: main concern is that if we change the behavior of user file indexing, it could have a side effect here

PR #5058, opened 2025-07-22
fix: handle empty doc batches

evan-onyx:
## Description

Fixes https://linear.app/danswer/issue/DAN-2237/empty-doc-batches
Previously connectors that retrieved no documents wouldn't ever finish because docprocessing always expected batches with at least one doc. Now batches with 0 docs are not issued, allowing 0-doc attempts to complete.

## How Has This Been Tested?

tested in UI

---

## Summary by cubic
Fixed an issue where connectors with no documents would hang instead of completing. Now, empty document batches are skipped so 0-document attempts finish as expected.

### DIFF
--- a/.vscode/launch.template.jsonc
+++ b/.vscode/launch.template.jsonc
@@ -430,7 +430,7 @@
       },
       "args": [
         "--filename",
-        "generated/openapi.json",
+        "generated/openapi.json"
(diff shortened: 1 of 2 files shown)


raunakab (approved):
Fire

PR #5060, opened 2025-07-22
Fix: too many internet chunks

wenxi-onyx:
## Description

- Limit internet search result chunks to a settable parameter

## How Has This Been Tested?

Locally

### DIFF
--- a/backend/onyx/tools/tool_implementations/internet_search/internet_search_tool.py
+++ b/backend/onyx/tools/tool_implementations/internet_search/internet_search_tool.py
@@ -26,7 +26,6 @@ from onyx.connectors.models import TextSection
 from onyx.context.search.enums import SearchType
 from onyx.context.search.models import InferenceChunk
 from onyx.context.search.models import InferenceSection
-from onyx.context.search.utils import inference_section_from_chunks
 from onyx.db.models import Persona
 from onyx.db.search_settings import get_current_search_settings
 from onyx.indexing.chunker import Chunker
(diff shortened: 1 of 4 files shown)


backend/onyx/tools/tool_implementations/internet_search/internet_search_tool.py
```diff
@@ -96,11 +95,13 @@ def __init__(
         answer_style_config: AnswerStyleConfig,
         provider: str | None = None,
         num_results: int = 10,
+        max_chunks: int = 50,
```
evan-onyx: would prefer having this stored in a shared constant value; makes it easier for people who want to change this to find it without feeling like they have to dive into our code

backend/onyx/tools/tool_implementations/internet_search/internet_search_tool.py
```diff
@@ -333,37 +334,23 @@ def _combine_chunks_into_sections(
...
-                center_chunk = doc_chunks[0]
-            else:
-                center_chunk = max(doc_chunks, key=lambda x: x.score or 0)
-
-            # Create section using the utility function
-            section = inference_section_from_chunks(
-                center_chunk=center_chunk,
-                chunks=sorted_chunks,
+        for inference_chunk in sorted_inference_chunks:
+            new_section = InferenceSection(
```
evan-onyx, line 350: leave comment about how prune_and_merge is called after this to handle coalescing sections

backend/onyx/tools/tool_implementations/internet_search/internet_search_tool.py
wenxi-onyx: done

backend/onyx/tools/tool_implementations/internet_search/internet_search_tool.py
wenxi-onyx: done

evan-onyx (approved):

PR #5065, opened 2025-07-23
fix: sharepoint lg files issue

Subash-Mohan:
## Description

This pull request introduces enhancements to the SharePoint connector in the `backend/onyx` package, focusing on rate-limiting handling, file size threshold enforcement, and retry logic for SharePoint queries. 
Linear link -> https://linear.app/danswer/issue/DAN-2245/sharepoint-connector-failing

## How Has This Been Tested?
By creating connector

---

## Summary by cubic
Improved the SharePoint connector by adding file size checks, rate limit handling, and retry logic to prevent errors and skip large files.

- **Bug Fixes**
  - Skips SharePoint files larger than 20MB.
  - Retries SharePoint queries on rate limit or server errors, with exponential backoff.

### DIFF
--- a/backend/onyx/connectors/sharepoint/connector.py
+++ b/backend/onyx/connectors/sharepoint/connector.py
@@ -112,6 +112,10 @@ def _convert_driveitem_to_document(
 
     # Proceed with download if size is acceptable or not available
     content = _sleep_and_retry(driveitem.get_content(), "get_content")
+    if content is None:
+        logger.warning(f"Could not access content for '{driveitem.name}'")
+        return None
+
     file_text = extract_file_text(
         file=io.BytesIO(content.value),
         file_name=driveitem.name,
--- a/backend/onyx/connectors/sharepoint/connector.py
+++ b/backend/onyx/connectors/sharepoint/connector.py
@@ -1,5 +1,6 @@
 import io
 import os
+import time
 from collections.abc import Generator
 from datetime import datetime
 from datetime import timezone
@@ -11,6 +12,7 @@ from office365.graph_client import GraphClient  # type: ignore
 from office365.onedrive.driveitems.driveItem import DriveItem  # type: ignore
 from office365.onedrive.sites.site import Site  # type: ignore
 from office365.onedrive.sites.sites_with_root import SitesWithRoot  # type: ignore
+from office365.runtime.client_request import ClientRequestException
 from pydantic import BaseModel
 
 from onyx.configs.app_configs import INDEX_BATCH_SIZE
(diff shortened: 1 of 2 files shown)


backend/onyx/connectors/sharepoint/connector.py
```diff
@@ -46,12 +49,71 @@ class SiteDescriptor(BaseModel):
     folder_path: str | None
 
 
+def _sleep_and_retry(query_obj: Any, method_name: str, max_retries: int = 3) -> Any:
+    """
+    Execute a SharePoint query with retry logic for rate limiting.
+    """
+    for attempt in range(max_retries + 1):
+        try:
+            return query_obj.execute_query()
```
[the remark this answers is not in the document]
Subash-Mohan: Added a none check before using the result from the method

backend/onyx/connectors/sharepoint/connector.py
```diff
@@ -46,12 +49,75 @@ class SiteDescriptor(BaseModel):
...
+            else:
+                # Either not a rate limit error, or we've exhausted retries
+                if e.response and e.response.status_code == 429:
+                    logger.error(
+                        f"Rate limit retry exhausted for {method_name} after {max_retries} attempts"
+                    )
+                raise e
+
+    # This should never be reached, but included for completeness
+    raise RuntimeError(f"Unexpected end of retry loop for {method_name}")
```
raunakab: Small nit, but:

Is the Python LSP throwing an error in your IDE? If not, then we should maybe consider removing these lines since they'll never be hit.

If you want to keep these lines for completeness, we could then remove the `else:` case (on line 77) and move that logic outside of the for-loop. Therefore, if the exception is hit and `attempt >= max_retries`, we exit the for-loop and hit this tail-end of the algorithm.

backend/onyx/connectors/sharepoint/connector.py
```diff
@@ -46,12 +49,75 @@ class SiteDescriptor(BaseModel):
...
+            logger.warning(
+                f"Could not access file size for '{driveitem.name}' Proceeding with download."
+            )
+    except (ValueError, TypeError, AttributeError) as e:
+        logger.info(
+            f"Could not access file size for '{driveitem.name}': {e}. Proceeding with download."
+        )
+
+    # Proceed with download if size is acceptable or not available
+    content = _sleep_and_retry(driveitem.get_content(), "get_content")
```
raunakab, line 111: It's possible that `_sleep_and_retry` throws an error, correct? When the exception is caught and `attempt >= max_retries`, which then raises an error (on line 83).

Should we catch that? Or did you want to purposely let that exception bubble up?

backend/onyx/connectors/sharepoint/connector.py
```diff
@@ -46,12 +49,75 @@ class SiteDescriptor(BaseModel):
     folder_path: str | None
 
 
+def _sleep_and_retry(query_obj: Any, method_name: str, max_retries: int = 3) -> Any:
```
raunakab, line 52: This sleep + retry logic is nice. I don't believe we have a common utility function for this, so maybe in a future PR, we should pull this function out into a more generic file so that we can also reuse it elsewhere.

raunakab (approved):
Looks good! Some minor questions, but other than that, we can definitely get this merged in.

backend/onyx/connectors/sharepoint/connector.py
Subash-Mohan: LSP is not throwing an error, I just removed it now.

backend/onyx/connectors/sharepoint/connector.py
Subash-Mohan: I intentionally want that exception to be raised. If it remains unhandled after the retries, it indicates that there are issues with the API, so it’s better to halt the executions.

raunakab: Sounds good!

Issue #4978, opened 2025-07-01
Error in API container when using init-letsencrypt.sh


mikeswiz:
Was following this documentation:
[https://docs.onyx.app/production/aws/ec2](https://docs.onyx.app/production/aws/ec2)

Getting the following error in API container: 
```
botocore.exceptions.EndpointConnectionError: Could not connect to the endpoint URL: "http://minio:9000/onyx-file-store-bucket"

ERROR:    Application startup failed. Exiting.
```
I see the container defined in the yaml but when running `docker ps` I do not see the container running or a record of it ever running.

Dcooley1350:
I can provide more context on this as I just ran into the issue and dug into it.

I was attempting to follow this [guide to deploying Onyx on GCP](https://docs.onyx.app/production/gcp). Whenever I ran the `init-letsencrypt.sh` script I got hung up on a repeating message `Nginx is not ready yet, retrying in 5 seconds...`.

When I checked the logs of the Nginx container, the container is hung up waiting on the API Server container to be ready. The Nginx container repeats the message `API server responded with 000, retrying in 5 seconds...`.

When I checked the logs of the API Server container, the API Server container is logging:

```sh
raise EndpointConnectionError(endpoint_url=request.url, error=e)
botocore.exceptions.EndpointConnectionError: Could not connect to the endpoint URL: "http://minio:9000/onyx-file-store-bucket"
```

The issue appears to be that the API server is dependent on Minio, but the docker-compose files lack this dependency definition. Indeed, if I edit the docker-compose to add the minio service as a dependency of the api-server service, I can get past this blockage.

Note that the issue crops up only in the init-letsencrypt.sh script because that script tries to start only the nginx service, but the nginx service is dependent on the api-server, which causes the api-server to spin up without it's dependency minio available.

PR #5066, opened 2025-07-23
fix: explicit api_server dependency on minio in docker compose files

Dcooley1350:
## Description

Adds explicit dependency from api_server service to minio service in Docker compose files.

Fixes issue #4978 

## How Has This Been Tested?

In a [comment on issue #4978](https://github.com/onyx-dot-app/onyx/issues/4978#issuecomment-3104978972) I describe how I was able to overcome the issue with the `init-letsencrypt.sh` script not progressing. Explicitly defining in the Docker compose file the `api_server` service's dependency on the `minio` service causes the API server to be able to start, which in turn allows nginx to start, which allows the `init-letsencrypt.sh` script to progress.

The actual test was done on a GCP compute instance. The test actions were:

1. Ran the `init-letsencrypt.sh` script, experienced a hang while the script waiting for nginx to be ready.
2. Found in nginx logs that nginx was waiting for the api_server to be ready.
3. Found in the api_server logs that it was waiting for minio to be ready.
4. I edited [this depends_on list]( https://github.com/onyx-dot-app/onyx/blob/991d5e42037a8f259e33b9f5bfa24890dec5f349/deployment/docker_compose/docker-compose.prod.yml#L12) in the docker-compose.prod.yml file to include `- minio` as a dependency.
5. I re-ran the `init-letsencrypt.sh` script and it was able to progress successfully.

## Further Note

It appears that there may also be a dependency from the `background` service to the `minio` service. Is that the case and should I include that in this PR?

---

## Summary by cubic
Added an explicit dependency from the api_server service to the minio service in all Docker Compose files to ensure correct startup order and prevent initialization issues.

### DIFF
--- a/deployment/docker_compose/docker-compose.dev.yml
+++ b/deployment/docker_compose/docker-compose.dev.yml
@@ -13,6 +13,7 @@ services:
       - index
       - cache
       - inference_model_server
+      - minio
     restart: unless-stopped
     ports:
       - "8080:8080"
--- a/deployment/docker_compose/docker-compose.gpu-dev.yml
+++ b/deployment/docker_compose/docker-compose.gpu-dev.yml
@@ -13,6 +13,7 @@ services:
       - index
       - cache
       - inference_model_server
+      - minio
     restart: unless-stopped
     ports:
       - "8080:8080"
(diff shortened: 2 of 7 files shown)


justin-tahara:
@Dcooley1350 for the `background` service change, we can handle that in a separate PR but for now let's keep this scoped for just the `minio` change and we can further investigate on the `background` issue in a followup PR.

PR #5091, opened 2025-07-25
feat: Search UI

raunakab:
## Description

This PR adds Search UI.

Addresses: https://linear.app/danswer/issue/DAN-2213/work-on-updated-search-ui.

## How Has This Been Tested?

Tests incoming.


backend/onyx/chat/process_message.py
```diff
@@ -522,6 +527,45 @@ def _process_tool_response(
...
+    db_session: Session,
+    user: User | None,
+    api_search_request: ApiSearchRequest,
+) -> Generator[SavedSearchDoc]:
+    search_request = SearchRequest(
+        query=api_search_request.query,
+        limit=SEARCH_REQUEST_LIMIT,
+    )
+
+    llm, fast_llm = get_llms_for_persona(
```
evan-onyx, line 540: I don't think it's valid for persona to be None; maybe I just don't understand the new search UI but as of now which assistant is specified is a key part of determining what knowledge the user searches over

backend/onyx/chat/process_message.py
```diff
@@ -522,6 +527,45 @@ def _process_tool_response(
...
+        persona=None,
+        llm_override=api_search_request.llm_override,
+    )
+
+    search_pipeline = SearchPipeline(
```
evan-onyx, line 545: there are a bunch of settings that typically get passed into a new SearchPipeline based on stuff like retrieval options or a DocumentSearchRequest; if these are explicitly being ignored in this case we should comment as to why (but I imagine they should be passed in)

backend/onyx/chat/process_message.py
```diff
@@ -522,6 +527,45 @@ def _process_tool_response(
     return info_by_subq
 
 
+def _stream_search_result_objects(
+    db_session: Session,
+    user: User | None,
+    api_search_request: ApiSearchRequest,
+) -> Generator[SavedSearchDoc]:
+    search_request = SearchRequest(
```
evan-onyx, line 535: Typically this object is created with many more fields; probably they should be passed in from the frontend as in other cases like document-search

backend/onyx/server/query_and_chat/models.py
```diff
@@ -320,3 +320,8 @@ class ChatSearchRequest(BaseModel):
 
 class CreateChatResponse(BaseModel):
     chat_session_id: str
+
+
+class SearchRequest(BaseModel):
```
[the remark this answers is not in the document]
evan-onyx: I agree with cubic

backend/onyx/server/query_and_chat/search_backend.py
```diff
@@ -0,0 +1,42 @@
...
+from onyx.server.query_and_chat.chat_utils import is_connected
+from onyx.server.query_and_chat.models import SearchRequest
+from onyx.utils.logger import setup_logger
+
+
+logger = setup_logger()
+router = APIRouter(prefix="/search")
+
+
+@router.post("/send-query")
```
evan-onyx, line 21: this endpoint should be named something like stream-search-results probably. just anything that more directly describes what the endpoint does is fine

backend/onyx/server/query_and_chat/search_backend.py
```diff
@@ -0,0 +1,42 @@
...
+    def stream_generator() -> Generator[str, None, None]:
+        try:
+            for packet in stream_search_results(
+                api_search_request=search_request,
+                user=user,
+            ):
+                yield packet
+
+        except Exception as e:
+            logger.exception("Error in chat message streaming")
```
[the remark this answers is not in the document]
evan-onyx: agreed

evan-onyx (commented):
some things to discuss/fix/leave comments about if you've already talked it through with other folks

PR #5133, opened 2025-07-30
fix: Make ungrounded types have a default name when sending to the frontend

raunakab:
## Description

When compiling stats on entities, ungrounded entities were throwing errors because the logic expected that `grounded_source_name` to be non-null. However, it is possible for `grounded_source_name` to be null (hence the "ungrounded source"), which caused an error in pydantic serialization.

This PR fixes that issue by assigning a default name to all ungrounded entities when compiling the stats for the frontend.

Addresses: https://linear.app/danswer/issue/DAN-2255/fix-kg-admin-deserialization-serialization-issue.

## How Has This Been Tested?

Untested.
    

---

## Summary by cubic
Fixed errors when compiling entity stats by assigning a default name to ungrounded types before sending data to the frontend, ensuring proper serialization.

- **Bug Fixes**
  - Ungrounded entities now use "Ungrounded" as a default name if no grounded source name is present.

### DIFF
--- a/backend/onyx/db/entities.py
+++ b/backend/onyx/db/entities.py
@@ -330,7 +330,7 @@ def get_entity_stats_by_grounded_source_name(
         .all()
     )
 
-    # `row.grounded_source_name`` is NULLABLE in the database schema.
+    # `row.grounded_source_name` is NULLABLE in the database schema.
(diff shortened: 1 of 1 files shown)


evan-onyx (approved):
LGTM

PR #5135, opened 2025-07-30
fix: accept multiple zip types and fallback to extension

wenxi-onyx:
## Description

- zips uploaded from windows may not have mime application/zip 

## How Has This Been Tested?

tested by/credit to @ranok 


evan-onyx (approved):
Looks good, could you extract this into a utility function and check whether we're determining whether smth is a zip anywhere else in the codebase and use it there too?

PR #5142, opened 2025-07-31
Perf/indexing status page

Subash-Mohan:
## Description
This pull request introduces a new job that runs hourly to delete index attempt records older than 8 days. The 8-day threshold is set to ensure alignment with the checkpoint deletion job, which removes checkpoint records older than 7 days by referencing index attempt records—hence the additional day for buffer. Additionally, a new endpoint has been created to support pagination on the indexing status page, along with related changes.

## How Has This Been Tested?

---

## Summary by cubic
Added a new hourly job to clean up old index attempt records and rebuilt the indexing status admin page with backend pagination, filtering, and improved UI for large connector lists.

- **New Features**
  - Hourly background job deletes index attempt records older than 8 days to keep database size in check and align with checkpoint cleanup logic.
  - Indexing status admin page now supports backend pagination, filtering, and federated connector display for better performance and usability.
  - Added new API endpoint and related backend logic to support paginated, filterable connector status queries.

- **Refactors**
  - Updated frontend components to use the new paginated API, reducing load times and improving UX for large datasets.
  - Cleaned up and simplified frontend table code, including new loading skeletons and search/filter controls.


backend/onyx/background/celery/tasks/docprocessing/tasks.py
```diff
@@ -914,6 +916,94 @@ def cleanup_checkpoint_task(
...
+    # these tasks should never overlap
+    if not lock.acquire(blocking=False):
+        task_logger.info(
+            f"check_for_index_attempt_cleanup - Lock not acquired: tenant={tenant_id}"
+        )
+        return None
+
+    try:
+        locked = True
+        batch_size = 500
```
evan-onyx: nit: make batch_size a constant at the top of file

backend/onyx/background/indexing/index_attempt_utils.py
```diff
@@ -0,0 +1,27 @@
+from datetime import timedelta
+
+from sqlalchemy.orm import Session
+
+from onyx.configs.constants import NUM_DAYS_TO_KEEP_INDEX_ATTEMPTS
+from onyx.db.engine.time_utils import get_db_current_time
+from onyx.db.models import IndexAttempt
+
+
+def get_index_attempts(
```
evan-onyx: nit: should be get_old_index_attempts

backend/onyx/background/indexing/checkpointing_utils.py
```diff
@@ -163,13 +164,13 @@ def get_latest_valid_checkpoint(
...
 ) -> list[IndexAttempt]:
     """Get all index attempts with checkpoints older than the specified number of days.
 
     Args:
         db_session: The database session
-        days_to_keep: Number of days to keep checkpoints for (default: 7)
+        days_to_keep: Number of days to keep checkpoints for (default: NUM_DAYS_TO_KEEP_CHECKPOINTS)
 
     Returns:
         Number of checkpoints deleted
```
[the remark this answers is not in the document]
evan-onyx: agreed

backend/onyx/background/indexing/index_attempt_utils.py
```diff
@@ -0,0 +1,27 @@
...
+    cutoff_date = get_db_current_time(db_session) - timedelta(days=days_to_keep)
+    return (
+        db_session.query(IndexAttempt)
+        .filter(IndexAttempt.time_created < cutoff_date)
+        .all()
+    )
+
+
+def cleanup_index_attempts(db_session: Session, index_attempt_ids: list[int]) -> None:
+    """Clean up an index attempt"""
```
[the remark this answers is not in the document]
evan-onyx: agreed

backend/onyx/background/indexing/index_attempt_utils.py
```diff
@@ -0,0 +1,27 @@
...
+    db_session.query(IndexAttempt).filter(
+        IndexAttempt.id.in_(index_attempt_ids)
+    ).delete()
+    db_session.commit()
```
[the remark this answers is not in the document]
evan-onyx: ^^ this seems good since we commit afterwards

backend/onyx/server/documents/connector.py
```diff
@@ -868,6 +882,403 @@ def get_connector_indexing_status(
     return indexing_statuses
 
 
+@router.post("/admin/connector/indexing-status-with-pagination")
```
evan-onyx: nit: slightly prefer indexing-status-paginated

backend/onyx/server/documents/connector.py
```diff
@@ -868,6 +882,403 @@ def get_connector_indexing_status(
...
+            latest_finished_index_attempts,
+            non_editable_cc_pairs,
+        ) = run_functions_tuples_in_parallel(parallel_functions)
+
+    # Cast results to proper types
+    non_editable_cc_pairs = cast(list[ConnectorCredentialPair], non_editable_cc_pairs)
+    editable_cc_pairs = cast(list[ConnectorCredentialPair], editable_cc_pairs)
+    federated_connectors = cast(list[FederatedConnector], federated_connectors)
+    latest_index_attempts = cast(list[IndexAttempt], latest_index_attempts)
+    latest_finished_index_attempts = cast(
```
evan-onyx, line 801: do we need to cast here AND define the types above (lines 918-922)? Feels redundant but it's okay if it's necessary to get proper typing in the function

backend/onyx/server/documents/connector.py
```diff
@@ -868,6 +882,403 @@ def get_connector_indexing_status(
...
+
+        status = _get_connector_indexing_status_lite(
+            cc_pair, latest_attempt, latest_finished_attempt, True, doc_count
+        )
+        if status:
+            editable_statuses.append(status)
+
+    # Process non-editable cc_pairs
+    non_editable_statuses: list[ConnectorIndexingStatusLite] = []
+    for cc_pair in non_editable_cc_pairs:
```
evan-onyx, line 869: nit: put this for loop into a helper function (passing in the True/False for _get_indexing_connector_status_lite) to avoid code duplication

evan-onyx: if there's a ton of variables to pass in, it's fine to just define the function locally to avoid that

backend/onyx/server/documents/connector.py
```diff
@@ -868,6 +882,403 @@ def get_connector_indexing_status(
...
+    for federated_connector in federated_connectors:
+        federated_status = FederatedConnectorStatus(
+            id=federated_connector.id,
+            source=federated_connector.source,
+            name=f"{federated_connector.source.replace('_', ' ').title()}",
+        )
+
+        federated_statuses.append(federated_status)
+
+    source_to_summary: dict[DocumentSource, SourceSummary] = {}
```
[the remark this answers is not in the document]
evan-onyx: ^^ what is the summary used for? please verify whether this comment is valid. If we DO want the unfiltered connector summary, would be nice to have a comment here explaining why!

backend/onyx/server/documents/connector.py
```diff
@@ -868,6 +882,403 @@ def get_connector_indexing_status(
...
+        statuses = source_to_all_statuses[source]
+        # Get current page for this source (default to page 1, 1-indexed)
+        current_page = request.source_to_page.get(source, 1)
+
+        # Calculate start and end indices for pagination (convert to 0-indexed)
+        start_idx = (current_page - 1) * _INDEXING_STATUS_PAGE_SIZE
+        end_idx = start_idx + _INDEXING_STATUS_PAGE_SIZE
+
+        # Get the page slice for this source
+        page_statuses = statuses[start_idx:end_idx]
```
evan-onyx: doesn't this mean that we materialize all the statuses (for an arbitrary number of connectors) and then just send back some of them? Ideally we would be doing something like offset-ing and limit-ing a db query right? If you've tested this and it's still significantly faster than before then it's fine for now, but leaving a TODO explaining this would be good

web/src/app/admin/indexing/status/CCPairIndexingStatusTable.tsx
```diff
@@ -279,503 +278,168 @@ border border-border dark:border-neutral-700
...
-
-    setConnectorsToggled(connectors);
-    Cookies.set(TOGGLED_CONNECTORS_COOKIE_NAME, JSON.stringify(connectors));
-  };
-
-  const shouldExpand =
-    Object.values(connectorsToggled).filter(Boolean).length <
-    sortedSources.length;
-
+  // Filter connectors based on search quer
```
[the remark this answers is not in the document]
evan-onyx: true

web/src/app/admin/indexing/status/CCPairIndexingStatusTable.tsx
```diff
@@ -279,503 +278,168 @@ border border-border dark:border-neutral-700
...
+                                    borderRight: "1px solid var(--border)",
+                                    borderLeft: "1px solid var(--border)",
+                                  }
+                                : {}
+                            }
+                          >
+                            {isLastDummyRow ? (
+                              <TableCell
+                                colSpan={
+                                  isPaidEnterpriseFeaturesEnabled ? 6 : 5
```
evan-onyx: would be good to make these numbers programmatic or at least constants

web/src/app/admin/indexing/status/CCPairIndexingStatusTable.tsx
```diff
@@ -279,503 +278,168 @@ border border-border dark:border-neutral-700
...
+                    <TableCell
+                      colSpan={isPaidEnterpriseFeaturesEnabled ? 6 : 5}
+                    >
+                      <div className="flex justify-center">
+                        <PageSelector
+                          currentPage={ccPairStatus.current_page}
+                          totalPages={ccPairStatus.total_pages}
+                          onPageChange={(newPage) =>
+                            onPageChange(ccPairStatus.source, newPage)
+                          }
```
evan-onyx, line 445: These changes look pretty extensive, should get @Weves to look through the UI changes

web/src/app/admin/indexing/status/SearchAndFilterControls.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+  // Debounce the search query
+  useEffect(() => {
+    const timer = setTimeout(() => {
+      resetPagination();
+      onSearchChange(localSearchValue);
+    }, 300);
+
+    return () => clearTimeout(timer);
+  }, [localSearchValue, onSearchChange]);
```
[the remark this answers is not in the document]
evan-onyx: ^^ wow both code reviewers think this is an issue! give it a look :)

web/src/lib/hooks.ts
```diff
@@ -78,8 +88,125 @@ export const useObjectState = <T>(
...
+    () => ({
+      secondary_index: false,
+      access_type_filters: [],
+      last_status_filters: [],
+      docs_count_operator: null,
+      docs_count_value: null,
+      source_to_page: sourcePagesRef.current, // Use current pagination state
+      ...filters,
+    }),
+    [filters]
```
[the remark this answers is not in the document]
evan-onyx: ^^

web/src/lib/types.ts
```diff
@@ -175,6 +175,42 @@ export interface ConnectorIndexingStatus<
   docs_indexed: number;
 }
 
+export interface ConnectorIndexingStatusLite {
```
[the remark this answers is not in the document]
evan-onyx: not sure what the bot is cooking, maybe just do a quick check for duplicates

evan-onyx (approved):
mostly nits, one bigger comment about pagination

backend/onyx/background/celery/tasks/docprocessing/tasks.py
```diff
@@ -914,6 +916,94 @@ def cleanup_checkpoint_task(
...
+        with get_session_with_current_tenant() as db_session:
+            old_attempts = get_index_attempts(db_session)
+            # We need to batch this because during the initial run, the system might have a large number
+            # of index attempts since they were never deleted. After that, the number will be
+            # significantly lower.
+            if len(old_attempts) == 0:
+                task_logger.info(
+                    "check_for_index_attempt_cleanup - No index attempts to cleanup"
+                )
```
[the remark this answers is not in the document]
Subash-Mohan: If list is empty, iteration won't happen, but I added the return now

backend/onyx/server/documents/connector.py
Subash-Mohan: Just removed the predeclaration of variables

backend/onyx/server/documents/connector.py
Subash-Mohan: I also felt that indexing-status-paginated is better, so I’ve updated it.

backend/onyx/server/documents/connector.py
Subash-Mohan: Added a local function to deduplicate the code.

backend/onyx/server/documents/connector.py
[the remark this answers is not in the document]
Subash-Mohan: I think so if the filter is applied it should show the summary of filtered connectors. I am moving the logic after the filter is applied.

backend/onyx/server/documents/connector.py
Subash-Mohan: Since we need a summary for all connectors, we need to fetch all of them for the summary calculation. Also, the connectors should be displayed in descending order for each source. Given that the number of connectors isn’t expected to be very large I believe it’s more efficient and simpler to load all connectors into memory rather than introducing complex queries. Also, with around 2000 connectors, we're able to achieve a load time of under one second.

web/src/app/admin/indexing/status/CCPairIndexingStatusTable.tsx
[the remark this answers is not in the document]
Subash-Mohan: Whole comment is not needed😅

web/src/lib/hooks.ts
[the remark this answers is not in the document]
Subash-Mohan: We intentionally left it out to avoid triggering a full page reload.

web/src/lib/types.ts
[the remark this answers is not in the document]
Subash-Mohan: There are no duplicates; I think it's getting confused with other interfaces.

backend/onyx/server/documents/connector.py
evan-onyx: got it, I've seen it get slow when there are 10k+ connectors unfortunately; one thing to consider here is that each My Documents file is its own connector at the moment. As long as your query doesn't pull those in we should be fine, but I think it's very plausible that we hit even 100k connectors for larger deployments when My Documents (user files) are included!

Maybe not a subject for this PR, but it's somewhat odd to me that we NEED to get this summary info, I'm somewhat in favor of getting rid of it for efficiency. Could you leave a comment summarizing this discussion, i.e. why we're materializing everything in memory and that the query will likely become really slow if anyone accidentally makes it include user files?

evan-onyx (approved):
looks good, just left a suggestion to add a comment

backend/onyx/server/documents/connector.py
```diff
@@ -868,6 +882,389 @@ def get_connector_indexing_status(
     return indexing_statuses
 
 
+@router.post("/admin/connector/indexing-status-paginated")
```
Weves: hmm, why do we now have two endpoints for this?

Weves: We should probably just replace the `indexing-status` endpoint with this?

Weves (changes requested):

backend/onyx/server/documents/connector.py
Subash-Mohan: Still tests, scripts and Search settings updation page is dependent on this endpoint.

backend/onyx/server/documents/connector.py
```diff
@@ -693,27 +702,13 @@ def get_connector_status(
     ]
 
 
-@router.get("/admin/connector/indexing-status")
-def get_connector_indexing_status(
-    secondary_index: bool = False,
+@router.post("/admin/connector/indexing-status-paginated")
```
Weves: can we not call this `-paginated`? I think by default endpoints should be paginated / the fact that they are paginated should be determined by whether or not they provide a limit / start param.

Also why is this a POST now?

Subash-Mohan: We’re using a nested pagination structure where each source has its own page number and additional filter properties that need to be sent with every request. That’s why we are using to `POST`. In fact, the paginated URL has been `POST` right from the start.

Subash-Mohan: I’ve renamed both the method and the endpoint now.

Weves (approved):

PR #5158, opened 2025-08-06
Feat/salesforce custom query

evan-onyx:
## Description

Allow users to specify custom salesforce object queries. Rather than just specify top level object which leads to all children of those object(s) and all fields being pulled in, the user can now specify exactly which fields and child objects they want indexed. At the moment this goes "one level deep" i.e. the user specifies only top level objects and the children of those objects, but in principle it is fine to extend it to grandchildren by adding new top level objects that look like <parent>.<child> or something similar. Not doing this for now because it would require a bigger overhaul to the whole connector.

## How Has This Been Tested?

Tested in UI, unit tests

### DIFF
--- a/backend/onyx/connectors/salesforce/connector.py
+++ b/backend/onyx/connectors/salesforce/connector.py
@@ -209,8 +209,6 @@ class SalesforceConnector(LoadConnector, PollConnector, SlimConnector):
         # Validate and store custom query config
         if custom_query_config:
             config_json = json.loads(custom_query_config)
-            if not _validate_custom_query_config(config_json):
-                raise ValueError("Invalid custom_query_config structure")
             self.custom_query_config: dict[str, Any] | None = config_json
             # If custom query config is provided, use the object types from it
             self.parent_object_list = list(config_json.keys())
@@ -218,7 +216,7 @@ class SalesforceConnector(LoadConnector, PollConnector, SlimConnector):
             self.custom_query_config = None
             # Use the traditional requested_objects approach
             self.parent_object_list = (
-                [obj.capitalize() for obj in requested_objects]
+                [obj.strip().capitalize() for obj in requested_objects]
                 if requested_objects
                 else _DEFAULT_PARENT_OBJECT_TYPES
             )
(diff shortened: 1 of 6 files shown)


backend/onyx/connectors/salesforce/connector.py
```diff
@@ -87,6 +88,72 @@ class SalesforceConnectorContext:
...
+        return None, {}
+
+    obj_config = config[object_type]
+    fields = obj_config.get("fields")
+    associations = obj_config.get("associations", {})
+
+    return fields, associations
+
+
+def _get_all_object_types_from_config(config: dict[str, Any]) -> set[str]:
```
justin-tahara: Delete given this is not in use.

backend/onyx/connectors/salesforce/connector.py
```diff
@@ -87,6 +88,72 @@ class SalesforceConnectorContext:
...
+    Returns:
+        tuple of (fields_list, associations_dict)
+        - fields_list: List of fields to query, or None if not specified (use all)
+        - associations_dict: Dict mapping association names to their config
+    """
+    if object_type not in config:
+        return None, {}
+
+    obj_config = config[object_type]
+    fields = obj_config.get("fields")
```
justin-tahara, line 110: Should we add a default value if the fields is not populated? Or is None okay here

backend/onyx/connectors/salesforce/connector.py
```diff
@@ -87,6 +88,72 @@ class SalesforceConnectorContext:
...
+
+    return all_types
+
+
+def _validate_custom_query_config(config: dict[str, Any]) -> bool:
+    """
+    Validate the structure of the custom query configuration.
+    """
+
+    for object_type, obj_config in config.items():
```
justin-tahara, line 121: Is it okay if config is empty? It will just return True in this case.

backend/onyx/connectors/salesforce/connector.py
```diff
@@ -831,25 +907,55 @@ def _make_context(
...
+            if self.custom_query_config:
+                custom_fields, associations_config = (
+                    _extract_fields_and_associations_from_config(
+                        self.custom_query_config, parent_type
+                    )
+                )
+                custom_fields = custom_fields or []
+
+                # Get custom fields for parent type
+                field_set = set(custom_fields)
```
justin-tahara, line 922: nit: If custom_fields was a dict this would be cleaner but will leave up to you on the implementation. Just strange that we cast to set then recast back to a list

justin-tahara: Might be worth using some helpers or at least making this _make_context function smaller.

backend/onyx/connectors/salesforce/onyx_salesforce.py
```diff
@@ -60,7 +60,7 @@ def is_blacklisted(self, object_type: str) -> bool:
                 return True
 
         for suffix in SALESFORCE_BLACKLISTED_SUFFIXES:
-            if object_type_lower.endswith(prefix):
+            if object_type_lower.endswith(suffix):
```
justin-tahara, line 63: Nice catch

justin-tahara (commented):
Initial review. I also need to get ramped up on the context but looks good so far

backend/onyx/connectors/salesforce/connector.py
evan-onyx: None is fine

backend/onyx/connectors/salesforce/connector.py
evan-onyx: cleaned this up to throw reasonable errors, and yeah empty config is fine (will just get the default User and Account objects)

backend/onyx/connectors/salesforce/connector.py
evan-onyx: removed the recast back to list, code is simpler/better in other places now

backend/onyx/connectors/salesforce/connector.py
```diff
@@ -831,25 +899,54 @@ def _make_context(
...
+                # these are expected and used during doc conversion
+                field_set.add("Name")
```
Weves: add constants for these!

backend/onyx/connectors/salesforce/connector.py
```diff
@@ -858,19 +955,25 @@ def _make_context(
...
                 parent_to_child_relationships[parent_type].add(child_relationship)
 
                 # map relationship to queryable fields of the target table
-                queryable_fields = sf_client.get_queryable_fields_by_type(child_type)
+                if config_fields := (
+                    associations_config and associations_config.get(child_type)
+                ):
+                    field_set = set(config_fields)
+                    # these are expected and used during doc conversion
+                    field_set.add("Name")
```
Weves: same here

backend/onyx/connectors/salesforce/connector.py
```diff
@@ -899,9 +1002,17 @@ def _make_context(
 
         logger.info(f"All object types: num={len(all_types)} list={all_types}")
 
+        # Ensure User and Account have queryable fields if they weren't already processed
+        essential_types = ["User", "Account"]
```
Weves: constants here as well

Weves (approved):

PR #5164, opened 2025-08-07
feat: JIRA support for custom JQL filter

sktbcpraha:
## Description

Adds user defined JQL support to the Jira connector 

- Custom JQL query support in connector configuration
- Basic validation of JQL syntax
- Integration with existing time-based filtering
- UI support for JQL query input

Snímek obrazovky 2025-08-07 113534

## How Has This Been Tested?

- Added new test case with defined JQL filter
- Tested in UI configurations with JQL and without
- Tested on private Jira instance both configuration with and without JQL

---

## Summary by cubic
Added support for custom JQL queries in the Jira connector, allowing users to filter issues using their own JQL expressions.

- **New Features**
  - Users can enter a custom JQL query in the connector settings UI.
  - The connector validates JQL syntax and prevents using both a project key and JQL at the same time.


wenxi-onyx:
@cubic-dev-ai review this

backend/onyx/connectors/jira/connector.py
```diff
@@ -262,6 +269,11 @@ def _get_jql_query(
 
         time_jql = f"updated >= '{start_date_str}' AND updated <= '{end_date_str}'"
 
+        # If custom JQL query is provided, use it and combine with time constraints
+        if self.jql_query:
```
wenxi-onyx, line 274: Is there anything stopping the user from submitting time constraints? I believe that would interfere with the checkpointed connector system. Maybe this needs to be clarified on the front-end instructions?

wenxi-onyx: To add to this, any JQL that the connector inherently adds can't/shouldn't be modified by the user --> I suppose the JQL query will just be invalid if the user attempts this.

backend/onyx/connectors/jira/connector.py
```diff
@@ -371,9 +383,51 @@ def retrieve_all_slim_documents(
...
+                [_ for _ in _perform_jql_search(
+                    jira_client=self.jira_client,
+                    jql=self.jql_query,
+                    start=0,
+                    max_results=1,
+                )]
+            except Exception as e:
+                status_code = getattr(e, "status_code", None)
+
+                if status_code == 401:
```
wenxi-onyx: I think you can just log the error instead of processing each error code separately. The response from Jira should specify, in detail, the issue. I was attempting to test with project = AS, but gettin unspecified 400. Turns out AS is a protected keyword that needs quotes!

backend/onyx/connectors/jira/connector.py
```diff
@@ -371,9 +383,51 @@ def retrieve_all_slim_documents(
     def validate_connector_settings(self) -> None:
         if self._jira_client is None:
             raise ConnectorMissingCredentialError("Jira")
+            
+        # Check if both jql_query and project_key are defined
+        if self.jql_query and self.jira_project:
```
wenxi-onyx: I believe this is impossible to do via the UI. Is there a reason you check for this?

backend/onyx/connectors/jira/connector.py
```diff
@@ -371,9 +383,51 @@ def retrieve_all_slim_documents(
...
+            try:
+                # Try to execute the JQL query with a small limit to validate its syntax
+                # We use a list comprehension to force evaluation of the generator
+                # without actually processing the results
+                [_ for _ in _perform_jql_search(
+                    jira_client=self.jira_client,
+                    jql=self.jql_query,
+                    start=0,
+                    max_results=1,
+                )]
```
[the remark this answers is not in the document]
wenxi-onyx: Perhaps irrelevant since you do max_results 1.

backend/tests/daily/connectors/jira/test_jira_basic.py
```diff
@@ -26,6 +26,23 @@ def jira_connector() -> JiraConnector:
     return connector
 
 
+@pytest.fixture
+def jira_connector_with_jql() -> JiraConnector:
+    connector = JiraConnector(
+        jira_base_url="https://danswerai.atlassian.net",
+        jql_query="project = AS AND issuetype = Story",
```
wenxi-onyx: AS is a protected JQL term, so it has to be wrapped in quotes. 

Exact error
`response text = {"errorMessages":["Error in the JQL Query: 'AS' is a reserved JQL word. You must surround it in quotation marks to use it in a query. (line 1, character 12)"],"errors":{}}`

wenxi-onyx:
Hey @sktbcpraha - thanks for the contribution! This is definitely useful. Would you mind addressing the comments? Thank you!

sktbcpraha:
@wenxi-onyx thank you for you feedback. I will try to fix it...

backend/onyx/connectors/jira/connector.py
sktbcpraha: I’ve improved the front-end instruction. Is that enough, or do we also need to validate it on the back-end?

backend/onyx/connectors/jira/connector.py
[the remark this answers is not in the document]
sktbcpraha: I have changed it anyway...

backend/onyx/connectors/jira/connector.py
sktbcpraha: It should not be necessary. I have removed this check...

backend/tests/daily/connectors/jira/test_jira_basic.py
sktbcpraha: It should be fixed

backend/onyx/connectors/jira/connector.py
sktbcpraha: I tried to implement a helper function to handle the Jira API exceptions for all three cases (custom JQL, project definition, whole project). I kept the most 'special' error code cases. For others (e.g., 400, 404), I try to extract the original error message from Jira. I tested it, and it seems to work.
image

PR #5167, opened 2025-08-07
fix: sanitize slack payload before logging

wenxi-onyx:
## Description

- title, remove slack msg content from logs

## How Has This Been Tested?

locally


backend/onyx/onyxbot/slack/listener.py
```diff
@@ -570,6 +570,16 @@ def shutdown(self, signum: int | None, frame: FrameType | None) -> None:
         sys.exit(0)
 
 
+def sanitize_slack_payload(payload: dict) -> dict:
+    """Remove message content from Slack payload for logging"""
+    sanitized = {k: v for k, v in payload.items() if k not in {"text", "blocks"}}
```
evan-onyx: nit: since you're re-using {"text", "blocks"} and it's constant just define it as a constant outside the func

wenxi-onyx: done

PR #5174, opened 2025-08-08
max tokens param

evan-onyx:
## Description

Gpt-5 and o3 use a newer version of the max tokens param

## How Has This Been Tested?

n/a should be totally safe and fall back to default value at worst


justin-tahara (approved):
Beautiful 🤩

Issue #5178, opened 2025-08-10
Loopio connector will fail to index libraries that contain complianceOption entries that have no answer text


joseat4g:
Some Loopio questions are compliance only questions. The response values are stored in `complianceOption` as part of the question object  and not as answer text.  This results in an indexing error  `AttributeError: '<class 'bs4.BeautifulSoup'>' object has no attribute 'contents'` as the answer has no text value.

PR #5179, opened 2025-08-10
fix(connector): #5178 Add error handling and logging for empty answer text in Loopio Connector

joseat4g:
## Description

Add handler for Loopio library entries that are compliance option only and have no answer text.

## How Has This Been Tested?

This was tested on a local installation with a Loopio connector, indexing an actual Loopio library that contains complianceOption only entries. Before this change, the Library fails to index. After this change, the Library indexing completes, skipping the entries that have no answer text.

- [x ] This PR should be backported (make sure to check that the backport attempt succeeds)

    

---

## Summary by cubic
Added error handling and logging for Loopio library entries with empty answer text to prevent indexing failures.

- **Bug Fixes**
 - Skips entries without answer text and logs a warning.
 - Catches and logs errors when parsing answer HTML.


backend/onyx/connectors/loopio/connector.py
```diff
@@ -119,7 +119,20 @@ def _process_entries(
                     part["name"] for part in entry["location"].values() if part
                 )
 
-                answer = parse_html_page_basic(entry.get("answer", {}).get("text", ""))
+                if entry.get("answer", {}).get("text", "") == "":
```
wenxi-onyx: Minor nit: Since you re-use `entry.get("answer", {}).get("text", "")`, can you move it to a var before this if? 

e.g. something like
```
answer_text = entry.get("answer", {}).get("text", "")
if not answer_text
...
try:
   parsed_answer = parse_html_page_basic(answer_text)
```

wenxi-onyx:
@joseat4g Thanks for the contribution! One small nit and we can get it merged :)

backend/onyx/connectors/loopio/connector.py
```diff
@@ -119,7 +119,20 @@ def _process_entries(
...
+                    )
+                    continue
+
+                try:
+                    answer = parse_html_page_basic(
+                        entry.get("answer", {}).get("text", "")
+                    )
+                except Exception as e:
+                    logger.error(f"Error parsing HTML for entry {entry['id']}: {e}")
+                    answer = ""
```
wenxi-onyx: I'm not super familiar with loopio. Does it make sense to continue creating a Document with answer = ""? I know this is the original connector's logic. Would this just create nearly blank documents?

josebanez: You're correct. I updated it to just skip a document that cannot be parsed.

backend/onyx/connectors/loopio/connector.py
josebanez: I've updated it to re-use the value of `entry.get("answer", {}).get("text", "")` as you recommended.

PR #5173, opened 2025-08-08
feat: make sharepoint documents and sharepoint pages optional

nsklei:
## Description

The Feature allows users to choose whether they want to index only SharePoint pages, only SharePoint documents, or both.
The PR adds the backend logic to the sharepoint connector.py and the necessary UI elements to the connectors.tsx.

image

## How Has This Been Tested?

I tested the UI elements and its functionality manually by building the images locally and starting the application.
I created three Sharepoint connectors with permutation of all three options (see below) and tested them in an assistant.

- include_site_pages: bool = True, include_site_documents: bool = True
- include_site_pages: bool = True, include_site_documents: bool = False
- include_site_pages: bool = False, include_site_documents: bool = True

## Backporting (check the box to trigger backport action)

Note: You have to check that the action passes, otherwise resolve the conflicts manually and tag the patches.

- [ ] This PR should be backported (make sure to check that the backport attempt succeeds)
- [x] [Optional] Override Linear Check

    

---

## Summary by cubic
Added options to let users choose whether to index SharePoint documents, SharePoint pages, or both when setting up a connector.

- **New Features**
  - Added checkboxes in the UI for selecting documents and/or pages.
  - Updated backend logic to support indexing based on user selection.

PR #5183, opened 2025-08-11
feat: make sharepoint documents and sharepoint pages optional

nsklei:
The Feature allows users to choose whether they want to index only SharePoint pages, only SharePoint documents, or both.
The PR adds the backend logic to the sharepoint connector.py and the necessary UI elements to the connectors.tsx.

image

## How Has This Been Tested?

include_site_pages: bool = True, include_site_documents: bool = True
include_site_pages: bool = True, include_site_documents: bool = False
include_site_pages: bool = False, include_site_documents: bool = True## Description

I tested the UI elements and its functionality manually by building the images locally and starting the application.
I created three Sharepoint connectors with permutation of all three options (see below) and tested them in an assistant.

---

## Summary by cubic
Added options to let users choose whether to index SharePoint documents, SharePoint pages, or both when setting up a connector.

- **New Features**
 - Added checkboxes in the UI for selecting documents and/or pages.
 - Updated backend logic to support these options.

### DIFF
--- a/backend/onyx/connectors/sharepoint/connector.py
+++ b/backend/onyx/connectors/sharepoint/connector.py
@@ -494,6 +494,7 @@ class SharepointConnector(
         batch_size: int = INDEX_BATCH_SIZE,
         sites: list[str] = [],
         include_site_pages: bool = True,
+        include_site_documents: bool = True,
     ) -> None:
         self.batch_size = batch_size
         self._graph_client: GraphClient | None = None
@@ -502,6 +503,7 @@ class SharepointConnector(
         )
         self.msal_app: msal.ConfidentialClientApplication | None = None
         self.include_site_pages = include_site_pages
+        self.include_site_documents = include_site_documents
         self.sp_tenant_domain: str | None = None
 
     @property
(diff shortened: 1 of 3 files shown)


wenxi-onyx:
Hey @nsklei - Looking at #5173, I think some commits were lost when you swapped over to PR from a new branch. Would you mind taking a look at this?

nsklei:
Hey @wenxi-onyx - Looks good from my side. Originally it were just one commit with 3 files changed.
The second commit in #5173 was just the unintended result of syncing the main branch of my fork.

web/src/lib/connectors/connectors.tsx
```diff
@@ -666,14 +666,31 @@ See our docs for more details.`,
...
 `,
       },
     ],
-    advanced_values: [],
+    advanced_values: [
+      {
+        type: "checkbox",
+        query: "Index Documents:",
+        label: "Index Documents",
+        name: "include_site_documents",
```
wenxi-onyx, line 676: Thanks for the contribution! Bit of a long ask here - could you please remove the existing site indexing logic in favor of your contribution? I'll outline the ask across multiple comments.

Could you please expand the description of this checkbox to something like: "if this is checked, all ASPX files/sites will be indexed. This ignores drive or folder restrictions you may have entered above."

backend/onyx/connectors/sharepoint/connector.py
```diff
@@ -558,18 +560,21 @@ def _fetch_from_sharepoint(
...
+                    # Convert driveitem to document with size checking
+                    doc = _convert_driveitem_to_document(driveitem, drive_name)
+                    if doc is not None:
+                        doc_batch.append(doc)
+
+                    if len(doc_batch) >= self.batch_size:
+                        yield doc_batch
+                        doc_batch = []
 
             # Fetch SharePoint site pages (.aspx files)
```
wenxi-onyx: Would you mind removing the logic here L579:586? Instead of checking for a drive or folder restriction, just index all sites if the checkbox is ticked

backend/onyx/connectors/sharepoint/connector.py
```diff
@@ -302,6 +303,7 @@ def __init__(
         )
         self.msal_app: msal.ConfidentialClientApplication | None = None
         self.include_site_pages = include_site_pages
+        self.include_site_documents = include_site_documents
```
wenxi-onyx, line 515: Could you please implement validate_connector_settings so that at least 1 checkbox is checked?

User could theoretically uncheck both and have 0 docs indexed --> may think that connector is broken instead of bad config

wenxi-onyx (changes requested):

backend/onyx/connectors/sharepoint/connector.py
nsklei: Validation implemented in L509:L514

Image

web/src/lib/connectors/connectors.tsx
nsklei: The description is now a bit more informative.

Image

backend/onyx/connectors/sharepoint/connector.py
nsklei: Totally makes sense to remove this logic.
Now if include_site_pages is True all SharePoint sites will be indexed, even if a library or folder is specified (see L1279:L1308)

wenxi-onyx:
Hey @nsklei, would you mind fixing one last issue I found in this connector (not caused by you)? 

Could you please add
```
# Exclude personal sites because GET personal site pages returns 404
if "-my.sharepoint" in site_descriptor.url:
   return []
```

At the start of _fetch_site_pages()  
https://github.com/onyx-dot-app/onyx/pull/5183/files#diff-0e02c6711532c1b02c5f40fe38179e67d7bdb918fc826c6f3e1a1c6017ae14e0R794

Thank you so much!!

PR #5185, opened 2025-08-12
thread safe approach to docprocessing logging

evan-onyx:
## Description

docprocessing multithreading was causing wrong logging tags which made debugging docprocessing difficult

## How Has This Been Tested?

testing in UI


justin-tahara:
@greptileai

backend/onyx/utils/logger.py
```diff
@@ -34,30 +35,6 @@ def reset() -> None:
         doc_permission_sync_ctx.set(dict())
 
 
-class TaskAttemptSingleton:
```
justin-tahara, line 37: We're sure that this is no longer needed?

justin-tahara (approved):

PR #5200, opened 2025-08-14
Decouple helm chart from bitnami

eric-zadara:
## Description

The original layout for the single secret and single configmap made it difficult to deploy the application with anything other than the bundled dependencies, making the `[postgresql|redis|vespa|nginx].enabled` configuration value moot.

This change maintains the functional defaults, but breaks up the secrets for postgres/redis/etc so that they can be exchanged with user defined alternatives. Essentially it enables:
* Using as-is with self-contained dependencies
* Swapping bitnami postgresql for RDS, cloudnative-pg.io or other database solutions/operators
* Swapping bitnami redis for hosted redis solutions or redis operators
* Swapping bitnami minio for any other S3 compatible endpoint
* Using an external secrets solution such as sealedsecrets or Hashicorp vault

Since the `auth` section has become iterative, all existing secrets configurations may be disabled and a user can arbitrarily configure as many as they like to be added as environment variables to all onyx pods.
Also empty configmap variables are now NOT generated in the configmap to prevent k8s from reporting environment variable collisions between the configmap and secrets IF a user decides to use a secret instead.

Our environment uses cloudnative-pg by default for all postgres dependencies, which the original chart could not support due to cloudnative-pg creating it's own secrets configuration.

I tried to keep changes to a minimum and consistent with existing configuration decisions to not make the values.yaml changes too confusing/drastic.

## How Has This Been Tested?

Mostly by hand in my k3s sandbox cluster and with argocd.
I primarily tested swapping postgresql between bitnami and cloudnative-pg, as all behavior stems from range loops, configuration-wise it is sound for postgresql/redis/vespa/nginx/minio.
Cloudnative-pg by default generates it's own credentials and loads them into a secret.

Below are the argo configurations I used for testing bitnami and cloudnative-pg, and I've included an "RDS" example as well that is roughly the "manual" entry.

[onyx-stack.stock.yml](https://github.com/user-attachments/files/21781197/onyx-stack.stock.yml)
[onyx-stack.cnpg.yml](https://github.com/user-attachments/files/21781195/onyx-stack.cnpg.yml)
[onyx-stack.rds.yml](https://github.com/user-attachments/files/21781196/onyx-stack.rds.yml)


justin-tahara:
@greptileai

deployment/helm/charts/onyx/values.yaml
```diff
@@ -599,34 +599,77 @@ letsencrypt:
...
+    # -- Overwrite the default secret name, ignored if existingSecret is defined
+    secretName: 'onyx-redis'
+    # -- Use a secret specified elsewhere
+    existingSecret: ""
+    # -- This defines the env var to secret map, key is always upper-cased as an env var
+    secretKeys:
+      REDIS_PASSWORD: redis_password
+    # -- Secrets values IF existingSecret is empty. Key here must match the value in secretKeys to be used
+    values:
+      redis_password: "password"
```
justin-tahara, line 744: I know this is nit but can we align and make this the `redis_password` that we have from above. Just in case someone did not update these values and are using the default values?

deployment/helm/charts/onyx/templates/configmap.yaml
```diff
@@ -6,13 +6,21 @@ metadata:
     {{- include "onyx-stack.labels" . | nindent 4 }}
 data:
   INTERNAL_URL: "http://{{ include "onyx-stack.fullname" . }}-api-service:{{ .Values.api.service.port | default 8080 }}"
+  {{- if .Values.postgresql.enabled }}
```
justin-tahara, line 9: Should these be .Values.auth.postgresql.enabled? Just double checking that the indentation is correct in the values.yaml file.

justin-tahara: Check the following as well

deployment/helm/charts/onyx/Chart.yaml
```diff
@@ -5,7 +5,7 @@ home: https://www.onyx.app/
 sources:
   - "https://github.com/onyx-dot-app/onyx"
 type: application
-version: 0.2.9
+version: 0.3.0
```
justin-tahara, line 8: Might need to rebase again. Sorry!

deployment/helm/charts/onyx/values.yaml
eric-zadara: The original actually is `password`, this is mimicking the original behavior
`.redis.secretKeys.REDIS_PASSWORD` points to `.redis.values.redis_password` which contains `password`.
REDIS_PASSWORD is the env var and it's value is the secret's key.

https://github.com/onyx-dot-app/onyx/blob/cc48a0c38eda9f4e8028719bbb4beead970f217d/deployment/helm/charts/onyx/values.yaml#L615
https://github.com/onyx-dot-app/onyx/blob/cc48a0c38eda9f4e8028719bbb4beead970f217d/deployment/helm/charts/onyx/values.yaml#L628

When I duplicated the behavior, i used `.values` instead of `.secrets` compared to the original. I think I wasn't sure how close to the original behavior I would get.

The original lower->uppercase behavior is present in the template as well, but I uppercased these entries in the config anyway to prevent myself from getting them mixed up.

deployment/helm/charts/onyx/templates/configmap.yaml
eric-zadara: No actually, this should be `.Values.postgresql.enabled`.

This entry in the configmap hardcodes an environment variable to explicitly point to the bitnami postgresql chart.
Regardless of the secret, if the bitnami chart is unused, this should NOT be configured as k8s will error/block chart deployment for conflicting key definitions in the yaml.
Linking it's removal to the secret's presence would force the user to disable the postgresql secret and have to create another secret in the `.auth` section with a different name for the same purpose.

`.Values.auth` is actually not ideal naming, but I was trying not to stray too far from the original chart.
The entirety of `.Values.auth` is responsible for generating multiple secrets and configuring environment variables to point to them. I think a more appropriate label would be `.Values.secrets`.

The postgresql secret may still be used without the bitnami chart, if `.Values.postgresql.enabled` is false but `.Values.auth.postgresql.enabled` is true, the user may now populate `.Values.auth.postgresql.[secretKeys|values]` with another manual solution, such as AWS RDS. Or in my case with cloudnativePG, this would be how I configure `.Values.auth.postgresql.existingSecret` and `.Values.auth.postgresql.secretKeys` to use CNPG's secret and define the host/user/pass/etc environment variables.

I believe this is more clear in my argo examples in the original post, which are effectively some values configs wrapped in an argo definition.

https://github.com/onyx-dot-app/onyx/blob/cc48a0c38eda9f4e8028719bbb4beead970f217d/deployment/helm/charts/onyx/Chart.yaml#L24
https://github.com/onyx-dot-app/onyx/blob/cc48a0c38eda9f4e8028719bbb4beead970f217d/deployment/helm/charts/onyx/Chart.yaml#L28
https://github.com/onyx-dot-app/onyx/blob/cc48a0c38eda9f4e8028719bbb4beead970f217d/deployment/helm/charts/onyx/Chart.yaml#L36
https://github.com/onyx-dot-app/onyx/blob/cc48a0c38eda9f4e8028719bbb4beead970f217d/deployment/helm/charts/onyx/Chart.yaml#L40

deployment/helm/charts/onyx/Chart.yaml
eric-zadara: Of all the conflicts, this is at least the least dangerous :-)

deployment/helm/charts/onyx/values.yaml
justin-tahara: You are correct. This is my bad. I misread the setup. Feel free to resolve with no changes

eric-zadara:
And here's the results of `helm template . > manifest.yaml` as of the latest commit e10cf264a2e3ca7a9859872fea274051e8c1971f in my fork.

[manifest.yaml](https://github.com/user-attachments/files/22223198/manifest.yaml)

justin-tahara:
@greptileai

PR #5208, opened 2025-08-17
fix: sharepoint group not found error and url with apostrophe

Subash-Mohan:
## Description
This pull request enhances SharePoint external permissions handling by improving error handling, logging, and reliability when interacting with SharePoint and Azure AD groups. The most important changes include better handling of missing groups, improved logging for debugging, and a more robust approach to retrieving site page permissions.

## How Has This Been Tested?
Tested by creating the connector

---

## Summary by cubic
Fixes SharePoint permission fetching by skipping deleted Azure AD groups and correctly handling site page URLs with apostrophes. Adds safer retries and logging to improve reliability.

- **Bug Fixes**
  - Skip Azure AD groups that return 404 during traversal; log and continue instead of failing.
  - Resolve site page permission lookups for URLs with apostrophes by using server‑relative paths.
  - Wrap permission queries with retries and null‑safe error checks; improve 429 rate‑limit logging.
  - Add targeted logs for users, members, and role assignments to help debug permission issues.

### DIFF
--- a/backend/ee/onyx/external_permissions/sharepoint/permission_utils.py
+++ b/backend/ee/onyx/external_permissions/sharepoint/permission_utils.py
@@ -234,7 +234,7 @@ def _get_sharepoint_groups(
         nonlocal groups, user_emails
 
         for user in users:
-            logger.info(f"User: {user.to_json()}")
+            logger.debug(f"User: {user.to_json()}")
             if user.principal_type == USER_PRINCIPAL_TYPE and hasattr(
                 user, "user_principal_name"
             ):
@@ -289,7 +289,7 @@ def _get_azuread_groups(
 
         for member in members:
             member_data = member.to_json()
-            logger.info(f"Member: {member_data}")
+            logger.debug(f"Member: {member_data}")
             # Check for user-specific attributes
             user_principal_name = member_data.get("userPrincipalName")
             mail = member_data.get("mail")
@@ -370,6 +370,7 @@ def _get_groups_and_members_recursively(
     client_context: ClientContext,
     graph_client: GraphClient,
     groups: set[SharepointGroup],
+    is_group_sync: bool = False,
 ) -> GroupsResult:
     """
     Get all groups and their members recursively.
@@ -377,6 +378,7 @@ def _get_groups_and_members_recursively(
     group_queue: deque[SharepointGroup] = deque(groups)
     visited_groups: set[str] = set()
     visited_group_name_to_emails: dict[str, set[str]] = {}
+    found_public_group = False
     while group_queue:
         group = group_queue.popleft()
         if group.login_name in visited_groups:
@@ -397,8 +399,14 @@ def _get_groups_and_members_recursively(
             try:
                 # if the site is public, we have default groups assigned to it, so we return early
                 if _is_public_login_name(group.login_name):
-                    return GroupsResult(groups_to_emails={}, found_public_group=True)
-
+                    found_public_group = True
+                    if not is_group_sync:
+                        return GroupsResult(
+                            groups_to_emails={}, found_public_group=True
+                        )
+                    else:
+                        # we don't want to sync public groups, so we skip them
+                        continue
                 group_info, user_emails = _get_azuread_groups(
                     graph_client, group.login_name
                 )
@@ -415,7 +423,8 @@ def _get_groups_and_members_recursively(
                 raise e
 
     return GroupsResult(
-        groups_to_emails=visited_group_name_to_emails, found_public_group=False
+        groups_to_emails=visited_group_name_to_emails,
+        found_public_group=found_public_group,
     )
 
 
@@ -440,7 +449,7 @@ def get_external_access_from_sharepoint(
     ) -> None:
         nonlocal user_emails, groups
         for assignment in role_assignments:
-            logger.info(f"Assignment: {assignment.to_json()}")
+            logger.debug(f"Assignment: {assignment.to_json()}")
             if assignment.role_definition_bindings:
                 is_limited_access = True
                 for role_definition_binding in assignment.role_definition_bindings:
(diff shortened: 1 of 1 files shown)


backend/ee/onyx/external_permissions/sharepoint/permission_utils.py
```diff
@@ -231,6 +234,7 @@ def process_users(users: list[Any]) -> None:
         nonlocal groups, user_emails
 
         for user in users:
+            logger.info(f"User: {user.to_json()}")
```
wenxi-onyx: Is this necessary to log this as info? Is there any sensitive data? Ideally put in debug log or don't log at all

backend/ee/onyx/external_permissions/sharepoint/permission_utils.py
```diff
@@ -285,7 +289,7 @@ def process_members(members: list[Any]) -> None:
 
         for member in members:
             member_data = member.to_json()
-
+            logger.info(f"Member: {member_data}")
```
wenxi-onyx: Same as above comment for users

backend/ee/onyx/external_permissions/sharepoint/permission_utils.py
```diff
@@ -427,6 +440,7 @@ def add_user_and_group_to_sets(
     ) -> None:
         nonlocal user_emails, groups
         for assignment in role_assignments:
+            logger.info(f"Assignment: {assignment.to_json()}")
```
wenxi-onyx: This might generate a lot of logs. Prefer debug

backend/onyx/connectors/sharepoint/connector.py
```diff
@@ -101,7 +101,7 @@ def sleep_and_retry(
             return query_obj.execute_query()
         except ClientRequestException as e:
             if (
-                e.response
+                e.response is not None
```
wenxi-onyx, line 104: Can we use the `@retry_builder` decorator instead of a Sharepoint-specific retry func here?

https://github.com/onyx-dot-app/onyx/blob/main/backend/onyx/utils/retry_wrapper.py#L19

backend/ee/onyx/external_permissions/sharepoint/permission_utils.py
```diff
@@ -390,16 +394,25 @@ def _get_groups_and_members_recursively(
             if group_info:
                 group_queue.extend(group_info)
         if group.principal_type == AZURE_AD_GROUP_PRINCIPAL_TYPE:
-            # if the site is public, we have default groups assigned to it, so we return early
-            if _is_public_login_name(group.login_name):
-                return GroupsResult(groups_to_emails={}, found_public_group=True)
+            try:
+                # if the site is public, we have default groups assigned to it, so we return early
```
wenxi-onyx, line 400: I'm not confident in this logic.
- I created a "Communications Site" which is a public site that has the default public groups --> https://danswerai.sharepoint.com/sites/PublicSitewithPrivateDocs
- I created a custom group called "Wonyx Group"
- I removed public group access from my doc and added Wonyx Group

As it stands, would this document be indexed as a public doc?

backend/ee/onyx/external_permissions/sharepoint/permission_utils.py
```diff
@@ -390,16 +394,25 @@ def _get_groups_and_members_recursively(
...
-            if group_info:
-                group_queue.extend(group_info)
+                group_info, user_emails = _get_azuread_groups(
+                    graph_client, group.login_name
+                )
+                visited_group_name_to_emails[group.name].update(user_emails)
+                if group_info:
+                    group_queue.extend(group_info)
+            except ClientRequestException as e:
+                # If the group is not found, we skip it. There is a chance that group is still referenced
```
wenxi-onyx, line 417: What happens when this is the case? Say a Site/Document is restricted to "SomeInvalidADGroup." With current behavior, does this mean that this doc is indexed with no perm sync groups? Could this result private docs indexing as public?

backend/ee/onyx/external_permissions/sharepoint/permission_utils.py
```diff
@@ -503,12 +517,19 @@ def add_user_and_group_to_sets(
         )
     elif site_page:
         site_url = site_page.get("webUrl")
-        site_pages = client_context.web.lists.get_by_title("Site Pages")
-        client_context.load(site_pages)
-        client_context.execute_query()
-        site_pages.items.get_by_url(site_url).role_assignments.expand(
-            ["Member", "RoleDefinitionBindings"]
-        ).get_all(page_loaded=add_user_and_group_to_sets).execute_query()
+        # Prefer server-relative URL to avoid OData filters that break on apostrophes
```
wenxi-onyx, line 529: Clarification question: Does this end up re-retrieving the contents of the site? If so, does it make sense to do content and groups at the same time (just a thought)?

wenxi-onyx (changes requested):

backend/onyx/connectors/sharepoint/connector.py
Subash-Mohan: The decorator doesn’t capture the `retry-after` header or allow handling specific status codes, so we can’t use it.

backend/ee/onyx/external_permissions/sharepoint/permission_utils.py
Subash-Mohan: No, it won’t fetch the entire content.

backend/ee/onyx/external_permissions/sharepoint/permission_utils.py
Subash-Mohan: If a site or document has only one group and that group is invalid, then the document won’t have any access. But I don’t think that situation will occur—sites and documents usually have multiple groups, and in this case, we’re just skipping the invalid one.

backend/ee/onyx/external_permissions/sharepoint/permission_utils.py
Subash-Mohan: Permission syncing happens individually for each document, so a document won’t end up being public.
However, this causes an issue with group sync—earlier, if we found a public group, we simply returned an empty list. Now, we just skip the public group and continue processing the remaining groups.

PR #5214, opened 2025-08-19
fix(admin): [DAN-2202] Remove users from invited users after accept

rguan72:
## Description

https://linear.app/danswer/issue/DAN-2202/current-users-still-show-up-under-invited-after-accepting-invite

## How Has This Been Tested?

Integration tested in the UI (all login types, very painstakingly)


evan-onyx (approved):
code cleanup stuff that we talked about irl

PR #5268, opened 2025-08-26
feat: add api/versions to onyx

wenxi-onyx:
## Description

- Replicates app.danswer.ai/api/versions in onyx 

## How Has This Been Tested?

---

## Summary by cubic
Adds a public GET /versions endpoint that returns the latest stable and beta Onyx Docker image versions across components, mirroring app.danswer.ai/api/versions. Also whitelists the route in auth to allow unauthenticated reads.

- **New Features**
  - New GET /versions returns AllVersions {stable, dev, migration} with image tags for: danswer (Onyx), relational_db, index, and nginx.
  - Pulls tags from DockerHub for onyxdotapp/onyx-model-server, onyxdotapp/onyx-backend, and onyxdotapp/onyx-web-server.
  - Uses strict patterns (vX.Y.Z and vX.Y.Z-beta.N), finds tags common to all repos, and selects the latest for each stream.
  - Returns fixed companion tags: postgres:15.2-alpine, vespaengine/vespa:8.277.17, nginx:1.23.4-alpine; migration image set to airgapped-intfloat-nomic-migration.

### DIFF
--- a/backend/onyx/configs/app_configs.py
+++ b/backend/onyx/configs/app_configs.py
@@ -1,6 +1,5 @@
 import json
 import os
-import re
 import urllib.parse
 from datetime import datetime
 from datetime import timezone
@@ -26,13 +25,6 @@ APP_PORT = 8080
 APP_API_PREFIX = os.environ.get("API_PREFIX", "")
 
 SKIP_WARM_UP = os.environ.get("SKIP_WARM_UP", "").lower() == "true"
-
-#####
-# Version Pattern Configs
-#####
-# Version patterns for Docker image tags
-STABLE_VERSION_PATTERN = re.compile(r"^v(\d+)\.(\d+)\.(\d+)$")
-DEV_VERSION_PATTERN = re.compile(r"^v(\d+)\.(\d+)\.(\d+)-beta\.(\d+)$")
 
 #####
 # User Facing Features Configs
--- a/backend/onyx/configs/app_configs.py
+++ b/backend/onyx/configs/app_configs.py
@@ -1,5 +1,6 @@
 import json
 import os
+import re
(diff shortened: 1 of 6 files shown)


backend/onyx/configs/app_configs.py
```diff
@@ -26,6 +27,13 @@
 
 SKIP_WARM_UP = os.environ.get("SKIP_WARM_UP", "").lower() == "true"
 
+#####
+# Version Pattern Configs
+#####
+# Version patterns for Docker image tags
```
Weves: nit: should not be in configs, these are all meant to be configuable

backend/onyx/server/manage/get_state.py
```diff
@@ -28,3 +37,102 @@ def get_auth_type() -> AuthTypeResponse:
 @router.get("/version")
 def get_version() -> VersionResponse:
     return VersionResponse(backend_version=__version__)
+
+
+@router.get("/versions")
+def get_versions() -> AllVersions:
+    """
+    Fetches the latest stable and beta versions of Onyx Docker images
```
Weves: nit: can we add a comment as to why this is here?

backend/onyx/configs/app_configs.py
wenxi-onyx: How about `backend/onyx/configs/constants.py`?

backend/onyx/server/manage/get_state.py
wenxi-onyx: done

backend/onyx/configs/app_configs.py
Weves: sounds good

PR #5275, opened 2025-08-27
feat(slack bot): add federated search

jessicasingh7:
## Description

[DAN-2328](https://linear.app/danswer/issue/DAN-2328/add-support-for-slack-search-into-the-slack-bot)

## How Has This Been Tested?

---

## Summary by cubic
Adds federated Slack search to the Slack bot using a user OAuth token, enabling public-channel search results in chat. This delivers the core of DAN-2328 for public channels; channel scoping and private/DM coverage will follow.

- **New Features**
  - Added user_token to Slack bot (DB, models, API, UI) and server-side validation (requires xoxp- and Slack auth.test).
  - Slack OAuth now requests both bot scopes and user scopes; callback returns bot access_token and user_token.
  - Federated retrieval uses Slack search in bot context when an enabled Slack bot has a user_token.
  - Slack retrieval uses search.messages for public channels, improves logging, username fallback, and skips thread expansion with user tokens to avoid extra scopes.
  - New SlackSearch tool for the bot to fetch and return Slack snippets.
  - Small fixes: timezone handling in Slack blocks; typing/instance checks.

- **Migration**
  - Run backend DB migrations.
  - In Slack: add user scopes search:read.public and channels:history, then reinstall to obtain a user OAuth token.
  - In Admin > Bots, add the optional “Slack User Token” (xoxp-). Without it, bot-context federated Slack search is disabled.

### DIFF
--- a/backend/onyx/connectors/slack/onyx_retry_handler.py
+++ b/backend/onyx/connectors/slack/onyx_retry_handler.py
@@ -103,11 +103,18 @@ class OnyxRedisSlackRetryHandler(RetryHandler):
                     "OnyxRedisSlackRetryHandler.prepare_for_next_attempt: retry-after header name is None"
                 )
 
-            retry_after_value = response.headers.get(retry_after_header_name)
-            if not retry_after_value:
+            retry_after_header_value = response.headers.get(retry_after_header_name)
+            if not retry_after_header_value:
                 raise ValueError(
                     "OnyxRedisSlackRetryHandler.prepare_for_next_attempt: retry-after header value is None"
                 )
+
+            # Handle case where header value might be a list
+            retry_after_value = (
+                retry_after_header_value[0]
+                if isinstance(retry_after_header_value, list)
+                else retry_after_header_value
+            )
 
             retry_after_value_int = int(
                 retry_after_value


.prettierignore
```diff
@@ -1 +1,2 @@
 backend/tests/integration/tests/pruning/website
+web/src/app/chat/utils/thinkingTokens.ts
```
Weves: what's this one about 🧐

backend/alembic/versions/5ae8240accb3_add_research_agent_database_tables_and_.py
```diff
@@ -0,0 +1,115 @@
+"""add research agent database tables and chat message research fields
```
Weves, line 1: are we supposed to have these migrations?

backend/tests/external_dependency_unit/slack_bot/test_slack_bot_federated_search.py
```diff
@@ -0,0 +1,518 @@
...
+            post_message_calls = mock_client.web_client.chat_postMessage.call_args_list
+            last_call = post_message_calls[-1]
+            assert (
+                last_call[1]["channel"] == channel_id
+            ), f"Response should be sent to {channel_id}"
+
+            response_text = last_call[1].get("text", "")
+            assert len(response_text) > 0, "Bot should have sent a non-empty response"
+
+            response_text = last_call[1].get("text", "")
```
Weves: nit: repeated

backend/tests/external_dependency_unit/slack_bot/test_slack_bot_federated_search.py
```diff
@@ -0,0 +1,518 @@
...
+            ), "Should be testing dev-team channel"
+
+        finally:
+            self._teardown_common_mocks(patches)
+
+    def test_slack_bot_dm_filtering(self, db_session: Session) -> None:
+        """Test that slack bot in DM sees all messages (no filtering)"""
+        self._setup_test_environment(db_session)
+
+        channel_id = "D1234567890"  # DM
```
Weves, line 536: is it true that dms still have a channel ID? If so, can we add a comment that states that + links the relevant docs, since it's a bit confusing

backend/tests/external_dependency_unit/slack_bot/test_slack_bot_federated_search.py
```diff
@@ -0,0 +1,518 @@
...
+            post_message_calls = mock_client.web_client.chat_postMessage.call_args_list
+            last_call = post_message_calls[-1]
+            assert (
+                last_call[1]["channel"] == channel_id
+            ), f"Response should be sent to {channel_id}"
+
+            response_text = last_call[1].get("text", "")
+            assert len(response_text) > 0, "Bot should have sent a non-empty response"
+
+            response_text = last_call[1].get("text", "")
```
Weves: nit: duplicate

backend/onyx/onyxbot/slack/listener.py
```diff
@@ -658,9 +659,21 @@ def prefilter_requests(req: SocketModeRequest, client: TenantSocketModeClient) -
...
+        # 1. app_mention - when someone uses @OnyxBot
+        # 2. message in DMs (channel_type == "im") - when someone messages the bot directly
+        if event_type == "app_mention":
+            # Someone used @OnyxBot - allow this
+            pass
+        elif event_type == "message" and channel_type == "im":
+            # This is a DM with the bot - allow this
+            pass
+        else:
+            # Block everything else: regular messages in channels, other event types, etc.
```
Weves: we also allow people to set bots up that respond to every message in regular channels, so we should still allow this

backend/onyx/onyxbot/slack/listener.py
```diff
@@ -864,6 +877,17 @@ def build_request_details(
         if tagged:
             logger.debug("User tagged OnyxBot")
 
+        # Build Slack context for federated search
+        # Get proper channel type from Slack API instead of relying on event.channel_type
+        channel_type = get_channel_type_from_id(client.web_client, channel)
+
+        slack_context = {
```
Weves: can we make this a pydantic model and use that everywhere instead of `dict[str, str]`

backend/onyx/onyxbot/slack/utils.py
```diff
@@ -83,6 +83,39 @@ def get_onyx_bot_auth_ids(
     return user_id, bot_id
 
 
+def get_channel_type_from_id(web_client: WebClient, channel_id: str) -> str:
```
Weves: nit: can we make this return a str Enum? e.g. `class ChannelType(str, Enum):`

backend/onyx/federated_connectors/federated_retrieval.py
```diff
@@ -35,11 +37,78 @@ def get_federated_retrieval_functions(
     user_id: UUID | None,
     source_types: list[DocumentSource] | None,
     document_set_names: list[str] | None,
+    slack_context: dict[str, str] | None = None,  # Add Slack context parameter
 ) -> list[FederatedRetrievalInfo]:
+    # Check for Slack bot context first (regardless of user_id)
+    if slack_context:
```
Weves, line 42: could we move all of this logic into the slack/federated_connector.py?

Weves (commented):
@jessicasingh7 overall looks great! Left a few initial comments, let me know if you have any questions about them.

.prettierignore
jessicasingh7: oh yes I always get this file whenever I run precommit hooks

backend/alembic/versions/5ae8240accb3_add_research_agent_database_tables_and_.py
jessicasingh7: oops sorry removed

backend/tests/external_dependency_unit/slack_bot/test_slack_bot_federated_search.py
jessicasingh7: Yes of course, added!

backend/onyx/federated_connectors/federated_retrieval.py
jessicasingh7: Used search function here instead as per our discussion

backend/onyx/configs/constants.py
```diff
@@ -207,7 +208,7 @@ class DocumentSource(str, Enum):
 class FederatedConnectorSource(str, Enum):
     FEDERATED_SLACK = "federated_slack"
 
-    def to_non_federated_source(self) -> DocumentSource | None:
+    def to_non_federated_source(self) -> Union[DocumentSource, None]:
```
evan-onyx: nit: prefer the way it was before

package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
evan-onyx: What's this file for?

backend/onyx/connectors/slack/onyx_retry_handler.py
```diff
@@ -75,7 +75,7 @@ def prepare_for_next_attempt(
         """
         ttl_ms: int | None = None
 
-        retry_after_value: list[str] | None = None
+        retry_after_value: list[str] | str | None = None
```
evan-onyx: Why is this change needed? If possible I'd like to keep it as it was to reduce type confusion

backend/onyx/context/search/federated/slack_search.py
```diff
@@ -64,11 +66,39 @@ def build_slack_queries(query: SearchQuery, llm: LLM) -> list[str]:
...
+    is_mpim = channel.get("is_mpim", False)
+    is_im = channel.get("is_im", False)
+
+    # A public channel is: a channel that is NOT private, NOT a group, NOT mpim, NOT im
+    is_public = (
+        is_channel and not is_private and not is_group and not is_mpim and not is_im
+    )
+
+    # Add detailed logging for debugging
+    logger.debug(
```
evan-onyx: somewhat worried this could get spammy; is it true that this log will fire once per public channel per slackbot response?

backend/onyx/context/search/federated/slack_search.py
```diff
@@ -64,11 +66,39 @@ def build_slack_queries(query: SearchQuery, llm: LLM) -> list[str]:
...
+
+    return is_public
+
+
 def query_slack(
     query_string: str,
     original_query: SearchQuery,
     access_token: str,
     limit: int | None = None,
+    allowed_private_channel: str | None = None,  # Add allowed private channel parameter
```
evan-onyx: nit: remove the first two comments here

backend/onyx/context/search/federated/slack_search.py
```diff
@@ -103,6 +150,74 @@ def query_slack(
...
+                    logger.warning(
+                        f"Could not determine channel type for {channel_id}, filtering out: {e}"
+                    )
+                    filtered_count += 1
+                    continue
+        elif include_dm:
+            # Include direct messages - no filtering needed
+            pass
+        elif (
+            allowed_private_channel is None and not include_dm and bot_token is not None
```
evan-onyx: should be able to simplify this to `bot_token is not None` bc its in an `elif`

backend/onyx/context/search/federated/slack_search.py
```diff
@@ -103,6 +150,74 @@ def query_slack(
         ):
             continue
 
+        # Apply channel filtering if needed
```
evan-onyx, line 183: nit: put this filtering code in a helper function named like _should_skip_channel()

backend/onyx/context/search/federated/slack_search.py
```diff
@@ -103,6 +150,74 @@ def query_slack(
...
+        if allowed_private_channel is not None:
```
evan-onyx: When an allowed_private_channel is passed in, it's the only filtering applied; is this intentional? If so would be good to have a comment explaining that in the docstring or next to the param!

backend/onyx/context/search/federated/slack_search.py
```diff
@@ -103,6 +150,74 @@ def query_slack(
...
+            # Private channel context: only allow the specific private channel + public channels
+            if channel_id == allowed_private_channel:
+                # This is the allowed private channel - keep it
+                pass
```
evan-onyx: nit: prefer to avoid `pass`; can invert the condition and use the else case body

backend/onyx/context/search/federated/slack_search.py
```diff
@@ -103,6 +150,74 @@ def query_slack(
...
+            else:
+                # Check if this is a public channel
+                try:
+                    # Use bot token if available (has full permissions), otherwise fall back to user token
+                    token_to_use = bot_token if bot_token else access_token
```
evan-onyx: nit: `bot_token or access_token`

backend/onyx/context/search/federated/slack_search.py
```diff
@@ -103,6 +150,74 @@ def query_slack(
...
+        elif include_dm:
+            # Include direct messages - no filtering needed
+            pass
+        elif (
+            allowed_private_channel is None and not include_dm and bot_token is not None
+        ):
+            # Slack bot context (has bot_token but no specific channel context): apply default filtering (only public channels)
+            try:
+                # Use bot token if available (has full permissions), otherwise fall back to user token
+                token_to_use = bot_token if bot_token else access_token
```
evan-onyx: nit: can directly use `bot_token` in this case since the if case guarantees it isn't None

backend/onyx/context/search/federated/slack_search.py
```diff
@@ -291,14 +411,40 @@ def slack_retrieval(
     access_token: str,
     db_session: Session,
     limit: int | None = None,
+    slack_event_context: SlackContext | None = None,
+    bot_token: str | None = None,  # Add bot token parameter
```
evan-onyx, line 384: nit: remove comment

backend/onyx/context/search/federated/slack_search.py
```diff
@@ -307,15 +453,22 @@ def slack_retrieval(
...
-    thread_texts: list[str] = run_functions_tuples_in_parallel(
-        [
-            (get_contextualized_thread_text, (slack_message, access_token))
-            for slack_message in slack_messages
-        ]
-    )
-    for slack_message, thread_text in zip(slack_messages, thread_texts):
-        slack_message.text = thread_text
+    # Check if we're in a bot context by looking at the access token prefix
+    if access_token.startswith("xoxp-"):
```
evan-onyx: nit: make this a string constant like SLACK_BOT_TOKEN_PREFIX or something similar

backend/onyx/context/search/federated/slack_search.py
```diff
@@ -307,15 +453,22 @@ def slack_retrieval(
...
+        logger.info(
+            "Bot context detected (user OAuth token): skipping thread context to avoid additional scope requirements"
```
evan-onyx: For my own knowledge: do we just not need the contextualized thread text when running the slackbot?

backend/onyx/db/slack_bot.py
```diff
@@ -78,11 +82,12 @@ def fetch_slack_bots(db_session: Session) -> Sequence[SlackBot]:
 
 def fetch_slack_bot_tokens(
     db_session: Session, slack_bot_id: int
-) -> dict[str, str] | None:
+) -> dict[str, str | None] | None:
```
evan-onyx: I don't think this function is actually used (from before this PR too), we should probably delete it

backend/onyx/federated_connectors/federated_retrieval.py
```diff
@@ -60,11 +132,12 @@ def get_federated_retrieval_functions(
             pair
         )
 
+    # At this point, user_id is guaranteed to be not None since we're in the else branch
+    assert user_id is not None
```
evan-onyx, line 136: Does removing this cause mypy to complain? It _should_ be fine without this assert, but of course if we need to keep it then no problem

backend/onyx/server/manage/validate_tokens.py
```diff
@@ -41,3 +41,38 @@ def validate_app_token(app_token: str) -> bool:
         )
 
     return True
+
+
+def validate_user_token(user_token: str | None) -> bool:
+    """
+    Validate that the user_token is a valid user OAuth token (xoxp-...)
+    and not a bot token (xoxb-...)
```
evan-onyx, line 51: nit: these prefixes should probably be defined in an existing models, utils, or consts file in case they ever change

backend/onyx/server/manage/validate_tokens.py
```diff
@@ -41,3 +41,38 @@ def validate_app_token(app_token: str) -> bool:
...
+    """
+    if user_token is None:
+        # user_token is optional, so None is valid
+        return True
```
evan-onyx: nit: given that we're raising errors instead of returning False when validation fails, would prefer this function returning None. And then we can add to the docstring something like "raises a HttpException if validation fails"

backend/onyx/server/manage/validate_tokens.py
```diff
@@ -41,3 +41,38 @@ def validate_app_token(app_token: str) -> bool:
         )
 
     return True
+
+
+def validate_user_token(user_token: str | None) -> bool:
+    """
+    Validate that the user_token is a valid user OAuth token (xoxp-...)
```
evan-onyx, line 50: probably should mention that None is valid too, as I don't think people would assume that

backend/onyx/tools/tool_implementations/images/image_generation_tool.py
```diff
@@ -348,7 +348,10 @@ def generate_all_images() -> None:
             # Yield a heartbeat packet to prevent timeout
             yield ToolResponse(
                 id=IMAGE_GENERATION_HEARTBEAT_ID,
-                response=None,
+                response={
```
evan-onyx, line 370: this doesn't seem related to this PR?

backend/onyx/context/search/federated/slack_search.py
jessicasingh7: Sure, removed!

package-lock.json
jessicasingh7: not needed, deleted!

backend/onyx/context/search/federated/slack_search.py
jessicasingh7: that's a good call out I was avoiding it for user context actually but I'll keep it to be consistent with how it was previously

backend/onyx/tools/tool_implementations/images/image_generation_tool.py
jessicasingh7: needed it for a failing unit test!

backend/onyx/federated_connectors/federated_retrieval.py
jessicasingh7: yep unfortunately mypy was complaining

backend/onyx/connectors/slack/onyx_retry_handler.py
jessicasingh7: oh yes I think it should be str | None , I think earlier mypy automatically added list[str] and str

jessicasingh7:
Slack bot- fed search

PR #5281, opened 2025-08-29
fix: web search bugs [DAN-2351]

rguan72:
## Description

1. Pass branch questions to deep research
2. Deduplicate urls before fetching
3. Fix duplicate section end sent
4. Code cleanup -- make required fields required

Instead of Query -> Urls, do Urls -> Query

Then can stream individual urls to update the UI + avoid dupes in fetching but still pass proper context to subquestions

Only stream deduped urls

Thoughtful mode and Deep research

Context window should contain all the original documents

Context window should also contain arbitrary mappings from citations to original documents [so we don't have to duplicate them in context]

Make sure research_agent_iteration and research_agent_iteration_sub_step tables are as expected

## How Has This Been Tested?

Locally

relevant linear: https://linear.app/danswer/issue/DAN-2351/duplicate-urls-sometimes-searched
https://linear.app/danswer/issue/DAN-2350/sub-search-queries-arent-being-passed-to-deep-research-summarization

---

## Summary by cubic
Deduplicates URLs and routes fetches per query in the internet search pipeline to reduce redundant requests and improve accuracy. Refactors state handling and graph edges to pass query–URL pairs through a new dedup step.

- **Refactors**
  - Added InternetSearchInput and FetchInput states; moved internet search–specific state to a new states module.
  - New dedup_urls node groups unique URLs by originating query.
  - Updated routers: search -> dedup_urls; dedup_urls -> fetch; adjusted fallback to reducer.
  - web_search now returns query–URL pairs and InternetSearchInput; BranchUpdate no longer carries urls_to_open.
  - Enforced BranchInput.branch_question as required; removed nullable usage and empty-string fallbacks in fetch.

- **Bug Fixes**
  - Prevents duplicate URL fetches across parallel queries.
  - Ensures prompts and IterationAnswer use the actual branch_question.
  - Fixes fetch routing to process URLs per query instead of arbitrary pairing.


backend/onyx/agents/agent_search/dr/sub_agents/web_search/utils.py
```diff
@@ -0,0 +1,77 @@
...
+            metadata={},
+            match_highlights=[],
+            doc_summary="",
+            chunk_context="",
+            updated_at=result.published_date,
+            image_file_id=None,
+        ),
+        chunks=[],
+        combined_content="",
+    )
```
joachim-danswer, line 77: Looks good. We had discussed a number of recommended changes which I see implemented. I also ran it locally.

joachim-danswer (approved):
Looks good. We had discussed a number of recommended changes which I see implemented. I also ran it locally.

PR #5298, opened 2025-08-31
fix: make radios and checkboxes actually clickable

wenxi-onyx:
## Description

- Original issue: certain radios/checkboxes required users to click specifically just the box component

- Nested labels cause text to not be clickable
- Introduce htmlFor that associates html elements with the form input item using a unique id
- CheckField.tsx had bugged logic that reset the form value on click

## How Has This Been Tested?

- locally


evan-onyx (approved):
l e g e n d a r y

PR #5323, opened 2025-09-03
Bump LiteLLM version, fix introduced issue about concatenating sets

emerzon:
## Description
* Bumps LiteLLM to latest version, updated model DB with newer entries;
* LiteLLM changed the type of model lists from list to set, causing issues when they are concatenated. Fixed.
    

---

## Summary by cubic
Upgrade LiteLLM to 1.76.1 and fix Bedrock model aggregation for the new set types. Prevents runtime errors from list concatenation and keeps model data current.

- **Bug Fixes**
  - Replace list concatenation with set union when combining bedrock_models and bedrock_converse_models.

- **Dependencies**
  - Bump litellm from 1.72.2 to 1.76.1 in default and model_server requirements.

### DIFF
--- a/backend/requirements/default.txt
+++ b/backend/requirements/default.txt
@@ -52,7 +52,7 @@ msal==1.28.0
 nltk==3.9.1
 Office365-REST-Python-Client==2.5.9
 oauthlib==3.2.2
-openai==1.75.0
+openai==1.106.1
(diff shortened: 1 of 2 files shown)


emerzon:
Deprecated by #5349

PR #5330, opened 2025-09-04
fix: various bug bash improvements

rguan72:
## Description

1. Fix inline math rendering issue (re-enable LaTeX math rendering by replacing `\(...\)` with `$$...$$` but leaving `$...$` inline latex rendering disabled which helps reduce mistakes with `$` in variable names or referencing money 
2. Newly created assistants pinned by default
3. Simplified Azure form

https://linear.app/danswer/issue/DAN-2370/when-creating-a-new-assistant-it-should-be-default-pinned-right-after

## How Has This Been Tested?

tested locally

### DIFF
--- a/web/src/app/admin/configuration/llm/LLMProviderUpdateForm.tsx
+++ b/web/src/app/admin/configuration/llm/LLMProviderUpdateForm.tsx
@@ -58,6 +58,13 @@ export function LLMProviderUpdateForm({
     api_key: existingLlmProvider?.api_key ?? "",
     api_base: existingLlmProvider?.api_base ?? "",
     api_version: existingLlmProvider?.api_version ?? "",
+    // For Azure OpenAI, combine api_base and api_version into target_uri
+    target_uri:
+      llmProviderDescriptor.name === "azure" &&
+      existingLlmProvider?.api_base &&
+      existingLlmProvider?.api_version
+        ? `${existingLlmProvider.api_base}/openai/deployments/your-deployment?api-version=${existingLlmProvider.api_version}`
+        : "",
     default_model_name:
       existingLlmProvider?.default_model_name ??
       (llmProviderDescriptor.default_model ||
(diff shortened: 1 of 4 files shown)


web/src/app/admin/configuration/llm/LLMProviderUpdateForm.tsx
```diff
@@ -98,12 +105,42 @@ export function LLMProviderUpdateForm({
     api_key: llmProviderDescriptor.api_key_required
       ? Yup.string().required("API Key is required")
       : Yup.string(),
-    api_base: llmProviderDescriptor.api_base_required
-      ? Yup.string().required("API Base is required")
-      : Yup.string(),
-    api_version: llmProviderDescriptor.api_version_required
-      ? Yup.string().required("API Version is required")
-      : Yup.string(),
+    api_base:
```
wenxi-onyx, line 108: @rguan72 you'll want to make this change to llm_provider_options.py --> the WellKnownLLMProviderDescriptor for Azure

web/src/app/admin/configuration/llm/LLMProviderUpdateForm.tsx
```diff
@@ -98,12 +105,42 @@ export function LLMProviderUpdateForm({
...
+      llmProviderDescriptor.api_base_required &&
+      llmProviderDescriptor.name !== "azure"
```
rguan72, line 110: chatted sync with Wenxi -- right now, the abstraction assumes 1:1 between UI and back end specified requirements, but we're creating divergence in this case (also with AWS bedrock). IMO it's no longer the right abstraction -- if we want it to be back end driven, we need different fields for requirements + UI displays and some mapping between UI displays and requirements. If we want it to be front end driven, we need to hardcode providers. Or we can do something in between.

I think it's ok to hardcode providers

Weves (approved):

PR #5420, opened 2025-09-15
Added OpenRouter as model provider

staru09:
## Description

This PR adds OpenRouter as a dedicated provider in Onyx, replacing the current workaround of using the "Custom Inference Provider." The integration uses our existing LiteLLM infrastructure without adding any extra dependency.

## How Has This Been Tested?

- Built locally and tested OpenRouter provider end-to-end
- Verified API key configuration and model fetching
- Tested chat functionality with various OpenRouter models
- Confirmed existing providers remain unaffected


staru09:
custom-model
multi-llm
api
main

here are the screenshots of the change

PR #5425, opened 2025-09-16
fix: Improve datetime replacement

Weves:
## Description

## How Has This Been Tested?

---

## Summary by cubic
Replaces [[CURRENT_DATETIME]] in DR system prompts with the current timestamp (YYYY-MM-DD HH:MM:SS) and applies it to both persona and default prompts in the clarifier. This prevents the placeholder from leaking and ensures time-aware responses.


backend/tests/external_dependency_unit/answer/test_current_datetime_replacement.py
```diff
@@ -0,0 +1,109 @@
...
+from onyx.db.chat import create_chat_session
+from onyx.db.models import User
+from onyx.db.persona import get_persona_by_id
+from onyx.server.query_and_chat.models import CreateChatMessageRequest
+from onyx.server.query_and_chat.streaming_models import MessageDelta
+from tests.external_dependency_unit.answer.conftest import ensure_default_llm_provider
+from tests.external_dependency_unit.conftest import create_test_user
+
+
+def test_stream_chat_current_date_response(
```
rguan72, line 20: Interesting! I've always avoided tests that hit external APIs (seems we're hitting an LLM api) but as a smoke test it's probably pretty reasonable

I guess integration tests also hit external APIs

rguan72 (approved):

PR #5427, opened 2025-09-16
fix(slack): swapped checkpoint index

jessicasingh7:
## Description

## How Has This Been Tested?

### DIFF
--- a/backend/onyx/connectors/slack/connector.py
+++ b/backend/onyx/connectors/slack/connector.py
@@ -837,7 +837,7 @@ class SlackConnector(
 
             channel_message_ts = checkpoint.channel_completion_map.get(channel_id)
             if channel_message_ts:
-                latest = channel_message_ts
+                oldest = channel_message_ts
 
             logger.debug(
                 f"Getting messages for channel {channel} within range {oldest} - {latest}"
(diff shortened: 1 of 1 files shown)


backend/onyx/connectors/slack/connector.py
```diff
@@ -837,7 +837,7 @@ def _load_from_checkpoint(
 
             channel_message_ts = checkpoint.channel_completion_map.get(channel_id)
             if channel_message_ts:
-                latest = channel_message_ts
+                oldest = channel_message_ts
```
Weves, line 841: can we add a comment here as to why we updated `oldest` rather than updating `latest`?

backend/onyx/connectors/slack/connector.py
```diff
@@ -855,7 +855,7 @@ def _load_from_checkpoint(
                 f"{latest=}"
             )
 
-            new_latest = message_batch[-1]["ts"] if message_batch else latest
+            new_latest = message_batch[0]["ts"] if message_batch else latest
```
Weves: this is now `new_oldest`, right?

Weves: I think some of the logic below (e.g. `new_latest_seconds_epoch = SecondsSinceUnixEpoch(new_latest)` and below) might be wrong now?

backend/onyx/connectors/slack/connector.py
jessicasingh7: Done!

backend/onyx/connectors/slack/connector.py
jessicasingh7: Yep resolved!

Weves (approved):

PR #5392, opened 2025-09-11
feat: Add Jira Service Management Connector

melmathari:
/claim #2281

## Description

This PR adds a new connector for Jira Service Management (JSM).

The connector uses the standard Jira API instead of JSM-specific APIs because:
- JSM tickets are built on Jira Core infrastructure 
- Jira API provides better bulk document retrieval capabilities
- JQL search is more powerful for indexing operations
- Maintains consistency with existing Jira connector patterns

The connector filters to only service desk projects using `projectType = 'service_desk'` JQL to prevent overlap with the existing Jira connector.

Features:
- Support for specific project or all service desk projects
- JSM-specific credential fields to avoid confusion with regular Jira

---

- [ ] This PR should be backported (make sure to check that the backport attempt succeeds)  
- [x] [Optional] Override Linear Check  

### VIDEO

ALL + PROJECT ID = https://www.youtube.com/watch?v=J7Piw9fgmBU
JQL = https://www.youtube.com/watch?v=J7Piw9fgmBU

## How Has This Been Tested?

We tested this with real JSM tickets and made sure it:
- Only grabs customer support tickets (not development tickets) with `projectType = 'service_desk'` JQL filtering
- Works with different company setups
- Handles errors - retry logic and error handling
- Verified document retrieval and indexing with demo tickets
- No overlap with existing Jira connector as it has separate credential fields and service desk filtering
- Time range handling

### Backend API Test Output:
```
=== JSM CONNECTOR API TEST OUTPUT ===
Generated JQL Query: project in projectsWhereUserHasPermission('Browse Projects') AND projectType = 'service_desk' AND updated >= '1970-01-01 00:00' AND updated <= '2025-09-11 10:35'
Found 5 issues
Issue: DEMO-5 - I need help connecting my laptop to the level 3 printer
  Created: 2025-09-11T09:02:31.397+0200
  Updated: 2025-09-11T09:02:32.936+0200
  Project: DEMO (Demo service project)
  Status: Waiting for support

Issue: DEMO-4 - Collecting custom request details
  Created: 2025-09-11T09:02:30.576+0200
  Updated: 2025-09-11T09:02:31.157+0200
  Project: DEMO (Demo service project)
  Status: Waiting for support

Issue: DEMO-3 - Agents & customers
  Created: 2025-09-11T09:02:29.884+0200
  Updated: 2025-09-11T09:02:30.347+0200
  Project: DEMO (Demo service project)
  Status: Waiting for support

[...]

PR #5437, opened 2025-09-17
chore: [Running GitHub actions for #5392]

wenxi-onyx:
Automated mirror of PR #5392 so private CI can run.


melmathari:
Thanks for the review @wenxi-onyx do you need anything from my end?

melmathari:
@wenxi-onyx I've fixed the TypeScript error by adding the missing jira_service_management property to the autoSyncConfigBySource object in my original PR

PR #5442, opened 2025-09-17
fix(Federated Slack): Persist Document Set for Federated Connectors

justin-tahara:
## Description

A customer was noticing that the Federated Slack Connector being the only connector in the document set would cause the document set to get deleted or disappear immediately. This is a bug in which we deleted any document set that was just a federated slack connector since it wouldn't have a `connector_credential_pair`. This PR aims to fix this and allow the Document Set to persist if it is only a Federated Slack Connector. 

## How Has This Been Tested?

Tested this locally as well as created tests to ensure that this is tracked and checked when we have any new Federated Connectors. 

---

## Summary by cubic
Fixes deletion of document sets that only include Federated connectors (e.g., Federated Slack). These sets now persist without connector_credential_pairs; only truly empty sets are deleted.

- **Bug Fixes**
  - Updated monitor_document_set_taskset to keep document sets alive if federated_connectors are present; delete only when both connector_credential_pairs and federated_connectors are empty.
  - Added unit tests for federated-only persistence and empty-set deletion.

### DIFF
--- a/backend/onyx/background/celery/tasks/vespa/tasks.py
+++ b/backend/onyx/background/celery/tasks/vespa/tasks.py
@@ -414,8 +414,14 @@ def monitor_document_set_taskset(
         get_document_set_by_id(db_session=db_session, document_set_id=document_set_id),
     )  # casting since we "know" a document set with this ID exists
     if document_set:
-        if not document_set.connector_credential_pairs:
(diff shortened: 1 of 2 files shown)


jessicasingh7 (approved):
LGTM!

PR #5453, opened 2025-09-18
fix(infra): create pre commit script and port vertex as lazy import

edwin-onyx:
## Description

created pre commit script to prevent brittleness of inline lazy imports it make sure that packages which are lazy loaded must not be imported directly

couldnt find any solid lazy loading python libraries, top results that come up r smth like this https://pypi.org/project/lazy-imports/ which has liek 10 starts, i think just doing python native and having script to enforce is prob best?

ported over the vertex ai import to this structure and tested using script in https://github.com/onyx-dot-app/onyx/pull/5456/files

per celery worker ->
without vertex ai lazy loading-Peak RSS Memory: 796.9 MB
with - Peak RSS Memory: 653.0 MB, 18% decrease

moving forward will start porting other big packages / connector specific pkgs that likely won't be used by core application flow onto new lazy loading abstraction to further reduce mem usage

https://linear.app/danswer/issue/DAN-2558/investigate-lazy-loading

## How Has This Been Tested?

created unit tests for both lazy loading abstraction and pre commit script

---

## Summary by cubic
Introduces lazy imports to reduce baseline memory and only load heavy Google/VertexAI libs when needed, addressing DAN-2558 (Reduce Memory usage in Onyx). Adds tools to measure Celery worker memory and per-package import cost.

- **Refactors**
  - Added lazy_imports.py with LazyImport/LazyClass and wrappers for googleapiclient, google.auth, and vertexai.
  - Switched Gmail and Google Drive connectors plus google_utils (auth/resources) to lazy imports; routed HttpError/RefreshError via wrappers; replaced Resource typing with light placeholders; tightened types with Union and TYPE_CHECKING.
  - Vertex AI embeddings now lazily load service account creds and language_models; behavior unchanged.

- **New Features**
  - measure_celery_memory.py: launch a Celery worker and sample RSS/USS/PSS with a summary.
  - measure_package_size.py: import packages from requirements and report per-import memory deltas.


backend/onyx/lazy_handling/lazy_import_registry.py
```diff
@@ -0,0 +1,8 @@
+from typing import TYPE_CHECKING
+
+from onyx.lazy_handling.lazy_module import LazyModule
+
+if TYPE_CHECKING:
+    import vertexai  # type: ignore[import-untyped]
+
+lazy_vertexai: "vertexai" = LazyModule("vertexai")  # noqa: F821
```
Weves: why is `# noqa: F821` needed?

backend/onyx/lazy_handling/lazy_module.py
```diff
@@ -0,0 +1,33 @@
+import importlib
+import threading
+from typing import Any
+
+
+class LazyModule:
```
Weves: does this handle both `from X import Y` and `import X`?

backend/onyx/lazy_handling/lazy_module.py
```diff
@@ -0,0 +1,33 @@
...
+        if self._import_failed:
+            raise ImportError(f"Previous import of '{self.module_name}' failed")
+
+        if self._module is None:
+            with self._lock:
+                if self._module is None and not self._import_failed:
+                    try:
+                        self._module = importlib.import_module(self.module_name)
+                    except (ImportError, ValueError) as e:
+                        self._import_failed = True
```
Weves: nit: maybe store the exception and re-raise it on repeated failure to more closely mimic the normal behavior?

backend/onyx/lazy_handling/lazy_module.py
Weves: also, I assume you've already looked but I'm curious why not use an existing lazy load library/other approach?

Maybe could add a quick comment here on why this approach is best/preferred

backend/tests/unit/onyx/lazy_handling/test_lazy_module.py
```diff
@@ -0,0 +1,367 @@
+"""
```
Weves: Not an expert on imports, but `from vertexai.language_models import TextEmbeddingInput` works but:

```
import vertexai

vertexai.language_models.TextEmbeddingInput
```

doesn't work. Since I think the `LazyModule` doesn't support the former, this might be a problem?

Weves (approved):

PR #5458, opened 2025-09-19
fix: braintrust masking was over truncating

rguan72:
## Description

was masking to 1000 characters (too aggressive)

## How Has This Been Tested?

tested locally

---

## Summary by cubic
Fixes over-aggressive Braintrust tracing masking to preserve useful context while still limiting very large payloads. Increases the mask threshold to 20,000 characters and applies masking only when needed.

- **Bug Fixes**
  - Raised masking length to 20,000 chars with an 80/20 head/tail split.
  - Added a guard to skip masking for payloads under the threshold.
  - Standardized the truncation marker to include original length and target size.

### DIFF
--- a/backend/onyx/evals/tracing.py
+++ b/backend/onyx/evals/tracing.py
@@ -7,18 +7,24 @@ from braintrust_langchain.callbacks import BraintrustCallbackHandler
 from onyx.configs.app_configs import BRAINTRUST_API_KEY
 from onyx.configs.app_configs import BRAINTRUST_PROJECT
 
+MASKING_LENGTH = 20000
 
-def _truncate_str(s: str, head: int = 800, tail: int = 200) -> str:
-    if len(s) <= head + tail:
-        return s
-    return f"{s[:head]}…{s[-tail:]}[TRUNCATED {len(s)} chars to 10,000]"
+
+def _truncate_str(s: str) -> str:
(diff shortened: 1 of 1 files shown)


backend/onyx/evals/tracing.py
```diff
@@ -7,18 +7,24 @@
...
 
-def _truncate_str(s: str, head: int = 800, tail: int = 200) -> str:
-    if len(s) <= head + tail:
-        return s
-    return f"{s[:head]}…{s[-tail:]}[TRUNCATED {len(s)} chars to 10,000]"
+
+def _truncate_str(s: str) -> str:
+    tail = MASKING_LENGTH // 5
+    head = MASKING_LENGTH - tail
+    return f"{s[:head]}…{s[-tail:]}[TRUNCATED {len(s)} chars to {MASKING_LENGTH}]"
```
evan-onyx, line 16: technically the "[TRUNCATED ..." str adds characters, but if this is just approx anyways then doesnt matter

backend/onyx/evals/tracing.py
```diff
@@ -7,18 +7,24 @@
...
+
+
+def _should_mask(data: Any) -> bool:
```
evan-onyx: using Any here feels bad, would prefer strong typing

PR #5461, opened 2025-09-20
feat: Easy Install

yuhongsun96:
## Description

1 command install of Onyx

## How Has This Been Tested?

Ran locally, some concerns:
- Changed the file name to be default docker compose to simplify things, may confuse some people who have existing deployments
- The deployment is simply called onyx now instead of onyx-stack
- Not tested on Linux/Windows machines
- chmod +x on the file may in tiny percentage of servers be blocked

---

## Summary by cubic
Adds a one-command Docker Compose install with a guided script, a single compose file, and an env template to make setup fast and consistent. Also standardizes config names (DANSWER_BOT_* → ONYX_BOT_*, LOG_DANSWER_* → LOG_ONYX_*) and cleans up nginx/web configs.

- **New Features**
  - Added deployment/docker_compose/install.sh for a guided “1 command” install.
  - Unified to one deployment/docker_compose/docker-compose.yml and a new env.template.
  - Simplified nginx configs; introduced app.conf.template.prod for HTTPS setups.
  - Updated ECS/Compose to reference the right nginx template variants.

- **Migration**
  - Rename all DANSWER_BOT_* envs to ONYX_BOT_* (including Slack tokens) and LOG_DANSWER_MODEL_INTERACTIONS to LOG_ONYX_MODEL_INTERACTIONS.
  - Helm values and VSCode templates updated; align any custom overrides accordingly.
  - Removed NEXT_PUBLIC_DISABLE_STREAMING and NEXT_PUBLIC_NEW_CHAT_DIRECTS_TO_SAME_PERSONA from web build/runtime.
  - For Compose users, use the new install flow: cd deployment/docker_compose && ./install.sh; for prod, use the .prod nginx template.


backend/onyx/configs/onyxbot_configs.py
```diff
@@ -3,16 +3,14 @@
 #####
 # Onyx Slack Bot Configs
 #####
-DANSWER_BOT_NUM_RETRIES = int(os.environ.get("DANSWER_BOT_NUM_RETRIES", "5"))
+ONYX_BOT_NUM_RETRIES = int(os.environ.get("ONYX_BOT_NUM_RETRIES", "5"))
```
yuhongsun96, line 6: These slackbot changes are minimally risky imo. Since they are default off riskier behaviors like responding in every channel, it shouldn't cause any issues if they no longer apply. Also some of these I suspect don't work anyway.

deployment/aws_ecs_fargate/cloudformation/services/onyx_web_server_service_template.yaml
```diff
@@ -126,9 +126,6 @@ Resources:
               HostPort: 3000
               Protocol: tcp
           Environment:
-            - Name: NEXT_PUBLIC_DISABLE_STREAMING
```
yuhongsun96, line 129: Not used I believe, might want to double check

deployment/data/nginx/app.conf.template
```diff
@@ -4,24 +4,6 @@ log_format custom_main '$remote_addr - $remote_user [$time_local] "$request" '
                 '"$http_user_agent" "$http_x_forwarded_for" '
```
yuhongsun96, line 4: Following same convention as docker compose file naming, there's a default version and a prod version, just renamed the files, the diffs here are just from git not understanding the file renamings

deployment/data/nginx/app.conf.template.prod
```diff
@@ -4,6 +4,24 @@ log_format custom_main '$remote_addr - $remote_user [$time_local] "$request" '
                 '"$http_user_agent" "$http_x_forwarded_for" '
                 'rt=$request_time';
 
+# Map X-Forwarded-Proto or fallback to $scheme
```
yuhongsun96, line 7: May want to quickly sanity check that nothing is broken from the renaming. I did not test the prod setup

.github/workflows/pr-integration-tests.yml
```diff
@@ -248,6 +248,7 @@ jobs:
       - name: Start Docker containers
         run: |
           cd deployment/docker_compose
+          cp env.template .env
```
justin-tahara: Is this required to get the test to pass? Or are there other dependencies that require this to be set.

.github/workflows/pr-playwright-tests.yml
```diff
@@ -182,21 +182,22 @@ jobs:
...
           IMAGE_TAG=test \
-          docker compose -f docker-compose.dev.yml -p danswer-stack up -d
+          docker compose up -d
         id: start_docker
 
       - name: Wait for service to be ready
         run: |
           echo "Starting wait-for-service script..."
 
-          docker logs -f danswer-stack-api_server-1 &
```
justin-tahara, line 199: As long as the onyx name is correctly populated I think this is fine.

.vscode/env_template.txt
```diff
@@ -39,8 +39,8 @@ FAST_GEN_AI_MODEL_VERSION=gpt-4o
 
 # For Danswer Slack Bot, overrides the UI values so no need to set this up via UI every time
 # Only needed if using DanswerBot
-#DANSWER_BOT_SLACK_APP_TOKEN=<REPLACE THIS>
-#DANSWER_BOT_SLACK_BOT_TOKEN=<REPLACE THIS>
+#ONYX_BOT_SLACK_APP_TOKEN=<REPLACE THIS>
+#ONYX_BOT_SLACK_BOT_TOKEN=<REPLACE THIS>
```
justin-tahara, line 43: Beautiful

backend/ee/onyx/background/celery/tasks/external_group_syncing/tasks.py
```diff
@@ -93,7 +93,7 @@ def _is_external_group_sync_due(cc_pair: ConnectorCredentialPair) -> bool:
 
     if cc_pair.access_type != AccessType.SYNC:
         task_logger.error(
-            f"Recieved non-sync CC Pair {cc_pair.id} for external "
+            f"Received non-sync CC Pair {cc_pair.id} for external "
```
justin-tahara, line 96: Nice catch

backend/onyx/configs/onyxbot_configs.py
justin-tahara: This also implies that the new env variable is set properly and can be retrieved. I think it's a matter of making sure the new one is there not so much that the old was stays or not.

Do feel like this is quite risky.

backend/onyx/onyxbot/slack/handlers/handle_message.py
```diff
@@ -3,8 +3,8 @@
 from slack_sdk import WebClient
 from slack_sdk.errors import SlackApiError
 
-from onyx.configs.onyxbot_configs import DANSWER_BOT_FEEDBACK_REMINDER
 from onyx.configs.onyxbot_configs import DANSWER_REACT_EMOJI
```
justin-tahara: Should we not change this one as well?

backend/onyx/onyxbot/slack/handlers/handle_regular_answer.py
```diff
@@ -14,11 +14,11 @@
 from onyx.chat.process_message import stream_chat_message_objects
 from onyx.configs.app_configs import DISABLE_GENERATIVE_AI
 from onyx.configs.constants import DEFAULT_PERSONA_ID
-from onyx.configs.onyxbot_configs import DANSWER_BOT_DISABLE_DOCS_ONLY_ANSWER
-from onyx.configs.onyxbot_configs import DANSWER_BOT_DISPLAY_ERROR_MSGS
-from onyx.configs.onyxbot_configs import DANSWER_BOT_NUM_RETRIES
 from onyx.configs.onyxbot_configs import DANSWER_REACT_EMOJI
```
justin-tahara: Same as above here.

backend/onyx/onyxbot/slack/utils.py
```diff
@@ -89,22 +89,22 @@ def check_message_limit() -> bool:
     High traffic at the end of one period and start of another could cause
     the limit to be exceeded.
     """
-    if DANSWER_BOT_RESPONSE_LIMIT_PER_TIME_PERIOD == 0:
+    if ONYX_BOT_RESPONSE_LIMIT_PER_TIME_PERIOD == 0:
```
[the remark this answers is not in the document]
justin-tahara: I think this is a safe change that we should probably just do to be safe

deployment/data/nginx/app.conf.template
justin-tahara, line 1: Is the understanding that everything should be routed through the .prod template instead of this one? Should we even keep this one?

deployment/docker_compose/install.sh
justin-tahara, line 1: What was the reasoning behind the Bash script? Was this more along the lines of with Bash you don't need to download anything but with Python you would need a working version of python? 

I think as someone who is using docker, it probably can be assumed that they have python and so writing a python script would be nice to maintain and easier to debug in my opinion. 

Open to hearing what your thoughts are

justin-tahara (commented):
Overall I think this PR is a bit too dangerous to just merge. 

My opinion is to break this up into smaller PR's that are easier to digest and easier to read through. 

Ideally the Slack bot changes are all one PR, then the compose unification is another. Helm can be it's own PR and lastly the script comes in once all of the individual workflows have been tested. 

From personal experience, I do prefer Python Scripts for maintainability and ease of understanding but if this is truly a one line command approach to onyx then I can be convinced to do bash instead.

Maybe we can even have both options be available but again two things to maintain is not as easy as one. 

The last thing for this is that many workflows will most likely break across the board. Many customers that have hardcoded workflows will definitely complain about this upgrade and thus we should think a bit more about the migration path and how we would introduce this to existing users of Onyx in a non-invasive way such that they are able to also upgrade easily to this new workflow. 

Also if we are going to introduce the bash script or any script in general, we should have a integration or test suite to ensure that it never breaks and is the source of truth to upgrading or installing onyx.

.github/workflows/pr-integration-tests.yml
yuhongsun96: Let me try to make it not required

yuhongsun96: done!

.github/workflows/pr-playwright-tests.yml
yuhongsun96: Ya it's the default now in the compose script, a lot of things are simplified now

backend/ee/onyx/background/celery/tasks/external_group_syncing/tasks.py
yuhongsun96: lmao I wanted to fix a bot comment but I realized this same typo appeared a bunch of times so just fixed it across the board

backend/onyx/configs/onyxbot_configs.py
yuhongsun96: The ones that don't work (I removed some), don't feel super critical. It's outside the scope of this PR. My only concern is with variable renaming when some people might still be deploying with old vars and this becoming a "regression" even though it's not.

But reading through what they do, I'm not too concerned about that case.

backend/onyx/onyxbot/slack/handlers/handle_message.py
yuhongsun96: Ah it's not even a thing I think people should change to be honest... ya we should change it

yuhongsun96: done!

backend/onyx/onyxbot/slack/utils.py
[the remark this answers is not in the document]
yuhongsun96: LMAO that's so funny that the bot is patting itself on the back for this

done

deployment/data/nginx/app.conf.template
yuhongsun96: This is the default non SSL one. I think it's good to keep, the default deployment uses this. HTTPS setup is kind of a pain in the ass, I still don't fully understand it and we don't want to make the base deployment too painful.

deployment/docker_compose/install.sh
yuhongsun96: pip is the python package manager, you can reuse it to manage other packages but it's less intuitive, it's like installing Onyx with npm, probably there's some way to get that working but it's not the right tool.

The other thing is with requiring a python installation which doesn't come out of the box with most operating systems whereas bash typically does. Maybe on Windows most people have pip instead of bash but meh...

PR #5475, opened 2025-09-23
feat: pass slack thread context to llm as message_history

brijsiyag-meesho:
## Description

Previously, Slack integration was using single_message_history but wasn't passing this to LLM for previous context. 
This solution just passes all thread messages as message_history so LLM can get context of the thread. 
Also we are not storing anything at onyx side so it shouldn't be issue in terms of slack

## How Has This Been Tested?

Manually

---

## Summary by cubic
Added per-message history with token counting for Slack threads to improve context management and prevent token overflow in long conversations.

- **New Features**
  - Introduced message_history parameter in stream_chat_message_objects.
  - Slack handler builds a list of PreviousMessage with token_count using the persona’s tokenizer.
  - Passes message_history and single_message_history to chat processing, with a safe fallback when message_history is None.

### DIFF
--- a/backend/onyx/chat/process_message.py
+++ b/backend/onyx/chat/process_message.py
@@ -3,7 +3,7 @@ import time
 import traceback
 from collections.abc import Callable
 from collections.abc import Iterator
-from typing import cast
+from typing import cast, List
 from typing import Protocol
 
 from sqlalchemy.orm import Session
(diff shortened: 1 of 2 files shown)


brijsiyag-meesho:
@Weves currently this PR contains some patchy solutions, if team thinks we can proceed with this than I will polish the code further.

jessicasingh7 (commented):
Overall looks good @brijsiyag-meesho ! Would recommend looking at cubic comment and then getting mypy and other tests to pass and then would be good to merge!

brijsiyag-meesho:
Okay, will update it and let you know

PR #5488, opened 2025-09-25
fix(infra): lazy import litellm and some more pkgs and add layer to connector instantiation for lazy loading

edwin-onyx:
## Description

moves litellm to lazy import to prevent import on some worker startup codepaths

create lazy loading for connectors so workers don't have to load into memory if they don't use that connector

in terms of impact, reduces some workers even more:

Worker | Before | After | Change
primary | 674.7 MB | 570.3 MB | -104.4 MB
light | 584.7 MB | 285.6 MB | -299.1 MB
heavy | 424.4 MB | 209.7 MB | -214.7 MB
kg_processing | 376.0 MB | 242.9 MB | -133.1 MB

api server also down like 100 MB

total -851.3 MB

https://linear.app/danswer/issue/DAN-2636/ensure-workers-are-slim-and-not-in-dependency-hell

## How Has This Been Tested?

## Additional Options

---

## Summary by cubic
Refactored connectors and heavy dependencies to lazy-load at call time, making workers slimmer and faster to start. Addresses Linear DAN-2636 by reducing import overhead and memory usage in Onyx.

- **Refactors**
  - Connector factory now lazily imports connector classes via a module/class map with caching; updated identify_connector_class and improved error handling.
  - Added lazy-loading for federated connectors with similar mapping and cache.
  - Moved SearchDoc conversion into SearchDoc.from_chunks_or_sections; removed the old utility and updated all call sites.
  - Made heavy deps lazy: openai, litellm, markitdown, tiktoken, unstructured; scoped imports inside functions and used TYPE_CHECKING where needed.
  - Moved PromptSnapshot to chat/prompt_builder/schemas.py and fixed imports.
  - Expanded check_lazy_imports script to enforce lazy imports for the new modules; added unit tests validating connector and federated factory mappings, lazy load, and caching.


backend/onyx/connectors/factory.py
```diff
@@ -72,77 +29,50 @@ class ConnectorMissingException(Exception):
...
-        DocumentSource.AIRTABLE: AirtableConnector,
-        DocumentSource.HIGHSPOT: HighspotConnector,
-        DocumentSource.IMAP: ImapConnector,
-        DocumentSource.BITBUCKET: BitbucketConnector,
-        # just for integration tests
-        DocumentSource.MOCK_CONNECTOR: MockConnector,
-    }
-    connector_by_source = connector_map.get(source, {})
-
-    if isinstance(connector_by_source, dict):
```
edwin-onyx, line 135: IIUC this branch is redundant? the only input source where it would be a dict is slack and both input types return slack connectors, lmk if im missing something but i just removed it

backend/onyx/file_processing/unstructured.py
```diff
@@ -2,7 +2,6 @@
 from typing import cast
 from typing import IO
 
-from unstructured.staging.base import dict_to_elements
 from unstructured_client import UnstructuredClient  # type: ignore
 from unstructured_client.models import operations  # type: ignore
 from unstructured_client.models import shared
```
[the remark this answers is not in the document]
edwin-onyx: fine if unstructured_client is imported, just not unstructured

backend/onyx/llm/chat_llm.py
```diff
@@ -85,8 +80,11 @@ def _base_msg_to_role(msg: BaseMessage) -> str:
 
 
 def _convert_litellm_message_to_langchain_message(
-    litellm_message: litellm.Message,
+    litellm_message: "litellm.Message",
 ) -> BaseMessage:
+    configure_litellm()
```
Weves: hmm, this doesn't seem ideal (people have to remember to do this every time they use litellm). 

Can we create our own version of litellm that has some automatic caching built in? Or ideally, a standardized wrapper that we can wrap any lib in to make it lazy + add some optional initialization behavior (kind of like before)?

backend/onyx/connectors/factory.py
```diff
@@ -72,101 +29,75 @@ class ConnectorMissingException(Exception):
...
-                and not issubclass(connector, EventConnector)
-            ),
-        ]
-    ):
+    mapping = CONNECTOR_CLASS_MAP[source]
+
+    try:
+        module = importlib.import_module(mapping.module_path)
+        connector_class = getattr(module, mapping.class_name)
+        _connector_cache[source] = connector_class
```
Weves, line 49: why do we not need `global` here? Feels a bit weird. I believe it works due to the tests, but still

edwin-onyx: dont need global if mutating dict, only need it if reassigning

backend/onyx/llm/chat_llm.py
edwin-onyx: ye fair, so the issue with the previous impl i tested was that it doesn't import submodules so having to do that for every submodule i think isnt very extensible

i did it for at least the base module for now

PR #5511, opened 2025-09-26
fix(source selection): enable all by default and persist choice

jessicasingh7:
## Description

## How Has This Been Tested?

## Additional Options


web/package-lock.json
Weves, line 1: why is this file changing? 🧐

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -446,10 +448,49 @@ export function ActionToggle({
...
   const [showFadeMask, setShowFadeMask] = useState(false);
   const [showTopShadow, setShowTopShadow] = useState(false);
-  // Use existing filter system
-  const filterManager = useFilters();
   const { selectedSources, setSelectedSources } = filterManager;
   const [mcpServers, setMcpServers] = useState<MCPServer[]>([]);
+  const [sourcesInitialized, setSourcesInitialized] = useState(false);
+
+  // Load saved source preferences from localStorage
+  const loadSavedSourcePreferences = () => {
```
Weves: can we move all of this into a hook (this plus the below func + useEffect + toggleSource + selectedSources)?

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -446,10 +448,49 @@ export function ActionToggle({
...
+    if (typeof window === "undefined") return null;
+    const saved = localStorage.getItem("chat-selected-sources");
```
Weves: nit: use contstant here

Weves (changes requested):

web/package-lock.json
jessicasingh7: Minor changes will keep as per our discussion- adding dev:true flag and duplicate @radix-ui/react-use-effect-event

Weves (approved):

PR #5557, opened 2025-09-30
fix(SAML): Add additional Email Fields

justin-tahara:
## Description

Customer setting up SAML for Entra ID is hitting a new issue. In our `saml.py` file we hardcode to look for an email field. If this field is not found we error out with a generic issue in the FE and they cannot log in properly. 

This PR aims to add additional email address fields that we should check and extract variables out of in order to properly get through.

## How Has This Been Tested?

I stood up my own instance of a SAML application in Entra ID and then stood up a local setup to then test. I was able to reproduce the issue the customer was hitting and then once I made the change, I could get through without any updates to my Entra ID configuration. 

The customer can manually unblock as well by creating a new custom claim that is just titled `email` and that should also allow the customer to get through to their application. 

## Additional Options

### DIFF
--- a/backend/ee/onyx/server/saml.py
+++ b/backend/ee/onyx/server/saml.py
@@ -37,6 +37,19 @@ from onyx.utils.logger import setup_logger
 logger = setup_logger()
 router = APIRouter(prefix="/auth/saml")
 
+# Azure AD / Entra ID often returns the email attribute under different keys.
+# Keep a list of common variations so we can fall back gracefully if the IdP
+# does not send the plain "email" attribute name.
+EMAIL_ATTRIBUTE_KEYS = {
+    "email",
+    "emailaddress",
+    "mail",
+    "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress",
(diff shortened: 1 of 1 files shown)


backend/ee/onyx/server/saml.py
```diff
@@ -204,16 +217,37 @@ async def _process_saml_callback(
...
-        )
+    user_email: str | None = None
+
+    # The OneLogin toolkit normalizes attribute keys, but still performs a
+    # case-sensitive lookup. Try the common keys first and then fall back to a
+    # case-insensitive scan of all returned attributes.
+    for attribute_key in EMAIL_ATTRIBUTE_KEYS:
+        attribute_values = auth.get_attribute(attribute_key)
+        if attribute_values:
+            user_email = attribute_values[0]
```
Weves, line 228: this will get the first letter of the email?

justin-tahara: Nope this is getting a list of attributes using the OneLogin toolkit so it will get the user's email not just the first letter

Weves (approved):

PR #5558, opened 2025-09-30
.

edwin-onyx:
## Description

## How Has This Been Tested?

## Additional Options


.github/workflows/pr-helm-chart-testing.yml
```diff
@@ -225,4 +372,4 @@ jobs:
         helm list --all-namespaces
       # the following would install only changed charts, but we only have one chart so 
       # don't worry about that for now
-      # run: ct install --target-branch ${{ github.event.repository.default_branch }}
+      # run: ct install --target-branch ${{ github.event.repository.default_branch }}
```
Weves: nit: add newline

Weves (commented):
porque no all workers in one?

Also for kubernetes deploys, I think we can only support one deployment mode (all workers as separate pods).

The way I'm thinking about it is:
1. Lightweight mode. Lightest we can go, aiming for <1gb. Primarily for folks using the pure chat use case / testing locally / little connected data / don't use any of the background-heavy features (user groups, permission syncing, etc.). Default for local/small scale deploys.

2. Standard mode. Exactly as is today.

edwin-onyx:
> porque no all workers in one?
> 
> Also for kubernetes deploys, I think we can only support one deployment mode (all workers as separate pods).
> 
> The way I'm thinking about it is:
> 
> 1. Lightweight mode. Lightest we can go, aiming for <1gb. Primarily for folks using the pure chat use case / testing locally / little connected data / don't use any of the background-heavy features (user groups, permission syncing, etc.). Default for local/small scale deploys.
> 2. Standard mode. Exactly as is today.

as mentioned in pr description wanted to rollout consolidation over phases for safety regarding load considerations given that this would be the default mode

> in terms of consolidation i think there may be possibly performance concerns but dont think there is an easy way to load test so i think just starting w consolidating a couple is safe enough?

edwin-onyx:
updated to remove the kubernetes changes

backend/scripts/supervisord_entrypoint.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+# Entrypoint script for supervisord that sets environment variables
+# for controlling which celery workers to start
+
+# Default to lightweight mode if not set
+if [ -z "$USE_LIGHTWEIGHT_BACKGROUND_WORKER" ]; then
+    export USE_LIGHTWEIGHT_BACKGROUND_WORKER="true"
+fi
+
+# Set the complementary variable for supervisord
```
Weves, line 10: nit: can we add another comment as to why this is necessary? I'm guessing it's because we can't do `autostart=%(not ENV_USE_LIGHTWEIGHT_BACKGROUND_WORKER)s`

backend/onyx/configs/app_configs.py
```diff
@@ -351,29 +351,35 @@
...
-    CELERY_WORKER_USER_FILE_PROCESSING_CONCURRENCY = (
-        CELERY_WORKER_USER_FILE_PROCESSING_CONCURRENCY_DEFAULT
-    )
+
+# Consolidated background worker (merges heavy, kg_processing, monitoring, user_file_processing)
+CELERY_WORKER_BACKGROUND_CONCURRENCY = int(
+    os.environ.get("CELERY_WORKER_BACKGROUND_CONCURRENCY") or 6
+)
+
+# Individual worker concurrency settings (used when USE_LIGHTWEIGHT_BACKGROUND_WORKER is False)
```
Weves: nit: `or on Kubernetes deployments`

backend/ee/onyx/background/celery/apps/background.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+@shared_task(
+    name=OnyxCeleryTask.EXPORT_QUERY_HISTORY_TASK,
+    ignore_result=True,
+    soft_time_limit=JOB_TIMEOUT,
+    bind=True,
+    trail=False,
+)
+def export_query_history_task(
```
Weves: this should probably be moved to `ee/onyx/background/celery/tasks/query_history` or something like that?

.vscode/launch.template.jsonc
```diff
@@ -311,58 +307,6 @@
       },
       "consoleTitle": "Celery beat Console"
     },
-    {
```
Weves, line 314: can we leave this in for now in case someone wants to test the separated worker set up? Maybe in a different section

PR #5563, opened 2025-10-01
Feat/add download users

trial-danswer:
## Description

Adds functionality to download all users in a csv format file with three columns - Email, Role, and Status.

## How Has This Been Tested?

Functional testing was done by going into admin/users page and clicking on Download All Users button. Doing so, downloaded a users.csv file with three columns. 

## Additional Options

### DIFF
--- a/backend/onyx/server/manage/users.py
+++ b/backend/onyx/server/manage/users.py
@@ -1,3 +1,5 @@
+import csv
+import io
 import re
 from datetime import datetime
 from datetime import timedelta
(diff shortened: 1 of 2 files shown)


trial-danswer:
I think the 2 failing checks are unrelated since last PR's also have them failing.

Weves:
Screenshot 2025-10-01 at 3 50 29 PM

Could we actually replicate something like the ^? Specifically, just the Invite / Download / Search bar layout part?

Weves:
Also, could we add a spinner that covers the screen when the download is happening? We do that in a few other places in the app.

Weves (approved):
lgtm!

PR #5583, opened 2025-10-02
Handle Confluence attachments as individual documents

emerzon:
## Description

This PR changes the current behavior that considers page attachments as part of the page itself, to individual documents.

Mainly due to 3 reasons:

1) Documents only show up once in search results - Pages with lots of attachments might make the search lose context;
2) Lack of direct reference to the document can make very hard to compare source when page has lots of attachments;
3) Small updates of the page can trigger large updates forcing all of the related attachments to be reindexed, even if they didn't change.

## How Has This Been Tested?
Index a Confluence page with attachments, check each attachment is treated as individual doc.

## Additional Options

### DIFF
--- a/backend/onyx/connectors/confluence/connector.py
+++ b/backend/onyx/connectors/confluence/connector.py
@@ -245,9 +245,26 @@ class ConfluenceConnector(
         page_query += " order by lastmodified asc"
         return page_query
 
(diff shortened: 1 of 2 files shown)


emerzon:
Closing as addressed in #5641

PR #5586, opened 2025-10-02
Feature/confluence attachments size limits

emerzon:
## Description
This PR adds configurable parameters under advanced settings for Confluence allowing customization of the file size/char count limit thresholds. Today this is managed via the global env vars ` CONFLUENCE_MAX_ATTACHMENT_SIZE` and `CONFLUENCE_CONNECTOR_ATTACHMENT_CHAR_COUNT_THRESHOLD` , however a global value is not adequate for all use cases that require different threshold levels.

### DIFF
--- a/backend/onyx/connectors/confluence/connector.py
+++ b/backend/onyx/connectors/confluence/connector.py
@@ -245,9 +245,26 @@ class ConfluenceConnector(
         page_query += " order by lastmodified asc"
         return page_query
 
-    def _construct_attachment_query(self, confluence_page_id: str) -> str:
+    def _construct_attachment_query(
+        self,
+        confluence_page_id: str,
+        start: SecondsSinceUnixEpoch | None = None,
+        end: SecondsSinceUnixEpoch | None = None,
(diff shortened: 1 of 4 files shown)


backend/onyx/connectors/confluence/connector.py
```diff
@@ -247,9 +251,26 @@ def _construct_page_cql_query(
         page_query += " order by lastmodified asc"
         return page_query
 
-    def _construct_attachment_query(self, confluence_page_id: str) -> str:
+    def _construct_attachment_query(
+        self,
+        confluence_page_id: str,
+        start: SecondsSinceUnixEpoch | None = None,
```
wenxi-onyx: The timestamp params here can be made non-optional (i.e., remove | None) since both implementations that call this method will always have timestamps available.

Additionally/related, see comment about implementing timestamps for the `retrieve_slim_documents`

emerzon:
This one needs to be back ported to current code - some changes (ie handle attachments as separate documents) were implemented separately in different commits

PR #5587, opened 2025-10-02
Feature/confluence cql pagination sanitize

emerzon:
## Description
Confluence sometimes echoes query params passed via `params` back into  the `_links.next` URL, which can cause duplicate `expand` and  `body-format` entries to accumulate across pages. To avoid URL growth and eventual 400s, strip these from the path before issuing the request and re-apply them via the `params` argument.

    

---

## Summary by cubic
Sanitizes Confluence CQL pagination URLs to stop duplicated expand/body-format params that bloat URLs and cause 400s. Also handles missing space-permissions JSON-RPC results more gracefully.

- **Bug Fixes**
  - Strip duplicate expand/body-format from _links.next, rebuild a clean path, and merge existing expansions with body.atlas_doc_format.
  - Fallback to original URL if parsing fails.
  - Downgrade log to info and return [] when space permissions JSON-RPC has no result (likely insufficient permissions).

### DIFF
--- a/backend/onyx/connectors/confluence/onyx_confluence.py
+++ b/backend/onyx/connectors/confluence/onyx_confluence.py
@@ -919,10 +919,10 @@ class OnyxConfluence:
         response = self.post(url, data=data)
         logger.debug(f"jsonrpc response: {response}")
         if not response.get("result"):
-            logger.warning(
-                f"No jsonrpc response for space permissions for space {space_key}"
-                f"\nResponse: {response}"
+            logger.info(
+                f"No jsonrpc result for space permissions for space {space_key}; likely insufficient permissions."
             )
+            return []
 
         return response.get("result", [])
 


backend/onyx/connectors/confluence/onyx_confluence.py
```diff
@@ -523,13 +524,49 @@ def _paginate_url(
 
         while url_suffix:
             logger.debug(f"Making confluence call to {url_suffix}")
+            # Confluence sometimes echoes query params passed via `params` back into
+            # the `_links.next` URL, which can cause duplicate `expand` and
+            # `body-format` entries to accumulate across pages. To avoid URL
+            # growth and eventual 400s, strip these from the path before
+            # issuing the request and re-apply them via the `params` argument.
+            try:
```
wenxi-onyx, line 532: If i'm understanding this code correctly, the only duplicated items (passed via `params`) are 
```
params={
   "body-format": "atlas_doc_format",
   "expand": "body.atlas_doc_format",
}
```
If this is the case, can we just always pop `body-format` and `expand` instead of iterating through and building a unique list? I.e
```
q.pop("expand", None)
q.pop("body-format", None)

sanitized_url_suffix = parsed.path
if q:
    sanitized_url_suffix += "?" + urlencode({k: v[0] for k, v in q.items()})

try:
    raw_response = self.get(
        path=sanitized_url_suffix,
        advanced_mode=True,
        params={
            "body-format": "atlas_doc_format",
            "expand": "body.atlas_doc_format",
        },
    )
```

wenxi-onyx: If I'm wrong about this, a few follow-up nits
- Would prefer variable names instead of single letter variables (I know it's obvious here, but might as well try to stay consistent with this style)
- I think the iteration logic can be simplified if you just use a single set like `existing_expansions: set[str]`
```
if part := part.strip(): 
    existing_expansions.add(part) # Duplicate values auto handled
```
- Prefer using `urlencode({k: v[0] for k, v in q.items()})` instead of f string + quote

backend/onyx/connectors/confluence/onyx_confluence.py
```diff
@@ -523,13 +524,49 @@ def _paginate_url(
...
+                    for part in val.split(","):
+                        part = part.strip()
+                        if part and part not in seen_expansions:
+                            existing_expansions.append(part)
+                            seen_expansions.add(part)
+                q.pop("expand", None)
+                q.pop("body-format", None)
+                # rebuild sanitized path
+                base = url_suffix.split("?")[0]
+                sanitized_query = "&".join(
```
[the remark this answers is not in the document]
wenxi-onyx: Seems reasonable? Thoughts? (only relevant if my `urlencode` comment above is invalid)

wenxi-onyx (changes requested):
Hey @emerzon - apologies for the wait on this. I've left a comment about the param parsing logic. Additionally, could you please rebase on main? Will run tests once ready!

PR #5601, opened 2025-10-04
Celery log level

emerzon:
## Description

## How Has This Been Tested?

## Additional Options


emerzon:
> 1. Can you please confirm that `LOG_LEVEL=debug` doesn't cause crazy spam from celery? Most of the development team has this permanently set and making this change may hurt DX. If this is the case, maybe we should have an additional env var like `OVERRIDE_DEFAULT_CELERY_LOG_LEVEL`? Open to suggestions
> 2. If we do make this change, should we remove the `backend/supervisord.conf` hardcoded default?

1) Yes, I can confirm that the debug from celery is quite verbose. So indeed would make sense to have a separate config for it. 

2) We should, to avoid confusion.

emerzon:
Closing in favor of individual flag for Celery

PR #5620, opened 2025-10-06
Fix/doc id migration task

Subash-Mohan:
## Description
This pull request introduces several updates related to the user file document ID migration process, including a new database migration, Celery task scheduling adjustments, and supporting code changes to facilitate and control the migration. The primary focus is on ensuring that the document ID migration for user files is handled robustly and can be managed via scheduled background tasks with appropriate locking and configuration.

**Database Migration and Model Changes:**

* Added a new Alembic migration (`40926a4dab77`) to reset the `document_id_migrated` field for all records in the `user_file` table, with upgrade and downgrade logic to toggle the field as needed.

**Celery Task Scheduling and Locking:**

* Modified the Celery beat schedule:
  - Added a new periodic task `check-for-user-file-project-sync` scheduled every 20 seconds.
  - Changed the `user-file-docid-migration` task to run every 10 minutes (instead of every 1 minute) and moved it to the `USER_FILES_INDEXING` queue.
* Registered the `user_file_processing` tasks module in both the docfetching and primary Celery apps to ensure the doc ID migration task is available during the migration period. 
* Introduced new lock timeout constants for user file processing, project sync, and doc ID migration, and added a Redis lock key for the doc ID migration process. 

**Document Index and Migration Support:**

* Updated the `VespaDocumentFields` dataclass and the Vespa index update logic to support updating the `document_id` field during migration, with clear comments indicating this is a temporary, migration-only change. 

## How Has This Been Tested?
Tested from UI

## Additional Options


backend/onyx/background/celery/tasks/user_file_processing/tasks.py
```diff
@@ -359,6 +363,15 @@ def process_single_user_file_project_sync(
 
     try:
         with get_session_with_current_tenant() as db_session:
+
+            # 20 is the documented default for httpx max_keepalive_connections
```
Weves, line 367: this should be done in the `on_worker_init` function, right?

backend/onyx/background/celery/tasks/user_file_processing/tasks.py
```diff
@@ -436,127 +453,148 @@ def _visit_chunks(
...
-    with get_vespa_http_client() as http_client:
-        continuation: str | None = None
-        while True:
-            docs, continuation = _visit_chunks(
-                http_client=http_client,
-                index_name=index_name,
-                selection=selection,
-                continuation=continuation,
+def update_legacy_plaintext_file_records() -> None:
+
```
Weves, line 464: nit: can we add a comment describing what this function is trying to do? It's not clear to me from the name

backend/onyx/background/celery/tasks/user_file_processing/tasks.py
```diff
@@ -436,127 +453,148 @@ def _visit_chunks(
...
+    updated_count = 0
     try:
+        update_legacy_plaintext_file_records()
+        # Track lock renewal
+        last_lock_time = time.monotonic()
         with get_session_with_current_tenant() as db_session:
+
+            # 20 is the documented default for httpx max_keepalive_connections
+            if MANAGED_VESPA:
```
Weves, line 571: same comment here

backend/onyx/background/celery/tasks/user_file_processing/tasks.py
```diff
@@ -580,120 +618,138 @@ def user_file_docid_migration_task(self: Task, *, tenant_id: str) -> bool:
...
-                        db_session.execute(
-                            sa.select(FileRecord).where(
-                                FileRecord.file_origin == FileOrigin.PLAINTEXT_CACHE,
-                                FileRecord.file_id.like("plaintext_%"),
-                            )
+                    # Count all chunks for this document
+                    chunk_count = 0
+                    continuation = None
+                    while True:
+                        docs, continuation = _visit_chunks(
```
Weves, line 685: can we get the chunk count from postgres? In the `document` table there is a `chunk_count` column

backend/onyx/document_index/interfaces.py
```diff
@@ -109,6 +109,9 @@ class VespaDocumentFields:
     hidden: bool | None = None
     aggregated_chunk_boost_factor: float | None = None
 
+    # document_id is added for migration purposes, ideally we should not be updating this field
+    document_id: str | None = None
```
Weves, line 114: can we add a TODO to remove this once the migration is done?

backend/alembic/versions/40926a4dab77_reset_userfile_document_id_migrated_.py
```diff
@@ -0,0 +1,31 @@
...
+    # Set all existing records to not migrated
+    op.execute(
+        "UPDATE user_file SET document_id_migrated = FALSE "
+        "WHERE document_id_migrated IS DISTINCT FROM FALSE;"
+    )
+
+
+def downgrade() -> None:
+    # Revert all records back to migrated
+    op.execute(
```
Weves: hmm, maybe this should be a no-op? We don't want to mark things that haven't been migrated as migrated on the downgrade

backend/onyx/background/celery/tasks/user_file_processing/tasks.py
Subash-Mohan: I just followed the same pattern as docprocessing worker. Do you want me to move it to worker_init ?

backend/onyx/background/celery/tasks/user_file_processing/tasks.py
Subash-Mohan: We don't have the document records. We already deleted it in the migrations.

PR #5632, opened 2025-10-07
Fix/confluence pruning perms

evan-onyx:
## Description

Replacement for #5583 

## How Has This Been Tested?

tested in UI + CI tests

## Additional Options

---

## Summary by cubic
Fixes permission pruning by separating slim doc retrieval into “IDs only” and “IDs + permissions,” and updating the pruning job to use the right path. Confluence now indexes attachments as separate documents and applies time filters during permission sync.

- **Bug Fixes**
  - Pruning job now handles SlimConnector vs SlimConnectorWithPermSync correctly and rate-limits ID retrieval.
  - Confluence slim perm sync applies start/end time to attachment queries and handles bad attachment URLs gracefully.
  - Clearer warnings when Confluence space permissions are missing.

- **Refactors**
  - New SlimConnectorWithPermSync interface with retrieve_all_slim_docs_perm_sync; added retrieve_all_slim_docs for ID-only cases. Updated all connectors and tests.
  - Confluence attachments split into standalone docs with metadata/owners; page text includes inline attachment markers instead of merging attachment content.

### DIFF
--- a/backend/ee/onyx/external_permissions/confluence/space_access.py
+++ b/backend/ee/onyx/external_permissions/confluence/space_access.py
@@ -124,9 +124,9 @@ def get_space_permission(
         and not space_permissions.external_user_group_ids
     ):
         logger.warning(
-            f"No permissions found for space '{space_key}'. This is very unlikely"
-            "to be correct and is more likely caused by an access token with"
-            "insufficient permissions. Make sure that the access token has Admin"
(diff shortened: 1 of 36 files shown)


evan-onyx:
oops, re-used an old branch on accident. See https://github.com/onyx-dot-app/onyx/pull/5641

PR #5638, opened 2025-10-07
fix(Blob Storage): Add Chunking + Size Limits

justin-tahara:
## Description

Our Google Cloud Storage connector currently does not check for sizes and it also does not chunk at all. 

This causes our docprocessing pods to OOM. This PR aims to add a threshold of 20MB as the ceiling and then also adds chunking functionality to ensure that we can prevent files that don't necessarily have a size attribute to be properly evicted if they get too large.

## How Has This Been Tested?

Tested locally

## Additional Options


backend/onyx/connectors/blob/connector.py
```diff
@@ -195,11 +204,48 @@ def _refresh_credentials() -> dict[str, str]:
...
+            if not chunk:
+                continue
+            chunks.append(chunk)
+            bytes_read += len(chunk)
+
+            if bytes_read > self.size_threshold + SIZE_THRESHOLD_BUFFER:
+                logger.warning(
+                    f"{key} exceeds size threshold of {self.size_threshold}. Skipping."
+                )
+                return None
```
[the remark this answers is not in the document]
justin-tahara: This is to add a bit of a buffer so that the file can be exactly 20MB plus a lil bit to handle for overflow @greptileai

backend/onyx/connectors/blob/connector.py
```diff
@@ -266,6 +312,18 @@ def _yield_blob_objects(
                 key = obj["Key"]
                 link = self._get_blob_link(key)
 
+                size_bytes = obj.get("Size")
```
evan-onyx: Is this standardized? if not we could check for size, fileSize, file_size, etc

backend/onyx/connectors/blob/connector.py
```diff
@@ -195,11 +204,48 @@ def _refresh_credentials() -> dict[str, str]:
 
         return None
 
-    def _download_object(self, key: str) -> bytes:
+    def _should_enforce_size_limit(self) -> bool:
+        return self.bucket_type == BlobType.GOOGLE_CLOUD_STORAGE and (
```
evan-onyx: should make this generic imo

evan-onyx (approved):

backend/onyx/connectors/blob/connector.py
justin-tahara: Adding in some edge case handling for different types of file metadatas

PR #5645, opened 2025-10-07
fix: align text with icon

Weves:
## Description

Fixes https://linear.app/danswer/issue/DAN-2800/align-text-with-assistant-icon

## How Has This Been Tested?

local

## Additional Options


raunakab (approved):
Fine with merging this in! Icon placement will be updated in the future, but this will work for now!

PR #5649, opened 2025-10-07
refactor: improve modal behavior

Weves:
## Description

Fixes https://linear.app/danswer/issue/DAN-2804/clean-up-modal-logic

## How Has This Been Tested?

local

## Additional Options


web/src/app/chat/layout.tsx
```diff
@@ -59,12 +60,14 @@ export default async function Layout({
...
-          </div>
-        </ProjectsProvider>
+        <ChatModalProvider>
+          <ProjectsProvider initialProjects={projects}>
+            <div className="flex flex-row w-full h-full">
+              <AppSidebar />
+              {children}
+            </div>
+          </ProjectsProvider>
+        </ChatModalProvider>
```
raunakab, line 70: Hmm, is there a reason we're moving the `ChatModalProvider` away from a more global context into only the `chat` context?

If, in the future, we want to have modals inside of the admin panel (or any non-chat page), we'd need to move this back.

raunakab: I guess I am planning on removing this, since it's a bit of an ugly pattern which I'm no longer fond of. But just curious.

raunakab (approved):
Looks good, but had a small question.

PR #5664, opened 2025-10-08
add chatsession drag and drop functionality

Subash-Mohan:
## Description
This pull request introduces drag-and-drop functionality for chat sessions and projects in the sidebar, refactors related components for modularity and clarity, and improves user feedback for empty states and error handling. 

## How Has This Been Tested?
Tested from UI

## Additional Options

### DIFF
--- a/web/src/sections/sidebar/constants.ts
+++ b/web/src/sections/sidebar/constants.ts
@@ -1,0 +1,11 @@
+export const DRAG_TYPES = {
+  CHAT: "chat",
+  PROJECT: "project",
+  RECENTS: "recents",
+} as const;
+
+export const LOCAL_STORAGE_KEYS = {
+  HIDE_MOVE_CUSTOM_AGENT_MODAL: "onyx:hideMoveCustomAgentModal",
+} as const;
+
+export const DEFAULT_PERSONA_ID = 0;
--- a/web/src/components/sidebar/Projects.tsx
+++ b/web/src/components/sidebar/Projects.tsx
@@ -6,6 +6,7 @@ import {
   useProjectsContext,
 } from "@/app/chat/projects/ProjectsContext";
 import NavigationTab from "@/refresh-components/buttons/NavigationTab";
+import Text from "@/refresh-components/Text";
 import SvgFolder from "@/icons/folder";
 import SvgEdit from "@/icons/edit";
 import { PopoverMenu } from "@/components/ui/popover";
@@ -20,7 +21,11 @@ import {
   useChatModal,
 } from "@/refresh-components/contexts/ChatModalContext";
 import { SEARCH_PARAM_NAMES } from "@/app/chat/services/searchParams";
-import { noProp } from "@/lib/utils";
+import { cn, noProp } from "@/lib/utils";
+import { OpenFolderIcon } from "@/components/icons/CustomIcons";
+
+import { SvgProps } from "@/icons";
+import { useDroppable } from "@dnd-kit/core";
 
 interface ProjectFolderProps {
   project: Project;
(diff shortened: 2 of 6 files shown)


web/src/sections/sidebar/constants.ts
```diff
@@ -0,0 +1,11 @@
...
+  CHAT: "chat",
+  PROJECT: "project",
+  RECENTS: "recents",
+} as const;
+
+export const LOCAL_STORAGE_KEYS = {
+  HIDE_MOVE_CUSTOM_AGENT_MODAL: "onyx:hideMoveCustomAgentModal",
+} as const;
+
+export const DEFAULT_PERSONA_ID = 0;
```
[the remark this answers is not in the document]
Subash-Mohan: ChatSessionMorePopup component will be removed going forward

web/src/sections/sidebar/ProjectFolderButton.tsx
```diff
@@ -94,14 +137,13 @@ function ProjectFolderButtonInner({ project }: ProjectFolderProps) {
...
+            leftIcon={getFolderIcon()}
             active={
               params(SEARCH_PARAM_NAMES.PROJECT_ID) === String(project.id)
             }
-            onClick={() => {
-              setOpen((prev) => !prev);
-              route({ projectId: project.id });
-            }}
+            onIconClick={handleIconClick}
+            onIconHover={setIsHoveringIcon}
```
raunakab: I see the reason for having the `onIconClick` functionality now!

However, I propose a different solution to it, *just to keep the SidebarTab API clean*:"

```tsx
<SidebarTab
leftIcon={({ className }) => {
  /* Put your IconButton here! */
  return (
    <IconButton className={className} icon={getFolderIcon()} onClick={handleIconClick} ... />
  );
}}
onClick={handleTextClick}
...
>
{ ... }
</SidebarTab>
```

The reason for the above is this:
1. It keeps the API to the `SidebarTab` clean and simple.
2. It still allows you to specify a custom `IconButton` instead of a simple, non-clickable icon. You can still specify your `onIconClick` and `onIconHover` handlers.

Ideally, we should try our best to *not* modify any components since they've been carefully curated as per @duo-onyx's mocks.

web/src/sections/sidebar/constants.ts
raunakab, line 1: Awesome! I like this placement of constants in this new, "sidebar"-specific location.

web/src/sections/sidebar/ChatButton.tsx
```diff
@@ -327,51 +365,109 @@ function ChatButtonInner({ chatSession, project }: ChatButtonProps) {
...
+                    </PopoverTrigger>
+                    <PopoverContent side="right" align="end">
+                      {popoverItems}
+                    </PopoverContent>
+                  </>
+                }
+              >
+                {renaming ? (
+                  <ButtonRenaming
+                    initialName={chatSession.name}
```
raunakab: Are the internals of these two conditionals duplicated?

I.e.:

```tsx
draggable ? (
  <div>
    {component}
  </div
) : (
  {component}
)
```

If so, then we should pull out the duplicated component into its own *memoized* variable and then use it in the subsequent logic.

web/src/sections/sidebar/ProjectFolderButton.tsx
Subash-Mohan: Thanks for the approach!. Now i implemented the logic in the IconButton.

web/src/sections/sidebar/ChatButton.tsx
Subash-Mohan: Now I moved it to variable.

Subash-Mohan: Since its just jsx, not even a separate component. I am not sure whether we need memo here.

PR #5681, opened 2025-10-10
feat: tables/migration for permission syncing attempts

nmgarza5:
## Description

Partial completion of: https://linear.app/danswer/issue/DAN-2494/sync-permissions-database-and-backend-updates

Adding two two tables related to auto sync permissions:
1. doc_permission_sync_attempt
2. external_group_permission_sync_attempt

These will be used to track history of sync attempts and display them to the user.

PR Stack:
1. [feat: tables/migration for permission syncing attempts ](https://github.com/onyx-dot-app/onyx/pull/5681)-> you are here
2. feat: basic db methods to create and update permission sync attempts
3. feat: modify existing permission sync tasks workflow to create/update attempt records
4. feat: read latest permission sync from the frontend

## How Has This Been Tested?

Ran a successful alembic upgrade/downgrade 

---

## Summary by cubic
Add database tables and models to track permission sync attempts for documents and external groups, including status, counts, timestamps, and errors. Progress toward Linear DAN-2494 to surface sync history in the UI.

- **New Features**
  - Added tables: doc_permission_sync_attempt and external_group_permission_sync_attempt (group sync can be global; cc pair nullable).
  - Introduced PermissionSyncStatus enum with helpers for terminal/success states.
  - Created ORM models and indexes for time_created, connector_credential_pair_id + time_created, and status + time_finished DESC.
  - Included error_message fields and CASCADE FKs; all timestamps are timezone-aware.

### DIFF
--- a/backend/alembic/versions/03d710ccf29c_add_permission_sync_attempt_tables.py
+++ b/backend/alembic/versions/03d710ccf29c_add_permission_sync_attempt_tables.py
@@ -1,0 +1,153 @@
+"""add permission sync attempt tables
+
+Revision ID: 03d710ccf29c
+Revises: 96a5702df6aa
+Create Date: 2025-09-11 13:30:00.000000
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "03d710ccf29c"  # Generate a new unique ID
+down_revision = "96a5702df6aa"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # Create the permission sync status enum
+    permission_sync_status_enum = sa.Enum(
+        "not_started",
+        "in_progress",
+        "success",
+        "canceled",
+        "failed",
+        "completed_with_errors",
+        name="permissionsyncstatus",
+        native_enum=False,
+    )
+
+    # Create doc_permission_sync_attempt table
+    op.create_table(
+        "doc_permission_sync_attempt",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("connector_credential_pair_id", sa.Integer(), nullable=False),
+        sa.Column("status", permission_sync_status_enum, nullable=False),
+        sa.Column("total_docs_synced", sa.Integer(), nullable=True),
+        sa.Column("docs_with_permission_errors", sa.Integer(), nullable=True),
+        sa.Column("error_message", sa.Text(), nullable=True),
+        sa.Column(
+            "time_created",
+            sa.DateTime(timezone=True),
+            server_default=sa.text("now()"),
+            nullable=False,
+        ),
+        sa.Column("time_started", sa.DateTime(timezone=True), nullable=True),
+        sa.Column("time_finished", sa.DateTime(timezone=True), nullable=True),
+        sa.ForeignKeyConstraint(
+            ["connector_credential_pair_id"],
+            ["connector_credential_pair.id"],
+        ),
+        sa.PrimaryKeyConstraint("id"),
+    )
+
+    # Create indexes for doc_permission_sync_attempt
+    op.create_index(
+        "ix_doc_permission_sync_attempt_time_created",
+        "doc_permission_sync_attempt",
+        ["time_created"],
+        unique=False,
+    )
+    op.create_index(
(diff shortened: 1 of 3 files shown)


backend/onyx/db/enums.py
```diff
@@ -25,6 +25,32 @@ def is_successful(self) -> bool:
...
+            PermissionSyncStatus.COMPLETED_WITH_ERRORS,
+            PermissionSyncStatus.CANCELED,
+            PermissionSyncStatus.FAILED,
+        }
+        return self in terminal_states
+
+    def is_successful(self) -> bool:
+        return (
+            self == PermissionSyncStatus.SUCCESS
+            or self == PermissionSyncStatus.COMPLETED_WITH_ERRORS
```
justin-tahara, line 50: Does this really mean it was successful? Should we make this function like is_completed instead? A bit misleading to do successful even if there are errors. 

Let me know what you think

backend/onyx/db/models.py
```diff
@@ -3600,3 +3601,145 @@ class MCPConnectionConfig(Base):
...
+    """
+
+    __tablename__ = "doc_permission_sync_attempt"
+
+    id: Mapped[int] = mapped_column(primary_key=True)
+
+    connector_credential_pair_id: Mapped[int] = mapped_column(
+        ForeignKey("connector_credential_pair.id", ondelete="CASCADE"),
+        nullable=False,
+    )
```
justin-tahara, line 3624: I'm a bit worried about this `ondelete="CASCADE"`. This potentially could be a very long process if there are customers with a ton of connector credential pairs I believe

backend/onyx/db/models.py
```diff
@@ -3600,3 +3601,145 @@ class MCPConnectionConfig(Base):
...
+            "status",
+            desc("time_finished"),
+        ),
+    )
+
+    def __repr__(self) -> str:
+        return f"<DocPermissionSyncAttempt(id={self.id!r}, " f"status={self.status!r})>"
+
+    def is_finished(self) -> bool:
+        return self.status.is_terminal()
```
justin-tahara, line 3673: Maybe we can keep this consistent with the name we choose in the enums.py file

backend/onyx/db/models.py
```diff
@@ -3600,3 +3601,145 @@ class MCPConnectionConfig(Base):
...
+
+    __tablename__ = "external_group_permission_sync_attempt"
+
+    id: Mapped[int] = mapped_column(primary_key=True)
+
+    # Can be tied to a specific connector or be a global group sync
+    connector_credential_pair_id: Mapped[int | None] = mapped_column(
+        ForeignKey("connector_credential_pair.id", ondelete="CASCADE"),
+        nullable=True,  # Nullable for global group syncs across all connectors
+    )
```
justin-tahara, line 3690: Just another delete that we should be careful about

backend/alembic/versions/03d710ccf29c_add_permission_sync_attempt_tables.py
```diff
@@ -0,0 +1,155 @@
...
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "03d710ccf29c"  # Generate a new unique ID
+down_revision = "96a5702df6aa"
```
justin-tahara, line 15: Double check that this is the latest revision before yours

backend/onyx/db/enums.py
nmgarza5: eh well an attempt is still considered successful even it completes with errors...and I pattern matched with what we do for indexing attempts. going to leave as is

backend/onyx/db/models.py
nmgarza5: good callout. will remove the cascade delete in favor of explicit deletes. will add that in the 2nd pr in the chain

backend/alembic/versions/03d710ccf29c_add_permission_sync_attempt_tables.py
nmgarza5: already done. i rebased this morning before requesting review. will double check again before merging

backend/onyx/db/models.py
nmgarza5: I pattern matched again here 😅 going to leave as is

PR #5682, opened 2025-10-10
feat: basic db methods to create, update, delete permission sync attempts 

nmgarza5:
## Description

Completing subtasks within: https://linear.app/danswer/issue/DAN-2494/sync-permissions-database-and-backend-updates

Some basic operations to read, create and update sync attempts in the db. nothing is actually being called yet. These operations will be hooked up to celery tasks in the next PR

PR Stack:
1. [feat: tables/migration for permission syncing attempts ](https://github.com/onyx-dot-app/onyx/pull/5681)
2. [feat: basic db methods to create and update permission sync attempts](https://github.com/onyx-dot-app/onyx/pull/5682) -> you are here
3. feat: modify existing permission sync tasks workflow to create/update attempt records
4. feat: read latest permission sync from the frontend

## How Has This Been Tested?

unit tests added in the pr

---

## Summary by cubic
Adds database models, migrations, and helper methods to track permission sync attempts for documents and external groups. Supports creating, updating, and querying attempts with progress and status, aligned with Linear DAN-2494.

- **New Features**
  - New tables and ORM models: doc_permission_sync_attempt and external_group_permission_sync_attempt, with useful indexes.
  - PermissionSyncStatus enum with terminal/success helpers.
  - CRUD helpers: create/get attempts, mark in_progress/success/failed/completed_with_errors, and increment progress counters (row-locked updates).
  - Optional telemetry on status changes; eager-load of related connector/credential.
  - Group sync attempts support per-CC pair or global (no CC pair).

- **Migration**
  - Run Alembic upgrade to create new tables and indexes.
  - No data backfill required.

### DIFF
--- a/backend/onyx/db/permission_sync_attempt.py
+++ b/backend/onyx/db/permission_sync_attempt.py
@@ -1,0 +1,485 @@
+"""Permission sync attempt CRUD operations and utilities.
+
+This module contains all CRUD operations for both DocPermissionSyncAttempt
+and ExternalGroupPermissionSyncAttempt models, along with shared utilities.
+"""
+
+from typing import Any
+from typing import cast
+
+from sqlalchemy import delete
+from sqlalchemy import func
+from sqlalchemy import select
+from sqlalchemy.engine.cursor import CursorResult
+from sqlalchemy.orm import joinedload
+from sqlalchemy.orm import Session
+
+from onyx.db.enums import PermissionSyncStatus
+from onyx.db.models import ConnectorCredentialPair
+from onyx.db.models import DocPermissionSyncAttempt
+from onyx.db.models import ExternalGroupPermissionSyncAttempt
+from onyx.utils.logger import setup_logger
+from onyx.utils.telemetry import optional_telemetry
+from onyx.utils.telemetry import RecordType
+
+logger = setup_logger()
+
+
+# =============================================================================
+# DOC PERMISSION SYNC ATTEMPT CRUD
+# =============================================================================
+
+
+def create_doc_permission_sync_attempt(
+    connector_credential_pair_id: int,
+    db_session: Session,
+) -> int:
+    """Create a new doc permission sync attempt.
+
+    Args:
+        connector_credential_pair_id: The ID of the connector credential pair
+        db_session: The database session
+
+    Returns:
+        The ID of the created attempt
+    """
+    attempt = DocPermissionSyncAttempt(
+        connector_credential_pair_id=connector_credential_pair_id,
+        status=PermissionSyncStatus.NOT_STARTED,
+    )
+    db_session.add(attempt)
+    db_session.commit()
+
+    return attempt.id
+
+
+def get_doc_permission_sync_attempt(
+    db_session: Session,
+    attempt_id: int,
+    eager_load_connector: bool = False,
+) -> DocPermissionSyncAttempt | None:
+    """Get a doc permission sync attempt by ID.
+
+    Args:
+        db_session: The database session
+        attempt_id: The ID of the attempt
+        eager_load_connector: If True, eagerly loads the connector and cc_pair relationships
+
+    Returns:
+        The attempt if found, None otherwise
+    """
+    stmt = select(DocPermissionSyncAttempt).where(
+        DocPermissionSyncAttempt.id == attempt_id
+    )
+
+    if eager_load_connector:
+        stmt = stmt.options(
+            joinedload(DocPermissionSyncAttempt.connector_credential_pair).joinedload(
+                ConnectorCredentialPair.connector
+            )
+        )
+
+    return db_session.scalars(stmt).first()
+
+
+def get_recent_doc_permission_sync_attempts_for_cc_pair(
+    cc_pair_id: int,
+    limit: int,
+    db_session: Session,
+) -> list[DocPermissionSyncAttempt]:
+    """Get recent doc permission sync attempts for a cc pair, most recent first."""
+    return list(
+        db_session.execute(
+            select(DocPermissionSyncAttempt)
+            .where(DocPermissionSyncAttempt.connector_credential_pair_id == cc_pair_id)
+            .order_by(DocPermissionSyncAttempt.time_created.desc())
+            .limit(limit)
+        ).scalars()
+    )
+
+
+def mark_doc_permission_sync_attempt_in_progress(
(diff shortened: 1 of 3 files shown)


nmgarza5:
@greptile

backend/onyx/db/permission_sync_attempt.py
```diff
@@ -0,0 +1,518 @@
...
+) -> DocPermissionSyncAttempt | None:
+    """Get a doc permission sync attempt by ID."""
+    stmt = select(DocPermissionSyncAttempt).where(
+        DocPermissionSyncAttempt.id == attempt_id
+    )
+
+    if eager_load_cc_pair:
+        stmt = stmt.options(
+            joinedload(DocPermissionSyncAttempt.connector_credential_pair).joinedload(
+                ConnectorCredentialPair.connector
```
evan-onyx, line 78: is this second one intentional? if what you really need is the connector I'd go with calling it `eager_load_connector`. Also there are some tradeoffs between joinedload and selectinload, just wanted to make sure you've evaluated that since in my experience selectinload is a more default option

backend/onyx/db/permission_sync_attempt.py
```diff
@@ -0,0 +1,518 @@
...
+                "cc_pair_id": attempt.connector_credential_pair_id,
+            },
+        )
+        return attempt
+    except Exception:
+        db_session.rollback()
+        raise
+
+
+def mark_doc_permission_sync_attempt_completed_with_errors(
```
evan-onyx: this seems basically the same as the previous function, slight preference for extracting the implementation into _mark_doc_permission_sync_attempt_done() and passing in the status you want to mark it with

backend/onyx/db/permission_sync_attempt.py
```diff
@@ -0,0 +1,518 @@
...
+                "cc_pair_id": attempt.connector_credential_pair_id,
+            },
+        )
+        return attempt
+    except Exception:
+        db_session.rollback()
+        raise
+
+
+def mark_doc_permission_sync_attempt_failed(
```
evan-onyx, line 130: honestly you could roll this one in too and have the error message be optional in the implementation function

backend/onyx/db/permission_sync_attempt.py
```diff
@@ -0,0 +1,518 @@
...
+) -> None:
+    """Update the progress of a doc permission sync attempt."""
+    try:
+        attempt = db_session.execute(
+            select(DocPermissionSyncAttempt)
+            .where(DocPermissionSyncAttempt.id == attempt_id)
+            .with_for_update()
+        ).scalar_one()
+
+        attempt.total_docs_synced = (attempt.total_docs_synced or 0) + total_docs_synced
```
evan-onyx, line 193: somewhat weird to me that total_docs_synced and docs_with_permission_errors are nullable but its fine if there's a good reason for it

backend/onyx/db/permission_sync_attempt.py
```diff
@@ -0,0 +1,518 @@
...
+    """Get an external group sync attempt by ID."""
+    stmt = select(ExternalGroupPermissionSyncAttempt).where(
+        ExternalGroupPermissionSyncAttempt.id == attempt_id
+    )
+
+    if eager_load_cc_pair:
+        stmt = stmt.options(
+            joinedload(
+                ExternalGroupPermissionSyncAttempt.connector_credential_pair
+            ).joinedload(ConnectorCredentialPair.connector)
```
evan-onyx, line 274: same as above

backend/onyx/db/permission_sync_attempt.py
```diff
@@ -0,0 +1,518 @@
...
+                "cc_pair_id": attempt.connector_credential_pair_id,
+            },
+        )
+        return attempt
+    except Exception:
+        db_session.rollback()
+        raise
+
+
+def mark_external_group_sync_attempt_completed_with_errors(
```
evan-onyx: same comment about extracting implementation of these similar functions

backend/onyx/db/permission_sync_attempt.py
```diff
@@ -0,0 +1,518 @@
...
+        cc_pair_id: The connector credential pair ID
+
+    Returns:
+        The number of attempts deleted
+    """
+    stmt = delete(ExternalGroupPermissionSyncAttempt).where(
+        ExternalGroupPermissionSyncAttempt.connector_credential_pair_id == cc_pair_id
+    )
+    result = db_session.execute(stmt)
+    return result.rowcount or 0  # type: ignore
```
evan-onyx: why does this need a type: ignore?

evan-onyx (commented):
some things to do rq

backend/onyx/db/permission_sync_attempt.py
nmgarza5: yeah 2nd is intentional because I want the connector. will update to `eager_load_connector` 👍 

we are only loading one attempt here so `joinedload` should be ok. iiuc `selectinload` would be better to prevent duplicate rows

backend/onyx/db/permission_sync_attempt.py
nmgarza5: this is because when the attempt is initially kicked off or marked as in progress for the first time via `mark_doc_permission_sync_attempt_in_progress`, there are no docs synced at this time. I'm just going to remove this func entirely and consolidate into a completed func

backend/onyx/db/permission_sync_attempt.py
nmgarza5: I agree with you. I was just trying to be explicit with the function name and abstract any optional parameters away from the consumers/callsites. I will go with the following middle ground:

use two funcs
1. mark_attempt_completed (optional arg for errors)
2. mark_attempt_failed (keeping separate because the

backend/onyx/db/permission_sync_attempt.py
nmgarza5: without: 
`onyx/db/permission_sync_attempt.py:496: error: "Result[Any]" has no attribute "rowcount"  [attr-defined]`

we do this in other places of the codebase so i figured it was ok. I did try to add proper typing by setting the type as `Result` and `CursorResult` and both didn't work. 

....
kept testing and I can cast to CursorResult to get us a proper type 
`cast(CursorResult[Any], db_session.execute(stmt))`

PR #5697, opened 2025-10-13
fix(entra): JWT Passthrough for Entra

justin-tahara:
## Description

A customer uses Entra and when doing so, they have a different public keys endpoint and thus we need to handle for the new workflow in order to authenticate their JWT tokens.

## How Has This Been Tested?

Created a OIDC Entra Application to test the workflow end to end:

```
justintahara@justins-macbook-pro onyx % curl -X POST \
  http://localhost:3000/api/chat/create-chat-session \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{"persona_id": 0}'
{"chat_session_id":"1dc3b073-76fc-407b-b981-39983ea76a6a"}
```

Some additional information about how to test: 
I first used this command:
```
curl -X POST "https://login.microsoftonline.com/f28492f8-0adc-4bd8-995e-6dae854b9b72/oauth2/v2.0/devicecode" \
  -H "Content-Type: application/x-www-form-urlencoded" \
  -d "client_id=a6d03e72-c251-4543-a62a-32be414b6f20" \
  -d "scope=openid email profile User.Read"
```
I already updated the tenant id in the devicecode url and the client id to match what you gave me.
This then tells me to go to https://microsoft.com/devicelogin and then use the code for example `B2K663DAU` <- (expired code)
to authenticate and in the same message that it tells me to do this, there is a
```
"device_code":"BAQABIQEAAABlMNzVhAPUTrARzfQjWPtKatZGEKCHSNt4X1hv....
```
Then with another curl command you do the following:
```
curl -X POST "https://login.microsoftonline.com/f28492f8-0adc-4bd8-995e-6dae854b9b72/oauth2/v2.0/token" \
  -H "Content-Type: application/x-www-form-urlencoded" \
  -d "grant_type=urn:ietf:params:oauth:grant-type:device_code" \
  -d "client_id=a6d03e72-c251-4543-a62a-32be414b6f20" \
  -d "device_code=BAQABIQEAAABlMNzVhAPUTrARzfQjWPtKatZGEKCHSNt4X1hv...."
```
Which returns you a giant json blob.
In that blob there is a id_token JWT token. I believe what you have is the access_token JWT token.

The specific steps to test are:

1. Create an OIDC application in Entra
2. Then use some cURL commands to get a valid JWT token
3. Update my .env file to have a `JWT_PUBLIC_KEY_URL`
4. Switch to OIDC
5. Then use the above cURL command to test that the JWT passthrough was working properly

## Additional Options

### DIFF
--- a/backend/ee/onyx/auth/users.py
+++ b/backend/ee/onyx/auth/users.py
@@ -184,7 +184,7 @@ async def verify_jwt_token(token: str, async_db_session: AsyncSession) -> User |
         logger.warning(
             "JWT token decoded successfully but no email claim found; skipping auth"
         )
-        return None
+        break
 
     return None
 


justin-tahara:
@greptileai

backend/ee/onyx/auth/users.py
```diff
@@ -30,43 +34,147 @@
...
+
+    if "application/json" in content_type or body_lstripped.startswith("{"):
+        try:
+            data = response.json()
+        except ValueError:
+            logger.error("JWT public key URL returned invalid JSON")
+            return None
 
+        if isinstance(data, dict) and "keys" in data:
+            return data, "jwks"
```
Weves: this should probably be an enum with "jwks" and "pem" as the two options

backend/ee/onyx/auth/users.py
```diff
@@ -30,43 +34,147 @@
...
+                continue
+            return None
+        except PyJWTError as e:
+            logger.error(f"JWT decoding error: {str(e)}")
+            if attempt < _PUBLIC_KEY_FETCH_ATTEMPTS - 1:
+                get_public_key.cache_clear()
+                continue
+            return None
+        else:
+            email = payload.get("email")
```
Weves: nit: this all could be moved out of the `else` block? Slightly cleaner that way imo

backend/ee/onyx/auth/users.py
```diff
@@ -30,43 +34,147 @@
...
-        if public_key_pem is None:
+        return RSAAlgorithm.from_jwk(json.dumps(jwk))
+    except ValueError as e:
+        logger.error(f"Failed to construct RSA key from JWK: {str(e)}")
+        return None
+
+
+async def verify_jwt_token(token: str, async_db_session: AsyncSession) -> User | None:
+    for attempt in range(_PUBLIC_KEY_FETCH_ATTEMPTS):
+        public_key_payload = get_public_key()
```
Weves: could `get_public_key` handle the stuff below? It feels cleaner to have the function always return one type

justin-tahara:
@greptileai

PR #5700, opened 2025-10-13
feat(vectordb): New Document Index Interface

yuhongsun96:
## Description

No live code in it

## How Has This Been Tested?

## Additional Options

---
## Summary by cubic
Introduces a unified document index interface with clear contracts for indexing, updates, deletions, verification, and retrieval. This is scaffolding only; no live code paths or runtime impact.

- **New Features**
  - Models: DocumentInsertionRecord, DocumentSectionRequest, IndexingMetadata, MetadataUpdateRequest.
  - Abstract methods: SchemaVerifiable.verify_and_create_index_if_necessary, Indexable.index, Deletable.delete, Updatable.update, IdRetrievalCapable.id_based_retrieval, HybridCapable.hybrid_retrieval, RandomCapable.random_retrieval; DocumentIndex composes these capabilities.
  - Enum: QueryType (keyword, semantic) for hybrid search.

<sup>Written for commit 9c8b0edf2fa2f097de032d9148ca9f0a54d491b3. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/document_index/interfaces_new.py
+++ b/backend/onyx/document_index/interfaces_new.py
@@ -24,7 +24,6 @@ from shared_configs.model_server_models import Embedding
 __all__ = [
     # Main interfaces - these are what you should inherit from
     "DocumentIndex",
-    "BaseIndex",
     # Data models - used in method signatures
     "DocumentInsertionRecord",
     "DocumentSectionRequest",
@@ -113,7 +112,7 @@ class SchemaVerifiable(abc.ABC):
         self.tenant_id = tenant_id
 
     @abc.abstractmethod
-    def ensure_indices_exist(
+    def verify_and_create_index_if_necessary(
         self,
         embedding_dim: int,
         embedding_precision: EmbeddingPrecision,
@@ -269,14 +268,9 @@ class HybridCapable(abc.ABC):
         """
         Run hybrid search and return a list of inference chunks.
 
-        NOTE: the query passed in here is the unprocessed plain text query. Preprocessing is
-        expected to be handled by this function as it may depend on the index implementation.
-        Things like query expansion, synonym injection, stop word removal, lemmatization, etc. are
-        done here.
-
         Parameters:
         - query: unmodified user query. This may be needed for getting the matching highlighted
-                keywords
+                keywords or for logging purposes
         - query_embedding: vector representation of the query, must be of the correct
                 dimensionality for the primary index
         - final_keywords: Final keywords to be used from the query, defaults to query if not set
@@ -306,28 +300,26 @@ class RandomCapable(abc.ABC):
         raise NotImplementedError
 
 
-class BaseIndex(
+class DocumentIndex(
     SchemaVerifiable,
     Indexable,
     Updatable,
     Deletable,
+    HybridCapable,
+    IdRetrievalCapable,
+    RandomCapable,
     abc.ABC,
-):
-    """
-    All basic document index functionalities excluding the actual querying approach.
-
-    As a summary, document indices need to be able to
-    - Verify the schema definition is valid
-    - Index new documents
-    - Update specific attributes of existing documents
-    - Delete documents
-    """
-
-
-class DocumentIndex(
-    HybridCapable, IdRetrievalCapable, RandomCapable, BaseIndex, abc.ABC
 ):
     """
     A valid document index that can plug into all Onyx flows must implement all of these
     functionalities.
+
+    As a high level summary, document indices need to be able to
+    - Verify the schema definition is valid
+    - Index new documents
+    - Update specific attributes of existing documents
+    - Delete documents
+    - Run hybrid search
+    - Retrieve document or sections of documents based on document id
+    - Retrieve sets of random documents
     """


backend/onyx/document_index/interfaces_new.py
```diff
@@ -0,0 +1,338 @@
...
+    As a summary, document indices need to be able to
+    - Verify the schema definition is valid
+    - Index new documents
+    - Update specific attributes of existing documents
+    - Delete documents
+    """
+
+
+class DocumentIndex(
+    HybridCapable, BaseIndex, IdRetrievalCapable, RandomCapable, abc.ABC
```
edwin-onyx: maybe we can look to use composition over mixins? inheritance gets quite messy in cases like this, think composition makes dependencies much more explicit

backend/onyx/document_index/interfaces_new.py
```diff
@@ -0,0 +1,338 @@
...
+
+    connector_id: int
+    credential_id: int
+    document_id: str
+    semantic_identifier: str
+    first_link: str
+    doc_updated_at: datetime | None = None
+    # Emails, not necessarily attached to users
+    # Users may not be in Onyx
+    primary_owners: list[str] | None = None
```
edwin-onyx: nit, can rename to primary_owner_emails for explicitness?

backend/onyx/document_index/interfaces_new.py
```diff
@@ -0,0 +1,338 @@
...
+    secondary_owners: list[str] | None = None
+    from_ingestion_api: bool = False
```
edwin-onyx: me no ctx what is this?

backend/onyx/document_index/interfaces_new.py
```diff
@@ -0,0 +1,338 @@
...
+
+class Verifiable(abc.ABC):
+    """
+    Class must implement document index schema verification. For example, verify that all of the
+    necessary attributes for indexing, querying, filtering, and fields to return from search are
+    all valid in the schema.
+    """
+
+    @abc.abstractmethod
+    def __init__(
```
edwin-onyx, line 103: why does the index name need to be here?, could it be a param of the method?

backend/onyx/document_index/interfaces_new.py
```diff
@@ -0,0 +1,338 @@
...
+    def random_retrieval(
+        self,
+        filters: IndexFilters,
+        num_to_retrieve: int = 10,
+    ) -> list[InferenceChunkUncleaned]:
+        """Retrieve random chunks matching the filters"""
+        raise NotImplementedError
+
+
+class BaseIndex(
```
edwin-onyx: why even use mixins here in the first place, it seems that all of these are core functionality for a vector store?

backend/onyx/document_index/interfaces_new.py
evan-onyx: there's a way to index documents via our ingestion api instead of in a connector; people typically use it for large files

backend/onyx/document_index/interfaces_new.py
evan-onyx: yah this is weird

backend/onyx/document_index/interfaces_new.py
```diff
@@ -0,0 +1,335 @@
...
+    user_project_ids: list[int] | None = None
+
+
+class MetadataUpdateRequest(BaseModel):
+    """
+    Updates to the documents that can happen without there being an update to the contents of the document.
+    """
+
+    document_ids: list[str]
+    doc_id_to_chunk_cnt: dict[str, int]
```
evan-onyx, line 86: can't we just use the keys of this instead of explicitly passing document_ids? if this is not valid then these fields definitely need comments or aren't named well

backend/onyx/document_index/interfaces_new.py
```diff
@@ -0,0 +1,335 @@
...
+
+
+class IndexingMetadata(BaseModel):
+    """
+    Information about chunk counts for efficient cleaning / updating of document chunks. A common pattern to ensure
+    that no chunks are left over is to delete all of the chunks for a document and then re-index the document. This
+    information allows us to only delete the extra "tail" chunks when the document has gotten shorter.
+    """
+
+    doc_id_to_old_chunk_cnt: dict[str, int]
```
evan-onyx: imo `doc_id_to_chunk_cnt_change: dct[str, tuple[int, int]]`

but also I don't understand how this is supposed to work, how do you know which chunks to delete? How do you detect a pure deletion vs deletion plus edits?

backend/onyx/document_index/interfaces_new.py
```diff
@@ -0,0 +1,335 @@
...
+    large_chunks_enabled: bool  # Meaning larger chunks should also get deleted
+
+
+class DocumentMetadata(BaseModel):
+    """
+    Document information that needs to be inserted into Postgres on first time encountering this
+    document during indexing across any of the connectors.
+    """
+
+    connector_id: int
```
evan-onyx: "during indexing across any of the connectors" then why is there cc pair info in here? What happens when the one cc pair this document is associated with gets deleted but the document is actually indexed by several connectors?

backend/onyx/document_index/interfaces_new.py
```diff
@@ -0,0 +1,335 @@
...
+
+    document_ids: list[str]
+    doc_id_to_chunk_cnt: dict[str, int]
+    access: DocumentAccess | None = None
+    document_sets: set[str] | None = None
+    boost: float | None = None
+    hidden: bool | None = None
+
+
+class Verifiable(abc.ABC):
```
evan-onyx: this name is way too broad, no? VerifiableIndexSchema maybe?

backend/onyx/document_index/interfaces_new.py
edwin-onyx: hm then maybe better to have this be a creation_source field with IngestionAPI as an enum option

backend/onyx/document_index/interfaces_new.py
```diff
@@ -0,0 +1,335 @@
...
+        does not exist.
+
+        Parameters:
+        - embedding_dim: Vector dimensionality for the vector similarity part of the search
+        - embedding_precision: Precision of the vector similarity part of the search
+        """
+        raise NotImplementedError
+
+
+class Indexable(abc.ABC):
```
evan-onyx, line 132: this name sounds like "something that can be indexed" rather than "something that CAN index chunks". I'd prefer smth like "Indexer"

backend/onyx/document_index/interfaces_new.py
```diff
@@ -0,0 +1,335 @@
...
+
+        Returns:
+            List of document ids which map to unique documents and are used for deduping chunks
+            when updating, as well as if the document is newly indexed or already existed and
+            just updated
+        """
+        raise NotImplementedError
+
+
+class Deletable(abc.ABC):
```
evan-onyx, line 166: same as above

backend/onyx/document_index/interfaces_new.py
```diff
@@ -0,0 +1,335 @@
...
+        """
+        Takes a list of document chunks and indexes them in the document index. This is often a batch operation
+        including chunks from multiple documents.
+
+        NOTE: When a document is reindexed/updated here, it must clear all of the existing document
+        chunks before reindexing. This is because the document may have gotten shorter since the
+        last run. Therefore, upserting the first 0 through n chunks may leave some old chunks that
+        have not been written over.
+
+        NOTE: The chunks of a document are never separated into separate index() calls. So there is
```
evan-onyx, line 150: this doesn't scale: it requires materializing the entire document in memory whenever we index it. I suppose I'm okay with that being the case going forward, but it's definitely a significant design limitation

backend/onyx/document_index/interfaces_new.py
```diff
@@ -0,0 +1,335 @@
...
+    def index(
+        self,
+        chunks: list[DocMetadataAwareIndexChunk],
+        indexing_metadata: IndexingMetadata,
+    ) -> set[DocumentInsertionRecord]:
+        """
+        Takes a list of document chunks and indexes them in the document index. This is often a batch operation
+        including chunks from multiple documents.
+
+        NOTE: When a document is reindexed/updated here, it must clear all of the existing document
```
evan-onyx: this comment seems to go against the whole point of IndexingMetadata as described above, although tbf I don't understand how that was intended to work

backend/onyx/document_index/interfaces_new.py
```diff
@@ -0,0 +1,335 @@
...
+        - doc_id: document id as represented in Onyx
+        - chunk_count: number of chunks in the document
+
+        Returns:
+            number of chunks deleted
+        """
+        raise NotImplementedError
+
+
+class Updatable(abc.ABC):
```
evan-onyx, line 193: Same as above

backend/onyx/document_index/interfaces_new.py
```diff
@@ -0,0 +1,335 @@
...
+    """
+    Class must implement the ability to update certain attributes of a document without needing to
+    update all of the fields. Specifically, needs to be able to update:
+    - Access Control List
+    - Document-set membership
+    - Boost value (learning from feedback mechanism)
+    - Whether the document is hidden or not, hidden documents are not returned from search
```
evan-onyx, line 200: presumably these could be added to in the future, but not a big deal

backend/onyx/document_index/interfaces_new.py
evan-onyx: +1 imo this kind of thing makes more sense if we were implementing multiple vector stores. I suppose it's reasonable if the point is we think open source folks will WANT to add their own vector store implementations that have a subset of this functionality. And then the point is that parts of the codebase that use the vector store can declare different subsets of these as the input vector store type based on what they need

backend/onyx/document_index/interfaces_new.py
yuhongsun96: Ya I was thinking that too, I'm not opposed to this change. This is useful because you can possibly delete by id and you know which ids there are by constructing it from document id and chunk index. This is also useful in updating docs since you don't need to pre-delete all the chunks.

backend/onyx/document_index/interfaces_new.py
yuhongsun96: Comment is inaccurate, this is just what's stored to Postgres. There's an argument that maybe this isn't the right place for this class but it already existed here.

yuhongsun96: I guess ya removing it from here makes more sense

backend/onyx/document_index/interfaces_new.py
yuhongsun96: Ya, it is redundant

yuhongsun96: I think it is more clear though, if you just had the dict, it's a bit unexpected that you're supposed to be acting on all the keys

backend/onyx/document_index/interfaces_new.py
yuhongsun96: fine

backend/onyx/document_index/interfaces_new.py
yuhongsun96: It is enforcing that the class must be created with the index name, I think we may want to add tenant here too honestly

backend/onyx/document_index/interfaces_new.py
yuhongsun96: image
looks like this old man still has a few tricks to teach

backend/onyx/document_index/interfaces_new.py
yuhongsun96: Isn't it more clear to think of it as "this class must implement this thing" rather than "this class must have something that implements this thing"

backend/onyx/document_index/interfaces_new.py
yuhongsun96: Previously we had hybrid search, keyword search, vector search, and other searches. We had hybrid be implemented by two separate search engines (one did keyword only, one did vector only), both implemented BaseIndex. Just as a bit of history. But I guess at this point there's only one so it can all be crammed into one but this doesn't feel like it hurts in any way?

backend/onyx/document_index/interfaces_new.py
```diff
@@ -0,0 +1,333 @@
...
+    document_sets: set[str] | None = None
+    boost: float | None = None
+    hidden: bool | None = None
+    secondary_index_updated: bool | None = None
+    project_ids: set[int] | None = None
+
+
+class SchemaVerifiable(abc.ABC):
+    """
+    Class must implement document index schema verification. For example, verify that all of the
```
acaprau, line 98: from `BaseIndex`, which inherits this: "All basic document index functionalities excluding the actual querying approach." but this interface concerns itself with querying. this seems like abstraction leak.

backend/onyx/document_index/interfaces_new.py
acaprau: +1. is there a technical reason inherent to vector dbs for why we cant supply different chunks in different batches?

backend/onyx/document_index/interfaces_new.py
acaprau: Just documenting my own understanding, we ensure indices exist on onyx start, and the DocumentIndex instance will always correspond to a tenant so I think this makes sense. If it is possible to switch index name mid onyx instance then maybe it makes less sense for that to be in the init vs tenant id?

backend/onyx/document_index/interfaces_new.py
```diff
@@ -0,0 +1,333 @@
...
+        tenant_id: int | None,
+        *args: Any,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self.index_name = index_name
+        self.tenant_id = tenant_id
+
+    @abc.abstractmethod
+    def ensure_indices_exist(
```
acaprau: Super nit comment but I don't think this name sufficiently indicates that this method can also create a schema. Maybe something like check_for_and_create_index_if_necessary? Just a thought.

backend/onyx/document_index/interfaces_new.py
```diff
@@ -0,0 +1,333 @@
...
+        self,
+        embedding_dim: int,
+        embedding_precision: EmbeddingPrecision,
+    ) -> None:
```
acaprau, line 119: More general thought that is not specific to this work, I'm assuming this is meant to throw some exception if the schema is invalid. I don't think it's generally good practice to rely on exceptions, in an ideal world we should be returning statuses when possible.

backend/onyx/document_index/interfaces_new.py
```diff
@@ -0,0 +1,333 @@
...
+
+
+class MetadataUpdateRequest(BaseModel):
+    """
+    Updates to the documents that can happen without there being an update to the contents of the document.
+    """
+
+    document_ids: list[str]
+    # Passed in to help with potential optimizations of the implementation
+    doc_id_to_chunk_cnt: dict[str, int]
```
acaprau, line 86: Just to be extra explicit, as opposed to relying on an empty dict to imply this, can we make this optional like `delete`? Shouldn't be a part of the required API, even an empty dict.

backend/onyx/document_index/interfaces_new.py
```diff
@@ -0,0 +1,333 @@
...
+
+    @abc.abstractmethod
+    def id_based_retrieval(
+        self,
+        chunk_requests: list[DocumentSectionRequest],
+    ) -> list[InferenceChunk]:
+        """
+        Fetch chunk(s) based on document id
+
+        NOTE: This is used to reconstruct a full document or an extended (multi-chunk) section
```
acaprau, line 234: Just to be clear, we mean reconstruct a document embedding?

backend/onyx/document_index/interfaces_new.py
```diff
@@ -0,0 +1,333 @@
...
+        query_type: QueryType,
+        filters: IndexFilters,
+        num_to_retrieve: int,
+        offset: int = 0,
+    ) -> list[InferenceChunk]:
+        """
+        Run hybrid search and return a list of inference chunks.
+
+        NOTE: the query passed in here is the unprocessed plain text query. Preprocessing is
+        expected to be handled by this function as it may depend on the index implementation.
```
acaprau: Just clarifying, you mean how the db implements indices, or the index embedding? For the former I agree, for the latter I don't see why this method cares about that implementation detail.

backend/onyx/document_index/interfaces_new.py
```diff
@@ -0,0 +1,333 @@
...
+
+class HybridCapable(abc.ABC):
+    """
+    Class must implement hybrid (keyword + vector) search functionality
+    """
+
+    @abc.abstractmethod
+    def hybrid_retrieval(
+        self,
+        query: str,
```
acaprau, line 260: Which of these params are inherent to hybrid search, and which are specific to Vespa? Are all of these relevant to all vector dbs?

backend/onyx/document_index/interfaces_new.py
evan-onyx: ChatGPT is agreeing w/me.... T.T Indexable == "can be indexed", and objects of this class cannot be indexed. This class implements behavior to index Other stuff

backend/onyx/document_index/interfaces_new.py
evan-onyx: nothing inherent to vector dbs, but there are some external reasons why we do full documents at a time at other stages in the pipeline (to make summarization more convenient and making it simpler to detect when a full document is done being indexed)

backend/onyx/document_index/interfaces_new.py
evan-onyx: These are all pretty generic, although ofc the form that the embedding and filters take might differ across vector db implementations

backend/onyx/document_index/interfaces_new.py
yuhongsun96: Have you tried asking st-dev?

PR #5704, opened 2025-10-13
fix(slack): Fix Fed Slack Gear Button Error

justin-tahara:
## Description

Trying to fix the Gear Button that exists for the Federated Connector Selector. The current behavior before the change is that the Gear button works when we are creating the Document Set but when we create a new document set and then we try to click on the gear icon afterwards, it automatically closes the page and goes back to the main Document Sets page. 

This PR aims to fix this functionality due to the button component not having a proper type setup. 

At the same time, I am using the new components UI to ensure that the fixes represent the new UI changes the team has been working on. 

## How Has This Been Tested?

Brought up my local instance and set up a new Fed Slack Connector to test this workflow. 

https://github.com/user-attachments/assets/7ba0227b-1671-47fb-ab6b-52e22fa171fd

## Additional Options

### DIFF
--- a/web/src/components/FederatedConnectorSelector.tsx
+++ b/web/src/components/FederatedConnectorSelector.tsx
@@ -6,13 +6,14 @@ import {
   ValidSources,
 } from "@/lib/types";
 import { SourceIcon } from "@/components/SourceIcon";
-import SvgSearch from "@/icons/search";
 import SvgX from "@/icons/x";
 import SvgSettings from "@/icons/settings";
 import { Label } from "@/components/ui/label";
 import { ErrorMessage } from "formik";
-import { Input } from "@/components/ui/input";
+import Text from "@/refresh-components/Text";
 import IconButton from "@/refresh-components/buttons/IconButton";
+import { Input } from "@/components/ui/input";
+import InputTypeIn from "@/refresh-components/inputs/InputTypeIn";
 import {
   Dialog,
   DialogContent,
@@ -374,48 +375,41 @@ export const FederatedConnectorSelector = ({
     <div className="flex flex-col w-full space-y-2 mb-4">
       {label && <Label className="text-base font-medium">{label}</Label>}
 
-      <p className="text-xs text-neutral-500 dark:text-neutral-400">
+      <Text mainUiMuted text03>
         Documents from selected federated connectors will be searched in
         real-time during queries.
-      </p>
+      </Text>
       <div className="relative">
-        <div
-          className={`flex items-center border border-input rounded-md border-neutral-200 dark:border-neutral-700 ${
-            allConnectorsSelected ? "bg-neutral-50 dark:bg-neutral-800" : ""
-          } focus-within:ring-1 focus-within:ring-ring focus-within:border-neutral-400 dark:focus-within:border-neutral-500 transition-colors`}
-        >
-          <SvgSearch className="absolute left-3 h-4 w-4 stroke-current text-neutral-500 dark:text-neutral-400" />
-          <input
-            ref={inputRef}
-            type="text"
-            value={searchQuery}
-            onChange={(e) => {
-              setSearchQuery(e.target.value);
+        <InputTypeIn
+          ref={inputRef}
+          leftSearchIcon
+          placeholder={effectivePlaceholder}
+          value={searchQuery}
+          disabled={isInputDisabled}
+          onChange={(e) => {
+            setSearchQuery(e.target.value);
+            setOpen(true);
+          }}
+          onKeyDown={handleKeyDown}
+          onFocus={() => {
+            if (!allConnectorsSelected) {
               setOpen(true);
-            }}
-            onFocus={() => {
-              if (!allConnectorsSelected) {
-                setOpen(true);
-              }
-            }}
-            onKeyDown={handleKeyDown}
-            placeholder={effectivePlaceholder}
-            className={`h-9 w-full pl-9 pr-10 py-2 bg-transparent dark:bg-transparent text-sm outline-none disabled:cursor-not-allowed disabled:opacity-50 ${
-              allConnectorsSelected
-                ? "text-neutral-500 dark:text-neutral-400"
-                : ""
-            }`}
-            disabled={isInputDisabled}
-          />
-        </div>
+            }
+          }}
+          className={
+            allConnectorsSelected
+              ? "rounded-12 bg-background-neutral-01"
+              : "rounded-12"
+          }
+        />
 
         {open && !allConnectorsSelected && (
           <div
             ref={dropdownRef}
-            className="absolute z-50 w-full mt-1 rounded-md border border-neutral-200 dark:border-neutral-700 bg-white dark:bg-neutral-900 shadow-md default-scrollbar max-h-[300px] overflow-auto"
+            className="absolute z-50 w-full mt-1 rounded-12 border border-border-02 bg-background-neutral-00 shadow-md default-scrollbar max-h-[300px] overflow-auto"
           >
             {filteredUnselectedConnectors.length === 0 ? (
-              <div className="py-4 text-center text-xs text-neutral-500 dark:text-neutral-400">
+              <div className="py-4 text-center text-xs text-text-03">
                 {searchQuery
                   ? "No matching federated connectors found"
                   : "No more federated connectors available"}
@@ -425,7 +419,7 @@ export const FederatedConnectorSelector = ({
                 {filteredUnselectedConnectors.map((connector) => (
                   <div
                     key={connector.id}
-                    className="flex items-center justify-between py-2 px-3 cursor-pointer hover:bg-neutral-50 dark:hover:bg-neutral-800 text-xs"
+                    className="flex items-center justify-between py-2 px-3 cursor-pointer hover:bg-background-neutral-01 text-xs"
                     onClick={() => selectConnector(connector.id)}
                   >
                     <div className="flex items-center truncate mr-2">
@@ -460,7 +454,7 @@ export const FederatedConnectorSelector = ({
               return (
                 <div
                   key={connector.id}
-                  className="flex items-center bg-white dark:bg-neutral-800 rounded-md border border-neutral-300 dark:border-neutral-700 transition-all px-2 py-1 max-w-full group text-xs"
+                  className="flex items-center bg-background-neutral-00 rounded-12 border border-border-02 transition-all px-2 py-1 max-w-full group text-xs"
                 >
                   <div className="flex items-center overflow-hidden">
                     <div className="mr-1 flex-shrink-0">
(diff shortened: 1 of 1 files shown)


web/src/components/FederatedConnectorSelector.tsx
```diff
@@ -9,15 +9,15 @@ import { SourceIcon } from "@/components/SourceIcon";
 import { X, Search, Settings } from "lucide-react";
 import { Label } from "@/components/ui/label";
 import { ErrorMessage } from "formik";
-import { Button } from "@/components/ui/button";
+import { Button as UiButton } from "@/components/ui/button";
```
raunakab: Would prefer to delete usage of this button and only use the new one!

web/src/components/FederatedConnectorSelector.tsx
```diff
@@ -9,15 +9,15 @@ import { SourceIcon } from "@/components/SourceIcon";
 import { X, Search, Settings } from "lucide-react";
```
raunakab: Could you use the icons from the `src/icons` directory, please?

```ts
import SvgSearch from "@/icons/search";
import SvgX from "@/icons/x";
import SvgSettings from "@/icons/settings";
```

web/src/components/FederatedConnectorSelector.tsx
```diff
@@ -9,15 +9,15 @@ import { SourceIcon } from "@/components/SourceIcon";
...
+import { Button as UiButton } from "@/components/ui/button";
 import { Input } from "@/components/ui/input";
 import {
   Dialog,
   DialogContent,
   DialogHeader,
   DialogTitle,
-  DialogTrigger,
 } from "@/components/ui/dialog";
+import RefreshButton from "@/refresh-components/buttons/Button";
```
raunakab: If you delete the usage of the previous one, then you should rename this to `Button`.

web/src/components/FederatedConnectorSelector.tsx
```diff
@@ -470,21 +470,25 @@ export const FederatedConnectorSelector = ({
...
-                    <button
-                      className="flex-shrink-0 rounded-full w-4 h-4 flex items-center justify-center bg-neutral-100 dark:bg-neutral-700 text-neutral-500 dark:text-neutral-400 hover:bg-neutral-200 dark:hover:bg-neutral-600 hover:text-neutral-700 dark:hover:text-neutral-300 transition-colors"
-                      onClick={() => openConfigDialog(connector.id)}
+                    <RefreshButton
+                      tertiary
+                      type="button"
                       aria-label="Configure entities"
                       title="Configure entities"
+                      onClick={() => openConfigDialog(connector.id)}
+                      className="flex-shrink-0 !p-1 !gap-0 !rounded-full !w-6 !h-6 !bg-neutral-100 !text-neutral-500 hover:!bg-neutral-200 hover:!text-neutral-700 dark:!bg-neutral-700 dark:!text-neutral-400 dark:hover:!bg-neutral-600 dark:hover:!text-neutral-300 [&>div:first-child]:hidden [&>div:last-child]:hidden"
```
raunakab: Let's remove all of these custom stylings from the "refresh"-d buttons. `dark:` modifiers should *definitely* not be used anymore, since colour-scheming is all handled at the CSS DOM level (by `src/app/figma/colors.css`).

web/src/components/FederatedConnectorSelector.tsx
```diff
@@ -470,21 +470,25 @@ export const FederatedConnectorSelector = ({
...
-                    <button
-                      className="flex-shrink-0 rounded-full w-4 h-4 flex items-center justify-center bg-neutral-100 dark:bg-neutral-700 text-neutral-500 dark:text-neutral-400 hover:bg-neutral-200 dark:hover:bg-neutral-600 hover:text-neutral-700 dark:hover:text-neutral-300 transition-colors"
-                      onClick={() => removeConnector(connector.id)}
+                    </RefreshButton>
+                    <RefreshButton
+                      tertiary
+                      type="button"
                       aria-label="Remove connector"
+                      onClick={() => removeConnector(connector.id)}
+                      className="flex-shrink-0 !p-1 !gap-0 !rounded-full !w-6 !h-6 !bg-neutral-100 !text-neutral-500 hover:!bg-neutral-200 hover:!text-neutral-700 dark:!bg-neutral-700 dark:!text-neutral-400 dark:hover:!bg-neutral-600 dark:hover:!text-neutral-300 [&>div:first-child]:hidden [&>div:last-child]:hidden"
```
raunakab: Would prefer to remove these as well!

web/src/components/FederatedConnectorSelector.tsx
raunakab: Unless, of course, there are some alignment styles that you'd like to preserve (i.e., usages of padding [`px-5`, etc.]).

web/src/components/FederatedConnectorSelector.tsx
```diff
@@ -470,21 +482,24 @@ export const FederatedConnectorSelector = ({
...
-                      title="Configure entities"
-                    >
-                      <Settings className="h-2.5 w-2.5" />
-                    </button>
-                    <button
-                      className="flex-shrink-0 rounded-full w-4 h-4 flex items-center justify-center bg-neutral-100 dark:bg-neutral-700 text-neutral-500 dark:text-neutral-400 hover:bg-neutral-200 dark:hover:bg-neutral-600 hover:text-neutral-700 dark:hover:text-neutral-300 transition-colors"
-                      onClick={() => removeConnector(connector.id)}
+                      onClick={() => openConfigDialog(connector.id)}
+                      icon={SvgSettings}
+                      className="!p-1 !rounded-full"
```
raunakab: How come this is needed?

web/src/components/FederatedConnectorSelector.tsx
```diff
@@ -470,21 +482,24 @@ export const FederatedConnectorSelector = ({
...
+                      internal
+                      type="button"
                       aria-label="Remove connector"
-                    >
-                      <X className="h-2.5 w-2.5" />
-                    </button>
+                      tooltip="Remove connector"
+                      onClick={() => removeConnector(connector.id)}
+                      icon={SvgX}
+                      className="!p-1 !rounded-full"
```
raunakab: Same question here: How come this is needed? We should try as much as possible to not modify the design of each component!

web/src/components/FederatedConnectorSelector.tsx
```diff
@@ -372,7 +384,7 @@ export const FederatedConnectorSelector = ({
             allConnectorsSelected ? "bg-neutral-50 dark:bg-neutral-800" : ""
           } focus-within:ring-1 focus-within:ring-ring focus-within:border-neutral-400 dark:focus-within:border-neutral-500 transition-colors`}
         >
-          <Search className="absolute left-3 h-4 w-4 text-neutral-500 dark:text-neutral-400" />
+          <SvgSearch className="absolute left-3 h-4 w-4 stroke-current text-neutral-500 dark:text-neutral-400" />
```
raunakab: Ideally, we don't use the `dark:` modifiers at all. Is there a reason why we need to break into custom dark-mode colours as opposed to the ones which are auto-provided by `colors.css`?

raunakab:
@justin-tahara I see, for example, that the `SvgX` that is present in the button above has full-roundedness. This goes against the unified design that @duo-onyx has created, so in order to adhere to principles, we should avoid overriding the refreshed components' stylings!

web/src/components/FederatedConnectorSelector.tsx
```diff
@@ -362,17 +375,20 @@ export const FederatedConnectorSelector = ({
...
+          className={cn(
+            "flex items-center border rounded-12 transition-colors bg-background-neutral-00 border-border",
+            "focus-within:border-border-05",
+            allConnectorsSelected && "bg-background-neutral-01",
+            isInputDisabled && "cursor-not-allowed"
+          )}
         >
-          <Search className="absolute left-3 h-4 w-4 text-neutral-500 dark:text-neutral-400" />
+          <SvgSearch className="absolute left-3 h-4 w-4 stroke-current text-text-02" />
           <input
```
raunakab: Not super pressing, but we should move towards using the `InputTypeIn` component for all of these styles of inputs!

web/src/components/FederatedConnectorSelector.tsx
```diff
@@ -362,17 +375,20 @@ export const FederatedConnectorSelector = ({
     <div className="flex flex-col w-full space-y-2 mb-4">
       {label && <Label className="text-base font-medium">{label}</Label>}
 
-      <p className="text-xs text-neutral-500 dark:text-neutral-400">
+      <p className="text-xs text-text-03">
         Documents from selected federated connectors will be searched in
         real-time during queries.
       </p>
```
raunakab: Not super pressing, but we should aim to render texts with `@/refresh-components/Text`.

This one isn't pressing, per se, because the default font-styles have been applied uniformly by the `src/app/globals.css` file. As a result, `p`, `div`, etc. tags all will still default to using @duo-onyx's styles. The benefit to using the refreshed `Text` component, however, is that it helps restrict the colours, sizings, etc. to fixed enumerations which can be activated using "flags" in `JSX` instead.

But this is definitely NOT pressing, so don't worry about it. This is just context for *future* changes.

web/src/components/FederatedConnectorSelector.tsx
raunakab: Just for additional context: the `InputTypeIn` is a component specifically design by @duo-onyx to also contain the `SvgSearch` icon in the left hand side, and it contains a whole bunch of other default goodies (such as focus state, hover state, disabled state, etc.). We want to absolutely unify in the coming future!

web/src/components/FederatedConnectorSelector.tsx
```diff
@@ -388,22 +404,21 @@ export const FederatedConnectorSelector = ({
             }}
             onKeyDown={handleKeyDown}
             placeholder={effectivePlaceholder}
-            className={`h-9 w-full pl-9 pr-10 py-2 bg-transparent dark:bg-transparent text-sm outline-none disabled:cursor-not-allowed disabled:opacity-50 ${
-              allConnectorsSelected
-                ? "text-neutral-500 dark:text-neutral-400"
-                : ""
-            }`}
+            className={cn(
```
raunakab: Thank you so much for using `cn` instead of raw string-formatting! This is why you're the 🐐.

raunakab:
@justin-tahara, if you get time, it'd be nice if you could update that `<SvgSearch {..} /><input {..} />` thing to just be `<InputTypeIn {..} />`! It will contain all the goodies that you're looking for.

If this is highly time-sensitive, then no worries. @duo-onyx and I will be running a revamp of the Admin panel in the coming future, so this will probably be cleaned up then anyways.

PR #5736, opened 2025-10-15
feat(onboarding): Pin Featured Agents to New Users

justin-tahara:
## Description

There was an old workflow in which we would show Featured Assistants as pinned assistants for new users since their `pinned_assistants` would be `NULL` meaning in the FE we were rendering the Featured Assistants as pinned. 

This functionality was removed due to the new UI refresh that happened and thus the FE was no longer handling this logic and thus customers were seeing their pinned assistants disappearing. 

## How Has This Been Tested?

I created a dummy user and then signed up after being invited and saw that the Agent was pinned properly. 

Also added tests in order to ensure we maintain this functionality.

## Additional Options

---

## Summary by cubic
Automatically pin Featured Agents for new users at signup so they see recommended assistants by default. Restores the behavior lost during the UI refresh.

- **Bug Fixes**
  - Set pinned_assistants on user creation (including OAuth) when none are set.
  - Select personas marked default, public, visible, and not deleted; ordered by display_priority then id.
  - Persist persona IDs on the user to remove the old frontend fallback.

### DIFF
--- a/backend/tests/unit/onyx/auth/test_user_default_pins.py
+++ b/backend/tests/unit/onyx/auth/test_user_default_pins.py
@@ -36,6 +36,7 @@ async def test_assign_default_pinned_assistants_populates_ids(
     assert db_session.execute.await_count == 1
     user_db.update.assert_awaited_once()
     await_args = user_db.update.await_args
+    assert await_args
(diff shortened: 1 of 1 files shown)


justin-tahara:
@greptileai

PR #5740, opened 2025-10-16
fix: s3 connector citation bugs

wenxi-onyx:
## Description

## How Has This Been Tested?

## Additional Options

---

## Summary by cubic
Fix citation links in the S3 connector by detecting the bucket’s actual region and URL-encoding object keys. Also trim bucket names to prevent malformed URLs.

- **Bug Fixes**
  - Detect and cache S3 bucket region via head_bucket; use it when building links.
  - URL-encode object keys (keep “/” unencoded) for S3, R2, GCS, and OCI links.
  - Strip whitespace from bucket_name during initialization.

### DIFF
--- a/backend/onyx/connectors/blob/connector.py
+++ b/backend/onyx/connectors/blob/connector.py
@@ -7,6 +7,7 @@ from io import BytesIO
 from numbers import Integral
 from typing import Any
 from typing import Optional
+from urllib.parse import quote
 
 import boto3  # type: ignore
 from botocore.client import Config  # type: ignore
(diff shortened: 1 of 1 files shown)


backend/onyx/connectors/blob/connector.py
```diff
@@ -257,21 +285,25 @@ def _get_blob_link(self, key: str) -> str:
         if self.s3_client is None:
             raise ConnectorMissingCredentialError("Blob storage")
 
+        # URL encode the key to handle special characters, spaces, etc.
+        # safe='/' keeps forward slashes unencoded for proper path structure
+        encoded_key = quote(key, safe="/")
```
justin-tahara, line 290: Does this change work across all blob storages?

backend/onyx/connectors/blob/connector.py
```diff
@@ -257,21 +285,25 @@ def _get_blob_link(self, key: str) -> str:
...
 
         elif self.bucket_type == BlobType.S3:
-            region = self.s3_client.meta.region_name
-            return f"https://{self.bucket_name}.s3.{region}.amazonaws.com/{key}"
+            region = self._bucket_region or self.s3_client.meta.region_name
+            return f"https://{self.bucket_name}.s3.{region}.amazonaws.com/{encoded_key}"
 
         elif self.bucket_type == BlobType.GOOGLE_CLOUD_STORAGE:
-            return f"https://storage.cloud.google.com/{self.bucket_name}/{key}"
+            return f"https://storage.cloud.google.com/{self.bucket_name}/{encoded_key}"
```
justin-tahara, line 301: I worry that it might be different for GCS

justin-tahara (commented):
This should be scoped for S3 only

PR #5743, opened 2025-10-16
feat: user info personalization

evan-onyx:
## Description

Adds the ability for users to give onyx info about themselves and an initial purely user-directed memory system (mostly preparation for the larger memory effort)

Screenshot 2025-10-16 at 11 29 39 AM

## How Has This Been Tested?

tested in UI + new unit/integration tests

## Additional Options

---

## Summary by cubic
Adds user personalization (name, role) and a simple user-directed memory system that’s injected into prompts to personalize answers. Includes UI, API, and DB support; works for anonymous mode too.

- New Features
  - Backend: new User fields (personal_name, personal_role, use_memories) and Memory table; memories are fetched via a callback and added to system prompts.
  - API: PATCH /user/personalization to update name, role, use_memories, and memories; supports no-auth via KV store.
  - Prompts: handle_memories adds user info and memories to system messages (standard and v2) and DR clarifier.
  - UI: new “Personalization” tab in User Settings to edit name/role, toggle “Use memories,” and manage memory entries.
  - Tests: integration test for personalization round-trip and unit tests for prompt builders.

- Migration
  - Run Alembic migrations to add personalization fields and the memory table.
  - Existing users default to use_memories=true; no backfill required.

### DIFF
--- a/web/src/app/chat/components/modal/UserSettings.tsx
+++ b/web/src/app/chat/components/modal/UserSettings.tsx
@@ -20,58 +20,23 @@ import { Loader2, Monitor, Moon, Sun } from "lucide-react";
 import { useTheme } from "next-themes";
 import Button from "@/refresh-components/buttons/Button";
 import { Input } from "@/components/ui/input";
-import { Textarea } from "@/components/ui/textarea";
 import { deleteAllChatSessions } from "@/app/chat/services/lib";
 import { SourceIcon } from "@/components/SourceIcon";
-import { UserPersonalization, ValidSources } from "@/lib/types";
+import { ValidSources } from "@/lib/types";
 import { getSourceMetadata } from "@/lib/sources";
 import SvgTrash from "@/icons/trash";
 import SvgExternalLink from "@/icons/external-link";
 import { useFederatedOAuthStatus } from "@/lib/hooks/useFederatedOAuthStatus";
 import { useCCPairs } from "@/lib/hooks/useCCPairs";
 import { useLLMProviders } from "@/lib/hooks/useLLMProviders";
+import { useUserPersonalization } from "@/lib/hooks/useUserPersonalization";
+import { AutoResizeTextarea } from "@/components/ui/auto-resize-textarea";
 
 type SettingsSection =
   | "settings"
   | "password"
   | "connectors"
   | "personalization";
-
-interface AutoResizeTextareaProps {
-  value: string;
-  onChange: (value: string) => void;
-  placeholder?: string;
-}
-
-function AutoResizeTextarea({
-  value,
-  onChange,
-  placeholder,
-}: AutoResizeTextareaProps) {
-  const textareaRef = useRef<HTMLTextAreaElement | null>(null);
-
-  useEffect(() => {
-    if (textareaRef.current) {
-      textareaRef.current.style.height = "auto";
-      textareaRef.current.style.height = `${textareaRef.current.scrollHeight}px`;
-    }
-  }, [value]);
-
-  return (
-    <Textarea
-      ref={textareaRef}
-      value={value}
-      placeholder={placeholder}
-      className="min-h-[3rem] resize-none"
-      onChange={(event) => {
-        onChange(event.target.value);
-        const target = event.currentTarget;
-        target.style.height = "auto";
-        target.style.height = `${target.scrollHeight}px`;
-      }}
-    />
-  );
-}
 
 interface UserSettingsProps {
   onClose: () => void;
@@ -106,14 +71,6 @@ export function UserSettings({ onClose }: UserSettingsProps) {
   const [isDeleteAllLoading, setIsDeleteAllLoading] = useState(false);
   const [showDeleteConfirmation, setShowDeleteConfirmation] = useState(false);
   const [isDisconnecting, setIsDisconnecting] = useState<number | null>(null);
-  const [personalizationValues, setPersonalizationValues] =
-    useState<UserPersonalization>({
-      name: "",
-      role: "",
-      memories: [],
-      use_memories: true,
-    });
-
   const { popup, setPopup } = usePopup();
 
   // Fetch federated-connector info so the modal can list/refresh them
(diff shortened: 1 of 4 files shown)


backend/alembic/versions/6f4f86aef280_add_queries_and_is_web_fetch_to_.py
```diff
@@ -1,7 +1,7 @@
 """add queries and is web fetch to iteration answer
 
 Revision ID: 6f4f86aef280
-Revises: 96a5702df6aa
+Revises: 03d710ccf29c
```
Weves, line 4: huh

Weves: nvm, please ignore

backend/onyx/agents/agent_search/dr/nodes/dr_a0_clarification.py
```diff
@@ -491,7 +493,16 @@ def clarifier(
             + PROJECT_INSTRUCTIONS_SEPARATOR
             + graph_config.inputs.project_instructions
         )
+    user = (
+        graph_config.tooling.search_tool.user
+        if graph_config.tooling.search_tool
```
justin-tahara, line 498: Is it true that if `graph_config.tooling.search_tool` is available that the `user` field would be there? Do we need to also check that the user is also available?

backend/onyx/auth/noauth_user.py
```diff
@@ -33,6 +41,15 @@ def load_no_auth_user_preferences(store: KeyValueStore) -> UserPreferences:
 def fetch_no_auth_user(
     store: KeyValueStore, *, anonymous_user_enabled: bool | None = None
 ) -> UserInfo:
+    personalization = UserPersonalization()
+    try:
+        personalization_data = cast(
+            Mapping[str, Any], store.load(KV_NO_AUTH_USER_PERSONALIZATION_KEY)
+        )
+        personalization = UserPersonalization(**personalization_data)
+    except KvKeyNotFoundError:
```
justin-tahara, line 50: Curious why we put this in a try catch instead of checking the store before hand and see if anything exists. 

It just seems weird to me that we have a try catch just to pass in the case that the key value is not found.

backend/onyx/chat/memories.py
```diff
@@ -0,0 +1,29 @@
...
+) -> Callable[[], list[str]]:
+    def memories_callback() -> list[str]:
+        if user is None:
+            return []
+
+        user_info = [
+            f"User's name: {user.personal_name or ''}",
+            f"User's role: {user.personal_role or ''}",
+            f"User's email: {user.email}",
+        ]
```
justin-tahara, line 21: Should we do a bit better where if the user.<item> item is missing we do a different string or just not add this string at all instead of an empty string

backend/onyx/db/models.py
```diff
@@ -183,6 +183,10 @@ class User(SQLAlchemyBaseUserTableUUID, Base):
     )
     auto_scroll: Mapped[bool | None] = mapped_column(Boolean, default=None)
     shortcut_enabled: Mapped[bool] = mapped_column(Boolean, default=False)
+    # personalization fields are exposed via the chat user settings "Personalization" tab
+    personal_name: Mapped[str | None] = mapped_column(String, nullable=True)
+    personal_role: Mapped[str | None] = mapped_column(String, nullable=True)
+    use_memories: Mapped[bool] = mapped_column(Boolean, nullable=False, default=True)
```
justin-tahara, line 189: Should this be default true? I would check if this is default true for other systems like Claude and Chat

backend/onyx/server/manage/models.py
```diff
@@ -67,6 +67,13 @@ class UserPreferences(BaseModel):
     assistant_specific_configs: UserSpecificAssistantPreferences | None = None
 
 
+class UserPersonalization(BaseModel):
+    name: str = ""
+    role: str = ""
+    use_memories: bool = True
```
justin-tahara, line 73: Another place to check for default

backend/onyx/agents/agent_search/dr/nodes/dr_a0_clarification.py
evan-onyx: we don't, this is also a hack for now since this is mostly intended to be used with the new framework

backend/onyx/auth/noauth_user.py
evan-onyx: It's the pattern we use all over the codebase, just tryna be consistent. It's also one less call

backend/onyx/chat/memories.py
evan-onyx: oh yeah good point I meant to do this but forgot oops

backend/onyx/db/models.py
evan-onyx: If they do default false I think they're trolling

evan-onyx: the main thing that matters is that users have input when a new memory is created so they can cultivate useful ones. Only power users should have a reason to turn this off, and then they'll go looking for it

backend/onyx/server/manage/models.py
evan-onyx: same as above

web/src/components/ui/textarea.tsx
```diff
@@ -10,26 +10,7 @@ const Textarea = React.forwardRef<HTMLTextAreaElement, TextareaProps>(
     return (
       <textarea
         className={cn(
-          [
```
Weves, line 13: what's going on here?

web/src/app/chat/components/modal/UserSettings.tsx
```diff
@@ -20,17 +20,58 @@ import { Loader2, Monitor, Moon, Sun } from "lucide-react";
...
+  | "connectors"
+  | "personalization";
+
+interface AutoResizeTextareaProps {
+  value: string;
+  onChange: (value: string) => void;
+  placeholder?: string;
+}
+
+function AutoResizeTextarea({
```
Weves: This should probably be a standardized component under `comonents/ui`

web/src/app/chat/components/modal/UserSettings.tsx
```diff
@@ -64,6 +106,13 @@ export function UserSettings({ onClose }: UserSettingsProps) {
   const [isDeleteAllLoading, setIsDeleteAllLoading] = useState(false);
   const [showDeleteConfirmation, setShowDeleteConfirmation] = useState(false);
   const [isDisconnecting, setIsDisconnecting] = useState<number | null>(null);
+  const [personalizationValues, setPersonalizationValues] =
```
Weves: prefer having a hook called `useUserPersonalization` which holds all the useEffect / `handleSavePersonalization` logic

Weves (changes requested):

web/src/components/ui/textarea.tsx
evan-onyx: now I understand why frontend PRs have so many lines

Weves (approved):

PR #5770, opened 2025-10-17
fix(bedrock): Make Region Selectable

justin-tahara:
## Description

Making our AWS Bedrock configurable with a region dropdown rather than having the user input the region name to avoid errors.

## How Has This Been Tested?

Validated that model retrieval still worked and we not have a cleaner UI. 

Screenshot 2025-10-17 at 2 37 10 PM

## Additional Options

---

## Summary by cubic
Make AWS Bedrock region selectable via a dropdown in the admin LLM settings. This adds a new “select” config type and auto-populates valid regions (via boto3 with a safe fallback) to prevent typos and setup errors.

- **New Features**
  - Backend: Introduced select config type and options model; generated Bedrock region options from boto3 (bedrock + bedrock-runtime) with a conservative fallback list; applied to AWS_REGION_NAME.
  - Frontend: Added select field rendering in the LLM provider form; supports default value, optional reset, and option descriptions; updated types to include select.

### DIFF
--- a/backend/onyx/llm/llm_provider_options.py
+++ b/backend/onyx/llm/llm_provider_options.py
@@ -17,6 +17,15 @@ class CustomConfigKeyType(Enum):
     # i.e., file based credentials (e.g., "/path/to/credentials/file.json")
     FILE_INPUT = "file_input"
 
+    # used for configuration values that require a selection from predefined options
+    SELECT = "select"
+
+
+class CustomConfigOption(BaseModel):
+    label: str
+    value: str
+    description: str | None = None
+
 
 class CustomConfigKey(BaseModel):
     name: str
@@ -26,6 +35,7 @@ class CustomConfigKey(BaseModel):
     is_secret: bool = False
     key_type: CustomConfigKeyType = CustomConfigKeyType.TEXT_INPUT
     default_value: str | None = None
+    options: list[CustomConfigOption] | None = None
 
 
 class WellKnownLLMProviderDescriptor(BaseModel):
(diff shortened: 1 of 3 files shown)


backend/onyx/llm/llm_provider_options.py
```diff
@@ -88,6 +98,42 @@ class WellKnownLLMProviderDescriptor(BaseModel):
...
+        "ap-east-1",
+        "ca-central-1",
+        "eu-central-1",
+        "eu-west-2",
+    ]
+
+
+def _build_bedrock_region_options() -> list[CustomConfigOption]:
+    try:
+        import boto3
```
nmgarza5, line 121: ive seen this in a few places but is it best practice to add import statements like this vs the top of the file?

backend/onyx/llm/llm_provider_options.py
```diff
@@ -88,6 +98,42 @@ class WellKnownLLMProviderDescriptor(BaseModel):
...
+
+        session = boto3.session.Session()
+        regions = set(session.get_available_regions("bedrock"))
+        regions.update(session.get_available_regions("bedrock-runtime"))
+        if not regions:
+            raise ValueError("No Bedrock regions returned from boto3")
+        sorted_regions = sorted(regions)
+    except Exception:
+        sorted_regions = _fallback_bedrock_regions()
```
nmgarza5, line 130: nit: add a log statement if the boto3 call fails

nmgarza5 (approved):

backend/onyx/llm/llm_provider_options.py
justin-tahara: It's just to avoid issues when it comes to dependencies

backend/onyx/llm/llm_provider_options.py
justin-tahara: We don't currently have a logger in here and this might make the PR larger so going to do in a followup

PR #5772, opened 2025-10-18
feat: openrouter support

wenxi-onyx:
## Description

## How Has This Been Tested?

## Additional Options

### DIFF
--- a/backend/onyx/llm/utils.py
+++ b/backend/onyx/llm/utils.py
@@ -667,17 +667,14 @@ def model_supports_image_input(model_name: str, model_provider: str) -> bool:
         )
 
     # Fallback to looking up the model in the litellm model_cost dict
-    model_map = get_model_map()
     try:
-        model_obj = find_model_obj(
-            model_map,
-            model_provider,
-            model_name,
-        )
+        model_obj = find_model_obj(get_model_map(), model_provider, model_name)
         if not model_obj:
-            raise RuntimeError(
-                f"No litellm entry found for {model_provider}/{model_name}"
+            logger.warning(
+                f"No litellm entry found for {model_provider}/{model_name}, "
+                "this model may or may not support image input."
             )
+            return False
         return model_obj.get("supports_vision", False)
     except Exception:
         logger.exception(


backend/onyx/llm/llm_provider_options.py
```diff
@@ -223,15 +156,41 @@ def _get_provider_to_models_map() -> dict[str, list[str]]:
...
         VERTEXAI_PROVIDER_NAME: VERTEXAI_MODEL_NAMES,
         OLLAMA_PROVIDER_NAME: [],
+        OPENROUTER_PROVIDER_NAME: [],
     }
 
 
+def get_bedrock_model_names() -> list[str]:
+    import litellm
+
+    litellm._turn_on_debug()
```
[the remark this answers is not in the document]
nmgarza5: agree with greptile. remove this

backend/onyx/server/manage/llm/api.py
```diff
@@ -577,3 +579,96 @@ def get_ollama_available_models(
...
+) -> list[OpenRouterFinalModelResponse]:
+    """Fetch available models from OpenRouter `/models` endpoint.
+
+    Parses id, context_length, and architecture.input_modalities to infer vision support.
+    """
+    if not request.api_base or not request.api_key:
+        raise HTTPException(
+            status_code=400,
+            detail="API base and API key are required to fetch OpenRouter models.",
+        )
```
nmgarza5: lets separate out missing api base and key errors. more explicit error messages == less time debugging

backend/onyx/server/manage/llm/api.py
```diff
@@ -577,3 +579,96 @@ def get_ollama_available_models(
...
+            if not model_id or not isinstance(model_id, str):
+                continue
+
+            # Filter out embedding models
+            if "embed" in model_id.lower():
+                continue
+
+            context_len = item.get("context_length")
+            if not isinstance(context_len, int):
+                context_len = GEN_AI_MODEL_FALLBACK_MAX_TOKENS
```
nmgarza5: what else could this be if not an int? a str? can we just convert it instead of using the fallback?

backend/onyx/server/manage/llm/api.py
```diff
@@ -577,3 +579,96 @@ def get_ollama_available_models(
...
+
+            architecture = item.get("architecture") or {}
+            input_modalities = architecture.get("input_modalities") or []
```
nmgarza5:
```suggestion
            architecture = item.get("architecture", {})
            input_modalities = architecture.get("input_modalities", [])
```

backend/onyx/server/manage/llm/api.py
```diff
@@ -577,3 +579,96 @@ def get_ollama_available_models(
...
+            supports_image_input = False
+            try:
+                supports_image_input = isinstance(input_modalities, list) and any(
+                    str(m).lower() == "image" for m in input_modalities
+                )
+            except Exception:
+                supports_image_input = False
```
nmgarza5:
```suggestion
            supports_image_input = isinstance(input_modalities, list) and any(
                str(m).lower() == "image" for m in input_modalities
            )
```

web/src/app/admin/configuration/llm/LLMProviderUpdateForm.tsx
```diff
@@ -623,6 +678,14 @@ export function LLMProviderUpdateForm({
                   />
                   {showAdvancedOptions && (
                     <>
+                      {llmProviderDescriptor.name === "openrouter" && (
+                        <div className="w-full">
+                          <ModelConfigurationField
+                            name="model_configurations"
+                            formikProps={formikProps as any}
```
nmgarza5: why are we casting to any? can we use the actual type

```
import type { FormikProps } from "formik";
```

backend/onyx/llm/llm_provider_options.py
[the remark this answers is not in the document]
wenxi-onyx: I really thought I deleted this before pushing 😅

backend/onyx/server/manage/llm/api.py
wenxi-onyx: These checks are actually redundant since my pydantic model requires both of these. Note, not sure if the model validation error will show exactly what is missing. Additionally, the request model should be generic with provider name as a field (in the future when this is refactored). Deleting these checks and exceptions for this PR

backend/onyx/server/manage/llm/api.py
wenxi-onyx: For some reason I thought this could be None. Per open router docs, I think its always set

backend/onyx/server/manage/llm/api.py
wenxi-onyx: ChatGPT codex 😓. Removed this in favor of pydantic model_validate()

backend/onyx/server/manage/llm/api.py
wenxi-onyx: Moved the model details to a separate pydantic model (similar to Ollama)

web/src/app/admin/configuration/llm/LLMProviderUpdateForm.tsx
wenxi-onyx: Consistency with @raunakab and also (according to Claude code) need to define a custom interface with all the fields. I think it's probably still better than any so I'm replacing both Ronnie and my any.

PR #5795, opened 2025-10-20
fix(ui): Set as Default for LLM

justin-tahara:
## Description

The `Set as default` button for the LLM configuration in the admin panel is not working and not clickable. 

This PR aims to restore the functionality that was previously there.

## How Has This Been Tested?

Tested this locally. 

https://github.com/user-attachments/assets/b702ff75-5ea4-438d-976e-a6aa95dcefa5

## Additional Options

---

## Summary by cubic
Restored the “Set as default” action for LLM providers in the admin panel. It’s now clickable and keyboard-accessible, with clear feedback and an automatic refresh.

- **Bug Fixes**
  - Added a handler to POST the default action, show success/error popups, and mutate the provider list.
  - Enabled Text to forward DOM attributes (role, tabIndex, events) and updated its types to support interactive usage.
  - Added focus-visible styles and Enter/Space key handling for accessibility.

### DIFF
--- a/web/src/app/admin/configuration/llm/ConfiguredLLMProviderDisplay.tsx
+++ b/web/src/app/admin/configuration/llm/ConfiguredLLMProviderDisplay.tsx
@@ -3,7 +3,7 @@ import { LLMProviderView, WellKnownLLMProviderDescriptor } from "./interfaces";
 import { Modal } from "@/components/Modal";
 import { LLMProviderUpdateForm } from "./LLMProviderUpdateForm";
 import { CustomLLMProviderUpdateForm } from "./CustomLLMProviderUpdateForm";
-import { useCallback, useState } from "react";
+import { useState } from "react";
 import { LLM_PROVIDERS_ADMIN_URL } from "./constants";
 import { mutate } from "swr";
 import { Badge } from "@/components/ui/badge";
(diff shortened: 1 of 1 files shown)


web/src/app/admin/configuration/llm/ConfiguredLLMProviderDisplay.tsx
```diff
@@ -87,28 +110,22 @@ function LLMProviderDisplay({
...
+              tabIndex={0}
+              className={cn(
+                "text-action-link-05",
+                "cursor-pointer",
+                "focus-visible:underline",
+                "focus-visible:text-action-link-06"
+              )}
+              onClick={(): void => {
+                void handleSetAsDefault();
+              }}
```
raunakab: Small nit: could we just do `onClick={handleSetAsDefault}` instead?

web/src/app/admin/configuration/llm/ConfiguredLLMProviderDisplay.tsx
```diff
@@ -87,28 +110,22 @@ function LLMProviderDisplay({
...
+              onKeyDown={(event): void => {
+                if (event.key === "Enter" || event.key === " ") {
+                  event.preventDefault();
+                  void handleSetAsDefault();
                 }
```
raunakab: Maybe I'm missing some context here, but is the idea here to have `Enter + Space` be the keybind for setting an LLM as a default?

raunakab: If so, how is one specific LLM provider selected? Does it have to be focused?

justin-tahara: removed

PR #5819, opened 2025-10-21
feat: llm access controls

nmgarza5:
## Description

This PR implements fine-grained role-based access control (RBAC) for LLM providers at the persona (assistant) level, extending the existing user group-based restrictions. This allows administrators to restrict specific LLM providers to only be used with specific assistants, providing better cost control and governance.

### What's New

Previously, LLM providers could only be restricted to user groups (controlling **who** can use them). This PR adds **persona-level restrictions** (controlling **which assistants** can use them), enabling administrators to:

- Restrict expensive/specialized models to specific assistants
- Enforce model choice policies per assistant
- Prevent users from selecting inappropriate models for certain use cases
- Layer multiple access controls (user groups AND persona restrictions)

### Database Schema Changes

**New Table:** `llm_provider__persona`
- Many-to-many association table linking LLM providers to personas
- If no rows exist for a provider → accessible by all personas (backward compatible)
- If rows exist → only those personas can use the provider
- Includes composite index for performance

**Updated Models:**
- `LLMProvider.personas`: relationship to restricted personas
- `Persona.allowed_by_llm_providers`: inverse relationship

**Migration:** `a4f23d6b71c8_add_llm_provider_persona_restrictions.py`

### Core RBAC Logic (`backend/onyx/db/llm.py`)

New function `can_user_access_llm_provider()` implements access control:

**Access Rules:**
1. `is_public=True` → Everyone has access (public override)
2. `is_public=False`:
   - Both groups AND personas set → Must satisfy BOTH (admins bypass groups only)
   - Only groups set → Must be in one of the groups (admins bypasa
   - Only personas set → Must use one of the personas (applies to admins too)
   - Neither set → Admin-only access

**Key New Functions:**
- `update_llm_provider_persona_relationships__no_commit()`: Manage persona associations
- `fetch_persona_with_groups()`: Eagerly load persona with groups for access checks
- `validate_persona_ids_exist()`: Validate persona IDs before creating restrictions
- `get_personas_using_provider()`: Find personas affected by provider changes

### API Changes (`backend/onyx/server/manage/llm/api.py`)

**New Endpoints:**
- `GET /llm/persona/{persona_id}/providers`: List LLM providers available for a specific persona

**Updated Endpoints:**
- `PUT /admin/llm/provider`: Now accepts `personas` field (list of persona IDs)
- `GET /llm/provider`: Enhanced to filter providers by user's group memberships
- Validates persona IDs exist and deduplicates before saving
- API keys are now masked in responses (show first 4 + last 4 chars)

### Frontend Changes

**LLM Provider Configuration UI** (`web/src/app/admin/configuration/llm/LLMProviderUpdateForm.tsx`):

Added new "Access Controls" section with:
- **User Group Selector** (existing, now visually grouped)
- **Assistant Whitelist** (NEW) - Multi-select for choosing which assistants can use this provider
- Automatically disabled when provider is public with helpful message
- Real-time validation of persona IDs

───────────────────────────────────────────────`

**LLM Provider Selection:**
- Chat UI now filters available providers based on current persona
- Assistant editor shows only accessible providers when editing
- Provider list API respects persona restrictions

### Backward Compatibility

✅ Existing providers without persona restrictions work unchanged
✅ Empty `personas` array = accessible by all personas (default behavior)
✅ UI defaults to no restrictions (maintains existing UX)
✅ Database migration only creates new table (no data migration needed)
✅ Indexed for query performance

### Example Usage

**Scenario:** Restrict GPT-4 to only "Code Review Assistant" and "Architecture Advisor"

1. Admin navigates to LLM Provider configuration
2. Edits GPT-4 provider:
   - Sets `is_public: false`
   - Adds "Engineering Team" to user groups
   - Adds "Code Review Assistant" and "Architecture Advisor" to Assistant Whitelist
3. Save

**Result:** Only users in "Engineering Team" can use GPT-4, and only when using those two specific assistants.

### Files Changed

**Backend:**
- Core RBAC logic in `db/llm.py`
- API endpoints in `server/manage/llm/api.py`
- Database migration and model updates
- Comprehensive test coverage

**Frontend:**
- 3 new reusable components (will change with the ui refresh so they arent perfect)
- Enhanced LLM provider configuration form
- Updated provider selection throughout chat UI

---

### Backend Integration Tests

1. **`test_llm_provider.py`** (169 lines)
   - Provider creation with persona restrictions
   - Validation of persona IDs during provider creation
   - Edge cases (empty restrictions, invalid persona IDs)
   - Database relationship integrity

2. **`test_llm_provider_access_control.py`** (353 lines)
   - Public vs private provider access
   - User group-based restrictions
   - Persona-based restrictions
   - Combined group + persona restrictions (AND logic)
   - Admin bypass behavior for groups but not personas
   - Anonymous user access patterns

3. **`test_llm_provider_persona_access.py`** (247 lines)
   - Persona-provider association creation and updates
   - User access validation through different personas
   - Cascade deletion behavior (persona/provider deletion)
   - Provider visibility in different contexts

### Frontend E2E Tests

**`web/tests/e2e/assistants/llm_provider_rbac.spec.ts`** (153 lines)
- Creates test fixtures: restricted user group and restricted LLM provider
- Creates random user (not in restricted group)
- Navigates to assistant creation page
- Verifies restricted provider doesn't appear in LLM selector dropdown
- Tests complete admin-to-end-user flow

---

## Additional Options

### DIFF
--- a/backend/onyx/llm/factory.py
+++ b/backend/onyx/llm/factory.py
@@ -90,7 +90,7 @@ def get_llm_config_for_persona(
         llm_provider = fetch_default_provider(db_session)
         if not llm_provider:
             raise ValueError("No default LLM provider found")
-        model_name = llm_provider.default_model_name
+        model_name: str | None = llm_provider.default_model_name
     else:
         llm_provider = fetch_llm_provider_view(db_session, provider_name)
         if not llm_provider:
--- a/web/next.config.js
+++ b/web/next.config.js
@@ -60,6 +60,17 @@ const nextConfig = {
           },
         ],
       },
+      {
+        // Cache static assets (images, icons, fonts, etc.) to prevent refetching and re-renders
+        // This helps eliminate icon flickering and improves performance
+        source: "/_next/static/:path*", // Matches static assets served by Next.js
+        headers: [
+          {
+            key: "Cache-Control",
+            value: "public, max-age=31536000, immutable", // Cache for 1 year, mark as immutable
+          },
+        ],
+      },
     ];
   },
   async rewrites() {
--- a/web/src/app/admin/assistants/AssistantEditor.tsx
+++ b/web/src/app/admin/assistants/AssistantEditor.tsx
@@ -90,7 +90,7 @@ import CoreModal from "@/refresh-components/modals/CoreModal";
 import UserFilesModalContent from "@/components/modals/UserFilesModalContent";
 import { TagIcon, UserIcon, FileIcon, InfoIcon, BookIcon } from "lucide-react";
 import { LLMSelector } from "@/components/llm/LLMSelector";
-import useSWR from "swr";
+import useSWR, { mutate } from "swr";
 import { errorHandlingFetcher } from "@/lib/fetcher";
 import { ConfirmEntityModal } from "@/components/modals/ConfirmEntityModal";
 import {
@@ -207,6 +207,7 @@ export function AssistantEditor({
     [llmProviders.length]
   );
   const isUpdate = existingPersona !== undefined && existingPersona !== null;
+
   const defaultProvider = llmProviders.find(
     (llmProvider) => llmProvider.is_default_provider
   );
@@ -783,6 +784,14 @@ export function AssistantEditor({
             }
 
             await refreshAgents();
+
+            // Force refetch LLM provider cache for this agent
+            // This ensures the chat page shows the updated provider list
+            await mutate(
+              `/api/llm/persona/${assistantId}/providers`,
+              undefined,
+              { revalidate: true }
+            );
 
             router.push(
               isAdminPage
@@ -1366,7 +1375,7 @@ export function AssistantEditor({
                                 <BooleanFormField
                                   name={`enabled_tools_map.${imageGenerationTool.id}`}
                                   label={imageGenerationTool.display_name}
-                                  subtext="Generate and manipulate images using AI-powered tools"
+                                  subtext="Generate and manipulate images using AI-powered tools."
                                   disabled={!currentLLMSupportsImageOutput}
                                   disabledTooltip={
                                     !currentLLMSupportsImageOutput
@@ -1461,12 +1470,19 @@ export function AssistantEditor({
                   <LLMSelector
                     llmProviders={llmProviders}
                     currentLlm={
-                      values.llm_model_version_override
-                        ? structureValue(
-                            values.llm_model_provider_override,
-                            "",
-                            values.llm_model_version_override
-                          )
+                      values.llm_model_version_override &&
+                      values.llm_model_provider_override
+                        ? (() => {
+                            const provider = llmProviders.find(
+                              (p) =>
+                                p.name === values.llm_model_provider_override
+                            );
+                            return structureValue(
+                              values.llm_model_provider_override,
+                              provider?.provider || "",
+                              values.llm_model_version_override
+                            );
+                          })()
                         : null
                     }
                     requiresImageGeneration={
@@ -1479,7 +1495,7 @@ export function AssistantEditor({
                         setFieldValue("llm_model_version_override", null);
                         setFieldValue("llm_model_provider_override", null);
                       } else {
-                        const { modelName, provider, name } =
+                        const { modelName, name } =
                           parseLlmDescriptor(selected);
                         if (modelName && name) {
                           setFieldValue(
--- a/web/src/app/chat/components/ChatPage.tsx
+++ b/web/src/app/chat/components/ChatPage.tsx
@@ -111,14 +111,8 @@ export function ChatPage({
   const router = useRouter();
   const searchParams = useSearchParams();
 
-  const {
-    chatSessions,
-    ccPairs,
-    tags,
-    documentSets,
-    llmProviders,
-    refreshChatSessions,
-  } = useChatContext();
+  const { chatSessions, ccPairs, tags, documentSets, refreshChatSessions } =
+    useChatContext();
 
   const {
     currentMessageFiles,
@@ -211,18 +205,21 @@ export function ChatPage({
     llmDescriptors,
   } = useOnboardingState();
 
-  // On first render, open onboarding if there are no configured LLM providers.
-  // Intentionally exclude llmProviders from deps to avoid changing the state
-  // when data refreshes.
-  useEffect(() => {
-    setShowOnboarding(llmProviders.length === 0);
-  }, []);
+  const llmManager = useLlmManager(selectedChatSession, liveAssistant);
 
-  const llmManager = useLlmManager(
-    llmProviders,
-    selectedChatSession,
-    liveAssistant
-  );
+  // Track if we've done the initial onboarding check
+  const hasCheckedOnboarding = useRef(false);
+
+  // On first render, open onboarding if there are no configured LLM providers.
+  // Only check once to avoid re-triggering onboarding when data refreshes.
+  useEffect(() => {
+    if (!hasCheckedOnboarding.current) {
+      setShowOnboarding(
+        !llmManager.llmProviders || llmManager.llmProviders.length === 0
+      );
+      hasCheckedOnboarding.current = true;
+    }
+  }, [llmManager.llmProviders]);
 
   const noAssistants = liveAssistant === null || liveAssistant === undefined;
 
@@ -930,8 +927,8 @@ export function ChatPage({
                           textAreaRef={textAreaRef}
                           setPresentingDocument={setPresentingDocument}
                           disabled={
-                            llmProviders.length === 0 ||
-                            (llmProviders.length === 0 &&
+                            llmManager.llmProviders?.length === 0 ||
+                            (llmManager.llmProviders?.length === 0 &&
                               !user?.personalization?.name) ||
                             onboardingState.currentStep !==
                               OnboardingStep.Complete
--- a/web/src/app/chat/components/input/ChatInputBar.tsx
+++ b/web/src/app/chat/components/input/ChatInputBar.tsx
@@ -131,7 +131,6 @@ function ChatInputBarInner({
   disabled,
 }: ChatInputBarProps) {
   const { user } = useUser();
-
   const { forcedToolIds, setForcedToolIds } = useAgentsContext();
   const { currentMessageFiles, setCurrentMessageFiles } = useProjectsContext();
 
@@ -626,7 +625,6 @@ function ChatInputBarInner({
     </div>
   );
 }
-
 const ChatInputBar = React.memo(ChatInputBarInner);
 ChatInputBar.displayName = "ChatInputBar";
 
--- a/web/src/components/AgentsMultiSelect.tsx
+++ b/web/src/components/AgentsMultiSelect.tsx
@@ -1,0 +1,49 @@
+import { FormikProps } from "formik";
+import { GenericMultiSelect } from "@/components/GenericMultiSelect";
+
+export type AgentsMultiSelectFormType = {
+  personas: number[];
+};
+
+interface Agent {
+  id: number;
+  name: string;
+  description: string;
+}
+
+interface AgentsMultiSelectProps<T extends AgentsMultiSelectFormType> {
+  formikProps: FormikProps<T>;
+  agents: Agent[] | undefined;
+  isLoading?: boolean;
+  error?: any;
+  label?: string;
+  subtext?: string;
+  disabled?: boolean;
+  disabledMessage?: string;
+}
+
+export function AgentsMultiSelect<T extends AgentsMultiSelectFormType>({
+  formikProps,
+  agents,
+  isLoading = false,
+  error,
+  label = "Agents",
+  subtext = "",
+  disabled = false,
+  disabledMessage,
+}: AgentsMultiSelectProps<T>) {
+  return (
+    <GenericMultiSelect
+      formikProps={formikProps}
+      fieldName="personas"
+      label={label}
+      subtext={subtext}
+      items={agents}
+      isLoading={isLoading}
+      error={error}
+      emptyMessage="No agents available. Please create an agent first from the Agents page."
+      disabled={disabled}
+      disabledMessage={disabledMessage}
+    />
+  );
+}
--- a/web/src/components/GenericMultiSelect.tsx
+++ b/web/src/components/GenericMultiSelect.tsx
@@ -1,0 +1,146 @@
+import { FormikProps, ErrorMessage } from "formik";
+import Text from "@/refresh-components/texts/Text";
+import Button from "@/refresh-components/buttons/Button";
+import { SearchMultiSelectDropdown } from "@/components/Dropdown";
+import { cn } from "@/lib/utils";
+import SvgX from "@/icons/x";
+
+export type GenericMultiSelectFormType<T extends string> = {
+  [K in T]: number[];
+};
+
+interface GenericItem {
+  id: number;
+  name: string;
+}
+
+interface GenericMultiSelectProps<
+  T extends string,
+  F extends GenericMultiSelectFormType<T>,
+> {
+  formikProps: FormikProps<F>;
+  fieldName: T;
+  label: string;
+  subtext?: string;
+  items: GenericItem[] | undefined;
+  isLoading: boolean;
+  error: any;
+  emptyMessage: string;
+  disabled?: boolean;
+  disabledMessage?: string;
+}
+
+export function GenericMultiSelect<
+  T extends string,
+  F extends GenericMultiSelectFormType<T>,
+>({
+  formikProps,
+  fieldName,
+  label,
+  subtext,
+  items,
+  isLoading,
+  error,
+  emptyMessage,
+  disabled = false,
+  disabledMessage,
+}: GenericMultiSelectProps<T, F>) {
+  if (isLoading) {
+    return (
+      <div className="flex flex-col gap-2 w-full">
+        <Text mainUiAction>{label}</Text>
+        <div className="animate-pulse bg-background-neutral-02 h-10 w-full rounded-08" />
+      </div>
+    );
+  }
+
+  if (error) {
+    return (
+      <div className="flex flex-col gap-2 w-full">
+        <Text mainUiAction>{label}</Text>
+        <Text text03 className="text-action-danger-05">
+          Failed to load {label.toLowerCase()}. Please try again.
+        </Text>
+      </div>
+    );
+  }
+
+  if (!items || items.length === 0) {
+    return (
+      <div className="flex flex-col gap-2 w-full">
+        <Text mainUiAction>{label}</Text>
+        <Text text03>{emptyMessage}</Text>
+      </div>
+    );
+  }
+
+  const selectedIds = (formikProps.values[fieldName] as number[]) || [];
+  const selectedItems = items.filter((item) => selectedIds.includes(item.id));
+
+  const handleSelect = (option: { name: string; value: string | number }) => {
+    if (disabled) return;
+    const currentIds = (formikProps.values[fieldName] as number[]) || [];
+    const numValue =
+      typeof option.value === "string"
+        ? parseInt(option.value, 10)
+        : option.value;
+    if (!currentIds.includes(numValue)) {
+      formikProps.setFieldValue(fieldName, [...currentIds, numValue]);
+    }
+  };
+
+  const handleRemove = (itemId: number) => {
+    if (disabled) return;
+    const currentIds = (formikProps.values[fieldName] as number[]) || [];
+    formikProps.setFieldValue(
+      fieldName,
+      currentIds.filter((id) => id !== itemId)
+    );
+  };
+
+  return (
+    <div className="flex flex-col gap-2 w-full">
+      <Text mainUiAction>{label}</Text>
+
+      {subtext && <Text text03>{disabled ? disabledMessage : subtext}</Text>}
+
+      <div className={cn(disabled && "opacity-50 pointer-events-none")}>
+        <SearchMultiSelectDropdown
+          options={items
+            .filter((item) => !selectedIds.includes(item.id))
+            .map((item) => ({
+              name: item.name,
+              value: item.id,
+            }))}
+          onSelect={handleSelect}
+        />
+      </div>
+
+      {selectedItems.length > 0 && (
+        <div className="flex flex-wrap gap-2">
+          {selectedItems.map((item) => (
+            <Button
+              key={item.id}
+              main
+              secondary
+              disabled={disabled}
+              rightIcon={SvgX}
+              onClick={() => handleRemove(item.id)}
+              className="!px-2 !py-1"
+            >
+              {item.name}
+            </Button>
+          ))}
+        </div>
+      )}
+
+      <ErrorMessage name={fieldName} component="div">
+        {(msg) => (
+          <Text text03 className="text-action-danger-05">
+            {msg}
+          </Text>
+        )}
+      </ErrorMessage>
+    </div>
+  );
+}
--- a/web/src/lib/assistants/fetchPersonaEditorInfoSS.ts
+++ b/web/src/lib/assistants/fetchPersonaEditorInfoSS.ts
@@ -22,10 +22,17 @@ export async function fetchAssistantEditorInfoSS(
     ]
   | [null, string]
 > {
+  // When editing an existing persona, fetch only the providers available to that persona
+  // When creating a new persona, fetch all providers
+  const llmProvidersUrl =
+    personaId !== undefined
+      ? `/llm/persona/${personaId}/providers`
+      : "/llm/provider";
+
   const tasks = [
     fetchSS("/manage/connector-status"),
     fetchSS("/manage/document-set"),
-    fetchSS("/llm/provider"),
+    fetchSS(llmProvidersUrl),
     // duplicate fetch, but shouldn't be too big of a deal
     // this page is not a high traffic page
     getCurrentUserSS(),
--- a/web/src/lib/hooks.ts
+++ b/web/src/lib/hooks.ts
@@ -41,6 +41,8 @@ import { SEARCH_TOOL_ID } from "@/app/chat/components/tools/constants";
 import { updateTemperatureOverrideForChatSession } from "@/app/chat/services/lib";
 import { usePathname, useSearchParams } from "next/navigation";
 import { SEARCH_PARAM_NAMES } from "@/app/chat/services/searchParams";
+import { useLLMProviders } from "./hooks/useLLMProviders";
+import { useChatContext } from "@/refresh-components/contexts/ChatContext";
 
 export function useIsMounted() {
   const [mounted, setMounted] = useState(false);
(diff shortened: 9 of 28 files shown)


nmgarza5:
@greptile what do you think?

backend/onyx/db/llm.py
```diff
@@ -46,6 +50,99 @@ def update_group_llm_provider_relationships__no_commit(
...
+                    select(User__UserGroup.user_group_id).where(
+                        User__UserGroup.user_id == user.id
+                    )
+                ).all()
+            )
+        )
+        if resolved_user_group_ids & allowed_group_ids:
+            return True
+
+    persona_id = persona.id if isinstance(persona, Persona) else None
```
Weves: nit: `persona.id if persona else None` is probably better

backend/onyx/db/llm.py
```diff
@@ -273,6 +413,11 @@ def remove_llm_provider(db_session: Session, provider_id: int) -> None:
             LLMProvider__UserGroup.llm_provider_id == provider_id
         )
     )
+    db_session.execute(
```
Weves: prefer to have a on delete CASCADE handle this. I know it's a bit inconsistent, but I think generally that approach is better

backend/onyx/llm/factory.py
```diff
@@ -55,6 +58,7 @@ def get_main_llm_from_tuple(
 
 def get_llms_for_persona(
     persona: Persona | PersonaOverrideConfig | None,
+    user: User | None = None,
```
Weves: should we make this required? That way we don't accidentally not pass it in?

backend/onyx/server/manage/llm/api.py
```diff
@@ -133,7 +150,7 @@ def test_llm_configuration(
         client_error_msg = litellm_exception_to_error_msg(
             error, llm, fallback_to_error_msg=True
         )
-        raise HTTPException(status_code=400, detail=client_error_msg)
+        raise HTTPException(status_code=400, detail={"message": client_error_msg})
```
Weves: why is it better to have a dict here?

backend/onyx/server/manage/llm/api.py
```diff
@@ -210,14 +226,42 @@ def put_llm_provider(
...
-            detail=f"LLM Provider with name {llm_provider_upsert_request.name} does not exist",
+            detail={
+                "message": f"LLM Provider with name {llm_provider_upsert_request.name} does not exist"
+            },
         )
 
+    persona_ids = llm_provider_upsert_request.personas
+    if persona_ids:
+        fetched_persona_ids = set(
+            db_session.scalars(
```
Weves: prefer having no DB operations outside of the `db/` folder if possible

backend/onyx/server/manage/llm/api.py
```diff
@@ -269,9 +313,62 @@ def delete_llm_provider(
...
     db_session: Session = Depends(get_session),
 ) -> None:
+    provider = db_session.get(LLMProviderModel, provider_id)
+    if not provider:
+        raise HTTPException(
+            status_code=404, detail={"message": "LLM Provider not found"}
+        )
+
+    in_use_personas = list(
+        db_session.scalars(
```
Weves: same comment here

backend/onyx/server/manage/llm/api.py
```diff
@@ -269,9 +313,62 @@ def delete_llm_provider(
...
+    db_session: Session = Depends(get_session),
+) -> LLMProviderUsageResponse:
+    provider = db_session.get(LLMProviderModel, provider_id)
+    if not provider:
+        raise HTTPException(
+            status_code=404, detail={"message": "LLM Provider not found"}
+        )
+
+    personas = list(
+        db_session.scalars(
```
Weves: same

backend/onyx/server/manage/llm/api.py
```diff
@@ -336,14 +433,51 @@ def get_vision_capable_providers(
...
     user: User | None = Depends(current_chat_accessible_user),
     db_session: Session = Depends(get_session),
 ) -> list[LLMProviderDescriptor]:
     start_time = datetime.now(timezone.utc)
     logger.debug("Starting to fetch basic LLM providers for user")
 
+    # If persona_id provided, fetch persona and filter by persona access
+    persona: Persona | None = None
+    if persona_id is not None:
+        persona = db_session.scalar(
```
Weves: same comment here

backend/onyx/server/manage/llm/api.py
```diff
@@ -336,14 +433,51 @@ def get_vision_capable_providers(
...
+        if not persona:
+            raise HTTPException(
+                status_code=404, detail={"message": "Persona not found"}
+            )
+
+    # Get user's group IDs once
+    user_group_ids: set[int] | None = None
+    if user:
+        user_group_ids = set(
+            db_session.scalars(
```
Weves: same here

backend/onyx/server/manage/llm/api.py
```diff
@@ -336,14 +433,51 @@ def get_vision_capable_providers(
 
 @basic_router.get("/provider")
 def list_llm_provider_basics(
+    persona_id: int | None = Query(
```
Weves: should this be bundled as part of this API? First impression is that it should be a separate API

backend/onyx/server/manage/llm/api.py
```diff
@@ -404,6 +538,95 @@ def get_provider_contextual_cost(
     return costs
 
 
+@admin_router.get("/persona/{persona_id}/available-providers")
+def get_available_providers_for_persona(
+    persona_id: int,
+    _: User | None = Depends(current_admin_user),
+    db_session: Session = Depends(get_session),
+) -> list[LLMProviderView]:
+    persona = db_session.scalar(
```
Weves: same here

Weves: going to stop adding this comment everywhere

web/src/lib/hooks/useLLMProviders.ts
```diff
@@ -2,9 +2,13 @@ import useSWR from "swr";
 import { LLMProviderDescriptor } from "@/app/admin/configuration/llm/interfaces";
 import { errorHandlingFetcher } from "@/lib/fetcher";
 
-export function useLLMProviders() {
+export function useLLMProviders(personaId?: number) {
+  const url = personaId
```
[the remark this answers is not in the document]
Weves: +1

web/src/lib/assistants/fetchPersonaEditorInfoSS.ts
```diff
@@ -22,10 +22,16 @@ export async function fetchAssistantEditorInfoSS(
     ]
   | [null, string]
 > {
+  // When editing an existing persona, fetch only the providers available to that persona
+  // When creating a new persona, fetch all providers
+  const llmProvidersUrl = personaId
```
Weves: from cubic:

```
Treating personaId as a truthy value means persona_id=0 (the default persona) skips the persona-specific endpoint, returning unrestricted providers and bypassing persona-level RBAC.
```

web/src/components/PersonasMultiSelect.tsx
```diff
@@ -0,0 +1,45 @@
+import { FormikProps } from "formik";
+import useSWR from "swr";
+import { errorHandlingFetcher } from "@/lib/fetcher";
+import { GenericMultiSelect } from "@/components/GenericMultiSelect";
+
+export type PersonasMultiSelectFormType = {
```
Weves: let's go with Agents everywhere as a push for consistency (need to make sure we re-name everything in the future 😓 )

web/src/refresh-components/LLMPopover.tsx
```diff
@@ -121,32 +124,42 @@ export default function LLMPopover({
...
-                    setOpen(false);
-                  }}
-                >
-                  {getDisplayNameForModel(modelName)}
-                </LineItem>
-              );
-            }
-          )}
+          {(() => {
+            const renderedModels: string[] = [];
```
Weves: what are these needed for?

web/src/refresh-components/modals/CoreModal.tsx
```diff
@@ -34,7 +34,7 @@ export default function CoreModal({
 
   const modalContent = (
     <div
-      className="fixed inset-0 z-[2000] flex items-center justify-center bg-mask-03 backdrop-blur-xl"
+      className="fixed inset-0 z-[2000] flex items-center justify-center bg-mask-03 backdrop-blur-[1px]"
```
Weves: is this intended?

web/tests/e2e/assistants/llm_provider_rbac.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+  const llmOptions = await getLLMProviderOptions(page);
+
+  // Verify that we have some options (at least the default provider)
+  expect(llmOptions.length).toBeGreaterThan(0);
+
+  // Check that no restricted providers appear
+  // Note: In a real test, you'd need to set up a restricted provider first
+  // and verify its name doesn't appear in the options.
+  // For now, we just verify the selector is working
+  const hasDefaultOption = llmOptions.some(
```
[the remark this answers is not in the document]
Weves: seems valid

backend/onyx/server/manage/llm/api.py
```diff
@@ -404,6 +538,95 @@ def get_provider_contextual_cost(
...
+    descriptors: list[LLMProviderDescriptor] = []
+    for provider in fetch_existing_llm_providers(db_session):
+        if not llm_provider_is_public(provider):
+            continue
+        descriptors.append(LLMProviderDescriptor.from_model(provider))
+    return descriptors
+
+
+@admin_router.post("/persona/{persona_id}/grant-provider-access")
+def grant_persona_access_to_provider(
```
Weves: are these APIs used (this + `unrestricted-providers`)?

web/src/refresh-components/modals/CoreModal.tsx
nmgarza5: oops no...was testing something for another pr. will remove

backend/onyx/server/manage/llm/api.py
nmgarza5: ah this was a a claude thing. will revert back.

backend/onyx/chat/user_files/parse_user_files.py
```diff
@@ -49,6 +49,11 @@ def parse_user_files(
     if not user_file_ids and not project_id:
         return [], [], None
 
+    if user is None:
```
Weves: this will blow things up for folks with `AUTH_TYPE=disabled`

backend/onyx/server/manage/llm/api.py
```diff
@@ -269,6 +298,18 @@ def delete_llm_provider(
     _: User | None = Depends(current_admin_user),
     db_session: Session = Depends(get_session),
 ) -> None:
+    provider = db_session.get(LLMProviderModel, provider_id)
```
Weves: same here — prefer not having any DB operations outside of `db/` for consistency

backend/onyx/db/llm.py
```diff
@@ -46,6 +52,150 @@ def update_group_llm_provider_relationships__no_commit(
...
+    Providers with no group or persona restrictions are treated as public.
+    If any restrictions are set (groups or personas), the provider is restricted
+    regardless of the is_public flag.
+    """
+    return not provider.groups and not provider.personas
+
+
+def can_user_access_llm_provider(
+    *,
+    db_session: Session,
```
Weves: prefer requiring the user group ids to be passed in. Currently the `db_session` is unused if you pass in `user_group_ids` / the behavior of the function is reasonably different if `user_group_ids` is passed in or not.

If we want two versions, prefer making them separate functions

backend/onyx/db/llm.py
```diff
@@ -178,30 +347,54 @@ def fetch_existing_llm_providers_for_user(
     db_session: Session,
     user: User | None = None,
 ) -> list[LLMProviderModel]:
-    # if user is anonymous
-    if not user:
-        # Only fetch public providers if auth is turned on
-        return fetch_existing_llm_providers(
-            db_session, only_public=AUTH_TYPE != AuthType.DISABLED
-        )
+    """Fetch LLM providers accessible to the user using database-level filtering.
```
Weves: I don't think is actually used? When should we use this vs fetching all and doing post-filtering. It seems like this might be better since it enforces that the access checks are done vs the other approach you could easily forget?

backend/onyx/db/models.py
```diff
@@ -3055,6 +3067,19 @@ class Persona__UserGroup(Base):
     )
 
 
+class LLMProvider__Persona(Base):
+    """Association table restricting LLM providers to specific personas."""
```
Weves: nit: maybe add a comment here that if no such rows exist for a LLM then it is accessible by all personas.

backend/onyx/llm/factory.py
```diff
@@ -84,10 +88,30 @@ def get_llms_for_persona(
...
     with get_session_with_current_tenant() as db_session:
-        llm_provider = fetch_llm_provider_view(db_session, provider_name)
-
-    if not llm_provider:
-        raise ValueError("No LLM provider found")
+        provider_model = fetch_existing_llm_provider(provider_name, db_session)
+        if not provider_model:
+            raise ValueError("No LLM provider found")
+
+        persona_model = persona if isinstance(persona, Persona) else None
```
Weves, line 151: Can we add a comment as to why we don't need to do this check for `PersonaOverrideConfig`?

backend/onyx/server/query_and_chat/chat_backend.py
```diff
@@ -568,9 +569,16 @@ def get_max_document_tokens(
     except ValueError:
         raise HTTPException(status_code=404, detail="Persona not found")
 
+    # For anonymous users, we need a User object to compute max tokens
```
Weves: hmm, this seems not ideal. Why do we need to do this?

backend/onyx/server/query_and_chat/chat_backend.py
```diff
@@ -586,10 +594,16 @@ def get_available_context_tokens_for_session(
     db_session: Session = Depends(get_session),
 ) -> AvailableContextTokensResponse:
     """Return available context tokens for a chat session based on its persona."""
+    if user is None:
```
Weves: this will break `AUTH_TYPE=disabled`?

backend/onyx/server/query_and_chat/chat_backend.py
```diff
@@ -632,9 +647,15 @@ class ChatSeedResponse(BaseModel):
 def seed_chat(
     chat_seed_request: ChatSeedRequest,
     # NOTE: realistically, this will be an API key not an actual user
-    _: User | None = Depends(current_user),
+    user: User | None = Depends(current_user),
     db_session: Session = Depends(get_session),
 ) -> ChatSeedResponse:
+    if user is None:
```
Weves: same here

web/src/app/admin/assistants/AssistantEditor.tsx
```diff
@@ -1362,7 +1362,7 @@ export function AssistantEditor({
                                 <BooleanFormField
                                   name={`enabled_tools_map.${imageGenerationTool.id}`}
                                   label={imageGenerationTool.display_name}
-                                  subtext="Generate and manipulate images using AI-powered tools"
+                                  subtext="Generate and manipulate images using AI-powered tools. If selected, only models that support image generation will be displayed."
```
Weves: I don't think this is true? All models are supported even if you have image gen enabled

web/src/app/chat/components/ChatPage.tsx
```diff
@@ -211,8 +212,22 @@ export function ChatPage({
...
     useState<MinimalOnyxDocument | null>(null);
 
+  // Fetch persona-filtered providers for the llmManager
+  const {
+    llmProviders: personaFilteredProviders,
+    isLoading: isLoadingPersonaProviders,
+  } = useLLMProviders(liveAssistant?.id);
+
+  // Use persona-filtered providers when loaded and not empty, otherwise use context providers
+  const llmManagerProviders =
```
Weves: this will cause a potential flicker, right? E.g. one LLM shows up and then it's removed after the second request completes? Ideally we'd fetch this at the same time as the `llmProviders` I think?

Weves: Screenshot 2025-10-24 at 11 05 20 AM

Also I get this despite having an LLM provider set up now 👀

web/src/components/AgentsMultiSelect.tsx
```diff
@@ -0,0 +1,45 @@
...
+interface AgentsMultiSelectProps<T extends AgentsMultiSelectFormType> {
+  formikProps: FormikProps<T>;
+  label?: string;
+  subtext?: string;
+}
+
+export function AgentsMultiSelect<T extends AgentsMultiSelectFormType>({
+  formikProps,
+  label = "Agents",
+  subtext = "Select which agents can use this LLM provider. If none selected, all agents can use it.",
```
Weves: this would make the component a bit non-generic, right? Prefer to just pass this in have and have no subtext by default

web/src/components/AgentsMultiSelect.tsx
```diff
@@ -0,0 +1,45 @@
...
+}: AgentsMultiSelectProps<T>) {
+  const {
+    data: agents,
+    isLoading,
+    error,
+  } = useSWR<Agent[]>("/api/persona", errorHandlingFetcher);
```
Weves: this approach causes a flicker, right? Prefer enforcing it to be pre-fetched and passed in

web/src/components/GenericMultiSelect.tsx
```diff
@@ -0,0 +1,137 @@
...
+  fieldName: T;
+  label: string;
+  subtext?: string;
+  items: GenericItem[] | undefined;
+  isLoading: boolean;
+  error: any;
+  emptyMessage: string;
+}
+
+export function GenericMultiSelect<
```
Weves, line 33: @raunakab can you take a look at this? Want to make sure new components we add are following our general best practices

backend/onyx/chat/user_files/parse_user_files.py
nmgarza5: This was a concern of mine as well but I've been running locally with auth disabled and it works 👍

backend/onyx/server/manage/llm/api.py
nmgarza5: ahhhh why is this still here...my bad..though i got all of them. will fix

backend/onyx/db/llm.py
nmgarza5: yeah makes sense

backend/onyx/server/query_and_chat/chat_backend.py
nmgarza5: was being overly cautious. no need. its handled downstream appropriately. but this does mean that restricted llm providers will not be accessible by anonymous users. which makes sense actually

web/src/app/admin/assistants/AssistantEditor.tsx
nmgarza5: good catch... i could have swore during testing when selecting lighter models, img generation was failing. so i added some conditional behavior. going to remove it

web/src/components/PersonasMultiSelect.tsx
nmgarza5: I'll respect this for the component but for all the backend stuff, I'm going to keep using persona. Would prefer to make the switch in 1 go vs having both persona/agent present 😅

web/src/components/GenericMultiSelect.tsx
nmgarza5: These components will likely change with the ui refresh. I dont see menu item mocks yet so if design is an issue can we just plan on updating as part of ui refresh? 🙏

web/src/refresh-components/buttons/Chip.tsx
raunakab, line 1: @nmgarza5, were these based off of the chips available in Figma? Usually, those chips use varying background colours, not opacities, to convey abled / disabled state. I would assume that it carries over to this `Chip` component as well. I may be wrong, though.

web/src/app/chat/components/input/ChatInputBar.tsx
```diff
@@ -11,6 +11,7 @@ import LLMPopover from "@/refresh-components/popovers/LLMPopover";
 import { InputPrompt } from "@/app/chat/interfaces";
 import { FilterManager, LlmManager, useFederatedConnectors } from "@/lib/hooks";
 import { useChatContext } from "@/refresh-components/contexts/ChatContext";
+import { LLMProviderDescriptor } from "@/app/admin/configuration/llm/interfaces";
```
raunakab: Remove.

web/src/app/chat/components/input/ChatInputBar.tsx
```diff
@@ -605,7 +605,6 @@ function ChatInputBarInner({
               <LLMPopover
                 llmManager={llmManager}
                 requiresImageGeneration={false}
-                disabled={disabled}
```
raunakab, line 608: What's the purpose of removing this?

web/src/refresh-components/onboarding/useOnboardingState.ts
```diff
@@ -49,7 +49,7 @@ export function useOnboardingState(): {
 
   // If there are any configured LLM providers already present, skip to the final step
   useEffect(() => {
-    if (Array.isArray(llmProviders) && llmProviders.length > 0) {
+    if (llmProviders && llmProviders.length > 0) {
```
raunakab: Will `llmProviders?.length > 0` not work here?

raunakab: Similarly for the other diffs in this file.

web/src/refresh-components/buttons/Chip.tsx
nmgarza5: im just going to remove and use a regular button

web/src/app/chat/components/input/ChatInputBar.tsx
nmgarza5: adding back

backend/alembic/versions/a4f23d6b71c8_add_llm_provider_persona_restrictions.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+    op.create_index(
+        "ix_llm_provider__persona_composite",
+        "llm_provider__persona",
+        ["persona_id", "llm_provider_id"],
+    )
+
+
+def downgrade() -> None:
+
```
Weves: nit: extra newline

backend/onyx/chat/user_files/parse_user_files.py
```diff
@@ -99,6 +98,7 @@ def parse_user_files(
     available_tokens = compute_max_document_tokens_for_persona(
         persona=persona,
         db_session=db_session,
+        user=None,  # We only have user_id, not the full user object
```
Weves: nit: should this function take in `user`? It requires the persona object already to call, so we should have failed earlier if the user doesn't have access

Weves (approved):

backend/onyx/chat/user_files/parse_user_files.py
nmgarza5: user is only passed for a subsequent function call to fetch the models `get_llms_for_persona(persona=persona, user=user)`. I'll create another helper to use instead so we dont need to pass user.

web/src/refresh-components/popovers/LLMPopover.tsx
```diff
@@ -143,32 +135,36 @@ export default function LLMPopover({
...
-                    llmManager.updateCurrentLlm({
-                      modelName,
-                      provider,
-                      name,
-                    } as LlmDescriptor);
-                    onSelect?.(structureValue(name, provider, modelName));
-                    setOpen(false);
-                  }}
+                  key="loading"
+                  icon={() => <LoadingSpinner size="small" />}
```
raunakab: Would prefer to use `SimpleLoader` here!

web/src/refresh-components/popovers/LLMPopover.tsx
raunakab, line 1: Yo, awesome usage of the existing components! Keeps the UI consistent and up to par. Huge shoutout.

web/src/refresh-components/onboarding/useOnboardingState.ts
```diff
@@ -49,7 +49,7 @@ export function useOnboardingState(): {
 
   // If there are any configured LLM providers already present, skip to the final step
   useEffect(() => {
-    if (Array.isArray(llmProviders) && llmProviders.length > 0) {
+    if (llmProviders) {
```
raunakab: Ooh, I think this is a bug. Try running the following in a `node` shell:

```ts
> if ([]) console.log(0)
```

It prints out `0`, which implies that `[]` is truthy. I believe the logic you want is that `llmProviders?.length > 1`, which is technically different than what you have here.

web/src/lib/hooks.ts
```diff
@@ -576,12 +580,27 @@ providing appropriate defaults for new conversations based on the available tool
 */
 
 export function useLlmManager(
```
raunakab, line 582: Not a nit for now, but could it be possible that we turn all "important" state (such as this `LlmManager` object) into a React Context one day? Would be very slick, I'm sure.

web/src/components/GenericMultiSelect.tsx
raunakab: I'm fine for merging in substitutes for Opal components which are not yet implemented / mocked up yet (i.e., multi-selects).

@nmgarza5, would probably ask you, at a future date, to implement Opal's very own `MultiSelect` component!

web/src/refresh-components/popovers/LLMPopover.tsx
```diff
@@ -143,32 +134,33 @@ export default function LLMPopover({
...
-                  }}
-                >
-                  {getDisplayNameForModel(modelName)}
-                </LineItem>
-              );
-            }
-          )}
+          {isLoadingProviders
+            ? [
+                <LineItem key="loading" icon={() => <SimpleLoader />}>
```
raunakab: Super small nit, but you could just pass in `SimpleLoader` here!

raunakab: I.e.,

```tsx
<LineItem icon={SimpleLoader} {...} />
```

The reason being that `(SimpleLoader) === (() => <SimpleLoader />)`.

web/next.config.js
raunakab, line 1: Could you explain to me what this part does? I'm not the best with config files, lol.

web/src/lib/hooks.ts
nmgarza5: i considered it but didnt want to do it as a part of this PR. good callout though!

web/src/refresh-components/popovers/LLMPopover.tsx
nmgarza5: oops yeah can do

PR #5823, opened 2025-10-21
fix: decrease background blur in modals

nmgarza5:
## Description

Fix for: https://linear.app/danswer/issue/DAN-2711/sidebar-disappears-when-opening-up-agents-tab

Decreases modal blur so the user can still see what the background is. This is to prevent confusion surfaced by the linear ticket.

Screenshot 2025-10-21 at 11 03 05 AM

## How Has This Been Tested?

manual testing

## Additional Options


raunakab:
Force-merging in since everything looks safe.

PR #5831, opened 2025-10-21
fix: always trigger DR

evan-onyx:
## Description

Fixes https://linear.app/danswer/issue/DAN-2902/deep-research-is-broken
when the user selects deep research, it should always be triggered

## How Has This Been Tested?

tested in UI + integration test for DR

## Additional Options

---

## Summary by cubic
Ensure Deep Research always runs by preventing LLM-only shortcuts, so selecting Deep Research will either invoke a tool or ask a clarification, even for simple prompts.

- **Bug Fixes**
  - Block LLM-only path when research_type is DEEP in clarification and streaming logic.
  - Update test response analyzer to handle message_start without final_documents (clarification case).
  - Add integration test to assert DR triggers a tool or asks clarification on simple prompts.

- **Refactors**
  - Simplify tenant cleanup signal handler by removing subprocess termination tracking.

### DIFF
--- a/backend/tests/integration/tests/streaming_endpoints/test_chat_stream.py
+++ b/backend/tests/integration/tests/streaming_endpoints/test_chat_stream.py
@@ -1,4 +1,3 @@
-from onyx.db.chat import ResearchAnswerPurpose
 from tests.integration.common_utils.managers.chat import ChatSessionManager
 from tests.integration.common_utils.managers.llm_provider import LLMProviderManager
 from tests.integration.common_utils.test_models import DATestUser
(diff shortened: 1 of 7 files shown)


backend/tests/integration/tests/streaming_endpoints/test_chat_stream.py
```diff
@@ -4,6 +4,32 @@
...
+    test_chat_session = ChatSessionManager.create(user_performing_action=admin_user)
+
+    response = ChatSessionManager.send_message(
+        chat_session_id=test_chat_session.id,
+        message="Hello",
+        user_performing_action=admin_user,
+        use_agentic_search=True,
+    )
+
+    # Success if either:
```
Weves: nit: add `TODO` that we should improve this

evan-onyx: Instead, I just improved it rn

PR #5836, opened 2025-10-21
feat: add DB models for OAuthTools

Weves:
## Description

DB models + migration to support OAuth Tools.

## How Has This Been Tested?

👀 

## Additional Options

---

## Summary by cubic
Add DB support for OAuth-enabled tools. Tools can now link to a shared OAuth provider config, and users can store encrypted OAuth tokens.

- **New Features**
  - OAuthConfig table for provider endpoints and client credentials, reusable across tools.
  - OAuthUserToken table for per-user encrypted tokens; unique per (user, config); cascades on delete.
  - Tool model now has oauth_config_id and relationship; null for non-OAuth tools.

- **Migration**
  - Creates oauth_config and oauth_user_token tables and adds oauth_config_id to tool (FK SET NULL).
  - Run alembic upgrade head. Ensure encryption keys are set for EncryptedString/EncryptedJson.

### DIFF
--- a/backend/onyx/db/models.py
+++ b/backend/onyx/db/models.py
@@ -248,6 +248,11 @@ class User(SQLAlchemyBaseUserTableUUID, Base):
         cascade="all, delete-orphan",
         lazy="selectin",
     )
+    oauth_user_tokens: Mapped[list["OAuthUserToken"]] = relationship(
+        "OAuthUserToken",
+        back_populates="user",
+        cascade="all, delete-orphan",
+    )
 
     @validates("email")
     def validate_email(self, key: str, value: str) -> str:


backend/onyx/db/models.py
```diff
@@ -2524,6 +2531,92 @@ class Tool(Base):
...
+        server_default=func.now(),
+        onupdate=func.now(),
+        nullable=False,
+    )
+
+    # Relationships
+    oauth_config: Mapped["OAuthConfig"] = relationship(
+        "OAuthConfig", back_populates="user_tokens"
+    )
+    user: Mapped["User"] = relationship("User")
```
evan-onyx, line 2617: do we want a relationship from User for conveniently accessing a user's oauth tokens?

Weves: Seems good, added!

PR #5842, opened 2025-10-21
chore(debug): packet stream utility

rguan72:
## Description

A dev utility to debug what is rendered for a given packet stream
I think we should be able to make this better 🤔 

## How Has This Been Tested?

Tested locally

## Additional Options


rguan72:
I think it's better to have a utility to mock packet responses from the back end

PR #5844, opened 2025-10-21
feat: oauth tools/fe

Weves:
## Description

## How Has This Been Tested?

## Additional Options

---

## Summary by cubic
Adds shared OAuth support for tools with admin-managed provider configs, per-user encrypted tokens with automatic refresh, and frontend to attach OAuth to a tool and prompt users to authenticate.

- **New Features**
  - Database: oauth_config and oauth_user_token tables; tools can reference an OAuth config.
  - Backend APIs: admin CRUD for OAuth configs; user endpoints to initiate OAuth, handle callback, check status, and revoke tokens.
  - Token management: OAuthTokenManager for code exchange, refresh, expiry checks, and auth URL building.
  - Tool integration: create/update APIs accept oauth_config_id; tool snapshots include config info.
  - Frontend: OAuthConfigSelector in ActionEditor, OAuthPromptModal, and oauth API client; interfaces updated.
  - Tests: unit suites for OAuth config/token CRUD and token manager behavior.

- **Migration**
  - Run Alembic migration to create new tables.
  - Ensure WEB_DOMAIN is set for correct OAuth redirect.
  - Create OAuth configs in admin, then link tools via oauth_config_id.
  - Passthrough auth is disabled when an OAuth config is selected for a tool.

### DIFF
--- a/web/tests/e2e/admin/oauth_config/test_tool_oauth.spec.ts
+++ b/web/tests/e2e/admin/oauth_config/test_tool_oauth.spec.ts
@@ -4,7 +4,6 @@ import { loginAs } from "../../utils/auth";
 
 // --- Locator Helper Functions ---
 const getNameInput = (page: Page) => page.locator('input[name="name"]');
-const getProviderInput = (page: Page) => page.locator('input[name="provider"]');
 const getAuthorizationUrlInput = (page: Page) =>
   page.locator('input[name="authorization_url"]');
 const getTokenUrlInput = (page: Page) =>
@@ -103,9 +102,9 @@ test("Tool OAuth Configuration: Creation, Selection, and Assistant Integration",
   // Verify OAuth Config Selector is visible
   await expect(page.getByText("OAuth Configuration:")).toBeVisible();
 
-  // Click "Create New OAuth Config" button
+  // Click "New OAuth Configuration" button
   const createNewOAuthButton = page.getByRole("button", {
-    name: "Create New OAuth Config",
+    name: "New OAuth Configuration",
   });
   await createNewOAuthButton.click();
 
@@ -114,7 +113,6 @@ test("Tool OAuth Configuration: Creation, Selection, and Assistant Integration",
 
   // Fill in OAuth config details
   await getNameInput(page).fill(configName);
-  await getProviderInput(page).fill(provider);
   await getAuthorizationUrlInput(page).fill(authorizationUrl);
   await getTokenUrlInput(page).fill(tokenUrl);
   await getClientIdInput(page).fill(clientId);


web/src/app/admin/oauth-configs/OAuthConfigForm.tsx
```diff
@@ -0,0 +1,251 @@
...
+  const isUpdate = config !== undefined;
+
+  return (
+    <CoreModal onClickOutside={onClose} className="w-[60%] max-h-[80vh]">
+      <div className="overflow-y-auto p-6">
+        <h2 className="text-xl font-bold flex mb-4">
+          {isUpdate
+            ? "Update OAuth Configuration"
+            : "Create OAuth Configuration"}
+        </h2>
```
raunakab: Could we use the `@/refresh-components/texts/Text` component here? Should have all the permutations of texts available to it as flags!

```tsx

<Text headingH2>{isUpdate ? "..." : "..."}</Text>
```

web/src/components/oauth/OAuthConfigSelector.tsx
```diff
@@ -0,0 +1,106 @@
...
+        }}
+      />
+      <Button
+        onClick={() => setShowModal(true)}
+        type="button"
+        secondary
+        leftIcon={SvgPlusCircle}
+      >
+        New OAuth Configuration
+      </Button>
```
raunakab: Super minor nit, but there is a `CreateButton` in `refresh-components/buttons` which I created due to the reason of the use-case for a "create" button being so common.

raunakab: The API should be very similar to what you have here, except that you don't have to specify the `leftIcon` and `secondary` (they're default provided internally).

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -177,6 +185,36 @@ function ActionItem({
...
+                          ? "OAuth token expired - click to re-authenticate"
+                          : "Click to authenticate"
+                      }
+                    />
+                  ) : (
+                    <FiCheck
+                      size={16}
+                      className="text-green-500"
+                      title="Authenticated"
+                    />
```
raunakab: Could we use the `check.tsx` icon instead here? Also, we should use the new colour-scheme that @duo-onyx laid out. All the available colours that we can choose from are in `colors.css`!

```tsx
import SvgCheck from "@/src/icons/check";

// Our green colour is called `status-text-success-05`.
// We can specify the stroke-colour of the checkmark by specifying `stroke-status-text-success-05`.
<SvgCheck className="stroke-status-text-success-05" />
```

raunakab: Keep in mind that those refreshed components don't have a `size: number` prop on them.

If you want to set their height + width, you'll have to specify it (ideally in REMs) through the Tailwind classes.

The above example should be:

```tsx
<SvgCheck className="stroke-... w-[1rem] h-[1rem]" />
```

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -177,6 +185,36 @@ function ActionItem({
...
+                  {!oauthStatus.has_token || oauthStatus.is_expired ? (
+                    <FiKey
+                      size={16}
+                      className="transition-colors cursor-pointer text-yellow-500 hover:text-yellow-600"
+                      title={
+                        oauthStatus.is_expired
+                          ? "OAuth token expired - click to re-authenticate"
+                          : "Click to authenticate"
+                      }
+                    />
```
raunakab: @duo-onyx's created a new key icon in Figma!

If you want to import it, you can either import it via the `web/src/icons/README.md` instructions (manual and a bit tiresome), OR you could use the Figma MCP tool to import it.

P.s., if you do the latter, you should remove the `stroke="..."` and `strokeOpacity="..."` props from the imported `key.tsx` file. Those props hardcode the colours and opacities into the component, thus preventing the end-developer from setting the colour of the icon themselves.

raunakab: We should ideally move away from using icons provided by `lucide-icons`, `react-icons`, etc. in our codebase, especially as we start bolting down our design language and aesthetic.

web/src/app/admin/actions/ActionEditor.tsx
```diff
@@ -292,32 +328,37 @@ function ActionForm({
...
                           />
                         </div>
                       </TooltipTrigger>
-                      {values.customHeaders.some(
+                      {(values.customHeaders.some(
                         (header) => header.key.toLowerCase() === "authorization"
-                      ) && (
+                      ) ||
+                        values.oauth_config_id !== null) && (
                         <TooltipContent side="top" align="center">
```
raunakab: This would actually be a really good spot to use the `SimpleTooltip` exposed in the `refresh-components` directory.

raunakab: The only hitch here is that `SimpleTooltip` doesn't have a `disabled?: boolean` prop on it.

If you'd rather not modify that component in this PR, then no worries. We can stick with this approach for the time-being and I can update it later.

web/src/app/admin/actions/ActionEditor.tsx
```diff
@@ -292,32 +328,37 @@ function ActionForm({
...
                           <Text className="bg-background-900 max-w-[200px] mb-1 text-sm rounded-lg p-1.5 text-white">
```
raunakab: I know this isn't part of your changes, but could you remove the `className="..."` prop that is passed into `Text`?

We should ideally not pass *ANY* custom class-name stylings into `Text`. It has been very meticulously designed by Duo and I to only express the exact combinations of text fonts, colours, sizes, weights, and line heights as per the mocks.

For example, explicitly setting `className="text-sm"` only sets the font-size and does not modify the line-height. However, @duo-onyx's mocks have specifically set certain font-sizes to have specific line-heights associated with them. Other combinations of such will lead to inconsistent UIs which will need to be cleaned up later... :(

raunakab: Btw, having *spacing* class-names on the `Text` component is permissible (e.g., `"mb-4"`). *Styling* class-names (e.g., `"text-sm"`) are going to be a strict no-go going forward!

web/src/components/chat/OAuthPromptModal.tsx
```diff
@@ -0,0 +1,97 @@
...
+      await initiateOAuthFlow(oauthConfigId, currentPath);
+      // initiateOAuthFlow will redirect to OAuth provider
+    } catch (err: any) {
+      setIsAuthenticating(false);
+      setError(err.message || "Failed to initiate authentication");
+    }
+  };
+
+  const applicationName =
+    settings?.enterpriseSettings?.application_name || "Onyx";
```
raunakab: Hmm interesting...

I've seen this use-case a fair number of times in the codebase (specifically `settings.enterpriseSettings?.application_name || "Onyx"`). Although I don't think this PR should solve that problem right now, this is a problem which can easily be solved by creating a more thought-out context API...

raunakab: Not requesting a solution to this problem right now; I'll work on this issue later in a future PR.

web/src/components/chat/OAuthPromptModal.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  return (
+    <Modal
+      title="Authentication Required"
+      onOutsideClick={onClose}
+      width="w-[500px]"
+      icon={KeyIcon}
```
raunakab: Please replace with the `SvgKey` after you import it! (Import instructions should be in another comment).

web/src/components/chat/OAuthPromptModal.tsx
```diff
@@ -0,0 +1,97 @@
...
+      <div className="space-y-4">
+        <Text>
+          This tool requires authentication with{" "}
+          <span className="font-semibold">
+            {providerName || oauthConfigName}
+          </span>
+          .
+        </Text>
+
+        <Text className="text-sm text-subtle">
```
raunakab: Please remove the custom styling class-names!

raunakab: If you want to modify the text font, you can choose from a wide variety of flags:

```txt
headingH1
headingH2
headingH3
headingH3Muted
mainContentBody
mainContentMuted
mainContentEmphasis
mainContentMono
mainUiBody
mainUiMuted
mainUiAction
mainUiMono
secondaryBody
secondaryAction
secondaryMono
figureSmallLabel
figureSmallValue
figureKeystroke
```

If you want to modify the text colour, you can choose from some more flags as well:

```txt
text05
text04
text03
text02
text01
inverted // This one inverts the text colour
```

raunakab: All the above were designed and signed off by @duo-onyx! Each flag also sets the line-height, font-weight, and some other goodies with it. When you custom override with `"text-sm"`, you only set the font size (but not the font-weight, line-height, etc.). Therefore, we break congruency!

web/src/components/chat/OAuthPromptModal.tsx
```diff
@@ -0,0 +1,97 @@
...
+          To use this tool, you need to authorize {applicationName} to access
+          your account. You&apos;ll be redirected to the provider&apos;s login
+          page to complete the authentication process.
+        </Text>
+
+        {error && (
+          <div className="p-3 bg-red-50 dark:bg-red-900/20 border border-red-200 dark:border-red-800 rounded-lg">
+            <Text className="text-sm text-red-800 dark:text-red-200">
```
raunakab: Same comment as above.

We also definitely should *NOT* specify `dark:` modifiers from here on out! All the colour-scheming is fully done by the `colors.css` file and Tailwind!

raunakab: If you want to use red, please refer to the `colors.css` file to see what reds are available to you.

Please don't use the "Base Colors" (at the top of the file). They're the raw colour definitions. Please instead use the variables defined in either "Brand Colors" or "Light Colors" (the light-mode colours will automatically switch over for you in dark-mode).

web/src/components/oauth/OAuthConfigSelector.tsx
```diff
@@ -0,0 +1,106 @@
...
+        options={options}
+        subtext="Select an OAuth configuration for this tool. Users will be prompted to authenticate when using this tool."
+        onSelect={(selected) => {
+          // SelectorFormField passes the value string directly, not an object
+          let configId: number | null;
+          if (
+            !selected ||
+            selected === "null" ||
+            selected === -1 ||
+            selected === "-1"
```
raunakab, line 113: Do `null`, `-1` and `"-1"` have some sort of semantic meaning? If so, are they documented somewhere? Furthermore, would it be better if we labelled those ones as constants (either at the top of this file or in the `constants.ts` file)?

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -177,6 +185,36 @@ function ActionItem({
               </span>
             </div>
             <div className="flex items-center gap-2">
+              {/* OAuth Authentication Indicator */}
+              {tool?.oauth_config_id && oauthStatus && (
+                <div
+                  className="flex items-center gap-2 transition-opacity duration-200 opacity-0 group-hover:opacity-100"
```
raunakab, line 176: This is not a concern for this PR, but we have a lot of custom animations being specified all over the place (with varying durations, varying easing functions, etc. being used). Creates a bit of a haphazard experience.

I'll look to address this in a future PR. Will probably create a fixed set of animations in the `globals.css` file (e.g., `.slow { .. }`, `.icon { .. }`, etc.) for developers to easily choose from.

raunakab (commented):
Some styistic nits!

web/src/app/admin/oauth-configs/OAuthConfigForm.tsx
Weves: Done!

web/src/components/oauth/OAuthConfigSelector.tsx
Weves: done!

web/src/components/chat/OAuthPromptModal.tsx
Weves: Done

web/src/app/admin/actions/ActionEditor.tsx
Weves: done

web/src/app/chat/components/input/ActionManagement.tsx
Weves: Done!

web/src/app/chat/components/input/ActionManagement.tsx
Weves: Done!

PR #5865, opened 2025-10-22
fix: mcp chat frontend

evan-onyx:
## Description

Unified MCP tools chat UX with other frontend components

## How Has This Been Tested?

Tested in UI

## Additional Options

---

## Summary by cubic
Unified the MCP tools chat UI into the main action popover with server selection, search, and tool management. Removes the floating MCP popup, reduces flicker, and adds bulk enable/disable and inline re-auth.

- **New Features**
  - MCP tools shown as a secondary view inside the action popover with back navigation.
  - Search tools, toggle per-tool, and see descriptions via tooltip.
  - Enable/Disable all tools per server and display enabled count.
  - Inline “Re-Authenticate” row when user auth is required.
  - Consistent scroll shadows and fade mask in secondary views.

- **Refactors**
  - Replaced portal-based MCP popup with SecondaryViewState-driven views.
  - Simplified OAuth flow and loading state management on errors.
  - Consolidated scroll handling to a single container (action-submenu-scroll).
  - MCPServerItem now uses isActive/onSelect and shows disabled state; removed preventCloseRef and outside-click guards.

### DIFF
--- a/web/src/app/chat/components/input/ActionManagement.tsx
+++ b/web/src/app/chat/components/input/ActionManagement.tsx
@@ -41,6 +41,7 @@ import SimpleLoader from "@/refresh-components/loaders/SimpleLoader";
 import SvgSlash from "@/icons/slash";
 import { SvgProps } from "@/icons";
 import SimpleTooltip from "@/refresh-components/SimpleTooltip";
+import LineItem from "@/refresh-components/buttons/LineItem";
 
 // Get source metadata for configured sources - deduplicated by source type
 function getConfiguredSources(
@@ -109,9 +110,9 @@ function ActionItem({
 
   return (
     <SimpleTooltip tooltip={tool?.description}>
-      <Button
+      <LineItem
         data-testid={`tool-option-${toolName}`}
-        active={isForced}
+        // active={isForced}
         onClick={() => {
           // If no connectors, don't allow forcing the tool
           if (isSearchToolWithNoConnectors) return;
@@ -119,9 +120,9 @@ function ActionItem({
           if (onToggle && disabled) onToggle();
           onForceToggle();
         }}
-        disabled={isSearchToolWithNoConnectors || disabled}
-        leftIcon={Icon}
-        rightIcon={() => (
+        // disabled={isSearchToolWithNoConnectors || disabled}
+        icon={Icon}
+        rightChildren={
           <div className="flex items-center">
             {!isSearchToolWithNoConnectors && (
               <IconButton
@@ -148,13 +149,14 @@ function ActionItem({
                 icon={
                   isSearchToolWithNoConnectors ? SvgSettings : SvgChevronRight
                 }
+                internal
               />
             )}
           </div>
-        )}
+        }
       >
         {label}
-      </Button>
+      </LineItem>
     </SimpleTooltip>
   );
 }
@@ -232,14 +234,14 @@ function MCPServerItem({
   const allToolsDisabled = enabledTools.length === 0 && tools.length > 0;
 
   return (
-    <Button
-      className="group flex items-center justify-between mx-1"
-      disabled={allToolsDisabled}
+    <LineItem
+      // className="group flex items-center justify-between mx-1"
+      // disabled={allToolsDisabled}
       onClick={handleClick}
       data-mcp-server-id={server.id}
       data-mcp-server-name={server.name}
-      leftIcon={getServerIcon}
-      rightIcon={({ className }) => (
+      icon={getServerIcon}
+      rightChildren={
         <div className="flex items-center gap-1">
           {showReauthButton && (
             <IconButton
@@ -247,17 +249,14 @@ function MCPServerItem({
               tertiary
               aria-label="Re-authenticate MCP server"
               title="Re-authenticate"
-              onClick={(event) => {
-                event.stopPropagation();
-                onAuthenticate();
-              }}
+              onClick={noProp(onAuthenticate)}
             />
           )}
-          {isAuthenticated && tools.length > 0 && (
+          {/*{isAuthenticated && tools.length > 0 && (
             <SvgChevronRight className={className} />
-          )}
+          )}*/}
         </div>
-      )}
+      }
     >
       <div className="flex items-center gap-2 flex-1 min-w-0">
         <Text
@@ -278,7 +277,7 @@ function MCPServerItem({
             </Text>
           )}
       </div>
-    </Button>
+    </LineItem>
   );
 }
 
@@ -676,7 +675,7 @@ export function ActionToggle({
 
   const mcpFooter = showActiveReauthRow ? (
     <div className="sticky bottom-0 bg-background-neutral-00 border-t border-border z-[1] rounded-b-lg">
-      <Button
+      <LineItem
         role="button"
         tabIndex={0}
         onClick={handleFooterReauthClick}
@@ -686,7 +685,7 @@ export function ActionToggle({
             handleFooterReauthClick();
           }
         }}
-        leftIcon={
+        icon={
           selectedMcpServerData?.isLoading
             ? SimpleLoader
             : ({ className }) => (
@@ -702,12 +701,12 @@ export function ActionToggle({
                 />
               )
         }
-        rightIcon={
-          selectedMcpServerData?.isLoading ? undefined : SvgChevronRight
+        rightChildren={
+          selectedMcpServerData?.isLoading ? undefined : <SvgChevronRight />
         }
       >
         Re-Authenticate
-      </Button>
+      </LineItem>
     </div>
   ) : undefined;
 
--- a/web/src/refresh-components/buttons/LineItem.tsx
+++ b/web/src/refresh-components/buttons/LineItem.tsx
@@ -4,11 +4,13 @@ import React from "react";
 import Text from "@/refresh-components/texts/Text";
 import { cn } from "@/lib/utils";
 import { SvgProps } from "@/icons";
+import Truncated from "@/refresh-components/texts/Truncated";
 
-interface LineItemProps {
+interface LineItemProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
   icon?: React.FunctionComponent<SvgProps>;
   description?: string;
   children?: string | React.ReactNode;
+  rightChildren?: React.ReactNode;
   onClick?: React.MouseEventHandler<HTMLButtonElement>;
 }
 
@@ -16,13 +18,14 @@ export default function LineItem({
   icon: Icon,
   description,
   children,
+  rightChildren,
   onClick,
 }: LineItemProps) {
   return (
     <button
       type="button"
       className={cn(
-        "flex flex-col w-full justify-center items-start p-spacing-interline hover:bg-background-tint-02 rounded-08"
+        "flex flex-col w-full justify-center items-start p-spacing-interline hover:bg-background-tint-02 rounded-08 group/LineItem"
       )}
       onClick={onClick}
     >
@@ -33,12 +36,13 @@ export default function LineItem({
           </div>
         )}
         {typeof children === "string" ? (
-          <Text mainUiMuted text04 className="text-left w-full">
+          <Truncated mainUiMuted text04 className="text-left w-full">
             {children}
-          </Text>
+          </Truncated>
         ) : (
           children
         )}
+        {rightChildren}
       </div>
       {description && (
         <div className="flex flex-row">
@@ -49,7 +53,7 @@ export default function LineItem({
             </>
           )}
 
-          <Text secondaryBody text03 className="">
+          <Text secondaryBody text03>
             {description}
           </Text>
         </div>


web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -335,20 +342,32 @@ function MCPServerItem({
         rounded-lg
         py-2
         mx-1
-        ${isExpanded ? "bg-accent-100 hover:bg-accent-200" : ""}
+        ${isActive ? "bg-accent-100 hover:bg-accent-200" : ""}
+        ${allToolsDisabled ? "opacity-60" : ""}
```
raunakab: Instead of using string-formatting to pass conditional logic to the className, could you please try using the `cn` utility function, please?

```tsx
import { cn } from "@/lib/utils";
```

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -335,20 +342,32 @@ function MCPServerItem({
...
-          </span>
-        )}
+        <span
+          className={`text-sm font-medium select-none truncate max-w-[120px] inline-block align-middle ${
+            allToolsDisabled ? "line-through" : ""
+          }`}
+          title={server.name}
+        >
+          {server.name}
+        </span>
```
raunakab: We should aim to always render text using the refreshed `Text` component!

```tsx
```

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -335,20 +342,32 @@ function MCPServerItem({
...
+        {isAuthenticated &&
+          tools.length > 0 &&
+          enabledTools.length > 0 &&
+          tools.length !== enabledTools.length && (
+            <span className="text-xs whitespace-nowrap ml-1 text-neutral-400 dark:text-neutral-500">
+              <span className="text-blue-500">{enabledTools.length}</span>
+              {` of ${tools.length}`}
+            </span>
```
raunakab: Refreshed `Text` here too, please!

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -367,7 +386,7 @@ function MCPServerItem({
         )}
         {isAuthenticated && tools.length > 0 && (
           <FiChevronRight
-            className={`transition-transform ${isExpanded ? "rotate-90" : ""}`}
+            className={`transition-transform ${isActive ? "rotate-90" : ""}`}
             size={14}
           />
```
raunakab: We're trying to move away from using icons provided by `react-icons/fi` (and other icon libraries)!

There should be a nice `SvgChevronRight` available to you in the codebase already which @duo-onyx created!

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -439,79 +481,255 @@ function MCPToolsList({
...
+              onClick={() => {
+                setSearchTerm("");
+                onBack();
+              }}
+              className="absolute left-1 top-1/2 transform -translate-y-1/2 text-text-400 hover:text-text-300 z-10 w-4 h-4 flex items-center justify-center transition-colors"
+              style={{ borderRadius: "8px" }}
+              aria-label="Back to action menu"
+            >
+              <FiChevronLeft size={16} />
+            </button>
```
raunakab: Let's use the refreshed `IconButton` here instead!

```tsx
import IconButton from "@/refresh-components/buttons/IconButton";
```

raunakab: You should also *not* have to specify *any* custom classNames or styles on the `IconButton`; everything (including light/dark-modes should all be handled for you).

Usage of it should look like this:

```tsx
<IconButton
  icon={SvgChevronLeft}

  // This is a flag which specifies what "variant" of an IconButton this is.
  // There are other variants also available; try playing around with them:
  // - primary
  // - secondary
  // - tertiary
  internal

  aria-label="Back to action menu"

  onClick={...}
/>
```

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -865,6 +1065,32 @@ export function ActionToggle({
     return server.name.toLowerCase().includes(searchLower);
   });
 
+  const activeMcpServerId =
```
Weves: nit: thoughts on `selectedMcpServerId` instead (same for all the below)? :thonk:

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -439,79 +481,255 @@ function MCPToolsList({
...
+                py-2
+                bg-transparent
+                rounded-lg
+                text-sm
+                outline-none
+                text-neutral-700 dark:text-neutral-300
+                placeholder:text-neutral-400 dark:placeholder:text-neutral-500
+              "
+              autoFocus
+            />
```
raunakab: Let's use the `InputTypeIn` component here! You should also not have to specify any custom classNames or stylings here as well; they will all be default provided to you!

```tsx
import InputTypeIn from "@/refresh-components/inputs/InputTypeIn";
```

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -439,79 +481,255 @@ function MCPToolsList({
...
+                py-2
+                bg-transparent
+                rounded-lg
+                text-sm
+                outline-none
+                text-neutral-700 dark:text-neutral-300
+                placeholder:text-neutral-400 dark:placeholder:text-neutral-500
+                cursor-pointer
+              "
+            />
```
raunakab: `InputTypeIn` here as well, please!

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -439,79 +481,255 @@ function MCPToolsList({
...
+                          className="max-w-xs"
+                        >
+                          <Text inverted>{tool.description}</Text>
+                        </TooltipContent>
+                      </Tooltip>
+                    </TooltipProvider>
+                  ) : (
+                    label
+                  )}
+                  <button
```
Weves: there should probably be a component for this?

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -439,79 +481,255 @@ function MCPToolsList({
...
+              alt={allToolsDisabled ? "Enable All Tools" : "Disable All Tools"}
+              className="absolute right-3 top-1/2 transform -translate-y-1/2 w-4 h-4 pointer-events-none"
+              style={{
+                filter:
+                  theme === "dark"
+                    ? "invert(0.6) sepia(1) saturate(0) hue-rotate(0deg) brightness(1.2)"
+                    : "invert(0.6) sepia(1) saturate(0) hue-rotate(0deg) brightness(0.6)",
+                opacity: theme === "dark" ? 0.9 : 0.8,
+              }}
+            />
```
raunakab: Using `img` to render SVGs is usually bad form.

@duo-onyx has created the appropriate plug and unplug SVGs in Figma. You can import them into `web/src/icons` by running through the manual instructions laid out in `web/src/icons/README.md`. OR, you could instead ask the Figma MCP tool to directly import them into the `web/src/icons` directory!

Note that if you opt for the latter, you will need to either tell the AI tool to remove the `stroke="..."` and `strokeOpacity={...}` props, or do it yourself after import. Those props hardcode the stroke colour and opacity, thus making it impossible for the end-user to specify it themselves.

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -439,79 +481,255 @@ function MCPToolsList({
...
+                bg-transparent
+                rounded-lg
+                text-sm
+                outline-none
+                text-neutral-700 dark:text-neutral-300
+                placeholder:text-neutral-400 dark:placeholder:text-neutral-500
+                cursor-pointer
+              "
+            />
+            <img
```
Weves: ideally should be pulled out as a dedicated icon (maybe in Icon.tsx) if possible.

cc @raunakab to advise

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -439,79 +481,255 @@ function MCPToolsList({
     );
   });
 
+  useEffect(() => {
```
Weves: what does this do? 😰

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -439,79 +481,255 @@ function MCPToolsList({
...
+        </div>
+        <div className="border-b border-border mx-1 mt-1" />
+        <div
+          className="mx-1 h-2 -mb-2 transition-opacity ease-out"
+          style={{
+            background:
+              "linear-gradient(to bottom, rgba(0, 0, 0, 0.06), transparent)",
+            opacity: showTopShadow ? 1 : 0,
+          }}
+        />
```
raunakab: I believe the effect which is trying to be achieved here is the masking effect. I already have a component written which achieves this. It's called `VerticalShadowScroller` (located in the `refresh-components` directory). Try wrapping whatever scrollable content with that component instead!

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -439,79 +481,255 @@ function MCPToolsList({
...
+      <div
+        id="action-submenu-scroll"
+        className="flex-1 overflow-y-auto min-h-0 relative"
+        onScroll={(e) => onScrollStateChange(e.currentTarget)}
+      >
+        <div className="space-y-1.5 pb-2 pt-2">
+          {filteredTools.length === 0 ? (
+            <div className="text-center py-4 text-neutral-400 dark:text-neutral-500">
+              No matching tools found
+            </div>
```
raunakab: Render with `Text` here too, please!

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -439,79 +481,255 @@ function MCPToolsList({
...
+                  <span
+                    className={`text-sm font-medium ${
+                      isEnabled
+                        ? "text-neutral-700 dark:text-neutral-200"
+                        : "text-neutral-400 dark:text-neutral-500"
+                    }`}
+                  >
+                    {tool.display_name || tool.name}
+                  </span>
+                </div>
```
raunakab: Is this a "tag"? Or just a simple text component? If the former is true, please try using the `Tag` component. If the latter is true, please render using the refreshed `Text` component, please!

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -439,79 +481,255 @@ function MCPToolsList({
...
+                        <TooltipTrigger asChild>{label}</TooltipTrigger>
+                        <TooltipContent
+                          side="right"
+                          align="start"
+                          className="max-w-xs"
+                        >
+                          <Text inverted>{tool.description}</Text>
+                        </TooltipContent>
+                      </Tooltip>
+                    </TooltipProvider>
```
raunakab: The usage of this is awesome! Actually, since this pattern is so commonplace, I've created a neat little helper called `SimpleTooltip` in the `refresh-components` directory. Its usage is as follows:

```tsx
import SimpleTooltip from "@/refresh-components/SimpleTooltip";

<SimpleTooltip tooltip="...">
  {...}
</SimpleTooltip>
```

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -439,79 +481,255 @@ function MCPToolsList({
...
+                    <div
+                      className={`absolute top-[2px] left-[2px] h-[12px] w-[12px] rounded-full transition-transform duration-200 ease-in-out ${
+                        isEnabled
+                          ? "translate-x-[12px] bg-white"
+                          : "translate-x-0 bg-white dark:bg-neutral-900"
+                      }`}
+                      style={{
+                        boxShadow: "0 0 1px 1px rgba(0, 0, 0, 0.05)",
+                      }}
+                    />
```
raunakab: What is this supposed to be rendering?

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -439,79 +481,255 @@ function MCPToolsList({
...
+                  </button>
```
raunakab: Let's use either the refreshed `IconButton` (if this button is only rendering an icon) OR the refreshed `Button` (if this is rendering text PLUS an icon).

Both are located in `web/src/refresh-components/buttons/`.

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -439,79 +481,255 @@ function MCPToolsList({
...
+                  Re-Authenticate
+                </span>
+              </div>
+              {!isReauthLoading && (
+                <FiChevronRight
+                  size={14}
+                  className="text-neutral-400 dark:text-neutral-500 transition-colors"
+                />
+              )}
+            </button>
```
raunakab: Seems like we're rendering icon buttons here too. Please use the refreshed `IconButton`!

web/src/app/chat/components/input/ActionManagement.tsx
evan-onyx: codex was cooking

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -439,79 +481,255 @@ function MCPToolsList({
...
+                <span className="text-sm font-medium text-neutral-700 dark:text-neutral-300">
+                  Re-Authenticate
+                </span>
+              </div>
+              {!isReauthLoading && (
+                <FiChevronRight
+                  size={14}
+                  className="text-neutral-400 dark:text-neutral-500 transition-colors"
+                />
+              )}
```
raunakab: Let's use the icons in the `web/src/icons` directory instead! There should be an `SvgChevronRight`. @Weves is bringing in the `SvgKey`, so you can wait until his PR lands to use it. The `FiLoader` should be replaced with the `SvgRefreshCw`, I believe.

web/src/app/chat/components/input/ActionManagement.tsx
```diff
@@ -958,14 +1154,17 @@ export function ActionToggle({
...
           {/* Options */}
           <div className="pt-2 flex-1 flex flex-col mx-1 relative overflow-hidden">
-            {showSourceManagement ? (
+            {secondaryView?.type === "sources" ? (
               <>
                 {/* Fixed Header */}
                 <div className="bg-transparent flex-shrink-0">
                   <div className="mx-1">
                     <div className="relative">
                       <button
```
raunakab: Let's use the refreshed `IconButton` here too, please!

raunakab: There's also a raw `input` component below this. Could you also replace that with the refreshed `InputTypeIn` as well, please? Thanks!

web/src/app/chat/components/input/ActionManagement.tsx
raunakab: ➕!!!

web/src/app/chat/components/input/ActionManagement.tsx
evan-onyx: this is the lil circle in the toggle

web/src/app/chat/components/input/ActionManagement.tsx
evan-onyx: I changed the whole file to follow this

web/src/app/chat/components/input/ActionManagement.tsx
evan-onyx: This doesn't work in my component without a ton of adaptation; I'm going to leave a TODO on this one

evan-onyx:
closing in favor of #5913 and a future follow-up

PR #5886, opened 2025-10-23
fix: don't set env vars if they're empty (aws bedrock)

wenxi-onyx:
## Description

- Empty custom config values were being set as env vars, breaking the boto session

## How Has This Been Tested?

## Additional Options

---

## Summary by cubic
Stop setting empty custom config values as environment variables so AWS Bedrock can fall back to IAM roles and avoid boto session failures.

- **Bug Fixes**
  - Filtered out empty strings and None from custom_config on upsert; set to None if the dict is empty.
  - Only set env vars for non-empty values in chat_llm and factory; unset env vars when values are empty.

### DIFF
--- a/backend/onyx/llm/factory.py
+++ b/backend/onyx/llm/factory.py
@@ -403,12 +403,7 @@ def get_llm_model_and_settings(
     model_kwargs = model_kwargs or {}
     if custom_config:
         for k, v in custom_config.items():
-            # If there are any empty or null values
-            # They MUST NOT be set in the env
-            if v is not None and v.strip():
-                os.environ[k] = v
-            else:
-                os.environ.pop(k, None)
+            os.environ[k] = v
     if custom_config and provider == "vertex_ai":
         for k, v in custom_config.items():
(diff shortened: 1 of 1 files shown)


backend/onyx/llm/factory.py
```diff
@@ -403,7 +403,12 @@ def get_llm_model_and_settings(
     model_kwargs = model_kwargs or {}
     if custom_config:
         for k, v in custom_config.items():
-            os.environ[k] = v
+            # If there are any empty or null values
+            # They MUST NOT be set in the env
+            if v is not None and v.strip():
+                os.environ[k] = v
+            else:
+                os.environ.pop(k, None)
```
justin-tahara: Only nit would be if we're doing the same thing in the factory.py and the chat_llm.py, we should just make this a function in like llm/utils.py and then just call it in the places we need it

wenxi-onyx: Actually, since we don't store empty values in custom config anymore, this check is actually not necessary. Reverting to what it was originally.

PR #5892, opened 2025-10-23
fix: update playwright tests for agent creation with knowledge

nmgarza5:
## Description

A few changes here to allow testing knowledge enablement.

1. creating an onyx api client that can be used to talk directly to the backend -> this client should be used for setup/teardown operations. In this PR, I'm specifically adding methods for creating/deleting connectors and document sets.
2. making the assistant tests serial to prevent connector existence from causing flaky failures
3. actually adding a user document set to an assistant (does not test using this knowledge in chat)

closes: https://linear.app/danswer/issue/DAN-2816/edit-e2e-test-to-seed-a-document-first-prior-to-running-a-test-so-that

## How Has This Been Tested?

existing -> unskipping tests dependent on pre-seeded file/connector existence 

## Additional Options

### DIFF
--- a/web/tests/README.md
+++ b/web/tests/README.md
@@ -113,59 +113,59 @@ Use queries in this priority order (most accessible first):
 
 ```typescript
 // Buttons
-screen.getByRole("button", { name: /submit/i })
-screen.getByRole("button", { name: /cancel/i })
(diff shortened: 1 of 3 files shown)


nmgarza5:
@cubic redo your summary

PR #5896, opened 2025-10-23
fix: chat feedback

wenxi-onyx:
## Description

- Actually accept text feedback
- Actually display feedback in query history
- Show visual changes when feedback is clicked
- Add ability to remove feedback instead of just overwriting 
- Optimistically update UI 
- Claude code alleges ~20ms chat page improvement due to memoization of chat state 

## How Has This Been Tested?

https://github.com/user-attachments/assets/3ce80f72-9266-4280-872f-3af7a59a2676

## Additional Options

### DIFF
--- a/web/src/app/chat/components/ChatPage.tsx
+++ b/web/src/app/chat/components/ChatPage.tsx
@@ -19,7 +19,6 @@ import { usePopup } from "@/components/admin/connectors/Popup";
 import { SEARCH_PARAM_NAMES } from "@/app/chat/services/searchParams";
 import { useFederatedConnectors, useFilters, useLlmManager } from "@/lib/hooks";
 import { useFederatedOAuthStatus } from "@/lib/hooks/useFederatedOAuthStatus";
-import { FeedbackType } from "@/app/chat/interfaces";
 import { OnyxInitializingLoader } from "@/components/OnyxInitializingLoader";
 import { FeedbackModal } from "@/app/chat/components/modal/FeedbackModal";
 import { FiArrowDown } from "react-icons/fi";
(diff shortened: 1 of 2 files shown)


web/src/refresh-components/buttons/IconButton.tsx
```diff
@@ -30,7 +30,7 @@ const buttonClasses = (active: boolean | undefined) =>
     },
     tertiary: {
       main: [
-        active ? "bg-background-tint-00" : "bg-transparent",
+        active ? "bg-background-tint-02" : "bg-transparent",
```
wenxi-onyx, line 36: The feedback buttons are the only ones that actually use this active state. The reason I changed this is because those buttons live on the main chat background, which is white. 00 active state is also white, making the shading invisible. I believe 00 makes sense if the background is gray (e.g. the side panel). Not sure if editing this base value is valid.

web/src/app/chat/stores/useChatSessionStore.ts
```diff
@@ -276,6 +287,49 @@ export const useChatSessionStore = create<ChatSessionStore>()((set, get) => ({
...
+      const updatedSession = {
+        ...session,
+        lastAccessed: new Date(),
+      };
+      const newSessions = new Map(state.sessions);
+      newSessions.set(sessionId, updatedSession);
+
+      return { sessions: newSessions };
+    });
+  },
```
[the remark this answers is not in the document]
wenxi-onyx: resolved

backend/onyx/server/query_and_chat/models.py
```diff
@@ -266,6 +266,7 @@ class ChatMessageDetail(BaseModel):
     files: list[FileDescriptor]
     tool_call: ToolCallFinalResult | None
     error: str | None = None
+    current_feedback: str | None = None  # "like" | "dislike" | null
```
Weves, line 269: nit: should use an Enum type

web/src/app/chat/components/ChatPage.tsx
```diff
@@ -274,9 +279,77 @@ export function ChatPage({
   const filterManager = useFilters();
   const [isChatSearchModalOpen, setIsChatSearchModalOpen] = useState(false);
 
-  const [currentFeedback, setCurrentFeedback] = useState<
-    [FeedbackType, number] | null
-  >(null);
+  // Get the store action for updating message feedback
```
Weves: can this be a hook to avoid polluting the ChatPage.tsx file? :big_grin:

wenxi-onyx: done

backend/onyx/server/query_and_chat/models.py
wenxi-onyx: per convo, moving to separate pr

silberda:
Feedback Issue seems to be reintroduced in v2.4.0. It worked for v2.2 up until v2.3.
So you're only able to submit feedback after refreshing the page after the first message of a new chat.

PR #5904, opened 2025-10-24
chore(citation): processing

rguan72:
## Description

Fixes citation processing in new agent framework.
Also refactors context processing for new agent framework to have distinct handlers for each manual transformation we want to do in the context in between tool calls.

Sorry the diff is so large -- most of it is from vibe refactoring the tests

## How Has This Been Tested?

Tested locally + unit tests

## Additional Options

---

## Summary by cubic
Adds context handlers to assign sequential citation numbers and refresh the task prompt on each agent turn, improving deterministic inline citations and prompt control. Also sets sensible LLM timeouts based on model type.

- **New Features**
  - Citation handler assigns document_citation_number to LlmDoc in tool outputs, tracked via ChatTurnContext.documents_cited_count.
  - Task prompt handler replaces the last user task prompt using build_task_prompt_reminders_v2 and should_cite_documents.
  - Default LLM timeout now uses QA_TIMEOUT (10x for reasoning models) and is passed to providers.
  - Added LlmDoc.document_citation_number and comprehensive unit tests for both handlers and their combined flow.

- **Refactors**
  - Agent loop now separates chat_history, current_user_message, and agent_turn_messages, then applies handlers in order.
  - Removed old task prompt mutation logic and standardized to REQUIRE_CITATION_STATEMENT.
  - Threaded timeout through get_llm_model_and_settings_*; process_message passes timeout=None to use the new defaults.

### DIFF
--- a/backend/onyx/agents/agent_sdk/message_format.py
+++ b/backend/onyx/agents/agent_sdk/message_format.py
@@ -5,8 +5,9 @@ from langchain.schema.messages import BaseMessage
 from onyx.agents.agent_sdk.message_types import AgentSDKMessage
 from onyx.agents.agent_sdk.message_types import AssistantMessageWithContent
 from onyx.agents.agent_sdk.message_types import ImageContent
+from onyx.agents.agent_sdk.message_types import InputTextContent
+from onyx.agents.agent_sdk.message_types import OutputTextContent
 from onyx.agents.agent_sdk.message_types import SystemMessage
-from onyx.agents.agent_sdk.message_types import TextContent
 from onyx.agents.agent_sdk.message_types import UserMessage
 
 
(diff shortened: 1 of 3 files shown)


backend/onyx/agents/agent_search/dr/sub_agents/web_search/utils.py
```diff
@@ -75,3 +77,20 @@ def dummy_inference_section_from_internet_search_result(
         chunks=[],
         combined_content="",
     )
+
+
+def llm_doc_from_web_content(web_content: WebContent) -> LlmDoc:
+    """Create an LlmDoc from WebContent with the INTERNET_SEARCH_DOC_ prefix"""
+    return LlmDoc(
+        document_id="INTERNET_SEARCH_DOC_" + web_content.link,
```
evan-onyx, line 88: this is somewhat concerning to me; are we sure there aren't any flows downstream that expect a real document with a corresponding id to exist? If there isn't already a comment in LLMDoc saying so, we should mention that instances of this class do NOT necessarily correspond to fragments of indexed docs

backend/onyx/chat/stream_processing/answer_response_handler.py
```diff
@@ -49,6 +46,8 @@ class CitationResponseHandler(AnswerResponseHandler):
     def __init__(
         self,
         context_docs: list[LlmDoc],
+        # TODO: Right now these are always the same at runtime in prod
+        # one of them should be removed
```
evan-onyx: So considering
 a) the order documents are presented to the LLM 
b) the order documents are first displayed to the user
c) the final order the user sees

is there no longer a difference between b and c?

backend/onyx/chat/turn/context_handler/citation.py
```diff
@@ -0,0 +1,75 @@
...
+    """Result of assigning citation numbers to recent tool calls."""
+
+    updated_messages: Sequence[dict[str, Any]]
+    num_docs_cited: int
+    num_tool_calls_cited: int
+    new_llm_docs: list[LlmDoc]
+
+
+def assign_citation_numbers_recent_tool_calls(
+    agent_turn_messages: Sequence[dict[str, Any]],
```
evan-onyx: could we make these message dicts at least TypedDicts rather than str -> Any? ideal would be BaseModel but if you need the typedDict's flexibility that's still way preferrable over having Any's floating around

backend/onyx/chat/turn/context_handler/citation.py
```diff
@@ -0,0 +1,75 @@
...
+                            )
+                    if updated_citation_number:
+                        new_message = message.copy()
+                        new_message["output"] = json.dumps(
+                            [doc.model_dump(mode="json") for doc in llm_docs]
+                        )
+                        num_tool_calls_cited += 1
+                        new_llm_docs.extend(llm_docs)
+            curr_tool_call_idx += 1
+        if new_message is not None:
```
evan-onyx: replace this if/else with:
`updated_messages.append(new_message or message)`

assuming you can do the TypedDict or BaseModel thing above or can guarantee new_message is not an empty dict (or if you'd actually like to default to message in that case too)

backend/onyx/chat/turn/context_handler/citation.py
```diff
@@ -0,0 +1,75 @@
...
+    docs_cited_so_far = ctx.documents_cited_count
+    tool_calls_cited_so_far = ctx.tool_calls_cited_count
+    num_tool_calls_cited = 0
+    num_docs_cited = 0
+    curr_tool_call_idx = 0
+    new_llm_docs: list[LlmDoc] = []
+    for message in agent_turn_messages:
+        new_message = None
+        if message.get("type") == "function_call_output":
+            if curr_tool_call_idx >= tool_calls_cited_so_far:
```
evan-onyx, line 38: prefer connecting these conditions with `and` to reduce indentation

backend/onyx/chat/turn/context_handler/task_prompt.py
```diff
@@ -0,0 +1,32 @@
+"""Task prompt context handler for updating task prompts in agent messages."""
+
+from collections.abc import Sequence
+
+from onyx.chat.models import PromptConfig
+from onyx.prompts.prompt_utils import build_task_prompt_reminders_v2
+
+
+def update_task_prompt(
+    current_user_message: dict,  # TODO should this be more strongly typed?
```
evan-onyx: I would certainly like this to be more strongly typed!

backend/onyx/chat/turn/models.py
```diff
@@ -71,3 +74,10 @@ class ChatTurnContext:
     )
     web_fetch_results: list[dict] = dataclasses.field(default_factory=list)
     should_cite_documents: bool = False
+    documents_cited_count: int = 0
```
evan-onyx, line 69: fwiw cubic is right, just for consistency with the rest of the codebase I'd like to use BaseModel here

backend/onyx/tools/tool_implementations_v2/web.py
```diff
@@ -106,33 +83,29 @@ def _web_search_core(
...
+    all_hits: list[WebSearchResult] = []
     for result_id in search_results_dict:
         hits = search_results_dict[result_id]
         if hits:
             all_hits.extend(hits)
 
     # Convert hits to WebSearchResult objects
     results = []
-    for i, r in enumerate(all_hits):
+    for _, r in enumerate(all_hits):
```
evan-onyx: remove enumerate

backend/tests/unit/onyx/chat/stream_processing/test_citation_processing.py
```diff
@@ -93,6 +93,13 @@ def process_text(
     return final_answer_text, citations
 
 
+def test_citation_extraction_v2(mock_data: tuple[list[LlmDoc], dict[str, int]]) -> None:
+    input_tokens = ["Hi", "[", "1", "].\n\n"]
+    final_answer_text, citations = process_text(input_tokens, mock_data)
+    print(final_answer_text)
```
evan-onyx: should have some kind of assert rather than print?

backend/onyx/agents/agent_search/dr/sub_agents/web_search/utils.py
rguan72: Hmmm, I just carried over this pattern from what we're doing with deep research today... let me leave a comment and a TODO

backend/onyx/chat/stream_processing/answer_response_handler.py
rguan72: yeah -- at the very least the field is literally the same no matter what lol

rguan72: we're not doing re-ranking anymore

backend/onyx/chat/turn/models.py
rguan72: i'm sorry evan I'm going to tell cubic to ignore this -- there is some weird circular import thing when I try to use BaseModel

PR #5908, opened 2025-10-24
feat: save user's theme preference

wenxi-onyx:
## Description

## How Has This Been Tested?

## Additional Options

---
## Summary by cubic
Persisted the user's theme preference (light, dark, system) to the database and synced it across sessions. The UI now saves theme changes immediately, with a safe fallback if the server update fails.

- **New Features**
  - Added ThemePreference enum and a non-null user.theme_preference column (default: system) via Alembic migration.
  - Exposed theme_preference in UserPreferences and added an API to update it, with no-auth KV fallback.
  - Updated UserSettings to save theme changes to the server; applies locally even if the save fails.
  - Added UserProvider logic to update theme preference and silently migrate existing localStorage theme to the DB.
  - Updated types to include ThemePreference.


web/src/app/chat/components/modal/UserSettings.tsx
```diff
@@ -391,22 +400,30 @@ export function UserSettings({ onClose }: UserSettingsProps) {
...
-                    value="system"
+                    value={ThemePreference.SYSTEM}
                     icon={<Monitor className="h-4 w-4" />}
                   >
                     System
                   </SelectItem>
-                  <SelectItem value="light" icon={<Sun className="h-4 w-4" />}>
+                  <SelectItem
+                    value={ThemePreference.LIGHT}
+                    icon={<Sun className="h-4 w-4" />}
```
raunakab, line 409: @duo-onyx is creating icons for these (will work on it next week). *Not a blocker for this PR*, but I'll go in and update these icons when they land.

web/src/app/chat/components/modal/UserSettings.tsx
```diff
@@ -90,6 +92,13 @@ export function UserSettings({ onClose }: UserSettingsProps) {
     }
   }, [defaultModel, currentDefaultModel]);
 
+  // Sync selectedTheme with current theme from next-themes
+  useEffect(() => {
+    if (theme) {
+      setSelectedTheme(theme);
```
raunakab: What's the difference between `theme` and `setTheme` items (provided by the `useTheme` hook) and `selectedTheme` and `setSelectedTheme` items (provided by the `useState` local state)?

If this is duplicated logic, maybe this would be a good PR to remove them in.

web/src/components/theme/ThemeSync.tsx
raunakab, line 1: How come we've turned this into a `React.FunctionComponent` instead of a React hook? Something like `useThemeSync` may suffice (and doesn't need to return null). You could also simply just call it from `AppProvider` without having to hook it into the component-tree.

backend/alembic/versions/09995b8811eb_add_theme_preference_to_user.py
raunakab, line 1: Oh interesting. I thought we had a `UserPreferences` table, but apparently we don't. Was going to suggest to use that instead, but never mind.

web/src/app/chat/components/modal/UserSettings.tsx
```diff
@@ -391,22 +400,30 @@ export function UserSettings({ onClose }: UserSettingsProps) {
                 onValueChange={(value) => {
                   setSelectedTheme(value);
                   setTheme(value);
+                  updateUserThemePreference(value as ThemePreference).catch(
+                    () => {
+                      // Theme still applied locally even if DB save fails
+                    }
+                  );
```
raunakab: Do we need the `.catch(...)` block here if it's not doing anything? Maybe the comment inside the `catch` can go above the call to `updateUserThemePreference`.

web/src/app/chat/components/modal/UserSettings.tsx
wenxi-onyx: - `theme` and `setTheme` are from next-themes and updates the DOM
- `selected` is just for the selection component --> having this allows the _component_ to update immediately without waiting for next-themes 

tl;dr I think this can be removed as it provides little value (IMO optimistic update not necessary as the theme switch is not that slow)

backend/alembic/versions/09995b8811eb_add_theme_preference_to_user.py
wenxi-onyx: We should have a user preferences or user settings table IMO. @evan-onyx FYI. The user table is getting a bit burdensome

web/src/components/theme/ThemeSync.tsx
wenxi-onyx: Because Claude code told me so :despair: --> converted and moved into UserProvider (needs user provider to be rendered first)

web/src/app/chat/components/modal/UserSettings.tsx
wenxi-onyx: No - the original intention was that we should just silently fail the db update, but this is actually just unnecessary since it's handled in the func

PR #5910, opened 2025-10-24
chore(gha): disable docker caching for backend images

trial2onyx:
## Description

We're still seeing [failed image pulls](https://onyx-company.slack.com/archives/C09DHFK3220/p1761278693845939) after [re-enabling caching](https://github.com/onyx-dot-app/onyx/pull/5839). 

My impression is these [COPY's](https://github.com/onyx-dot-app/onyx/blob/af243b0ef5d87141534d91669b6a250313d148ac/backend/Dockerfile#L104-L119) in the backend image is what is causing issues [[1](https://github.com/moby/buildkit/issues/2198#issuecomment-924064592)][[2](https://github.com/moby/buildkit/issues/1980#issuecomment-783492400)]. Rather than turning off caching wholesale, disable it just for the backend image. The model-server takes longer to build and changes less often, so it benefits the most from keep caching enabled. The [integration image is composed of the backend image now](https://github.com/onyx-dot-app/onyx/pull/5873), so need to disable caching for it as well.

## How Has This Been Tested?

Captured by presubmit

## Additional Options

- [x] Override Linear Check

### DIFF
--- a/.github/workflows/pr-integration-tests.yml
+++ b/.github/workflows/pr-integration-tests.yml
@@ -95,7 +95,7 @@ jobs:
           tags: ${{ env.PRIVATE_REGISTRY }}/integration-test-onyx-backend:test-${{ github.run_id }}
           push: true
           outputs: type=registry
-          no-cache: ${{ vars.DOCKER_NO_CACHE == 'true' }}
+          no-cache: true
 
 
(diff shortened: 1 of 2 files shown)


trial2onyx:
@greptileai

PR #5917, opened 2025-10-24
fix: input bar thickness

Weves:
## Description

Fixes https://linear.app/danswer/issue/DAN-2935/main-input-bar-height-is-off

New:

Screenshot 2025-10-24 at 2 52 23 PM

Old:

Screenshot 2025-10-24 at 2 52 42 PM

## How Has This Been Tested?

👀 

## Additional Options

### DIFF
--- a/web/src/app/chat/components/input/ChatInputBar.tsx
+++ b/web/src/app/chat/components/input/ChatInputBar.tsx
@@ -506,8 +506,8 @@ function ChatInputBarInner({
           </div>
         )}
 
-        <div className="flex justify-between items-center w-full p-spacing-inline">
-          <div className="flex flex-row items-center gap-spacing-inline">
+        <div className="flex justify-between items-center w-full p-1">
+          <div className="flex flex-row items-center gap-1">
             <FilePicker
               onFileClick={handleFileClick}
               onPickRecent={(file: ProjectFile) => {


web/src/app/chat/components/input/ChatInputBar.tsx
```diff
@@ -421,7 +415,20 @@ function ChatInputBarInner({
...
+            "placeholder:text-text-03",
+            "whitespace-normal",
+            "break-word",
+            "overscroll-contain",
+            "overflow-y-auto",
+            // NOTE: need custom padding for textarea specifically to get to the 44px desired
+            // total height
+            "px-[12px]",
+            "pb-[8px]",
+            "pt-[12px]"
```
raunakab: Change looks good.

We have constants defined (in REMs) which map over to these values. Would prefer we use those to avoid hardcoding values directly into the app.

```txt
12px -> 0.75rem -> padding-button (px-padding-button pt-padding-button)
8px -> 0.5rem -> spacing-interline (pb-spacing-interline)
```

The constants are defined in `web/src/app/globals.css`.

raunakab: If you want to convert from `px`s to `REM`s, you can take the value and divide by `16`.

(This isn't a purely static mapping, however; the benefit to using REM is that it scales as the zoom-level in the browser scales; `1rem` at normal zoom will be `16px`, but at 200% zoom, it'll be `32px`).

PR #5918, opened 2025-10-24
feat(tracing): custom tracing processor which injects tenant_id

trial2onyx:
## Description

Includes `tenant_id` along with braintrust traces which allows for per-tenant usage tracking.

I'm not entirely sure why the why the custom processor doesn't capture the `@traced` decorator. AFAICT, it seems like it should -- it does look like it uses the same `span.log()` machinery, but couldn't get those callsites to even register the custom processor. Defining the metadata along with the decorator seems non-ideal but also not too burdensome. There is also a non-zero chance there is a bug/quirk in the braintrust sdk that is being exposed.

Also upgrades braintrust sdk to latest as that appeared to fix a bug relevant to the `TenantContextTracingProcessor`.

## How Has This Been Tested?

Tested on existing `invoke llm`, `stream llm` and `fast_chat_turn` traces and confirmed their metadata appears in Braintrust,

Screenshot 2025-10-24 at 3 01 33 PM

## Additional Options

- [x] Override Linear Check

---
## Summary by cubic
Add a custom Braintrust tracing processor that injects tenant_id into trace metadata and updates LLM trace decorators to include tenant_id, enabling per-tenant usage tracking. Also upgrade Braintrust SDK to v0.3.5 to resolve a processor issue.

- **New Features**
  - TenantContextTracingProcessor injects tenant_id on trace start.
  - Added tenant_id from CURRENT_TENANT_ID_CONTEXTVAR to @traced for invoke llm, stream llm, and clarifier.
  - Switched set_trace_processors to use the new processor.

- **Dependencies**
  - braintrust[openai-agents]: 0.2.6 → 0.3.5

### DIFF
--- a/backend/onyx/tracing/braintrust_tracing.py
+++ b/backend/onyx/tracing/braintrust_tracing.py
@@ -1,4 +1,5 @@
 import os
+from collections.abc import Callable
 from functools import wraps
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


trial2onyx:
@greptileai

PR #5920, opened 2025-10-24
fix(slack bot): add thread context

jessicasingh7:
## Description

Tested in Slack channel with Slack bot: https://onyx-company.slack.com/archives/C09CS1EPE5C/p1761593363377169

## How Has This Been Tested?

https://onyx-company.slack.com/archives/C09CS1EPE5C/p1762285359355479

Example thread about lunch orders- responding to reply with context of thread

## Additional Options

### DIFF
--- a/backend/onyx/onyxbot/slack/handlers/handle_regular_answer.py
+++ b/backend/onyx/onyxbot/slack/handlers/handle_regular_answer.py
@@ -162,7 +162,7 @@ def handle_regular_answer(
     # Convert ThreadMessage objects to PreviousMessage for query rephrasing
     thread_previous_messages: list[PreviousMessage] | None = None
     if history_messages:
-        llm, _ = get_llms_for_persona(persona)
+        llm, _ = get_llms_for_persona(persona, user)
         llm_tokenizer = get_tokenizer(
             model_name=llm.config.model_name,
             provider_type=llm.config.model_provider,


backend/onyx/onyxbot/slack/handlers/handle_regular_answer.py
```diff
@@ -155,6 +158,33 @@ def handle_regular_answer(
...
+                PreviousMessage(
+                    message=thread_msg.message,
+                    token_count=token_count,
+                    message_type=thread_msg.role,
+                    files=[],
+                    tool_call=None,
+                    refined_answer_improvement=None,
+                    research_answer_purpose=None,
+                )
+            )
```
justin-tahara: If there is a giant thread, can we in theory blow up the context window? Should we limit the number of thread messages that we take in? 

What happens if there is a thread that has 100s of messages? I think the token count would be rather large. 

Worth double checking since I can't account for all types of slack threads

justin-tahara:
Also would appreciate some sort of testing. Either in the description or with unit tests or something

backend/onyx/onyxbot/slack/handlers/handle_regular_answer.py
jessicasingh7: Makes sense, limited it to max 3000 tokens

Weves (approved):

PR #5921, opened 2025-10-24
fix(tracing): include token count with "invoke llm" and "stream llm"

trial2onyx:
## Description

Updates the `invoke llm` and `stream llm` log events to include the token counts. For streaming events, ~~use the relevant tokenizer to estimate the token counts and includes `is_estimate` metadata which can be filtered out in situations accuracy is relevant to the counts~~ `include_usage` when applicable.

I don't think the braintrust sdk upgrade is relevant here, but is getting upgraded in https://github.com/onyx-dot-app/onyx/pull/5918, so included for my convenience. 

## How Has This Been Tested?

Confirmed disabling braintrust does not cause errors.

Confirmed in braintrust values appear as expected with reasonable values,

Screenshot 2025-10-24 at 3 45 08 PM

## Additional Options

- [x] Override Linear Check

---
## Summary by cubic
Add token usage metrics to LLM tracing. “invoke llm” logs exact counts when available; “stream llm” logs estimated counts with an is_estimate flag.

- **New Features**
  - Log prompt_tokens, completion_tokens, and total_tokens via current_span for invoke responses (from response.usage when present).
  - Estimate streaming token counts with the model’s tokenizer; include metadata is_estimate: true and estimate_method: tokenizer_estimate.
  - Logging is best-effort and never impacts the request path.

- **Dependencies**
  - Upgrade braintrust[openai-agents] to 0.3.5.

### DIFF
--- a/backend/onyx/llm/chat_llm.py
+++ b/backend/onyx/llm/chat_llm.py
@@ -40,10 +40,7 @@ from onyx.llm.interfaces import LLMConfig
 from onyx.llm.interfaces import ToolChoiceOptions
 from onyx.llm.llm_provider_options import VERTEX_CREDENTIALS_FILE_KWARG
 from onyx.llm.llm_provider_options import VERTEX_LOCATION_KWARG
-from onyx.llm.utils import check_message_tokens
-from onyx.llm.utils import check_number_of_tokens
 from onyx.llm.utils import model_is_reasoning_model
-from onyx.natural_language_processing.utils import get_tokenizer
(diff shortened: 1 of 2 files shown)


trial2onyx:
Replaced the estimated token counting with provider usage data. Confirmed it worked with `stream llm` events using mistralai/devstral-small & openrouter and `clarifier stream and process` events using gpt-5-nano & openrouter. Also confirmed the option is ignored when streaming is disabled (include the case `DISABLE_LITELLM_STREAMING=True`).

PR #5930, opened 2025-10-26
chore(force): tool use fix

rguan72:
## Description

This fixes force tool use for the new agent framework

## How Has This Been Tested?

Tested locally + added integration test

## Additional Options

---
## Summary by cubic
Fixes agent tool selection so web_search uses our function tool instead of the OpenAI-hosted tool, restoring expected tool calls during FastChat turns. Also updates OpenAI and openai-agents to compatible versions.

- **Bug Fixes**
  - Monkey-patch convert_tool_choice to map "web_search" to a function tool.
  - Apply the patch at the start of _run_agent_loop in fast_chat_turn.

- **Dependencies**
  - openai: 1.107.1 → 2.6.1 (default, model_server)
  - openai-agents: 0.3.3 → 0.4.2


backend/onyx/chat/turn/models.py
```diff
@@ -46,14 +40,12 @@
 class ChatTurnDependencies:
     llm_model: Model
     model_settings: ModelSettings
+    # TODO we can delete this field (combine them)
```
evan-onyx, line 43: let's do it now

backend/onyx/natural_language_processing/search_nlp_models.py
```diff
@@ -200,7 +200,7 @@ async def _embed_openai(
             response = await client.embeddings.create(
                 input=text_batch,
                 model=model,
-                dimensions=reduced_dimension or openai.NOT_GIVEN,
+                dimensions=reduced_dimension if reduced_dimension else openai.omit,
```
evan-onyx: `reduced_dimension or openai.omit`

PR #5953, opened 2025-10-27
chore(add): tests for citation and force tool use and some cleanup

rguan72:
## Description

Tested locally, but it's mostly just cleanup and new tests anyways

## How Has This Been Tested?

The pr adds new tests

## Additional Options

---
## Summary by cubic
Refactored turn saving to decouple from ChatTurnContext and ensure correct citation extraction. Added a unit test that verifies only cited docs are stored and tool iteration steps are persisted.

- **Refactors**
  - Renamed save_iteration to save_turn with explicit args: model_name, model_provider, iteration_instructions, global_iteration_responses.
  - Updated fast_chat_turn to use save_turn and import extract_final_answer_from_packets from the new module.
  - Citations now derive from rendered [[n]] links mapped to ordered_fetched_documents; token count uses the provided model config.

### DIFF
--- a/backend/alembic/versions/2b75d0a8ffcb_user_file_schema_cleanup.py
+++ b/backend/alembic/versions/2b75d0a8ffcb_user_file_schema_cleanup.py
@@ -12,6 +12,7 @@ from alembic import op
 import sqlalchemy as sa
 from sqlalchemy import text
 import logging
+import fastapi_users_db_sqlalchemy
 
 logger = logging.getLogger("alembic.runtime.migration")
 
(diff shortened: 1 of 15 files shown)


backend/onyx/evals/models.py
```diff
@@ -23,11 +23,7 @@ class EvalConfiguration(BaseModel):
 
 
 class EvalConfigurationOptions(BaseModel):
-    builtin_tool_types: list[str] = list(
-        tool_name
-        for tool_name in BUILT_IN_TOOL_MAP.keys()
-        if tool_name != "OktaProfileTool"
```
rguan72, line 29: confirmed with chris + yuhong this is not needed

PR #5961, opened 2025-10-28
chore(turn): fix sending multiple messages

rguan72:
## Description

I introduced a bug (behind feature flag) where sending multiple messages doesn't work (lol)

## How Has This Been Tested?

tested locally and added an integration test to catch the issue

## Additional Options

---
## Summary by cubic
Fix chat stream parsing to always capture the assistant_message_id and surface errors, enabling reliable multi-turn chat. Adds an integration test that sends two messages and verifies correct history and parent_message_id usage.

- **Bug Fixes**
  - Parse reserved_assistant_message_id from stream and require it; raise if missing.
  - Return structured ErrorResponse with error and stack_trace from the stream.
  - Correctly build full_message and track top_documents, tool deltas, and heartbeat packets.
  - Added integration test for two consecutive messages.

### DIFF
--- a/backend/requirements/default.txt
+++ b/backend/requirements/default.txt
@@ -40,7 +40,7 @@ langgraph==0.2.72
 langgraph-checkpoint==2.0.13
 langgraph-sdk==0.1.44
 lazy_imports==1.0.1
-litellm==1.79.0 
+litellm==1.76.2 
 lxml==5.3.0
 Mako==1.2.4
 markitdown[pdf, docx, pptx, xlsx, xls]==0.1.2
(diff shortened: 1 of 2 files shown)


backend/tests/integration/common_utils/managers/chat.py
```diff
@@ -132,79 +133,74 @@ def analyze_response(response: Response) -> StreamedResponse:
...
-                )
-                continue
-
-            if packet_type == "internal_search_tool_delta":
-                ind_to_tool_use[ind].queries.extend(data_obj.get("queries", []))
-
-                documents = data_obj.get("documents", [])
-                ind_to_tool_use[ind].documents.extend(
-                    [SavedSearchDoc(**doc) for doc in documents]
+            if data.get("reserved_assistant_message_id"):
```
evan-onyx: I'd really prefer that these are string constants or enums (depending on context) but if you just wanna get this in then just make sure its on your TODO list

rguan72: makes sens

PR #5962, opened 2025-10-28
fix: deprecated confluence oauth api

evan-onyx:
## Description

User's with OAuth credentials weren't able to create confluence connectors due to a deprecated API for fetching spaces.
Added the new required scope and updated code.

## How Has This Been Tested?

tested in UI and will rely on confluence connector tests to prevent regressions

## Additional Options

### DIFF
--- a/backend/tests/unit/file_store/test_file_store.py
+++ b/backend/tests/unit/file_store/test_file_store.py
@@ -23,11 +23,11 @@ from onyx.file_store.file_store import get_default_file_store
 from onyx.file_store.file_store import S3BackedFileStore
 
 
-class TestDBBase(DeclarativeBase):
+class DBBaseTest(DeclarativeBase):
     pass
 
 
-class FileRecord(TestDBBase):
+class FileRecord(DBBaseTest):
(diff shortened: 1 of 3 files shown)


backend/tests/unit/file_store/test_file_store.py
```diff
@@ -23,11 +23,11 @@
 from onyx.file_store.file_store import S3BackedFileStore
 
 
-class TestDBBase(DeclarativeBase):
```
wenxi-onyx, line 26: @evan-onyx Did you accidentally commit this?

backend/tests/unit/onyx/connectors/zendesk/test_zendesk_rate_limit.py
```diff
@@ -39,7 +39,6 @@ def raise_for_status(self) -> None:
         return None
 
 
-@pytest.mark.unit
```
wenxi-onyx, line 42: Was this intended?

wenxi-onyx (changes requested):

PR #5963, opened 2025-10-28
chore(ollama): ollama support

rguan72:
## Description

Fix Ollama

## How Has This Been Tested?

tested locally

## Additional Options

---
## Summary by cubic
Add full Ollama Chat support with reasoning and tool-calling, using LiteLLM patches for correct request/stream handling. Renames the provider to ollama_chat and updates the admin UI; also bumps litellm to 1.79.0.

- **New Features**
  - Add LiteLLM singleton with Ollama config and monkey patches (reasoning_content, tool_calls, usage, finish_reason) and register common Ollama models.
  - Change provider key to ollama_chat; update admin model list endpoint and icon mapping.
  - Add Agent SDK ReasoningMessage type for reasoning outputs.

- **Bug Fixes**
  - Support unicode citation brackets (【】, ［］) and mixed formats; add unit tests.
  - Set OLLAMA_API_BASE from api_base for Ollama requests.
  - Mask Authorization headers in Braintrust traces.
  - Make image generation base64-only; remove URL outputs and additional headers; update saving and tests.

<sup>Written for commit e729331. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/tools/tool_implementations/images/image_generation_tool.py
+++ b/backend/onyx/tools/tool_implementations/images/image_generation_tool.py
@@ -242,7 +242,8 @@ class ImageGenerationTool(Tool[None]):
                 api_key=self.api_key,
                 api_base=self.api_base or None,
                 api_version=self.api_version or None,
-                response_format="b64_json",
+                # response_format parameter is not supported for gpt-image-1
+                response_format=None if self.model == "gpt-image-1" else "b64_json",
                 size=size,
                 n=1,
             )


backend/onyx/llm/litellm_singleton/monkey_patches.py
```diff
@@ -0,0 +1,129 @@
...
+    # Monkey patches for issues that are hopefully resolved soon
+    # This is needed for litellm==1.79.0
+    # Should be removed once this PR is merged: https://github.com/BerriAI/litellm/pull/
+    if (
+        getattr(OllamaChatConfig.transform_request, "__name__", "")
+        == "_patched_transform_request"
+    ):
+        return
+
+    def _patched_transform_request(
```
rguan72, line 40: this is copy pasted from ollama with a small tweak

backend/onyx/llm/litellm_singleton/monkey_patches.py
```diff
@@ -0,0 +1,129 @@
...
+            ollama_message = OllamaChatCompletionMessage(
+                role=cast(str, m.get("role")),
+            )
+            if reasoning_content is not None:
+                ollama_message["thinking"] = reasoning_content
+            if content_str is not None:
+                ollama_message["content"] = content_str
+            if images is not None:
+                ollama_message["images"] = images
+            if new_tools:
```
rguan72, line 94: this is the tweak -- litellm just throws the `new_tools` variable away

backend/onyx/llm/chat_llm.py
```diff
@@ -312,7 +313,8 @@ def __init__(
                     os.environ[k] = v
                 else:
                     os.environ.pop(k, None)
-
+        if model_provider == OLLAMA_PROVIDER_NAME:
+            os.environ["OLLAMA_API_BASE"] = api_base
```
evan-onyx, line 318: is this necessary? I'd really prefer avoiding on-the-fly modification of env vars if there's a way around it

rguan72: setting it in .env won't work because then it adds overhead to the deployment setup when it doesn't need to

rguan72: let me think about this a bit more

rguan72: good news -- I think this was actually a red herring and I thought it fixed something but it didn't

PR #6001, opened 2025-10-30
feat: View permissioning information for documents in documents explorer

trial2onyx:
## Description
Display permissioning (access control) information for documents in Document Explorer.

## How Has This Been Tested?
Screenshot 2025-10-30 at 12 18 49 PM

## Additional Options

---
## Summary by cubic
Show document permissions in the Documents Explorer so you can quickly see who has access. The app now surfaces the ACL from Vespa and displays it per document.

- **New Features**
  - Added access_control_list to chunk and document models and selectively fetched it from Vespa.
  - Parsed ACL keys from Vespa hits and passed them through to SearchDoc.
  - Rendered “Permissions: …” in Explorer with a fallback to “None” when no ACL is present.

<sup>Written for commit b99b5ec. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/context/search/models.py
+++ b/backend/onyx/context/search/models.py
@@ -257,6 +257,7 @@ class InferenceChunk(BaseChunk):
     large_chunk_reference_ids: list[int] = Field(default_factory=list)
 
     is_federated: bool = False
+    access_control_list: list[str] | None = None
 
     @property
     def unique_id(self) -> str:
@@ -356,6 +357,7 @@ class SearchDoc(BaseModel):
     primary_owners: list[str] | None = None
     secondary_owners: list[str] | None = None
     is_internet: bool = False
+    access_control_list: list[str] | None = None
 
     @classmethod
     def from_chunks_or_sections(
@@ -389,6 +391,7 @@ class SearchDoc(BaseModel):
                 primary_owners=chunk.primary_owners,
                 secondary_owners=chunk.secondary_owners,
                 is_internet=False,
+                access_control_list=chunk.access_control_list,
             )
             for item in items
         ]
@@ -450,6 +453,7 @@ class SavedSearchDoc(SearchDoc):
             primary_owners=None,
             secondary_owners=None,
             is_internet=True,
+            access_control_list=[],
         )
 
     def __lt__(self, other: Any) -> bool:
--- a/backend/onyx/document_index/vespa/shared_utils/vespa_request_builders.py
+++ b/backend/onyx/document_index/vespa/shared_utils/vespa_request_builders.py
@@ -3,17 +3,34 @@ from datetime import timedelta
 from datetime import timezone
 
 from onyx.configs.constants import INDEX_SEPARATOR
+from onyx.configs.constants import SOURCE_TYPE
 from onyx.context.search.models import IndexFilters
 from onyx.document_index.interfaces import VespaChunkRequest
+from onyx.document_index.vespa.deletion import CONTENT_SUMMARY
 from onyx.document_index.vespa_constants import ACCESS_CONTROL_LIST
+from onyx.document_index.vespa_constants import AGGREGATED_CHUNK_BOOST_FACTOR
+from onyx.document_index.vespa_constants import BLURB
+from onyx.document_index.vespa_constants import BOOST
+from onyx.document_index.vespa_constants import CHUNK_CONTEXT
 from onyx.document_index.vespa_constants import CHUNK_ID
+from onyx.document_index.vespa_constants import CONTENT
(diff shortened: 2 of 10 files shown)


backend/onyx/document_index/vespa_constants.py
```diff
@@ -100,6 +101,7 @@
     f"{METADATA_SUFFIX}, "
     f"{DOC_SUMMARY}, "
     f"{CHUNK_CONTEXT}, "
-    f"{CONTENT_SUMMARY} "
+    f"{CONTENT_SUMMARY}, "
+    f"{ACCESS_CONTROL_LIST} "
```
Weves: can we only fetch this when we need it?

backend/onyx/document_index/vespa_constants.py
```diff
@@ -74,6 +74,7 @@
 
 # Specific to Vespa, needed for highlighting matching keywords / section
 CONTENT_SUMMARY = "content_summary"
+ACCESS_CONTROL_LIST = "access_control_list"
```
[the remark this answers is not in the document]
Weves: +1!

web/src/lib/search/interfaces.ts
```diff
@@ -71,6 +71,7 @@ export interface OnyxDocument extends MinimalOnyxDocument {
   db_doc_id?: number;
   is_internet: boolean;
   validationState?: null | "good" | "bad";
+  access_control_list: string[];
```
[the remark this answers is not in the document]
Weves: +1!

backend/onyx/context/search/models.py
```diff
@@ -257,6 +257,7 @@ class InferenceChunk(BaseChunk):
     large_chunk_reference_ids: list[int] = Field(default_factory=list)
 
     is_federated: bool = False
+    access_control_list: list[str] | None = None
```
Weves, line 260: nit: add a comment as to when this is present vs not

web/src/app/admin/documents/explorer/Explorer.tsx
```diff
@@ -92,6 +92,9 @@ const DocumentDisplay = ({
             <CustomCheckbox checked={!document.hidden} />
           </div>
         </div>
+        <div className="px-1 py-0.5 bg-accent-background-hovered rounded flex">
+          Permissions: {document.access_control_list?.join(", ") || "None"}
```
Weves: We may need to handle cases where this list is very long. Ideally, it truncates + you can hover it to see the full list

backend/onyx/document_index/vespa_constants.py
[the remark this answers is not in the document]
trial2onyx: oh oops, didn't see that thanks for catching!

web/src/lib/search/interfaces.ts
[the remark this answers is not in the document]
trial2onyx: im going to make the backend return a non null list eveyrtime, so I will remove the or None annotation here and keep the same type annotations for the frontend fields

backend/onyx/context/search/models.py
trial2onyx: im going to make the backend return a non null list eveyrtime, so I will remove the | None annotation here and keep the same type annotations for the frontend fields

backend/onyx/document_index/vespa_constants.py
```diff
@@ -100,6 +99,5 @@
     f"{METADATA_SUFFIX}, "
     f"{DOC_SUMMARY}, "
     f"{CHUNK_CONTEXT}, "
-    f"{CONTENT_SUMMARY} "
-    f"from {{index_name}} where "
+    f"{CONTENT_SUMMARY}, "
```
Weves: I think this comma may break things for the case where we don't add the `ACCESS_CONTROL_LIST`

backend/onyx/document_index/vespa/shared_utils/vespa_request_builders.py
```diff
@@ -232,3 +233,11 @@ def build_vespa_id_based_retrieval_yql(
 
     id_based_retrieval_yql_section += ")"
     return id_based_retrieval_yql_section
+
+
+def build_yql_base(index_name: str, include_acl: bool = False) -> str:
+    yql_base = YQL_BASE.format(index_name=index_name)
```
Weves: nit: prefer moving the YQL_BASE inside the funciton since this is the only place it's used? Nothing should use it directly afaik, they should all only use this function. 

This goes to the principle of cohesion

Weves: `.format(index_name=index_name)` this part isn't necessary anymore, right?

PR #6020, opened 2025-10-31
feat: user-level api keys

trial-danswer:
## Description

https://linear.app/danswer/issue/DAN-2739/user-level-api-keys

## How Has This Been Tested?

Tested creating, deleting, updating API keys on local Onyx server.

## Additional Options

---
## Summary by cubic
Adds user-level API keys (Personal Access Tokens) that mirror the current user's permissions, alongside service account keys with explicit roles. Updates backend and admin UI to create, manage, and display both key types. Addresses Linear issue DAN-2739.

- **New Features**
  - Personal Access Tokens: create keys linked to the current user; role mirrors the owner and cannot be changed on update.
  - Service Account Keys: create with a specific role; can update name and role; PATs only allow name updates.
  - API changes: type is required on create; role is required for service accounts and omitted for PATs; ApiKeyDescriptor now includes api_key_type and user_email; safer deletion only removes fake service account users.
  - Admin UI: tabs to choose Personal Access Token or Service Account Key; edit flow respects key type; keys list shows Type and User columns.

- **Migration**
  - Add user_type enum to users (HUMAN, SERVICE_ACCOUNT) with backfill based on email pattern.
  - Models and logic use user_type to distinguish real users from service account users.

<sup>Written for commit d9ac0546a750f6b5520052f731a7ac5d1d19bfcb. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/db/users.py
+++ b/backend/onyx/db/users.py
@@ -14,6 +14,7 @@ from sqlalchemy.sql.elements import KeyedColumnElement
 
 from onyx.auth.invited_users import remove_user_from_invited_users
 from onyx.auth.schemas import UserRole
+from onyx.auth.schemas import UserType
 from onyx.db.api_key import DANSWER_API_KEY_DUMMY_EMAIL_DOMAIN
 from onyx.db.models import DocumentSet__User
 from onyx.db.models import Persona__User
@@ -232,6 +233,7 @@ def _generate_slack_user(email: str) -> User:
         email=email,
         hashed_password=hashed_pass,
         role=UserRole.SLACK_USER,
+        user_type=UserType.HUMAN,
     )
 
 
@@ -278,6 +280,7 @@ def _generate_ext_permissioned_user(email: str) -> User:
         email=email,
         hashed_password=hashed_pass,
         role=UserRole.EXT_PERM_USER,
+        user_type=UserType.HUMAN,
     )
 
 
--- a/backend/tests/integration/common_utils/chat.py
+++ b/backend/tests/integration/common_utils/chat.py
@@ -1,5 +1,6 @@
 import requests
 
+from onyx.auth.schemas import UserType
 from onyx.db.engine.sql_engine import get_session_with_current_tenant
 from onyx.db.models import User
 
@@ -7,7 +8,11 @@ from onyx.db.models import User
 def test_create_chat_session_and_send_messages() -> None:
     # Create a test user
     with get_session_with_current_tenant() as db_session:
-        test_user = User(email="test@example.com", hashed_password="dummy_hash")
+        test_user = User(
+            email="test@example.com",
+            hashed_password="dummy_hash",
+            user_type=UserType.HUMAN,
+        )
         db_session.add(test_user)
         db_session.commit()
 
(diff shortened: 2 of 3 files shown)


backend/onyx/db/api_key.py
```diff
@@ -64,74 +68,139 @@ def get_api_key_fake_email(
     return f"{DANSWER_API_KEY_PREFIX}{name}@{unique_id}{DANSWER_API_KEY_DUMMY_EMAIL_DOMAIN}"
 
 
+def is_api_key_user(user: User) -> bool:
```
trial-danswer: Maybe I should call this is_service_account_user() to standardize terminology

Weves: yea, I think that would be better

backend/onyx/auth/schemas.py
```diff
@@ -33,6 +33,16 @@ def is_web_login(self) -> bool:
         ]
 
 
+class UserType(str, Enum):
+    HUMAN = "human"
+    API_KEY = "api_key"
```
Weves: probably change this to `service_account`?

backend/onyx/db/api_key.py
```diff
@@ -64,74 +68,139 @@ def get_api_key_fake_email(
...
 def insert_api_key(
     db_session: Session, api_key_args: APIKeyArgs, user_id: uuid.UUID | None
 ) -> ApiKeyDescriptor:
-    std_password_helper = PasswordHelper()
-
     # Get tenant_id from context var (will be default schema for single tenant)
     tenant_id = get_current_tenant_id()
 
+    # Type is required when creating a new API key
+    if api_key_args.type is None:
```
Weves: prefer going with an approach that has static-type checking for things like this. Could we have different models for insert / update?

backend/onyx/db/api_key.py
```diff
@@ -64,74 +68,139 @@ def get_api_key_fake_email(
...
-    )
-    db_session.add(api_key_user_row)
+    if api_key_args.type == ApiKeyType.PERSONAL_ACCESS_TOKEN:
+        # Personal Access Token: Link to real user
+        api_key_user = db_session.scalar(select(User).where(User.id == user_id))
+        if api_key_user is None:
+            raise ValueError(f"User with id {user_id} does not exist")
+    else:
+        # Service Account: Create fake service account user with specific role
+        if api_key_args.role is None:
```
Weves: same here

backend/onyx/db/api_key.py
```diff
@@ -64,74 +68,139 @@ def get_api_key_fake_email(
...
-    api_key_user.role = api_key_args.role
+    # Determine the API key type
+    api_key_type = get_api_key_type(api_key_user)
+
+    # Update the API key's name
+    existing_api_key.name = api_key_args.name
+
+    if api_key_type == ApiKeyType.SERVICE_ACCOUNT:
+        # Service Account: can update name, email, and role
+        if api_key_args.role is None:
```
Weves: same comment here

Weves: we may need to restructure some of the APIs / db operation functions to allow this

PR #6023, opened 2025-10-31
fix(oom): Images Dressed up as PDF files

justin-tahara:
## Description

## How Has This Been Tested?
If there are files with the wrong labels like images that are saved as PDF files our API pods continue to try and read the file and end up causing issues. 

```
ERROR:    10/31/2025 04:34:12 PM           extract_file_text.py  329: [API:fZJIyZPp] Invalid PDF file
Traceback (most recent call last):
  File "/Users/justintahara/onyx/onyx/backend/onyx/file_processing/extract_file_text.py", line 280, in read_pdf_file
    pdf_reader = PdfReader(file)
                 ^^^^^^^^^^^^^^^
  File "/Users/justintahara/onyx/onyx/.venv/lib/python3.11/site-packages/pypdf/_reader.py", line 131, in __init__
    self._initialize_stream(stream)
  File "/Users/justintahara/onyx/onyx/.venv/lib/python3.11/site-packages/pypdf/_reader.py", line 153, in _initialize_stream
    self.read(stream)
  File "/Users/justintahara/onyx/onyx/.venv/lib/python3.11/site-packages/pypdf/_reader.py", line 592, in read
    self._find_eof_marker(stream)
  File "/Users/justintahara/onyx/onyx/.venv/lib/python3.11/site-packages/pypdf/_reader.py", line 698, in _find_eof_marker
    line = read_previous_line(stream)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/justintahara/onyx/onyx/.venv/lib/python3.11/site-packages/pypdf/_utils.py", line 316, in read_previous_line
    raise PdfStreamError(STREAM_TRUNCATED_PREMATURELY)
pypdf.errors.PdfStreamError: Stream has ended unexpectedly
```
This doesn't necessarily cause issues but it is definitely an error we should be handling gracefully.

The old workflow would continuously look for an EOF and would keep scanning the file even after it failed once to find this but this marking never exists in an image file which is why it reads the entire thing into memory.

I created a test file that is a screenshot that I saved as a PDF that I then uploaded and validated that I was able to see the error before the changes and after the changes we gracefully handle this error.

## Additional Options

---
## Summary by cubic
Prevented non-PDF files (e.g., images) from being parsed as PDFs and improved fallback text extraction for binary uploads to reduce errors and bad outputs.

- **Bug Fixes**
  - Validate PDFs by checking for a "%PDF" header (after skipping whitespace) and a trailing "%%EOF"; skip PDF parsing if missing.
  - In _build_content, try UTF‑8 first, then use the generic extractor; safely handle mislabeled text/binary files and only show a placeholder if no readable text is found.
  - Handle missing Unstructured API key and extraction failures with targeted warnings, reducing noisy logs and crashes.

<sup>Written for commit 8504053e1ee2ae9ac395990e10544c2af801220f. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/file_processing/extract_file_text.py
+++ b/backend/onyx/file_processing/extract_file_text.py
@@ -180,40 +180,6 @@ def to_bytesio(stream: IO[bytes]) -> BytesIO:
     return BytesIO(data)
 
 
-def _has_pdf_header(sample: bytes) -> bool:
-    whitespace_chars = b"\x00\x09\x0a\x0c\x0d\x20"
-    return sample.lstrip(whitespace_chars).startswith(b"%PDF")
-
-
-PDF_EOF_SCAN_WINDOW = 8192
-
-
-def _has_pdf_eof_marker(file: IO[Any]) -> bool:
-    """
-    Returns True if the underlying stream appears to contain the PDF %%EOF marker.
-    Falls back to True if the stream is not seekable, since we cannot verify.
-    """
-    seekable = getattr(file, "seekable", None)
-    if callable(seekable) and not seekable():
-        return True
-
-    current_pos = file.tell()
-    try:
-        file.seek(0, os.SEEK_END)
-        file_size = file.tell()
-        if file_size <= 0:
-            return False
-        read_size = min(file_size, PDF_EOF_SCAN_WINDOW)
-        file.seek(file_size - read_size)
-        tail = file.read(read_size)
-        return b"%%EOF" in tail
-    except OSError:
-        # Some file-like objects may not fully support SEEK_END; err on the side of True.
-        return True
-    finally:
-        file.seek(current_pos)
-
-
 def _safe_get_unstructured_api_key() -> str | None:
     try:
         return get_unstructured_api_key()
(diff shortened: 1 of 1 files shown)


backend/onyx/file_processing/extract_file_text.py
```diff
@@ -169,6 +169,23 @@ def to_bytesio(stream: IO[bytes]) -> BytesIO:
     return BytesIO(data)
 
 
+def _has_pdf_header(sample: bytes) -> bool:
```
Weves: thoughts on:

```
def _has_pdf_header(sample: bytes) -> bool:
    whitespace_chars = b"\x00\x09\x0A\x0C\x0D\x20"
    return sample.lstrip(whitespace_chars).startswith(b"%PDF")
```

feels simpler / easier to understand

backend/onyx/llm/utils.py
```diff
@@ -204,15 +221,61 @@ def _build_content(
         try:
             file_content = file.content.decode("utf-8")
         except UnicodeDecodeError:
-            # Try to decode as binary
-            try:
-                file_content, _, _ = read_pdf_file(io.BytesIO(file.content))
```
Weves, line 209: can we move the logic into this func?

justin-tahara: Rereading the function we should not rely on this function since this can be used for all of the file types

PR #6033, opened 2025-11-02
fix: too many redirects

marcinkubica:
## Description

Fixes "too many redirects" error when upgrading from Onyx 2.0.3 to 2.1.2 with disabled authentication.

**Root Cause:** The `ANONYMOUS_USER_ENABLED` environment variable was never read by the backend code, causing the frontend to receive incorrect metadata (`{"anonymousUserEnabled": false}`) and redirect to `/auth/login` despite authentication being disabled.

**Changes Made:**
1. Added `ANONYMOUS_USER_ENABLED` environment variable reading in [app_configs.py](cci:7://file:///Users/marcin/Projects/_marcin/onyx/backend/onyx/configs/app_configs.py:0:0-0:0)
2. Updated [settings/store.py](cci:7://file:///Users/marcin/Projects/_marcin/onyx/backend/onyx/server/settings/store.py:0:0-0:0) to use the environment variable as default instead of hardcoded `False`
3. Fixed Redis initialization to store the environment variable value

**Impact:** Restores proper functionality for deployments using `AUTH_TYPE=disabled` with `ANONYMOUS_USER_ENABLED=true`.

## How Has This Been Tested?

1. **Environment Setup:** Kubernetes deployment with Onyx 2.1.2, `AUTH_TYPE=disabled`, `ANONYMOUS_USER_ENABLED=true`
2. **Pre-Fix Verification:** Confirmed redirect loop - endpoints returned 307 redirects to `/auth/login`
3. **Code Analysis:** Pinpointed missing environment variable handling in backend settings initialization
4. **Fix Implementation:** Added proper environment variable reading and Redis initialization
5. **Post-Fix Testing:** Verified API returns correct metadata and frontend no longer redirects
6. **Configuration Validation:** Confirmed both environment variables and Redis values are properly synchronized

---
## Summary by cubic
Fixes a redirect loop after upgrading to 2.1.2 when auth is disabled by correctly honoring ANONYMOUS_USER_ENABLED and syncing it with Redis. Restores anonymous access so the app no longer redirects to /auth/login.

- **Bug Fixes**
  - Read ANONYMOUS_USER_ENABLED from env in app_configs.py.
  - Use the env value as the default when Redis is empty; persist it as 1/0.
  - On Redis errors, fall back to the env value instead of hardcoded false.

<sup>Written for commit 155ae103eee3cdd68ee0c745e6777b603e9e4272. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/configs/app_configs.py
+++ b/backend/onyx/configs/app_configs.py
@@ -36,6 +36,9 @@ DISABLE_GENERATIVE_AI = os.environ.get("DISABLE_GENERATIVE_AI", "").lower() == "
 # Controls whether users can use User Knowledge (personal documents) in assistants
 DISABLE_USER_KNOWLEDGE = os.environ.get("DISABLE_USER_KNOWLEDGE", "").lower() == "true"
 
+# Controls whether anonymous user access is enabled when auth is disabled
+ANONYMOUS_USER_ENABLED = os.environ.get("ANONYMOUS_USER_ENABLED", "").lower() == "true"
+
 # If set to true, will show extra/uncommon connectors in the "Other" category
 SHOW_EXTRA_CONNECTORS = os.environ.get("SHOW_EXTRA_CONNECTORS", "").lower() == "true"
 
--- a/backend/onyx/server/settings/store.py
+++ b/backend/onyx/server/settings/store.py
@@ -1,6 +1,7 @@
 from onyx.configs.app_configs import DISABLE_USER_KNOWLEDGE
 from onyx.configs.app_configs import ONYX_QUERY_HISTORY_TYPE
 from onyx.configs.app_configs import SHOW_EXTRA_CONNECTORS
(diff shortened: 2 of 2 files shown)


wenxi-onyx:
@marcinkubica Thanks for the contribution! I think this is a good addition. 

Note, I believe the intention was not to read this via env. Do you have the anonymous users enabled checkbox ticked in the Admin workspace settings page? I believe the value is added to the settings via that page?

marcinkubica:
Hi @wenxi-onyx 
In short deploying 2.1.2 from scratch (ie.with docker compose) will give "too many redirects" so users can't tick anything yet (and anonymous users is enabled by default).

wenxi-onyx:
@marcinkubica Maybe we continue this convo in discord or slack (lmk if you need links) --> I just launched v2.1.2 completely fresh and did not get this issue.

PR #6048, opened 2025-11-04
enhancement(onboarding) : Replacing Select input with combobox

Subash-Mohan:
## Description
This pull request introduces a new, highly flexible `InputComboBox` component to the frontend, designed to replace and extend the functionality of the existing input/select controls. It provides a unified input and filterable dropdown experience with strong accessibility, validation, and integration features. The implementation includes modular subcomponents for dropdown rendering and option management, and robust keyboard/mouse interaction handling.

**Backend Bug Fix:**

* Fixes a bug in `model_supports_image_input` in `llm/utils.py` to properly handle cases where `supports_vision` is `None`, ensuring a boolean is always returned.

## How Has This Been Tested?
Tested from UI

## Additional Options

- [] [Optional] Override Linear Check

---
## Summary by cubic
Replaced onboarding model selects with a new InputComboBox that supports typing to filter, keyboard navigation, and better validation/accessibility. Keeps InputSelect API compatibility for a smooth adoption.

- **New Features**
  - Added InputComboBox (dual input/select) with type-to-filter, strict mode, ARIA support, and collision-aware portal positioning via useDropdownPosition.
  - Supports both onChange and onValueChange, integrates with FormField, and includes inline validation with onValidationError.
  - Updated DynamicProviderFields and LLMConnectionFieldsBasic to use InputComboBox; simplified flow by removing API-key blur/model-change test triggers; kept the “Fetch models” refresh button.
  - Introduced FieldMessage for consistent inline feedback; integrated into FormField and InputKeyValue. Enhanced useClickOutside to handle multiple refs.
  - Improved modal and onboarding UX: dropdown clicks no longer close modals, Enter submits ProviderModal, and onboarding/provider checks use hasAnyProvider with liveAssistant-aware useOnboardingState.

- **Bug Fixes**
  - model_supports_image_input now handles supports_vision=None and always returns a boolean.

<sup>Written for commit 7188d40da92a761915b285b998e4f8c2c3dfe7a4. Summary will update automatically on new commits.</sup>

### DIFF
--- a/web/src/refresh-components/onboarding/useOnboardingState.ts
+++ b/web/src/refresh-components/onboarding/useOnboardingState.ts
@@ -66,17 +66,10 @@ export function useOnboardingState(liveAssistant?: MinimalPersonaSnapshot): {
         type: OnboardingActionType.UPDATE_DATA,
         payload: { llmProviders: llmProviders.map((p) => p.provider) },
       });
-      if (userName) {
-        dispatch({
-          type: OnboardingActionType.GO_TO_STEP,
-          step: OnboardingStep.Complete,
-        });
-      } else {
-        dispatch({
-          type: OnboardingActionType.GO_TO_STEP,
-          step: OnboardingStep.Name,
-        });
-      }
+      dispatch({
+        type: OnboardingActionType.GO_TO_STEP,
+        step: OnboardingStep.Complete,
+      });
       return;
     }
     if (userName && state.currentStep === OnboardingStep.Welcome) {


web/src/refresh-components/inputs/InputComboBox/components/OptionItem.tsx
raunakab, line 1: How does this component differ from `LineItem`? I think @duo-onyx mentioned that we should stick to using the existing components instead of creating new ones.

web/src/refresh-components/inputs/InputComboBox/hooks.ts
```diff
@@ -0,0 +1,363 @@
...
+    if (timeout) clearTimeout(timeout);
+    timeout = setTimeout(() => func.apply(this, args), wait);
+  } as T & { cancel: () => void };
+
+  debounced.cancel = () => {
+    if (timeout) clearTimeout(timeout);
+  };
+
+  return debounced;
+}
```
raunakab: Do we not already have a `debounce` function? Or should we use `lodash`'s `debounce` function (i.e., `import { debounce } from "lodash"`).

web/src/refresh-components/inputs/InputComboBox/hooks.ts
```diff
@@ -0,0 +1,363 @@
...
+    };
+
+    if (isOpen) {
+      document.addEventListener("mousedown", handleClickOutside);
+      return () => {
+        document.removeEventListener("mousedown", handleClickOutside);
+      };
+    }
+  }, [isOpen, refs, onClickOutside]);
+}
```
raunakab: Do we have an existing function for detecting if a `click-outside`?

If we don't, seems like this function would be better inside of `refresh-components/hooks.ts` since it's more universal (living inside of `refresh-components/inputs/components/hooks.ts` makes it seem as if this function is specialized to only inputs).

web/src/refresh-components/onboarding/components/DynamicProviderFields.tsx
raunakab, line 1: Hm, I think `refresh-components` is designed to be a universal, feature-agnostic library. Feature-specific code should ideally live inside of `web/src/sections` (that's where I've put the `Settings` dropdown stuff, for example).

raunakab: Are the components in this folder not global? Are they *only* meant to work with the onboarding feature?

web/src/refresh-components/inputs/InputComboBox/InputComboBox.tsx
```diff
@@ -0,0 +1,397 @@
...
+                  {rightSection}
+                </div>
+              )}
+              {hasOptions && (
+                <IconButton
+                  internal
+                  onClick={(e) => {
+                    e.stopPropagation();
+                    toggleDropdown();
+                  }}
```
raunakab: We can use the `noProp` helper here! Usage will be like:

```tsx
onClick={noProp(toggleDropdown)}
```

web/src/refresh-components/inputs/InputComboBox/InputComboBox.tsx
```diff
@@ -0,0 +1,397 @@
...
+          <Text
+            id={`${fieldId}-error`}
+            text03
+            secondaryBody
+            className="ml-0.5"
+            role="alert"
+          >
+            {errorMessage}
+          </Text>
+        </div>
```
raunakab: This may be a good place to extract "error message" chips out into their own component.

web/src/refresh-components/inputs/InputComboBox/hooks.ts
raunakab, line 1: Seems like there are a lot of hooks in here which are not *specific* to inputs (they're more universal).

I think `refresh-components/hooks.ts` would be a better place for them to live!

web/src/refresh-components/onboarding/components/LLMConnectionFieldsBasic.tsx
```diff
@@ -353,68 +348,44 @@ export const LLMConnectionFieldsBasic: React.FC<Props> = ({
...
+                rightSection={
+                  canFetchModels ? (
+                    <IconButton
+                      internal
+                      icon={SvgRefreshCw}
+                      onClick={(e) => {
+                        e.preventDefault();
+                        e.stopPropagation();
+                        onFetchModels?.();
+                      }}
```
raunakab: Could use the `noProp` here as well:

```tsx
noProp((event) => {
  event.preventDefault();
  onFetchModels?.();
})
```

web/src/refresh-components/onboarding/components/LLMConnectionFieldsBasic.tsx
```diff
@@ -353,68 +348,44 @@ export const LLMConnectionFieldsBasic: React.FC<Props> = ({
...
+                      tooltip="Fetch available models"
+                      disabled={disabled || isFetchingModels}
+                      className={isFetchingModels ? "animate-spin" : ""}
```
raunakab: This will spin the entire component, which would not be very desirable. A better option would be to "supercharge" the icon which is passed in. Remember, the `icon` prop is just a function-component at the end of the day. You can pass in whatever you want, as long it's a function-component:

```tsx
<IconButton
  icon={({ className }) => <SvgRefreshCw className={cn(className, isFetchingModels && "animate-spin")}
{...}
/>
```

In fact, that above pattern is so common, I've created a `SimpleLoader` component which does exactly that. The usage is as follows:

```tsx
<IconButton
  icon={SimpleLoader}
  {...}
/>
```

(`SimpleLoader === ({ className }) => <SvgRefreshCw className={cn(className, "animate-spin")} />`).

web/src/refresh-components/inputs/InputComboBox/components/ComboBoxDropdown.tsx
raunakab, line 1: How come we're using this component instead of `Popovers`?

Subash-Mohan: We’re not using Popover for the combo box because Popover is designed for menus, not form inputs. A combo box must follow the WAI-ARIA Combobox pattern (input retains focus, role="listbox"/"option", aria-activedescendant, type‑ahead), whereas Popover implements a menu pattern (role="menu"/"menuitem") with different focus and keyboard semantics. Using Popover would break accessibility for screen readers and keyboard users.

web/src/refresh-components/inputs/InputComboBox/components/OptionItem.tsx
Subash-Mohan: I don’t want to extend the LineItem to include everything needed for this particular component. If we keep adding more responsibilities to LineItem, the component may become bloated. Even a small mistake in LineItem could cause issues across the entire application.

Also, the ComboBox is already an individual component. Keeping separate components gives us the flexibility to extend or modify them as needed, without affecting other parts of the system.

I strongly believe we need to draw a line for where LineItem should be used and where it should not. In this case, I personally feel that we need a different component.

What do you think?

web/src/refresh-components/inputs/InputComboBox/hooks.ts
Subash-Mohan: Now I moved the hooks that feels universal to different file under hooks folder.

web/src/refresh-components/inputs/InputComboBox/hooks.ts
Subash-Mohan: Now we use the debounce method from lodash.

web/src/refresh-components/inputs/InputComboBox/hooks.ts
Subash-Mohan: Now we use the existing clickoutside hook with small correction in it. Now the useClickOutside hook accept multiple refs.

web/src/refresh-components/onboarding/components/DynamicProviderFields.tsx
Subash-Mohan: Yeah, I will create separate PR to move these components to sections folder. Thanks for letting me know!

PR #6064, opened 2025-11-05
chore(simplified): tools

rguan72:
## Description

Makes tool responses minimal according to this spec: https://docs.google.com/document/d/1RP1_9MmbzLkF7b9zpaOSKHUVZhwoo5hQUwKyuwKRs70/edit?tab=t.0

and allows the LLM to cite web search results

## How Has This Been Tested?

updated unit tests + ran locally

## Additional Options

---
## Summary by cubic
Unifies document caching keyed by normalized URLs or document IDs across internal and web search to reduce duplication and stabilize citations. Simplifies tool outputs and updates return types/JSON schemas for internal_search, web_search, and open_url, with centralized URL normalization.

- **Refactors**
  - InternalSearch returns list[LlmInternalSearchResult]; WebSearch returns list[LlmWebSearchResult]; OpenUrl returns list[LlmOpenUrlResult]. All include document_citation_number, unique_identifier_to_strip_away, and a type discriminator.
  - Added fetched_documents_cache to ChatTurnContext for dedup and citation tracking; removed unordered_fetched_inference_sections and ordered_fetched_documents; save_turn now maps citations directly from the cache.
  - Centralized normalize_url in onyx.utils.url and adopted in web tools and DR summarize.
  - Citation handler now decodes mixed tool result types, assigns/reuses numbers from the cache, and strips unique_identifier_to_strip_away and type from LLM-facing outputs; FastChatTurn builds context docs from the cache for streaming; updated unit tests.
  - Switched to Pydantic TypeAdapter for JSON output; web_search now emits an initial SearchToolDelta with queries before documents.

- **Migration**
  - Consumers must parse discriminated JSON arrays of the new simplified result models instead of LlmDoc/WebSearchResponse.
  - Use unique_identifier_to_strip_away (normalized URL or document_id) as the stable key for dedup/citation mapping; note it and type may be removed from final tool outputs after citation processing.

<sup>Written for commit aa5fea84ec2e594a14b4a6074e907620491de71a. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/tools/tool_implementations_v2/web.py
+++ b/backend/onyx/tools/tool_implementations_v2/web.py
@@ -216,19 +216,15 @@ def _open_url_core(
         for doc in docs
     ]
     for doc in docs:
-        if doc.link not in run_context.context.fetched_documents_cache:
-            run_context.context.fetched_documents_cache[doc.link] = (
-                FetchedDocumentCacheEntry(
-                    inference_section=dummy_inference_section_from_internet_content(
-                        doc
-                    ),
-                    document_citation_number=DOCUMENT_CITATION_NUMBER_EMPTY_VALUE,
-                )
-            )
-        else:
-            run_context.context.fetched_documents_cache[doc.link].inference_section = (
-                dummy_inference_section_from_internet_content(doc)
-            )
+        cache = run_context.context.fetched_documents_cache
+        entry = cache.setdefault(
+            doc.link,
+            FetchedDocumentCacheEntry(
+                inference_section=dummy_inference_section_from_internet_content(doc),
+                document_citation_number=DOCUMENT_CITATION_NUMBER_EMPTY_VALUE,
+            ),
+        )
+        entry.inference_section = dummy_inference_section_from_internet_content(doc)
     run_context.context.iteration_instructions.append(
         IterationInstructions(
             iteration_nr=index,


backend/onyx/chat/turn/context_handler/citation.py
```diff
@@ -39,44 +54,70 @@ def assign_citation_numbers_recent_tool_calls(
...
+                tool_call_results = _decode_tool_call_result(content)
 
-                if llm_docs:
+                if tool_call_results:
                     updated_citation_number = False
-                    for doc in llm_docs:
+                    for result in tool_call_results:
                         if (
-                            doc.document_citation_number
+                            result.unique_identifier_to_strip_away is not None
```
evan-onyx, line 63: since we added a bunch of code in here, I kind of like the if not condition: continue pattern here to reduce indentation

backend/onyx/chat/turn/context_handler/citation.py
```diff
@@ -39,44 +54,70 @@ def assign_citation_numbers_recent_tool_calls(
...
-        new_llm_docs=new_llm_docs,
     )
+
+
+def _decode_tool_call_result(
+    content: str,
+) -> list[LlmInternalSearchResult | LlmOpenUrlResult | LlmWebSearchResult]:
+    try:
+        return _tool_result_adapter.validate_json(content)
+    except ValidationError:
```
evan-onyx, line 123: maybe we should have an error log here? So we know when this happens at least

backend/onyx/tools/tool_implementations_v2/web.py
```diff
@@ -89,23 +94,49 @@ def _web_search_core(
...
-            WebSearchResult(
+            LlmWebSearchResult(
+                document_citation_number=DOCUMENT_CITATION_NUMBER_EMPTY_VALUE,
+                url=r.link,
                 title=r.title,
-                link=r.link,
                 snippet=r.snippet or "",
-                author=r.author,
-                published_date=r.published_date,
+                unique_identifier_to_strip_away=normalize_url(r.link),
```
evan-onyx: nit: store the normalized result rather than calling the func 3 times

backend/onyx/tools/tool_implementations_v2/web.py
```diff
@@ -175,7 +208,28 @@ def _open_url_core(
...
+        LlmOpenUrlResult(
+            document_citation_number=DOCUMENT_CITATION_NUMBER_EMPTY_VALUE,
+            content=truncate_search_result_content(doc.full_content),
+            unique_identifier_to_strip_away=normalize_url(doc.link),
+        )
+        for doc in docs
+    ]
+    for doc in docs:
+        if normalize_url(doc.link) not in run_context.context.fetched_documents_cache:
+            run_context.context.fetched_documents_cache[normalize_url(doc.link)] = (
```
evan-onyx: can we make this cleaner with a defaultdict

backend/onyx/chat/turn/context_handler/citation.py
rguan72: for this case prefer not to log since not all tool calls will be in this format, and when we get the validation error it's expected [it's likely just not a tool call for that]

PR #6084, opened 2025-11-06
feat(testrail): Implement a read-only custom connector for Testrail

sashank-rayapudi-ai:
## Description

TestRail Connector: Read-only connector that indexes test cases from TestRail projects

Features:

- Fetches from /api/v2/get_cases with pagination, automatically maps field IDs to labels via /api/v2/get_case_fields
- Indexes: case metadata, test content (preconditions, steps with expected results), custom fields (priority, team, automation, environments, etc.)
- Steps fallback: Uses custom_steps_separated or falls back to legacy custom_steps + custom_expected for compatibility
- Project filtering: Optional comma-separated project IDs field in UI to index specific projects
- Configurable: Pagination size, max pages, document size limits
- HTML sanitization and clean text extraction

Screenshot 2025-11-20 at 2 18 37 PM

Screenshot 2025-11-21 at 2 28 16 AM

## How Has This Been Tested?

Tested locally

## Additional Options

---
## Summary by cubic
Implements a read-only TestRail connector to index test cases into Onyx with clean text extraction, project filtering, and paginated sync. Adds full UI and credential support plus clear HTTP error handling.

- **New Features**
  - Backend: project/suite discovery, optional project ID filtering, paginated case fetch with time-window polling, per-case links, field ID→label mapping, HTML→text sanitization (removes images and markdown image refs), steps/expected parsing, case_key metadata, and a configurable size guard.
  - App & Web: registers TESTRAIL source and icon; new connector form with Project IDs and advanced options (cases_page_size, max_pages, skip_doc_absolute_chars); credentials (testrail_base_url, testrail_username, testrail_api_key).
  - Errors: 401→credential expired, 403→insufficient permissions; surfaces unexpected validation errors.

<sup>Written for commit 3996727515c2b2b730b108262c562f2d4dc6edb9. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/connectors/testrail/connector.py
+++ b/backend/onyx/connectors/testrail/connector.py
@@ -69,8 +69,13 @@ class TestRailConnector(LoadConnector, PollConnector):
         self.batch_size = batch_size
 
         # Parse project_ids from string if needed
-        if isinstance(project_ids, str) and project_ids.strip():
-            self.project_ids = [int(x.strip()) for x in project_ids.split(",") if x.strip()]
+        # None = all projects (no filtering), [] = no projects, [1,2,3] = specific projects
+        if isinstance(project_ids, str):
+            if project_ids.strip():
+                self.project_ids = [int(x.strip()) for x in project_ids.split(",") if x.strip()]
+            else:
+                # Empty string from UI means "all projects"
+                self.project_ids = None
         else:
             self.project_ids = project_ids
 
@@ -450,7 +455,8 @@ class TestRailConnector(LoadConnector, PollConnector):
 
         for project in projects:
             project_id = project.get("id")
-            if project_filter and project_id not in project_filter:
+            # None = index all, [] = index none, [1,2,3] = index only those
+            if project_filter is not None and project_id not in project_filter:
                 continue
 
             suites = self._list_suites(project_id)


backend/onyx/connectors/testrail/connector.py
```diff
@@ -0,0 +1,385 @@
...
+            return ""
+        html = str(value)
+        # Replace explicit data:image URIs with a marker to avoid huge inlined payloads
+        html = re.sub(
+            r"data:image/[^\"' )]+;base64,[A-Za-z0-9+/=]+", "[image removed]", html
+        )
+        soup = BeautifulSoup(html, "html.parser")
+        # Drop <img> tags entirely
+        for img in soup.find_all("img"):
+            img.decompose()
```
justin-tahara: Is there any other potential things that we would need to sanitize for like Gifs or videos? I just want to make sure that we handle as much as we can.

backend/onyx/connectors/testrail/connector.py
```diff
@@ -0,0 +1,385 @@
...
+    # --- Rich text sanitization helpers ---
+    @staticmethod
+    def _sanitize_rich_text(value: Any) -> str:
+        if value is None:
+            return ""
+        html = str(value)
+        # Replace explicit data:image URIs with a marker to avoid huge inlined payloads
+        html = re.sub(
+            r"data:image/[^\"' )]+;base64,[A-Za-z0-9+/=]+", "[image removed]", html
+        )
```
justin-tahara: Ideally we make this global and then call it in here. I like this pattern so if we can use it other places I would like that

backend/onyx/connectors/testrail/connector.py
```diff
@@ -0,0 +1,385 @@
...
+        # Use shared formatter to produce clean, structured text
+        text = format_document_soup(soup)
+        # Unescape any remaining HTML entities just in case
+        return html_lib.unescape(text)
+
+    def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None:
+        # Expected keys from UI credential JSON
+        self.base_url = str(credentials["testrail_base_url"]).rstrip("/")
+        self.username = str(credentials["testrail_username"])  # email or username
+        self.api_key = str(credentials["testrail_api_key"])  # API key (password)
```
justin-tahara, line 128: Is there a concept of api key when doing auth? I remember there only being a password. If so can we just call this a password to keep it simple?

backend/onyx/configs/constants.py
```diff
@@ -618,4 +619,5 @@ class OnyxCallTypes(str, Enum):
     DocumentSource.AIRTABLE: "airtable - database",
     DocumentSource.HIGHSPOT: "highspot - CRM data",
     DocumentSource.IMAP: "imap - email data",
+    DocumentSource.TESTRAIL: "testrail - test management data",
```
justin-tahara: Maybe we can make this a bit more constructive like for QA or something just to give more detail.

backend/onyx/connectors/testrail/connector.py
```diff
@@ -0,0 +1,385 @@
...
+    def _list_projects(self) -> list[dict[str, Any]]:
+        projects = self._api_get("get_projects")
+        if isinstance(projects, list):
+            return projects
+        if isinstance(projects, dict):
+            # Handle wrapped responses like {"projects": [...]} or {"items": [...]} or {"data": [...]}
+            for key in ("projects", "items", "data", "results"):
+                val = projects.get(key)
+                if isinstance(val, list):
+                    return val
```
justin-tahara: Does the api return return different types? Bit odd but I wonder if we can do something cleaner. 

For the dict path, if you ever find a val that is a project it exits early. Is this intended?

backend/onyx/connectors/testrail/connector.py
```diff
@@ -0,0 +1,385 @@
...
+        gracefully fallback to calling get_cases without suite_id.
+        """
+        suites = self._api_get(f"get_suites/{project_id}")
+        if isinstance(suites, list):
+            return suites
+        if isinstance(suites, dict):
+            for key in ("suites", "items", "data", "results"):
+                val = suites.get(key)
+                if isinstance(val, list):
+                    return val
```
justin-tahara: Same as above

backend/onyx/connectors/testrail/connector.py
```diff
@@ -0,0 +1,385 @@
...
+
+    def _iter_cases(
+        self,
+        project_id: int,
+        suite_id: Optional[int] = None,
+        start: Optional[SecondsSinceUnixEpoch] = None,
+        end: Optional[SecondsSinceUnixEpoch] = None,
+    ) -> Iterator[dict[str, Any]]:
+        # Pagination: TestRail supports 'limit' and 'offset' for many list endpoints
+        limit = 250
```
justin-tahara: Can we make this a global variable that is configurable? also where did you get this 250 limit from?

backend/onyx/connectors/testrail/connector.py
```diff
@@ -0,0 +1,385 @@
...
+        offset = 0
+        while True:
```
justin-tahara: Ideally we avoid infinite loops

backend/onyx/connectors/testrail/connector.py
```diff
@@ -0,0 +1,385 @@
...
+        case_key = f"C{case_id}" if case_id is not None else None
+        suite_id = suite.get("id") if suite else None
+        suite_name = suite.get("name") if suite else None
+        section_name = case.get("section_id")
+
+        # Convert epoch seconds to aware datetime if available
+        updated = case.get("updated_on") or case.get("created_on")
+        updated_dt = (
+            datetime.fromtimestamp(updated, tz=timezone.utc) if isinstance(updated, (int, float)) else None
+        )
```
justin-tahara, line 358: If we are okay with this implementation I think this is fine but it can probably result in some wonky issues like if you ask what issues were created in the past hour or something

backend/onyx/connectors/testrail/connector.py
```diff
@@ -0,0 +1,385 @@
...
+            text_lines.append(f"Steps: {steps}")
+        exp = self._sanitize_rich_text(case.get("custom_expected"))
+        if exp:
+            text_lines.append(f"Expected: {exp}")
+
+        link = self._build_case_link(project_id, case_id)
+
+        # Build full text and apply size policies
+        full_text = "\n".join(text_lines)
+        if len(full_text) > SKIP_DOC_ABSOLUTE_CHARS:
```
justin-tahara: Are we doing an explicit text string length difference? 

I wonder what the best way to do this is.

justin-tahara: Ideally we can make this configurable as well by the end user

backend/onyx/connectors/testrail/connector.py
```diff
@@ -0,0 +1,385 @@
...
+        if suite_id is not None:
+            doc_metadata["suite_id"] = str(suite_id)
+        if suite_name is not None:
+            doc_metadata["suite_name"] = str(suite_name)
+
+        # Include the human-friendly case key in identifiers for easier search
+        display_title = f"{case_key}: {title}" if case_key else title
+
+        # Apply truncation if needed
+        if len(full_text) > MAX_DOC_CHARS:
```
justin-tahara: So this is the max if its under the `SKIP_DOC_ABSOLUTE_CHARS`? Do we want this behavior?

backend/onyx/connectors/testrail/connector.py
```diff
@@ -0,0 +1,385 @@
...
+
+
+
+if __name__ == "__main__":
+    # Simple local test harness similar to other connectors.
+    # Env vars (examples):
+    #   TESTRAIL_BASE_URL=https://yourcompany.testrail.io
+    #   TESTRAIL_USERNAME=you@example.com
+    #   TESTRAIL_API_KEY=your_api_key_or_password
+    #   TESTRAIL_PROJECT_ID=123            (optional)
```
justin-tahara: I think we would need to update the `app_configs.py` if we do want to enable something like this in the future.

backend/onyx/connectors/testrail/connector.py
```diff
@@ -0,0 +1,385 @@
...
+    import os
+
+    base = os.environ.get("TESTRAIL_BASE_URL", "").strip()
+    user = os.environ.get("TESTRAIL_USERNAME", "").strip()
+    key = os.environ.get("TESTRAIL_API_KEY", "").strip()
+    project_id = os.environ.get("TESTRAIL_PROJECT_ID")
```
justin-tahara: I personally don't really like this since this kind of logic should be handled in the app_config.py

justin-tahara: Doesn't this also mean that the credentials that are set at the start are never used?

web/src/lib/connectors/credentials.ts
```diff
@@ -267,6 +267,12 @@ export interface ImapCredentialJson {
   imap_password: string;
 }
 
+export interface TestRailCredentialJson {
+  testrail_base_url: string;
+  testrail_username: string;
+  testrail_api_key: string;
```
justin-tahara, line 273: Same question as before. If this is just the password, let's make it just the password and not an API setup

backend/onyx/connectors/testrail/connector.py
sashank-rayapudi-ai: Yes! I was able to use the connector with both the api key and password. I'll leave it as api key for now to follow best practices but I can change it to password if needed.

web/src/lib/connectors/credentials.ts
sashank-rayapudi-ai: https://github.com/onyx-dot-app/onyx/pull/6084#discussion_r2535887239

backend/onyx/connectors/testrail/connector.py
sashank-rayapudi-ai: I cleaned this up, I was initially handling the return types defensively.

backend/onyx/connectors/testrail/connector.py
sashank-rayapudi-ai: 250 is the default response size: https://support.testrail.com/hc/en-us/articles/7077292642580-Cases#getcases

backend/onyx/connectors/testrail/connector.py
sashank-rayapudi-ai: I won't be storing testrail case results so I think current implementation is fine

backend/onyx/connectors/testrail/connector.py
sashank-rayapudi-ai: All Gifs are stored as `img` so we are covered there. I also added some code to remove `span` tags and added a `remove_markdown_image_references` function in `file_processing.py`

backend/onyx/connectors/testrail/connector.py
sashank-rayapudi-ai: I don't think we actually need this pattern since I don't see it in any of the indexed docs, going to remove this

backend/onyx/connectors/testrail/connector.py
sashank-rayapudi-ai: I removed this truncation logic since Onyx's chunker handles this in [backend/onyx/indexing/chunker.py](https://github.com/onyx-dot-app/onyx/blob/main/backend/onyx/indexing/chunker.py)

backend/onyx/connectors/testrail/connector.py
sashank-rayapudi-ai: I added a few fields to `connectors.tsx` to make this configurable

backend/onyx/connectors/testrail/connector.py
sashank-rayapudi-ai: Updated this logic to live in app_config.py

I initially had this here since I was following the logic for the confluence connector: https://github.com/onyx-dot-app/onyx/blob/main/backend/onyx/connectors/confluence/connector.py#L795

backend/onyx/connectors/testrail/connector.py
justin-tahara: Probs worth adding as a detail

justin-tahara (approved):

PR #6092, opened 2025-11-06
fix: setDisplayComplete not called for ollama

Weves:
## Description

We weren't marking non-tool indices as complete when a new index # is seen. I believe it's safe to assume this for all packet types.

## How Has This Been Tested?

tested locally

## Additional Options


web/src/app/chat/message/messageComponents/AIMessage.tsx
```diff
@@ -362,19 +352,6 @@ export default function AIMessage({
   const updateCurrentSelectedNodeForDocDisplay = useChatSessionStore(
     (state) => state.updateCurrentSelectedNodeForDocDisplay
   );
-  // Calculate unique source count
```
Weves, line 365: this was unused

PR #6098, opened 2025-11-06
fix(onboarding): Header animated icon

Subash-Mohan:
## Description
This pull request refactors the onboarding flow UI and logic for LLM provider setup, focusing on improved code organization, maintainability, and user experience. The main changes include extracting the `LLMFormikEffects` component into its own file, introducing a progress indicator for onboarding steps, and standardizing LLM provider references. These updates enhance clarity in both the codebase and the onboarding interface.

## How Has This Been Tested?
Tested from UI

## Additional Options

---
## Summary by cubic
Replaced the onboarding header’s static step icon with an animated circular progress indicator and standardized LLM provider handling, including a new ProviderModal, to improve clarity and maintainability.

- **UI Improvements**
  - Added ProgressSteps for a smooth, animated step indicator.
  - Header now uses percentage (STEP_CONFIG.iconPercentage) and shows a check on completion.
  - Introduced InlineExternalLink and replaced the modal layout with a new ProviderModal; tall modals now have a max-height to prevent overflow.

- **Refactors**
  - Extracted LLMFormikEffects into its own component and wired it in LLMConnectionModal.
  - Added LLMProviderName enum and updated constants, maps, and effects to use it.
  - Integrated ProviderModal with modal state via LLMStep/LLMProvider and removed useModal in LLMConnectionModal.

<sup>Written for commit 27829e9af8eb8945480f1c56f31649efffd066b6. Summary will update automatically on new commits.</sup>

### DIFF
--- a/web/src/refresh-components/onboarding/components/LLMProvider.tsx
+++ b/web/src/refresh-components/onboarding/components/LLMProvider.tsx
@@ -12,6 +12,7 @@ import SvgCheckCircle from "@/icons/check-circle";
 import { OnboardingActions, OnboardingState } from "../types";
 import { cn, noProp } from "@/lib/utils";
 import { LLMConnectionModalProps } from "./LLMConnectionModal";
+import { ModalCreationInterface } from "@/refresh-components/contexts/ModalContext";
 
 export interface LLMProviderProps {
   title: string;
@@ -24,6 +25,7 @@ export interface LLMProviderProps {
   onboardingState: OnboardingState;
   onboardingActions: OnboardingActions;
   onOpenModal?: () => void;
+  modal: ModalCreationInterface;
 }
 
 function LLMProviderInner({
@@ -37,6 +39,7 @@ function LLMProviderInner({
   onboardingActions,
   onClick,
   onOpenModal,
+  modal,
 }: LLMProviderProps) {
   const [isHovered, setIsHovered] = useState(false);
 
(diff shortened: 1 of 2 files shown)


web/src/refresh-components/inputs/ProgressSteps.tsx
raunakab, line 1: This seems like a nice generic component. Could we move this to `refresh-components/inputs` instead? Seems like this would live better over there.

I'm sure that there are more components in the future that could utilize this exact component for themselves, so having `ProgressSteps` live inside of `onboarding` doesn't make too much sense.

web/src/refresh-components/onboarding/constants.tsx
```diff
@@ -93,166 +95,150 @@ export const PROVIDER_ICON_MAP: Record<
...
+  <a
+    href={href}
+    target="_blank"
+    rel="noopener noreferrer"
+    className={className ?? "underline"}
+  >
+    {children}
+  </a>
+);
+
```
raunakab: Could we move this to its own separate component too, please? Seems like it would be very useful as its own isolated component!

web/src/refresh-components/onboarding/components/LLMFormikEffects.tsx
raunakab, line 1: Hmm, not super confident on having form based stuff live inside of `refresh-components`, especially since `refresh-components` is supposed to be an ad-hoc components library.

We can merge this in for now, but we should think of a better place for this logic to live in the coming future.

Subash-Mohan: Yeah, I’ll move the components related to onboarding to the section directory once it goes in.

PR #6128, opened 2025-11-07
chore(llm): interface decoupled from langchain

rguan72:
## Description

Decoupling the LLM interface from Langchain so we can use this class for non langchain purposes

## How Has This Been Tested?

Tested locally + unit tests

## Additional Options

---
## Summary by cubic
Decoupled the LLM interface from LangChain and introduced a generic ModelResponse API backed by LiteLLM. Flows and tests now call explicit LangChain methods via invoke_langchain/stream_langchain.

- **Refactors**
  - Added invoke_langchain/stream_langchain and updated call sites across the codebase.
  - Implemented generic invoke/stream in ChatLLM using LiteLLM; kept LangChain-specific implementations as _invoke_implementation_langchain/_stream_implementation_langchain.
  - Introduced onyx.llm.model_response with converters from LiteLLM; generic calls return ModelResponse/ModelResponseStream.
  - Removed legacy CustomLLM.
  - Added backend/onyx/evals/one_off/*.json to .gitignore.

- **Migration**
  - Use llm.invoke_langchain and llm.stream_langchain for LangChain-based code.
  - For non-LangChain backends, implement the generic invoke/stream that return ModelResponse/ModelResponseStream.

<sup>Written for commit b0b5b3adfae03809642948f7311dacd00d532ce7. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/tests/unit/onyx/indexing/test_indexing_pipeline.py
+++ b/backend/tests/unit/onyx/indexing/test_indexing_pipeline.py
@@ -306,7 +306,7 @@ def test_contextual_rag(
     mock_llm.config.max_input_tokens = get_max_input_tokens(
         model_provider="openai", model_name="gtp-4o"
     )
-    mock_llm.invoke = mock_llm_invoke
+    mock_llm.invoke_langchain = mock_llm_invoke
 
     chunker = Chunker(
         tokenizer=embedder.embedding_model.tokenizer,


backend/onyx/llm/interfaces.py
```diff
@@ -96,11 +97,30 @@ def invoke(
         structured_response_format: dict | None = None,
         timeout_override: int | None = None,
         max_tokens: int | None = None,
+    ) -> None:
+        self._invoke_implementation(
```
evan-onyx: why are we calling this but not returning?

rguan72: good catch

PR #6152, opened 2025-11-10
feat(gha): python tests use uv w/ caching

jmelahman:
## Description

Migrates the python unit tests to use `uv` with runs-on caching. Also reduces the size of the instance runner from 8 cores to 2 since pytest runs in serial by default and the test is still pretty small.

## How Has This Been Tested?

Captured by presubmit

## Additional Options

- [x] Override Linear Check

---
## Summary by cubic
Switch CI Python installs to uv with caching to speed up test runs and standardize setup. Adds a reusable composite action used by pr-python-tests and prepare-build.

- **Refactors**
  - Added .github/actions/setup-python-and-install-dependencies to set up Python 3.11, install deps via uv, and cache ~/.cache/uv.
  - Updated pr-python-tests and prepare-build to use the new composite action; switched pr-python-tests to a 2cpu arm64 runner and enabled CPU/memory metrics.
  - Note: setup-uv’s built-in cache remains off until a uv.lock is added; caching uses runs-on/cache for now.

<sup>Written for commit 65bcff0a450de38fabdcbbd14dc79f49e80247c9. Summary will update automatically on new commits.</sup>

### DIFF
--- a/.github/workflows/pr-python-tests.yml
+++ b/.github/workflows/pr-python-tests.yml
@@ -22,8 +22,6 @@ jobs:
       
     steps:
     - uses: runs-on/action@cd2b598b0515d39d78c38a02d529db87d2196d1e # ratchet:runs-on/action@v2
-      with:
-        metrics: cpu,memory 
 
     - name: Checkout code
       uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # ratchet:actions/checkout@v4


.github/workflows/pr-python-tests.yml
```diff
@@ -20,25 +21,15 @@ jobs:
       SF_SECURITY_TOKEN: ${{ secrets.SF_SECURITY_TOKEN }}
       
     steps:
+    - uses: runs-on/action@cd2b598b0515d39d78c38a02d529db87d2196d1e # ratchet:runs-on/action@v2
+      with:
+        metrics: cpu,memory
```
jmelahman:
```suggestion
```

PR #6180, opened 2025-11-12
chore(fix): agent sdk replacement message formatting

rguan72:
## Description

Add message format for agent sdk framework replacement
Also fix Azure form validation

## How Has This Been Tested?

Tested locally

## Additional Options

---
## Summary by cubic
Replaces the agent SDK with a lightweight agent framework, adds stateful message building with streaming events, and moves the LLM to OpenAI Responses-style messages with reasoning.

- **New Features**
  - query() now returns QueryResult with a stream and new_messages_stateful (assistant/tool messages), and enforces reasoning_done before message/tool events.
  - Added ChatCompletion message types and a base_messages_to_chat_completion_msgs converter.
  - Upgraded LitellmLLM with tool_choice, parallel_tool_calls, and reasoning support; logs both legacy and new prompt formats; comprehensive tests for streaming, parallel tool calls, and event ordering.

- **Refactors and Migration**
  - LLM interfaces accept LanguageModelInput (ChatCompletionMessage[] or str) while keeping legacy LangChain input compatible; query() accepts LanguageModelInput.
  - LLM routes OpenAI/Azure calls to /responses to enable reasoning tokens and sets thinking budgets; frontend validation now allows /openai/responses; disables litellm function injection to preserve tool call formatting.

<sup>Written for commit 0c856ba34c75759f31c66d9655c419dbb6e3988e. Summary will update automatically on new commits.</sup>


backend/onyx/agents/agent_framework/query.py
```diff
@@ -1,19 +1,28 @@
...
+from onyx.llm.interfaces import LanguageModelInput
 from onyx.llm.interfaces import LLM
 from onyx.llm.interfaces import ToolChoiceOptions
+from onyx.llm.message_types import ChatCompletionMessage
 from onyx.llm.model_response import ModelResponseStream
 from onyx.tools.tool import RunContextWrapper
 from onyx.tools.tool import Tool
 
 
+@dataclass
```
rguan72, line 21: I'm sorry evan but there are some subtle bugs from using Pydantic BaseModel + we shouldn't be able to convert this to json anyways since an Iterator => JSON doesn't make sense

backend/onyx/llm/chat_llm.py
```diff
@@ -437,9 +508,7 @@ def _completion(
                 # streaming choice
                 stream=stream,
                 # model params
-                temperature=(
-                    1 if "gpt-5" in self.config.model_name else self._temperature
-                ),
+                temperature=(1 if is_reasoning else self._temperature),
```
yuhongsun96, line 511: Is this logic correct?

backend/onyx/llm/interfaces.py
```diff
@@ -13,13 +16,15 @@
...
 from onyx.configs.app_configs import LOG_ONYX_MODEL_INTERACTIONS
+from onyx.llm.message_types import ChatCompletionMessage
 from onyx.llm.model_response import ModelResponse
 from onyx.utils.logger import setup_logger
 
 logger = setup_logger()
 
 STANDARD_TOOL_CHOICE_OPTIONS = ("required", "auto", "none")
 ToolChoiceOptions = Union[Literal["required", "auto", "none"], str]
+LanguageModelInput = Union[Sequence[ChatCompletionMessage], str]
```
yuhongsun96, line 27: Would argue this should not allow a str type but this is ok for now

yuhongsun96 (approved):

backend/onyx/agents/agent_framework/query.py
evan-onyx: putting an iterator in a class like this is definitely spicy lol

I'm not a dataclass hater across the board, it's just for consistency with the rest of the repo. For cases like this I think they're reasonable!

backend/onyx/agents/agent_framework/query.py
```diff
@@ -1,19 +1,38 @@
...
 from onyx.llm.interfaces import ToolChoiceOptions
+from onyx.llm.message_types import ChatCompletionMessage
+from onyx.llm.message_types import ToolCall
 from onyx.llm.model_response import ModelResponseStream
 from onyx.tools.tool import RunContextWrapper
 from onyx.tools.tool import Tool
 
 
+@dataclass
+class QueryResult:
```
evan-onyx, line 22: Upon looking below and seeing how this is used, I think it's probably nicer to make this a normal Class, and move the stream_generator into it. My reasoning is having state coupling between the iterator and the messages list is a common pattern in classes that have a coherent purpose, i.e. its reasonable for instance variables in a Connector to change in correlated ways. But imo dataclasses (and BaseModels) are more for grouping data that are typically not coupled in this way.

Not a big deal really, if you strongly disagree we can leave it as is

backend/onyx/llm/chat_llm.py
```diff
@@ -399,11 +449,19 @@ def _completion(
...
         timeout_override: int | None = None,
         max_tokens: int | None = None,
+        is_legacy_langchain: bool = False,
     ) -> Union["ModelResponse", "CustomStreamWrapper"]:
         # litellm doesn't accept LangChain BaseMessage objects, so we need to convert them
         # to a dict representation
-        processed_prompt = _prompt_to_dict(prompt)
-        self._record_call(processed_prompt)
+        processed_prompt: LegacyPromptDict | LanguageModelInput
+        if is_legacy_langchain:
```
evan-onyx, line 457: inline conditional might look nicer

PR #6216, opened 2025-11-13
chore(runs-on): define custom AMI specs

jmelahman:
## Description

I think runs-on is only referring to this `.github/runs-on.yml` at HEAD, so check in [the custom image definitions](https://runs-on.com/configuration/custom-images/).

## How Has This Been Tested?

no-op

## Additional Options

- [x] Override Linear Check

---
## Summary by cubic
Migrated runs-on configuration to extend from .github.private, sourcing custom AMI specs for Onyx Ubuntu 24 full images on x64 and arm64.
Workflows still use onyx_x64_image and onyx_arm64_image.

<sup>Written for commit e0f6e80ea3d20c4f8fb1e8a5cc0c4b21d2ffc4d5. Summary will update automatically on new commits.</sup>

### DIFF
--- a/.github/runs-on.yml
+++ b/.github/runs-on.yml
@@ -1,1 +1,1 @@
-_extend: .github.private
+_extend: .github-private


justin-tahara (approved):

.github/runs-on.yml
```diff
@@ -1,4 +1 @@
-runners:
-  default:
-    family: ["m7", "c7"]
-    spot: price-capacity-optimized
+_extend: .github.private
```
jmelahman:
```suggestion
_extend: .github-private
```

PR #6244, opened 2025-11-14
fix: playwright test

Weves:
## Description

## How Has This Been Tested?

## Additional Options

---
## Summary by cubic
Isolates the Disable Default Assistant Playwright test by tagging it @exclusive and configuring Playwright/CI to run @exclusive tests serially in a separate project. This prevents state leaks and flakiness.

<sup>Written for commit 173f913b30f25c3be3b4145333a85e3d12d5a859. Summary will update automatically on new commits.</sup>

### DIFF
--- a/web/playwright.config.ts
+++ b/web/playwright.config.ts
@@ -64,6 +64,7 @@ export default defineConfig({
         storageState: "admin_auth.json",
       },
       grep: /@exclusive/,
+      workers: 1,
     },
   ],
 });


web/playwright.config.ts
```diff
@@ -44,13 +44,24 @@ export default defineConfig({
...
       },
+      grepInvert: /@exclusive/,
+    },
+    {
+      name: "tests-that-run-independently-and-serially",
+      use: {
+        ...devices["Desktop Chrome"],
+        viewport: { width: 1280, height: 720 },
+        storageState: "admin_auth.json",
+      },
```
jmelahman, line 65:
```suggestion
      },
      workers: 1,
```
do we want something like [this](https://playwright.dev/docs/api/class-testconfig#test-config-workers) to force the tests to run serially?

.github/workflows/pr-playwright-tests.yml
```diff
@@ -119,7 +119,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        project: [admin, no-auth]
+        project: [admin, no-auth, tests-that-run-independently-and-serially]
```
nmgarza5: nit: kind of prefer to keep the project name 1:1 with the tag and then add a detailed comment to the project configuration stating this suite runs independently and serially + slower and we should be cautious about bloating the suite.
```suggestion
        project: [admin, no-auth, exclusive]
```

web/playwright.config.ts
```diff
@@ -44,13 +44,24 @@ export default defineConfig({
...
+      grep: /@exclusive/,
```
nmgarza5, line 66: Need to set workers to 1 for this to run serially
```suggestion
      grep: /@exclusive/,
      workers: 1,
```

nmgarza5 (approved):

web/playwright.config.ts
Weves: added

.github/workflows/pr-playwright-tests.yml
Weves: seems good!

PR #6245, opened 2025-11-14
fix: gpt-5 output formatting

Weves:
## Description

^

## How Has This Been Tested?

## Additional Options

---
## Summary by cubic
Fixes incorrect markdown/code block formatting in responses from OpenAI reasoning models (gpt-5, gpt-5.1, o1, o3). Replaces a fragile provider check with a robust model detector to re-enable formatting when needed.

- **Bug Fixes**
  - Added model_needs_formatting_reenabled with regex matching for gpt-5/gpt-5.1/o1/o3 variants (including prefixes/suffixes and slashes).
  - Updated answer_prompt_builder to use the new detector and removed the provider-based check.
  - Added comprehensive unit tests covering positive matches and non-matches.

<sup>Written for commit 252ee1a318865603dc90ac099ff0ac9cb7210686. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/llm/utils.py
+++ b/backend/onyx/llm/utils.py
@@ -1,6 +1,7 @@
 import copy
 import io
 import json
+import re
 from collections.abc import Callable
 from collections.abc import Iterator
 from functools import lru_cache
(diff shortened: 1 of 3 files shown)


backend/onyx/llm/utils.py
```diff
@@ -914,3 +915,24 @@ def is_true_openai_model(model_provider: str, model_name: str) -> bool:
             f"Failed to determine if {model_provider}/{model_name} is a true OpenAI model"
         )
         return False
+
+
+def model_needs_formatting_reenabled(model_name: str) -> bool:
```
justin-tahara, line 920: We should probably make a similar function for the temperature setup and just do the same thing instead of the if 'gpt-5' in string or something. Can be a followup

backend/tests/unit/onyx/llm/test_formatting_reenabled.py
justin-tahara, line 1: I think if we can also do something like azure/litellm/gpt-5 for double slashes that would be nice to see in the tests.

hopefully no one does that but you never know how the model names are configured

rguan72 (approved):

backend/tests/unit/onyx/llm/test_formatting_reenabled.py
Weves: this should handle it already!

PR #6247, opened 2025-11-14
chore(internal): search prune sections

rguan72:
## Description

Uses prune_sections to burn through significantly fewer tokens on internal search :)
I don't fully understand how prune_sections works, but seems to reduce input tokens by 40-50% even when we are relatively far from the limit and it probably gets better when you're closer to the limit

## How Has This Been Tested?

Tested locally

## Additional Options

---
## Summary by cubic
Switches internal search to token-aware pruning using the chat’s current input tokens, removing the 15-doc cap and pruning/merging sections before emitting results.

- **Refactors**
  - Compute ChatTurnContext.current_input_tokens in fast_chat_turn and use it for pruning.
  - Updated prune_and_merge to accept existing_input_tokens (optional prompt_config); added compatibility in pipeline/search_tool via query token count.
  - Internal search aggregates sections from SEARCH_RESPONSE_SUMMARY_ID, prunes/merges, updates caches, and emits pruned sections; citation flag respects pruned results.

<sup>Written for commit 93b440f8da10398b26c4c0d86020203c5c243d39. Summary will update automatically on new commits.</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,11 @@
     </a>
 </p>
 
+<p align="center">
+  <a href="https://trendshift.io/repositories/12516" target="_blank">
+    <img src="https://trendshift.io/api/badge/repositories/12516" alt="onyx-dot-app/onyx | Trendshift" style="width: 250px; height: 55px;" />
+  </a>
+</p>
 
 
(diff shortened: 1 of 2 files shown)


rguan72:
@greptile

PR #6264, opened 2025-11-15
chore(migrate): to new framework

rguan72:
## Description

Cut over to the new framework :)

Reminder: figure out something weird with custom tool calls

This:
1. Rips out agent SDK so we can do a huge deletion
2. Removes Asyncio from the agent loop, fixing all kind of strange errors (and likely making something like this much more debuggable: https://onyx-company.slack.com/archives/C06V12Z46P2/p1764024286106059)

## How Has This Been Tested?

Tested locally very thoroughly, although this is going to be cursed so I will have to actively monitor braintrust
My last contribution aside from code deletion -- will need y'all's help most likely for fixing some leftover bugs, but I think this is most of my vision for the new framework

## Additional Options

---
## Summary by cubic
Migrates chat orchestration to the new agent framework, fully removes the legacy Agent SDK, and upgrades streaming/tracing. Also enables JSON-to-tool-call fallback so non-tool-calling models can execute tools, adds an Open URL tool, and expands web search with citation numbering and progress streaming.

- **New Features**
  - query(...) now accepts structured_response_format and forwards it to the LLM.
  - Auto-convert assistant JSON content into tool_calls for non-tool-calling models, with synthetic IDs.
  - Stream tool_call_output even when tools are missing (returns “Tool X not found”).
  - Patch OpenAI Responses API streaming to Chat Completions format (tool calls, text deltas, reasoning) via LiteLLM; keep and apply Ollama stream patches.
  - Add tracing: spans/traces with agent/function/generation spans, pluggable processors, error attachment, and Braintrust + OpenInference/Langfuse integration.
  - Stream token usage (prompt/completion/total/cache) via include_usage and ModelResponseStream.
  - Search tool integrates DR iteration models, streams progress, assigns citation numbers, and respects stop signals.
  - Add Open URL tool and wire it into the tool constructor.

- **Refactors**
  - Refactored fast_chat_turn to the new query loop and stream events; rebuilt tests for the new framework.
  - Added tests for tool execution, prompt chaining, cancellation (pre/mid-stream), citations, JSON-to-tool-call behavior, Anthropic function tool_choice handling, and Open URL/Web Search tools.
  - Moved context handlers and prompt utils to ChatCompletionMessage and added base_messages_to_chat_completion_msgs.
  - Removed deprecated llm_model/model_settings from ChatTurnDependencies.
  - Fixed tool accounting to support class methods and accurate run_step tracking.
  - Exclude reasoning config for Anthropic models when using function tool_choice.
  - Standardized tool run_v2 to return JSON strings; updated Custom, MCP, Python, Image, Search, Web Search, and Open URL tools and tests.

<sup>Written for commit 9ee36e356bc4ad92633aa5d5e9429596674964d4. Summary will update automatically on new commits.</sup>


backend/onyx/chat/turn/fast_chat_turn.py
```diff
@@ -145,44 +121,32 @@ def _run_agent_loop(
         ctx.current_input_tokens = _extract_tokens_from_messages(current_messages)
 
         if not available_tools:
-            tool_choice = None
+            tool_choice = "none"
```
evan-onyx, line 126: this seems odd to me but if we want to do it this way then it would be good for this to be a string constant

backend/onyx/chat/turn/fast_chat_turn.py
```diff
@@ -339,47 +308,87 @@ def _process_stream(
...
-            tool_call_events.append(cast(ResponseFunctionToolCall, ev.item.raw_item))
-    if agent_stream.streamed is None:
-        raise ValueError("agent_stream.streamed is None")
-    return agent_stream.streamed, tool_call_events
+
+        if isinstance(event, RunItemStreamEvent):
+            if event.type == "message_start" and not message_section_open:
+                _start_message_section(dependencies, ctx)
+                message_section_open = True
+            elif event.type == "message_done" and message_section_open:
```
evan-onyx, line 330: these strs should probably be part of an enum

PR #6266, opened 2025-11-15
fix(index attempts): Preserve some attempts

justin-tahara:
## Description

We currently purge all of the indexing attempts if it's been more than 7 days and the connector has been paused. 

This PR aims to preserve at least 10 attempts so that we have some history for the user in order to ensure that we can display the correct statuses for the connectors.

## How Has This Been Tested?

Manually created entires locally and purged or updated the dates to ensure that they get in the purge queue

## Additional Options

---
## Summary by cubic
Always keep the 10 most recent indexing attempts per connector/search settings when purging old attempts. This preserves useful history and keeps connector statuses accurate.

<sup>Written for commit 75bf86e14fc30102bbaf015fd19556cb98d21d8f. Summary will update automatically on new commits.</sup>


Subash-Mohan (approved):
lgtm!

PR #6272, opened 2025-11-15
chore(deployments): fix cross-platform related issues

jmelahman:
## Description

Some follow fixes related to https://github.com/onyx-dot-app/onyx/commit/530d6d8284b5516f0842690bf955d317859cafc4

- Keeps the `Docker Login` for dry-runs. I think this is required for caching to work for them.
- Removes the rust toolchain from the model-server. I can try alternative methods of installing it if need, but it doesn't seem necessary and otherwise saves 700mb + we don't have to deal with packages like `perl`
- Freeze's the version of `uv`. Should help with caching + more secure
- Uses smaller instances for deployments. These steps are mostly network bound especially with caching fully working rather than CPU bound.

## How Has This Been Tested?

Partially captured by presubmit, dry-run deployment @ https://github.com/onyx-dot-app/onyx/actions/runs/19383244648

## Additional Options

- [x] Override Linear Check

---
## Summary by cubic
Fix cross-platform deployment issues by right-sizing x64 runners, enabling Docker login during dry runs, standardizing builds to x64 only, and adding 30GB disk for model-server builds. Pin uv and remove the Rust toolchain to stabilize caching and shrink the image.

- **Refactors**
  - Switch web to 4cpu x64 and backend to 2cpu x64; keep model-server at 2cpu x64.
  - Standardize Docker build platforms to x64; drop arm64.
  - Add volume=30gb for model-server builds; allow the volume tag in actionlint.
  - Always run Docker Hub login to ensure cache hits on dry runs.

- **Dependencies**
  - Pin uv to 0.9.9 across backend and model-server.
  - Remove the Rust toolchain from the model-server image to reduce size and avoid extra packages.
  - Drop the redundant uv copy in the integration tests image.

<sup>Written for commit cd6ac60358c75b42bdeb0c13763395620da624cb. Summary will update automatically on new commits.</sup>

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -119,7 +119,7 @@ jobs:
         with:
           context: ./web
           file: ./web/Dockerfile
-          platforms: linux/amd64,linux/x64
+          platforms: linux/amd64,linux/arm64
           push: ${{ env.IS_DRY_RUN != 'true' }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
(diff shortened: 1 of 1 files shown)


justin-tahara (approved):

justin-tahara (approved):

jmelahman:
@greptile

PR #6286, opened 2025-11-17
chore(docker): generate OpenAPI schema/client with docker

jmelahman:
## Description

Re-use the backend image to avoid needing to setup the same python environment. Also, generate the schema and client in parallel with the installing the dev dependencies.

## How Has This Been Tested?

Captured by presubmit

## Additional Options

- [x] Override Linear Check

---
## Summary by cubic
Generates the OpenAPI schema and Python client during the integration test Docker build using multi-stage builds on the backend image. This makes schema/client generation reproducible and faster.

- **New Features**
  - Multi-stage Dockerfile: build schema with the backend image and generate the client via openapi-generator in parallel, then include both in the final integration image.
  - Reuses the backend base image to avoid setting up a separate Python environment.

- **Refactors**
  - Removed the prepare-build step from the PR integration workflows; generation now happens inside the Docker build.
  - Added Docker Hub login in CI to reliably pull openapi-generator-cli and avoid unauthenticated rate limits.

<sup>Written for commit 83ca962e17760eb767c5ae9b4ca5015793d15b63. Summary will update automatically on new commits.</sup>

### DIFF
--- a/.github/workflows/pr-mit-integration-tests.yml
+++ b/.github/workflows/pr-mit-integration-tests.yml
@@ -123,11 +123,14 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@e468171a9de216ec08956ac3ada2f0791b6bd435 # ratchet:docker/setup-buildx-action@v3
 
-      - name: Prepare build
-        uses: ./.github/actions/prepare-build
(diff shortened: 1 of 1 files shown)


jmelahman:
@greptile

PR #6290, opened 2025-11-17
refactor: Input select

raunakab:
## Description

This PR modifies the existing `InputSelect` to make it more adherent to the mocks. For example, we have instances in which the selected item's icon needs to be displayed as well.

Addresses: https://linear.app/danswer/issue/DAN-3079/refactor-inputselect-component-to-make-it-more-adherent-to-the-mocks.

---
## Summary by cubic
Refactored InputSelect to match the mocks and display the selected option’s icon in the trigger. Replaced the old select with a new component built on Radix primitives and updated usage across the app. Addresses Linear issue DAN-3079.

- **New Features**
  - Trigger shows the selected option’s icon and label, with a clear placeholder when none is selected.
  - Error and disabled states with consistent styles; chevron rotates on open; rightSection still supported.
  - Closes on click outside and Escape; dropdown renders in a portal with proper positioning; supports option descriptions.

- **Refactors**
  - New API: pass children as InputSelect.Item; added Trigger and Content subcomponents; removed options prop; use error (not isError).
  - Updated onboarding, admin, auth, settings, and PAT components to the new API and placeholders.
  - LineItem now renders as a button and drops the main prop; selection uses heavyForced.
  - Restored Escape key handling; simplified RawModal outside-click; removed unused ui/select exports.
  - Deleted unused shadcn chart component.

<sup>Written for commit 84081900755d482ce58c583521deb8bf96fba723. Summary will update automatically on new commits.</sup>


web/src/refresh-components/inputs/InputSelect.tsx
```diff
@@ -1,213 +1,228 @@
...
+  ...props
+}: ItemProps) {
+  return (
+    <SelectPrimitive.Item
+      value={value}
+      className="outline-none focus:outline-none"
+      onSelect={(event) => {
+        // allow consumers to hook into clicks without breaking radix behaviour
+        onClick?.(event as unknown as React.MouseEvent<HTMLButtonElement>);
+      }}
```
nmgarza5: this smells a bit. i think we can add onClick to ItemProps like so to prevent this unknown casting. 

```
interface ItemProps extends Omit<LineItemProps, "heavyForced"> {
  value: string;
  selected?: boolean;
  onClick?: (event: Event) => void; // accept a standard DOM event
}
...
...
...
<SelectPrimitive.Item
   value={value}
   className="outline-none focus:outline-none"
   onSelect={(event) => onClick?.(event)}
/>
```

web/src/refresh-components/inputs/InputSelect.tsx
```diff
@@ -1,213 +1,228 @@
...
-
-  useEffect(() => {
-    const previousCursor = document.body.style.cursor;
-
-    // if disabled, set cursor to "not-allowed"
-    if (hovered) {
-      document.body.style.cursor = disabled ? "not-allowed" : "pointer";
-    } else {
-      document.body.style.cursor = ""; // revert to stylesheet default when not hovered
+  React.useEffect(() => {
```
nmgarza5: is there a preference to use `React.<hook>` instead of importing the hook directly?

web/src/components/ui/chart.tsx
nmgarza5, line 1: any reason why you want these components removed in this PR? I think it should be separate. Refering to chart, command, and menubar

web/src/components/llm/LLMSelector.tsx
```diff
@@ -112,41 +107,44 @@ export const LLMSelector: React.FC<LLMSelectorProps> = ({
...
   const currentLlmName = currentDescriptor?.modelName;
+  const defaultLabel = userSettings ? "System Default" : "User Default";
+
+  const wrapIcon = (
+    IconComponent: ReturnType<typeof getProviderIcon>
+  ): React.FC<SvgProps> => {
+    return function SelectIcon({ className }: SvgProps) {
+      return IconComponent({ size: 16, className });
+    };
+  };
```
nmgarza5: should be able to replace this with [createIcon](https://github.com/onyx-dot-app/onyx/blob/2b6e02a775fee920ad098314dd3f6b790722f479/web/src/components/icons/icons.tsx#L137) if you export it.

web/src/hooks/useKeyPress.ts
```diff
@@ -7,18 +7,18 @@ export function useKeyPress(
...
+    function handleKeyDown(event: KeyboardEvent) {
+      if (event.key !== key) return;
+      event.preventDefault();
+      callback();
+    }
+    document.addEventListener("keydown", handleKeyDown);
+    return () => {
+      document.removeEventListener("keydown", handleKeyDown);
+    };
+  }, [callback, enabled, key]);
```
nmgarza5, line 21: do you know why this was commented out previously? Are we sure this wont cause any bugs? Should probably add a unit test for the keystroke behavior

nmgarza5 (commented):
some comments. generally looks good but I do have one major critique....

To use this component, we import the root and items via 
`import * as InputSelect` and then always nest Items in the root like so:
```
<InputSelect.Root>
	<InputSelect.Item />
	<InputSelect.Item />
</InputSelect.Root>
```

I think it would be cleaner if we didnt force the consumer to have to use both of the components. Can we instead do something like this?

```
const values = [value1, value2, value3]
...
<InputSelect>
	values={values}
	selectedValue={selectedValue}
</InputSelect>
```

Let me know what you think. Happy to chat live

web/src/hooks/useKeyPress.ts
raunakab: Yup! I was the one who commented this out earlier.

The reason why was because it was a hotfix solution to another poor API decision that I created. Essentially, previously, our modals were implemented with a global `ModalProvider`. Then, modals would be attached to the DOM, and they'd look like this:

```tsx
// MyModal.tsx
export function MyModal() {
  const { isOpen, ... } = useModal();
  // other `useStates` here, etc.
  useEscape(...);

  if (!isOpen("MyModal") return null;

  return <div> { ... } </div;
}

// Consumer of MyModal.tsx
export function Parent() {
  return (
    <>
      <MyModal />
      <div { ... } />
    </>
  );
}
```

The problem with the above approach was that `MyModal` would be *always* mounted to the DOM. Even if `isOpen("MyModal")` returned false, the `useStates` and *all other hooks* would still run, because the `MyModal` component still exists. It's just returning `null`, but it's *still mounted*.

Now, if you had 2 or more components which all used `useEscape`, then you'd have 2+ components *all listening to "Escape" keypresses*. This caused very strange behaviour in which components which will "invisible" were still hijacking keypresses.

To hotfix that solution, I just commented out that code. However, @Subash-Mohan and I reimplemented the Modal API (Subash actually has more changes which are landing). With those fixes which are already in, modals are actually now being dismounted for real. Therefore, there's no more keypress hijacking, meaning that it's safe to uncomment the contents of this hook!

web/src/components/llm/LLMSelector.tsx
raunakab: Nice catch! Updated.

web/src/components/ui/chart.tsx
raunakab: I saw that they were completely unused, so I figured.. eh, why not? Would you like them to split out in a separate PR? I don't mind.

web/src/refresh-components/inputs/InputSelect.tsx
raunakab: Oof... hm. I haven't been enforcing any style, tbh. Do you have a preference?

web/src/refresh-components/inputs/InputSelect.tsx
raunakab: Fair point! Updated.

web/src/refresh-components/buttons/LineItem.tsx
```diff
@@ -47,26 +44,23 @@ export default function LineItem({
...
-        : "main";
+  const variant = strikethrough
+    ? "strikeThrough"
+    : forced || heavyForced
+      ? "forced"
+      : "main";
 
   const content = (
-    <div
+    <button
```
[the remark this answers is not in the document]
raunakab: Will address in a future PR.

web/src/refresh-components/inputs/InputSelect.tsx
nmgarza5: I have a preference to pull in the hooks directly. In most cases you want to import only what you need.

web/src/components/ui/chart.tsx
nmgarza5: not a huge deal you can leave them in if thats easier for you

web/src/refresh-components/inputs/InputSelect.tsx
raunakab: That sounds good to me! I'll update the PR.

raunakab:
> some comments. generally looks good but I do have one major critique....
> 
> To use this component, we import the root and items via `import * as InputSelect` and then always nest Items in the root like so:
> 
> ```
> <InputSelect.Root>
> 	<InputSelect.Item />
> 	<InputSelect.Item />
> </InputSelect.Root>
> ```
> 
> I think it would be cleaner if we didnt force the consumer to have to use both of the components. Can we instead do something like this?
> 
> ```
> const values = [value1, value2, value3]
> ...
> <InputSelect>
> 	values={values}
> 	selectedValue={selectedValue}
> </InputSelect>
> ```
> 
> Let me know what you think. Happy to chat live

@nmgarza5 I think that's a totally valid approach. I think the general pattern that we've been employing in our codebase is that we have a more "involved" API, and then a "simplified" version also available.

A good example are the `Popovers`, for example. We have a simpler popover, called `SimplePopover`, which makes the API simpler, but also thus makes it more limited.

So maybe we also expose a `SimpleInputSelect`?

PR #6300, opened 2025-11-17
chore(deps): upgrade psutil: 5.9.8->7.1.3

jmelahman:
## Description

This is pulled in transitively by the model_server. Without https://github.com/onyx-dot-app/onyx/pull/6291, the model_server installs this latest version. With that change, this `5.9.5` version fails to install in the model_server image (it seems there are no prebuilt wheels at the version and building from source fails). 

## How Has This Been Tested?

Captured by presubmit

## Additional Options

- [x] Override Linear Check


jmelahman:
@greptile

justin-tahara (approved):

PR #6305, opened 2025-11-18
part 1: no-op PC provider and interfaces

evan-onyx:
## Description

first step in prompt caching support: basic pass-through provider and interfaces

## How Has This Been Tested?

n/a, will test full system

## Additional Options

---
## Summary by cubic
Introduces the prompt caching framework scaffolding with a pass-through (no-op) provider and shared interfaces. No behavior changes yet; all requests continue unchanged.

- **New Features**
  - CacheManager to store and retrieve CacheMetadata in PgRedisKVStore with tenant-aware keys.
  - Deterministic cache key hashing from normalized messages plus provider, model, and tenant.
  - Provider base interface for message preparation, metadata extraction, and TTL.
  - NoOpPromptCacheProvider that leaves messages unchanged; handles optional continuation when the prefix isn’t a string.
  - Utilities to normalize LanguageModelInput and content; public exports added in package __init__.

<sup>Written for commit 1c27d414f1ca76e33ce6878c3e1af2b8a631ad07. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/llm/prompt_cache/cache_manager.py
+++ b/backend/onyx/llm/prompt_cache/cache_manager.py
@@ -1,0 +1,199 @@
+"""Cache manager for storing and retrieving prompt cache metadata."""
+
+import hashlib
+import json
+from datetime import datetime
+from datetime import timezone
+
+from onyx.key_value_store.store import PgRedisKVStore
+from onyx.llm.interfaces import LanguageModelInput
+from onyx.llm.prompt_cache.interfaces import CacheMetadata
+from onyx.llm.prompt_cache.utils import normalize_language_model_input
+from onyx.utils.logger import setup_logger
+from shared_configs.contextvars import get_current_tenant_id
+
+logger = setup_logger()
+
+REDIS_KEY_PREFIX = "prompt_cache:"
+# Cache TTL multiplier - store caches slightly longer than provider TTL
+# This allows for some clock skew and ensures we don't lose cache metadata prematurely
+CACHE_TTL_MULTIPLIER = 1.2
+
+
+class CacheManager:
+    """Manages storage and retrieval of prompt cache metadata."""
+
+    def __init__(self, kv_store: PgRedisKVStore | None = None) -> None:
+        """Initialize the cache manager.
+
+        Args:
+            kv_store: Optional key-value store. If None, creates a new PgRedisKVStore.
+        """
+        self._kv_store = kv_store or PgRedisKVStore()
+
+    def _build_cache_key(
+        self,
+        provider: str,
+        model_name: str,
+        cache_key_hash: str,
+        tenant_id: str | None = None,
+    ) -> str:
+        """Build a Redis/PostgreSQL key for cache metadata.
+
+        Args:
+            provider: LLM provider name (e.g., "openai", "anthropic")
+            model_name: Model name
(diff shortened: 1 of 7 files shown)


backend/onyx/llm/prompt_cache/cache_manager.py
```diff
@@ -0,0 +1,199 @@
...
+            tenant_id: Tenant ID. If None, uses current tenant from context.
+
+        Returns:
+            Cache key string
+        """
+        if tenant_id is None:
+            tenant_id = get_current_tenant_id()
+        return f"{REDIS_KEY_PREFIX}{tenant_id}:{provider}:{model_name}:{cache_key_hash}"
+
+    def store_cache_metadata(
```
yuhongsun96, line 56: Why do we hash some of the keys and not here? How would the fetching actually work?

backend/onyx/llm/prompt_cache/interfaces.py
```diff
@@ -0,0 +1,20 @@
+"""Interfaces and data structures for prompt caching."""
```
yuhongsun96, line 1: If not adding an actual interface, might as well call this models.py for consistency in the code base since those files store pydantic models

backend/onyx/llm/prompt_cache/utils.py
```diff
@@ -0,0 +1,30 @@
+"""Utility functions for prompt caching."""
+
+from collections.abc import Sequence
+from typing import Any
+
+from onyx.llm.interfaces import LanguageModelInput
+from onyx.llm.message_types import ChatCompletionMessage
```
yuhongsun96, line 7: Seems a rebase thing but on your branch these models don't exist. Prob fine but FYI

yuhongsun96: Actually ya they don't exist there on latest main, might want to update that

backend/onyx/llm/prompt_cache/utils.py
```diff
@@ -0,0 +1,30 @@
...
+from onyx.llm.message_types import UserMessageWithText
+
+
+def normalize_language_model_input(
+    input: LanguageModelInput,
```
yuhongsun96, line 12: I think we should just not allow str, the cases where it's a str, we just use UserMessage anyway

backend/onyx/llm/prompt_cache/utils.py
```diff
@@ -0,0 +1,30 @@
...
+    """Normalize LanguageModelInput to Sequence[ChatCompletionMessage]."""
+    if isinstance(input, str):
+        # Convert string to user message
+        return [UserMessageWithText(role="user", content=input)]
+    else:
+        return input
+
+
+def normalize_content(
+    content: str | list[str | dict[str, Any]] | Any,
```
yuhongsun96, line 23: the typing here is kind of oof

yuhongsun96 (approved):
lgtm, small comments/questions but good to go!

backend/onyx/llm/prompt_cache/cache_manager.py
evan-onyx: the redis key is based on the generated hash and some other stuff

evan-onyx:
the third branch is what got merged

PR #6318, opened 2025-11-18
fix: slack config forms + scope issues

nmgarza5:
## Description

Some fixes for slack bot configuration and scoping. The changes here allow document sets to be added to slack bots and prevent missing slack app scopes from crashing the entire workflow.

Also adding a small change for recency queries. I noticed with recency slack bot queries, we were only triggering one slack search. Now if it is a content + recency query we use an llm query for keyword extraction.  

## How Has This Been Tested?

tested local slack bot setup in this channel: https://onyx-company.slack.com/archives/C09ULM10RUG

## Additional Options

---
## Summary by cubic
Fixes Slack bot configuration and scope handling so searches keep working even when some Slack scopes are missing. Also improves recency queries and unblocks adding document sets to Slack bots.

- **Bug Fixes**
  - Prevent crashes on missing Slack scopes by mapping scopes to channel types, retrying without the missing types, and returning partial channel metadata.
  - Fix Slack admin form to read connector source correctly, restoring the “Relevant Connectors” list and enabling document set selection.

- **New Features**
  - Smarter recency handling: detect pure recency vs content+recency, parse phrases like “last Monday,” and use LLM keyword extraction when content+recency is detected.
  - Added tests for single and multiple missing-scope scenarios (e.g., mpim, im).

<sup>Written for commit 887a1abadb43a91f0f15f9e190ffc4c6dbff30c7. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/context/search/federated/slack_search.py
+++ b/backend/onyx/context/search/federated/slack_search.py
@@ -162,11 +162,17 @@ def fetch_and_cache_channel_metadata(
 
         except SlackApiError as e:
             last_exception = e
-            error_response = e.response.get("error", "") if e.response else ""
+
+            # Extract all needed fields from response upfront
+            if e.response:
+                error_response = e.response.get("error", "")
+                needed_scope = e.response.get("needed", "")
+            else:
+                error_response = ""
+                needed_scope = ""
 
             # Check if this is a missing_scope error
-            if error_response == "missing_scope" and e.response:
-                needed_scope = e.response.get("needed", "")
+            if error_response == "missing_scope":
 
                 # Get the channel type that requires this scope
                 missing_channel_type = get_channel_type_for_missing_scope(needed_scope)
--- a/backend/onyx/context/search/federated/slack_search_utils.py
+++ b/backend/onyx/context/search/federated/slack_search_utils.py
@@ -29,6 +29,9 @@ DAYS_PER_WEEK = 7
 DAYS_PER_MONTH = 30
 MAX_CONTENT_WORDS = 3
 
+# Punctuation to strip from words during analysis
+WORD_PUNCTUATION = ".,!?;:\"'#"
+
 RECENCY_KEYWORDS = ["recent", "latest", "newest", "last"]
 
 
@@ -130,7 +133,7 @@ def is_recency_query(query: str) -> bool:
     # Count content words (not stop words, length > 2)
     content_word_count = 0
     for word in words:
-        clean_word = word.strip(".,!?;:\"'#")
+        clean_word = word.strip(WORD_PUNCTUATION)
         if clean_word and len(clean_word) > 2 and clean_word not in all_stop_words:
             content_word_count += 1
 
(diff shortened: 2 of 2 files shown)


backend/onyx/context/search/federated/slack_search.py
```diff
@@ -157,6 +162,36 @@ def fetch_and_cache_channel_metadata(
 
         except SlackApiError as e:
             last_exception = e
+            error_response = e.response.get("error", "") if e.response else ""
+
+            # Check if this is a missing_scope error
+            if error_response == "missing_scope" and e.response:
```
justin-tahara: Feel like it might be redundant to check the e.response here again but let me know what you think. Maybe we can just get all the fields that we need ahead of time or something

backend/onyx/context/search/federated/slack_search.py
```diff
@@ -157,6 +162,36 @@ def fetch_and_cache_channel_metadata(
...
+                # Get the channel type that requires this scope
+                missing_channel_type = get_channel_type_for_missing_scope(needed_scope)
+
+                if (
+                    missing_channel_type
+                    and missing_channel_type in available_channel_types
+                ):
+                    # Remove the problematic channel type and retry
+                    available_channel_types.remove(missing_channel_type)
+                    logger.warning(
```
justin-tahara, line 186: Should this be warning or info? This won't show up with the default log level

backend/onyx/context/search/federated/slack_search_utils.py
```diff
@@ -41,6 +41,48 @@ class ChannelTypeString(str, Enum):
...
+
+    Examples:
+        >>> get_channel_type_for_missing_scope('mpim:read')
+        'mpim'
+        >>> get_channel_type_for_missing_scope('im:read')
+        'im'
+        >>> get_channel_type_for_missing_scope('unknown:scope')
+        None
+    """
+    return SCOPE_TO_CHANNEL_TYPE_MAP.get(scope)
```
justin-tahara, line 86: Should we have a default?

backend/onyx/context/search/federated/slack_search_utils.py
```diff
@@ -64,11 +106,40 @@ def _parse_llm_code_block_response(response: str) -> str:
...
+    all_stop_words = _get_combined_stop_words()
+
+    # Extract content words (excluding stop words)
+    query_lower = query.lower()
+    words = query_lower.split()
+
+    # Count content words (not stop words, length > 2)
+    content_word_count = 0
+    for word in words:
+        clean_word = word.strip(".,!?;:\"'#")
```
justin-tahara: Should we set this filter somewhere at the top so we don't have to keep updating it?

backend/onyx/context/search/federated/slack_search_utils.py
```diff
@@ -413,6 +517,23 @@ def build_channel_override_query(channel_references: set[str], time_filter: str)
...
+def _get_combined_stop_words() -> set[str]:
+    """Get combined NLTK + Slack-specific stop words.
+
+    Returns a set of stop words for filtering content words.
+    Falls back to just Slack-specific stop words if NLTK is unavailable.
+    """
+    try:
+        from nltk.corpus import stopwords  # type: ignore
+
+        nltk_stop_words = set(stopwords.words("english"))
```
justin-tahara, line 538: There are users who use different languages so don't know how you want to handle that

justin-tahara (commented):
Minor nits

backend/onyx/context/search/federated/slack_search_utils.py
nmgarza5: ooof....ok will think on that

nmgarza5: adding a todo and stating we need to support multiple languages. I dont want to add that in this PR.

backend/onyx/context/search/federated/slack_search_utils.py
nmgarza5: keeping as is - this is the correct behavior to signal "unknown scope" and allows the calling code to handle when a scope isnt set

backend/onyx/context/search/federated/slack_search.py
nmgarza5: warnings still surface. leaving as is

PR #6332, opened 2025-11-19
fix: Some minor touch-ups for the new modal

raunakab:
## Description

A few minor touch ups here and there.

Addresses: https://linear.app/danswer/issue/DAN-3087/revamp-modal-component.

---
## Summary by cubic
Polished the Modal component to align with DAN-3087: simplified size API, improved animations, and moved the close button into the header. Includes small UI tweaks to header shadow, description text, and footer layout.

- **Refactors**
  - Use boolean size variants on Modal.Content: large, medium (default), small, tall, mini.
  - Overlay animation class order fixed; content slides vertically from center.
  - Close button is placed inside Modal.Header; removed custom focus border.
  - Header bottom shadow uses shadow-01; description uses text03; footer is flex-aligned to end.

- **Migration**
  - Import as default: import Modal from "@/refresh-components/Modal".
  - Update usage: <Modal.Content large> (replaces size="main") and other boolean flags.
  - Move <Modal.CloseButton /> inside <Modal.Header>.
  - Tall sets a max-height to keep content within the viewport.
  - ProviderModal drops startAdornment; pass an icon component via the icon prop.

<sup>Written for commit 30e8b84a5f7e92c4edc0e9c411b2a2aebde987a4. Summary will update automatically on new commits.</sup>

### DIFF
--- a/web/src/refresh-components/Modal.tsx
+++ b/web/src/refresh-components/Modal.tsx
@@ -91,7 +91,7 @@ const sizeClassNames = {
   large: ["w-[80dvw]", "h-[80dvh]"],
   medium: ["w-[60rem]", "h-fit"],
   small: ["w-[32rem]", "h-[30rem]"],
-  tall: ["w-[32rem]"],
+  tall: ["w-[32rem]", "max-h-[calc(100dvh-4rem)]"],
   mini: ["w-[32rem]", "h-fit"],
 } as const;
 


web/src/refresh-components/Modal.tsx
```diff
@@ -57,22 +57,41 @@ const ModalOverlay = React.forwardRef<
...
 /**
  * Size class names mapping for modal variants
  */
 const sizeClassNames = {
-  main: ["w-[80dvw]", "h-[80dvh]"],
+  large: ["w-[80dvw]", "h-[80dvh]"],
   medium: ["w-[60rem]", "h-fit"],
   small: ["w-[32rem]", "h-[30rem]"],
-  tall: ["w-[32rem]", "max-h-[calc(100dvh-4rem)]"],
+  tall: ["w-[32rem]"],
```
Subash-Mohan: We need to set a maximum height here because if the content grows, the modal overflows the screen instead of becoming scrollable.

Subash-Mohan: Screenshot 2025-11-19 at 7 02 09 PM

web/src/refresh-components/Modal.tsx
```diff
@@ -415,7 +417,6 @@ const ModalCloseButton = React.forwardRef<
       <div
         ref={closeButtonRef as React.RefObject<HTMLDivElement>}
         tabIndex={-1}
-        className="rounded-12 !outline-none !border-[3px] !border-transparent focus:!border-action-link-05 transition-colors duration-200"
```
Subash-Mohan, line 418: We need the class names; if we remove them, the accidental modal-closing mechanism won’t highlight the close icon on the first click.

raunakab: Oh, sorry about that. Good catch!

raunakab: Reverted.

web/src/refresh-components/Modal.tsx
raunakab: Ah I see. Okay. Added back. Thanks for the catch!

Subash-Mohan (approved):

PR #6333, opened 2025-11-19
Fix/user file modal

Subash-Mohan:
## Description
This pull request refactors the user file modal functionality in the admin and chat project panels by replacing the old `UserFilesModalContent` component with a new, more flexible `UserFilesModal` component. Additionally, it introduces a reusable `ScrollIndicatorDiv` for improved scroll feedback in modal dialogs. The changes streamline modal usage, improve UI consistency, and enhance user experience when managing files.

## How Has This Been Tested?
Tested from UI

## Additional Options

---
## Summary by cubic
Replaced the old user file modal with a new controlled UserFilesModal and added scroll indicators for smoother scrolling. This standardizes file management across admin, project, and popover flows and improves usability.

- **New Features**
  - Controlled modal with search, upload, selection, view/delete, and a Done action.
  - Show-only-selected toggle and Deselect All controls.
  - ScrollIndicatorDiv adds gradient/shadow indicators during scroll.
  - Modal.Content option to disable accidental-close protection.

- **Refactors**
  - Swapped UserFilesModalContent for UserFilesModal in AssistantEditor, ProjectContextPanel, and FilePickerPopover.
  - Removed UserFilesModalContent and ProviderModalLayout; modal open/close now uses ModalContext and drops onClose props.
  - Unified file status icons and labels.

<sup>Written for commit 79b1ae7c5d3dc6b3df765f088fe1c7c25e2cdf4c. Summary will update automatically on new commits.</sup>

### DIFF
--- a/web/src/refresh-components/Modal.tsx
+++ b/web/src/refresh-components/Modal.tsx
@@ -248,6 +248,13 @@ const ModalContent = React.forwardRef<
     // Handle escape key and outside clicks
     const handleInteractOutside = React.useCallback(
       (e: Event) => {
+        // If preventAccidentalClose is disabled, always allow immediate close
+        if (!preventAccidentalClose) {
+          setHasAttemptedClose(false);
+          return;
+        }
+
+        // If preventAccidentalClose is enabled, check if user has modified inputs
         if (hasModifiedInputs()) {
           if (!hasAttemptedClose) {
             // First attempt: prevent close and focus the close button
@@ -261,7 +268,7 @@ const ModalContent = React.forwardRef<
             setHasAttemptedClose(false);
           }
         } else {
-          // No modified inputs or feature disabled: allow immediate close
+          // No modified inputs: allow immediate close
           setHasAttemptedClose(false);
         }
       },


web/src/components/modals/UserFilesModal.tsx
```diff
@@ -0,0 +1,306 @@
...
+          </Modal.Header>
+
+          <Modal.Body className="flex flex-col flex-1 overflow-hidden bg-background-tint-01">
+            {/* File display section */}
+            {filtered.length === 0 ? (
+              <div className="p-4 flex w-full h-full items-center justify-center">
+                <Text text03>No files found</Text>
+              </div>
+            ) : (
+              <ScrollIndicatorDiv className="px-2 pt-2 gap-2" variant="shadow">
```
Subash-Mohan, line 226: I created this component because it behaves differently from `OverflowDiv`:

| Feature | OverflowDiv | ScrollIndicatorDiv |
|--------|-------------|--------------------|
| **Bottom indicator** | ✅ Always visible | ✅ Only shows when scrollable |
| **Top indicator** | ❌ None | ✅ Shows when scrolled down |
| **Scroll detection** | ❌ No | ✅ Yes (via `ResizeObserver`) |
| **Variants** | Gradient only | Gradient or shadow |

In this scenario, having scroll-aware top and bottom shadows is required.

raunakab: If this is a superior component to `OverflowDiv`, then we should aim to always use this one and delete that one!

web/src/refresh-components/Modal.tsx
```diff
@@ -133,174 +138,180 @@ const ModalContent = React.forwardRef<
...
-      // Cleanup previous listener if exists
-      if (containerNodeRef.current) {
-        containerNodeRef.current.removeEventListener(
-          "input",
-          handleInput,
-          true
-        );
+>(
+  (
+    { className, children, size, preventAccidentalClose = true, ...props },
```
raunakab: Should we instead prefer `preventAccidentalClose = false`? I don't mind either ways, just wondering why we chose `true`.

web/src/components/modals/UserFilesModal.tsx
```diff
@@ -0,0 +1,306 @@
...
+                      onDelete={
+                        onDelete ? () => onDelete(projectFle) : undefined
+                      }
+                    />
+                  );
+                })}
+              </ScrollIndicatorDiv>
+            )}
+          </Modal.Body>
+          <Modal.Footer className="flex items-center justify-between p-4 border-t">
```
raunakab: Is the `border-t` here used to draw a separator? If so, I would prefer that we use the actual `Separator` component instead.

web/src/app/admin/assistants/AssistantEditor.tsx
```diff
@@ -846,36 +846,29 @@ export default function AssistantEditor({
...
-                          status: "completed" as const,
-                        }
-                      );
+              <UserFilesModal
+                open={userFilesModal.isOpen}
+                onOpenChange={userFilesModal.toggle}
+                title="User Files"
+                description="All files selected for this assistant"
+                icon={SvgFiles}
+                recentFiles={values.user_file_ids.map((userFileId: string) => {
```
raunakab: Hm. Eliminating the `userFilesModal.Provider` wrapper here seems dangerous. Now, `UserFilesModal` is *always* attached to the DOM, no? It may just not render if `open === false`, but all of the state inside of `UserFilesModal` will still be attached...

Subash-Mohan: Sorry, I just handled it manually. I’ve wrapped the modals with the provider now.

web/src/components/modals/UserFilesModal.tsx
Subash-Mohan: OverflowDiv is used in the sidebar. I’m not sure whether the new component fully matches the old one. Since you created the original OverflowDiv, could you confirm this?

web/src/components/modals/UserFilesModal.tsx
Subash-Mohan: We don’t need it. I’ve removed it now.

web/src/refresh-components/Modal.tsx
Subash-Mohan: I believe there are only a few cases where we have input fields but still allow the modal to close accidentally, such as the search bar in the UserFileModal.

raunakab: Ah okay. Sounds good then!

web/src/components/modals/UserFilesModal.tsx
raunakab: Ya, the overflow is used in the `AppSidebar`, and was previously used in this `UserFilesModal`.

However, if your new component is truly better, then I will create a follow-up PR to delete  `OverflowDiv` and replace it with `ScrollIndicatorDiv`.

PR #6334, opened 2025-11-19
fix(Projects): file ordering in project panel

Subash-Mohan:
## Description
This pull request introduces a new `created_at` timestamp to the `project__user_file` association table to track when files are added to projects. The changes update the database schema, ORM model, and API logic to use this new timestamp for sorting files in a project.

Ticket -> https://linear.app/danswer/issue/DAN-2884/newly-added-files-should-appear-on-the-left-of-the-list

## How Has This Been Tested?
Tested from UI

## Additional Options

---
## Summary by cubic
Fixes project file ordering so newly added files appear first (left) in the project panel by sorting on when they were added to the project, not when they were uploaded. Addresses Linear DAN-2884.

- **Bug Fixes**
  - Added created_at to project__user_file (DB + ORM) to track when a file is added to a project, with an index for faster ordering.
  - Updated API to join the association and order by Project__UserFile.created_at desc, keeping existing filters.

- **Migration**
  - Run alembic upgrade head to apply the new column and index.

<sup>Written for commit 99887eadb4694d5c3e693c197cb98a1aa3469af2. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/alembic/versions/6436661d5b65_add_created_at_in_project_userfile.py
+++ b/backend/alembic/versions/6436661d5b65_add_created_at_in_project_userfile.py
@@ -1,8 +1,8 @@
 """add_created_at_in_project_userfile
 
-Revision ID: 3e380d25c3c3
-Revises: 1f2a3b4c5d6e
-Create Date: 2025-11-19 14:10:49.515889
+Revision ID: 6436661d5b65
+Revises: c7e9f4a3b2d1
+Create Date: 2025-11-24 11:50:24.536052
 
 """
 
@@ -10,8 +10,8 @@ from alembic import op
 import sqlalchemy as sa
 
 # revision identifiers, used by Alembic.
-revision = "3e380d25c3c3"
-down_revision = "1f2a3b4c5d6e"
+revision = "6436661d5b65"
+down_revision = "c7e9f4a3b2d1"
 branch_labels = None
 depends_on = None
 


backend/onyx/db/models.py
```diff
@@ -3503,6 +3503,9 @@ class Project__UserFile(Base):
     user_file_id: Mapped[UUID] = mapped_column(
         ForeignKey("user_file.id"), primary_key=True
     )
+    created_at: Mapped[datetime.datetime] = mapped_column(
```
Weves, line 3506: we should add some indices to make this query fast

Subash-Mohan: Now I added the index for create_at.

backend/alembic/versions/6436661d5b65_add_created_at_in_project_userfile.py
```diff
@@ -0,0 +1,42 @@
...
+        "project__user_file",
+        sa.Column(
+            "created_at",
+            sa.DateTime(timezone=True),
+            server_default=sa.text("now()"),
+            nullable=False,
+        ),
+    )
+    # Add index on created_at column
+    op.create_index(
```
Weves, line 31: I believe we need to do a multi-column index, since the query is complex. See https://chatgpt.com/share/69209793-adc0-8005-aba7-eeb439bb11d1 for a brief explanation.

backend/onyx/db/models.py
```diff
@@ -3503,6 +3503,9 @@ class Project__UserFile(Base):
...
+        DateTime(timezone=True), server_default=func.now(), index=True
+    )
```
Weves, line 3516: can we also add the index definition here as well?

Weves: some of the other tables have them defined via SQLAlchemy, you can use those for reference

PR #6345, opened 2025-11-19
chore: unify checkbox implementation across the frontend

nmgarza5:
## Description

Closes https://linear.app/danswer/issue/DAN-3089/cleanup-checkbox-component-delete-old-instances

Refreshing the checkbox component and unifying usage across the codebase. We do add special UnLabeledCheckbox for formik forms though. 

## How Has This Been Tested?

unit + manual 

## Additional Options

---
## Summary by cubic
Unified all checkbox usage to a single, accessible Checkbox component with consistent behavior across the app. Addresses Linear DAN-3089 by removing legacy implementations, adding indeterminate support, and improving tests.

- **New Features**
  - New Checkbox component with controlled/uncontrolled modes, indeterminate and disabled states, ARIA support, and ref forwarding.
  - Added Formik integrations: CheckboxField and UnlabeledCheckboxField.
  - Comprehensive unit tests; E2E selectors now use aria-checked and aria-label.

- **Refactors**
  - Replaced CustomCheckbox and ui/checkbox imports across the frontend; removed both components.
  - Standardized event to onCheckedChange (replacing onChange where needed).
  - Updated ToolList to use the indeterminate prop.
  - Removed @radix-ui/react-checkbox from dependencies.
  - Fixed checkbox click propagation in DocumentSetSelectable; corrected Slack channel config to use CheckboxField.

<sup>Written for commit efd58ec5a979386e5e0303cc2ef203b160ab169a. Summary will update automatically on new commits.</sup>

### DIFF
--- a/web/tests/e2e/mcp/default-assistant-mcp.spec.ts
+++ b/web/tests/e2e/mcp/default-assistant-mcp.spec.ts
@@ -493,7 +493,7 @@ test.describe("Default Assistant MCP Integration", () => {
     const finalChecked =
       await firstToolCheckboxAfter.getAttribute("aria-checked");
     console.log(`[test] Final tool state: ${finalChecked}`);
-    expect(Boolean(finalChecked)).toEqual(!Boolean(initialChecked));
+    expect(finalChecked).not.toEqual(initialChecked);
   });
 
   test("Instructions persist when saving default assistant", async ({


web/src/icons/minus.tsx
```diff
@@ -0,0 +1,20 @@
...
+const SvgMinus = (props: SVGProps<SVGSVGElement>) => (
+  <svg
+    viewBox="0 0 16 16"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+    {...props}
+  >
+    <path
+      d="M4 8H12"
+      stroke="currentColor"
```
raunakab: Remove `stroke="currentColor"`. This hardcodes the colour, making it unmodifiable from the calling context. I.e.,

```tsx
<SvgMinus className="h-4 w-4 stroke-red-500" />
```

won't make draw an SVG with a red stroke, it'll hardcode keep it some default colour.

Removing that option will then allow us to update the colour from the calling context. Note that this will require you to add `"stroke-${colour}"` to wherever you're calling this from (presumably `Checkbox.tsx`).

web/src/refresh-components/inputs/Checkbox.tsx
```diff
@@ -0,0 +1,173 @@
...
+          }
+        }}
+      >
+        {(checked || indeterminate) && (
+          <div className="text-white">
+            {indeterminate ? (
+              <SvgMinus className="h-3 w-3" stroke="currentColor" />
+            ) : (
+              <SvgCheck className="h-3 w-3" stroke="currentColor" />
+            )}
```
raunakab, line 163: I believe in the mocks, the colour of the strokes are:

```
<Svg___ className="h-3 w-3 stroke-text-light-05" />
```

web/src/refresh-components/inputs/Checkbox.tsx
```diff
@@ -0,0 +1,173 @@
...
+          "hover:border-border-03",
+        ],
+  disabled:
+    checked || indeterminate
+      ? ["bg-background-neutral-04"]
+      : ["bg-background-neutral-03", "border", "border-border-02"],
+  focus:
+    checked || indeterminate
+      ? ["focus-visible:border-border-05"]
+      : ["focus-visible:border-border-05"],
```
raunakab: Super small nit, but maybe we could just throw these both as part of the `main` key-value?

Also, I believe the focus state is forgetting the `focus-shadow` (that class is present in the `globals.css`, if you want to check it out). It gives the component a "dual-bordered" look.

image

So, all in all, the final class names could look something like:

```tsx
const getRootClasses = (checked: boolean, indeterminate: boolean) => ({
  main:
    checked || indeterminate
      ? [
          "bg-action-link-05",
          "hover:bg-action-link-04",
          "focus-visible:border-border-05",
          "focus-visible:focus-shadow",
        ]
      : [
          "bg-background-neutral-00",
          "border",
          "border-border-02",
          "hover:border-border-03",
          "focus-visible:border-border-05",
          "focus-visible:focus-shadow",
        ],
  disabled:
    checked || indeterminate
      ? ["bg-background-neutral-04"]
      : ["bg-background-neutral-03", "border", "border-border-02"],
});
```

web/src/components/Bubble.tsx
```diff
@@ -36,7 +36,7 @@ export function Bubble({
       <div className="my-auto">{children}</div>
       {showCheckbox && (
         <div className="pl-2 my-auto">
-          <CustomCheckbox checked={isSelected} onChange={() => null} />
+          <Checkbox checked={isSelected} onCheckedChange={() => null} />
```
raunakab: Small nit, but maybe we could remove the `onCheckedChange` handler here?

web/src/refresh-components/inputs/Checkbox.tsx
```diff
@@ -0,0 +1,175 @@
...
+        role="checkbox"
+        aria-checked={indeterminate ? "mixed" : checked}
+        aria-label={ariaLabel}
+        aria-labelledby={ariaLabelledby}
+        aria-describedby={ariaDescribedby}
+        tabIndex={disabled ? -1 : 0}
+        className={cn(
+          "flex h-4 w-4 shrink-0 items-center justify-center rounded-04 transition-colors",
+          "focus-visible:outline-none focus-visible:ring-0",
+          "peer-focus-visible:border peer-focus-visible:border-border-05",
```
raunakab: Maybe these two lines need to be removed? Not sure why they're here, given that we are setting the `focus-visible` via the `rootClasses[variant]`.

nmgarza5: yeah seems they can be removed. will do. small oversight on my end. my b

PR #6346, opened 2025-11-20
chore(gha): require playwright passing on merge

jmelahman:
## Description

I think playwright stopped being required(?) when we introduced test projects in https://github.com/onyx-dot-app/onyx/commit/860bdd3c0f40ac8c9c79515f783acff4cf167753 as it changed the name of the Playwright Tests step

Also simplify the `require` step.

## Additional Options

- [x] Override Linear Check

---
## Summary by cubic
Make Playwright tests a required check on PRs so merges only happen when the suite passes. Simplifies CI “require” steps with a lightweight status check.

- **Bug Fixes**
  - Add a required job in pr-playwright-tests that fails if playwright-tests failed, were cancelled, or were skipped.
  - Restores the required status that was lost after the Playwright project rename.

- **Refactors**
  - Replace actions/github-script with a simple exit-on-failure check in pr-integration-tests and pr-mit-integration-tests (fail on failure/cancelled/skipped).
  - Run the required job on ubuntu-slim for faster queue times.

<sup>Written for commit 0cf27e881d30372b7d0001fbe9ae37d936cc325d. Summary will update automatically on new commits.</sup>

### DIFF
--- a/.github/workflows/pr-playwright-tests.yml
+++ b/.github/workflows/pr-playwright-tests.yml
@@ -309,7 +309,7 @@ jobs:
           name: docker-logs-${{ matrix.project }}-${{ github.run_id }}
           path: ${{ github.workspace }}/docker-compose.log
 
-  required:
+  playwright-required:
     # NOTE: Github-hosted runners have about 20s faster queue times and are preferred here.
     runs-on: ubuntu-slim
     needs: [playwright-tests]


jmelahman:
> did you mean to make the integration test changes?

Ya, these steps came up in the zizmor changes and I figured this is safer and faster

jmelahman:
I'm not sure how Github will handle branches that don't have this change (i.e. old branches are stuck waiting for a step that doesn't exist), so might need to wait until tomorrow or the day after before enabling to let folks catch up.

PR #6405, opened 2025-11-24
fix(AppSidebar): prevent overlay flash on sidebar for chatpage

Subash-Mohan:
## Description
Prevent the mobile sidebar overlay from rendering during initial mount to eliminate the unwanted shadow/overlay effect that appeared when refreshing the chat page.

## How Has This Been Tested?
Tested from UI

## Additional Options

---
## Summary by cubic
Prevents the mobile sidebar overlay from flashing on the chat page during load/refresh by rendering it only after the component mounts. Removes the unwanted shadow while keeping sidebar behavior unchanged.

- **Bug Fixes**
  - Gate overlay rendering with useIsMounted so it appears only after mount and still respects the folded state.

<sup>Written for commit 2f5c72fc648af09f8a1e1e2de915a2b58fadbf96. Summary will update automatically on new commits.</sup>

### DIFF
--- a/web/src/sections/sidebar/AppSidebar.tsx
+++ b/web/src/sections/sidebar/AppSidebar.tsx
@@ -1,11 +1,12 @@
 "use client";
 
-import { useCallback, memo, useMemo, useState, useEffect } from "react";
+import { useCallback, memo, useMemo, useState } from "react";
 import { useSettingsContext } from "@/components/settings/SettingsProvider";
 import { MinimalPersonaSnapshot } from "@/app/admin/assistants/interfaces";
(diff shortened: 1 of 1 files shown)


raunakab:
@Subash-Mohan, this doesn't address the "root" of the issue, which is that the application (while still loading up), defaults to a "mobile responsiveness mode".

I have a PR up which addresses this: https://github.com/onyx-dot-app/onyx/pull/6422. It essentially just updates the default responsiveness mode to be "desktop" instead (i.e., `isMobile === false` while the application is still mounting).

raunakab:
I'll close this PR in favour of that one!

PR #6427, opened 2025-11-25
feat(api): Add Users to Group Endpoint

justin-tahara:
## Description

Historically users needed to add users individually by concatenating to the existing list of users to add users to the user groups. 

This was done because our FE does not have a need for this endpoint. This is purely for our dev users. 

## How Has This Been Tested?

Tested locally and added some unit tests as well.

## Additional Options

---
## Summary by cubic
Adds an admin API to add multiple users to a user group in one call. Deduplicates user IDs and preserves existing group data.

- **New Features**
  - POST /manage/admin/user-group/{user_group_id}/add-users accepts user_ids and returns the updated group.
  - Requires curator/admin auth; 404 on missing group or invalid user IDs; respects “modifiable” checks.
  - Keeps existing users and cc_pairs; ignores duplicates.
  - Includes an EE integration test and test helpers.

- **Bug Fixes**
  - PATCH /manage/admin/user-group/{user_group_id} now validates added users exist and returns 404 if not.

<sup>Written for commit 55bbfd450ee47b33f96e299f3fcd353456ed2a38. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/ee/onyx/db/user_group.py
+++ b/backend/ee/onyx/db/user_group.py
@@ -591,8 +591,6 @@ def add_users_to_user_group(
     if db_user_group is None:
         raise ValueError(f"UserGroup with id '{user_group_id}' not found")
 
-    _check_user_group_is_modifiable(db_user_group)
-
     missing_users = [
         user_id for user_id in user_ids if fetch_user_by_id(db_session, user_id) is None
     ]
@@ -600,6 +598,8 @@ def add_users_to_user_group(
         raise ValueError(
             f"User(s) not found: {', '.join(str(user_id) for user_id in missing_users)}"
         )
+
+    _check_user_group_is_modifiable(db_user_group)
 
     current_user_ids = [user.id for user in db_user_group.users]
     current_user_ids_set = set(current_user_ids)
@@ -645,6 +645,17 @@ def update_user_group(
     added_user_ids = list(updated_user_ids - current_user_ids)
     removed_user_ids = list(current_user_ids - updated_user_ids)
 
+    if added_user_ids:
+        missing_users = [
+            user_id
+            for user_id in added_user_ids
+            if fetch_user_by_id(db_session, user_id) is None
+        ]
+        if missing_users:
+            raise ValueError(
+                f"User(s) not found: {', '.join(str(user_id) for user_id in missing_users)}"
+            )
+
     # LEAVING THIS HERE FOR NOW FOR GIVING DIFFERENT ROLES
     # ACCESS TO DIFFERENT PERMISSIONS
     # if (removed_user_ids or added_user_ids) and (
@@ -660,15 +671,6 @@ def update_user_group(
         )
 
     if added_user_ids:
-        missing_users = [
-            user_id
-            for user_id in added_user_ids
-            if fetch_user_by_id(db_session, user_id) is None
-        ]
-        if missing_users:
-            raise ValueError(
-                f"User(s) not found: {', '.join(str(user_id) for user_id in missing_users)}"
-            )
         _add_user__user_group_relationships__no_commit(
             db_session=db_session,
             user_group_id=user_group_id,


justin-tahara:
@greptileai can you review one more time?

backend/ee/onyx/db/user_group.py
```diff
@@ -581,6 +581,48 @@ def update_user_curator_relationship(
...
+    user_group_id: int,
+    user_ids: list[UUID],
+) -> UserGroup:
+    db_user_group = fetch_user_group(db_session=db_session, user_group_id=user_group_id)
+    if db_user_group is None:
+        raise ValueError(f"UserGroup with id '{user_group_id}' not found")
+
+    _check_user_group_is_modifiable(db_user_group)
+
+    missing_users = [
```
wenxi-onyx, line 594: Do this check first before fetching the group to fail earlier

backend/ee/onyx/db/user_group.py
```diff
@@ -581,6 +581,48 @@ def update_user_curator_relationship(
...
+        user_id for user_id in user_ids if fetch_user_by_id(db_session, user_id) is None
+    ]
+    if missing_users:
+        raise ValueError(
+            f"User(s) not found: {', '.join(str(user_id) for user_id in missing_users)}"
+        )
+
+    current_user_ids = [user.id for user in db_user_group.users]
```
wenxi-onyx, line 604: This likely doesn't matter, but set operations is a more pythonic way to do this. I could be convinced that this is more difficult to read and the list operations are fine since user groups should theoretically be very small n
```
current_user_ids = {user.id for user in db_user_group.users}
new_user_ids = set(user_ids) - current_user_ids

if not new_user_ids:
    return db_user_group

user_group_update = UserGroupUpdate(
    user_ids=list(current_user_ids | new_user_ids),
    cc_pair_ids=[cc_pair.id for cc_pair in db_user_group.cc_pairs],
)
```

backend/ee/onyx/db/user_group.py
```diff
@@ -581,6 +581,48 @@ def update_user_curator_relationship(
...
+
+    if not new_user_ids:
+        return db_user_group
+
+    user_group_update = UserGroupUpdate(
+        user_ids=current_user_ids + new_user_ids,
+        cc_pair_ids=[cc_pair.id for cc_pair in db_user_group.cc_pairs],
+    )
+
+    return update_user_group(
```
wenxi-onyx, line 618: I think there's technically a race condition here.

User A: hits this endpoint --> existing group users are retrieved from DB
User B: adds a someone to the group at the same time via UI (user A does not see this person)
User A: commits updated user group _without_ the person that User B adds

Perhaps low risk. Risk increases if there's a huge amount of users being processed in this operation. 
Maybe just document in a comment?

backend/ee/onyx/db/user_group.py
```diff
@@ -618,6 +660,15 @@ def update_user_group(
         )
 
     if added_user_ids:
+        missing_users = [
```
wenxi-onyx: Same as above comment, I think this should happen first to fail fast.

Additionally, maybe worth creating a helper func since this code appears twice

wenxi-onyx (approved):
Left a few comments, approving to unblock

backend/ee/onyx/db/user_group.py
justin-tahara: I feel like the current way is a bit more readable. I don't think we need to overcomplicate. Going to keep as is.

backend/ee/onyx/db/user_group.py
justin-tahara: I think your concern is valid but given this is an API that is only used for development I think it's fine for now. If we say this, there are a ton of more endpoints that we would need to update in order for this statement to be addressed across the board.

backend/ee/onyx/db/user_group.py
wenxi-onyx: I will say, I pref the simplicity of
```
current_user_ids = {user.id for user in db_user_group.users} 
new_user_ids = set(user_ids) - current_user_ids
```

Another option is to used the named operators
```
current_user_ids = {user.id for user in db_user_group.users}
incoming_user_ids = set(user_ids)

new_user_ids = incoming_user_ids.difference(current_user_ids)

if not new_user_ids:
    return db_user_group

user_group_update = UserGroupUpdate(
    user_ids=current_user_ids.union(new_user_ids),
    cc_pair_ids=[cc_pair.id for cc_pair in db_user_group.cc_pairs],
)
```

The combo of these two is probably the most concise and readable
```
current_user_ids = {user.id for user in db_user_group.users}
new_user_ids = set(user_ids) - current_user_ids 

if not new_user_ids:
    return db_user_group

user_group_update = UserGroupUpdate(
    user_ids=list(current_user_ids.union(new_user_ids)),
    cc_pair_ids=[cc_pair.id for cc_pair in db_user_group.cc_pairs]
)
```
Fine to just merge though

PR #6439, opened 2025-11-25
chore(google): Naming update from Vertex to Gemini

justin-tahara:
## Description

Might break up this PR but this is just slowly doing the migration for Gemini

## How Has This Been Tested?

## Additional Options

---
## Summary by cubic
Rename the Vertex AI provider to Google Gemini (Google Gen AI) with full backward compatibility. Updates backend mappings, model defaults, and UI copy/icons while keeping existing vertex_ai configs working.

- **Refactors**
  - Added google_genai provider and alias set with vertex_ai; normalize provider names across the app.
  - Keep litellm compatibility by mapping google_genai → vertex_ai where required.
  - Updated default/visible Gemini models and provider descriptor names.
  - Ensure vertex_credentials and vertex_location are passed when provider is in the alias set.
  - Use normalized provider for reasoning detection, token limits, and image-support checks.
  - Renamed embedding defaults (DEFAULT_GEMINI_MODEL) and added GOOGLE_GENAI_EMBEDDING_LOCAL_BATCH_SIZE (falls back to VERTEXAI_EMBEDDING_LOCAL_BATCH_SIZE).
  - Updated UI labels, icons, validation, docs/pricing links, and onboarding E2E to “Google Gen AI”.

- **Migration**
  - No breaking changes; existing vertex_ai configs continue to work.
  - Prefer GOOGLE_GENAI_EMBEDDING_LOCAL_BATCH_SIZE; the old env var still works.
  - New setups can use provider google_genai; vertex_ai remains supported.

<sup>Written for commit 475d2395d6432c9572e6ceffc2c8921d0a8c2dbd. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/llm/chat_llm.py
+++ b/backend/onyx/llm/chat_llm.py
@@ -42,6 +42,7 @@ from onyx.llm.interfaces import LLMConfig
 from onyx.llm.interfaces import STANDARD_TOOL_CHOICE_OPTIONS
 from onyx.llm.interfaces import ToolChoiceOptions
 from onyx.llm.llm_provider_options import AZURE_PROVIDER_NAME
+from onyx.llm.llm_provider_options import GOOGLE_GENAI_PROVIDER_ALIASES
 from onyx.llm.llm_provider_options import OLLAMA_PROVIDER_NAME
 from onyx.llm.llm_provider_options import VERTEX_CREDENTIALS_FILE_KWARG
 from onyx.llm.llm_provider_options import VERTEX_LOCATION_KWARG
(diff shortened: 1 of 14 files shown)


wenxi-onyx (changes requested):
Overall looks good - can you confirm that this will work seamlessly for people with vertex already configured in Onyx? 

E.g. test: configure vertex off of `main` --> switch to this branch --> confirm that gemini provider still works.

Additionally, I don't remember exactly how the model configurations are stored and updated in the DB. But we currently see annoying things where legacy models like `claude-2` are saved to the DB and then never removed because the users don't 'edit and update' their LLM provider config. 

I'm concerned that legacy vertex providers may continue to work, but will immediately break when we remove `LEGACY_VERTEX_PROVIDER_NAME` because the DB config is never changed.

PR #6441, opened 2025-11-25
fix: spinner during CSV load

Weves:
## Description

## How Has This Been Tested?

## Additional Options

---
## Summary by cubic
Shows a spinner while CSV data loads and prevents the mobile sidebar overlay from flashing on initial page load.

- **Bug Fixes**
  - CSV preview: added isFetching state, imported spinner.css, and replaced the skeleton with a spinner to cover both initial and fetch states.
  - Mobile sidebar: render the backdrop only after mount via useIsMounted to avoid the overlay flash.

<sup>Written for commit 889c3a2a93315414a8d97fd33809a219d18ded71. Summary will update automatically on new commits.</sup>

### DIFF
--- a/web/src/components/tools/CSVContent.tsx
+++ b/web/src/components/tools/CSVContent.tsx
@@ -10,7 +10,7 @@ import {
 } from "@/components/ui/table";
 import { ContentComponentProps } from "./ExpandableContentWrapper";
 import { WarningCircle } from "@phosphor-icons/react";
-import "../spinner.css";
+import SimpleLoader from "@/refresh-components/loaders/SimpleLoader";
 
 const CsvContent: React.FC<ContentComponentProps> = ({
   fileDescriptor,
@@ -79,7 +79,7 @@ const CsvContent: React.FC<ContentComponentProps> = ({
   if (isLoading || isFetching) {
     return (
       <div className="flex items-center justify-center h-[300px]">
-        <div className="loader ease-linear rounded-full border-8 border-t-8 border-background-200 h-8 w-8"></div>
+        <SimpleLoader />
       </div>
     );
   }


web/src/components/tools/CSVContent.tsx
```diff
@@ -68,25 +71,15 @@ const CsvContent: React.FC<ContentComponentProps> = ({
...
-            <div className="w-full space-y-3">
-              <div className="grid grid-cols-3 gap-4">
-                <div className="h-2 bg-background-200 rounded col-span-2"></div>
-                <div className="h-2 bg-background-200 rounded col-span-1"></div>
-              </div>
-              <div className="h-2 bg-background-200 rounded"></div>
-            </div>
-          </div>
-        </div>
+        <div className="loader ease-linear rounded-full border-8 border-t-8 border-background-200 h-8 w-8"></div>
```
raunakab: Is this the spinner? If so, I've created a nice utility spinner based off of @duo-onyx's spinner icon that he created. The name is `SimpleLoader`. It should already define a height, width, and stroke-colour (all overridable by passing in custom `className`).

web/src/components/tools/CSVContent.tsx
```diff
@@ -10,6 +10,7 @@ import {
 } from "@/components/ui/table";
 import { ContentComponentProps } from "./ExpandableContentWrapper";
 import { WarningCircle } from "@phosphor-icons/react";
+import "../spinner.css";
```
raunakab: If we remove this and use the `SimpleLoader`, will this be used anywhere else? If not, then let's remove it. No point in keeping dead-code around.

web/src/components/tools/CSVContent.tsx
raunakab: Updated.

PR #6454, opened 2025-11-26
chore(dev): run mypy and uv-sync on pre-commit

jmelahman:
## Description

Adds a pre-commit hook to run `mypy` via `uv`. 

This hook is expected to take a long time on the first run as it downloads all dependencies including torch's cuda deps, but afterwards, it should execute on the order of 100's of ms. This is also expected to be shared with the developer's venv going forward, so they typically will already have them installed anyways.

Folks may opt-out of these checks by setting `SKIP` in their shell, for example `export SKIP=uv-run` to skip the mypy hook.

Replaces the `optional-dependencies` with `dependency-groups`. Using dependency groups allows me to include them as `default-groups` which is what `uv sync` and `uv run` use to determine which packages to install. This in turn allows all packages to be installed when running `uv run mypy`. 

Adds a `uv-sync` hook. This should sync the packages when checking out or pulling new changes. Should help avoid using stale packages in dev. See also, https://github.com/astral-sh/uv-pre-commit. This requires a custom `default_install_hook_types`, so devs may need to re-run `pre-commit install` to install the hooks which trigger on checkout/pull.

## How Has This Been Tested?

`uv sync` and confirmed all packages are installed.

`uv run mypy .` (no errors except import-untyped for generated directory -- maybe we should stub this).

Confirmed mypy was running on pre-commit + incremental (only ran on changed files) + w/ caching

## Additional Options

- [x] Override Linear Check

---
## Summary by cubic
Add pre-commit hooks to run mypy via uv and auto-sync dependencies on checkout/merge. Switch to uv dependency groups to ensure the right packages are installed for local typing and consistent dev environments.

- **New Features**
  - Run mypy with uv-run in pre-commit; incremental on changed backend/*.py files; cached for speed.
  - Add uv-sync hook on post-checkout/merge/rewrite to install default groups and avoid stale deps.
  - Skip mypy in the CI pre-commit action to rely on the Python Checks job’s caching.
  - Action required: run pre-commit install to register the new hook types.

- **Dependencies**
  - Replace optional-dependencies with dependency-groups; set tool.uv default-groups = "all".
  - Update compile_requirements.py and requirement exports to use uv export --group with --no-default-groups.
  - Refresh backend/requirements/*.txt and README to use uv sync --group commands.

<sup>Written for commit ec10a96ac291576057453e51ccf61975b81f041a. Summary will update automatically on new commits.</sup>

### DIFF
--- a/.github/workflows/pr-quality-checks.yml
+++ b/.github/workflows/pr-quality-checks.yml
@@ -27,7 +27,7 @@ jobs:
         uses: hashicorp/setup-terraform@b9cd54a3c349d3f38e8881555d616ced269862dd # ratchet:hashicorp/setup-terraform@v3
       - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # ratchet:pre-commit/action@v3.0.1
         env:
-          # Mypy is covered by the Python Checks which caches dependencies better.
-          SKIP: mypy
+          # uv-run is mypy's id and mypy is covered by the Python Checks which caches dependencies better.
+          SKIP: uv-run
         with:
           extra_args: ${{ github.event_name == 'pull_request' && format('--from-ref {0} --to-ref {1}', github.event.pull_request.base.sha, github.event.pull_request.head.sha) || '' }}


.github/workflows/pr-quality-checks.yml
```diff
@@ -26,5 +26,8 @@ jobs:
       - name: Setup Terraform
         uses: hashicorp/setup-terraform@b9cd54a3c349d3f38e8881555d616ced269862dd # ratchet:hashicorp/setup-terraform@v3
       - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # ratchet:pre-commit/action@v3.0.1
+        env:
+          # Mypy is covered by the Python Checks which caches dependencies better.
+          SKIP: mypy
```
jmelahman:
```suggestion
          # uv-run is mypy's id and mypy is covered by the Python Checks which caches dependencies better.
          SKIP: uv-run
```

Weves (approved):

PR #6460, opened 2025-11-26
chore(deps): make backend/ a uv workspace

jmelahman:
## Description

Makes `backend/` a uv [workspace](https://docs.astral.sh/uv/concepts/projects/workspaces/#getting-started).

I think the two biggest reasons against this are:

1. it makes it slightly more cumbersome to sync the root venv (requires `uv sync --all-extras` instead of `uv sync`)
2. it might be difficult to support/QA for `backend/.venv/`, in particular enabling automatic syncing via a pre-commit hook -- devs might forget to sync this and use stale packages which would be confusing. `uv` seems to pretty aggressively steer users away from these sub-environments, requiring `--active` to be passed explicitly when preferring the non-default environment [[1](https://docs.astral.sh/uv/guides/migration/pip-to-project/#project-environments)].

Re: 1., we have the pre-commit hook to do this automatically, so I don't think it's that big of a deal. I think in the ideal workflow, this is done automatically and devs never need to run `uv sync` anyways.

Re: 2., I don't think it's too severe of an issue and I don't mind deferring to the engineer's discretion here.

Conversely, I think if we're going to have `pyproject.toml` files in the repo, it makes sense to consolidate them. If nothing else, this should avoid pitfalls where commands like `uv run` break in subdirectories needlessly. Also, there is precedence for a `backend/.venv/` and without this would be considered a breaking change to the workflow which I think is needless. This is doubly true if we introduce more proper python packages (like [chore(devtools): introduce the Onyx Developer Script, ods](https://github.com/onyx-dot-app/onyx/pull/6559)) in the future.

With a single environment setup, there is also an issue of undeclared dependencies. A package like the model_server might import a module without including it in its optional-dependency group. This will work in dev since all the packages are available from the root venv, but may fail in CI or when using a container. This is more or less the purpose of the integration/e2e tests, but with things like lazy_imports used widely in the codebase, I think the risk of not catching these issues in these tests increases. In other words, developing in narrowly scoped environments is safer.

## How Has This Been Tested?

```
uv venv
source .venv/bin/activate
uv sync --all-extras
  Audited 428 packages in 3ms
mypy --version
uv run --active mypy --version
```

```shell
cd backend
uv venv
source .venv/bin/activate
uv sync --active
  Audited 343 packages in 4ms
mypy --version
uv run --active mypy --version
```

## Additional Options

- [x] Override Linear Check

---
## Summary by cubic
Converted backend/ into a uv workspace with its own pyproject, and switched to optional extras for dependency management. Exports now use --extra, and tooling aligns with the workspace.

- **Dependencies**
  - Added backend/pyproject.toml (onyx-backend, Python >=3.11,<3.13) and depend on onyx[backend,ee] via workspace source.
  - Updated root uv.lock with workspace manifest members including onyx-backend.
  - Moved backend, ee, and model_server to [project.optional-dependencies]; dev stays in [dependency-groups].
  - Defined [tool.uv.workspace] members = ["backend", "tools/ods"].
  - Updated compile_requirements.py to use uv export --extra; regenerated backend requirements files.
  - Updated pre-commit: uv-sync runs with --locked and --all-extras.

<sup>Written for commit 09d505afef6c0405576cefa5e855099fbfe8f8e1. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/requirements/README.md
+++ b/backend/requirements/README.md
@@ -80,7 +80,7 @@ uv sync --extra backend
 uv sync --extra backend --group dev
 
 # For backend with EE (shared + backend + ee)
-uv sync --extra backend --group ee
+uv sync --extra backend --extra ee
 
 # For model server (shared + model_server, NO backend deps!)
 uv sync --extra model_server


backend/pyproject.toml
```diff
@@ -1,3 +1,14 @@
+[project]
+name = "onyx-backend"
+version = "0.0.0"
+requires-python = ">=3.11,<3.13"
+dependencies = [
+    "onyx[backend]",
```
jmelahman:
```suggestion
    "onyx[backend,ee]",
```
maybe?

Weves: I think it should include ee

PR #6469, opened 2025-11-27
feat: mcp pass through oauth

evan-onyx:
## Description

Addresses https://linear.app/danswer/issue/DAN-3121/mcp-pass-through-oauth

Implement pass-through oauth for MCP servers for google_oauth and oidc Onyx instances.

Deployers of Onyx can set up their own MCP servers that use the OAuth access token their users have as a result of logging in to Onyx for authn/z.

## How Has This Been Tested?

Tested in UI using new google oauth MCP server script `run_mcp_server_google_oauth.py`. Also added some external dep unit tests, since we currently don't have a way to run i.e. playwright tests in auth modes other than basic. I have an uncommitted test file for that eventuality too.

## Additional Options

---
## Summary by cubic
Adds pass-through OAuth (PT_OAUTH) for MCP servers, letting Onyx forward the user’s login OAuth token (Google OAuth or OIDC) to MCP via the Authorization header. This enables auth without storing server credentials and is selectable in the admin UI when OAuth is enabled.

- **New Features**
  - Added PT_OAUTH auth type and backend handling to treat users as authenticated and skip stored connection configs.
  - Discovery and tool runs send the user’s OAuth token in Authorization; MCPTool supports token passthrough at runtime.
  - Admin UI: new “Pass-through OAuth” option (shown only when OAuth is enabled) with updated validation and guidance.
  - Test coverage for PT_OAUTH flow plus an e2e helper and a Google OAuth MCP test server script for local validation.

- **Refactors**
  - Web auth constants converted to an AuthType enum and replaced string checks across auth pages, routing, and settings.

<sup>Written for commit 09f8917f05ccacacfc10f8f59404290dc9c225c3. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/tests/external_dependency_unit/tools/test_mcp_passthrough_oauth.py
+++ b/backend/tests/external_dependency_unit/tools/test_mcp_passthrough_oauth.py
@@ -401,7 +401,7 @@ class TestMCPPassThroughOAuth:
             print(response.rich_response)
             assert isinstance(response.rich_response, CustomToolCallSummary)
             print(response.rich_response.tool_result)
-            assert response.rich_response.tool_result == mocked_response
+            assert response.rich_response.tool_result["tool_result"] == mocked_response
 
         # Verify Authorization header was set with the user's OAuth token
         assert "Authorization" in captured_headers


backend/onyx/db/enums.py
```diff
@@ -86,6 +86,7 @@ class MCPAuthenticationType(str, PyEnum):
     NONE = "NONE"
     API_TOKEN = "API_TOKEN"
     OAUTH = "OAUTH"
+    PT_OAUTH = "PT_OAUTH"
```
justin-tahara: I know we want to abbreviate but is this standard to use PT?

web/src/app/admin/actions/ActionEditor.tsx
```diff
@@ -85,7 +86,8 @@ function ActionForm({
   const [methodSpecs, setMethodSpecs] = methodSpecsState;
   const [showAdvancedOptions, setShowAdvancedOptions] = useState(false);
   const authType = useAuthType();
-  const isOAuthEnabled = authType === "oidc" || authType === "google_oauth";
+  const isOAuthEnabled =
+    authType === AuthType.OIDC || authType === AuthType.GOOGLE_OAUTH;
```
justin-tahara, line 90: Thanks for cleaning this!

web/src/lib/tools/interfaces.ts
```diff
@@ -2,6 +2,7 @@ export enum MCPAuthenticationType {
   NONE = "NONE",
   API_TOKEN = "API_TOKEN",
   OAUTH = "OAUTH",
+  PT_OAUTH = "PT_OAUTH",
```
justin-tahara: Just going back to original comment of should we rename this. I keep thinking physical training

web/tests/e2e/utils/mcpServer.ts
```diff
@@ -210,3 +210,61 @@ export async function startMcpApiKeyServer(
...
+  const pythonBinary = options.pythonBinary || "python3";
+  const readyTimeout = options.readyTimeoutMs ?? READY_TIMEOUT_MS;
+  const requiredScopes = options.requiredScopes || [];
+
+  const scriptPath =
+    options.scriptPath ||
+    path.resolve(
+      __dirname,
+      "../../../..",
+      "backend/tests/integration/mock_services/mcp_test_server/run_mcp_server_google_oauth.py"
```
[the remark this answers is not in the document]
justin-tahara: Is this valid?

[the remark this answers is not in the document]
evan-onyx: I added the file, but also the tests that use this can't be run anyways in the current paradigm where we just use basic auth to test

backend/onyx/db/enums.py
evan-onyx: probably not... open to suggestions! The annoying thing is it can't be >9 characters or we have to do a migration for the db column

PR #6475, opened 2025-11-28
refactor(MCP): mcp backend and schema

Subash-Mohan:
## Description
This pull request introduces significant improvements to how MCP servers are managed, especially focusing on configuration status tracking, nullable authentication fields, and filtering of MCP-connected tools. The main changes include adding a new `status` field to the `MCPServer` model and database table, making several authentication-related fields nullable, updating queries and APIs to use the new status, and providing new endpoints and filtering logic for MCP tools.

## How Has This Been Tested?

## Additional Options

---
## Summary by cubic
Adds MCP server status tracking and makes auth fields optional to support staged setup. Users only see tools from connected MCP servers, with new admin endpoints and a lightweight admin UI to create/manage servers and OpenAPI actions, run auth, refresh tools, and toggle enablement.

- **New Features**
  - Added MCPServer.status (CREATED, AWAITING_AUTH, FETCHING_TOOLS, CONNECTED, DISCONNECTED).
  - Made transport, auth_type, auth_performer nullable for simple server creation.
  - Filtered tool listings to only connected MCP servers (user list_tools, persona queries).
  - New admin endpoints:
    - GET /server/{id}/tools/snapshots?source=db|mcp
    - PATCH /server/{id}/status
    - POST /server (simple create), PATCH /server/{id} (simple update)
    - GET /tools (all MCP tools), PATCH /tool/status (bulk enable/disable)
    - GET /oauth-config/{id}
  - API models: MCPServer includes status and tool_count; ToolSnapshot includes enabled.
  - Admin UI for MCP and OpenAPI actions: add/edit servers/tools, authenticate (admin/user), refresh discovery, per-tool toggle, bulk enable/disable.

- **Migration**
  - Alembic: adds status column and makes auth fields nullable; existing servers set to CONNECTED.
  - No manual steps.

<sup>Written for commit 63ba6ec3f1b35bc04d05fd6cef7fbf2606158a19. Summary will update automatically on new commits.</sup>


backend/alembic/versions/e8f0d2a38171_add_status_to_mcp_server_and_make_auth_.py
```diff
@@ -0,0 +1,113 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # Make auth fields nullable
+    op.alter_column(
+        "mcp_server",
+        "transport",
+        existing_type=sa.Enum(MCPTransport, name="mcp_transport", native_enum=False),
```
evan-onyx, line 30: Why would we transport to be nullable? It's set to a default on the frontend and passed through so imo we should never need to?

backend/alembic/versions/e8f0d2a38171_add_status_to_mcp_server_and_make_auth_.py
```diff
@@ -0,0 +1,113 @@
...
+        nullable=True,
+    )
+
+    op.alter_column(
+        "mcp_server",
+        "auth_type",
+        existing_type=sa.Enum(
+            MCPAuthenticationType, name="mcp_authentication_type", native_enum=False
```
evan-onyx, line 38: same here

backend/alembic/versions/e8f0d2a38171_add_status_to_mcp_server_and_make_auth_.py
```diff
@@ -0,0 +1,113 @@
...
+        ),
+        nullable=True,
+    )
+
+    op.alter_column(
+        "mcp_server",
+        "auth_performer",
+        existing_type=sa.Enum(
+            MCPAuthenticationPerformer,
```
evan-onyx, line 47: same

backend/onyx/db/enums.py
```diff
@@ -101,6 +101,14 @@ class MCPAuthenticationPerformer(str, PyEnum):
     PER_USER = "PER_USER"
 
 
+class MCPServerStatus(str, PyEnum):
+    CREATED = "CREATED"  # Server created, needs auth configuration
+    AWAITING_AUTH = "AWAITING_AUTH"  # Auth configured, pending user authentication
+    FETCHING_TOOLS = "FETCHING_TOOLS"  # Auth complete, fetching tools
+    CONNECTED = "CONNECTED"  # Fully configured and connected
```
evan-onyx, line 109: The problem with these states is that Oauth servers will end up out of sync with what's in the DB because tokens expire. I also don't think we need a status for fetching tools since that's an extremely transient state and logically overlaps with i.e. "connected" or "disconnected". I also don't like that this is effectively duplicate information stored in the db: as I mentioned we can't actually use connected/disconnected due to how OAuth auth works, awaiting_auth is equivalent to "is there a connection config saved", and created is just "there is no connection config saved". If you want to write a function that returns an element of this enum based on the db model I'm fine with that, but I'm also doubtful it's that helpful

backend/alembic/versions/e8f0d2a38171_add_status_to_mcp_server_and_make_auth_.py
```diff
@@ -0,0 +1,113 @@
...
+        ),
+    )
+
+    # For existing records, mark status as CONNECTED
+    bind = op.get_bind()
+    bind.execute(
+        sa.text(
+            """
+        UPDATE mcp_server
+        SET status = 'CONNECTED'
```
evan-onyx, line 71: see my comment below w.r.t why I don't think these should be stored in the db, but regardless this is DEFINITELY wrong. using UPDATE....SET without a where clause will just make everything connected, which would be inconsistent with partially configured servers. Stuff like this is further reason for not storing this status in the db.

backend/alembic/versions/e8f0d2a38171_add_status_to_mcp_server_and_make_auth_.py
```diff
@@ -0,0 +1,113 @@
+"""add status to mcp server and make auth fields nullable
```
evan-onyx, line 1: long story short I don't believe we need this migration, see below comments

backend/onyx/db/mcp.py
```diff
@@ -61,9 +62,14 @@ def get_mcp_servers_for_persona(
     if not mcp_server_ids:
         return []
 
-    # Fetch the MCP servers
+    # Fetch the MCP servers (include connected servers only)
```
evan-onyx: this does not feel right to me; as I mentioned before you don't even _know_ what servers are truly connected until you try talking to them. Even API key servers can have keys changed, None auth servers could be taken down, etc.

backend/onyx/db/mcp.py
```diff
@@ -90,9 +96,9 @@ def create_mcp_server__no_commit(
     name: str,
     description: str | None,
     server_url: str,
-    auth_type: MCPAuthenticationType,
-    transport: MCPTransport,
-    auth_performer: MCPAuthenticationPerformer,
+    auth_type: MCPAuthenticationType | None,
+    transport: MCPTransport | None,
+    auth_performer: MCPAuthenticationPerformer | None,
```
evan-onyx, line 96: again, why? these should have sensible defaults that are passed in from the frontend. There's no reason to let a user just leave these fields empty imo.

evan-onyx (changes requested):
There's probably some context I'm missing for why these things are necessary/useful, happy to chat about it!

backend/alembic/versions/e8f0d2a38171_add_status_to_mcp_server_and_make_auth_.py
Subash-Mohan: Changed the auth fields to nullable because of UI changes.

backend/alembic/versions/e8f0d2a38171_add_status_to_mcp_server_and_make_auth_.py
Subash-Mohan: Changed the auth fields to nullable because of UI changes

backend/alembic/versions/e8f0d2a38171_add_status_to_mcp_server_and_make_auth_.py
Subash-Mohan: Changed the auth fields to nullable because of UI changes

backend/alembic/versions/e8f0d2a38171_add_status_to_mcp_server_and_make_auth_.py
Subash-Mohan: Now i added the where condition to check whether the mcp server has `admin_connection_config_id`.

backend/onyx/db/enums.py
Subash-Mohan: Since we discussed this yesterday, I am closing this.

PR #6484, opened 2025-11-29
[FIX] Fix citation document mismatch and standardize citation format

weijia619:
### Problem 1
There was a frontend rendering issue where the wrong citation is being displayed. The root cause was the citation number was incorrectly used as a direct array index. 

### Solution 1
Now the citations map is used to find which document_id corresponds to citation number, then find that document in the docs array.

### Problem 2
Frontend had to handle two different citation map formats:
     - Streaming: `{citation_num: document_id}` (e.g., `{1: "doc_abc"}`)
     - Database: `{citation_num: db_doc_id}` (e.g., `{1: 45}`)
     
### Solution 2
Convert citations to streaming format `{citation_num:document_id}` when loading from database

### Known Limitations
  - **Assumes `document_id` is unique within one message**: The citation map uses `document_id` as the key, which works correctly as long as each document has a unique ID within a single message context. 

## How Has This Been Tested?

Screenshot 2025-11-28 at 9 05 40 PM

## Additional Options

---
## Summary by cubic
Fixed mismatched citation rendering by resolving documents via the citation map and standardized all citation maps to the streaming format to keep frontend logic consistent.

- **Bug Fixes**
  - Use the citation map to map citation number → document_id before rendering the cited doc.
  - Convert DB citation maps to streaming format `{citation_num: document_id}` to match live stream behavior.

- **Refactors**
  - Unified chat history helper to `create_chat_history_chain` for simpler message loading.
  - Added migrations for new chat history, internet search providers (incl. Exa seeding), the Open URL tool, and the Python Code Interpreter tool.
  - Hardened CI: explicit workflow permissions, disabled checkout credentials, Slack failure notifications, and added Zizmor scanning.
  - Docker/build: run backend as non-root, pin uv version, split bake targets for backend/model server, and improved cache settings.
  - Agent framework updates: new message formatting utilities, tracing spans, and web search client scaffolding; changed encoder error logs to warnings.

<sup>Written for commit 3b83d2579c3e0efd14848fbc702dd908636dfec2. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/server/query_and_chat/models.py
+++ b/backend/onyx/server/query_and_chat/models.py
@@ -247,8 +247,8 @@ class ChatMessageDetail(BaseModel):
     tool_call: ToolCallFinalResult | None = None
     message_type: MessageType
     context_docs: RetrievalDocs | None = None
-    # Dict mapping document_id to citation number (matches streaming format)
-    citations: dict[str, int] | None = None
+    # Dict mapping citation number to document_id
+    citations: dict[int, str] | None = None
     time_sent: datetime
     files: list[FileDescriptor]
     error: str | None = None
--- a/backend/onyx/db/chat.py
+++ b/backend/onyx/db/chat.py
@@ -896,8 +896,7 @@ def translate_db_message_to_chat_message_detail(
         if latest_feedback.is_positive is not None:
             current_feedback = "like" if latest_feedback.is_positive else "dislike"
 
-    # Convert citations from {citation_num: db_doc_id} to {document_id: citation_num}
-    # to match the streaming format
+    # Convert citations from {citation_num: db_doc_id} to {citation_num: document_id}
     converted_citations = None
     if chat_message.citations and chat_message.search_docs:
         # Build lookup map: db_doc_id -> document_id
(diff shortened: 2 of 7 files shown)


backend/onyx/server/query_and_chat/models.py
```diff
@@ -243,8 +243,8 @@ class ChatMessageDetail(BaseModel):
     tool_call: ToolCallFinalResult | None = None
     message_type: MessageType
     context_docs: RetrievalDocs | None = None
-    # Dict mapping citation number to db_doc_id
-    citations: dict[int, int] | None = None
+    # Dict mapping document_id to citation number (matches streaming format)
+    citations: dict[str, int] | None = None
```
yuhongsun96: Should this use CitationInfo class?

yuhongsun96 (approved):
Nice!

backend/onyx/server/query_and_chat/models.py
weijia619: updated the code to use `CitationMap = { [citation_num: number]: string };` enableing O(1) complexity for looking up the citation numbers while list[CitationInfo] makes it into O(n)

PR #6491, opened 2025-12-01
fix: icon coloring in Renderer

Weves:
## Description

Screenshot 2025-11-30 at 6 30 44 PM

Icon was previously invisible

## How Has This Been Tested?

## Additional Options

---
## Summary by cubic
Fixed icon color inheritance in chat renderers so icons match theme and loading shimmer across light/dark modes.

- **Bug Fixes**
  - Icons now inherit color (SvgCode uses currentColor; Python CodeIcon forwards size/props).
  - MultiToolRenderer applies shimmer color to icons and keeps loading shimmer on status text only.
  - Added shimmer-base and shimmer-highlight variables and Tailwind tokens for consistent theming.

<sup>Written for commit a4c52a51605521b037563127d784750336e94be4. Summary will update automatically on new commits.</sup>

### DIFF
--- a/web/src/app/chat/message/messageComponents/MultiToolRenderer.tsx
+++ b/web/src/app/chat/message/messageComponents/MultiToolRenderer.tsx
@@ -233,7 +233,7 @@ export default function MultiToolRenderer({
 
                           <div
                             className={cn(
-                              "text-base flex items-center gap-1 loading-text mb-2",
+                              "text-base flex items-center gap-1 mb-2",
                               toolsToDisplay.length > 1 &&
                                 isLastItem &&
                                 "cursor-pointer hover:text-text-900 transition-colors"
--- a/web/src/app/globals.css
+++ b/web/src/app/globals.css
@@ -460,7 +460,7 @@ textarea {
   background-clip: text;
   -webkit-background-clip: text;
   -webkit-text-fill-color: transparent;
-  animation: shimmerTransition 2.5s ease-out infinite;
+  animation: shimmerTransition 1.8s ease-out infinite;
 }
 
 .dark .loading-text {


web/src/icons/code.tsx
```diff
@@ -9,6 +9,7 @@ const SvgCode = (props: SVGProps<SVGSVGElement>) => (
   >
     <path
       d="M10.6667 12L14.6667 8L10.6667 4M5.33334 4L1.33334 8L5.33334 12"
+      stroke="currentColor"
```
raunakab, line 12: Hmm, I believe this hardcodes the stroke-colour, making setting the stroke-colour from the calling context to be ignored (i.e., `<SvgCode className="... stroke-red-500" />` will not work).

web/src/app/css/colors.css
```diff
@@ -228,6 +228,10 @@
 
 /* Light Colors */
 :root {
+  /* Shimmer colors for loading animations */
+  --shimmer-base: #a3a3a3;
+  --shimmer-highlight: #000000;
```
raunakab, line 233: Ideally, we have `shimmer-base` and `shimmer-highlight` be variable *references* only, instead of defining the raw colours themselves.

raunakab: I also believe these are not @duo-onyx defined? They don't follow the naming scheme that he's been using 😅..

web/src/app/chat/message/messageComponents/renderers/PythonToolRenderer.tsx
```diff
@@ -47,12 +47,8 @@ function constructCurrentPythonState(packets: PythonToolPacket[]) {
...
 }
 
-function CodeIcon(props: IconProps) {
-  return (
-    <div className="w-4 h-4">
-      <SvgCode />
-    </div>
-  );
+function CodeIcon({ size = 16, ...props }: IconProps) {
+  return <SvgCode width={size} height={size} {...props} />;
```
raunakab, line 51: Hmm, the `height` and `width` props don't do anything on our SVG components, from what I recall.

Ideally, we just directly call `SvgCode` from whatever context it is required in; there's no real need for a wrapper here, I believe.

PR #6516, opened 2025-12-02
feat(agents pagination): FE changes for pagination to the agents admin page

acaprau:
## Description

This PR adds a frontend component for pagination, and uses this component for the agents admin page.

Note with this change the drag and drop display priority feature will only work for agents all on the same page.

Some screenshots of what the component looks like for different item counts:
Screenshot 2025-12-04 at 15 13 15
The following screenshots have button styling from a previous version of this PR, but still show the different possible button arrangements.
Screenshot 2025-12-03 at 16 34 02
Screenshot 2025-12-03 at 16 35 36
Screenshot 2025-12-03 at 16 35 56
Screenshot 2025-12-03 at 16 36 03

## How Has This Been Tested?

Manually adding agents in the UI with a small enough page size to get pagination. Switching between pages, reordering agents within a page, reloading the page all seem to work fine. Also tried deleting agents to the point where pagination is no longer relevant and that worked fine too.

Small observation, by default agents have no display priority. This means if you go to some page > 1 and rearrange the agents there, those relevant agents will now have some non-null display prio and jump to the earliest available page. Technically "correct" behavior but I wonder if we should assign some default display prio value when creating new agents.

## Additional Options

---
## Summary by cubic
Added pagination to the Admin Assistants page and connected it to the paginated agents API. This makes large assistant lists easier to browse and keeps display priorities correct across pages.

- **New Features**
  - Backend: added include_default to both admin and user agents pagination; total counts respect it.
  - Admin UI: new Pagination component with page controls (PAGE_SIZE = 20) wired through useAdminPersonas to the paginated endpoint.
  - Display priority: PersonasTable now offsets priorities by page to maintain global ordering.
  - Tests: integration tests updated to pass include_default.

<sup>Written for commit cbc28283f6cbc3102c9d523f04e2b9fbf588ea04. Summary will update automatically on new commits.</sup>

### DIFF
--- a/web/src/app/admin/assistants/page.tsx
+++ b/web/src/app/admin/assistants/page.tsx
@@ -13,7 +13,7 @@ import { ThreeDotsLoader } from "@/components/Loading";
 import { ErrorCallout } from "@/components/ErrorCallout";
 import SvgOnyxOctagon from "@/icons/onyx-octagon";
 import { useState, useEffect } from "react";
-import { Pagination } from "@/refresh-components/Pagination";
+import Pagination from "@/refresh-components/Pagination";
 
 const PAGE_SIZE = 20;
 
--- a/web/src/refresh-components/Pagination.tsx
+++ b/web/src/refresh-components/Pagination.tsx
@@ -11,7 +11,7 @@ interface PaginationProps {
   onPageChange: (page: number) => void;
 }
 
-export function Pagination({
+export default function Pagination({
   currentPage,
   totalPages,
   onPageChange,
@@ -113,7 +113,6 @@ export function Pagination({
       <IconButton
         onClick={() => onPageChange(currentPage + 1)}
         disabled={currentPage === totalPages}
-        main
         tertiary
         icon={SvgChevronRight}
       />


backend/onyx/server/features/persona/api.py
```diff
@@ -196,6 +196,9 @@ def get_agents_admin_paginated(
     get_editable: bool = Query(
         False, description="If true, only returns editable personas."
     ),
+    include_default: bool = Query(
```
acaprau, line 203: we add this arg bc the fe was previously filtering for default agents. in order for pagination to work correctly, we need to do this filtering in the be.

web/src/components/pagination/Pagination.tsx
```diff
@@ -0,0 +1,114 @@
...
+          </Button>
+        );
+      })}
+
+      {/* Next button */}
+      <Button
+        onClick={() => onPageChange(currentPage + 1)}
+        disabled={currentPage === totalPages}
+        tertiary
+        className="px-3 py-2"
```
raunakab: Should avoid using custom padding! The `Button` component has been created with the appropriate paddings already applied.

web/src/components/pagination/Pagination.tsx
```diff
@@ -0,0 +1,114 @@
...
+      >
+        <ChevronRight className="h-4 w-4" />
```
raunakab: Two things here:

1. Prefer to use icons provided in the `web/src/icons` directory rather than icons from an external library. Here, you could use `SvgChevronRight` instead.
2. The `Button` component already exposes a `leftIcon` and `rightIcon` component. You can pass in icons to one of those props instead! `<Button leftIcon={SvgChevronRight} ... />`.

web/src/components/pagination/Pagination.tsx
```diff
@@ -0,0 +1,114 @@
...
+        const isActive = pageNum === currentPage;
+
+        return (
+          <Button
+            key={pageNum}
+            onClick={() => onPageChange(pageNum)}
+            tertiary={!isActive}
+            className={`px-3 py-2 min-w-[40px] ${
+              isActive ? "bg-background-strong text-inverted" : ""
+            }`}
```
raunakab: Oof, we should definitely remove this. Custom padding should be discouraged, and the text/background colours should not be custom specified either.

raunakab: If you want `active`, try giving the `transient` prop on the `Button` component instead. It's a boolean flag that can be specified.

web/src/components/pagination/Pagination.tsx
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        const pageNum = page as number;
+        const isActive = pageNum === currentPage;
+
+        return (
+          <Button
+            key={pageNum}
+            onClick={() => onPageChange(pageNum)}
+            tertiary={!isActive}
```
raunakab: `tertiary` isn't exactly a "state", per se, but rather a "variant" of a button. For example, we have "main" buttons, "action" buttons, and "danger" buttons. They all denote different "styles" of buttons. Those styles shouldn't be toggled on or off by state, however.

If you want to alter the state of a button, try looking at `disabled` / `transient` instead!

web/src/components/pagination/Pagination.tsx
```diff
@@ -0,0 +1,114 @@
...
+        <ChevronLeft className="h-4 w-4" />
+      </Button>
+
+      {/* Page numbers */}
+      {pageNumbers.map((page, index) => {
+        if (page === "...") {
+          return (
+            <span key={`ellipsis-${index}`} className="px-3 py-2 text-text-500">
+              ...
+            </span>
```
raunakab: `text-text-500` is a Tailwind specified colour. We're trying to move away from those!

Let's try to always render all text with a `refresh-components/texts/Text` tag!

```tsx

<Text as="span" ...>
  ...
</Text>
```

web/src/components/pagination/Pagination.tsx
```diff
@@ -0,0 +1,114 @@
...
+  return (
+    <div className="flex items-center justify-center gap-1 mt-4">
+      {/* Previous button */}
+      <Button
+        onClick={() => onPageChange(currentPage - 1)}
+        disabled={currentPage === 1}
+        tertiary
+        className="px-3 py-2"
+      >
+        <ChevronLeft className="h-4 w-4" />
```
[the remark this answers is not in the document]
raunakab: Yup! Agree with Greptile here.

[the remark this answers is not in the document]
raunakab: Also, throw `SvgChevronLeft` into the `leftIcon` or `rightIcon` prop that is available directly on the `Button` component.

web/src/refresh-components/Pagination.tsx
raunakab, line 1: If this is a reusable component, please throw it inside of `refresh-components` instead!

raunakab: `web/src/components` is where all the old, legacy components are. We're slowly trying to migrate away from them!

web/src/app/admin/assistants/page.tsx
```diff
@@ -12,16 +12,30 @@ import { useAdminPersonas } from "./hooks";
 import { Persona } from "./interfaces";
 import { ThreeDotsLoader } from "@/components/Loading";
 import { ErrorCallout } from "@/components/ErrorCallout";
+import { useState } from "react";
+import { Pagination } from "@/components/pagination/Pagination";
+
+const PAGE_SIZE = 3;
```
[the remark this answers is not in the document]
raunakab: Seems like `3` was used for debugging purposes?

raunakab (commented):
Added a few small comments on the usage of certain FE components!

web/src/components/pagination/Pagination.tsx
raunakab: Oh, in fact, you should use an `IconButton` here.

The general gist is that when you need a regular button with text as children (with optional left / right icons), you use the `Button` component.

When you have to render *just* an icon, you use the `IconButton`.

web/src/hooks/useAdminPersonas.ts
raunakab, line 1: Hmm, this is a *super* small nit, but could we move this hook to `web/src/hook`, please? I've been trying to clean up all our disparate hook locations into one, unified place. I think the `web/src/hooks` directory would be a fine place for this one to live!

raunakab: Unless there are other considerations to this that I'm missing.

web/src/refresh-components/Pagination.tsx
```diff
@@ -0,0 +1,122 @@
...
+            {pageNum}
+          </Button>
+        );
+      })}
+
+      {/* Next button */}
+      <IconButton
+        onClick={() => onPageChange(currentPage + 1)}
+        disabled={currentPage === totalPages}
+        main
```
raunakab: Don't have to fix this, since it's a nanoscopic nit, but `main` here is technically not needed, since that is the default variant!

web/src/refresh-components/Pagination.tsx
```diff
@@ -0,0 +1,122 @@
...
+        tertiary
+        icon={SvgChevronRight}
```
raunakab, line 117: Awesome!

web/src/refresh-components/Pagination.tsx
```diff
@@ -0,0 +1,122 @@
...
+        const isActive = pageNum === currentPage;
+
+        return (
+          <Button
+            key={pageNum}
+            onClick={() => onPageChange(pageNum)}
+            main
+            tertiary
+            transient={isActive}
+            className={cn("min-w-[40px]")}
```
raunakab, line 105: Do we need this one here?

web/src/refresh-components/Pagination.tsx
```diff
@@ -0,0 +1,122 @@
...
+import Text from "@/refresh-components/texts/Text";
+import { cn } from "@/lib/utils";
+
+interface PaginationProps {
+  currentPage: number;
+  totalPages: number;
+  onPageChange: (page: number) => void;
+}
+
+export function Pagination({
```
raunakab: This is a super useful component! Thanks for creating it <3!

One minor thing: could we make this the *default* export of this file, please?

web/src/refresh-components/Pagination.tsx
acaprau: claude says this is nice bc otherwise the button component will fit its content, which can make inconsistent spacing for page 1 vs 10 for example. lmk of we shouldnt be doing this

PR #6524, opened 2025-12-02
chore(gha): persist docker cache intra-PR builds

jmelahman:
## Description

ECR cache cost has been negligible, so we can cache more aggressively. There's a decent amount of thrashing when these images change from package versions bumps which I'd like to make commonplace, so optimize for that a bit.

In short, this means if a docker image was built once for a PR, we'll continue building that image from the cache.

Previously, we only stored layers from the previous 1 build (global) in the cache, so if the current branch or the previous builds branch are sufficient diverged, the likelihood of cache hit was low.

## How Has This Been Tested?

Captured by existing

## Additional Options

- [x] Override Linear Check

---
## Summary by cubic
Persist Docker build caches per PR in the integration, MIT integration, and Playwright CI workflows to speed up repeat builds and avoid cross-branch cache collisions.

- **New Features**
  - Generate a cache suffix from the PR number or a sanitized branch name.
  - Read/write per-PR cache tags for backend, model-server, web, and integration images, with shared cache and latest-image fallbacks.
  - Applied to pr-integration-tests.yml, pr-mit-integration-tests.yml, and pr-playwright-tests.yml using Docker Buildx/Bake; respects DOCKER_NO_CACHE.

- **Dependencies**
  - Increase Dependabot cooldown from 4 to 7 days for GitHub Actions and npm.

<sup>Written for commit cc6930ee44f103953f8f438ecf54056395d35b48. Summary will update automatically on new commits.</sup>

### DIFF
--- a/.github/workflows/pr-integration-tests.yml
+++ b/.github/workflows/pr-integration-tests.yml
@@ -84,6 +84,7 @@ jobs:
           if [ -n "${PR_NUMBER}" ]; then
             CACHE_SUFFIX="${PR_NUMBER}"
           else
+            # shellcheck disable=SC2001
             CACHE_SUFFIX=$(echo "${REF_NAME}" | sed 's/[^A-Za-z0-9._-]/-/g')
           fi
           echo "cache-suffix=${CACHE_SUFFIX}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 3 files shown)


jmelahman:
@greptile

jmelahman:
@greptile

jmelahman:
@greptile

justin-tahara (approved):

PR #6532, opened 2025-12-02
chore(tracing): add tracing to new backend

rguan72:
## Description

add tracing to new framework

## How Has This Been Tested?

tested locally

## Additional Options

---
## Summary by cubic
Adds structured tracing to the new backend, instrumenting the agent loop, model generations, and tool calls, and wiring Braintrust and Langfuse through a unified tracing framework. This improves visibility into inputs, outputs, token usage, and errors for each request.

- **New Features**
  - Wrapped run_llm_loop in a top-level trace with tenant_id.
  - Wrapped run_llm_step in generation_span; records input, usage, final output.
  - Added function_span for tools; captures args/output and attaches error details on failure.

- **Refactors**
  - Migrated Braintrust to the new tracing framework (BraintrustTracingProcessor + set_trace_processors); removed group_id from exported traces.
  - Integrated Langfuse via OpenInference (OITracer + OpenInferenceTracingProcessor) and registered through set_trace_processors.

<sup>Written for commit a23adbee1fddc76089e27c2b9d561dfac4930a0d. Summary will update automatically on new commits.</sup>


backend/onyx/chat/llm_loop.py
```diff
@@ -520,94 +526,110 @@ def run_llm_step(
...
     accumulated_answer = ""
 
-    for packet in llm.stream(
-        prompt=llm_msg_history,
-        tools=tool_definitions,
-        tool_choice=tool_choice,
-        structured_response_format=None,  # TODO
-    ):
-        delta = packet.choice.delta
-        finish_reason = packet.choice.finish_reason
```
evan-onyx, line 530: is finish_reason not needed?

rguan72: yeah, I had to delete finish_reason in order to get the usage packet at the end -- but in general I think we should be ending the stream when the stream ends and we don't need to end it early

rguan72: tested and it doesn't impact functionality at all

PR #6535, opened 2025-12-02
fix: eager load persona in slack channel config

nmgarza5:
## Description

Fix Slack bot falling back to default assistant instead of using the configured persona. See thread: https://onyx-dot-app.slack.com/archives/C056265VB1N/p1763721492214309

**Root Cause**: The persona relationship on SlackChannelConfig uses lazy loading. When `handle_regular_answer` accesses slack_channel_config.persona, the SQLAlchemy session context may have changed, causing the lazy load to fail silently and return None. This triggers the fallback to `DEFAULT_PERSONA_ID` (the default "Assistant"), which doesn't have the custom tools configured on the user's intended persona.

**Fix**: Add joinedload(SlackChannelConfig.persona) to eagerly load the persona relationship in fetch_slack_channel_config_for_channel_or_default, ensuring the persona is loaded in the same query and eliminating dependency on lazy loading.

Additional: Added info/warning logs to track which assistant is being resolved for Slack messages, making future debugging easier.

## How Has This Been Tested?
  - Unit tests
  - Manual testing with Slack bot to verify correct persona is loaded
  - Verified logs show correct assistant resolution without fallback warnings
  
## Additional Options

---
## Summary by cubic
Fixes the Slack bot using the wrong assistant by eagerly loading the channel’s configured persona. Prevents fallback to the default assistant and restores persona-specific tools.

- **Bug Fixes**
  - Eagerly load SlackChannelConfig.persona via joinedload in fetch_slack_channel_config_for_channel_or_default.
  - Add info/warning logs in handle_regular_answer to trace which persona is used.

<sup>Written for commit c7f9b4adb9639a81d9da7fec4e6947921264ee98. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/db/slack_channel_config.py
+++ b/backend/onyx/db/slack_channel_config.py
@@ -2,6 +2,7 @@ from collections.abc import Sequence
 from typing import Any
 
 from sqlalchemy import select
+from sqlalchemy.orm import joinedload
 from sqlalchemy.orm import Session
 
 from onyx.configs.chat_configs import MAX_CHUNKS_FED_TO_CHAT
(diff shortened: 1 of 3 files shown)


nmgarza5:
@greptile

evan-onyx (approved):
great find! imo we should stop passing SQLAlchemy objects around except for cases where there's a specific reason we want lazy loading, and even once we do that we should typically be eager loading to avoid extra db queries

PR #6537, opened 2025-12-02
chore(deployment): check if tagged correctly

jmelahman:
## Description

Adds a step to the deployment pipeline to assert the git tag is correct. That is to say, the previous stable tag is an ancestor, the version bump is continuous, and no larger tag is an ancestor.

**valid**
```
zxczxc - (v0.0.3)
asdasd - (v0.0.2)
qweqwe - (v0.0.1)
```

**valid**
```
foobar - (v0.0.3-cloud)
zxczxc - (v0.0.3)
asdasd - (v0.0.2)
qweqwe - (v0.0.1)
```

**invalid**
```
zxczxc - (v0.0.2)
asdasd - (v0.0.3)
qweqwe - (v0.0.1)
```

**invalid**
```
zxczxc - (v0.0.4)
asdasd - (v0.0.2)
qweqwe - (v0.0.1)
```
etc.

Generally, any tag generated with `uv run tag` would be an acceptable tag.

This is currently non-blocking but posts to #monitor-deployments on failure. We can change this if there's interest.

## Additional Options

---
## Summary by cubic
Adds a tag validation step to the deployment workflow to catch mis-tagged releases early, and sends a Slack alert if the check fails.

- **New Features**
  - Adds check-version-tag job to deployment.yml.
  - Installs release-tag and runs uv run --no-sync tag --check to verify tag ancestry and continuous version bump.
  - Adds a Slack notification job that alerts on tag check failure.

- **Dependencies**
  - Adds release-tag >=0.4.3 to dev dependencies.

<sup>Written for commit bf8b594a14ec5cc0e965296300be6967f00dcd07. Summary will update automatically on new commits.</sup>

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -83,6 +83,51 @@ jobs:
             echo "sanitized-tag=$SANITIZED_TAG"
           } >> "$GITHUB_OUTPUT"
 
+  check-version-tag:
+    runs-on: ubuntu-slim
+    timeout-minutes: 10
+    if: ${{ !startsWith(github.ref_name, 'nightly-latest') && github.event_name != 'workflow_dispatch' }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # ratchet:actions/checkout@v6
+        with:
+          persist-credentials: false
+
+      - name: Setup uv
+        uses: astral-sh/setup-uv@1e862dfacbd1d6d858c55d9b792c756523627244 # ratchet:astral-sh/setup-uv@v7.1.4
+        with:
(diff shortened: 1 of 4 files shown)


.github/workflows/deployment.yml
```diff
@@ -83,6 +83,51 @@ jobs:
             echo "sanitized-tag=$SANITIZED_TAG"
           } >> "$GITHUB_OUTPUT"
 
+  check-version-tag:
+    runs-on: ubuntu-slim
```
justin-tahara, line 87: We're okay with these instances? Or should we move to runs on?

justin-tahara (approved):
One comment

.github/workflows/deployment.yml
jmelahman: Yep, they're fine. They're free since we're OSS and this step is small/isn't doing anything that'd benefit from AWS infra.

PR #6539, opened 2025-12-02
fix: expand special casing around sharepoint shared drives

evan-onyx:
## Description

Shared Drives are differently named across languages, but they appear to share the property that there is a discrepancy between the internal name for the drive and its user-facing name. We've expanded the special casing to be consistent across the board and added cases for german and spanish.

## How Has This Been Tested?

Unit tests

## Additional Options

---
## Summary by cubic
Handle SharePoint default document library names across English, German, and Spanish by mapping internal drive names to their user-facing library names. Fixes drive matching and permissions so syncs work regardless of site language.

- **Bug Fixes**
  - Added SHARED_DOCUMENTS_MAP (+ reverse) to translate between Graph drive names and SharePoint list titles.
  - Applied mapping in drive discovery, item fetching, and checkpoint resume logic; updated permission utils to use reverse mapping.
  - Added unit tests covering English, German, and Spanish name pairs to prevent regressions.
  - Updated SharePoint connector help text to note current support for English, Spanish, and German when specifying sites.

<sup>Written for commit c8ce6247ccc2223d30f3475afe240d6a8f2cb777. Summary will update automatically on new commits.</sup>

### DIFF
--- a/web/src/lib/connectors/connectors.tsx
+++ b/web/src/lib/connectors/connectors.tsx
@@ -813,6 +813,7 @@ export const connectorConfigs: Record<
         description: `• If no sites are specified, all sites in your organization will be indexed (Sites.Read.All permission required).
 • Specifying 'https://onyxai.sharepoint.com/sites/support' for example only indexes this site.
 • Specifying 'https://onyxai.sharepoint.com/sites/support/subfolder' for example only indexes this folder.
+• Specifying sites currently works for SharePoint instances using English, Spanish, or German. Contact the Onyx team if you need another language supported.
 `,
       },
     ],


web/src/lib/connectors/connectors.tsx
```diff
@@ -813,6 +813,7 @@ export const connectorConfigs: Record<
         description: `• If no sites are specified, all sites in your organization will be indexed (Sites.Read.All permission required).
 • Specifying 'https://onyxai.sharepoint.com/sites/support' for example only indexes this site.
 • Specifying 'https://onyxai.sharepoint.com/sites/support/subfolder' for example only indexes this folder.
+• Specifying sites currently works for SharePoint instances using English, Spanish, or German. Contact the Onyx team if you need another language supported.
```
wenxi-onyx, line 816: Please make sure to add this to Sharepoint docs

PR #6550, opened 2025-12-03
feat: Some UI enhancements for tools

yuhongsun96:
## Description

- OpenURL now has a start packet and has a backup if the page fails to open
- UI no longer shows sub-text under the step
- No longer flashes updates to the text
- Internal Search now visually is 2 steps
- Updated icons for some of the built-in actions
- Minor copy change

## How Has This Been Tested?

Ran locally for sending messages and replaying sessions

## Additional Options

---
## Summary by cubic
Introduced a 3-stage OpenURL streaming flow (start → urls → documents) and updated the UI to show a stable “Reading” experience without flicker. Internal search now renders as two steps. Increased the agent cycle limit to 6 to allow one more tool step before answering.

- **New Features**
  - Added OpenURL packets: open_url_start, open_url_urls, open_url_documents, and updated backend emitters/translators to send them.
  - Frontend handles these packets with a consistent “Reading” state, showing documents when ready or URLs if none are produced.
  - Raised MAX_LLM_CYCLES from 5 to 6.

- **UI Improvements**
  - Removed sub-text under steps and stopped text from flashing during updates.
  - Internal Search now visually shows two steps (queries → documents).
  - Updated icons for built-in actions and made minor copy tweaks.

<sup>Written for commit 07d3bbdba0d100bef5431c51f4cc8748758ff50f. Summary will update automatically on new commits.</sup>

### DIFF
--- a/web/src/app/chat/message/messageComponents/MultiToolRenderer.test.tsx
+++ b/web/src/app/chat/message/messageComponents/MultiToolRenderer.test.tsx
@@ -10,7 +10,14 @@ import {
   createToolGroups,
   createMockChatState,
   renderMultiToolRenderer,
+  createInternalSearchToolGroup,
 } from "@tests/setup/multiToolTestHelpers";
+
+// The search tool renderers use ResultIcon, which pulls in complex source metadata.
+// For these tests we only care about statuses/text, so mock it to avoid heavy deps.
+jest.mock("@/components/chat/sources/SourceCard", () => ({
+  ResultIcon: () => <div data-testid="result-icon" />,
+}));
 
 // Mock the RendererComponent to return predictable, simple output
 jest.mock("./renderMessageComponent", () => ({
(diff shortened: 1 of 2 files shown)


backend/onyx/server/query_and_chat/streaming_utils.py
```diff
@@ -209,14 +211,31 @@ def create_custom_tool_packets(
...
+                turn_index=turn_index,
+                obj=OpenUrlStart(),
+            )
+        )
+        # Emit URLs packet
         packets.append(
             Packet(
                 turn_index=turn_index,
-                obj=OpenUrl(documents=SearchDoc.from_saved_search_docs(fetch)),
+                obj=OpenUrlUrls(urls=urls),
```
Subash-Mohan: Is it expected that we send the same set of URLs for every packet?

yuhongsun96: Ah true true

yuhongsun96: Some weird leftover code from the previous implementation where it was tied to the web search tool, let me update

PR #6552, opened 2025-12-03
fix: Reasoning Block Linebreaks

yuhongsun96:
## Description

Previously the reasoning blocks would not correctly have newlines. This fixes it. It has to do with the monkey patching of LiteLLM and handling the different summary items.

## How Has This Been Tested?

Verified locally with streaming flows and replaying past sessions.

## Additional Options

---
## Summary by cubic
Fixes incorrect line breaks in reasoning blocks so multi-part summaries render as separate paragraphs in both streaming and non-streaming responses.

- **Bug Fixes**
  - Streaming (OpenAI Responses API): Track summary_index and prepend double newlines when a new summary part starts.
  - Non-streaming: Patch transform_response to concatenate multiple summary items with double newlines and update reasoning_content in the result.

<sup>Written for commit 123a531469fe65891708687efb986ac8c6295aaa. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/llm/litellm_singleton/monkey_patches.py
+++ b/backend/onyx/llm/litellm_singleton/monkey_patches.py
@@ -12,6 +12,9 @@ from typing import Union
 
 from litellm import AllMessageValues
 from litellm.completion_extras.litellm_responses_transformation.transformation import (
+    LiteLLMResponsesTransformationHandler,
+)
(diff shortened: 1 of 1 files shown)


rguan72:
i don't 100% understand this, but for newlines, might make sense to have this processing on a higher level of abstraction than the Litellm monkey patch (for example, in our rendering logic)

rguan72:
altho this is also fine

PR #6565, opened 2025-12-03
feat(agents admin page): Make display priority adjustments PATCH instead of PUT, allowing granular edits + small cleanups

acaprau:
## Description

This PR allows adjusting the display priorities of agents in the agents admin page granularity as opposed to all at once for all agents. This implies a change from PUT to PATCH for the relevant method. This PR also includes some small hygiene around resource naming and the persona db layer.

This PR is needed for #6516, which introduces pagination to the agents admin page, and for which bulk edits to display priorities will not work.

## How Has This Been Tested?

I dragged agents around in the admin page, examined network activity, and refreshed the admin page, and everything seems to work as intended.

## Additional Options

---
## Summary by cubic
Scoped agent display-priority updates to personas the current admin can access and moved the endpoint to the admin agents router. This secures the operation, supports partial updates, and aligns with the Linear task to update agent display priority.

- **Refactors**
  - Replaced update_all_personas_display_priority with update_personas_display_priority (validates IDs against user-accessible personas; updates only provided IDs).
  - Moved route to admin_agents_router PATCH /display-priorities and renamed handler to patch_agents_display_priorities.
  - get_raw_personas_for_user now uses a shared base query.

- **Migration**
  - Update clients to call the admin agents display-priorities endpoint; request body stays the same.

<sup>Written for commit 9934ae6799893c1f74adb1a1c08be048a1d8df41. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/tests/unit/onyx/db/test_persona_display_priority.py
+++ b/backend/tests/unit/onyx/db/test_persona_display_priority.py
@@ -13,6 +13,7 @@ def _persona(persona_id: int, display_priority: int) -> SimpleNamespace:
 def test_update_display_priority_updates_subset(
     monkeypatch: pytest.MonkeyPatch,
 ) -> None:
+    # Precondition
     persona_a = _persona(1, 5)
     persona_b = _persona(2, 6)
     db_session = MagicMock()
@@ -22,16 +23,19 @@ def test_update_display_priority_updates_subset(
         lambda user, db_session, **kwargs: [persona_a, persona_b],
     )
 
+    # Under test
     update_personas_display_priority(
         {persona_a.id: 0}, db_session, user, commit_db_txn=True
     )
 
+    # Postcondition
     assert persona_a.display_priority == 0
     assert persona_b.display_priority == 6
     db_session.commit.assert_called_once_with()
 
 
 def test_update_display_priority_invalid_ids(monkeypatch: pytest.MonkeyPatch) -> None:
+    # Precondition
     persona_a = _persona(1, 5)
     db_session = MagicMock()
     user = MagicMock()
@@ -40,9 +44,14 @@ def test_update_display_priority_invalid_ids(monkeypatch: pytest.MonkeyPatch) ->
         lambda user, db_session, **kwargs: [persona_a],
     )
 
+    # Under test
     with pytest.raises(ValueError):
         update_personas_display_priority(
-            {persona_a.id: 0, 99: 1}, db_session, user, commit_db_txn=True
+            {persona_a.id: 0, 99: 1},
+            db_session,
+            user,
+            commit_db_txn=True,
         )
 
+    # Postcondition
     db_session.commit.assert_not_called()


backend/onyx/db/persona.py
```diff
@@ -702,19 +712,54 @@ def mark_delete_persona_by_name(
...
+
+    for persona_id, priority in display_priority_map.items():
+        if persona_id not in available_personas_map:
+            raise ValueError(
+                "Invalid persona ID provided: Persona with ID {} was not found for this user."
+            )
+
+        available_personas_map[persona_id].display_priority = priority
+
+    if commit_db_txn:
```
acaprau, line 760: opinion weakly held but this design allows for composing different queries under one commit and callers can choose when to commit, if we dont like it i can revert this diff i just thought it might make things more explicit.

backend/onyx/server/features/persona/api.py
```diff
@@ -155,15 +155,16 @@ def patch_persona_default_status(
         raise HTTPException(status_code=403, detail=str(e))
 
 
-@admin_router.put("/display-priority")
-def patch_persona_display_priority(
+@admin_agents_router.patch("/display-priorities")
```
acaprau, line 158: improving the semantics of this endpoint and resource

backend/onyx/db/persona.py
```diff
@@ -648,15 +659,14 @@ def get_raw_personas_for_user(
...
-    stmt = _build_persona_filters(
-        stmt, include_default, include_slack_bot_personas, include_deleted
+    stmt = _build_persona_base_query(
+        user, get_editable, include_default, include_slack_bot_personas, include_deleted
     )
     return db_session.scalars(stmt).all()
 
 
 def get_personas(db_session: Session) -> Sequence[Persona]:
+    """WARNING: Unsafe, can fetch personas from all users."""
```
justin-tahara, line 669: What's the point of this. Is this more of a note?

backend/onyx/server/features/persona/api.py
```diff
@@ -155,16 +155,22 @@ def patch_persona_default_status(
...
+    try:
+        update_personas_display_priority(
+            display_priority_map=display_priority_request.display_priority_map,
+            db_session=db_session,
+            user=user,
+            commit_db_txn=True,
+        )
+    except ValueError as e:
+        logger.exception("Failed to update agent display priorities.")
+        raise HTTPException(status_code=403, detail=str(e))
```
justin-tahara, line 173: Is 403 the correct error code? I guess this is saying that this operation was forbidden for whatever reason?

backend/onyx/db/persona.py
acaprau: so this was a pre-existing docstring on `get_personas_by_ids`, and i notice this function has the same issue. long term we should delete these but im not doing this now, this should just help remind devs on the day to day to think before using this function

backend/onyx/server/features/persona/api.py
acaprau: yep, the client tried to reprio an agent for which it doesnt have access

PR #6568, opened 2025-12-03
refactor(API): replace redundant api key dep from ingestion endpoints

wenxi-onyx:
- Simplify optional_user code --> no functional change
- Remove api_key_dep as it is only used for the ingestion API and is redundant with current_user dep

## Description

## How Has This Been Tested?

## Additional Options

---
## Summary by cubic
Replaced the ingestion endpoints’ API key dependency with the standard current_user and simplified auth checks to reduce duplication and ensure consistent behavior across SAML/JWT, PAT, and API key.

- **Refactors**
  - Removed api_key_dep and its references in router auth checks.
  - Ingestion endpoints now use Depends(current_user).
  - optional_user now returns immediately on successful SAML/JWT, then falls back to PAT before API key, with safer error handling and a warning on token issues.
  - Ingestion endpoints now accept PATs in addition to API keys.

<sup>Written for commit 74a0bc8f77fc4b9fbf77b08e64becb6b91034f88. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/auth/users.py
+++ b/backend/onyx/auth/users.py
@@ -1250,7 +1250,7 @@ async def optional_user(
             user = await fetch_user_for_pat(hashed_pat, async_db_session)
         elif hashed_api_key := get_hashed_api_key_from_request(request):
             user = await fetch_user_for_api_key(hashed_api_key, async_db_session)
-    except Exception:
+    except ValueError:
         logger.warning("Issue with validating authentication token")
         return None
 


backend/onyx/server/onyx_api/ingestion.py
```diff
@@ -43,7 +43,7 @@
 @router.get("/connector-docs/{cc_pair_id}")
 def get_docs_by_connector_credential_pair(
     cc_pair_id: int,
-    _: User | None = Depends(api_key_dep),
+    _: User | None = Depends(current_user),
     db_session: Session = Depends(get_session),
```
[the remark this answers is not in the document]
wenxi-onyx: This will be addressed in a follow up PR that clarifies 401 403 semantics in our API

backend/onyx/auth/users.py
```diff
@@ -1240,22 +1240,19 @@ async def optional_user(
...
-    if user is None:
-        try:
-            hashed_api_key = get_hashed_api_key_from_request(request)
-        except ValueError:
-            hashed_api_key = None
-        if hashed_api_key:
+    try:
+        if hashed_pat := get_hashed_pat_from_request(request):
+            user = await fetch_user_for_pat(hashed_pat, async_db_session)
+        elif hashed_api_key := get_hashed_api_key_from_request(request):
```
evan-onyx, line 1251: technically this is different from before; we used to check for user by api key if a pat existed but fetch_user_for_pat returned None. This new behavior is probably fine, but imo people probably shouldn't be specifying both? Maybe we can tell the user they're not valid if the request contains both?

evan-onyx (approved):
one thought + agreed we should solidify when we're giving back 401s and 403s in general

backend/onyx/auth/users.py
wenxi-onyx: AI review also said this, and I think it's invalid.

Both PAT and API key are passed as bearer tokens in auth header. You _can_ also pass via `X-Onyx-api-key`, but this is not documented anywhere except in the codebase. The only way you could have both PAT and API key is if you had both of these headers, which would be extremely weird to do in the first place IMO. Yes, we could tell people "don't have both set," but i'm honestly in favor of just removing the custom header. I'm assuming it was added because someone was trying to use Onyx API through some auth gateway that needed separate creds

PR #6605, opened 2025-12-04
feat: prompt cache 3

evan-onyx:
## Description

Actual implementation and test for prompt caching

## How Has This Been Tested?

external dep unit tests to check that the framework itself works. unit tests to ensure that the framework is used TBD

## Additional Options

---
## Summary by cubic
Adds a provider-agnostic prompt caching framework and wires it into chat and indexing flows to cut prompt costs on repeated context. Supports OpenAI, Anthropic, and Vertex AI with best-effort behavior and no API changes for callers.

- New Features
  - Added process_with_prompt_cache for strings and chat messages, including continuation handling.
  - Provider adapters: OpenAI (implicit), Anthropic (cache_control ephemeral), Vertex AI (explicit with content-level cache_control).
  - Chat integration: marks cacheable system/history messages and splits the cacheable prefix from the live suffix automatically.
  - Message schema adds cache_control at message/content-part levels; tenant-aware CacheManager with deterministic keys and metadata.
  - ModelResponse now includes usage and cache metrics (prompt/completion tokens, cache_creation_input_tokens, cache_read_input_tokens).
  - Configs: ENABLE_PROMPT_CACHING (default on) and PROMPT_CACHE_REDIS_TTL_MULTIPLIER. README and external dependency tests validate cost reductions.

- Migration
  - Enabled by default; set ENABLE_PROMPT_CACHING=false to disable.
  - No API changes required; existing llm.invoke flows continue to work.
  - Optional: use process_with_prompt_cache directly in custom LLM calls for fine-grained control.

<sup>Written for commit b06001518400ce2f7d7625d91c5c419f9cbe4b63. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/llm/prompt_cache/providers/factory.py
+++ b/backend/onyx/llm/prompt_cache/providers/factory.py
@@ -19,7 +19,7 @@ def get_provider_adapter(provider: str) -> PromptCacheProvider:
     """
     if provider == LlmProviderNames.OPENAI:
         return OpenAIPromptCacheProvider()
-    elif provider == LlmProviderNames.ANTHROPIC:
+    elif provider in [LlmProviderNames.ANTHROPIC, LlmProviderNames.BEDROCK]:
         return AnthropicPromptCacheProvider()
     elif provider == LlmProviderNames.VERTEX_AI:
         return VertexAIPromptCacheProvider()


backend/onyx/chat/llm_loop.py
```diff
@@ -125,6 +125,8 @@ def construct_message_history(
     if history_token_budget < 0:
         raise ValueError("Not enough tokens available to construct message history")
 
+    system_prompt.should_cache = True
```
yuhongsun96: Is anything needed for the deep research flow or is it all covered?

backend/onyx/chat/llm_step.py
```diff
@@ -277,15 +278,31 @@ def _extract_tool_call_kickoffs(
...
     Converts ChatMessageSimple messages to ChatCompletionMessage format,
     handling different message types and image files for multimodal support.
     """
     messages: list[ChatCompletionMessage] = []
+    last_cacheable_msg_idx = -1
+    all_previous_msgs_cacheable = True
+
+    for idx, msg in enumerate(history):
+        # if the message is being added to the history
+        if msg.message_type in [
```
yuhongsun96, line 295: Does it cost anything to cache the tool call response? might as well if it's cheap/free. It may be reused with sequential tool calls (also heavily in deep research)

backend/onyx/chat/llm_step.py
```diff
@@ -395,7 +412,7 @@ def translate_history_to_llm_format(
             assistant_msg_with_tool = AssistantMessage(
                 role="assistant",
                 content=None,  # The tool call is parsed, doesn't need to be duplicated in the content
-                tool_calls=tool_calls if tool_calls else None,
+                tool_calls=tool_calls or None,
```
yuhongsun96, line 416: brooo so truuuuu

backend/onyx/configs/model_configs.py
```diff
@@ -128,3 +128,17 @@
         LITELLM_EXTRA_BODY = json.loads(_LITELLM_EXTRA_BODY_RAW)
     except Exception:
         pass
+
+#####
+# Prompt Caching Configs
+#####
+# Enable prompt caching framework
+ENABLE_PROMPT_CACHING = (
```
yuhongsun96, line 136: Can it just always be on? Why would one want to turn it off?

backend/onyx/llm/prompt_cache/README.md
```diff
@@ -0,0 +1,167 @@
...
+## Quick Start
+
+### Basic Usage
+
+```python
+from onyx.llm.prompt_cache import process_with_prompt_cache
+from onyx.llm.factory import get_default_llms
+
+# Get your LLM instance
+llm, _ = get_default_llms()
```
yuhongsun96: Some of this is outdated

backend/onyx/llm/prompt_cache/README.md
```diff
@@ -0,0 +1,167 @@
+# Prompt Caching Framework
```
yuhongsun96, line 1: Do we support Bedrock/Azure?

Also what happens if there is a cache miss (like the day line in the prompt ticked over to the next one). I'm guessing it fails gracefully because of a redis miss and then we cache everything for the next run?

yuhongsun96 (approved):
lgtm

backend/onyx/chat/llm_loop.py
evan-onyx: good catch, need to do DR

evan-onyx: actually jk I accounted for this, doing the caching in construct_message_history means it gets used for both flows. We could in theory mess around with it more for DR but I think this is good for now

backend/onyx/chat/llm_step.py
evan-onyx: to my understanding this is the right thing to do, we want to try to cache a strict prefix of the actual messages we're sending

backend/onyx/configs/model_configs.py
evan-onyx: I just think it's nice to have an easy way to turn new stuff off, it's default on though! Because yeah it should be a strict positive

backend/onyx/llm/prompt_cache/README.md
evan-onyx: I haven't tested it, but according to their docs I can use my anthropic cache provider! made that change. Azure _should_ also be implicit.

and yeah we don't have to worry about cache misses

PR #6619, opened 2025-12-05
feat(llm): fetch dynamic provider models directly from source APIs

nmgarza5:
## Description

  For dynamic LLM providers (OpenRouter, Bedrock, Ollama), fetch models directly from source APIs instead of using LiteLLM's static lists. Source
  APIs are more up-to-date and provide better metadata.

  Changes

  Backend - Model Fetching
  - Remove LiteLLM filter from Bedrock model fetch
  - Return empty built-in model lists for dynamic providers
  - Add display name generators for Bedrock and Ollama models
  - Strip redundant vendor prefixes from OpenRouter display names (e.g., "Microsoft: Phi 4" → "Phi 4")

  Backend - Auto-Sync to Database
  - New models are automatically synced to DB when fetching (no save/update click required)
  - Existing model preferences (is_visible, max_input_tokens) are preserved during sync
  - Add sync_model_configurations() function with INSERT ON CONFLICT DO NOTHING

  Backend - Token Limit Fallback Chain
  - Add get_bedrock_token_limit() with lookup order: model ID suffix (:200k) → LiteLLM → hardcoded mapping → 4096 default

  Database
  - Add migration for display_name column in model_configuration table

  Frontend
  - Pass provider_name when fetching models to trigger auto-sync
  - Use display_name from backend instead of generating locally

  Code Organization
  - Add onyx/server/manage/llm/utils.py for provider utilities
  - Add onyx/llm/constants.py for model name mappings and token limits

  Tests
  - test_utils.py - Display name generation, model validation, vision/reasoning inference (34 tests)
  - test_bedrock_token_limit.py - Token limit fallback chain (18 tests)
  - test_llm_sync.py - DB sync logic (5 tests)
  - test_fetch_models_api.py - Endpoint flow with mocked HTTP (8 tests)

## How Has This Been Tested?

unit tests and manually in the UI

## Additional Options

---
## Summary by cubic
Fetch dynamic LLM models directly from provider APIs (OpenRouter, Bedrock, Ollama) for fresher model lists and clearer display names. Persist display_name end-to-end and skip LiteLLM enrichment/static lists for these providers.

- **New Features**
  - Fetch models from source APIs for OpenRouter, Bedrock, and Ollama; built‑in lists now empty; management API can sync fetched models to DB when provider_name is provided.
  - Persist display_name in model_configuration; backend and UI pass it through on save.
  - Bedrock: prefer inference profiles (cross‑region), include display names, context limits, and vision support, with a fallback name generator; removed LiteLLM filtering.
  - OpenRouter: include display_name, strip redundant vendor prefixes, and return context limits and vision support.
  - Ollama: generate display_name from tags and return context limits and vision support.

- **Migration**
  - Adds display_name column to model_configuration. Run DB migrations.

<sup>Written for commit fc9b951d6bcf00b62bb78c0fd2ee42ad459ce1f9. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/llm/utils.py
+++ b/backend/onyx/llm/utils.py
@@ -432,43 +432,40 @@ def get_llm_contextual_cost(
     return usd_per_prompt + usd_per_completion
 
 
-def get_llm_max_tokens(
+def llm_max_input_tokens(
     model_map: dict,
     model_name: str,
     model_provider: str,
 ) -> int:
-    """Best effort attempt to get the max tokens for the LLM"""
+    """Best effort attempt to get the max input tokens for the LLM."""
     if GEN_AI_MAX_TOKENS:
         # This is an override, so always return this
         logger.info(f"Using override GEN_AI_MAX_TOKENS: {GEN_AI_MAX_TOKENS}")
         return GEN_AI_MAX_TOKENS
 
-    try:
-        model_obj = find_model_obj(
-            model_map,
-            model_provider,
-            model_name,
-        )
-        if not model_obj:
-            raise RuntimeError(
-                f"No litellm entry found for {model_provider}/{model_name}"
-            )
-
-        if "max_input_tokens" in model_obj:
-            max_tokens = model_obj["max_input_tokens"]
-            return max_tokens
-
-        if "max_tokens" in model_obj:
-            max_tokens = model_obj["max_tokens"]
-            return max_tokens
-
-        raise RuntimeError("No max tokens found for LLM")
-    except Exception:
+    model_obj = find_model_obj(
+        model_map,
+        model_provider,
+        model_name,
+    )
+    if not model_obj:
         logger.warning(
             f"Model '{model_name}' not found in LiteLLM. "
             f"Falling back to {GEN_AI_MODEL_FALLBACK_MAX_TOKENS} tokens."
         )
         return GEN_AI_MODEL_FALLBACK_MAX_TOKENS
+
+    if "max_input_tokens" in model_obj:
+        return model_obj["max_input_tokens"]
+
+    if "max_tokens" in model_obj:
+        return model_obj["max_tokens"]
+
+    logger.warning(
+        f"No max tokens found for '{model_name}'. "
+        f"Falling back to {GEN_AI_MODEL_FALLBACK_MAX_TOKENS} tokens."
+    )
+    return GEN_AI_MODEL_FALLBACK_MAX_TOKENS
 
 
 def get_llm_max_output_tokens(
@@ -476,31 +473,32 @@ def get_llm_max_output_tokens(
     model_name: str,
     model_provider: str,
 ) -> int:
-    """Best effort attempt to get the max output tokens for the LLM"""
-    try:
-        model_obj = model_map.get(f"{model_provider}/{model_name}")
-        if not model_obj:
-            model_obj = model_map[model_name]
-        else:
-            pass
+    """Best effort attempt to get the max output tokens for the LLM."""
+    default_output_tokens = int(GEN_AI_MODEL_FALLBACK_MAX_TOKENS)
 
-        if "max_output_tokens" in model_obj:
-            max_output_tokens = model_obj["max_output_tokens"]
-            return max_output_tokens
+    model_obj = model_map.get(f"{model_provider}/{model_name}")
+    if not model_obj:
+        model_obj = model_map.get(model_name)
 
-        # Fallback to a fraction of max_tokens if max_output_tokens is not specified
-        if "max_tokens" in model_obj:
-            max_output_tokens = int(model_obj["max_tokens"] * 0.1)
-            return max_output_tokens
-
-        raise RuntimeError("No max output tokens found for LLM")
-    except Exception:
-        default_output_tokens = int(GEN_AI_MODEL_FALLBACK_MAX_TOKENS)
+    if not model_obj:
         logger.warning(
             f"Model '{model_name}' not found in LiteLLM. "
             f"Falling back to {default_output_tokens} output tokens."
         )
         return default_output_tokens
+
+    if "max_output_tokens" in model_obj:
+        return model_obj["max_output_tokens"]
+
+    # Fallback to a fraction of max_tokens if max_output_tokens is not specified
+    if "max_tokens" in model_obj:
+        return int(model_obj["max_tokens"] * 0.1)
+
+    logger.warning(
+        f"No max output tokens found for '{model_name}'. "
+        f"Falling back to {default_output_tokens} output tokens."
+    )
+    return default_output_tokens
 
 
 def get_max_input_tokens(
@@ -517,7 +515,7 @@ def get_max_input_tokens(
     litellm_model_map = get_model_map()
 
     input_toks = (
-        get_llm_max_tokens(
+        llm_max_input_tokens(
             model_name=model_name,
             model_provider=model_provider,
             model_map=litellm_model_map,


nmgarza5:
@greptile

backend/onyx/server/manage/llm/api.py
```diff
@@ -573,48 +579,104 @@ def get_bedrock_available_models(
...
+
+        # Filter patterns for non-LLM models (image gen, embeddings, etc.)
+        non_llm_patterns = {"embed", "stable-", "titan-image", "titan-embed"}
+
+        for model in model_summaries:
+            model_id = model.get("modelId", "")
+            model_id_lower = model_id.lower()
+
+            # Skip non-LLM models
+            if not model_id:
```
wenxi-onyx: I personally pref concision here
```
if not mode_id
   or any()
   or not model.get():
   continue
```

backend/onyx/server/manage/llm/api.py
```diff
@@ -573,48 +579,104 @@ def get_bedrock_available_models(
...
-        # Available inference profiles. Invoking these allows cross-region inference (preferred over base models).
+        model_info: dict[str, dict] = {}
+        available_models: set[str] = set()
+
+        # Filter patterns for non-LLM models (image gen, embeddings, etc.)
+        non_llm_patterns = {"embed", "stable-", "titan-image", "titan-embed"}
+
+        for model in model_summaries:
+            model_id = model.get("modelId", "")
+            model_id_lower = model_id.lower()
```
wenxi-onyx: since this is only used once, maybe just have the lower inside the `any()`?

backend/onyx/server/manage/llm/api.py
```diff
@@ -573,48 +579,104 @@ def get_bedrock_available_models(
...
             ).get("inferenceProfileSummaries", [])
             for profile in inference_profiles:
-                if profile_id := profile.get("inferenceProfileId"):
-                    profile_ids.add(profile_id)
-
-                    # The model id is everything after the first period in the profile id
-                    if "." in profile_id:
-                        model_id = profile_id.split(".", 1)[1]
-                        cross_region_models.add(model_id)
+                profile_id = profile.get("inferenceProfileId")
```
wenxi-onyx: Are we feeling anti walrus operator? I'm a bit of a walrus enjoyer, but I can see how it makes the code slightly harder to read. It does save a line here though...

backend/onyx/server/manage/llm/api.py
wenxi-onyx: This chunk of code seems quite similar to the inference profile check. Should this be a helper function? Something like 
```
def _valid_bedrock_model(model_id) -> bool:
   # Empty model id can be explicitly early returned or just allowed to go to default return
   if any(pattern in model_id.lower() etc....
      or  not model.get(responseStreaming)
         return False
   return True
```
You may need to confirm whether or not inference profiles always stream --> there doesn't seem to be a check in that section

backend/onyx/server/manage/llm/api.py
```diff
@@ -573,48 +579,104 @@ def get_bedrock_available_models(
...
+                                "supports_image_input": False,
+                            }
         except Exception as e:
-            # Cross-region inference isn't guaranteed; ignore failures here.
             logger.warning(f"Couldn't fetch inference profiles for Bedrock: {e}")
 
         # Prefer profiles: de-dupe available models, then add profile IDs
         candidates = (available_models - cross_region_models) | profile_ids
 
-        # Keep only models we support (compatibility with litellm)
```
wenxi-onyx, line 615: If we don't cross reference litellm, we may not have token limit/context limit data for certain models. This results in a 4k token limit (quite low, but conservative). 

Is it possible to retrieve it from Bedrock similar to how we retrieve it from Ollama and OpenRouter?

backend/onyx/server/manage/llm/api.py
```diff
@@ -573,48 +579,104 @@ def get_bedrock_available_models(
...
+                            "display_name": f"{base_info['display_name']} ({region})",
+                            "supports_image_input": base_info["supports_image_input"],
+                        }
+                    else:
+                        # Generate display name from profile name if available
+                        profile_name = profile.get("inferenceProfileName", "")
+                        if profile_name:
+                            model_info[profile_id] = {
+                                "display_name": f"{profile_name} ({region})",
+                                "supports_image_input": False,
```
wenxi-onyx: If a cross region inference profile is available, but the base model is not (i.e. someone in EU east accessing models only hosted on EU west), this erroneously sets `supports_image_input` to False.

backend/onyx/server/manage/llm/api.py
```diff
@@ -573,48 +579,104 @@ def get_bedrock_available_models(
...
-        available_models = {
-            model.get("modelId", "")
-            for model in model_summaries
-            if model.get("modelId")
-            and "embed" not in model.get("modelId", "").lower()
-            and model.get("responseStreamingSupported", False)
-        }
-
-        # Available inference profiles. Invoking these allows cross-region inference (preferred over base models).
+        model_info: dict[str, dict] = {}
```
wenxi-onyx: Should this/can the inner dict here be more strictly typed? E.g. TypedDict or TypedTuple. 

Maybe something like
```
class ModelMetadata(TypedDict):
   model_id: str
   display_name: str
   supports_image_input: bool

model_info = dict[str, ModelMetadata] = {}
```

Then, retrieving these values can assume that they're never None and the expected type

backend/onyx/server/manage/llm/models.py
```diff
@@ -199,13 +235,37 @@ def from_model(
...
+            )
+
+            return cls(
+                name=model_configuration_model.name,
+                is_visible=model_configuration_model.is_visible,
+                max_input_tokens=model_configuration_model.max_input_tokens,
+                supports_image_input=(
+                    model_configuration_model.supports_image_input or False
+                ),
+                supports_reasoning=False,  # Dynamic providers don't have this metadata
```
wenxi-onyx: This one is a tiny bit sus to me. Not exactly sure what to do if the providers don't actually have this metadata, but OpenRouter, for example, has GPT 5 models and open weight thinking models. 

Would be good to double check this.

backend/onyx/server/manage/llm/models.py
```diff
@@ -305,6 +372,7 @@ class OpenRouterModelDetails(BaseModel):
     model_config = {"extra": "ignore"}
 
     id: str
+    name: str  # Display name from OpenRouter (e.g., "OpenAI: GPT-5 Pro")
```
wenxi-onyx: pref `display_name` to stay consistent with others

backend/onyx/server/manage/llm/api.py
```diff
@@ -628,6 +690,153 @@ def get_bedrock_available_models(
         )
 
 
+def _generate_bedrock_display_name(model_id: str) -> str:
```
wenxi-onyx: Is there an argument to move these to a separate utils file of some kind. Purely for keeping the file a bit shorter and more focused on the api aspects. Additionally, are we feeling confident about how fragile/reliable these generate display name functions are?

nmgarza5:
@greptile

backend/onyx/llm/utils.py
```diff
@@ -462,11 +462,11 @@ def get_llm_max_tokens(
             max_tokens = model_obj["max_tokens"]
             return max_tokens
 
-        logger.error(f"No max tokens found for LLM: {model_name}")
         raise RuntimeError("No max tokens found for LLM")
```
wenxi-onyx: I had this fix in my branch, but it's not going in anytime soon. Can you change this to not raise an error as this is not actually an error?

```
logger.warning("No max tokens found")
return fallback constant
```

Since the Except block was written for this raise, may make sense to modify about how a real exception occured.

wenxi-onyx: I know this is slightly out of scope, but can we rename this function to `llm_max_input_tokens`? Hopefully just the vscode rename will get it one shot...

backend/onyx/llm/utils.py
```diff
@@ -493,13 +493,12 @@ def get_llm_max_output_tokens(
             max_output_tokens = int(model_obj["max_tokens"] * 0.1)
             return max_output_tokens
 
-        logger.error(f"No max output tokens found for LLM: {model_name}")
         raise RuntimeError("No max output tokens found for LLM")
```
wenxi-onyx: Same thing here about not raising as above comment.

wenxi-onyx (approved):

backend/onyx/llm/utils.py
nmgarza5: ah yeah my bad, i though i already did this...i saw the same thing and also thought it should be a warning. will make the change

PR #6628, opened 2025-12-05
chore(docs): prefer uv over pip

jmelahman:
## How Has This Been Tested?

Captured by existing

## Additional Options

- [x] Override Linear Check

---
## Summary by cubic
Standardize Python setup on uv and simplify backend tooling commands in CONTRIBUTING. Also slim the mock connector Docker image by avoiding pip cache.

- **Refactors**
  - Replace pip-based setup with uv (uv venv, uv sync --all-extras).
  - Use uv run for Playwright, pre-commit, and mypy.
  - Remove redundant pip requirements and editable install steps.

- **Dependencies**
  - Add --no-cache-dir to pip install in mock connector server Dockerfile to reduce image size.

<sup>Written for commit a0bb84edda637da997ff28ea94eb1a66bc563bbf. Summary will update automatically on new commits.</sup>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,7 +71,7 @@ If using a higher version, sometimes some libraries will not be available (i.e. 
 
 #### Backend: Python requirements
 
-Currently, we use [uv](https://docs.astral.sh/uv/) and recommend creating a virtual environment.
+Currently, we use [uv](https://docs.astral.sh/uv/) and recommend creating a [virtual environment](https://docs.astral.sh/uv/pip/environments/#using-a-virtual-environment).
 
 For convenience here's a command for it:
 


CONTRIBUTING.md
```diff
@@ -71,12 +71,12 @@ If using a higher version, sometimes some libraries will not be available (i.e.
...
 #### Backend: Python requirements
 
-Currently, we use pip and recommend creating a virtual environment.
+Currently, we use [uv](https://docs.astral.sh/uv/) and recommend creating a virtual environment.
 
 For convenience here's a command for it:
 
 ```bash
-python -m venv .venv
+uv venv .venv --python 3.11
```
wenxi-onyx, line 79: Any idea why I didn't have to do this? I'm not sure exactly what happened, but I think I might've run `uv sync` or even just `uv` and the `.venv` was auto-created for me

jmelahman: `uv sync` (and `uv run`) creates this by default, but it will use the system default python interpreter (though each has an equivalent `--python` option).

PR #6642, opened 2025-12-05
fix(sharepoint): set tenant domain for client secret auth

nmgarza5:
## Description

The `sp_tenant_domain` was only being set when using certificate authentication, causing permissions sync to fail with "MSAL app or tenant domain is not set" error for connectors using client secret auth when "Auto Sync Permissions" is enabled.

This PR moves the tenant domain fetching logic outside the certificate-only block so it runs for all authentication methods.

  Why it's needed for all auth methods:
  - Graph API uses https://graph.microsoft.com/.default (works without tenant domain)
  - SharePoint REST API uses https://{tenant}.sharepoint.com/.default (needs tenant domain)
  - Permissions sync uses the SharePoint REST API via ClientContext, so it needs the tenant domain regardless of how you authenticated

## How Has This Been Tested?

- Added unit tests for `load_credentials` that verify `sp_tenant_domain` is set for both CLIENT_SECRET and CERTIFICATE auth methods
- All 3 new tests pass

## Additional Options

---
## Summary by cubic
Fixes SharePoint permission sync for client secret auth by always setting the tenant domain during credential loading. Applies tenant domain lookup for all auth methods to unblock Auto Sync Permissions.

- **Bug Fixes**
  - SharePoint: fetch and set sp_tenant_domain for client secret and certificate flows; added unit tests covering both paths and domain extraction.

<sup>Written for commit c53b731e312261085b1124cfa58850dd62163b62. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/connectors/sharepoint/connector.py
+++ b/backend/onyx/connectors/sharepoint/connector.py
@@ -1258,22 +1258,21 @@ class SharepointConnector(
             return token
 
         self._graph_client = GraphClient(_acquire_token_for_graph)
-        if auth_method == SharepointAuthMethod.CERTIFICATE.value:
-            org = self.graph_client.organization.get().execute_query()
-            if not org or len(org) == 0:
-                raise ConnectorValidationError("No organization found")
 
-            tenant_info: Organization = org[
-                0
-            ]  # Access first item directly from collection
-            if not tenant_info.verified_domains:
-                raise ConnectorValidationError("No verified domains found for tenant")
+        org = self.graph_client.organization.get().execute_query()
+        if not org or len(org) == 0:
+            raise ConnectorValidationError("No organization found")
 
-            sp_tenant_domain = tenant_info.verified_domains[0].name
-            if not sp_tenant_domain:
-                raise ConnectorValidationError("No verified domains found for tenant")
-            # remove the .onmicrosoft.com part
-            self.sp_tenant_domain = sp_tenant_domain.split(".")[0]
+        tenant_info: Organization = org[0]
+        if not tenant_info.verified_domains:
+            raise ConnectorValidationError("No verified domains found for tenant")
+
+        sp_tenant_domain = tenant_info.verified_domains[0].name
+        if not sp_tenant_domain:
+            raise ConnectorValidationError("No verified domains found for tenant")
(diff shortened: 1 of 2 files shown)


backend/tests/unit/onyx/connectors/sharepoint/test_load_credentials.py
```diff
@@ -0,0 +1,153 @@
...
+    mock_graph_client_class: MagicMock,
+    client_secret_credentials: dict[str, Any],
+) -> None:
+    """Test that load_credentials sets sp_tenant_domain when using client secret auth."""
+    # Setup mocks
+    mock_msal_class.return_value = _create_mock_msal_app()
+
+    mock_org = MockOrganization([MockVerifiedDomain("contoso.onmicrosoft.com")])
+    mock_graph_client = MockGraphClient([mock_org])
+    mock_graph_client_class.return_value = mock_graph_client
```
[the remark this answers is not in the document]
nmgarza5: fake news

backend/onyx/connectors/sharepoint/connector.py
```diff
@@ -1258,22 +1258,21 @@ def _acquire_token_for_graph() -> dict[str, Any]:
...
+        tenant_info: Organization = org[0]
+        if not tenant_info.verified_domains:
+            raise ConnectorValidationError("No verified domains found for tenant")
+
+        sp_tenant_domain = tenant_info.verified_domains[0].name
+        if not sp_tenant_domain:
+            raise ConnectorValidationError("No verified domains found for tenant")
+        # remove the .onmicrosoft.com part
+        self.sp_tenant_domain = sp_tenant_domain.split(".")[0]
+
```
Subash-Mohan, line 1275: Shall we revert it? As far as I remember, accessing organization data may require extra permissions for users using client auth credentials. The connector will break for customers using the client secret authentication method.

Subash-Mohan: > Note: we don't populate the sp_tenant_domain for non-certificate authentication methods because it's unnecessary outside of permission sync. Permission sync is only supported with certificate authentication.

Subash-Mohan (changes requested):

PR #6643, opened 2025-12-05
feat: standardize placeholders in default system prompt

rohoswagger:
## Description

Standardizing the placeholders we had in our default system prompt. 

Addresses https://linear.app/danswer/issue/DAN-3142/need-to-remove-this-from-the-ui

## How Has This Been Tested?

Locally
Screenshot 2025-12-05 at 13 36 04

## Additional Options

### DIFF
--- a/backend/alembic/versions/87c52ec39f84_update_default_system_prompt.py
+++ b/backend/alembic/versions/87c52ec39f84_update_default_system_prompt.py
@@ -1,9 +1,8 @@
 """update_default_system_prompt
 
-
-Revision ID: d21c3867fa79
-Revises: e8f0d2a38171
-Create Date: 2025-12-05 11:16:26.812404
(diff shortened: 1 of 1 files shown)


yuhongsun96 (approved):
Would be ever so slightly nicer if we can add a preview option to it so the user can see how it would look after the change. But this is not required

backend/onyx/chat/prompt_utils.py
```diff
@@ -166,7 +160,13 @@ def build_system_prompt(
                 memory.strip() for memory in memories if memory.strip()
             )
 
-    if should_cite_documents or include_all_guidance:
+    # If citation guidance placeholder was not present and we need citations, append after company context
```
yuhongsun96: Nit: Ever so slightly cleaner if it's handled also within the function but not important

PR #6658, opened 2025-12-06
feat(new vector db interface): Implement index for Vespa

acaprau:
## Description

This PR implements VespaIndex's index method to the new vector db interface we want to have, introduced in #5700. This PR comes with some small changes to the proposed interface as they became apparent through implementation, and some other small hygiene.

## How Has This Been Tested?

#6659 tests it.

## Additional Options

---
## Summary by cubic
Implements VespaDocumentIndex indexing with safe ID handling, pre-delete of stale chunks, and batched inserts via a thread pool. Refines the DocumentIndex interface for clearer indexing inputs/outputs.

- New Features
  - Implemented VespaDocumentIndex index() with parallel deletes and inserts.
  - Added TenantState to manage tenant ID and multitenant mode.

- Refactors
  - IndexingMetadata now uses a ChunkCounts model; Indexable.index now takes/returns lists.
  - Improved docstrings and naming in Vespa utilities, HTTP client setup, and batching helper; constants unchanged functionally.

<sup>Written for commit b28b64df39722ac196e518cdd76da166f41efd4f. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/document_index/interfaces_new.py
+++ b/backend/onyx/document_index/interfaces_new.py
@@ -6,16 +6,6 @@ from onyx.access.models import DocumentAccess
 from onyx.context.search.enums import QueryType
 from onyx.context.search.models import IndexFilters
 from onyx.context.search.models import InferenceChunk
-from dataclasses import dataclass
-from datetime import datetime
-from typing import Any
-
-from pydantic import BaseModel
-
-from onyx.access.models import DocumentAccess
-from onyx.context.search.enums import QueryType
-from onyx.context.search.models import IndexFilters
-from onyx.context.search.models import InferenceChunk
 from onyx.db.enums import EmbeddingPrecision
 from onyx.indexing.models import DocMetadataAwareIndexChunk
 from shared_configs.model_server_models import Embedding
@@ -25,12 +15,6 @@ from shared_configs.model_server_models import Embedding
 # document chunks. By the terminology of most search engines / vector databases,
 # the individual objects stored are called documents, but in this case it refers
 # to a chunk.
-
-# Outside of searching and update capabilities, the document index must also
-# implement the ability to port all of the documents over to a secondary index.
-# This allows for embedding models to be updated and for porting documents to
-# happen in the background while the primary index still serves the main
-# traffic.
 
 
 __all__ = [
--- a/backend/onyx/document_index/vespa/vespa_document_index.py
+++ b/backend/onyx/document_index/vespa/vespa_document_index.py
@@ -159,14 +159,16 @@ class VespaDocumentIndex(DocumentIndex):
             len(doc_id_to_chunk_cnt_diff)
             == len(doc_id_to_previous_chunk_cnt)
             == len(doc_id_to_new_chunk_cnt)
-        )
+        ), "Bug: Doc ID to chunk maps have different lengths."
 
         # Vespa has restrictions on valid characters, yet document IDs come from
         # external w.r.t. this class. We need to sanitize them.
         cleaned_chunks: list[DocMetadataAwareIndexChunk] = [
             clean_chunk_id_copy(chunk) for chunk in chunks
         ]
-        assert len(cleaned_chunks) == len(chunks)
+        assert len(cleaned_chunks) == len(
+            chunks
+        ), "Bug: Cleaned chunks and input chunks have different lengths."
 
         # Needed so the final DocumentInsertionRecord returned can have the
         # original document ID. cleaned_chunks might not contain IDs exactly as


backend/onyx/document_index/interfaces_new.py
```diff
@@ -12,13 +10,17 @@
...
+# NOTE: "Document" in the naming convention is used to refer to the entire
+# document as represented in Onyx. What is actually stored in the index is the
+# document chunks. By the terminology of most search engines / vector databases,
+# the individual objects stored are called documents, but in this case it refers
+# to a chunk.
 
-# Outside of searching and update capabilities, the document index must also implement the ability to port all of
-# the documents over to a secondary index. This allows for embedding models to be updated and for porting documents
-# to happen in the background while the primary index still serves the main traffic.
+# Outside of searching and update capabilities, the document index must also
```
yuhongsun96: We've actually decided this would happen outside of the index and would be handled by the logic outside, can probably remove this comment

yuhongsun96 (approved):
Seems good

backend/onyx/document_index/vespa/vespa_document_index.py
```diff
@@ -0,0 +1,285 @@
...
+            # exiting a context manager.
+            self._httpx_client_context = TemporaryHTTPXClientContext(
+                get_vespa_http_client
+            )
+        self._multitenant = tenant_state.multitenant
+
+    def verify_and_create_index_if_necessary(
+        self, embedding_dim: int, embedding_precision: EmbeddingPrecision
+    ) -> None:
+        raise NotImplementedError
```
evan-onyx, line 142: is this addressed in a followup PR? Is it intended that we/other people further subclass this?

backend/onyx/document_index/vespa/vespa_document_index.py
```diff
@@ -0,0 +1,285 @@
...
+        doc_id_to_chunk_cnt_diff = indexing_metadata.doc_id_to_chunk_cnt_diff
+        doc_id_to_previous_chunk_cnt = {
+            doc_id: chunk_cnt_diff.old_chunk_cnt
+            for doc_id, chunk_cnt_diff in doc_id_to_chunk_cnt_diff.items()
+        }
+        doc_id_to_new_chunk_cnt = {
+            doc_id: chunk_cnt_diff.new_chunk_cnt
+            for doc_id, chunk_cnt_diff in doc_id_to_chunk_cnt_diff.items()
+        }
+        assert (
```
evan-onyx, line 158: nit: we decided to reduce the usage of assert in new code, so we prefer raising an error here

backend/onyx/document_index/interfaces_new.py
```diff
@@ -67,24 +70,37 @@ class DocumentSectionRequest(BaseModel):
...
+
+    A common pattern to ensure that no chunks are left over is to delete all of
+    the chunks for a document and then re-index the document. This information
+    allows us to only delete the extra "tail" chunks when the document has
+    gotten shorter.
     """
 
-    # The tuple is (old_chunk_cnt, new_chunk_cnt)
-    doc_id_to_chunk_cnt_diff: dict[str, tuple[int, int]]
+    class ChunkCounts(BaseModel):
```
evan-onyx, line 76: seems odd to define a BaseModel inside a BaseModel?

evan-onyx (approved):
some tiny asks

backend/onyx/document_index/vespa/vespa_document_index.py
acaprau: yes, this sequence of prs will slowly implement this interface. in any case the existing hotpath uses its own implementation for this functionality and it has to explicitly be moved over to this interface, so no harm in leaving something unimplemented for now

backend/onyx/document_index/vespa/vespa_document_index.py
acaprau: will fix this in this pr and all other ones

acaprau: wait hang on, why? there are valid use-cases for assert, namely internal invariant checks which i would argue this is one of those @evan-onyx

acaprau: im gonna merge this pr to make forward progress but ill come back to this based on the outcome of this slack thread: https://onyx-company.slack.com/archives/C0771QKDBPE/p1766016410469499

backend/onyx/document_index/interfaces_new.py
acaprau: maybe, this used to be just a tuple where callers had to know which index theyre interested in, i think a struct-like object with named fields is better. and it's not used anywhere outside the context of IndexingMetadata so might as well be in here no?

PR #6666, opened 2025-12-08
feat(mcp): auth and deletion changes

Subash-Mohan:
## Description
This pull request introduces several backend and frontend changes to support tracking when an MCP server's tools were last refreshed, improves authentication handling, and refactors types for MCP servers in the frontend for consistency. The most important changes are grouped below.

## How Has This Been Tested?
From UI

## Additional Options

---
## Summary by cubic
Tracks when MCP server tools were last refreshed, adds safer auth change cleanup, and simplifies auth modals. Also refactors frontend types to a single MCPServer model and splits Actions into separate MCP and OpenAPI pages.

- **New Features**
  - Added last_refreshed_at to MCP servers; set on tool refresh and shown in UI with a refresh button.
  - Introduced upsertMCPServer API/client for saving auth configs.
  - Auth change cleanup: OAUTH deletes all user connection configs; API_TOKEN deletes admin config and admin user configs.

- **Refactors**
  - Replaced MCPServerWithStatus with a unified MCPServer type across the web app.
  - Centralized ApiResponse and OAuthConfig types; adjusted services to return MCPServer.
  - Simplified MCP and OpenAPI authentication modals.
  - Split Actions into dedicated MCP and OpenAPI pages with a new layout and sidebar links.

<sup>Written for commit 17cc8e8d10420c0c1c6470ef87c09b94d4f0c070. Summary will update automatically on new commits.</sup>


backend/onyx/db/mcp.py
```diff
@@ -330,3 +335,15 @@ def delete_user_connection_configs_for_server(
         db_session.delete(config)
 
     db_session.commit()
+
+
+def delete_all_user_connection_configs_for_server(
```
evan-onyx: nit: add __no_commit to db functions that don't commit

backend/onyx/db/mcp.py
```diff
@@ -330,3 +335,15 @@ def delete_user_connection_configs_for_server(
...
+    server_id: int, db_session: Session
+) -> None:
+    """Delete all user connection configs for a specific MCP server (excluding admin config)"""
```
evan-onyx: cubic's comment is correct, either the doc string needs to be changed or the function does depending on the desired behavior

web/src/sections/actions/modals/MCPAuthenticationModal.tsx
```diff
@@ -670,24 +629,17 @@ export default function MCPAuthenticationModal({
...
+                  Cancel
                 </Button>
                 <Button
                   main
                   primary
                   type="submit"
-                  disabled={!isValid || isSubmitting}
+                  disabled={
+                    !isValid || isSubmitting || (isEditConfigsFlow && !dirty)
+                  }
```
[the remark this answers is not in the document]
evan-onyx: I agree that this is something we should support, we should get Duo's take too

web/src/sections/actions/MCPActionCard.tsx
```diff
@@ -8,6 +8,7 @@ import React, {
   useCallback,
 } from "react";
 import SvgServer from "@/icons/server";
+import SvgRefreshCw from "@/icons/refresh-cw";
```
raunakab: This one will need to be updated with the new `lib/opal` updates!

web/src/sections/actions/MCPPageContent.tsx
```diff
@@ -47,12 +48,10 @@ export default function MCPPageContent() {
...
   // Local state
-  const [selectedServer, setSelectedServer] =
-    useState<MCPServerWithStatus | null>(null);
+  const [selectedServer, setSelectedServer] = useState<MCPServer | null>(null);
   const [serverToDisconnect, setServerToDisconnect] =
-    useState<MCPServerWithStatus | null>(null);
-  const [serverToManage, setServerToManage] =
-    useState<MCPServerWithStatus | null>(null);
+    useState<MCPServer | null>(null);
+  const [serverToManage, setServerToManage] = useState<MCPServer | null>(null);
```
raunakab: Can these states be non-null all at the same time? I.e., is the following possible?

```tsx
selectedServer !== null && serverToDisconnect !== null && serverToManage !== null
```

raunakab: If they're all mutually separate from one another, then what exists is fine. However, if it's invalid for two or more of them to be true, then we should maybe combine them into an enum.

web/src/sections/actions/useServerTools.ts
raunakab, line 1: If this is a generic hook, let's move it to `web/src/hooks`.

web/src/sections/actions/MCPActionCard.tsx
```diff
@@ -207,6 +213,43 @@ export default function MCPActionCard({
...
+        )}
+      </div>
+    );
+  }, [
+    server.last_refreshed_at,
+    serverId,
+    mutate,
+    onRefreshTools,
+    isToolsRefreshing,
+  ]);
```
raunakab, line 275: If the `leftAction` is only used in one place, can we just inline it? I.e.,

```tsx
<ToolsList
  leftAction={(
    ...
  )}
  ...
/>
```

web/src/sections/actions/ToolsList.tsx
```diff
@@ -87,27 +91,37 @@ const ToolsList: React.FC<ToolsListProps> = ({
...
             )}
+
+            {/* Right action area */}
+            <div className="flex items-center gap-1 ml-auto">
+              <Text mainUiBody className="text-action-link-05">
+                {enabledCount}
+              </Text>
+              <Text text03 mainUiBody>
+                of {totalCount} tool{totalCount !== 1 ? "s" : ""} enabled
+              </Text>
```
raunakab: Text which contains formatting escapes inside of it will not be treated as strings, but rather a `JSX` component. You should update the children to be:

```tsx
<Text text03 mainUiBody>
  {`of ${totalCount} tool${totalCount !== 1 ? "s" : ""} enabled`}
```

raunakab: I think this component is going to be used in many more places (the "X of Y enabled` paradigm). Maybe you can create a new component that handles this?

web/src/sections/actions/ToolsList.tsx
```diff
@@ -87,27 +91,37 @@ const ToolsList: React.FC<ToolsListProps> = ({
...
-                  showOnlyEnabled ? "Show all tools" : "Show only enabled"
-                }
-                aria-label={
-                  showOnlyEnabled ? "Show all tools" : "Show only enabled tools"
-                }
-              />
+          <div className="flex items-center justify-between gap-2 w-full">
+            {/* Left action area */}
+            {leftAction && (
+              <div className="flex items-center">{leftAction}</div>
```
raunakab: Do you need a wrapper `div` for this? Maybe just

```tsx
{leftAction}
```

will suffice?

raunakab (commented):
Looks good!

backend/onyx/db/mcp.py
Subash-Mohan: doc string is wrong. I changed it now.

web/src/sections/actions/MCPActionCard.tsx
Subash-Mohan: I added it at the top just to keep the code clean. Since it is memoized, I don’t think it will cause any issues. What do you think?

web/src/sections/actions/MCPPageContent.tsx
Subash-Mohan: We don’t need three state variables there. We can merge them into a single variable. Now we only have one state variable.

web/src/sections/actions/ToolsList.tsx
Subash-Mohan: No we don't need it.

web/src/sections/actions/useServerTools.ts
Subash-Mohan: No, this is specific to MCP servers.

raunakab:
Oh, there are some Playwright tests that are failing here. Could we get them passing before merge?

PR #6680, opened 2025-12-09
fix: use consistent INSTALL_ROOT instead of pwd for deployment paths

vinitngr:
## Summary
This patch corrects path handling in the installation script.  
Previously, several operations relied on the current working directory (`pwd`) rather than the intended installation root. This caused deployment files to be created wherever the user happened to run the script.  

The update standardizes all filesystem operations on `INSTALL_ROOT`, ensuring deterministic behavior across environments.

## Changes
- Replaced all `pwd`-dependent path usage with consistent `INSTALL_ROOT` references.
- Added quoting to all path operations (`cd`, `cp`, `sed`, etc.) to support home directories that contain spaces.
- Replaced unquoted subshell path calls with safely quoted forms:  
  `cd "${INSTALL_ROOT}/deployment" …`
- Removed implicit working-directory side effects.
- Ensured Docker Compose calls always execute inside the correct deployment directory.

## Why This Matters
1. **Predictable install location**  
   Using `pwd` meant installation output varied depending on where the script was executed. This caused confusion and inconsistent filesystem layouts.

2. **Cross-platform stability**  
   Users with spaces in their home directory (common on Windows Git Bash, macOS user accounts, and even some Linux desktop setups) experienced unexpected failures due to unquoted paths.  
   These failures manifested as:
   - `cd: too many arguments`
   - missing directories  
   - incorrectly resolved installation paths  

   Quoting fixes these issues consistently across all platforms.

3. **CI and automation compatibility**  
   Deterministic paths allow GitHub Actions and automated setups to work reliably without depending on the runner’s working directory.

## Testing
- Verified installation directory resolution under:
  - Windows Git Bash
  - WSL2
  - Linux shells (bash/zsh)
  - macOS shells (bash/zsh)
- Confirmed that `.onyx_data` is consistently created under `$HOME` or `$INSTALL_PREFIX`.
- Checked that Docker Compose commands resolve correctly when run inside the fixed deployment directory.

## Notes
This patch does not change deployment logic or service behavior.  
It only stabilizes and corrects path resolution for reliable cross-platform installs.

---
## Summary by cubic
Standardizes install.sh to use INSTALL_ROOT for all deployment paths instead of pwd. This makes installs predictable and fixes failures on systems with spaces in paths.

- **Bug Fixes**
  - Replaced all pwd/onyx_data references with ${INSTALL_ROOT} (defaults to $PWD/onyx_data; respects $INSTALL_PREFIX when set).
  - Quoted all path operations to handle spaces reliably.
  - Always run Docker Compose from ${INSTALL_ROOT}/deployment; fixed container name parsing.
  - Ensured shutdown/delete flows target ${INSTALL_ROOT} and handle missing directories gracefully.

<sup>Written for commit c204c9d586fbb500e88402f1a590dc3a7e1570c9. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/chat/llm_loop.py
+++ b/backend/onyx/chat/llm_loop.py
@@ -444,7 +444,7 @@ def run_llm_loop(
                 tool_definitions=[tool.tool_definition() for tool in final_tools],
                 tool_choice=tool_choice,
                 llm=llm,
-                turn_index=llm_cycle_count + reasoning_cycles,
+                placement=Placement(turn_index=llm_cycle_count + reasoning_cycles),
                 citation_processor=citation_processor,
                 state_container=state_container,
                 # The rich docs representation is passed in so that when yielding the answer, it can also
--- a/backend/onyx/chat/llm_step.py
+++ b/backend/onyx/chat/llm_step.py
@@ -394,7 +394,7 @@ def run_llm_step_pkt_generator(
     tool_definitions: list[dict],
     tool_choice: ToolChoiceOptions,
     llm: LLM,
-    turn_index: int,
+    placement: Placement,
     state_container: ChatStateContainer,
     citation_processor: DynamicCitationProcessor | None,
     reasoning_effort: ReasoningEffort | None = None,
@@ -403,7 +403,6 @@ def run_llm_step_pkt_generator(
     custom_token_processor: (
         Callable[[Delta | None, Any], tuple[Delta | None, Any]] | None
     ) = None,
-    sub_turn_index: int | None = None,
 ) -> Generator[Packet, None, tuple[LlmStepResult, bool]]:
     """Run an LLM step and stream the response as packets.
     NOTE: DO NOT TOUCH THIS FUNCTION BEFORE ASKING YUHONG, this is very finicky and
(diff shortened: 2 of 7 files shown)


jmelahman:
Appreciate the change. Would it make sense to break this up into two changes? One which fixes the path handling and another which changes the default installation directory? I'd be happy to stamp the fix, but the behavioral change might require more thought (I'm not sure `$HOME/.onyx_data` is the correct location for this) + likely requires documentation updates. It's also easily overridden by your `INSTALL_PREFIX`.

vinitngr:
Thanks for the feedback - agreed.

I’ve updated this PR to only include the path-handling fixes (consistent INSTALL_ROOT usage + quoting) and reverted the default INSTALL_ROOT resolution so the install location behavior matches the previous onyx_data behavior.

We can discuss alternative default locations separately if needed.

deployment/docker_compose/install.sh
```diff
@@ -44,6 +44,8 @@ while [[ $# -gt 0 ]]; do
     esac
 done
 
+INSTALL_ROOT="${INSTALL_PREFIX:-$PWD/onyx_data}"
```
jmelahman:
```suggestion
INSTALL_ROOT="${INSTALL_PREFIX:-onyx_data}"
```
This is closer to the previous behavior. I'm not sure how every shell handles `$PWD`

vinitngr:
Good point, updated to avoid relying on $PWD and keep behavior consistent.

jmelahman:
FYI, we had a test breakage at HEAD. These are real failures but unrelated, so you'll need to merge in HEAD. Sorry

PR #6692, opened 2025-12-12
fix: mcp server name and desc updates

evan-onyx:
## Description

Addresses https://linear.app/danswer/issue/DAN-3143/mcp-name-not-updating

Previously the update endpoint was only intended to handle the user changing tools, but it's unintuitive for it to not handle name and description changes too (Especially since we now auto-list tools). The user can still change the server url without any repercussions until they successfully list tools (at which point it gets updated), but since this UI is going away soon anyways I think it's reasonable

## How Has This Been Tested?

tested in UI

## Additional Options

---
## Summary by cubic
Allow editing MCP server name and description through the tools update endpoint so users can update details and tools in one step (fixes Linear issue DAN-3143). The response now returns the updated server name.

- **Bug Fixes**
  - API: Accept optional name/description and return server_name; apply updates before syncing tools.
  - Web: ToolList sends name and description to attachMCPTools; types updated to include server_name.
  - Connectors: Use gzip/deflate only in Accept-Encoding to avoid flaky Brotli decoding in connectivity checks.
  - Users: Skip DNS deliverability checks in invite flows to allow test domains.
  - Tests: Relax flaky Google Drive service account doc count assertion to 51–52.

- **Dependencies**
  - ODS tools: Downgraded Go version to 1.24 to fix CI.

<sup>Written for commit 702a93c1469de4dc9dda4f7e82163a43b2dcc409. Summary will update automatically on new commits.</sup>

### DIFF
--- a/tools/ods/go.mod
+++ b/tools/ods/go.mod
@@ -1,6 +1,6 @@
 module github.com/onyx-dot-app/onyx/tools/ods
 
-go 1.25.0
+go 1.24.0
 
 require (
 	github.com/sirupsen/logrus v1.9.3


tools/ods/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/onyx-dot-app/onyx/tools/ods
 
-go 1.25.0
+go 1.24.0
```
jmelahman, line 3: Please undo

PR #6722, opened 2025-12-15
fix: remove unnecessary chat display tabindex

jmelahman:
## Description

This element is inheriting a `tabIndex` of `0` causing it to be focusable which in turn causes an `outline`,

20251214_17h36m33s_grim

## Additional Options

- [x] Override Linear Check

---
## Summary by cubic
Prevent unwanted focus ring in the Chat page by making the chat display container non-focusable and disabling its outline. Set tabIndex to -1 and apply outline: none so it no longer inherits focus or shows an outline during navigation.

<sup>Written for commit c27acb040c6224a3dd9c8d102c9c44da1f814850. Summary will update automatically on new commits.</sup>


jmelahman:
@greptile

PR #6731, opened 2025-12-15
fix(llm): restore default models and  filter obsolete/duplicate models from API

nmgarza5:
## Description

1. Restores the list of default models for openai, anthropic and vertexai. 
2. Filters out models known to be obsolete and duplicate models (same version with multiple release dates -> prefer the non dated version).
3. moves some utils functions from `backend/onyx/server/manage/llm/models.py` to  `backend/onyx/server/manage/llm/utils.py`

## How Has This Been Tested?

local testing + existing

## Additional Options

---
## Summary by cubic
Restored sensible default-visible models and cleaned up provider model lists by filtering obsolete and dated-duplicate entries. This makes the model picker clearer and sets better defaults in the UI.

- **New Features**
  - Added curated default-visible lists for OpenAI, Anthropic, and Vertex (e.g., gpt-5, claude-sonnet-4-5, gemini-2.5-flash) marked is_visible=true.

- **Bug Fixes**
  - Filtered obsolete models: GPT-3 and legacy text-*; Claude 2/instant; Gemini 1.0; Palm/Bison.
  - Hid dated duplicates when a base model exists (e.g., gpt-4-20241122 vs gpt-4).
  - Applied filtering in both non-admin and admin views (LLMProviderDescriptor.from_model and LLMProviderView.from_model) via a shared utility.

<sup>Written for commit c4c16f16fac31979f8b554d52e6a2bbd4dffda44. Summary will update automatically on new commits.</sup>


nmgarza5:
@greptile

backend/onyx/server/manage/llm/models.py
```diff
@@ -11,6 +12,75 @@
 from onyx.server.manage.llm.utils import is_reasoning_model
 
 
+def is_obsolete_model(model_name: str, provider: str) -> bool:
```
Weves: this feels like it better belongs in llm_provider_options.py

Weves: also prefer to keep the models.py files basically just the models w/o any utils

backend/onyx/llm/llm_provider_options.py
```diff
@@ -490,7 +527,7 @@ def fetch_model_configurations_for_provider(
         configs.append(
             ModelConfigurationView(
                 name=model_name,
-                is_visible=False,
+                is_visible=model_name in visible_models,
```
Weves, line 592: will this cause a problem for model providers that aren't openai/anthropic/vertex?

backend/onyx/server/manage/llm/models.py
nmgarza5: i had it there at first but was running into a circular import issue. I think i can handle that with lazy loading if you still prefer it there

backend/onyx/llm/llm_provider_options.py
nmgarza5: no this is the behavior we had before. we didn't surface any default models names for providers outside of the big three -> openai/anthropic/vertex

backend/onyx/server/manage/llm/models.py
```diff
@@ -64,8 +67,28 @@ def from_model(
...
+        from onyx.llm.llm_provider_options import is_obsolete_model
 
         provider = llm_provider_model.provider
+
+        # Get all model names for deduplication check
+        all_model_names = {mc.name for mc in llm_provider_model.model_configurations}
+
+        # Filter out obsolete and duplicate models
+        filtered_configs = []
+        for model_configuration in llm_provider_model.model_configurations:
```
Weves: nit: this seems to be repeated twice, should we add another util for this?

Weves (approved):

backend/onyx/server/manage/llm/models.py
nmgarza5: yeah makes sense

backend/onyx/server/manage/llm/utils.py
```diff
@@ -235,3 +241,104 @@ def is_reasoning_model(model_id: str, display_name: str) -> bool:
...
+
+
+def should_filter_as_dated_duplicate(
+    model_name: str, all_model_names: set[str]
+) -> bool:
+    """Check if this model is a dated variant and a non-dated version exists."""
+    base = extract_base_model_name(model_name)
+    if base and base in all_model_names:
+        return True
+    return False
```
nmgarza5, line 269: @wenxi-onyx

PR #6777, opened 2025-12-16
fix(feedback): More fixes

justin-tahara:
## Description

Fixing the Feedback API backend calls to ensure we are not failing API calls.

## How Has This Been Tested?

## Additional Options

---
## Summary by cubic
Fixes feedback creation and removal when the frontend sends a TOOL_CALL/TOOL_CALL_RESPONSE message by attaching feedback to the related assistant message. Also passes messageId from the shared chat UI to prevent failed API calls.

- **Bug Fixes**
  - Backend: Traverse parent messages to find the assistant message and use its ID for feedback add/remove; error only if none exists.
  - Frontend: Send messageId in SharedChatDisplay so feedback actions target the correct message.

<sup>Written for commit 0953230ae0ca8ecb4855a59a8c105a66d758e1f7. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/db/feedback.py
+++ b/backend/onyx/db/feedback.py
@@ -243,17 +243,26 @@ def _find_assistant_message_id(
     if chat_message.message_type == MessageType.ASSISTANT:
         return chat_message.id
 
+    # Fetch all messages in the session at once to avoid N+1 queries
+    stmt = (
+        select(ChatMessage)
+        .where(ChatMessage.chat_session_id == chat_message.chat_session_id)
+        .order_by(ChatMessage.id)
+    )
+    all_messages = db_session.execute(stmt).scalars().all()
+
+    # Build a lookup map for efficient parent traversal
+    message_map = {msg.id: msg for msg in all_messages}
+
     # Traverse up the message tree to find the parent ASSISTANT message
     current_msg = chat_message
     max_depth = 100  # Safeguard against infinite loops or extremely deep chains
     depth = 0
 
     while current_msg.parent_message_id is not None and depth < max_depth:
-        parent_msg = get_chat_message(
-            chat_message_id=current_msg.parent_message_id,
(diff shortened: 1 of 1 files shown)


backend/onyx/db/feedback.py
```diff
@@ -227,6 +228,40 @@ def delete_document_feedback_for_documents__no_commit(
...
+    """
+    if chat_message.message_type == MessageType.ASSISTANT:
+        return chat_message.id
+
+    # Traverse up the message tree to find the parent ASSISTANT message
+    current_msg = chat_message
+    max_depth = 100  # Safeguard against infinite loops or extremely deep chains
+    depth = 0
+
+    while current_msg.parent_message_id is not None and depth < max_depth:
```
Weves, line 262: could we instead fetch the last 100 messages for the chat session at once rather than up to 100 distinct DB calls?

backend/onyx/db/feedback.py
```diff
@@ -249,11 +284,15 @@ def create_chat_message_feedback(
         chat_message_id=chat_message_id, user_id=user_id, db_session=db_session
     )
 
-    if chat_message.message_type != MessageType.ASSISTANT:
-        raise ValueError("Can only provide feedback on LLM Outputs")
+    # Find the ASSISTANT message (handles TOOL_CALL/TOOL_CALL_RESPONSE by traversing up)
+    target_message_id = _find_assistant_message_id(
```
Weves, line 301: why do we need this? Can we add a comment explaining the scenario?

justin-tahara: This is needed because the code creates separate `ChatMessage` records with different types like `TOOL_CALL` and `TOOL_CALL_RESPONSE` which are children of the main `ASSISTANT` message. 

This can happen accidentally if the FE returns the tool call message ID instead of the assistant message id.

PR #6798, opened 2025-12-17
chore(tf): Update VPC calling

justin-tahara:
## Description

Update the terraform code to pass in the VPC endpoint id instead of grabbing the VPC data 

## How Has This Been Tested?

Already using in our internal clusters

## Additional Options

---
## Summary by cubic
Switched Terraform AWS modules to use a provided S3 VPC endpoint ID. The VPC module now creates the S3 gateway endpoint and outputs its ID, and the S3 bucket policy restricts access to that endpoint.

- **Refactors**
  - VPC: creates the S3 gateway VPC endpoint and outputs s3_vpc_endpoint_id.
  - Onyx: passes s3_vpc_endpoint_id from VPC output or uses the user-provided value when reusing an existing VPC.
  - S3: removes VPC/region inputs and endpoint creation; bucket policy uses s3_vpc_endpoint_id.
  - README updated to document the new input and outputs.

- **Migration**
  - When create_vpc is false, provide s3_vpc_endpoint_id for the existing VPC endpoint (validated).
  - Update any usage of the S3 module to remove vpc_id/region and supply s3_vpc_endpoint_id.

<sup>Written for commit 72f745281ed342151d2877d549e234b300a8ebcd. Summary will update automatically on new commits.</sup>

### DIFF
--- a/deployment/terraform/modules/aws/vpc/outputs.tf
+++ b/deployment/terraform/modules/aws/vpc/outputs.tf
@@ -13,3 +13,8 @@ output "public_subnets" {
 output "vpc_cidr_block" {
   value = module.vpc.vpc_cidr_block
 }
+
+output "s3_vpc_endpoint_id" {
+  description = "ID of the S3 gateway VPC endpoint created for this VPC"
+  value       = aws_vpc_endpoint.s3.id
+}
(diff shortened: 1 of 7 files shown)


deployment/terraform/modules/aws/vpc/outputs.tf
```diff
@@ -13,3 +13,8 @@ output "public_subnets" {
 output "vpc_cidr_block" {
   value = module.vpc.vpc_cidr_block
 }
+
+output "s3_vpc_endpoint_id" {
+  description = "ID of the S3 gateway VPC endpoint created for this VPC"
+  value       = aws_vpc_endpoint.s3.id
```
jmelahman, line 19: Should this error if defined _and_ create_vpc is true? Otherwise, this would get silently ignored which seems like a potential pitfall

justin-tahara: Given this is the outputs.tf I'm not too worried. There's no where we reference the output as the input for the value so I think it's fine for now.

PR #6809, opened 2025-12-17
feat: ee license tracking - Data Plane Models + Database Schema

nmgarza5:
## Description

Foundational models and database schema for EE license tracking infrastructure.

**Includes:**
- Pydantic models: `LicensePayload`, `LicenseData`, `LicenseMetadata`, `PlanType`, `LicenseSource`
- SQLAlchemy model: `License` (singleton pattern - one row max)
- `ApplicationStatus` enum additions: `ACTIVE`, `GRACE_PERIOD`, `GATED_ACCESS`
- Alembic migration with unique index for singleton enforcement

**Review focus:** Data structures and schema design. No business logic in this PR.

## How Has This Been Tested?

- Migration tested locally
- Models are data-only, tested in subsequent PRs

## Additional Options

- [x] [Optional] Override Linear Check

---
## Summary by cubic
Adds models and database schema for EE license tracking. Creates a singleton License table to store the signed license blob.

- **New Features**
  - Pydantic models for license payload, signed data, metadata, status responses, and seat usage.
  - SQLAlchemy License model with timestamps.
  - Alembic migration that creates the license table and a unique index to enforce a single row; the singleton index is also defined in the SQLAlchemy model.
  - ApplicationStatus enum updated to include ACTIVE, GRACE_PERIOD, and GATED_ACCESS.

<sup>Written for commit 0bf5cf5d01a6a3c41444642eacbabc06d4b74055. Summary will update automatically on new commits.</sup>

PR #6810, opened 2025-12-17
feat: ee license tracking - Crypto Verification Utils

nmgarza5:
## Description

RSA-4096 signature verification utilities for license validation.

**Includes:**
- `verify_license_signature()`: Decode base64 license, verify RSA signature
- `get_license_status()`: Compute status (ACTIVE → GRACE_PERIOD → GATED_ACCESS)
- `is_license_valid()`: Simple expiry check helper
- Placeholder public key embedded (to be replaced with real key)

**Review focus:** Cryptographic verification logic. Pure functions, no side effects.

**Stack:** Depends on #6809 (Models + Schema)

## How Has This Been Tested?

Comprehensive unit tests with mocked RSA key pairs covering:
- Valid/invalid signatures
- Expired licenses
- Grace period logic
- Malformed input handling

## Additional Options

---
## Summary by cubic
Add RSA-4096 license verification and status helpers, plus license models and a singleton DB table to store signed licenses and gate access.

- **New Features**
  - verify_license_signature(): verify RSA-SHA256 (PSS) and return payload.
  - get_license_status(): ACTIVE, GRACE_PERIOD, or GATED_ACCESS.
  - is_license_valid(): simple expiry check.
  - LicensePayload/LicenseData models and License SQLAlchemy model with singleton index + migration.
  - Unit tests cover valid/invalid signatures, tampering, expiry, grace period, and malformed input.

- **Migration**
  - Set LICENSE_PUBLIC_KEY_PEM to the control-plane RSA-4096 public key.
  - Run DB migration to create the license table.

<sup>Written for commit 00fe50049ab462e4615c66c70cd3e68280375036. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/ee/onyx/utils/license.py
+++ b/backend/ee/onyx/utils/license.py
@@ -2,6 +2,7 @@
 
 import base64
 import json
+import os
 from datetime import datetime
 from datetime import timezone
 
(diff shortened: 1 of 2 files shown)


backend/ee/onyx/utils/license.py
```diff
@@ -0,0 +1,133 @@
...
+0000000000000000000000000000000000000000000000000000000000000000000
+0000000000000000000000000000000000000000000000000000000000000000000
+0000000000000000000000000000000000000000000000000000000000000000000
+0000000000000000000000000000000000000000000000000000000000000000000
+0000000000000000000000000000000000000000000000000000000000000000000
+0000000000000000000000000000000000000000000000000000000000000000000
+0000000000000000000000000000000000000000000000000000000000000000000
+0000000000000000000000000000000000000000000000000000000000000000000
+00000000000000000000000000000000000000000000000000000000000AQAB
+-----END PUBLIC KEY-----"""
```
justin-tahara: Even this we might want to store in a secret or reference it from elsewhere. Should be fine but I think we don't need to hardcode keys like this

nmgarza5: makes sense. updated to load the key from an env var. how that env var gets set will be determined by the deployment type

PR #6813, opened 2025-12-17
fix: Open sub menu on tool force

raunakab:
## Description

This PR opens up the subtool menu when forcing a tool in the `ActionsPopover`.

Addresses: https://linear.app/danswer/issue/DAN-2971/tools-selector-enabling-tools-should-also-open-the-sub-menu.

## Video

https://github.com/user-attachments/assets/fbe7705f-e37c-4ed1-988c-4556257fb2db

---
## Summary by cubic
Enabling the Search tool now auto-opens the Sources sub menu so users can configure connectors immediately (Linear DAN-2971).

- **Bug Fixes**
  - Keep ActionsPopover open only when forcing Search outside a project and open the Sources view; close after forcing other tools.
  - Use ProjectsContext; show “Project Search” in project context and hide Search when the project has no files.
  - In non-project context, show Settings when no connectors (link to add connectors) and Chevron to configure sources otherwise.

<sup>Written for commit 947e66ecef816384db2aa361361f6fb2b964ddb0. Summary will update automatically on new commits.</sup>

### DIFF
--- a/web/src/refresh-components/popovers/ActionsPopover/ActionLineItem.tsx
+++ b/web/src/refresh-components/popovers/ActionsPopover/ActionLineItem.tsx
@@ -12,7 +12,7 @@ import { cn, noProp } from "@/lib/utils";
 import type { IconProps } from "@opal/types";
 import { SvgChevronRight, SvgKey, SvgSettings, SvgSlash } from "@opal/icons";
 import { useProjectsContext } from "@/app/chat/projects/ProjectsContext";
-import { useRouter } from "next/router";
+import { useRouter } from "next/navigation";
 
 export interface ActionItemProps {
   tool?: ToolSnapshot;
@@ -26,6 +26,7 @@ export interface ActionItemProps {
   hasNoConnectors?: boolean;
   toolAuthStatus?: ToolAuthStatus;
   onOAuthAuthenticate?: () => void;
+  onClose?: () => void;
 }
 
 export default function ActionLineItem({
@@ -40,6 +41,7 @@ export default function ActionLineItem({
   hasNoConnectors = false,
   toolAuthStatus,
   onOAuthAuthenticate,
+  onClose,
 }: ActionItemProps) {
   const router = useRouter();
   const { currentProjectId } = useProjectsContext();
@@ -70,6 +72,7 @@ export default function ActionLineItem({
             onForceToggle();
             if (isSearchToolAndNotInProject && !isForced)
               onSourceManagementOpen?.();
+            else onClose?.();
           }}
           selected={isForced}
           strikethrough={disabled || isSearchToolWithNoConnectors}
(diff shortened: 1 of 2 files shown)


web/src/refresh-components/popovers/ActionsPopover/ActionLineItem.tsx
```diff
@@ -100,32 +106,28 @@ export default function ActionLineItem({
...
+                    else onSourceManagementOpen?.();
+                  })}
+                  internal
+                  className={cn(
+                    isSearchToolWithNoConnectors &&
+                      "invisible group-hover/LineItem:visible"
+                  )}
+                  tooltip={
+                    isSearchToolWithNoConnectors
+                      ? "Setup Connectors"
```
nmgarza5: I feel like we can be more prescriptive here. Can this just be "Add connectors"?

web/src/refresh-components/popovers/ActionsPopover/index.tsx
```diff
@@ -523,15 +522,11 @@ export default function ActionsPopover({
             disabled={disabledToolIds.includes(tool.id)}
             isForced={forcedToolIds.includes(tool.id)}
             onToggle={() => toggleToolForCurrentAssistant(tool.id)}
-            onForceToggle={() => {
-              toggleForcedTool(tool.id);
-              setOpen(false);
```
nmgarza5, line 528: just to be clear...we are now leaving the submenu open after selecting a tool?

web/src/refresh-components/popovers/ActionsPopover/ActionLineItem.tsx
```diff
@@ -100,32 +106,28 @@ export default function ActionLineItem({
...
-                  />
-                )}
+              {isSearchToolAndNotInProject && (
+                <IconButton
+                  icon={
+                    isSearchToolWithNoConnectors ? SvgSettings : SvgChevronRight
+                  }
+                  onClick={noProp(() => {
+                    if (isSearchToolWithNoConnectors)
+                      window.location.href = "/admin/add-connector";
```
nmgarza5: since you are here....lets not user window.location for navigation. prefer to use next's built in router: https://nextjs.org/docs/pages/api-reference/functions/use-router 
```
import { useRouter } from "next/navigation";
const router = useRouter();
...
...
if (isSearchToolWithNoConnectors) {
	router.push("/admin/add-connector");
}
```

or... we can use Link components: https://nextjs.org/docs/app/api-reference/components/link

jmelahman: +1 Link components. Not being able to right-click and open links in a new tab is a pet peeve of mine

web/src/refresh-components/popovers/ActionsPopover/index.tsx
raunakab: Yup!

web/src/refresh-components/popovers/ActionsPopover/ActionLineItem.tsx
raunakab: Oh good catch! Updated.

web/src/refresh-components/popovers/ActionsPopover/ActionLineItem.tsx
raunakab: Updated.

web/src/refresh-components/popovers/ActionsPopover/index.tsx
raunakab: Actually, I updated the behaviour to close the popover for the other tool types.

PR #6822, opened 2025-12-17
feat(tools): extend open_url to handle indexed content urls

jessicasingh7:
## Description
[DAN-2327](https://linear.app/danswer/issue/DAN-2327/single-file-picker-tool)
Vague query with title or info -> internal search -> open_url
Exact url of GDrive/JIRA -> open_url

## How Has This Been Tested?

Unit tests and manual testing in UI

## Additional Options

---
## Summary by cubic
Extended OpenURL to fetch indexed documents via document identifiers and connector-specific URL normalization, with permission-aware retrieval and parallel crawling that prefers indexed content. Enables the single-file picker flow (DAN-2327) by opening exact Drive/Notion/Slack/Linear links or search result identifiers.

- **New Features**
  - Accepts URLs and document_identifiers.
  - Connector-aware URL normalization (Drive, Notion, Slack, Linear) to map to Document.id or link, with link-based fallback when needed (e.g., Linear).
  - Permission-aware indexed retrieval (ACLs, tenant), with crawl fallback.
  - Runs indexed retrieval and crawling in parallel, merging results and preferring indexed content.
  - Includes document_identifier in LLM and UI outputs.

- **Refactors**
  - Cache DocumentIndex in tool_constructor to avoid repeated initialization.
  - OpenURLTool now receives document_index and user; uses tenant-aware DB sessions.
  - Added url_normalization module with unit tests.
  - Added fast DB checks for existing Document.id and link-based lookup.

<sup>Written for commit c3ae36ac3c7996ee1cced259cd9a48e431908183. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/connectors/interfaces.py
+++ b/backend/onyx/connectors/interfaces.py
@@ -25,6 +25,18 @@ GenerateSlimDocumentOutput = Iterator[list[SlimDocument]]
 CT = TypeVar("CT", bound=ConnectorCheckpoint)
 
 
+class NormalizationResult(BaseModel):
+    """Result of URL normalization attempt.
+
+    Attributes:
+        normalized_url: The normalized URL string, or None if normalization failed
+        use_default: If True, fall back to default normalizer. If False, return None.
+    """
+
+    normalized_url: str | None
+    use_default: bool = False
+
+
 class BaseConnector(abc.ABC, Generic[CT]):
     REDIS_KEY_PREFIX = "da_connector_data:"
 
@@ -75,13 +87,13 @@ class BaseConnector(abc.ABC, Generic[CT]):
         based on the application level image analysis setting."""
 
     @classmethod
-    def normalize_url(cls, url: str) -> str | None:
+    def normalize_url(cls, url: str) -> "NormalizationResult":
         """Normalize a URL to match the canonical Document.id format used during ingestion.
 
         Connectors that use URLs as document IDs should override this method.
-        Returns None if the URL doesn't belong to this connector or normalization fails.
+        Returns NormalizationResult with use_default=True if not implemented.
         """
-        return None
+        return NormalizationResult(normalized_url=None, use_default=True)
 
     def build_dummy_checkpoint(self) -> CT:
         # TODO: find a way to make this work without type: ignore
--- a/backend/onyx/connectors/slack/connector.py
+++ b/backend/onyx/connectors/slack/connector.py
@@ -42,6 +42,7 @@ from onyx.connectors.interfaces import CheckpointOutput
 from onyx.connectors.interfaces import CredentialsConnector
 from onyx.connectors.interfaces import CredentialsProviderInterface
 from onyx.connectors.interfaces import GenerateSlimDocumentOutput
+from onyx.connectors.interfaces import NormalizationResult
 from onyx.connectors.interfaces import SecondsSinceUnixEpoch
 from onyx.connectors.interfaces import SlimConnectorWithPermSync
 from onyx.connectors.models import BasicExpertInfo
@@ -629,29 +630,29 @@ class SlackConnector(
 
     @classmethod
     @override
-    def normalize_url(cls, url: str) -> str | None:
+    def normalize_url(cls, url: str) -> NormalizationResult:
         """Normalize a Slack URL to extract channel_id__thread_ts format."""
         parsed = urlparse(url)
         if "slack.com" not in parsed.netloc.lower():
-            return None
+            return NormalizationResult(normalized_url=None, use_default=False)
 
         # Slack document IDs are format: channel_id__thread_ts
         # Extract from URL pattern: .../archives/{channel_id}/p{timestamp}
         path_parts = parsed.path.split("/")
         if "archives" not in path_parts:
-            return None
+            return NormalizationResult(normalized_url=None, use_default=False)
 
         archives_idx = path_parts.index("archives")
         if archives_idx + 1 >= len(path_parts):
-            return None
+            return NormalizationResult(normalized_url=None, use_default=False)
 
         channel_id = path_parts[archives_idx + 1]
         if archives_idx + 2 >= len(path_parts):
-            return None
+            return NormalizationResult(normalized_url=None, use_default=False)
 
         thread_part = path_parts[archives_idx + 2]
         if not thread_part.startswith("p"):
-            return None
+            return NormalizationResult(normalized_url=None, use_default=False)
 
         # Convert p1234567890123456 to 1234567890.123456 format
         timestamp_str = thread_part[1:]  # Remove 'p' prefix
@@ -661,7 +662,8 @@ class SlackConnector(
         else:
             thread_ts = timestamp_str
 
-        return f"{channel_id}__{thread_ts}"
+        normalized = f"{channel_id}__{thread_ts}"
+        return NormalizationResult(normalized_url=normalized, use_default=False)
 
     @staticmethod
     def make_credential_prefix(key: str) -> str:
--- a/backend/onyx/tools/tool_implementations/open_url/url_normalization.py
+++ b/backend/onyx/tools/tool_implementations/open_url/url_normalization.py
@@ -49,7 +49,10 @@ def normalize_url(url: str, source_type: DocumentSource | None = None) -> str | 
         try:
             connector_class = identify_connector_class(source_type)
             result = connector_class.normalize_url(url)
-            return result
+
+            if result.use_default:
+                return _default_url_normalizer(url)
+            return result.normalized_url  # Could be None if failed
         except Exception as exc:
             logger.debug(
                 "Failed to normalize URL for source %s: %s. Using default normalizer.",
--- a/backend/onyx/connectors/google_drive/connector.py
+++ b/backend/onyx/connectors/google_drive/connector.py
@@ -66,6 +66,7 @@ from onyx.connectors.google_utils.shared_constants import USER_FIELDS
 from onyx.connectors.interfaces import CheckpointedConnectorWithPermSync
 from onyx.connectors.interfaces import CheckpointOutput
 from onyx.connectors.interfaces import GenerateSlimDocumentOutput
+from onyx.connectors.interfaces import NormalizationResult
 from onyx.connectors.interfaces import SecondsSinceUnixEpoch
 from onyx.connectors.interfaces import SlimConnectorWithPermSync
 from onyx.connectors.models import ConnectorFailure
@@ -285,7 +286,7 @@ class GoogleDriveConnector(
 
     @classmethod
     @override
-    def normalize_url(cls, url: str) -> str | None:
+    def normalize_url(cls, url: str) -> NormalizationResult:
         """Normalize a Google Drive URL to match the canonical Document.id format.
 
         Reuses the connector's existing document ID creation logic from
@@ -298,7 +299,7 @@ class GoogleDriveConnector(
             netloc.startswith("docs.google.com")
             or netloc.startswith("drive.google.com")
         ):
-            return None
+            return NormalizationResult(normalized_url=None, use_default=False)
 
         # Handle ?id= query parameter case
         query_params = parse_qs(parsed.query)
@@ -310,8 +311,10 @@ class GoogleDriveConnector(
             params = ""
             query = ""
             fragment = ""
-            normalized = urlunparse((scheme, netloc, path, params, query, fragment))
-            return normalized.rstrip("/")
+            normalized = urlunparse(
+                (scheme, netloc, path, params, query, fragment)
+            ).rstrip("/")
+            return NormalizationResult(normalized_url=normalized, use_default=False)
 
         # Extract file ID and use connector's function
         path_parts = parsed.path.split("/")
@@ -322,11 +325,12 @@ class GoogleDriveConnector(
                 break
 
         if not file_id:
-            return None
+            return NormalizationResult(normalized_url=None, use_default=False)
 
         # Create minimal file object for connector function
         file_obj = {"webViewLink": url, "id": file_id}
-        return onyx_document_id_from_drive_file(file_obj).rstrip("/")
+        normalized = onyx_document_id_from_drive_file(file_obj).rstrip("/")
+        return NormalizationResult(normalized_url=normalized, use_default=False)
 
     # TODO: ensure returned new_creds_dict is actually persisted when this is called?
     def load_credentials(self, credentials: dict[str, Any]) -> dict[str, str] | None:
--- a/backend/onyx/connectors/linear/connector.py
+++ b/backend/onyx/connectors/linear/connector.py
@@ -19,6 +19,7 @@ from onyx.connectors.cross_connector_utils.miscellaneous_utils import (
 from onyx.connectors.cross_connector_utils.miscellaneous_utils import time_str_to_utc
 from onyx.connectors.interfaces import GenerateDocumentsOutput
 from onyx.connectors.interfaces import LoadConnector
+from onyx.connectors.interfaces import NormalizationResult
 from onyx.connectors.interfaces import OAuthConnector
 from onyx.connectors.interfaces import PollConnector
 from onyx.connectors.interfaces import SecondsSinceUnixEpoch
@@ -317,7 +318,7 @@ class LinearConnector(LoadConnector, PollConnector, OAuthConnector):
 
     @classmethod
     @override
-    def normalize_url(cls, url: str) -> str | None:
+    def normalize_url(cls, url: str) -> NormalizationResult:
         """Extract Linear issue identifier from URL.
 
         Linear URLs are like: https://linear.app/team/issue/IDENTIFIER/...
@@ -327,7 +328,7 @@ class LinearConnector(LoadConnector, PollConnector, OAuthConnector):
         netloc = parsed.netloc.lower()
 
         if "linear.app" not in netloc:
-            return None
+            return NormalizationResult(normalized_url=None, use_default=False)
 
         # Extract identifier from path: /team/issue/IDENTIFIER/...
         # Pattern: /{team}/issue/{identifier}/...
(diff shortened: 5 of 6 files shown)


jessicasingh7:
@greptile

jessicasingh7:
@cubic-dev-ai review this PR

backend/onyx/tools/tool_implementations/open_url/open_url_tool.py
```diff
@@ -213,43 +315,83 @@ def run(
...
-        # Filter out failed fetches and convert to InferenceSections
-        inference_sections = []
-        for content in web_contents:
-            if content.scrape_successful and content.full_content:
-                inference_sections.append(
-                    inference_section_from_internet_page_scrape(content)
-                )
-            else:
-                logger.warning(f"Failed to fetch content from URL: {content.link}")
+        db_session = self._get_thread_safe_session()
```
evan-onyx: nit: replace the try-finally with a context manager pattern:
```
with self._session_factory() as db_session:
```

backend/onyx/tools/tool_implementations/open_url/open_url_tool.py
```diff
@@ -172,8 +255,16 @@ def tool_definition(self) -> dict:
                             "items": {"type": "string"},
                             "description": "List of URLs to open and read. Can be a single URL or multiple URLs.",
                         },
+                        DOCUMENT_IDENTIFIERS_FIELD: {
+                            "type": "array",
+                            "items": {"type": "string"},
+                            "description": (
+                                "List of Onyx document identifiers (returned by internal search). "
+                                "Use when available for faster, permission-aware retrieval."
```
evan-onyx: all retrieval is permission aware so we can just say "Use when available" to save tokens!

jessicasingh7:
@greptile

backend/onyx/tools/tool_implementations/open_url/open_url_tool.py
```diff
@@ -193,18 +280,29 @@ def run(
         override_kwargs: OpenURLToolOverrideKwargs,
         **llm_kwargs: Any,
     ) -> ToolResponse:
-        """Execute the open URL tool to fetch content from the specified URLs.
+        """Execute the OpenURL tool using document identifiers when possible."""
+        raw_urls = llm_kwargs.get(URLS_FIELD, [])
+        if isinstance(raw_urls, str):
+            raw_urls = [raw_urls]
+        urls = _dedupe_preserve_order(
+            [str(url).strip() for url in cast(list[str], raw_urls) if str(url).strip()]
```
yuhongsun96: The typing here is really redundant and hard to read, let's clean it up

backend/onyx/tools/tool_implementations/open_url/open_url_tool.py
```diff
@@ -17,17 +31,66 @@
...
+    return ordered
+
+
+def _url_lookup_variants(url: str) -> set[str]:
+    """Generate URL variants (with/without trailing slash) for database lookup.
+
+    This is used after normalize_url() to create variants for fuzzy matching
+    in the database, since URLs may be stored with or without trailing slashes.
+    """
+    from onyx.tools.tool_implementations.open_url.url_normalization import (
```
yuhongsun96: These should be module level imports. 

Performance – Import overhead happens once at startup, not on every function call. Python caches imported modules, but there's still a small lookup cost each time.
Clarity & Discoverability – All dependencies are visible at the top of the file. You can immediately see what the module depends on.
Fail-Fast – Import errors surface immediately when the module loads, not later during runtime when a function happens to be called.
PEP 8 Convention – Python's style guide recommends imports at the top.

backend/onyx/tools/tool_implementations/open_url/open_url_tool.py
```diff
@@ -172,8 +251,16 @@ def tool_definition(self) -> dict:
                             "items": {"type": "string"},
                             "description": "List of URLs to open and read. Can be a single URL or multiple URLs.",
                         },
+                        DOCUMENT_IDENTIFIERS_FIELD: {
```
yuhongsun96: I don't think we should complicate the open_url tool with additional params, good chance this will trip up the LLM for a lot of flows. Why is this needed?

backend/onyx/tools/tool_implementations/open_url/open_url_tool.py
```diff
@@ -263,3 +399,206 @@ def run(
             ),
             llm_facing_response=docs_str,
         )
+
+    def _build_requests_from_identifiers(
```
yuhongsun96: This is a local helper function which is just a list comprehension and is only used once, I would say this doesn't need to be a function at all

backend/onyx/tools/tool_implementations/open_url/open_url_tool.py
```diff
@@ -193,18 +280,29 @@ def run(
...
+        urls = _dedupe_preserve_order(
+            [str(url).strip() for url in cast(list[str], raw_urls) if str(url).strip()]
+        )
 
-        Args:
-            turn_index: The current turn index in the conversation.
-            override_kwargs: Override arguments including starting citation number
-                and existing citation_mapping to reuse citations for already-cited URLs.
-            **llm_kwargs: Arguments provided by the LLM, including the 'urls' field.
+        raw_identifiers = llm_kwargs.get(DOCUMENT_IDENTIFIERS_FIELD, [])
```
yuhongsun96: We don't want to have the LLM decide if it's an existing doc in the index or not. We should just try to handle it from just the URLS_FIELD in parallel.

backend/onyx/tools/tool_implementations/open_url/open_url_tool.py
```diff
@@ -193,18 +280,29 @@ def run(
...
+                for identifier in cast(list[str], raw_identifiers)
+                if str(identifier).strip()
+            ]
+        )
 
-        Returns:
-            ToolResponse containing the fetched content and citation mapping.
-        """
-        urls = cast(list[str], llm_kwargs[URLS_FIELD])
+        if not urls and not document_identifiers:
```
yuhongsun96: A lot of downstream logic comes from this wrong premise of having the LLM choose if the link is an external or internal link. I won't comment on other similar ones from here on

backend/onyx/tools/tool_implementations/open_url/open_url_tool.py
```diff
@@ -263,3 +399,206 @@ def run(
...
+        seen: set[str] = set()
+        deduped: list[IndexedDocumentRequest] = []
+        for request in requests:
+            if request.document_id in seen:
+                continue
+            seen.add(request.document_id)
+            deduped.append(request)
+        return deduped
+
+    def _resolve_urls_to_document_ids(
```
yuhongsun96: These functions that don't rely on the object state should not be scoped to the object and require self.

backend/onyx/tools/tool_implementations/open_url/open_url_tool.py
```diff
@@ -263,3 +399,206 @@ def run(
...
+                    unresolved.append(url)
+
+        if not normalized_map:
+            return matches, unresolved
+
+        # Query database with all normalized variants
+        all_variants = {
+            variant for variants in normalized_map.values() for variant in variants
+        }
+        document_rows = (
```
yuhongsun96: image
Also all DB table related operations should be under the db directory

backend/onyx/tools/tool_implementations/open_url/url_normalization.py
```diff
@@ -0,0 +1,314 @@
...
+    return None
+
+
+# ============================================================================
+# Connector Normalizers
+# Each connector registers its normalization logic here
+# ============================================================================
+
+
+def _register_google_drive_normalizer() -> None:
```
yuhongsun96: This should be part of the connector interface somehow, that way implementing classes are aware of this and the functionality can simply be reused from there.

backend/onyx/tools/tool_implementations/open_url/url_normalization.py
```diff
@@ -0,0 +1,314 @@
...
+from onyx.configs.constants import DocumentSource
+from onyx.utils.logger import setup_logger
+
+logger = setup_logger()
+
+# Registry: DocumentSource -> normalization function
+_NORMALIZERS: dict[DocumentSource, Callable[[str], str | None]] = {}
+
+
+def register_normalizer(
```
yuhongsun96: image

But the google drive util is a simple pure function, this should not be allowed to fail

yuhongsun96: Regardless of this, I think it should not be handled in a separate module. It should be part of the connectors themselves and we call their function for this.

backend/onyx/tools/tool_implementations/open_url/url_normalization.py
```diff
@@ -0,0 +1,314 @@
...
+    """
+    try:
+        parsed = urlparse(url)
+        if not parsed.netloc:
+            return None
+
+        # Strip query params and fragment, normalize trailing slash
+        normalized = urlunparse(
+            (
+                parsed.scheme or "https",
```
yuhongsun96: for functions with a lot of args, especially some of these mysterious empty strings, pref providing the arg names in front. This also prevents future refactors from mixing up the orders etc just because the types match.

backend/onyx/tools/tool_implementations/open_url/open_url_tool.py
jessicasingh7: I wanted to make sure if open_url is called after internal search it can handle file connectors that don't have urls but have removed this param now and made the description more open

backend/onyx/tools/tool_implementations/open_url/open_url_tool.py
```diff
@@ -163,17 +258,34 @@ def tool_definition(self) -> dict:
             "type": "function",
             "function": {
                 "name": self.name,
-                "description": "Open and read the content of one or more URLs. Returns the text content of the pages.",
+                "description": (
```
yuhongsun96: I would not change the description of the tool and how it's used

backend/onyx/tools/tool_implementations/open_url/open_url_tool.py
```diff
@@ -163,17 +258,34 @@ def tool_definition(self) -> dict:
...
+                                "or (3) Non-URL document identifiers for file connectors "
+                                "(e.g., 'FILE_CONNECTOR__abc-123'). "
+                                "Use the 'document_identifier' field from search results when 'url' is not available. "
+                                "You can extract URLs or document_identifier values from search results "
+                                "to read those documents in full."
+                            ),
                         },
                     },
-                    "required": [URLS_FIELD],
+                    "required": [],
```
yuhongsun96: this should still be required, this part of the code just shouldn't change imo

backend/onyx/tools/tool_implementations/open_url/open_url_tool.py
```diff
@@ -213,43 +318,92 @@ def run(
...
-            if content.scrape_successful and content.full_content:
-                inference_sections.append(
-                    inference_section_from_internet_page_scrape(content)
-                )
-            else:
-                logger.warning(f"Failed to fetch content from URL: {content.link}")
+        with get_session_with_current_tenant() as db_session:
+            # Resolve URLs to document IDs for indexed retrieval
+            # Handles both raw URLs and already-normalized document IDs
+            url_requests, _ = OpenURLTool._resolve_urls_to_document_ids(
```
yuhongsun96: Why make this part of the class?

backend/onyx/connectors/slack/connector.py
```diff
@@ -626,6 +627,37 @@ def __init__(
...
+                    if thread_part.startswith("p"):
+                        # Convert p1234567890123456 to 1234567890.123456 format
+                        timestamp_str = thread_part[1:]  # Remove 'p' prefix
+                        if len(timestamp_str) == 16:
+                            # Insert dot at position 10 to match canonical format
+                            thread_ts = f"{timestamp_str[:10]}.{timestamp_str[10:]}"
+                        else:
+                            thread_ts = timestamp_str
+                        return f"{channel_id}__{thread_ts}"
+        except (ValueError, IndexError):
```
yuhongsun96: Should avoid silently swallowing errors generally, why are these ok and expected? Maybe it's fine but I feel this may cause unknown errors that silently degrade product quality

backend/onyx/connectors/interfaces.py
```diff
@@ -76,6 +76,15 @@ def set_allow_images(self, value: bool) -> None:
         """Implement if the underlying connector wants to skip/allow image downloading
         based on the application level image analysis setting."""
 
+    @classmethod
```
yuhongsun96, line 89: Can we be more strict with the typing? Does it need to have a return None? Would one expect to call this function and have to handle a None case?

backend/onyx/tools/tool_implementations/open_url/url_normalization.py
```diff
@@ -0,0 +1,122 @@
...
+        fragment = ""  # Fragment/hash (removed)
+
+        normalized = urlunparse((scheme, netloc, path, params, query, fragment))
+        return normalized or None
+    except Exception as e:
+        logger.debug(f"Default normalizer failed for URL {url}: {e}")
+        return None
+
+
+def normalize_url(url: str, source_type: DocumentSource | None = None) -> str | None:
```
yuhongsun96, line 39: Looking at this code, I can see several opportunities to make it cleaner and more readable. The main issues are:
Duplicated logic - The "try connector → fall back to default" pattern is repeated twice
Excessive nesting - Multiple levels of try/except and if/else
Verbose logging - Logging statements scattered throughout add noise

I think in general there is an over-usage of try/excepts and this code feels like it can be organized differently. This may be a stylistic difference, maybe it's worth a brief discussion.

backend/onyx/tools/tool_implementations/search/search_tool.py
```diff
@@ -657,15 +603,8 @@ def expand_section_safe(
...
-            )
-
-            # End overall timing
-            overall_elapsed = time.time() - overall_start_time
-            logger.debug(
-                f"Search tool - Total execution time: {overall_elapsed:.3f} seconds "
-                f"(query expansion: {query_expansion_elapsed:.3f}s, "
-                f"document selection: {document_selection_elapsed:.3f}s, "
-                f"document expansion: {document_expansion_elapsed:.3f}s)"
+                include_link=True,
```
yuhongsun96: Why are we adding these?

backend/onyx/tools/tool_implementations/web_search/web_search_tool.py
```diff
@@ -214,6 +214,7 @@ def run(
             limit=None,  # Already truncated
             include_source_type=False,
             include_link=True,
+            include_document_id=True,
```
yuhongsun96: I don't think it should include this, this might cause confusion to the LLM

backend/onyx/tools/tool_implementations/utils.py
```diff
@@ -96,6 +72,8 @@ def convert_inference_sections_to_llm_string(
                 # source_links is dict[int, str], get the first value
                 link = next(iter(chunk.source_links.values()), None)
             result["url"] = link
+        if include_document_id:
```
yuhongsun96: What does this look like for web search?

backend/onyx/tools/tool_implementations/open_url/open_url_tool.py
```diff
@@ -101,26 +179,43 @@ def _convert_sections_to_llm_string_with_citations(
...
+            project_id: Optional project scope for filters.
             content_provider: Optional content provider. If not provided,
                 will use the default provider from the database or fall back
                 to the built-in Onyx web crawler.
         """
         super().__init__(emitter=emitter)
         self._id = tool_id
+        self._document_index = document_index
+        self._user = user
+        self._project_id = project_id
```
yuhongsun96: I think it's fine to not scope it to this honestly. If the LLM generates a specific URL or the user asks for a specific URL/doc which isn't part of the project, I think it's ok to show it. Thoughts?

yuhongsun96 (commented):
I think we're quite close, a few comments on style and product related things which we should address. Also 1 point to discuss with the team, and should clean up the bot comments (such as unused params).

Also please rebase before requesting a next look!

backend/onyx/tools/tool_implementations/open_url/open_url_tool.py
jessicasingh7: I agree, removed!

backend/onyx/tools/tool_implementations/utils.py
jessicasingh7: Made this param False for web search

backend/onyx/connectors/interfaces.py
jessicasingh7: None falls back to generic url normalization so I think it should be okay

backend/onyx/tools/tool_implementations/search/search_tool.py
jessicasingh7: This is so that if internal search returns a Google doc it returns a url or if it's an internal file then it can return the document_id, and the appropriate one is used later by open_url tool

backend/onyx/connectors/interfaces.py
jessicasingh7: None would fallback to the default url normalization method if there is no specific one for that connector.

backend/onyx/tools/tool_implementations/open_url/open_url_tool.py
jessicasingh7: I agree, moved it out of the class!

backend/onyx/connectors/interfaces.py
yuhongsun96: mk seems fine

backend/onyx/tools/tool_implementations/search/search_tool.py
yuhongsun96: I am very cautious about what is sent to the LLM. I would not change this one or the definition of the open_url tool

PR #6850, opened 2025-12-18
fix: clean up disable-simple-agent-framework feature flag and send backend request to stop generation always

rohoswagger:
## Description

Previously, the stopGenerating function conditionally called the backend stop API based on the `disable-simple-agent-framework` feature flag. When the flag was enabled (true), the backend stop was skipped entirely, leaving the backend processing indefinitely even after the frontend aborted.
Changes:
- Always call stopChatSession() to set Redis fence and signal backend to stop
- Always call abortSession() to immediately cancel the frontend fetch
- Always clean up incomplete tool calls and inject STOP packet for proper UI state
- Remove feature flag checks from SearchToolRenderer for consistent V2 rendering
This fixes the issue where users couldn't refresh the page after stopping generation because the backend was still processing.

## How Has This Been Tested?

Locally
Stopping actually works!

## Additional Options


rohoswagger:
@greptile

web/src/app/chat/hooks/useChatController.ts
```diff
@@ -309,66 +306,59 @@ export function useChatController({
...
+      newMessageTree.set(lastMessage.nodeId, updatedMessage);
+      updateSessionMessageTree(currentSession, newMessageTree);
     }
 
-    // Only do the subsequent cleanup if the message was agent search or feature flag is enabled to disable it
-    if (isDeepResearch || isSimpleAgentFrameworkDisabled) {
-      abortSession(currentSession);
+    // Ensure UI reflects a STOP event by appending a STOP packet to the
+    // currently streaming assistant message if one exists and doesn't already
+    // contain a STOP. This makes AIMessage behave as if a STOP packet arrived.
```
nmgarza5, line 336: for my own understanding..can you explain why the agent wouldn't ALWAYS send a STOP packet?

Also, let's update the comment from `assistant` -> `agent`. This is a migration we are slowly rolling out but all new references should use agent.

web/src/app/chat/message/messageComponents/MultiToolRenderer.tsx
```diff
@@ -23,7 +23,7 @@ import {
   SearchToolStep1Renderer,
   SearchToolStep2Renderer,
   constructCurrentSearchState,
-} from "./renderers/SearchToolRendererV2";
+} from "./renderers/SearchToolRenderer";
```
nmgarza5, line 26: was the v2 implementation tied to the simple agent framework? seems like the removal should be in a separate PR if not. This comment applies to the actual deletion of the component too

nmgarza5: one of the comments in the pr description is as follows so im a little confused

```
Remove feature flag checks from SearchToolRenderer for consistent V2 rendering
This fixes the issue where users couldn't refresh the page after stopping generation because the backend was still processing.
```

web/src/app/chat/message/messageComponents/renderers/SearchToolRenderer.tsx
```diff
@@ -76,31 +78,161 @@ const constructCurrentSearchState = (
...
+    isInternetSearch,
+  };
 };
 
-export const SearchToolRenderer: MessageRenderer<
-  SearchToolPacket,
-  { researchType?: string | null }
-> = ({
+// Step 1 Renderer: "Searching internally" with queries
+export function SearchToolStep1Renderer({
```
nmgarza5, line 96: I think naming could be better here. Something like `SourceRetrievalRenderer` or something similar would be more descriptive

web/src/app/chat/message/messageComponents/renderers/SearchToolRenderer.tsx
```diff
@@ -1,5 +1,5 @@
-import React, { useEffect, useState, useRef, useMemo } from "react";
-import { FiSearch, FiGlobe } from "react-icons/fi";
+import React, { useEffect, useState, useRef, useMemo, JSX } from "react";
+import { FiSearch, FiGlobe, FiBookOpen } from "react-icons/fi";
```
nmgarza5, line 2: since you're here....lets replace these icons with opal icons -> `web/lib/opal/src/icons/`

web/src/app/chat/message/messageComponents/renderers/SearchToolRenderer.tsx
```diff
@@ -76,31 +78,161 @@ const constructCurrentSearchState = (
...
+          )}
+          {queries.length === 0 && <BlinkingDot />}
+        </div>
+      </div>
+    ),
+  });
+}
+
+// Step 2 Renderer: "Reading" with documents
+export function SearchToolStep2Renderer({
```
nmgarza5, line 158: same thing about naming. we can be a little bit more clear

web/src/app/chat/message/messageComponents/renderers/SearchToolRenderer.tsx
```diff
@@ -76,31 +78,161 @@ const constructCurrentSearchState = (
...
+                  );
+                }}
+              />
+            </div>
+          )}
+          {results.length === 0 && <BlinkingDot />}
+        </div>
+      </div>
+    ),
+  });
```
nmgarza5, line 226: most of this seems shared with the SearchTool1 component. we should consolidate into a generic SearchToolStepRenderer (can disregard previous comments about naming) and pass icon + status args (and anything else I'm missing)

web/src/app/chat/message/messageComponents/renderers/SearchToolRenderer.tsx
```diff
@@ -255,107 +388,109 @@ export const SearchToolRenderer: MessageRenderer<
...
+                  <SourceChip2
+                    icon={<ResultIcon doc={result} size={10} />}
+                    title={truncateString(
+                      result.semantic_identifier || "",
+                      MAX_TITLE_LENGTH
+                    )}
+                    onClick={() => {
+                      if (result.link) {
+                        window.open(result.link, "_blank");
+                      }
```
nmgarza5, line 459: lets use a react/next Link component instead of window.open

nmgarza5 (changes requested):
mostly looks fine but l left a number of comments on things to consolidate. I'll give a stamp after that is done 😄

PR #6851, opened 2025-12-18
chore: better cloud metrics

Weves:
## Description

Also drops the `milestone` table since it feels a bit silly + wasn't really used.

## How Has This Been Tested?

eyes

## Additional Options

---
## Summary by cubic
Replaced milestone DB tracking with direct cloud telemetry to fix miscounted PostHog metrics and remove DB overhead. Drops the milestone table and switches all milestone events to mt_cloud_telemetry across signup, connectors, chat, personas, Slack bot, and indexing.

- **Refactors**
  - Replaced create_milestone_and_report with mt_cloud_telemetry and added consistent tenant_id properties.
  - Standardized distinct_id (user.email when available; else tenant_id).
  - Removed MULTIPLE_USERS event; always emit USER_SIGNED_UP.
  - Deleted milestone-related modules and code paths (db/milestone.py, chat_milestones.py).
  - Removed deprecated legacy indexing runner; emit CONNECTOR_SUCCEEDED on success in Celery path.

- **Migration**
  - Run Alembic upgrade to drop the milestone table (b8c9d0e1f2a3).
  - No data migration needed; app no longer reads milestone records.
  - Update any dashboards that referenced MULTIPLE_USERS or milestone table.

<sup>Written for commit 2978ce83c85083777f5537a016bdef4aa4ad1f1e. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/utils/telemetry.py
+++ b/backend/onyx/utils/telemetry.py
@@ -156,7 +156,7 @@ def mt_cloud_telemetry(
 
     # Automatically include tenant_id in properties
     all_properties = {**properties} if properties else {}
-    if "tenant_id" in properties:
+    if properties and "tenant_id" in properties:
         logger.warning(
             f"tenant_id already in properties: {properties}. "
             f"Overwriting with new value {tenant_id}."


backend/onyx/configs/constants.py
```diff
@@ -332,7 +332,6 @@ class FileType(str, Enum):
 class MilestoneRecordType(str, Enum):
     TENANT_CREATED = "tenant_created"
     USER_SIGNED_UP = "user_signed_up"
-    MULTIPLE_USERS = "multiple_users"
```
Weves, line 335: just having a used_signed_up and now relying on the sql query to figure out if there are multiple users

backend/onyx/background/indexing/run_docfetching.py
```diff
@@ -272,583 +249,6 @@ def _check_failure_threshold(
         )
 
 
-# NOTE: this is the old run_indexing function that the new decoupled approach
```
Weves, line 275: see this comment

backend/onyx/configs/constants.py
wenxi-onyx: Just FYI - `MULTIPLE_USERS` was fired only once for every tenant where a second user joined.

Additionally, `USER_SIGNED_UP` was only fired once after a tenant was created and the user joined (according to you, this was to track tenant provisioning failures). 

Make sure to understand the actual use and semantic meaning of the events are not the same --> your change will result in many more `USER_SIGNED_UP` events and we will no longer have a "user successfully entered entered their newly provisioned tenant."

backend/onyx/background/indexing/run_docfetching.py
wenxi-onyx: Feel free to commit farm and make this a separate PR :big_grin:

backend/onyx/utils/telemetry.py
```diff
@@ -145,34 +146,28 @@ def telemetry_logic() -> None:
 
 
 def mt_cloud_telemetry(
+    tenant_id: str,
     distinct_id: str,
     event: MilestoneRecordType,
     properties: dict | None = None,
```
wenxi-onyx, line 152: You can default to an empty dict 
`properties: dict | None = {}`

To allow:
```
# Automatically include tenant_id in properties
    all_properties = {**properties}
    if "tenant_id" in properties:
        logger.warning(
            f"tenant_id already in properties: {properties}. "
            f"Overwriting with new value {tenant_id}."
        )
    all_properties["tenant_id"] = tenant_id
 ```

PR #6853, opened 2025-12-18
fix(admin): Sidebar Scroll

justin-tahara:
## Description

The admin sidebar would not stay static and would jump around when you clicked on a separate option like Groups or Users 

## How Has This Been Tested?

Tested this locally to make sure that it works properly.

## Additional Options

---
## Summary by cubic
Fixes sidebar scroll jumping by persisting scroll position and preventing automatic scroll on navigation. Applies to both Admin and App sidebars.

- **Bug Fixes**
  - Added scrollKey to SidebarBody/OverflowDiv to store and restore scrollTop per sidebar.
  - Persisted scroll in sessionStorage and restored on route changes with useLayoutEffect.
  - Set Link scroll={false} in SidebarTab to stop Next.js from resetting scroll.

<sup>Written for commit 34e21bdca46db0dbe88aa4617a1b10a6cb78f1f0. Summary will update automatically on new commits.</sup>

### DIFF
--- a/web/src/refresh-components/OverflowDiv.tsx
+++ b/web/src/refresh-components/OverflowDiv.tsx
@@ -23,8 +23,7 @@ export interface VerticalShadowScrollerProps
   scrollKey?: string;
 }
 
-// Global map to store scroll positions across renders
-const scrollPositions = new Map<string, number>();
+const SCROLL_POSITION_PREFIX = "onyx-scroll-";
 
 export default function OverflowDiv({
   disableMask,
@@ -45,9 +44,9 @@ export default function OverflowDiv({
     const scrollElement = scrollRef.current;
     if (!scrollElement) return;
 
-    const key = scrollKey; // Capture for closure
+    const storageKey = `${SCROLL_POSITION_PREFIX}${scrollKey}`;
     const handleScroll = () => {
-      scrollPositions.set(key, scrollElement.scrollTop);
+      sessionStorage.setItem(storageKey, scrollElement.scrollTop.toString());
     };
 
     scrollElement.addEventListener("scroll", handleScroll, { passive: true });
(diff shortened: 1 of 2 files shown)


justin-tahara:
@greptileai please review once more

@cubic-dev-ai can you review once more

web/src/refresh-components/OverflowDiv.tsx
```diff
@@ -9,19 +10,53 @@ export interface VerticalShadowScrollerProps
   disableMask?: boolean;
   backgroundColor?: string;
   height?: string;
+  // Unique key to identify this scroll container (required for scroll persistence)
+  scrollKey: string;
 }
 
+// Global map to store scroll positions across renders
+const scrollPositions = new Map<string, number>();
```
raunakab: Instead of a map, could we use the `sessionStorage` instead?

web/src/refresh-components/OverflowDiv.tsx
```diff
@@ -9,19 +10,53 @@ export interface VerticalShadowScrollerProps
   disableMask?: boolean;
   backgroundColor?: string;
   height?: string;
+  // Unique key to identify this scroll container (required for scroll persistence)
+  scrollKey: string;
```
raunakab: Let's make this optional? If `scrollKey` is *not* defined, then the invoker has opted out of scroll-position-saving.

web/src/sections/sidebar/SidebarBody.tsx
```diff
@@ -7,17 +7,21 @@ export interface SidebarBodyProps {
   actionButton?: React.ReactNode;
   children?: React.ReactNode;
   footer?: React.ReactNode;
+  scrollKey: string;
```
raunakab: This one can stay non-optional. `SidebarBody` can enforce that it itself is scroll-save-able, and thus definitely requires this parameter to be passed into it.

PR #6861, opened 2025-12-18
fix(slack): limit thread context fetch to top N messages by relevance

nmgarza5:
## Description

Fixes thread pool exhaustion and potential OOM on API pods when Slack federated search hits rate limits.

**Problem:**
When a Slack search returns many messages (e.g., 65+), the code spawns unbounded parallel threads to fetch thread context for each message via `conversations.replies` API. When Slack rate limits (429), these threads hang waiting for retries, saturating the API server's thread pool and blocking all other requests. With `max_messages_per_query=25` (old default), a single user query could spawn 25+ threads that all hang on rate limit, causing cascading failures across the pod.

**Solution:**
1. **Batch processing with rate limit detection** - Process thread context fetches in controlled batches (default 5 concurrent). When a 429 is detected, immediately stop sending additional batches and gracefully degrade remaining messages to their original text.

2. **Limit enrichment to top N messages** - Only fetch thread context for the top 5 most relevant messages (configurable via `MAX_SLACK_THREAD_CONTEXT_MESSAGES`). Messages beyond this limit keep their original text, which is sufficient since only top results matter for answer generation.

3. **Reduce default query size** - Lower `max_messages_per_query` from 25 to 10. Fetching 25 messages when only 5 get enriched wastes API calls and increases rate limit risk.

**Env vars for hot-fix tuning:**
- `SLACK_THREAD_CONTEXT_BATCH_SIZE` (default 5) - concurrent API calls per batch
- `MAX_SLACK_THREAD_CONTEXT_MESSAGES` (default 5) - max messages to enrich with thread context

## How Has This Been Tested?

- Unit tests for rate limit detection (429 raises `SlackRateLimitError`)
- Unit tests for batch processing (respects batch size, stops on rate limit)
- Unit tests for max_messages limiting (only top N get enriched, rest keep original text)

## Additional Options

### DIFF
--- a/backend/onyx/context/search/federated/slack_search.py
+++ b/backend/onyx/context/search/federated/slack_search.py
@@ -829,14 +829,21 @@ def fetch_thread_contexts_with_rate_limit_handling(
 
     results: list[str] = []
     rate_limited = False
+    total_batches = (len(messages_for_context) + batch_size - 1) // batch_size
+    rate_limit_batch = 0
 
     # Process in batches
     for i in range(0, len(messages_for_context), batch_size):
+        current_batch = i // batch_size + 1
+
         if rate_limited:
             # Skip remaining batches, use original message text
             remaining = messages_for_context[i:]
+            skipped_batches = total_batches - rate_limit_batch
             logger.warning(
-                f"Skipping {len(remaining)} thread context fetches due to rate limiting"
+                f"Slack rate limit: skipping {len(remaining)} remaining messages "
+                f"({skipped_batches} of {total_batches} batches). "
+                f"Successfully enriched {len(results)} messages before rate limit."
             )
             results.extend([msg.text for msg in remaining])
             break
@@ -866,18 +873,20 @@ def fetch_thread_contexts_with_rate_limit_handling(
                 logger.error(f"Unexpected None result for message {j} in batch")
                 results.append(batch[j].text)
                 rate_limited = True
+                rate_limit_batch = current_batch
             elif result.is_rate_limited:
                 # Rate limit hit - use original text, stop further batches
                 results.append(result.text)
                 rate_limited = True
+                rate_limit_batch = current_batch
             else:
                 # Success or recoverable error - use the text (enriched or original)
                 results.append(result.text)
 
         if rate_limited:
             logger.warning(
-                f"Rate limit detected at batch {i // batch_size + 1}, "
-                f"stopping further thread context fetches"
+                f"Slack rate limit (429) hit at batch {current_batch}/{total_batches} "
+                f"while fetching thread context. Stopping further API calls."
             )
 
     # Add original text for messages we didn't fetch context for


nmgarza5:
@greptile

nmgarza5:
@greptile

backend/onyx/configs/app_configs.py
```diff
@@ -583,6 +583,16 @@ def get_current_tz_offset() -> int:
 SLACK_NUM_THREADS = int(os.getenv("SLACK_NUM_THREADS") or 8)
 MAX_SLACK_QUERY_EXPANSIONS = int(os.environ.get("MAX_SLACK_QUERY_EXPANSIONS", "5"))
 
+# Slack federated search thread context settings
+# Batch size for fetching thread context (controls concurrent API calls per batch)
+SLACK_THREAD_CONTEXT_BATCH_SIZE = int(
+    os.environ.get("SLACK_THREAD_CONTEXT_BATCH_SIZE", "5")
+)
```
justin-tahara, line 590: I think we should specify that this is the sub questions that we do search slack for. I think this is what this variable is controlling.

backend/onyx/context/search/federated/slack_search.py
```diff
@@ -658,19 +682,44 @@ def get_contextualized_thread_text(
         response.validate()
         messages: list[dict[str, Any]] = response.get("messages", [])
     except SlackApiError as e:
-        logger.error(f"Slack API error in get_contextualized_thread_text: {e}")
-        return message.text
+        # Check for rate limit error specifically
+        if e.response and e.response.status_code == 429:
```
justin-tahara, line 686: If we can do a quick search to see if there are any other error codes that we may hit, I think we should take a quick peek while we're at it.

backend/onyx/context/search/federated/slack_search.py
```diff
@@ -658,19 +682,44 @@ def get_contextualized_thread_text(
...
+        logger.error(f"Slack API error in thread context fetch: {e}")
+        return ThreadContextResult.error(message.text)
+    except Exception as e:
+        # Network errors, timeouts, etc - treat as recoverable error
+        logger.error(f"Unexpected error in thread context fetch: {e}")
+        return ThreadContextResult.error(message.text)
 
-    # make sure we didn't get an empty response or a single message (not a thread)
+    # If empty response or single message (not a thread), return original text
     if len(messages) <= 1:
```
justin-tahara, line 700: Any change messages is empty?

backend/onyx/context/search/federated/slack_search.py
```diff
@@ -658,19 +682,44 @@ def get_contextualized_thread_text(
...
-        return message.text
+        return ThreadContextResult.success(message.text)
+
+    # Build thread text from thread starter + context window around matched message
+    thread_text = _build_thread_text(
+        messages, message_id, thread_id, access_token, team_id, slack_client
+    )
+    return ThreadContextResult.success(thread_text)
```
justin-tahara, line 707: Are we sure we are only doing this for the most relevant messages?

backend/onyx/context/search/federated/slack_search.py
```diff
@@ -747,6 +786,106 @@ def get_contextualized_thread_text(
...
+
+    Returns:
+        List of thread texts, one per input message.
+        Messages beyond max_messages or after rate limit get their original text.
+    """
+    if not slack_messages:
+        return []
+
+    # Limit how many messages we fetch thread context for (if max_messages is set)
+    if max_messages is not None and max_messages < len(slack_messages):
```
justin-tahara: Can we just do:
```suggestion
    if max_messages and max_messages < len(slack_messages):
```

backend/onyx/context/search/federated/slack_search.py
```diff
@@ -747,6 +786,106 @@ def get_contextualized_thread_text(
...
+    results: list[str] = []
+    rate_limited = False
+
+    # Process in batches
+    for i in range(0, len(messages_for_context), batch_size):
+        if rate_limited:
+            # Skip remaining batches, use original message text
+            remaining = messages_for_context[i:]
+            logger.warning(
+                f"Skipping {len(remaining)} thread context fetches due to rate limiting"
```
justin-tahara: I'm sure we will see this a ton in our logs

justin-tahara: Just something to note for sentry

backend/onyx/context/search/federated/slack_search.py
```diff
@@ -964,11 +1103,12 @@ def slack_retrieval(
...
-        [
-            (get_contextualized_thread_text, (slack_message, access_token, team_id))
-            for slack_message in slack_messages
-        ]
+    # Fetch thread context with rate limit handling and message limiting
+    # Messages are already sorted by relevance (slack_score), so top N get full context
+    thread_texts = fetch_thread_contexts_with_rate_limit_handling(
+        slack_messages=slack_messages,
+        access_token=access_token,
+        team_id=team_id,
```
justin-tahara, line 1120: Should we consider anything with workers or are we assuming that we just have 1 worker?

backend/onyx/federated_connectors/slack/models.py
```diff
@@ -47,10 +47,10 @@ class SlackEntities(BaseModel):
 
     # Message count per slack request
     max_messages_per_query: int = Field(
-        default=25,
+        default=10,
```
justin-tahara, line 50: I'm still a bit hesitant but we can try with this and lower it later if we are still throttling.

backend/onyx/configs/app_configs.py
nmgarza5: this sets the num of max workers we are running in batch. so not the number of sub queries

backend/onyx/context/search/federated/slack_search.py
nmgarza5: maybe. would you prefer to just not log this though? i feel like we should

backend/onyx/federated_connectors/slack/models.py
nmgarza5: we dont end up processing all of the messages anyways (getting context). all of these messages are still pulled from a single query. so i think this is fine for now

backend/onyx/context/search/federated/slack_search.py
nmgarza5: within `fetch_thread_contexts_with_rate_limit_handling` we set `max_workers = SLACK_THREAD_CONTEXT_BATCH_SIZE` which is 5

backend/onyx/context/search/federated/slack_search.py
nmgarza5: yes. we only process for messages_for_context

```
        messages_for_context = slack_messages[:max_messages]
        messages_without_context = slack_messages[max_messages:]
```

where `max_messages = MAX_SLACK_THREAD_CONTEXT_MESSAGES`

backend/onyx/context/search/federated/slack_search.py
nmgarza5: probably not but doesnt hurt to account for it

backend/onyx/context/search/federated/slack_search.py
nmgarza5: eh all other errors will still surface them. we just need explicit handling for the rate limited errors so as to not prevent other messages from being processed. I think we are good here for now unless we need to handle other errors more specifically -> [slack errors](https://docs.slack.dev/tools/node-slack-sdk/reference/web-api/enumerations/ErrorCode/)

backend/onyx/context/search/federated/slack_search.py
justin-tahara: I think it's fine for now and we update it later but maybe just add some info

justin-tahara (approved):
Let's do some testing on this before we promote this to stable. Definitely going to be in a better state than we are currently though

PR #6862, opened 2025-12-18
fix: small MCP UI changes

evan-onyx:
## Description

Addressing the following small UI problems:
- in tool selection popover, mcp tool toggles were not left justified due to the containing span (part of SimpleTooltip) having inline-flex
- could not create MCP servers for local development (removed erroring on user is None, as that is the local dev auth=disabled case)
- when configuring a per-user API key, the admin-configured header values would only appear after editing some text when re-entering the modal

## How Has This Been Tested?

tested in UI. I really want to get some per user API key tests in, but that will have to wait

## Additional Options

---
## Summary by cubic
Fixes small MCP UI issues and unblocks creating servers in local dev. Improves tool selection layout and makes per-user API key config reflect required fields immediately.

- **Bug Fixes**
  - Tool selection popover: remove inline-flex wrapper in SimpleTooltip so MCP tool toggles are left-aligned.
  - Local dev: allow creating MCP servers when auth is disabled (user is None) by using an empty owner_email.
  - Per-user API key modal: compute required_fields from header placeholders on load and on change; keep placeholder examples inline.

<sup>Written for commit 4fdb6c449f274320f1b18e620afd8ab8a4ae1d18. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/server/features/mcp/api.py
+++ b/backend/onyx/server/features/mcp/api.py
@@ -1730,12 +1730,12 @@ def get_mcp_server_db_tools(
 
 
 @admin_router.post("/servers/create", response_model=MCPServerCreateResponse)
-def upsert_mcp_server_with_tools(
+def upsert_mcp_server(
     request: MCPToolCreateRequest,
     db_session: Session = Depends(get_session),
     user: User | None = Depends(current_curator_or_admin_user),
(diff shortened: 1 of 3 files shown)


Subash-Mohan (approved):
lgtm!

PR #6913, opened 2025-12-20
fix(install): handle non-semver docker-compose versions

jmelahman:
## Description

For some odd reason, my docker compose version is `dev`...

Fixes https://github.com/onyx-dot-app/onyx/commit/6b39d8eed9#r173235322

## Additional Options

- [x] Override Linear Check

---
## Summary by cubic
Handle non-semver Docker Compose versions (e.g., "dev") in the install script so setup doesn’t fail on version parsing. This avoids incorrect “older than 2.24.0” warnings and allows installs to proceed for both plugin and standalone builds.

- **Bug Fixes**
  - Detect non-semver Compose versions and treat them as "dev" to continue installation.
  - Skip the 2.24.0 minimum-version check when the Compose version is "dev".

<sup>Written for commit 0853794f797ff2a8725275f40c9ede41d03e3751. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/deep_research/dr_loop.py
+++ b/backend/onyx/deep_research/dr_loop.py
@@ -444,6 +444,7 @@ def run_deep_research_llm_loop(
                 break
 
             research_results = run_research_agent_calls(
+                # The tool calls here contain the placement information
                 research_agent_calls=research_agent_calls,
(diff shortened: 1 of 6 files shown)


jmelahman:
@greptile

PR #6924, opened 2025-12-20
fix(fe): make recent chat sidebar buttons links

jmelahman:
## Description

These should be links so they support right-click open new + a11y

## Additional Options

- [x] Override Linear Check

---
## Summary by cubic
Made recent chat sidebar items real links, so users can right-click to open in a new tab and screen readers treat them as navigation. Replaced onClick routing with href to /chat?chatId={chatSessionId}.

<sup>Written for commit 999fcd35c2bb910ab37f93b2acb7323a33dd7330. Summary will update automatically on new commits.</sup>

### DIFF
--- a/web/src/sections/sidebar/ChatButton.tsx
+++ b/web/src/sections/sidebar/ChatButton.tsx
@@ -436,7 +436,7 @@ const ChatButton = memo(
       >
         <PopoverAnchor>
           <SidebarTab
-            onClick={() => route({ chatSessionId: chatSession.id })}
+            href={`/chat?chatId=${chatSession.id}`}
             active={active}
             rightChildren={rightMenu}
             focused={renaming}


web/src/sections/sidebar/ChatButton.tsx
```diff
@@ -436,7 +436,7 @@ const ChatButton = memo(
       >
         <PopoverAnchor>
           <SidebarTab
-            onClick={() => route({ chatSessionId: chatSession.id })}
+            href={`/chat?chatId=${chatSession.id}`}
```
jmelahman, line 439: @greptile is this the most idiomatic way to do this?

jmelahman: @greptile is the route object more type-safe and thus could be preferred?

PR #6936, opened 2025-12-21
feat: deep research

rohoswagger:
## Description

deep research!

Open Issues that will be addressed in future prs:
- Setup Posthog for FE feature flag
- Context window exceeded error
- parallel tool tabs having some funky open/close functionality (will align w duo here)
- other nits that come up
- use new FE standards (will align w ronnie here)

## How Has This Been Tested?

lots and lots and lots of local testing
https://github.com/user-attachments/assets/7260f150-35c2-4319-b7a0-812114062a64

## Additional Options

---
## Summary by cubic
Adds the deep research experience with streaming plan and research agent UIs, plus backend packet generation to power them. Enables the deep research toggle in the chat bar and instruments the backend loop with tracing.

- **New Features**
  - Frontend renderers for Deep Research Plan and Research Agent with animated, collapsible content.
  - Shared hook to animate packet display and auto-collapse on completion.
  - Early renderer selection for deep research packets; AIMessage shows content when plan/agent starts.
  - New packet models and unions for plan, intermediate reports, reasoning_done, citation info, error packets, and top-level branching; tool utils and icons updated.
  - MultiToolRenderer supports deep research, nested tools, error states, and parallel branch timing; status can render JSX.
  - Backend citation utilities and processor: update mappings from citeable tools, collapse/renumber citations, and optionally replace markers with links; ResearchAgent emits start/intermediate/report/section packets and cited docs.
  - Deep Research toggle enabled in ChatInputBar; request param renamed useAgentSearch → deepResearch (payload: deep_research).
  - Trace instrumentation for run_deep_research_llm_loop with tenant_id metadata.
  - Validate LLMs have ≥50k max input tokens before running deep research.

- **Bug Fixes**
  - More robust streaming: close dangling reasoning, correct SECTION_END counting, treat ERROR as completion, handle PacketException, and skip failed research agent calls.
  - Treat content before tool calls as reasoning when tool_choice=REQUIRED to avoid mixing answers with thinking.
  - Preserve reasoning on non-reasoning models by converting think_tool to reasoning and unescaping JSON escape sequences for proper newline/tab rendering; bump urllib3 to 2.6.2.
  - Fix chat content links to use proper href targets instead of mouse handlers.
  - Desktop: Linux-specific build fixes and config tweaks (version set to 0.0.0-dev, window background color, signing identity).
  - More correct packet handling and UI: show all tool statuses during streaming, detect completion via sub_turn_index, mark “Completed with errors,” and set stop packet turn_index correctly; tests updated.

<sup>Written for commit 99c4ddd53b1e23ef2cba735801e5dfb7bf5d6480. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/deep_research/dr_loop.py
+++ b/backend/onyx/deep_research/dr_loop.py
@@ -314,10 +314,7 @@ def run_deep_research_llm_loop(
                     Packet(
                         # Marks the last turn end which should be the plan generation
                         placement=Placement(
-                            turn_index=packet.placement.turn_index
-                            + (1 if reasoned else 0),
-                            tab_index=packet.placement.tab_index,
-                            sub_turn_index=packet.placement.sub_turn_index,
+                            turn_index=1 if reasoned else 0,
                         ),
                         obj=SectionEnd(),
                     )


backend/onyx/chat/llm_step.py
```diff
@@ -542,72 +548,102 @@ def run_llm_step_pkt_generator(
...
-                                    tab_index=tab_index,
-                                    sub_turn_index=sub_turn_index,
-                                ),
-                                obj=result,
-                            )
-                else:
-                    # When citation_processor is None, use delta.content directly without modification
-                    accumulated_answer += delta.content
-                    # Save answer incrementally to state container
+                # When tool_choice is REQUIRED, content before tool calls is reasoning/thinking
```
rohoswagger, line 552: @yuhongsun96 see this

yuhongsun96: I don't think we need this logic because the frontend needs to handle the case where it outputs some normal tokens and then calls a tool anyway. Like just in the normal flow it might do this. So I would rather that be a consistent experience for those models (the Claude ones basically).

backend/onyx/deep_research/dr_loop.py
```diff
@@ -143,6 +149,8 @@ def generate_final_report(
         # but we'd still want to capture the reasoning from the think_tool of theprevious turn.
         state_container.set_reasoning_tokens(saved_reasoning)
 
+    return has_reasoned
```
yuhongsun96: good catch!

backend/onyx/tools/fake_tools/research_agent.py
```diff
@@ -127,26 +129,32 @@ def generate_intermediate_report(
             if isinstance(packet.obj, AgentResponseStart):
                 emitter.emit(
                     Packet(
-                        placement=packet.placement,
+                        placement=placement,  # Use original placement
```
yuhongsun96: Why is this the case? We just want to change the packet type for clarity. The reasoning may cause an offset of 1 here also. We want to allow for reasoning to be a separate substep here right?

yuhongsun96: I think there is a separate issue here of needing to pass in the tab but I can add that

backend/onyx/tools/fake_tools/research_agent.py
```diff
@@ -507,7 +512,15 @@ def run_research_agent_call(
         emitter.emit(
             Packet(
                 placement=Placement(turn_index=turn_index, tab_index=tab_index),
-                obj=PacketException(type="error", exception=e),
+                obj=PacketException(type=StreamingType.ERROR.value, exception=e),
+            )
+        )
+        # Emit SectionEnd so the frontend knows this research agent is complete (even if it failed)
+        # Without this, the frontend will hang waiting for the SectionEnd packet
+        emitter.emit(
```
yuhongsun96: I think the exception is supposed to imply end so we don't need this I feel

rohoswagger:
@greptile

rohoswagger:
@cubic-dev-ai review this PR

PR #6945, opened 2025-12-21
perf(chat): avoid re-rendering chat on ChatInput change

jmelahman:
## Description

We're currently re-rendering all of the ChatUI (which is very expensive) when the ChatInput changes. This causes the FPS to drop to ~10 frames/s when typing new messages.

**before**
20251221_13h33m34s_grim

**after**
20251221_14h59m08s_grim

## How Has This Been Tested?

:grimacing: 

## Additional Options

- [x] Override Linear Check

---
## Summary by cubic
Stopped re-rendering the entire Chat UI on every ChatInput keystroke by making the input manage its own state and exposing only minimal callbacks. Typing is now smooth and no longer drops FPS to ~10.

- **Refactors**
  - Made ChatInputBar self-contained: replaced message/setMessage props with internal state and a new initialMessage prop.
  - Changed onSubmit to accept the message string and added onHeightChange for scroll adjustments.
  - Added ChatInputBarHandle (reset, focus) via forwardRef; replaced textAreaRef usage in useChatSessionController.
  - Used a ResizeObserver to detect input height changes without triggering ChatUI re-renders.
  - Memoized ChatUI and stabilized prop identities; exposed scrollBy via the ChatUI ref for smoother auto-scroll behavior.

- **Bug Fixes**
  - Fixed auto-scroll when switching chat sessions by resetting scroll tracking and scrolling to bottom on load.

<sup>Written for commit 4f8ac5731aef7b54c3660b3e6632cf5d31f9f810. Summary will update automatically on new commits.</sup>

### DIFF
--- a/web/src/sections/ChatUI.tsx
+++ b/web/src/sections/ChatUI.tsx
@@ -100,6 +100,7 @@ const ChatUI = React.memo(
       const endDivRef = useRef<HTMLDivElement>(null);
       const scrollDist = useRef<number>(0);
       const scrolledForSession = useRef<string | null>(null);
+      const prevMessageCount = useRef<number>(0);
       const [aboveHorizon, setAboveHorizon] = useState(false);
       const debounceNumber = 100;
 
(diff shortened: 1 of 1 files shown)


web/src/sections/ChatUI.tsx
raunakab, line 1: This file's diffs are a bit difficult to parse through. Was the only change here encompassing the `ChatUI` component with a `React.memo`?

jmelahman: I would strongly recommend ignoring whitespace, https://onyx-company.slack.com/archives/C0771QKDBPE/p1763056759979369. These components desperately need refactoring and there were notable changes to `useEffect`

PR #6969, opened 2025-12-22
fix(docprocessing): Move to Sync

justin-tahara:
## Description

The previous event pooling change seems to be working now! 

There still seems to be an issue with the way we are batching and handling issues with the Google GenAI client which I hope to fix with this PR.

## How Has This Been Tested?

Still need to look into this but it should be a better setup relying on native google api calls.

## Additional Options

---
## Summary by cubic
Switch Google Vertex embeddings from async to sync to stop memory leaks from connection pooling during doc processing. Adds safer HTTP settings and proper client cleanup for more reliable runs.

- **Bug Fixes**
  - Replaced async embedding with `_embed_vertex_sync` using the Google GenAI synchronous API.
  - Configured httpx client limits, 60s timeout, and reduced retries (3 attempts) to prevent resource buildup.
  - Kept batching; now processes batches synchronously and collects embeddings safely.
  - Updated `embed()` to run the sync call in a thread, route EmbeddingProvider.GOOGLE to the sync path, and ensure clients close cleanly.

<sup>Written for commit 6ac79cf6e14ccabfd1bdcf714b75920042d2c4ef. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/natural_language_processing/search_nlp_models.py
+++ b/backend/onyx/natural_language_processing/search_nlp_models.py
@@ -388,12 +388,16 @@ class CloudEmbedding:
 
             return all_embeddings
         finally:
-            # Close the synchronous client - no async/await needed
+            # Close both clients independently to ensure cleanup even if one fails
             try:
                 client.close()
+            except Exception as e:
+                logger.warning(f"Error closing Google GenAI client: {e}")
+
+            try:
                 sync_client.close()
             except Exception as e:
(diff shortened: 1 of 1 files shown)


justin-tahara:
@greptileai & @cubic-dev-ai can you both review once more? 

The idea is that we want to move away from asyncio workflows as this causes a lot of issues within the code that is harder to control

PR #6987, opened 2025-12-23
refactor: add more helpful utility hooks

raunakab:
## Description

- Add useMcpServers hook for fetching MCP server configuration
- Add useOpenApiTools hook for fetching OpenAPI tools configuration
- Update MCPPageContent to use useMcpServers hook
- Update OpenApiPageContent to use useOpenApiTools hook
- Both hooks poll every 10 seconds for fresh data

## Notes

Since this PR just shifts things around, no screenshots / tests are necessary.

---
## Summary by cubic
Refactored data fetching by adding reusable hooks for MCP servers and OpenAPI tools, and updated their pages to use them. Also added a full-detail agent hook and a content size hook, and fixed app focus detection for agents routes.

- **New Features**
  - Added useAgent for fetching a single agent with full details.
  - Added useContentSize for measuring element dimensions.

- **Refactors**
  - Added useMcpServers and useOpenApiTools hooks and removed 10s polling.
  - Updated MCPPageContent and OpenApiPageContent to use the new hooks.
  - Updated useAppFocus to handle /chat/agents subpaths.

<sup>Written for commit 74db388cc450fd5e3fcf7db3a90a22522dcaa1cc. Summary will update automatically on new commits.</sup>

### DIFF
--- a/web/src/hooks/useMcpServers.ts
+++ b/web/src/hooks/useMcpServers.ts
@@ -8,8 +8,7 @@ import useSWR from "swr";
  * Fetches MCP (Model Context Protocol) servers configuration.
  *
  * MCP servers provide additional tools and capabilities to agents through
- * the Model Context Protocol. This hook polls the MCP servers endpoint
- * every 10 seconds to keep the data fresh.
+ * the Model Context Protocol.
  *
  * @returns Object containing:
  *   - mcpData: MCPServersResponse data or null if not loaded
@@ -30,8 +29,7 @@ export default function useMcpServers() {
     mutate: mutateMcpServers,
   } = useSWR<MCPServersResponse>(
     "/api/admin/mcp/servers",
-    errorHandlingFetcher,
-    { refreshInterval: 10000 }
+    errorHandlingFetcher
   );
 
   return {
--- a/web/src/hooks/useOpenApiTools.ts
+++ b/web/src/hooks/useOpenApiTools.ts
@@ -8,8 +8,7 @@ import useSWR from "swr";
  * Fetches OpenAPI tools configuration.
  *
  * OpenAPI tools provide custom actions and integrations to agents through
- * OpenAPI specifications. This hook polls the OpenAPI tools endpoint
- * every 10 seconds to keep the data fresh.
+ * OpenAPI specifications.
  *
  * @returns Object containing:
  *   - openApiTools: ToolSnapshot[] data or null if not loaded
@@ -28,9 +27,7 @@ export default function useOpenApiTools() {
     error,
     isLoading: isOpenApiLoading,
     mutate: mutateOpenApiTools,
-  } = useSWR<ToolSnapshot[]>("/api/tool/openapi", errorHandlingFetcher, {
-    refreshInterval: 10000,
-  });
+  } = useSWR<ToolSnapshot[]>("/api/tool/openapi", errorHandlingFetcher);
 
   return {
     openApiTools: openApiTools ?? null,


web/src/hooks/useContentSize.ts
```diff
@@ -0,0 +1,57 @@
+import { useRef, useEffect, useState } from "react";
+
+interface ContentSize {
+  width: number;
+  height: number;
+}
+
+export function useContentSize(
```
nmgarza5, line 92: can you add a docstring for this one as well? I'm a big fan

web/src/hooks/useContentSize.ts
```diff
@@ -0,0 +1,57 @@
...
+      };
+      setSize(newSize);
+    }
+  };
+
+  // Measure on dependencies change
+  useEffect(() => {
+    measureSize();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, dependencies);
```
nmgarza5, line 118: mind adding a brief comment on why we are ignoring this eslint rule? My guess is you added this because the dependencies list is provided by the consumer of the hook so we cant define as a static array.

web/src/hooks/useMcpServers.ts
```diff
@@ -0,0 +1,43 @@
...
+export default function useMcpServers() {
+  const {
+    data: mcpData,
+    error,
+    isLoading: isMcpLoading,
+    mutate: mutateMcpServers,
+  } = useSWR<MCPServersResponse>(
+    "/api/admin/mcp/servers",
+    errorHandlingFetcher,
+    { refreshInterval: 10000 }
```
nmgarza5: 10 seconds kinda seems like overkill. Are MCP configs really changing that frequently?

web/src/hooks/useOpenApiTools.ts
```diff
@@ -0,0 +1,41 @@
...
+ * return <OpenApiToolsList tools={openApiTools} />;
+ */
+export default function useOpenApiTools() {
+  const {
+    data: openApiTools,
+    error,
+    isLoading: isOpenApiLoading,
+    mutate: mutateOpenApiTools,
+  } = useSWR<ToolSnapshot[]>("/api/tool/openapi", errorHandlingFetcher, {
+    refreshInterval: 10000,
```
nmgarza5: same comment as above

web/src/hooks/useMcpServers.ts
nmgarza5: I think this MCP config is defined by an admin anyways. I think we should removing polling and call `mutate()` after any config changes. As is we are just spamming traffic to the backend

web/src/hooks/useContentSize.ts
raunakab: Added!

web/src/hooks/useMcpServers.ts
raunakab: I agree. I'll remove that. This was pretty much a pure extraction from @Subash-Mohan's changes, so I want to ping him for visibility here.

web/src/hooks/useContentSize.ts
raunakab: Added!

web/src/hooks/useMcpServers.ts
raunakab: Removed.

web/src/hooks/useOpenApiTools.ts
raunakab: Removed.

PR #6998, opened 2025-12-23
chore: cleanup drupal connector nits

wenxi-onyx:
- **cleanup drupals headers processing**
- **duplicate var**
- **erroneous __init__.py**
- **remove max page limit**
- **change some log levels**
- **use rstrip**

## Description

## How Has This Been Tested?

## Additional Options

---
## Summary by cubic
Cleaned up the Drupal Wiki connector’s header handling, logging, and URL processing to reduce noise and fix edge cases. Also removed the artificial page cap so we fetch all available pages.

- **Bug Fixes**
  - Initialize headers with Accept and set Authorization from a trimmed API token; correct headers for binary downloads.
  - Normalize base_url with rstrip to prevent double slashes in document IDs.
  - Fix attachment type detection using get_file_ext and warn on unsupported types.
  - Remove erroneous backend/__init__.py.

- **Refactors**
  - Simplify space scope handling and switch most info logs to debug; add clearer warnings/errors.
  - Remove the max page limit when listing pages.
  - Drop duplicate/unused variables and rely on OnyxFileExtensions directly.

<sup>Written for commit 3869deb01dca7b3a866c80885a748b1032dce7e6. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/connectors/drupal_wiki/utils.py
+++ b/backend/onyx/connectors/drupal_wiki/utils.py
@@ -1,6 +1,3 @@
-from datetime import datetime
-from datetime import timezone
-
 from onyx.utils.logger import setup_logger
 
 logger = setup_logger()
@@ -11,9 +8,3 @@ def build_drupal_wiki_document_id(base_url: str, page_id: int) -> str:
     # Ensure base_url ends with a slash
     base_url = base_url.rstrip("/") + "/"
     return f"{base_url}node/{page_id}"
-
-
-def datetime_from_timestamp(timestamp: int) -> datetime:
-    """Convert a Unix timestamp to a datetime object in UTC"""
-
-    return datetime.fromtimestamp(timestamp, tz=timezone.utc)


backend/onyx/connectors/drupal_wiki/connector.py
```diff
@@ -535,18 +515,14 @@ def _process_page(self, page: DrupalWikiPage) -> Document | ConnectorFailure:
                     )
                     if error:
                         logger.warning(
-                            "Error processing attachment %s: %s",
-                            attachment.get("name", "Unknown"),
-                            error,
+                            f"Error processing attachment {attachment.get('name', 'Unknown')}: {error}"
```
jmelahman, line 518: I think the previous behavior was preferred (though I don't feel strongly). iiuc, f-strings in logging statements may have runtime side-effects even when the logger level is above the given statement's level.

See also, https://docs.astral.sh/ruff/rules/logging-f-string/

wenxi-onyx: Just doing this for consistency because I'm pretty sure we do this everywhere else in the codebase. If we want to swap over, we can one-shot convert the entire codebase

jmelahman: I think that just means we have a lot of potential bugs :)

PR #7007, opened 2025-12-24
fix(fe): right sidebar buttons dont inherit href

jmelahman:
## Description

Fixes an issue with sidebar items when the item has an `href` and there is a right button. By moving the right button out of the Link, it no longer inherits the `href` causing https://onyx-company.slack.com/archives/C0832RVRVG8/p1766433472201599.

**hover right button, no hyperlink behavior**
20251223_16h37m19s_grim

**hover content, hyperlink behavior**
20251223_16h37m32s_grim

## Additional Options

- [x] Override Linear Check

---
## Summary by cubic
Fix sidebar tab links so right-side action buttons don’t inherit the href, and make recent chat items navigate via link. This prevents accidental navigation from right menu clicks and restores consistent chat navigation.

- **Bug Fixes**
  - SidebarTab: use an absolute Link overlay and disable pointer events on main content unless focused; keep rightChildren outside the overlay so right buttons don’t trigger navigation.
  - ChatButton: pass href (/chat?chatId={id}) to SidebarTab instead of onClick routing.

<sup>Written for commit 48b3135c1f5d0046ec1b8665ec1c19a823042b8b. Summary will update automatically on new commits.</sup>


jmelahman:
> Can we replace the existing `content` with the following instead?

Discussed offline, but I think there are semantic and practical reasons why positioning the `Link` absolutely is preferred in this case. I think the only downside of this approach is that `Link` requires `tabIndex=-1`, but I think that's about it.

jmelahman:
I was going to try and improve the tabindexing, but given there are changes that depend on this & given that will likely increase the scope of this change, I'll just follow up with that change after dinner.

PR #7051, opened 2025-12-26
fix(chat): Custom Agent Chat Rename

justin-tahara:
## Description

Currently when you try and chat with a custom agent, it never renames the chat. This is to fix this workflow 

## How Has This Been Tested?

Tested locally with multiple different agents including the default one and custom ones.

## Additional Options

---
## Summary by cubic
Fixes chat auto-naming for custom agents. New chats are named once after the first assistant reply when no explicit title is provided, preventing unnamed sessions.

- **Bug Fixes**
  - Auto-name only when there’s no ?title, no prior user messages, and the session has no description.
  - Rename trigger now runs when history has 2+ messages (was exactly 2), covering custom agent flows.

<sup>Written for commit e6b0c3defedae014e4e8b60b10ee20a9497a1702. Summary will update automatically on new commits.</sup>

### DIFF
--- a/web/src/app/chat/hooks/useChatController.ts
+++ b/web/src/app/chat/hooks/useChatController.ts
@@ -483,6 +483,11 @@ export function useChatController({
 
       const searchParamBasedChatSessionName =
         searchParams?.get(SEARCH_PARAM_NAMES.TITLE) || null;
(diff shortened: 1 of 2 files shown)


jmelahman (approved):
Is it worth adding tests for this?

PR #6464, opened 2025-11-26
feat: move to client side rendering + incremental loading

Weves:
Add reusable hooks for decentralized data fetching as part of the client-side rendering refactor. These hooks enable components to fetch their own data directly using SWR with automatic deduplication.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Description

[Provide a brief description of the changes in this PR]

## How Has This Been Tested?

[Describe the tests you ran to verify your changes]

## Additional Options

- [x] [Optional] Override Linear Check

---
## Summary by cubic
Shifted to client-side data fetching with reusable SWR hooks and removed legacy contexts/SSR. Updated ChatPage, ChatInputBar, AppSidebar, NRFPage, and layouts for smoother loading and onboarding.

- **New Features**
  - Added hooks: useAgents, usePinnedAgents, usePinnedAgentsWithDetails, useCurrentAgent, useIsDefaultAgent, useForcedTools, useAvailableTools, useChatSessions, useDocumentSets, useInputPrompts, useProjects, useTags, useAssistantPreferences.
  - Unified return shape: data, isLoading, error, and mutate helpers (refresh, refreshChatSessions, refreshProjects).
  - SWR config: revalidateOnFocus disabled; deduping 60s (most) and 30s (chat sessions, projects).
  - Pinning helpers: pinAgents to persist pinned assistant order.
  - UI polish: stable input/textarea behavior; defer controls/sidebar until data loads; welcome/assistant name test IDs; onboarding waits for provider loading.

- **Refactors**
  - Removed ChatContext, AgentsContext, and fetchChatData; layouts now use requireAuth and ProjectsProvider (no SSR data fetching).
  - Replaced AgentsProvider in AppProvider; Root/Admin/EE pages now client-only for data.
  - AppSidebar now syncs pinned assistants reorder to the server (optimistic).
  - ProjectsProvider switched to useProjects; refresh paths updated across components (e.g., ChatButton, ChatSessionMorePopup).
  - Assistant preferences moved to SWR; forced tools moved to a small Zustand store; LlmManager pulls providers via useLLMProviders.

<sup>Written for commit 4df9b976872f641a88892b8cdb81736f2bb013b0. Summary will update automatically on new commits.</sup>

PR #7054, opened 2025-12-26
feat(connectors): Add Deletion Popup

justin-tahara:
## Description

Currently when you go to delete a connector, there is no warning or message explaining what it is that you are doing. 

This PR adds a new workflow to show a popup when someone is trying to delete a connector. 

## How Has This Been Tested?

Tested this locally with the popup but also confirmed that the connector would delete after clicking on the final delete button.
Screenshot 2025-12-26 at 12 51 09 PM

## Additional Options

---
## Summary by cubic
Add a confirmation modal before deleting connectors to prevent accidental removals. The modal explains that deletion schedules a job to remove indexed documents and deletes the connector for all users.

- **New Features**
  - Show a confirm dialog when deleting a connector from the connector details page and the Failed Re-indexing Attempts list.
  - Confirmation is shown for admins and global curators; deletion is scheduled only after explicit confirmation, with duplicate-click protection.

- **Bug Fixes**
  - Prevent chat welcome message hydration flicker by rendering a deterministic greeting first, then randomizing on mount.
  - Fix delete button disabled state in Failed Re-indexing Attempts so deletable connectors can be removed.

<sup>Written for commit fda5e4b5876ac6111782c5e24a6a7b0c8f70509c. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/app/chat/components/WelcomeMessage.tsx
+++ b/web/src/app/chat/components/WelcomeMessage.tsx
@@ -8,7 +8,8 @@ import {
 import AgentAvatar from "@/refresh-components/avatars/AgentAvatar";
 import Text from "@/refresh-components/texts/Text";
 import { MinimalPersonaSnapshot } from "@/app/admin/assistants/interfaces";
-import { useEffect, useState } from "react";
+import useOnMount from "@/hooks/useOnMount";
+import { useState } from "react";
 
 export interface WelcomeMessageProps {
   agent?: MinimalPersonaSnapshot;
@@ -21,9 +22,7 @@ export default function WelcomeMessage({
 }: WelcomeMessageProps) {
   const [greeting, setGreeting] = useState(() => GREETING_MESSAGES[0] ?? "");
 
-  useEffect(() => {
-    setGreeting(getRandomGreeting());
-  }, []);
+  useOnMount(() => setGreeting(getRandomGreeting()));
 
   let content: React.ReactNode = null;
 


web/src/app/chat/components/WelcomeMessage.tsx
```diff
@@ -16,10 +19,15 @@ export default function WelcomeMessage({
   agent,
   isDefaultAgent,
 }: WelcomeMessageProps) {
+  const [greeting, setGreeting] = useState(() => GREETING_MESSAGES[0] ?? "");
+
+  useEffect(() => {
+    setGreeting(getRandomGreeting());
+  }, []);
+
```
raunakab: Would prefer to use `useOnMount` here!

raunakab:
```tsx
import useOnMount from "@/hooks";

useOnMount(() => setGreeting(getRandomGreeting()));
```

web/src/app/chat/components/WelcomeMessage.tsx
```diff
@@ -16,10 +19,15 @@ export default function WelcomeMessage({
   agent,
   isDefaultAgent,
 }: WelcomeMessageProps) {
+  const [greeting, setGreeting] = useState(() => GREETING_MESSAGES[0] ?? "");
```
raunakab, line 23: Nice! You're extracting the hook out into an unconditional pathway.

I'm curious as to why this wasn't done before. If the `git blame` points to me, it's probably because I move the file and formatted it a bit.

justin-tahara: Pretty sure it's this: https://github.com/onyx-dot-app/onyx/pull/6464

PR #7057, opened 2025-12-26
fix: disable responses api/reasoning param for gpt chat models

wenxi-onyx:
## Description

- OpenAI API does not accept reasoning.effort for gpt chat models like gpt-5-chat-latest
- Litellm has these chat models listed as reasoning models
- OpenAI documentation implies they are reasoning models

Additionally
- We cannot designate them as non-reasoning models in Onyx because 1) our code passes self.temperature for non-reasoning models 2) these chat models will only accept temperature 1

## How Has This Been Tested?

Tested gpt-5, 5.1, and 5.2 chat models locally

## Additional Options
Fixes https://linear.app/danswer/issue/DAN-3245/gpt-5-chat-latest

### DIFF
--- a/backend/onyx/llm/multi_llm.py
+++ b/backend/onyx/llm/multi_llm.py
@@ -248,10 +248,14 @@ class LitellmLLM(LLM):
         )
 
         # Needed to get reasoning tokens from the model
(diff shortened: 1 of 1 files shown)


wenxi-onyx:
Closed in favor of #7065

PR #7058, opened 2025-12-26
chore: removed unnecessary monkey patch

wenxi-onyx:
## Description

This bug was fixed by litellm in our current version.

## How Has This Been Tested?

Tested openai models locally

## Additional Options

---
## Summary by cubic
Removed the obsolete monkey patch that forced LiteLLM Responses tool messages to use "input_text". This deletes _patch_openai_responses_tool_content_type and its call in apply_monkey_patches, relying on upstream behavior and simplifying maintenance.

<sup>Written for commit d8c04216c90197be86a9e55731f00d69b1c2e5db. Summary will update automatically on new commits.</sup>

### DIFF
--- a/backend/onyx/llm/litellm_singleton/monkey_patches.py
+++ b/backend/onyx/llm/litellm_singleton/monkey_patches.py
@@ -631,56 +631,6 @@ def _patch_openai_responses_transform_response() -> None:
     LiteLLMResponsesTransformationHandler.transform_response = _patched_transform_response  # type: ignore[method-assign]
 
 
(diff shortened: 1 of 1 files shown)


wenxi-onyx:
@justin-tahara 
> Should we add some sort of test so that we can catch this if anything changes?

No, this was an extremely niche, specific bug that broke many openai models through litellm. What we actually need is 
1) Litellm implements CI test for every major model and provider
2) We implement similar CI test that just checks every major model and provider are configurable and useable (Chris is already working on this on the side)

PR #7096, opened 2025-12-29
chore(fe): make Text component default to span

jmelahman:
## Description

Updates the `<Text>` component to default to an HTML `<span>`. I spot-checked a handful of current usages and all of them:
- Are semantically incorrect as `<p>`'s
- All have (would be unnecessary) flex container parents which are providing an inline appearance

See also, https://onyx-company.slack.com/archives/C09BHKH5PML/p1767037628936109

## How Has This Been Tested?

Expecting no change in behavior here.

Expecting the type-checker to work here -- made the `as` prop required and ensured all instances had an explicit Tag specified. The legacy `components/ui/text.tsx` should also fail the type-checker if unexpectedly updated.

## Additional Options

- [x] Override Linear Check

---
## Summary by cubic
Set the Text component to default to an inline span instead of p, improving semantic correctness and avoiding unintended block-level layout. Also narrowed the as prop to only p and span.

- **Refactors**
  - Default element is now span when as is not set.
  - Removed div from TextProps.as; only p and span are allowed.
  - No visual or behavioral changes expected.

<sup>Written for commit ef0ef44414c496d40de29234a3277df4ae96abab. Summary will update automatically on new commits.</sup>

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -10298,6 +10298,7 @@
     },
     "node_modules/fsevents": {
       "version": "2.3.2",
+      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [


web/src/refresh-components/Card.tsx
```diff
@@ -29,8 +29,8 @@
  *
  * // Multiple children - automatically spaced
  * <Card>
- *   <Text headingH3>Section 1</Text>
- *   <Text body>Some content</Text>
+ *   <Text as="p" headingH3>Section 1</Text>
+ *   <Text as="p" body>Some content</Text>
```
jmelahman, line 33: I think this is correct/desired

web/src/refresh-components/texts/Text.tsx
```diff
@@ -93,7 +93,7 @@ export interface TextProps
   textDark05?: boolean;
 
   // Tag type override
-  as?: "p" | "span" | "div";
```
jmelahman, line 96: FYI, don't feel too strongly, but it seems weird to want to ever use `div` here

PR #7101, opened 2025-12-30
feat: save reasoning details and append to assistant messages

wenxi-onyx:
## Description

- Certain models like Gemini 3.0 pro require that reasoning steps, especially for tool calls, should be appended to the assistant message and sent back to the LLM.
- Litellm automatically handles this for Gemini and Claude models by appending reasoning_details and removing reasoning altogether, respectively
- Litellm does not handle this for Gemini via OpenRouter --> it just blows up

- tested flagship model for every provider

## Additional Options

---
## Summary by cubic
Adds support for capturing provider-specific reasoning details from streaming (thinking_blocks/reasoning_details), echoing them back with grouped parallel tool calls in a single assistant message, and budgeting their tokens so Anthropic and Gemini can use verified reasoning reliably.

- **New Features**
  - Capture extra_reasoning_details in-memory on ChatMessageSimple and LlmStepResult; parse and accumulate from deltas with provider keys preserved.
  - Flatten extra_reasoning_details into AssistantMessage top-level fields; group parallel tool calls into one assistant message and order calls before responses.
  - Count extra_reasoning_tokens excluding signature/data and add them to the token budget, attributed to the first tool call.
  - Keep extra_reasoning_details only for the current turn (not persisted).

<sup>Written for commit cface0eb2aae74af3007b7caadff0d38e077de8a. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/llm/litellm_singleton/config.py
+++ b/backend/onyx/llm/litellm_singleton/config.py
@@ -13,7 +13,13 @@ def configure_litellm_settings() -> None:
     # parameters like frequency and presence, just ignore them
     litellm.drop_params = True
     litellm.telemetry = False
-    litellm.modify_params = True
+    # whuang: Claude 4-5 sonnet DR is emitting tool calls with NO thinking blocks like
+    # tool call (with thinking block) --> we send tool response with thinking block -->
+    # --> 2nd tool call (no thinking block) --> 2nd tool response no thinking block
+    # On our tool response message, litellm.modify_params=True drops the thinking param (forcing thinking=None)
+    # Anthropic then says 'thinking=None we can't accept thought signatures from your first tool call'
+    # Disabling this setting seems to completely fix this issue and I'm not yet sure why
+    # litellm.modify_params = False
     litellm.add_function_to_prompt = False
 
 


backend/onyx/chat/chat_state.py
```diff
@@ -69,6 +75,11 @@ def get_answer_tokens(self) -> str | None:
         with self._lock:
             return self.answer_tokens
 
+    def get_reasoning_details(self) -> list[dict] | None:
+        """Thread-safe getter for reasoning_details."""
+        with self._lock:
+            return self.reasoning_details
```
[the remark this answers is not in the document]
wenxi-onyx: addressed

backend/onyx/db/chat.py
```diff
@@ -661,6 +661,7 @@ def create_new_chat_message(
     commit: bool = True,
     reserved_message_id: int | None = None,
     reasoning_tokens: str | None = None,
+    reasoning_details: list[dict] | None = None,
```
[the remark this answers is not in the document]
wenxi-onyx: addressed

backend/onyx/chat/llm_step.py
```diff
@@ -343,6 +343,7 @@ def translate_history_to_llm_format(
                 role="assistant",
                 content=msg.message or None,
                 tool_calls=None,
+                reasoning_details=msg.reasoning_details,
             )
```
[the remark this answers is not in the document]
wenxi-onyx: i believe this is not an issue based on testing

wenxi-onyx:

backend/onyx/chat/chat_utils.py
```diff
@@ -683,13 +689,17 @@ def convert_chat_history(
                             )
                         )
 
+                    simple_messages.extend(tool_call_messages)
```
yuhongsun96: Why are we stacking them like this now?

backend/onyx/chat/README.md
```diff
@@ -76,6 +76,21 @@ Instead of dropping the Tool Call response, we might summarize it using an LLM s
 this is questionable value add because anything relevant and useful should be already captured in the Agent response.
 
 
+## Extended Thinking / Reasoning Verification
```
yuhongsun96: I would move this down to the lower section on the architecture stuff

backend/onyx/chat/chat_utils.py
wenxi-onyx: This is actually how the models expect parallel tool calls if you check their docs/cookbooks. A single assistant message with a list of tool calls.

Gemini explodes if
- You attempt to save and return thought signature
- Parallel tool calls are split into
-- Tool call 1 --> Tool response 1 --> Tool call 2 --> Tool response 2

wenxi-onyx:
@cubic-dev-ai @greptile review this

backend/onyx/chat/chat_utils.py
```diff
@@ -692,7 +692,12 @@ def convert_chat_history(
                     # Sort by tool_id within the turn for consistent ordering
                     turn_tool_calls.sort(key=lambda tc: tc.tool_id)
 
-                    # Add each tool call as a separate message with the tool arguments
+                    # Parallel tool calls are grouped together (sequentially in the history)
```
yuhongsun96: I can't make sense of this comment

backend/onyx/chat/llm_loop.py
```diff
@@ -518,6 +518,9 @@ def run_llm_loop(
                 simple_chat_history.extend(failure_messages)
```
yuhongsun96, line 518: Do we need to address the deep research loop?

backend/onyx/chat/llm_loop.py
```diff
@@ -630,6 +643,12 @@ def run_llm_loop(
         if not llm_step_result or not llm_step_result.answer:
             raise RuntimeError("LLM did not return an answer.")
 
+        # Set extra_reasoning_details on state_container for the final answer
+        if llm_step_result.extra_reasoning_details:
```
yuhongsun96: Is the state_container's extra reasoning details ever used?

backend/onyx/chat/chat_utils.py
yuhongsun96: This is not making sense to me, seems this is adding all the tool calls then all the tool responses.

backend/onyx/chat/llm_loop.py
```diff
@@ -575,6 +579,7 @@ def run_llm_loop(
                     tool_call_response=tool_response.llm_facing_response,
                     search_docs=search_docs,
                     generated_images=generated_images,
+                    extra_reasoning_details=llm_step_result.extra_reasoning_details,
```
yuhongsun96: ToolCallInfo is to get saved into the DB no? Seems the README comment also states that it saves to the DB, but there are no migrations and no changes to the saving functions. I'm quite confused

backend/onyx/chat/chat_utils.py
wenxi-onyx: Before:
```
Assistant message 1: Parallel tool call 1
Tool result 1
Assistant message 2: Parallel tool call 2
Tool result 2
```
After:
```
Assistant message 1: Parallel tool call 1 + parallel tool call 2
Tool result 1
Tool result 2
```

PR #7006, opened 2025-12-24
refactor: llm provider forms

Weves:
## Description

Moves configuration of LLM configuration form structure to the frontend. Each form is it's own dedicated component that uses shared utils. 

Does not apply to the onboarding flow unfortunately, would need to be a separate PR.

## How Has This Been Tested?

Tested creating every provider in the admin panel.

## Additional Options

- [x] [Optional] Override Linear Check

---
## Summary by cubic
Refactored the LLM provider configuration into modular, provider-specific forms with shared components. This streamlines admin setup, improves model discovery, and replaces the legacy in-chat modal flow.

- **Refactors**
  - Added per-provider forms and shared components (FormWrapper, AdvancedOptions, DisplayModels, FetchModelsButton, ApiKeyField, DisplayNameField, SingleDefaultModelField); removed legacy forms, ApiKeyModal/ApiKeyForm, and ConfiguredLLMProviderDisplay.
  - Simplified the admin LLMConfiguration page; NRF now shows a “Set up an LLM” button linking to the admin page instead of an inline modal.
  - Centralized model fetching and provider capability checks (fetchModels, canProviderFetchModels); added backend fetch_model_configurations_for_provider.
  - Minor improvements: labels now use htmlFor; improved OpenAI vs custom provider detection for the correct form.

- **New Features**
  - Backend default model selection via fetch_default_model_for_provider; OpenAI default set to gpt-5.2 and Vertex AI defaults included.
  - API and UI support fetching and selecting visible models (e.g., Bedrock, OpenRouter, Ollama).
  - Added CustomForm and updated tests to cover the end-to-end configuration flow.

<sup>Written for commit e2c2ea007a30e80f750d46d042b725d9ff1eace6. Summary will update automatically on new commits.</sup>

PR #7105, opened 2025-12-30
refactor: onboarding forms

Weves:
## Description

Makes the onboarding forms match the structure of https://github.com/onyx-dot-app/onyx/pull/7006

## How Has This Been Tested?

Set up every model provider locally + verified the experience was identical to the old experience on the happy path

## Additional Options

---
## Summary by cubic
Refactored LLM provider onboarding into modular, provider-specific forms with a shared wrapper. This replaces the old modal/field components, simplifies the UI, centralizes model fetching, and trims the provider descriptor schema.

- **Refactors**
  - Replaced LLMConnectionModal and dynamic field components with provider-specific forms (OpenAI, Anthropic, Azure, Bedrock, VertexAI, OpenRouter, Ollama, Custom).
  - Added OnboardingFormWrapper for shared context (API status, model fetching, modal control).
  - Updated LLMStep to use getOnboardingForm and LLMProviderCard; removed legacy validation schema, Formik effects, and tab configs from constants.
  - Simplified WellKnownLLMProviderDescriptor on backend and UI types to only include needed fields; removed MODAL_CONTENT_MAP and moved display info into getOnboardingForm.

- **Tests**
  - Added unit tests for all provider onboarding forms with shared test helpers; updated Bedrock daily test to use a local default model constant.

<sup>Written for commit 30384ee46068764c567b5da37f709a6029d2b91c. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/refresh-components/onboarding/forms/__tests__/AnthropicOnboardingForm.test.tsx
+++ b/web/src/refresh-components/onboarding/forms/__tests__/AnthropicOnboardingForm.test.tsx
@@ -259,7 +259,8 @@ describe("AnthropicOnboardingForm", () => {
 
       mockFetch
         .mockResolvedValueOnce(mockResponses.testApiSuccess)
-        .mockResolvedValueOnce(mockResponses.createProviderSuccess(1));
+        .mockResolvedValueOnce(mockResponses.createProviderSuccess(1))
+        .mockResolvedValueOnce(mockResponses.setDefaultSuccess);
 
       render(<AnthropicOnboardingForm {...defaultProps} />);
 
@@ -284,7 +285,8 @@ describe("AnthropicOnboardingForm", () => {
 
       mockFetch
         .mockResolvedValueOnce(mockResponses.testApiSuccess)
-        .mockResolvedValueOnce(mockResponses.createProviderSuccess(1));
+        .mockResolvedValueOnce(mockResponses.createProviderSuccess(1))
+        .mockResolvedValueOnce(mockResponses.setDefaultSuccess);
 
       render(<AnthropicOnboardingForm {...defaultProps} />);
 
@@ -309,7 +311,8 @@ describe("AnthropicOnboardingForm", () => {
 
       mockFetch
         .mockResolvedValueOnce(mockResponses.testApiSuccess)
-        .mockResolvedValueOnce(mockResponses.createProviderSuccess(1));
+        .mockResolvedValueOnce(mockResponses.createProviderSuccess(1))
+        .mockResolvedValueOnce(mockResponses.setDefaultSuccess);
 
       render(
         <AnthropicOnboardingForm
@@ -335,7 +338,8 @@ describe("AnthropicOnboardingForm", () => {
 
       mockFetch
         .mockResolvedValueOnce(mockResponses.testApiSuccess)
-        .mockResolvedValueOnce(mockResponses.createProviderSuccess(1));
+        .mockResolvedValueOnce(mockResponses.createProviderSuccess(1))
+        .mockResolvedValueOnce(mockResponses.setDefaultSuccess);
 
       render(
         <AnthropicOnboardingForm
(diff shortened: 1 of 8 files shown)


backend/tests/daily/llm/test_bedrock.py
```diff
@@ -9,6 +9,9 @@
 from onyx.llm.llm_provider_options import WellKnownLLMProviderDescriptor
 
 
+_DEFAULT_BEDROCK_MODEL = "anthropic.claude-3-5-sonnet-20241022-v2:0"
+
+
 @pytest.fixture
 def bedrock_provider() -> WellKnownLLMProviderDescriptor:
```
Subash-Mohan: `bedrock_provider` fixture is not used anywhere.

web/src/refresh-components/onboarding/forms/OllamaOnboardingForm.tsx
```diff
@@ -0,0 +1,428 @@
...
+  // Set hidden fields based on active tab
+  useEffect(() => {
+    if (activeTab === "cloud") {
+      formikProps.setFieldValue("api_base", "https://ollama.com");
+    } else {
+      if (formikProps.values.api_base === "https://ollama.com") {
+        formikProps.setFieldValue("api_base", "http://127.0.0.1:11434");
+      }
+    }
+  }, [activeTab]);
```
Subash-Mohan, line 120: Here we’re only changing the API base per method. What happens if a user enters an API key, then switches the tab to Self-Hosted and proceeds with that option? How are we handling clearing or nullifying the form fields for the inactive tabs?

This applies to Bedrock and other providers that support multiple authentication methods as well.

web/src/refresh-components/onboarding/forms/OnboardingFormWrapper.tsx
```diff
@@ -0,0 +1,369 @@
...
+    // Use fetched model configurations if available
+    let modelConfigsToUse =
+      fetchedModelConfigurations.length > 0
+        ? fetchedModelConfigurations
+        : llmDescriptor?.model_configurations.map((model) => ({
+            name: model.name,
+            is_visible: true,
+            max_input_tokens: model.max_input_tokens,
+            supports_image_input: model.supports_image_input,
+          })) ?? [];
```
Subash-Mohan: Is this valid? Since we removed the default models from the LLM options in the backend, I assume we’re now fetching models dynamically for all providers.

web/src/refresh-components/onboarding/forms/OnboardingFormWrapper.tsx
```diff
@@ -0,0 +1,369 @@
...
+            <ProviderModal
+              open={open}
+              onOpenChange={onOpenChange}
+              title={title}
+              description={description}
+              icon={icon}
+              onSubmit={formikProps.submitForm}
+              submitDisabled={
+                isCustomProvider
+                  ? !formikProps.isValid || !formikProps.dirty
```
Subash-Mohan: Redundant Ternary:
 ```tsx
   submitDisabled={
    isCustomProvider
      ? !formikProps.isValid || !formikProps.dirty
      : !formikProps.isValid || !formikProps.dirty  // Both branches identical
  }```

web/src/refresh-components/onboarding/forms/CustomOnboardingForm.tsx
```diff
@@ -1,43 +1,63 @@
-import React, { useState } from "react";
+import React, { useMemo, useState } from "react";
+import * as Yup from "yup";
+import { FormikProps, useFormikContext } from "formik";
```
Subash-Mohan: `useFormikContext` is unused.

web/src/refresh-components/onboarding/forms/OnboardingFormWrapper.tsx
Weves: I think this should be correct? What's wrong with it?

web/src/refresh-components/onboarding/forms/OllamaOnboardingForm.tsx
Weves: Added useEffect's to clear the fields when they aren't used!

Weves:
@greptile re-review

PR #7114, opened 2025-12-30
feat: add z indices for confirmation modal

wenxi-onyx:
## Description

- There are places in the app where a modal has a second modal popup like a confirmation dialog. Realistically, these should probably be re-designed away, but nevertheless they exist
- Adds confirmation modal z indices that are slightly higher than modal index so make them visible
- Personal access token deletion is one of these "double modal" instances

## Additional Options

---
## Summary by cubic
Fixes stacking order so confirmation modals always appear above other modals and overlays (e.g., PAT confirmation), preventing them from being hidden.

- **Bug Fixes**
  - Added z-index token and utility for confirmation modals: --z-confirmation and z-confirmation.
  - Updated RawModal to use z-modal-overlay and z-modal instead of hardcoded values.
  - Applied z-confirmation to ConfirmationModalLayout content.
  - Standardized Modal.Content to use the z-modal class.

<sup>Written for commit ce982938980628e7d4728aa9b9efc7be0a55453f. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/app/css/z-index.css
+++ b/web/src/app/css/z-index.css
@@ -7,8 +7,7 @@
   /* Interactive overlays */
   --z-modal-overlay: 900;
   --z-modal: 1000;
-  --z-confirmation-overlay: 1050;
-  --z-confirmation: 1075;
+  --z-confirmation: 1050;
   --z-toast: 1100;
   --z-popover: 1200;
   --z-tooltip: 1300;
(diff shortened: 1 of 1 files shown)


web/src/app/css/z-index.css
raunakab, line 1: Is `z-confirmation-overlay` used? I think it'll be fine to keep the overlay at the same z-index, but promote confirmation modals to have a slightly higher z-index (than regular modals, that is).

wenxi-onyx: To double check, the overlay is the darkening of the background behind the modal right? In testing, the overlay actually correctly hides the "first modal," but I can see that it isn't used in this pr... Will investigate and clean this up

PR #7127, opened 2025-12-30
fix: suppress Jest act() warning spam in test output

nmgarza5:
## Description

Fixes the massive console.error spam from Jest tests that was making it hard to identify actual test failures. The spam looked like:

```
console.error
  The current testing environment is not configured to support act(...)
  ...
  at setRef (node_modules/@radix-ui/react-compose-refs/dist/index.js:42:12)
  ... (hundreds of lines)
```

This is a known issue with Radix UI + React + Jest. Radix UI's `compose-refs` package triggers state updates during component unmount, which happens in React's commit phase outside of any `act()` boundary.

Changes:
- Added a `SUPPRESSED_ERRORS` array pattern (following [testing-library/user-event#1114](https://github.com/testing-library/user-event/issues/1114#issuecomment-1876164351)) to filter the specific non-actionable warning
- Added `act` import to MultiToolRenderer tests and wrapped timer execution in `act()` to properly handle timer-triggered state updates

The actionable warning "An update to X inside a test was not wrapped in act" is NOT suppressed - only the configuration warning that we can't fix due to Radix UI internals.

## How Has This Been Tested?

- Ran `npm test` - all 121 tests pass
- Verified the act() warning spam is gone from test output
- Legitimate warnings (HTML nesting, input values, accessibility) are still shown

## Additional Options

- [x] Override Linear Check

### DIFF
--- a/web/src/app/chat/message/messageComponents/MultiToolRenderer.test.tsx
+++ b/web/src/app/chat/message/messageComponents/MultiToolRenderer.test.tsx
@@ -4,7 +4,13 @@
  */
 
 import React from "react";
-import { render, screen, waitFor, setupUser } from "@tests/setup/test-utils";
+import {
+  act,
+  render,
+  screen,
+  waitFor,
+  setupUser,
+} from "@tests/setup/test-utils";
 import MultiToolRenderer from "./MultiToolRenderer";
 import {
   createToolGroups,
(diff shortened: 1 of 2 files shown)


nmgarza5:
@greptile

Weves (approved):

PR #7133, opened 2025-12-31
feat(opensearch): Add OpenSearch queries

acaprau:
## Description
PR chain:
- #7133 (you are here)
    - #7137
        - #7155
        - #7143

## How Has This Been Tested?
#7155 tests these queries.

## Additional Options

- [x] [Optional] Override Linear Check

---
## Summary by cubic
Adds a hybrid OpenSearch query that mixes vector, keyword, and phrase matches, with filters for public and non-hidden documents. Includes normalization pipelines and tuned weights to favor semantic content for better relevance.

- **New Features**
  - DocumentQuery builds five subqueries: title vector, content vector, title keyword (multi_match), content keyword, and content phrase; plus filters for public:true and hidden:false.
  - Added normalization pipeline configs (min-max and z-score) that combine clause scores via arithmetic mean.
  - Updated default hybrid weights: content vector 0.50, content keyword 0.35, title vector 0.05, title keyword 0.05, content phrase 0.05.

<sup>Written for commit e6b635226d03616ad94791fca39dabe06f428e51. Summary will update on new commits.</sup>

PR #7137, opened 2025-12-31
feat(opensearch): Add OpenSearch client

acaprau:
## Description
PR chain:
- #7133
    - #7137 (you are here)
        - #7155
        - #7143

## How Has This Been Tested?
#7155 tests this client.

## Additional Options

---
## Summary by cubic
Introduce a typed OpenSearch client and hybrid search query builder, plus basic OpenSearch app configs and a dev script update. This prepares the codebase to use OpenSearch as a document index backend.

- **New Features**
  - Added OpenSearchClient: index CRUD, document create/get/delete, search, search pipeline create/delete, ping/close, and basic index validation.
  - Added hybrid search query builder and normalization pipeline configs (min-max, z-score) with default weights.
  - Added OpenSearch config vars: OPENSEARCH_HOST, OPENSEARCH_REST_API_PORT, OPENSEARCH_ADMIN_USERNAME, OPENSEARCH_ADMIN_PASSWORD.
  - Updated restart_opensearch_container.sh: reads OPENSEARCH_REST_API_PORT from env, requires OPENSEARCH_ADMIN_PASSWORD, and passes it to the container.
  - Updated DocumentChunk schema: title fields default to None; chunk ID includes source_type to avoid collisions.

- **Migration**
  - Add OPENSEARCH_ADMIN_PASSWORD to .vscode/.env (rename from OPENSEARCH_INITIAL_ADMIN_PASSWORD if you used it locally).
  - Optionally set OPENSEARCH_REST_API_PORT in .env.
  - Run backend/scripts/restart_opensearch_container.sh to recreate the container.

<sup>Written for commit ed8abd8e8cb71df8b142b0aa638d0066ac760ae6. Summary will update on new commits.</sup>


backend/onyx/document_index/opensearch/client.py
```diff
@@ -0,0 +1,336 @@
...
+    def delete_index(self) -> bool:
+        """Deletes the index.
+
+        Raises:
+            Exception: There was an error deleting the index.
+
+        Returns:
+            True if the index was deleted, False if it did not exist.
+        """
+        if self._client.indices.exists(index=self._index_name):
```
evan-onyx: nit: imo it's a little nicer to do
```
if self._client.indices.exists(index=self._index_name):
    logger.warning(f"attempted to delete index {self._index_name} but it does not exist")
    return False
```
and then the rest of the function. but if you prefer the other way the main thing I want is the warning log. I know it's unambiguous right now, but might make diagnosing an issue faster in the future

backend/onyx/document_index/opensearch/client.py
```diff
@@ -0,0 +1,336 @@
...
+        # invoke client.indices.get on an index that does not exist, so we check
+        # for existence explicitly just to be sure.
+        exists_response = self._client.indices.exists(index=self._index_name)
+        if not exists_response:
+            logger.warning(
+                f"Tried to validate index {self._index_name} but it does not exist."
+            )
+            return False
+        get_result = self._client.indices.get(index=self._index_name)
+        # I don't understand this response schema yet so just print for
```
evan-onyx: hopefully this is documented somewhere, if it isn't then it's doubly important we have our own documentation on it. But assuming they have a link it would be great to have here!

backend/onyx/document_index/opensearch/client.py
```diff
@@ -0,0 +1,336 @@
...
+        return DocumentChunk.model_validate_json(document_chunk_source)
+
+    def create_search_pipeline(
+        self,
+        pipeline_id: str,
+        pipeline_body: dict[str, Any],
+    ) -> None:
+        """Creates a search pipeline.
+
+        See the OpenSearch documentation for more information on the search
```
evan-onyx, line 292: link here would be great

backend/onyx/document_index/opensearch/client.py
```diff
@@ -0,0 +1,336 @@
...
+
+    def ping(self) -> bool:
+        """Pings the OpenSearch cluster.
+
+        Returns:
+            True if OpenSearch could be reached, False if it could not.
+        """
+        return self._client.ping()
+
+    def close(self) -> None:
```
evan-onyx, line 373: does this need to be called explicitly by the caller? If so I'd like to use a context manager (with) pattern

backend/onyx/document_index/opensearch/client.py
```diff
@@ -0,0 +1,336 @@
...
+            True if the document was deleted, False if it was not found.
+        """
+        try:
+            result = self._client.delete(index=self._index_name, id=document_chunk_id)
+        except TransportError as e:
+            if e.status_code == 404:
+                return False
+            else:
+                raise e
+        else:
```
evan-onyx: can remove the `else`

evan-onyx (approved):

backend/onyx/document_index/opensearch/client.py
```diff
@@ -0,0 +1,336 @@
...
+            The document chunk.
+        """
+        result = self._client.get(index=self._index_name, id=document_chunk_id)
+        found_result: bool = result.get("found", False)
+        if not found_result:
+            raise RuntimeError(
+                f'Document chunk with ID "{document_chunk_id}" was not found.'
+            )
+
+        document_chunk_source: str = result.get("_source", "")
```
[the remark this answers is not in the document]
acaprau: i know, i fix it in #7155, this isn't in any hotpath i just wanna get it in

backend/onyx/document_index/opensearch/client.py
```diff
@@ -0,0 +1,336 @@
...
+        hits_first_layer: dict[Any, Any] = result.get("hits", {})
+        if not hits_first_layer:
+            raise RuntimeError(
+                f"Hits field missing from response when trying to search index {self._index_name}."
+            )
+        hits_second_layer: list[dict] = hits_first_layer.get("hits", [])
+
+        result_chunks: list[DocumentChunk] = []
+        for hit in hits_second_layer:
+            document_chunk_source: str = hit.get("_source", "")
```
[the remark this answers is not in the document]
acaprau: i know, i fix it in #7155, this isn't in any hotpath i just wanna get it in

backend/onyx/document_index/opensearch/client.py
acaprau: good question. i dont have evidence to say yes here is why callers should always close but i think they should anyway, stuff like this is how we get ooms in prod. thing is the caller for this is always going to be an OpenSearchDocumentIndex, which will have this client object exist for the lifetime of the document index object. no other part of the code should be concerned about this client. so i will wrap this client in a with context pattern, but it will be when i implement the document index interface, and it is possible that interface may actually be the one whose lifetime should always be within a context block

backend/onyx/document_index/opensearch/client.py
```diff
@@ -0,0 +1,379 @@
...
+            )
+        index_mapping_properties: dict[str, Any] = index_info.get("mappings", {}).get(
+            "properties", {}
+        )
+        expected_mapping_properties: dict[str, Any] = expected_mappings.get(
+            "properties", {}
+        )
+        assert (
+            expected_mapping_properties
+        ), "Bug: No properties were found in the provided expected mappings."
```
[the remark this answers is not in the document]
acaprau: yes cubic i know, this is a contract check between internal code, not from a third party library. if someone puts up a pr where they pass in a bad mapping, CI would catch this among other things breaking. by the time we get to prod the code should not have bugs of this type and disabling asserts to improve perf makes sense.

PR #7144, opened 2025-12-31
feat: add tool call validation to eval cli

rohoswagger:
## Description

we should be able to specify if any tools should be required/expected in eval runs

## How Has This Been Tested?

ran an eval with required/expected tools

## Additional Options

---
## Summary by cubic
Adds tool call validation to the eval CLI with per-test tool forcing and Braintrust scoring. Also fixes a chat UI bug where responses were hidden after reasoning steps.

- **New Features**
  - Per-test config: force_tools, expected_tools, require_all_tools.
  - Streaming now records tools_called and details; supports forced_tool_ids in chat requests.
  - Braintrust score “tool_assertion” (1.0/0.0) plus metadata (tools_called, count, details).
  - New CLI flag: --no-send-logs.
  - README updated with examples and available tool types.

- **Bug Fixes**
  - Removed circular dependency in tool display timing (allToolsDisplayed no longer depends on isFinalAnswerComing).
  - Added isActualToolCallPacket and updated AIMessage to ignore reasoning packets so final answers render without refresh.

<sup>Written for commit 42e8706393c192ae6a1d000b7e146cc744251e49. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/evals/eval.py
+++ b/backend/onyx/evals/eval.py
@@ -298,7 +298,7 @@ def _get_answer_with_tools(
                 skip_gen_ai_answer_generation=False,
                 llm_override=full_configuration.llm,
                 allowed_tool_ids=full_configuration.allowed_tool_ids,
-                forced_tool_ids=forced_tool_ids if forced_tool_ids else None,
+                forced_tool_ids=forced_tool_ids or None,
             )
 
             packets = stream_chat_message_objects(


backend/onyx/evals/eval.py
```diff
@@ -82,19 +298,41 @@ def _get_answer(
                 skip_gen_ai_answer_generation=False,
                 llm_override=full_configuration.llm,
                 allowed_tool_ids=full_configuration.allowed_tool_ids,
+                forced_tool_ids=forced_tool_ids if forced_tool_ids else None,
```
evan-onyx: nit: `forced_tool_ids or None`

evan-onyx (approved):

PR #7149, opened 2025-12-31
feat: add new models to cloud

Weves:
## Description

Adds gemini + open source models + turns on auto updating for new tenants.

## Additional Options

---
## Summary by cubic
Add Gemini (Vertex AI) and OpenRouter to Cloud provisioning and enable auto model updates for new tenants. Defaults and visible models now come from our recommended-models config, fetched from GitHub when available, to reduce manual setup.

- **New Features**
  - Provision Vertex AI (Gemini) and OpenRouter when env keys are set.
  - Create Anthropic, OpenAI, Vertex AI, and OpenRouter in auto mode with defaults from recommendations, and set the first configured provider as the default.
  - Set model visibility via recommendations; OpenRouter uses dynamic visible models with display names.

- **Migration**
  - Optional envs: VERTEXAI_DEFAULT_CREDENTIALS, VERTEXAI_DEFAULT_LOCATION (default: global), OPENROUTER_DEFAULT_API_KEY.

<sup>Written for commit a644b1a040ed99b4ff032698529065eaf8eba032. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/llm/well_known_providers/llm_provider_options.py
+++ b/backend/onyx/llm/well_known_providers/llm_provider_options.py
@@ -3,6 +3,7 @@ import pathlib
 
 from onyx.llm.constants import LlmProviderNames
 from onyx.llm.constants import PROVIDER_DISPLAY_NAMES
+from onyx.llm.constants import WELL_KNOWN_PROVIDER_NAMES
 from onyx.llm.utils import get_max_input_tokens
 from onyx.llm.utils import model_supports_image_input
 from onyx.llm.well_known_providers.auto_update_models import LLMRecommendations
@@ -230,7 +231,7 @@ def fetch_available_well_known_llms() -> list[WellKnownLLMProviderDescriptor]:
     llm_recommendations = get_recommendations()
 
     well_known_llms = []
-    for provider_name in LlmProviderNames:
+    for provider_name in WELL_KNOWN_PROVIDER_NAMES:
         model_configurations = model_configurations_for_provider(
             provider_name, llm_recommendations
         )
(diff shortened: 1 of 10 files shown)


backend/onyx/llm/well_known_providers/llm_provider_options.py
```diff
@@ -38,11 +41,11 @@ def _get_provider_to_models_map() -> dict[str, list[str]]:
     }
 
 
-def _get_reccomendations() -> LLMRecommendations:
+def get_reccomendations() -> LLMRecommendations:
     """Get the recommendations from the GitHub config."""
-    # recommendations_from_github = fetch_llm_recommendations_from_github()
-    # if recommendations_from_github:
-    #     return recommendations_from_github
+    recommendations_from_github = fetch_llm_recommendations_from_github()
```
Weves, line 46: oops

backend/ee/onyx/server/tenants/provisioning.py
```diff
@@ -262,59 +274,149 @@ async def rollback_tenant_provisioning(tenant_id: str) -> None:
...
+            if not has_set_default_provider:
+                update_default_provider(provider.id, db_session)
+                has_set_default_provider = True
+        except Exception as e:
+            logger.error(f"Failed to configure {request.provider} provider: {e}")
+
+    # Configure OpenAI provider
+    if OPENAI_DEFAULT_API_KEY:
+        default_model = recommendations.get_default_model(OPENAI_PROVIDER_NAME)
+        default_model_name = default_model.name if default_model else "gpt-4o"
```
nmgarza5: do we want the default to be something newer for openai? gpt 5?

nmgarza5: when would we even hit this actually. wont the recommendation always be available? same for the other providers

nmgarza5: eh i guess the call to github could fail. do we have to do that? can we just embed the json file somewhere?

Weves: yea we should always have a recommendation. A bit dangerous to throw an exception here though, since it will stop new tenants from being crated. Will log an error tho + update the default

PR #7167, opened 2026-01-02
New send message api

yuhongsun96:
## Description

A cleaner send message backend API. Also fixes a bug with the message editing not showing arrows on the user message.

## How Has This Been Tested?
- Sending a message
- Regenerate a root message
- Regenerating a message further in the history
- Editing root message
- Editing message further in history
- Forcing tools
- Running queries w search filters
- Replaying sessions
- The new API directly via requests

## Additional Options

---
## Summary by cubic
Introduced a new send-chat-message API with a simpler request model and automatic chat session creation, and fixed message switching arrows when editing user messages. Deprecated the old /send-message endpoint and improved chat history handling and project-aware tool selection.

- **New Features**
  - Added POST /send-chat-message to stream responses as JSON lines over text/event-stream; streams chat_session_id when a session is auto-created.
  - New SendMessageRequest with:
    - message, llm_override, allowed_tool_ids, forced_tool_id
    - file_descriptors, internal_search_filters, deep_research
    - parent_message_id placement: -1 (latest), null (root), or specific id
    - chat_session_id or chat_session_info (auto-creates session)
  - Centralized session creation via create_chat_session_from_request with project access validation.
  - Auto-creates a root message when missing.
  - Project search behavior: with the default persona in projects, disables search (and any forced search tool) when project files are already loaded or when the project has no files; otherwise enables it.

- **Migration**
  - Use /send-chat-message and SendMessageRequest instead of /send-message.
  - Replace forced_tool_ids with a single forced_tool_id; validated against available tools.
  - Parent placement logic changed; pass parent_message_id per the new semantics.
  - Removed CHAT_TARGET_CHUNK_PERCENTAGE; chunk percentage logic is no longer used.

<sup>Written for commit 292fb9b7f78b5155a19a2de7e3744a059cbb6f1b. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/chat/process_message.py
+++ b/backend/onyx/chat/process_message.py
@@ -5,8 +5,6 @@ An overview can be found in the README.md file in this directory.
 
 import re
 import traceback
-from collections.abc import Callable
-from collections.abc import Iterator
 from uuid import UUID
 
 from sqlalchemy.orm import Session
@@ -15,6 +13,7 @@ from onyx.chat.chat_state import ChatStateContainer
 from onyx.chat.chat_state import run_chat_loop_with_state_containers
 from onyx.chat.chat_utils import convert_chat_history
 from onyx.chat.chat_utils import create_chat_history_chain
+from onyx.chat.chat_utils import create_chat_session_from_request
 from onyx.chat.chat_utils import get_custom_agent_prompt
 from onyx.chat.chat_utils import is_last_assistant_message_clarification
 from onyx.chat.chat_utils import load_all_chat_files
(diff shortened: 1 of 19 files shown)


web/src/app/chat/services/lib.tsx
```diff
@@ -102,78 +102,55 @@ export type PacketType =
...
+  signal?: AbortSignal;
+  deepResearch?: boolean;
+  enabledToolIds?: number[];
+  // Single forced tool ID (new API uses singular, not array)
+  forcedToolId?: number | null;
+  // LLM override parameters
   modelProvider?: string;
   modelVersion?: string;
   temperature?: number;
+  // Fields kept for frontend use but not sent to new API
```
Weves: why do these need to be kept?

backend/onyx/chat/process_message.py
```diff
@@ -246,130 +241,81 @@ def _get_project_search_availability(
...
-    chat_session_id: UUID,
-    db_session: Session,
-    use_existing_user_message: bool = False,
-) -> ChatMessage:
-    root_message = get_or_create_root_message(
-        chat_session_id=chat_session_id, db_session=db_session
-    )
+            if user_force_search_tool:
+                disable_forced_tool = True
+    # We do not set it to ENABLED if the user has forced it, it will be determined later if the forcing
```
Weves: this is pretty confusing

backend/onyx/server/query_and_chat/models.py
```diff
@@ -80,6 +79,44 @@ def check_is_positive_or_feedback_text(self) -> "ChatFeedbackRequest":
...
+
+    file_descriptors: list[FileDescriptor] = []
+
+    internal_search_filters: BaseFilters | None = None
+
+    deep_research: bool = False
+
+    # Placement information for the message in the conversation tree:
+    # - -1: auto-place after latest message in chain
+    # - null: regeneration from root (first message)
```
Weves, line 98: why do we want this option?

backend/onyx/chat/process_message.py
yuhongsun96: Agreed lmao, let me see if there's a better way

web/src/app/chat/services/lib.tsx
yuhongsun96: Asked this to the LLM so that it doesn't need to refactor more frontend code which I can't understand very well. Let me ask it to clean it up and I'll see if I'm concerned about it or not

PR #7169, opened 2026-01-02
fix(braintrust): Implement actual TTFA Metric

justin-tahara:
## Description

In this PR: https://github.com/onyx-dot-app/onyx/pull/7168 the metric was incorrectly named so actually introducing the correct metric

## How Has This Been Tested?

Setup braintrust locally to validate the metric was correct.

## Additional Options

---
## Summary by cubic
Implement the correct Time To First Action for LLM streams. We now measure from stream start to the first content/tool action and log it as time_to_first_action_seconds; total end-to-end time is logged as total_latency_seconds.

- **Bug Fixes**
  - Compute time to first action when the first content/reasoning delta or tool call arrives.
  - Add time_to_first_action_seconds to GenerationSpanData and export to Braintrust/OpenInference.
  - Replace previous TTFT (span duration) with total_latency_seconds.

- **Migration**
  - Update dashboards/queries to use time_to_first_action_seconds.
  - Use total_latency_seconds for full request duration.

<sup>Written for commit 6879a253a520cea36a46a1c082a05c11c8fa6ab3. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/chat/llm_step.py
+++ b/backend/onyx/chat/llm_step.py
@@ -430,7 +430,7 @@ def _increment_turns(
         return turn_index, sub_turn_index + 1
 
 
-def _delta_has_content(delta: Delta) -> bool:
+def _delta_has_action(delta: Delta) -> bool:
     return bool(delta.content or delta.reasoning_content or delta.tool_calls)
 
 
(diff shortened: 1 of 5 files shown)


yuhongsun96 (commented):
This should probably capture any delta sent imo. If it calls a tool or does reasoning, that counts as something has happened for the user.

PR #7171, opened 2026-01-02
chore(playwright): prefer baseURL

jmelahman:
## Description

Makes the tests cleaner and allows me to easily override this value (I run my dev server on `http://localhost:3001`).

## How Has This Been Tested?

Captured by existing

## Additional Options

- [x] Override Linear Check

---
## Summary by cubic
Switch tests to Playwright’s baseURL and relative routes to remove hardcoded hosts and make e2e runs portable. Supports overriding BASE_URL so devs can run against different ports without changing tests.

- **Refactors**
  - Set use.baseURL in Playwright config (defaults to http://localhost:3000; can be overridden via BASE_URL).
  - Replace absolute URLs in page navigation and waitForURL with relative paths.
  - Build API request URLs using BASE_URL (including OnyxApiClient and page.request calls).
  - Exclude /tests from Docker build context to avoid bundling e2e assets.

<sup>Written for commit d647040e55ad1628299b15afddf1631a39b8a259. Summary will update on new commits.</sup>

### DIFF
--- a/web/tests/e2e/global-setup.ts
+++ b/web/tests/e2e/global-setup.ts
@@ -4,19 +4,22 @@ import { inviteAdmin2AsAdmin1, loginAs } from "./utils/auth";
 async function globalSetup(config: FullConfig) {
   const browser = await chromium.launch();
 
-  const adminContext = await browser.newContext();
+  // Get baseURL from config, fallback to localhost:3000
+  const baseURL = config.projects[0]?.use?.baseURL || "http://localhost:3000";
+
+  const adminContext = await browser.newContext({ baseURL });
   const adminPage = await adminContext.newPage();
   await loginAs(adminPage, "admin");
   await adminContext.storageState({ path: "admin_auth.json" });
   await adminContext.close();
 
-  const userContext = await browser.newContext();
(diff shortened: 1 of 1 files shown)


web/tests/e2e/admin/default-assistant.spec.ts
```diff
@@ -66,11 +62,10 @@ test.describe("Default Assistant Admin Page", () => {
 
     // Proactively log tool availability and current config
     try {
-      const toolsResp = await page.request.get(
-        "http://localhost:3000/api/tool"
-      );
+      const baseURL = process.env.BASE_URL || "http://localhost:3000";
+      const toolsResp = await page.request.get(`${baseURL}/api/tool`);
```
jmelahman, line 66: FYI, apparently these `page.request.get`s doesn't inherit the global config context (`baseURL`), so might replace these with `goto`, but unsure of the effort required.

justin-tahara (approved):

web/tests/e2e/admin/default-assistant.spec.ts
jmelahman: Eh, it's contained to this file alone, so fine I guess.

PR #7181, opened 2026-01-03
feat: nonstreaming send chat message api

yuhongsun96:
## Description

Provide a stream parameter so that developers can use it more easily without reconstructing from packets.

## How Has This Been Tested?

Works locally with UI (for testing existing path)
Tested via API calls directly with personal API key for nonstream path:
- With tools
- With message editing
- With not specifying chat session id

## Additional Options

---
## Summary by cubic
Adds non‑streaming support to /send-chat-message via a new stream flag, returning a complete response when stream=false. This removes the need to reconstruct SSE packets while keeping streaming as the default.

- **New Features**
  - SendMessageRequest now includes stream (default true); when false, the API returns a ChatFullResponse instead of SSE.
  - Added ChatFullResponse and ToolCallResponse models with answer, citations, documents, reasoning, tool calls, and generated images.
  - Backend aggregates packets with gather_stream_full and supports an external state container for non‑streaming accumulation.

<sup>Written for commit 55ae80837e95bca706c9204490ad3838bdde5cb6. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/chat/models.py
+++ b/backend/onyx/chat/models.py
@@ -228,6 +228,8 @@ class ToolCallResponse(BaseModel):
     tool_result: str
     search_docs: list[SearchDoc] | None = None
     generated_images: list[GeneratedImage] | None = None
+    # Reasoning that led to the tool call
+    pre_reasoning: str | None = None
 
 
 class ChatFullResponse(BaseModel):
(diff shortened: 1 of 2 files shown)


backend/onyx/chat/models.py
```diff
@@ -219,6 +220,37 @@ class ChatBasicResponse(BaseModel):
...
+    message_id: int
+    chat_session_id: UUID | None = None
+    error_msg: str | None = None
+
+    # Documents & citations
+    top_documents: list[SearchDoc]
+    citation_info: list[CitationInfo]
+
+    # Reasoning (null when model doesn't produce reasoning)
+    reasoning: str | None = None
```
Weves: isn't reasoning interwoven within tool calls?

Weves (approved):

PR #7192, opened 2026-01-04
feat: cloud usage limits

evan-onyx:
## Description

We've seen people abusing the Onyx trial process by using very expensive models for very large non-productivity workloads. This is the implementation of a system that ensures that folks doing this must provide their own keys when usage exceeds a certain limit.

Also implements captcha, but this won't work until we set up the secrets on our cloud

## How Has This Been Tested?

ALERT this is pretty hard to test, so I haven't verified that the fix actually works. At best I've ran the UI and made sure that normal calls still run successfully.

added some basic unit tests but they don't cover much

## Additional Options

---
## Summary by cubic
Add weekly cloud usage limits and enforcement for Onyx-managed LLM keys, plus CAPTCHA on signup to reduce abuse. Tracks tenant usage (LLM cost, indexing, API calls) and blocks over-limit requests unless users bring their own keys.

- New Features
  - Track usage per-tenant per window (tenant_usage): LLM cost (cents via litellm), chunks indexed, API calls, non-streaming calls.
  - Enforce limits for Onyx-managed keys: pre-check LLM cost, gate non-streaming endpoints, limit chunk indexing, and count API key/PAT calls.
  - Trial vs paid limits with EE trial detection; defaults configurable and aligned to weekly windows.
  - Added CAPTCHA (reCAPTCHA v3) on signup: backend verification and frontend hook passing token in header/body.
  - Unit tests for usage windowing, increments, and limit checks.

- Migration
  - Run Alembic migration 2b90f3af54b8 to create tenant_usage.
  - Configure env vars as needed:
    - USAGE_LIMITS_ENABLED (defaults to MULTI_TENANT), USAGE_LIMIT_WINDOW_SECONDS, and per-tier limits (LLM cost, chunks, API calls).
    - OPENAI_DEFAULT_API_KEY, ANTHROPIC_DEFAULT_API_KEY, COHERE_DEFAULT_API_KEY, OPENROUTER_DEFAULT_API_KEY (moved to OSS configs).
    - CAPTCHA_ENABLED, RECAPTCHA_SECRET_KEY, RECAPTCHA_SCORE_THRESHOLD, and NEXT_PUBLIC_RECAPTCHA_SITE_KEY (web).

<sup>Written for commit d4c08c919621f5d67940a8faf5c315a7a3fe96cb. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/unit/onyx/db/test_usage.py
+++ b/backend/tests/unit/onyx/db/test_usage.py
@@ -43,32 +43,40 @@ class TestGetCurrentWindowStart:
 class TestGetOrCreateTenantUsage:
     """Tests for get_or_create_tenant_usage function."""
 
-    def test_creates_new_usage_record_when_none_exists(self) -> None:
-        """Test that a new usage record is created when none exists."""
+    def test_creates_or_gets_usage_record(self) -> None:
+        """Test that get_or_create returns a usage record via atomic upsert."""
+        mock_usage = MagicMock()
+        mock_usage.llm_cost_cents = 0.0
+        mock_usage.chunks_indexed = 0
+
         mock_session = MagicMock()
-        mock_session.execute.return_value.scalar_one_or_none.return_value = None
+        # The new implementation uses INSERT ... ON CONFLICT with RETURNING
+        # which calls execute().scalar_one()
+        mock_session.execute.return_value.scalar_one.return_value = mock_usage
 
         window_start = datetime(2024, 1, 1, tzinfo=timezone.utc)
-        get_or_create_tenant_usage(mock_session, window_start)
+        usage = get_or_create_tenant_usage(mock_session, window_start)
 
-        # Verify a new TenantUsage was added
-        mock_session.add.assert_called_once()
+        # Verify execute was called (with the INSERT ... ON CONFLICT statement)
+        mock_session.execute.assert_called_once()
         mock_session.flush.assert_called_once()
+        assert usage == mock_usage
 
-    def test_returns_existing_usage_record(self) -> None:
-        """Test that existing usage record is returned if it exists."""
+    def test_returns_usage_record_from_atomic_upsert(self) -> None:
+        """Test that the returned usage record comes from the atomic upsert."""
         mock_usage = MagicMock()
         mock_usage.llm_cost_cents = 100.0
         mock_usage.chunks_indexed = 500
 
         mock_session = MagicMock()
-        mock_session.execute.return_value.scalar_one_or_none.return_value = mock_usage
+        mock_session.execute.return_value.scalar_one.return_value = mock_usage
 
         window_start = datetime(2024, 1, 1, tzinfo=timezone.utc)
         usage = get_or_create_tenant_usage(mock_session, window_start)
 
         assert usage == mock_usage
-        mock_session.add.assert_not_called()
+        assert usage.llm_cost_cents == 100.0
+        assert usage.chunks_indexed == 500
 
 
 class TestGetTenantUsageStats:


backend/ee/onyx/configs/app_configs.py
justin-tahara, line 1: This is just cleanup right?

backend/ee/onyx/server/tenants/provisioning.py
justin-tahara, line 1: More cleanup?

backend/ee/onyx/server/usage_limits.py
```diff
@@ -0,0 +1,47 @@
...
+
+logger = setup_logger()
+
+
+def is_tenant_on_trial(tenant_id: str) -> bool:
+    """
+    Determine if a tenant is currently on a trial subscription.
+
+    In multi-tenant mode, we fetch billing information from the control plane
+    to determine if the tenant has an active trial.
```
justin-tahara, line 20: I don't think we need to personally put this down in a comment but I guess it's fine.

backend/ee/onyx/server/usage_limits.py
justin-tahara, line 1: This might be useful for other enforcement we should do later

backend/onyx/auth/captcha.py
```diff
@@ -0,0 +1,106 @@
...
+
+class RecaptchaResponse(BaseModel):
+    """Response from Google reCAPTCHA verification API."""
+
+    success: bool
+    score: float | None = None  # Only present for reCAPTCHA v3
+    action: str | None = None
+    challenge_ts: str | None = None
+    hostname: str | None = None
+    error_codes: list[str] | None = None
```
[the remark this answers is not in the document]
justin-tahara: Would double check since this will affect the behavior of the captcha I believe

backend/onyx/auth/captcha.py
justin-tahara, line 1: Personally for me this is more of a nice to have. I think we could just follow up with this later if need be but if you think it's essential then we can work to get this in

backend/onyx/auth/captcha.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def is_captcha_enabled() -> bool:
+    """Check if captcha verification is enabled."""
+    return CAPTCHA_ENABLED and bool(RECAPTCHA_SECRET_KEY)
```
justin-tahara, line 34: Does this make sure that it's a cloud only thing?

backend/onyx/chat/llm_step.py
```diff
@@ -543,6 +543,14 @@ def run_llm_step_pkt_generator(
                     "cache_read_input_tokens": usage.cache_read_input_tokens,
                     "cache_creation_input_tokens": usage.cache_creation_input_tokens,
                 }
+                # Track usage in state container for cost calculation
```
justin-tahara, line 546: Do we have to do a transformation by hitting litellm to get the actual costs?

backend/onyx/db/models.py
```diff
@@ -3962,3 +3962,46 @@ class License(Base):
     updated_at: Mapped[datetime.datetime] = mapped_column(
         DateTime(timezone=True), server_default=func.now(), onupdate=func.now()
     )
+
+
+class TenantUsage(Base):
```
justin-tahara, line 3967: I would really want to pass this through to Braintrust as well since seeing this information just in the DB can be kind of annoying

backend/onyx/llm/cost.py
```diff
@@ -0,0 +1,39 @@
+"""LLM cost calculation utilities."""
+
+from onyx.utils.logger import setup_logger
+
+logger = setup_logger()
+
+
+def calculate_llm_cost_cents(
```
justin-tahara, line 8: We can reuse this for a new span attribute for costs in braintrust

backend/onyx/server/query_and_chat/query_backend.py
```diff
@@ -273,6 +277,22 @@ def get_answer_with_citation(
...
+            usage_type=UsageType.NON_STREAMING_API_CALLS,
+            tenant_id=tenant_id,
+            pending_amount=1,
+        )
+        increment_usage(
+            db_session=db_session,
+            usage_type=UsageType.NON_STREAMING_API_CALLS,
+            amount=1,
+        )
+        db_session.commit()
```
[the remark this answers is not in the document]
evan-onyx: the purpose of this is to prevent people from excessively _hitting_ the endpoints i.e. some kind of denial of service, so this is appropriate

backend/onyx/server/usage_limits.py
```diff
@@ -0,0 +1,192 @@
...
+    Uses fetch_versioned_implementation to get the EE version if available,
+    otherwise falls back to the non-EE version that returns False.
+    """
+    fn: Callable[[str], bool] = fetch_versioned_implementation(
+        "onyx.server.usage_limits", "is_tenant_on_trial"
+    )
+    return fn(tenant_id)
+
+
+def get_limit_for_usage_type(usage_type: UsageType, is_trial: bool) -> int:
```
justin-tahara, line 74: I think we may want to work on this a bit more since the chunks indexed is the stuff we talked with Andrei and I think it might be a lil more tricky but we can keep this for now. Might be good to keep a mental note to make sure that we should double check this.

backend/onyx/server/query_and_chat/chat_backend.py
```diff
@@ -521,6 +537,21 @@ def handle_send_chat_message(
     # Non-streaming path: consume all packets and return complete response
     if not chat_message_req.stream:
         with get_session_with_current_tenant() as db_session:
+            # Check and track non-streaming API usage limits
+            if is_usage_limits_enabled():
```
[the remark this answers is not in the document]
evan-onyx: I'm okay with living on the edge for this one

backend/onyx/server/api_key_usage.py
```diff
@@ -0,0 +1,56 @@
...
+    is_api_key_request = get_hashed_api_key_from_request(request) is not None
+    is_pat_request = get_hashed_pat_from_request(request) is not None
+
+    if not is_api_key_request and not is_pat_request:
+        return
+
+    tenant_id = get_current_tenant_id()
+
+    # Check usage limit
+    check_usage_and_raise(
```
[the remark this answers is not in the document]
evan-onyx: I think we kind of just have to live with this is we want to let people do things in parallel. We can try to address if the bad actors want to put in lots of effort to exploit this (which we'll see pretty quickly)

backend/ee/onyx/configs/app_configs.py
evan-onyx: these were moved to non-ee so we can access them in non-ee without needing a fetch_versioned_implementation

backend/ee/onyx/server/tenants/provisioning.py
evan-onyx: they're now imported from the non ee configs file

backend/ee/onyx/server/usage_limits.py
evan-onyx: I feel like its fine in the ee section of the repo

backend/onyx/auth/captcha.py
evan-onyx: this won't be active during this PR bc we need to set up the env vars anyways. imo it IS pretty necessary, otherwise people can pretty easily just programmtically make a new tenant when they run out of usage

backend/onyx/auth/captcha.py
evan-onyx: well it's default off, in theory non cloud could use captchas too if they want to

backend/onyx/chat/llm_step.py
evan-onyx: yeah later on we get the total usage to get costs

backend/onyx/db/models.py
evan-onyx: idk how to do this off the top of my head, probably this can be a follow up PR

backend/onyx/llm/cost.py
evan-onyx: sg but yeah follow up

backend/onyx/server/usage_limits.py
evan-onyx: idr what we were yapping about but yeah sure happy to iterate on this later

backend/onyx/auth/captcha.py
justin-tahara: I see that makes sense. I guess if they get really intricate they can just keep swapping over and over again

backend/onyx/db/models.py
justin-tahara: I can probs do it so dw

PR #7196, opened 2026-01-04
fix: llm usage tracking for dr

evan-onyx:
## Description

we were previously not counting all LLM calls; now we are

## How Has This Been Tested?

not tested beyond unit tests, but basic flows work.
## Additional Options

---
## Summary by cubic
Fixes LLM usage and cost tracking so all model calls, including deep research and streaming, are counted accurately. Accounting is now centralized in multi_llm and respects usage limits and Onyx-managed keys.

- **Bug Fixes**
  - Centralized cost tracking in multi_llm with _track_llm_cost invoked after invoke/stream.
  - Added Usage to ModelResponse and parse usage from LiteLLM responses.
  - Removed token tally in ChatState and manual aggregation in chat_backend.
  - Compute cost via calculate_llm_cost_cents and persist to usage; log failures without affecting requests.

<sup>Written for commit 426831044384b4cba4a6403b2a18cfbc3c823b71. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/llm/multi_llm.py
+++ b/backend/onyx/llm/multi_llm.py
@@ -238,14 +238,15 @@ class LitellmLLM(LLM):
 
         if not is_usage_limits_enabled():
             return
-        # Import here to avoid circular imports
-        from onyx.db.engine.sql_engine import get_session_with_current_tenant
-        from onyx.db.usage import increment_usage
-        from onyx.db.usage import UsageType
+
         from onyx.server.usage_limits import is_onyx_managed_api_key
 
         if not is_onyx_managed_api_key(self._api_key):
             return
+        # Import here to avoid circular imports
+        from onyx.db.engine.sql_engine import get_session_with_current_tenant
+        from onyx.db.usage import increment_usage
+        from onyx.db.usage import UsageType
 
         # Calculate cost in cents
         cost_cents = calculate_llm_cost_cents(


backend/onyx/llm/multi_llm.py
```diff
@@ -222,6 +224,48 @@ def _record_error(
                 category=_LLM_PROMPT_LONG_TERM_LOG_CATEGORY,
             )
 
+    def _track_llm_cost(self, usage: Usage) -> None:
```
acaprau, line 227: who uses this respectable method? i don't think i see an invocation in this pr, will it be plugged in later?

backend/onyx/llm/multi_llm.py
```diff
@@ -222,6 +224,48 @@ def _record_error(
...
+        """
+        Track LLM usage cost for Onyx-managed API keys.
+
+        This is called after every LLM call completes (streaming or non-streaming).
+        Cost is only tracked if:
+        1. Usage limits are enabled for this deployment
+        2. The API key is one of Onyx's managed default keys
+        """
+
+        from onyx.server.usage_limits import is_usage_limits_enabled
```
acaprau, line 237: this is a nit shouldn't block the pr but i rly don't dynamic imports. in general this makes app performance hard to predict after startup and makes the code harder to read, and in particular this import and the one below it will essentially almost always be invoked bc i assume this method will get called for every chat. below the `is_onyx_managed_api_key` line i could maybe buy into dynamic imports.

backend/onyx/llm/multi_llm.py
```diff
@@ -222,6 +224,48 @@ def _record_error(
...
+
+        if not is_usage_limits_enabled():
+            return
+
+        from onyx.server.usage_limits import is_onyx_managed_api_key
+
+        if not is_onyx_managed_api_key(self._api_key):
+            return
+        # Import here to avoid circular imports
```
acaprau, line 246: i did say i could buy into this dynamic import, although this comment indicates that the below modules should be factored in a better way.

acaprau (approved):

PR #7197, opened 2026-01-05
feat: per-tenant usage limits

evan-onyx:
## Description

Some of our cloud tenants are customers with which we have explicit agreements about usage limits (or lack thereof). We can now fetch per-tenant limits from the control plane (use defaults when the request fails).

## How Has This Been Tested?

n/a

## Additional Options

---
## Summary by cubic
Add per-tenant usage limits with control-plane overrides. Tenants can now have custom limits or unlimited usage; defaults are used when no override is present or fetch fails.

- **New Features**
  - EE: Fetch tenant override limits from the control plane, cache in memory, and load at server startup via versioned implementation.
  - Override fields for trial/paid: llm_cost_cents_trial/paid, chunks_indexed_trial/paid, api_calls_trial/paid, non_streaming_calls_trial/paid. -1 means unlimited; None falls back to env defaults.
  - Updated limit checks to be tenant-aware and skip enforcement when NO_LIMIT (-1) is set.
  - Non-EE: Stubbed overrides (always None) so non-EE deployments use defaults only.

<sup>Written for commit 5bbd628ea91c0ef6a6caa22544f47c64e4a37c1f. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/server/tenant_usage_limits.py
+++ b/backend/onyx/server/tenant_usage_limits.py
@@ -4,7 +4,7 @@ Non-EE version of tenant usage limit overrides.
 In non-EE deployments, there are no tenant-specific overrides - all tenants
 use the default limits from environment variables.
 
-The EE version (ee.onyx.server.tenants.usage_limits) fetches per-tenant
+The EE version (ee.onyx.server.tenant_usage_limits) fetches per-tenant
 overrides from the control plane.
 """
 
--- a/backend/onyx/server/usage_limits.py
+++ b/backend/onyx/server/usage_limits.py
@@ -12,8 +12,8 @@ from onyx.configs.app_configs import OPENROUTER_DEFAULT_API_KEY
 from onyx.db.usage import check_usage_limit
 from onyx.db.usage import UsageLimitExceededError
 from onyx.db.usage import UsageType
-from onyx.server.tenants.usage_limits import TenantUsageLimitKeys
-from onyx.server.tenants.usage_limits import TenantUsageLimitOverrides
+from onyx.server.tenant_usage_limits import TenantUsageLimitKeys
+from onyx.server.tenant_usage_limits import TenantUsageLimitOverrides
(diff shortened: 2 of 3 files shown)


backend/onyx/server/tenant_usage_limits.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    return None
+
+
+def load_usage_limit_overrides() -> None:
+    """
+    Load tenant usage limit overrides from the control plane.
+
+    Non-EE version is a no-op since there's no control plane to fetch from.
+    """
```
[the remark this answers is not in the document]
Weves: +1

backend/onyx/server/usage_limits.py
```diff
@@ -71,29 +71,80 @@ def is_tenant_on_trial_fn(tenant_id: str) -> bool:
...
+    """
+    Get the appropriate limit based on usage type, trial status, and tenant overrides.
+
+    Returns:
+        - Positive int: The usage limit
+        - NO_LIMIT (-1): No limit (unlimited) for this tenant
+    """
+    # Determine field names for this usage type
+    if usage_type == UsageType.LLM_COST:
+        trial_field = "llm_cost_cents_trial"
```
Weves: maybe should be an enum?

backend/onyx/main.py
```diff
@@ -294,6 +294,15 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
     else:
         setup_multitenant_onyx()
 
+        # Load tenant-specific usage limit overrides from control plane
+        try:
+            load_overrides_fn = fetch_versioned_implementation(
+                "onyx.server.tenants.usage_limits", "load_usage_limit_overrides"
+            )
+            load_overrides_fn()
```
Weves: do we need to load this for background workers (e.g. chunk limit)?

evan-onyx: ah good catch

PR #7198, opened 2026-01-05
seeds(config): image gen from llm providers

Subash-Mohan:
## Description
This pull request implements automatic creation of a default image generation configuration when a new tenant is provisioned, specifically leveraging an OpenAI API key if available. The changes include a new Alembic migration for seeding the default config, a new helper function for creating the config programmatically, integration into the tenant provisioning flow, and an integration test to verify correct behavior.

## How Has This Been Tested?
For Alembic, I tested it by upgrading the schema. For tenant provisioning, I wasn’t able to test it locally, so I added an integration test to verify it.

## Additional Options

---
## Summary by cubic
Automatically creates a default image generation config for new tenants using the OpenAI API key, so image generation works out of the box. This adds a safe migration and provisioning hook to set up gpt-image-1 as the default.

- **New Features**
  - Added alembic migration to seed a default image generation config (provider id: openai_gpt_image_1, model: gpt-image-1) using the OpenAI provider’s API key.
  - Introduced create_default_image_gen_config_from_api_key and integrated it into tenant provisioning; skips if a config already exists and logs errors without blocking provisioning.
  - Added an integration test that verifies a single default config is created with gpt-image-1.

- **Migration**
  - Run alembic upgrade; no manual steps required. No-op when configs already exist or no OpenAI provider is present.
  - Set OPENAI_DEFAULT_API_KEY to enable automatic creation during tenant provisioning.

<sup>Written for commit 9112437dea4cdeb84ac2b0d6dce5313b1971cd63. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/configs/app_configs.py
+++ b/backend/onyx/configs/app_configs.py
@@ -800,6 +800,11 @@ ENTERPRISE_EDITION_ENABLED = (
     os.environ.get("ENABLE_PAID_ENTERPRISE_EDITION_FEATURES", "").lower() == "true"
 )
 
+#####
+# Image Generation Configuration (DEPRECATED)
+# These environment variables will be deprecated soon.
+# To configure image generation, please visit the Image Generation page in the Admin Panel.
+#####
 # Azure Image Configurations
 AZURE_IMAGE_API_VERSION = os.environ.get("AZURE_IMAGE_API_VERSION") or os.environ.get(
     "AZURE_DALLE_API_VERSION"


backend/alembic/versions/9087b548dd69_seed_default_image_gen_config.py
```diff
@@ -0,0 +1,136 @@
...
+        {
+            "name": f"Image Gen - {IMAGE_PROVIDER_ID}",
+            "provider": PROVIDER_NAME,
+            "api_key": api_key,
+            "default_model_name": MODEL_NAME,
+            "is_public": True,
+            "is_auto_mode": False,
+        },
+    )
+    new_provider_id = result.scalar()
```
evan-onyx, line 85: if we already have an openai provider, why do we need to make a new one specificaly for image gen?

backend/onyx/db/image_generation.py
```diff
@@ -153,3 +163,84 @@ def delete_image_generation_config__no_commit(
...
+    # Check if any image generation configs already exist (optimization to avoid work)
+    existing_configs = get_all_image_generation_configs(db_session)
+    if existing_configs:
+        logger.info("Image generation config already exists, skipping default creation")
+        return None
+
+    try:
+        # Create new LLM provider for image generation
+        new_provider = LLMProvider(
+            name=f"Image Gen - {image_provider_id}",
```
evan-onyx, line 200: Can we use this name in the migration too for standardization?

backend/onyx/db/image_generation.py
```diff
@@ -153,3 +163,84 @@ def delete_image_generation_config__no_commit(
...
+        )
+
+        model_config = ModelConfiguration(
+            llm_provider_id=new_provider.id,
+            name=model_name,
+            is_visible=True,
+            max_input_tokens=max_input_tokens,
+        )
+        db_session.add(model_config)
+        db_session.flush()
```
evan-onyx, line 225: Since we're using the same session for this I don't think we need to flush()

evan-onyx (approved):
approving to unblock!

backend/alembic/versions/9087b548dd69_seed_default_image_gen_config.py
Subash-Mohan: We create a new llm provider to ensure that api key changes in the llm provider admin page and image generation admin page won't interfere with each other.

backend/onyx/db/image_generation.py
Subash-Mohan: Both are same name.

PR #7224, opened 2026-01-06
feat(image-generation): Add Azure OpenAI GPT image models 

Subash-Mohan:
## Description
This pull request added the gpt image models for azure and fixed the azure specific issue.
Screenshot 2026-01-06 at 8 09 54 PM

## How Has This Been Tested?
Tested by generating the image.

## Additional Options

---
## Summary by cubic
Adds support for Azure OpenAI GPT Image 1 and 1.5, and fixes Azure model name formatting so image generation works with LiteLLM deployments.

- **New Features**
  - Added Azure GPT Image 1 and 1.5 options in admin image-generation config.
  - Model name is derived from the Azure deployment in the target URI.

- **Bug Fixes**
  - Backend now formats Azure model names as azure/<deployment_name> (fallback to configured name) for LiteLLM.
  - Prevents image generation failures caused by incorrect Azure model identifiers.

<sup>Written for commit 83b66416b34de6e516f4949c2853ab3bb328edd1. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/app/admin/configuration/image-generation/constants.ts
+++ b/web/src/app/admin/configuration/image-generation/constants.ts
@@ -45,12 +45,12 @@ export const IMAGE_PROVIDER_GROUPS: ProviderGroup[] = [
     name: "Azure OpenAI",
     providers: [
       {
-        image_provider_id: "azure_dalle_3",
+        image_provider_id: "azure_gpt_image_1_5",
         model_name: "", // Extracted from deployment in target URI
         provider_name: "azure",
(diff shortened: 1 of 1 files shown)


wenxi-onyx (approved):
Do you know if we store azure LLMs as `azure/model-name` in LLMConfig.model_name? 

I don't think it matters too much, but it would be nice to be consistent here I think. For reference, LitellmLLM._completion creates the "final model name" immediately before the litellm.completion call (L312).

PR #7233, opened 2026-01-06
fix(chat): enable exclusion of failed chat sessions from api

Danelegend:
Failed chats are currently being included in the recent tab. We don't want this to be the case. This occurs because the tab hits the "/get-user-chat-sessions" endpoint which currently includes failed chats.

## Description
Add a query parameter to the endpoint to allow to opt out of failed chat inclusion. By default this is true.
In the database level, we use a subquery to filter out chat-sessions with no messages in them (except for system).

## How Has This Been Tested?

Manual / Visual test:

Before View -> Example w/ 2 failed chat sessions (Note the two unnamed conversations)
Screenshot 2026-01-06 at 2 43 54 PM

After View -> 2 Failed chat sessions no longer visible
Screenshot 2026-01-06 at 2 44 40 PM

## Additional Options
closes https://linear.app/onyx-app/issue/ENG-3250/failed-chats-should-not-show

---
## Summary by cubic
Stop showing failed chat sessions in the Recent tab by default to match expected behavior and reduce noise.

- **Bug Fixes**
  - Filter out chat sessions that only contain system messages using an EXISTS subquery.
  - Added include_failed_chats query param (default false) to /get-user-chat-sessions to allow opting in when needed.
  - Addresses Linear ENG-3250.

<sup>Written for commit fbbd03045d8f6980ac15f41d72538d59c1162217. Summary will update on new commits.</sup>


justin-tahara (approved):

Danelegend:
Closes https://linear.app/onyx-app/issue/ENG-3250/failed-chats-should-not-show

PR #7241, opened 2026-01-07
fix(sidebar): Allow unpinning all agents and fix icon flicker

nmgarza5:
## Description

Fixes agent pinning issues in the sidebar:
- Allow unpinning all agents (distinguish `null` vs `[]` for pinned preferences)
- Prevent icon flicker when unpinning
- Persist pins when clicking sidebar agents or chat sessions
- Hide unpin button for current agent

## How Has This Been Tested?

Closes https://linear.app/onyx-app/issue/ENG-3311/cant-unpin-agent

### DIFF
--- a/web/src/sections/sidebar/AgentButton.tsx
+++ b/web/src/sections/sidebar/AgentButton.tsx
@@ -2,14 +2,13 @@
 
 import React, { memo } from "react";
 import { MinimalPersonaSnapshot } from "@/app/admin/assistants/interfaces";
-import { usePinnedAgents } from "@/hooks/useAgents";
+import { usePinnedAgents, useCurrentAgent } from "@/hooks/useAgents";
 import { useAppRouter } from "@/hooks/appNavigation";
 import { cn, noProp } from "@/lib/utils";
 import SidebarTab from "@/refresh-components/buttons/SidebarTab";
 import IconButton from "@/refresh-components/buttons/IconButton";
 import { useSortable } from "@dnd-kit/sortable";
 import { CSS } from "@dnd-kit/utilities";
-import useAppFocus from "@/hooks/useAppFocus";
 import useOnMount from "@/hooks/useOnMount";
 import AgentAvatar from "@/refresh-components/avatars/AgentAvatar";
 import { SvgPin, SvgX } from "@opal/icons";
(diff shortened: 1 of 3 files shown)


nmgarza5:
> A recent change made by Yuhong will cause agents to be pinned when a user tries to chat with them. Because of this change, I think we can remove some logic around `currentAgentIsPinned` and the new `pinned` prop. Please let me know if this is not possible. Yuhong's PR -> #7261

makes sense. if the agentButton is in the sidebar, we will just always pin. I'm going to keep the pinned prop though to prevent icon flicker when unpinning

web/src/sections/sidebar/AgentButton.tsx
```diff
@@ -46,35 +45,47 @@ function SortableItem({ id, children }: SortableItemProps) {
...
-  const activeSidebarTab = useAppFocus();
+  const currentAgent = useCurrentAgent();
   const { pinnedAgents, togglePinnedAgent } = usePinnedAgents();
-  const pinned = pinnedAgents.some(
-    (pinnedAgent) => pinnedAgent.id === agent.id
-  );
+  // Actual pinned state from server (for logic)
+  const isActuallyPinned = pinnedAgents.some((a) => a.id === agent.id);
+  // Display state: use prop if provided to prevent flicker
+  const pinned = pinnedProp ?? isActuallyPinned;
```
nmgarza5: I don't like that I have to use both of these `pinnedProp` and `isActuallyPinned`. `pinnedProp` is needed to prevent the icon flicker when unpinning an agent and I haven't been able to figure out a way around that. 


The prop solves this

web/src/sections/sidebar/AppSidebar.tsx
```diff
@@ -495,6 +495,7 @@ const MemoizedAppSidebarInner = memo(
                         <AgentButton
                           key={visibleAgent.id}
                           agent={visibleAgent}
+                          pinned
```
raunakab: `AgentButton` is only ever called and used in one place. Setting `pinned={true}` essentially is hardcoding the value of `pinned`.

Are we sure we need to expose this?

web/src/sections/sidebar/AgentButton.tsx
raunakab: `pinned: pinnedProp` always resolves to `true`, since `AgentButton` is only used in one place, and `pinned` is ALWAYS set unconditionally at that invocation site.

Technically, this will always statically evaluate to `const pinned = true ?? isActuallyPinned`, no?

web/src/sections/sidebar/AppSidebar.tsx
nmgarza5: yes, this prevents the icon flicker

nmgarza5: it has to do with the component unmounting lifecycle. the icon state updates before we unmount if i dont use the prop causing the flicker

The Flicker Sequence (without the prop):

  1. User clicks X to unpin an agent
  2. togglePinnedAgent(agent, false) is called
  3. Optimistic update: setLocalPinnedAgents removes the agent from the local state
  4. pinnedAgents updates → pinned = pinnedAgents.some(...) becomes false
  5. React re-renders: icon changes from SvgX to SvgPin (the flicker!)
  6. visibleAgents recalculates, no longer includes this agent
  7. Component unmounts

  The problem is step 5 happens before step 7. The component re-renders with pinned=false (showing Pin icon) for a brief moment before it unmounts.

  The Fix:

  By passing pinned={true} as a prop (or defaulting to true), the display state is decoupled from the hook state. The icon stays as SvgX throughout the unmount process because it doesn't re-derive from pinnedAgents.

  We still need isActuallyPinned from the hook for the logic in handleClick (to know whether to call togglePinnedAgent), but we use the prop for display.

raunakab: Ahhh okay. I'm fine with this then!

PR #7245, opened 2026-01-07
fix(files): propagate file error from backend

Danelegend:
The backend was separating files into 3 separate lists, sending those lists to the frontend where the frontend would then infer the error from the list the file is in. This is not the most scalable for adding differentiating errors. This PR provides a different mechanism in order to assist the solving of a follow-up linear issue.

## Description
Moves from a inferred error model to a backend defined error model for file errors. Introduces 'RejectedFile' model which is a wrapper of a filename + reason. Files that are not accepted are denoted as rejected, and propagated up to the frontend.

The current downside is that errors are not collapsed. For example File type not accepted: jpg, png vs File type not accepted: jpg | File type not accepted: png. This could be fixed with later effort however

## How Has This Been Tested?
Manually tested. Near functionally the same

## Additional Options
https://linear.app/onyx-app/issue/ENG-3187/unclear-user-fileproject-file-error
part 1 of 2

---
## Summary by cubic
Return explicit file rejection reasons from the backend and surface them in the UI. Users now see clear, distinct upload errors instead of inferred lists.

- **Refactors**
  - Replace non_accepted_files and unsupported_files with rejected_files using a RejectedFile {file_name, reason} model.
  - Centralize reasons in categorize_uploaded_files (unsupported type, >100k tokens, unreadable, processing failure).
  - Update API types and frontend (ProjectsContext, projectsService) to show unique error messages and map failed optimistic uploads by temp_id.
  - Improve server logs to include filename and rejection reason.

<sup>Written for commit 257b0f6fe55ffbc08785e1d05ea7ca3d982955c1. Summary will update on new commits.</sup>

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -11,7 +11,6 @@ on:
 permissions: {}
 
 env:
-  IS_DRY_RUN: ${{ github.event_name == 'workflow_dispatch' }}
   EDGE_TAG: ${{ startsWith(github.ref_name, 'nightly-latest') }}
 
 jobs:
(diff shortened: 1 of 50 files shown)


Danelegend:
@greptileai

backend/onyx/server/features/projects/models.py
```diff
@@ -52,10 +52,14 @@ class TokenCountResponse(BaseModel):
     total_tokens: int
 
 
+class RejectedFile(BaseModel):
```
acaprau, line 55: any reason to have what is essentially duplicated models here and in projects file utils?

Danelegend: Rationale is that this one is exposed to the client while the other is more internal to the system. Just seems to be more explicit and follows the pattern that seems to be here

Danelegend:
> lgtm, im curious is there an intended outcome in the frontend or something that this supports, or this this more of a hygiene thing?

Frontend wants to display more meaningful errors to users

PR #7261, opened 2026-01-07
chore: agent pin behavior

yuhongsun96:
## Description
When chatting with an agent or selecting it from the agent nav page, it should pin.

Addresses https://linear.app/onyx-app/issue/ENG-3158/starting-a-chat-with-an-unpinned-agent-from-the-list-should-pin-it-at

## How Has This Been Tested?
Works as expected locally

## Additional Options

---
## Summary by cubic
Automatically pin agents when starting a chat or sending a message, so unpinned agents appear in the sidebar. Satisfies Linear ENG-3158.

- **New Features**
  - AgentCard: Start Chat pins the agent if not already pinned, then opens chat.
  - useChatController: Sending a message auto-pins the current agent if not already pinned.

<sup>Written for commit d856b6e1c26bd6c178046bea040ee317fce2c31b. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/app/chat/hooks/useChatController.ts
+++ b/web/src/app/chat/hooks/useChatController.ts
@@ -450,7 +450,9 @@ export function useChatController({
           (agent) => agent.id === liveAssistant.id
         );
         if (!isAlreadyPinned) {
-          togglePinnedAgent(liveAssistant, true);
+          togglePinnedAgent(liveAssistant, true).catch((err) => {
+            console.error("Failed to auto-pin agent:", err);
+          });
         }
       }
 


web/src/refresh-components/AgentCard.tsx
```diff
@@ -59,6 +59,14 @@ export default function AgentCard({ agent }: AgentCardProps) {
...
   const [hovered, setHovered] = React.useState(false);
 
+  // Start chat and auto-pin unpinned agents to the sidebar
+  const handleStartChat = useCallback(() => {
+    if (!pinned) {
+      togglePinnedAgent(agent, true);
+    }
+    route({ agentId: agent.id });
+  }, [pinned, togglePinnedAgent, agent, route]);
+
```
Subash-Mohan, line 69: Why do we need to pin the agent in two places? I think pinning the agent only when we send a message (`useChatController`) is enough. Sometimes a user may click on the wrong agent, and they definitely wouldn’t want that to be pinned.
What do you think? Or am I missing any edge cases?

web/src/app/chat/hooks/useChatController.ts
```diff
@@ -442,6 +444,16 @@ export function useChatController({
         return;
       }
 
+      // Auto-pin the agent to sidebar when sending a message if not already pinned
+      if (liveAssistant) {
+        const isAlreadyPinned = pinnedAgents.some(
+          (agent) => agent.id === liveAssistant.id
+        );
+        if (!isAlreadyPinned) {
+          togglePinnedAgent(liveAssistant, true);
```
Subash-Mohan: I think we should catch the error from togglePinnedAgent, if it fails here, it may breaks the send button. 
example:
```tsx
togglePinnedAgent(liveAssistant, true).catch((err) => {
  console.error("Failed to auto-pin agent:", err);
});

web/src/refresh-components/AgentCard.tsx
yuhongsun96: If they start a chat with an Agent, I think it's ok to pin, the other check enforces flows where the user is directed to the chat without having interacted with it. One example is when they finalize the agent creation. In that situation, if they send a message, that's also enough signal.

PR #7265, opened 2026-01-07
fix(files): Display protected file errors

Danelegend:
## Description
We want to be able to tell the user when a file is rejected because it is a protected file.

This PR implements checking for common files that can be protected, and determines if they are protected. The categorization pipeline has also been updated to check if a file is protected, and if so, error it back up to the user

## How Has This Been Tested?

Uploaded protected .docx, .pptx and .pdf files, and checked that the appropriate error appeared.

Microsoft handles .docx, .pptx and .xlsx the same so there is an assumption that this works for excel.
## Additional Options
Closes https://linear.app/onyx-app/issue/ENG-3187/unclear-user-fileproject-file-error

---
## Summary by cubic
Show a clear error when users upload password-protected documents. Detects protected PDF, DOCX, PPTX, and XLSX files during categorization and rejects them with a specific message.

- **Bug Fixes**
  - Detect and reject password-protected PDFs and Office docs; preserve file cursor and log a warning.
  - Addresses Linear ENG-3187 by replacing the unclear error with a specific, user-facing message.

<sup>Written for commit ed40e8eb78302e4e830d6f77485de4a1131c8648. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/file_processing/password_validation.py
+++ b/backend/onyx/file_processing/password_validation.py
@@ -1,4 +1,5 @@
 from collections.abc import Callable
+from collections.abc import Generator
 from contextlib import contextmanager
 from typing import Any
 from typing import IO
@@ -17,7 +18,7 @@ PASSWORD_PROTECTED_FILES = [
 
 
 @contextmanager
-def preserve_position(file: IO[Any]):
+def preserve_position(file: IO[Any]) -> Generator[IO[Any], None, None]:
     """Preserves the file's cursor position"""
     pos = file.tell()
     try:
@@ -49,7 +50,7 @@ def is_xlsx_protected(file: IO[Any]) -> bool:
 
 
 def is_office_file_protected(file: IO[Any]) -> bool:
-    import msoffcrypto
+    import msoffcrypto  # type: ignore[import-untyped]
 
     with preserve_position(file):
         office = msoffcrypto.OfficeFile(file)


backend/onyx/server/features/projects/projects_file_utils.py
```diff
@@ -187,6 +188,19 @@ def categorize_uploaded_files(files: list[UploadFile]) -> CategorizedFiles:
...
+                    file_name=filename,
+                    extension=extension,
+                ):
+                    logger.warning(f"{filename} is password protected")
+                    results.rejected.append(
+                        RejectedFile(
+                            filename=filename, reason="Document is password protected"
+                        )
+                    )
+                    continue
```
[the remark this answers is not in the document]
Danelegend: That seems reasonable to give off failed to process upload errors in those cases

backend/onyx/file_processing/password_validation.py
```diff
@@ -0,0 +1,87 @@
...
+def is_pptx_protected(file: IO[Any]) -> bool:
+    return is_office_file_protected(file)
+
+
+def is_xlsx_protected(file: IO[Any]) -> bool:
+    return is_office_file_protected(file)
+
+
+def is_office_file_protected(file: IO[Any]) -> bool:
+    import msoffcrypto  # type: ignore[import-untyped]
```
evan-onyx, line 53: if this is a new lib we'll need the uv.lock, requirements.txt, and pyproject.toml

backend/onyx/server/features/projects/projects_file_utils.py
```diff
@@ -187,6 +188,19 @@ def categorize_uploaded_files(files: list[UploadFile]) -> CategorizedFiles:
 
             # Otherwise, handle as text/document: extract text and count tokens
             elif extension in OnyxFileExtensions.ALL_ALLOWED_EXTENSIONS:
+                if is_file_password_protected(
```
evan-onyx, line 191: is this feature intended specifically for Projects, i.e. not intended to touch files uploaded via connectors?

backend/onyx/file_processing/password_validation.py
Danelegend: Already been added. Spoke w/ jamison yesterday about it

backend/onyx/server/features/projects/projects_file_utils.py
Danelegend: At the moment yeah. We're just covering the case where a user uploads a password protected file.

PR #7266, opened 2026-01-07
fix(fe): fix InputComboBox shrinking when disabled

jmelahman:
## Description

Closes (part 1) of https://linear.app/onyx-app/issue/ENG-3271/minor-onboarding-nits

## Additional Options

- [x] Override Linear Check

---
## Summary by cubic
Fixes InputComboBox shrinking when disabled so the field size stays consistent in forms. Addresses ENG-3271 (onboarding nits).

- Bug Fixes
  - Keep disabled inputs the same size by using a transparent 1px border instead of removing the border.
  - Remove an extra relative wrapper around the input to prevent layout side effects.

<sup>Written for commit adac241274883da283a2abf38d65d56db774e978. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/refresh-components/inputs/InputComboBox/InputComboBox.tsx
+++ b/web/src/refresh-components/inputs/InputComboBox/InputComboBox.tsx
@@ -364,7 +364,7 @@ const InputComboBox = ({
 
   return (
     <div ref={refs.setReference} className={cn("relative w-full", className)}>
-      <div className="relative">
+      <>
         <InputTypeIn
           ref={inputRef}
           placeholder={placeholder}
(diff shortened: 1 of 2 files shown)


web/src/refresh-components/inputs/InputComboBox/InputComboBox.tsx
```diff
@@ -364,7 +364,7 @@ const InputComboBox = ({
 
   return (
     <div ref={refs.setReference} className={cn("relative w-full", className)}>
-      <div className="relative">
+      <>
```
jmelahman, line 367: FYI, this isn't strictly necessary, but I don't think it's doing anything, so I just removed.

PR #7270, opened 2026-01-07
fix(web crawler): Fixing decoding bytes issue

justin-tahara:
## Description

There is some formatting issues when parsing some websites that get decoded into bytes. This PR aims to fix this properly

## How Has This Been Tested?

Tested it by asking about certain websites that had historically return bytes instead of actual text.

Also added a simple test to prevent regression

## Additional Options
Closes https://linear.app/onyx-app/issue/ENG-3256/some-websites-are-parsed-like-this

---
## Summary by cubic
Fix HTML decoding in the web crawler so pages served as raw bytes are correctly converted to text. Prevents garbled output and missing titles/text on affected sites and addresses ENG-3256.

- Bug Fixes
  - Added decode_html_bytes utility (uses UnicodeDammit), honoring Content-Type charset and response encodings with a safe fallback.
  - Crawler now decodes response.content before cleanup, fixing sites previously parsed as bytes.

<sup>Written for commit a105f202d41f8689c9a0336927bf086c96df52dc. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/utils/web_content.py
+++ b/backend/onyx/utils/web_content.py
@@ -4,7 +4,7 @@ import io
 from urllib.parse import unquote
 from urllib.parse import urlparse
 
-from bs4 import UnicodeDammit
+from bs4.dammit import UnicodeDammit
 
 from onyx.file_processing.extract_file_text import read_pdf_file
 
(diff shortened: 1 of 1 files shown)


jessicasingh7 (approved):
lgtm!- you might want to check the failing mypy error

backend/onyx/utils/web_content.py
```diff
@@ -16,6 +18,37 @@
...
+    fallback_encoding: str | None = None,
+) -> str:
+    override_encodings: list[str] = []
+    charset = _charset_from_content_type(content_type)
+    if charset:
+        override_encodings.append(charset)
+    if fallback_encoding and fallback_encoding not in override_encodings:
+        override_encodings.append(fallback_encoding)
+
+    dammit = UnicodeDammit(content, override_encodings=override_encodings or None)
```
jessicasingh7: maybe we can rename this variable? lol

PR #7275, opened 2026-01-07
chore: limit Deep Research to sequential calls only

yuhongsun96:
## Description
Disallow DR to call multiple different tools in a single cycle, we have no way  of showing that in the UI today.

Also a small edit to a prompt

Addresses https://linear.app/onyx-app/issue/ENG-3335/enforce-deep-research-1-tool-call-at-a-time

## How Has This Been Tested?
Works locally

## Additional Options

---
## Summary by cubic
Limits Deep Research to one tool call per cycle and drops extras to match the UI, and clarifies the final report prompt to avoid tool usage. Addresses Linear ENG-3335.

- **New Features**
  - Added max_concurrent_tools to run_tool_calls; unknown tools are dropped; cap enforced and passed to max_workers.
  - Deep Research sets max_concurrent_tools=1 in research_agent; llm_loop defaults to None for other flows.
  - Prevents parallel nested calls so packets stay consistent with the UI.

<sup>Written for commit c662d45ebefc9eeb3c0a4f46000858fe61aa04b9. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/prompts/deep_research/research_agent.py
+++ b/backend/onyx/prompts/deep_research/research_agent.py
@@ -42,6 +42,8 @@ Your job is now to organize the findings to return a comprehensive report that p
 The report will be seen by another agent instead of a user so keep it free of formatting or commentary and instead focus on the facts only. \
 Do not give it a title, do not break it down into sections, and do not provide any of your own conclusions/analysis.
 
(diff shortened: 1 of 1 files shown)


Danelegend (approved):
Looks pretty good

PR #7280, opened 2026-01-08
chore: Refactor tool exceptions

yuhongsun96:
## Description

Allow tool implementations to fail with specific handling to inform the LLM of the error

## How Has This Been Tested?
Only tested positive path, seems good

## Additional Options

---
## Summary by cubic
Adds structured tool error handling with ToolCallException and a new ParallelToolCallResponse for parallel runs. Improves web search reliability by filtering empty results and giving the LLM clear, actionable error messages on failures.

- **Refactors**
  - Introduced ToolCallException in tools.models with an llm_facing_message for LLM-friendly errors.
  - Added _safe_run_single_tool that handles ToolCallException and unexpected errors, always emitting SectionEnd.
  - run_tool_calls now returns ParallelToolCallResponse (tool_responses + updated_citation_mapping).
  - Merges repeated calls for Search/WebSearch/OpenURL and assigns distinct citation ranges to avoid collisions.
  - Updated chat loop and research agent to consume the new response shape.
  - WebSearchTool wraps provider calls, logs partial failures, and raises ToolCallException when all queries fail or no results.

- **Migration**
  - Update imports to use onyx.tools.models.ToolCallException.
  - Update call sites of run_tool_calls to read .tool_responses and .updated_citation_mapping instead of tuple unpacking.
  - Tool implementations should raise ToolCallException with a user-facing llm_facing_message for expected failures.

<sup>Written for commit fa7a46b9a017ae4285d0e99fb91d93f91f87b0f8. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/integration/tests/web_search/test_web_search_api.py
+++ b/backend/tests/integration/tests/web_search/test_web_search_api.py
@@ -71,10 +71,10 @@ class TestOnyxWebCrawler:
         assert response.status_code == 200, response.text
         data = response.json()
 
-        # Should return a result but with empty content
-        assert len(data["results"]) == 1
-        result = data["results"][0]
-        assert result["content"] == ""
+        assert data["provider_type"] == WebContentProviderType.ONYX_WEB_CRAWLER.value
+
+        # The API filters out docs with no title/content, so unreachable domains return no results
+        assert data["results"] == []
 
     def test_handles_404_page(self, admin_user: DATestUser) -> None:
         """Test that the crawler handles 404 responses gracefully."""
@@ -86,8 +86,10 @@ class TestOnyxWebCrawler:
         assert response.status_code == 200, response.text
         data = response.json()
 
-        # Should return a result (possibly with empty content for 404)
-        assert len(data["results"]) == 1
+        assert data["provider_type"] == WebContentProviderType.ONYX_WEB_CRAWLER.value
+
+        # Non-200 responses are treated as non-content and filtered out
+        assert data["results"] == []
 
     def test_https_url_with_path(self, admin_user: DATestUser) -> None:
         """Test that the crawler handles HTTPS URLs with paths correctly."""


backend/onyx/tools/tool_implementations/web_search/clients/exa_client.py
```diff
@@ -41,20 +40,34 @@ def search(self, query: str) -> list[WebSearchResult]:
...
-                    else None
-                ),
+        # Exa can return results without a title and/or highlight snippet.
+        # Downstream uses these for display + prompting; skip empty ones rather than
+        # returning WebSearchResult objects with empty strings.
+        valid_results: list[WebSearchResult] = []
+        for result in response.results:
+            title = (result.title or "").strip()
+            snippet = (
+                (result.highlights[0] if result.highlights else "") or ""
```
acaprau: do we need this second or ""

backend/onyx/tools/tool_implementations/web_search/clients/exa_client.py
acaprau, line 32: prob want a docstring here explaining the contract of the method, namely that it only returns "valid" results

backend/onyx/tools/tool_implementations/web_search/clients/exa_client.py
```diff
@@ -93,16 +106,27 @@ def contents(self, urls: Sequence[str]) -> list[WebContent]:
             livecrawl="preferred",
         )
 
-        return [
-            WebContent(
-                title=result.title or "",
-                link=result.url,
-                full_content=result.text or "",
-                published_date=(
-                    time_str_to_utc(result.published_date)
```
acaprau, line 102: docstring. also this function name can be improved, really it's get_contents or something like that

backend/onyx/tools/tool_implementations/web_search/clients/exa_client.py
```diff
@@ -41,20 +40,34 @@ def search(self, query: str) -> list[WebSearchResult]:
...
+        # returning WebSearchResult objects with empty strings.
+        valid_results: list[WebSearchResult] = []
+        for result in response.results:
+            title = (result.title or "").strip()
+            snippet = (
+                (result.highlights[0] if result.highlights else "") or ""
+            ).strip()
+
+            if not title and not snippet:
+                continue
```
acaprau: i would log this

backend/onyx/tools/tool_implementations/web_search/clients/exa_client.py
```diff
@@ -93,16 +106,27 @@ def contents(self, urls: Sequence[str]) -> list[WebContent]:
...
-                ),
+        # Exa can return partial/empty content entries; skip those to avoid
+        # downstream prompt + UI pollution.
+        valid_contents: list[WebContent] = []
+        for result in response.results:
+            title = (result.title or "").strip()
+            full_content = (result.text or "").strip()
+
+            if not title and not full_content:
+                continue
```
acaprau: same etc

backend/onyx/tools/tool_implementations/web_search/clients/exa_client.py
```diff
@@ -41,20 +40,34 @@ def search(self, query: str) -> list[WebSearchResult]:
...
+                        time_str_to_utc(result.published_date)
+                        if result.published_date
+                        else None
+                    ),
+                )
             )
-            for result in response.results
-        ]
+
+        return valid_results
```
acaprau: if callers are interested in which of their queries failed you can also return a list of failed results which contain reasons or something like that. then callers can decide what to do. this all falls under the "it depends" paradigm, idk enough about tool calls of this makes sense

backend/onyx/tools/tool_implementations/web_search/web_search_tool.py
```diff
@@ -149,22 +162,46 @@ def run(
...
+        search_results_with_errors: list[
+            tuple[list[WebSearchResult] | None, str | None]
+        ] = run_functions_tuples_in_parallel(
+            functions_with_args,
+            allow_failures=False,  # Our wrapper handles errors internally
         )
 
+        # Separate successful results from failures
+        valid_results: list[list[WebSearchResult]] = []
+        failed_queries: dict[str, str] = {}
```
acaprau, line 186: ya so in reference to one of my above comments, we already track failed queries. if callers would find this interesting we could return this too

backend/onyx/tools/tool_implementations/web_search/clients/exa_client.py
```diff
@@ -41,20 +40,34 @@ def search(self, query: str) -> list[WebSearchResult]:
...
+        # Downstream uses these for display + prompting; skip empty ones rather than
+        # returning WebSearchResult objects with empty strings.
+        valid_results: list[WebSearchResult] = []
+        for result in response.results:
+            title = (result.title or "").strip()
+            snippet = (
+                (result.highlights[0] if result.highlights else "") or ""
+            ).strip()
+
+            if not title and not snippet:
```
Danelegend: With web tools, is it worth passing everything back as is (ie. if title is not present, leave as None) and let the higher level deal setting defaults and/or skipping rather than keeping this in the individual tool layer? 

Basically keeping the client implementations dumb and letting the manager manage

backend/onyx/tools/tool_implementations/web_search/clients/exa_client.py
yuhongsun96: tru, must be LLM code

backend/onyx/tools/tool_implementations/web_search/clients/exa_client.py
yuhongsun96: I think it depends on what you think each layer is responsible for. In this case I feel either is valid. Client might be responsible for giving back valid objects.

yuhongsun96: Probably makes sense to move it up the stack though given that many clients probably have the same invalid check

backend/onyx/tools/tool_implementations/web_search/clients/exa_client.py
yuhongsun96: Ya, I totally agree, should give a list of failures or something even if we allow partial successes. I am however lazy :(

backend/onyx/tools/tool_implementations/web_search/clients/exa_client.py
yuhongsun96: Is this a comment on the time_str_to_utc func?

backend/onyx/tools/tool_implementations/web_search/web_search_tool.py
yuhongsun96: makes sense

Danelegend (approved):
lgtm

PR #7281, opened 2026-01-08
fix(layout): absolute chat header for large screens

Subash-Mohan:
## Description
This pull request introduces the absolute chat headers for large screens which allows the messages to reach top of the view port.
ticket -> https://linear.app/onyx-app/issue/ENG-3240/no-top-header

## How Has This Been Tested?
Test from UI
https://github.com/user-attachments/assets/b2cdc339-baa5-46f4-9aeb-14eb84854172

## Additional Options

---
## Summary by cubic
Make the chat header absolute on large (2xl) screens so messages reach the top of the viewport. Addresses Linear ENG-3240 (no top header).

- **Bug Fixes**
  - Added --z-chat-header token and .z-chat-header utility.
  - Made AppHeader 2xl:absolute with transparent background and z-chat-header when no custom header.
  - Set AppRoot to 2xl:relative to anchor the header.

<sup>Written for commit dc45863d67a913e3404dd8625a29a59e8fc3c53b. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/layouts/app-layouts.tsx
+++ b/web/src/layouts/app-layouts.tsx
@@ -305,7 +305,13 @@ function AppHeader() {
       )}
 
       {(isMobile || customHeaderContent || currentChatSessionId) && (
-        <header className="w-full flex flex-row justify-center items-center py-3 px-4 h-16">
+        <header
+          className={cn(
+            "w-full flex flex-row justify-center items-center py-3 px-4 h-16",
+            !customHeaderContent &&
+              "2xl:absolute 2xl:top-0 2xl:left-0 2xl:right-0 2xl:bg-transparent 2xl:z-chat-header"
+          )}
+        >
           {/* Left - contains the icon-button to fold the AppSidebar on mobile */}
           <div className="flex-1">
             <IconButton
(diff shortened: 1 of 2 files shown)


web/src/layouts/app-layouts.tsx
```diff
@@ -305,7 +305,13 @@ function AppHeader() {
       )}
 
       {(isMobile || customHeaderContent || currentChatSessionId) && (
-        <header className="w-full flex flex-row justify-center items-center py-3 px-4 h-16">
+        <header
+          className={cn(
+            "w-full flex flex-row justify-center items-center py-3 px-4 h-16",
+            !customHeaderContent &&
```
jmelahman, line 311: Is it possible to instead remove the background color? I don't love the idea of conditionally changing the positioning

jmelahman: Also, it's probably negligible, but painting is the most performance intensive styling, so removing them when possible is nice

Subash-Mohan: Initially, I thought the same. However, the header is occupying space in the DOM, and the chats end before the header. Removing the header’s background doesn’t help either, as the header doesn’t have a background in the first place.

jmelahman: Ya, I noticed that afterwards (I was confused because you're setting a transparent one below, so I assumed you were overriding something).

I think using positioning is a bit of a hack (also, the first message should probably respect that whitespace i.e. we want it to take up whitespace even on large screens at least initially, closer in behavior to `sticky`). I think the proper solution here will be to make the top `overflow` of the chat content to be `visible`. I'm not super strong with `overflow`, but I can try to help.

Subash-Mohan: Thanks for the insight. Let me try the sticky approach.

PR #7289, opened 2026-01-08
fix: auto-pause

Weves:
## Description

Previously, once a connector had 5 failures in a row any attempt to unpause it and re-run indexing would be thwarted by it immediately pausing again.

## How Has This Been Tested?

eyes

## Additional Options

---
## Summary by cubic
Prevent connectors from auto-pausing immediately after a user unpauses them. We now only mark repeated error and auto-pause (Cloud only) when entering that state, so manual restarts don’t get re-paused.

- **Bug Fixes**
  - Skip setting repeated error and auto-pause if the connector is already flagged, allowing unpaused connectors to run.
  - Apply auto-pause on entering repeated error state only for Cloud; self-hosted is unaffected.

<sup>Written for commit b39f22b60bf59f4a8932e0ba5856451fb39b8688. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/background/celery/tasks/docprocessing/tasks.py
+++ b/backend/onyx/background/celery/tasks/docprocessing/tasks.py
@@ -64,6 +64,7 @@ from onyx.db.connector_credential_pair import (
 )
 from onyx.db.connector_credential_pair import get_connector_credential_pair_from_id
 from onyx.db.connector_credential_pair import set_cc_pair_repeated_error_state
+from onyx.db.connector_credential_pair import update_connector_credential_pair_from_id
 from onyx.db.engine.sql_engine import get_session_with_current_tenant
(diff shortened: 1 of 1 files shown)


backend/onyx/background/celery/tasks/docprocessing/utils.py
```diff
@@ -126,18 +125,9 @@ def progress(self, tag: str, amount: int) -> None:
 
 
 def is_in_repeated_error_state(
-    cc_pair_id: int, search_settings_id: int, db_session: Session
+    cc_pair: ConnectorCredentialPair, search_settings_id: int, db_session: Session
```
Weves, line 128: everywhere we used this already had a CC pair object

Weves:
@greptile re-review

justin-tahara (approved):

PR #7290, opened 2026-01-08
chore: allow tenant cleanup script to skip control plane if tenant not found

wenxi-onyx:
## Additional Options

---
## Summary by cubic
Allow tenant cleanup and connector-marking scripts to skip control plane steps when the tenant is not found in the control plane. In force mode, we continue with data plane only; otherwise we prompt.

- **New Features**
  - Detect tenant not found in control plane (missing table or record) and raise TenantNotFoundInControlPlaneError.
  - In --force, skip control-plane steps and proceed with data plane cleanup.
  - In non-force, prompt before proceeding.
  - Applied to bastion and no-bastion scripts, including mark-connectors.

- **Migration**
  - No-bastion scripts now require --data-plane-context and --control-plane-context.

<sup>Written for commit 3e7ce56fbd46d4f72ceeaabdfd66d637ea741e25. Summary will update on new commits.</sup>

### DIFF
--- a/backend/scripts/tenant_cleanup/cleanup_utils.py
+++ b/backend/scripts/tenant_cleanup/cleanup_utils.py
@@ -159,20 +159,18 @@ def get_tenant_status(tenant_id: str) -> str | None:
             return status
         else:
             print("⚠ Tenant not found in control plane")
-            return None
-
+            raise TenantNotFoundInControlPlaneError(
+                f"Tenant {tenant_id} not found in control plane database"
+            )
+    except TenantNotFoundInControlPlaneError:
+        # Re-raise without wrapping
+        raise
     except subprocess.CalledProcessError as e:
         error_msg = e.stderr if e.stderr else str(e)
         print(
             f"✗ Failed to get tenant status for {tenant_id}: {error_msg}",
             file=sys.stderr,
         )
-        # Check if this is a "not found" error (tenant table doesn't exist)
-        error_str = str(error_msg).lower()
-        if 'relation "tenant" does not exist' in error_str:
-            raise TenantNotFoundInControlPlaneError(
-                f"Tenant table/relation not found in control plane: {error_msg}"
-            )
         return None
 
 
--- a/backend/scripts/tenant_cleanup/no_bastion_cleanup_utils.py
+++ b/backend/scripts/tenant_cleanup/no_bastion_cleanup_utils.py
@@ -229,8 +229,7 @@ def get_tenant_status(pod_name: str, tenant_id: str, context: str) -> str | None
         Tenant status string (e.g., 'GATED_ACCESS', 'ACTIVE') or None if not found
 
     Raises:
-        TenantNotFoundInControlPlaneError: If the tenant table/relation does not exist,
-            or if the tenant record is not found in the table
+        TenantNotFoundInControlPlaneError: If the tenant record is not found in the table
     """
     print(f"Fetching tenant status for tenant: {tenant_id}")
 


backend/scripts/tenant_cleanup/cleanup_tenants.py
```diff
@@ -457,8 +461,25 @@ def cleanup_tenant(tenant_id: str, pod_name: str, force: bool = False) -> bool:
             if response.lower() != "yes":
                 print("Cleanup aborted - could not verify tenant status")
                 return False
+    except TenantNotFoundInControlPlaneError as e:
+        # Tenant/table not found in control plane
+        error_str = str(e)
+        print(f"⚠️  WARNING: Tenant not found in control plane: {error_str}")
```
jmelahman, line 467: this is fine, but it's probably preferred to use a logger for these so they're sent to stderr by default. slightly less likely to be missed that way

backend/scripts/tenant_cleanup/cleanup_utils.py
```diff
@@ -7,6 +7,10 @@
 from pathlib import Path
 
 
+class TenantNotFoundInControlPlaneError(Exception):
```
jmelahman, line 10: this is fine, but very nitpicky: it's nice to keep exceptions in a dedicated module. that's much more important with public interfaces though which this isn't really

backend/scripts/tenant_cleanup/cleanup_utils.py
```diff
@@ -152,15 +159,18 @@ def get_tenant_status(tenant_id: str) -> str | None:
...
+                f"Tenant {tenant_id} not found in control plane database"
+            )
+    except TenantNotFoundInControlPlaneError:
+        # Re-raise without wrapping
+        raise
     except subprocess.CalledProcessError as e:
+        error_msg = e.stderr if e.stderr else str(e)
         print(
-            f"✗ Failed to get tenant status for {tenant_id}: {e}",
+            f"✗ Failed to get tenant status for {tenant_id}: {error_msg}",
```
jmelahman, line 171: nit: while you're here wanna convert to an emoji for visibility (ideally just a colored logger imo but im boring)

backend/scripts/tenant_cleanup/no_bastion_mark_connectors.py
```diff
@@ -205,45 +217,43 @@ def main() -> None:
     if len(sys.argv) < 2:
         print(
             "Usage: PYTHONPATH=. python scripts/tenant_cleanup/"
-            "no_bastion_mark_connectors.py <tenant_id> [--force] [--concurrency N]"
+            "no_bastion_mark_connectors.py <tenant_id> \\"
```
jmelahman, line 220: i know you're not introducing these prints, but the python argparser is pretty extensible and supports custom long messages. this is like a decades old pattern left over from shell scripting

backend/scripts/tenant_cleanup/no_bastion_cleanup_utils.py
```diff
@@ -10,19 +10,19 @@
...
     """
-    print(
-        f"Finding user file processing worker pod{f' in context {context}' if context else ''}..."
-    )
+    print(f"Finding user file processing worker pod in context {context}...")
 
-    cmd = ["kubectl", "get", "po"]
-    if context:
-        cmd.extend(["--context", context])
+    cmd = ["kubectl", "get", "po", "--context", context]
```
jmelahman, line 25: optional nitpick, but this would be so much better with a KubectlHelper class that stored the `"kubectl"` (preferably as `/usr/bin/kubectl`) command and `context` as private variables.

backend/scripts/tenant_cleanup/no_bastion_cleanup_tenants.py
```diff
@@ -650,6 +654,21 @@ def main() -> None:
         except ValueError:
             pass
 
+    # Validate required contexts
+    if not data_plane_context:
+        print(
+            "Error: --data-plane-context is required",
+            file=sys.stderr,
+        )
+        sys.exit(1)
```
jmelahman, line 663: nit: i personally prefer `raise SystemExit(1)`

backend/scripts/tenant_cleanup/no_bastion_cleanup_tenants.py
```diff
@@ -528,8 +527,14 @@ def cleanup_tenant(
     else:
         print("Step 2 skipped by user")
 
-    # Step 3: Clean up control plane
-    if confirm_step(
+    # Step 3: Clean up control plane (skip if tenant not found in control plane with --force)
+    if tenant_not_found_in_control_plane:
+        print(f"\n{'=' * 80}")
```
jmelahman, line 532: nit: I feel like this would be safer without an f-string

backend/scripts/tenant_cleanup/no_bastion_cleanup_tenants.py
```diff
@@ -620,7 +624,7 @@ def main() -> None:
         )
         sys.exit(1)
 
-    # Parse contexts
+    # Parse contexts (required)
```
jmelahman, line 627: nitttt: these really should be required by the argparser and the corresponding types should not be optional

backend/scripts/tenant_cleanup/no_bastion_cleanup_tenants.py
```diff
@@ -41,12 +44,12 @@ def signal_handler(signum: int, frame: object) -> None:
     sys.exit(1)
 
 
-def setup_scripts_on_pod(pod_name: str, context: str | None = None) -> None:
+def setup_scripts_on_pod(pod_name: str, context: str) -> None:
```
jmelahman, line 47: its hard to tell if this was required for this change, but would have been really nice as a separate change so the behavioral changes related to skipping the control plane were more prominent

jmelahman (approved):
many small nits and I think they're all optional. Please do at least read them though

backend/scripts/tenant_cleanup/no_bastion_cleanup_tenants.py
wenxi-onyx: This was required. The intention of this script is to do actions in both planes, but the script would attempt to access control plane via data plane pods (which is impossible).

PR #7308, opened 2026-01-09
refactor: Components updates

raunakab:
## Description

---
## Summary by cubic
Standardized the Modal API and styling across the app and aligned shared components for consistent spacing and scroll behavior. Removed ad‑hoc classNames in consumers and simplified the UserFilesModal search experience.

- **Refactors**
  - Modal.Header/Body/Footer now use Section layout props (padding, gap, alignment); remove className overrides.
  - Modal.Body handles overflow automatically for fixed-height sizes and supports twoTone backgrounds (default on).
  - Modal.Footer rebuilt on Section with consistent spacing; all footers updated to rely on defaults.
  - Section now supports justifyContent="between" and is ref-forwarded.
  - Cleaned up z-index tokens (removed z-confirmation) and unified modal sizing/overflow; TextView uses large size.
  - UserFilesModal: replaced useFilter with local search state, simplified layout with Section, and tightened empty state.
  - Separator fills its axis (w-full horizontal, h-full vertical) for consistent dividers.
  - ShadowDiv wrapper uses h-full for accurate scroll shadows in flexible layouts.
  - InputSelect viewport uses flex column with gap; added InputSelect.Separator built on Separator with sensible defaults.

- **Migration**
  - Do not pass className to Modal.Header/Body/Footer; use Section props (padding, gap, flexDirection, justifyContent, alignItems).
  - Rely on Modal for overflow and sizing; remove local max-h/px/p- classes previously applied to Body/Content/Footer.
  - Use twoTone={false} on Modal.Body when a flat background is desired.
  - Remove any z-confirmation class/variable usage.
  - Use InputSelect.Separator for dropdown dividers instead of custom elements.

<sup>Written for commit 9604076fc023ef52cde8d45ff8240ab904f5d142. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/refresh-components/ShadowDiv.tsx
+++ b/web/src/refresh-components/ShadowDiv.tsx
@@ -105,7 +105,7 @@ export default function ShadowDiv({
   }, [containerRef, checkScroll]);
 
   return (
-    <div className="relative h-full">
+    <div className="relative">
       <div
         ref={containerRef}
         className={cn("overflow-y-auto", className)}


web/src/refresh-components/ShadowDiv.tsx
```diff
@@ -105,7 +105,7 @@ export default function ShadowDiv({
   }, [containerRef, checkScroll]);
 
   return (
-    <div className="relative">
+    <div className="relative h-full">
```
jmelahman: Maybe an example would be helpful, but this seems wrong as a default? Feel like `ShadowDiv` should behave like a `div`.

jmelahman: Or maybe rename the component to make it clear this has default layout properties

raunakab: Hmm, maybe @nmgarza5 can update the name? He was the one who originally named it.

jmelahman: Is this change load-bearing? Can we follow up with this change? It's a functional change and realistically does not belong in a refactor. I also think it's incorrect, so we should resolve this before merge imo.

raunakab:
Actually, @jmelahman, I can just revert the changes to `ShadowDiv`. They're not particularly important for my changes.

PR #7311, opened 2026-01-09
fix(chat): wide tables can be scrolled

jmelahman:
## Description

Tables can now be scrolled when they overflow (especially common on mobile) while having their content maximally visible w/ default positioning inline with container content. There is no scroll when tables fit the container.

Unfortunately, this was unrelated to https://linear.app/onyx-app/issue/ENG-3361/no-lower-bound-for-scroll-behavior

## How Has This Been Tested?

20260109_08h43m50s_grim

20260109_08h43m55s_grim

20260109_08h50m18s_grim

## Additional Options

- [x] Override Linear Check


jmelahman:
@greptile

jmelahman:
@greptile

PR #7313, opened 2026-01-09
feat(posthog): track message origin analytics in posthog

rohoswagger:
## Description

add message origin analytics in posthog for all chat messages

## How Has This Been Tested?

in posthog

## Additional Options

---
## Summary by cubic
Track the origin of every chat message (web app, Chrome extension, API, Slack) and send analytics to PostHog. Improves visibility into usage across surfaces without changing user behavior.

- **New Features**
  - Added MessageOrigin enum and threaded it through SendMessageRequest/CreateChatMessageRequest.
  - Set origin automatically per entry point: webapp (default), chrome_extension (UI), api (server APIs), slackbot (Slack handler), unknown (fallback). Force API origin when authenticated via API key or PAT.
  - Backend emits user_message_sent with origin, has_files, has_project, has_persona, deep_research, and tenant_id.
  - Frontend (extension) emits extension_chat_query with extension_context, assistant_id, has_files, and deep_research.
  - Safe telemetry loading via versioned fetch with noop fallback to avoid runtime errors when telemetry is unavailable.

<sup>Written for commit 3ef78fe22bb03fd5df738dab1f52fe3c7c5bba4a. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/server/query_and_chat/models.py
+++ b/backend/onyx/server/query_and_chat/models.py
@@ -45,6 +45,7 @@ class MessageOrigin(str, Enum):
     API = "api"
     SLACKBOT = "slackbot"
     UNKNOWN = "unknown"
+    UNSET = "unset"
 
 
 if TYPE_CHECKING:
@@ -105,8 +106,7 @@ class SendMessageRequest(BaseModel):
     deep_research: bool = False
 
     # Origin of the message for telemetry tracking
-    # Defaults to UNKNOWN - callers should set this explicitly
-    origin: MessageOrigin = MessageOrigin.UNKNOWN
+    origin: MessageOrigin = MessageOrigin.UNSET
 
     # Placement information for the message in the conversation tree:
     # - -1: auto-place after latest message in chain


rohoswagger:
@greptile

backend/onyx/server/query_and_chat/models.py
```diff
@@ -93,6 +104,10 @@ class SendMessageRequest(BaseModel):
 
     deep_research: bool = False
 
+    # Origin of the message for telemetry tracking
+    # Defaults to UNKNOWN - callers should set this explicitly
+    origin: MessageOrigin = MessageOrigin.UNKNOWN
```
wenxi-onyx: Thoughts on making this `MesageOrigin.NOT_SET` instead of UNKNOWN? 

Just so that, if we see this value, we know that we missed a spot to set the origin and maybe should go and set it. 

Your call, not committed to this idea

web/src/app/chat/hooks/useChatController.ts
```diff
@@ -132,6 +134,7 @@ export function useChatController({
   const { forcedToolIds } = useForcedTools();
   const { fetchProjects, setCurrentMessageFiles, beginUpload } =
     useProjectsContext();
+  const posthog = usePostHog();
```
wenxi-onyx, line 137: Does it make sense to only do this if we know it's the chrome extension (since we only emit for chrome ext)? I.e. could there be unnecessary performance impact in web app?

TBH, I'm really not sure about this, so relying on you to decide :)

wenxi-onyx (approved):
Two minor comments :)

backend/onyx/server/query_and_chat/models.py
rohoswagger: i like this, added!

web/src/app/chat/hooks/useChatController.ts
rohoswagger: idt the performance impact would be measurable, our chat is not super quick anyways LOL

PR #7315, opened 2026-01-09
fix: always start onboarding at first step and track by user

wenxi-onyx:
## Description

- Onboarding completion stored in browser --> can pollute if more than one user logs in on the same browser
- Onboarding always skips to final step if LLM providers available
    - On cloud, all new users get shunted to the final step even though they have completed step 1 (name)

## How Has This Been Tested?

https://github.com/user-attachments/assets/369edcc9-84cb-466f-9a78-fc1b89f97329

## Additional Options

---
## Summary by cubic
Ensure onboarding begins at the first incomplete step and tracks completion per user. Prevents skipped steps and cross-user state leaks in shared browsers.

- **Bug Fixes**
  - Compute the earliest incomplete step (Welcome → Name → LLM setup → Complete) and navigate there; no auto-skip to Complete when providers exist.
  - Pre-populate state from existing name/providers and set button state appropriately.
  - Track completion per user via a user-scoped localStorage key; wait for providers, chat sessions, and user to load before deciding.
  - Centralize finish handling by passing a finish handler through the flow instead of writing to localStorage in the header.

<sup>Written for commit 165080d41428dbf950f2c21c6df7625eea4373f5. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/hooks/useShowOnboarding.ts
+++ b/web/src/hooks/useShowOnboarding.ts
@@ -37,22 +37,24 @@ export function useShowOnboarding({
     ? `${HAS_FINISHED_ONBOARDING_KEY}_${userId}`
     : HAS_FINISHED_ONBOARDING_KEY;
 
-  // On first render, open onboarding if there are no configured LLM providers
-  // OR if the user hasn't explicitly finished onboarding yet.
-  // Wait until both providers AND chat sessions have loaded before making this decision.
-  // Skip onboarding entirely if the user has any existing chat sessions.
-  const hasCheckedOnboarding = useRef(false);
+  // Track which user we've already evaluated onboarding for.
+  // Re-check when userId changes (logout/login, account switching without full reload).
+  const hasCheckedOnboardingForUserId = useRef<string | undefined>(undefined);
+
+  // Evaluate onboarding once per user after data loads.
+  // Show onboarding if no LLM providers OR user hasn't finished onboarding.
+  // Skip entirely if user has existing chat sessions.
   useEffect(() => {
-    // Only check once, and only after both providers, chat sessions, and user have loaded
-    if (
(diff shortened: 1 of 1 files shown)


web/src/refresh-components/onboarding/useOnboardingState.ts
```diff
@@ -58,52 +58,58 @@ export function useOnboardingState(liveAssistant?: MinimalPersonaSnapshot): {
...
+    // All steps complete - go to Complete step
+    dispatch({
+      type: OnboardingActionType.SET_BUTTON_ACTIVE,
+      isButtonActive: true,
+    });
+    dispatch({
+      type: OnboardingActionType.GO_TO_STEP,
+      step: OnboardingStep.Complete,
+    });
   }, [llmProviders, isLoadingProviders]);
```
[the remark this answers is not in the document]
wenxi-onyx: don't want to auto advance as soon as the username value is modified. may result in form completion before the user finishes typing/unexpected jump

web/src/hooks/useShowOnboarding.ts
```diff
@@ -30,31 +32,39 @@ export function useShowOnboarding({
...
+  // Skip entirely if user has existing chat sessions.
   useEffect(() => {
-    // Only check once, and only after both providers and chat sessions have loaded
-    if (
-      hasCheckedOnboarding.current ||
-      isLoadingProviders ||
-      isLoadingChatSessions
-    ) {
+    // Wait for data to load
+    if (isLoadingProviders || isLoadingChatSessions || userId === undefined) {
```
Weves, line 49: will this break the flow for auth_type = disabled?

wenxi-onyx: Tested locally, no

PR #7317, opened 2026-01-09
fix: make autocomplete opt in 

nmgarza5:
## Summary

Basically make autocomplete opt in instead of opt out

- Rename `autoCompleteDisabled` to `autoCompleteEnabled` (inverted logic)
- Default autocomplete to OFF for all password-type TextFormField inputs
- Only enable autocomplete on actual user password forms (reset-password)
- Remove redundant props across admin forms

Closes https://linear.app/onyx-app/issue/ENG-3232

## Test plan
- [x] Verify API key fields don't trigger browser password autofill

---
## Summary by cubic
Prevents browser password managers from autofilling non-login fields (e.g., API keys) while keeping autocomplete on the reset-password flow. Aligns with ENG-3232: saved passwords should only appear on login-related forms.

- **Refactors**
  - Renamed TextFormField prop from autoCompleteDisabled to autoCompleteEnabled (inverted logic).
  - TextFormField now disables autocomplete by default; opt in with autoCompleteEnabled.
  - Enabled autocomplete on reset-password fields and removed redundant autocomplete props across admin forms.

<sup>Written for commit c333db8df24b0580ac41ead98a515b01b6343c2f. Summary will update on new commits.</sup>


web/src/components/Field.tsx
```diff
@@ -215,7 +215,7 @@ export function TextFormField({
   includeRevert,
   isTextArea = false,
   disabled = false,
-  autoCompleteDisabled = true,
+  autoCompleteEnabled,
```
jmelahman:
```suggestion
  autoCompleteEnabled = false,
```
nit, but I think it's helpful to readers to be explicit (and seems to match local conventions). Same comment for the other components as well

nmgarza5: makes sense. will make the change

PR #7318, opened 2026-01-09
fix(chat): Chat in-progress messages

Danelegend:
## Description
There is an on-going PR that allows a message to continue processing even if the user refreshes and the streaming session being invalid. We still want to tell the frontend that this message is in-progress so that don't think it is lost.

Currently the flow is: Create and reserve message in database w/ initial default message that writes some generic error message. Create a message string with the llm loop. Save the message string to the database.

If the client requests the chat while the llm loop is on-going, the last message they see will be that default message that currently tells the user to refresh. 

We now change that by tracking in redis whether a chat session is currently in progress. 

## How Has This Been Tested?
Manually tested

## Additional Options
closes https://linear.app/onyx-app/issue/ENG-3356/need-to-allow-chats-to-continue-if-any-of-the-below

---
## Summary by cubic
Marks in-progress chat messages across refreshes by tracking processing state in Redis and returning a clear loading message instead of the default error. Supports Linear ENG-3356 to let chats continue even if the streaming session becomes invalid.

- **Bug Fixes**
  - Added Redis-based processing fence per chat session (30 min TTL).
  - Set/clear processing status around the chat loop (finally ensures cleanup).
  - When a session is processing, override the last message to: "Message is loading... Please refresh the page soon."

<sup>Written for commit a53c221fb69bfda8628876b7f222fa2a9e09bc46. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/server/query_and_chat/chat_backend.py
+++ b/backend/onyx/server/query_and_chat/chat_backend.py
@@ -297,9 +297,9 @@ def get_chat_session(
         is_processing = is_chat_session_processing(session_id, get_redis_client())
         # Edit the last message to indicate loading (Overriding default message value)
         if is_processing and chat_message_details:
-            chat_message_details[-1].message = (
-                "Message is loading... Please refresh the page soon."
-            )
+            last_msg = chat_message_details[-1]
+            if last_msg.message_type == MessageType.ASSISTANT:
+                last_msg.message = "Message is loading... Please refresh the page soon."
     except Exception:
         logger.exception(
             "An error occurred while checking if the chat session is processing"


backend/onyx/server/query_and_chat/chat_backend.py
```diff
@@ -284,6 +285,13 @@ def get_chat_session(
         prefetch_top_two_level_tool_calls=True,
     )
 
+    is_processing = is_chat_session_processing(session_id, get_redis_client())
+    # Edit the last message to indicate loading (Overriding default message value)
```
Danelegend: Bit of a code smell here. Would probs be better to pass is_processing flag to FE and let them handle, but this works for now.

My FE knowledge seems to be the limiting factor here

backend/onyx/chat/process_message.py
```diff
@@ -528,6 +529,12 @@ def handle_stream_message_objects(
         def check_is_connected() -> bool:
             return check_stop_signal(chat_session.id, redis_client)
 
+        set_processing_status(
+            chat_session.id,
+            redis_client,
+            True,
+        )
```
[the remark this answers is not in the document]
Danelegend: Processing status is unrelated to cancel status.

Processing status is set just before the main thread begins, and is deleted once the pipeline is complete (either due to a cancel or normal finish).

Danelegend:
@greptile

backend/onyx/chat/process_message.py
```diff
@@ -530,6 +535,12 @@ def handle_stream_message_objects(
         def check_is_connected() -> bool:
             return check_stop_signal(chat_session.id, redis_client)
 
+        set_processing_status(
+            chat_session.id,
```
yuhongsun96: nit: prefer named args but doesn't matter

Danelegend:
@greptile

PR #7330, opened 2026-01-10
fix: limit web crawler memory consumption with two-pass relevance extraction

nmgarza5:
## Description

Add relevance-based content extraction to `OnyxWebCrawler` to reduce memory consumption when parsing large HTML pages.

**Why this helps memory:**
In many crawlers, peak memory is often dominated by HTML parsing rather than the final extracted text. Libraries like BeautifulSoup construct an in-memory parse tree with significant per-node overhead, which can increase memory usage by an order of magnitude depending on page structure and parser backend.

**Before:** Parse the full HTML → extract text → truncate afterward (peak memory already incurred)

**After (two-pass):**
1. Pass 1 (cheap): Decode HTML and scan for keyword hits (regex/string search)
2. Pass 2 (targeted): Parse only the surrounding regions that are likely relevant

If keywords occur in a few spots, this can reduce the parsed input from megabytes to tens of kilobytes and substantially lower peak memory in practice.

**Changes:**
- `onyx_web_crawler.py`: Two-pass extraction with keyword scanning, region merging, tag boundary snapping
- `models.py`: Add `query_context` to `OpenURLToolOverrideKwargs`
- `tool_runner.py`: Pass `last_user_message` as query context
- `open_url_tool.py`: Thread query to provider
- `WebContentProvider` interface: Add optional `query` parameter
- External providers (Firecrawl, Exa, Serper): Accept query param for interface compatibility

## How Has This Been Tested?

- Unit tested keyword extraction, position finding, region merging, and tag boundary snapping
- Manual testing with large HTML pages
- Verified fallback to simple extraction for small pages

## Additional Options

### DIFF
--- a/backend/onyx/tools/tool_implementations/open_url/onyx_web_crawler.py
+++ b/backend/onyx/tools/tool_implementations/open_url/onyx_web_crawler.py
@@ -298,8 +298,8 @@ class OnyxWebCrawler(WebContentProvider):
         # Use existing NLTK stopword removal
         keywords = remove_stop_words_and_punctuation(words)
 
-        # Filter out very short words (likely not meaningful) and lowercase
-        keywords = [w.lower() for w in keywords if len(w) > 2]
+        # Lowercase for case-insensitive matching
+        keywords = [w.lower() for w in keywords]
 
         return keywords
 


nmgarza5:
@greptile

backend/onyx/tools/tool_implementations/open_url/onyx_web_crawler.py
```diff
@@ -129,3 +150,253 @@ def _fetch_url(self, url: str) -> WebContent:
...
+        words = [w for w in words if w]  # Remove empty strings
+
+        if not words:
+            return []
+
+        # Use existing NLTK stopword removal
+        keywords = remove_stop_words_and_punctuation(words)
+
+        # Filter out very short words (likely not meaningful) and lowercase
+        keywords = [w.lower() for w in keywords if len(w) > 2]
```
evan-onyx: imo this should be handled by stopword removal, words like ID probably shouldn't be removed

backend/onyx/tools/tool_implementations/open_url/onyx_web_crawler.py
```diff
@@ -129,3 +150,253 @@ def _fetch_url(self, url: str) -> WebContent:
...
+                merged[-1] = (merged[-1][0], max(merged[-1][1], end))
+            else:
+                merged.append((start, end))
+
+        # Snap boundaries to tag edges to avoid cutting mid-tag
+        return [
+            self._snap_to_tag_boundaries(start, end, html_str) for start, end in merged
+        ]
+
+    def _snap_to_tag_boundaries(
```
evan-onyx, line 368: it's possible to end up in a situation like this:

```
// some code
      </div>
    </div>
    // start is here
  </div>
</div>
<div>
//more stuff
```

so the slice can be invalid html right?I think it's hard to make this correct without some more complicated tag matching alg that detects those situations

evan-onyx (approved):
approving to unblock, probably still an improvement even without the more complicated splitting logic (but yeah lets see if the alternative parser fixes the issues we're seeing first)

backend/onyx/tools/tool_implementations/open_url/onyx_web_crawler.py
nmgarza5: This is an accepted edge case. BeautifulSoup _should_ be able to handle this gracefully.

backend/onyx/tools/tool_implementations/open_url/onyx_web_crawler.py
nmgarza5: ah makes sense. i'll update

PR #7345, opened 2026-01-11
fix(chat): chat header should transparent in large screens

Subash-Mohan:
## Description
This pull request refactors the layout logic for chat pages to improve sticky header behavior and clarify component responsibilities. The main change is splitting the header functionality into a new `ChatHeader` component, which is now rendered inside the chat scroll container for proper sticky positioning, while the root layout only handles the footer and a mobile sidebar toggle fallback. This makes the codebase more modular and the UI more consistent across devices.

https://github.com/user-attachments/assets/e2db6a42-362d-4ab0-8f30-19073ea50e77

addresses -> https://linear.app/onyx-app/issue/ENG-3240/no-top-header
## How Has This Been Tested?
Tested from UI

## Additional Options

---
## Summary by cubic
Made the chat header sticky inside the chat scroll container and transparent on large screens to fix the missing/incorrect desktop header. Header stays solid on mobile for readability.

- **Bug Fixes**
  - Sticky header now renders correctly while scrolling.
  - Transparent header on 2xl+ screens; solid tint on mobile/tablet.
  - Addresses Linear ENG-3240 (“no top header”).

- **Refactors**
  - Added ChatHeader component and render it inside ChatUI; AppRoot no longer renders a header.
  - Removed SmallScreenSidebarFallback and simplified layout; ChatUI always renders and manages the sticky header.
  - Added StickyHeader wrapper for consistent sticky styling.

<sup>Written for commit c3e9ac0c4c26a0c2b4a41d5644311b14f28afd78. Summary will update on new commits.</sup>


web/src/layouts/app-layouts.tsx
```diff
@@ -407,37 +413,29 @@ function AppFooter() {
...
- * // The footer will show custom disclaimer (if configured)
- * ```
+ * Note: ChatHeader is rendered inside ChatUI's scroll container
+ * for sticky behavior, not in this root component.
+ */
+/**
+ * Mobile sidebar toggle shown when no chat session is active.
+ * ChatHeader handles this when a chat session exists.
  */
+function MobileSidebarFallback() {
```
jmelahman: small nit, but I would avoid using "mobile" in-place of "responsive" or "small screen". There are some truly mobile-specific settings, things around hover and lack of mouse support, which I think should be reserved for "mobile". Also, I think it's semantically more correct and easier for readers to understand the distinction (i.e. they don't ask, "I'm not on mobile, why is this behavior being applied", etc.).

web/src/layouts/app-layouts.tsx
```diff
@@ -89,7 +84,21 @@ const footerMarkdownComponents = {
...
+ *
+ * Features:
+ * - Sticky positioning within scroll container
+ * - Transparent on desktop, solid on mobile
+ * - Share chat functionality
+ * - Move chat to project (with confirmation for custom agents)
+ * - Delete chat with confirmation
+ * - Mobile-responsive sidebar toggle
+ */
+function ChatHeader() {
```
jmelahman: I feel like this should be two separate components: a `StickyHeader` which takes the implementation of `ChatHeader` as `children`. That way, we can re-use the `StickyHeader` behavior in places the SharedChat UI even though the contents are slightly different.

We should also update the SharedChat UI to use this, but I think fine to follow up to reduce the blast radius of this change. I am still trying to reason through https://linear.app/onyx-app/issue/ENG-3361/no-lower-bound-for-scroll-behavior and something like this might very easily aggravate it. That issue also seems to be specific to particular chat sessions and browsers, in the case of ENG-3361, chats with inline math text and only on Chrome, so we should QA this reasonably well.

web/src/layouts/app-layouts.tsx
```diff
@@ -448,11 +446,11 @@ function AppRoot({ children }: AppRootProps) {
...
-      <AppHeader />
+      <MobileSidebarFallback />
       <div className="flex-1 overflow-auto h-full w-full">{children}</div>
       <AppFooter />
     </div>
   );
 }
 
-export { AppRoot as Root };
+export { AppRoot as Root, ChatHeader };
```
jmelahman: @raunakab might not like this, but I personally don't have strong objections. I think it makes sense in a separate module, but maybe not.

web/src/layouts/app-layouts.tsx
jmelahman: I think this should go above the comment that precedes it? The comment still seems to be describing the `AppRoot` below. Also, some of that is probably still relevant.

Subash-Mohan: I got it, now I changed the name to `SmallScreenSidebarFallback`.

web/src/layouts/app-layouts.tsx
Subash-Mohan: Now `StickyHeader` is a wrapper component used in `ChatHeader`.

web/src/layouts/app-layouts.tsx
Subash-Mohan: Sorry, I just moved it above the comment.

Subash-Mohan: Now I removed the component itself.

web/src/layouts/app-layouts.tsx
```diff
@@ -89,282 +52,27 @@ const footerMarkdownComponents = {
...
-  }, [
-    showMoveOptions,
-    filteredProjects,
-    currentChatSession,
-    setDeleteConfirmationModalOpen,
-    handleMoveClick,
-  ]);
+interface StickyHeaderProps {
+  children: React.ReactNode;
+  className?: string;
```
jmelahman, line 57: Let's remove and replace with an `isTransparent` bool which can be toggled as needed.

web/src/sections/ChatUI.tsx
```diff
@@ -33,6 +33,8 @@ import { useUser } from "@/components/user/UserProvider";
 import { HORIZON_DISTANCE_PX } from "@/lib/constants";
 import Spacer from "@/refresh-components/Spacer";
 import { SvgChevronDown } from "@opal/icons";
+import ChatHeader from "@/app/chat/components/ChatHeader";
```
jmelahman, line 36: I feel like this belongs in the `ChatPage`? I know that might complicate the overflow, but I think we should at least try putting it there. 

That would also allow for this API which is probably the ideal,
```
<AppLayouts.Root>

  <AppLayouts.StickyHeader>
    <ChatHeaderContents>
  <AppLayouts.StickyHeader>

  <ChatInterface />

</AppLayouts.Root>
```
This

PR #7352, opened 2026-01-12
fix(tools): Tool name should align with what llm knows

Danelegend:
## Description
When reconstructing the chat history for the llm, the provided tool name is pulled from the database. Meanwhile, the llm is told the tool name from from a prompt. Currently there is a misalignment between the tool name in the database and the tool name in the prompt.

This migration ensures that tool.name is aligned with what we expect the llm to see. Although this doesn't solve the fundamental problem of needing to align database + prompt name, it solves the immediate problem. A larger followup PR can come to refactor this logic.

## How Has This Been Tested?
Run migration and ensure that the database presents us with correct tools.
Validate on braintrust that tool history is consistent with the action that it took.

## Additional Options
closes https://linear.app/onyx-app/issue/ENG-3360/tool-inconsistency

---
## Summary by cubic
Aligns tool names in the database with the names the LLM is instructed to use, fixing mismatched tool calls and history reconstruction. Addresses ENG-3360 (tool inconsistency).

- **Migration**
  - Adds Alembic migration to set tool.name to expected LLM-visible names (e.g., internal_search, web_search, research_agent); downgrade restores class-name values.
  - Clarifies Tool.name is the LLM-facing name in the model.

- **Code Updates**
  - Deep Research now looks up the Research Agent by LLM name (research_agent) and compares using in_code_tool_id (ResearchAgent).
  - Replaces RESEARCH_AGENT_DB_NAME with RESEARCH_AGENT_IN_CODE_ID where appropriate and simplifies seed migration usage.

<sup>Written for commit e38f2c215cbeddc38aac37021d563c6f93355ab2. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/deep_research/dr_loop.py
+++ b/backend/onyx/deep_research/dr_loop.py
@@ -219,35 +219,90 @@ def run_deep_research_llm_loop(
             else ""
         )
         if not skip_clarification:
-            clarification_prompt = CLARIFICATION_PROMPT.format(
-                current_datetime=get_current_llm_day_time(full_sentence=False),
-                internal_search_clarification_guidance=internal_search_clarification_guidance,
-            )
+            with function_span("clarification_step") as span:
+                clarification_prompt = CLARIFICATION_PROMPT.format(
+                    current_datetime=get_current_llm_day_time(full_sentence=False),
+                    internal_search_clarification_guidance=internal_search_clarification_guidance,
+                )
+                system_prompt = ChatMessageSimple(
+                    message=clarification_prompt,
+                    token_count=300,  # Skips the exact token count but has enough leeway
+                    message_type=MessageType.SYSTEM,
+                )
+
+                truncated_message_history = construct_message_history(
+                    system_prompt=system_prompt,
+                    custom_agent_prompt=None,
+                    simple_chat_history=simple_chat_history,
+                    reminder_message=None,
+                    project_files=None,
+                    available_tokens=available_tokens,
+                    last_n_user_messages=MAX_USER_MESSAGES_FOR_CONTEXT,
+                )
+
+                llm_step_result, _ = run_llm_step(
+                    emitter=emitter,
+                    history=truncated_message_history,
+                    tool_definitions=get_clarification_tool_definitions(),
+                    tool_choice=ToolChoiceOptions.AUTO,
+                    llm=llm,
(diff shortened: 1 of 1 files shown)


backend/alembic/versions/c1d2e3f4a5b6_add_deep_research_tool.py
```diff
@@ -19,7 +18,7 @@
 
 
 DEEP_RESEARCH_TOOL = {
-    "name": RESEARCH_AGENT_DB_NAME,
```
Danelegend, line 22: Not good to leave this as variable since if code changes (as it does here), then the migration changes. Hardcoding this value.

backend/alembic/versions/c1d2e3f4a5b6_add_deep_research_tool.py
```diff
@@ -19,7 +18,7 @@
...
+    "name": "ResearchAgent",
```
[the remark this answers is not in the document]
Danelegend: No, we want to keep it consistent with what the name was at that point in time.

backend/alembic/versions/d25168c2beee_tool_name_consistency.py
```diff
@@ -0,0 +1,87 @@
...
+# revision identifiers, used by Alembic.
+revision = "d25168c2beee"
+down_revision = "8405ca81cc83"
+branch_labels = None
+depends_on = None
+
+
+# Mapping of in_code_tool_id to the NAME constant from each tool class
+# These are the currently seeded tool names
+CURRENT_TOOL_NAME_MAPPING = {
```
evan-onyx: the keys here look wrong, should be i.e. internal_search

Danelegend: This is a mapping in_code_id -> name for how it currently looks (for the downgrade).

Basically this is just how it currently looks now, so that we can downgrade appropriately

Danelegend:
> how does get_built_in_tool_by_id work now? I'm surprised that tool calling with internal tools still works after the migration

get_built_in_tool_by_id takes in in_code_tool_id. This hasn't been changed. We're only updating the tool.name value

Danelegend:
@greptile

evan-onyx (approved):

PR #7354, opened 2026-01-12
fix(chat): correct doc selection on replay

jessicasingh7:
## Description

Only display selected documents when chat is replayed to user

[ENG-3135](https://linear.app/onyx-app/issue/ENG-3135/doc-selection-on-replay)

## How Has This Been Tested?

ran locally- verified same number of sources after replay as seen below:
image

## Additional Options

---
## Summary by cubic
Fixes chat replay to show only the LLM-selected documents, so replays match the original answer. Addresses ENG-3135 by saving the full retrieval set for the assistant message but linking only the displayed docs to each tool call, and loading those on replay.

<sup>Written for commit 7a0674743815649a5f53eeb735322efe56149119. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/chat/save_chat.py
+++ b/backend/onyx/chat/save_chat.py
@@ -222,14 +222,16 @@ def save_chat_turn(
                     )
                     search_doc_key_to_id[search_doc_key] = db_search_doc.id
 
-        # Collect displayed_docs IDs for tool call linking
+        # Collect displayed_docs IDs for tool call linking (deduplicated)
         if tool_call_info.displayed_docs:
-            displayed_ids: list[int] = []
+            displayed_ids: set[int] = set()
             for search_doc_py in tool_call_info.displayed_docs:
                 search_doc_key = _create_search_doc_key(search_doc_py)
                 if search_doc_key in search_doc_key_to_id:
-                    displayed_ids.append(search_doc_key_to_id[search_doc_key])
-            tool_call_to_displayed_doc_ids[tool_call_info.tool_call_id] = displayed_ids
+                    displayed_ids.add(search_doc_key_to_id[search_doc_key])
+            tool_call_to_displayed_doc_ids[tool_call_info.tool_call_id] = list(
+                displayed_ids
+            )
 
     # 3. Collect all unique SearchDoc IDs for ChatMessage (full set from all tool calls)
     all_search_doc_ids_set: set[int] = set(search_doc_key_to_id.values())


jessicasingh7:
@greptile

jessicasingh7:
@greptile

jessicasingh7:
@greptile

backend/onyx/chat/llm_loop.py
```diff
@@ -690,6 +692,7 @@ def run_llm_loop(
                     tool_call_arguments=tool_call.tool_args,
                     tool_call_response=saved_response,
                     search_docs=search_docs,
+                    displayed_docs=displayed_docs,
```
yuhongsun96, line 697: Why do ToolCallInfo objects need both the search docs and display docs? What is the search docs used for?

backend/onyx/chat/save_chat.py
```diff
@@ -143,10 +145,14 @@ def _create_and_link_tool_calls(
     for tool_call_obj in valid_tool_calls:
         search_doc_ids = tool_call_to_search_doc_ids.get(tool_call_obj.tool_call_id, [])
         if search_doc_ids:
+            displayed_doc_ids = tool_call_to_displayed_doc_ids.get(
```
yuhongsun96: Why is all this needed to save this info to the Tool Calls in the DB? Is it not enough to just save the final documents used to the tool call and the whole set to the final agent message?

backend/onyx/chat/llm_loop.py
jessicasingh7: search_docs are linked to ChatMessage but only displayed docs are shown in ToolCall

jessicasingh7:
@greptile

backend/onyx/chat/llm_loop.py
yuhongsun96: Does the ToolCallInfo even need the search_docs?

backend/onyx/document_index/vespa/chunk_retrieval.py
```diff
@@ -167,25 +167,22 @@ def get_chunks_via_visit_api(
...
-        and filters.access_control_list
-        and acl_fieldset_entry not in field_set_list
-    ):
-        field_set_list.append(acl_fieldset_entry)
-
-    if MULTI_TENANT:
-        tenant_id_fieldset_entry = f"{TENANT_ID}"
-        if tenant_id_fieldset_entry not in field_set_list:
-            field_set_list.append(tenant_id_fieldset_entry)
+    # NOTE: If field_names is None/empty, we want ALL fields (field_set=None).
```
yuhongsun96, line 170: Why is this stuff being changed?

jessicasingh7:
another fix was merged

PR #7364, opened 2026-01-12
fix: improve scroll behavior

nmgarza5:
## Description

Refactors chat scroll behavior into a dedicated `ChatScrollContainer` component that handles both auto-scroll ON and OFF modes.

**Key changes:**
- New `ChatScrollContainer` component encapsulates all scroll logic (MutationObserver, ResizeObserver, scroll state tracking)
- Positions anchor message at top of viewport with dynamic spacer
- Auto-scroll follows new content when user is at bottom
- Only disables auto-scroll on explicit upward user scroll (fixes issues with code blocks/tables breaking auto-scroll)
- Cross-browser `scrollTo` instead of `scrollIntoView` for Safari compatibility
- Fade overlays indicate content above/below viewport
- Hides content until scroll positioning completes to prevent flicker

**ChatInputBar:**
- Fixes scroll jump when typing by only resizing textarea when lines change

## How Has This Been Tested?

Auto Scroll On
https://github.com/user-attachments/assets/fc0c7386-f5f7-46cb-a44d-2be98fab55b6

Auto Scroll Off
https://github.com/user-attachments/assets/44b22400-4a24-4e0c-909b-eacc4826ac10

Centering ChatUI
Before:
Screenshot 2026-01-15 at 10 44 02 PM

After:
Screenshot 2026-01-15 at 10 43 49 PM

## Additional Options

- [ ] Override Linear Check

closes https://linear.app/onyx-app/issue/ENG-3353/make-scrolling-work-nicely-with-auto-scroll-off

### DIFF
--- a/web/src/sections/ChatUI.tsx
+++ b/web/src/sections/ChatUI.tsx
@@ -1,24 +1,11 @@
 "use client";
 
-import React, {
-  ForwardedRef,
-  useCallback,
-  useEffect,
-  useImperativeHandle,
-  useMemo,
-  useRef,
-  useState,
-} from "react";
+import React, { ForwardedRef, useImperativeHandle, useRef } from "react";
 import { Message } from "@/app/chat/interfaces";
-import { OnyxDocument, MinimalOnyxDocument } from "@/lib/search/interfaces";
-import HumanMessage from "@/app/chat/message/HumanMessage";
-import { ErrorBanner } from "@/app/chat/message/Resubmit";
+import { MinimalOnyxDocument } from "@/lib/search/interfaces";
 import { MinimalPersonaSnapshot } from "@/app/admin/assistants/interfaces";
 import { LlmDescriptor, LlmManager } from "@/lib/hooks";
-import AIMessage from "@/app/chat/message/messageComponents/AIMessage";
 import { ProjectFile } from "@/app/chat/projects/projectsService";
-import { useScrollonStream } from "@/app/chat/services/lib";
-import useScreenSize from "@/hooks/useScreenSize";
 import {
   useCurrentChatState,
   useCurrentMessageHistory,
@@ -28,37 +15,10 @@ import {
 } from "@/app/chat/stores/useChatSessionStore";
 import useChatSessions from "@/hooks/useChatSessions";
 import { useUser } from "@/components/user/UserProvider";
-import { cn } from "@/lib/utils";
-import Spacer from "@/refresh-components/Spacer";
-
-// Size constants (1rem = 16px)
-const MESSAGE_TOP_OFFSET_PX = 16; // 1rem - gap between viewport top and new message
-const SCROLL_DEBOUNCE_MS = 100;
-const FADE_GRADIENT_THRESHOLD_PX = 80; // 5rem - scroll distance for fade to appear
-const FADE_OVERLAY_HEIGHT = "h-8"; // 2rem
-const SCROLL_BUTTON_THRESHOLD_PX = 32; // 2rem - scroll distance for button to appear
-
-const FadeOverlay = React.memo(
-  ({ show, position }: { show: boolean; position: "top" | "bottom" }) => {
-    if (!show) return null;
-    const isTop = position === "top";
-    return (
-      <div
-        aria-label={isTop ? "Top fade overlay" : "Bottom fade overlay"}
-        className={cn(
-          `absolute left-0 right-0 ${FADE_OVERLAY_HEIGHT} z-sticky pointer-events-none`,
-          isTop ? "top-0" : "bottom-0"
-        )}
-        style={{
-          background: `linear-gradient(${
-            isTop ? "to bottom" : "to top"
-          }, var(--background-tint-01) 0%, transparent 100%)`,
-        }}
-      />
-    );
-  }
-);
-FadeOverlay.displayName = "FadeOverlay";
+import ChatScrollContainer, {
+  ChatScrollContainerHandle,
+} from "@/components/chat/ChatScrollContainer";
+import MessageList from "@/components/chat/MessageList";
 
 export interface ChatUIHandle {
   scrollToBottom: () => boolean;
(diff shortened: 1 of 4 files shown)


web/src/sections/ChatUI.tsx
```diff
@@ -34,6 +34,13 @@ import { HORIZON_DISTANCE_PX } from "@/lib/constants";
 import Spacer from "@/refresh-components/Spacer";
 import { SvgChevronDown } from "@opal/icons";
 
+// Gap between top of viewport and new message (should be >= fade gradient height)
+const MESSAGE_TOP_OFFSET = 16;
+// Debounce time for scroll-related updates (ms)
+const SCROLL_DEBOUNCE_MS = 100;
+// Scroll distance from top (px) at which fade gradient appears
+const FADE_GRADIENT_THRESHOLD = 80;
```
nmgarza5: gonna leave these constants here because they _should_ only be used in this file

nmgarza5:
@greptile

web/tests/e2e/utils/chatActions.ts
```diff
@@ -32,21 +32,24 @@ export async function navigateToAssistantInHistorySidebar(
 }
 
 export async function sendMessage(page: Page, message: string) {
+  // Count existing AI messages before sending
+  const existingMessageCount = await page
```
jmelahman, line 36: unused? missing assert?

web/src/sections/ChatUI.tsx
```diff
@@ -183,73 +280,203 @@ const ChatUI = React.memo(
...
+             NOTE: We can't use ShadowDiv here because ChatUI requires direct control
+             over the scroll container for custom scroll behavior. */}
+          <FadeOverlay show={hasContentAbove} position="top" />
+          <FadeOverlay show={hasContentBelow} position="bottom" />
+
           <div
             key={currentChatSessionId}
             ref={scrollContainerRef}
             className="flex flex-1 justify-center min-h-0 overflow-y-auto overflow-x-hidden default-scrollbar"
+            style={{ overflowAnchor: "none" }}
```
jmelahman: Is this supported by safari? Would we be worried about that? https://bugs.webkit.org/show_bug.cgi?id=171099

web/src/sections/ChatUI.tsx
```diff
@@ -183,73 +280,203 @@ const ChatUI = React.memo(
...
+              className="w-[min(50rem,100%)] px-4"
+              data-scroll-ready={isScrollReady || messages.length === 0}
+              style={{
+                // Hide content until scroll positioning is complete to prevent flicker
+                visibility:
+                  isScrollReady || messages.length === 0 ? "visible" : "hidden",
+              }}
+            >
+              {/* Top spacer ensures consistent scroll positioning for all conversations */}
+              <Spacer px={MESSAGE_TOP_OFFSET} />
```
jmelahman: `px`? Is this because it's `flex-col`? wouldn't the top_offset use `py`? consider leaving a comment.

web/src/refresh-components/Spacer.tsx
```diff
@@ -1,12 +1,20 @@
 export interface SpacerProps {
   vertical?: boolean;
   horizontal?: boolean;
+  /** Size in rem units (default: 1) */
   rem?: number;
+  /** Size in pixels - takes precedence over rem */
```
jmelahman: Can we instead make these mutually exclusive and enforced by the type-checker?

web/src/sections/ChatUI.tsx
```diff
@@ -31,9 +31,17 @@ import useChatSessions from "@/hooks/useChatSessions";
 import { useDeepResearchToggle } from "../app/chat/hooks/useDeepResearchToggle";
 import { useUser } from "@/components/user/UserProvider";
 import { HORIZON_DISTANCE_PX } from "@/lib/constants";
+import { cn } from "@/lib/utils";
 import Spacer from "@/refresh-components/Spacer";
 import { SvgChevronDown } from "@opal/icons";
 
+// Gap between top of viewport and new message (should be >= fade gradient height)
```
jmelahman: Consider moving `h-8` to a const and referring to it by name here as to not get it confused with `FADE_GRADIENT_THRESHOLD`

web/src/sections/ChatUI.tsx
```diff
@@ -183,73 +280,203 @@ const ChatUI = React.memo(
...
+        show: boolean;
+        position: "top" | "bottom";
+      }) => {
+        if (!show) return null;
+        const isTop = position === "top";
+        return (
+          <div
+            aria-label={isTop ? "Top fade overlay" : "Bottom fade overlay"}
+            className={cn(
+              "absolute left-0 right-0 h-8 z-10 pointer-events-none",
```
jmelahman: nit: prefer named a `z-index`, https://github.com/onyx-dot-app/onyx/blob/main/web/src/app/css/z-index.css (sticky is probably fine)

web/src/sections/ChatUI.tsx
```diff
@@ -99,8 +107,72 @@ const ChatUI = React.memo(
...
+              const distanceFromBottom =
+                contentEnd - (container.scrollTop + container.clientHeight);
+              // Show scroll-to-bottom button when content is below viewport
+              setAboveHorizon(distanceFromBottom > 0);
+            }
+          });
+        };
+
+        // Use MutationObserver to detect content changes during streaming (auto-scroll OFF only)
+        let mutationObserver: MutationObserver | null = null;
```
jmelahman: What's the TL;DR on using a `MutationObserver` over `useLayoutEffect`?

web/src/sections/ChatUI.tsx
jmelahman: I just realized this is pixels. Is TypeScript expressive enough that we could use a unit-agnostic variable. Something like,

size: number | `${number}px` | `${number}rem`;


Or some library which provides types for css values

web/src/app/chat/components/ChatPage.tsx
```diff
@@ -271,14 +271,14 @@ export default function ChatPage({ firstMessage }: ChatPageProps) {
...
-    },
-    [autoScrollEnabled]
-  );
+  const handleInputHeightChange = useCallback((delta: number) => {
+    if (delta > 0) {
+      // Chat bar grew - scroll to keep content visible
+      chatUiRef.current?.scrollBy(delta);
+    }
+    // When chat bar shrinks (delta < 0), no scroll needed -
+    // the flex layout gives ChatUI more space automatically
```
jmelahman: I haven't looked at this in-depth yet, but maybe briefly, what is the utility of having the ChatInputBar know about scrolling adjacent content? If I am understanding the issue correctly and maybe naively, I'd assume that problem would be resolved entirely with the layout by having the ChatInputElement simply shrink the ChatUI flexbox?

web/src/sections/ChatUI.tsx
nmgarza5: - useLayoutEffect: Runs synchronously after React renders, but only when React re-renders
 - MutationObserver: Detects ongoing DOM changes asynchronously, including streaming content additions

For streaming responses where text is continuously appended to the DOM (outside React's render cycle), MutationObserver catches these incremental changes. useLayoutEffect would miss them.

web/src/app/chat/components/ChatPage.tsx
nmgarza5: You're correct that flex layout handles the container sizing automatically. The issue is scroll position when the input bar grows: as ChatUI shrinks, content that was visible gets pushed up out of the viewport. Without the scroll adjustment, the user would lose sight of what they were reading. When the input bar shrinks, no adjustment is needed since the scroll position naturally stays correct.

Input bar grows → ChatUI shrinks → content gets pushed UP out of view → scroll DOWN to compensate
Input bar shrinks → ChatUI grows → scroll position stays fine naturally → no adjustment needed

web/src/sections/ChatUI.tsx
nmgarza5: i considered using some unit-agnostic thing but nah. rem and pixel values are too different. you can choose one or the other now

web/src/sections/ChatUI.tsx
nmgarza5: yes. that was from 2017......

jmelahman: It's unresolved lol this was linked from the MDN docs, https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/overflow-anchor which claims it has limited support for safari

nmgarza5: oops lol

idk if its that style causing the issue but safari does have unexpected behavior. going to hold off on this pr for now

web/src/sections/ChatUI.tsx
jmelahman: Ugh. I had another comment about refactoring this `ChatUI` component, but I guess it got lost on the train's flaky wifi.

The other comment went basically:

Can we refactor this such that there is a single components (`ChatUI`/`ChatContainer`) that is solely responsible for layout and overflow/scrolling? Does the ChatUI even need context of the `messages` here? At a glance, it seems like `MutationObserver` > `useLayoutEffect` > `useEffect`?

Part of the issue imo is that there are like 3 or 4 different ways which can trigger scrolling (scroll to bottom callback, chat input bar, useEffect, MutationObserver) and I was wondering if we could have a very clear separation of concerns by (mostly) removing knowledge of chat messages from the component which manages scrolling.

web/src/app/chat/components/ChatPage.tsx
jmelahman: Ugh. Are we even sure the flex shifting is desired? It almost seems like we want the ChatInputBar to be floating/sticky and never actually affecting chat sizing?

nmgarza5: posted in slack about this to get @duo-onyx's opinion. ChatGPT and Claude seem to use the sticky approach

web/src/sections/ChatUI.tsx
nmgarza5: just saw this but yeah that makes sense

jmelahman (approved):
took a brief pass. i'm happy if you're happy. If you want an in-depth pass tomorrow, let me know

nmgarza5:
> took a brief pass. i'm happy if you're happy. If you want an in-depth pass tomorrow, let me know

Refactored a tad last night to move most of the scroll logic to `ChatScrollContainer.tsx`. I'm happy with where things are for now. There are some other techniques I want to try but this is already an improvement so going to go with this for now.

Some additional things to try:
- IntersectionObservers
- virtualization (I know you dont like this one lol)

PR #7386, opened 2026-01-13
refactor(img-gen): encapsulate provider quirks

Danelegend:
## Description
Currently with image gen, various reformattings and quirks are done on the api level. This is fine for the current setup w/ 2 providers, but with more inbound, this creates difficulty and bloats the api layer with various if statements to check different provider setups.

This PR refactors the image gen by creating ImageGenerateionProviders. These providers then contain their own logic for things like building from credentials, or sending image generation requests.

## How Has This Been Tested?
All tests still pass - non-functional change.
Manually verify that the change works

## Additional Options

---
## Summary by cubic
Encapsulated image generation provider logic behind a common interface and factory, removing provider-specific conditionals from the API and tools. This makes it easier to add providers (OpenAI, Azure) and simplifies credentials handling.

- **Refactors**
  - Added ImageGenerationProvider interface, credentials model, factory, and OpenAI/Azure providers.
  - Switched API test endpoint and ImageGenerationTool to use provider abstraction.
  - Centralized credentials validation with clear errors for invalid providers or missing fields.
  - Added IMAGE_MODEL_PROVIDER env var (default: "openai").

- **Migration**
  - Set IMAGE_MODEL_PROVIDER to the chosen provider.
  - Update tool construction to pass ImageGenerationProviderCredentials and provider.
    - Azure: api_key, api_base, api_version required; deployment_name optional (falls back to model_name).
    - OpenAI: api_key required (api_base optional).

<sup>Written for commit abdd3499933eec0d574c19fb30ae7d6f996939d8. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/tools/tool_constructor.py
+++ b/backend/onyx/tools/tool_constructor.py
@@ -77,12 +77,9 @@ def _get_image_generation_config(llm: LLM, db_session: Session) -> LLMConfig:
 
     llm_provider = default_config.model_configuration.llm_provider
 
-    # For Azure, format model name as azure/<deployment_name> for LiteLLM
-    model_name = default_config.model_configuration.name
-
     return LLMConfig(
         model_provider=llm_provider.provider,
-        model_name=model_name,
+        model_name=default_config.model_configuration.name,
         temperature=GEN_AI_TEMPERATURE,
         api_key=llm_provider.api_key,
         api_base=llm_provider.api_base,


backend/onyx/server/manage/image_generation/api.py
```diff
@@ -186,56 +190,45 @@ def test_image_generation(
         if not source_provider:
             raise HTTPException(
                 status_code=404,
-                detail=f"Source LLM provider with id {test_request.source_llm_provider_id} not found",
+                details=f"Source LLM provider with id {test_request.source_llm_provider_id} not found",
             )
+
         api_key = source_provider.api_key
         provider = source_provider.provider
```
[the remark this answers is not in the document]
Danelegend: The implemented flow is we get whatever credentials we get. These are then passed into get_image_generation_provider. This will then build the appropriate provider. During the building stage, if that provider doesn't have the right credentials they throw ImageProviderCredentialsError.

The original code had the elif/else statement to handle only 2 providers. With more providers coming there are different configurations required (for example, api_key = None, custom_config = ...) that are valid.

Hence, we leave it to the builder to determine if a combination is correct

Danelegend:
@greptile

backend/onyx/server/manage/image_generation/api.py
```diff
@@ -188,54 +192,50 @@ def test_image_generation(
...
-    elif test_request.api_key is not None and test_request.provider is not None:
-        # Use directly provided credentials
-        api_key = test_request.api_key
-        provider = test_request.provider
-    else:
+
```
jessicasingh7, line 211: maybe we can add a check if api_key is None too?

Danelegend: In the current setup, both azure + openai need an api_key. But with google incoming, they don't require an api key.

On the api layer, I'm essentially assuming that we have all the credentials we need, and when we pass those into the factory, I'm letting the provider factory throw an error if there are credentials missing (and we then catch that error in the api layer).

Danelegend:
@greptile

backend/onyx/server/manage/image_generation/api.py
```diff
@@ -188,54 +205,63 @@ def test_image_generation(
...
+        api_version=test_request.api_version,
+        deployment_name=(test_request.deployment_name or test_request.model_name),
+    )
 
-            # Make a minimal image generation request using LiteLLM
-            # Use descriptive prompt to avoid Azure content policy rejection
-            image_generation(
-                prompt="a simple blue circle on white background",
-                model=model,
+    if not validate_credentials(provider, provider_credentials):
```
jessicasingh7: I think you validate_credentials downstream so this is not needed: get_image_generation_provider -> build_from_credentials

backend/onyx/image_gen/factory.py
```diff
@@ -0,0 +1,31 @@
...
+from onyx.image_gen.providers.openai_img_gen import OpenAIImageGenerationProvider
+
+PROVIDERS: dict[str, type[ImageGenerationProvider]] = {
+    AzureImageGenerationProvider.NAME: AzureImageGenerationProvider,
+    OpenAIImageGenerationProvider.NAME: OpenAIImageGenerationProvider,
+}
+
+
+def get_image_generation_provider(
+    provider: str,
```
jessicasingh7, line 21: Could we make provider an enum?

backend/onyx/image_gen/factory.py
```diff
@@ -0,0 +1,31 @@
...
+def validate_credentials(
+    provider: str,
+    credentials: ImageGenerationProviderCredentials,
+) -> bool:
+    provider_cls = _get_image_generation_provider_cls(provider)
+    return provider_cls.validate_credentials(credentials)
+
+
+def _get_image_generation_provider_cls(provider: str) -> type[ImageGenerationProvider]:
+    if provider not in PROVIDERS:
```
jessicasingh7: same as above- this would be cleaner if provider was an enum

backend/onyx/image_gen/interfaces.py
```diff
@@ -0,0 +1,67 @@
...
+                f"Invalid image generation credentials: {credentials}"
+            )
+        return cls._build_from_credentials(credentials)
+
+    @classmethod
+    @abc.abstractmethod
+    def _build_from_credentials(
+        cls,
+        credentials: ImageGenerationProviderCredentials,
+    ) -> "ImageGenerationProvider":
```
jessicasingh7: stylistic nit- you can make the return type ImageGenerationProvider without the quotes with this import from __future__ import annotations

jessicasingh7: can be done for rest of code as well if you do this

Danelegend:
@greptile

backend/onyx/image_gen/factory.py
Danelegend: Spoke to evan re putting the provider class inside the enum and dropping the table. For consistency we decided on having an enum + provider map as that's alrdy in the codebase

PR #7403, opened 2026-01-14
feat(img-gen): Add nanobanana to backend

Danelegend:
## Description
Implements nano banana to the backend so that the config can be set, and it can be selected and used as a provider

## How Has This Been Tested?
Integration tests were written.
Manually verified that the flow worked by hitting BE endpoints.
Linked up with the FE and checked that we got expected results.

## Additional Options

---
## Summary by cubic
Adds a unified image generation provider system with OpenAI, Azure, and Vertex support, plus credential validation and provider selection via config and env. Updates server APIs and tools to use the new factory and safer custom config handling.

- New Features
  - Added ImageGenerationProvider interface, factory, and providers for openai, azure, vertex_ai.
  - Validates credentials before use; clear errors for bad providers or credentials.
  - ImageGenerationTool now builds a provider from credentials and validates availability.
  - Server endpoints accept and use custom_config; test/create/update paths route through the provider factory.
  - Added IMAGE_MODEL_PROVIDER env var (default: openai). LLMConfig adds custom_config; sensitive values are masked in logs.
  - Tests added for provider building; integration tests updated.

- Migration
  - Set IMAGE_MODEL_PROVIDER to the desired provider.
  - OpenAI: api_key (api_base optional).
  - Azure: api_key, api_base, api_version, deployment_name or model name.
  - Vertex: provide custom_config with vertex_credentials (JSON) and vertex_location; project_id is read from the credentials JSON.
  - Update ImageGenerationConfig/Test requests to include provider and custom_config where required.
  - LLMConfig: credentials_file removed; use custom_config.

<sup>Written for commit c999065deb3aea9a3782a44c5a8b2c5acb6334ba. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/image_gen/providers/vertex_img_gen.py
+++ b/backend/onyx/image_gen/providers/vertex_img_gen.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import json
 from typing import Any
 from typing import TYPE_CHECKING
@@ -19,8 +21,6 @@ class VertexCredentials(BaseModel):
 
 
 class VertexImageGenerationProvider(ImageGenerationProvider):
-    NAME = "vertex_ai"
-
     def __init__(
         self,
         vertex_credentials: VertexCredentials,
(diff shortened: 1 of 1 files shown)


Danelegend:
@greptile

backend/tests/integration/common_utils/managers/image_generation.py
```diff
@@ -9,6 +11,18 @@
 from tests.integration.common_utils.test_models import DATestUser
 
 
+def _serialize_custom_config(
```
jessicasingh7, line 14: do we need this method?- I see it's only used for the test, maybe we could just convert it in the Test to the right type `custom_config`

PR #7431, opened 2026-01-15
feat: Backend Search APIs

yuhongsun96:
## Description
Add the send search query and getting search queries APIs

## How Has This Been Tested?
Tested locally, simple enough so did not add unit/integration tests

## Additional Options

---
## Summary by cubic
Introduces backend search APIs with optional SSE streaming to return search results in real time. Adds endpoints to run document search and fetch saved search history for authenticated users.

- **New Features**
  - POST /send-search-message with stream support:
    - stream=false returns SearchFullResponse.
    - stream=true returns SSE JSON lines (SearchQueriesPacket, SearchDocsPacket, SearchErrorPacket).
  - GET /search-history returns recent queries for the signed-in user.
  - New helpers: stream_search_query and gather_search_stream.
  - Persists executed queries to the DB for signed-in users.
  - New models: SendSearchQueryRequest, SearchFullResponse, SearchQueriesPacket, SearchDocsPacket, SearchErrorPacket, SearchQueryResponse, SearchHistoryResponse.
  - Query expansion and LLM doc selection are stubbed and return NotImplemented errors.

- **Migration**
  - Clients should call /send-search-message and set stream based on needs.
  - Replace any usage of DocumentSearchRequest/Response with SearchDoc-based results and the new packet formats.

<sup>Written for commit 7a2bfb7e025478f282fa96c5dcd6429f850cb8c0. Summary will update on new commits.</sup>

### DIFF
--- a/backend/ee/onyx/server/query_and_chat/search_backend.py
+++ b/backend/ee/onyx/server/query_and_chat/search_backend.py
@@ -146,7 +146,7 @@ def get_search_history(
             detail="filter_days must be greater than 0",
         )
 
-    # TODO remove this
+    # TODO(yuhong) remove this
     if user is None:
         # Return empty list for unauthenticated users
         return SearchHistoryResponse(search_queries=[])


backend/ee/onyx/server/query_and_chat/streaming_models.py
```diff
@@ -0,0 +1,20 @@
+from typing import Literal
+
+from pydantic import BaseModel
+
+from onyx.context.search.models import SearchDoc
+
+
+class SearchQueriesPacket(BaseModel):
+    type: Literal["search_queries"] = "search_queries"
+    all_executed_queries: list[str]
```
acaprau, line 13: do we wanna be able to mutate these packets in this file? i hope the answer is no bc that just makes things more complicated, if it is no let's make these frozen with `model_config = {"frozen": True}`

backend/ee/onyx/server/query_and_chat/search_backend.py
```diff
@@ -51,3 +61,51 @@ def search_flow_classification(
...
+        try:
+            packets = stream_search_query(request, user, db_session)
+            return gather_search_stream(packets)
+        except NotImplementedError as e:
+            return SearchFullResponse(
+                all_executed_queries=[],
+                search_docs=[],
+                error=str(e),
+            )
+        except Exception as e:
```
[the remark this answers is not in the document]
acaprau: wait no i dont think this was addressed, this path still will return a 200. id have to look at how @router.post("/send-search-message", response_model=None) works w.r.t. exceptions but we should prob be raising some fast api exception here with the desired code, or just reraise. the response object will contain a reason which clients can parse

backend/ee/onyx/server/query_and_chat/search_backend.py
```diff
@@ -51,3 +65,96 @@ def search_flow_classification(
...
+    """
+    logger.debug(f"Received search query: {request.search_query}")
+
+    # Non-streaming path
+    if not request.stream:
+        try:
+            packets = stream_search_query(request, user, db_session)
+            return gather_search_stream(packets)
+        except NotImplementedError as e:
+            return SearchFullResponse(
```
acaprau: this pattern seems strange, why would we not expect something to be implemented?

backend/ee/onyx/server/query_and_chat/search_backend.py
```diff
@@ -51,3 +65,96 @@ def search_flow_classification(
...
+@router.post("/send-search-message", response_model=None)
+def handle_send_search_message(
+    request: SendSearchQueryRequest,
+    user: User | None = Depends(current_user),
+    db_session: Session = Depends(get_session),
+) -> StreamingResponse | SearchFullResponse:
+    """
+    Execute a search query with optional streaming.
+
+    When stream=True: Returns StreamingResponse with SSE
```
acaprau, line 79: what is sse?

backend/ee/onyx/server/query_and_chat/search_backend.py
```diff
@@ -51,3 +65,96 @@ def search_flow_classification(
...
+    limit: int = 100,
+    filter_days: int | None = None,
+    user: User | None = Depends(current_user),
+    db_session: Session = Depends(get_session),
+) -> SearchHistoryResponse:
+    """
+    Fetch past search queries for the authenticated user.
+
+    Args:
+        limit: Maximum number of queries to return (default 10)
```
acaprau: looks like it's 100 in the code

backend/ee/onyx/server/query_and_chat/search_backend.py
```diff
@@ -51,3 +65,96 @@ def search_flow_classification(
...
+        filter_days: Only return queries from the last N days (optional)
+
+    Returns:
+        SearchHistoryResponse with list of search queries, ordered by most recent first.
+    """
+    # TODO remove this
```
acaprau:
```suggestion
    # TODO(yuhong) remove this
```
todos tend to disappear forever if a name isnt associated with it. makes it easy to search for your todos as well bc you just have to search for your name in the codebase

backend/ee/onyx/server/query_and_chat/search_backend.py
```diff
@@ -51,3 +65,96 @@ def search_flow_classification(
...
+    if user is None:
```
acaprau, line 150: why not return a 403? sorry i dont have context on who consumes this api but i imagine it's less clear for users if they click their history and it's just empty and they dont know why

backend/ee/onyx/server/query_and_chat/search_backend.py
```diff
@@ -51,3 +64,92 @@ def search_flow_classification(
...
+        # Return empty list for unauthenticated users
+        return SearchHistoryResponse(search_queries=[])
+
+    search_queries = fetch_search_queries_for_user(
```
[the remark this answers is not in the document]
acaprau: seems reasonable

backend/ee/onyx/server/query_and_chat/models.py
```diff
@@ -16,17 +17,38 @@ class StandardAnswerResponse(BaseModel):
...
+
+
+class SearchFullResponse(BaseModel):
+    all_executed_queries: list[str]
+    search_docs: list[SearchDoc]
+    # Reasoning tokens output by the LLM for the document selection
+    doc_selection_reasoning: str | None = None
+    # This a list of document ids that are in the search_docs list
+    llm_selected_doc_ids: list[str] | None = None
+    # Error message if the search failed partway through
```
acaprau, line 43: do we expect to get any valuable data if this happens. just quickly scanning `handle_send_search_message` i dont think it does? see comment below but seems like we might just wanna use regular http/rest semantics here and return some non 2xx with error message

backend/ee/onyx/search/process_search_query.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def stream_search_query(
+    request: SendSearchQueryRequest,
+    user: User | None,
+    db_session: Session,
+) -> Generator[SearchQueriesPacket | SearchDocsPacket | SearchErrorPacket, None, None]:
+    """
+    Core search function that yields streaming packets.
+    Used by both streaming and non-streaming endpoints.
```
acaprau, line 27: prob worth mentioning that the non streaming endpoint simply exhausts the generator, otherwise this comment is confusing to someone just looking at this file and not the other one

backend/ee/onyx/search/process_search_query.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    Aggregate all streaming packets into SearchFullResponse.
+    """
+    all_executed_queries: list[str] = []
+    search_docs: list[SearchDoc] = []
+    error: str | None = None
+
+    for packet in packets:
+        if isinstance(packet, SearchQueriesPacket):
+            all_executed_queries = packet.all_executed_queries
```
acaprau, line 92: this clobbers not appends. is that intended? same below

backend/ee/onyx/search/process_search_query.py
yuhongsun96: I'm not sure I follow

yuhongsun96: The nonstreaming one just gathers the packets, feels like this function is pretty straightforward

backend/ee/onyx/search/process_search_query.py
yuhongsun96: That's fine, probably this is preferable

backend/ee/onyx/server/query_and_chat/models.py
yuhongsun96: Intention is it can for example run search and pass, then exceptions when calling the LLM, and instead of failing the whole thing, we can give back at least the search docs. That gives more flexibility to the caller to handle the cases

backend/ee/onyx/server/query_and_chat/search_backend.py
yuhongsun96: server side events I believe

backend/ee/onyx/server/query_and_chat/search_backend.py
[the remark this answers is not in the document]
yuhongsun96: I'm pretty ok with a top level exception handling for these types of flows but if we prefer, we can just not catch any exceptions here outside the NotImplemented. The other exceptions aren't raised anyway so it's more just scaffolding for future extensions. But ya I think not catching is better

backend/ee/onyx/server/query_and_chat/search_backend.py
yuhongsun96: This is because Jessica is actively removing the "or None" to the user dependency. I'm just doing this so she doesn't need to refactor the underlying functions and database models. Otherwise the user foreign key would need to be nullable also and she'd need to handle that also.

backend/ee/onyx/server/query_and_chat/search_backend.py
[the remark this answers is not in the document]
yuhongsun96: tru

backend/ee/onyx/server/query_and_chat/streaming_models.py
yuhongsun96: solid point, making the change

PR #7435, opened 2026-01-15
fix: bedrock non-anthropic prompt caching

evan-onyx:
## Description

Trying to do anthropic-style prompt caching for non-anthropic models served by bedrock was causing litellm to complain, hence we remove it

## How Has This Been Tested?

n/a its rlly small trust

## Additional Options

---
## Summary by cubic
Fixes prompt caching on Bedrock by only using Anthropic-style caching for Anthropic models. Prevents litellm errors on non-Anthropic Bedrock models.

- **Bug Fixes**
  - Updated provider factory to accept LLMConfig and use AnthropicPromptCacheProvider only for Anthropic models, including Bedrock models whose model_name contains "anthropic.".
  - Non-Anthropic Bedrock models now use the no-op cache provider. OpenAI and Vertex logic unchanged.

<sup>Written for commit ba6ab5b64281371516d1a784aa9e443970ecce37. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/llm/prompt_cache/providers/factory.py
+++ b/backend/onyx/llm/prompt_cache/providers/factory.py
@@ -8,7 +8,7 @@ from onyx.llm.prompt_cache.providers.noop import NoOpPromptCacheProvider
 from onyx.llm.prompt_cache.providers.openai import OpenAIPromptCacheProvider
 from onyx.llm.prompt_cache.providers.vertex import VertexAIPromptCacheProvider
 
-ANTHROPIC_BEDROCK_PREFIX = "anthropic.bedrock"
+ANTHROPIC_BEDROCK_TAG = "anthropic."
 
 
 def get_provider_adapter(llm_config: LLMConfig) -> PromptCacheProvider:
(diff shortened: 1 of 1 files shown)


backend/onyx/llm/prompt_cache/providers/factory.py
```diff
@@ -1,14 +1,17 @@
...
 
 from onyx.llm.constants import LlmProviderNames
+from onyx.llm.interfaces import LLMConfig
 from onyx.llm.prompt_cache.providers.anthropic import AnthropicPromptCacheProvider
 from onyx.llm.prompt_cache.providers.base import PromptCacheProvider
 from onyx.llm.prompt_cache.providers.noop import NoOpPromptCacheProvider
 from onyx.llm.prompt_cache.providers.openai import OpenAIPromptCacheProvider
 from onyx.llm.prompt_cache.providers.vertex import VertexAIPromptCacheProvider
 
+ANTHROPIC_BEDROCK_PREFIX = "anthropic.bedrock"
```
nmgarza5: this is prefix for the model string correct?

I think depending on the region, sometimes you might see:
`us.anthropic.bedrock`, `global.anthropic.bedrock`, etc

PR #7437, opened 2026-01-15
feat(opensearch): Implement match highlighting

acaprau:
## Description
This PR implements a feature that returns which part of a chunk's content contributed to a text match.

It also bumbs the OpenSearch image tag to 3.4.0 because somewhere between 3.0.0 and that release they dropped a fix for match highlighting; it did not work for hybrid queries beforehand. See https://github.com/opensearch-project/neural-search/issues/1215

Note that the highest version of the python client is 3.1.0, and we happen to be on 3.0.0. What a mess lol. Anyway from my testing things seem to work, I can't imagine that OpenSearch removed client-facing features from 3.0 to 3.4, just added things that maybe the client doesn't reflect.

## How Has This Been Tested?
`test_opensearch_client`

## Additional Options

---
## Summary by cubic
Adds match highlighting to OpenSearch search results so we return the exact content snippets that matched a query. Highlights use <hi> tags and work for hybrid search.

- **New Features**
  - Enable unified highlighter in hybrid queries (fragment_size 100, 4 fragments, <hi> tags).
  - Capture OpenSearch highlight data in SearchHit.match_highlights and pass to inference chunks (content field).
  - Update content and title fields to use index_options: offsets for faster highlighting.
  - Add tests that assert highlights are present and correctly tagged.

- **Dependencies**
  - Requires OpenSearch 3.4.0+ for hybrid query highlighting; dev container bumped to 3.4.0.

<sup>Written for commit e13798aa63583f8312015c724a1834230148e9f0. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/external_dependency_unit/opensearch/test_opensearch_client.py
+++ b/backend/tests/external_dependency_unit/opensearch/test_opensearch_client.py
@@ -638,7 +638,7 @@ class TestOpenSearchClient:
             assert chunk.score
 
         # Make sure there is some kind of match highlight for the first hit. We
-        # don't expect hightlights for any other hit.
+        # don't expect highlights for any other hit.
         assert results[0].match_highlights.get(CONTENT_FIELD_NAME, [])
 
     def test_search_with_pipeline(


backend/tests/external_dependency_unit/opensearch/test_opensearch_client.py
```diff
@@ -635,6 +637,10 @@ def test_search_basic(
             # or 0).
             assert chunk.score
 
+        # Make sure there is some kind of match highlight for the first hit. We
+        # don't expect hightlights for any other hit.
```
evan-onyx: highlights*

PR #7443, opened 2026-01-15
feat(auth): update default auth

jessicasingh7:
## Description

removed no auth and default is now basic and/or google auth

[Linear project](https://linear.app/onyx-app/project/update-default-auth-61e6eac3f333/overview)

## How Has This Been Tested?

ran with different auth env vars, tested migration with no auth case w local db

## Additional Options

---
## Summary by cubic
Removed “no auth” mode. Default auth is now Basic (legacy AUTH_TYPE=disabled/none falls back to Basic with a warning). Added an anonymous user for restricted unauthenticated access, plus safe migrations that preserve no‑auth data and transfer it to the first real user on signup. Aligns with Linear project “update-default-auth”.

- **New Features**
  - Default AUTH_TYPE=BASIC; removed disabled‑auth paths; enforced across API and UI; deprecated AUTH_TYPE=disabled/none now maps to BASIC with a warning.
  - Optional OAuth login when OAUTH_CLIENT_ID/SECRET are set; backend routes wired; login shows a provider button when available; PAT routes are always available.
  - Introduced a permanent anonymous user for limited, public‑only use; updated permissions and global‑only rate limits; excluded from analytics and user lists.

- **Migration**
  - Placeholder user created to hold all NULL user_id records from no‑auth; a DB trigger transfers data to the first real user on signup.
  - Added a migration to create a permanent anonymous user and move any remaining NULL user_id records to it.
  - Evals now require a real user via search_permissions_email and an API key; removed no‑auth eval flow.

<sup>Written for commit 9b0dc5b158b24a6569ec297e6575833c441f515f. Summary will update on new commits.</sup>

### DIFF
--- a/backend/alembic/versions/f7ca3e2f45d9_migrate_no_auth_data_to_placeholder.py
+++ b/backend/alembic/versions/f7ca3e2f45d9_migrate_no_auth_data_to_placeholder.py
@@ -8,7 +8,7 @@ A database trigger is installed that automatically transfers all data from
 the placeholder user to the first real user who registers, then drops itself.
 
 Revision ID: f7ca3e2f45d9
-Revises: 849b21c732f8
+Revises: 78ebc66946a0
 Create Date: 2026-01-15 12:49:53.802741
 
 """
@@ -23,7 +23,7 @@ from shared_configs.configs import MULTI_TENANT
 
 # revision identifiers, used by Alembic.
 revision = "f7ca3e2f45d9"
-down_revision = "849b21c732f8"
+down_revision = "78ebc66946a0"
 branch_labels = None
 depends_on = None
 


backend/alembic/versions/f7ca3e2f45d9_migrate_no_auth_data_to_placeholder.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+def upgrade() -> None:
+    """
+    Create a placeholder user and assign all NULL user_id records to it.
+    """
+    connection = op.get_bind()
+
+    # Check if there are any NULL user_id records that need migration
+    tables_to_check = [
```
Danelegend, line 153: usage_reports and user_projects also have nullable user_ids. Are these ok to skip?

backend/alembic/versions/f7ca3e2f45d9_migrate_no_auth_data_to_placeholder.py
```diff
@@ -0,0 +1,159 @@
...
+            )
+            if result.fetchone():
+                has_null_records = True
+                break
+        except Exception:
+            # Table might not exist
+            pass
+
+    if not has_null_records:
+        print("No NULL user_id records found. Skipping placeholder user creation.")
```
Danelegend: probs not worth printing this

backend/alembic/versions/f7ca3e2f45d9_migrate_no_auth_data_to_placeholder.py
```diff
@@ -0,0 +1,159 @@
...
+                "id": NO_AUTH_PLACEHOLDER_USER_UUID,
+                "email": NO_AUTH_PLACEHOLDER_USER_EMAIL,
+                "hashed_password": "",  # Empty password - user cannot log in
+                "is_active": False,  # Inactive - user cannot log in
+                "is_superuser": False,
+                "is_verified": False,
+                "role": "BASIC",
+            },
+        )
+        print(f"Created placeholder user: {NO_AUTH_PLACEHOLDER_USER_EMAIL}")
```
Danelegend: Can remove

backend/alembic/versions/f7ca3e2f45d9_migrate_no_auth_data_to_placeholder.py
```diff
@@ -0,0 +1,159 @@
...
+    for table in tables_to_check:
+        try:
+            # Exclude public credential (id=0) which must remain user_id=NULL
+            # Exclude builtin tools (in_code_tool_id IS NOT NULL) which must remain user_id=NULL
+            if table == "credential":
+                condition = "user_id IS NULL AND id != 0"
+            elif table == "tool":
+                condition = "user_id IS NULL AND in_code_tool_id IS NULL"
+            else:
+                condition = "user_id IS NULL"
```
Danelegend: kinda nitty, but would be better to write condition = "user_id IS NULL" above the conditionals, then build. on top of that

backend/alembic/versions/f7ca3e2f45d9_migrate_no_auth_data_to_placeholder.py
```diff
@@ -0,0 +1,159 @@
...
+                    SET user_id = :user_id
+                    WHERE {condition}
+                    """
+                ),
+                {"user_id": NO_AUTH_PLACEHOLDER_USER_UUID},
+            )
+            if result.rowcount > 0:
+                print(f"Updated {result.rowcount} rows in {table}")
+        except Exception as e:
+            print(f"Skipping {table}: {e}")
```
Danelegend, line 228: Why is a silent failure ok here? And when would we expect it to occur?

backend/alembic/versions/f7ca3e2f45d9_migrate_no_auth_data_to_placeholder.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+def downgrade() -> None:
+    """
+    Set placeholder user's records back to NULL and delete the placeholder user.
+    """
+    connection = op.get_bind()
+
+    tables_to_update = [
```
Danelegend, line 251: Move tables_to_update outside both functions so you only have to write it once. Maintains a single source of truth aspect

backend/alembic/versions/f7ca3e2f45d9_migrate_no_auth_data_to_placeholder.py
```diff
@@ -0,0 +1,159 @@
...
+        except Exception:
+            # Table might not exist
+            pass
+
+    if not has_null_records:
+        print("No NULL user_id records found. Skipping placeholder user creation.")
+        return
+
+    # Check if placeholder user already exists
+    result = connection.execute(
```
Danelegend: Afaik migrations have a guaranteed order. So I don't think we'd expect another placeholder user, and so erroring would maybe be ideal if that is the case? Could be wrong though

backend/alembic/versions/f7ca3e2f45d9_migrate_no_auth_data_to_placeholder.py
```diff
@@ -0,0 +1,159 @@
...
+                ),
+                {"user_id": NO_AUTH_PLACEHOLDER_USER_UUID},
+            )
+        except Exception:
+            pass
+
+    # Delete the placeholder user
+    connection.execute(
+        sa.text('DELETE FROM "user" WHERE id = :user_id'),
+        {"user_id": NO_AUTH_PLACEHOLDER_USER_UUID},
```
Danelegend, line 280: Just jumping on my comment before about how this guy should not exist. The current code-path suggests there is a chance he could exist (although I don't think he does). If he were to alrdy exist and we delete him, wouldn't that potentially create problems / things would mutate away from the expected flow.

I think the trick is to assume this guy doesn't exist and error if he does. That way we can assume here for sure that he doesn't exist

backend/onyx/auth/users.py
```diff
@@ -500,6 +501,85 @@ async def _assign_default_pinned_assistants(
         )
         user.pinned_assistants = default_persona_ids
 
+    async def _migrate_no_auth_data_to_user(
```
Danelegend: Hmm, this code-path is gonna be very inactive / triggered very infrequently. It might be possible to write this as a database trigger instead that listens for user creation and runs once when that happens (and ports the data over). It can then stop running in the future (or even drop itself after event completion). That way this logic only exists for users who need to be ported over, and we don't bloat the codebase.

Might be worth talking to Evan or Yuhong about this one

backend/onyx/auth/users.py
```diff
@@ -500,6 +501,85 @@ async def _assign_default_pinned_assistants(
...
+        to the first registered user. This transfers ownership of all data that
+        was created when AUTH_TYPE=disabled was in use.
+
+        The placeholder user is deleted after successful migration.
+        """
+        placeholder_uuid = uuid.UUID(NO_AUTH_PLACEHOLDER_USER_UUID)
+
+        # Check if placeholder user exists
+        result = await db_session.execute(
+            text('SELECT id FROM "user" WHERE id = :user_id'),
```
Danelegend: Should probs use ORM

backend/onyx/auth/users.py
```diff
@@ -500,6 +501,85 @@ async def _assign_default_pinned_assistants(
...
+                        "placeholder_user_id": str(placeholder_uuid),
+                    },
+                )
+                if result.rowcount > 0:  # type: ignore[attr-defined]
+                    logger.info(f"Migrated {result.rowcount} rows in {table}")  # type: ignore[attr-defined]
+            except Exception as e:
+                logger.warning(f"Could not migrate {table}: {e}")
+
+        # Make the user an admin since they had admin access in no-auth mode
+        # Use raw SQL to keep everything in the same transaction
```
Danelegend: I think we can swap between text + ORM and keep it in the same transaction as long as we call .commit at the end

backend/onyx/main.py
```diff
@@ -447,14 +447,14 @@ def get_application(lifespan_override: Lifespan | None = None) -> FastAPI:
             prefix="/users",
         )
 
-    if AUTH_TYPE == AuthType.GOOGLE_OAUTH:
-        # For Google OAuth, refresh tokens are requested by:
-        # 1. Adding the right scopes
-        # 2. Properly configuring OAuth in Google Cloud Console to allow offline access
+    # Register Google OAuth when AUTH_TYPE is GOOGLE_OAUTH, or when
+    # AUTH_TYPE is BASIC and OAuth credentials are configured
+    if AUTH_TYPE == AuthType.GOOGLE_OAUTH or (
```
Danelegend, line 458: What's the difference between the two. When would we be in the second one but not the first?

Danelegend (changes requested):
Removals look good. I'm just interested if we can find a way to move the migration out of the codebase and into the database or something. It seems like something that can be solved with DB triggers and means that this niche path won't exist in the codebase.

Might be worth a discussion

jessicasingh7:
@greptile

backend/onyx/main.py
jessicasingh7: when GOOGLE_OAUTH is set only google_oauth login is allowed and if it's BASIC with google oauth creds then both login options are shown

backend/alembic/versions/f7ca3e2f45d9_migrate_no_auth_data_to_placeholder.py
jessicasingh7: this is a part of the downgrade flow where anonymous users may have been created with the migration

backend/alembic/versions/f7ca3e2f45d9_migrate_no_auth_data_to_placeholder.py
jessicasingh7: if usage_reports are none it means it was auto generated and user_projects cannot have null user_ids because nullable=False

backend/alembic/versions/f7ca3e2f45d9_migrate_no_auth_data_to_placeholder.py
jessicasingh7: Since this is a best effort migration of anonymous data, we wouldn't want a missing table to fail the migration so we just log and continue

backend/alembic/versions/f7ca3e2f45d9_migrate_no_auth_data_to_placeholder.py
Danelegend: I think what I meant is that in the upgrade script you have this:

sa.text('SELECT id FROM "user" WHERE id = :user_id'),
        {"user_id": NO_AUTH_PLACEHOLDER_USER_UUID},
which is used to check if the placeholder user exists. If they don't exist we create them.

But in the downgrade script, we delete them. So just by reasoning from the code, there exists a world where the above said there was a user (so we don't create a new one). And then the downgrade deletes that pre-existing user which is a side-effect in this case.

I'm gonna go under the assumption that this user doesn't exist (and the upgrade check should never actually return True). But worth a quick thought

backend/alembic/versions/e7f8a9b0c1d2_create_anonymous_user.py
```diff
@@ -0,0 +1,143 @@
...
+
+def downgrade() -> None:
+    """
+    Set anonymous user's records back to NULL and delete the anonymous user.
+    """
+    connection = op.get_bind()
+
+    # Note: agent__search_metrics was dropped in migration a1b2c3d4e5f7
+    tables_to_update = [
+        "chat_session",
```
Danelegend: Nit: Personally I'd make this an out-of-function constant used in both upgrade and downgrade. Saves time repeating + minimises risk of misalignment

backend/alembic/versions/e7f8a9b0c1d2_create_anonymous_user.py
```diff
@@ -0,0 +1,143 @@
...
+                    WHERE user_id = :user_id
+                    """
+                ),
+                {"user_id": ANONYMOUS_USER_UUID},
+            )
+        except Exception:
+            pass
+
+    # Delete the anonymous user
+    connection.execute(
```
Danelegend, line 122:
```
 # Check if anonymous user already exists
    result = connection.execute(
        sa.text('SELECT id FROM "user" WHERE id = :user_id'),
        {"user_id": ANONYMOUS_USER_UUID},
    )
    if result.fetchone():
        print("Anonymous user already exists. Skipping creation.")
```

If result.fetchone() is true (so anon user alrdy exists), is deleting here going to be a problem? Since this hypothetical would be affecting beyond just this migration.
Is result.fetchone() even a code-path we expect?

backend/alembic/versions/f7ca3e2f45d9_migrate_no_auth_data_to_placeholder.py
```diff
@@ -0,0 +1,281 @@
...
+depends_on = None
+
+# Must match constants in onyx/configs/constants.py file
+NO_AUTH_PLACEHOLDER_USER_UUID = "00000000-0000-0000-0000-000000000001"
+NO_AUTH_PLACEHOLDER_USER_EMAIL = "no-auth-placeholder@onyx.app"
+
+
+# Trigger function that migrates data from placeholder to first real user
+MIGRATE_NO_AUTH_TRIGGER_FUNCTION = """
+CREATE OR REPLACE FUNCTION migrate_no_auth_data_to_user()
```
Danelegend: Can be extract out the function name and trigger name, make them constants and then insert them into the string via formatting. It'll just ensure a bit more safety cause  it's a bit long this function

Danelegend:
@greptile

backend/alembic/versions/f7ca3e2f45d9_migrate_no_auth_data_to_placeholder.py
```diff
@@ -0,0 +1,281 @@
...
+    connection.execute(sa.text(MIGRATE_NO_AUTH_TRIGGER))
+    print("Installed trigger for automatic data migration on first user registration")
+
+
+def downgrade() -> None:
+    """
+    Drop trigger and function, set placeholder user's records back to NULL,
+    and delete the placeholder user.
+    """
+    connection = op.get_bind()
```
Danelegend, line 245: probs add a skip for MULTI-TENANT for consistency

backend/ee/onyx/db/analytics.py
```diff
@@ -334,11 +335,10 @@ def fetch_assistant_unique_users_total(
...
 # or if they are an admin
 def user_can_view_assistant_stats(
-    db_session: Session, user: User | None, assistant_id: int
+    db_session: Session, user: User, assistant_id: int
 ) -> bool:
-    # If user is None and auth is disabled, assume the user is an admin
-
-    if user is None or user.role == UserRole.ADMIN:
+    # Anonymous users and admins can view all stats
+    if str(user.id) == ANONYMOUS_USER_UUID or user.role == UserRole.ADMIN:
```
Danelegend: Just want to double check that this is something we want to support for anon users. Based on the comments it seems like anon users might have been overlooked here

backend/ee/onyx/db/token_limit.py
```diff
@@ -18,13 +18,15 @@
...
-    stmt: Select, user: User | None, get_editable: bool = True
-) -> Select:
-    # If user is None and auth is disabled, assume the user is an admin
-    if (user is None and DISABLE_AUTH) or (user and user.role == UserRole.ADMIN):
+def _add_user_filters(stmt: Select, user: User, get_editable: bool = True) -> Select:
+    if user.role == UserRole.ADMIN:
         return stmt
 
+    # If anonymous user, only show global/public token_rate_limits
+    if str(user.id) == ANONYMOUS_USER_UUID:
```
Danelegend: Maybe make this a property on the user model instead (.is_anonymous)

backend/ee/onyx/db/user_group.py
```diff
@@ -144,7 +145,8 @@ def validate_object_creation_for_user(
     if object_is_perm_sync and not target_group_ids:
         return
 
-    if not user or user.role == UserRole.ADMIN:
+    # Anonymous users and admins are allowed
+    if str(user.id) == ANONYMOUS_USER_UUID or user.role == UserRole.ADMIN:
```
Danelegend: Again, maybe just validate that anon users are meant to have admin level perms

backend/onyx/chat/process_message.py
```diff
@@ -318,12 +319,14 @@ def handle_stream_message_objects(
     chat_session: ChatSession | None = None
     redis_client: Redis | None = None
 
-    user_id = user.id if user is not None else None
-    llm_user_identifier = (
-        user.email
-        if user is not None and getattr(user, "email", None)
-        else (str(user_id) if user_id else "anonymous_user")
-    )
+    user_id = user.id if user else None
```
Danelegend: Why can user be None in this function? The upstream api function ('send-chat-message') that calls this function has been typed to just User

backend/onyx/configs/app_configs.py
```diff
@@ -71,8 +74,16 @@
 #####
 # Auth Configs
 #####
-AUTH_TYPE = AuthType((os.environ.get("AUTH_TYPE") or AuthType.DISABLED.value).lower())
-DISABLE_AUTH = AUTH_TYPE == AuthType.DISABLED
+# Upgrades users from disabled auth to basic auth and shows warning.
+_auth_type_str = (os.environ.get("AUTH_TYPE") or AuthType.BASIC.value).lower()
```
Danelegend: If we're wanting this warning, shouldn't we get rid of 'or AuthType.BASIC.value'

backend/onyx/server/settings/api.py
```diff
@@ -70,63 +57,6 @@ def fetch_settings(
 
     return UserSettings(
         **general_settings.model_dump(),
-        notifications=settings_notifications,
+        notifications=[],
         needs_reindexing=needs_reindexing,
     )
-
-
-def get_settings_notifications(
```
Danelegend, line 83: Seems like there's user and admin logic that's getting removed here?

Danelegend (changes requested):
Couple more things. I definetely think we can make an is_anonymous property to clear up a few more LOC.

Also gonna check out in my local branch tmr just to make sure there are no more User | None missing

Danelegend:
I think there's also a couple new routes and flows that need to have the User | None removed (re search_backend.py). Maybe prepare that in a separate PR though as this ones getting pretty big

Danelegend:
One more comment, I see handle_stream_message_objects (and stream_chat_message_objects) still support User | None because it looks like the eval pipeline can put in a None user. What is we remove the potential for the eval pipeline to put in a None user, so then the downstream affects are we no longer need to handle None in this core flow

backend/onyx/configs/app_configs.py
jessicasingh7: we'll upgrade users to basic if they are have auth_type disabled or unset and show this warning

backend/alembic/versions/e7f8a9b0c1d2_create_anonymous_user.py
jessicasingh7: removed

backend/onyx/server/settings/api.py
jessicasingh7: I think we were discussing this code is not being used but won't have this as a part of this pr

web/src/app/auth/login/page.tsx
```diff
@@ -76,7 +71,14 @@ export default async function Page(props: PageProps) {
   let authUrl: string | null = null;
   if (authTypeMetadata) {
     try {
-      authUrl = await getAuthUrlSS(authTypeMetadata.authType, nextUrl!);
+      authUrl = await getAuthUrlSS(authTypeMetadata.authType, nextUrl);
```
Danelegend: Nit: Maybe use if (...) else ... instead here. I think the ai comments are valid

backend/alembic/versions/f7ca3e2f45d9_migrate_no_auth_data_to_placeholder.py
```diff
@@ -0,0 +1,275 @@
...
+    """
+    Create a placeholder user and assign all NULL user_id records to it.
+    Install a trigger that migrates data to the first real user and self-destructs.
+    Only runs if no real users exist (true AUTH_TYPE=disabled deployment).
+
+    Skipped in multi-tenant mode - each tenant starts fresh with no legacy data.
+    """
+    # Skip in multi-tenant mode - this migration handles single-tenant
+    # AUTH_TYPE=disabled -> auth transitions only
+    if MULTI_TENANT:
```
yuhongsun96, line 140: What happens if it's a new deployment with no users? Thoughts on just doing this stuff only if AUTH_TYPE=disabled env var is set? I'm much more concerned about negative consequences for real users than issues for the auth disabled users

backend/ee/onyx/db/analytics.py
```diff
@@ -334,11 +334,10 @@ def fetch_assistant_unique_users_total(
...
 # or if they are an admin
 def user_can_view_assistant_stats(
-    db_session: Session, user: User | None, assistant_id: int
+    db_session: Session, user: User, assistant_id: int
 ) -> bool:
-    # If user is None and auth is disabled, assume the user is an admin
-
-    if user is None or user.role == UserRole.ADMIN:
+    # Anonymous users and admins can view all stats
+    if user.is_anonymous or user.role == UserRole.ADMIN:
```
yuhongsun96: This feels kind of strange to me that anonymous users would be able to do this but basic users cannot

yuhongsun96: Honestly with the UI how it is now, I think this check doesn't even need to exist. It's just agent metadata, I think it's fine to remove the check entirely

backend/ee/onyx/db/user_group.py
```diff
@@ -144,7 +144,8 @@ def validate_object_creation_for_user(
     if object_is_perm_sync and not target_group_ids:
         return
 
-    if not user or user.role == UserRole.ADMIN:
+    # Anonymous users and admins are allowed
```
yuhongsun96: This should not be the case. Anonymous users have very limited scope and should not be able to do this

backend/ee/onyx/external_permissions/post_query_censoring.py
```diff
@@ -34,15 +34,15 @@ def _get_all_censoring_enabled_sources() -> set[DocumentSource]:
...
+    user: User,
 ) -> list[InferenceChunk]:
     """
     This function checks all chunks to see if they need to be sent to a censoring
     function. If they do, it sends them to the censoring function and returns the
     censored chunks. If they don't, it returns the original chunks.
     """
-    if user is None:
-        # if user is None, permissions are not enforced
+    # Anonymous users skip permission enforcement
```
yuhongsun96: Whoa whoa whoa, this is very much not correct, this means if someone does not sign in, they can bipass ACL checks

yuhongsun96: I do think the anonymous users feature was poorly designed and built but we should definitely not be doing things like this. It hasn't been a problem because the users of the feature are not connecting up sensitive information but this enables exploits which are quite bad

backend/ee/onyx/search/process_search_query.py
```diff
@@ -101,8 +101,7 @@ def stream_search_query(
     # Build list of all executed queries for tracking
     all_executed_queries = [original_query] + keyword_expansions
 
-    # TODO remove this check, user should not be None
-    if user is not None:
+    if not user.is_anonymous:
```
yuhongsun96, line 104: The check doesn't need to be here, it should just require a user which is now enforced at the function level, the reason it has a "if user is not None" is because create_search_query does not allow for a None user.

We should not be blanket replacing everything with user.is_anonymous

backend/ee/onyx/server/query_and_chat/search_backend.py
```diff
@@ -146,9 +145,8 @@ def get_search_history(
             detail="filter_days must be greater than 0",
         )
 
-    # TODO(yuhong) remove this
-    if user is None:
-        # Return empty list for unauthenticated users
+    if user.is_anonymous:
```
yuhongsun96: This check doesn't need to be here

backend/onyx/auth/users.py
```diff
@@ -1420,14 +1425,11 @@ async def current_user(
...
 async def current_curator_or_admin_user(
-    user: User | None = Depends(current_user),
-) -> User | None:
-    if DISABLE_AUTH:
-        return None
-
-    if not user or not hasattr(user, "role"):
+    user: User = Depends(current_user),
+) -> User:
+    if not hasattr(user, "role"):
```
yuhongsun96: Is this not enforced by the User role field being a not nullable enum? I would remove this check

backend/onyx/auth/users.py
```diff
@@ -1439,11 +1441,8 @@ async def current_curator_or_admin_user(
     return user
 
 
-async def current_admin_user(user: User | None = Depends(current_user)) -> User | None:
-    if DISABLE_AUTH:
-        return None
-
-    if not user or not hasattr(user, "role") or user.role != UserRole.ADMIN:
+async def current_admin_user(user: User = Depends(current_user)) -> User:
+    if not hasattr(user, "role") or user.role != UserRole.ADMIN:
```
yuhongsun96: Same here

backend/onyx/db/models.py
```diff
@@ -93,6 +94,9 @@
 from shared_configs.enums import RerankerProvider
 from onyx.context.search.enums import RecencyBiasSetting
 
+# TODO: After anonymous user migration has been deployed, make user_id columns NOT NULL
```
yuhongsun96, line 103: I think it's ok if we want to have this as a follow up PR but I think a lot of foreign keys can now be not nullable

backend/onyx/db/memory.py
```diff
@@ -5,9 +5,9 @@
 from onyx.db.models import User
 
 
-def get_memories(user: User | None, db_session: Session) -> list[str]:
+def get_memories(user: User, db_session: Session) -> list[str]:
     # Anonymous users and None users don't have memories
-    if not user or user.is_anonymous:
+    if user.is_anonymous:
```
Danelegend: Don't need cause no memories

backend/onyx/server/kg/api.py
```diff
@@ -127,9 +127,9 @@ def enable_or_disable_kg(
             # If persona doesn't exist or can't be restored, create a new one below
             pass
 
-    # Create KG Beta persona
-    user_ids = [user.id] if user else []
-    is_public = len(user_ids) == 0
+    # Create KG Beta persona (private to the admin who enabled KG)
+    user_ids = [user.id]
+    is_public = False
```
Danelegend: Can probs remove these variables and put them in the UpsertRequest directly

PR #7444, opened 2026-01-15
fix: Fix invite buttons

raunakab:
## Description

- `NewTenantModal.tsx`:
  - Moved from `components/modals` to `sections/modals`.
  - Refactored to use the new `Modal` component and `opal` icons.
  - Refactored and fixed buttons to be consistent with new style.
- `input-layouts.tsx`:
  - Created new `ErrorTextLayout` component (reusable error/warning text display).
  - `ErrorTextLayout` is being used inside of the `NewTenantModal` now.
- `Modal.tsx`:
  - Updated docs.
  - Remove bad-practice `titleClassName`. No negative UI effects (since the only usage was applying `text-text-04`, and that's the default text colour anyways).

## Screenshots

image

image

## Additional Options

### DIFF
--- a/backend/onyx/document_index/opensearch/client.py
+++ b/backend/onyx/document_index/opensearch/client.py
@@ -45,6 +45,9 @@ class SearchHit(BaseModel, Generic[SchemaDocumentModel]):
     # relevant for "fuzzy searches"; this will be None for direct queries where
     # score is not relevant like direct retrieval on ID.
     score: float | None = None
+    # Maps schema property name to a list of highlighted snippets with match
+    # terms wrapped in tags (e.g. "something <hi>keyword</hi> other thing").
+    match_highlights: dict[str, list[str]] = {}
 
 
 class OpenSearchClient:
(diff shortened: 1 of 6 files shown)


web/src/refresh-components/Modal.tsx
```diff
@@ -390,9 +371,9 @@ const ModalHeader = React.forwardRef<HTMLDivElement, ModalHeaderProps>(
             )}
           </Section>
           <DialogPrimitive.Title asChild>
-            <Truncated headingH3 as="span" className={titleClassName}>
+            <Text headingH3 as="span">
```
jmelahman:
```suggestion
            <Text headingH3>
```

web/src/layouts/input-layouts.tsx
```diff
@@ -257,37 +257,34 @@ function ErrorLayout({ name }: FieldErrorLayoutProps) {
...
-          >
-            {warning}
-          </Text>
-        </>
-      )}
+  return (
+    <div className="px-1">
+      <Section flexDirection="row" gap={0.25}>
+        <Icon size={12} className={strokeClass} />
+        <Text as="span" secondaryBody className={colorClass} role="alert">
```
jmelahman:
```suggestion
        <Text secondaryBody className={colorClass} role="alert">
```

jmelahman (approved):

web/src/refresh-components/Modal.tsx
raunakab: Straight fire 🔥

web/src/layouts/input-layouts.tsx
raunakab: Absolute bars 🔥

PR #7475, opened 2026-01-16
refactor: Layout fixes

raunakab:
## Description

- `general-layouts.tsx`:
  - Added `LineItemLayout` component.
  - Replaced `fit` boolean with `width`/`height` props ("auto" | "fit" | "full").
- `actions-layouts.tsx`:
  - Refactored `ActionsHeader`, `ActionsTool`, `ActionsToolSkeleton` to use `LineItemLayout`.
- `input-layouts.tsx`:
  - Simplified to use `Section` instead.
- `AgentEditorPage.tsx`

## How Has This Been Tested?

No visual changes; screenshots / testing not required.

## Additional Options

### DIFF
--- a/web/src/refresh-components/cards/Card.tsx
+++ b/web/src/refresh-components/cards/Card.tsx
@@ -36,16 +36,7 @@
  * ```
  */
 
-import { cn } from "@/lib/utils";
 import { Section, SectionProps } from "@/layouts/general-layouts";
-
-const classNames = {
-  main: ["bg-background-tint-00 border"],
-  translucent: ["bg-transparent border border-dashed"],
-  disabled: [
-    "cursor-not-allowed pointer-events-none bg-background-tint-00 border opacity-50",
-  ],
-} as const;
 
 export interface CardProps extends SectionProps {
   // card variants
@@ -68,11 +59,9 @@ export default function Card({
   return (
     <div
       ref={ref}
-      className={cn(
-        "rounded-16 w-full h-full",
-        classNames[variant],
-        borderless && "border-none"
-      )}
+      className="card"
+      data-variant={variant}
+      data-borderless={borderless || undefined}
     >
       <Section alignItems="start" padding={padding} height="fit" {...props} />
     </div>
--- a/web/src/app/css/card.css
+++ b/web/src/app/css/card.css
@@ -1,0 +1,19 @@
+.card {
+  @apply rounded-16 w-full;
+}
+
+.card[data-variant="main"] {
+  @apply bg-background-tint-00 border;
+}
+
+.card[data-variant="translucent"] {
+  @apply bg-transparent border border-dashed;
+}
+
+.card[data-variant="disabled"] {
+  @apply cursor-not-allowed pointer-events-none bg-background-tint-00 border opacity-50;
+}
+
+.card[data-borderless] {
+  @apply border-none;
+}
--- a/web/src/app/globals.css
+++ b/web/src/app/globals.css
@@ -3,6 +3,7 @@
 
 @import "css/attachment-button.css";
 @import "css/button.css";
+@import "css/card.css";
 @import "css/code.css";
 @import "css/color-swatch.css";
 @import "css/colors.css";


web/src/layouts/general-layouts.tsx
```diff
@@ -77,38 +90,46 @@ export interface SectionProps
...
+  height?: Length;
+
   gap?: number;
   padding?: number;
-  fit?: boolean;
   wrap?: boolean;
-}
 
+  // Debugging utilities
+  dbg?: boolean;
```
nmgarza5, line 101: I wonder.... is there a way to propagate this prop to all components? 🤔 would be nice to tack this on anywhere

web/src/layouts/general-layouts.tsx
```diff
@@ -118,4 +139,65 @@ const Section = forwardRef<HTMLDivElement, SectionProps>(
...
+          {typeof title === "string" ? (
+            <Text
+              mainContentEmphasis
+              className={cn(strikethrough && "line-through")}
+            >
+              {title}
+            </Text>
+          ) : (
+            title
+          )}
```
nmgarza5: whats this typecheck for? what else could we be rendering besides text? some custom div?

can we change this to string? `title: React.ReactNode;`

web/src/layouts/input-layouts.tsx
```diff
@@ -173,52 +153,50 @@ function HorizontalInputLayout({
...
+        flexDirection="row"
+        justifyContent={center ? "center" : "start"}
+        gap={0}
+      >
+        <Text mainContentEmphasis text04>
+          {title}
         </Text>
         {optional && (
-          <Text text03 mainContentMuted as="span">
+          <Text text03 mainContentMuted>
```
nmgarza5: save to remove the span and p tags for `<Text />`? Seeing it in a few places

web/src/layouts/actions-layouts.tsx
```diff
@@ -455,32 +426,32 @@ function ActionsTool({
...
+              "bg-background-neutral-02 rounded-04 animate-pulse"
+            )}
+          />
+        )}
+        title={
+          <div className="h-4 bg-background-neutral-02 rounded-04 w-1/3 animate-pulse" />
+        }
+        description={
+          <div className="h-3 bg-background-neutral-02 rounded-04 w-2/3 animate-pulse" />
+        }
```
nmgarza5: why are empty divs set as these a title/description?

nmgarza5 (commented):
lgtm but we should be stricter with title/description imo. I don't reallllly understand why we allow react nodes to be passed for those props but convince me and I'll approve

web/src/layouts/general-layouts.tsx
raunakab: This is a great callout! I was originally planning on super-restricting this to be just `title: string` as well.

However, I realized after discussing with @duo-onyx that there could be cases in which we do want to render something more complex than just a regular string. For example, a "two-toned" title where there is a bolded major name, and a subscripted minor name. There are some use-cases in `MCPActionCard` as well where we render the "Edit Name" `IconButton` of the MCP server right beside the title (see some of @Subash-Mohan's changes in `ActionCard.tsx`).

web/src/layouts/input-layouts.tsx
raunakab: I generally check to see if there are any weird layout issues, but yes, it's generally safe to remove the `as="span"`. But always check first prior to removing.

In this case, everything looked fine to me (no changes at all in the UI).

web/src/layouts/actions-layouts.tsx
raunakab: This is for the skeleton. We want to render an animated, pulsing box to indicate to the user that something is loading.

web/src/layouts/general-layouts.tsx
raunakab: I would like this as well! Mainly for other components, I just do `className="... dbg-red"`. There are other debugging classes as well (`dbg-blue` and `dbg-green`) to help with layouts.

Since `Section` does not expose `className` or `style` props, we have to explicitly expose this `dbg` flag.

web/src/layouts/general-layouts.tsx
```diff
@@ -118,4 +139,101 @@ const Section = forwardRef<HTMLDivElement, SectionProps>(
...
+          <Text
+            mainContentEmphasis={!secondary}
+            text03={secondary}
+            className={cn(strikethrough && "line-through")}
+          >
+            {title}
+          </Text>
+        ) : (
+          title
+        )}
```
nmgarza5, line 210: we should just be able to do this now

```suggestion
        {loading ? (
          <div className="h-4 bg-background-neutral-01 rounded-08 w-1/3 animate-pulse" />
        ) :  (
          <Text
            mainContentEmphasis={!secondary}
            text03={secondary}
            className={cn(strikethrough && "line-through")}
          >
            {title}
          </Text>
        )}
```

web/src/layouts/general-layouts.tsx
```diff
@@ -118,4 +139,101 @@ const Section = forwardRef<HTMLDivElement, SectionProps>(
...
+
+        {/* Row 2: Description (column 2, or column 1 if no icon) */}
+        {loading && description ? (
+          <div className="h-6 bg-background-neutral-01 rounded-08 w-2/3 animate-pulse" />
+        ) : description ? (
+          <div className={cn("leading-none", Icon && "col-start-2")}>
+            {typeof description === "string" ? (
```
nmgarza5: same as above

web/src/layouts/general-layouts.tsx
raunakab: Oh right, totally forgot about this.

raunakab: Fixed!

web/src/layouts/general-layouts.tsx
raunakab: Fixed!

web/src/layouts/input-layouts.tsx
```diff
@@ -100,48 +97,31 @@ function VerticalInputLayout({
...
-        <div className="flex flex-col items-end">{children}</div>
-      </label>
-      {name && <ErrorLayout name={name} />}
-    </div>
+    <label htmlFor={name} className="cursor-pointer w-full">
+      <Section gap={0.25} alignItems="start">
+        <Section
+          flexDirection="row"
+          justifyContent="start"
+          alignItems={center ? "center" : "start"}
```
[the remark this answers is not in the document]
raunakab: This comment is incorrect. The intended behaviour is correct.

web/src/refresh-components/cards/Card.tsx
```diff
@@ -68,13 +59,11 @@ export default function Card({
   return (
     <div
       ref={ref}
-      className={cn(
-        "rounded-16 w-full h-full",
-        classNames[variant],
-        borderless && "border-none"
-      )}
+      className="card"
+      data-variant={variant}
```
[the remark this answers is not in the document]
raunakab: This is intended.

PR #7488, opened 2026-01-17
feat(url): Open url around snippet

Danelegend:
## Description
Currently the agent may decide to make a web_search against some query. This results in it getting a list of urls with snippets associated with those urls. The llm then determines which urls it wants to read based on title + snippet, and we then read that website via open_url tool and parse the first 15000 tokens of the website into the LLMs context window.

The snippet is the primary motivator of the LLM to choose a website to read. So we should base our reading around the snippet. This PR locates where the snippet is in the content of the website, and loads the content such that the snippet is in the middle and surrounding content above and below is loaded in. This way we are more likely to load content that the LLM is likely to find more relevant.

## How Has This Been Tested?
Played around with prompting on localhost.

There is a wide array of unit tests to test the output
(~3.8k lines of testing + data)

## Additional Options

---
## Summary by cubic
Center the open_url content around the search snippet so the LLM reads the most relevant parts of a page. This improves context quality and reduces wasted tokens.

- **New Features**
  - Find the snippet in page content (normalization + fuzzy token match) and truncate around it to a ~15k char window.
  - Extract a url→snippet map from search results and pass it through tool_runner to open_url; use it when building InferenceSections.
  - Fallback to the old truncation when no snippet is found.
  - Added comprehensive unit tests for snippet matching across HTML entities, zero-width chars, punctuation, and unicode cases.

<sup>Written for commit 35a04a19a381f3167d9f68d735daa61faef8aa25. Summary will update on new commits.</sup>


Danelegend:
@greptile

backend/tests/unit/onyx/tools/tool_implementations/websearch/data/tartan.txt
jessicasingh7, line 1: we could probably use git lfs/another file store for our test files (slack-https://onyx-company.slack.com/archives/C0771QKDBPE/p1768864722539549)

backend/tests/unit/onyx/tools/tool_implementations/websearch/test_websearch_utils.py
```diff
@@ -0,0 +1,210 @@
...
+
+    assert len(section.combined_content) == len(
+        TRUNCATED_CONTENT_PREFIX
+    ) + MAX_NUM_CHARS_WEB_CONTENT + len(TRUNCATED_CONTENT_SUFFIX)
+
+    no_prefix = section.combined_content[len(TRUNCATED_CONTENT_PREFIX) :]
+    no_affix = no_prefix[:MAX_NUM_CHARS_WEB_CONTENT]
+
+    # expected start index of the snippet
+    expected_start_idx = 215398
```
jessicasingh7: would prefer no magic numbers- can either add calculation/constant here!

backend/tests/unit/onyx/tools/tool_implementations/websearch/test_websearch_utils.py
```diff
@@ -0,0 +1,210 @@
...
```
jessicasingh7: same as above^

backend/tests/unit/onyx/tools/tool_implementations/open_url/data/test_snippet_finding_data.json
```diff
@@ -0,0 +1,929 @@
...
+						"snippet_located": true,
+						"expected_start_idx": 235,
+						"expected_end_idx": 448
+					},
+					"allow_buffer": true
+				}
+			]
+		},
+		{
+			"category": "no_match",
```
jessicasingh7, line 529: not sure if we need as verbose "content" for no_match case

backend/onyx/tools/tool_implementations/open_url/snippet_matcher.py
```diff
@@ -0,0 +1,328 @@
...
+
+    # Sort entities by length (longest first) for greedy matching
+    sorted_entities = sorted(html_entities.keys(), key=len, reverse=True)
+
+    result_chars = []
+    result_map = []
+    i = 0
+    last_was_space = True  # Track to avoid leading spaces
+
+    def normalize_char(c: str) -> str:
```
jessicasingh7: I wonder if we want to put this in a utils file especially if it's just based on text and generic- up to personal preference

backend/onyx/tools/tool_implementations/open_url/snippet_matcher.py
```diff
@@ -0,0 +1,328 @@
...
+        return NegativeSnippetMatchResult
+
+    best_score = 0.0
+    best_word_pos = -1
+
+    # Sliding window through content words
+    for i in range(0, len(content_words) - window_size + 1):
+        window = " ".join(content_words[i : i + window_size])
+
+        score = fuzz.ratio(processed_snippet, window)
```
jessicasingh7: this would be a good use case for `fuzz.partial_ratio_alignment` which gives similarity score and does sliding window for you

backend/tests/unit/onyx/tools/tool_implementations/open_url/data/test_snippet_finding_data.json
Danelegend: I'd disagree here. A false positive here is probably worse than a false negative. I'd argue that because we are fuzzy searching (which although isn't a black box, let's just think of it as a black box that spits out a somewhat random number), we want to be sure that we're not going to have false positives

Danelegend:
@greptile

backend/tests/unit/onyx/tools/tool_implementations/websearch/data/tartan.txt
Danelegend: Just had a chat w/ jamison. We decided the best course of action is to leave this here for now

backend/onyx/tools/tool_implementations/open_url/snippet_matcher.py
```diff
@@ -0,0 +1,241 @@
...
+
+    res = fuzz.partial_ratio_alignment(processed_content, processed_snippet)
+
+    if not res:
+        return NegativeSnippetMatchResult
+
+    score = res.score
+
+    if score >= (min_threshold * 100):
+        start_idx = res.src_start
```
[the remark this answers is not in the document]
jessicasingh7: @Danelegend is this valid?

PR #7531, opened 2026-01-19
fix(chrome extension): user panel and admin settings removed from new tab page

rohoswagger:
## Description

user panel and admin settings removed from new tab page

## How Has This Been Tested?

Locally

## Additional Options

---
## Summary by cubic
Stops scrolling on the NRF new tab page and simplifies the NRF UI, adding a quick “Open in Onyx” action. Adds a display mode to distinguish new-tab vs side-panel behavior.

- **Bug Fixes**
  - New tab page: container now uses overflow hidden (removed min-height: 100vh) to prevent page scroll.
  - Introduced NRFDisplayMode enum and replaced isSidePanel prop; pages pass NEW_TAB or SIDE_PANEL explicitly.
  - Side panel: hide the entire sidebar and the header/toggle.
  - All NRF pages: hide settings and admin panel entries in the sidebar.

<sup>Written for commit 77f287835cf5298ef5e23f2e162862a4e32d87e6. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/app/css/nrf.css
+++ b/web/src/app/css/nrf.css
@@ -37,24 +37,31 @@
   gap: 0.5rem;
 }
 
-/* Settings button */
-.nrf-settings-button-container {
+/* Header buttons (Open in Onyx + Settings) */
+.nrf-header-buttons-container {
   position: absolute;
   top: 0;
   right: 0;
   padding: 1rem;
   z-index: 10;
+  display: flex;
+  align-items: center;
+  gap: 0.5rem;
 }
 
-.nrf-settings-button {
-  background-color: var(--mask-02) !important;
-  backdrop-filter: blur(12px);
-  border-radius: var(--border-radius-full);
-  box-shadow: 0px 2px 12px 0px var(--shadow-02);
+.nrf-header-button {
+  background-color: var(--background-tint-02) !important;
+  border-radius: var(--border-radius-12);
 }
 
-.nrf-settings-button:hover {
-  background-color: var(--mask-03) !important;
+/* Settings icon button specific sizing */
+button.nrf-header-button[aria-label="Open settings"] {
+  width: 2.5rem !important;
+  height: 2.5rem !important;
+}
+
+.nrf-header-button:hover {
+  background-color: var(--background-tint-03) !important;
 }
 
 /* Dropzone container */
@@ -153,6 +160,9 @@
   transform: translate(-50%, -50%);
   width: 90%;
   max-width: 48rem;
+  display: flex;
+  flex-direction: column;
+  gap: 1.5rem;
 }
 
 @media (min-width: 1024px) {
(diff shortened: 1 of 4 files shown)


web/src/layouts/app-layouts.tsx
```diff
@@ -304,66 +308,67 @@ function AppHeader() {
...
-          {/* Left - contains the icon-button to fold the AppSidebar on mobile */}
-          <div className="flex-1">
-            <IconButton
-              icon={SvgSidebar}
-              onClick={() => setFolded(false)}
-              className={cn(!isMobile && "invisible")}
-              internal
-            />
-          </div>
+      {!isNRFSidePanel &&
```
jmelahman, line 311: Do we need any of this component i.e. the modals, etc.? It might be an anti-pattern in react, but I'd prefer just exiting early,

```
if (isNRFSidePanel) return null;
```

I also plan on moving a lot of this stuff out of `app-layouts` in [fix(fe): chat header is sticky and transparent](https://github.com/onyx-dot-app/onyx/pull/7487) because I don't think it belongs here (and your change kinda proves that point). I would follow [this](https://github.com/onyx-dot-app/onyx/pull/7487/changes#diff-8d1768d2ca4931ea25d6a1e94aca5b90c80e7f5acae53d74a607794521ec0737R665-R667) pattern (rather than implement a custom header for your page -- can may be a separate change).

I also think making the header default and opt-out is wrong -- each page should just define its header IMO (at least for now, make it makes more sense in the future with a more consistent design language).

web/src/app/css/nrf.css
```diff
@@ -37,24 +37,31 @@
...
+  gap: 0.5rem;
 }
 
-.nrf-settings-button {
-  background-color: var(--mask-02) !important;
-  backdrop-filter: blur(12px);
-  border-radius: var(--border-radius-full);
-  box-shadow: 0px 2px 12px 0px var(--shadow-02);
+.nrf-header-button {
+  background-color: var(--background-tint-02) !important;
```
jmelahman, line 53: I'm down to iterate quickly and these changes should be isolated enough (all of the classes are prefix/namespaced with `nrf-`), but `!important` is a footgun we should avoid when possible lol

web/src/sections/sidebar/AppSidebar.tsx
```diff
@@ -546,6 +552,12 @@ MemoizedAppSidebarInner.displayName = "AppSidebar";
 export default function AppSidebar() {
   const { folded, setFolded } = useAppSidebarContext();
   const { isMobile } = useScreenSize();
+  const pathname = usePathname();
+
+  // Hide entire sidebar on NRF side-panel page
+  if (pathname === "/chat/nrf/side-panel") {
```
jmelahman, line 558: Similar to the header comment. Kinda hacky imo. Would rather the NRFPage not even hit this code path -- probably add a prop higher up which conditionally tries to render this component. I'm not immediately sure exactly where `AppSidebar` comes into play tho.

PR #7533, opened 2026-01-19
fix(chrome extension): auto focus on input bar

rohoswagger:
## Description

when you open a new tab and click tab, it auto focuses on the chat input bar

## How Has This Been Tested?

locally

## Additional Options

---
## Summary by cubic
Auto-focuses the chat input and sets tab order so the NRF new tab focuses it first; makes the “Open in Onyx” and settings header buttons focusable via Tab.

<sup>Written for commit 0054adaafe150d9152cdc2172888638014ee0647. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/app/chat/components/input/ChatInputBar.tsx
+++ b/web/src/app/chat/components/input/ChatInputBar.tsx
@@ -110,6 +110,7 @@ export interface ChatInputBarProps {
   setPresentingDocument?: (document: MinimalOnyxDocument) => void;
   toggleDeepResearch: () => void;
   disabled: boolean;
+  tabIndex?: number;
 }
 
 const ChatInputBar = React.memo(
@@ -136,6 +137,7 @@ const ChatInputBar = React.memo(
         toggleDeepResearch,
         setPresentingDocument,
         disabled,
+        tabIndex,
       },
       ref
     ) => {
@@ -511,6 +513,7 @@ const ChatInputBar = React.memo(
                 "pt-3"
               )}
               autoFocus
+              tabIndex={tabIndex}
               style={{ scrollbarWidth: "thin" }}
               role="textarea"
               aria-multiline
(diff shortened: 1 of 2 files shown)


web/src/app/chat/nrf/NRFPage.tsx
```diff
@@ -316,15 +320,25 @@ export default function NRFPage({ isSidePanel = false }: NRFPageProps) {
...
+      {/* Header buttons */}
       {!isSidePanel && (
-        <div className="nrf-settings-button-container">
+        <div className="nrf-header-buttons-container">
+          <Button
+            tertiary
+            rightIcon={SvgExternalLink}
+            onClick={handleOpenInOnyx}
+            className="nrf-header-button"
+            tabIndex={-1}
```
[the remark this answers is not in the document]
jmelahman: I agree with this comment and for the most part, we probably shouldn't be setting `-1` tab indices on these elements

web/src/app/chat/nrf/NRFPage.tsx
```diff
@@ -328,6 +328,7 @@ export default function NRFPage({
             rightIcon={SvgExternalLink}
             onClick={handleOpenInOnyx}
             className="nrf-header-button"
+            tabIndex={0}
```
jmelahman, line 331:
```suggestion
```
`0` should be the default for buttons

web/src/app/chat/components/input/ChatInputBar.tsx
```diff
@@ -110,6 +110,7 @@ export interface ChatInputBarProps {
   setPresentingDocument?: (document: MinimalOnyxDocument) => void;
   toggleDeepResearch: () => void;
   disabled: boolean;
+  tabIndex?: number;
```
jmelahman, line 113: I would leave a comment with as much technical detail for this as possible. "This ensures the ChatInputBar is focusable on tab" is good, but ideal imo is something like "Some pages, for example the NRFPage, are served as an iframe which breaks `autofocus`, so we allow the tab priority of the input textarea to be increased", etc.

PR #7534, opened 2026-01-19
fix(chrome extension): remove bottom padding in chrome extension once message is sent

rohoswagger:
## Description

remove bottom padding in chrome extension once message is sent

## How Has This Been Tested?

Fixes
image

Locally

## Additional Options

---
## Summary by cubic
Remove extra bottom spacing in the NRF Chrome extension chat so the content sits flush after sending messages. Makes the input clearance margin configurable and disables it in NRF.

- **Bug Fixes**
  - Disable bottom margin in NRFPage (applyBottomMargin=false) to remove unused space in the extension.

- **Refactors**
  - Add applyBottomMargin prop to ChatScrollContainer and ChatUI (default: true).

<sup>Written for commit 7b40fb9a13bd3839a2b68101cbf6c39114eca036. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/app/chat/nrf/NRFPage.tsx
+++ b/web/src/app/chat/nrf/NRFPage.tsx
@@ -374,6 +374,7 @@ export default function NRFPage({
                         stopGenerating={stopGenerating}
                         handleResubmitLastMessage={handleResubmitLastMessage}
                         deepResearchEnabled={deepResearchEnabled}
+                        applyBottomMargin={false}
                       />
                     </div>
                   </div>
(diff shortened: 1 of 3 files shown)


web/src/app/chat/nrf/NRFPage.tsx
```diff
@@ -374,6 +374,7 @@ export default function NRFPage({
                         stopGenerating={stopGenerating}
                         handleResubmitLastMessage={handleResubmitLastMessage}
                         deepResearchEnabled={deepResearchEnabled}
+                        applyBottomMargin={false}
```
jmelahman, line 377: This smells like a hack. If I were to wager a guess, it's because the ChatPage and NRFPage have diverged too much already :/


Let's table this change until I land [fix(fe): chat header is sticky and transparent](https://github.com/onyx-dot-app/onyx/pull/7487). That should simplify the structure here a bit and likely the best fix is to share as much implementation here as with the ChatPage.

PR #7538, opened 2026-01-19
fix: tool call handling for weak models

yuhongsun96:
## Description
Some bad LLMs will fail to output tool calls when it should. This could be due to the model or the serving layer. Added failure case handling to try to recover from those cases.

## How Has This Been Tested?
Tested by asking LLMs to explicitly output tool calls in the reasoning or output channels instead of the tool channels. Hardcoded some temporary logic to ensure this can trigger even if there is actual final output.

## Additional Options

---
## Summary by cubic
Adds a fallback that extracts tool calls from LLM response text when models fail to emit structured tool calls. Improves reliability for forced-tool flows and low-quality or non-tool-calling models.

- **New Features**
  - Fallback parses JSON-like snippets in answer, then reasoning, and maps them to known tools.
  - Supports common formats: {"name","arguments"}, {"function":{...}}, {"tool_name": {...}}, or args matching a tool’s schema.
  - Triggers when tool_choice is REQUIRED with no tool calls, or when there’s reasoning but no answer or tools.
  - Only one fallback attempt per turn to avoid loops; logs extracted call count.
  - Clearer error message when no answer is returned, pointing to tool-calling configuration or invalid tool outputs.

<sup>Written for commit 58e34f467979442f6dc037bfc505c520af60e2c6. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/chat/llm_step.py
+++ b/backend/onyx/chat/llm_step.py
@@ -49,6 +49,7 @@ from onyx.tools.models import ToolCallKickoff
 from onyx.tracing.framework.create import generation_span
 from onyx.utils.b64 import get_image_type_from_bytes
 from onyx.utils.logger import setup_logger
+from onyx.utils.text_processing import find_json_objects_in_text
 
 logger = setup_logger()
 
@@ -278,40 +279,6 @@ def _extract_tool_call_kickoffs(
     return tool_calls
 
 
-def _find_json_objects_in_text(text: str) -> list[dict[str, Any]]:
-    """Find all JSON objects in a text string.
-
-    Searches for JSON objects by looking for balanced braces and attempting to parse.
-    Returns a list of successfully parsed JSON objects.
-    """
-    json_objects: list[dict[str, Any]] = []
-    i = 0
-
-    while i < len(text):
-        if text[i] == "{":
-            # Try to find a matching closing brace
-            brace_count = 0
-            start = i
-            for j in range(i, len(text)):
-                if text[j] == "{":
-                    brace_count += 1
-                elif text[j] == "}":
-                    brace_count -= 1
-                    if brace_count == 0:
-                        # Found potential JSON object
-                        candidate = text[start : j + 1]
-                        try:
-                            parsed = json.loads(candidate)
-                            if isinstance(parsed, dict):
-                                json_objects.append(parsed)
-                        except json.JSONDecodeError:
-                            pass
-                        break
-        i += 1
-
-    return json_objects
-
-
 def extract_tool_calls_from_response_text(
     response_text: str | None,
     tool_definitions: list[dict],
@@ -347,7 +314,7 @@ def extract_tool_calls_from_response_text(
         return []
 
     # Find all JSON objects in the response text
-    json_objects = _find_json_objects_in_text(response_text)
+    json_objects = find_json_objects_in_text(response_text)
 
     tool_calls: list[ToolCallKickoff] = []
     tab_index = 0
@@ -370,10 +337,9 @@ def extract_tool_calls_from_response_text(
             )
             tab_index += 1
 
-    if tool_calls:
-        logger.info(
-            f"Extracted {len(tool_calls)} tool call(s) from response text as fallback"
-        )
+    logger.info(
+        f"Extracted {len(tool_calls)} tool call(s) from response text as fallback"
+    )
 
     return tool_calls
 
@@ -439,8 +405,8 @@ def _try_match_json_to_tool(
         if not properties:
             continue
 
-        # Check if all required parameters are present
-        if required and all(req in json_obj for req in required):
+        # Check if all required parameters are present (empty required = all optional)
+        if all(req in json_obj for req in required):
             # Check if any of the tool's properties are in the JSON object
             matching_props = [prop for prop in properties if prop in json_obj]
             if matching_props:
--- a/backend/onyx/utils/text_processing.py
+++ b/backend/onyx/utils/text_processing.py
@@ -21,6 +21,28 @@ ESCAPE_SEQUENCE_RE = re.compile(
     re.UNICODE | re.VERBOSE,
 )
 
+_INITIAL_FILTER = re.compile(
+    "["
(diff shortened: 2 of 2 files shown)


backend/onyx/chat/llm_step.py
```diff
@@ -278,6 +278,179 @@ def _extract_tool_call_kickoffs(
     return tool_calls
 
 
+def _find_json_objects_in_text(text: str) -> list[dict[str, Any]]:
```
evan-onyx: surely there's a util for this

backend/onyx/chat/llm_step.py
```diff
@@ -278,6 +278,179 @@ def _extract_tool_call_kickoffs(
...
+    """Find all JSON objects in a text string.
+
+    Searches for JSON objects by looking for balanced braces and attempting to parse.
+    Returns a list of successfully parsed JSON objects.
+    """
```
evan-onyx: with nested json this returns all inner json objs too, is that really what we want?

backend/onyx/chat/llm_step.py
```diff
@@ -278,6 +278,179 @@ def _extract_tool_call_kickoffs(
...
+    Returns:
+        List of ToolCallKickoff objects for any matched tool calls
+    """
+    if not response_text or not tool_definitions:
+        return []
+
+    # Build a map of tool names to their definitions
+    tool_name_to_def: dict[str, dict] = {}
+    for tool_def in tool_definitions:
+        if tool_def.get("type") == "function" and "function" in tool_def:
```
evan-onyx, line 307: string constants

evan-onyx: this check also feels hacky to me, are we sure this is the right way to extract tools from definitions?

backend/onyx/chat/llm_step.py
```diff
@@ -278,6 +278,179 @@ def _extract_tool_call_kickoffs(
...
+    tool_calls: list[ToolCallKickoff] = []
+    tab_index = 0
+
+    for json_obj in json_objects:
+        matched_tool_call = _try_match_json_to_tool(json_obj, tool_name_to_def)
+        if matched_tool_call:
+            tool_name, tool_args = matched_tool_call
+            tool_calls.append(
+                ToolCallKickoff(
+                    tool_call_id=f"extracted_{uuid.uuid4().hex[:8]}",
```
evan-onyx, line 328: we should probably have a util func for this

backend/onyx/chat/llm_step.py
```diff
@@ -278,6 +278,179 @@ def _extract_tool_call_kickoffs(
...
+                    placement=Placement(
+                        turn_index=placement.turn_index,
+                        tab_index=tab_index,
+                        sub_turn_index=placement.sub_turn_index,
+                    ),
+                )
+            )
+            tab_index += 1
+
+    if tool_calls:
```
evan-onyx: imo this can be unconditional, if this function ends up being called its good to have more info about what happened

backend/onyx/chat/llm_step.py
```diff
@@ -278,6 +278,179 @@ def _extract_tool_call_kickoffs(
...
+    if "function" in json_obj and isinstance(json_obj["function"], dict):
+        func_obj = json_obj["function"]
+        if "name" in func_obj and func_obj["name"] in tool_name_to_def:
+            tool_name = func_obj["name"]
+            arguments = func_obj.get("arguments", func_obj.get("parameters", {}))
+            if isinstance(arguments, str):
+                try:
+                    arguments = json.loads(arguments)
+                except json.JSONDecodeError:
+                    arguments = {}
```
evan-onyx, line 388: this feels pretty sus to me, I think we should just "fail" here

backend/onyx/chat/llm_step.py
```diff
@@ -278,6 +278,179 @@ def _extract_tool_call_kickoffs(
...
+    """
+    # Format 1: Direct tool call format {"name": "...", "arguments": {...}}
+    if "name" in json_obj and json_obj["name"] in tool_name_to_def:
+        tool_name = json_obj["name"]
+        arguments = json_obj.get("arguments", json_obj.get("parameters", {}))
+        if isinstance(arguments, str):
+            try:
+                arguments = json.loads(arguments)
+            except json.JSONDecodeError:
+                arguments = {}
```
evan-onyx, line 374: this feels pretty sus to me, I think we should just "fail" here

backend/onyx/chat/llm_step.py
```diff
@@ -278,6 +278,179 @@ def _extract_tool_call_kickoffs(
...
+
+    Args:
+        json_obj: The JSON object to match
+        tool_name_to_def: Map of tool names to their function definitions
+
+    Returns:
+        Tuple of (tool_name, tool_args) if matched, None otherwise
+    """
+    # Format 1: Direct tool call format {"name": "...", "arguments": {...}}
+    if "name" in json_obj and json_obj["name"] in tool_name_to_def:
```
evan-onyx, line 367: strings constants

backend/onyx/chat/llm_step.py
yuhongsun96: It's necessary, I'll add a comment on why

true-42:
Hello, thanks for the fix. Most models work better, mistral-large-3 has unfortunately now a problem, as follows:

With Onyx 2.10, many models had a problem with tool call web_search. 
With Onyx 2.11, these model work. BUT: one of my favorites, mistral-large-3 675B (access by Ollama cloud) worked before with no problems. Now, it is failing as follows

- Chat input "news today?"
- Websearch starts
- Connection Error: API connection error: Failed to connect to the API. Please check your internet connection. Model: mistral-large-3:675b (ollama_chat)
- Log: [onyx_error_log_websearch_mistral-large-3.txt](https://github.com/user-attachments/files/25468606/onyx_error_log_websearch_mistral-large-3.txt)
- An update to litellm Version: 1.81.6 was tested without improvement
- 
Any chance to check this? Kind regards

PR #7546, opened 2026-01-20
fix: handle missing tenant_id in license enforcement middleware

nmgarza5:
## Description

The license enforcement middleware was crashing when the tenant_id context variable wasn't set yet. This can happen if the middleware runs before tenant identification.

Now passes through gracefully if tenant context isn't established.

## Additional Options

---
## Summary by cubic
Fixed a crash in the license enforcement middleware when tenant_id isn’t set yet. The middleware now reads CURRENT_TENANT_ID_CONTEXTVAR and skips enforcement when None, allowing the request to pass through until tenant identification runs.

<sup>Written for commit 85ab8e207c7405861a0e1b2323e3fd3de1c040a3. Summary will update on new commits.</sup>


nmgarza5:
not needed

PR #7573, opened 2026-01-20
feat: Add Langfuse tracing integration with trace ID persistence and scoring

ntgussoni:
### Summary
Adds direct Langfuse integration with trace ID persistence and a framework for trace scoring.

### Changes

**Tracing Integration**
- Added `LangfuseTracingProcessor` for direct Langfuse SDK integration
- Replaces OpenTelemetry/OpenInference path with direct SDK calls
- Captures user ID, session ID, input/output, and generation metrics

**Trace ID Persistence**
- Added `trace_id` column to `ChatMessage` model with migration
- Sets `trace_id` on assistant messages when traces start
- Enables linking user feedback to traces after completion

**Scoring Framework**
- Added optional `score_trace()` method to tracing processor interface
- Implemented scoring in `LangfuseTracingProcessor` for user feedback
- Refactored feedback system to use tracing framework instead of direct Langfuse calls

### Benefits
- Consistent trace IDs between internal system and Langfuse
- Persistent trace IDs enable post-completion feedback scoring
- Extensible scoring framework for future tracing backends

---
## Summary by cubic
Integrated Langfuse tracing with deterministic trace IDs derived from session and message, enabling feedback scoring without storing IDs in the database, and removed the OpenTelemetry/OpenInference path.

- **New Features**
  - Added LangfuseTracingProcessor for direct Langfuse SDK integration.
  - Deterministic trace_id generation in run_llm_loop; feedback reconstructs and scores traces via score_trace.

- **Refactors**
  - Removed trace_id from ChatMessage and DB persistence; tracing setup now uses add_trace_processor.

<sup>Written for commit 1b2b9bb1f8874351bcd9b336d7d4d184248288cd. Summary will update on new commits.</sup>


backend/alembic/versions/087e708d6e26_add_trace_id_to_chat_message.py
```diff
@@ -0,0 +1,28 @@
...
+
+def upgrade() -> None:
+    op.add_column(
+        "chat_message",
+        sa.Column("trace_id", sa.String(), nullable=True),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("chat_message", "trace_id")
```
justin-tahara: Pretty sure this is quite a dangerous operation as this column is large for many customers and making any edits should be carefully checked before we do this. 

Why do we want to add a trace_id to the chat_message? Is this not something we can just extract later?

backend/onyx/chat/llm_loop.py
```diff
@@ -373,15 +374,38 @@ def run_llm_loop(
...
     chat_session_id: str | None = None,
     include_citations: bool = True,
+    assistant_message: Any | None = None,
 ) -> None:
+    # Get the user's input message (last user message in history)
+    user_input = None
+    for msg in reversed(simple_chat_history):
+        if msg.message_type == MessageType.USER:
+            user_input = msg.message
+            break
```
justin-tahara: Quite a costly operation and we should try to avoid. @yuhongsun96 may have a better idea on how we can do this.

backend/onyx/db/feedback.py
```diff
@@ -263,6 +263,30 @@ def create_chat_message_feedback(
...
+                f"feedback_text: {feedback_text}"
+            )
+            score_value = 1.0 if is_positive else 0.0
+            from onyx.tracing.framework import score_trace
+
+            score_trace(
+                trace_id=chat_message.trace_id,
+                score=score_value,
+                comment=feedback_text,
+            )
```
justin-tahara: Does this work for langfuse and braintrust or just langfuse?

backend/onyx/db/models.py
justin-tahara, line 1: Again something we need to discuss before we allowing

justin-tahara (commented):
Main comments are around the design of how we are adding to existing structure

backend/onyx/chat/llm_loop.py
ntgussoni: I removed this completely

backend/onyx/db/feedback.py
ntgussoni: the method is in the abstract class and **can** be overwritten by the implementor if needed, but not mandatory. Im unfamiliar with braintrust but I assume they have a way as well

ntgussoni:
@yuhongsun96  @justin-tahara Thanks for reviewing, 

I've updated and removed most of the sensitive code you mentioned.

The scoring is connected to the original trace by the trace_id hence why I was storing it along with the message before.
I've now changed it so that the trace_id is created deterministically based on the `thread_id + chat_message_id`.
This way there's no need to store anything else, or modify the schema, and we can re-create it when feedback happens.

There was a history.reverse being done on the llm_loop, I removed that as well because the original message can be obtained from the processor.

I hope this makes it easier to review now

PR #7582, opened 2026-01-20
feat(ee): Add daily license refresh Celery task

nmgarza5:
## Description

For self-hosted EE deployments, adds a daily Celery task that runs at 2 AM to keep the license fresh:

1. Fetches fresh license from control plane
2. Verifies the license signature
3. Persists to database and updates cache

If the control plane is unreachable, the task falls back to the existing cached license. This supports air-gapped deployments where the license remains valid based on its `expires_at` field.

Closes https://linear.app/onyx/issue/ENG-3398

## How Has This Been Tested?


nmgarza5:
cancelling for now. can reopen if necessary

PR #7589, opened 2026-01-20
feat(billing): hardcode Stripe publishable key for self-hosted deployments

nmgarza5:
## Description

Hardcodes the production Stripe publishable key so self-hosted deployments can access Stripe billing without requiring manual configuration.

- The env var `NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY` override is preserved for testing with `pk_test_*` keys
- Added explicit comments explaining why this is safe (publishable keys are designed to be public)
- Publishable keys can only initialize Stripe.js and tokenize payment info - they cannot make charges or access sensitive data

## Additional Options

- [x] Override Linear Check


justin-tahara (approved):

nmgarza5:
Superseded by #7595 - using S3 instead of hardcoding

PR #7590, opened 2026-01-20
chore(tests): Mock LLM

Danelegend:
## Description
Adds a framework for testing with an LLM that we can control.

Allows us to set what the response should be and forward through tokens during runtime.

## How Has This Been Tested?
This is the test

## Additional Options

---
## Summary by cubic
Adds a packet-based streaming chat test suite using a controllable mock LLM to validate AnswerStreamPart ordering and placement.

- **New Features**
  - Added tests that verify stream flow: MessageResponseIDInfo → AgentResponseStart → per-token AgentResponseDelta packets → OverallStop.
  - Introduced MockLLM with a SyncStreamController (forward and forward_till_end) and a patch for get_llm_for_persona to control token emission.

<sup>Written for commit 974bea0c06a0ef89175e1f261fb869218dde0b23. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/external_dependency_unit/mock_llm.py
+++ b/backend/tests/external_dependency_unit/mock_llm.py
@@ -35,7 +35,7 @@ class MockLLMController(abc.ABC):
 
 
 class MockLLM(LLM, MockLLMController):
-    def __init__(self):
+    def __init__(self) -> None:
         self.stream_controller: SyncStreamController | None = None
 
     def set_response(self, response_tokens: list[str]) -> None:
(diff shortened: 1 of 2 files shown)


backend/tests/external_dependency_unit/mock_llm.py
```diff
@@ -0,0 +1,162 @@
...
+        prompt: LanguageModelInput,
+        tools: list[dict] | None = None,
+        tool_choice: ToolChoiceOptions | None = None,
+        structured_response_format: dict | None = None,
+        timeout_override: int | None = None,
+        max_tokens: int | None = None,
+        reasoning_effort: ReasoningEffort | None = None,
+        user_identity: LLMUserIdentity | None = None,
+    ) -> ModelResponse:
+        raise NotImplementedError("We only care about streaming atm")
```
evan-onyx, line 76: might as well implement this by collecting the streamcontroller? If it's annoying then it's fine

evan-onyx (approved):

backend/tests/external_dependency_unit/mock_llm.py
Danelegend: It be annoying .-.

PR #7618, opened 2026-01-21
chore(fe): move chatpage footer inside background element

jmelahman:
## Description

Disables the AppFooter in the ChatPage and instead exports it so it can be placed inside the element with the custom background.

## How Has This Been Tested?
]
20260120_19h54m24s_grim

20260120_19h54m17s_grim

## Additional Options

- [x] Override Linear Check

---
## Summary by cubic
Moves the ChatPage and NRFPage footers inside their background containers so they match the custom background and spacing. Adds a way to disable the default layout footer and use the exported Footer in both pages.

- **Refactors**
  - AppLayouts.Root: add disableFooter prop (deprecated) to hide the default footer.
  - Rename AppFooter to Footer and export it.
  - ChatPage and NRFPage: render <AppLayouts.Footer /> inside the background container and change offset from bottom-0 to bottom-6 when a session has messages.

<sup>Written for commit d3463f959959a9ca9dba3b314de88cedc70eff70. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/app/chat/nrf/NRFPage.tsx
+++ b/web/src/app/chat/nrf/NRFPage.tsx
@@ -31,6 +31,7 @@ import MessageList from "@/components/chat/MessageList";
 import ChatScrollContainer from "@/components/chat/ChatScrollContainer";
 import WelcomeMessage from "@/app/chat/components/WelcomeMessage";
 import useChatSessions from "@/hooks/useChatSessions";
+import * as AppLayouts from "@/layouts/app-layouts";
 import { cn } from "@/lib/utils";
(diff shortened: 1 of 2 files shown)


web/src/layouts/app-layouts.tsx
```diff
@@ -180,18 +180,19 @@ function AppFooter() {
  * ```
  */
 export interface AppRootProps {
+  DEPRECATEDDisableFooter?: boolean;
```
jmelahman: I think there is a more idiomatic way to indicate this shouldn't be used (which I'll push in a sec), but you get the idea

PR #6743, opened 2025-12-15
feat: add slash commands for navigation and tool toggling

nmgarza5:
## Description

Add a slash command system to the chat input with a unified, collapsible menu. Users can type "/" to see available commands organized in three groups:

**Tools**:
- `/upload` - Open file picker to upload files
- Tool toggle commands: dynamically generated from assistant's available tools (e.g., `/web-search`, `/internal-search`, `/image-generation`, etc.)

**Navigation**:
- `/new-chat`, `/agents`, `/create-agent`
- Admin-only: `/settings`, `/connectors`, `/users`, `/document-sets`, `/llm-config`, `/add-connector`

**Prompts** :
- Prompt shortcuts: dynamically generated from saved prompts
- `/new-prompt` - Create a new prompt shortcut

## How Has This Been Tested?

Manual testing

Non-admin

Admin

E2E flow with tool calling + prompts
<video src="https://github.com/user-attachments/assets/d5969885-decd-4359-8647-8c13e22342e1" width="500" controls></video>

## Additional Options

- [x] Override Linear Check

---
## Summary by cubic
Adds slash commands to the chat input for quick navigation, tool toggling, and prompt shortcuts. Type “/” to open a grouped menu and run commands without sending a chat message.

- **New Features**
  - Menu appears on “/”; Arrow keys navigate; Tab/Enter select; Esc cancels. Groups are collapsible: Tools, Navigation, Prompts.
  - Navigation: /new-chat, /agents, /create-agent; admin-only: /settings, /connectors, /users, /document-sets, /llm-config, /add-connector.
  - Tools: commands generated from the assistant’s tools; disabled tools omitted; selection toggles forced tools; added /upload to open the file picker.
  - Prompts: commands generated from saved prompts; selection inserts content; added /new-prompt to create a prompt. The slash menu replaces the old prompt shortcuts UI.

<sup>Written for commit 1845069415e6057ccb176d4c3cc7406f43bcc67c. Summary will update automatically on new commits.</sup>

PR #7619, opened 2026-01-21
feat: improve prompt shortcuts with uniqueness constraints and enhancements

raunakab:
## Description

Backend:
- Add unique constraint on `(prompt, user_id)` for user-owned prompts
- Add partial unique index for public prompts (`user_id IS NULL`)
- Add validation to prevent duplicate prompt shortcut names
- Fix authorization to properly restrict public prompt modifications

Frontend:
- Improve prompt selection UI with `Popover` component
- Add proper keyboard navigation (Tab, Shift+Tab, Arrow keys)
- Use `useFilter` hook for efficient prompt filtering
- Add `activePromptShortcuts` to hook for filtering active prompts
- Remove standalone `input-prompts/page.tsx` (management moved to settings, merging in a future PR)

## Additional Options

---
## Summary by cubic
Enforces unique prompt shortcut names per user and for public prompts, and upgrades the in-chat prompt picker with a popover and full keyboard navigation. Prevents duplicates, tightens auth for public prompts, and moves prompt management to Settings.

- **New Features**
  - Enforced uniqueness: (prompt, user_id) and a partial index for public prompts.
  - Server-side validation returns 409 on duplicate names.
  - Prompt picker uses a Popover with Tab, Shift+Tab, and Arrow key navigation.
  - Efficient filtering via useFilter; added activePromptShortcuts to the hook.
  - Removed the /chat/input-prompts page; management lives in Settings.

- **Bug Fixes**
  - Correctly restrict modifications to public prompts; only owners can edit user-owned prompts.
  - Fixed InputPrompt__User.user_id foreign key to reference user.id.

<sup>Written for commit 4798fd232afd6476e8c61a890d4164f5ef9d8381. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/app/chat/components/input/ChatInputBar.tsx
+++ b/web/src/app/chat/components/input/ChatInputBar.tsx
@@ -316,6 +316,11 @@ const ChatInputBar = React.memo(
       [filteredPrompts]
     );
 
+    // Reset tabbingIconIndex when filtered prompts change to avoid out-of-bounds
+    useEffect(() => {
+      setTabbingIconIndex(0);
+    }, [filteredPrompts]);
+
     const handlePromptInput = useCallback(
       (text: string) => {
         if (text.startsWith("/")) {


web/src/app/chat/components/input/ChatInputBar.tsx
```diff
@@ -110,385 +120,334 @@ export interface ChatInputBarProps {
   setPresentingDocument?: (document: MinimalOnyxDocument) => void;
   toggleDeepResearch: () => void;
   disabled: boolean;
+  ref?: React.Ref<ChatInputBarHandle>;
```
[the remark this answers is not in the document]
raunakab: Invalid. We're using React 19, so passing `ref`s directly is valid.

web/src/app/chat/components/input/ChatInputBar.tsx
```diff
@@ -110,385 +120,334 @@ export interface ChatInputBarProps {
...
+    ]);
+
+    function handleKeyDown(e: React.KeyboardEvent<HTMLTextAreaElement>) {
+      if (!showPrompts) return;
+
+      if (e.key === "Enter") {
+        e.preventDefault();
+        if (tabbingIconIndex === filteredPrompts.length) {
+          // "Create a new prompt" is selected
+          window.open("/chat/settings", "_self");
```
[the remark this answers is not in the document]
raunakab: Prompt-shortcuts will be dead for a bit. The links are going to be broken for now. The reason is because the refreshed user-settings page will fix this, but since prompt-shortcut improvements are sufficiently different from user-settings, we ideally want to merge these in separately.

The actual, proper link will be appropriately set in the user-settings PR.

web/src/app/chat/components/input/ChatInputBar.tsx
```diff
@@ -110,385 +120,334 @@ export interface ChatInputBarProps {
...
+                compactImages={shouldCompactImages}
+              />
+            ))}
+          </div>
+        )}
+
+        {/* Input area */}
+        <Popover
+          open={user?.preferences.shortcut_enabled && showPrompts}
+          onOpenChange={setShowPrompts}
```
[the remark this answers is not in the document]
raunakab: Fixed.

raunakab:
@cubic-dev-ai, please re-review this.

backend/onyx/db/input_prompt.py
```diff
@@ -18,6 +18,36 @@
...
+    """
+    query = select(InputPrompt).where(InputPrompt.prompt == prompt)
+
+    if user_id is not None:
+        query = query.where(InputPrompt.user_id == user_id)
+    else:
+        query = query.where(InputPrompt.user_id.is_(None))
+
+    if exclude_id is not None:
+        query = query.where(InputPrompt.id != exclude_id)
```
Subash-Mohan: When user_id is None and exclude_id is provided, the SQL query becomes:

  SELECT * FROM inputprompt
  WHERE prompt = 'some_name'
    AND user_id IS NULL
    AND id != exclude_id
My understanding is that if the user_id is not present, it is a public shortcut, so we are searching only among the public shortcuts. Is that correct?

backend/onyx/db/input_prompt.py
```diff
@@ -64,12 +94,21 @@ def insert_input_prompt(
     user: User | None,
     db_session: Session,
 ) -> InputPrompt:
+    user_id = user.id if user is not None else None
+
+    # Check for duplicate prompt shortcut
+    if not check_prompt_is_unique(prompt, user_id, db_session):
```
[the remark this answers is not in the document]
evan-onyx: yeah this is valid, lets just use the on conflict do update pattern

backend/onyx/db/input_prompt.py
```diff
@@ -64,12 +94,21 @@ def insert_input_prompt(
     user: User | None,
     db_session: Session,
 ) -> InputPrompt:
+    user_id = user.id if user is not None else None
```
evan-onyx: nit: `user.id if user else None`

evan-onyx (commented):
backend seems mostly fine to me, leaving the approval for frontend folks since it seems that's most of the PR

web/src/app/chat/components/input/ChatInputBar.tsx
```diff
@@ -532,189 +491,223 @@ const ChatInputBar = React.memo(
...
-        </>
-      );
-    }
-  )
+        </div>
+      </div>
+    );
+  }
 );
-ChatInputBar.displayName = "ChatInputBar";
```
Subash-Mohan, line 718: shall we add this displayName again?

backend/onyx/db/input_prompt.py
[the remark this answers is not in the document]
raunakab: Nice catch; updating right now.

web/src/app/chat/components/input/ChatInputBar.tsx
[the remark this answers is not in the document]
raunakab: This is on purpose. The reason being is that the `chat-preferences` page will be merged in in a soon coming PR.

backend/onyx/db/input_prompt.py
raunakab: Fixed.

backend/onyx/db/input_prompt.py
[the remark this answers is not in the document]
raunakab: Should be fixed now. Thanks for the help, @evan-onyx.

raunakab:
@cubic-dev-ai, please re-review this.

web/src/app/chat/components/input/ChatInputBar.tsx
```diff
@@ -110,385 +120,334 @@ export interface ChatInputBarProps {
...
+      return message.slice(1);
+    }, [message]);
+
+    const { filtered: filteredPrompts, setQuery: setPromptFilterQuery } =
+      useFilter(activePromptShortcuts, (prompt) => prompt.prompt);
+
+    // Sync the filter query when message changes
+    useEffect(() => {
+      setPromptFilterQuery(promptFilterQuery);
+    }, [promptFilterQuery, setPromptFilterQuery]);
```
Subash-Mohan: My understanding:
When the `message` changes, it updates the value of `promptFilterQuery`, which in turn triggers the `useEffect` to set a new filter inside the `useFilter` hook. That hook then returns the result.
I feel this can be simplified to:

```
const filteredPrompts = useMemo(() => {
  if (!message.startsWith("/")) return activePromptShortcuts;
  const query = message.slice(1).toLowerCase();
  if (!query) return activePromptShortcuts;
  return activePromptShortcuts.filter(p =>
    p.prompt.toLowerCase().includes(query)
  );
}, [message, activePromptShortcuts]);
```
This eliminates:
- The `useEffect` used for syncing state
- The extra `useMemo` for promptFilterQuery
- The indirect state update via `setPromptFilterQuery`

web/src/app/chat/components/input/ChatInputBar.tsx
```diff
@@ -532,189 +491,223 @@ const ChatInputBar = React.memo(
...
+                    <LineItem
+                      key={prompt.id}
+                      selected={tabbingIconIndex === index}
+                      emphasized={tabbingIconIndex === index}
+                      description={prompt.content?.trim()}
+                      onClick={() => updateInputPrompt(prompt)}
+                    >
+                      {prompt.prompt}
+                    </LineItem>
+                  )),
```
Subash-Mohan: Every time the component re-renders (typing, mouse move, any state change), this runs again — even if filteredPrompts hasn't changed.

1. [...filteredPrompts] → Creates a new array copy (to avoid mutating original)
2. sort((a, b) => a.id - b.id) → Creates another new array and sorts it
3. map(...) → Creates yet another new array with React elements

  Visual Example

  Render 1: filteredPrompts = [{id: 3}, {id: 1}, {id: 2}]
    → Copy array     → [3, 1, 2]  (new array)
    → Sort           → [1, 2, 3]  (new array)
    → Map to JSX     → [<LineItem/>, <LineItem/>, <LineItem/>]  (new array)

  Render 2: filteredPrompts = [{id: 3}, {id: 1}, {id: 2}]  // SAME DATA!
    → Copy array     → [3, 1, 2]  (new array)  ← Unnecessary!
    → Sort           → [1, 2, 3]  (new array)  ← Unnecessary!
    → Map to JSX     → [<LineItem/>, <LineItem/>, <LineItem/>]  (new array)

Solution: Memoize the Sorted Array

```
  // Memoize the sorted prompts - only recalculates when filteredPrompts changes
  const sortedPrompts = useMemo(
    () => [...filteredPrompts].sort((a, b) => a.id - b.id),
    [filteredPrompts]
  );

  // In JSX - just map, no sort needed
  <Popover.Menu>
    {sortedPrompts.map((prompt, index) => (
      <LineItem
        key={prompt.id}
        selected={tabbingIconIndex === index}
        emphasized={tabbingIconIndex === index}
        description={prompt.content?.trim()}
        onClick={() => updateInputPrompt(prompt)}
      >
        {prompt.prompt}
      </LineItem>
    ))}
```

web/src/app/chat/components/input/ChatInputBar.tsx
```diff
@@ -532,189 +491,223 @@ const ChatInputBar = React.memo(
...
+                filteredPrompts.length > 0 ? null : undefined,
```
Subash-Mohan: May i know why we need to pass undefined or null here?

web/src/app/chat/components/input/ChatInputBar.tsx
```diff
@@ -110,385 +120,334 @@ export interface ChatInputBarProps {
...
-          );
-          const totalTokens =
-            (currentSessionFileTokenCount || 0) + currentFilesTokenTotal;
-          // Hide processing state when files are within context limits
-          return totalTokens < availableContextTokens;
+    // Extract everything after the last `/` for filtering
+    const promptFilterQuery = useMemo(() => {
+      if (!message.startsWith("/")) return "";
+      // Get everything after the `/`
+      return message.slice(1);
```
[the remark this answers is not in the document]
raunakab: Multi-word shortcuts are acceptable. This comment is invalid.

Subash-Mohan (changes requested):
lgtm! added few comments. Also claude mentioned below as well, I am not sure this is valid.
```
  When filter changes, tabbingIconIndex is not reset:
  // User selects index 5 with 10 prompts
  // Then types to filter down to 3 prompts
  // tabbingIconIndex is still 5 → out of bounds!
```

web/src/app/chat/components/input/ChatInputBar.tsx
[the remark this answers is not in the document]
raunakab: Updated. Since this keeps coming up over and over again, no matter how many times I say that this is on purpose and the appropriate change will be coming in a follow-up PR, I'm just deleting this code portion.

nmgarza5: would prefer to use a Link component

nmgarza5: and why the change to `/chat/settings`? Are we going to extend this page?

web/src/app/chat/components/input/ChatInputBar.tsx
```diff
@@ -110,385 +120,334 @@ export interface ChatInputBarProps {
...
+      );
+    }, [
+      selectedAssistant?.tools,
+      combinedSettings?.settings?.deep_research_enabled,
+    ]);
+
+    function handleKeyDown(e: React.KeyboardEvent<HTMLTextAreaElement>) {
+      if (!showPrompts) return;
+
+      if (e.key === "Enter") {
```
nmgarza5: since you're refactoring...my plan when i was working on slash commands was to separate out all of the keyboard operations to clean up this component. also to provide it as a utility for other input components.

I think it was pretty clean. Take a look: https://github.com/onyx-dot-app/onyx/pull/6743/files#diff-c1c18da8198e5984e245ca19b3553b59b2920eba6a3509af79abed34839349f0

web/src/app/chat/components/input/ChatInputBar.tsx
raunakab: Updated!

web/src/app/chat/components/input/ChatInputBar.tsx
raunakab: Oh yes, good point. Not sure why that was removed in the first place.

web/src/app/chat/components/input/ChatInputBar.tsx
raunakab: The `PopoverMenu` has some logic that says if an element in the array is `null`, then it renders a separator. If it's undefined, it skips over it.

I realize that this API is sorta ugly; I wrote it much earlier before we started using the composite pattern.

web/src/app/chat/components/input/ChatInputBar.tsx
raunakab: There is no `/chat/settings` currently. This is dead code. It will be created in the soon-upcoming `user-settings` PR.

backend/onyx/db/input_prompt.py
raunakab: From Claude:

Yes, your understanding is correct. The uniqueness constraints are scoped:
1. User-owned prompts (user_id is not NULL): Unique constraint on `(prompt, user_id)` — each user can only have one shortcut with a given name.
2. Public prompts (`user_id IS NULL`): Partial unique index on (prompt) `WHERE user_id IS NULL` — there can only be one public shortcut with a given name.

So when `user_id` is `None`, the query checks uniqueness only among public prompts. This means:
- User A can have a shortcut named "summarize"
- User B can also have a shortcut named "summarize"
- A public shortcut can also be named "summarize"
- But there can only be one public shortcut named "summarize"

The scopes are intentionally separate — a user's personal shortcuts don't conflict with public ones or other users' shortcuts.

web/src/app/chat/components/input/ChatInputBar.tsx
raunakab: Simplified logic!

backend/onyx/db/input_prompt.py
```diff
@@ -64,16 +66,36 @@ def insert_input_prompt(
...
+        user_id=user_id,
     )
-    db_session.add(input_prompt)
-    db_session.commit()
 
+    # Use the appropriate constraint based on whether this is a user-owned or public prompt
+    if user_id is not None:
+        stmt = stmt.on_conflict_do_nothing(constraint="uq_inputprompt_prompt_user_id")
+    else:
+        stmt = stmt.on_conflict_do_nothing(constraint="uq_inputprompt_prompt_public")
```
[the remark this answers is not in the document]
raunakab: Already solved!

web/src/app/chat/components/input/ChatInputBar.tsx
raunakab: I like this change. However, I think I'd prefer it if you'd take the front on merging that on in since I'd ideally want to keep this PR minimal for now.

raunakab: I'm going to create a ticket for it! Overall, the implementation looks good to me.

raunakab:
@cubic-dev-ai, please re-review this.

PR #7654, opened 2026-01-22
chore: Redirect user to create account

yuhongsun96:
## Description
If no accounts, user obviously cannot sign in, redirect them to the create account page instead.

## How Has This Been Tested?
Tested locally

## Additional Options

---
## Summary by cubic
Redirects users without existing accounts from the login page to the signup page to streamline initial setup for BASIC auth. Adds a has_users flag to auth metadata to drive this logic.

- **New Features**
  - Backend: /manage/auth/type now includes has_users (via get_user_count) and is async.
  - Frontend: /auth/login auto-redirects to /auth/signup when has_users is false for BASIC.
  - Sign In link uses ?autoRedirectToSignup=false to avoid redirect loops.
  - Types updated to include hasUsers in AuthTypeMetadata.

<sup>Written for commit 4159da27266fb71d29b4e8977f542f6a8d2dace2. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/app/auth/login/page.tsx
+++ b/web/src/app/auth/login/page.tsx
@@ -50,8 +50,7 @@ export default async function Page(props: PageProps) {
     authTypeMetadata &&
     !authTypeMetadata.hasUsers &&
     !autoRedirectToSignupDisabled &&
-    (authTypeMetadata.authType === AuthType.BASIC ||
-      authTypeMetadata.authType === AuthType.CLOUD)
+    authTypeMetadata.authType === AuthType.BASIC
   ) {
     return redirect("/auth/signup");
   }


web/src/app/auth/login/page.tsx
```diff
@@ -19,6 +19,8 @@ export interface PageProps {
 export default async function Page(props: PageProps) {
   const searchParams = await props.searchParams;
   const autoRedirectDisabled = searchParams?.disableAutoRedirect === "true";
+  const autoRedirectToSignupDisabled =
+    searchParams?.autoRedirectToSignup === "false";
```
nmgarza5, line 23: seems unnecessary

web/src/app/auth/login/page.tsx
```diff
@@ -42,6 +44,18 @@ export default async function Page(props: PageProps) {
...
+  // (only for auth types that support self-service signup)
+  if (
+    authTypeMetadata &&
+    !authTypeMetadata.hasUsers &&
+    !autoRedirectToSignupDisabled &&
+    (authTypeMetadata.authType === AuthType.BASIC ||
+      authTypeMetadata.authType === AuthType.CLOUD)
+  ) {
+    return redirect("/auth/signup");
+  }
```
nmgarza5, line 56: Why do we need to check auth type? All we care about is if there are users or not.
```suggestion
  // if there are no users, redirect to signup page for initial setup
  if (!authTypeMetadata?.hasUsers ) {
    return redirect("/auth/signup");
  }
```

when does this come into play? `// (only for auth types that support self-service signup)` dont we just offer basic and cloud now? no auth is no longer and option i though

web/src/components/auth/AuthFlowContainer.tsx
```diff
@@ -35,7 +35,7 @@ export default function AuthFlowContainer({
         <div className="text-sm mt-6 text-center w-full text-text-03 mainUiBody mx-auto">
           Already have an account?{" "}
           <Link
-            href="/auth/login"
+            href="/auth/login?autoRedirectToSignup=false"
```
nmgarza5, line 38: seems unnecessary

nmgarza5 (commented):
If we are doing this, we also probably need to rework the entire no-auth playwright suite.

web/src/components/auth/AuthFlowContainer.tsx
yuhongsun96: If the user clicks the "Sign In" button on the page, they shouldn't be just rejected and redirected to the create account page.

web/src/app/auth/login/page.tsx
yuhongsun96: We don't need to do this for things like SAML/OIDC since user provisioning just happens and they don't need to go to this page to create a user

web/src/app/auth/login/page.tsx
yuhongsun96: See other comment

backend/onyx/server/manage/get_state.py
```diff
@@ -28,12 +29,14 @@ def healthcheck() -> StatusResponse:
 
 
 @router.get("/auth/type", tags=PUBLIC_API_TAGS)
-def get_auth_type() -> AuthTypeResponse:
+async def get_auth_type() -> AuthTypeResponse:
```
Weves, line 32: why async?

web/src/components/auth/AuthFlowContainer.tsx
nmgarza5: yes i understand what the redirect is doing. i dont think the url param is necessary to enforce the behavior. seems like unnecessary complexity

web/src/app/auth/login/page.tsx
nmgarza5: see my response. the url param seems unnecessary

web/src/app/auth/login/page.tsx
nmgarza5: this would still work to redirect the user no?  without the need for the url param
```suggestion
  // if there are no users, redirect to signup page for initial setup
  // (only for auth types that support self-service signup)
  if (
    !authTypeMetadata?.hasUsers &&
    (authTypeMetadata.authType === AuthType.BASIC ||
      authTypeMetadata.authType === AuthType.CLOUD)
  ) {
    return redirect("/auth/signup");
  }
```

backend/onyx/server/manage/get_state.py
yuhongsun96: good point, LLM just threw this in, didn't really think much of it. I mean it's probably fine but let's not do that actually

web/src/app/auth/login/page.tsx
nmgarza5: and for OIDC/SAML,  if users are provisioned already, the `hasUsers` check is enough. no need to check authtype too

backend/onyx/server/manage/get_state.py
yuhongsun96: oh actually it's because get_user_count is async lol. well... this is probably fine

Weves (approved):

PR #7658, opened 2026-01-22
fix(ui): User Groups Connectors Fix

justin-tahara:
## Description

Currently the UI for the User Groups page is messed up and you aren't able to click and select any connectors. 

This prevents the use of user groups since it doesn't do anything right now. Even if you try to edit it, it won't work. 

This PR aims to restore the functionality that was there previously.

## How Has This Been Tested?

Ran locally
Before:
Screenshot 2026-01-21 at 5 09 08 PM

After:
Screenshot 2026-01-21 at 5 08 26 PM
Screenshot 2026-01-21 at 5 08 42 PM

## Additional Options

---
## Summary by cubic
Restored connector selection on the User Groups page by fixing the multi-select dropdown behavior. Users can now add, remove, and edit connectors reliably, with clear messages when all are selected or none are available.

- **Bug Fixes**
  - Corrected allConnectorsSelected logic (requires connectors > 0).
  - Stopped auto-blur/auto-close; input stays usable and dropdown remains open.
  - Improved empty states: all selected, no matches, or no private connectors available.

<sup>Written for commit ad3da281a07eef7c64a6b554886aa0635e303db1. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/components/ConnectorMultiSelect.tsx
+++ b/web/src/components/ConnectorMultiSelect.tsx
@@ -41,7 +41,8 @@ export const ConnectorMultiSelect = ({
     (connector) => !selectedIds.includes(connector.cc_pair_id)
   );
 
-  const allConnectorsSelected = unselectedConnectors.length === 0;
+  const allConnectorsSelected =
+    connectors.length > 0 && unselectedConnectors.length === 0;
 
   const filteredUnselectedConnectors = unselectedConnectors.filter(
     (connector) => {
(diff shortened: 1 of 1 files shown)


web/src/components/ConnectorMultiSelect.tsx
```diff
@@ -111,7 +103,7 @@ export const ConnectorMultiSelect = ({
     ? "All connectors selected"
     : placeholder;
 
-  const isInputDisabled = disabled || allConnectorsSelected;
+  const isInputDisabled = disabled;
```
raunakab, line 106: Hmm, this seems unnecessary now. Not anything you need to solve though lol.

PR #7683, opened 2026-01-23
feat: Maintain correct docs on replay

yuhongsun96:
## Description
Previously, when replaying a session, it would show all the docs that came back from the search, now it shows it same as the first pass. Addresses: [ENG-3135](https://linear.app/onyx-app/issue/ENG-3135/doc-selection-on-replay)

Additionally now sorts the cited sources on replay based on the order they appear in the text.

## How Has This Been Tested?
Verified with main chat loop with some tools
Verified with deep research

## Additional Options

---
## Summary by cubic
Fixes doc selection on chat replay so it shows the same displayed docs as the original turn, not the full search results. Addresses ENG-3135 (doc selection on replay).

- **Bug Fixes**
  - Added displayed_docs to SearchDocsResponse and saved those per tool call; replay now uses this subset.
  - Tracked all fetched search docs and emitted citation numbers in ChatStateContainer; saved only emitted citations, deduped docs, and ordered citations by first appearance.
  - Updated save logic to create DB entries from all_search_docs, link tool calls to displayed docs, and build citations from the emitted mapping.

<sup>Written for commit 84e6067156e627300cc0dd86582edb549c7e4782. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/sections/document-sidebar/DocumentsSidebar.tsx
+++ b/web/src/sections/document-sidebar/DocumentsSidebar.tsx
@@ -100,18 +100,26 @@ const DocumentsSidebar = memo(
       ? currentMessageTree?.get(idOfMessageToDisplay)
       : null;
 
-    // Separate cited documents from other documents
-    const citedDocumentIds = useMemo(() => {
+    // Get citations in order and build a set of cited document IDs
+    const { citedDocumentIds, citationOrder } = useMemo(() => {
       if (!selectedMessage) {
-        return new Set<string>();
+        return {
+          citedDocumentIds: new Set<string>(),
+          citationOrder: new Map<string, number>(),
+        };
       }
 
       const citedDocumentIds = new Set<string>();
+      const citationOrder = new Map<string, number>();
       const citations = getCitations(selectedMessage.packets);
-      citations.forEach((citation) => {
+      citations.forEach((citation, index) => {
         citedDocumentIds.add(citation.document_id);
+        // Only set the order for the first occurrence
+        if (!citationOrder.has(citation.document_id)) {
+          citationOrder.set(citation.document_id, index);
+        }
       });
-      return citedDocumentIds;
+      return { citedDocumentIds, citationOrder };
     }, [idOfMessageToDisplay, selectedMessage?.packets.length]);
 
     // if these are missing for some reason, then nothing we can do. Just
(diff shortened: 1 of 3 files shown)


backend/onyx/tools/fake_tools/research_agent.py
```diff
@@ -499,7 +505,7 @@ def run_research_agent_call(
                             or most_recent_reasoning,
                             tool_call_arguments=tool_call.tool_args,
                             tool_call_response=tool_response.llm_facing_response,
-                            search_docs=search_docs,
+                            search_docs=displayed_docs or search_docs,
```
jessicasingh7, line 508: when would this be the case that displayed_docs are None?

web/src/sections/document-sidebar/DocumentsSidebar.tsx
```diff
@@ -100,18 +100,26 @@ const DocumentsSidebar = memo(
       ? currentMessageTree?.get(idOfMessageToDisplay)
       : null;
 
-    // Separate cited documents from other documents
-    const citedDocumentIds = useMemo(() => {
+    // Get citations in order and build a set of cited document IDs
```
jessicasingh7, line 103: Why are we adding a citation order? Earlier citations were in order of importance already I thought

backend/onyx/tools/fake_tools/research_agent.py
jessicasingh7: Doesn't the research agents need all search docs?

web/src/sections/document-sidebar/DocumentsSidebar.tsx
yuhongsun96: Mmm it definitely wasn't showing in order in the UI. We just load them from the DB and pass them to the frontend via the relationship table and there is no sorting or anything prior to this so I'm pretty sure it wasn't sorted

backend/onyx/tools/fake_tools/research_agent.py
yuhongsun96: LLM filtering step could fail, I think it's ok to default to the larger set. I think it's likely ok to just let it fail also

yuhongsun96: Changing this to the display docs is only for saving the tool call in the DB. It's for replaying and it's for the internal search / web search tool

PR #7698, opened 2026-01-23
feat(filesys): APIs for attaching hiernodes

evan-onyx:
## Description

APIs that the FE can call to support the "attaching arbitrary knowledge to assistants" UI

## How Has This Been Tested?

not yet

## Additional Options

---
## Summary by cubic
Adds backend APIs to list accessible hierarchy nodes and their documents to support attaching knowledge to assistants. Also fixes Google Drive hierarchy extraction to avoid race conditions and correctly detect My Drive and Shared Drive roots.

- **New Features**
  - Endpoints: GET /hierarchy-nodes (list nodes by source) and POST /hierarchy-nodes/documents (list documents for a node).
  - Access control: filters by public items (public connectors mark nodes as public), user email, and external group overlap; external group IDs fetched via versioned EE implementation.
  - Pagination: 50 items per page with cursor derived from last_modified, last_synced, and document_id; stable ordering on those fields.
  - Responses: node summaries (id, title, link, parent_id); document summaries include last_modified and last_synced; next_cursor provided when more results are available.
  - Integration: hierarchy router registered in the app; DB helpers added for accessible nodes and paginated documents.

<sup>Written for commit 91381793a1d04269195bfe6b580ad05b4c74c51b. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/connectors/google_drive/models.py
+++ b/backend/onyx/connectors/google_drive/models.py
@@ -3,6 +3,7 @@ from typing import Any
 
 from pydantic import BaseModel
 from pydantic import ConfigDict
+from pydantic import Field
 from pydantic import field_serializer
 from pydantic import field_validator
 
@@ -150,7 +151,10 @@ class GoogleDriveCheckpoint(ConnectorCheckpoint):
     # Hierarchy node raw IDs that have already been yielded.
     # Used to avoid yielding duplicate hierarchy nodes across checkpoints.
     # Thread-safe because multiple impersonation threads access this concurrently.
-    seen_hierarchy_node_raw_ids: ThreadSafeSet[str] = ThreadSafeSet()
+    # Uses default_factory to ensure each checkpoint instance gets a fresh set.
+    seen_hierarchy_node_raw_ids: ThreadSafeSet[str] = Field(
+        default_factory=ThreadSafeSet
+    )
 
     # Hierarchy node raw IDs where we have successfully walked up to a terminal
     # node (a drive root with no parent). This is separate from seen_hierarchy_node_raw_ids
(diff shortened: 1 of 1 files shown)


backend/onyx/server/features/hierarchy/models.py
```diff
@@ -0,0 +1,55 @@
+from datetime import datetime
+
+from pydantic import BaseModel
+
+from onyx.configs.constants import DocumentSource
+from onyx.server.features.hierarchy.constants import DOCUMENT_PAGE_SIZE
+
+
+class HierarchyNodesRequest(BaseModel):
```
acaprau, line 9: for all of these models dont we want to make them frozen?

backend/onyx/server/features/hierarchy/api.py
```diff
@@ -0,0 +1,97 @@
...
+
+def _get_user_access_info(
+    user: User | None, db_session: Session
+) -> tuple[str | None, list[str]]:
+    if not user:
+        return None, []
+    return user.email, get_user_external_group_ids(db_session, user)
+
+
+@router.post(HIERARCHY_NODES_LIST_PATH)
```
acaprau: nit but i think this and the one below it are probably gets not posts in rest api semantics. i understand if the request body is very complicated it might be more convenient to use an endpoint that supports request bodies which get technically doesnt but scanning this logic the request body is pretty simple it could fit nicely into params no?

acaprau: unless these endpoints secretly have side effects then post makes sense but i dont think they do?

backend/onyx/server/features/hierarchy/models.py
evan-onyx: we don't do this elsewhere so I'm going to leave it off here, but if we want to do a repo wide refactor later it's probably a good idea

PR #7730, opened 2026-01-24
fix: Fix layout issues with `AgentEditorPage`

raunakab:
## Description

This PR fixed a whole bunch of issues around widths and layouts inside of the `AgentEditorPage`.

## Screenshots

image

image

image

## Additional Options

---
## Summary by cubic
Improved the AgentEditorPage layout and tool controls, and shipped reliability fixes across migrations, billing, caching, auth, and permissions. This reduces UI glitches, prevents permission loss during indexing, and adds annual billing support.

- **Bug Fixes**
  - AgentEditorPage: tighter layout, show tools only when enabled, fold button only when tools exist, preserve input focus when toggling fold, inline code wraps correctly, better alignment; optional tools default off for new assistants to avoid “Tool not available” errors.
  - Backend reliability: run migrations atomically and clean up legacy notifications; document upserts now preserve existing permissions via COALESCE to avoid overwrites.
  - Prompt cache: use full LLMConfig and correctly handle Bedrock non‑Anthropic models.
  - MCP auth: save per‑user header substitutions and update UI to mark servers authenticated after credential submission.
  - Tooltips/visuals: require ref‑forwarding children and add Card ref support; fix Coda logo rendering and minor source tile styling.

- **New Features**
  - Annual billing support: subscription checkout accepts monthly or annual; seat updates keep the customer’s existing price.

<sup>Written for commit 74a40463a45ac28addb54b49562464ea8820b83b. Summary will update on new commits.</sup>


web/src/layouts/actions-layouts.tsx
```diff
@@ -121,25 +112,56 @@ const ActionsLayoutContext = createContext<
...
+  const [hasContent, setHasContent] = useState(false);
+
+  // Registration function for ActionsContent to announce its presence
+  const registerContent = useMemo(
+    () => () => {
+      setHasContent(true);
+      return () => setHasContent(false);
+    },
+    []
+  );
```
nmgarza5, line 124: feels kinda hacky

raunakab: I think this is a common "registration" pattern to detect if a child is present. I believe it's standard (according to Claude's resources, at least).

PR #7736, opened 2026-01-24
fix(chat): Internal search enablement matches source enablement

Danelegend:
## Additional Options

---
## Summary by cubic
Makes internal search automatically match source enablement. When any source is enabled, search turns on; when none are enabled, search turns off. Also preserves selected sources when switching chat sessions.

- **Bug Fixes**
  - Syncs the search tool with source state: auto-enables when sources are enabled, auto-disables when none are.
  - Toggling the search tool updates sources: enabling restores previous sources (or enables all), disabling saves current sources then clears them.
  - Adds enableSources to persist preferences; enableAllSources now uses it.
  - Stops clearing selectedSources on session switch to keep user choices.

<sup>Written for commit 54b7b5c87eadbd56d6c95650987339b64d273c5a. Summary will update on new commits.</sup>


Danelegend:
@greptile

PR #7745, opened 2026-01-24
fix(chat): prevent streaming text from appearing in bursts after citations

nmgarza5:
## Description

Fixes a regression where AI response text appeared in large bursts instead of streaming smoothly.

### Root Cause

The `AIMessage` component uses `React.memo` with a custom `arePropsEqual` function that compared `prev.rawPackets.length === next.rawPackets.length`. However, the packets array was being **mutated in place** with `.push()` and the same array reference was passed to the store on each update.

This violated React's immutability expectation: `prev.rawPackets` and `next.rawPackets` pointed to the **same object**, so the length comparison always returned `true` (comparing identical objects), preventing re-renders during streaming.

### Solution

Added a `packetsVersion` counter that increments with each new packet. The memo comparison now checks `prev.packetsVersion === next.packetsVersion` instead of comparing array lengths.

### Why Version Counter Over Array Copying?

An alternative fix would be to create a new array on each update: `packets: [...packets]`. However:

| Approach | Complexity per packet | Memory |
|----------|----------------------|--------|
| `[...packets]` | O(n) - copies entire array | Creates n arrays of sizes 1, 2, 3, ..., n |
| `packetsVersion++` | O(1) - increment integer | Single integer |

For a response with 500 packets, array copying would perform ~125,000 copy operations total, while the version counter approach performs 500 increments. This matters for long AI responses.

## Additional Options

### DIFF
--- a/web/src/app/chat/hooks/useChatController.ts
+++ b/web/src/app/chat/hooks/useChatController.ts
@@ -643,6 +643,7 @@ export function useChatController({
       let toolCall: ToolCallMetadata | null = null;
       let files = projectFilesToFileDescriptors(currentMessageFiles);
       let packets: Packet[] = [];
+      let packetsVersion = 0;
 
       let newUserMessageId: number | null = null;
       let newAssistantMessageId: number | null = null;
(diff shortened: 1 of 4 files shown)


nmgarza5:
@greptile

nmgarza5:
Failing exa test is unrelated. going to merge

PR #7758, opened 2026-01-25
fix: Stop header from collapsing repeatedly

raunakab:
## Description

- Fixes flickering in the "Thinking block" / steps UI during message streaming.
  - The issue was caused by `isComplete` (mapped to `finalAnswerComing`) toggling between true and false as message and tool packets interleaved, causing the UI to flip between the streaming view (expanded timeline) and complete view (collapsed "N steps" header).
  - Changed the condition from `isComplete` to `stopPacketSeen`, which only transitions from false to true once, providing a stable UI state during streaming.

## Video

https://github.com/user-attachments/assets/32d71842-77cb-4c6f-8da6-fd9a85803a2f

## Additional Options

---
## Summary by cubic
Stops the “N steps” header from collapsing repeatedly during streaming by using stopPacketSeen instead of isComplete. This removes flicker in the Thinking/steps UI.

- **Bug Fixes**
  - Replace isComplete (finalAnswerComing) with stopPacketSeen in MultiToolRenderer to prevent toggling between streaming and complete states when packets interleave.
  - Keeps the timeline expanded until a stop packet is received for a stable view.

<sup>Written for commit 372bc1de16a92fcce41a66b156f94ac0e67df757. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/app/chat/message/messageComponents/MultiToolRenderer.tsx
+++ b/web/src/app/chat/message/messageComponents/MultiToolRenderer.tsx
@@ -750,7 +750,11 @@ export default function MultiToolRenderer({
   };
 
   // If still processing, show tools progressively with timing
-  if (!isComplete) {
+  // Use stopPacketSeen instead of isComplete to prevent flickering.
+  // isComplete (finalAnswerComing) toggles back and forth when message and tool
+  // packets interleave, causing the UI to flip between streaming and complete views.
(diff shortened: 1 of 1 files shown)


yuhongsun96:
1. there is no option to collapse it as the steps are running
2. the view to only show the last step that is happening is gone
3. Multiple Reading steps are failing without any indication of what is happening (seems unrelated to the change though)

PR #7762, opened 2026-01-25
Fix: Action block blinking and collapsing

yuhongsun96:
## Description
Reworked the display of the blocks. There was an issue of showing the latest block and the list of blocks and the steps collapsing on its own. This is an ok experience but not final.

## How Has This Been Tested?
Tested with basic tool calls and deep research.

https://github.com/user-attachments/assets/c3c8e44d-9877-4251-943c-df70320be789

Accepted downsides:
- It shows a Done after each step
- There is not a smooth transition between the steps. There is a "Done" type state in between rather than continuing the current step until the next packet shows up.
- There is not a nice view which just shows the current active action.

## Additional Options

---
## Summary by cubic
Fixes blinking and auto-collapsing of action blocks during streaming by simplifying MultiToolRenderer. The steps list now defaults to expanded and stays stable; the “Done” node renders only when the run is complete.

- **Bug Fixes**
  - Removed height/opacity transitions and render expanded content conditionally to prevent flicker.
  - Unified streaming/complete rendering so items don’t reflow between code paths.
  - Show only visible tools while streaming; render all tools when complete.
  - Display “Done” only after completion; keep connector lines consistent.

- **Refactors**
  - Deleted ExpandedToolItem and use ToolItemRow for all cases.
  - Moved the summary header outside the box and made it clickable; default isExpanded = true.
  - Centralized turn grouping and parallel-tab handling; simplified loading state checks.

<sup>Written for commit 51a92c5296dc1da9cfe43afe42bdf44d7263b8c3. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/app/chat/message/messageComponents/MultiToolRenderer.tsx
+++ b/web/src/app/chat/message/messageComponents/MultiToolRenderer.tsx
@@ -461,92 +461,6 @@ function ParallelToolTabs({
   );
 }
 
-// Shared component for expanded tool rendering
-function ExpandedToolItem({
-  icon,
-  content,
-  status,
-  isLastItem,
-  showClickableToggle = false,
-  onToggleClick,
-  defaultIconColor = "text-text-300",
-  expandedText,
-}: {
-  icon: ((props: { size: number }) => JSX.Element) | null;
-  content: JSX.Element | string;
-  status: string | JSX.Element | null;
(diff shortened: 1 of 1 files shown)


raunakab:
Closing in favour of https://github.com/onyx-dot-app/onyx/pull/7767.

PR #7810, opened 2026-01-26
feat(chat history): summarize older messages

jessicasingh7:
## Description

Summarize older messages instead of truncating
[History Compression Linear](https://linear.app/onyx-app/project/history-compression-8c04a68b51ac)

## How Has This Been Tested?

ran and checked summary ChatMessage local db entries

## Additional Options

---
## Summary by cubic
Summarizes older chat messages instead of truncating to keep context within model limits. Maintains a branch-aware summary message while keeping recent messages verbatim. Addresses Linear “History Compression”.

- **New Features**
  - Progressive summary stored as a ChatMessage; branch-aware via parent_message_id of the last message when triggered.
  - Tracks last_summarized_message_id to set the cutoff and filter recent messages.
  - Auto-trigger based on COMPRESSION_TRIGGER_RATIO and model max_input_tokens; reserves tokens for recent messages.
  - Prepends summary to LLM input and only sends messages after the cutoff.

- **Migration**
  - Adds chat_message.last_summarized_message_id (Alembic migration).
  - Run alembic upgrade.
  - Optional: set COMPRESSION_TRIGGER_RATIO (default 0.75).

<sup>Written for commit 88c915702d94e6d9a976e2ff2609787d4b1ba607. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/unit/onyx/chat/test_compression.py
+++ b/backend/tests/unit/onyx/chat/test_compression.py
@@ -1,5 +1,8 @@
 """Unit tests for chat history compression module."""
 
+from datetime import datetime
+from datetime import timedelta
+from datetime import timezone
 from unittest.mock import MagicMock
 
 from onyx.chat.compression import find_summary_for_branch
@@ -7,6 +10,9 @@ from onyx.chat.compression import get_compression_params
 from onyx.chat.compression import get_messages_to_summarize
 from onyx.chat.compression import SummaryContent
 from onyx.configs.constants import MessageType
+
+# Base time for generating sequential timestamps
+BASE_TIME = datetime(2024, 1, 1, 12, 0, 0, tzinfo=timezone.utc)
 
 
 def create_mock_message(
@@ -27,6 +33,8 @@ def create_mock_message(
     mock.chat_session_id = chat_session_id
     mock.parent_message_id = parent_message_id
     mock.last_summarized_message_id = last_summarized_message_id
+    # Generate time_sent based on id for chronological ordering
+    mock.time_sent = BASE_TIME + timedelta(minutes=id)
     return mock
 
 


jessicasingh7:
@greptile

backend/onyx/chat/compression.py
```diff
@@ -0,0 +1,365 @@
...
+from onyx.prompts.compression_prompts import PROGRESSIVE_USER_REMINDER
+from onyx.prompts.compression_prompts import SUMMARIZATION_CUTOFF_MARKER
+from onyx.prompts.compression_prompts import SUMMARIZATION_PROMPT
+from onyx.prompts.compression_prompts import USER_FINAL_REMINDER
+from onyx.utils.logger import setup_logger
+
+logger = setup_logger()
+
+# Tokens reserved for system prompt, tool definitions, current user message, and safety margin
+RESERVED_CONTEXT_TOKENS = 3000
```
Danelegend: Is it possible to get a more accurate estimate based tool availability + system prompt size?

backend/onyx/chat/compression.py
```diff
@@ -0,0 +1,365 @@
...
+    Calculate compression parameters based on model's context window.
+
+    Args:
+        llm: The LLM instance (used to get max_input_tokens)
+        current_history_tokens: Current total tokens in chat history
+
+    Returns:
+        CompressionParams indicating whether to compress and token budgets
+    """
+    max_context = llm.config.max_input_tokens
```
Danelegend: If we just need max_input_tokens, might not make as much sense to pass in the entire llm

backend/onyx/chat/compression.py
```diff
@@ -0,0 +1,365 @@
...
+            return summary
+
+    return None
+
+
+def get_messages_to_summarize(
+    chat_history: list[ChatMessage],
+    existing_summary: ChatMessage | None,
+    tokens_for_recent: int,
+) -> tuple[list[ChatMessage], list[ChatMessage]]:
```
Danelegend: nit: earring on bit yucky. Might consider a model class but up to you

backend/onyx/chat/compression.py
```diff
@@ -0,0 +1,365 @@
...
+        existing_summary: Existing summary for this branch (if any)
+        tokens_for_recent: Token budget for recent messages to keep
+
+    Returns:
+        Tuple of (messages_to_summarize, messages_to_keep)
+    """
+    # Filter to messages after the existing summary's cutoff
+    if existing_summary and existing_summary.last_summarized_message_id:
+        cutoff_id = existing_summary.last_summarized_message_id
+        messages = [m for m in chat_history if m.id > cutoff_id]
```
Danelegend: All good for now but worth noting that this only holds cause id is a int in the db. If it ever gets refactored to UUID or something, finding this is gonna be grim.

backend/onyx/chat/compression.py
```diff
@@ -0,0 +1,365 @@
...
+            ]
+            formatted.append(f"[assistant used tools: {', '.join(tool_names)}]")
+            continue
+
+        # Skip standalone tool call/response messages - captured above
+        if msg.message_type in (MessageType.TOOL_CALL, MessageType.TOOL_CALL_RESPONSE):
+            continue
+
+        role = msg.message_type.value
+        formatted.append(f"[{role}]: {msg.message}")
```
Danelegend, line 222: Worth mentioning who said it as well?

backend/onyx/chat/compression.py
```diff
@@ -0,0 +1,365 @@
...
+    recent_messages: list[ChatMessage],
+    llm: LLM,
+    tool_id_to_name: dict[int, str],
+    existing_summary: str | None = None,
+) -> str:
+    """
+    Generate a summary using cutoff marker approach.
+
+    The cutoff marker tells the LLM to summarize only older messages,
+    while using recent messages as context to inform what's important.
```
Danelegend, line 237: Interested to hear why we lean away from just older messages. It makes sense that newer messages can help us extract more useful summarisation, but there is possibly also a train of thought where we'd want the summarisation to be a summary of that point in time (not leaning either way).

Not saying either way is wrong btw, but just interested to hear about the thought process and whether there was any experimentation?

backend/onyx/chat/process_message.py
```diff
@@ -604,6 +626,26 @@ def llm_loop_completion_callback(
                 assistant_message=assistant_response,
             )
 
+            # Check if compression is needed after saving the message
+            # Rebuild chat_history to include the newly saved messages
```
Danelegend: This should be in the llm_loop_completion_handle.

This idea behind the callback is to be a wrapper around the handle

Danelegend (changes requested):
Looking pretty good. I wanna give it one more pass over tmr morning when I'm more awake haha. 

Given that this is a greenfield project also, it should be somewhat straight forward to add some tests. That would be great to see

backend/onyx/chat/compression.py
jessicasingh7: noted as TODO

Danelegend (approved):
Small change with the message id comparison (consider using time_sent instead). 

One small follow-up would be writing an external dependency test to check that the compression gets called once a certain capacity is hit.

But otherwise, lgtm once the top one is fixed

PR #7820, opened 2026-01-26
chore(api): Add tests for Tenantless Public Endpoints

justin-tahara:
## Description

Adding tests for endpoints that require some special care in order to prevent outages with our Cloud Multi-tenant setup

## How Has This Been Tested?

Tests were run locally.

## Additional Options

---
## Summary by cubic
Added tests to verify tenantless public endpoints (/health, /auth/type, /version) return successfully without a tenant context. Also routed these endpoints to the default schema in middleware and updated anonymous_user_enabled to treat missing tenant IDs and Redis errors as disabled, preventing pre-login errors.

<sup>Written for commit 41d316e7378ced15a3ad88c36cfc5095a9f7dd1f. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/auth/users.py
+++ b/backend/onyx/auth/users.py
@@ -55,6 +55,8 @@ from httpx_oauth.integrations.fastapi import OAuth2AuthorizeCallback
 from httpx_oauth.oauth2 import BaseOAuth2
 from httpx_oauth.oauth2 import OAuth2Token
 from pydantic import BaseModel
+from redis.exceptions import ConnectionError as RedisConnectionError
+from redis.exceptions import TimeoutError as RedisTimeoutError
 from sqlalchemy import nulls_last
 from sqlalchemy import select
 from sqlalchemy.exc import IntegrityError
@@ -200,13 +202,18 @@ def user_needs_to_be_verified() -> bool:
 def anonymous_user_enabled(*, tenant_id: str | None = None) -> bool:
     try:
         redis_client = get_redis_client(tenant_id=tenant_id)
+        value = redis_client.get(OnyxRedisLocks.ANONYMOUS_USER_ENABLED)
     except RuntimeError as exc:
         if tenant_id is None and MULTI_TENANT and "Tenant ID is not set" in str(exc):
             # Pre-login/public endpoints may be hit before a tenant context exists.
             # Treat as disabled rather than erroring.
             return False
         raise
-    value = redis_client.get(OnyxRedisLocks.ANONYMOUS_USER_ENABLED)
+    except (RedisConnectionError, RedisTimeoutError) as exc:
+        logger.warning(
+            "Redis unavailable; treating anonymous user as disabled", exc_info=exc
+        )
+        return False
 
     if value is None:
         return False
--- a/backend/onyx/server/auth_check.py
+++ b/backend/onyx/server/auth_check.py
@@ -95,6 +95,26 @@ def is_route_in_spec_list(
     return False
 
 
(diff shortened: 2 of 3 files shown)


backend/tests/api/test_tenantless_public_endpoints.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def test_tenantless_public_endpoints_do_not_require_tenant() -> None:
+    env = os.environ.copy()
+    env["MULTI_TENANT"] = "true"
+    env["AUTH_TYPE"] = "cloud"
+    env["ENABLE_PAID_ENTERPRISE_EDITION_FEATURES"] = "True"
+    backend_root = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", ".."))
+
+    script = textwrap.dedent(
```
yuhongsun96, line 14: Do we absolutely have to use this script/subprocess approach? This feels like an antipattern to me

backend/tests/api/test_tenantless_public_endpoints.py
```diff
@@ -0,0 +1,42 @@
+import os
+import subprocess
+import sys
+import textwrap
+
+
+def test_tenantless_public_endpoints_do_not_require_tenant() -> None:
```
yuhongsun96, line 7: I would write another negative test to assert that some other routes which do require tenant info do in fact fail if tenant info does not exist. That provides more certainty about the coverage of this test if that makes sense

backend/onyx/server/auth_check.py
```diff
@@ -87,6 +95,26 @@ def is_route_in_spec_list(
     return False
 
 
+def is_request_in_spec_list(
+    path: str, method: str, public_endpoint_specs: list[tuple[str, set[str]]]
```
yuhongsun96, line 99: This might be an existing pattern but we should fix instances of complex nested types: list[tuple[str, set[str]]]

backend/onyx/auth/users.py
```diff
@@ -198,8 +200,20 @@ def user_needs_to_be_verified() -> bool:
 
 
 def anonymous_user_enabled(*, tenant_id: str | None = None) -> bool:
-    redis_client = get_redis_client(tenant_id=tenant_id)
-    value = redis_client.get(OnyxRedisLocks.ANONYMOUS_USER_ENABLED)
+    try:
+        redis_client = get_redis_client(tenant_id=tenant_id)
+        value = redis_client.get(OnyxRedisLocks.ANONYMOUS_USER_ENABLED)
+    except RuntimeError as exc:
+        if tenant_id is None and MULTI_TENANT and "Tenant ID is not set" in str(exc):
```
yuhongsun96, line 207: Not a big fan of these magic strings that link to other parts of the code base. Also should this not just raise? Why return False?

backend/tests/api/test_tenantless_public_endpoints.py
```diff
@@ -0,0 +1,42 @@
...
+        from onyx.server.auth_check import TENANTLESS_PUBLIC_ENDPOINT_SPECS
+
+        app = fetch_versioned_implementation(
+            module="onyx.main", attribute="get_application"
+        )()
+        client = TestClient(app)
+
+        for path, methods in TENANTLESS_PUBLIC_ENDPOINT_SPECS:
+            if "GET" in methods:
+                response = client.get(path)
```
yuhongsun96, line 27: This only checks GETs, maybe it's ok for now but makes it easy to bipass this check accidentally

PR #7840, opened 2026-01-27
feat(ee): update api to claim license via cloud proxy

nmgarza5:
## Description

Update license API for self-hosted deployments to use cloud data plane proxy.

**Changes:**
- Rename `POST /license/fetch` → `POST /license/claim` with required `session_id` parameter
- Add `CLOUD_DATA_PLANE_URL` config (default: `https://cloud.onyx.app/api`)
- Block mutation endpoints for `MULTI_TENANT` (cloud) deployments
- Remove tenant_id validation - license verified cryptographically only

## How Has This Been Tested?

full e2e test locally

## Additional Options

### DIFF
--- a/backend/ee/onyx/server/license/api.py
+++ b/backend/ee/onyx/server/license/api.py
@@ -1,4 +1,14 @@
-"""License API endpoints."""
+"""License API endpoints for self-hosted deployments.
+
+These endpoints allow self-hosted Onyx instances to:
+1. Claim a license after Stripe checkout (via cloud data plane proxy)
+2. Upload a license file manually (for air-gapped deployments)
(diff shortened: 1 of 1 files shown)


backend/ee/onyx/server/license/api.py
```diff
@@ -79,81 +88,80 @@ async def get_seat_usage(
     )
 
 
-@router.post("/fetch")
-async def fetch_license(
+@router.post("/claim")
```
justin-tahara, line 91: Will this break for some time?

nmgarza5: nah its good. the license system isnt turned on yet

PR #7841, opened 2026-01-27
fix(craft): only insert onyx user context when demo data not enabled

rohoswagger:
## Description

insert user context into AGENTS.template.md only when we are not in demo data mode, this way model doesn't get confused with demo data role/name and real person role/name

## How Has This Been Tested?

locally

## Additional Options

---
## Summary by cubic
Only include Onyx user context in AGENTS.md when demo data is disabled, protecting personal info in demo sessions.

- **Bug Fixes**
  - Added a use_demo_data flag to instruction generation across Kubernetes, Local, and Directory managers.
  - AGENTS.md generation skips user context when use_demo_data is True.

<sup>Written for commit a72e2a583f6fe590bef8b094b4dd77eec3766810. Summary will update on new commits.</sup>


backend/onyx/server/features/build/sandbox/local/local_sandbox_manager.py
```diff
@@ -826,3 +827,42 @@ def sync_files(
         """
         logger.debug(f"sync_files called for local sandbox {sandbox_id} - no-op")
         return True
+
+    def cancel_message(
```
Weves: this is not used, right?

rohoswagger: oop remnants on stuff i didn't stash, removed

PR #7852, opened 2026-01-27
feat(filesys): filter on assistant info

evan-onyx:
## Description

add hierarchynode and document info from the assistant to the opensearch query

## How Has This Been Tested?

not yet

## Additional Options

---
## Summary by cubic
Scopes search results to knowledge attached to an assistant, and updates auth to use an anonymous user for unauthenticated access. ACLs are still enforced; anonymous sessions only see public content.

- **New Features**
  - Added AssistantKnowledgeFilters (attached_document_ids, hierarchy_node_ids) to IndexFilters; pipeline reads persona attachments and OpenSearch applies a document-ID OR ancestor-folder filter in both ID-based and hybrid queries.
  - Introduced an anonymous user; all endpoints now operate with a non-null User, with anonymous users limited to public/global objects and global-only rate limits.
  - Added migrations to create a permanent anonymous user and migrate legacy no-auth data; AUTH_TYPE='disabled' now defaults to 'basic'.

<sup>Written for commit 62eeb9721a36dd8a424efc54202183474dd1cd03. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/document_index/opensearch/search.py
+++ b/backend/onyx/document_index/opensearch/search.py
@@ -15,6 +15,7 @@ from onyx.document_index.opensearch.constants import SEARCH_CONTENT_VECTOR_WEIGH
 from onyx.document_index.opensearch.constants import SEARCH_TITLE_KEYWORD_WEIGHT
 from onyx.document_index.opensearch.constants import SEARCH_TITLE_VECTOR_WEIGHT
 from onyx.document_index.opensearch.schema import ACCESS_CONTROL_LIST_FIELD_NAME
+from onyx.document_index.opensearch.schema import ANCESTOR_HIERARCHY_NODE_IDS_FIELD_NAME
 from onyx.document_index.opensearch.schema import CHUNK_INDEX_FIELD_NAME
 from onyx.document_index.opensearch.schema import CONTENT_FIELD_NAME
 from onyx.document_index.opensearch.schema import CONTENT_VECTOR_FIELD_NAME
@@ -174,6 +175,8 @@ class DocumentQuery:
             max_chunk_index=max_chunk_index,
             max_chunk_size=max_chunk_size,
             document_id=document_id,
+            attached_document_ids=index_filters.attached_document_ids,
+            hierarchy_node_ids=index_filters.hierarchy_node_ids,
         )
         final_get_ids_query: dict[str, Any] = {
             "query": {"bool": {"filter": filter_clauses}},
@@ -288,6 +291,8 @@ class DocumentQuery:
             time_cutoff=index_filters.time_cutoff,
             min_chunk_index=None,
             max_chunk_index=None,
+            attached_document_ids=index_filters.attached_document_ids,
+            hierarchy_node_ids=index_filters.hierarchy_node_ids,
         )
         match_highlights_configuration = (
             DocumentQuery._get_match_highlights_configuration()
@@ -407,6 +412,9 @@ class DocumentQuery:
         max_chunk_index: int | None,
         max_chunk_size: int | None = None,
         document_id: str | None = None,
+        # Assistant knowledge filters
+        attached_document_ids: list[str] | None = None,
+        hierarchy_node_ids: list[int] | None = None,
     ) -> list[dict[str, Any]]:
         """Returns filters to be passed into the "filter" key of a search query.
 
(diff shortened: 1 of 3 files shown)


backend/onyx/document_index/opensearch/search.py
```diff
@@ -562,6 +576,51 @@ def _get_chunk_index_filter(
...
+            for doc_id in doc_ids:
+                doc_id_filter["bool"]["should"].append(
+                    {"term": {DOCUMENT_ID_FIELD_NAME: {"value": doc_id}}}
+                )
+            return doc_id_filter
+
+        def _get_hierarchy_node_filter(
+            node_ids: list[int],
+        ) -> dict[str, Any]:
+            """Filter for chunks whose ancestors include any of the given hierarchy nodes.
```
acaprau, line 594: super nit you dont have to take action i can change this in one of my prs once this merged but fyi this docstring styling doesnt conform to the rest of the file, we're using 80 char column cutoff, this technically should be on a new line after the """ and should wrap into a second line after 80 chars.

backend/onyx/document_index/opensearch/search.py
```diff
@@ -562,6 +576,51 @@ def _get_chunk_index_filter(
...
+
+            Uses a terms query to check if ancestor_hierarchy_node_ids contains
+            any of the specified node IDs.
+            """
+            return {"terms": {ANCESTOR_HIERARCHY_NODE_IDS_FIELD_NAME: node_ids}}
```
acaprau, line 599: not a huge deal but for my and anyone else's future reference, let's just link the documentation that confirms that you can pass in a list of things into a term query and opensearch will return a match on any of them. i dont think i added this in any of my docstrings in this file.
```suggestion
            # Will return a match if the field contains any of the ids in node_ids. See https://docs.opensearch.org/latest/query-dsl/term/terms/
            return {"terms": {ANCESTOR_HIERARCHY_NODE_IDS_FIELD_NAME: node_ids}}
```

backend/onyx/document_index/opensearch/search.py
```diff
@@ -407,6 +412,9 @@ def _get_search_filters(
         max_chunk_index: int | None,
         max_chunk_size: int | None = None,
         document_id: str | None = None,
+        # Assistant knowledge filters
+        attached_document_ids: list[str] | None = None,
```
acaprau, line 416: this is not a huge deal but if possible i'd prefer to just restrict this typing to list, since there is no semantic difference between an empty list and None in this case. contrast this to access_control_list, where an empty list is semantically different from None. in general we should avoid optionals in new code we write if we can

backend/onyx/document_index/opensearch/search.py
```diff
@@ -562,6 +576,51 @@ def _get_chunk_index_filter(
...
 
+        def _get_attached_document_id_filter(
+            doc_ids: list[str],
+        ) -> dict[str, Any]:
+            """Filter for documents explicitly attached to an assistant."""
+            # Logical OR operator on its elements.
+            doc_id_filter: dict[str, Any] = {"bool": {"should": []}}
+            for doc_id in doc_ids:
+                doc_id_filter["bool"]["should"].append(
+                    {"term": {DOCUMENT_ID_FIELD_NAME: {"value": doc_id}}}
```
acaprau, line 587: i guess this doesnt have to be wrapped in a bool should block and we can just pass in a list of ids into a single term block like you do below for `_get_hierarchy_node_filter`. this is an oversight on my part, i dont think i do this in some of the above functions. in terms of performance i'm guessing it might be slightly more efficient but idk theres no way it matters too much one way or another

backend/onyx/document_index/opensearch/search.py
acaprau, line 460: note we have a warning here that if you supply doc id and user file ids at the same time theyre applying filtering on the same field. this is also the case for attached_document_ids. two questions for you:

1. is it ever possible to be supplied attached_document_ids and one of document_id or user_file_ids? if so, this filtering might break and return 0 docs.
2. can you pls add a warning similar to the one above?

evan-onyx: 1 we definitely might have attached_document_ids and user_file_ids, I would've thought those conditions would be OR'ed? 2 yah

backend/onyx/document_index/opensearch/search.py
evan-onyx: we've had some back and forth on explicit None's in the past, afaik the current leaning is preferring explicit Nones when something isn't being used. But definitely open to changing this in cases where the effect is the same

PR #7853, opened 2026-01-27
feat(ee): improve license enforcement middleware

nmgarza5:
## Description

License enforcement middleware for **self-hosted deployments only**. This provides license-based access control for self-hosted users.

**Note:** Multi-tenant (cloud) deployments use the control plane gating via `/tenants/product-gating` and `is_tenant_gated()` checks instead. This middleware is not registered for MT deployments.

**Enforcement states (self-hosted):**
- **No license** (never subscribed): Allow community features, block EE-only paths
- **GATED_ACCESS** (fully expired): Block all except billing/auth/license paths
- **Valid license** (ACTIVE, GRACE_PERIOD, PAYMENT_REMINDER): Full EE access with seat limits enforced
  - GRACE_PERIOD/PAYMENT_REMINDER are for notifications only, not blocking

**Changes:**
- Add EE-only path prefixes that require valid license (analytics, user groups, query history, etc.)
- Add user management paths to allowed list (needed to remove users when over seat limit)
- Add `check_ee_features_enabled()` helper for conditional EE feature checks
- Add runtime seat limit check as safety net (returns 402 when exceeded)
- Refresh license from DB if not in cache
- Fail open on Redis errors to avoid blocking users due to connectivity issues
- Only register middleware for non-MT deployments in main.py

## How Has This Been Tested?

Manual testing locally

## Additional Options
closes https://linear.app/onyx-app/issue/ENG-3424/enforce-onyx-lockout-upon-license-expirationbilling-failure

### DIFF
--- a/backend/ee/onyx/server/middleware/license_enforcement.py
+++ b/backend/ee/onyx/server/middleware/license_enforcement.py
@@ -102,15 +102,14 @@ def add_license_enforcement_middleware(
             # If no cached metadata, check database (cache may have been cleared)
             if not metadata:
                 logger.debug(
-                    f"[license_enforcement] No cached license for tenant {tenant_id}, "
-                    "checking database..."
+                    "[license_enforcement] No cached license, checking database..."
                 )
                 try:
                     with get_session_with_current_tenant() as db_session:
                         metadata = refresh_license_cache(db_session, tenant_id)
                         if metadata:
                             logger.info(
-                                f"[license_enforcement] Loaded license from DB for tenant {tenant_id}"
+                                "[license_enforcement] Loaded license from database"
                             )
                 except SQLAlchemyError as db_error:
                     logger.warning(
@@ -130,7 +129,7 @@ def add_license_enforcement_middleware(
                     # when users are added/removed
                     if metadata.used_seats > metadata.seats:
                         logger.info(
-                            f"Blocking request for tenant {tenant_id}: "
+                            f"[license_enforcement] Blocking request: "
                             f"seat limit exceeded ({metadata.used_seats}/{metadata.seats})"
                         )
                         return JSONResponse(
@@ -149,7 +148,7 @@ def add_license_enforcement_middleware(
                 # Allow community features, but block EE-only features
                 if _is_ee_only_path(path):
                     logger.info(
-                        f"[license_enforcement] Blocking EE-only path for unlicensed tenant {tenant_id}: {path}"
+                        f"[license_enforcement] Blocking EE-only path (no license): {path}"
                     )
                     return JSONResponse(
                         status_code=402,
@@ -162,8 +161,7 @@ def add_license_enforcement_middleware(
                         },
                     )
                 logger.debug(
-                    f"[license_enforcement] No license for tenant {tenant_id}, "
-                    "allowing community features"
+                    "[license_enforcement] No license, allowing community features"
                 )
                 is_gated = False
         except RedisError as e:
@@ -172,7 +170,9 @@ def add_license_enforcement_middleware(
             is_gated = False
 
         if is_gated:
-            logger.info(f"Blocking request for gated tenant: {tenant_id}, path={path}")
+            logger.info(
+                f"[license_enforcement] Blocking request (license expired): {path}"
+            )
 
             return JSONResponse(
                 status_code=402,


nmgarza5:
@greptile

nmgarza5:
@greptile

backend/ee/onyx/server/settings/api.py
```diff
@@ -12,21 +12,51 @@
...
+    tenant_id = get_current_tenant_id()
+    try:
+        metadata = get_cached_license_metadata(tenant_id)
+        if metadata and metadata.status != _BLOCKING_STATUS:
+            # Has a valid license (GRACE_PERIOD/PAYMENT_REMINDER still allow EE features)
+            return True
+    except RedisError as e:
+        logger.warning(f"Failed to check license for EE features: {e}")
+        # Fail open - don't break the UI if Redis is down
+        return True
```
nmgarza5: hmmm considering returning False here. If Redis is down, we can't effectively gate access based on license presence. Feels like other things in the app would break too so I think its fine if UI breaks

backend/ee/onyx/main.py
```diff
@@ -85,10 +85,11 @@ def get_application() -> FastAPI:
...
         add_api_server_tenant_id_middleware(application, logger)
-
-    # Add license enforcement middleware (runs after tenant tracking)
-    # This blocks access when license is expired/gated
-    add_license_enforcement_middleware(application, logger)
+    else:
+        # License enforcement middleware for self-hosted deployments only
+        # Checks LICENSE_ENFORCEMENT_ENABLED at runtime (can be toggled without restart)
+        # MT deployments use control plane gating via is_tenant_gated() instead
+        add_license_enforcement_middleware(application, logger)
```
justin-tahara, line 92: Was this an issue before?

backend/ee/onyx/server/middleware/license_enforcement.py
```diff
@@ -33,10 +72,45 @@
...
+    "/manage/admin/delete-user",
+    "/users",
+    # Notifications - needed for UI to load properly
+    "/notifications",
+}
+
+# EE-only paths that require a valid license.
+# Users without a license (community edition) cannot access these.
+# These are blocked even when user has never subscribed (no license).
+EE_ONLY_PATH_PREFIXES = {
```
justin-tahara: Can we just double check that this doesn't exist anywhere else? I just get worried when we maintain these types of lists

backend/ee/onyx/server/middleware/license_enforcement.py
```diff
@@ -68,37 +147,82 @@ async def enforce_license(
...
+
             return JSONResponse(
                 status_code=402,
-                content={
-                    "detail": {
-                        "error": "license_expired",
-                        "message": "Your subscription has expired. Please update your billing.",
-                    }
-                },
+                content={"detail": message},
```
justin-tahara: We are not keeping the error?

backend/tests/unit/ee/onyx/server/middleware/test_license_enforcement.py
```diff
@@ -49,6 +51,33 @@ def test_path_allowlist(self, path: str, expected: bool) -> None:
         assert _is_path_allowed(path) is expected
 
 
+class TestEEOnlyPaths:
+    """Tests for EE-only path detection."""
+
+    @pytest.mark.parametrize(
```
justin-tahara, line 72: Just another place that I feel like will be hard to maintain over time

backend/ee/onyx/main.py
nmgarza5: no because I was just passing through requests if the system wasn't available.

```
        if not LICENSE_ENFORCEMENT_ENABLED:
            return await call_next(request)
```

i've since decided not to use licenses for mulit-tenant so there is no reason to register the middleware

backend/tests/unit/ee/onyx/server/middleware/test_license_enforcement.py
nmgarza5: moved the constants to a config file

backend/ee/onyx/server/settings/api.py
```diff
@@ -12,21 +12,51 @@
...
+    tenant_id = get_current_tenant_id()
+    try:
+        metadata = get_cached_license_metadata(tenant_id)
+        if metadata and metadata.status != _BLOCKING_STATUS:
+            # Has a valid license (GRACE_PERIOD/PAYMENT_REMINDER still allow EE features)
+            return True
+    except RedisError as e:
+        logger.warning(f"Failed to check license for EE features: {e}")
+        # Fail closed - if Redis is down, other things will break anyway
+        return False
```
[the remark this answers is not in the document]
nmgarza5: this is fine

backend/ee/onyx/configs/license_enforcement_config.py
justin-tahara, line 1: Beautiful

justin-tahara (approved):
Nice

PR #7857, opened 2026-01-27
feat(ee): Add unified billing API 

nmgarza5:
## Description

Add billing API endpoints that work for both cloud and self-hosted deployments:
- `POST /admin/billing/checkout-session` - Create Stripe checkout session
- `POST /admin/billing/portal-session` - Create portal session to manage subscription
- `GET /admin/billing/information` - Get subscription details
- `POST /admin/billing/seats` - Update seat count
- `GET /admin/billing/stripe-publishable-key` - Get Stripe publishable key

**Routing based on deployment type:**
- **Self-hosted**: Routes through Cloud DP proxy (`/proxy/*` endpoints)
- **Cloud (MULTI_TENANT)**: Routes directly to control plane

**Flow diagram:**
```
Self-hosted:
  Frontend → Backend /admin/billing/* → Cloud DP /proxy/* → Control Plane

Cloud:
  Frontend → Backend /admin/billing/* → Control Plane (direct)
```

**Code improvements:**
- Consolidated HTTP request logic in `service.py` with `_make_billing_request()` helper
- Removed duplicate Pydantic models from `proxy.py` (imports from `models.py` instead)
- Removed inline DEPRECATED/TODO comments from `billing_api.py` (kept in module docstring)

**Tests added (30 total):**
- `test_billing_api.py` - 10 tests for API endpoints
- `test_billing_service.py` - 10 tests for service layer
- `test_proxy.py` - 10 tests for proxy endpoints
- `conftest.py` - Shared test utilities (`make_license_payload`, `make_mock_response`, `make_mock_http_client`)

**NOTE:** License claiming is handled by `/license/claim` endpoint (PR #7840), not this billing API.

## Additional Options

### DIFF
--- a/backend/ee/onyx/server/billing/models.py
+++ b/backend/ee/onyx/server/billing/models.py
@@ -16,7 +16,7 @@ class CreateCheckoutSessionRequest(BaseModel):
 class CreateCheckoutSessionResponse(BaseModel):
     """Response containing the Stripe checkout session URL."""
 
-    url: str
+    stripe_checkout_url: str
 
 
 class CreateCustomerPortalSessionRequest(BaseModel):
@@ -28,7 +28,7 @@ class CreateCustomerPortalSessionRequest(BaseModel):
 class CreateCustomerPortalSessionResponse(BaseModel):
     """Response containing the Stripe customer portal URL."""
 
-    url: str
+    stripe_customer_portal_url: str
 
 
 class BillingInformationResponse(BaseModel):
--- a/backend/ee/onyx/server/billing/service.py
+++ b/backend/ee/onyx/server/billing/service.py
@@ -166,7 +166,7 @@ async def create_checkout_session(
         body=body,
         error_message="Failed to create checkout session",
     )
-    return CreateCheckoutSessionResponse(url=data["url"])
+    return CreateCheckoutSessionResponse(stripe_checkout_url=data["url"])
 
 
 async def create_customer_portal_session(
@@ -197,7 +197,7 @@ async def create_customer_portal_session(
         body=body,
         error_message="Failed to create customer portal session",
     )
-    return CreateCustomerPortalSessionResponse(url=data["url"])
+    return CreateCustomerPortalSessionResponse(stripe_customer_portal_url=data["url"])
 
 
 async def get_billing_information(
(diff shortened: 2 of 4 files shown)


nmgarza5:
@greptile

backend/ee/onyx/server/billing/api.py
```diff
@@ -0,0 +1,264 @@
...
+        return None
+    license_record = get_license(db_session)
+    return license_record.license_data if license_record else None
+
+
+def _get_tenant_id() -> str | None:
+    """Get tenant ID for cloud deployments."""
+    if MULTI_TENANT:
+        return get_current_tenant_id()
+    return None
```
justin-tahara, line 84: I thought we didn't care about multi-tenant or is this just a pre-caution?

backend/ee/onyx/server/billing/api.py
```diff
@@ -0,0 +1,264 @@
...
+    # Build redirect URL for after checkout completion
+    redirect_url = f"{WEB_DOMAIN}/admin/billing?checkout=success"
+
+    try:
+        return await create_checkout_service(
+            billing_period=billing_period,
+            email=email,
+            license_data=license_data,
+            redirect_url=redirect_url,
+            tenant_id=tenant_id,
```
justin-tahara, line 116: We don't seem to do any checks around these values that are passed in prior to them being passed through

backend/ee/onyx/server/billing/api.py
justin-tahara, line 1: A bit of repetition but I don't think we can do much about it

backend/ee/onyx/server/billing/models.py
```diff
@@ -0,0 +1,75 @@
...
+    """Request to create a Stripe checkout session."""
+
+    billing_period: Literal["monthly", "annual"] = "monthly"
+    email: str | None = None
+
+
+class CreateCheckoutSessionResponse(BaseModel):
+    """Response containing the Stripe checkout session URL."""
+
+    url: str
```
justin-tahara: Should we make this a bit more descriptive or you think the class is the detailed part so it's okay

backend/ee/onyx/server/billing/api.py
nmgarza5: the idea here is to eventually have one uniform backend billing api that serves both multi-tenant and self hosted. 

So depending on the presence of `MULTI_TENANT` this api will either route requets to the date plane proxy or directly to the control plane

backend/ee/onyx/server/billing/api.py
nmgarza5: Pydantic validates the request schema (types, required fields). Business logic validation (e.g., valid billing period, email domain restrictions, seat limits) is handled by the control plane - we're just passing through to avoid duplicating that logic.

backend/ee/onyx/server/billing/api.py
nmgarza5: 😢

PR #7858, opened 2026-01-27
feat(ee):  fe - add billing hooks and actions

nmgarza5:
## Description

Simplifies the billing frontend to use two SWR hooks following codebase patterns:

- **`useLicense`** - License status for self-hosted deployments
- **`useBillingInformation`** - Billing info from Stripe (cloud + self-hosted)
- **`actions.ts`** - Simple async functions for mutations (checkout, portal, seat updates, license upload)

### Changes
- Add clean TypeScript interfaces matching backend Pydantic models
- Add `hasActiveSubscription` type guard
- Add `uploadLicense` action for air-gapped deployments

### Pattern
```typescript
// Data fetching
const { data, isLoading, error, refresh } = useLicense();
const { data, isLoading, error, refresh } = useBillingInformation();

// Mutations
await createCheckoutSession({ billing_period: "monthly" });
await createCustomerPortalSession();
await updateSeatCount({ new_seat_count: 10 });
```

Follow up PR will use these hooks/actions

## How Has This Been Tested?

e2e locally


web/src/hooks/useBillingInformation.ts
raunakab, line 1: Could we move this hooks to `/web/src/hooks` instead? If you don't like that location, then we should aim to move all *those* hooks to `/web/src/lib/hooks` instead. All in all, we should aim to generalize the location of our hooks.

web/src/hooks/useLicense.ts
raunakab, line 1: Same for this one as well!

web/src/lib/billing/interfaces.ts
raunakab, line 1: I've been playing around with the idea of consolidating interfaces into `/web/src/interfaces` or `/web/src/app/interfaces`. Same argument where we have so many arbitrary hooks and util functions all scattered throughout the app.

In this change, we still have feature-specific interfaces, but they all live in a centralized place (for instance, `/web/src/app/interfaces/billing.ts` for this feature). That way, anyone who wants to go through all the interfaces for a specific feature knows EXACTLY where to find it.

web/src/app/ee/admin/billing/SubscriptionSummary.tsx
```diff
@@ -1,32 +1,36 @@
...
 
 interface SubscriptionSummaryProps {
-  billingInformation: any;
+  billingInformation: BillingInformation;
 }
 
 export function SubscriptionSummary({
   billingInformation,
 }: SubscriptionSummaryProps) {
+  const formatDate = (dateStr: string | null): string => {
```
raunakab: We should move all date-formatting utilities into `/web/src/lib/utils.ts`. I guess this is very simple, but just for completeness's sake.

web/src/lib/billing/svc.ts
raunakab, line 1: Minor nit, but Subash has been calling them `svc` in his implementations (short for service, I believe). I don't have a preference for which name is used, but we should consolidate what BE API hitters are called (whether that be `svg`, `actions`, or whatever else).

raunakab (approved):
Looks good to me! I would maybe tackle some of the organizational comments that I have. However, those shouldn't block you. Approving!

web/src/hooks/useBillingInformation.ts
nmgarza5: oops yeah my bad. meant to put it there

web/src/lib/billing/svc.ts
nmgarza5: yeah doesn tmatter to me!

PR #7873, opened 2026-01-27
fix(llm): Hide private models from Agent Creation

justin-tahara:
## Description

Updating the backend api check for LLM management in the Agent Creation page

## How Has This Been Tested?

Tested locally by making an LLM config private and setting myself to a basic user. 

Also added a proper test

Basic User:
Screenshot 2026-01-27 at 11 01 49 AM

Admin User:
Screenshot 2026-01-27 at 11 02 38 AM

## Additional Options

---
## Summary by cubic
Hide private LLM providers from the Agent Creation page by enforcing centralized access control in the listing API. Non-public, unrestricted providers no longer appear for basic users; admins still see all.

- **Bug Fixes**
  - Switched /llm/provider listing to use can_user_access_llm_provider(persona=None) for consistent checks.
  - Non-public providers with no groups/personas are now admin-only; persona-restricted providers are hidden without a persona context.
  - Added integration test validating visibility for basic users vs. admins.

<sup>Written for commit 10d83fe1b1315d0c692034f825adb0a0c967d27b. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/server/manage/llm/api.py
+++ b/backend/onyx/server/manage/llm/api.py
@@ -410,7 +410,7 @@ def list_llm_provider_basics(
 
     all_providers = fetch_existing_llm_providers(db_session)
     user_group_ids = fetch_user_group_ids(db_session, user) if user else set()
-    is_admin = user and user.role == UserRole.ADMIN
+    is_admin = user is not None and user.role == UserRole.ADMIN
 
     accessible_providers = []
 


backend/onyx/server/manage/llm/api.py
```diff
@@ -415,21 +415,15 @@ def list_llm_provider_basics(
...
-                accessible_providers.append(LLMProviderDescriptor.from_model(provider))
-        elif not provider.personas:
-            # No restrictions = accessible
+        # Use centralized access control logic with persona=None since we're
+        # listing providers without a specific persona context. This correctly:
+        # - Includes all public providers
+        # - Includes providers user can access via group membership
+        # - Excludes persona-only restricted providers (requires specific persona)
+        # - Excludes non-public providers with no restrictions (admin-only)
+        if can_user_access_llm_provider(
```
yuhongsun96, line 424: Perhaps this function should take a User object instead. Also I think providers can be shared with users, I'm pretty sure this functionality does not allow for that logic

justin-tahara: Can take a look as a followup PR

Issue #7381, opened 2026-01-13
`npm install` changes the lockfile on Linux


jmelahman:
I personally get:

```
diff --git a/web/package-lock.json b/web/package-lock.json
index 812e5fc10..afbb9947f 100644
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -10308,7 +10308,6 @@
     },
     "node_modules/fsevents": {
       "version": "2.3.2",
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
```

and depdendabot get these `"peer": true,` additions, https://github.com/onyx-dot-app/onyx/pull/6847/changes#r2632344724.

Mostly for personal note-keeping, I've tried,

`.npmrc`

```
omit=optional
os=darwin
```

Moving `fsevents` to `optionalDependencies`.

Personally migrating to `bun`/`yarn`/`pnpm` but all use separate lockfile syntax.

[chore(pre-commit): run npm install after web/package.json changes](https://github.com/onyx-dot-app/onyx/pull/7373) should make my necessity for running `npm install` explicitly much less common, but we'd still get these useless diffs when updating packages including via dependabot.

Moving off `npm` is unironically probably the move.

PR #7888, opened 2026-01-27
fix(citations): enable citation sidebar w/ web_search-only assistants

jmelahman:
## Description

Makes `retrievalEnabled` on the ChatPage consistent with `personaIncludesRetrieval`. In chats with only web searches, this `retrievalEnabled` was incorrectly being set to `false`.

I think we could also just always render this component? That way, the only logic involved with it appearing is whether the citation button is clicked, but this is the smallest, correct fix to resolve the issue for now.

Will be cherry-picking to `v2.9`, `v2.10`, and `v2.11`.

## How Has This Been Tested?

Created an assistant with only Web Search enabled and _not_ `Open URL` setting, submitted a prompt which executed a web search and confirmed the citation sidebar opens as expected.

This is very much something we could captured by saving such a query in the seeded db snapshot and ensure we can click and see the sidebar with playwright. Will thinking about it.

## Additional Options

Closes https://linear.app/onyx-app/issue/ENG-3520/the-sources-sidebar-is-broken

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -10365,6 +10365,7 @@
     },
     "node_modules/fsevents": {
       "version": "2.3.2",
+      "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [


web/package-lock.json
```diff
@@ -10365,7 +10365,6 @@
     },
     "node_modules/fsevents": {
       "version": "2.3.2",
-      "dev": true,
```
nmgarza5, line 10368: intended?

jmelahman: No :rage: https://github.com/onyx-dot-app/onyx/issues/7381. This is actually incredibly cumbersome for me :/

jmelahman: For context, I undo this change, then the `npm-install-check` pre-commit hook "fixes" it while blocking my push and I'm left like this,
20260127_12h35m35s_grim

PR #7907, opened 2026-01-27
chore(persona): Persona model override refactor

Danelegend:
## Description
Models are currently associated with personas via the provider name + model name. A better way to do this is linking via model_configuration_id so that we point to the correct model. This changes the database table for Personas, migrating old columns to the new column. Mentions of llm_provider_override and llm_model_override are then replaced for model_configuration_id_override. Upstream changes are appropriately changed.

## How Has This Been Tested?
Passes all current tests.
Check that migration w/ data before moves data to after.
Screenshot 2026-01-28 at 1 06 32 PM

## Additional Options

---
## Summary by cubic
Switch persona model selection to a default_model_configuration_id instead of provider/version strings. This makes model choice explicit, enforces access controls, and improves validation.

- **Refactors**
  - Replace llm_model_provider_override and llm_model_version_override with default_model_configuration_id across DB, server models, APIs, and seeding.
  - Update LLM factory to resolve provider via model configuration ID or explicit override, then fall back to the default provider; remove unused LLMConfig path.
  - Add helpers to fetch model/provider by model configuration ID.
  - Rename get_valid_model_names_for_persona to get_valid_models_for_persona, returning ModelConfigurationView (with id).
  - Validate persona’s default model on fetch; if the ID isn’t accessible, set it to the first valid model or None.
  - When removing an LLM provider, clear personas’ default_model_configuration_id that reference it.
  - Introduce a shared ModelConfiguration base model and update well-known provider descriptors to use it.

- **Migration**
  - Update clients to send/consume default_model_configuration_id in upserts and snapshots.
  - Use get_valid_models_for_persona and the returned ModelConfigurationView.id for selection.
  - Apply the DB migration to add the FK and migrate existing overrides to matching model configurations with sensible fallbacks.
  - Ensure seed data uses default_model_configuration_id.

<sup>Written for commit 89696a7f57e3f02c1345d1181c5e171a1be00c84. Summary will update on new commits.</sup>


backend/onyx/llm/factory.py
```diff
@@ -53,61 +55,18 @@ def _build_provider_extra_headers(
     return {}
 
 
-def get_llm_config_for_persona(
```
Danelegend, line 56: Function not used anywhere

backend/alembic/versions/1b5455f99105_persona_usees_model_configuration_id.py
```diff
@@ -0,0 +1,153 @@
...
+            WHERE persona.llm_model_provider_override IS NOT NULL
+              AND persona.llm_model_version_override IS NULL
+              AND persona.model_configuration_id_override IS NULL
+              AND lp.name = persona.llm_model_provider_override
+              AND mc.name = lp.default_model_name
+        """
+        )
+    )
+
+    # Strategy 3: Only model is present - try to find the model in any provider
```
Danelegend: Hmm, I think this might be worth removing as we're just choosing any provider.

For example, user sets up gpt-4o on azure. azure gets deleted. This will default to gpt-4o on openai I assume.

Imo, we should port this over to the global default model

web/src/app/admin/configuration/llm/ModelConfigurationField.tsx
```diff
@@ -144,6 +144,7 @@ export function ModelConfigurationField({
               <CreateButton
                 onClick={() => {
                   arrayHelpers.push({
+                    id: -1,
```
raunakab, line 147: Why negatives here?

raunakab (commented):
Approving FE changes! Will leave it up to @yuhongsun96 to tackle the rest and give the final stamp.

web/src/app/admin/configuration/llm/ModelConfigurationField.tsx
Danelegend: Arbitrary id for type. Cursor said before sending it will filter out empty names (ie. this guy)

Danelegend:
@greptile

backend/onyx/llm/factory.py
nmgarza5: it used to be 🥲 
someone else did a poor refactor 
https://github.com/onyx-dot-app/onyx/commit/9caf136f0e4c2822a71c8e2e5ad60c2069f508ca

web/src/app/admin/configuration/llm/ModelConfigurationField.tsx
nmgarza5: this smells but if it works then so be it 😄

web/src/lib/llm/utils.ts
```diff
@@ -41,26 +46,24 @@ export function getLLMProviderOverrideForPersona(
   liveAssistant: MinimalPersonaSnapshot,
   llmProviders: LLMProviderDescriptor[]
 ): LlmDescriptor | null {
-  const overrideProvider = liveAssistant.llm_model_provider_override;
-  const overrideModel = liveAssistant.llm_model_version_override;
+  const overrideModelId = liveAssistant.default_model_configuration_id;
```
nmgarza5:
```suggestion
  const defaultModelConfigurationId = liveAssistant.default_model_configuration_id;
```

nmgarza5: should be consistent

nmgarza5 (approved):
Changes look good to me. I like it! No major concerns other than testing that llm access controls still work as expected

backend/alembic/versions/1b5455f99105_persona_usees_model_configuration_id.py
```diff
@@ -0,0 +1,131 @@
...
+            FROM model_configuration mc
+            JOIN llm_provider lp ON mc.llm_provider_id = lp.id
+            WHERE persona.llm_model_provider_override IS NULL
+              AND persona.llm_model_version_override IS NOT NULL
+              AND persona.default_model_configuration_id IS NULL
+              AND lp.is_default_provider = true
+              AND mc.name = lp.default_model_name
+        """
+        )
+    )
```
justin-tahara, line 79: This is frightening especially trying to do this on the cloud with over 12k tenants now

Danelegend (commented):
Pending merge till next week

PR #7910, opened 2026-01-28
feat(craft): update github action for craft latest

rohoswagger:
## Description

update github action for craft-latest, so we can generate a tag

## How Has This Been Tested?

careful review

```
git tag craft-latest
git push origin craft-latest
```

## Additional Options

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -1214,6 +1214,8 @@ jobs:
             latest=false
           tags: |
             type=raw,value=craft-latest
+            # TODO: Consider aligning craft-latest tags with regular backend builds (e.g., latest, edge, beta)
+            # to keep tagging strategy consistent across all backend images
 
       - name: Create and push manifest
         env:


.github/workflows/deployment.yml
```diff
@@ -1003,6 +1018,215 @@ jobs:
...
+
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@c299e40c65443455700f0fdfc63efafe5b349051 # ratchet:docker/metadata-action@v5
+        with:
+          images: ${{ env.REGISTRY_IMAGE }}
+          flavor: |
+            latest=false
+          tags: |
+            type=raw,value=craft-latest
```
jmelahman, line 1216: what does `value=craft-latest` here mean?

jmelahman: Oh this is tags. I was thinking this was something else

jmelahman, line 1216: This smells a bit. It seems like we want to have ~about the same tags as the backend image since these are interchangeable in the `docker-compose`. Consider at least a TODO

.github/workflows/deployment.yml
```diff
@@ -1003,6 +1018,215 @@ jobs:
...
+    if: needs.determine-builds.outputs.build-backend-craft == 'true'
+    runs-on:
+      - runs-on
+      - runner=2cpu-linux-x64
+      - run-id=${{ github.run_id }}-merge-backend-craft
+      - extras=ecr-cache
+    timeout-minutes: 90
+    environment: release
+    env:
+      REGISTRY_IMAGE: onyxdotapp/onyx-backend
```
jmelahman, line 1181: Is this correct? Does it need `-craft` namespaced?

deployment/docker_compose/install.sh
```diff
@@ -63,7 +63,7 @@ NC='\033[0m' # No Color
 
 # Step counter variables
 CURRENT_STEP=0
-TOTAL_STEPS=8
+TOTAL_STEPS=10
```
jmelahman, line 66: not a big deal, but ideally this is a separate PR (dunno what this does -- not reviewing, etc.)

rohoswagger:
CURSOR RESPONSE

**Comment 1**: Regarding `value=craft-latest` - confirmed, this is the Docker image tag that will be applied to the final manifest.

**Comment 2**: Regarding tag strategy - You make a good point about consistency with regular backend builds. The craft-latest tag currently only has one tag value. However, this is intentional for the initial rollout:
- Regular backend builds have: `latest`, `edge`, `beta` tags (for production/stable versions)
- Craft-latest is a dedicated tag specifically for Craft-enabled builds
- Adding a TODO is a good idea for future enhancement to align the tagging strategy

**Comment 3**: Regarding REGISTRY_IMAGE namespace - The current approach keeps craft builds in the same `onyxdotapp/onyx-backend` repository with just a different tag (`craft-latest`). This is intentional because:
- The Craft version IS a backend image, just built with Craft dependencies enabled
- Users can switch between `latest` and `craft-latest` tags in docker-compose without changing the image name
- Makes deployment simpler and more flexible

**Comment 4**: Regarding TOTAL_STEPS change - Understood. The install.sh fix is technically unrelated to this PR. However, it was needed because the original script had TOTAL_STEPS=8 but actually had 10 print_step calls, causing the progress display to show "Step 10/9". Would you prefer we:
- Revert this change to a separate PR?
- Add a comment explaining the fix?
- Keep as-is since it corrects an existing bug?

.github/workflows/deployment.yml
rohoswagger: nope, cause ideally its the same backend image (they are interchangeable), just a different tag. all craft images also work with chat so didn't see a reason to have a separate image

deployment/docker_compose/install.sh
rohoswagger: oop yea mb

PR #7955, opened 2026-01-28
refactor(craft): fix pre-provisioning state management, fix demo data state management

wenxi-onyx:
## Additional Options

---
## Summary by cubic
Fixes Craft pre-provisioning and demo data handling so new sessions always match the user’s setting, and the sandbox is reliably ready before input. Adds a demo_data_enabled flag to BuildSession and makes the demo toggle cookie-driven.

- **Bug Fixes**
  - Align pre-provisioned sessions with the user’s demo-data preference by storing demo_data_enabled on BuildSession and only reusing empty sessions with a matching setting.
  - Harden pre-provisioning: single in-flight promise, controller-managed triggers, exponential backoff on failure, and tab-focus revalidation; input is disabled while provisioning or after a failure.
  - Always consume the pre-provisioned session on first message; remove fallback create path; wait/retry file uploads before streaming; delay session naming to avoid races.
  - Make demo-data a cookie-backed single source of truth; invalidate/delete mismatched pre-provisioned sessions when the setting changes; show the demo pill only on the welcome page; auto-enable demo data if no connector has ever synced.
  - Safer session loading to prevent duplicate API calls; add sandbox status indicator on Configure.

- **Migration**
  - Adds demo_data_enabled BOOLEAN to build_session (default true). Run: alembic upgrade head.
  - No config changes; existing sessions default to demo mode.

<sup>Written for commit f8a9853bd3778dadda425f00d004d302411369ac. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/server/features/build/db/build_session.py
+++ b/backend/onyx/server/features/build/db/build_session.py
@@ -108,9 +108,11 @@ def get_empty_session_for_user(
         user_id: The user ID
         db_session: Database session
         max_age_minutes: Maximum age of session to consider (default 30)
-        demo_data_enabled: If provided, only match sessions with this demo_data setting.
+        demo_data_enabled: Match sessions with this demo_data setting.
                           This ensures pre-provisioned sessions match the user's current
                           preferences. If None, matches any session regardless of setting.
+                          Note: None is only used internally for operations that need to
+                          match any session (e.g., deletion).
     """
     cutoff = datetime.utcnow() - timedelta(minutes=max_age_minutes)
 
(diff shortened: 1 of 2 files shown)


backend/onyx/server/features/build/db/build_session.py
```diff
@@ -90,20 +100,33 @@ def get_empty_session_for_user(
...
-            ~exists().where(BuildMessage.session_id == BuildSession.id),
-        )
-        .first()
+    query = db_session.query(BuildSession).filter(
+        BuildSession.user_id == user_id,
+        BuildSession.created_at > cutoff,
+        ~exists().where(BuildMessage.session_id == BuildSession.id),
     )
 
+    if demo_data_enabled is not None:
```
rohoswagger, line 125: why not just directly filter on demo_data_enabled? demo_data_enabled can default to False instead of None

wenxi-onyx: Need to leave a none default for deletion endpoint.

PR #7967, opened 2026-01-29
chore: Better enforcement of masking

yuhongsun96:
## Description
Standardizing masking of sensitive fields. This static time check now prevents developers from forgetting to mask fields as the value is unusuable and will now raise an error if not explicitly extracted with either a mask applied or not.

## How Has This Been Tested?
Test setting/updating LLM keys
Test setting/updating web search keys
Test setting/updating connector credentials
Test Inference flow with tools

## Additional Options

---
## Summary by cubic
Enforces explicit masking for all secrets by wrapping encrypted columns with SensitiveValue. Backend code unwraps for internal use; API responses mask by default across connectors, OAuth, bots, KV, MCP, tools, and manage/federated APIs.

- **New Features**
  - Introduced SensitiveValue with get_value(apply_mask=...). EncryptedString/EncryptedJson return SensitiveValue on reads and accept str/SensitiveValue on writes.
  - Centralized mask_string and mask_credential_dict in onyx.utils.encryption; removed legacy masking from server/utils.
  - Broad adoption with unwrapping before use: connectors and permission sync (GitHub/Gmail/Drive/Jira/SharePoint/Slack/Teams), OAuth token manager and status API, Slack/Discord bots (backend unwraps; manage APIs mask), KV store, MCP (extract_connection_data and tool headers), tools (web search/image/MCP), federated APIs/retrieval, and documents APIs.
  - DB helpers merge credential_json using unwrapped values and expire ORM instances after writes so columns reload as SensitiveValue.
  - Testing and typing: make_mock_sensitive_value helper, SensitiveValue unit/typing tests, and mypy-safe updates across models/APIs.

- **Migration**
  - In backend code, unwrap with .get_value(apply_mask=False); use apply_mask=True for anything returned to clients.
  - Update tests/mocks to use make_mock_sensitive_value and expire ORM instances after writes when you need SensitiveValue reloaded.
  - Always pass unwrapped API keys/tokens to SDKs or external clients. No database schema changes.

<sup>Written for commit c6bb925a8cf78e5eff4abef1c97ce04b70e2cc06. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/server/features/mcp/api.py
+++ b/backend/onyx/server/features/mcp/api.py
@@ -80,6 +80,7 @@ from onyx.server.features.tool.models import ToolSnapshot
 from onyx.tools.tool_implementations.mcp.mcp_client import discover_mcp_tools
 from onyx.tools.tool_implementations.mcp.mcp_client import initialize_mcp_client
 from onyx.tools.tool_implementations.mcp.mcp_client import log_exception_group
+from onyx.utils.encryption import mask_string
 from onyx.utils.logger import setup_logger
 
 logger = setup_logger()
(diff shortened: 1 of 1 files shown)


evan-onyx (approved):
make sure u fix tests before merging but approach looks fine

backend/onyx/db/mcp.py
```diff
@@ -269,7 +282,7 @@ def update_connection_config(
     config = get_connection_config_by_id(config_id, db_session)
 
     if config_data is not None:
-        config.config = config_data
+        config.config = config_data  # type: ignore[assignment]
```
yuhongsun96, line 288: Here and in a lot of places it's because the assignment requires it to be a str type but SqlAlchemy doesn't really differentiate between read/writes. Assignment should be a string but the type on read is the SensitiveValue type

backend/onyx/db/models.py
justin-tahara, line 1: This file is the one that I am most worried about but I think we should just get this deployed internally before pushing to cloud

PR #7989, opened 2026-01-29
feat: agent view only modal

raunakab:
## Description

This PR adds an agent-view-only modal to the agents navigation page.

Addresses: https://linear.app/onyx-app/project/agent-view-only-page-674ef7981d12.

## Screenshots

image

## Additional Options

### DIFF
--- a/web/src/lib/time.ts
+++ b/web/src/lib/time.ts
@@ -1,4 +1,4 @@
-import { User } from "./types";
+import { User } from "@/lib/types";
 
 const conditionallyAddPlural = (noun: string, cnt: number) => {
   if (cnt > 1) {


web/src/lib/utils.ts
```diff
@@ -187,3 +187,8 @@ export function mergeRefs<T>(
     });
   };
 }
+
+export function formatMmDdYyyy(d: string): string {
```
nmgarza5: sir... what are we doing here 😆 


`web/src/lib/dateUtils.ts` or `web/src/components/dateRangeSelectors/dateUtils.ts` please. We can consolidate them later

web/src/sections/modals/AgentViewerModal.tsx
```diff
@@ -0,0 +1,420 @@
...
+}
+
+function ViewerMCPServerCard({ server, tools }: ViewerMCPServerCardProps) {
+  const [folded, setFolded] = useState(false);
+  const serverIcon = getActionIcon(server.server_url, server.name);
+
+  return (
+    <ExpandableCard.Root isFolded={folded} onFoldedChange={setFolded}>
+      <ExpandableCard.Header>
+        <div className="p-2">
```
nmgarza5, line 62: cn()

web/src/sections/modals/AgentViewerModal.tsx
```diff
@@ -0,0 +1,420 @@
...
+        </div>
+      </ExpandableCard.Header>
+    </ExpandableCard.Root>
+  );
+}
+
+const EMPTY_DOCS: [] = [];
+
+/**
+ * Floating ChatInputBar below the AgentViewerModal.
```
nmgarza5, line 123: lets chat when you get into the office, this one seems interesting. why do we wanna be able to send a chat message from this modal

web/src/sections/modals/AgentViewerModal.tsx
```diff
@@ -0,0 +1,420 @@
...
+
+/**
+ * AgentViewerModal - A read-only view of an agent's configuration
+ *
+ * This modal is the view-only counterpart to `AgentEditorPage.tsx`. While
+ * AgentEditorPage allows creating and editing agents with forms and inputs,
+ * AgentViewerModal displays the same information in a read-only format.
+ *
+ * Key differences from AgentEditorPage:
+ * - Modal presentation instead of full page
```
nmgarza5, line 164: are we surfacing all agent details or a limited view? As the number of Agent attributes increases, i want to be sure we dont overcrowd the modal. nothing actionable here, just calling out a concern if you haven't thought about it yet

web/src/sections/modals/AgentViewerModal.tsx
```diff
@@ -0,0 +1,420 @@
...
+  const hasActions = mcpServersWithTools.length > 0 || openApiTools.length > 0;
+  const defaultModel = getDisplayName(agent, llmProviders ?? []);
+
+  return (
+    <Modal
+      open={agentViewerModal.isOpen}
+      onOpenChange={agentViewerModal.toggle}
+    >
+      <Modal.Content
+        width="md-sm"
```
nmgarza5, line 247: call me crazy but i like "sm-md" more lol SUPER NITTY comment 😆

nmgarza5:
canceled auto merge for the date formatter

web/src/lib/utils.ts
raunakab: Moved `formatMmDdYyyy` to `web/src/lib/dateUtils.ts` and updated the import in AgentViewerModal.

web/src/sections/modals/AgentViewerModal.tsx
raunakab: This is a static string (`"p-2"`), so `cn()` isn't needed here — it's only useful for conditional or merged classes.

web/src/sections/modals/AgentViewerModal.tsx
raunakab: The idea is to let users quickly start a conversation with an agent directly from the viewer modal — saves the extra step of closing the modal, finding the agent, and clicking Start Chat. The input bar pre-fills the message and navigates to the chat page. Happy to chat more about this in person!

web/src/sections/modals/AgentViewerModal.tsx
raunakab: Good callout. Right now it's a curated subset — the "More Info" section is collapsed by default via `SimpleCollapsible` so it doesn't visually overwhelm. As agent attributes grow we can revisit which ones make the cut for this modal vs. requiring the full editor page.

web/src/sections/modals/AgentViewerModal.tsx
raunakab: `md-sm` is the existing key name in Modal's `widthClasses`. Renaming to `sm-md` would be a separate Modal API change — can do that in a follow-up if you feel strongly about it\!

PR #7996, opened 2026-01-30
chore(gha): build desktop app in CI

jmelahman:
## Description

CI job to confirm the Desktop app builds

## How Has This Been Tested?

Should run on this PR

## Additional Options

---
## Summary by cubic
Adds a GitHub Actions workflow that builds the Desktop app in CI to catch build issues early and publish artifacts for review. Runs on PRs, merge queues, and version tags.

- **New Features**
  - Triggers on PRs to main/release, merge_group, and v*.*.* tags; cancels superseded runs.
  - Builds on Linux (x86_64-unknown-linux-gnu) producing deb and rpm bundles; macOS and Windows temporarily disabled.
  - Uses Node 24, Rust stable, and npm/Cargo caching.
  - Installs Tauri Linux dependencies; runs npm ci and builds via Tauri in desktop/.
  - Uploads unsigned bundles from desktop/src-tauri/target/release/bundle as artifacts (kept 7 days).

<sup>Written for commit 3c65317522338016a6d84d9cfbb541308b5c4325. Summary will update on new commits.</sup>

### DIFF
--- a/.github/workflows/pr-desktop-build.yml
+++ b/.github/workflows/pr-desktop-build.yml
@@ -37,10 +37,11 @@ jobs:
           #  os: macos-latest
           #  target: universal-apple-darwin
           #  args: "--target universal-apple-darwin"
-          - platform: windows
-            os: windows-latest
-            target: x86_64-pc-windows-msvc
-            args: ""
+          # TODO: Fix and enable the Windows build.
+          #- platform: windows
+          #  os: windows-latest
+          #  target: x86_64-pc-windows-msvc
+          #  args: ""
 
     steps:
       - name: Checkout code


.github/workflows/pr-desktop-build.yml
```diff
@@ -0,0 +1,103 @@
...
+  pull_request:
+    branches:
+      - main
+      - "release/**"
+    paths:
+      - "desktop/**"
+      - ".github/workflows/pr-desktop-build.yml"
+  push:
+    tags:
+      - "v*.*.*"
```
justin-tahara, line 17: Do we want to consider other tags?

justin-tahara: Likely nightly or you think it's okay

justin-tahara (commented):
Does this handle the matrix? Like Apple, Linux and Windows?

.github/workflows/pr-desktop-build.yml
jmelahman: Desktop requires semver tags, so those would just fail anyways?

jmelahman: In this case, we might use the fallback dev version tag for the actual build? but at any rate im not too worried yet. if the nightly builds fail, the actual build will fail which has notifications, etc already

jmelahman:
> Does this handle the matrix? Like Apple, Linux and Windows?

The macOS build is failing, so just going to drop. If we run into issues of the macOS build failing, we can revisit I think. Hopefully, just Linux (and maybe Windows) catches most issues (deps + compiler issues) anyways.

justin-tahara (approved):

PR #8025, opened 2026-01-30
feat(provider): create flow mapping table

Danelegend:
## Description
Introduces a new flow mapping table allowing each model configuration to be able to be mapped to a particular flow that it should be apart of.

## How Has This Been Tested?
Migration successful

## Additional Options

---
## Summary by cubic
Adds an LLMModelFlow table to map model configurations to flow types (chat, vision, embeddings) and enforce a single default per flow type. This enables per‑flow model selection with a clear default.

- **New Features**
  - LLMModelFlowType enum: chat, vision, embeddings.
  - LLMModelFlow model with:
    - Unique llm_model_flow_type + model_configuration_id.
    - Partial unique index to allow only one default per llm_model_flow_type.
    - Cascade delete with its model configuration.
  - ModelConfiguration.llm_model_flows relationship and llm_model_flow_types property.

- **Migration**
  - Run Alembic migrations.
  - Set is_default=true on exactly one row per llm_model_flow_type if a default is required.

<sup>Written for commit dee1bfe02c3df5990ec35c8f9b8d2dfd901c7397. Summary will update on new commits.</sup>

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -149,6 +149,11 @@ RUN if [ "$ENABLE_CRAFT" = "true" ]; then \
         ENABLE_CRAFT=true /app/scripts/setup_craft_templates.sh; \
     fi
 
+# Set Craft template paths to the in-image locations
+# These match the paths where setup_craft_templates.sh creates the templates
+ENV OUTPUTS_TEMPLATE_PATH=/app/onyx/server/features/build/sandbox/kubernetes/docker/templates/outputs
+ENV VENV_TEMPLATE_PATH=/app/onyx/server/features/build/sandbox/kubernetes/docker/templates/venv
+
 # Put logo in assets
 COPY --chown=onyx:onyx ./assets /app/assets
 


Danelegend:
@greptile

Danelegend:
@greptile

Danelegend:
@greptile

backend/alembic/versions/f220515df7b4_add_flow_mapping_table.py
```diff
@@ -0,0 +1,57 @@
...
+# revision identifiers, used by Alembic.
+revision = "f220515df7b4"
+down_revision = "cbc03e08d0f3"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "model_flow",
```
justin-tahara: Is this the name that we are going with?

backend/onyx/db/enums.py
```diff
@@ -285,3 +285,9 @@ class HierarchyNodeType(str, PyEnum):
 
     # Slack
     CHANNEL = "channel"
+
+
+class ModelFlowType(str, PyEnum):
+    CONVERSATION = "conversation"
+    VISION = "vision"
+    EMBEDDINGS = "embeddings"
```
justin-tahara: Should this be chat or conversation? I don't think I see many places myself where I see conversation but could be wrong.

Would Web Search and things like this fall under here or those are different non-model related flows so we don't want to group them here

backend/alembic/versions/f220515df7b4_add_flow_mapping_table.py
Danelegend: Spoke w/ Evan and we ended up going with this

Danelegend:
@greptile

justin-tahara (approved):

PR #8040, opened 2026-01-31
chore(fe): allow inverse inset for inline elements

jmelahman:
## Description

This was causing layout issues (unexpected scrollbars) on the chat page.

I think the issue is that the FrostedDiv on the chat page wraps a block-level element (in theory, they shouldn't be block-level, but w.e.) whereas this WelcomeMessage's wraps an inline element which needs this.

Given this seems like a bit of a footgun, it's off by default w/ a message to care when used.

## How Has This Been Tested?

Note the dark bloom behind the `How can I help you?`:

#### Before

20260130_17h18m56s_grim

#### After

20260130_17h18m44s_grim

## Additional Options

---
## Summary by cubic
Added an opt-in exceedBounds prop to FrostedDiv to let the blur bloom outside its container when needed. Default is off to avoid layout issues.

- **New Features**
  - FrostedDiv supports exceedBounds; when true, the overlay uses a negative inset based on blur to extend beyond bounds.

- **Bug Fixes**
  - Removes unexpected scrollbars on the chat page by enabling exceedBounds for WelcomeMessage.

<sup>Written for commit 7bb7329b959e11d1bc802efe4ce519fc91d5fa42. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/refresh-components/FrostedDiv.tsx
+++ b/web/src/refresh-components/FrostedDiv.tsx
@@ -79,7 +79,7 @@ export default function FrostedDiv({
         className={cn("absolute pointer-events-none", overlayClassName)}
         style={{
           // Extend beyond bounds to allow blur to bloom outward
-          inset: exceedBounds ? `-${blur}` : undefined,
+          inset: exceedBounds ? `calc(-1 * ${blur})` : undefined,
           borderRadius,
           background: backgroundColor,
           filter: `blur(${blur})`,


web/src/refresh-components/FrostedDiv.tsx
```diff
@@ -71,6 +78,8 @@ export default function FrostedDiv({
       <div
         className={cn("absolute pointer-events-none", overlayClassName)}
         style={{
+          // Extend beyond bounds to allow blur to bloom outward
+          inset: exceedBounds ? `-${blur}` : undefined,
```
jmelahman: P2: Negating `blur` with `-${blur}` breaks when `blur` is a CSS variable or non-literal length (e.g., `var(--frost-blur)`), producing an invalid CSS value. Use `calc(-1 * ${blur})` so variable-based lengths still work.

<details>
<summary>Prompt for AI agents</summary>

```text
Check if this issue is valid — if so, understand the root cause and fix it. At web/src/refresh-components/FrostedDiv.tsx, line 82:

<comment>Negating `blur` with `-${blur}` breaks when `blur` is a CSS variable or non-literal length (e.g., `var(--frost-blur)`), producing an invalid CSS value. Use `calc(-1 * ${blur})` so variable-based lengths still work.</comment>

<file context>
@@ -71,6 +78,8 @@ export default function FrostedDiv({
         className={cn("absolute pointer-events-none", overlayClassName)}
         style={{
+          // Extend beyond bounds to allow blur to bloom outward
+          inset: exceedBounds ? `-${blur}` : undefined,
           borderRadius,
           background: backgroundColor,
</file context>
```

</details>

```suggestion
          inset: exceedBounds ? `calc(-1 * ${blur})` : undefined,
```

✅ Addressed in [`99b86d3`](https://github.com/onyx-dot-app/onyx/commit/99b86d3081a2a56c63288d063aa2d18a3781edb8)

jmelahman:
Ugh, I think this might actually be needed for the chat page elements as well. This is behavior without the vignette at HEAD :/

20260130_17h34m21s_grim

Converted to draft until I can figure out how to achieve the frost without affecting layout.

PR #8061, opened 2026-02-01
fix: attaching user files to assistant

evan-onyx:
## Description

there was a bug where user files wouldn't actually get attached to an assistant AND included in search filters, now its fixed and tested E2E

## How Has This Been Tested?

tested in Ui and added automated tests for the two bugs

## Additional Options

---
## Summary by cubic
Fixes a bug where user files attached to an assistant were not saved or included in search. Search now includes user_file_ids with OR logic and the editor reliably persists them when knowledge is enabled.

- **Bug Fixes**
  - OpenSearch: include user_file_ids in the assistant knowledge filter (OR with attached documents and hierarchy nodes) and remove the separate AND filter; handle cases with only user_file_ids.
  - Web: simplify knowledge settings by removing the team/user toggle and persisting user_file_ids, document_set_ids, hierarchy_node_ids, and document_ids whenever knowledge is enabled.
  - UI: add ariaLabelPrefix for user file rows to stabilize selection; add “Celery user_file_processing” to launch configs.

- **Tests**
  - Added unit tests for the assistant knowledge filter to ensure user_file_ids are included with OR logic and no duplicate filtering.
  - Added E2E tests to verify user files are attached on create and remain selected on edit, including multi-file cases.

<sup>Written for commit d50ec32486dba4ff8a554142266837677dd360c9. Summary will update on new commits.</sup>

### DIFF
--- a/web/tests/e2e/assistants/user_file_attachment.spec.ts
+++ b/web/tests/e2e/assistants/user_file_attachment.spec.ts
@@ -102,10 +102,6 @@ async function uploadTestFile(
         }`
       );
       if (attempt === maxRetries) {
-        // Take a screenshot for debugging
-        await page.screenshot({
-          path: `upload-failure-${fileName}-${Date.now()}.png`,
-        });
         throw error;
       }
       await page.waitForTimeout(1000);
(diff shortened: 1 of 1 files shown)


.vscode/launch.json
```diff
@@ -25,6 +25,7 @@
         "Celery heavy",
         "Celery docfetching",
         "Celery docprocessing",
+        "Celery user_file_processing",
```
yuhongsun96, line 28: Ideally we consolidate these but maybe not right now

yuhongsun96 (approved):

.vscode/launch.json
evan-onyx: yeah I think for now its better to not have to start it explicitly during local dev

PR #8103, opened 2026-02-02
fix: LiteLLM in threading context

yuhongsun96:
## Description
Fixes "Bad file descriptor" errors occurring during concurrent LLM streaming requests (e.g., deep research with multiple parallel agents).

Root cause: When multiple threads stream from OpenAI simultaneously, they share litellm's internal HTTP connection pool. This causes race conditions where one thread's connection can be closed or returned to the pool while another thread is still reading from it.

Fix: Create an isolated HTTPHandler per streaming request in LitellmLLM.stream(). This ensures each concurrent stream has its own HTTP connection, preventing pool conflicts.

Key detail: Must use litellm's HTTPHandler wrapper (not raw httpx.Client) because response_api_handler checks isinstance(client, HTTPHandler) and falls back to the shared pool otherwise.

## How Has This Been Tested?
Works with nontrivially parallel deep research now

## Additional Options

---
## Summary by cubic
Fixes "Bad file descriptor" errors during concurrent LLM streaming by isolating HTTP connections per request.

- **Bug Fixes**
  - Create a dedicated litellm.HTTPHandler per stream and close it after use.
  - Pass the handler through _completion so streaming requests bypass the shared connection pool.
  - Close the handler in a finally block so connections are cleaned up when the stream finishes or is closed.

<sup>Written for commit 572e40be68615c8c21dca43b93f06a9a303d1d21. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/unit/onyx/llm/test_multi_llm.py
+++ b/backend/tests/unit/onyx/llm/test_multi_llm.py
@@ -1,3 +1,4 @@
+from unittest.mock import ANY
 from unittest.mock import patch
 
 import litellm
@@ -248,6 +249,7 @@ def test_multiple_tool_calls(default_multi_llm: LitellmLLM) -> None:
             temperature=0.0,  # Default value from GEN_AI_TEMPERATURE
             timeout=30,
             max_tokens=None,
+            client=None,
             parallel_tool_calls=True,
             mock_response=MOCK_LLM_RESPONSE,
             allowed_openai_params=["tool_choice"],
@@ -402,6 +404,7 @@ def test_multiple_tool_calls_streaming(default_multi_llm: LitellmLLM) -> None:
             temperature=0.0,  # Default value from GEN_AI_TEMPERATURE
             timeout=30,
             max_tokens=None,
+            client=ANY,  # HTTPHandler instance created per-stream
             stream_options={"include_usage": True},
             parallel_tool_calls=True,
             mock_response=MOCK_LLM_RESPONSE,


backend/onyx/llm/multi_llm.py
```diff
@@ -439,8 +444,15 @@ def stream(
         user_identity: LLMUserIdentity | None = None,
     ) -> Iterator[ModelResponseStream]:
         from litellm import CustomStreamWrapper as LiteLLMCustomStreamWrapper
+        from litellm import HTTPHandler
```
justin-tahara, line 446: Should we just move this to the top of the file? I agree that if it was only used in one place but it seems like we may use it more in the future as well.

backend/onyx/llm/multi_llm.py
```diff
@@ -454,14 +466,38 @@ def stream(
...
+                # Track LLM cost when usage info is available (typically in the last chunk)
+                if model_response.usage:
+                    self._track_llm_cost(model_response.usage)
+
+                yield model_response
+        except GeneratorExit:
+            client.close()
+            raise
+        finally:
+            client.close()
```
justin-tahara: When you get a GeneratorExit are you not closing the client twice? Or are you saying the `raise` is supposed to prevent this? 

Just an odd pattern if the finally is already catching and doing the same thing.

backend/onyx/llm/multi_llm.py
```diff
@@ -439,8 +444,15 @@ def stream(
...
 
+        # Create an isolated HTTP handler for this streaming request to avoid
+        # "Bad file descriptor" errors from connection pool conflicts when
+        # multiple threads stream concurrently (e.g., during deep research).
+        # Must use litellm's HTTPHandler wrapper, not raw httpx.Client, as
+        # litellm's response_api_handler checks for this specific type.
+        client = HTTPHandler()
         response = cast(
             LiteLLMCustomStreamWrapper,
             self._completion(
```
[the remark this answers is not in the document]
justin-tahara: Seems like a valid issue that we should handle

backend/onyx/llm/multi_llm.py
```diff
@@ -439,8 +444,15 @@ def stream(
...
         from litellm import CustomStreamWrapper as LiteLLMCustomStreamWrapper
+        from litellm import HTTPHandler
         from onyx.llm.model_response import from_litellm_model_response_stream
 
+        # Create an isolated HTTP handler for this streaming request to avoid
+        # "Bad file descriptor" errors from connection pool conflicts when
+        # multiple threads stream concurrently (e.g., during deep research).
+        # Must use litellm's HTTPHandler wrapper, not raw httpx.Client, as
+        # litellm's response_api_handler checks for this specific type.
+        client = HTTPHandler()
```
acaprau, line 460: why dont you just put this in a context manager? like cubic says you dont guarantee .close() if _completion raises. a context manager accomplishes what you want in _stream_with_cleanup but in a more pythonic way and with safer guarantees.

backend/onyx/llm/multi_llm.py
acaprau, line 446: nooo this can just go at the top imo.

backend/onyx/llm/multi_llm.py
```diff
@@ -51,6 +52,7 @@
 
 if TYPE_CHECKING:
     from litellm import CustomStreamWrapper
+    from litellm import HTTPHandler
```
acaprau, line 54: same comment as below.

backend/onyx/llm/multi_llm.py
yuhongsun96: :P That was the other option... but is it even necessary, what case am I even fixing for?

backend/onyx/llm/multi_llm.py
yuhongsun96: These are lazily imported on purpose. We have a check for this actually for heavy libs in the precommit hooks. The reason is that these are heavy libs and not used by background jobs so we don't want to load these.

backend/onyx/llm/multi_llm.py
yuhongsun96: To break it down more:
1. When the generator is fully consumed → finally runs immediately
2. When generator.close() is called → finally runs (GeneratorExit is raised inside the generator)
3. When the generator goes out of scope in CPython → refcounting finalizes it immediately, finally runs

The other edge case covering is that httpx has timeouts as well so regardless it will always be done without too much delay. We don't end up with any memory leak issues.

backend/onyx/llm/multi_llm.py
justin-tahara: Is HTTPHandler that heavy? I feel like this should be relatively light weight

justin-tahara:
@greptileai & @cubic-dev-ai can you review one more time?

evan-onyx (approved):

PR #8114, opened 2026-02-02
chore: LiteLLM bump version

yuhongsun96:
## Description
Bump LiteLLM version and update some relevant monkeypatches. More minimal now yay.

Major changes:
- LiteLLM version bumped. Needed for a client parameter to allow us to create httpx clients across parallel threads to avoid issues like shared connections being closed by other threads.
- Some monkeypatches were removed with the LiteLLM version bump.
- Azure handling for non-streaming was removed as a better solution was found and introduced which is to recognize Azure LLMs as streaming capable
- OpenAI patch is made more minimal (see below)

The old _patch_openai_responses_chunk_parser function was replaced with a new, more focused _patch_openai_responses_parallel_tool_calls function.

  Key Differences

  Old patch (~300 lines):
  - Completely re-implemented the entire chunk_parser method
  - Handled ALL streaming event types: response.created, response.output_item.added, response.output_item.done, response.output_text.delta, response.function_call_arguments.delta,
  response.reasoning_summary_text.delta, response.completed
  - Included Azure-specific logic to track _has_streamed_tool_calls and extract tool calls from response.completed (because Azure batches tool calls there instead of streaming them)

  New patch (~200 lines):
  - Only intercepts 4 specific event types that need fixes:
    - response.output_item.added (tool calls)
    - response.function_call_arguments.delta (tool call arguments)
    - response.output_item.done (tool call completion)
    - response.reasoning_summary_text.delta (reasoning newlines)
  - Delegates all other events to LiteLLM's original translate_responses_chunk_to_openai_stream static method

  Why the Azure _has_streamed_tool_calls Logic Was Removed

  The old patch tracked whether tool calls were streamed incrementally (OpenAI behavior) vs batched in response.completed (Azure behavior). This was removed because:

  1. It was overly defensive - The Azure Responses API behavior may have been a temporary quirk or specific to certain deployments
  2. The _patch_azure_responses_should_fake_stream patch already forces real streaming for Azure, which should make Azure behave like OpenAI
  3. Simpler is better - If Azure still batches tool calls in response.completed, that event will fall through to LiteLLM's original handler, which returns finish_reason="stop". This may need
  revisiting if it causes issues, but removing untested complexity is preferable.

  The Core Fix: output_index for Parallel Tool Calls

  The actual bug was simple: LiteLLM hardcodes index=0 for all tool calls in streaming:

  # LiteLLM's code (broken for parallel tool calls)
  tool_call_chunk = ChatCompletionToolCallChunk(
      id=output_item.get("call_id"),
      index=0,  # <-- Always 0, breaks parallel tool calls
      ...
  )

  The fix uses output_index from the OpenAI Responses API event:

  # Our fix
  output_index = parsed_chunk.get("output_index", 0)
  tool_call_chunk = ChatCompletionToolCallChunk(
      id=output_item.get("call_id"),
      index=output_index,  # <-- Correct index for each parallel tool call
      ...
  )

## How Has This Been Tested?
Works with Anthropic, OpenAI, AzureOpenAI, Ollama. The rest should not be affected and are not tested.

## Additional Options

---
## Summary by cubic
Upgraded LiteLLM to 1.81.6 and simplified our patches to fix streaming and tool-call behavior; also silences noisy ResponseAPIUsage warnings. This improves parallel tool calls, reasoning summaries, and time to first token, especially on Azure Responses; Ollama thinking content now streams correctly.

- **Bug Fixes**
  - OpenAI Responses: support parallel tool calls by using output_index; join multi-part reasoning summaries.
  - Azure Responses: force native streaming for custom deployments (no fake buffering) to improve TTFB.
  - Ollama: stream reasoning/thinking content across all chunks, not just the first ones.
  - Responses API usage: normalize usage format and avoid logging mutation to prevent serialization warnings.

- **Dependencies**
  - litellm: 1.80.11 → 1.81.6 (updated in requirements, pyproject, and lockfiles).

<sup>Written for commit 1f87fed62dd8c6155bbbecd89b964012ef956731. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/llm/litellm_singleton/monkey_patches.py
+++ b/backend/onyx/llm/litellm_singleton/monkey_patches.py
@@ -622,10 +622,10 @@ def _patch_responses_api_usage_format() -> None:
                         )
 
         # Call original model_construct (need to call it as unbound method)
-        return original_model_construct.__func__(cls, _fields_set, **values)
+        return original_model_construct.__func__(cls, _fields_set, **values)  # type: ignore[attr-defined]
 
     _patched_model_construct._is_patched = True  # type: ignore[attr-defined]
-    ResponsesAPIResponse.model_construct = _patched_model_construct  # type: ignore[method-assign]
+    ResponsesAPIResponse.model_construct = _patched_model_construct  # type: ignore[method-assign, assignment]
 
 
 def _patch_logging_assembled_streaming_response() -> None:


backend/onyx/llm/litellm_singleton/config.py
```diff
@@ -17,6 +18,14 @@ def configure_litellm_settings() -> None:
     litellm.add_function_to_prompt = False
     litellm.suppress_debug_info = True
 
+    # Suppress Pydantic serialization warning for ResponseAPIUsage type mismatch.
+    # LiteLLM's completions->responses API bridge passes usage data in chat completion
+    # format (completion_tokens, prompt_tokens) instead of Responses API format
+    # (output_tokens, input_tokens). This is a LiteLLM bug but harmless.
+    warnings.filterwarnings(
+        "ignore", message=".*ResponseAPIUsage.*", category=UserWarning
+    )
```
justin-tahara: As long as this is ignoring for the LiteLLM use case then I think this is fine.

backend/scripts/decrypt.py
justin-tahara, line 1: This seems unrelated. Can we please remove?

backend/onyx/llm/litellm_singleton/monkey_patches.py
```diff
@@ -1,186 +1,67 @@
-import json
+"""
+LiteLLM Monkey Patches
+
+This module addresses the following issues in LiteLLM:
+
+Status checked against LiteLLM v1.81.6-nightly (2026-02-02):
```
justin-tahara, line 6: I'm worried this will get stale very quickly. Hopefully the LLM or whoever edits this file remembers but this seems like something that will be confusing in the future.

backend/onyx/llm/litellm_singleton/monkey_patches.py
```diff
@@ -315,158 +196,177 @@ def _patched_chunk_parser(self: Any, chunk: dict) -> ModelResponseStream:
...
-                        type="function",
-                        function=ChatCompletionToolCallFunctionChunk(
-                            name=parsed_chunk.get("name", None),
-                            arguments="",  # responses API sends everything again, we don't
-                        ),
-                    ),
-                    is_finished=False,
-                    finish_reason="",
-                    usage=None,
+                provider_specific_fields = output_item.get("provider_specific_fields")
```
justin-tahara, line 352: Should we add a default here? Like a {}

backend/onyx/llm/litellm_singleton/monkey_patches.py
```diff
@@ -315,158 +196,177 @@ def _patched_chunk_parser(self: Any, chunk: dict) -> ModelResponseStream:
...
-                        type="function",
-                        function=ChatCompletionToolCallFunctionChunk(
-                            name=output_item.get("name", None),
-                            arguments=parsed_chunk.get("arguments", ""),
-                        ),
-                    ),
-                    is_finished=False,
-                    finish_reason="",
-                    usage=None,
+                provider_specific_fields = output_item.get("provider_specific_fields")
```
justin-tahara, line 283: Can we add default value here as well?

justin-tahara (approved):
Some small comments. I think overall logic is sound but let's not cherry pick anywhere and we can just keep testing models with this change in to make sure that things are working as intended

backend/onyx/llm/litellm_singleton/monkey_patches.py
yuhongsun96: It's pinned to a  major-minor-patch, it doesn't particularly minute what nightly etc. imo. I'm not sure I understand the concern

backend/onyx/llm/litellm_singleton/monkey_patches.py
yuhongsun96: Why do you want it?

PR #8129, opened 2026-02-03
feat(opensearch): Add tenant ID to the document chunk ID

acaprau:
## Description
It doesn't look like document IDs are guaranteed to be unique across tenants, so add tenant ID to the document chunk ID so essentially eliminate that as a source of collisions.

## How Has This Been Tested?
Tightened up a test in `backend/tests/external_dependency_unit/opensearch/test_opensearch_client.py` to exercise the specific case when document IDs collide, other than that CI should get me.

## Additional Options

---
## Summary by cubic
Add tenant-aware, sanitized OpenSearch chunk IDs to prevent cross-tenant collisions and invalid IDs. Updates indexing and migration paths to use filtered document IDs and real (unsanitized) doc IDs where appropriate.

- **New Features**
  - OpenSearch chunk ID now includes a short tenant ID in multitenant mode and uses a filtered document ID to avoid collisions.
  - Added filter_and_validate_document_id to remove unsafe characters and enforce ID length; used for ID generation and queries.
  - Indexing now strips invalid unicode from title, content, metadata, and blurb to avoid OpenSearch ingestion issues.
  - Migration transformer uses the real document ID (not Vespa’s sanitized version) and logs mismatches.

- **Migration**
  - If you call get_opensearch_doc_chunk_id or OpenSearchClient.index_document directly, update your code to pass tenant_state.

<sup>Written for commit 0772feeadc77f29e4da53844de7d9944ca2cf92e. Summary will update on new commits.</sup>

### DIFF
--- a/backend/Dockerfile.model_server
+++ b/backend/Dockerfile.model_server
@@ -48,6 +48,7 @@ WORKDIR /app
 # Utils used by model server
 COPY ./onyx/utils/logger.py /app/onyx/utils/logger.py
 COPY ./onyx/utils/middleware.py /app/onyx/utils/middleware.py
+COPY ./onyx/utils/tenant.py /app/onyx/utils/tenant.py
 
 # Place to fetch version information
 COPY ./onyx/__init__.py /app/onyx/__init__.py


backend/onyx/utils/tenant.py
```diff
@@ -0,0 +1,15 @@
...
+    """Gets a short string representation of a full tenant id.
+
+    Args:
+        tenant_id: The full tenant id.
+
+    Returns:
+        str: The first 8 characters of the tenant id after removing the prefix.
+    """
+    tenant_display = tenant_id.removeprefix(TENANT_ID_PREFIX)
+    short_tenant = tenant_display[:8] if len(tenant_display) > 8 else tenant_display
```
evan-onyx: the ppl before u were being silly, slicing with too large of an index just gets the whole string

evan-onyx (approved):

PR #8140, opened 2026-02-03
fix(craft): pod restoration race, recovery from unexpected state, and updating heartbeat on session creation

wenxi-onyx:
## Additional Options

---
## Summary by cubic
Fixes race conditions during sandbox session creation and Kubernetes pod restoration that caused unexpected states and occasional data loss. Adds a per-user lock and idempotent pod/service provisioning to safely reuse existing resources.

- **Bug Fixes**
  - Add Redis lock (300s) to serialize session creation per user; return 503 if lock wait times out.
  - Make Kubernetes provision idempotent: reuse healthy pods, handle 409 conflicts, ensure Service exists, and wait for readiness.
  - Avoid cleanup when a healthy pod likely belongs to a concurrent request; reduce accidental deletion.
  - Refresh sandbox heartbeat on RUNNING transitions by using the status update helper; remove manual status writes.

- **Dependencies**
  - Bump SANDBOX_CONTAINER_IMAGE default to onyxdotapp/sandbox:v0.1.2.

<sup>Written for commit cf5533fd11bb6d97e5cbe58da7052835d7ade2a6. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/server/features/build/sandbox/kubernetes/kubernetes_sandbox_manager.py
+++ b/backend/onyx/server/features/build/sandbox/kubernetes/kubernetes_sandbox_manager.py
@@ -834,13 +834,13 @@ done
                 if e.status == 409:
                     # Pod was created by another concurrent request
                     # Check if it's healthy and reuse it
-                    logger.info(
-                        f"Pod {pod_name} already exists (409 conflict), "
+                    logger.warning(
+                        f"Pod {pod_name} already exists (409 conflict, this shouldn't normally happen), "
                         "checking if it's healthy..."
                     )
                     if self._pod_exists_and_healthy(pod_name):
-                        logger.info(
-                            f"Pod {pod_name} created by concurrent request, reusing"
+                        logger.warning(
+                            f"During provisioning, discovered that pod {pod_name} already exists. Reusing"
                         )
                         # Continue to ensure service exists and wait for ready
                     else:
@@ -864,7 +864,9 @@ done
             except ApiException as e:
                 if e.status != 409:  # Ignore AlreadyExists
                     raise
-                logger.debug(f"Service {service_name} already exists, reusing")
+                logger.warning(
+                    f"During provisioning, discovered that service {service_name} already exists. Reusing"
+                )
 
             # 3. Wait for pod to be ready
             logger.info(f"Waiting for pod {pod_name} to become ready...")


backend/onyx/server/features/build/api/sessions_api.py
```diff
@@ -88,44 +92,73 @@ def create_session(
...
+        # Session was just created, so it's loaded in the sandbox
+        return DetailedSessionResponse.from_session_response(
+            base_response, session_loaded_in_sandbox=True
+        )
+    finally:
+        try:
+            lock.release()
+        except Exception:
+            # Lock may have expired or been released already
+            pass
```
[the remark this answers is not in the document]
rohoswagger: agree with this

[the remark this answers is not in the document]
wenxi-onyx: just removing the exception altogether and using if lock.owned()...

wenxi-onyx:

backend/onyx/server/features/build/api/sessions_api.py
```diff
@@ -74,6 +74,10 @@ def list_sessions(
     )
 
 
+# Lock timeout for session creation (should be longer than max provision time)
+SESSION_CREATE_LOCK_TIMEOUT_SECONDS = 300
```
rohoswagger, line 78: 5 minutes is a long time no? prolly 2-3 mins is enough, we can prolly add a timeout on session create as well

wenxi-onyx: This should never happen. The lock is released as soon as the session creation finishes. The longest possible sequence is 
```
- no pod exists (no sandbox)
- sandbox creation + image pull + data copying
- symlink creation
- session returned
```
tl;dr I think 5 mins is on the safe end, but likely 2-3 may not be enough on fresh starts

backend/onyx/server/features/build/sandbox/kubernetes/kubernetes_sandbox_manager.py
```diff
@@ -743,18 +825,46 @@ def provision(
...
-            )
+            try:
+                self._core_api.create_namespaced_pod(
+                    namespace=self._namespace,
+                    body=pod,
+                )
+            except ApiException as e:
+                if e.status == 409:
+                    # Pod was created by another concurrent request
+                    # Check if it's healthy and reuse it
```
rohoswagger, line 836: is this neccessary? given that we handle race conditions via the lock and earlier in the code

rohoswagger: generally this logic seems pretty messy, can prolly clean it up / simplify a bit

backend/onyx/server/features/build/session/manager.py
```diff
@@ -506,8 +510,10 @@ def create_session__no_commit(
                         tenant_id=tenant_id,
                         llm_config=llm_config,
                     )
-                    sandbox.status = sandbox_info.status
-                    self._db_session.flush()
+                    # Use update function to also set heartbeat when transitioning to RUNNING
+                    update_sandbox_status__no_commit(
```
rohoswagger, line 514: ideally sandbox_manager.provision() should handle this

backend/onyx/server/features/build/sandbox/kubernetes/kubernetes_sandbox_manager.py
wenxi-onyx: _Defense in depth :big_grin:_

The big thing here is if, for whatever reason, pod service attempts to create a pod, it should never catastrophically fail if the pod already exists. IMO this should stay

backend/onyx/server/features/build/session/manager.py
```diff
@@ -494,8 +497,9 @@ def create_session__no_commit(
                     self._sandbox_manager.terminate(sandbox_id)
 
                     # Mark as terminated and re-provision
-                    sandbox.status = SandboxStatus.TERMINATED
-                    self._db_session.flush()
+                    update_sandbox_status__no_commit(
+                        self._db_session, sandbox_id, SandboxStatus.TERMINATED
```
rohoswagger, line 501: ideally sandbox_manager.terminate() should handle this

backend/onyx/server/features/build/session/manager.py
```diff
@@ -539,9 +545,10 @@ def create_session__no_commit(
                 llm_config=llm_config,
             )
 
-            # Update sandbox record with status from provisioning
-            sandbox.status = sandbox_info.status
-            self._db_session.flush()
+            # Update sandbox status (also refreshes heartbeat when transitioning to RUNNING)
+            update_sandbox_status__no_commit(
+                self._db_session, sandbox_id, sandbox_info.status
```
rohoswagger, line 550: ditto

backend/onyx/server/features/build/session/manager.py
wenxi-onyx: The sandbox manager doesn't have access to the DB --> i agree with you, but we have a very consistent pattern here. Idt the refactor to do this belongs in this PR

backend/onyx/server/features/build/api/sessions_api.py
rohoswagger: kk fair, can prolly tune this later if need be

backend/onyx/server/features/build/sandbox/kubernetes/kubernetes_sandbox_manager.py
rohoswagger: IMO this is a good case where we should fail loudly, cause something really is wrong

backend/onyx/server/features/build/session/manager.py
rohoswagger: k fair not worth it

rohoswagger (approved):

PR #8149, opened 2026-02-04
fix(craft): file upload

rohoswagger:
## Description

file upload works

+ corresponding AGENTS.md changes, we inject the ATTACHMENTS_SECTION dynamically on first file upload for a session

## How Has This Been Tested?

tested locally pretty comprehensively, will test on craft-dev

## Additional Options

---
## Summary by cubic
Fixes file uploads end-to-end by streaming tar data with an explicit byte count in the Kubernetes sandbox and centralizing client upload logic so attachments are reliable from the first message. This addresses the K8s Python client limitation where stdin cannot be closed without dropping the WebSocket.

- **Bug Fixes**
  - Read exactly tar_size bytes with head -c before extracting and avoid closing stdin to keep the WebSocket open.
  - Centralize web uploads in a single context: keep the session in sync, auto-upload pending files when ready, fetch existing attachments on session change, and validate size/count/types.
  - Inject the “Attachments (PRIORITY)” section into AGENTS.md on first upload (K8s and local) and remove the template placeholder, so agents always see uploaded files as high-priority context.
  - Encode path segments on DELETE to handle spaces/special characters.
  - Prevent deleted attachments from reappearing by skipping refetch during active deletions, deduping by path when merging, and rolling back the UI if backend deletion fails.

<sup>Written for commit 63b12a2e2d4040bd051e3ad03b17fdda09a926ee. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/app/craft/contexts/UploadFilesContext.tsx
+++ b/web/src/app/craft/contexts/UploadFilesContext.tsx
@@ -15,6 +15,7 @@ import {
   deleteFile as deleteFileApi,
   fetchDirectoryListing,
 } from "@/app/craft/services/apiServices";
+import { useBuildSessionStore } from "@/app/craft/hooks/useBuildSessionStore";
 
 /**
  * Upload File Status - tracks the state of files being uploaded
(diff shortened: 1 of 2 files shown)


rohoswagger:
@greptile

rohoswagger:
@cubic-dev

rohoswagger:
@greptile

rohoswagger:
@cubic-dev

PR #8157, opened 2026-02-04
fix(imap): improve robustness for malformed and legacy emails

yvesgrolet:
## Problem

The IMAP connector crashes when processing emails that have:

1. **Unknown encodings** (e.g., `unknown-8bit`) - Python raises `LookupError` when trying to decode headers with unrecognized charset names
2. **Commas in display names** (e.g., `"Doe, John" <john@example.com>`) - Address parsing breaks on RFC 2822 compliant addresses  
3. **Missing From header** - Some spam/automated emails lack a sender, causing Pydantic validation failures
4. **Missing Message-ID or Date** - Old or malformed emails may lack these required fields

These issues prevent the connector from indexing entire mailboxes when even a single problematic email is encountered.

## Solution

1. **Unknown encodings** - Try specified encoding first, fall back to `latin-1` on `LookupError`
2. **Address parsing** - Use `email.utils.getaddresses()` and graceful fallbacks instead of raising exceptions
3. **Missing required fields** - Generate UUID for missing Message-ID, use `datetime.now()` for missing Date, use "Unknown Sender" for missing From

## Testing

Tested on a production mailbox with ~76,000 emails spanning 20+ years, including many malformed/legacy messages. Before the fix: connector would crash repeatedly. After: successfully indexed 20,000+ emails and counting.

## Changes

- `backend/onyx/connectors/imap/models.py`: Added fallbacks for unknown encodings and missing required fields
- `backend/onyx/connectors/imap/connector.py`: Improved address parsing with graceful error handling

---
## Summary by cubic
Improves the IMAP connector to handle malformed and legacy emails without crashing during indexing. Adds resilient parsing, UTC-safe fallbacks, and deterministic IDs when headers are missing.

- **Bug Fixes**
  - Decode headers using declared charset; fall back to latin-1 for unknown encodings.
  - Parse addresses with email.utils.getaddresses; _parse_singular_addr adds safe fallbacks for quoted names, commas, and malformed values.
  - Fill missing headers: generate deterministic Message-ID (uuid5 over subject/date/from/to plus a body hash; use uuid4 if everything is empty); use datetime.now(timezone.utc) for missing/invalid Date; default From to “Unknown Sender” and Subject to “Unknown Subject”.

<sup>Written for commit 2df3150e3b4a65fa8f6e434dedd93f5b34ac319b. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/connectors/imap/connector.py
+++ b/backend/onyx/connectors/imap/connector.py
@@ -7,6 +7,7 @@ from datetime import datetime
 from datetime import timezone
 from email.message import Message
 from email.utils import parseaddr
+from email.utils import getaddresses
 from enum import Enum
 from typing import Any
 from typing import cast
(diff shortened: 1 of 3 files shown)


yvesgrolet:
Thanks for the review! I've addressed all the feedback:

✅ **Timezone consistency**: Changed `datetime.now()` to `datetime.now(timezone.utc)` for proper timezone-aware datetime
✅ **Fixed import**: Corrected `load_all_docs_from_checkpoint_connector` → `load_all_from_connector` and added `.documents` accessor
✅ **Deprecated method**: Replaced all `logger.warn()` calls with `logger.warning()`
✅ **UUID collision risk**: Improved fallback ID generation to include sender field, and use random `uuid4()` when all fields are empty

wenxi-onyx:
@yvesgrolet Thanks for the contribution! Would you mind adding a few tests to `backend/tests/daily/connectors/imap/test_imap_connector.py` that cover the issues you've resolved? Thanks again!

yvesgrolet:
@cubic-dev-ai I've addressed both issues from your latest review:

✅ **Timezone-aware tests** (test_imap_edge_cases.py:314): Replaced `datetime.now()` with `datetime.now(timezone.utc)` and removed the `.replace(tzinfo=None)` workaround. All datetime comparisons are now consistently timezone-aware (UTC).

✅ **UUID fallback reachability** (models.py:112): Added `raw_subject` and `raw_from` variables to store the original values *before* applying default placeholders. The emptiness check now uses these raw values, so `uuid4()` is correctly reached when all original fields are empty/None.

Commit: affe08b4e

yvesgrolet:
@wenxi-onyx I added tests in `backend/tests/unit/connectors/imap/test_imap_edge_cases.py` rather than in `tests/daily/` — these are pure unit tests that mock email messages and don't require a live IMAP server, so `tests/unit/` felt like a better fit.

The test file covers all four edge cases:
- Unknown encoding (`unknown-8bit`) → latin-1 fallback
- Commas in display names (`"Doe, John" <john@example.com>`) → correct RFC 2822 parsing
- Missing `From:` header → "Unknown Sender" fallback
- Missing `Message-ID` / `Date` headers → generated UUID / `datetime.now(timezone.utc)`

I also rebased the branch on the latest `main`. Let me know if you'd prefer the tests moved to `tests/daily/` instead!

PR #8160, opened 2026-02-04
fix(github): Fixing the Release Tags

justin-tahara:
## Description

Trying to fix the issue of the the duplicate tags being created. I suspect that the different desktop ships are competing and getting to the push their version thus creating multiple duplicate releases at times. 

The Beta tag is also always looking for `beta.X` so updating to handle just `beta` tag for some of the new tags we've been cutting. 

## How Has This Been Tested?

Running in CI

## Additional Options

---
## Summary by cubic
Fix CI release tagging to stop duplicate releases and correctly detect beta tags. Releases now use the pushed tag name, and builds run one at a time to avoid races.

- **Bug Fixes**
  - Expand beta tag detection to match vX.Y.Z-beta and vX.Y.Z-beta.N.
  - Set max-parallel: 1 in the matrix to prevent concurrent runners from publishing the same tag.
  - Resolve the release tag from github.ref_name and pre-create a draft release; dev runs use v0.0.0-dev+<sha>, and assets are uploaded to that release.

<sup>Written for commit ccecbfe1d3cbfb7d94ebd7e832b9062a3b95bb84. Summary will update on new commits.</sup>

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -223,8 +223,8 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           RELEASE_TAG: ${{ steps.resolve-tag.outputs.release-tag }}
         run: |
-          if ! gh release view "${RELEASE_TAG}" >/dev/null 2>&1; then
-            gh release create "${RELEASE_TAG}" \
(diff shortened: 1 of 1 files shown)


justin-tahara:
@greptile can you review the PR one more time?

justin-tahara:

PR #8168, opened 2026-02-04
fix: Anthropic DR requires setting reasoning limit if we want to set output limit

yuhongsun96:
## Description
Anthropic inference wants to set a reasoning budget at least as high as the output limit. Weird but ok...
This just guarantees it on our end.

## How Has This Been Tested?
DR works with Anthropic locally.
Unit tests added

## Additional Options

---
## Summary by cubic
Ensures Anthropic Claude reasoning requests set an explicit thinking budget and bumps max_tokens to avoid API errors. Defaults to AUTO and applies provider-specific mapping.

- **Bug Fixes**
  - For Claude with reasoning, set thinking.budget_tokens (low=1024, medium=2048, high=4096) and enforce max_tokens >= budget_tokens + 1.
  - Default reasoning_effort to AUTO across interfaces; map AUTO to medium budget for Claude and to “auto” for OpenAI, and use low for other providers. Align DR loop tool call max_tokens to 1024 and remove env-based DEFAULT_REASONING_EFFORT.

<sup>Written for commit bf239191ed850942633bb968b32626c21eff303f. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/app/admin/configuration/llm/forms/CustomForm.tsx
+++ b/web/src/app/admin/configuration/llm/forms/CustomForm.tsx
@@ -131,10 +131,15 @@ export function CustomForm({
                   return;
                 }
 
+                const selectedModelNames = modelConfigurations.map(
+                  (config) => config.name
+                );
+
                 await submitLLMProvider({
                   providerName: values.provider,
                   values: {
                     ...values,
+                    selected_model_names: selectedModelNames,
                     custom_config: customConfigProcessing(
                       values.custom_config_list
                     ),


backend/onyx/llm/multi_llm.py
```diff
@@ -315,10 +319,37 @@ def _completion(
                         "effort": OPENAI_REASONING_EFFORT[reasoning_effort],
                         "summary": "auto",
                     }
+
+            if is_claude_model:
+                budget_tokens: int | None = None
+                if reasoning_effort is None:
```
evan-onyx: put in dict pls

evan-onyx: also kinda weird for reasoning effort to be None or an enum, it should probably just be always in the enum yah?

PR #8172, opened 2026-02-04
fix(ci): Cleaning up Release Tags

justin-tahara:
## Description

Updating the workflow to create the release before and then allow the desktop ships to point to the created release. 

## Additional Options

---
## Summary by cubic
Create the GitHub release before desktop builds and have all matrix jobs upload assets to that single release. This fixes duplicate tags/releases and race conditions in desktop publishing.

- **Bug Fixes**
  - Added a create-release job that determines the tag (dev: v0.0.0-dev+SHORT_SHA; prod: ref name) and creates a draft release, exposing release-id.
  - Made build-desktop depend on create-release and configured tauri-action to use releaseId for asset uploads.

<sup>Written for commit 798777ff0c28e151a1984d21e6dd121db80e0a7f. Summary will update on new commits.</sup>

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -182,8 +182,52 @@ jobs:
           title: "🚨 Version Tag Check Failed"
           ref-name: ${{ github.ref_name }}
 
-  build-desktop:
+  # Create GitHub release first, before desktop builds start.
+  # This ensures all desktop matrix jobs upload to the same release instead of
+  # racing to create duplicate releases.
+  create-release:
     needs: determine-builds
+    if: needs.determine-builds.outputs.build-desktop == 'true'
+    runs-on: ubuntu-slim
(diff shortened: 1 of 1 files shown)


jmelahman (approved):
I would double-check we can test this. I think I tried testing the other PR(?) and it didn't seem to work, https://github.com/onyx-dot-app/onyx/releases/tag/untagged-1f8909a2915f80a743f4. Non-prod artifacts may have also broken elsewhere (could run a  build off main to get a reference)

I looked at the tauri-action github repo and seemed like they don't support appending to existing releases (there was an open PR which looked like it added that feature), but if it works, fine with me.

It's also weird to me that we're getting partial platform uploads. I.e. the Linux build uploads a `.rpm` and a `.deb` and they're in separate 2.11.3-beta* releases (rather than the Linux build being together and separate from say the Windows artifacts).

PR #8173, opened 2026-02-04
feat(opal): extract Hoverable into Interactive atom

raunakab:
## Description

- Moves `Hoverable` from `refresh-components` into the Opal design system as `Interactive`, a compound component (`Interactive.Base`, `Interactive.Container`, `Interactive.ChevronContainer`) that serves as a base atom for building higher-level components
- Adds `@opal/atoms`, `@opal/utils`, and `WithoutStyles` type to `@opal/types`; adds an opal-local `tsconfig.json` for IDE path resolution within the excluded `lib/opal` directory
- Renames props to design-system conventions (`nonInteractive` → `disableHover`, `reducedRounding` → `roundingVariant`, `noPadding` → `paddingVariant`), removes unused `asChild`/`href` codepaths, and updates all 4 consumer files

## How Has This Been Tested?

No UI changes. Current usages of `Hoverable` all continue to remain the same.

### Usage in the `FileCard`:

https://github.com/user-attachments/assets/008815ec-7343-48c6-afd8-d58d1416ae9d

### Usage in the `AgentCard`:

https://github.com/user-attachments/assets/605a3916-7e76-452b-99c5-d044deeb9875

## Additional Options

### DIFF
--- a/web/lib/opal/src/core/interactive/components.tsx
+++ b/web/lib/opal/src/core/interactive/components.tsx
@@ -8,10 +8,25 @@ import type { WithoutStyles } from "@opal/types";
 // Types
 // ---------------------------------------------------------------------------
 
-/** Controls background color styling on the interactive element. */
+/**
+ * Background color variant for interactive elements.
+ *
+ * Each variant defines a base background tint and corresponding hover/active
+ * state colors, providing visual hierarchy across the UI:
+ *
+ * - `"primary"` — Lightest tint (`tint-00`), used for primary surfaces
+ * - `"secondary"` — Medium tint (`tint-01`), used for secondary/nested surfaces
+ * - `"tertiary"` — Darker tint (`tint-02`), used for tertiary/emphasized surfaces
+ */
 export type InteractiveBaseVariant = "primary" | "secondary" | "tertiary";
 
-/** Height presets for `Interactive.Container`. */
+/**
+ * Height presets for `Interactive.Container`.
+ *
+ * - `"standard"` — Default height of 2.25rem (36px), suitable for most buttons/items
+ * - `"compact"` — Reduced height of 1.75rem (28px), for denser UIs or inline elements
+ * - `"full"` — Expands to fill parent height (`h-full`), for flexible layouts
+ */
 export type InteractiveContainerHeightVariant =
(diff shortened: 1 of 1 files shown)


web/lib/opal/src/core/interactive/styles.css
nmgarza5, line 1: would prefer we structure as such:
```
atoms/
   | -- interactive/
                 | -- index.ts
                 | -- interfaces.ts
                 | -- interactive.css
                 | -- interactive.tsx
```

we also have a `types.ts` which imo should re-export all types from opal so you only ever import from `@opal/src/types`

web/lib/opal/package.json
```diff
@@ -2,13 +2,21 @@
   "name": "@onyx/opal",
   "version": "0.0.1",
   "exports": {
+    "./atoms": {
```
nmgarza5: idk how i feel about the name atoms. why not just components?

nmgarza5 (approved):
approving because all my comments are subjective. if you like this setup we can move forward. also happy to chat live if you want

web/lib/opal/package.json
raunakab: So ideally, these components should not really be used by end-users too much. I mean occasionally, yes, but they should aim to reach for other components first.

I renamed it to "core" to signify that these are the "core" parts of the library. Similar to how Rust has the "core" lib (very low level functions and compiler intrinsics) but also has the "std" lib (the actual standard library containing all the useful interfaces).

raunakab: The `Interactive` is aiming to solve how fragmented our interactive components (`Button`s, `FilterButton`s, `IconButton`s are currently). Each one, right now, has their own stylings for "primary", "secondary", and "tertiary" hover and active states. A lot of copy-paste.

Therefore, the `Interactive` components would be the primal layer of interactivity: the "base glue" which defines what our design language dictates for different hover, active, disabled, etc. states mean.

web/lib/opal/src/core/interactive/styles.css
raunakab: Updated!

raunakab: I like this style more too!

PR #8177, opened 2026-02-05
chore: Sanitize LLM tool call args

yuhongsun96:
## Description
This is just covering hallucination cases from the LLM. It cleans up the bad chars and also prevents failed writes to the DB as the DB does not allow these null chars in strings. Typically only encountered in long complex flows like Deep Research.

## How Has This Been Tested?
Unit test included

## Additional Options

---
## Summary by cubic
Sanitizes LLM tool call arguments to remove invalid characters, keeping JSON parsing stable and avoiding PostgreSQL write errors in long flows like Deep Research.

- **Bug Fixes**
  - Added _sanitize_llm_output to strip NULL bytes and UTF-16 surrogate code points.
  - Applied sanitization to dict string values and raw string args before JSON parsing.

<sup>Written for commit a8c6badfdaab2efa0019440dbcc4ceca28749ffe. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/unit/onyx/chat/test_llm_step.py
+++ b/backend/tests/unit/onyx/chat/test_llm_step.py
@@ -1,0 +1,139 @@
+"""Tests for llm_step.py, specifically sanitization and argument parsing."""
+
+from onyx.chat.llm_step import _parse_tool_args_to_dict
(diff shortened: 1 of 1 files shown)


Danelegend (approved):
lgtm

PR #8179, opened 2026-02-05
chore: Slight tweaks of DR

yuhongsun96:
## Description
Change agent cycle count incrementing every web search to every cycle. Reason behind this:

When there is a long set of similar tool calls that repeat many times, the LLM can catch onto that pattern and will end up attempting to call tools in its history even when those tools are not present anymore in the tool config json. This happens especially with worse models but even with Claude 4.5 Opus it happens reliably across 100 deep research questions. This fix addresses it by having a different way of tracking cycles such that the last cycle isn't just removing web search, but it automatically goes into the report generation.

## How Has This Been Tested?
Ran locally many times

## Additional Options

---
## Summary by cubic
Switch research cycle counting from per-search to per-cycle for clearer control, and raise the cycle cap to 8 with automatic report generation on the final cycle.

- **Refactors**
  - Cycle count now increments once per agent cycle, not per web search.
  - Increased RESEARCH_CYCLE_CAP from 3 to 8.
  - On the last cycle, auto-generate an intermediate report instead of continuing tool calls.
  - Updated prompts to reflect the new cycle logic and clarify Think tool cadence; improved final report instruction wording.

<sup>Written for commit d90ddd26977a395e2b3a62e23116e875f4c02eb4. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/prompts/deep_research/orchestration_layer.py
+++ b/backend/onyx/prompts/deep_research/orchestration_layer.py
@@ -139,7 +139,7 @@ Provide inline citations in the format [1], [2], [3], etc. based on the citation
 
 
 USER_FINAL_REPORT_QUERY = f"""
-Provide a comprehensive answer to my previous query. CRITICAL: be as detailed as possible, stay on topic, and provide clear organization in your response.
+Provide a comprehensive, detailed, and insightful answer to my previous query. CRITICAL: be as detailed as possible, stay on topic, and provide clear organization in your response, addressing all aspects of the query.
 
 Ignore the format styles of the intermediate {RESEARCH_AGENT_TOOL_NAME} reports, those are not end user facing and different from your task.
 
--- a/backend/onyx/prompts/deep_research/research_agent.py
+++ b/backend/onyx/prompts/deep_research/research_agent.py
@@ -14,13 +14,12 @@ NEVER output normal response tokens, you must only call tools.
 For context, the date is {{current_datetime}}.
 
 # Tools
-You have a limited number of cycles of searches to complete your research but you do not have to use all cycles. \
-Each set of web searches increments the cycle by 1 (only web searches increment the cycle count). You are on cycle {{current_cycle_count}} of 3.\
+You have a limited number of cycles to complete your research and you do not have to use all cycles. You are on cycle {{current_cycle_count}} of {MAX_RESEARCH_CYCLES}.\
 {{optional_internal_search_tool_description}}\
 {{optional_web_search_tool_description}}\
 {{optional_open_url_tool_description}}
 ## {THINK_TOOL_NAME}
-CRITICAL - use the think tool after every set of searches and reads. \
+CRITICAL - use the think tool after every set of searches and reads (so search, read some pages, then think and repeat). \
 You MUST use the {THINK_TOOL_NAME} before calling the web_search tool for all calls to web_search except for the first call. \
 Use the {THINK_TOOL_NAME} before calling the {GENERATE_REPORT_TOOL_NAME} tool.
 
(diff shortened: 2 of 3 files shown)


backend/onyx/prompts/deep_research/orchestration_layer.py
```diff
@@ -139,7 +139,7 @@
 
 
 USER_FINAL_REPORT_QUERY = f"""
-Provide a comprehensive answer to my previous query. CRITICAL: be as detailed as possible, stay on topic, and provide clear organization in your response.
+Provide a comprehensive, detailed, and insightful answer to my previous query. CRITICAL: be as detailed as possible, stay on topic, and provide clear organization in your response, addressing all aspects of the query.
```
justin-tahara, line 142: nit: but the additional comma after response is not necessary imo

backend/onyx/prompts/deep_research/research_agent.py
```diff
@@ -14,13 +14,12 @@
...
 # Tools
-You have a limited number of cycles of searches to complete your research but you do not have to use all cycles. \
-Each set of web searches increments the cycle by 1 (only web searches increment the cycle count). You are on cycle {{current_cycle_count}} of 3.\
+You have a limited number of cycles to complete your research and you do not have to use all cycles. You are on cycle {{current_cycle_count}} of {MAX_RESEARCH_CYCLES}.\
 {{optional_internal_search_tool_description}}\
 {{optional_web_search_tool_description}}\
 {{optional_open_url_tool_description}}
 ## {THINK_TOOL_NAME}
-CRITICAL - use the think tool after every set of searches and reads. \
+CRITICAL - use the think tool after every set of searches and reads (so search, read some pages, then think and repeat). \
```
justin-tahara, line 22: Any reason we want it to think about pages? Is this in context of google results pages or like within a specific page to read through multiple links?

backend/onyx/tools/fake_tools/research_agent.py
```diff
@@ -73,7 +73,7 @@
 logger = setup_logger()
 
 
-RESEARCH_CYCLE_CAP = 3
+RESEARCH_CYCLE_CAP = 8
```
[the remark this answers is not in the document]
justin-tahara: I agree that this isn't going to break things but why not keep them consistent if we bumped the value here? I feel like if anyone sees this in the future they will just be confused.

backend/onyx/tools/fake_tools/research_agent.py
```diff
@@ -264,12 +264,9 @@ def run_research_agent_call(
                     break
 
                 if research_cycle_count == RESEARCH_CYCLE_CAP:
-                    # For the last cycle, do not use any more searches, only reason or generate a report
-                    current_tools = [
-                        tool
-                        for tool in tools
-                        if tool.name not in {SearchTool.NAME, WebSearchTool.NAME}
-                    ]
+                    # Auto-generate report on last cycle
```
justin-tahara, line 267: I feel like we should explain within this comment that the break takes us to the `generate_intermediate_report` function which is why we break because we have in theory hit 8 cycles but because of the comments from above, that won't actually happen so this will still be hit after 3 cycles

backend/onyx/tools/fake_tools/research_agent.py
```diff
@@ -593,6 +583,7 @@ def run_research_agent_call(
                 # If it reached this point, it did not call reasoning, so here we wipe it to not save it to multiple turns
                 most_recent_reasoning = None
                 llm_cycle_count += 1
+                research_cycle_count += 1
```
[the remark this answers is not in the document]
justin-tahara: Maybe worth consolidating unless there's plans to use this variable in order ways in the future?

justin-tahara (commented):
Leaving some comments. I think we should probably just revert and have this reviewed by folks so more people can learn about the DR path

PR #8186, opened 2026-02-05
chore(chat): Cleaning Error Codes + Tests

justin-tahara:
## Description

Customer reported that there are issues with our endpoints not returning the correct error codes and thus adding some proper coverage + tests to cover the different workflows. 

The customer was reporting that a user who has a chat id of a chat session that was not shared with them would receive a 500 instead of a 403 so this is trying to correct the process and improve the workflow.

## How Has This Been Tested?

Running the tests that were added. 

## Additional Options

---
## Summary by cubic
Fixes chat session retrieval to return the right error codes: 403 for unauthorized access and 404 for not found or deleted sessions. Adds integration tests to cover private, shared, and deleted access flows and prevents 500s when users access sessions they shouldn’t.

- **Bug Fixes**
  - Map internal errors to explicit HTTP responses in get-chat-session.
  - Return 403 when session isn’t shared or user lacks access.
  - Return 404 for non-existent or deleted sessions (unless owner uses include_deleted).
  - Update test helper to expect 404 for missing/deleted sessions.

- **Tests**
  - Add integration tests for private, public-shared, deleted, and unknown chat session IDs.
  - Verify include_deleted works for owners and is blocked for non-owners.

<sup>Written for commit 620ab2c26e05f93ac56667b87f64ef3d27b0c1c3. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/integration/tests/chat/test_chat_session_access.py
+++ b/backend/tests/integration/tests/chat/test_chat_session_access.py
@@ -33,10 +33,11 @@ def second_user(admin_user: DATestUser) -> DATestUser:
         if response.status_code not in (400, 409):
             raise
         try:
-            detail = response.json().get("detail")
+            payload = response.json()
         except ValueError:
-            detail = None
-        if detail is None or "already exists" not in str(detail).lower():
+            raise
+        detail = payload.get("detail")
+        if not _is_user_already_exists_detail(detail):
             raise
         print("Second basic user already exists; logging in instead.")
         return UserManager.login_as_user(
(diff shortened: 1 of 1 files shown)


backend/onyx/server/query_and_chat/chat_backend.py
```diff
@@ -267,7 +268,32 @@ def get_chat_session(
             include_deleted=include_deleted,
         )
     except ValueError:
-        raise ValueError("Chat session does not exist or has been deleted")
+        try:
+            existing_chat_session = get_chat_session_by_id(
```
acaprau, line 275: i had to read this once or twice to understand what's going on, just left a quick comment with my understanding of how this works.
```suggestion
            # If we failed to get a chat session, try to retrieve the session with
            # less restrictive filters in order to identify what exactly mismatched
            # so we can bubble up an accurate code.
            existing_chat_session = get_chat_session_by_id(
```

backend/tests/integration/tests/chat/test_chat_session_access.py
```diff
@@ -0,0 +1,155 @@
...
+    """Reset all data once before running any tests in this module."""
+    reset_all()
+
+
+@pytest.fixture
+def second_user(admin_user: DATestUser) -> DATestUser:
+    # Ensure admin exists so this new user is created with BASIC role.
+    try:
+        return UserManager.create(name="second_basic_user")
+    except Exception as e:
```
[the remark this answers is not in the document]
acaprau: idk how this fixture works but the comment seems valid

backend/onyx/server/query_and_chat/chat_backend.py
justin-tahara: Going to update the working a bit but same idea

PR #8193, opened 2026-02-05
fix: LLM Read Timeout

yuhongsun96:
## Description
Bump the read timeout for Deep Research. It keeps failing on OpenAI models but works reliably with Anthropic.
The time between tokens is surprisingly long at times but the hope is it's one off cases and can recover if a more permissive timeout is given.

## How Has This Been Tested?
Ran normal inference and deep research with gpt5. both work

## Additional Options

---
## Summary by cubic
Increase and standardize LLM socket read timeouts to prevent Deep Research streams from failing on long pauses, especially with OpenAI. Adds per-request overrides for long report generation and improves connection reliability.

- **Bug Fixes**
  - Introduced LLM_SOCKET_READ_TIMEOUT (default 60s) as the default socket read timeout for all LLM requests.
  - Added timeout_override plumbed through llm_step functions; set to 300s for Deep Research intermediate and final reports.
  - Removed the special reasoning-model timeout multiplier in favor of a single read-timeout policy.
  - Created an isolated litellm HTTPHandler per request and closed it after use to avoid connection pool conflicts.

- **Migration**
  - Replace QA_TIMEOUT with LLM_SOCKET_READ_TIMEOUT in environment variables, Docker Compose, and Helm values.
  - Restart services to apply the new setting.

<sup>Written for commit 6b546da0d09f01aef416dcab2fd187bda5b5b6aa. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/unit/onyx/llm/test_multi_llm.py
+++ b/backend/tests/unit/onyx/llm/test_multi_llm.py
@@ -249,7 +249,7 @@ def test_multiple_tool_calls(default_multi_llm: LitellmLLM) -> None:
             temperature=0.0,  # Default value from GEN_AI_TEMPERATURE
             timeout=30,
             max_tokens=None,
-            client=None,
+            client=ANY,  # HTTPHandler instance created per-request
             parallel_tool_calls=True,
             mock_response=MOCK_LLM_RESPONSE,
             allowed_openai_params=["tool_choice"],


backend/onyx/llm/multi_llm.py
```diff
@@ -82,10 +81,6 @@ def _prompt_to_dicts(prompt: LanguageModelInput) -> list[dict[str, Any]]:
     return [prompt.model_dump(exclude_none=True)]
 
 
-def _prompt_as_json(prompt: LanguageModelInput) -> JSON_ro:
```
yuhongsun96, line 85: dead code

backend/onyx/llm/multi_llm.py
```diff
@@ -107,12 +102,14 @@ def __init__(
...
     ):
+        # Timeout in seconds for each socket read operation (i.e., max time between
+        # receiving data chunks/tokens). This is NOT a total request timeout - a
+        # request can run indefinitely as long as data keeps arriving within this
+        # window. If the LLM pauses for longer than this timeout between chunks,
+        # a ReadTimeout is raised.
         self._timeout = timeout
         if timeout is None:
-            if model_is_reasoning_model(model_name, model_provider):
-                self._timeout = QA_TIMEOUT * 10  # Reasoning models are slow
```
yuhongsun96, line 113: We get reasoning tokens streamed back so it doesn't impact the read timeout anyway

backend/onyx/llm/multi_llm.py
```diff
@@ -433,33 +430,41 @@ def invoke(
...
-                timeout_override=timeout_override,
-                max_tokens=max_tokens,
-                parallel_tool_calls=True,
-                reasoning_effort=reasoning_effort,
-                user_identity=user_identity,
-            ),
-        )
+        # Create an isolated HTTP handler for this request to avoid
+        # connection pool conflicts when multiple threads call concurrently.
+        client = HTTPHandler()
```
yuhongsun96, line 440: This is more complete, should have also done it for the non-stream case before, adding it now

backend/onyx/deep_research/dr_loop.py
```diff
@@ -161,6 +161,7 @@ def generate_final_report(
             max_tokens=MAX_FINAL_REPORT_TOKENS,
             is_deep_research=True,
             pre_answer_processing_time=pre_answer_processing_time,
+            timeout_override=300,  # 5 minute read timeout for long report generation
```
Weves, line 164: magic number

backend/onyx/tools/fake_tools/research_agent.py
```diff
@@ -138,6 +138,7 @@ def generate_intermediate_report(
             max_tokens=MAX_INTERMEDIATE_REPORT_LENGTH_TOKENS,
             use_existing_tab_index=True,
             is_deep_research=True,
+            timeout_override=300,  # 5 minute read timeout for long report generation
```
Weves, line 141: magic number + duplicated

backend/onyx/deep_research/dr_loop.py
yuhongsun96: tru

backend/onyx/tools/fake_tools/research_agent.py
yuhongsun96: Doesn't really need to be the same value, inherently they're 2 separate flows

PR #8207, opened 2026-02-05
feat: basic langfuse tracing + tracing consolidation

yuhongsun96:
## Description
Centralize tracing and enable Langfuse

## How Has This Been Tested?
Works locally with the env vars set. Will note that I have decided not to modify the deployment related files for this. If a deployment wants this, they need to ensure the env vars are set.

Additional accepted risk: Deep Research nesting is not fully correct but it's still quite informative.

## Additional Options

---
## Summary by cubic
Centralized tracing into setup_tracing() and switched Langfuse to the native SDK. Tracing spans API, Celery workers, and the eval CLI; inputs/outputs are masked, LLM costs are recorded, clients flush/shutdown cleanly, multiple providers can run together, and legacy setup modules/OpenInference processor were removed.

- **Migration**
  - To enable Langfuse: set LANGFUSE_SECRET_KEY, LANGFUSE_PUBLIC_KEY, and LANGFUSE_HOST (for self-hosted).
  - Braintrust initializes if BRAINTRUST_API_KEY and BRAINTRUST_PROJECT are set.
  - Optional: control masking size via TRACING_MASKING_LENGTH.
  - No deployment file changes; ensure env vars are present.
  - Known caveat: if some packets drop, nested spans may appear unnested.

<sup>Written for commit 44855c31d601de8fe1bb24e3c645121e20a89d89. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/tracing/langfuse_tracing_processor.py
+++ b/backend/onyx/tracing/langfuse_tracing_processor.py
@@ -175,6 +175,8 @@ class LangfuseTracingProcessor(TracingProcessor):
         """
         try:
             data = span.span_data
+            # Declare as Any since different code paths return different observation types
+            langfuse_span: Any = None
 
             # Get Langfuse IDs and metadata under lock for thread-safe access
             with self._lock:
@@ -215,26 +217,26 @@ class LangfuseTracingProcessor(TracingProcessor):
             # Create spans using trace_context (thread-safe ID-based approach)
             # In Langfuse SDK v3, use start_observation with as_type parameter
             if isinstance(data, GenerationSpanData):
-                langfuse_span = client.start_observation(
+                langfuse_span = client.start_observation(  # type: ignore[call-overload]
+                    trace_context=trace_context,
                     name=self._get_generation_name(data),
                     as_type="generation",
-                    trace_context=trace_context,
+                    metadata=trace_metadata,
                     model=data.model,
                     model_parameters=self._get_model_parameters(data),
-                    metadata=trace_metadata,
                 )
             elif isinstance(data, FunctionSpanData):
                 langfuse_span = client.start_observation(
+                    trace_context=trace_context,
                     name=data.name,
                     as_type="tool",
-                    trace_context=trace_context,
                     metadata=trace_metadata,
                 )
             elif isinstance(data, AgentSpanData):
                 langfuse_span = client.start_observation(
+                    trace_context=trace_context,
                     name=data.name,
                     as_type="agent",
-                    trace_context=trace_context,
                     metadata={
                         **(trace_metadata or {}),
                         "tools": data.tools,
@@ -244,9 +246,9 @@ class LangfuseTracingProcessor(TracingProcessor):
                 )
             else:
                 langfuse_span = client.start_observation(
+                    trace_context=trace_context,
                     name=data.type if hasattr(data, "type") else "unknown",
                     as_type="span",
-                    trace_context=trace_context,
                     metadata=trace_metadata,
                 )
 


backend/tests/unit/onyx/tracing/test_tracing_setup.py
```diff
@@ -2,47 +2,141 @@
...
+from onyx.tracing import setup as tracing_setup
 
 
-def test_setup_langfuse_if_creds_available_with_creds() -> None:
-    """Test that setup_langfuse_if_creds_available executes without error when credentials are available."""
-    # Set credentials to non-empty values to avoid early return
+def test_setup_tracing_with_no_creds() -> None:
+    """Test that setup_tracing returns empty list when no credentials are configured."""
+    # Ensure no tracing credentials are set
+    os.environ.pop("BRAINTRUST_API_KEY", None)
```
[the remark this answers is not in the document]
yuhongsun96: Ignoring, requires larger rework, out of scope of this

justin-tahara:
@greptile can you review this pr one more time

backend/onyx/main.py
```diff
@@ -275,8 +274,7 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
         logger.notice("Both OAuth Client ID and Secret are configured.")
 
     # Initialize tracing if credentials are provided
-    setup_braintrust_if_creds_available()
-    setup_langfuse_if_creds_available()
+    setup_tracing()
```
justin-tahara, line 277: Does this mean that the end user can run both traces if they want to? Do we only run one of them if we have both env vars set?

backend/onyx/tracing/setup.py
```diff
@@ -0,0 +1,100 @@
...
+    Uses add_trace_processor() to ADD processors rather than replacing them,
+    allowing multiple providers to receive trace events simultaneously.
+
+    This function is idempotent - calling it multiple times will only
+    initialize providers once.
+    """
+    global _initialized
+    if _initialized:
+        logger.debug("Tracing already initialized, skipping")
+        return []
```
justin-tahara, line 28: What's the point in having a return type if we don't use it? 

Even in the successful path it seems like we don't need this list in the call sites. Could we potentially get rid of it?

backend/onyx/tracing/langfuse_tracing_processor.py
```diff
@@ -0,0 +1,402 @@
...
+                    model_name=data.model,
+                    prompt_tokens=int(prompt_tokens),
+                    completion_tokens=int(completion_tokens),
+                )
+                if cost_cents > 0:
+                    # Convert cents to dollars for Langfuse
+                    return cost_cents / 100.0
+        except Exception as e:
+            logger.debug(f"Failed to calculate cost: {e}")
+        return None
```
justin-tahara, line 109: Do we not have a function that already does this elsewhere? I feel like this type of code is probably duplicated many times so ideally we just reuse the code

backend/onyx/tracing/openinference_tracing_processor.py
```diff
@@ -1,453 +0,0 @@
...
-MESSAGE_TOOL_CALLS = MessageAttributes.MESSAGE_TOOL_CALLS
-MESSAGE_TOOL_CALL_ID = MessageAttributes.MESSAGE_TOOL_CALL_ID
-
-TOOL_CALL_FUNCTION_ARGUMENTS_JSON = ToolCallAttributes.TOOL_CALL_FUNCTION_ARGUMENTS_JSON
-TOOL_CALL_FUNCTION_NAME = ToolCallAttributes.TOOL_CALL_FUNCTION_NAME
-TOOL_CALL_ID = ToolCallAttributes.TOOL_CALL_ID
-
-TOOL_JSON_SCHEMA = ToolAttributes.TOOL_JSON_SCHEMA
-
-JSON = OpenInferenceMimeTypeValues.JSON.value
```
justin-tahara, line 453: Do we not need any of these?

backend/onyx/tracing/langfuse_tracing_processor.py
justin-tahara, line 1: I think there are some similar functions that were implemented in [here](https://github.com/onyx-dot-app/onyx/pull/7573/files#diff-4433982272d65479717aa558618b8f600f02eb2a4c47b1579db782d250304f69) but I think if we can add some simple missing ones like the span_type or time_elapsed or agent/function/generation log that would help out the customer. 

I think you should be actually able to port these over and help improve this file relatively easily

backend/tests/unit/onyx/tracing/test_tracing_setup.py
```diff
@@ -2,47 +2,141 @@
...
+        assert "braintrust" in result
+
+    # Clean up
+    os.environ.pop("BRAINTRUST_API_KEY", None)
+    os.environ.pop("BRAINTRUST_PROJECT", None)
+    tracing_setup._initialized = False
+    importlib.reload(app_configs)
+
+
+def test_setup_tracing_with_langfuse_creds() -> None:
```
justin-tahara, line 83: Do we not need the host here?

backend/tests/unit/onyx/tracing/test_tracing_setup.py
```diff
@@ -2,47 +2,141 @@
...
     os.environ.pop("LANGFUSE_SECRET_KEY", None)
     os.environ.pop("LANGFUSE_PUBLIC_KEY", None)
+    tracing_setup._initialized = False
     importlib.reload(app_configs)
 
 
-def test_setup_braintrust_if_creds_available_with_creds() -> None:
-    """Test that setup_braintrust_if_creds_available executes without error when credentials are available."""
-    # Set credentials to non-empty values to avoid early return
+def test_setup_tracing_with_both_providers() -> None:
```
justin-tahara, line 110: Missing host here too I believe

backend/tests/unit/onyx/tracing/test_tracing_setup.py
justin-tahara, line 1: This covers setup but we might want a proper tracing set of tests but not urgent

backend/tests/unit/onyx/tracing/test_tracing_setup.py
yuhongsun96: not needed, it defaults to empty string which I think is like localhost or something. Richard had it that way I assume for good reason

backend/tests/unit/onyx/tracing/test_tracing_setup.py
yuhongsun96: Mhm, later if that's ok

backend/onyx/main.py
yuhongsun96: Ya, they can have both, I validated that it works

backend/onyx/tracing/openinference_tracing_processor.py
yuhongsun96: Nothing from this file is used/imported. Otherwise Mypy would fail

backend/onyx/tracing/setup.py
yuhongsun96: Fair point, let me check but seems good

backend/onyx/tracing/langfuse_tracing_processor.py
yuhongsun96: I lean we put this out and see what they ask for.

backend/onyx/tracing/langfuse_tracing_processor.py
yuhongsun96: Summary: The _calculate_cost method itself isn't pure duplication - it wraps the shared calculate_llm_cost_cents function and adds Langfuse-specific conversion (cents → dollars).

backend/onyx/tracing/setup.py
yuhongsun96: Turns out this is to make testing easier, I'm ok with that reasoning

PR #8215, opened 2026-02-06
fix: LiteLLM for OpenAI compatible models not using Responses route

yuhongsun96:
## Description
Clearly documented in the code comments.

## How Has This Been Tested?
Verified it works with OpenAI models via OpenAI
Reverting logic that was problematic for non "is_true_openai_model" models

## Additional Options

---
## Summary by cubic
Use LiteLLM’s Responses route only for true OpenAI models and route OpenAI‑compatible providers through the standard SDK path. This fixes api_key AttributeErrors and reduces “Bad file descriptor” errors during concurrent streams.

- **Bug Fixes**
  - Create HTTPHandler only when is_true_openai_model is true (invoke and stream).
  - Route non-OpenAI models using OpenAI provider through default client path.
  - Pass timeout to HTTPHandler and close it only when created.

<sup>Written for commit 9ac1a3920b69e0428114e826524799de16237410. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/unit/onyx/llm/test_multi_llm.py
+++ b/backend/tests/unit/onyx/llm/test_multi_llm.py
@@ -713,3 +713,43 @@ def test_bedrock_model_passes_no_client() -> None:
         mock_completion.assert_called_once()
         kwargs = mock_completion.call_args.kwargs
         assert kwargs["client"] is None
+
+
+def test_azure_openai_model_uses_httphandler_client() -> None:
+    """Test that Azure OpenAI models get an HTTPHandler client passed.
+
+    Azure OpenAI uses the same responses API as OpenAI, so it needs
+    the same HTTPHandler isolation to avoid connection pool conflicts.
+    """
+    from litellm import HTTPHandler
+
+    llm = LitellmLLM(
+        api_key="test_key",
+        timeout=30,
+        model_provider=LlmProviderNames.AZURE,
+        model_name="gpt-4o",
+        api_base="https://my-resource.openai.azure.com",
+        api_version="2024-02-15-preview",
+        max_input_tokens=128000,
+    )
+
+    with patch("litellm.completion") as mock_completion:
+        mock_response = litellm.ModelResponse(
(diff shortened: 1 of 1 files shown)


backend/onyx/llm/multi_llm.py
```diff
@@ -435,9 +435,44 @@ def invoke(
...
+        #    - Must use is_true_openai_model() NOT just check model_provider == "openai"
+        #    - Many OpenAI-compatible providers set model_provider="openai" but are NOT true
+        #      OpenAI models (glm-4.7, DeepSeek, local proxies, etc.)
+        #    - is_true_openai_model() checks both provider AND model name patterns
+        #
+        # This note may not be entirely accurate as there is a lot of complexity in the LiteLLM codebase around this
+        # and not every model path was traced thoroughly. It is also possible that in future versions of LiteLLM
+        # they will realize that their OpenAI handling is not threadsafe. Hope they will just fix it.
+        client = None
+        if is_true_openai_model(self.config.model_provider, self.config.model_name):
```
justin-tahara, line 473: Is there potentially weird cases with Azure's hosted Open AI models? 

LiteLLM Proxy is also checked in this function which I don't think should have issues but Azure could definitely have some weird edge cases like we've seen in the past.

backend/onyx/llm/multi_llm.py
acaprau, line 568: pls confirm/double check what happens if we pass a None client into this function call. i see it is an acceptable type for this argument but we need to be sure what the behavior is like.

yuhongsun96: It defaults under the hood to a None anyway

Image

backend/onyx/llm/multi_llm.py
yuhongsun96: I think for these cases if we use the OpenAI client it should be fine. Let me verify real quick with an Azure model

yuhongsun96: verified works for Azure

PR #8217, opened 2026-02-06
chore(migrations): database migration runner

jmelahman:
## Description

Fully enclosed script to run database migrations.

Notable functional changes include:
- running alembic migrations only on tenants who are currently not at the head revision
- continue running migrations in batches (default 50 schemas per batch with 6 workers). Now, however, if the batch fails, the batch is attempted again with `-x continue=true` & has stderr streamed to the console to ensure all tenants are attempted to be migrated + logs from any failed migrations
- updates the logging to include:
  - when a new batch is started
  - the duration of each finished batch
  - a 60s monitor that warns if a batch has been running for >60s (on small migrations in local testing, a batch takes ~20s)
  - the only logging from alembic is now when a batch fails -- happy path upgrades will have no alembic logging

## How Has This Been Tested?

`uv run pytest backend/tests/integration/multitenant_tests/migrations/test_run_multitenant_migrations.py`

(confirmed the test is running in [CI](https://github.com/onyx-dot-app/onyx/actions/runs/21741067692/job/62716635532?pr=8217#step:8:338))

#### Local end-to-end testing
```
# Get a fresh, multitenant backend running:
ods db drop --yes && \
  ods compose multitenant --down && \
  ods compose multitenant
```

```
# Restore a database dump with 100 empty tenants
ods db restore ./100tenants.sql
```

For example, a dump of 100 empty tenants: [100tenants.sql](https://github.com/user-attachments/files/25120778/100tenants.sql)

```
MULTI_TENANT=true ./alembic/run_multitenant_migrations.py -b 10 -j 4
```

**stuck warning log**
20260205_22h17m18s_grim

**happy path log updates**
20260205_22h16m47s_grim

## Additional Options

---
## Summary by cubic
Adds a parallel, batched Alembic migration runner to speed up tenant schema upgrades. It discovers tenant schemas, skips those already at head, runs upgrades concurrently, and exits non-zero if any fail or if no tenants are found.

- **New Features**
  - Runs alembic upgrade head in batched subprocesses via -x schemas=a,b with configurable workers (-j, default 6) and batch size (-b, default 50).
  - Discovers tenants via get_all_tenant_ids filtered by TENANT_ID_PREFIX; migrates only schemas not at head (or missing alembic_version).
  - Retries failed batches with -x continue=true so remaining schemas still upgrade; prints batch progress/durations with periodic “still running” status.
  - Uses SqlEngine.scoped_engine to init and clean up the DB engine.

- **Migration**
  - Run: python backend/alembic/run_multitenant_migrations.py -j 8 -b 100
  - Alembic env must accept -x schemas=a,b,c (and optional -x continue=true) to target multiple schemas per process.

<sup>Written for commit 7b35ed2db0beaa30a54ea315747232c9e323536b. Summary will update on new commits.</sup>

### DIFF
--- a/backend/alembic/run_multitenant_migrations.py
+++ b/backend/alembic/run_multitenant_migrations.py
@@ -80,13 +80,13 @@ def run_alembic_for_batch(schemas: list[str]) -> BatchResult:
 
     # At least one schema failed.  Print the initial error output, then
     # re-run with continue=true so the remaining schemas still get migrated.
+    if result.stdout:
+        print(f"Initial error output:\n{result.stdout}", file=sys.stderr, flush=True)
     print(
-        f"Batch failed (exit {result.returncode}), retrying with continue=true ...",
+        f"Batch failed (exit {result.returncode}), retrying with 'continue=true'...",
         file=sys.stderr,
         flush=True,
     )
-    if result.stdout:
-        print(result.stdout, file=sys.stderr, end="", flush=True)
 
     retry = subprocess.run(
         [*base_cmd, "-x", "continue=true", "upgrade", "head"],


jmelahman:
@greptile review please

jmelahman:
@greptile review please

jmelahman:
@greptile review please

jmelahman:
@greptile review please

jmelahman:
@cubic-dev-ai please review

backend/alembic/run_multitenant_migrations.py
```diff
@@ -0,0 +1,332 @@
...
+    # remaining schemas still get migrated.  The output will contain
+    # clear error messages for the problematic schema(s).
+    retry = subprocess.run(
+        [*base_cmd, "-x", "continue=true", "upgrade", "head"],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+        text=True,
+    )
+    elapsed = time.monotonic() - start
+    return BatchResult(schemas, False, retry.stdout or "", elapsed)
```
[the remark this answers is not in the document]
jmelahman: Can't because `-x continue=true` always returns `0`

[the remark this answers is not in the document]
jmelahman: https://github.com/onyx-dot-app/onyx/pull/8217#discussion_r2772629027

backend/alembic/run_multitenant_migrations.py
```diff
@@ -0,0 +1,343 @@
...
+
+    # At least one schema failed.  Print the initial error output, then
+    # re-run with continue=true so the remaining schemas still get migrated.
+    if result.stdout:
+        print(f"Initial error output:\n{result.stdout}", file=sys.stderr, flush=True)
+    print(
+        f"Batch failed (exit {result.returncode}), retrying with 'continue=true'...",
+        file=sys.stderr,
+        flush=True,
+    )
```
justin-tahara, line 89: I'm curious what this print looks like. 

I wonder if we can just create a temp table to mark tenants that were not upgraded properly or something

backend/alembic/run_multitenant_migrations.py
```diff
@@ -0,0 +1,343 @@
...
+    elapsed = time.monotonic() - start
+    return BatchResult(schemas, False, retry.stdout or "", elapsed)
+
+
+def get_head_revision() -> str | None:
+    """Get the head revision from the alembic script directory."""
+    alembic_cfg = Config("alembic.ini")
+    script = ScriptDirectory.from_config(alembic_cfg)
+    return script.get_current_head()
```
justin-tahara, line 105: This is per schema right?

backend/alembic/run_multitenant_migrations.py
```diff
@@ -0,0 +1,343 @@
...
+
+
+def get_schemas_needing_migration(
+    tenant_schemas: List[str], head_rev: str
+) -> List[str]:
```
justin-tahara, line 110: Will this run on the pre provisioned tenants as well? Like the seeded tenants?

backend/alembic/run_multitenant_migrations.py
```diff
@@ -0,0 +1,343 @@
...
+        type=int,
+        default=50,
+        metavar="N",
+        help="Schemas per alembic process (default: 50)",
+    )
+    args = parser.parse_args()
+    if args.jobs < 1:
+        parser.error("--jobs must be >= 1")
+    if args.batch_size < 1:
+        parser.error("--batch-size must be >= 1")
```
justin-tahara, line 295: Minor nit: May want an upper bound as well

justin-tahara (commented):
Overall I think we are going to be in a better spot. Just have some minor comments and then we can definitely get this in

backend/alembic/run_multitenant_migrations.py
jmelahman: Once per migration. This is the head migration they're all being migrated to.

backend/alembic/run_multitenant_migrations.py
jmelahman: I'm not sure what a seeded tenant is, but it will run on all tenants that are not at the head revision which likely includes these tenants.

backend/alembic/run_multitenant_migrations.py
jmelahman: It's a bit confusing since it includes the entire alembic output from start to error (it's also too large to get in a single screenshot), but it's surrounded by `Initial error output:` at the top and `Batch failed (exit {result.returncode}), retrying with 'continue=true'...` at the bottom + stands out reasonably well to other logs.

I think this is ideal since we want to know these errors. Also, it's hard to know which tenant specifically is failing since they're batched and I don't want to have to parse the alembic output.

backend/alembic/run_multitenant_migrations.py
justin-tahara: Okay I misunderstood

backend/alembic/run_multitenant_migrations.py
justin-tahara: Okay I think it should be fine but my worry is that every cloud deploy we will have like 1 or 2 tenants that need to be upgraded

PR #8219, opened 2026-02-06
fix(db): null out document set and persona ownership on user deletion

nmgarza5:
## Description

Fixes a `ForeignKeyViolation` error when deleting a user who owns document sets. The `document_set.user_id` FK constraint blocks user deletion because the DB constraint lacks the `ON DELETE CASCADE` that the ORM model declares.

Instead of cascade-deleting (which would destroy document sets used by other users), this PR:
- Nulls out `document_set.user_id` and `persona.user_id` for owned resources before deleting the user, preserving them for other users
- Treats orphaned document sets (`user_id=None`) as curator-manageable so curators can still edit/delete them after the original owner is removed

## Additional Options

---
## Summary by cubic
Fixes user deletion failing when the user owns document sets by clearing ownership instead of deleting data. Orphaned document sets remain editable by curators.

- **Bug Fixes**
  - Set DocumentSet.user_id and Persona.user_id to None before deleting the user to avoid FK violations and preserve shared data.
  - Update document set permission checks to treat ownerless sets as owned by the requester (enables curator edit/delete).
  - Add unit tests for ownership nulling, join-table cleanup, OAuth deletion, and commit behavior.

<sup>Written for commit 34eba6972f90df45eb5c21dc75d85de97087a744. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/db/users.py
+++ b/backend/onyx/db/users.py
@@ -17,7 +17,9 @@ from onyx.auth.schemas import UserRole
 from onyx.configs.constants import ANONYMOUS_USER_EMAIL
 from onyx.configs.constants import NO_AUTH_PLACEHOLDER_USER_EMAIL
 from onyx.db.api_key import DANSWER_API_KEY_DUMMY_EMAIL_DOMAIN
+from onyx.db.models import DocumentSet
 from onyx.db.models import DocumentSet__User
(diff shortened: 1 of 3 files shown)


yuhongsun96 (approved):
lgtm

PR #8220, opened 2026-02-06
fix(ui): Inconsistent LLM Provider Logo

justin-tahara:
## Description

There is a bug in which the logo for the LLM provider can potentially be swapped out if the name is the same between the different providers. 

For example: Azure GPT-5 model -> OpenAI GPT-5 model
The UI would jump from the Azure logo to the OpenAI logo when you started a new chat. 

This PR aims to fix this issue.

## How Has This Been Tested?

Reproduced locally and validated that the fix is working.

## Additional Options

---
## Summary by cubic
Fixes provider logo switching when models share the same name across providers. We now match by provider + model first, falling back to name-only if provider info is missing.

- **Bug Fixes**
  - Prioritize provider-scoped lookup in useLlmManager to ensure the correct logo (e.g., Azure vs OpenAI).
  - Add fallback to search by raw model name when model parsing fails.
  - Temperature and state fixes: remove unintended llmUpdate during init; cap Anthropic temps using session model when currentLlm isn’t ready; fix useEffect deps in source preferences to avoid stale state.

<sup>Written for commit 6643a0468b4bf3cc7952aed0f5f931e2c20f4ba0. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/lib/hooks.ts
+++ b/web/src/lib/hooks.ts
@@ -810,13 +810,12 @@ export function useLlmManager(
   ]);
 
   const updateTemperature = (temperature: number) => {
-    if (isAnthropic(currentLlm.provider, currentLlm.modelName)) {
-      setTemperature(Math.min(temperature, 1.0));
-    } else {
-      setTemperature(temperature);
-    }
+    const clampedTemp = isAnthropic(currentLlm.provider, currentLlm.modelName)
+      ? Math.min(temperature, 1.0)
+      : temperature;
+    setTemperature(clampedTemp);
     if (chatSession) {
-      updateTemperatureOverrideForChatSession(chatSession.id, temperature);
+      updateTemperatureOverrideForChatSession(chatSession.id, clampedTemp);
     }
   };
 


justin-tahara:
@greptile Can you please take a look here?

justin-tahara:
@greptile can you review the PR one more time?

web/src/lib/hooks.ts
```diff
@@ -651,6 +651,7 @@ export function useLlmManager(
 
     if (modelName) {
       const model = parseLlmDescriptor(modelName);
+      // If we have no parsed modelName, try to find the provider by the raw modelName string
       if (!(model.modelName && model.modelName.length > 0)) {
         const provider = llmProviders.find((p) =>
           p.model_configurations
```
nmgarza5, line 657: can we use model id?

justin-tahara:
@greptile can you review the PR one more time?

web/src/lib/hooks.ts
justin-tahara: Discussed offline, this will come in a separate PR given this will be quite the refactor

justin-tahara:
@greptile can you please review this one more time

PR #8244, opened 2026-02-06
feat: jwt-based auth

evan-onyx:
## Description

Support JWT-based Auth

## How Has This Been Tested?

Unit tested

## Additional Options

---
## Summary by cubic
Add a stateless JWT auth backend for single-tenant self-hosted deployments. It issues HS256-signed cookies with iat, validates audience, refreshes by minting a new token, and the web logout now reliably clears the auth cookie.

- **New Features**
  - Single-tenant JWT backend (HS256) with iat; refresh creates a new token; destroy is a no-op; guarded to block cloud/multi-tenant; requires USER_AUTH_SECRET; unit tests added.
  - Unified token-created-at lookup reads iat from the JWT cookie; frontend logout route now always clears the fastapiusersauth cookie.

- **Migration**
  - Set AUTH_BACKEND=jwt and USER_AUTH_SECRET (optional SESSION_EXPIRE_TIME_SECONDS) for self-hosted single-tenant.
  - No token store; server-side revocation is not supported. Cloud/multi-tenant should keep using redis or postgres backends.

<sup>Written for commit 2d06c624ece2052f6b745b0d1867f91fa8a1c798. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/app/auth/logout/route.ts
+++ b/web/src/app/auth/logout/route.ts
@@ -1,4 +1,3 @@
-import { NEXT_PUBLIC_CLOUD_ENABLED } from "@/lib/constants";
 import { getAuthTypeMetadataSS, logoutSS } from "@/lib/userSS";
 import { NextRequest } from "next/server";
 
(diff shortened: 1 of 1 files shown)


backend/onyx/auth/users.py
```diff
@@ -1046,6 +1047,56 @@ async def refresh_token(self, token: Optional[str], user: User) -> str:
...
+
+    async def write_token(self, user: User) -> str:
+        data = {
+            "sub": str(user.id),
+            "aud": self.token_audience,
+            "iat": int(datetime.now(timezone.utc).timestamp()),
+        }
+        return generate_jwt(
+            data, self.encode_key, self.lifetime_seconds, algorithm=self.algorithm
+        )
```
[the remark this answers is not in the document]
evan-onyx: generate_jwt from the fastapi_users library sets "exp" based on the lifetime_seconds value.

backend/onyx/auth/users.py
```diff
@@ -1058,6 +1109,19 @@ def get_database_strategy(
...
 
 
+def get_jwt_strategy() -> SingleTenantJWTStrategy:
+    return SingleTenantJWTStrategy(
+        secret=USER_AUTH_SECRET,
+        lifetime_seconds=SESSION_EXPIRE_TIME_SECONDS,
+    )
+
+
+if AUTH_BACKEND == AuthBackend.JWT and (MULTI_TENANT or AUTH_TYPE == AuthType.CLOUD):
```
yuhongsun96: if AUTH_BACKEND == AuthBackend.JWT:
      if not USER_AUTH_SECRET:
          raise ValueError("USER_AUTH_SECRET must be set when using JWT auth backend")
      if MULTI_TENANT or AUTH_TYPE == AuthType.CLOUD:
          raise ValueError(...)

^ Thoughts about something like this? Should we allow this being empty? Makes it really trivial to forge valid JWTs

backend/onyx/auth/users.py
```diff
@@ -1046,6 +1047,56 @@ async def refresh_token(self, token: Optional[str], user: User) -> str:
...
+    """Stateless JWT strategy for single-tenant deployments.
+
+    Tokens are self-contained and verified via signature — no Redis or DB
+    lookup required per request. An ``iat`` claim is embedded so that
+    downstream code can determine when the token was created without
+    querying an external store.
+
+    Refresh is implemented by issuing a brand-new JWT (the old one remains
+    valid until its natural expiry).  ``destroy_token`` is a no-op because
+    JWTs cannot be server-side invalidated.
```
yuhongsun96, line 1060: What happens when someone logs out? Should we bother with that? AI recommendation:

Add a token_version column to the User model. Include it in JWT claims and validate on each request.

  How it works:
  - On login: embed current token_version in JWT
  - On logout: increment token_version in DB
  - On validation: reject if JWT version != user's current version

  # In SingleTenantJWTStrategy.write_token()
  data = {
      "sub": str(user.id),
      "aud": self.token_audience,
      "iat": int(datetime.now(timezone.utc).timestamp()),
      "tv": user.token_version,  # token version
  }

  # Override read_token to check version
  async def read_token(self, token: str | None, user_manager) -> User | None:
      user = await super().read_token(token, user_manager)
      if user is None:
          return None

      payload = jwt.decode(token, options={"verify_signature": False})
      if payload.get("tv") != user.token_version:
          return None  # Token was invalidated
      return user

  Pros: Minimal DB change, instant invalidation, works for password changes too
  Cons: Requires one DB read per request (but you're already fetching the user)

evan-onyx: The way this should be handled normally is just throwing out the JWT on the frontend. I'm not a huge fan of this approach since it makes the login stateful, but I'll leave a note about it in the destroy_token func

PR #8268, opened 2026-02-09
fix(posthog): Only call when Key is available

justin-tahara:
## Description

Cleaning up Posthog calls to prevent blocking calls. Context is that there was a locked down environment that was not able to interact with their LLMs since they had posthog blocked within in their firewall. 

This then led to a blocking call meaning the LLM was never invoked. 

This PR aims to fix this behavior in order to ensure that we can properly skip this flow when the key is not set.

This needs to be cherry-picked to v2.9, v2.10, and v2.11

## How Has This Been Tested?

Ran tests.

## Additional Options

---
## Summary by cubic
Disable PostHog when no API key is configured to avoid blocking calls. Use a no‑op client so telemetry is skipped and feature flags fall back to safe defaults; also standardize user message telemetry with a new USER_MESSAGE_SENT event.

- **Bug Fixes**
  - Treat missing/whitespace POSTHOG_API_KEY as unset; expose POSTHOG_ENABLED.
  - Initialize real client only when enabled; otherwise use NoOpPosthogClient (identify/capture/flush/set are no‑ops).
  - Feature flags: when disabled, default to True; "onyx-craft-enabled" and "craft-has-usage-limits" default to False; provider returns fallbacks early.
  - Chat telemetry: send USER_MESSAGE_SENT via mt_cloud_telemetry; remove dynamic import fallback.
  - Add unit tests for disabled init, whitespace handling, and fallback flag defaults.

<sup>Written for commit 9359c4c1b4edae9e2ffddb00ac01c1b9c88ec797. Summary will update on new commits.</sup>

### DIFF
--- a/backend/ee/onyx/utils/posthog_client.py
+++ b/backend/ee/onyx/utils/posthog_client.py
@@ -18,6 +18,13 @@ def posthog_on_error(error: Any, items: Any) -> None:
     logger.error(f"PostHog error: {error}, items: {items}")
 
 
+# This module intentionally lives under `utils` (rather than `feature_flags`) because
+# PostHog is used by more than feature flags:
+# - Feature flag evaluation
+# - Cloud telemetry capture
+# - Marketing cookie/identity syncing
+
+# These flags are disabled by default when PostHog is not configured.
 POSTHOG_DISABLED_DEFAULT_OFF_FLAGS = frozenset(
     {
         # Build mode access should stay disabled unless explicitly enabled.
--- a/backend/ee/onyx/utils/telemetry.py
+++ b/backend/ee/onyx/utils/telemetry.py
@@ -1,5 +1,4 @@
 from ee.onyx.utils.posthog_client import posthog
-from ee.onyx.utils.posthog_client import POSTHOG_ENABLED
 from onyx.utils.logger import setup_logger
 
 logger = setup_logger()
@@ -9,9 +8,6 @@ def event_telemetry(
     distinct_id: str, event: str, properties: dict | None = None
 ) -> None:
     """Capture and send an event to PostHog, flushing immediately."""
-    if not POSTHOG_ENABLED:
-        return
-
     logger.info(f"Capturing PostHog event: {distinct_id} {event} {properties}")
     try:
         posthog.capture(distinct_id, event, properties)
--- a/backend/onyx/chat/process_message.py
+++ b/backend/onyx/chat/process_message.py
@@ -90,10 +90,6 @@ from onyx.tools.tool_constructor import SearchToolConfig
 from onyx.utils.logger import setup_logger
 from onyx.utils.telemetry import mt_cloud_telemetry
 from onyx.utils.timing import log_function_time
-from onyx.utils.variable_functionality import (
-    fetch_versioned_implementation_with_fallback,
-)
-from onyx.utils.variable_functionality import noop_fallback
 from shared_configs.contextvars import get_current_tenant_id
 
 logger = setup_logger()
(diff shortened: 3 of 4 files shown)


backend/ee/onyx/feature_flags/posthog_provider.py
```diff
@@ -34,6 +35,9 @@ def feature_enabled(
         Returns:
             True if the feature is enabled for the user, False otherwise.
         """
+        if not POSTHOG_ENABLED:
+            return False
```
wenxi-onyx: No posthog, all features should be enabled

backend/ee/onyx/utils/posthog_client.py
```diff
@@ -18,12 +18,38 @@ def posthog_on_error(error: Any, items: Any) -> None:
...
+        return None
+
+    def flush(self, *args: Any, **kwargs: Any) -> None:  # noqa: ARG002
+        return None
+
+    def set(self, *args: Any, **kwargs: Any) -> None:  # noqa: ARG002
+        return None
+
+    def feature_enabled(self, *args: Any, **kwargs: Any) -> bool:  # noqa: ARG002
+        return False
```
wenxi-onyx: ditto above

backend/ee/onyx/utils/posthog_client.py
```diff
@@ -18,12 +18,38 @@ def posthog_on_error(error: Any, items: Any) -> None:
     logger.error(f"PostHog error: {error}, items: {items}")
 
 
-posthog = Posthog(
-    project_api_key=POSTHOG_API_KEY,
-    host=POSTHOG_HOST,
-    debug=POSTHOG_DEBUG_LOGS_ENABLED,
-    on_error=posthog_on_error,
-)
+class NoOpPosthogClient:
```
wenxi-onyx, line 48: You might want to double check all currently feature-flagged items --> might be just Craft. I believe we left comments on whether or not self-hosted users should see craft (I believe it is no)

wenxi-onyx (changes requested):
Please verify default feature enabled behavior when no posthog

backend/ee/onyx/feature_flags/posthog_provider.py
```diff
@@ -34,6 +38,10 @@ def feature_enabled(
         Returns:
             True if the feature is enabled for the user, False otherwise.
         """
+        # If PostHog is not configured, fall back to static defaults.
+        if not POSTHOG_ENABLED:
```
wenxi-onyx, line 42: What is the case where we would get to this code?

If no POSTHOG_API_KEY --> we will not have a PostHogProvider --> although the bool is false, this code would never be reached

backend/ee/onyx/utils/posthog_client.py
```diff
@@ -18,13 +18,61 @@ def posthog_on_error(error: Any, items: Any) -> None:
     logger.error(f"PostHog error: {error}, items: {items}")
 
 
-posthog = Posthog(
-    project_api_key=POSTHOG_API_KEY,
-    host=POSTHOG_HOST,
-    debug=POSTHOG_DEBUG_LOGS_ENABLED,
-    on_error=posthog_on_error,
+POSTHOG_DISABLED_DEFAULT_OFF_FLAGS = frozenset(
```
wenxi-onyx, line 28: Thoughts on moving this entire file to `feature_flags/` dir?

Primarily because it feels confusing for future implementers that the posthog code is split into 2 places for (what I can tell) no real reason

backend/ee/onyx/utils/telemetry.py
```diff
@@ -8,6 +9,9 @@ def event_telemetry(
     distinct_id: str, event: str, properties: dict | None = None
 ) -> None:
     """Capture and send an event to PostHog, flushing immediately."""
+    if not POSTHOG_ENABLED:
```
wenxi-onyx: I believe there are 2 underlying issues here.

1) It is unclear whether or not we want to support self-hosted EE telemetry
2) Roshan implemented the wrong function

Prior to Roshan's PR `event_telemetry()` was **only** used in `mt_cloud_telemetry()`.

I believe we document chat telemetry as a self-hosted EE feature. However, `mt_cloud_telemetry` completely blocks this by checking for `MULTI_TENANT=true`. This is why `event_telemetry()` is implemented with `fetch_versioned_implementation_with_fallback()`, which should only be used to segregate EE functionality from CE functionality. 

Roshan should have used `mt_cloud_telemetry` since his intention was to log that info only for our Cloud.

--

tl;dr I don't think we should add this check in this function. The `NoOpPosthogClient` should silently pass this successfully. We can unblock the OG issue here by changing Roshan's code to use `mt_cloud_telemetry()`

backend/ee/onyx/utils/posthog_client.py
justin-tahara: Will look into doing this in a followup

justin-tahara:
closing for now as we are going to do this in a followup PR

PR #8269, opened 2026-02-09
fix: allow basic users to share agents

wenxi-onyx:
## Description
**Before**
basic-cant-share-agents

**After**
https://github.com/user-attachments/assets/24113180-b78f-4253-806b-9f3d786dffcc

## Additional Options

---
## Summary by cubic
Allow basic users to share agents by fixing the Share Agent modal to load real users (not API keys) and, when enterprise is enabled, their groups.

- **Bug Fixes**
  - Backend: new GET /user-groups/minimal returning MinimalUserGroupSnapshot; admins see all groups, others see only theirs.
  - Backend: /users supports include_api_keys (default false) and filters API key emails when false.
  - Frontend: replaced useUsers/useGroups with useShareableUsers/useShareableGroups; exclude the current user and pass includeApiKeys=false.
  - Enterprise gating: only fetch groups when enterprise features are enabled; otherwise show users only.

<sup>Written for commit 62a2bbdca9081bcc3cc3556c718526ce97bfc51f. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/hooks/useShareableGroups.ts
+++ b/web/src/hooks/useShareableGroups.ts
@@ -10,6 +10,9 @@ export interface MinimalUserGroupSnapshot {
   name: string;
 }
 
+// TODO (@raunakab):
+// Refactor this hook to live inside of a special `ee` directory.
+
 export default function useShareableGroups() {
   const combinedSettings = useContext(SettingsContext);
   const isPaidEnterpriseFeaturesEnabled =


wenxi-onyx:
> ### 💡 Codex Review
> https://github.com/onyx-dot-app/onyx/blob/7f4325baf51c2ec0b5226a4fdacf31ef807a8855/web/src/hooks/useShareableUsers.ts#L8-L10
> 
> 
> **P2 Badge Filter API-key users from shareable users list**
> Using `/api/users` here returns _all_ user rows from `get_all_users` without the `is_api_key_email_address` filter that the previous `/api/manage/users?include_api_keys=false` path applied. That means API-key pseudo-users (emails ending in the dummy API key domain) will now appear in the share dialog and can be selected for sharing. This is a regression versus the old hook and can unintentionally grant agent access to API keys rather than real users when any API key exists in the org. Consider filtering these users server-side or switching to an endpoint that excludes API-key accounts like the prior flow.
> 
> ℹ️ About Codex in GitHub

This was valid --> fixed

web/src/hooks/useShareableGroups.ts
```diff
@@ -0,0 +1,38 @@
...
+import useSWR from "swr";
+import { useContext } from "react";
+import { errorHandlingFetcher } from "@/lib/fetcher";
+import { SettingsContext } from "@/providers/SettingsProvider";
+
+export interface MinimalUserGroupSnapshot {
+  id: number;
+  name: string;
+}
+
```
raunakab, line 12: Minor comment here - please add the following todo for me to clean up later:

```
# TODO (@raunakab):
Refactor this hook to live inside of a special `ee` directory.
```

wenxi-onyx: added

PR #8289, opened 2026-02-10
fix(llm): Fixing seeding logic for LLM Model Flow

justin-tahara:
## Description

There is currently an issue if you upgrade your application code but you don't run a schema migration, you will be caught in a limbo state of not being able to access your previous LLM providers since you will have an overwritten model. 

This PR aims to fix this properly by checking if your LLMProviderModel is empty or not to make sure it's not overwritten with the seeding.

## How Has This Been Tested?

Tested this locally by going back into my schema migration and then ensuring that the application code is still intact without having ran the schema migrations yet.

## Additional Options

---
## Summary by cubic
Fixes LLM provider setup to avoid overwriting existing providers. Adds safe backfill for LLM model flow mappings and makes the migration idempotent to prevent the limbo state after upgrades without DB migrations.

- **Bug Fixes**
  - Only seed dev provider on truly fresh installs; skip when any provider exists.
  - Backfill missing flow mappings/defaults at startup using existing provider settings.
  - Make llm_model_flow migration idempotent and recompute defaults from legacy flags; added unit tests for fresh vs. existing/backfill paths.

<sup>Written for commit fdea0cab6e3ca0ac1a94241aa2d411c9275d90ee. Summary will update on new commits.</sup>

### DIFF
--- a/backend/alembic/versions/01f8e6d95a33_populate_flow_mapping_data.py
+++ b/backend/alembic/versions/01f8e6d95a33_populate_flow_mapping_data.py
@@ -17,8 +17,10 @@ depends_on = None
 
 
 def upgrade() -> None:
-    # Add each model config to the conversation flow, setting the global default if it exists
-    # Exclude models that are part of ImageGenerationConfig
+    # Add each model config to the conversation flow.
+    # Exclude models that are part of ImageGenerationConfig.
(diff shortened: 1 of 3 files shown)


justin-tahara:

PR #8299, opened 2026-02-10
fix(vertex): Updating masking workflow

justin-tahara:
## Description

The Vertex AI workflow was not working properly since we were not able to use Vertex AI models. 

This was due to the PR that was merged in recently. This PR ensures that Vertex is workflow properly with the masking workflow. 

## How Has This Been Tested?

Added new tests but also made sure to go through the process of adding a Vertex LLM Provider and then begin a chat with the model.

## Additional Options

---
## Summary by cubic
Fixes Vertex provider setup and chat by restoring masked secrets in custom_config during updates and tests. Vertex models now work again without losing stored credentials.

- **Bug Fixes**
  - Restore sensitive custom_config values (e.g., vertex_credentials) when the UI sends masked placeholders on provider update and test.
  - Run restoration before SSRF validation and change checks to prevent false diffs and accidental secret overwrites.
  - Add tests covering provider update and test flows to confirm masked values are preserved and non-sensitive fields can still change.

<sup>Written for commit c1212db4a519bb2e6ec3684ea66f7e8d2d7511a2. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/server/manage/llm/api.py
+++ b/backend/onyx/server/manage/llm/api.py
@@ -71,6 +71,7 @@ from onyx.server.manage.llm.utils import infer_vision_support
 from onyx.server.manage.llm.utils import is_valid_bedrock_model
 from onyx.server.manage.llm.utils import ModelMetadata
 from onyx.server.manage.llm.utils import strip_openrouter_vendor_prefix
+from onyx.utils.encryption import mask_string as mask_with_ellipsis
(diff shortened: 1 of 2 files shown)


yuhongsun96 (commented):
Can you use the existing provider configs rather than fetching it again? There is logic already to fetch the existing provider info since we need to get the API base for example.

PR #8301, opened 2026-02-10
fix(embedding): Updating Masking Logic

justin-tahara:
## Description

Refactoring the embedding models to use the new masking method

## How Has This Been Tested?

Tested locally by hitting this endpoint with an admin api key and ensured that things were working properly.

Obviously added a test as well

## Additional Options

---
## Summary by cubic
Updated embedding API key masking to use SensitiveValue during serialization, preventing raw key exposure and double-masking. Local embedding models now return no api_key.

- **Bug Fixes**
  - Mask cloud provider API keys when building EmbeddingModelDetail and CloudEmbeddingProvider responses.
  - Removed manual mask_string usage and redundant masking in admin endpoints.
  - Added unit tests for model/provider masking, local model behavior, and runtime access to the unmasked key.
  - Addressed mypy typing in EmbeddingModelDetail serialization and API helpers.

<sup>Written for commit e2664aadafda26be8186281fd08ba21f9c50a123. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/unit/onyx/server/manage/embedding/test_embedding_api.py
+++ b/backend/tests/unit/onyx/server/manage/embedding/test_embedding_api.py
@@ -101,4 +101,5 @@ def test_search_settings_api_key_property_returns_raw_value_for_runtime_use() ->
         cloud_provider=SimpleNamespace(api_key=_build_sensitive_value(raw_api_key))
     )
 
-    assert SearchSettings.api_key.fget(fake_search_settings) == raw_api_key
+    api_key_property = SearchSettings.__dict__["api_key"]
+    assert api_key_property.fget(fake_search_settings) == raw_api_key


backend/onyx/indexing/models.py
```diff
@@ -162,14 +162,21 @@ def from_db_model(
         cls,
         search_settings: "SearchSettings",
     ) -> "EmbeddingModelDetail":
+        api_key = None
+        if (
+            search_settings.cloud_provider is not None
+            and search_settings.cloud_provider.api_key is not None
+        ):
+            api_key = search_settings.cloud_provider.api_key.get_value(apply_mask=True)
```
evan-onyx, line 170: isn't this a change in the actual api key we're choosing to (mask and) send back?

PR #8308, opened 2026-02-10
fix(anthropic): Model Selection in Multi-Tenant

justin-tahara:
## Description

There is an issue with handling Anthropic Keys in the cloud so working on a proper fix.

## Additional Options

---
## Summary by cubic
Fixes Anthropic model selection in multi-tenant by treating empty api_base as unset. Prevents false “API base changed” detections so model-only updates don’t get blocked.

- **Bug Fixes**
  - Normalize ""/undefined api_base to None in upsert and change validation.
  - Compare normalized api_base values to avoid spurious changes in multi-tenant.
  - Frontend sends undefined for empty api_base; form submit normalizes "" to undefined.
  - Added test allowing empty-string api_base when existing value is None.

<sup>Written for commit 0fb04f29b909257d7d75eefea8050e205b7e4715. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/db/llm.py
+++ b/backend/onyx/db/llm.py
@@ -231,10 +231,7 @@ def upsert_llm_provider(
         # Set to None if the dict is empty after filtering
         custom_config = custom_config or None
 
-    api_base = llm_provider_upsert_request.api_base
-    if api_base == "":
-        api_base = None
-
+    api_base = llm_provider_upsert_request.api_base or None
     existing_llm_provider.provider = llm_provider_upsert_request.provider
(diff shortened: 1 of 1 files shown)


backend/onyx/db/llm.py
```diff
@@ -231,10 +231,14 @@ def upsert_llm_provider(
         # Set to None if the dict is empty after filtering
         custom_config = custom_config or None
 
+    api_base = llm_provider_upsert_request.api_base
```
evan-onyx: api_base = llm_provider_upsert_request.api_base or None

PR #8326, opened 2026-02-11
chore(admin): Improve Playwright test speeds

justin-tahara:
## Description

This is to just make the tests run faster. Purely optimizing the playwright tests to be faster.

## How Has This Been Tested?

Ran playwright locally

## Additional Options

---
## Summary by cubic
Speeds up Playwright E2E tests by replacing sleeps with deterministic waits and using lighter load states. Cuts runtime and flakiness across chat, scroll behavior, and MCP OAuth flows.

- **Refactors**
  - Use domcontentloaded instead of networkidle where full quiescence isn’t needed; remove post-send networkidle waits and fixed delays.
  - Replace waitForTimeout with state-based waits: locator/selector visibility (modals, inputs, tool toggles), waitForURL with regex, dialog visibility, and expect.poll for top-offset and scrollTop checks.
  - Harden MCP OAuth: faster polling (100→50ms, 250→100ms), wait for password field after username, shorter retry delays, assert toggles via data-state, and wait for auth modal fields instead of fixed sleeps.
  - Update auth/chat utils/tests to rely on URL regex + domcontentloaded and stable selectors; remove idle waits throughout.

<sup>Written for commit bd724dfb0439e3de9ddbd233d919ac5e70f262b5. Summary will update on new commits.</sup>

### DIFF
--- a/web/tests/e2e/mcp/mcp_oauth_flow.spec.ts
+++ b/web/tests/e2e/mcp/mcp_oauth_flow.spec.ts
@@ -279,7 +279,7 @@ async function waitForAnySelector(
 
 async function scrollToBottom(page: Page): Promise<void> {
   try {
-    await page.evaluate(async () => {
+    await page.evaluate(() => {
       const section = document.querySelector(
         '[data-testid="available-tools-section"]'
       );
@@ -288,11 +288,8 @@ async function scrollToBottom(page: Page): Promise<void> {
       } else {
         window.scrollTo(0, document.body.scrollHeight);
       }
-      // Let layout settle across two animation frames before continuing.
-      await new Promise<void>((resolve) =>
-        requestAnimationFrame(() => requestAnimationFrame(() => resolve()))
-      );
     });
+    await page.waitForTimeout(200);
   } catch {
     // ignore scrolling failures in test environment
   }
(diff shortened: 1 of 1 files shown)


web/tests/e2e/utils/auth.ts
```diff
@@ -136,14 +136,8 @@ export async function loginAsRandomUser(page: Page) {
...
   try {
-    // Wait for 2 seconds to ensure the signup process completes
-    await page.waitForTimeout(3000);
-    // Refresh the page to ensure everything is loaded properly
-    // await page.reload();
-
-    await page.waitForURL("/app?new_team=true");
-    // Wait for the page to be fully loaded after refresh
-    await page.waitForLoadState("networkidle");
+    await page.waitForURL(/\/app(\?.*)?$/, { timeout: 30000 });
```
jmelahman: Would double-check this is correct

web/tests/e2e/mcp/mcp_oauth_flow.spec.ts
```diff
@@ -387,7 +387,7 @@ async function performIdpLogin(page: Page): Promise<void> {
     await page
       .waitForLoadState("domcontentloaded", { timeout: 15000 })
       .catch(() => {});
-    await page.waitForTimeout(1000);
+    await page.waitForTimeout(300);
```
jmelahman: These can probably be removed as well in favor of an optimized `waitForSelector`. `waitForTimeout` might be something we call out as an explicit anti-pattern

PR #8328, opened 2026-02-11
feat(llm): Populate env w/ custom config

Danelegend:
## Description
Populate the .env with custom config values so that they are read by litellm. Does so in a thread-safe manner w/ a context manager so that there is no pollution from setting/unsetting the .env.

## How Has This Been Tested?
Unit tests
Tested with Mistral as a custom provider (Checked that concurrent streaming still works).
Also tested w/ various other providers

## Additional Options

---
## Summary by cubic
Populate per-request env vars from custom_config for LiteLLM calls under a global lock with automatic cleanup. invoke() now streams internally and reassembles the response to minimize lock time; when no custom_config is provided, we skip the lock.

- **New Features**
  - Added temporary_env_and_lock to set env vars under a global lock and restore them; when custom_config is absent, no lock is taken (via nullcontext).
  - invoke() uses stream=True and rebuilds via stream_chunk_builder so the lock is held only during setup; tests cover cleanup (including exceptions), multithreaded isolation for both invoke/stream, and lock-skipping without custom_config.

<sup>Written for commit 2ab6cb0cf9d53f0bf63cb013f013fb5bdf0a9924. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/integration/tests/web_search/test_web_search_api.py
+++ b/backend/tests/integration/tests/web_search/test_web_search_api.py
@@ -17,6 +17,7 @@ class TestOnyxWebCrawler:
     content from public websites correctly.
     """
 
+    @pytest.mark.skip(reason="Temporarily disabled")
     def test_fetches_public_url_successfully(self, admin_user: DATestUser) -> None:
         """Test that the crawler can fetch content from a public URL."""
         response = requests.post(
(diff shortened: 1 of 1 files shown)


backend/onyx/llm/multi_llm.py
```diff
@@ -475,22 +487,53 @@ def invoke(
...
+                        client=client,
+                    ),
+                )
+                chunks = list(stream_response)
+                response = cast(
+                    LiteLLMModelResponse,
+                    stream_chunk_builder(chunks),
+                )
+            else:
+                response = cast(
```
Danelegend: Not sure if this is a concern, but in this path, I believe there is a race condition where someone can jump in while another persons .env is set. 

This path should only be fired under our defined providers however (and only those that don't use custom config). Hence we wouldn't expect litellm to actually use the .env in this scenario

Danelegend:
@greptile

PR #8341, opened 2026-02-11
chore(playwright): remove unnecessary global auth checks

jmelahman:
## Description

Currently, there is a global check during setup that confirms that the `admin`, `admin2`, and `user` accounts can all login from the UI. We don't really need to check this -- we can just have the user creation pre-flight workflow fail in that case.

Moreover, replaces the `auth.ts` flow which performed browser-based login with an API-based for faster individual test running.

Also adds an explicit test for auth to cover the UI functionality spec (previously this was only assumed covered by these auth utils).

Seems to save about 2 min in CI for the `admin` playwright job :rocket: 

## How Has This Been Tested?

```
npx playwright test create_and_edit_assistant.spec.ts
```
Takes about `16s` locally now compared to `29s` previously + has fewer `403` and similar error messages spamming the console.

## Additional Options

---
## Summary by cubic
Removed global UI auth checks and switched Playwright auth to API login for faster, quieter runs. Global setup now saves storage state via API and ensures a default public LLM provider; added explicit login UI tests.

- **Refactors**
  - Global setup logs in via API request context and saves storage for admin, user, admin2; ensures a public LLM provider and sets it as default.
  - Auth utils use API; removed loginWithCredentials; loginAs and random-user flows call apiLogin (random user registers via API).
  - OnyxApiClient now takes APIRequestContext and optional baseUrl; lists via /admin/llm/provider; adds ensurePublicProvider and setProviderAsDefault.
  - Tests updated to pass page.request to OnyxApiClient; theme settings navigate before using localStorage, skip when Enterprise license is inactive, and guard cleanup when the form isn’t visible.

- **New Features**
  - Login UI tests: fields render, valid login, invalid password, unknown user.
  - LLM provider fixture to create a public provider when missing, with cleanup.

<sup>Written for commit 92f0d3f51d387982febd836574419d5306065adf. Summary will update on new commits.</sup>

### DIFF
--- a/web/tests/e2e/admin/theme/appearance_theme_settings.spec.ts
+++ b/web/tests/e2e/admin/theme/appearance_theme_settings.spec.ts
@@ -21,6 +21,16 @@ test.describe("Appearance Theme Settings", () => {
     await page.goto("/admin/theme");
     await page.waitForLoadState("networkidle");
 
+    // Skip the entire test when Enterprise features are not licensed.
+    // The /admin/theme page is gated behind ee_features_enabled and
+    // renders a license-required message instead of the settings form.
+    const eeLocked = page.getByText(
+      "This functionality requires an active Enterprise license."
+    );
+    if (await eeLocked.isVisible({ timeout: 1000 }).catch(() => false)) {
+      test.skip(true, "Enterprise license not active — skipping theme tests");
+    }
+
     // Clear localStorage to ensure consent modal shows
     await page.evaluate(() => {
       localStorage.removeItem("allUsersInitialPopupFlowCompleted");


web/tests/e2e/auth/login.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    const { email, password } = TEST_ADMIN_CREDENTIALS;
+
+    await page.goto("/auth/login");
+    await page.waitForLoadState("networkidle");
+
+    await page.getByTestId("email").fill(email);
+    await page.getByTestId("password").fill(password);
+    await page.getByRole("button", { name: "Sign In" }).click();
+
+    await expect(page).toHaveURL(/\/app/);
```
justin-tahara, line 31: I'm sure this shouldn't be an issue anymore but we don't need to check for any /chat endpoints right?

jmelahman: Shouldn't need to. We could maybe add a test that asserts `/chat` redirects to `/app` for backwards compatibility, but not sure it'd belong here (or is really all that interesting).

PR #8359, opened 2026-02-11
chore(opensearch): Tuning - Reduce k from 1000 to 50

acaprau:
## Description
We were seeing slow queries in the dev cluster. I'm still looking into this, there will be more PRs, but immediately I can say 1000 is almost certainly too large for `k` for `knn`, confirmed by a quick internet search using Onyx. Large `k` is a contributing factor to slow queries.

## How Has This Been Tested?
It wasn't.

## Additional Options

---
## Summary by cubic
Reduced k for OpenSearch kNN from 1000 to 50 to cut query latency and stabilize performance. Introduced DEFAULT_K_NUM_CANDIDATES, applied it in hybrid search, removed the num_candidates param, updated tests, and added comments/docstrings explaining the k choice.

<sup>Written for commit d27a6b727365bfefecfde653c031573af1f0a179. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/document_index/opensearch/search.py
+++ b/backend/onyx/document_index/opensearch/search.py
@@ -268,10 +268,6 @@ class DocumentQuery:
             tenant_state: Tenant state containing the tenant ID.
             index_filters: Filters for the hybrid search query.
             include_hidden: Whether to include hidden documents.
-            num_candidates: The number of neighbors to consider for vector
-                similarity search. Generally more candidates improves search
-                quality at the cost of performance. Defaults to
-                DEFAULT_NUM_CANDIDATES.
 
         Returns:
             A dictionary representing the final hybrid search query.


backend/onyx/document_index/opensearch/search.py
```diff
@@ -113,8 +113,13 @@
...
 
-# The default number of neighbors to consider for vector similarity search.
-DEFAULT_NUM_CANDIDATES = 1000
+# The default number of neighbors to consider for knn vector similarity search.
+# NOTE: Higher k slows down queries. Although this is a tuning question and
+# there is no correct value, a brief internet search will show that it is
+# generally accepted that k roughly equal to the number of hits you expect to
+# get is decent. I prefer to use a deterministic k rather than using num hits
+# directly which is dynamic; 50 seems reasonable.
+DEFAULT_K_NUM_CANDIDATES = 50
```
evan-onyx, line 122: talked offline; something like DEFAULT_CHUNKS_PER_SHARD that avoids overloading `K`

PR #8363, opened 2026-02-12
chore(llm): Backend Fallback Logic Tests

justin-tahara:
## Description

Introducing new tests to make sure we are covering fallback extraction paths and prevent regressions in tool-call reconstruction.

## How Has This Been Tested?

Ran the tests locally

## Additional Options

---
## Summary by cubic
Add unit and integration tests for fallback tool-call extraction when tool choice is REQUIRED, including deep_research mode and cases where tool-call JSON is embedded in assistant text. Update chat test helpers to pass deep_research in send_message and disconnect, and cover extraction from answer/reasoning, no-op when already attempted or tool_calls exist, and not-extractable cases.

<sup>Written for commit b1eae31f7682282e1e26bab0a886e15b1445ed1f. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/integration/tests/llm_workflows/test_mock_llm_tool_calls.py
+++ b/backend/tests/integration/tests/llm_workflows/test_mock_llm_tool_calls.py
@@ -95,7 +95,7 @@ def test_mock_llm_response_parallel_tool_call_debug(admin_user: DATestUser) -> N
     ]
 
 
-def test_mock_llm_response_reasoning_fallback_tool_call_debug(
+def test_mock_llm_response_embedded_json_fallback_tool_call_debug(
     admin_user: DATestUser,
 ) -> None:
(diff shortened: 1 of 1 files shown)


backend/tests/integration/common_utils/managers/chat.py
```diff
@@ -115,6 +115,7 @@ def send_message(
         forced_tool_ids: list[int] | None = None,
         chat_session: DATestChatSession | None = None,
         mock_llm_response: str | None = None,
+        deep_research: bool = False,
```
Danelegend, line 118: What was the reasoning behind these? Doesn't look like they're used

justin-tahara: Will try to specifically test out deep_research later but the arg was missing so there was no default value

PR #8368, opened 2026-02-12
feat(ee): gate access only when legacy EE flag is set and no license exists

nmgarza5:
## Description

Self-hosted license enforcement improvements:

- **Conditional gating on no-license**: `apply_license_status_to_settings` now checks `ENTERPRISE_EDITION_ENABLED` before setting `GATED_ACCESS` when no license is found. If the legacy EE flag is off (fresh install with only `LICENSE_ENFORCEMENT_ENABLED=true`), users get community features instead of the Access Restricted page. If the flag is on, access is gated because prior EE usage (e.g. permission syncing) may mean indexed data needs protection.
- **EE layout passthrough for GATED_ACCESS**: `ee/layout.tsx` now returns children when `application_status === "gated_access"`, deferring to the root layout's `GatedContentWrapper` which handles path-based exemptions (e.g. allowing `/ee/admin/billing` for license management).
- **Billing page redirect**: `BillingInformationPage` redirects to `/admin/billing` when there is no active subscription, so users land on the page with the full plans/checkout/license-upload flow.
- **Access Restricted page copy**: Uses the `useLicense` hook to distinguish "never had a license" vs "license lapsed" and shows appropriate messaging and CTAs for each case.
- **Billing link fix**: `AnnouncementBanner` and `AccessRestrictedPage` now link to `/admin/billing` instead of `/ee/admin/billing`.

## Additional Options

### DIFF
--- a/web/src/app/ee/layout.tsx
+++ b/web/src/app/ee/layout.tsx
@@ -26,7 +26,7 @@ export default async function AdminLayout({
       if (settings.ee_features_enabled === false) {
         // When the app is in GATED_ACCESS (expired or missing license), defer
         // to the root layout's GatedContentWrapper which handles path-based
-        // exemptions (e.g. allowing /ee/admin/billing for license management).
+        // exemptions (e.g. allowing /admin/billing for license management).
         if (settings.application_status === "gated_access") {
           return children;
         }


nmgarza5:
@greptile

PR #8372, opened 2026-02-12
refactor(opal): Interactive.Container always renders a <button>

raunakab:
## Description

`Interactive.Container` now always renders a `<button>` element instead of conditionally rendering a `<div>` or `<button>` based on the `type` prop. The `type` prop defaults to `"button"` (preventing accidental form submissions).

This is a semantic correctness fix — a component in the `Interactive` system should always produce a focusable, accessible `<button>` element. The `<div>` fallback path has been removed, along with the now-unnecessary ref and props type casts.

## How Has This Been Tested?

- `bun run types:check` passes with zero errors
- `bun run test` passes all 483 tests across 22 suites

## Additional Options

---
## Summary by cubic
Make Interactive.Container always render a semantic <button>, defaulting to type="button". This improves accessibility and prevents accidental form submissions.

- **Refactors**
  - Always render <button>; removed the <div> branch and related casts.
  - Default type="button"; map aria-disabled to disabled to block submits when disabled.

- **Migration**
  - If you relied on <div> semantics or selectors, update tests/CSS. Set type="submit" where form submission is intended.

<sup>Written for commit 73521055d13c70b78124617b6092e53073666a98. Summary will update on new commits.</sup>

### DIFF
--- a/web/lib/opal/src/core/interactive/components.tsx
+++ b/web/lib/opal/src/core/interactive/components.tsx
@@ -303,27 +303,16 @@ function InteractiveBase({
  * Extends standard `<div>` attributes (minus `className` and `style`).
  */
 interface InteractiveContainerProps
-  extends WithoutStyles<React.HTMLAttributes<HTMLDivElement>> {
+  extends WithoutStyles<React.HTMLAttributes<HTMLButtonElement>> {
   /**
-   * Ref forwarded to the underlying element.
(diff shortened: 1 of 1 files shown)


raunakab:
Closing because we don't always want to render a `button` (especially to avoid "HTML invalid button nesting" errors).

PR #8376, opened 2026-02-12
feat(search-settings): Add tests + contextual llm validation

Danelegend:
## Description
This PR adds integration and extenal dependecy tests for the contextual llm. It also adds some validation that the contextual llm must actually exist if it is wanted to be used.

Parts are currently commented out while opensearch changes are being made

## How Has This Been Tested?
Tests have been added to check it is a safe pathway.

## Additional Options

---
## Summary by cubic
Adds server-side validation for contextual RAG LLM settings and expands test coverage. Updates now require a real provider and model and return clear 400 errors when invalid; set-new remains disabled.

- **New Features**
  - DB-backed validation on update-inference-settings; errors for missing provider/model, provider not found, and model not found in provider.
  - Integration tests for search-settings endpoints: get current/all/secondary; set, unset, and switch contextual RAG provider/model; preserved fields in update-inference-settings.
  - External dependency tests for the indexing pipeline using the configured contextual LLM (and skipping when disabled); currently skipped while set-new is disabled.

<sup>Written for commit b3ffdb2690b3b07fd09ad5d54383b2793cfeae80. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/server/manage/search_settings.py
+++ b/backend/onyx/server/manage/search_settings.py
@@ -42,8 +42,8 @@ def set_new_search_settings(
     Gives an error if the same model name is used as the current or secondary index
     """
     # TODO(andrei): Re-enable.
-    # NOTE Enable external dependency tests in test_search_settings.py
-    # when this is reenabled
+    # NOTE Enable integration external dependency tests in test_search_settings.py
+    # when this is reenabled. They are currently skipped
     logger.error("Setting new search settings is temporarily disabled.")
     raise HTTPException(
         status_code=status.HTTP_501_NOT_IMPLEMENTED,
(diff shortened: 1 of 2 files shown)


backend/onyx/server/manage/search_settings.py
```diff
@@ -257,3 +281,25 @@ def delete_unstructured_api_key_endpoint(
     _: User = Depends(current_admin_user),
 ) -> None:
     delete_unstructured_api_key()
+
+
+def validate_contextual_rag_model(
```
evan-onyx, line 276: this pattern is a bit non-standard, typically a validate_...() function will error if validation fails. Would prefer making this a helper with an underscore in front, then have the public facing validate() function raise the 400

PR #8384, opened 2026-02-12
refactor: github connector

evan-onyx:
## Description

minor github change for consistency

## How Has This Been Tested?

n/a pure refactor

## Additional Options

---
## Summary by cubic
Consolidated GitHub repository selection into a single method to keep behavior consistent and remove duplicated logic. Added mypy-friendly type checks; no functional changes.

- **Refactors**
  - Added fetch_configured_repos() to centralize repo selection (single, multiple, or all).
  - Updated doc_sync and initial fetch to use it; added type hints and a github_client assert for mypy.

<sup>Written for commit df575b578098f54c423ad353836b9fe189f1cc96. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/connectors/github/connector.py
+++ b/backend/onyx/connectors/github/connector.py
@@ -530,6 +530,7 @@ class GithubConnector(CheckpointedConnectorWithPermSync[GithubConnectorCheckpoin
         Returns:
             list[Repository.Repository]: The configured repositories.
         """
+        assert self.github_client is not None  # mypy
         if self.repositories:
             if "," in self.repositories:
                 return self.get_github_repos(self.github_client)


backend/onyx/connectors/github/connector.py
```diff
@@ -523,6 +523,21 @@ def get_all_repos(
...
+            list[Repository.Repository]: The configured repositories.
+        """
+        if self.repositories:
+            if "," in self.repositories:
+                return self.get_github_repos(self.github_client)
+            else:
+                return [self.get_github_repo(self.github_client)]
+        else:
+            return self.get_all_repos(self.github_client)
+
```
justin-tahara, line 540: Can we just create a small test for this to test out the different types of repos?

I think it would help especially for me to understand what falls into the 3 different paths.

evan-onyx: this is just parsing out the different possibilities for the connector config. All the calls are hitting the github API, and tbh the logic is a bit too simple to unit test

PR #8385, opened 2026-02-12
fix(ollama): Passing Context Window through

justin-tahara:
## Description

We are not passing in any model token context window values to Ollama and thus in certain cases I believe we default to the 4096 token count meaning we will not be able to properly support models in Ollama. 

This changes makes it so we properly use the `num_ctx` parameter that Ollama looks for in order to properly set the context limit window and ensure that we are getting the proper count passed through.

## How Has This Been Tested?

Tested this with a number of different Ollama models that we have access to and wrote some tests as well

## Additional Options

---
## Summary by cubic
Fixes Ollama context window handling by passing each model’s configured max_input_tokens to max_input_tokens and, for Ollama, to num_ctx, with a safe fallback when unknown. Centralizes LLM creation via llm_from_provider and makes sure model_kwargs reach LitellmLLM.

- **Bug Fixes**
  - For OLLAMA_CHAT, set model_kwargs.num_ctx to the model’s configured max_input_tokens (>0); omit when unknown.
  - Set max_input_tokens from the configured value; otherwise fall back to get_max_input_tokens_from_llm_provider.
  - Never set num_ctx for non-Ollama providers.
  - Route all LLM creation (including vision) through llm_from_provider and propagate model_kwargs into get_llm/LitellmLLM.
  - Add unit tests for Ollama, non-Ollama, and unknown-context paths.

<sup>Written for commit 6d2474c233b887308ffcc6fe24a0e316f43c6d1b. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/unit/onyx/llm/test_factory.py
+++ b/backend/tests/unit/onyx/llm/test_factory.py
@@ -1,12 +1,12 @@
 from unittest.mock import patch
 
 from onyx.llm.constants import LlmProviderNames
+from onyx.llm.factory import _build_provider_extra_headers
 from onyx.llm.factory import get_llm
 from onyx.llm.factory import llm_from_provider
-from onyx.llm.factory import _build_provider_extra_headers
+from onyx.llm.well_known_providers.constants import OLLAMA_API_KEY_CONFIG_KEY
 from onyx.server.manage.llm.models import LLMProviderView
 from onyx.server.manage.llm.models import ModelConfigurationView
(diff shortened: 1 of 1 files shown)


redhell:
@justin-tahara Your PR overrides the system default by ollama.
-> e.g. your context window set by ollama is 8k and gpt-oss:20b has 128k context window.

We let the user should decide if the context window defined by ollama is taken oder or override it through the model. 

Right now I have massive offload to gpu/cpu

PR #8386, opened 2026-02-12
chore(devtools): `ods screenshot-diff` for visual regression testing

jmelahman:
## Additional Options

---
## Summary by cubic
Adds a Playwright-based visual regression workflow with a new ods screenshot-diff tool, S3-backed baselines, and a single PR comment linking to HTML diff reports. Replaces Chromatic with in-repo screenshots and adds light/dark admin page coverage.

- **New Features**
  - CLI: ods screenshot-diff compare | upload-baselines with project defaults, local-vs-S3 and cross-revision S3-vs-S3 compare, always writes summary.json, and only generates HTML when diffs exist (requires onyx-devtools 0.5.7).
  - Playwright: global screenshot thresholds (threshold 0.2, maxDiffPixelRatio 0.01), capture-only mode unless VISUAL_REGRESSION=true, outputs to web/output/playwright and web/output/screenshots, and an expectScreenshot helper (masking, full-page).
  - Tests: admin_pages.spec.ts captures light and dark screenshots across key admin routes and waits for networkidle.
  - CI: determines baseline rev (PR base, tag, or branch), configures OIDC AWS creds, runs ods compare via uv, uploads reports/summary only on diffs, posts a single combined PR comment with counts and S3 report links, updates baselines on main/release/tags, and uploads artifacts from web/output/playwright and web/output/screenshots.

- **Migration**
  - Install Playwright deps: npx playwright install; enable assert mode with VISUAL_REGRESSION=true if desired.
  - AWS CLI required for S3 sync; bucket defaults to onyx-playwright-artifacts (override via PLAYWRIGHT_S3_BUCKET).
  - Local use: ods screenshot-diff compare --project <name> to generate reports; CI updates baselines automatically on protected branches.

<sup>Written for commit 666f3e0d77120219acb4bede2b6da522763d390d. Summary will update on new commits.</sup>

### DIFF
--- a/.github/workflows/pr-playwright-tests.yml
+++ b/.github/workflows/pr-playwright-tests.yml
@@ -242,7 +242,6 @@ jobs:
   playwright-tests:
     needs: [build-web-image, build-backend-image, build-model-server-image]
     name: Playwright Tests (${{ matrix.project }})
-    environment: ci-protected
     permissions:
       id-token: write # Required for OIDC-based AWS credential exchange (S3 access)
       contents: read


.github/workflows/pr-playwright-tests.yml
```diff
@@ -52,6 +52,9 @@ env:
   MCP_SERVER_PUBLIC_HOST: host.docker.internal
   MCP_SERVER_PUBLIC_URL: http://host.docker.internal:8004/mcp
 
+  # Visual regression S3 bucket (shared across all jobs)
+  PLAYWRIGHT_S3_BUCKET: onyx-playwright-artifacts
```
justin-tahara, line 56: Is there any lifecycle policies on this bucket? I don't think we need to keep things around maybe after a month? Will let you figure out what you think is best for that but ideally we do set limits just so we don't add more cost

.github/workflows/pr-playwright-tests.yml
```diff
@@ -428,18 +434,124 @@ jobs:
...
+          path: ./web/output/screenshots/
+          retention-days: 30
+
+      # --- Visual Regression Diff ---
+      - name: Install the latest version of uv
+        if: always()
+        uses: astral-sh/setup-uv@61cb8a9741eeb8a550a1b8544337180c0fc8476b # ratchet:astral-sh/setup-uv@v7
+        with:
+          enable-cache: false
+          version: "0.9.9"
```
justin-tahara, line 468: Small nit: I assume you want to consolidate this at one point

.github/workflows/pr-playwright-tests.yml
```diff
@@ -457,6 +569,95 @@ jobs:
...
+          path: summaries/
+
+      - name: Post combined PR comment
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          RUN_ID: ${{ github.run_id }}
+          REPO: ${{ github.repository }}
+          S3_BUCKET: ${{ env.PLAYWRIGHT_S3_BUCKET }}
+        run: |
```
justin-tahara, line 601: I'm sure theres a good reason that you added this code block here but it does seem a bit odd since it is a pretty chunky block

web/tests/e2e/admin_pages.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+      {
+        timeout: 10000,
+      }
+    );
+  } catch (error) {
+    console.error(
+      `Failed to find admin-page title with text "${pageTitle}" for path "${path}"`
+    );
+    // NOTE: This is a temporary measure for debugging the issue
+    console.error(await page.content());
```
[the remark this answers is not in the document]
justin-tahara: Seems like a valid comment

tools/ods/internal/imgdiff/report.go
```diff
@@ -0,0 +1,345 @@
...
+func imageToDataURI(img image.Image) (string, error) {
+	var buf bytes.Buffer
+	if err := png.Encode(&buf, img); err != nil {
+		return "", err
+	}
+	encoded := base64.StdEncoding.EncodeToString(buf.Bytes())
+	return "data:image/png;base64," + encoded, nil
+}
+
+const htmlTemplate = `<!DOCTYPE html>
```
justin-tahara, line 137: Do you think we should put this in another file and then pull it in? I just think it might be a lil bit cleaner

.github/workflows/pr-playwright-tests.yml
jmelahman: We can play it be ear. Monthly was to support release comparisons since they're on the order of a few weeks. The github artifact retention policies don't matter that much as there's also a global 10GB limit or something, so none of those artifacts will ever live long enough to hit 30d

.github/workflows/pr-playwright-tests.yml
jmelahman: Probably eventually. Not in a rush atm

.github/workflows/pr-playwright-tests.yml
jmelahman: it's mostly just a `jq` wrapper. it's fine for now

tools/ods/internal/imgdiff/report.go
jmelahman: Maybe. This page can be improved a bit yet. `--report-template` might make sense, so we can iterate on it without needing new versions of `ods`. I'll think about it, but not too worried tbh

.github/workflows/pr-playwright-tests.yml
jmelahman: Honestly, might even make sense as a Github Variable and just duplicate the step everywhere

jmelahman:
```
      - name: Install the latest version of uv
        if: always()
        uses: astral-sh/setup-uv@61cb8a9741eeb8a550a1b8544337180c0fc8476b # ratchet:astral-sh/setup-uv@v7
        with:
          enable-cache: false
          version: "{{ variables.UV_VERSION }}"
```
etc.

web/tests/e2e/admin_pages.spec.ts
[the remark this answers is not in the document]
jmelahman: This test was a direct port from the Chromatic version, FYI, https://github.com/onyx-dot-app/onyx/pull/8339/changes#diff-a67c56bb3aa570ebdbb950554c83b9cf686547010fab0a70fde199885b1861c9L30-L31. It can probably be improved and is mostly just an example.

tools/ods/internal/imgdiff/report.go
jmelahman: Given this would require another release of `ods`, I'd rather just get this in for now.

PR #8388, opened 2026-02-12
chore(helm): Make the command we give to celery light templated on values

acaprau:
## Description
This should allow us to overwrite the commend per-cluster, for the purpose of setting log level, queues, etc.

Ideally we do this for every celery type, if @justin-tahara likes this PR I can do that.

## How Has This Been Tested?
I did not test it yet, ideally we try with the dev cluster to see that the override works, and that we correctly revert to the default when it's not set.

## Additional Options

---
## Summary by cubic
Made the celery-worker-light command Helm-templated so clusters can override log level, queues, and hostname without code changes. Defaults match the previous command; chart version bumped to 0.4.28.

- **Migration**
  - No action needed; default behavior is unchanged.
  - To customize per cluster, set values.celery_worker_light.command to your desired command array.

<sup>Written for commit a7729f2d331e968795ee87d37fc912eee7cb998b. Summary will update on new commits.</sup>

### DIFF
--- a/deployment/helm/charts/onyx/Chart.yaml
+++ b/deployment/helm/charts/onyx/Chart.yaml
@@ -5,7 +5,7 @@ home: https://www.onyx.app/
 sources:
   - "https://github.com/onyx-dot-app/onyx"
 type: application
-version: 0.4.27
+version: 0.4.28
 appVersion: latest
 annotations:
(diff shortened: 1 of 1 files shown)


acaprau:
justin did a better job than this in #8495

PR #8391, opened 2026-02-12
fix: Reminders polluting the query expansion

yuhongsun96:
## Description
This PR fixes several related issues:
- Reminder messages in the history would get passed through the tools and they would get treated as the last user message instead of the actual query.
- Reminder messages have weird artifacts, using the OpenCode/Anthropic tags approach for reminders for a cleaner approach.
- Some of the tracings were missing outputs, adding those to verify the prompts for the above ^
- Added a new MessageType (USER_REMINDER) which is handled separately with injecting the tags above ^ to ensure correctness.
- Removed some dead files and constants 

## Additional Options

---
## Summary by cubic
Prevents reminder messages from being treated as user queries by introducing a USER_REMINDER type wrapped in <system-reminder> tags. Also standardizes tracing to record full LLM responses (content, tool calls, reasoning, usage).

- **Bug Fixes**
  - Reminders are now MessageType.USER_REMINDER and do not influence query expansion or tool selection.
  - USER_REMINDER messages are wrapped with <system-reminder> tags and sent as user messages only for context.
  - System and chat-naming prompts explain how to handle reminder tags.
  - Deep Research and chat session naming inject reminders correctly without polluting the main query.

- **Refactors**
  - Added record_llm_response and updated flows to capture content, tool calls, reasoning, and usage in spans; enhanced record_llm_span_output to support tool_calls.
  - Updated section relevance classification scale from 1–4 to 0–3 and aligned parsing/mapping.
  - Removed unused prompt files (direct_qa_prompts, query_validation, custom_tool_prompts) and constants; localized SOURCES_KEY where needed.
  - Added tests for USER_REMINDER handling and LLM span recording, including usage and tool_calls.

<sup>Written for commit 18adae082f092f1088c4b876671c37fbda66fddd. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/prompts/chat_prompts.py
+++ b/backend/onyx/prompts/chat_prompts.py
@@ -1,6 +1,7 @@
 # ruff: noqa: E501, W605 start
 
 from onyx.prompts.constants import REMINDER_TAG_DESCRIPTION
+from onyx.prompts.constants import REMINDER_TAG_NO_HEADER
 
 
 DATETIME_REPLACEMENT_PAT = "{{CURRENT_DATETIME}}"
@@ -49,8 +50,6 @@ DO NOT provide any links following the citations. Cite inline as opposed to leav
 # Reminder message if any search tool has been run anytime in the chat turn
 CITATION_REMINDER = """
 Remember to provide inline citations in the format [1], [2], [3], etc. based on the "document" field of the documents.
-
-Do not acknowledge this hint in your response.
 """.strip()
 
 LAST_CYCLE_CITATION_REMINDER = """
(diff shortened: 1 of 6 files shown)


backend/onyx/deep_research/dr_loop.py
```diff
@@ -484,16 +484,13 @@ def run_deep_research_llm_loop(
                     system_prompt=system_prompt,
                     custom_agent_prompt=None,
                     simple_chat_history=simple_chat_history,
-                    reminder_message=None,
+                    reminder_message=first_cycle_reminder_message,
```
yuhongsun96, line 487: No logical change here, just the other way previously was not reusing provided utilities

backend/onyx/secondary_llm_flows/document_filter.py
```diff
@@ -133,25 +133,25 @@ def classify_section_relevance(
...
             classification = default_classification
         else:
-            # Parse the response to extract the situation number (1-4)
-            numbers = re.findall(r"\b[1-4]\b", llm_response)
+            # Parse the response to extract the situation number (0-3)
+            numbers = re.findall(r"\b[0-3]\b", llm_response)
             if numbers:
                 situation = int(numbers[-1])
                 # Map situation number to ContextExpansionType
                 situation_to_type = {
```
Danelegend, line 150: This feels like it should be some sort of constant and used inside the regex above b[{lowest number}-{highest number}]

yuhongsun96: Not sure I fully understand the suggestion, can you share a diff? I want to get this in for a customer so I'm going to let this merge but happy to do a follow up PR to change it

PR #8397, opened 2026-02-12
fix(billing): exclude inactive users from seat counts and allow users page when gated

nmgarza5:
## Description

Inactive users were being counted toward seat enforcement in the billing UI, inflating the "in use" seat count and forcing admins to purchase more seats than needed. This PR:

- Filters seat counts in `CheckoutView` and `BillingDetailsView` to only include active users (`is_active === true`), matching the backend's `get_used_seats()` behavior
- Allows navigation to `/admin/users` when the app is in gated access mode (expired/missing license), so admins can adjust users before subscribing
- Removes `size="sm"` from InputNumber reset button (prettier/lint fix)

## How Has This Been Tested?
manual testing

## Additional Options

---
## Summary by cubic
Fixes seat counts to include only active users, preventing inflated usage and unnecessary purchases. Also allows access to the Users page when licensing is gated so admins can deactivate users before subscribing.

- **Bug Fixes**
  - Filter accepted and Slack users by is_active in CheckoutView and BillingDetailsView to match backend seat logic.
  - Permit navigation to /admin/users under gated access via GatedContentWrapper.

- **Refactors**
  - Remove size="sm" from InputNumber reset button (lint/prettier cleanup).

<sup>Written for commit 347052262eefe9d125a3c74dc5992ee67cadccc6. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/refresh-components/inputs/InputNumber.tsx
+++ b/web/src/refresh-components/inputs/InputNumber.tsx
@@ -167,7 +167,6 @@ export default function InputNumber({
             onClick={handleReset}
             disabled={!canReset || isDisabled}
             prominence="tertiary"
-            size="sm"
           />
         )}
         <div className="flex flex-col">
--- a/web/src/app/admin/billing/BillingDetailsView.tsx
+++ b/web/src/app/admin/billing/BillingDetailsView.tsx
@@ -281,8 +281,10 @@ function SeatsCard({
   });
 
   const totalSeats = billing?.seats ?? license?.seats ?? 0;
-  const acceptedUsers = usersData?.accepted?.length ?? 0;
-  const slackUsers = usersData?.slack_users?.length ?? 0;
+  const acceptedUsers =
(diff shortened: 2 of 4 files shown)


web/src/refresh-components/inputs/InputNumber.tsx
```diff
@@ -167,7 +167,6 @@ export default function InputNumber({
             onClick={handleReset}
             disabled={!canReset || isDisabled}
             prominence="tertiary"
-            size="sm"
```
justin-tahara, line 170: Is this intentional?

justin-tahara (approved):

web/src/refresh-components/inputs/InputNumber.tsx
nmgarza5, line 170: Image

nmgarza5: yes

PR #8398, opened 2026-02-12
feat(contextual-llm): Populate and set w/ llm flow

Danelegend:
## Description
This PR populates the llm_model_flow for contextual rag type via a database migration and code changes. It does this concurrently w/ search_settings. Reading still takes place on search_settings. A separate PR will move the reading over to search_settings

The contextual rag llm to be used should be the one defined by the PRESENT search settings

## How Has This Been Tested?
Tests
Manual db migration verification

## Additional Options

---
## Summary by cubic
Adds a CONTEXTUAL_RAG LLM flow and keeps its default in llm_model_flow synced to PRESENT search settings; reads still use search_settings. Migration backfills flow rows and widens the flow type column.

- **New Features**
  - Introduced CONTEXTUAL_RAG flow type (replaces EMBEDDINGS) and helpers: fetch_default_contextual_rag_model, add_model_to_flow, update_default_contextual_model, update_no_default_contextual_rag_provider.
  - Syncs default on PRESENT updates and FUTURE→PRESENT swaps; clears when disabled or invalid. If the model is missing, default is cleared and PRESENT settings are disabled with provider/name reset.
  - Tests cover default before/after swap and updates, and that indexing uses the default when enabled (and skips when disabled).

- **Migration**
  - Backfills contextual RAG flow rows from search_settings, marks PRESENT as default, widens llm_model_flow_type to varchar(20), and downgrade removes contextual_rag rows and returns the column to varchar(10).

<sup>Written for commit b9f72e0495c0a45e1ffd72ba3ac6b9de19cfc0d0. Summary will update on new commits.</sup>

### DIFF
--- a/backend/alembic/versions/19c0ccb01687_migrate_to_contextual_rag_model.py
+++ b/backend/alembic/versions/19c0ccb01687_migrate_to_contextual_rag_model.py
@@ -1,7 +1,7 @@
 """Migrate to contextual rag model
 
 Revision ID: 19c0ccb01687
-Revises: b51c6844d1df
+Revises: 9c54986124c6
 Create Date: 2026-02-12 11:21:41.798037
 
 """
@@ -12,7 +12,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision = "19c0ccb01687"
-down_revision = "b51c6844d1df"
+down_revision = "9c54986124c6"
 branch_labels = None
 depends_on = None
 


backend/alembic/versions/19c0ccb01687_migrate_to_contextual_rag_model.py
```diff
@@ -0,0 +1,63 @@
...
+    # For every search_settings row that has contextual rag configured,
+    # create an llm_model_flow entry. is_default is TRUE if the row
+    # belongs to the PRESENT search settings, FALSE otherwise.
+    op.execute(
+        """
+        INSERT INTO llm_model_flow (llm_model_flow_type, model_configuration_id, is_default)
+        SELECT DISTINCT
+            'CONTEXTUAL_RAG',
+            mc.id,
+            (ss.status = 'PRESENT')
```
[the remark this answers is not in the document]
Danelegend: We only expect 1 search settings row to exist. There is a unique index on this as well

Danelegend:
@greptile

backend/onyx/db/llm.py
```diff
@@ -646,6 +652,73 @@ def update_default_vision_provider(
...
+        or not contextual_rag_llm_provider
+    ):
+        update_no_default_contextual_rag_provider(db_session=db_session)
+        return
+
+    provider = fetch_existing_llm_provider(
+        name=contextual_rag_llm_provider, db_session=db_session
+    )
+    if not provider:
+        raise ValueError(f"Provider '{contextual_rag_llm_provider}' not found")
```
Danelegend, line 692: I believe this scenario is possible if the provider gets deleted

backend/alembic/versions/19c0ccb01687_migrate_to_contextual_rag_model.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+
+
+def downgrade() -> None:
+    op.execute(
+        """
+        DELETE FROM llm_model_flow
+        WHERE llm_model_flow_type = 'CONTEXTUAL_RAG'
+        """
+    )
```
evan-onyx, line 63: we need to alter the column back to the old length right? we should test that you can upgrade -> downgrade 1 -> upgrade

PR #8402, opened 2026-02-12
chore: hide file reader

evan-onyx:
## Description

hides the file reader tool when not enabled ( this is a stopgap)

## How Has This Been Tested?

tested in UI

## Additional Options

---
## Summary by cubic
Hide the File Reader tool from the Actions popover when it isn’t available (e.g., vector DB disabled) so users don’t see a non-functional option.

<sup>Written for commit 4123ca1c8c41f5a7d06017d3a659e633ca8ce1a5. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/refresh-components/popovers/ActionsPopover/index.tsx
+++ b/web/src/refresh-components/popovers/ActionsPopover/index.tsx
@@ -1,6 +1,7 @@
 "use client";
 
 import {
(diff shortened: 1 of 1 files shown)


evan-onyx:
> Is this the best way to do it? I :sus: but whatever fine for now

yeah it is not but it works for now

PR #8403, opened 2026-02-12
fix: SearXNG works now

yuhongsun96:
## Description
SearXNG no work -> now work

## How Has This Been Tested?
Ran SearXNG locally, verified setup of other providers like Serper still work.

## Additional Options

---
## Summary by cubic
SearXNG now works without an API key. Server and tool logic use provider-specific key requirements and validate the SearXNG base URL.

- **Bug Fixes**
  - Added provider_requires_api_key and made build_search_provider_from_config accept an optional api_key.
  - SearXNG: removed API key requirement and now require searxng_base_url.
  - Updated the test_search_provider API and WebSearchTool to enforce keys only when needed and handle None keys.
  - Added unit tests for SearXNG and key requirements for EXA, Serper, and Google PSE.

<sup>Written for commit 4fcd3a894ac5f147f58d6f5b114380b90ab0dafc. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/tools/tool_implementations/web_search/providers.py
+++ b/backend/onyx/tools/tool_implementations/web_search/providers.py
@@ -36,7 +36,10 @@ logger = setup_logger()
 
 
 def provider_requires_api_key(provider_type: WebSearchProviderType) -> bool:
-    """Return True if the given provider type requires an API key."""
+    """Return True if the given provider type requires an API key.
+    This list is most likely just going to contain SEARXNG. The way it works is that it uses public search engines that do not
+    require an API key. You can also set it up in a way which requires a key but SearXNG itself does not require a key.
+    """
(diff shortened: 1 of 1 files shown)


backend/onyx/tools/tool_implementations/web_search/providers.py
```diff
@@ -35,14 +35,33 @@
 logger = setup_logger()
 
 
+def provider_requires_api_key(provider_type: WebSearchProviderType) -> bool:
+    """Return True if the given provider type requires an API key."""
+    return provider_type != WebSearchProviderType.SEARXNG
```
justin-tahara, line 43: Can we leave a comment as to why SEARXNG is here? I know you are just moving this function around but just reading this function makes no sense without context

justin-tahara (approved):

PR #8407, opened 2026-02-13
fix(celery): Respecting Limits for Celery Heavy Tasks

justin-tahara:
## Description

Currently heavy tasks on the cloud are not respecting the timeouts that are set because they are being spawned in a thread pool. This allows us to have tasks that run for multiple hours that end up causing a backlog of jobs in the queue. 

This PR aims to properly handle these jobs in order to get through the queue. 

Will look into a followup PR to figure out why these jobs are so long running.

## How Has This Been Tested?

Ran existing tests to validate that the code path is intact.

## Additional Options

---
## Summary by cubic
Enforce timeouts for heavy Celery tasks by adding internal stop/timeout checks, preventing multi-hour runs that block queues. Applies a consistent JOB_TIMEOUT across permission sync, external group sync, pruning, and doc processing, with runtime checks inside long loops.

- **Bug Fixes**
  - Add timeout_seconds to PermissionSyncCallback and IndexingCallback; track elapsed with time.monotonic and stop on fence or timeout.
  - In permission sync and external group sync loops, check should_stop or elapsed and raise RuntimeError with cc_pair and progress details.
  - Pass timeout_seconds=JOB_TIMEOUT when constructing pruning and permission sync callbacks; set a default timeout in external group sync.
  - Log a warning when timeouts are exceeded; enforce time limits in code since Celery soft_time_limit doesn’t apply to thread pools.

<sup>Written for commit 02a83f4fd5e8f17c9c66f6c18ac2631074bf3304. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/background/celery/tasks/docprocessing/utils.py
+++ b/backend/onyx/background/celery/tasks/docprocessing/utils.py
@@ -68,7 +68,7 @@ class IndexingCallbackBase(IndexingHeartbeatInterface):
             elapsed = time.monotonic() - self.start_monotonic
             if elapsed > self.timeout_seconds:
                 logger.warning(
-                    f"{self.__class__.__name__} - task timeout exceeded: "
+                    f"IndexingCallback Docprocessing - task timeout exceeded: "
                     f"elapsed={elapsed:.0f}s timeout={self.timeout_seconds}s "
                     f"cc_pair={self.redis_connector.cc_pair_id}"
                 )


backend/onyx/background/celery/tasks/docprocessing/utils.py
```diff
@@ -51,11 +52,29 @@ def __init__(
...
+            return True
+
+        # Check if the task has exceeded its timeout
+        # NOTE: Celery's soft_time_limit does not work with thread pools,
+        # so we must enforce timeouts internally.
+        if self.timeout_seconds is not None:
+            elapsed = time.monotonic() - self.start_monotonic
+            if elapsed > self.timeout_seconds:
+                logger.warning(
+                    f"{self.__class__.__name__} - task timeout exceeded: "
```
evan-onyx: probably just docprocessing here instead of __class__.__name__?

PR #8419, opened 2026-02-13
feat(scim): add SCIM database models and migration

nmgarza5:
## Description

Adds three new SQLAlchemy models (`ScimToken`, `ScimUserMapping`, `ScimGroupMapping`) and an Alembic migration to support SCIM 2.0 automated user/group provisioning from identity providers (Okta, Azure AD).

- **ScimToken**: Bearer tokens for IdP SCIM authentication (SHA256 hashed, display suffix, active flag, audit timestamps)
- **ScimUserMapping**: Maps SCIM `externalId` from IdP to Onyx `User` (UUID FK, unique indexed)
- **ScimGroupMapping**: Maps SCIM `externalId` from IdP to Onyx `UserGroup` (int FK, unique indexed)

Linear: https://linear.app/onyx-app/issue/ENG-3640/add-scim-database-models-and-migration

## Additional Options

---
## Summary by cubic
Adds SCIM 2.0 database models and an Alembic migration to enable automated user and group provisioning from IdPs like Okta and Azure AD. Implements token-based SCIM auth and externalId mappings to Onyx users/groups, aligning with ENG-3640.

- **New Features**
  - ScimToken: name, SHA256 hashed_token (64 chars, unique), token_display (last 4), created_by_id (User FK, cascade), is_active, created_at (non-null, default now()), last_used_at.
  - ScimUserMapping: external_id (unique, indexed) → user_id (UUID FK, cascade), created_at, updated_at (auto-updates).
  - ScimGroupMapping: external_id (unique, indexed) → user_group_id (FK, cascade), created_at, updated_at (auto-updates).
  - Migration creates scim_token, scim_user_mapping, scim_group_mapping with indexes, unique constraints, onupdate for updated_at, and created_by_id FK on scim_token.

<sup>Written for commit a024e321cbed2e80085a3bc187da02c41743a069. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/db/models.py
+++ b/backend/onyx/db/models.py
@@ -4907,7 +4907,7 @@ class ScimToken(Base):
         Boolean, server_default=text("true"), nullable=False
     )
     created_at: Mapped[datetime.datetime] = mapped_column(
-        DateTime(timezone=True), server_default=func.now()
+        DateTime(timezone=True), server_default=func.now(), nullable=False
     )
     last_used_at: Mapped[datetime.datetime | None] = mapped_column(
         DateTime(timezone=True), nullable=True


nmgarza5:
@greptile

nmgarza5:
@cubic

backend/onyx/db/models.py
```diff
@@ -4877,3 +4877,90 @@ class BuildMessage(Base):
             "ix_build_message_session_turn", "session_id", "turn_index", "created_at"
         ),
     )
+
+
+"""
+SCIM 2.0 Provisioning Models (Enterprise Edition only)
```
justin-tahara, line 4883: I don't think there is anything special we need to do with the tables for EE right?

backend/onyx/db/models.py
```diff
@@ -4877,3 +4877,90 @@ class BuildMessage(Base):
...
+class ScimToken(Base):
+    """Bearer tokens for IdP SCIM authentication."""
+
+    __tablename__ = "scim_token"
+
+    id: Mapped[int] = mapped_column(Integer, primary_key=True)
+    name: Mapped[str] = mapped_column(String, nullable=False)
+    hashed_token: Mapped[str] = mapped_column(
+        String(64), unique=True, nullable=False
+    )  # SHA256 = 64 hex chars
```
justin-tahara, line 4897: Is there some encryption things we should do?

backend/onyx/db/models.py
nmgarza5: nah we are good

backend/onyx/db/models.py
nmgarza5: eh maybe. let me look into this a bit more. we are storing the encrypted hash already but maybe we want another layer here in the db

nmgarza5: We store a SHA-256 hash of the token (same approach as `ApiKey.hashed_api_key`). The raw token is only shown once at creation time and never persisted. Since the tokens are high-entropy random values (`secrets.token_urlsafe(48)`), SHA-256 is sufficient — salting/bcrypt would add overhead without meaningful benefit for random secrets (unlike passwords which are low-entropy). See `auth.py` on the `nik/eng-3642` branch for the implementation.

PR #8422, opened 2026-02-13
feat(scim): add SCIM PATCH operation handler with unit tests

nmgarza5:
## Description

Add a SCIM PATCH operation handler (`patch.py`) that processes RFC 7644 PATCH requests from identity providers. Handles `replace`, `add`, and `remove` operations for both User and Group resources — including Okta's `active: false` deprovisioning pattern and group membership modifications via filter-based member removal.

Also includes the SCIM Pydantic schemas (`models.py`) that the patch handler depends on.

Comprehensive unit tests covering user deactivation/reactivation, group member add/remove/replace, multi-operation patches, and edge cases.

Linear: https://linear.app/onyx-app/issue/ENG-3644

## Additional Options

---
## Summary by cubic
Adds SCIM PATCH handlers for Users and Groups with unit tests. Supports replace/add/remove operations, including active=false deprovisioning and group membership changes with filter-based removal; returns added/removed member IDs for DB updates. Addresses Linear ENG-3644.

- **New Features**
  - Implemented apply_user_patch and apply_group_patch; handlers are pure (no DB writes).
  - Supported paths: users — active, userName, externalId, name.givenName/familyName/formatted, displayName→name.formatted; groups — displayName, externalId; members add/remove via members[value eq "id"] and full members replace with duplicate prevention; case-insensitive paths and top-level replace/add dicts.
  - Broadened ScimPatchOperation.value to accept mixed-type dicts; tests cover deprovision/reactivation, membership add/remove/replace, multi-op patches, and invalid paths.

<sup>Written for commit eee1282ae19e1fc2b19fa4f3e87ac93c14a6e6bf. Summary will update on new commits.</sup>

### DIFF
--- a/backend/ee/onyx/server/scim/patch.py
+++ b/backend/ee/onyx/server/scim/patch.py
@@ -1,0 +1,256 @@
+"""SCIM PATCH operation handler (RFC 7644 §3.5.2).
+
+Identity providers use PATCH to make incremental changes to SCIM resources
+instead of replacing the entire resource with PUT. Common operations include:
+
+  - Deactivating a user: ``replace`` ``active`` with ``false``
+  - Adding group members: ``add`` to ``members``
+  - Removing group members: ``remove`` from ``members[value eq "..."]``
+
+This module applies PATCH operations to Pydantic SCIM resource objects and
+returns the modified result. It does NOT touch the database — the caller is
+responsible for persisting changes.
+"""
+
+from __future__ import annotations
+
+import re
+
+from ee.onyx.server.scim.models import ScimGroupResource
+from ee.onyx.server.scim.models import ScimPatchOperation
+from ee.onyx.server.scim.models import ScimPatchOperationType
+from ee.onyx.server.scim.models import ScimUserResource
+
+
+class ScimPatchError(Exception):
+    """Raised when a PATCH operation cannot be applied."""
+
+    def __init__(self, detail: str, status: int = 400) -> None:
+        self.detail = detail
+        self.status = status
+        super().__init__(detail)
+
+
+# Pattern for member removal path: members[value eq "user-id"]
+_MEMBER_FILTER_RE = re.compile(
+    r'^members\[value\s+eq\s+"([^"]+)"\]$',
+    re.IGNORECASE,
+)
+
+
+def apply_user_patch(
+    operations: list[ScimPatchOperation],
+    current: ScimUserResource,
+) -> ScimUserResource:
+    """Apply SCIM PATCH operations to a user resource.
+
+    Returns a new ``ScimUserResource`` with the modifications applied.
+    The original object is not mutated.
+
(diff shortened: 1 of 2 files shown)


backend/ee/onyx/server/scim/patch.py
```diff
@@ -0,0 +1,224 @@
...
+    value: str | bool | dict | list | None,
+    data: dict,
+) -> None:
+    """Set a single field on group data by SCIM path."""
+    if path == "displayname":
+        data["displayName"] = value
+    elif path == "externalid":
+        data["externalId"] = value
+    else:
+        raise ScimPatchError(f"Unsupported path '{path}' for Group PATCH")
```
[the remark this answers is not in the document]
nmgarza5: Fixed — added `_replace_members` helper and updated `_apply_group_replace` to handle the `members` path (both explicit and in no-path dict). Also added tests for replace-with-path and replace-with-empty-list. The `current_members` and tracking lists are now passed through correctly.

backend/ee/onyx/server/scim/patch.py
```diff
@@ -0,0 +1,256 @@
...
+        name_data["givenName"] = value
+    elif path == "name.familyname":
+        name_data["familyName"] = value
+    elif path == "name.formatted":
+        name_data["formatted"] = value
+    elif path == "displayname":
+        # Some IdPs send displayName on users; map to formatted name
+        name_data["formatted"] = value
+    else:
+        raise ScimPatchError(f"Unsupported path '{path}' for User PATCH")
```
justin-tahara, line 113: Can we just make sure this works with all providers? I know you are going through a ton of options but maybe mentioning which is used in what systems would be nice

backend/ee/onyx/server/scim/patch.py
justin-tahara, line 1: Just maybe triple check we aren't missing any other patch commands like importing other fields out of user info and group info. I don't fully know what common fields may be filled at larger orgs like org or branch or team so that kind of stuff would potentially fall under this file I believe

backend/ee/onyx/server/scim/patch.py
nmgarza5: i will be doing that in a future phase. I'm laying the foundation before I start adding provider specific logic. Don't worry though, I will ensure all providers (starting with Okta and Azure) are covered

backend/ee/onyx/server/scim/patch.py
nmgarza5: We might be, but thats also why I said this might be incomplete. Once I get to the [Azure](https://linear.app/onyx-app/issue/ENG-3652/azure-ad-scim-compatibility) / [Okta](https://linear.app/onyx-app/issue/ENG-3651/okta-scim-compatibility) compatibility phase, I will add anything that is missing

justin-tahara (approved):

PR #8420, opened 2026-02-13
feat(scim): add SCIM 2.0 Pydantic schemas

nmgarza5:
## Description

Adds all Pydantic schemas needed for SCIM 2.0 provisioning at `backend/ee/onyx/server/scim/models.py`.

**SCIM Protocol Schemas ([RFC 7643](https://datatracker.ietf.org/doc/html/rfc7643) / [RFC 7644](https://datatracker.ietf.org/doc/html/rfc7644)):**
- `ScimUserResource` / `ScimGroupResource` — User and Group wire format for IdP communication
- `ScimListResponse` — Paginated list envelope
- `ScimPatchRequest` / `ScimPatchOperation` — Incremental update operations (deactivate user, modify group members)
- `ScimError` — Standard error response
- `ScimServiceProviderConfig` — Capability advertisement (GET /scim/v2/ServiceProviderConfig)
- `ScimResourceType` — Resource type definitions (GET /scim/v2/ResourceTypes)

**Admin API Schemas (Onyx-internal):**
- `ScimTokenCreate` / `ScimTokenResponse` / `ScimTokenCreatedResponse` — SCIM bearer token management

Linear: https://linear.app/onyx-app/issue/ENG-3641/create-scim-pydantic-schemas

## Additional Options

- [x] [Required] I have considered whether this PR needs to be cherry-picked to the latest beta branch.
- [ ] [Optional] Override Linear Check

---
## Summary by cubic
Adds SCIM 2.0 Pydantic schemas for user and group provisioning and admin token management, enabling Okta/Azure AD integration per RFC 7643/7644. Implements ENG-3641 in backend/ee/onyx/server/scim/models.py.

- **New Features**
  - Protocol models: User, Group, ListResponse, Patch (request/operation), Error, ServiceProviderConfig, ResourceType.
  - Admin models: token create, token response, created response with raw token.
  - Wire format: camelCase fields, schema URIs, PATCH op enum, pagination defaults; aliased fields via ConfigDict(populate_by_name=True).

- **Bug Fixes**
  - ScimPatchOperation.value accepts dicts with bool values (e.g., {"active": false}) for Okta payloads.

<sup>Written for commit c4a87a58e0c6890b5bfee1307e9528f9eb1cb217. Summary will update on new commits.</sup>

PR #8424, opened 2026-02-13
feat(scim): add SCIM database CRUD operations (DAL)

nmgarza5:
## Description

Add the SCIM data access layer (`scim.py`) with CRUD operations for tokens, user mappings, and group mappings. Also includes a reusable base DAL class (`dal.py`) for common CRUD patterns.

Depends on #8419 (DB models) and #8420 (Pydantic schemas) — includes those commits as the DAL builds directly on both.

New files:
- `backend/ee/onyx/db/scim.py` — Token, user mapping, and group mapping CRUD
- `backend/onyx/db/dal.py` — Base DAL with generic CRUD helpers
- Unit tests for both modules

Linear: https://linear.app/onyx-app/issue/ENG-3645

## Additional Options

---
## Summary by cubic
Adds a SCIM Data Access Layer (ScimDAL) and shared base DAL with explicit session control and tenant-aware context. Implements CRUD for tokens, user mappings, and group mappings with SCIM 1-based pagination to satisfy ENG-3645.

- **New Features**
  - Tokens: create, list (newest first), revoke, lookup by hash, update last_used.
  - User mappings: create, get by external_id/user_id, list with 1-based pagination and totals, update external_id, delete.
  - Group mappings: create, get by external_id/group_id, list with 1-based pagination and totals, delete.

- **Refactors**
  - Introduced backend/onyx/db/dal.py base DAL and class-based ScimDAL in ee/onyx/db/scim.py.
  - Added unit tests and shared fixtures for mocked sessions and Postgres-backed tests.

<sup>Written for commit 6c780f8e74f416f8bd1ddc4e66dca93a7353d996. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/external_dependency_unit/db/conftest.py
+++ b/backend/tests/external_dependency_unit/db/conftest.py
@@ -1,0 +1,83 @@
+"""Fixtures for testing DAL classes against a real PostgreSQL database.
+
+These fixtures build on the db_session and tenant_context fixtures from
+the parent conftest (tests/external_dependency_unit/conftest.py).
+
+Requires a running Postgres instance. Run with::
+
+    python -m dotenv -f .vscode/.env run -- pytest tests/external_dependency_unit/db/
+"""
+
+from collections.abc import Callable
+from collections.abc import Generator
+from uuid import UUID
+from uuid import uuid4
+
+import pytest
+from sqlalchemy.orm import Session
+
+from ee.onyx.db.scim import ScimDAL
+from onyx.db.models import ScimToken
+from onyx.db.models import UserGroup
+
+
+@pytest.fixture
+def scim_dal(db_session: Session) -> ScimDAL:
+    """A ScimDAL backed by the real test database session."""
+    return ScimDAL(db_session)
+
+
+@pytest.fixture
+def scim_token_factory(
+    db_session: Session,
+) -> Generator[Callable[..., ScimToken], None, None]:
+    """Factory that creates ScimToken rows and cleans them up after the test."""
+    created_ids: list[int] = []
+
+    def _create(
+        name: str = "test-token",
+        hashed_token: str | None = None,
+        token_display: str = "onyx_scim_****test",
+        created_by_id: UUID | None = None,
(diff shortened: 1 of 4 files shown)


backend/onyx/db/models.py
```diff
@@ -4877,3 +4877,90 @@ class BuildMessage(Base):
...
+    created_by: Mapped[User] = relationship("User", foreign_keys=[created_by_id])
+
+
+class ScimUserMapping(Base):
+    """Maps SCIM externalId from the IdP to an Onyx User."""
+
+    __tablename__ = "scim_user_mapping"
+
+    id: Mapped[int] = mapped_column(Integer, primary_key=True)
+    external_id: Mapped[str] = mapped_column(String, unique=True, index=True)
```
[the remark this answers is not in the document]
nmgarza5: Valid observation, but this code lives in `models.py` which was added in PR #8419 (already merged). The redundant `index=True` will be cleaned up in a follow-up PR against main.

backend/ee/onyx/db/scim.py
```diff
@@ -0,0 +1,258 @@
...
+    if not token:
+        raise ValueError(f"SCIM token with id {token_id} not found")
+    token.is_active = False
+
+
+def update_scim_token_last_used(token_id: int, db_session: Session) -> None:
+    """Update the last_used_at timestamp for a token."""
+    token = db_session.get(ScimToken, token_id)
+    if token:
+        token.last_used_at = func.now()  # type: ignore[assignment]
```
acaprau: don't need to change anything abt this pr but for future reference if you want the row to show the last time it was changed i think you can just define an on_update when you define the orm object and not have to worry abt calling something like this.

acaprau (approved):

backend/ee/onyx/db/scim.py
nmgarza5: little different. in this case, we want to track the last time someone used the token and not when it was updated. this token should actually never be updated.

so on_update won't really work for this use case

PR #8430, opened 2026-02-13
fix(search): Improve Speed

justin-tahara:
## Description

Trying to reduce the latency of the search and optimize the current behavior.

## How Has This Been Tested?

Tested locally on my corpus and saw the average down down from 400ms to 150ms

## Additional Options

---
## Summary by cubic
Lowered default results and capped Vespa candidate hits to cut query latency. Local tests show average search time down from ~400ms to ~150ms.

- **Refactors**
  - Set default num_hits to 30 (was 50) across backend and web.
  - Document set filter now uses the user's selection; falls back to persona sets only if none provided.
  - Cap Vespa rerank candidates: target_hits = min(max(4 * num_to_retrieve, 100), RERANK_COUNT).

<sup>Written for commit e24ef138b4cb2213092c427a3f1bc05e5fbe92fd. Summary will update on new commits.</sup>


backend/onyx/document_index/vespa/vespa_document_index.py
```diff
@@ -553,10 +553,9 @@ def hybrid_retrieval(
         num_to_retrieve: int,
     ) -> list[InferenceChunk]:
         vespa_where_clauses = build_vespa_filters(filters)
-        # Needs to be at least as much as the rerank-count value set in the
-        # Vespa schema config. Otherwise we would be getting fewer results than
-        # expected for reranking.
-        target_hits = max(10 * num_to_retrieve, RERANK_COUNT)
+        # Avoid over-fetching a very large candidate set for global-phase reranking.
+        # Keep enough headroom for quality while capping cost on larger indices.
+        target_hits = min(max(4 * num_to_retrieve, 100), RERANK_COUNT)
```
acaprau, line 559: my only concern is that this will reduce recall for chat queries @yuhongsun96

PR #8457, opened 2026-02-14
chore(llm): Adding more FE Unit Tests

justin-tahara:
## Description

Going through and adding tests for relevant LLM related FE workflows

## How Has This Been Tested?

Running all of the tests and ensuring that they are functioning properly.

## Additional Options

---
## Summary by cubic
Adds FE unit tests for LLM provider/model resolution and the LLMPopover option builder, and extracts shared helpers for reuse. Improves provider/model collision handling, default fallbacks, and provider/vendor grouping.

- **Refactors**
  - Exported getDefaultLlmDescriptor and added getValidLlmDescriptorForProviders; useLlmManager now reuses this.
  - Exported buildLlmOptions and groupLlmOptions from LLMPopover and moved LLM option types to interfaces.ts to dedupe and simplify UI code.

- **New Tests**
  - hooks.llmResolver.test: provider-specific resolution on model name collisions, default fallback when model missing, first-available when no default.
  - useLLMProviders.test: selects correct endpoint (public vs persona), validates loading/error, exposes refetch.
  - LLMPopover.test: dedupes identical provider+model combos, includes hidden selected model, groups/sorts aggregator options by provider/vendor.

<sup>Written for commit f957e621d22f0bc283d8e1f35dd70e24f16403ae. Summary will update on new commits.</sup>


web/src/refresh-components/popovers/llmPopoverUtils.ts
nmgarza5, line 1: dont think we need this file. Any utility like `buildLlmOptions` should just go to the top of the main component file. (unless we are going to use it in many different places, then we can abstract to some kind of utility file) 

typically we would do something like:

```
refresh-components/popovers/
					| - LLMPopover.tsx
					| - LLMPopover.test.tsx
					| - interfaces.ts -> for all types used within the component (not props)
```

web/src/refresh-components/popovers/LLMPopover.test.ts
```diff
@@ -0,0 +1,121 @@
...
+): LLMProviderDescriptor {
+  return {
+    name: overrides.name ?? "Provider",
+    provider: overrides.provider ?? "openai",
+    default_model_name: overrides.default_model_name ?? "default-model",
+    is_default_provider: overrides.is_default_provider ?? false,
+    model_configurations: overrides.model_configurations ?? [],
+    ...overrides,
+  };
+}
```
nmgarza5: you are using this in two different files. Should probably just move to `https://github.com/onyx-dot-app/onyx/blob/main/web/tests/setup/test-utils.tsx`

web/src/lib/hooks.llmResolver.test.ts
```diff
@@ -0,0 +1,147 @@
+import {
+  getDefaultLlmDescriptor,
+  getValidLlmDescriptorForProviders,
+} from "@/lib/hooks";
+import { structureValue } from "@/lib/llm/utils";
+import { LLMProviderDescriptor } from "@/app/admin/configuration/llm/interfaces";
+
+function makeProvider(
```
nmgarza5: same comment as the other one

web/src/refresh-components/popovers/LLMPopover.test.ts
justin-tahara: `test-utils.tsx` pulls in `@testing-library/user-event` and that package has side effects at import time so just moved the function to a central file

nmgarza5 (approved):

PR #8464, opened 2026-02-14
Fix: CheckpointedConnector pruning only processes first checkpoint step

yvesgrolet:
## Bug

The pruning code for `CheckpointedConnector` in `extract_ids_from_runnable_connector()` only calls `load_from_checkpoint()` once with the initial dummy checkpoint. It never loops on `checkpoint.has_more`.

Multi-step checkpointed connectors like IMAP are affected: the first `load_from_checkpoint()` call with the dummy checkpoint only initializes internal state (e.g. fetches mailbox list) and yields **zero documents**. Without looping to subsequent checkpoint steps, pruning sees 0 source document IDs and proceeds to **delete all indexed documents** from Vespa.

## Impact

**All `CheckpointedConnector` implementations** are affected. For connectors where the first checkpoint step yields no documents (confirmed with IMAP), pruning deletes the entire document set on every run.

Confirmed with:
- IMAP connector on Exchange (OVH) — all emails deleted on prune
- IMAP connector on Exchange (Microsoft 365) — same behavior

## Fix

Replace the single `load_from_checkpoint()` call with a `while` loop that:

1. Calls `load_from_checkpoint(start, end, checkpoint)` 
2. Iterates the generator via `CheckpointOutputWrapper` to extract document IDs
3. Updates the checkpoint from the generator's return value
4. Breaks when `checkpoint.has_more == False`

This mirrors the checkpoint loop pattern already used by `ConnectorRunner.run()` in `connector_runner.py`.

The fix is a new private helper `_checkpointed_batched_doc_ids()` that yields batched document IDs across all checkpoint steps, plugging directly into the existing rate-limiting loop.

## Changes

- **1 file modified**: `backend/onyx/background/celery/celery_utils.py`
- **+49 / -7 lines**
- No new dependencies
- Consistent with existing code patterns (`CheckpointOutputWrapper`, rate limiting, callback support)

---
## Summary by cubic
Fixes pruning for CheckpointedConnector by looping through all checkpoint steps to collect document IDs across the entire run. Prevents deletions when the first step returns zero documents (seen with IMAP).

- **Bug Fixes**
  - Iterate until checkpoint.has_more is false; update checkpoint between steps.
  - Add _checkpointed_batched_doc_ids using CheckpointOutputWrapper; collect IDs from documents and failures; batch size 32.
  - Integrate with existing rate limiting to stop zero-ID prune deletions for IMAP and other checkpointed connectors.

<sup>Written for commit 5587e616c6ed0d22131d5453af6e1098512501cd. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/background/celery/celery_utils.py
+++ b/backend/onyx/background/celery/celery_utils.py
@@ -10,7 +10,7 @@ import httpx
 
 from onyx.configs.app_configs import MAX_PRUNING_DOCUMENT_RETRIEVAL_PER_MINUTE
 from onyx.configs.app_configs import VESPA_REQUEST_TIMEOUT
-from onyx.connectors.connector_runner import batched_doc_ids
+from onyx.connectors.connector_runner import CheckpointOutputWrapper
 from onyx.connectors.cross_connector_utils.rate_limit_wrapper import (
     rate_limit_builder,
 )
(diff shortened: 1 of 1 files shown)


evan-onyx:
Thanks for catching this! FWIW this was only an issue for checkpointedconnectors that did not also implement the SlimConnector interface. Arguably the IMAP connector should also implement SlimConnector, but certainly as a stopgap this is a good fix to get in. Here's the mirror PR I merged (we have to do this to run our CI) https://github.com/onyx-dot-app/onyx/pull/8468

PR #8465, opened 2026-02-15
fix(craft): ephemeral ACP clients to prevent multi-replica session corruption

rohoswagger:
## Summary

- **Root cause**: When multiple API server replicas handle messages for the same Craft session, each replica creates its own `opencode acp` process inside the same sandbox container. Both processes share flat file session storage (`~/.local/share/opencode/storage/`), causing stale in-memory state and silently lost JSON-RPC responses — making follow-up messages hang indefinitely.
- **Fix**: Use ephemeral ACP clients — each message spins up a fresh `opencode acp` process, resumes the session from disk via `session/list` + `session/resume`, handles the prompt, then stops. This ensures only one process operates on a session's files at a time.
- Simplified session management: single `resume_or_create_session()` entry point, removed redundant two-layer cache that was dead code with ephemeral clients (~100 lines removed).
- Added `session/cancel` on timeout, `GeneratorExit`, and exceptions so abandoned prompts don't leave the agent spinning.
- Accept either JSON-RPC response or `prompt_response` notification as turn completion (first one wins).

## Test plan

- [x] Verified in production (multi-replica K8s) — follow-up messages no longer hang
- [x] Confirm first message (session/new) and follow-up messages (session/resume) both work
- [x] Confirm timeout + cancel behavior when agent takes too long

## Additional Options

### DIFF
--- a/backend/onyx/server/features/build/sandbox/kubernetes/internal/acp_exec_client.py
+++ b/backend/onyx/server/features/build/sandbox/kubernetes/internal/acp_exec_client.py
@@ -29,6 +29,7 @@ from dataclasses import field
 from queue import Empty
 from queue import Queue
 from typing import Any
+from typing import cast
 
 from acp.schema import AgentMessageChunk
 from acp.schema import AgentPlanUpdate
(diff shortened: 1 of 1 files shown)


rohoswagger:
@greptile

rohoswagger:
@cubic-dev

PR #8495, opened 2026-02-17
fix(helm): Log Level Passthrough for Celery

justin-tahara:
## Description

Currently you are not able to override the log level of the celery pods so this is to ensure that we are able to update this use case.

## How Has This Been Tested?

Helm Template

## Additional Options

---
## Summary by cubic
Enable configurable Celery log levels in the Helm chart by passing logLevel values to all Celery components instead of hard-coding INFO. This lets operators tune verbosity per worker while keeping INFO as the default.

- **New Features**
  - Added logLevel to values for celery_beat and all celery_worker deployments (default: INFO).
  - Templated commands now use --loglevel from values.
  - Bumped chart version to 0.4.29.

<sup>Written for commit f4a14d52d89e1dacba8ffcd119a6f0fa35aa9f77. Summary will update on new commits.</sup>

### DIFF
--- a/deployment/helm/charts/onyx/Chart.yaml
+++ b/deployment/helm/charts/onyx/Chart.yaml
@@ -5,7 +5,7 @@ home: https://www.onyx.app/
 sources:
   - "https://github.com/onyx-dot-app/onyx"
 type: application
-version: 0.4.28
+version: 0.4.29
 appVersion: latest
 annotations:
   category: Productivity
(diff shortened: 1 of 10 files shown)


nmgarza5 (approved):
lgtm! 

for my own understanding, how do we actually override this? Do we run some helm command and pass `celery_beat.logLevel` as an arg?

justin-tahara:
You can do either pass in as arg or add to your override values.yaml file

PR #8498, opened 2026-02-17
fix(ee): small ux fixes for licensing

nmgarza5:
## Description

Three issues fixed for self-hosted enterprise customers with manually uploaded licenses (where `stripe_customer_id` is null):

1. **PlansView showed "Get Business Plan" instead of recognizing the enterprise plan**: When billing information fetch fails (no Stripe customer), `hasActiveSubscription()` returns false, so PlansView didn't recognize the customer's existing plan. Added a `hasLicense` prop to PlansView — Enterprise card now shows "Your Current Plan" and Business card shows "Included in your plan" (with checkout disabled) when a valid license exists.

2. **Brief "Access Restricted" flash after Redis license cache TTL expires**: The `/settings` endpoint used `get_cached_license_metadata()` which only reads from Redis. When the 24h TTL expired, it returned `None` → `GATED_ACCESS`, causing an access-restricted page flash until `/license` re-warmed the cache. Added a DB fallback (via `refresh_license_cache()`) when the cache is empty, matching the pattern already used by `get_license_metadata()`. Wrapped in nested `try/except SQLAlchemyError` to match the middleware pattern.

3. **`refresh_license_cache` overwrote `source` to `AUTO_FETCH` for manual licenses**: After cache TTL expiry, refreshing from DB always set `source=AUTO_FETCH`, which broke air-gap detection for manual license customers. Now derives `source` from the payload's `stripe_customer_id` — present means `AUTO_FETCH`, absent means `MANUAL_UPLOAD`.

## How Has This Been Tested?

Screenshot 2026-02-17 at 12 39 15 PM

## Additional Options

### DIFF
--- a/backend/ee/onyx/db/license.py
+++ b/backend/ee/onyx/db/license.py
@@ -263,9 +263,15 @@ def refresh_license_cache(
 
     try:
         payload = verify_license_signature(license_record.license_data)
+        # Derive source from payload: manual licenses lack stripe_customer_id
+        source: LicenseSource = (
+            LicenseSource.AUTO_FETCH
+            if payload.stripe_customer_id
+            else LicenseSource.MANUAL_UPLOAD
+        )
         return update_license_cache(
             payload,
-            source=LicenseSource.AUTO_FETCH,
+            source=source,
             tenant_id=tenant_id,
         )
     except ValueError as e:
--- a/backend/ee/onyx/server/settings/api.py
+++ b/backend/ee/onyx/server/settings/api.py
@@ -1,10 +1,13 @@
 """EE Settings API - provides license-aware settings override."""
 
 from redis.exceptions import RedisError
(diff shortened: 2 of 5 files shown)


nmgarza5:
@greptile

backend/ee/onyx/db/license.py
```diff
@@ -263,9 +263,15 @@ def refresh_license_cache(
 
     try:
         payload = verify_license_signature(license_record.license_data)
+        # Derive source from payload: manual licenses lack stripe_customer_id
```
nmgarza5, line 266: this is because when we generate them on the fly and send to customers, we don't create a stripe customer. This is really only relevant for enterprise self hosted customers that pay outside of stripe

nmgarza5:
@greptile

PR #8499, opened 2026-02-17
chore(playwright): hide toast elements in screenshots

jmelahman:
## Description

Toasts are ephemeral and animated with opacity which trigger false positives on screenshot diffs, so hide them by default.

## How Has This Been Tested?

TODO verify the visual regression report

## Additional Options

---
## Summary by cubic
Hide toast UI in Playwright screenshots to avoid false positives in visual diffs. Adds a default hide selector and a data-testid on the toast container.

- **Bug Fixes**
  - Added data-testid="toast-container" to ToastProvider.
  - Screenshot helpers now hide toasts by default and merge with per-call hide selectors.

<sup>Written for commit 534db103fc8fba66976d6f840e2535893f126766. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/app/app/message/HumanMessage.tsx
+++ b/web/src/app/app/message/HumanMessage.tsx
@@ -214,7 +214,7 @@ const HumanMessage = React.memo(function HumanMessage({
                 </Text>
               </div>
             </div>
-            {onEdit && !isEditing && (!files || files.length === 0) && (
+            {onEdit && !isEditing && (
               <div className="flex flex-row p-1 opacity-0 group-hover:opacity-100 transition-opacity">
                 <CopyIconButton
                   getCopyText={() => content}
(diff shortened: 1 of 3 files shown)


web/src/providers/ToastProvider.tsx
```diff
@@ -65,6 +65,7 @@ function ToastContainer() {
 
   return (
     <div
+      data-testid="toast-container"
```
evan-onyx: I believe guidance from the FE folks was to use aria label for this kind of thing

evan-onyx (approved):

web/src/providers/ToastProvider.tsx
jmelahman: I think this is converted to an aria-label by next

jmelahman: They are different, but I don't think an aria-label makes sense on this particular element -- this specific div -- or at least, this is good enough and the accessibility concerns are best addressed separately imo.

PR #8501, opened 2026-02-17
chore(playwright): hide toast elements in screenshots

jmelahman:
## Description

Had a `git switch $(git branch --show-current) && git reset --hard origin/main && git push -f` incident in https://github.com/onyx-dot-app/onyx/pull/8499

## How Has This Been Tested?

The regression report in that PR looked good.

## Additional Options

---
## Summary by cubic
Hide toast notifications in Playwright screenshots by default to prevent false diffs and flaky tests. Add data-testid="toast-container", merge default hide selectors in the screenshot utility, and log in before each welcome page test to ensure a valid session.

<sup>Written for commit 3ae53ffeee83532df7ae797578e8afcd8973f0dd. Summary will update on new commits.</sup>

### DIFF
--- a/web/tests/e2e/welcome_page.spec.ts
+++ b/web/tests/e2e/welcome_page.spec.ts
@@ -13,8 +13,7 @@ const THEMES = ["light", "dark"] as const;
 for (const theme of THEMES) {
   test.describe(`Welcome page — /app (${theme} mode)`, () => {
     test.beforeEach(async ({ page }) => {
-      // Always log in before each test to ensure a valid session, regardless of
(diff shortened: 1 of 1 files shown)


web/src/providers/ToastProvider.tsx
```diff
@@ -65,6 +65,7 @@ function ToastContainer() {
 
   return (
     <div
+      data-testid="toast-container"
```
evan-onyx, line 68: guidance from the FE folks is to use aria label instead of data-testid

PR #8506, opened 2026-02-17
chore(cursor): playwright SKILL.md

jmelahman:
## Description

Used cursor to generate a playwright skill.

## How Has This Been Tested?

20260217_14h18m56s_grim

## Additional Options

---
## Summary by cubic
Added a Playwright E2E testing skill (SKILL.md) with pre-auth storage states, API-first auth, and a clear default for isolation: prefer loginAsRandomUser for parallel-safe tests. Exposed the skills folder to tooling and updated locator guidance to prefer aria-label alongside data-testid; best practices now recommend focused comments where helpful.

- **New Features**
  - SKILL.md covers projects (admin/exclusive), global setup (health check, user provisioning, API login, storage states, default provider), locator/assertion/waiting strategies (data-testid and aria-label first), best practices (use focused comments), and utilities (OnyxApiClient, chatActions, visualRegression, tools); avoid UI-driven login.
  - Tests default to an admin storage state, but prefer loginAsRandomUser for most feature tests; API-based user switching supported; visual regression is opt-in via VISUAL_REGRESSION.

<sup>Written for commit f2503c6e603f743023ed206f30fdadb40f6767ca. Summary will update on new commits.</sup>

### DIFF
--- a/.cursor/skills/playwright/SKILL.md
+++ b/.cursor/skills/playwright/SKILL.md
@@ -17,7 +17,10 @@ description: Write and maintain Playwright end-to-end tests for the Onyx applica
 ## Running Tests
 
 ```bash
+# Run a specific test file
 npx playwright test web/tests/e2e/chat/default_assistant.spec.ts
+
+# Run a specific project
 npx playwright test --project admin
 npx playwright test --project exclusive
 ```
@@ -64,6 +67,7 @@ test.describe("Feature Name", () => {
   test("should describe expected behavior clearly", async ({ page }) => {
     await page.goto("/app");
     await page.waitForLoadState("networkidle");
+    // Already authenticated as admin — go straight to testing
   });
 });
 ```
@@ -150,10 +154,19 @@ Use locators in this priority order:
 Use web-first assertions — they auto-retry until the condition is met:
 
 ```typescript
+// Visibility
 await expect(page.getByTestId("onyx-logo")).toBeVisible({ timeout: 5000 });
+
+// Text content
 await expect(page.getByTestId("assistant-name-display")).toHaveText("My Assistant");
+
+// Count
 await expect(page.locator('[data-testid="onyx-ai-message"]')).toHaveCount(2, { timeout: 30000 });
+
+// URL
 await expect(page).toHaveURL(/chatId=/);
+
+// Element state
 await expect(toggle).toBeChecked();
 await expect(button).toBeEnabled();
 ```
@@ -163,13 +176,17 @@ await expect(button).toBeEnabled();
 ## Waiting Strategy
 
 ```typescript
+// Wait for load state after navigation
 await page.goto("/app");
 await page.waitForLoadState("networkidle");
 
+// Wait for specific element
 await page.getByTestId("chat-intro").waitFor({ state: "visible", timeout: 10000 });
 
+// Wait for URL change
 await page.waitForFunction(() => window.location.href.includes("chatId="), null, { timeout: 10000 });
 
+// Wait for network response
 await page.waitForResponse(resp => resp.url().includes("/api/chat") && resp.status() === 200);
 ```
 


.cursor/skills/playwright/SKILL.md
```diff
@@ -0,0 +1,209 @@
...
+```typescript
+import { test, expect } from "@playwright/test";
+import { loginAs, loginAsRandomUser } from "@tests/e2e/utils/auth";
+import { OnyxApiClient } from "@tests/e2e/utils/onyxApiClient";
+
+test.describe("Feature Name", () => {
+  test.beforeAll(async ({ browser }) => {
+    // One-time setup with admin privileges (e.g., create resources via API)
+    const ctx = await browser.newContext({ storageState: "admin_auth.json" });
+    const page = await ctx.newPage();
```
jmelahman: I think this might be handled in the global_setup already

.cursor/skills/playwright/SKILL.md
```diff
@@ -0,0 +1,209 @@
...
+### `tools` (`@tests/e2e/utils/tools`)
+
+- `TOOL_IDS` — centralized `data-testid` selectors for tool options
+- `openActionManagement(page)` — opens the tool management popover
+
+## Locator Strategy
+
+Use locators in this priority order:
+
+1. **`data-testid`** — preferred for Onyx components
```
jmelahman: Do we feel strongly against this? I somewhat agree with it. I understand the convenience of using accessibility/aria-labels when possible, but feel like a better default is to separate the two when it makes sense. We can audit accessibility separately.

.cursor/skills/playwright/SKILL.md
```diff
@@ -0,0 +1,208 @@
...
+1. **Descriptive test names** — clearly state expected behavior: `"should display greeting message when opening new chat"`
+2. **API-first setup** — use `OnyxApiClient` for backend state; reserve UI interactions for the behavior under test
+3. **User isolation** — tests that modify visible app state (sidebar, chat history) should run as `"user"` to avoid polluting the admin session used by other parallel tests. Always cleanup API-created resources in `afterAll`
+4. **DRY helpers** — extract reusable logic into `utils/` with JSDoc comments
+5. **No hardcoded waits** — use `waitFor`, `waitForLoadState`, or web-first assertions
+6. **Parallel-safe** — no shared mutable state between tests; use unique names with timestamps (`\`test-${Date.now()}\``)
+7. **Error context** — catch and re-throw with useful debug info (page text, URL, etc.)
+8. **Tag slow tests** — mark serial/slow tests with `@exclusive` in the test title
+9. **Visual regression** — use `expectScreenshot()` for UI consistency checks
+10. **No code comments** — avoid inline comments in test code; use descriptive names and JSDoc on helpers instead
```
jmelahman: I can see these being useful. Maybe we remove this one

.cursor/skills/playwright/SKILL.md
jmelahman: Oh, maybe we just say `data-testid` or `aria-label` as #1. I do think we should shy a way a bit from defaulting to only `aria-label` (just so test-only changes are focused as not to introduce application changes/bad accessibility for the sake of testing).

nmgarza5: I disagree in using test ids. Aria-labels allows you to get the component in Jest tests as well. When does it make sense to using them separately? I don't see why a test id would be preferred over an aria label but pitch your case and I will listen

.cursor/skills/playwright/SKILL.md
nmgarza5: i think remove or just reword. Some comments are good but most are noise like the below

AI agents love to add meaningless comments:
```
// click the modal button
await modalButton.click()
```

.cursor/skills/playwright/SKILL.md
nmgarza5: > as not to introduce application changes/bad accessibility for the sake of testing

fair point. send it

jmelahman: I think https://github.com/onyx-dot-app/onyx/pull/8499#discussion_r2818567102 is a valid example when `data-testid` is preferred. An aria-label on this specific element wouldn't make sense from an accessibility POV. 

I've since updated this to prefer either which I think mostly satisfies both sides.

PR #8507, opened 2026-02-17
feat(code interpreter): improve produced code artifacts

Danelegend:
## Description
Currently when you create a code artifact, an anchor is produced that redirects you to a another tab. There is no actual way to download a file. This introduces a change such that a popup display first comes up showcasing the code, with a download button present to actually download the content.

Screenshot 2026-02-19 at 3 36 17 PM

## How Has This Been Tested?
Manual + playwright tests

## Additional Options
closes https://linear.app/onyx-app/issue/ENG-3709/improve-code-artifact-display

---
## Summary by cubic
Open code and text artifacts in an in-app modal with preview and a real download button, replacing the old link-out behavior. Closes ENG-3709 by making artifact display clearer and downloads straightforward.

- **New Features**
  - Intercepts same-origin /api/chat/file links to open a viewer modal; external links open normally.
  - Detects code via linguist-languages and shows a syntax-highlighted CodeViewModal with header info (language, line count, size); other files open in a TextViewModal with inline preview and a Download button.

- **Refactors**
  - Added fetchChatFile and getCodeLanguage; introduced CodeViewModal; AppPage routes code vs text.
  - fetchChatFile uses HTTP cache and object URLs are revoked. Code component allows hiding the copy button. Playwright tests cover link interception, modal selection, previews, and downloads.

<sup>Written for commit 778c1931ce349913fd8049447cd72d40cc630ddd. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/sections/modals/CodeViewModal.tsx
+++ b/web/src/sections/modals/CodeViewModal.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useState, useEffect, useCallback, useMemo } from "react";
+import { useState, useEffect, useMemo } from "react";
 import { MinimalOnyxDocument } from "@/lib/search/interfaces";
 import Modal from "@/refresh-components/Modal";
 import Text from "@/refresh-components/texts/Text";
@@ -56,8 +56,8 @@ export default function CodeViewModal({
     } · ${fileSize}`;
   }, [fileContent, language, lineCount, fileSize]);
 
-  const fetchFile = useCallback(
-    async (signal?: AbortSignal) => {
+  useEffect(() => {
+    (async () => {
       setIsLoading(true);
       setLoadError(null);
       setFileContent("");
@@ -66,8 +66,7 @@ export default function CodeViewModal({
         presentingDocument.document_id;
 
       try {
-        const response = await fetchChatFile(fileIdLocal, signal);
-
+        const response = await fetchChatFile(fileIdLocal);
         const blob = await response.blob();
         const url = window.URL.createObjectURL(blob);
         setFileUrl((prev) => {
@@ -76,34 +75,15 @@ export default function CodeViewModal({
           }
           return url;
         });
-
-        const originalFileName =
-          presentingDocument.semantic_identifier || "document";
-        setFileName(originalFileName);
-
-        const text = await blob.text();
-        setFileContent(text);
-      } catch (error) {
-        if (signal?.aborted) {
-          return;
-        }
+        setFileName(presentingDocument.semantic_identifier || "document");
+        setFileContent(await blob.text());
+      } catch {
         setLoadError("Failed to load document.");
       } finally {
-        if (!signal?.aborted) {
-          setIsLoading(false);
-        }
+        setIsLoading(false);
       }
-    },
-    [presentingDocument]
-  );
-
-  useEffect(() => {
-    const controller = new AbortController();
-    fetchFile(controller.signal);
-    return () => {
-      controller.abort();
-    };
-  }, [fetchFile]);
+    })();
+  }, [presentingDocument]);
 
   useEffect(() => {
     return () => {
--- a/web/src/lib/chat/svc.ts
+++ b/web/src/lib/chat/svc.ts
@@ -6,15 +6,11 @@ const CHAT_FILE_PREFIX = "/api/chat/file";
  * The caller is responsible for consuming the body (e.g. `.blob()`,
  * `.text()`) since different consumers need different formats.
  */
-export async function fetchChatFile(
-  fileId: string,
-  signal?: AbortSignal
-): Promise<Response> {
+export async function fetchChatFile(fileId: string): Promise<Response> {
   const response = await fetch(
     `${CHAT_FILE_PREFIX}/${encodeURIComponent(fileId)}`,
     {
       method: "GET",
-      signal,
       cache: "force-cache",
     }
   );


web/src/lib/languages.ts
```diff
@@ -0,0 +1,40 @@
...
+
+for (const lang of Object.values(languages) as LinguistLanguage[]) {
+  if (lang.type !== "programming") continue;
+
+  const name = lang.name.toLowerCase();
+  for (const ext of lang.extensions ?? []) {
+    // First language to claim an extension wins
+    if (!extensionMap.has(ext)) {
+      extensionMap.set(ext, name);
+    }
```
nmgarza5, line 22: i guess this makes sense but any chance we will actually have conflicts here?

What if there are multiple versions of Python as an example:
- Python 2 <- appears first and gets the .py extension
- Python 3

web/src/lib/languages.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Returns the language name for a given file name, or null if it's not a
+ * recognised code file. Looks up by extension first, then by exact filename
+ * (e.g. "Dockerfile", "Makefile"). Runs in O(1).
+ */
+export function getCodeLanguage(name: string): string | null {
+  const lower = name.toLowerCase();
+  const ext = lower.match(/\.[^.]+$/)?.[0];
+  return (ext && extensionMap.get(ext)) ?? filenameMap.get(lower) ?? null;
```
nmgarza5, line 39: if returning null, do we set anything? "Unset" maybe? that is what google docs code blocks do

web/src/refresh-components/buttons/DownloadIconButton.tsx
nmgarza5, line 1: I don't think we need this component. Can you just use a regular OpalButton and pass the props accordingly for tooltip and  icon

web/src/lib/languages.ts
nmgarza5: ah nvm, i saw later on we set to plaintext

web/src/sections/modals/CodeViewModal.tsx
```diff
@@ -0,0 +1,221 @@
...
+        presentingDocument.document_id;
+
+      try {
+        const response = await fetch(
+          `/api/chat/file/${encodeURIComponent(fileIdLocal)}`,
+          {
+            method: "GET",
+            signal,
+            cache: "force-cache",
+          }
```
nmgarza5: we dont like to add raw fetches to the component files. typically we add api calls like this to either a SWR hook `useChatFile` if we need request caching in the browser or we add the fetches to some svc.ts file (service api file). see this as an example https://github.com/onyx-dot-app/onyx/blob/main/web/src/lib/billing/svc.ts

web/src/sections/modals/CodeViewModal.tsx
```diff
@@ -0,0 +1,221 @@
...
+  }, [fileUrl]);
+
+  const handleDownload = () => {
+    const link = document.createElement("a");
+    link.href = fileUrl;
+    link.download = fileName || presentingDocument.document_id;
+    document.body.appendChild(link);
+    link.click();
+    document.body.removeChild(link);
+  };
```
nmgarza5: seems weird that we need to create an anchor tag on the fly

web/src/sections/modals/CodeViewModal.tsx
```diff
@@ -0,0 +1,221 @@
...
+              </Section>
+            ) : loadError ? (
+              <Section padding={1}>
+                <Text text03 mainUiBody>
+                  {loadError}
+                </Text>
+              </Section>
+            ) : (
+              <MinimalMarkdown
+                content={`\`\`\`${language}\n${fileContent}\n\`\`\``}
```
nmgarza5, line 131: dafuq is this lol

web/src/lib/languages.ts
Danelegend: True, but seems to be more of an edge case where there are multiple names for the same extension (ie. Python is just python). The alternative is we write our own map of extension -> Language which feels more icky

web/src/sections/modals/CodeViewModal.tsx
Danelegend: Same pattern that is done here:
https://github.com/onyx-dot-app/onyx/blob/5b54687077636245e9e693fe8502685bfb886842/web/src/sections/modals/TextViewModal.tsx#L183

web/src/sections/modals/CodeViewModal.tsx
Danelegend: \```js
// file contents here...
\```

Markdown rendering. Couldn't find any other nice components that do code highlighting and stuff (except one of the craft ones)

web/src/sections/modals/CodeViewModal.tsx
Danelegend: Same pattern as here: https://github.com/onyx-dot-app/onyx/blob/5b54687077636245e9e693fe8502685bfb886842/web/src/sections/modals/TextViewModal.tsx#L102

I was considering a refactor of common logic but decided against it. Lmk if you think it's worth doing, while modernising some of the bits

web/src/sections/modals/CodeViewModal.tsx
```diff
@@ -0,0 +1,208 @@
...
+        if (!signal?.aborted) {
+          setIsLoading(false);
+        }
+      }
+    },
+    [presentingDocument]
+  );
+
+  useEffect(() => {
+    const controller = new AbortController();
```
nmgarza5: why do we need this abort controller?

Danelegend: I think the idea was to cancel the request on tear down to stop potential document override race condition, but looking closer that doesn't seem necessary so removed.

PR #8508, opened 2026-02-17
fix: open_url broken on non-normalized urls and enable web crawl tests

wenxi-onyx:
## Description
- Web search/open URL tools normalize web URLs by removing query params (e.g. foo.com/?page=1 -> foo.com/). 
- The merge step in open_url was still matching crawled results against the raw input URL, so successful crawls are dropped as misses when the URL was normalized. 
    - This PR fixes that by also matching on the normalized URL during merge
- Additionally enables firecrawl and pse playwright tests

## Additional Options

---
## Summary by cubic
Enabled Firecrawl and Google PSE Playwright tests by adding CI secrets and a data-testid for the API key input, updating e2e selectors to use it for a stable setup. Fixed open_url merge to match crawled results by normalized URL (query/fragment stripped) to avoid dropped content.

<sup>Written for commit f3c4597cd22bfc869705721ea84af9f028246a16. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/tools/tool_implementations/open_url/open_url_tool.py
+++ b/backend/onyx/tools/tool_implementations/open_url/open_url_tool.py
@@ -804,9 +804,7 @@ class OpenURLTool(Tool[OpenURLToolOverrideKwargs]):
             indexed_section = indexed_by_doc_id.get(doc_id) if doc_id else None
             # WebContent.link is normalized (query/fragment stripped). Match on the
             # same normalized form to avoid dropping successful crawl results.
-            crawled_section = crawled_by_url.get(url) or crawled_by_url.get(
-                normalize_web_content_url(url)
-            )
+            crawled_section = crawled_by_url.get(normalize_web_content_url(url))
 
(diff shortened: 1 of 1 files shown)


backend/onyx/tools/tool_implementations/open_url/open_url_tool.py
```diff
@@ -801,7 +802,11 @@ def _merge_indexed_and_crawled_results(
         for url in all_urls:
             doc_id = url_to_doc_id.get(url)
             indexed_section = indexed_by_doc_id.get(doc_id) if doc_id else None
-            crawled_section = crawled_by_url.get(url)
+            # WebContent.link is normalized (query/fragment stripped). Match on the
+            # same normalized form to avoid dropping successful crawl results.
```
jessicasingh7, line 806: I would switch the order of first checking normalized version then fallback to original url, if that's how they are stored

jessicasingh7 (changes requested):

PR #8510, opened 2026-02-17
feat(craft): shareable webapp URLs with two-tier access control

rohoswagger:
## Description

Adds the ability to share Craft webapp URLs with configurable access control. Session owners can choose who can view their generated webapp:

- **Private** — owner only (default)
- **Organization** — any authenticated user on the Onyx instance
- **Public** — anyone with the link, no login required

**Backend:**
- Adds `sharing_scope` column to `build_session` (replaces `is_public` bool) with values `private` / `public_org` / `public_global`
- `_check_webapp_access` enforces scope: global → no auth; org → authenticated; private → owner only
- `PATCH /sessions/{session_id}/public` accepts `sharing_scope` field
- Proxy routes bypass auth middleware (access enforced in handler)
- Removed dead `/_next/{path:path}` route — asset paths are already rewritten by `_rewrite_asset_paths()`
- Offline page matches the default web template terminal UI (shown on 502/503/504)
- Auto-restart of Next.js on local backend when sandbox is running but port is dead

**Frontend:**
- `ShareButton` shows a 3-option popover (Private / Organization / Public) with copy-link row when shared
- `SharingScope` type added to `streamingTypes.ts`; `setSessionSharing` replaces `setSessionPublic`
- `ShareButton` moved into `UrlBar` for cleaner prop flow

## How Has This Been Tested?

Tested manually on dev deployment (`experimental-cc4a.72`):
- Sharing toggle persists across page refresh
- Unauthenticated access works for public_global sessions
- Unauthenticated access returns 401 for public_org sessions
- Private sessions return 404 for non-owners

## Additional Options

### DIFF
--- a/backend/onyx/server/features/build/api/api.py
+++ b/backend/onyx/server/features/build/api/api.py
@@ -33,7 +33,6 @@ from onyx.server.features.build.api.rate_limit import get_user_rate_limit_status
 from onyx.server.features.build.api.sessions_api import router as sessions_router
 from onyx.server.features.build.api.user_library import router as user_library_router
 from onyx.server.features.build.db.sandbox import get_sandbox_by_user_id
-from onyx.server.features.build.db.sandbox import update_sandbox_heartbeat
 from onyx.server.features.build.sandbox import get_sandbox_manager
 from onyx.server.features.build.session.manager import SessionManager
 from onyx.server.features.build.utils import is_onyx_craft_enabled
@@ -277,8 +276,8 @@ REWRITABLE_CONTENT_TYPES = {
 }
 
 
-def _get_sandbox_url(session_id: UUID, db_session: Session) -> tuple[str, UUID]:
-    """Get the internal URL and sandbox ID for a session's Next.js server.
+def _get_sandbox_url(session_id: UUID, db_session: Session) -> str:
+    """Get the internal URL for a session's Next.js server.
 
     Uses the sandbox manager to get the correct URL for both local and
     Kubernetes environments.
@@ -288,7 +287,7 @@ def _get_sandbox_url(session_id: UUID, db_session: Session) -> tuple[str, UUID]:
         db_session: Database session
 
     Returns:
-        Tuple of (internal URL to proxy requests to, sandbox ID for heartbeat)
+        Internal URL to proxy requests to
 
     Raises:
         HTTPException: If session not found, port not allocated, or sandbox not found
@@ -307,18 +306,14 @@ def _get_sandbox_url(session_id: UUID, db_session: Session) -> tuple[str, UUID]:
         raise HTTPException(status_code=404, detail="Sandbox not found")
 
     sandbox_manager = get_sandbox_manager()
-    return sandbox_manager.get_webapp_url(sandbox.id, session.nextjs_port), sandbox.id
+    return sandbox_manager.get_webapp_url(sandbox.id, session.nextjs_port)
 
 
 def _proxy_request(
     path: str, request: Request, session_id: UUID, db_session: Session
 ) -> StreamingResponse | Response:
-    """Proxy a request to the sandbox's Next.js server.
-
-    Updates the sandbox heartbeat on success to keep the sandbox alive
-    while the webapp is being viewed.
-    """
-    base_url, sandbox_id = _get_sandbox_url(session_id, db_session)
+    """Proxy a request to the sandbox's Next.js server."""
+    base_url = _get_sandbox_url(session_id, db_session)
 
     # Build the target URL
     target_url = f"{base_url}/{path.lstrip('/')}"
@@ -360,15 +355,12 @@ def _proxy_request(
                     media_type=content_type,
                 )
 
-            result: StreamingResponse | Response = StreamingResponse(
+            return StreamingResponse(
                 content=_stream_response(response),
                 status_code=response.status_code,
                 headers=response_headers,
                 media_type=content_type or None,
             )
-
-        update_sandbox_heartbeat(db_session, sandbox_id)
-        return result
 
     except httpx.TimeoutException:
         logger.error(f"Timeout while proxying request to {target_url}")


backend/onyx/server/features/build/api/api.py
```diff
@@ -365,71 +369,173 @@ def _proxy_request(
...
+    return session
+
+
+def _offline_html_response() -> Response:
+    """Return a branded Craft HTML page when the sandbox is not reachable.
+
+    Design mirrors the default Craft web template (outputs/web/app/page.tsx):
+    terminal window aesthetic with Minecraft-themed typing animation.
+    """
+    html = """<!DOCTYPE html>
```
wenxi-onyx: no inline html as string bruh

backend/onyx/server/features/build/api/api.py
```diff
@@ -365,71 +370,176 @@ def _proxy_request(
...
+def _check_webapp_access(
+    session_id: UUID, user: User | None, db_session: Session
+) -> BuildSession:
+    """Check if user can access a session's webapp.
 
+    - public_global: accessible by anyone (no auth required)
+    - public_org: accessible by any authenticated user
+    - private: only accessible by the session owner
+    """
+    session = db_session.get(BuildSession, session_id)
```
wenxi-onyx, line 382: have cursor 1 shot simplify this logic (can have fewer ifs for sure)

backend/onyx/server/features/build/api/api.py
```diff
@@ -365,71 +370,176 @@ def _proxy_request(
...
-    match = re.search(r"/api/build/sessions/([a-f0-9-]+)/webapp", referer)
-    if match:
-        try:
-            return UUID(match.group(1))
-        except ValueError:
-            return None
-    return None
+    _check_webapp_access(session_id, user, db_session)
+    try:
+        return _proxy_request("", request, session_id, db_session)
```
wenxi-onyx: might be nice to update the sandbox heartbeat here?

backend/onyx/server/features/build/api/api.py
```diff
@@ -365,71 +370,176 @@ def _proxy_request(
...
+    except HTTPException as e:
+        if e.status_code in (502, 503, 504):
+            return _offline_html_response()
```
wenxi-onyx: might be nice to create a notification here for the owner (that someone tried to access their shared webapp, but their sandbox is asleep) --> could get annoying tho

backend/onyx/server/features/build/api/api.py
```diff
@@ -365,71 +370,176 @@ def _proxy_request(
...
+        raise
 
 
-@nextjs_assets_router.get("/_next/{path:path}", response_model=None)
-def get_nextjs_assets(
+@public_build_router.get(
+    "/sessions/{session_id}/webapp/{path:path}", response_model=None
+)
+def get_webapp_path(
```
wenxi-onyx: Seems sus that this and the above have the exact same code. There's probably a way to combo these two routes onto one func?

backend/onyx/server/features/build/sandbox/local/local_sandbox_manager.py
```diff
@@ -766,6 +778,87 @@ def session_workspace_exists(
...
+        process_key = (sandbox_id, session_id)
+
+        with self._nextjs_lock:
+            # Already have a live tracked process
+            existing = self._nextjs_processes.get(process_key)
+            if existing is not None and existing.poll() is None:
+                return
+
+            # Already being started by another thread
+            if process_key in self._nextjs_starting:
```
wenxi-onyx: is this possible with the lock?

backend/onyx/server/features/build/sandbox/local/local_sandbox_manager.py
```diff
@@ -766,6 +778,87 @@ def session_workspace_exists(
...
+                return
+
+            # Check if the port is already responding (orphan process that survived restart)
+            try:
+                import httpx
+
+                with httpx.Client(timeout=1.0) as client:
```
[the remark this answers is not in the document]
wenxi-onyx: valid?

backend/onyx/server/features/build/sandbox/local/local_sandbox_manager.py
```diff
@@ -766,6 +778,87 @@ def session_workspace_exists(
...
+            if existing is not None and existing.poll() is None:
+                return
+
+            # Already being started by another thread
+            if process_key in self._nextjs_starting:
+                return
+
+            # Check if the port is already responding (orphan process that survived restart)
+            try:
+                import httpx
```
wenxi-onyx: import top of file

backend/onyx/server/features/build/sandbox/local/local_sandbox_manager.py
```diff
@@ -766,6 +778,87 @@ def session_workspace_exists(
...
+                pass  # Port is dead; proceed to restart
+
+            self._nextjs_starting.add(process_key)
+
+        logger.info(
+            f"Scheduling background Next.js restart for session {session_id} "
+            f"on port {nextjs_port}"
+        )
+
+        def _start_in_background() -> None:
```
wenxi-onyx: Maybe could look at threadpool_concurrency.py that has a lot of utils for stuff like this.

This func in general seems extremely sussy. The lock is not held for the actual `start_nextjs_server` and only held for setting the `[process_key]`, which seems wrong. Shouldn't the entire try hold the lock and `finally` release the lock? I could be misunderstanding this piece of code

web/src/app/craft/components/output-panel/UrlBar.tsx
```diff
@@ -38,6 +38,8 @@ export interface UrlBarProps {
   isDownloading?: boolean;
   /** Optional refresh callback — shows a refresh icon at the right edge of the URL pill */
   onRefresh?: () => void;
+  /** Optional share button node — shown on the far right of the URL bar */
+  shareButton?: React.ReactNode;
```
wenxi-onyx: Why is this a react node that is passed in as a prop? Shouldn't it be inlined with conditional rendering logic (if webapp &&)

web/src/app/craft/components/OutputPanel.tsx
```diff
@@ -615,6 +616,20 @@ const BuildOutputPanel = memo(({ onClose, isOpen }: BuildOutputPanelProps) => {
         onDownload={isMarkdownPreview ? handleDocxDownload : undefined}
         isDownloading={isExportingDocx}
         onRefresh={handleRefresh}
+        shareButton={
```
wenxi-onyx: related to above --> can we just inline this in urlbar?

web/src/app/craft/components/ShareButton.tsx
```diff
@@ -0,0 +1,181 @@
...
+  description: string;
+}[] = [
+  {
+    value: "private",
+    label: "Private",
+    description: "Only you can view this app.",
+  },
+  {
+    value: "public_org",
+    label: "Org",
```
wenxi-onyx: Pref `Organization`

web/src/app/craft/components/ShareButton.tsx
```diff
@@ -0,0 +1,181 @@
...
+        success = document.execCommand("copy");
+        document.body.removeChild(el);
+      } catch {}
+    }
+    setCopyState(success ? "copied" : "error");
+    setTimeout(() => setCopyState("idle"), 2000);
+  };
+
+  return (
+    <div className="relative flex-shrink-0" ref={popoverRef}>
```
wenxi-onyx: I think there's a pre-made/refreshed popover component you should use here. It handles things like "close when clicking outside the popover" which doesn't work currently here

wenxi-onyx:

backend/onyx/server/features/build/api/api.py
rohoswagger: too much work and not worth it atp

backend/onyx/server/features/build/api/api.py
```diff
@@ -365,71 +372,75 @@ def _proxy_request(
...
-        raise HTTPException(
-            status_code=400,
-            detail="Could not determine session from request context",
-        )
-    return _proxy_request(f"_next/{path}", request, session_id, db_session)
+    _check_webapp_access(session_id, user, db_session)
+    try:
+        result = _proxy_request(path, request, session_id, db_session)
+        # Update sandbox heartbeat — webapp access is user activity that keeps sandbox alive
+        session = db_session.get(BuildSession, session_id)
```
wenxi-onyx: I think we generally pref not doing db operations directly in api files --> i.e you would call some func from a db/file that does the db operations

wenxi-onyx: One option could be placing the heartbeat inside one of the downstream functions of `_check_webapp_access` or `_proxy_request` (i.e. something that retrieves web app details)

rohoswagger:
@greptile

rohoswagger:
@cubic-dev

rohoswagger:
### Code review

Found 1 issue:

1. **`update_sandbox_heartbeat` is never called for HTML/CSS/JS responses** — The heartbeat call is placed after the `with httpx.Client(...)` block exits, but the early `return Response(...)` for rewritable content types (HTML, CSS, JS) exits before reaching it. Since a webapp's root path always returns `text/html`, the heartbeat is skipped on every normal page view. Only binary assets would ever trigger it, which defeats the purpose of keeping the sandbox alive while the webapp is being viewed.

https://github.com/onyx-dot-app/onyx/blob/71e7a16e74929952e0df1bc40642f7b8965a90f2/backend/onyx/server/features/build/api/api.py#L350-L371

Fix: call `update_sandbox_heartbeat` before the early return (or immediately after the `with` block closes, unconditionally).

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #8519, opened 2026-02-18
fix(ollama): Cleaning up DeepSeek

justin-tahara:
## Description

DeepSeek on Ollama seems to have regressed. DeepSeek sends back XML formatted function calls which is something our FE does not know how to handle to pass back to the backend calls. 

This causes the FE to see something like: 
Screenshot 2026-02-17 at 4 58 01 PM

This is not usable. 

The PR aims to fix DeepSeek by ensuring we handle XML responses and also filtering XML strings from appearing from the frontend when they do get outputted. 

## How Has This Been Tested?

Added tests as well as testing locally with many different chats even asking it to format XML which it did fine.

## Additional Options

---
## Summary by cubic
Fixes DeepSeek on Ollama by parsing XML-style function calls and hiding them from the UI. Tool calls now run reliably, and XML payloads no longer appear in chat.

- **Bug Fixes**
  - Fallback extracts tool calls from JSON first, then XML <invoke> blocks, scanning answer, raw_answer, and reasoning.
  - Triggers fallback when no tool calls are returned (REQUIRED), and for AUTO when XML-style payload hints are detected.
  - Strips <function_calls> blocks from streamed content (handles split chunks, ignores lookalike tags) so the frontend only sees plain text.
  - Stores raw LLM text on LlmStepResult and adds tests for XML extraction, AUTO/REQUIRED fallback, and the streaming filter.

<sup>Written for commit 7091c1e3c87c0e550e9a365047193f5b1aeb41e9. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/chat/llm_loop.py
+++ b/backend/onyx/chat/llm_loop.py
@@ -138,7 +138,6 @@ def _try_fallback_tool_extraction(
         _looks_like_xml_tool_call_payload(llm_step_result.answer)
         or _looks_like_xml_tool_call_payload(llm_step_result.raw_answer)
         or _looks_like_xml_tool_call_payload(llm_step_result.reasoning)
-        or _looks_like_xml_tool_call_payload(llm_step_result.raw_reasoning)
     )
     should_try_fallback = (
         (tool_choice == ToolChoiceOptions.REQUIRED and no_tool_calls)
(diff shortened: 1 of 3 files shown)


justin-tahara:

justin-tahara:

yuhongsun96 (approved):

PR #8534, opened 2026-02-18
feat(opal): Add Content layout component

raunakab:
## Description

- Add `Content` component with two-axis architecture (`sizePreset` × `variant`) that routes to `HeadingLayout`, `LabelLayout`, or `BodyLayout`
- **HeadingLayout**: `headline`/`section` presets with `heading` (icon top) and `section` (icon inline) variants, editable titles
- **LabelLayout**: `main-content`/`main-ui`/`secondary` presets with `optional`, `auxIcon`, `tag`, and `editable` accessories
- **BodyLayout**: `inline`/`vertical`/`reverse` orientations with `default`/`muted` prominence
- Add `@opal/layouts` as new top-level import path (alongside `@opal/components`, `@opal/core`, `@opal/icons`)
- Add `2xs` button size for secondary preset edit buttons
- Auto-grow edit input on both HeadingLayout and LabelLayout
- Full storybook coverage for all layout combinations

**Depends on:** #8533

## How Has This Been Tested?

- `bun run build` passes
- `/storybook` renders all HeadingLayout, LabelLayout, and BodyLayout variants
- Editable title auto-grows at all size presets

## Additional Options


web/lib/opal/src/components/buttons/Button/components.tsx
```diff
@@ -15,13 +15,15 @@ const iconPaddingInRemVariants = {
...
   fit: "p-0.5",
 } as const;
 const iconSizeInRemVariants = {
   lg: 1,
   md: 1,
   sm: 1,
   xs: 0.75,
+  "2xs": 0.75,
   fit: 1,
 } as const;
```
Subash-Mohan: Since both of them use the same keys to access the values, shall we do it the following way? That would make it more cleaner.
```tsx
With a single merged object keyed by size, each value holding { padding, size }:

 const iconVariants = {
   lg:    { padding: "p-0.5", size: 1 },
   md:    { padding: "p-0.5", size: 1 },
   sm:    { padding: "p-0",   size: 1 },
   xs:    { padding: "p-0.5", size: 0.75 },
   "2xs": { padding: "p-0",   size: 0.75 },
   fit:   { padding: "p-0.5", size: 1 },
 } as const;

File: web/lib/opal/src/core/interactive/components.tsx [hunk 2]
```diff
@@ -54,21 +55,24 @@ const interactiveContainerHeightVariants = {
...
   fit: "",
 } as const;
 const interactiveContainerPaddingVariants = {
   lg: "p-2",
   md: "p-1",
   sm: "p-1",
   xs: "p-0.5",
-  fit: "",
+  "2xs": "p-0.5",
+  fit: "p-0",
```
Subash-Mohan: shall we do the same here as well?

web/lib/opal/src/layouts/Content/BodyLayout.tsx
```diff
@@ -0,0 +1,121 @@
...
+          />
+        </div>
+      )}
+
+      <span
+        className={`opal-content-body-title ${config.titleFont} ${titleColorClass}`}
+        style={{ height: config.lineHeight }}
+      >
+        {title}
+      </span>
```
Subash-Mohan, line 118: Are we moving away from Text component ?

web/lib/opal/src/layouts/Content/HeadingLayout.tsx
```diff
@@ -0,0 +1,193 @@
...
+      <div className="opal-content-heading-body">
+        <div className="opal-content-heading-title-row">
+          {editing ? (
+            <div className="opal-content-heading-input-sizer">
+              <span
+                className={`opal-content-heading-input-mirror ${config.titleFont}`}
+              >
+                {editValue || "\u00A0"}
+              </span>
+              <input
```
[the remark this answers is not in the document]
raunakab: Fixed in a52f6c930. `editValue` is now reset to `title` before `setEditing(false)`, so if `onBlur` → `commit()` fires, it sees `value === title` and skips calling `onTitleChange`.

web/lib/opal/src/layouts/Content/LabelLayout.tsx
```diff
@@ -0,0 +1,261 @@
...
+                onChange={(e) => setEditValue(e.target.value)}
+                size={1}
+                autoFocus
+                onFocus={(e) => e.currentTarget.select()}
+                onBlur={commit}
+                onKeyDown={(e) => {
+                  if (e.key === "Enter") commit();
+                  if (e.key === "Escape") setEditing(false);
+                }}
+                style={{ height: config.lineHeight }}
```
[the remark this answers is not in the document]
raunakab: Fixed in a52f6c930 — same approach as HeadingLayout: `editValue` is reset to `title` before closing, neutralizing the `onBlur` → `commit()` path.

web/lib/opal/src/layouts/Content/components.tsx
```diff
@@ -0,0 +1,141 @@
...
+};
+
+type ContentProps = HeadingContentProps | LabelContentProps | BodyContentProps;
+
+// ---------------------------------------------------------------------------
+// Content — routes to the appropriate internal layout
+// ---------------------------------------------------------------------------
+
+function Content(props: ContentProps) {
+  const { sizePreset = "headline", variant = "heading", ...rest } = props;
```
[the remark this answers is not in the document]
raunakab: This is intentional. When `sizePreset` is `"main-content"`, `"main-ui"`, or `"secondary"`, the `LabelLayout` is the correct default — it's the most common use case for those presets. The discriminated union type system enforces valid combinations at compile time: `LabelContentProps` restricts `variant` to `"section"` (or omitted), so TypeScript prevents invalid combos like `sizePreset="main-ui" variant="heading"` from being passed to `Content`.

The `variant === "heading"` branch in the router exists only as a runtime fallback for the default value — it correctly routes to `LabelLayout` since that's what label-sized presets should render.

web/lib/opal/src/layouts/Content/HeadingLayout.tsx
```diff
@@ -0,0 +1,193 @@
...
+  function commit() {
+    const value = editValue.trim();
+    if (value && value !== title) onTitleChange?.(value);
+    setEditing(false);
+  }
+
+  return (
+    <div
+      className="opal-content-heading"
+      data-icon-placement={iconPlacement}
```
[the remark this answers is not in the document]
raunakab: Fixed in a52f6c930. All template literal className concatenations across HeadingLayout, LabelLayout, and BodyLayout have been converted to use `cn()`.

web/lib/opal/src/layouts/Content/HeadingLayout.tsx
```diff
@@ -0,0 +1,179 @@
...
+    const value = inputRef.current.value.trim();
+    if (value && value !== title) onTitleChange?.(value);
+    setEditing(false);
+  }
+
+  return (
+    <div
+      className="opal-content-heading"
+      data-icon-placement={iconPlacement}
+      style={{ gap: iconPlacement === "left" ? config.gap : undefined }}
```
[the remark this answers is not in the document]
raunakab: This is intentional. When the icon is placed on top (\`variant="heading"\`), the icon container and body stack vertically inside a flex-column layout. The vertical spacing between the icon and body is controlled by CSS in the \`.opal-content-heading[data-icon-placement="top"]\` rule, not by the inline gap. The preset \`gap\` value is specifically for inline placement where the icon sits beside the title row.

Applying gap to both placements would double-space the top variant since the CSS already handles that case.

web/lib/opal/src/components/buttons/Button/components.tsx
raunakab: Good suggestion — merging the maps into a single object keyed by size would be cleaner. This is an existing pattern in Button that predates this PR though, so I'd prefer to address it in a separate refactor to keep this PR focused on the Content layout.

web/lib/opal/src/layouts/Content/BodyLayout.tsx
raunakab: Not moving away from Text — the \`Content\` layout component is a lower-level primitive in the opal design library (\`web/lib/opal/\`) that predates and sits beneath the app-level \`Text\` component in \`web/src/refresh-components/\`. It renders raw \`<span>\` elements with Tailwind font classes (like \`font-main-ui-action\`, \`font-heading-h2\`) to stay self-contained within the opal package without depending on app-level components.

App-level code should still use the \`Text\` component for general text rendering.

raunakab:
@Subash-Mohan this PR implements the "ContentContainer" component that @duo-onyx mapped out for us here: https://www.figma.com/design/Fms9uHkizpBuDEfsB8ujCc/Onyx-UI-Library?node-id=6195-195728&m=dev.

web/lib/opal/src/components/buttons/Button/components.tsx
raunakab: Done in 122f380b7 — merged `iconPaddingInRemVariants` + `iconSizeInRemVariants` into a single `iconVariants` object.

web/lib/opal/src/core/interactive/components.tsx
raunakab: Done in 122f380b7 — merged the three separate maps into a single `interactiveContainerSizeVariants` object with `{ height, minWidth, padding }` per key.

PR #8538, opened 2026-02-18
feat(scim): add admin SCIM token management API

nmgarza5:
## Description

Closes [ENG-3649](https://linear.app/onyx-app/issue/ENG-3649)
Closes [ENG-3650](https://linear.app/onyx-app/issue/ENG-3650)

Add admin endpoints for SCIM bearer token management, registered on the existing `enterprise_settings_admin_router`:

- **GET /admin/enterprise-settings/scim/token** — Return the active token's metadata (name, created date, last used) or 404 if none exists
- **POST /admin/enterprise-settings/scim/token** — Create a new token (auto-revokes all previous tokens; raw value returned once)

Design decisions:
- Only one active token at a time — creating a new token is the rotation mechanism
- No list/revoke/delete endpoints — to disable SCIM, customers revoke provisioning on the IdP side
- Revocation logic is inlined in `ScimDAL.create_token()`
- Uses standard Onyx admin session auth, not SCIM bearer tokens
- Endpoints live on `enterprise_settings_admin_router` alongside other EE admin settings (no separate SCIM admin router)

## Additional Options

### DIFF
--- a/backend/ee/onyx/db/scim.py
+++ b/backend/ee/onyx/db/scim.py
@@ -8,11 +8,11 @@ Usage from FastAPI::
     def get_scim_dal(db_session: Session = Depends(get_session)) -> ScimDAL:
         return ScimDAL(db_session)
 
-    @router.get("/tokens")
-    def list_tokens(dal: ScimDAL = Depends(get_scim_dal)) -> ...:
-        tokens = dal.list_tokens()
+    @router.post("/tokens")
+    def create_token(dal: ScimDAL = Depends(get_scim_dal)) -> ...:
+        token = dal.create_token(name=..., hashed_token=..., ...)
         dal.commit()
-        return tokens
+        return token
 
 Usage from background tasks::
 
@@ -53,7 +53,20 @@ class ScimDAL(DAL):
         token_display: str,
         created_by_id: UUID,
     ) -> ScimToken:
-        """Create a new SCIM bearer token."""
+        """Create a new SCIM bearer token.
+
+        Only one token is active at a time — this method automatically revokes
+        all existing active tokens before creating the new one.
+        """
+        # Revoke any currently active tokens
+        active_tokens = list(
+            self._session.scalars(
+                select(ScimToken).where(ScimToken.is_active.is_(True))
+            ).all()
+        )
+        for t in active_tokens:
+            t.is_active = False
+
         token = ScimToken(
             name=name,
             hashed_token=hashed_token,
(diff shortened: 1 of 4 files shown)


backend/ee/onyx/main.py
```diff
@@ -163,10 +164,14 @@ def get_application() -> FastAPI:
...
         include_router_with_global_prefix_prepended(application, tenants_router)
 
-    # SCIM 2.0 — service discovery endpoints (unauthenticated).
+    # SCIM 2.0 — protocol endpoints (unauthenticated by Onyx session auth;
+    # they use their own SCIM bearer token auth).
     # Not behind APP_API_PREFIX because IdPs expect /scim/v2/... directly.
     application.include_router(scim_router)
 
+    # SCIM admin — token management (standard Onyx admin auth)
+    include_router_with_global_prefix_prepended(application, scim_admin_router)
```
nmgarza5: eh maybe doesnt need to be its own router. let me know what you think @justin-tahara

backend/ee/onyx/db/scim.py
```diff
@@ -8,11 +8,11 @@
     def get_scim_dal(db_session: Session = Depends(get_session)) -> ScimDAL:
         return ScimDAL(db_session)
 
-    @router.get("/tokens")
-    def list_tokens(dal: ScimDAL = Depends(get_scim_dal)) -> ...:
-        tokens = dal.list_tokens()
+    @router.post("/tokens")
+    def create_token(dal: ScimDAL = Depends(get_scim_dal)) -> ...:
+        token = dal.create_token(name=..., hashed_token=..., ...)
```
justin-tahara, line 13: Is this something that we want to swap?

backend/ee/onyx/server/scim/admin.py
```diff
@@ -0,0 +1,83 @@
...
+        id=token.id,
+        name=token.name,
+        token_display=token.token_display,
+        is_active=token.is_active,
+        created_at=token.created_at,
+        last_used_at=token.last_used_at,
+    )
+
+
+@admin_router.post("/tokens", status_code=201)
```
justin-tahara: Any reason it's tokens? Is this the convention? Or should we just align with the get endpoint given it's only creating one token?

backend/ee/onyx/main.py
justin-tahara: I think it would be nice to be aligned with everything else. Would be odd to just deviate for SCIM only

backend/ee/onyx/db/scim.py
nmgarza5: this is just a doc string 🤦

backend/ee/onyx/server/scim/admin.py
nmgarza5: eh can it can be token

backend/ee/onyx/main.py
nmgarza5: everything else is not straight forward....we have tons of routers

backend/ee/onyx/db/scim.py
justin-tahara: oop my b

backend/ee/onyx/server/scim/admin.py
nmgarza5: Good call — renamed to `POST /admin/scim/token` (singular) to align with the GET endpoint. Pushed.

backend/ee/onyx/main.py
nmgarza5: Good point. We're using `include_router_with_global_prefix_prepended` because the admin router needs the global API prefix (like all other admin endpoints), while the SCIM protocol router (`/scim/v2`) is registered directly without a prefix since IdPs hardcode those exact paths. So it's intentional — the two routers have different registration requirements.

justin-tahara (approved):

PR #8580, opened 2026-02-19
perf(open-url): parallelize URL fetching with split connect/read timeouts

Subash-Mohan:
## Description
  - Parallelize URL fetching in OnyxWebCrawler.contents() using ThreadPoolExecutor (up to 5 workers), significantly reducing
  latency when opening multiple URLs
  - Split the single timeout into separate connect (5s) and read (15s) timeouts for faster failure detection on unreachable
  hosts
  - Isolate failures so one bad URL doesn't kill the entire batch — each URL's errors are caught independently
  - Update ssrf_safe_get / _make_ssrf_safe_request timeout type to accept float | tuple[float, float]

## How Has This Been Tested?
Tested with 10 URLs via chat; summarization worked perfectly, achieving a 4.7× performance boost.

## Additional Options

---
## Summary by cubic
Speeds up open_url by fetching URLs in parallel and using separate connect/read timeouts. Multi-URL opens are ~4.7× faster, with failures isolated per URL.

- **New Features**
  - Parallel URL fetching with ThreadPoolExecutor (up to 5 workers).
  - Separate connect (5s) and read (15s) timeouts for quicker failures.
  - Per-URL error handling so one bad URL doesn’t block the batch.
  - ssrf_safe_get/_make_ssrf_safe_request now accept float or (connect, read) timeout tuples.
  - Unit tests for parallelism, failure isolation, and timeout handling.

- **Bug Fixes**
  - Clearer error logging for SSRF blocks and unexpected exceptions.
  - Stabilized tests: set default encodings in mocks and validate tuple timeout form in ssrf_safe_get.

<sup>Written for commit 6fd57c15fb45ee5f155c04abc84fd8b5ed5a9c8a. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/unit/onyx/utils/test_url_ssrf.py
+++ b/backend/tests/unit/onyx/utils/test_url_ssrf.py
@@ -291,7 +291,7 @@ class TestSsrfSafeGet:
                 assert call_args[1]["headers"]["User-Agent"] == "TestBot/1.0"
 
     def test_passes_timeout(self) -> None:
-        """Test that timeout is passed through."""
+        """Test that timeout is passed through, including tuple form."""
         mock_response = MagicMock()
         mock_response.is_redirect = False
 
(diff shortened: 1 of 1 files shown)


backend/onyx/utils/url.py
```diff
@@ -204,7 +204,7 @@ def _make_ssrf_safe_request(
 def ssrf_safe_get(
     url: str,
     headers: dict[str, str] | None = None,
-    timeout: int = 15,
+    timeout: float | tuple[float, float] = 15,
```
jessicasingh7, line 207: can we only define timeout as tuple[float, float]

Subash-Mohan: I feel keeping float in the union lets callers who only care about a single overall timeout pass a simple number, which requests natively supports.

PR #8595, opened 2026-02-19
feat(web): Logical Hardening for Brave Web Search 2/3

justin-tahara:
## Description

PR built on the previous implementation. This is purely for hardening the workflow 

## How Has This Been Tested?

Added new unit tests

## Additional Options

---
## Summary by cubic
Hardened Brave Web Search with targeted retries, strict config parsing, and normalized optional params for safer, more predictable requests.

- **New Features**
  - Support for country, search_lang, ui_lang, safesearch, and freshness; values are normalized before sending.
  - timeout_seconds can be provided as a string in config; parsed and validated as > 0.

- **Bug Fixes**
  - Retries only on network errors and 429/5xx; other HTTP statuses raise clear ValueErrors without retry.
  - Ignore malformed result items and trim fields to avoid empty/invalid entries.

<sup>Written for commit 123ab6a6ec46ca46d0b3f768dabdff4a13429f0b. Summary will update on new commits.</sup>


backend/onyx/tools/tool_implementations/web_search/clients/brave_client.py
```diff
@@ -25,44 +31,97 @@ def __init__(
         *,
         num_results: int = 10,
         timeout_seconds: int = 10,
+        country: str | None = None,
```
jessicasingh7, line 34: do we need these extra params of country, etc.- asking because other web search providers we didn't add this

justin-tahara: Spoke offline, we are keeping this given this can be used via the API but as a followup we should probably customize the Web Search to be able to use this out of the box and configurable as well

PR #8596, opened 2026-02-19
feat: Modal Header no icon

Danelegend:
## Description
Currently the Modal Header requires an icon. This enables the Modal Header to accept no header

## How Has This Been Tested?
Example without icon:
Screenshot 2026-02-19 at 12 09 10 PM

## Additional Options

---
## Summary by cubic
Modal Header now supports no icon: the icon prop is optional and renders a minimal title/description header with the close button inline; with an icon, the close button sits next to it. Also removes an accidentally committed file.

<sup>Written for commit cd8de8ea1f007bb592288902cb4bb89311830ce9. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/refresh-components/Modal.tsx
+++ b/web/src/refresh-components/Modal.tsx
@@ -390,6 +390,7 @@ ModalContent.displayName = DialogPrimitive.Content.displayName;
  *
  * When `icon` is omitted the header renders a minimal variant: just the
  * title + description with the close button inline to the right of the title.
+ * This is JUST to be used for preview windows
  *
  * @example
  * ```tsx


web/src/refresh-components/Modal.tsx
```diff
@@ -415,45 +421,67 @@ const ModalHeader = React.forwardRef<HTMLDivElement, ModalHeaderProps>(
...
+                size={24}
+              />
+              {closeButton}
+            </Section>
+          )}
 
           <Section alignItems="start" gap={0}>
-            <DialogPrimitive.Title asChild>
-              <Text headingH3>{title}</Text>
-            </DialogPrimitive.Title>
```
nmgarza5, line 456: why do we have to change this? seems like Icon + closeButton only need to be handled in the section above this.

Danelegend: Currently the icon, title and description are all new lines. The close button is then aligned with the icon in this scenario. With the change of removing the icon, we then need to make the close button align with the title/description instead.

PR #8605, opened 2026-02-20
chore(app packaging): Ensure `curl` is included in the app image

acaprau:
## Description
We were nuking `curl` in the `apt-get remove -y --allow-remove-essential ...` step of the Dockerfile.

## How Has This Been Tested?
I did not test that this Dockerfile results in an image with `curl`, but I manually verified that running `apt-get update && \
    apt-get remove -y --allow-remove-essential \
        perl-base \
        xserver-common \
        xvfb \
        cmake \
        libldap-2.5-0 \
        libxmlsec1-dev \
        pkg-config \
        gcc`
removes `curl` from the system even though I manually installed it.

## Additional Options

---
## Summary by cubic
Ensure curl is included in the backend Docker image by installing it after the cleanup step. This fixes missing curl in built images and helps with debugging.

- **Bug Fixes**
  - Install curl in the final apt-get install block after the remove --allow-remove-essential step.
  - Add Dockerfile notes warning that the remove step can delete manually installed packages.

<sup>Written for commit 069f6f25fa81acdd801b1167f359d9c3d65b6e26. Summary will update on new commits.</sup>

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -28,13 +28,15 @@ COPY --from=ghcr.io/astral-sh/uv:0.9.9 /uv /uvx /bin/
 # Install system dependencies
 # cmake needed for psycopg (postgres)
 # libpq-dev needed for psycopg (postgres)
-# curl included just for users' convenience
 # zip for Vespa step futher down
 # ca-certificates for HTTPS
+# NOTE: Packages installed here are liable to be removed by `apt-get remove -y
+# --allow-remove-essential` below, even though they were manually installed
+# here. For useful packages like curl, prefer installing them below after the
+# remove step.
 RUN apt-get update && \
     apt-get install -y \
         cmake \
-        curl \
         zip \
         ca-certificates \
         libgnutls30 \
@@ -85,6 +87,7 @@ RUN uv pip install --system --no-cache-dir --upgrade \
     # xserver-common and xvfb included by playwright installation but not needed after
     # perl-base is part of the base Python Debian image but not needed for Onyx functionality
     # perl-base could only be removed with --allow-remove-essential
+    # NOTE: This removes packages that may have been manually installed above (such as curl).
     apt-get update && \
     apt-get remove -y --allow-remove-essential \
         perl-base \
@@ -98,6 +101,8 @@ RUN uv pip install --system --no-cache-dir --upgrade \
     # Install here to avoid some packages being cleaned up above
     apt-get install -y \
         libxmlsec1-openssl \
+        # curl is included for debugging convenience.
+        curl \
         # Install postgresql-client for easy manual tests
         postgresql-client && \
     apt-get autoremove -y && \


backend/Dockerfile
```diff
@@ -28,13 +28,15 @@ COPY --from=ghcr.io/astral-sh/uv:0.9.9 /uv /uvx /bin/
 # Install system dependencies
 # cmake needed for psycopg (postgres)
 # libpq-dev needed for psycopg (postgres)
-# curl included just for users' convenience
 # zip for Vespa step futher down
 # ca-certificates for HTTPS
+# NOTE: Packages installed here are liable to be removed by `apt-get remove -y
+# --allow-remove-essential` below, even though they were manually installed
+# here. For useful packages like curl, prefer installing them below after the
+# remove step.
```
jmelahman, line 36:
```suggestion
# here.
```
I think this is still the preferred place to install packages generally and this doesn't make that incredibly clear. If anything, something along the lines of "In that case, you may install them separately below."

backend/Dockerfile
```diff
@@ -85,6 +87,7 @@ RUN uv pip install --system --no-cache-dir --upgrade \
     # xserver-common and xvfb included by playwright installation but not needed after
     # perl-base is part of the base Python Debian image but not needed for Onyx functionality
     # perl-base could only be removed with --allow-remove-essential
+    # NOTE: This removes packages that may have been manually installed above (such as curl).
```
jmelahman, line 90: My guess is that removing these packages invalidates curl. I can see curl depending on `perl-base` or something and re-installing `curl` pulls `perl-base` back in implicitly. We should probably check which packages curl depends on (or inspect the new image or run the security scanner, etc.) to ensure we aren't invalidating this flow

acaprau:
closing this pr, installing curl necessarily introduces at least one CVE:

> curl → libcurl4 → libldap-2.5-0
libldap-2.5-0 is NOT in the base python:3.11.7-slim-bookworm image — it gets installed transitively (via playwright deps or Python packages), then explicitly removed at line 96 to reduce CVE exposure, and then curl brings it back as a hard dependency at line 104, along with several other packages (libsasl2-2, libldap-common, libsasl2-modules-db, etc.).
Are we making the image insecure?
Yes, somewhat. Installing curl reintroduces libldap-2.5-0 which has two known CVEs:
CVE-2023-2953 — null pointer dereference in ber_memalloc_x(). Marked "won't fix" in bookworm by Debian.
CVE-2025-68972 — improper verification of cryptographic signature in openldap.
The removal of libldap-2.5-0 at line 96 is effectively undone. A security scanner will flag it again.
Practical risk: curl only uses the LDAP path in libcurl if you're making ldap:// or ldaps:// requests. Debugging use with http:// / https:// doesn't trigger that code path. So the actual exploitability is low for a debugging-only curl install. That said, it still shows up as a vulnerability in scanners, which defeats the purpose of removing libldap in the first place.

PR #8609, opened 2026-02-20
fix: unify provider retrieval into swr hook

wenxi-onyx:
### Issue

On first load of /app, the application was making 6 duplicate requests to /api/llm/provider, causing unnecessary network overhead and potential performance issues.

### Root Cause

  1. ProviderContextProvider was using raw fetch() calls instead of SWR hooks, bypassing SWR's built-in request deduplication
  2. useOnboardingState was also making raw fetch() calls to the same endpoints and calling refreshLlmProviders() on mount, forcing unnecessary refetches
  3. Multiple components (ProviderContextProvider, useLlmManager, useOnboardingState) were all independently fetching the same data without coordination

### Resolution

  1. Created useLLMProviderOptions hook (web/src/lib/hooks/useLLMProviderOptions.ts)
    - Centralized fetching of /api/admin/llm/built-in/options
    - Uses SWR with 60-second deduplication interval
  2. Refactored ProviderContextProvider (web/src/components/chat/ProviderContext.tsx)
    - Replaced raw fetch("/api/llm/provider") with useLLMProviders() hook
    - Replaced raw fetch("/api/admin/llm/built-in/options") with useLLMProviderOptions() hook
    - Fixed problematic useEffect dependencies that were causing unnecessary re-renders
  3. Refactored useOnboardingState (web/src/refresh-components/onboarding/useOnboardingState.ts)
    - Removed refreshLlmProviders() call on mount
    - Replaced raw fetch("/api/admin/llm/built-in/options") with data from ProviderContextProvider

## Additional Options

---
## Summary by cubic
Unifies LLM provider and option retrieval with SWR and centralizes provider state in ProviderContext to eliminate duplicate requests. Keeps configuration prompts accurate and makes refresh consistent, re-testing the default provider on mount and on refresh if it hasn’t passed yet.

- **Refactors**
  - Added useLLMProviderOptions (SWR with errorHandlingFetcher, 60s dedupe, no focus revalidate).
  - ProviderContext now uses SWR hooks and exposes llmProviders/isLoadingProviders/hasProviders/providerOptions plus a single refreshProviderInfo that refetches both and conditionally re-tests the default provider.
  - Default provider test moved to a svc function (llm/svc.testDefaultProvider) instead of raw API calls.
  - Onboarding now consumes provider data via ProviderContext (useProviderStatus); removed manual fetches and redundant logic.
  - Removed dead code: initialSetup/welcome/lib.ts and initialSetup/welcome/constants.ts.

<sup>Written for commit a9412822830f7e9f96cec78d54946a60039f99f5. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/components/chat/ProviderContext.tsx
+++ b/web/src/components/chat/ProviderContext.tsx
@@ -13,6 +13,7 @@ import React, {
 import { useUser } from "@/providers/UserProvider";
 import { useLLMProviders } from "@/lib/hooks/useLLMProviders";
 import { useLLMProviderOptions } from "@/lib/hooks/useLLMProviderOptions";
+import { testDefaultProvider as testDefaultProviderSvc } from "@/lib/llm/svc";
 
 interface ProviderContextType {
   shouldShowConfigurationNeeded: boolean;
@@ -68,17 +69,10 @@ export function ProviderContextProvider({
       (!user || user.role === "admin");
 
     if (shouldCheck) {
-      try {
-        const response = await fetch("/api/admin/llm/test/default", {
-          method: "POST",
-        });
-        const success = response?.ok || false;
-        setDefaultCheckSuccessful(success);
-        if (success) {
-          setDefaultLLMProviderTestComplete();
-        }
-      } catch {
-        setDefaultCheckSuccessful(false);
+      const success = await testDefaultProviderSvc();
+      setDefaultCheckSuccessful(success);
+      if (success) {
+        setDefaultLLMProviderTestComplete();
       }
     }
   }, [user]);
--- a/web/src/lib/llm/svc.ts
+++ b/web/src/lib/llm/svc.ts
@@ -1,0 +1,23 @@
+/**
+ * LLM action functions for mutations.
+ *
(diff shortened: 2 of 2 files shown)


web/src/components/chat/ProviderContext.tsx
```diff
@@ -8,57 +11,107 @@ import React, {
...
+  const testDefaultProvider = useCallback(async () => {
+    const shouldCheck =
+      !checkDefaultLLMProviderTestComplete() &&
+      (!user || user.role === "admin");
 
+    if (shouldCheck) {
+      try {
+        const response = await fetch("/api/admin/llm/test/default", {
+          method: "POST",
+        });
```
nmgarza5: we are moving towards a new standard/pattern on the FE where all api calls that don't require swr caching are placed in a specific `svc.ts`. In this case, it probably makes sense to have a `web/src/lib/llm/svc.ts`

here is an example :big-grin
https://github.com/onyx-dot-app/onyx/blob/fb7e7e4395fc399a1d8834279131aada3c782de3/web/src/lib/billing/svc.ts#L4

nmgarza5 (approved):

PR #8633, opened 2026-02-20
fix(celery): Guardrail for User File Processing

justin-tahara:
## Description

Working on protecting the User File Processing queue. 

If the pods can't consume from the queue fast enough the queue builds up over time and can't handle the load and just continues to OOM. 

This is trying to prevent that build up from ever happening so the queues will eventually self recover. 

## How Has This Been Tested?

Adding tests

## Additional Options
Most likely need to cherry-pick back to a couple of releases

---
## Summary by cubic
Adds guardrails to the User File Processing Celery queue to prevent runaway backlogs and OOMs. Implements queue backpressure, per-file enqueue guards, and task expiry so the system self-recovers.

- **Bug Fixes**
  - Skip the beat cycle when queue depth exceeds USER_FILE_PROCESSING_MAX_QUEUE_DEPTH (500).
  - Set a per-file Redis guard key (USER_FILE_QUEUED_PREFIX) with TTL before enqueue; worker deletes it on pickup; logs enqueued and skipped_guard counts.
  - Enqueue tasks with expires=CELERY_USER_FILE_PROCESSING_TASK_EXPIRES (60s) to auto-drop stale items.
  - Added external-dependency unit tests (Redis/Postgres) covering backpressure, guard key TTL and cleanup, task expiry, and worker guard-key removal; exposed redis_pool.setex for guard setup.

<sup>Written for commit 8f1f5f304cb1bffaa1f06206c31399c7a4ea5848. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/background/celery/tasks/user_file_processing/tasks.py
+++ b/backend/onyx/background/celery/tasks/user_file_processing/tasks.py
@@ -195,21 +195,33 @@ def check_user_file_processing(self: Task, *, tenant_id: str) -> None:
             for user_file_id in user_file_ids:
                 # --- Protection 2: per-file queued guard ---
                 queued_key = _user_file_queued_key(user_file_id)
-                if redis_client.exists(queued_key):
+                guard_set = redis_client.set(
+                    queued_key,
+                    1,
+                    ex=CELERY_USER_FILE_PROCESSING_TASK_EXPIRES,
+                    nx=True,
+                )
+                if not guard_set:
                     skipped_guard += 1
                     continue
 
-                # --- Protection 3: task expiry + set guard ---
-                self.app.send_task(
-                    OnyxCeleryTask.PROCESS_SINGLE_USER_FILE,
-                    kwargs={"user_file_id": str(user_file_id), "tenant_id": tenant_id},
-                    queue=OnyxCeleryQueues.USER_FILE_PROCESSING,
-                    priority=OnyxCeleryPriority.HIGH,
-                    expires=CELERY_USER_FILE_PROCESSING_TASK_EXPIRES,
-                )
-                redis_client.setex(
-                    queued_key, CELERY_USER_FILE_PROCESSING_TASK_EXPIRES, 1
-                )
+                # --- Protection 3: task expiry ---
+                # If task submission fails, clear the guard immediately so the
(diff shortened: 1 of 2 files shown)


backend/onyx/background/celery/tasks/user_file_processing/tasks.py
```diff
@@ -135,7 +166,21 @@ def check_user_file_processing(self: Task, *, tenant_id: str) -> None:
...
+    skipped_guard = 0
     try:
+        # --- Protection 1: queue depth backpressure ---
+        r_celery = self.app.broker_connection().channel().client  # type: ignore
+        queue_len = celery_get_queue_length(
+            OnyxCeleryQueues.USER_FILE_PROCESSING, r_celery
+        )
+        if queue_len > USER_FILE_PROCESSING_MAX_QUEUE_DEPTH:
+            task_logger.warning(
+                f"check_user_file_processing - Queue depth {queue_len} exceeds "
```
acaprau, line 178: nit for future reference: task_logger already prepends `check_user_file_processing` to this log.

backend/onyx/background/celery/tasks/user_file_processing/tasks.py
```diff
@@ -135,7 +166,21 @@ def check_user_file_processing(self: Task, *, tenant_id: str) -> None:
...
+                f"{USER_FILE_PROCESSING_MAX_QUEUE_DEPTH}, skipping enqueue for "
+                f"tenant={tenant_id}"
+            )
+            return None
+
         with get_session_with_current_tenant() as db_session:
             user_file_ids = (
                 db_session.execute(
```
acaprau, line 186: @Subash-Mohan for future reference logic like this should ideally be encapsulated in some function in some dedicated module in `backend/onyx/db` or something like that.

backend/onyx/redis/redis_pool.py
```diff
@@ -109,6 +109,7 @@ def __getattribute__(self, item: str) -> Any:
             "unlock",
             "get",
             "set",
+            "setex",
```
acaprau, line 112: whats going on here?

justin-tahara: This was for the tests to properly pass.

PR #8644, opened 2026-02-20
fix(projects): Guardrails for Project User Files

justin-tahara:
## Additional Options

---
## Summary by cubic
Adds guardrails to Project User File sync: queue backpressure and per-file enqueue dedupe to prevent overload and duplicate work. Applies the same rules to API triggers and sets a 60s task expiry.

- **Bug Fixes**
  - Backpressure: skip enqueues when USER_FILE_PROJECT_SYNC depth > 500 (beat and API).
  - Dedupe: Redis NX+EX per user_file_id with 60s TTL; cleared on pickup; rolls back on publish failure; Celery task expires set to 60s.
  - API link/unlink uses guarded enqueue and defers when depth is high; tests cover backpressure, dedupe, expiry, and guard cleanup.

<sup>Written for commit bf86794699b65cc50d8113ba71a6dd77ad130057. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/unit/onyx/background/celery/tasks/test_user_file_project_sync_queue.py
+++ b/backend/tests/unit/onyx/background/celery/tasks/test_user_file_project_sync_queue.py
@@ -47,7 +47,7 @@ def test_check_for_user_file_project_sync_applies_queue_backpressure(
 
     task_app = MagicMock()
     with patch.object(check_for_user_file_project_sync, "app", task_app):
-        check_for_user_file_project_sync.run(tenant_id="test-tenant")  # type: ignore[misc]
+        check_for_user_file_project_sync.run(tenant_id="test-tenant")
 
     task_app.send_task.assert_not_called()
     lock.release.assert_called_once()
@@ -89,7 +89,7 @@ def test_check_for_user_file_project_sync_skips_duplicates(
 
     task_app = MagicMock()
     with patch.object(check_for_user_file_project_sync, "app", task_app):
-        check_for_user_file_project_sync.run(tenant_id="test-tenant")  # type: ignore[misc]
+        check_for_user_file_project_sync.run(tenant_id="test-tenant")
 
     assert mock_enqueue.call_count == 2
     lock.release.assert_called_once()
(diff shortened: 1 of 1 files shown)


backend/onyx/background/celery/tasks/user_file_processing/tasks.py
```diff
@@ -570,7 +623,16 @@ def check_for_user_file_project_sync(self: Task, *, tenant_id: str) -> None:
         return None
 
     enqueued = 0
+    skipped_guard = 0
     try:
+        queue_depth = get_user_file_project_sync_queue_depth(self.app)
+        if queue_depth > USER_FILE_PROJECT_SYNC_MAX_QUEUE_DEPTH:
+            task_logger.warning(
+                f"check_for_user_file_project_sync - Queue depth {queue_depth} exceeds "
```
acaprau: not a big deal don't need to change anything but remember the small nit from last time task_logger alr prefixes each log w check_for_user_file_project_sync 😅

backend/onyx/background/celery/tasks/user_file_processing/tasks.py
```diff
@@ -617,6 +683,8 @@ def process_single_user_file_project_sync(
     )
 
     redis_client = get_redis_client(tenant_id=tenant_id)
+    redis_client.delete(_user_file_project_sync_queued_key(user_file_id))
```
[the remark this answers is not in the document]
acaprau: technically true i guess, although only one of these locks can be re-set once at a time, that's the whole point of this lock, so i wouldn't say we're spamming the queue, just that there's a window of time where maybe we're re-enqueuing something once redundantly.

backend/onyx/configs/constants.py
```diff
@@ -157,6 +157,15 @@
 
 CELERY_USER_FILE_PROCESSING_LOCK_TIMEOUT = 30 * 60  # 30 minutes (in seconds)
 
+# How long a queued user-file-project-sync task remains valid.
+# Should be short enough to discard stale queue entries under load while still
+# allowing workers enough time to pick up new tasks.
+CELERY_USER_FILE_PROJECT_SYNC_TASK_EXPIRES = 60  # 1 minute (in seconds)
```
acaprau, line 173: this is just personal style preference don't need to change anything but for constants like this i like putting `_S` at the end to indicate what units it's in just by looking at it rather than having to come see this comment. not a big deal either way.

acaprau (approved):

backend/onyx/configs/constants.py
justin-tahara: Given the entire codebase has these issues I think we might be better off doing this in one big swoop

PR #8663, opened 2026-02-22
feat: code interpreter supports streaming

Danelegend:
## Description
Updates have been made to code-interpreter to support streaming. This allows the code-interpreter-client to use the streaming endpoint if the endpoint exists. Maintains backwards compatibility.

## How Has This Been Tested?
Manual + some tests

## Additional Options
closes https://linear.app/onyx-app/issue/ENG-3676/stream-code-interpreter-outputs

---
## Summary by cubic
Adds streaming to the Code Interpreter so users see live stdout/stderr updates, with automatic fallback to batch on older servers. Adds a warning log for incomplete SSE events. Addresses ENG-3676.

- **New Features**
  - CodeInterpreterClient: execute_streaming over SSE with typed Output/Result/Error events; robust SSE parser that warns on incomplete events; 404 fallback closes the stream and replays batch as stream while preserving the files param.
  - PythonTool: consumes stream events, emits incremental stdout/stderr deltas, uploads/cleans files after the result, then emits file_ids; builds final result from StreamResultEvent and raises on StreamErrorEvent.
  - Tests: unit tests for fallback and SSE parsing (stdout-only and files param); external dependency tests simulating SSE frames, asserting use of /v1/execute/stream, and verifying fallback when streaming is disabled.

<sup>Written for commit 53e30b7cbeda5fb4bfe1ef36bb6a3742625168b3. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/tools/tool_implementations/python/code_interpreter_client.py
+++ b/backend/onyx/tools/tool_implementations/python/code_interpreter_client.py
@@ -182,6 +182,12 @@ class CodeInterpreterClient:
             elif line.startswith("data:"):
                 data_lines.append(line[len("data:") :].strip())
 
+        if event_type is not None or data_lines:
+            logger.warning(
+                f"SSE stream ended with incomplete event: "
+                f"event_type={event_type}, data_lines={data_lines}"
+            )
+
     def _batch_as_stream(
         self,
         code: str,


backend/onyx/tools/tool_implementations/python/code_interpreter_client.py
```diff
@@ -45,32 +81,128 @@ def __init__(self, base_url: str | None = CODE_INTERPRETER_BASE_URL):
...
+                    if model_cls is not None:
+                        yield model_cls(**json.loads(data))
+                    else:
+                        logger.warning(f"Unknown SSE event type: {event_type}")
+                event_type = None
+                data_lines = []
+            elif line.startswith("event:"):
+                event_type = line[len("event:") :].strip()
+            elif line.startswith("data:"):
+                data_lines.append(line[len("data:") :].strip())
```
evan-onyx, line 183: I'd like at least a warning line at the end here if event_type or data_lines have stuff in them when we exit the for loop

PR #8673, opened 2026-02-23
fix: resolve OAuth token manager using masked secrets

Subash-Mohan:
## Description
This PR fixes the broken OpenAPI OAuth flow and includes a small UI fix.

Before:
Screenshot 2026-02-23 at 6 45 16 PM

After:
Screenshot 2026-02-23 at 6 45 27 PM

## How Has This Been Tested?
Test via the UI by creating the OpenAPI configuration and completing the OAuth flow successfully

## Additional Options

---
## Summary by cubic
Fixes OpenAPI OAuth by unwrapping masked client_id/client_secret and enforcing required checks so the auth URL, code exchange, and token refresh work again. Refactors token payloads and simplifies the OAuth button icon.

- **Bug Fixes**
  - Use _unwrap_sensitive_str to unmask client_id/client_secret and validate required fields (client_id for auth URL; client_id/client_secret for code exchange and refresh) across OAuthTokenManager; restores the full OAuth flow.

- **Refactors**
  - Build token request payloads via explicit data dicts and _unwrap_token_data; simplify ActionLineItem OAuth icon to SvgKey with secondary prominence and small size.

<sup>Written for commit d51eb14e239641c796027f1da368bceb317883bc. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/auth/oauth_token_manager.py
+++ b/backend/onyx/auth/oauth_token_manager.py
@@ -71,9 +71,9 @@ class OAuthTokenManager:
         data: dict[str, str] = {
             "grant_type": "refresh_token",
             "refresh_token": token_data["refresh_token"],
-            "client_id": self.oauth_config.client_id.get_value(apply_mask=False),
-            "client_secret": self.oauth_config.client_secret.get_value(
-                apply_mask=False
+            "client_id": self._unwrap_sensitive_str(self.oauth_config.client_id),
+            "client_secret": self._unwrap_sensitive_str(
+                self.oauth_config.client_secret
             ),
         }
         response = requests.post(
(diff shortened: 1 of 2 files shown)


backend/onyx/auth/oauth_token_manager.py
```diff
@@ -58,15 +58,25 @@ def refresh_token(self, user_token: OAuthUserToken) -> str:
...
         token_data = self._unwrap_token_data(user_token.token_data)
 
         response = requests.post(
             self.oauth_config.token_url,
             data={
                 "grant_type": "refresh_token",
                 "refresh_token": token_data["refresh_token"],
-                "client_id": self.oauth_config.client_id,
-                "client_secret": self.oauth_config.client_secret,
+                "client_id": self.oauth_config.client_id.get_value(apply_mask=False),
```
yuhongsun96: How come the type checking didn't catch these

Subash-Mohan: \`requests.post(data=...)\` accepts \`Mapping[Any, Any]\`, so mypy allows any value type in the dict — passing a \`SensitiveValue[str]\` instead of a \`str\` doesn't trigger a type error.

The \`SensitiveValue.__str__() -> NoReturn\` guard only fires at runtime (when the value is actually serialized), not during static analysis.

I've added explicit \`dict[str, str]\` annotations on the data dicts so mypy will catch this kind of mismatch going forward.

PR #8679, opened 2026-02-23
fix(db): Multitenant Schema migration update

justin-tahara:
## Description

The previous approach that we had would cause the database to lock since we were holding a lock on the entire set of schemas causing us to destroy the DB. This new approach creates a temp table that is used to maintain a running list instead of doing the entire query at once. 

## How Has This Been Tested?

Ran the live command on the DB and tested that it finished in a reasonable amount of time. Also added a new set of tests as well.

## Additional Options

---
## Summary by cubic
Prevents DB locks during multitenant migration checks by replacing the giant UNION with a server-side loop and temp tables that snapshot alembic versions per schema. Iterates only the requested schemas, keeps locks short, scales to large schema counts, and is safe to run repeatedly.

- **Bug Fixes**
  - Move schema check logic into tenant_utils; runner now imports get_schemas_needing_migration.
  - Use a PL/pgSQL DO block with temp input/snapshot tables to iterate schemas server-side.
  - Quote identifiers on the server; include schemas with missing tables or invalid names as needing migration.
  - Add integration tests for classification (head/empty/stale), idempotency, and empty input.

<sup>Written for commit 372dc9ece5ed92139c01c44cae68f667feb909bf. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/integration/multitenant_tests/migrations/test_run_multitenant_migrations.py
+++ b/backend/tests/integration/multitenant_tests/migrations/test_run_multitenant_migrations.py
@@ -11,12 +11,10 @@ Usage:
 
 from __future__ import annotations
 
-import importlib.util
 import os
 import subprocess
 import sys
 import uuid
-from collections.abc import Callable
 from collections.abc import Generator
 
 import pytest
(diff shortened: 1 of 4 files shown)


justin-tahara:
@greptile @cubic-dev-ai can you review the PR one more time?

justin-tahara:
@greptile @cubic-dev-ai Can you review the PR one more time?

backend/tests/integration/multitenant_tests/migrations/test_run_multitenant_migrations.py
```diff
@@ -28,6 +30,19 @@
     os.path.join(os.path.dirname(__file__), "..", "..", "..", "..")
 )
 
+# Load get_schemas_needing_migration directly from its file path to avoid the
```
jmelahman: Let's move this to backend/tests/unit/db/ instead. This seems like too much implementation knowledge for an integration test.

Also, we can refactor get_schemas_needing_migration into onyx/db/ somewhere and then unit test that module and just call it from the migration script. That implicitly also fixes the import collision.

---

I just realized that a unit test likely won't nor should it have access to a database which would be necessary here, but I think we should still move these cases out of this test. It can stay in `tests/integration/multitenant_tests/`, but I think should be a separate test focused on testing the new helper module which houses `get_schemas_needing_migration`.

PR #8691, opened 2026-02-23
fix(fe): inline code-blocks respect header font-size

jmelahman:
## Description

Closes https://linear.app/onyx-app/issue/ENG-3732/inline-code-block-should-follow-size-of-header-etc

## How Has This Been Tested?

Visual regression tests included

**before**
image

**after**
image

## Additional Options

---
## Summary by cubic
Inline code in chat now scales with heading font sizes (h1–h4) using 0.75em, fixing ENG-3732. Added e2e tests for h1–h4 with inline code and a data-testid on inline code for stable assertions; renamed references from “header” to “heading”.

<sup>Written for commit 04fcf685520f5768600eb97e3e03ed53feebe4a6. Summary will update on new commits.</sup>

### DIFF
--- a/web/tests/e2e/chat/chat_message_rendering.spec.ts
+++ b/web/tests/e2e/chat/chat_message_rendering.spec.ts
@@ -675,7 +675,7 @@ The platform architecture document provides additional context on how these impr
     });
 
     test.describe("Header Levels", () => {
-      const HEADERS_RESPONSE = `# Getting Started
+      const HEADINGS_RESPONSE = `# Getting Started
 
 This is the introductory paragraph.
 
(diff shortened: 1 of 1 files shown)


jmelahman:
@greptile please re-review

web/src/app/app/message/CodeBlock.tsx
```diff
@@ -62,12 +62,13 @@ export const CodeBlock = memo(function CodeBlock({
   if (typeof children === "string" && !language) {
     return (
       <span
+        data-testid="code-block"
```
nmgarza5, line 65: can there be multiple codeblocks at once? might want to make thsi more specifc in someway

nmgarza5 (approved):

web/src/app/app/message/CodeBlock.tsx
jmelahman: I think this is correct. You can chain selectors to find specific instances (like in the tests here). `testid` is not like an element `id` where they're expected to be unique

PR #8694, opened 2026-02-23
feat(llm-provider): main page refresh (pt1)

raunakab:
## Description

This PR implements a refreshed LLM Configuration Page. It only refreshes the core page; the popup-modals will be refreshed in a quick follow-up PR.

Addresses: https://linear.app/onyx-app/project/admin-llm-model-config-ffd46c18b87b.

## Screenshots

image

## Additional Options

---
## Summary by cubic
Refreshed the LLM Configuration page with SettingsLayouts, a provider‑grouped default model selector using global defaults, modern provider cards with a hover‑to‑reveal delete icon, and a safer delete flow. Part one of the Admin LLM Model Config refresh (Linear project).

- New Features
  - Default model selector groups by provider; reads global defaultText; updates via setDefaultLlmModel(provider_id, model_name) with toasts; shows an info state when no providers exist.
  - “Available Providers” list and “Add Provider” grid using ContentAction cards; Connect/Edit modals are controlled and pause fetches when closed; delete uses a confirmation modal with a warning if it’s the active default; added a hover‑to‑reveal trash icon on existing provider cards.
  - Test/PUT requests send model and provider id; added an error toast when a well‑known provider lacks a modal mapping.

- Refactors
  - Page moved to refresh-pages/admin/LLMConfigurationPage and wrapped in SettingsLayouts; sidebar label “LLM Models”; layout registered in ClientLayout.
  - Renamed lib/llm → lib/llmConfig; added svc.ts (testDefaultProvider, setDefaultLlmModel, deleteLlmProvider); updated provider display helpers and icons; consolidated hooks into hooks/useLLMProviders with useAdminLLMProviders; adapted to LLMProviderResponse with defaultText; LLMSelector accepts defaultText; restored getFinalLLM fallback.
  - Moved interfaces to interfaces/llm; removed deprecated default fields; ModelConfiguration now requires supports_image_input and supports_reasoning; provider_display_name required; unit and e2e tests updated; null‑safe isAnthropic.
  - Hardened cherry‑pick workflow to avoid code injection.

<sup>Written for commit fcb2e063c552d91ddfb01afa646ed78dc9a5724e. Summary will update on new commits.</sup>

### DIFF
--- a/web/tests/e2e/admin/llm_provider_setup.spec.ts
+++ b/web/tests/e2e/admin/llm_provider_setup.spec.ts
@@ -3,7 +3,7 @@ import type { Locator, Page } from "@playwright/test";
 import { loginAs } from "@tests/e2e/utils/auth";
 import { OnyxApiClient } from "@tests/e2e/utils/onyxApiClient";
 
-const LLM_SETUP_URL = "/admin/configuration/llm";
+const LLM_CONFIG_URL = "/admin/configuration/llm";
 const BASE_URL = process.env.BASE_URL || "http://localhost:3000";
 const PROVIDER_API_KEY =
   process.env.E2E_LLM_PROVIDER_API_KEY ||
@@ -13,18 +13,32 @@ const PROVIDER_API_KEY =
 type AdminLLMProvider = {
   id: number;
   name: string;
-  is_default_provider: boolean | null;
   is_auto_mode: boolean;
+};
+
+type AdminLLMProviderResponse = {
+  providers: AdminLLMProvider[];
+  default_text: { provider_id: number; model_name: string } | null;
 };
 
 function uniqueName(prefix: string): string {
   return `${prefix}-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
 }
 
-async function listAdminLLMProviders(page: Page): Promise<AdminLLMProvider[]> {
+async function listAdminLLMProviders(
+  page: Page
+): Promise<AdminLLMProviderResponse> {
   const response = await page.request.get(`${BASE_URL}/api/admin/llm/provider`);
   expect(response.ok()).toBeTruthy();
-  return (await response.json()) as AdminLLMProvider[];
+  return (await response.json()) as AdminLLMProviderResponse;
+}
+
+async function getProviderByName(
+  page: Page,
+  providerName: string
+): Promise<AdminLLMProvider | null> {
+  const { providers } = await listAdminLLMProviders(page);
+  return providers.find((provider) => provider.name === providerName) ?? null;
 }
 
 async function createPublicProvider(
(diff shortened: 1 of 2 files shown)


web/src/sections/modals/llmConfig/formUtils.ts
```diff
@@ -15,10 +15,7 @@ export const buildDefaultInitialValues = (
   existingLlmProvider?: LLMProviderView,
   modelConfigurations?: ModelConfiguration[]
 ) => {
-  const defaultModelName =
-    existingLlmProvider?.default_model_name ??
-    modelConfigurations?.[0]?.name ??
-    "";
+  const defaultModelName = modelConfigurations?.[0]?.name ?? "";
```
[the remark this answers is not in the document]
raunakab: `default_model_name` was intentionally removed from `LLMProviderView` — the backend no longer sends a per-provider default model (it is now a global setting tracked via the `DefaultModel` envelope in the API response). Restoring `existingLlmProvider?.default_model_name` would cause a TypeScript error since that field no longer exists on the type.

Each modal already overrides `default_model_name` with its own logic. For example, `AnthropicModal` uses `wellKnownLLMProvider?.recommended_default_model?.name`, and `CustomModal` has its own model configuration fields. The value from `buildDefaultInitialValues` is just a fallback that gets overwritten.

Resolving as not applicable.

web/src/lib/llmConfig/utils.ts
```diff
@@ -1,21 +1,23 @@
...
-  const defaultProvider = llmProviders.find(
-    (llmProvider) => llmProvider.is_default_provider
-  );
+  const defaultProvider = defaultText
+    ? llmProviders.find((p) => p.id === defaultText.provider_id)
+    : undefined;
 
   let provider = defaultProvider?.provider || "";
-  let model = defaultProvider?.default_model_name || "";
+  let model = defaultText?.model_name || "";
```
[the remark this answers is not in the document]
raunakab: Valid concern — fixed. When `defaultText` is not provided, `getFinalLLM` now falls back to the first provider with visible models (and uses the first visible model name), similar to how `getDefaultLlmDescriptor` already works.

The `useChatController` callsite passes `llmManager.currentLlm` which is already resolved from `defaultText` inside `useLlmManager`, so in practice this was covered at runtime. But the fallback makes `getFinalLLM` more robust for edge cases (e.g., initial render before `currentLlm` is populated).

web/src/hooks/useLLMProviders.ts
```diff
@@ -0,0 +1,150 @@
...
+ *    while loading.
+ * - `defaultText` — The global (or persona-overridden) default text model.
+ * - `defaultVision` — The global (or persona-overridden) default vision model.
+ * - `isLoading` — `true` until the first successful response or error.
+ * - `error` — The SWR error object, if any.
+ * - `refetch` — SWR `mutate` function to trigger a revalidation.
+ */
+export function useLLMProviders(personaId?: number) {
+  const url =
+    typeof personaId === "number"
```
nmgarza5: why do we need to this type check?

web/src/refresh-pages/admin/LLMConfigurationPage.tsx
```diff
@@ -0,0 +1,379 @@
...
+
+              <div className="flex flex-col gap-2">
+                {[...existingLlmProviders]
+                  .sort((a, b) => {
+                    const aIsDefault = defaultText?.provider_id === a.id;
+                    const bIsDefault = defaultText?.provider_id === b.id;
+                    if (aIsDefault && !bIsDefault) return -1;
+                    if (!aIsDefault && bIsDefault) return 1;
+                    return 0;
+                  })
```
nmgarza5: would rather have this in a helper func instead of inlined within JSX or sort ahead of time and pass `sortedLlmProviders`

web/src/refresh-pages/admin/LLMConfigurationPage.tsx
```diff
@@ -0,0 +1,379 @@
...
+                value={currentDefaultValue}
+                onValueChange={handleDefaultModelChange}
+              >
+                <InputSelect.Trigger placeholder="Select a default model" />
+                <InputSelect.Content>
+                  {existingLlmProviders.map((provider) => {
+                    const visibleModels = provider.model_configurations.filter(
+                      (m) => m.is_visible
+                    );
+                    if (visibleModels.length === 0) return null;
```
nmgarza5: same comment as previous. calculate this ahead of time and pass `filteredLlmProviders` instead

web/src/refresh-pages/admin/LLMConfigurationPage.tsx
```diff
@@ -0,0 +1,379 @@
...
+            title="Add Provider"
+            description="Onyx supports both popular providers and self-hosted models."
+            sizePreset="main-content"
+            variant="section"
+          />
+
+          <div className="grid grid-cols-2 gap-2">
+            {wellKnownLLMProviders?.map((provider) => {
+              const formFn = PROVIDER_MODAL_MAP[provider.name];
+              if (!formFn) return null;
```
nmgarza5: when would we hit this null check? is it needed? if we don't have the form, should we console error? seems like this would not be a situation we want to be in and logging would be required

web/src/refresh-pages/SettingsPage.tsx
```diff
@@ -326,8 +328,13 @@ function GeneralSettings() {
...
               center
             >
+              {/* next-themes injects a blocking <script> that reads the theme
+                 from localStorage before React hydrates, so `theme` is already
+                 "system" on the client's first render while the server saw
+                 `undefined`. Deferring to `mounted` keeps SSR and hydration
+                 in agreement, avoiding a data-placeholder mismatch. */}
               <InputSelect
-                value={theme}
+                value={mounted ? theme : undefined}
```
nmgarza5: why is this change needed in this PR?

web/src/hooks/useLLMProviders.ts
raunakab: Fixed — simplified to `personaId \!== undefined`. TypeScript already constrains the type to `number | undefined`, so the `typeof` check was redundant.

web/src/refresh-pages/admin/LLMConfigurationPage.tsx
raunakab: Fixed — extracted into a `sortedProviders` variable computed before the JSX return.

web/src/refresh-pages/admin/LLMConfigurationPage.tsx
raunakab: Fixed — extracted into a `providersWithVisibleModels` variable computed before the JSX return.

web/tests/e2e/admin/llm_provider_setup.spec.ts
```diff
@@ -186,70 +238,85 @@ test.describe("LLM Provider Setup @exclusive", () => {
...
-      secondProviderName
-    );
-    providersToCleanup.push(firstProviderId, secondProviderId);
+    const providerName = uniqueName("PW Default Test Provider");
+    const providerId = await createPublicProvider(page, providerName);
+    providersToCleanup.push(providerId);
 
-    try {
-      await apiClient.setProviderAsDefault(firstProviderId);
+    await apiClient.setProviderAsDefault(providerId, "gpt-4o");
```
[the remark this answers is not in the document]
raunakab: Fixed — wrapped the assertion blocks in `try/finally` so the initial default is always restored, even on test failure.

.github/workflows/post-merge-beta-cherry-pick.yml
```diff
@@ -75,10 +80,82 @@
...
+          {
+            echo "details<<EOF"
+            tail -n 40 "$output_file"
+            echo "EOF"
+          } >> "$GITHUB_OUTPUT"
+
+      - name: Mark workflow as failed if cherry-pick failed
+        if: steps.gate.outputs.should_cherrypick == 'true' && steps.run_cherry_pick.outputs.status == 'failure'
+        run: |
+          echo "::error::Automated cherry-pick failed (${{ steps.run_cherry_pick.outputs.reason }})."
```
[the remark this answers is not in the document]
raunakab: Fixed — moved `${{ steps.run_cherry_pick.outputs.reason }}` into an `env` variable (`CHERRY_PICK_REASON`) and referenced it as `${CHERRY_PICK_REASON}` in the shell script, preventing template injection.

web/src/refresh-pages/admin/LLMConfigurationPage.tsx
raunakab: This is purely defensive — it guards against a well-known provider being returned by the API without a corresponding modal implementation. Added an error toast so the issue is surfaced to the admin instead of silently swallowed.

web/src/refresh-pages/SettingsPage.tsx
raunakab: Good call — reverted from this PR. It was a hydration fix for the theme selector (next-themes SSR mismatch) that I noticed while working nearby, but it belongs in its own PR.

raunakab:
Closing in favour of https://github.com/onyx-dot-app/onyx/pull/8761.

PR #8695, opened 2026-02-23
refactor: add new `ContentAction` component

raunakab:
## Description

- Create `@opal/layouts/ContentAction`: a `Content` + `rightChildren` row layout with padding alignment via the shared size scale
- Add README docs for `ContentAction` and the `layouts` directory

## Rationale

`Content` purely deals with rendering the title, description, and icon. It is exactly up-to-spec with the mocks @duo-onyx has created (available here: https://www.figma.com/design/Fms9uHkizpBuDEfsB8ujCc/Onyx-UI-Library?node-id=1598-39416&m=dev).

`ContentAction` then expands on this by ALSO allowing a `rightChildren?: React.ReactNode` to be specified. The "right children" allows end-users to throw whatever content on the right side of the `Content` invocation. This is a very VERY common use-case in our entire application.

Now, you may be wondering why a whole new component needs to be created for this. Isn't this a simple `"flex flex-row"`? Well, no, not exactly. In Duo's mocks, it is possible for the `Content` to have *padding* around it in order for the title-line to appear inline with the *TEXT* of a button on the right side. Take this for example:

image

On the left, we have the standard `Content` component. On the right, we have the "right children", which just happens to be a `Button`. However, since the `Button` itself has some padding, the text is pushed down 8px. Therefore, in order to counteract that and make sure the title stays inline with the text of the button, the `Content` component itself needs to be wrapped in some padding (8px around, to be specific).

Therefore, due to this additional complexity detail, I would much rather prefer we have a standardized way of adding the *right* amount of padding (the padding amounts coincide directly with the button-padding variants).

## Screenshots

The below are examples of `ContentAction` in action. Note how the text of the button is *exactly* inline with the title.

image

(and hovered:)

image

image

## Additional Options

---
## Summary by cubic
Added ContentAction for row layouts with right-side actions and moved size variants to @opal/shared for consistent padding and height across components. Updated docs and examples to match the shared size scale.

- **New Features**
  - Added ContentAction with rightChildren and paddingVariant (uses shared SizeVariant).
  - Added README docs for ContentAction and layouts; fixed README example import (SvgCloud).

- **Refactors**
  - Extracted sizeVariants and SizeVariant to @opal/shared with module docs.
  - Interactive.Container now uses SizeVariant for heightVariant.
  - Updated Button, HeadingLayout, and LabelLayout to import SizeVariant from @opal/shared.
  - Removed InteractiveContainerHeightVariant from @opal/core exports.

<sup>Written for commit 23e8cb0963a685fc8e4e5ef9ce85ce45a7d987a6. Summary will update on new commits.</sup>

### DIFF
--- a/web/lib/opal/src/shared.ts
+++ b/web/lib/opal/src/shared.ts
@@ -1,8 +1,16 @@
+/**
+ * @opal/shared — Shared constants and types for the opal design system.
+ *
+ * This module holds design tokens that are referenced by multiple opal
+ * packages (core, components, layouts). Centralising them here avoids
+ * circular imports and gives every consumer a single source of truth.
+ */
+
 // ---------------------------------------------------------------------------
 // Size Variants
 //
-// Shared size-variant scale used across the opal design system.
-// Maps named size presets to Tailwind height, min-width, and padding classes.
+// A named scale of size presets (lg → 2xs, plus fit) that map to Tailwind
+// utility classes for height, min-width, and padding.
 //
 // Consumers:
 //   - Interactive.Container  (height + min-width + padding)
--- a/web/lib/opal/src/layouts/ContentAction/README.md
+++ b/web/lib/opal/src/layouts/ContentAction/README.md
@@ -68,7 +68,7 @@ import SvgSettings from "@opal/icons/settings";
 ```tsx
 import { ContentAction } from "@opal/layouts";
 import { Button } from "@opal/components";
-import { SvgArrowExchange } from "@opal/icons";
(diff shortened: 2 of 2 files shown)


web/lib/opal/src/shared.ts
nmgarza5, line 1: can you add a description at the top of the file for what its purpose? `shared.ts` tells me its for shared types/props across lib components but I think we should state that clearly

nmgarza5 (approved):

web/lib/opal/src/shared.ts
raunakab: Added!

PR #8696, opened 2026-02-23
chore(chat): consolidate chat message whitespacing style

jmelahman:
## Description

Standardizes the spaces between chat messages by applying `gap-12` on the message container rather than rely on individual messages aligning themselves with padding.

Closes https://linear.app/onyx-app/issue/ENG-3744/the-user-input-is-misaligned

## How Has This Been Tested?

Reviewed the visual regression report and all changes look desired

---
## Summary by cubic
Standardized chat message spacing by moving vertical gaps and right padding to the ChatUI container, and made HumanMessage action buttons responsive to prevent wrap-induced layout shifts. Fixes misaligned user input and keeps layouts consistent across the chat (Linear ENG-3744).

- **Refactors**
  - Applied flex flex-col with gap-12, pt-4, pb-8, pr-1 on the ChatUI container; removed the top Spacer and per-message padding; DynamicBottomSpacer now has w-full to satisfy Playwright.
  - HumanMessage: removed pt/pb and -mr-6; message row is md:flex and relative; action buttons are absolute on small screens and relative on md+ with z-content, right-anchored. AgentMessage: removed padding (including pr-1) and relies on container gap.

<sup>Written for commit 3a397cf135f1e1c9d83d343aac75822a42fbb453. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/components/chat/DynamicBottomSpacer.tsx
+++ b/web/src/components/chat/DynamicBottomSpacer.tsx
@@ -298,6 +298,7 @@ const DynamicBottomSpacer = React.memo(
         ref={spacerRef}
         data-dynamic-spacer="true"
         aria-hidden="true"
+        className="w-full"
         style={{
           height: "0px",
           flexShrink: 0,


web/src/app/app/message/HumanMessage.tsx
```diff
@@ -215,7 +215,11 @@ const HumanMessage = React.memo(function HumanMessage({
               </div>
             </div>
             {onEdit && !isEditing && (
-              <div className="flex flex-row p-1 opacity-0 group-hover:opacity-100 transition-opacity">
+              // HACK: In order to standardize a gap-12 spacing between messages, we apply a
+              // negative margin to the copy and edit buttons. On smaller screens when they wrap
+              // below the message instead of extend to the left, this -mb-11 offsets the size of
+              // the buttons (2.75rem) meaning the only whitespace is being applied by the gap-12.
+              <div className="flex flex-row p-1 -mb-11 opacity-0 group-hover:opacity-100 transition-opacity">
```
nmgarza5: would rather we don't fall into a pattern of applying negative margins

nmgarza5: How about `h-0 min-h-0 overflow-visible` instead? The buttons render and are clickable but contribute `0px` to the parent layout, so `gap-12` stays consistent regardless of screen size — and no magic negative margin number tied to the button height. 

`min-h-0` overrides the flex item default `min-height: auto` which would otherwise prevent the height from collapsing. `relative z-content` keeps the overflowed buttons clickable above the next message.

```suggestion
              <div className="h-0 min-h-0 overflow-visible relative z-content flex flex-row p-1 opacity-0 group-hover:opacity-100 transition-opacity">
```

jmelahman:
@greptile plrease re-review

PR #8719, opened 2026-02-24
Fix MCP admin listing when OAuth client_id is empty

### LINKED ISSUE #8660: MCP admin endpoint crashes with 500 when OAuth client_id is empty (ValueError breaks entire server list)
Hello,
  
The admin endpoint `GET /admin/mcp/servers` always returns HTTP 500 if any MCP server has OAuth client info stored with an empty client_id (or client_secret). This makes the entire MCP admin page unusable.

  Logs
```
  INFO:  api.py 1633: Fetching all MCP servers for admin display
  ERROR: api.py 1647: Failed to fetch MCP servers for admin: <class 'ValueError'>:Stored client info had empty client ID or secret
  INFO:  h11_impl.py 473: "GET /admin/mcp/servers HTTP/1.1" 500
```
  The error repeats on every page load/refresh.

  Steps to Reproduce

  1. Add an MCP server with OAuth authentication (e.g. Attio)
  2. Go through the OAuth flow — dynamic client registration may store incomplete credentials
  3. The page fails to load entirely with a 500

  Root Cause

  In `backend/onyx/server/features/mcp/api.py`, `_db_mcp_server_to_api_mcp_server` raises a `ValueError` when the stored `OAuthClientInformationFull` has an empty `client_id`. This exception
   is not caught locally and propagates up to the top-level handler of get_mcp_servers_for_admin, causing the entire endpoint to fail — all MCP servers are unavailable, not just the broken one.

  The problematic pattern appears in two places in the function (for admin-performer and per-user-performer flows).

  Expected Behavior

  The endpoint should handle a missing/empty `client_id` gracefully: log a warning, return `admin_credentials = {}` for that server, and continue returning the rest of the server list.

  This is already the pattern used when `client_info` is `None` entirely — it just needs to be applied to the client_id empty string case too.

benwu408:
## Description

Changed the MCP API to not raise ValueError when encountered an empty OAuth client-id in MCP Server.

## Additional Options
This PR closes: https://github.com/onyx-dot-app/onyx/issues/8660

---
## Summary by cubic
Fixes MCP admin listing to handle OAuth configs with an empty client_id without crashing. Now returns empty admin_credentials and logs a warning; masking remains when credentials exist.

- **Bug Fixes**
  - Prevent ValueError in _db_mcp_server_to_api_mcp_server when client_id is empty; log a warning and return empty admin_credentials.
  - Keep masking for client_id and client_secret when present.
  - Add unit tests for admin and per-user performers covering empty client_id and null client_secret.

<sup>Written for commit 1ecbcd5406d4d5233f88db957ef585774464e961. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/unit/onyx/server/features/mcp/test_mcp_server_conversion.py
+++ b/backend/tests/unit/onyx/server/features/mcp/test_mcp_server_conversion.py
@@ -1,0 +1,169 @@
+"""Unit tests for _db_mcp_server_to_api_mcp_server handling of empty OAuth credentials.
+
+These tests verify that the conversion function handles empty/missing OAuth
+client_id gracefully (warning + empty credentials) rather than raising ValueError.
+"""
+
+from datetime import datetime
+from datetime import timezone
+from unittest.mock import MagicMock
+from unittest.mock import patch
+
+from onyx.auth.schemas import UserRole
+from onyx.db.enums import MCPAuthenticationPerformer
+from onyx.db.enums import MCPAuthenticationType
+from onyx.db.enums import MCPServerStatus
+from onyx.server.features.mcp.api import _db_mcp_server_to_api_mcp_server
+
+
+def _make_db_server(
+    auth_type: MCPAuthenticationType,
+    auth_performer: MCPAuthenticationPerformer,
+    admin_config_data: dict | None = None,
+) -> MagicMock:
+    """Create a mock DbMCPServer with the given auth settings."""
+    db_server = MagicMock()
+    db_server.id = 1
+    db_server.name = "Test Server"
+    db_server.description = "A test server"
+    db_server.server_url = "https://example.com/mcp"
+    db_server.owner = "admin@example.com"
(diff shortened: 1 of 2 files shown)


backend/tests/unit/onyx/server/features/mcp/test_mcp_server_conversion.py
```diff
@@ -0,0 +1,169 @@
...
+    db_server = MagicMock()
+    db_server.id = 1
+    db_server.name = "Test Server"
+    db_server.description = "A test server"
+    db_server.server_url = "https://example.com/mcp"
+    db_server.owner = "admin@example.com"
+    db_server.transport = None
+    db_server.auth_type = auth_type
+    db_server.auth_performer = auth_performer
+    db_server.status = MCPServerStatus.CONNECTED
```
justin-tahara: Are we assuming we are just connected from the beginning? Want to understand if this is the default setup that we want to setup

PR #8722, opened 2026-02-24
fix: make name in ConnectorCredentialPairMetadata required

ciaransweet:
## Description

Makes `ConnectorCredentialPairMetadata`'s `name` field required and not `| None = None`

Current docs (as per the Pydantic Model) state that it _can_ be None, especially in: https://cloud.onyx.app/api/docs#/public/associate_credential_to_connector

However, if not provided, you get:

```
(psycopg2.errors.NotNullViolation) null value in column "name" of relation "connector_credential_pair" violates not-null constraint
```

As it's `NotNull` in the database.

I also had a back and forth with `@OnyxBot` for a bit and seems like the frontend always provides a value anyway. So the biggest affected group is anyone interacting with the API _outside_ of Onyx's own services, which I imagine is _pretty_ small.

## How Has This Been Tested?

Confirmed generated docs no longer say optional:

image

## Additional Options

---
## Summary by cubic
Make name required in ConnectorCredentialPairMetadata to match the database NOT NULL. Docs now show it as required, and requests without it fail fast.

- **Migration**
  - External API clients must send name when associating a credential to a connector. The Onyx frontend already does this.

<sup>Written for commit 5b57e20686c9d89506e09ccf127c2576d1679797. Summary will update on new commits.</sup>


wenxi-onyx:
Thanks for the PR! Closing in favor of #9408, which builds on top of your commit to make a few more callers not-null.

PR #8727, opened 2026-02-24
fix: graph client env

evan-onyx:
## Description

pass correct environment to microsoft graph client based on configured authority url

## How Has This Been Tested?

We can't actually test the integration, but unit tested

## Additional Options

---
## Summary by cubic
Routes Microsoft Graph requests to the correct national cloud by deriving AzureEnvironment from configured host URLs. Ensures SharePoint and Teams use the right endpoints and fixes GCC High/DoD routing.

- **Bug Fixes**
  - Added resolve_microsoft_environment to map graph/authority hosts to AzureEnvironment and SharePoint domain suffix.
  - SharePoint and Teams now pass environment to GraphClient and normalize hosts; warn if provided SharePoint suffix differs.
  - Validate and error on unknown graph hosts or mismatched authority/graph pairs.
  - Added unit tests for environment resolution.

<sup>Written for commit 31d413be738c5481da644f16d2b4c9e8010ec792. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/unit/onyx/connectors/test_microsoft_graph_env.py
+++ b/backend/tests/unit/onyx/connectors/test_microsoft_graph_env.py
@@ -1,0 +1,52 @@
+import pytest
+from office365.graph_client import AzureEnvironment  # type: ignore[import-untyped]
+
+from onyx.connectors.exceptions import ConnectorValidationError
+from onyx.connectors.microsoft_graph_env import resolve_microsoft_environment
+
+
+def test_resolve_global_defaults() -> None:
+    env = resolve_microsoft_environment(
+        "https://graph.microsoft.com", "https://login.microsoftonline.com"
+    )
+    assert env.environment == AzureEnvironment.Global
+    assert env.sharepoint_domain_suffix == "sharepoint.com"
+
+
+def test_resolve_gcc_high() -> None:
+    env = resolve_microsoft_environment(
+        "https://graph.microsoft.us", "https://login.microsoftonline.us"
+    )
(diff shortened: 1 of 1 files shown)


backend/onyx/connectors/microsoft_graph_env.py
```diff
@@ -0,0 +1,96 @@
...
+        authority_host="https://login.chinacloudapi.cn",
+        sharepoint_domain_suffix="sharepoint.cn",
+    ),
+    MicrosoftGraphEnvironment(
+        environment=AzureEnvironment.Germany,
+        graph_host="https://graph.microsoft.de",
+        authority_host="https://login.microsoftonline.de",
+        sharepoint_domain_suffix="sharepoint.de",
+    ),
+]
```
justin-tahara, line 60: Is this the full list of envs? Or do we need to keep updating this in the future? 

Want to make sure that we are not missing stuff

justin-tahara: I'm also assuming that these work since I don't know how to test these properly

evan-onyx: It should be pretty rare that this gets updated, but yeah it's not my favorite solution. I'm actually doing this because the SDK has already shown itself to be not what we need for various reasons (we had to paginate manually for example, only a few versions ago they didn't properly support environments themself...), so I want to maintain the flexibility to move away from it entirely. If it wasn't a ton of work and failure surface, I'd honestly rip out all the SDK code right now 😅

PR #8731, opened 2026-02-24
fix(user invitation): failed but no warning.

Bo-Onyx:
## Description

Ticket Link: https://linear.app/onyx-app/issue/ENG-3505

  Problem

  When an admin invites users, email notifications could silently fail to send in two scenarios:
  - ENABLE_EMAIL_INVITES=true but no SMTP/SendGrid is configured
  - ENABLE_EMAIL_INVITES=true, email is configured, but sending fails (e.g. bad credentials)

  In both cases the invite was saved but the admin always saw "Users invited!" with no indication that the invited user never received a notification.

  Changes

  Backend
  - Added EmailInviteStatus enum and BulkInviteResponse model to models.py
  - Changed bulk_invite_users return type from int to BulkInviteResponse, explicitly tracking the email delivery outcome across all code paths
  - Fixed a pre-existing bug where EMAIL_CONFIGURED was never checked before attempting to send — placeholder default values for SMTP_USER/SMTP_PASS
  caused EMAIL_CONFIGURED to always evaluate as True

  Frontend
  - BulkAdd.tsx parses the response and passes email_invite_status to onSuccess
  - page.tsx shows a warning toast alongside "Users invited!" when status is not_configured or send_failed; disabled and sent show no warning
Screenshot 2026-02-24 at 4 07 17 PM

## How Has This Been Tested?

Added unit tests

  - Fixed pre-existing broken test (test_trial_tenant_can_invite_within_limit was missing enforce_seat_limit patch and asserting against the old int
  return type)
  - Added 4 new unit tests covering all EmailInviteStatus outcomes

Add local test on UI before change:
Add bulk invitations show no warning.
After changes 
Screenshot 2026-02-24 at 4 08 41 PM

## Additional Options

---
## Summary by cubic
Shows clear warnings when invite emails don’t send, and returns delivery status in the bulk invite API. Addresses Linear ENG-3505.

- **Bug Fixes**
  - Backend: added EmailInviteStatus and BulkInviteResponse; bulk_invite_users now always returns { invited_count, email_invite_status } with DISABLED/NOT_CONFIGURED/SENT/SEND_FAILED and checks EMAIL_CONFIGURED.
  - Config: removed default SMTP server/user/pass so EMAIL_CONFIGURED is accurate.
  - Frontend: BulkAdd parses JSON and passes email_invite_status to onSuccess; AddUserButton shows warning toasts for NOT_CONFIGURED and SEND_FAILED, otherwise success.
  - Tests: added coverage for all status outcomes; fixed trial invite limit test by patching enforce_seat_limit.

- **Migration**
  - bulk_invite_users now returns JSON { invited_count, email_invite_status } instead of an int; update callers accordingly.

<sup>Written for commit 59a0a31e9bb9e209cfacda39f2608da9fc36374e. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/components/admin/users/BulkAdd.tsx
+++ b/web/src/components/admin/users/BulkAdd.tsx
@@ -16,7 +16,11 @@ const addUsers = async (url: string, { arg }: { arg: Array<string> }) => {
   });
 };
 
-export type EmailInviteStatus = "SENT" | "NOT_CONFIGURED" | "SEND_FAILED" | "DISABLED";
+export type EmailInviteStatus =
+  | "SENT"
+  | "NOT_CONFIGURED"
+  | "SEND_FAILED"
+  | "DISABLED";
 
 interface FormProps {
   onSuccess: (emailInviteStatus: EmailInviteStatus) => void;


backend/onyx/configs/app_configs.py
```diff
@@ -210,10 +210,10 @@
 REQUIRE_EMAIL_VERIFICATION = (
     os.environ.get("REQUIRE_EMAIL_VERIFICATION", "").lower() == "true"
 )
-SMTP_SERVER = os.environ.get("SMTP_SERVER") or "smtp.gmail.com"
+SMTP_SERVER = os.environ.get("SMTP_SERVER") or ""
```
wenxi-onyx, line 213: You should verify that our cloud deployment has this variable set and/or doesn't need this. There's a chance it's not and could break email invites on cloud by removing default 😅

web/src/app/admin/users/page.tsx
```diff
@@ -267,12 +267,22 @@ const SearchableTables = () => {
 function AddUserButton() {
   const [bulkAddUsersModal, setBulkAddUsersModal] = useState(false);
 
-  const onSuccess = () => {
+  const onSuccess = (emailInviteStatus: string) => {
```
wenxi-onyx: pref `emailInviteStatus` should be a strictly typed string union

backend/onyx/server/manage/users.py
```diff
@@ -427,15 +430,24 @@ def bulk_invite_users(
...
             for email in emails_needing_seats:
                 send_user_email_invite(email, current_user, AUTH_TYPE)
+            email_invite_status = EmailInviteStatus.SENT
         except Exception as e:
             logger.error(f"Error sending email invite to invited users: {e}")
+            email_invite_status = EmailInviteStatus.SEND_FAILED
 
     if not MULTI_TENANT or DEV_MODE:
-        return number_of_invited_users
+        return BulkInviteResponse(
```
wenxi-onyx: nit: duplicate code. This could be
```
if MULTI_TENANT or not DEV_MODE:
   try
      control plane writing code
   except

return BulkInviteResponse
```

backend/onyx/configs/app_configs.py
wenxi-onyx: Actually, cloud uses sendgrid, so maybe it's fine --> might want to check with Justin about other managed deployments? Just in case customers expect email invites, but never set this value. Shouldn't be a big deal

backend/onyx/server/manage/users.py
Bo-Onyx: good point, the flip of "not MULTI_TENANT or DEV_MODE" is `MULTI_TENANT and not DEV_MODE`

updated

backend/onyx/configs/app_configs.py
Bo-Onyx: @justin-tahara would you please help review? thanks

justin-tahara: We're good. Our Cloud has `SMTP_SERVER: smtp.gmail.com` set

PR #8735, opened 2026-02-24
chore(fe): foldable buttons display text via tooltip when disabled

jmelahman:
## Description

The `Deep Research` button typically reveals the text when hovered, but when it's disabled, so is this animation. In spite of that, show a tooltip with the same text when disabled.

When all of these buttons are disabled, it's the only one in the InputBar without a tooltip/hover animation.

20260224_14h40m46s_grim

## Additional Options

---
## Summary by cubic
Disabled foldable buttons now show their label in a tooltip by default, so the Deep Research button matches other InputBar controls and stays discoverable. If a tooltip prop is provided, it’s used; non-foldable buttons are unchanged.

<sup>Written for commit 8146787945c9b08e7254eac56507f675911eb485. Summary will update on new commits.</sup>

### DIFF
--- a/web/lib/opal/src/components/buttons/Button/components.tsx
+++ b/web/lib/opal/src/components/buttons/Button/components.tsx
@@ -162,7 +162,9 @@ function Button({
 
   const resolvedTooltip =
     tooltip ??
-    (interactiveBaseProps.disabled && children ? children : undefined);
+    (foldable && interactiveBaseProps.disabled && children
+      ? children
+      : undefined);
 
   if (!resolvedTooltip) return button;
 


web/lib/opal/src/components/buttons/Button/components.tsx
```diff
@@ -160,7 +160,11 @@ function Button({
     </Interactive.Base>
   );
 
-  if (!tooltip) return button;
+  const resolvedTooltip =
+    tooltip ??
+    (interactiveBaseProps.disabled && children ? children : undefined);
```
jmelahman: Per offline, we only want to apply this to `foldable` buttons

PR #8746, opened 2026-02-25
feat(scim): field round-tripping for IdP attribute preservation [2/3]

nmgarza5:
## Description

Second PR in the SCIM 2.0 Entra ID compatibility stack. Stacked on #8745. Ensures IdP-provided user attributes (name components, email metadata, department, manager) are persisted and echoed back verbatim on subsequent GET/LIST requests, which Entra ID requires for provisioning validation.

Adds field round-tripping so that IdP-provided attributes (name components, email metadata, enterprise extension fields) are persisted and returned verbatim in subsequent GET responses.

**Key changes:**
- `ScimMappingFields` dataclass — carries stored mapping fields (given_name, family_name, scim_emails_json, department, manager)
- `build_scim_name` concrete method on base provider — round-trips stored name components, falls back to splitting `personal_name` for legacy users
- Email serialization/deserialization (`serialize_emails`, `_deserialize_emails`)
- `user_schemas` property on base provider for schema extensibility
- DAL `create_user_mapping` and `sync_user_external_id` accept `fields` parameter
- `_mapping_to_fields` and `_fields_from_resource` helpers in api.py
- All user endpoints (`list`, `get`, `create`, `replace`, `patch`) thread `fields=` through

**Stack:** [PR 1/3: protocol compliance] → PR 2/3 → [PR 3/3: Entra support]

## How Has This Been Tested?

see 3rd pr in stack

## Additional Options

### DIFF
--- a/backend/ee/onyx/server/scim/api.py
+++ b/backend/ee/onyx/server/scim/api.py
@@ -31,6 +31,7 @@ from ee.onyx.server.scim.models import ScimError
 from ee.onyx.server.scim.models import ScimGroupMember
 from ee.onyx.server.scim.models import ScimGroupResource
 from ee.onyx.server.scim.models import ScimListResponse
+from ee.onyx.server.scim.models import ScimMappingFields
 from ee.onyx.server.scim.models import ScimName
 from ee.onyx.server.scim.models import ScimPatchRequest
 from ee.onyx.server.scim.models import ScimServiceProviderConfig
@@ -40,6 +41,7 @@ from ee.onyx.server.scim.patch import apply_user_patch
 from ee.onyx.server.scim.patch import ScimPatchError
 from ee.onyx.server.scim.providers.base import get_default_provider
 from ee.onyx.server.scim.providers.base import ScimProvider
+from ee.onyx.server.scim.providers.base import serialize_emails
 from ee.onyx.server.scim.schema_definitions import GROUP_RESOURCE_TYPE
 from ee.onyx.server.scim.schema_definitions import GROUP_SCHEMA_DEF
 from ee.onyx.server.scim.schema_definitions import SERVICE_PROVIDER_CONFIG
@@ -47,6 +49,7 @@ from ee.onyx.server.scim.schema_definitions import USER_RESOURCE_TYPE
 from ee.onyx.server.scim.schema_definitions import USER_SCHEMA_DEF
 from onyx.db.engine.sql_engine import get_session
 from onyx.db.models import ScimToken
+from onyx.db.models import ScimUserMapping
 from onyx.db.models import User
 from onyx.db.models import UserGroup
 from onyx.db.models import UserRole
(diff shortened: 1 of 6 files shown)


backend/ee/onyx/server/scim/models.py
```diff
@@ -71,6 +72,23 @@ class ScimUserGroupRef(BaseModel):
     display: str | None = None
 
 
+@dataclass
+class ScimMappingFields:
```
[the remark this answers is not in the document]
nmgarza5: `ScimMappingFields` is a simple data transfer container — it shuttles field values between the API layer and the DAL. It has no validation logic, no serialization behavior, and no public API. A dataclass is the right tool here; Pydantic would add unnecessary overhead for a struct that's never parsed from external input.

backend/ee/onyx/server/scim/api.py
```diff
@@ -229,6 +232,30 @@ def _scim_resource_response(
     )
 
 
+def _mapping_to_fields(
+    mapping: ScimUserMapping | None,
+) -> ScimMappingFields | None:
+    """Extract round-trip fields from a SCIM user mapping."""
+    if not mapping:
+        return None
+    return ScimMappingFields(
```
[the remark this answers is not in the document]
nmgarza5: This is intentional — department/manager extraction from the enterprise extension is added in PR #3 (nik/eng-3652-entra-support), which stacks on top of this PR. This PR lays the groundwork with the `ScimMappingFields` container and DAL columns; PR #3 wires up the enterprise extension parsing. The stacked PR structure keeps each diff focused.

nmgarza5:
@greptile

backend/ee/onyx/server/scim/models.py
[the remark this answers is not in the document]
nmgarza5: intentional in this case.

backend/ee/onyx/server/scim/providers/base.py
```diff
@@ -106,9 +129,24 @@ def build_group_resource(
...
+        fields: ScimMappingFields,
+    ) -> ScimName | None:
+        """Build SCIM name components for the response.
+
+        Round-trips stored ``given_name``/``family_name`` when available (so
+        the IdP gets back what it sent). Falls back to splitting
+        ``personal_name`` for users provisioned before we stored components.
+        Providers may override for custom behavior.
+        """
+        if fields.given_name is not None or fields.family_name is not None:
```
justin-tahara, line 144: Can we just do:

```suggestion
        if fields.given_name and fields.family_name:
```

or am I bugging?

backend/ee/onyx/server/scim/api.py
```diff
@@ -261,6 +264,30 @@ def _build_list_response(
...
+        family_name=mapping.family_name,
+        scim_emails_json=mapping.scim_emails_json,
+    )
+
+
+def _fields_from_resource(resource: ScimUserResource) -> ScimMappingFields:
+    """Build mapping fields from an incoming SCIM user resource."""
+    return ScimMappingFields(
+        given_name=resource.name.givenName if resource.name else None,
+        family_name=resource.name.familyName if resource.name else None,
```
justin-tahara, line 286: If the name resource is not none, does that mean it will always have givenName and familyName?

backend/ee/onyx/server/scim/providers/base.py
nmgarza5: `is not None` distinguishes "never set" from "explicitly sent empty string". An IdP could send `{"givenName": ""}` to clear a name, and we should round-trip  that empty string rather than falling through to the personal_name split fallback (current behavior if fields not set)

backend/ee/onyx/server/scim/api.py
nmgarza5: The guard is just to avoid `AttributeError` when name is None — if name exists but both fields are None, we still store None for each, which is correct and still valid SCIM (IdP sent a name object with no components).

backend/ee/onyx/server/scim/providers/base.py
nmgarza5: Needs to stay `is not None` with `or` — an IdP can send just `givenName` without `familyName` (or vice versa), and we should round-trip whichever was provided. Using `and` would require both to be present. And `is not None` distinguishes "never set" (`None`) from "explicitly sent empty" (`""`), which is a valid SCIM value.

PR #8749, opened 2026-02-25
feat: python tool not added when no code interpretter server

Danelegend:
## Description
When the code interpreter server is disabled, we should not show or allow the python tool call.

## How Has This Been Tested?
Unit + Integration tests
Manual

## Additional Options

---
## Summary by cubic
Hide and disable the Python tool when the code interpreter server is not configured or is turned off. Users won’t see or call run_python unless the server is enabled.

- **Bug Fixes**
  - PythonTool.is_available now returns true only when CODE_INTERPRETER_BASE_URL is set and the DB server_enabled flag is true (via fetch_code_interpreter_server).
  - Tests: unit tests for base URL and flag cases; an external-dependency unit test validates DB toggling; integration tests verify GET /tool visibility and add a fixture to restore server state between runs.

<sup>Written for commit 54ec892bae879ef2d68a71c779d70e5e6bd4f949. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/integration/tests/code_interpreter/test_code_interpreter_api.py
+++ b/backend/tests/integration/tests/code_interpreter/test_code_interpreter_api.py
@@ -1,12 +1,10 @@
 import requests
 
 from tests.integration.common_utils.constants import API_SERVER_URL
-from tests.integration.common_utils.managers.tool import ToolManager
 from tests.integration.common_utils.test_models import DATestUser
 
 CODE_INTERPRETER_URL = f"{API_SERVER_URL}/admin/code-interpreter"
 CODE_INTERPRETER_HEALTH_URL = f"{CODE_INTERPRETER_URL}/health"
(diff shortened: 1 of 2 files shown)


backend/tests/integration/tests/code_interpreter/test_code_interpreter_api.py
```diff
@@ -94,3 +96,35 @@ def test_code_interpreter_endpoints_require_admin(
...
+
+
+def test_python_tool_hidden_from_tool_list_when_disabled(
+    admin_user: DATestUser,
+) -> None:
+    """When code interpreter is disabled, the Python tool should not appear
+    in the GET /tool response (i.e. the frontend tool list)."""
+    # Disable
+    response = requests.put(
+        CODE_INTERPRETER_URL,
```
evan-onyx: cubic is right that we should make sure this is part of the integration test reset

PR #8750, opened 2026-02-25
fix(actions): persist source selection when toggling search tool in ActionsPopover

Subash-Mohan:
## Description
- Fixed source selection not persisting to localStorage when toggling the search tool on/off in the ActionsPopover

## How Has This Been Tested?
Tested by first disabling all the sources; at that point, the internal search tool was disabled. When I enabled the tool and refreshed the page, it stays enabled. Previously, it went back to a disabled state again.

## Additional Options

---
## Summary by cubic
Persist source selection when toggling the search tool in ActionsPopover, so the tool’s state and selected sources survive page reloads. Fixes the regression where the search tool reverted to disabled after refresh.

- **Bug Fixes**
  - Use enableSources(...) and baseDisableAllSources(...) to write changes to localStorage.
  - On enable: restore previous sources or enable all; on disable: save current sources, then disable all.

<sup>Written for commit 07b366285af0a24638c8861254249940042b785a. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/refresh-components/popovers/ActionsPopover/index.tsx
+++ b/web/src/refresh-components/popovers/ActionsPopover/index.tsx
@@ -816,18 +816,18 @@ export default function ActionsPopover({
 
     if (toolId === searchToolId) {
       if (wasDisabled) {
-        // Enabling - restore previous sources or enable all (no persistence)
+        // Enabling - restore previous sources or enable all (persisted to localStorage)
         const previous = previouslyEnabledSourcesRef.current;
         if (previous.length > 0) {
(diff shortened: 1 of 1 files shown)


Danelegend (approved):
Has this been tested with multiple and various scenarios?
ie. Suppose we have 2 internal sources hooked up. Select 1, press disable all, press enable all and expect that one to come up again. 

Likewise, you should be able to disable internal search, refresh and see it come back again.

Reason for asking is a feel like there was a scenario where this approach didn't work. But it could have just been a skill issue on me.

PR #8764, opened 2026-02-25
chore(fe): opal button implements responsiveHideText

jmelahman:
## Description

We've concluded all buttons should have some text associated with them. This is incompatible with this pattern of conditionally passing text on `isMobile`. Instead, implement this functionality inside the opal button. This is also nicer since it only uses CSS media queries.

## How Has This Been Tested?

20260225_11h55m55s_grim

---
## Summary by cubic
Adds a responsiveHideText prop to the Opal Button to hide the label and trailing icon on small screens using CSS, removing isMobile checks in consumers. Updates the “Share Chat” button to use this pattern while keeping text in the markup.

- **New Features**
  - Button prop: responsiveHideText (default false).
  - Hides label and right icon below md via CSS (shown at md and up).

- **Refactors**
  - Header “Share Chat” now uses icon, prominence="tertiary", and responsiveHideText; text always provided.
  - Tightened ButtonContentProps to allow responsiveHideText only with icon; adjusted imports to use Opal Button and kept RefreshButton for modal submit.

<sup>Written for commit bc7ab13fa6835e63f445272433143950d88ef5aa. Summary will update on new commits.</sup>

### DIFF
--- a/web/lib/opal/src/components/buttons/Button/components.tsx
+++ b/web/lib/opal/src/components/buttons/Button/components.tsx
@@ -59,18 +59,21 @@ type ButtonContentProps =
       icon: IconFunctionComponent;
       children: string;
       rightIcon?: IconFunctionComponent;
+      responsiveHideText?: never;
     }
   | {
       foldable?: false;
       icon?: IconFunctionComponent;
       children: string;
       rightIcon?: IconFunctionComponent;
+      responsiveHideText?: never;
     }
   | {
       foldable?: false;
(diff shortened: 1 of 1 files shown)


raunakab (approved):
Spoke offline. We agreed to move the `responsiveHideText?: boolean` to ONE specific type-discriminant inside of `ButtonContentProps` (and the other ones will have `responsiveHideText: never`) to enforce type-based discrimination.

PR #8767, opened 2026-02-25
fix(slack): sanitize HTML tags and broken citation links in bot responses

nmgarza5:
## Description

Comprehensive fix for Slack bot message formatting:

1. **HTML tags appearing as literal text** — LLM responses containing HTML from documentation sources (SharePoint, Confluence, Google Drive, etc.) showed `<br>`, `<p>`, `<div>` etc. as raw text in Slack. Block-level closing tags and `<br>` are now converted to newlines; all other HTML tags are stripped. Code blocks are preserved intact, and autolinks (`<https://...>`) are not touched.

2. **Broken citation/markdown links** — Links with special characters in URLs (parentheses, spaces) caused mistune to truncate URLs at the first `)`. All markdown link destinations are now wrapped in angle brackets (`[text](<url>)`) using a balanced-parenthesis parser so URLs with special characters render correctly.

3. **`&quot;` appearing as literal text** — `HTMLRenderer.text()` escapes `"` to `&quot;`, which Slack doesn't recognize. Overridden to only escape the three entities Slack supports: `&`, `<`, `>`.

4. **Slack-style links rendered as literal text** — LLMs sometimes output Slack mrkdwn `<url|text>` syntax directly. Mistune doesn't recognise this, so the angle brackets were escaped and Slack showed raw text instead of clickable links. Now pre-processed into standard markdown before rendering.

5. **Poor visual spacing** — Block-level elements (paragraphs, headings, lists, code blocks, quotes) now end with double newlines for proper visual separation in Slack.

6. **Extra blank line in code blocks** — Mistune's `code` parameter includes a trailing newline, which combined with our template produced a blank line at the bottom of every code block. Now stripped.

7. **Uncited reference documents shown** — Removed the fallback that displayed all top search results as "Reference Documents" when the LLM didn't cite any sources. Documents now only appear when actually cited.

Also overrides all remaining `HTMLRenderer` methods (`linebreak`, `thematic_break`, `block_quote`, `block_html`, `block_error`, `text`, `paragraph`) to ensure no raw HTML ever appears in Slack output.

## Additional Options

### DIFF
--- a/backend/onyx/onyxbot/slack/formatting.py
+++ b/backend/onyx/onyxbot/slack/formatting.py
@@ -1,4 +1,5 @@
 import re
+from collections.abc import Callable
 from typing import Any
 
 from mistune import create_markdown
@@ -39,23 +40,20 @@ def _sanitize_html(text: str) -> str:
     return text
 
 
-def _sanitize_for_slack(message: str) -> str:
-    """Pre-process LLM output before markdown rendering for Slack compatibility.
-
-    Strips HTML tags from documentation sources (SharePoint, Confluence, Google
-    Drive, etc.) that would otherwise appear as literal escaped text in Slack.
-    Code blocks are preserved intact to avoid mangling code samples containing HTML.
-    """
+def _transform_outside_code_blocks(
+    message: str, transform: Callable[[str], str]
+) -> str:
+    """Apply *transform* only to text outside fenced code blocks."""
     parts = _FENCED_CODE_BLOCK_PATTERN.split(message)
     code_blocks = _FENCED_CODE_BLOCK_PATTERN.findall(message)
 
-    sanitized: list[str] = []
+    result: list[str] = []
     for i, part in enumerate(parts):
-        sanitized.append(_sanitize_html(part))
+        result.append(transform(part))
         if i < len(code_blocks):
-            sanitized.append(code_blocks[i])
+            result.append(code_blocks[i])
 
-    return "".join(sanitized)
+    return "".join(result)
 
 
 def _extract_link_destination(message: str, start_idx: int) -> tuple[str, int | None]:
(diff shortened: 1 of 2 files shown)


backend/onyx/onyxbot/slack/blocks.py
justin-tahara, line 1: This is for the raw HTML stuff? Just want to make sure I understand why we are removing this

backend/onyx/onyxbot/slack/formatting.py
```diff
@@ -1,10 +1,59 @@
...
+_HTML_TAG_PATTERN = re.compile(
+    r"<(?!https?://|mailto:)/?[a-zA-Z][^>]*>",
+)
+
+# Matches fenced code blocks (``` ... ```) so we can skip sanitization inside them
+_FENCED_CODE_BLOCK_PATTERN = re.compile(r"```[\s\S]*?```")
+
+# Matches the start of any markdown link: [text]( or [[n]](
+# The inner group handles nested brackets for citation links like [[1]](.
+_MARKDOWN_LINK_PATTERN = re.compile(r"\[(?:[^\[\]]|\[[^\]]*\])*\]\(")
```
justin-tahara, line 24: I'd be curious to know if we do any of this stuff else where. Seems like something that we would have hit before but it's just a bit odd of a workflow

backend/onyx/onyxbot/slack/blocks.py
nmgarza5: these were specifically being used as document blocks to surface when no relevant sources are found

backend/onyx/onyxbot/slack/formatting.py
nmgarza5: There's \`html_utils.py\` in \`file_processing/\` that does full HTML-to-text conversion via BeautifulSoup — but it's built for indexing whole HTML pages from connectors (SharePoint, Confluence, etc.), not for pre-processing LLM output fragments before Slack rendering. BeautifulSoup parses into a full DOM tree which is overkill here — we're just stripping occasional HTML tags that leak through from documentation sources in the LLM's answer. Lightweight regexes are a better fit for this context.

backend/onyx/onyxbot/slack/blocks.py
nmgarza5: No, this is separate from the HTML stuff. This removes the fallback that showed top search results as "Reference Documents" when the LLM didn't cite anything. Cited sources still display — only the uncited fallback is removed.

nmgarza5: Intentional change — when the LLM cites sources, those citations still show up via \`_build_citations_blocks\`. This only removes the fallback that dumped the top search results when the LLM didn't cite anything, which was surfacing irrelevant documents.

PR #8776, opened 2026-02-25
chore(opensearch): MT cloud should verify index on document index init, and do cluster setup once at start

acaprau:
## Description
The state of the code before this PR is that we assume that when the product first starts we verify and setup the index if necessary in OpenSearch based on Onyx's current search settings. In MT cloud this does not work as each tenant has its own search settings. So this PR does the following:
- If we are on MT cloud (and the flag `VERIFY_CREATE_OPENSEARCH_INDEX_ON_INIT_MT` is enabled, which is is by default) then every time we create an `OpenSearchDocumentIndex` we verify and setup the index if necessary.
    - This adds a small network overhead each instantiation to check for index existence, and a non-trivial overhead each time a new index is created, which should be a one-time cost of roughly 30 creations for the entire MT cloud cluster across all tenants.
- If we are on MT cloud, on cluster setup we put cluster settings and search pipelines. This essentially achieves functional parity with every other Onyx deployment type with respect to cluster settings and search pipelines.
- Only if we are not on MT cloud, we put cluster settings and search pipelines while also verifying and possibly creating an index, which is exactly what was happening before this PR.

## How Has This Been Tested?
I am relying on existing automated tests to continue passing. Eventually we will want to do a quick manual test in an MT environment for all this stuff, plus the migration task, but this was not done in this PR.

## Additional Options

---
## Summary by cubic
Verify and create OpenSearch indices per tenant on init in MT cloud, and move cluster settings/pipeline setup to a one-time startup step. This prevents missing indices and aligns MT behavior with other deployments while simplifying client responsibilities.

- **New Features**
  - In MT cloud, each OpenSearchDocumentIndex init verifies/creates the index when VERIFY_CREATE_OPENSEARCH_INDEX_ON_INIT_MT=true (default). Index creation is a one-time cost per index.
  - On MT cluster startup, wait for OpenSearch and apply cluster settings and create search pipelines once when ENABLE_OPENSEARCH_INDEXING_FOR_ONYX is enabled. Non‑MT still sets cluster settings/pipelines during per-index verification, matching previous behavior.

- **Refactors**
  - Split client into OpenSearchClient (cluster ops) and OpenSearchIndexClient (index ops); updated usages and tests.
  - Use IndexingSetting to pass index_name, embedding_dim, and embedding_precision to OpenSearch constructors and migration tasks; OpenSearchDocumentIndex and factory now supply these on init.

<sup>Written for commit 193e46381a775d980d1a5670ca80bbdf10f8693c. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/background/celery/tasks/opensearch_migration/tasks.py
+++ b/backend/onyx/background/celery/tasks/opensearch_migration/tasks.py
@@ -153,7 +153,7 @@ def migrate_chunks_from_vespa_to_opensearch_task(
             indexing_setting = IndexingSetting.from_db_model(search_settings)
             opensearch_document_index = OpenSearchDocumentIndex(
                 tenant_state=tenant_state,
-                index_name=indexing_setting.index_name,
+                index_name=search_settings.index_name,
                 embedding_dim=indexing_setting.final_embedding_dim,
                 embedding_precision=indexing_setting.embedding_precision,
             )


backend/onyx/document_index/opensearch/client.py
```diff
@@ -125,6 +128,136 @@ def __init__(
             # your request body that is less than this value.
             timeout=timeout,
         )
+
+    @log_function_time(print_only=True, debug_only=True, include_args=True)
+    def create_search_pipeline(
```
acaprau, line 133: these functions are a pure move from the `OpenSearchIndexClient` class to this one, no functional change.

backend/onyx/document_index/opensearch/client.py
```diff
@@ -125,6 +128,136 @@ def __init__(
...
+        TODO(andrei): Can we have some way to auto close when the client no
+        longer has any references?
+
+        Raises:
+            Exception: There was an error closing the client.
+        """
+        self._client.close()
+
+
+class OpenSearchIndexClient(OpenSearchClient):
```
acaprau, line 212: this class now just extends the above, no other change.

backend/onyx/document_index/opensearch/client.py
```diff
@@ -948,10 +1002,7 @@ def wait_for_opensearch_with_timeout(
     made_client = False
     try:
         if client is None:
-            # NOTE: index_name does not matter because we are only using this object
-            # to ping.
-            # TODO(andrei): Make this better.
```
acaprau, line 953: we did it.

backend/onyx/document_index/opensearch/opensearch_document_index.py
```diff
@@ -258,19 +279,17 @@ def __init__(
             index_name=index_name,
             secondary_index_name=secondary_index_name,
```
acaprau: big if true.

backend/onyx/document_index/opensearch/opensearch_document_index.py
```diff
@@ -471,19 +484,37 @@ class OpenSearchDocumentIndex(DocumentIndex):
...
-        index_name: str,
         tenant_state: TenantState,
+        index_name: str,
+        embedding_dim: int,
+        embedding_precision: EmbeddingPrecision,
     ) -> None:
         self._index_name: str = index_name
         self._tenant_state: TenantState = tenant_state
-        self._os_client = OpenSearchClient(index_name=self._index_name)
+        self._client = OpenSearchIndexClient(index_name=self._index_name)
```
acaprau, line 512: simple rename bc my bp would rise each time i see `os`.

backend/tests/external_dependency_unit/opensearch/test_opensearch_client.py
```diff
@@ -1,4 +1,4 @@
-"""External dependency unit tests for OpenSearchClient.
+"""External dependency unit tests for OpenSearchIndexClient.
```
acaprau, line 1: simple type change, there is no functional change in this file.

backend/tests/external_dependency_unit/opensearch_migration/test_opensearch_migration_tasks.py
```diff
@@ -37,6 +37,7 @@
 from onyx.db.search_settings import get_active_search_settings
 from onyx.document_index.interfaces_new import TenantState
 from onyx.document_index.opensearch.client import OpenSearchClient
+from onyx.document_index.opensearch.client import OpenSearchIndexClient
```
acaprau, line 40: simple type change, there is no functional change in this file.

backend/onyx/document_index/opensearch/opensearch_document_index.py
```diff
@@ -258,19 +279,17 @@ def __init__(
             index_name=index_name,
             secondary_index_name=secondary_index_name,
         )
```
acaprau, line 260: big if true.

evan-onyx (approved):

PR #8781, opened 2026-02-26
chore(opensearch): OpenSearchClient implements context manager, also closes on del

acaprau:
## Description
This lets us use the client in a context manager session and also calls close on the underlying OpenSearch library client when it is garbage collected.

## How Has This Been Tested?
'twas not, other than automated testing.

## Additional Options

---
## Summary by cubic
OpenSearchClient now implements a context manager and automatically closes the underlying client on context exit and when the object is deleted. wait_for_opensearch_with_timeout now uses a with block (nullcontext if a client is passed) and removes the manual close/finally logic.

<sup>Written for commit 2dbd51b607166a7a2d86714e416bc8de801043f6. Summary will update on new commits.</sup>


backend/onyx/document_index/opensearch/client.py
```diff
@@ -999,11 +1007,7 @@ def wait_for_opensearch_with_timeout(
     Returns:
         True if OpenSearch is ready, False otherwise.
     """
-    made_client = False
-    try:
-        if client is None:
-            client = OpenSearchClient()
-            made_client = True
+    with nullcontext(client) if client else OpenSearchClient() as client:
```
evan-onyx, line 1010: you shouldn't need nullcontext here? and won't this cause the client to not use its enter/exit?

acaprau: that's intended, we want the client which is passed in to be untouched in that case. if no client is passed in at all, then yes use a real context manager.

PR #8787, opened 2026-02-26
fix(billing): handle manual license users without Stripe subscription

nmgarza5:
## Description

Closes: https://linear.app/onyx-app/issue/ENG-3752/billing-page-cleanup

Fixes billing page for customers who received a manual license but don't pay through Stripe. Previously, these users saw broken "Manage Plan" and "Update Seats" buttons that failed silently.

Changes:
- Add `isManualLicenseOnly` derived state to distinguish manual-license-without-Stripe from air-gapped and normal flows
- Replace "Manage Plan" button with sales contact text for manual license users
- Hide "Update Seats" button entirely for manual license users
- Move LicenseActivationCard inline between subscription and seats cards (instead of in footer)
- Restore back button on Plans page for users with a license (manual or otherwise)
- Add `hideClose` prop to LicenseActivationCard for permanently displayed inline usage

## How Has This Been Tested?

manually uploaded a license locally 

Screenshot 2026-02-25 at 8 12 44 PM

## Additional Options

### DIFF
--- a/web/src/app/admin/billing/BillingDetailsView.tsx
+++ b/web/src/app/admin/billing/BillingDetailsView.tsx
@@ -17,6 +17,7 @@ import {
   SvgPlus,
   SvgWallet,
   SvgFileText,
+  SvgOrganization,
 } from "@opal/icons";
 import { BillingInformation, LicenseStatus } from "@/lib/billing/interfaces";
 import {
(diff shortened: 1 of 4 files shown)


web/src/app/admin/billing/BillingDetailsView.tsx
```diff
@@ -211,7 +215,7 @@ function SubscriptionCard({
         height="auto"
       >
         <Section gap={0.25} alignItems="start" height="auto" width="auto">
-          <SvgUsers className="w-5 h-5 stroke-text-03" />
+          <PlanIcon className="w-5 h-5" />
```
Subash-Mohan, line 218: Don't we need to set icon colour here?

nmgarza5: No after looking at the mocks again, it shouldn't have been added in the first place

PR #8794, opened 2026-02-26
chore(opensearch): Allow programatic schema updates

acaprau:
## Description
If someone adds fields to the schema the product should still work automatically. This does not include invalid changes to the schema, such as changing the type of a field. The product will still fail on start in this case.

## How Has This Been Tested?
Added tests for the client `put_index` method.

## Additional Options

---
## Summary by cubic
Automatically applies OpenSearch mapping updates so new fields are added without manual reindexing. Incompatible type changes still fail fast and require reindexing.

- **New Features**
  - Added OpenSearchIndexClient.put_mapping for idempotent mapping updates (no-op for same fields, adds new fields, errors on type changes).
  - verify_and_create_index_if_necessary now updates mappings when the index exists, instead of only validating.

- **Tests**
  - Coverage for idempotency, adding new fields, failure on type changes, and non-existent index behavior.

<sup>Written for commit 326123c4b2163a33b5b29e211405e64f59b2c6dc. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/document_index/opensearch/client.py
+++ b/backend/onyx/document_index/opensearch/client.py
@@ -329,17 +329,17 @@ class OpenSearchIndexClient(OpenSearchClient):
     def put_mapping(self, mappings: dict[str, Any]) -> None:
         """Updates the index mapping in an idempotent manner.
 
-        This operation is idempotent:
-        - Existing fields with the same definition: No-op (succeeds silently)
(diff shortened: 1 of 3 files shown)


acaprau:
@greptile @cubic

yuhongsun96 (approved):

PR #8798, opened 2026-02-26
feat(opal): add Hoverable compound component

raunakab:
## Description

Adds a reusable `Hoverable` compound component to `@opal/core` for hover-to-reveal patterns (e.g., showing a trash icon when hovering over a card).

**Components:**
- `Hoverable.Item` — The core abstraction. On its own (no `group`), it applies local CSS `:hover` for the variant effect (e.g., `opacity-on-hover`). When `group` is specified, it reads hover state from the nearest matching `Hoverable.Root` — and **throws** if no matching Root is found.
- `Hoverable.Root` — Wraps children in a hover-tracking container for a named group. Provides hover state to descendant `Hoverable.Item`s with the same group name via React context.

**Behavior matrix:**

| Usage | Result |
|---|---|
| `<Hoverable.Item variant="opacity-on-hover">` | Local mode — CSS `:hover` on the item itself |
| `<Hoverable.Item group="x">` (no matching Root) | **Error** thrown |
| `<Hoverable.Root group="a">` → `<Hoverable.Item group="b">` | **Error** — group names don't match |
| `<Hoverable.Root group="x">` → `<Hoverable.Item group="x">` | Group mode — Root hover controls item visibility |

**Key design decisions:**
- Uses a **context-per-group** approach (`getOrCreateContext(group)`) so unrelated groups don't cause unnecessary re-renders
- Nesting works correctly — each inner `Root` shadows the parent's context provider
- Follows the existing opal data-attribute pattern for CSS styling (`data-hoverable-variant`, `data-hoverable-active`, `data-hoverable-local`)

## Rationale

We currently are doing a LOT of this:

```tsx
<div className="group/my-component">
  {/* ... */}
  <div className="opacity-0 group-hover/my-component:opacity-100 transition-all duration-200" />
  {/* ... */}
</div>
```

This means that each instance of hovering transitions have custom styling associated with them. This PR alleviates that problem by standardizing hoverable interactions.

```tsx
// Local mode — hovering the item itself reveals it
<Hoverable.Item variant="opacity-on-hover">
  <TrashIcon />
</Hoverable.Item>

// Group mode — hovering the root reveals the item
<Hoverable.Root group="my-component">
  {/* ... */}
  <Hoverable.Item group="my-component" variant="opacity-on-hover">
    <TrashIcon />
  </Hoverable.Item>
  {/* ... */}
</Hoverable.Root>
```

## Video

https://github.com/user-attachments/assets/38f213b0-650d-4873-b877-d8bd4fb916e6

## Additional Options

### DIFF
--- a/web/lib/opal/src/core/hoverable/components.tsx
+++ b/web/lib/opal/src/core/hoverable/components.tsx
@@ -74,10 +74,30 @@ interface HoverableItemProps
  * </Hoverable.Root>
  * ```
  */
-function HoverableRoot({ group, children, ...props }: HoverableRootProps) {
+function HoverableRoot({
+  group,
+  children,
+  onMouseEnter: consumerMouseEnter,
+  onMouseLeave: consumerMouseLeave,
+  ...props
+}: HoverableRootProps) {
   const [hovered, setHovered] = useState(false);
-  const onMouseEnter = useCallback(() => setHovered(true), []);
-  const onMouseLeave = useCallback(() => setHovered(false), []);
+
+  const onMouseEnter = useCallback(
+    (e: React.MouseEvent<HTMLDivElement>) => {
+      setHovered(true);
+      consumerMouseEnter?.(e);
+    },
+    [consumerMouseEnter]
+  );
+
+  const onMouseLeave = useCallback(
+    (e: React.MouseEvent<HTMLDivElement>) => {
+      setHovered(false);
+      consumerMouseLeave?.(e);
+    },
+    [consumerMouseLeave]
+  );
 
   const GroupContext = getOrCreateContext(group);
 


web/lib/opal/src/core/hoverable/components.tsx
```diff
@@ -0,0 +1,169 @@
...
+ * Provides two sub-components:
+ *
+ * - `Hoverable.Root` — A container that tracks hover state for a named group
+ *   and provides it via React context.
+ *
+ * - `Hoverable.Item` — An element that reads hover state from the nearest
+ *   matching `Hoverable.Root` and applies variant styles (e.g., opacity fade-in).
+ *
+ * Supports nesting: a child `Hoverable.Root` shadows the parent's context,
+ * so each group's items only respond to their own root's hover.
```
nmgarza5, line 201: Is the root solely so you can control the icon hover state from an outer container?

Can I use it without the root?

raunakab: Yup, updated the code such that it can be used without the `Root` too. Just the `Hoverable.Item` component works in isolation.

PR #8809, opened 2026-02-26
chore(fe): support featured agents w/o being public

jmelahman:
## Description

Closes https://linear.app/onyx-app/issue/ENG-3605/share-agent-menu-add-label-and-show-public-agent-message

## How Has This Been Tested?

featured agent shared only with user `admin2_user`

20260226_12h01m48s_grim

info message in modal

20260226_12h02m51s_grim

---
## Summary by cubic
Allow agents to be featured/default without forcing them public, and show a clear org-public notice in the Share Agent modal. Implements ENG-3605’s “show public agent message” requirement.

- **New Features**
  - Backend: allow default personas to be private on create and when toggling default (removed is_public validation and auto-set).
  - Share Agent modal: show “This agent is public to your organization.” banner when isPublic is true (non-dismissible).
  - Message component: add iconComponent prop typed as IconFunctionComponent to allow custom icons for the banner.

<sup>Written for commit 3d16272aac15ce078980039b8974a4deaaa17088. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/refresh-components/messages/Message.tsx
+++ b/web/src/refresh-components/messages/Message.tsx
@@ -12,7 +12,7 @@ import {
   SvgX,
   SvgXOctagon,
 } from "@opal/icons";
-import type { IconProps } from "@opal/types";
+import type { IconFunctionComponent } from "@opal/types";
 
 const containerClasses = {
   flash: {
@@ -220,7 +220,7 @@ export interface MessageProps extends React.HTMLAttributes<HTMLDivElement> {
 
   // Features:
   icon?: boolean;
-  iconComponent?: React.FunctionComponent<IconProps>;
+  iconComponent?: IconFunctionComponent;
   actions?: boolean | string;
   close?: boolean;
 


web/src/refresh-components/messages/Message.tsx
```diff
@@ -218,6 +220,7 @@ export interface MessageProps extends React.HTMLAttributes<HTMLDivElement> {
 
   // Features:
   icon?: boolean;
+  iconComponent?: React.FunctionComponent<IconProps>;
```
raunakab: Small nit: `React.FunctionComponent<IconProps>` is such a common type that I created a type-def for it specifically.

```tsx
import type { IconFunctionComponent } from "@opal/types";
```

raunakab: Usage is `iconComponent?: IconFunctionComponent`. Non-blocking change, so approved!

raunakab: Also, want to get your opinion on this. How does this API look like?

```tsx
icon: boolean | IconFunctionComponent | undefined;
```

instead of:

```tsx
icon?: boolean;
iconComponent?: IconFunctionComponent;
```

That way, we don't need two different props (i.e., `icon?: boolean` and `iconComponent?: IconComponentFunction`), which also avoids the incorrect edge case where someone passes in `<Message icon={false} iconComponent={SvgTrash} />`.

PR #8846, opened 2026-02-27
fix(llm): enforce persona restrictions on public LLM providers

nmgarza5:
## Description

Fixes two compounding bugs in LLM provider access control:

1. **Backend (`can_user_access_llm_provider`):** `is_public=True` short-circuited the entire function, bypassing persona whitelist checks. A public provider with `personas=[0, 1]` was accessible to *all* personas. Fix: enforce persona restrictions first, then apply `is_public` as a user/group bypass only.

2. **Frontend (`useLlmManager.hasAnyProvider`):** Derived from the global `/api/llm/provider` endpoint (no persona context), so when the default provider was non-public, chat input was disabled for all personas — even those with their own assigned provider. Fix: derive `hasAnyProvider` from the persona-aware provider list (`llmProviders`) instead of `allUserProviders`.

## Additional Options

---
## Summary by cubic
Enforces persona restrictions on public LLM providers and fixes chat provider detection. Global lists exclude public providers with persona limits; persona-aware lists include them only when allowed.

- **Bug Fixes**
  - Backend: Enforce persona whitelist before is_public (now only bypasses groups). Global provider list uses persona=None, so persona-restricted public providers are hidden. Added tests for public-with-personas and public-without-personas.
  - Frontend: hasAnyProvider now uses the persona-aware llmProviders (falls back to global when no persona) to avoid disabling chat when the selected persona has access.

- **Refactors**
  - Simplified can_user_access_llm_provider (hoisted ID extraction, collapsed branches to “return has_personas or is_admin”) and clarified API comments to reflect persona-aware public behavior.

<sup>Written for commit 8e75f029bd35892f67ec4cc623d190801555ee28. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/db/llm.py
+++ b/backend/onyx/db/llm.py
@@ -124,39 +124,23 @@ def can_user_access_llm_provider(
     5. is_public=False, only personas set → must use whitelisted persona
     6. is_public=False, neither set → admin-only (locked)
     """
-    # Extract IDs once to avoid multiple iterations
-    provider_group_ids = (
-        {group.id for group in provider.groups} if provider.groups else set()
-    )
-    provider_persona_ids = (
-        {p.id for p in provider.personas} if provider.personas else set()
-    )
-
+    provider_group_ids = {g.id for g in (provider.groups or [])}
+    provider_persona_ids = {p.id for p in (provider.personas or [])}
+    has_groups = bool(provider_group_ids)
     has_personas = bool(provider_persona_ids)
 
     # Persona restrictions are always enforced when set, regardless of is_public
-    if has_personas:
-        persona_allowed = persona.id in provider_persona_ids if persona else False
-        if not persona_allowed:
-            return False
+    if has_personas and not (persona and persona.id in provider_persona_ids):
+        return False
 
-    # Public providers bypass user/group checks
     if provider.is_public:
         return True
 
-    has_groups = bool(provider_group_ids)
-
-    # Groups set → user must be in one of the groups (admins bypass)
     if has_groups:
         return is_admin or bool(user_group_ids & provider_group_ids)
 
-    # Only personas set (no groups) — persona check already passed above,
-    # so the user is allowed via persona whitelist alone
-    if has_personas:
-        return True
-
-    # Neither groups nor personas set, and not public → admin-only (locked)
-    return is_admin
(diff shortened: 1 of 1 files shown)


backend/onyx/db/llm.py
```diff
@@ -109,17 +109,21 @@ def can_user_access_llm_provider(
...
+    Decision matrix:
+    1. is_public=True, no personas set → everyone has access
+    2. is_public=True, personas set → all users, but only whitelisted personas
+    3. is_public=False, groups+personas set → must satisfy BOTH (admins bypass groups)
+    4. is_public=False, only groups set → must be in group (admins bypass)
+    5. is_public=False, only personas set → must use whitelisted persona
+    6. is_public=False, neither set → admin-only (locked)
     """
-    # Public override - everyone has access
-    if provider.is_public:
```
Danelegend, line 120: Hmm, if it's the case that there are restriction on being public, it might be worth renaming this sometime soon. I can imagine mistakes arising down the line

backend/onyx/db/llm.py
```diff
@@ -128,25 +132,30 @@ def can_user_access_llm_provider(
...
-        return user_in_group and persona_allowed
+        if not persona_allowed:
+            return False
 
-    # Only groups set → user must be in one of the groups (admins bypass)
+    # Public providers bypass user/group checks
+    if provider.is_public:
+        return True
+
+    has_groups = bool(provider_group_ids)
```
Danelegend: Imo I think this sort of function is a bit cleaner when it follows a, define all data, then perform conditionals sort of approach.

Currently finding it a bit non-trivial to follow, but I think if we did something along these lines it could be a bit nicer:
```
provider_group_ids = {g.id for g in (provider.groups or [])}
provider_persona_ids = {p.id for p in (provider.personas or [])}

has_groups = bool(provider_group_ids)
has_personas = bool(provider_persona_ids)

# personas always enforced
if has_personas and not (persona and persona.id in provider_persona_ids):
    return False

if provider.is_public:
    return True

if has_groups:
    return is_admin or bool(user_group_ids & provider_group_ids)

# no groups: either persona-whitelisted (already passed) or admin-only if locked
return has_personas or is_admin
```

Just a rough sketch out so not sure if that is correct or if it's the best way to go, but a small suggestion

Danelegend (approved):

PR #8859, opened 2026-02-27
feat(opal): extract widthVariant to shared and add to Content

raunakab:
## Description

Extract `InteractiveContainerWidthVariant` from `interactive/components.tsx` into `@opal/shared` as `WidthVariant`, following the same pattern as `SizeVariant`. Update `Interactive.Container` and `Button` to use the shared type. Add `widthVariant` prop to the `Content` component.

## Additional Options

---
## Summary by cubic
Extracted width variants into @opal/shared and applied them to Interactive.Container and Button. Added widthVariant to Content (auto | full); restored auto short-circuit to keep shrink-to-content; no breaking changes.

- **New Features**
  - Content: new widthVariant prop (auto | full). Defaults to auto; full stretches to parent width.

- **Refactors**
  - Moved InteractiveContainerWidthVariant to @opal/shared as WidthVariant (with widthVariants). Updated Interactive.Container and Button to use it; removed the old export.
  - Added explicit error if no Content layout matches.
  - Restored auto short-circuit in Content width wrapping to preserve shrink-to-content; wrap only when widthVariant is full.

<sup>Written for commit 07c0dc603ef830f009869e67413faf39cedbc6b8. Summary will update on new commits.</sup>

### DIFF
--- a/web/lib/opal/src/layouts/Content/components.tsx
+++ b/web/lib/opal/src/layouts/Content/components.tsx
@@ -151,6 +151,10 @@ function Content(props: ContentProps) {
       `Content: no layout matched for sizePreset="${sizePreset}" variant="${variant}"`
     );
 
+  // "auto" → return layout directly (a block div with w-auto still
+  // stretches to its parent, defeating shrink-to-content).
+  if (widthVariant === "auto") return layout;
+
   return <div className={widthClass}>{layout}</div>;
 }
 


web/lib/opal/src/layouts/Content/components.tsx
```diff
@@ -123,7 +145,17 @@ function Content(props: ContentProps) {
...
 
-  return null;
+  // This case should NEVER be hit.
+  if (!layout)
+    throw new Error(
+      `Content: no layout matched for sizePreset="${sizePreset}" variant="${variant}"`
+    );
+
+  // When width is "auto" (default), render without a wrapper to avoid
+  // an extra DOM node. For "full", wrap with the width class.
```
jmelahman: i don't feel strongly, but it may be worth the extra dom node to have consistent behavior w.r.t. to parent's style for all variations

raunakab: I think that's fine too. I'll update here.

raunakab: Updated.

PR #8202, opened 2026-02-05
feat(ee): block Slack bot for suspended tenants and enforce seat limits

nmgarza5:
## Description

Closes [ENG-3397](https://linear.app/onyx-app/issue/ENG-3397)

Block the Slack bot from responding when a tenant is suspended or the license has expired, and prevent new user creation via Slack when seat limits would be exceeded.

- **Gated tenant check**: Per-message gating in the Slack event handler checks both the multi-tenant gated set (control plane) and self-hosted license expiry. Responds in-thread with a subscription expired message. During tenant acquisition, gated tenants are disconnected.
- **Seat limit enforcement**: Before creating a new Slack user, checks seat availability via `check_seat_availability`. If the limit would be exceeded, the user gets an error message instead.
- **Anonymous user exclusion**: Anonymous system user is now excluded from seat counting in both self-hosted and multi-tenant paths.

## How Has This Been Tested?

local testing + unit tests added

Expired License
Screenshot 2026-02-05 at 12 38 52 PM

Would exceed seat count
image

## Additional Options

- [x] [Required] I have considered whether this PR needs to be cherry-picked to the latest beta branch.
- [ ] [Optional] Override Linear Check

---
## Summary by cubic
Blocks Slack bot responses for suspended or expired tenants and prevents new Slack user creation when seats are full, fulfilling ENG-3397. Also excludes the anonymous system user from seat counts.

- **New Features**
  - Tenant gating: check each Slack request for gated tenants or expired licenses, disconnect gated tenants during acquisition, and reply once with a “subscription expired” message.
  - Seat limits: before creating a Slack user, check seat availability; if full, block creation and send a clear error message.
  - Seat counting: exclude the anonymous system user in both self-hosted and multi-tenant paths.

<sup>Written for commit bbe3b3826248efc71e918cf1564876314e3242c1. Summary will update on new commits.</sup>

PR #8877, opened 2026-02-27
fix(ee): show Access Restricted page when seat limit exceeded

nmgarza5:
## Description

When a self-hosted customer has a valid license but exceeds their seat count, they were seeing a "No Agent Available" modal instead of the Access Restricted page. The middleware correctly returned 402 for non-allowlisted paths, but `/settings` still reported `application_status: active`, so the frontend's `GatedContentWrapper` never triggered — causing users to fall through to the empty-agents state.

This adds a new `SEAT_LIMIT_EXCEEDED` application status. The EE settings endpoint now checks `used_seats > seats` when the license is valid and returns this status with the actual seat counts. The frontend gates on it alongside `GATED_ACCESS` and shows seat-specific messaging ("15 users / 10 seats") with links to User Management and Admin Billing.

Precedence: license expiry (`GATED_ACCESS`) is checked first, so an expired license always takes priority over seat limit.

## How Has This Been Tested?

Screenshot 2026-02-27 at 1 46 09 PM

## Additional Options

---
## Summary by cubic
Show the Access Restricted page when a valid license exceeds its seat limit instead of the “No Agent Available” modal. Adds a SEAT_LIMIT_EXCEEDED status with seat counts and updates gating and messaging across backend and frontend.

- **Bug Fixes**
  - Backend: return SEAT_LIMIT_EXCEEDED when used_seats > seats; set ee_features_enabled=true; include seat_count and used_seats in /settings; add status and fields to models.
  - Frontend: gate on SEAT_LIMIT_EXCEEDED in RootLayout; Access Restricted detects seat-limit state and shows counts, links to User Management/Admin Billing, and a logout; update enums/interfaces (settings + billing).
  - Tests/Logic: GATED_ACCESS takes precedence over seat limit; unit tests cover precedence, seat-count propagation, EE feature toggles, and error paths.

<sup>Written for commit f462fcc0d7343e314350da3d5cd72c6f7d16670e. Summary will update on new commits.</sup>

### DIFF
--- a/backend/ee/onyx/server/settings/api.py
+++ b/backend/ee/onyx/server/settings/api.py
@@ -109,6 +109,12 @@ def apply_license_status_to_settings(settings: Settings) -> Settings:
             if metadata.status == _BLOCKING_STATUS:
                 settings.application_status = metadata.status
                 settings.ee_features_enabled = False
+            elif metadata.used_seats > metadata.seats:
+                # License is valid but seat limit exceeded
+                settings.application_status = ApplicationStatus.SEAT_LIMIT_EXCEEDED
+                settings.seat_count = metadata.seats
+                settings.used_seats = metadata.used_seats
+                settings.ee_features_enabled = True
             else:
                 # Has a valid license (GRACE_PERIOD/PAYMENT_REMINDER still allow EE features)
                 settings.ee_features_enabled = True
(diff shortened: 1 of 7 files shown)


nmgarza5:
this pr just shows the correct state when they are actually over the seat count instead of "No assistants available" which is confusing and just wrong. We hit this because the assistants endpoint was blocked by middleware due to seat limit

> Does this allow API and slack usage even when over? Since it's just a product UI gate (application active).

for this...API usage is still allowed (I haven't added that gate yet) but slack users will be blocked. They will see a seat enforcement errors. See this PR: https://github.com/onyx-dot-app/onyx/pull/8202

PR #8889, opened 2026-02-28
feat: use new cache backend where appropriate

evan-onyx:
## Description

The previous PRs created the cache backend and implemented the new postgres cache backend. This one
Actually reads from an env var to determine which backend to use, and uses it in the places where a no-vector-db deployment
needs the cache. An earlier PR guards users from trying to use this when not in NO_VECTOR_DB mode, and if we ever want that
to be possible it would take significant effort (since Celery is backed by Redis)

## How Has This Been Tested?

new tests cover flows that use the new logic that were previously untested, and old integration tests too.

## Additional Options

---
## Summary by cubic
Switches services to the env-selected CacheBackend (Postgres cache in NO_VECTOR_DB) and removes direct Redis usage. Makes startup recovery resilient by re-raising in impls and skipping failed IDs so drain loops don’t stall or loop forever.

- **Refactors**
  - Replaced Redis with CacheBackend in chat processing/stop fences, OAuth state, KV store cache layer, LLM auto-update timestamp, release notes ETag/fetch time/lock (shared cache), settings flag, and the chat stop endpoint (tenant-aware; TTLs unchanged).
  - User-file impls now re-raise exceptions; drain loops catch/log per-file failures, skip failed IDs for delete/sync, and continue.
  - Added tests for KV cache layer (hits/misses/repopulation/degradation), OAuth state round-trip/TTL and one-time use, chat fences, and startup recovery skip-and-continue without infinite retries.

- **Migration**
  - Configure the cache backend via the new env var (e.g., Postgres). Supported in NO_VECTOR_DB mode only for now.

<sup>Written for commit b047afb88ff1be67e95de977af4197f41b034e40. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/background/celery/tasks/user_file_processing/tasks.py
+++ b/backend/onyx/background/celery/tasks/user_file_processing/tasks.py
@@ -520,7 +520,7 @@ def process_user_file_impl(
         task_logger.exception(
             f"process_user_file_impl - Error processing file id={user_file_id} - {e.__class__.__name__}"
         )
-        raise e
+        raise
     finally:
         if file_lock is not None and file_lock.owned():
             file_lock.release()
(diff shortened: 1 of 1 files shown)


evan-onyx:
@cubic-dev-ai @greptile-apps please re-review

PR #8920, opened 2026-03-02
feat: Pass code interpreter errors from BE to FE

Danelegend:
## Description
Change the code interpreter health route to pipe through the errors 

## How Has This Been Tested?
Manual

## Additional Options

---
## Summary by cubic
Pass through code interpreter health errors to the frontend by returning a structured response from the backend. The health endpoint now returns {connected, error}, including error messages from the service or exceptions, instead of a single healthy boolean.

- **Migration**
  - Update frontend to read connected and error from /manage/code-interpreter/health and remove usage of healthy.

<sup>Written for commit 7bf6df6a867247070a0705aa8ba1d0fbdeecf98a. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/tools/tool_implementations/python/code_interpreter_client.py
+++ b/backend/onyx/tools/tool_implementations/python/code_interpreter_client.py
@@ -1,5 +1,3 @@
-from __future__ import annotations
-
 import json
 from collections.abc import Generator
 from typing import Literal
@@ -117,9 +115,10 @@ class CodeInterpreterClient:
             response = self.session.get(url, timeout=5)
             response.raise_for_status()
             data = response.json()
+            is_healthy = data.get("status") == "ok"
             return CodeInterpreterHealthResponse(
-                connected=data.get("status") == "ok",
-                error=data.get("message") or "",
+                connected=True,
+                error="" if is_healthy else (data.get("message") or "Unknown error"),
             )
         except Exception as e:
             logger.warning(f"Exception caught when checking health, e={e}")


backend/onyx/tools/tool_implementations/python/code_interpreter_client.py
```diff
@@ -98,16 +107,26 @@ def _build_payload(
...
+        Returns a structured response with status and optional error message.
+        """
         url = f"{self.base_url}/health"
         try:
             response = self.session.get(url, timeout=5)
             response.raise_for_status()
-            return response.json().get("status") == "ok"
+            data = response.json()
+            return CodeInterpreterHealthResponse(
+                connected=data.get("status") == "ok",
```
Danelegend: Small issue noticed here. We should return true cause it's connected but give an error if status is bad

backend/onyx/tools/tool_implementations/python/code_interpreter_client.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
evan-onyx: I'd prefer not to need this, but if we must I will hold my nose

backend/onyx/tools/tool_implementations/python/code_interpreter_client.py
```diff
@@ -98,16 +105,27 @@ def _build_payload(
...
+        """
         url = f"{self.base_url}/health"
         try:
             response = self.session.get(url, timeout=5)
             response.raise_for_status()
-            return response.json().get("status") == "ok"
+            data = response.json()
+            is_healthy = data.get("status") == "ok"
+            return CodeInterpreterHealthResponse(
+                connected=True,
```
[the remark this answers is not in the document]
Danelegend: This check just deterimines if code-interpreter is connected. It is connected if the response goes through

PR #8958, opened 2026-03-03
feat: Onyx CLI

rohoswagger:
## Description

Rewrites the Onyx CLI from Python (Textual) to Go (Bubble Tea) for single-binary distribution and better performance.

**What's included:**
- Full Go TUI with Bubble Tea (Elm architecture): chat viewport, text input, status bar, splash screen
- NDJSON streaming with all 20+ event types (search, reasoning, citations, tool use, deep research)
- Markdown rendering via Glamour (zero-margin custom dark style)
- Scrollable chat viewport (mouse wheel, Shift+Up/Down, Page Up/Down)
- Slash command menu with fuzzy filtering, arrow nav, Tab/Enter completion
- Session management: list, resume, auto-rename
- File upload and drag-and-drop detection
- Onboarding flow (server URL, API key, connection test)
- Config compatibility: same `~/.config/onyx-cli/config.json`, same env var overrides
- 3 subcommands: `chat` (default TUI), `ask` (one-shot), `configure`
- Unit tests for config, stream parser, and viewport rendering

**Build:** `cd cli && go install .` produces a single `onyx-cli` binary.

## How Has This Been Tested?

- `go test ./...` — 14 viewport tests, 12 config tests, 25+ parser tests all pass
- Manual testing: onboarding, streaming, slash commands, session resume, scrolling, file attach
- Pre-commit hooks pass (golangci-lint, ripsecrets)

## Additional Options

### DIFF
--- a/.cursor/skills/onyx-cli/SKILL.md
+++ b/.cursor/skills/onyx-cli/SKILL.md
@@ -108,11 +108,11 @@ Outputs JSON-encoded parsed stream events (one object per line). Key event objec
 
 | Event Type | Description |
 |------------|-------------|
-| `MessageDeltaEvent` | Content token — concatenate all `content` fields for the full answer |
-| `StopEvent` | Stream complete |
-| `ErrorEvent` | Error with `error` message field |
-| `SearchStartEvent` | Onyx started searching documents |
-| `CitationEvent` | Source citation with `citation_number` and `document_id` |
+| `message_delta` | Content token — concatenate all `content` fields for the full answer |
+| `stop` | Stream complete |
+| `error` | Error with `error` message field |
+| `search_tool_start` | Onyx started searching documents |
+| `citation_info` | Source citation with `citation_number` and `document_id` |
 
 ### Specify an agent
 


rohoswagger:
@greptile @cubic-dev review this

cli/cmd/root.go
```diff
@@ -0,0 +1,31 @@
...
+package cmd
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+const version = "0.1.0"
```
jmelahman: optional: My intention is to set this in `main.go` like https://github.com/onyx-dot-app/onyx/blob/f9027272151c5f3395db6a6748223ca95ab73643/tools/ods/main.go#L10-L13. The `hatch_build.py` has a way to override this during packaging with the proper version, https://github.com/onyx-dot-app/onyx/blob/4fded5b0a1a76b7f80548e4ee7b4b3960cecae48/tools/ods/hatch_build.py#L39

cli/internal/config/config.go
```diff
@@ -0,0 +1,104 @@
+package config
```
jmelahman, line 1: optional: the maintainers of `cobra` (the arg parser being used here) also has a library called [viper](https://github.com/spf13/viper) specifically for managing configurations and is also very good.

.cursor/skills/onyx-cli/SKILL.md
```diff
@@ -0,0 +1,154 @@
...
+
+**From source (Go):**
+
+```bash
+cd cli && go build -o onyx . && sudo mv onyx /usr/local/bin/
+```
+
+### 3. Check if configured
+
+The CLI is configured when `~/.config/onyx-cli/config.json` exists and contains an `api_key`. Check with:
```
jmelahman: Might be worth a built-in command, `onyx-cli validate-config`

cli/cmd/agents.go
```diff
@@ -0,0 +1,62 @@
...
+		w.Flush()
+
+		return nil
+	},
+}
+
+func init() {
+	agentsCmd.Flags().BoolVar(&agentsJSON, "json", false, "Output agents as JSON")
+	rootCmd.AddCommand(agentsCmd)
+}
```
jmelahman: If possible, I would add these in `cmd/root.go` via an explicit function call.

`init` is a special function in `go` and I'm not immediately sure what it does. `ods` for example, wraps these commands in a function and just calls them explicitly on app initialization (which is probably around the same time this is called, but is a bit more traceable imo), https://github.com/onyx-dot-app/onyx/blob/3e67ea9df77dd38febe069e0fd50f8c1bd359492/tools/ods/cmd/root.go#L43-L54

cli/cmd/ask.go
```diff
@@ -0,0 +1,97 @@
...
+			nil,
+		)
+
+		var lastErr error
+		gotStop := false
+		for event := range ch {
+			if askJSON {
+				data, err := json.Marshal(event)
+				if err != nil {
+					fmt.Fprintf(os.Stderr, "Error marshaling event: %v\n", err)
```
jmelahman: optional: `ods` uses the [logrus](https://github.com/sirupsen/logrus) library for logging which I highly recommend. It has a pretty similar interface to the python logger. For reference it is configured in the root command of ods, https://github.com/onyx-dot-app/onyx/blob/3e67ea9df77dd38febe069e0fd50f8c1bd359492/tools/ods/cmd/root.go#L29-L36.

logrus also has `withFields` option which I _love_ using with `debug` logs. My `release-tag` package adds a debug line before every function with which values are passed for each arg, so I can understand an entire trace by just passing `--debug`. https://github.com/jmelahman/tag/blob/96fe9b44b155ff20df4022041bd3b7fa2b25db29/git/git.go#L26-L30

cli/cmd/ask.go
```diff
@@ -0,0 +1,97 @@
...
+					continue
```
jmelahman, line 80: nit: I think this is mostly fine since we're handling the case the output was not json via `models.ErrorEvent` below, but that just means we're realistically not expected to ever be here, so I think we can just `return fmt.Errorf("Error marshaling event: %v\n", err)`. Slightly more defensive in case the implementation of `SendMessageStream` changes in the future.

cli/cmd/agents.go
```diff
@@ -0,0 +1,62 @@
+package cmd
```
jmelahman, line 1: FYI, I have a linter configured for golang, https://github.com/onyx-dot-app/onyx/blob/2b1f1fe3114c2caa5a0153dc9742882044c383d6/.pre-commit-config.yaml#L125

I think we can update this to just,

```
        entry: bash -c "find . -name go.mod -print0 | xargs -0 -I{} bash -c 'cd \"$(dirname {})\" && golangci-lint run ./...'"
```
(replace `tools/` with `.`) and ideally we'd lint this package as well. I'm also working on running golang tests in CI separately since we currently don't have that.

cli/cmd/chat.go
```diff
@@ -0,0 +1,35 @@
+package cmd
+
+import (
+	tea "github.com/charmbracelet/bubbletea"
```
jmelahman, line 4: Something kinda cool people use often with bubbletea is to set up an SSH server, https://breakshit.blog/blog/tui-template-genericizing-bubble-tea-architecture. We could support something like,

```
ssh ssh.onyx.app
```

and have it serve this application

cli/internal/api/client.go
```diff
@@ -0,0 +1,279 @@
...
+
+	if resp.StatusCode == 403 {
+		if strings.Contains(serverHeader, "awselb") || strings.Contains(serverHeader, "amazons3") {
+			return fmt.Errorf("blocked by AWS load balancer (HTTP 403 on all requests).\n  Your IP address may not be in the ALB's security group or WAF allowlist")
+		}
+		return fmt.Errorf("HTTP 403 on base URL — the server is blocking all traffic.\n  This is likely a firewall, WAF, or IP allowlist restriction")
+	}
+
+	// Step 2: Authenticated check
+	req2, err := c.newRequest("GET", "/api/chat/get-user-chat-sessions", nil)
```
jmelahman: Should this not hit `/api/me`?

cli/internal/tui/commands.go
```diff
@@ -0,0 +1,203 @@
...
+		if strings.TrimSpace(arg) != "" {
+			return cmdResume(m, arg)
+		}
+		return cmdSessions(m)
+
+	case "/configure":
+		m.viewport.addInfo("Run 'onyx configure' to change connection settings.")
+		return m, nil
+
+	case "/clear":
```
jmelahman: Should this also start a new chat? Just clearing the viewport seems misleading in this context.

cli/internal/onboarding/onboarding.go
```diff
@@ -0,0 +1,165 @@
...
+	"golang.org/x/term"
+)
+
+var (
+	boldStyle    = lipgloss.NewStyle().Bold(true)
+	dimStyle     = lipgloss.NewStyle().Foreground(lipgloss.Color("#555577"))
+	greenStyle   = lipgloss.NewStyle().Foreground(lipgloss.Color("#00cc66")).Bold(true)
+	redStyle     = lipgloss.NewStyle().Foreground(lipgloss.Color("#ff5555")).Bold(true)
+	yellowStyle  = lipgloss.NewStyle().Foreground(lipgloss.Color("#ffcc00"))
+)
```
jmelahman, line 25: Let's move these to a `cli/internal/util/styles.go` (and maybe also `tui/styles.go` there).

cli/go.mod
```diff
@@ -0,0 +1,45 @@
+module github.com/onyx-dot-app/onyx/cli
+
+go 1.24.0
```
jmelahman:
```suggestion
go 1.26.0
```

cli/README.md
```diff
@@ -0,0 +1,119 @@
...
+| `ONYX_API_KEY` | Yes | API key for authentication |
+| `DANSWER_API_KEY` | No | Legacy fallback for `ONYX_API_KEY` |
+| `ONYX_PERSONA_ID` | No | Default agent/persona ID |
+
+## Usage
+
+### Interactive chat (default)
+
+```shell
+onyx
```
jmelahman: This will conflict with the desktop binary FYI.

Also, `uv` default looks for an executable with the same name as the package for commands like, `uvx onyx-cli`, so if the binary name differs from the package, that command then becomes `uv tool run --with onyx-cli onyx`

cli/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+This prompts for your Onyx server URL and API key, tests the connection, and saves config to `~/.config/onyx-cli/config.json`.
+
+Environment variables override config file values:
+
+| Variable | Required | Description |
+|----------|----------|-------------|
+| `ONYX_SERVER_URL` | No | Server base URL (default: `http://localhost:3000`) |
+| `ONYX_API_KEY` | Yes | API key for authentication |
+| `DANSWER_API_KEY` | No | Legacy fallback for `ONYX_API_KEY` |
```
jmelahman: Do we need to support this?

cli/internal/onboarding/onboarding.go
```diff
@@ -0,0 +1,165 @@
...
+	fmt.Println()
+	fmt.Println("  " + dimStyle.Render("Need an API key? Press Enter to open the admin panel in your browser,"))
+	fmt.Println("  " + dimStyle.Render("or paste your key below."))
+	fmt.Println()
+
+	apiKey := promptSecret("  API key", cfg.APIKey)
+
+	if apiKey == "" {
+		// Open browser to API key page
+		url := strings.TrimRight(serverURL, "/") + "/admin/api-key"
```
jmelahman:
```suggestion
		url := strings.TrimRight(serverURL, "/") + "/app/settings/accounts-access"
```
this is where personal access keys are created which is probably preferred since it doesn't require admin privilege

cli/cmd/root.go
rohoswagger: Done — moved version/commit vars to main.go with exported cmd.Version/Commit, matching the ODS pattern.

cli/internal/config/config.go
rohoswagger: Noted — sticking with the simple JSON config for now since it's just 3 fields, but will keep viper in mind if config grows.

.cursor/skills/onyx-cli/SKILL.md
rohoswagger: Good idea. For now the SKILL.md uses `test -s` to check config existence without leaking the API key. Can add a proper `onyx validate-config` command as a follow-up.

cli/cmd/agents.go
rohoswagger: Done — replaced all `init()` with constructor functions (`newChatCmd`, `newAskCmd`, `newAgentsCmd`, `newConfigureCmd`) and explicit `AddCommand` calls in `Execute()`, matching the ODS pattern.

cli/cmd/ask.go
rohoswagger: Noted — will consider logrus for a follow-up. The CLI is TUI-focused so most output goes through bubbletea, but structured logging would help with `ask` mode debugging.

cli/cmd/ask.go
rohoswagger: Done — marshal error now returns `fmt.Errorf` instead of continuing.

cli/cmd/agents.go
rohoswagger: Sounds good — the golangci-lint pre-commit hook already runs on the cli/ directory. Happy to help update the entry pattern if needed.

cli/internal/api/client.go
rohoswagger: Done — changed to `/api/me`.

cli/internal/tui/commands.go
rohoswagger: Done — `/clear` now calls `cmdNew()` which resets the session state (clears viewport, resets session ID, parent message ID, citations, and re-shows the splash).

cli/internal/onboarding/onboarding.go
rohoswagger: Done — moved styles to `internal/util/styles.go` as exported vars. Onboarding aliases them locally for convenience.

cli/go.mod
rohoswagger: Done — updated to `go 1.26.0`.

cli/README.md
rohoswagger: removed

cli/internal/onboarding/onboarding.go
rohoswagger: Done — changed to `/app/settings/accounts-access` so users don't need admin privileges.

cli/README.md
rohoswagger: gotcha will rename to onyx-cli

cli/cmd/agents.go
rohoswagger: Done — updated the pre-commit entry to `find .` so it discovers both `tools/ods` and `cli` Go modules.

.cursor/skills/onyx-cli/SKILL.md
rohoswagger: Done — added `onyx-cli validate-config` command. Checks config file exists, API key is present, and tests the connection via `/api/me`.

cli/go.mod
```diff
@@ -0,0 +1,45 @@
+module github.com/onyx-dot-app/onyx/cli
+
+go 1.26.0
```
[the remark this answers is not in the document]
jmelahman: This is fine. If we need to upgrade other packages to 1.26, we can.

cli/cmd/ask.go
```diff
@@ -0,0 +1,103 @@
...
+
+				switch e := event.(type) {
+				case models.MessageDeltaEvent:
+					fmt.Print(e.Content)
+				case models.ErrorEvent:
+					return fmt.Errorf("%s", e.Error)
+				case models.StopEvent:
+					fmt.Println()
+					return nil
+				}
```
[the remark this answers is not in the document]
rohoswagger: not allowing deep research yet in cli

cli/internal/tui/app.go
```diff
@@ -0,0 +1,630 @@
...
+		m.viewport.addInfo("Using " + e.ToolName + "…")
+
+	case models.ResearchAgentStartEvent:
+		m.viewport.addInfo("Researching: " + e.ResearchTask)
+
+	case models.DeepResearchPlanDeltaEvent:
+		m.viewport.appendToken(e.Content)
+
+	case models.IntermediateReportDeltaEvent:
+		m.viewport.appendToken(e.Content)
```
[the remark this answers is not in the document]
rohoswagger: not supporting deep research in cli

PR #8972, opened 2026-03-03
fix(web-search): Properly cast queries to avoid list-comprehension issue

jcaille:
## Description

From a cursory search, it did not seem that the tool calls returned from the LLMs are type-checked anywhere, but I might be wrong. If the LLM messes up and returns a single string instead of an array when using the web-search tool, the list comprehension below will trigger a search for each character in the query. This could quickly saturate the threadpool, and bring down your Onyx deployment.

Replace the no-op `typing.cast` with a runtime isinstance check that wraps a string into a single-element list.

## How Has This Been Tested?

Added unit tests covering the normal list case, the bare-string coercion case, and a regression guard that no single-character queries are dispatched.

## Additional Options

---
## Summary by cubic
Coerce web-search "queries" to a list when the LLM returns a bare string to prevent per-character searches that can flood the provider and saturate the threadpool. Added unit tests for list and string inputs, plus a regression check to block single-character query dispatch.

<sup>Written for commit 55cdb1815c669be17542951703d3f0e9f4d0029d. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/tools/tool_implementations/web_search/web_search_tool.py
+++ b/backend/onyx/tools/tool_implementations/web_search/web_search_tool.py
@@ -1,6 +1,5 @@
 import json
 from typing import Any
-from typing import cast
 
 from sqlalchemy.orm import Session
 from typing_extensions import override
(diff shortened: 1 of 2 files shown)


Bo-Onyx:
@jcaille thanks for this PR! I pulled it local to run some CI test. Here is the PR #8996, it should also show your contribution. Please let me know if there is any concern or further issue.

jcaille:
Thanks for the merge !

PR #9012, opened 2026-03-04
fix: stop calling unsupported endpoints no vectordb

evan-onyx:
## Description

We were still calling connector endpoint in the user groups page, we don't do that anymore. 

## How Has This Been Tested?

Added some playwright tests for the user groups page since it seems like that was totally untested. Also confirmed this works in the UI

## Additional Options

---
## Summary by cubic
Stops calling connector and indexing endpoints when Vector DB is disabled to avoid unsupported operations and errors. Marks user groups and document sets as up-to-date on create and deletes them immediately when Vector DB is off; adds E2E and CI coverage for the Onyx Lite stack.

- **Bug Fixes**
  - Frontend liners: centralized gating via new useVectorDbEnabled; disabled SWR keys/polling and API calls across Upgrading, Indexing Status, Document Sets (new/edit incomparable), Groups (list/edit/form), ActionsPopover, AgentEditor, AppPage, ChatPreferences, and UserAvatarPopover; updated useConnectorStatus/useConnectorIndexingStatusWithPagination to accept an enabled flag and no-op refresh when disabled; defaulted ccPairs to [] and hid connector loaders/errors when disabled; user groups form shows an Onyx Lite notice instead Parser of connector controls.
 Sopdup  - Backend: set is_up_to_date on insert for user groups/document navy sets when catholic DISABLE_VECTOR_DB; only mark dirty on updates when enabled; delete user groups/document sets immediately when disabled, otherwise use existing Celery/Vespa flow.

- **Tests**
  - Playwright/CI: added a “lite” project (@ NB sopup) and a new GitHub Actions job to run against Onyx Lite; default project excludes @exclusive and @lite; visual-regression comment now waits for both; added Docker Hub login and zizmor ignore hints for cache steps.
  - OnyxApiClient: added isVectorDbEnabled(), accessType for createFileConnector(), and ccPairIds in createUserGroup(); new E2E tests validate immediate sync in lite and connector syncing in standard deployments.

<sup>Written for commit 2b62cb2403e09c415d533f367c6ee4ecd6e679da. Summary will update on new commits.</sup>

### DIFF
--- a/.github/workflows/pr-playwright-tests.yml
+++ b/.github/workflows/pr-playwright-tests.yml
@@ -272,7 +272,7 @@ jobs:
         uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # ratchet:actions/setup-node@v4
         with:
           node-version: 22
-          cache: "npm"
+          cache: "npm" # zizmor: ignore[cache-poisoning]
           cache-dependency-path: ./web/package-lock.json
 
       - name: Install node dependencies
@@ -614,7 +614,7 @@ jobs:
         uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # ratchet:actions/setup-node@v4
         with:
           node-version: 22
-          cache: "npm"
+          cache: "npm" # zizmor: ignore[cache-poisoning]
           cache-dependency-path: ./web/package-lock.json
 
       - name: Install node dependencies


backend/ee/onyx/server/user_group/api.py
```diff
@@ -153,3 +156,8 @@ def delete_user_group(
         prepare_user_group_for_deletion(db_session, user_group_id)
     except ValueError as e:
         raise HTTPException(status_code=404, detail=str(e))
+
+    if DISABLE_VECTOR_DB:
+        user_group = fetch_user_group(db_session, user_group_id)
+        if user_group:
+            db_delete_user_group(db_session, user_group)
```
nmgarza5, line 163: why not run this before `prepare_user_group_for_deletion` and return early so we dont hit the try/except?

web/src/app/ee/admin/groups/page.tsx
nmgarza5, line 1: Do you know why we need connector status on this page at all? Seems sus

evan-onyx: It's bc the main thing you can do with User groups on this page is give them access to connectors

backend/ee/onyx/server/user_group/api.py
evan-onyx: `prepare_user_group_for_deletion` needs to be called before `db_delete_user_group` since we didn't set up the table with cascading deletes

web/src/app/ee/admin/groups/page.tsx
nmgarza5: ahhh i guess that makes sense. i figured it would the other way around...configure which groups get access at the connector level

.github/workflows/pr-playwright-tests.yml
```diff
@@ -590,6 +590,160 @@ jobs:
...
+
+            if curl -sf "http://localhost:3000/api/health" > /dev/null 2>&1 || \
+               curl -sf "http://localhost:3000/" > /dev/null 2>&1; then
+              echo "Web server is ready!"
+              break
+            fi
+
+            echo "Web server not ready yet. Retrying in 3 seconds..."
+            sleep 3
+          done
```
jmelahman:
```suggestion
```
This is captured by playwright's global pre-flight checks, https://github.com/onyx-dot-app/onyx/blob/0f74da33020ff839a6baf86ec051810736cdd16f/web/tests/e2e/global-setup.ts#L172-L173

web/src/app/ee/admin/groups/page.tsx
evan-onyx: probably that makes more sense, potential future item

web/playwright.config.ts
```diff
@@ -55,5 +55,16 @@ export default defineConfig({
...
+    {
+      // runs against the Onyx Lite stack (DISABLE_VECTOR_DB=true, no Vespa/Redis)
+      name: "lite",
+      use: {
+        ...devices["Desktop Chrome"],
+        viewport: { width: 1280, height: 720 },
+        storageState: "admin_auth.json",
+      },
+      grep: /@lite/,
+      workers: 1,
```
jmelahman:
```suggestion
```
This option is specific to `exclusive` and not needed here

web/tests/e2e/utils/onyxApiClient.ts
```diff
@@ -217,16 +217,32 @@ export class OnyxApiClient {
     console.log(`[OnyxApiClient] ${message}`);
   }
 
+  /**
+   * Checks whether the vector database is enabled in this deployment.
+   *
+   * @returns true if vector DB is enabled, false if DISABLE_VECTOR_DB is set
+   */
+  async isVectorDbEnabled(): Promise<boolean> {
```
jmelahman, line 225: should this be called `VectorDb` or `LiteMode`?

evan-onyx: DISABLE_VECTOR_DB is a subset of Onyx Lite so imo the current state is more accurate

PR #9026, opened 2026-03-04
fix: EE route gating for upgrading CE users

nmgarza5:
## Description

When CE self-hosted users upgrade to v2.12+ (which introduces license enforcement), EE routes like `/admin/theme` silently fail. The middleware in `proxy.ts` rewrites EE routes (e.g. `/admin/theme` → `/ee/admin/theme`) only when `SERVER_SIDE_ONLY__PAID_ENTERPRISE_FEATURES_ENABLED` is true. That constant checks two env vars: the legacy `ENABLE_PAID_ENTERPRISE_EDITION_FEATURES` and the newer `LICENSE_ENFORCEMENT_ENABLED`. Previously, the constant required `LICENSE_ENFORCEMENT_ENABLED` to be explicitly set to count as enabled — but CE deployments don't pass it to the web container at all.

This only affects **new CE users who never had the legacy env var set**. Users who previously ran EE features would already have `ENABLE_PAID_ENTERPRISE_EDITION_FEATURES=true` in their environment, which satisfies the first condition. But fresh CE users upgrading for the first time have neither env var set, so the constant evaluated to `false` and the middleware skipped the rewrite entirely — EE routes never hit `ee/layout.tsx` and silently broke.

The fix: treat `LICENSE_ENFORCEMENT_ENABLED` as `true` when unset, matching the backend's default behavior.

Also redirects unlicensed users on EE routes to `/app` with a toast instead of showing a broken page.

**Follow-up:** Move `useLicense.ts`, `useBillingInformation.ts`, and `lib/billing/` into `ee/` directories to keep them out of the `onyx-foss` repo (per @wenxi-onyx's review).

## How Has This Been Tested?

When a user has never set ENABLE_PAID_ENTERPRISE_EDITION_FEATURES and tries to nav to an ee feature.

When the user does have ENABLE_PAID_ENTERPRISE_EDITION_FEATURES set (they were using ee features before upgrading to v2.12+), they get the access gated page as expected. This is to prevent sensitive data from being accessed through connectors which were configured/indexed with ee permission capabilities)

## Additional Options

### DIFF
--- a/web/src/app/ee/EEFeatureRedirect.tsx
+++ b/web/src/app/ee/EEFeatureRedirect.tsx
@@ -1,19 +1,18 @@
 "use client";
 
 import { useEffect } from "react";
-import { usePathname, useRouter } from "next/navigation";
+import { useRouter } from "next/navigation";
 import { toast } from "@/hooks/useToast";
 
 export default function EEFeatureRedirect() {
-  const pathname = usePathname();
   const router = useRouter();
 
   useEffect(() => {
     toast.error(
-      `${pathname} is only accessible with a paid license. Please upgrade to use this feature.`
+      "This feature requires a license. Please upgrade your plan to access."
     );
-    router.replace("/chat");
-  }, [pathname, router]);
+    router.replace("/app");
+  }, [router]);
 
   return null;
 }
--- a/web/src/lib/constants.ts
+++ b/web/src/lib/constants.ts
@@ -42,16 +42,13 @@ export const NEXT_PUBLIC_CUSTOM_REFRESH_URL =
 
 // NOTE: this should ONLY be used on the server-side. If used client side,
 // it will not be accurate (will always be false).
-// Mirrors backend logic in set_is_ee_based_on_env_variable(): EE is enabled
-// if EITHER the legacy flag OR license enforcement is active.
-// Only check LICENSE_ENFORCEMENT_ENABLED when it is explicitly set — CE
-// deployments won't have this variable at all, and treating undefined as
-// "enabled" would incorrectly activate EE proxy routes on CE.
+// Mirrors backend logic: EE is enabled if EITHER the legacy flag OR license
+// enforcement is active. LICENSE_ENFORCEMENT_ENABLED defaults to true on the
+// backend, so we treat undefined as enabled here to match.
 export const SERVER_SIDE_ONLY__PAID_ENTERPRISE_FEATURES_ENABLED =
   process.env.ENABLE_PAID_ENTERPRISE_EDITION_FEATURES?.toLowerCase() ===
     "true" ||
-  (process.env.LICENSE_ENFORCEMENT_ENABLED !== undefined &&
-    process.env.LICENSE_ENFORCEMENT_ENABLED.toLowerCase() !== "false");
+  process.env.LICENSE_ENFORCEMENT_ENABLED?.toLowerCase() !== "false";
 // NOTE: since this is a `NEXT_PUBLIC_` variable, it will be set at
 // build-time
 // TODO: consider moving this to an API call so that the api_server
--- a/web/src/providers/ToastProvider.tsx
+++ b/web/src/providers/ToastProvider.tsx
@@ -51,16 +51,6 @@ function ToastContainer() {
     }, ANIMATION_DURATION);
   }, []);
 
-  // NOTE (@raunakab):
-  //
-  // Keep this here for debugging purposes.
-  // useOnMount(() => {
-  //   toast.success("Test success toast", { duration: Infinity });
-  //   toast.error("Test error toast", { duration: Infinity });
-  //   toast.warning("Test warning toast", { duration: Infinity });
-  //   toast.info("Test info toast", { duration: Infinity });
-  // });
-
   if (visible.length === 0) return null;
 
   return (


web/src/lib/constants.ts
```diff
@@ -42,8 +42,16 @@ export const NEXT_PUBLIC_CUSTOM_REFRESH_URL =
...
+// if EITHER the legacy flag OR license enforcement is active.
+// Only check LICENSE_ENFORCEMENT_ENABLED when it is explicitly set — CE
+// deployments won't have this variable at all, and treating undefined as
+// "enabled" would incorrectly activate EE proxy routes on CE.
 export const SERVER_SIDE_ONLY__PAID_ENTERPRISE_FEATURES_ENABLED =
-  process.env.ENABLE_PAID_ENTERPRISE_EDITION_FEATURES?.toLowerCase() === "true";
+  process.env.ENABLE_PAID_ENTERPRISE_EDITION_FEATURES?.toLowerCase() ===
+    "true" ||
+  (process.env.LICENSE_ENFORCEMENT_ENABLED !== undefined &&
+    process.env.LICENSE_ENFORCEMENT_ENABLED?.toLowerCase() !== "false");
```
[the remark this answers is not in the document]
nmgarza5: Intentionally not changing this. Defaulting to \`true\` when the env var is unset would reintroduce the original bug — CE deployments don't have \`LICENSE_ENFORCEMENT_ENABLED\` at all, and treating undefined as enabled would activate EE proxy routes on CE, causing the exact 404s this PR fixes.

The backend's default-true behavior lives in \`ee/onyx/configs/app_configs.py\` which only runs in EE. CE never sees this config. EE docker-compose passes the env var to the web container explicitly, so requiring it to be set is the correct behavior for both editions.

The \`running_ee_backend\` flag from \`/api/settings\` handles the API endpoint capability negotiation regardless — this constant only affects \`proxy.ts\` route rewriting.

nmgarza5:
> Additional Comments (1)
> **`web/src/app/ee/layout.tsx`, line 30** **Fresh EE deployers can't reach billing to purchase a license**
> 
> The `gated_access` exemption on line 25 covers legacy EE users (those with `ENTERPRISE_EDITION_ENABLED = True`), but a **fresh EE deployer** with no prior license has `application_status = ACTIVE` (the Pydantic default per `models.py:57`). When the backend finds no license:
> 
> * If `ENTERPRISE_EDITION_ENABLED = True`: sets `application_status = "gated_access"` → layout returns `children` ✓
> * If `ENTERPRISE_EDITION_ENABLED = False` (fresh deployer): leaves `application_status = "active"` → layout returns `<EEFeatureRedirect />` and redirects to `/chat` ✗
> 
> This creates a circular dependency: a fresh EE deployer cannot reach `/admin/billing` to purchase the license they need to unlock EE features — contradicting the PR's stated design: _"running_ee_backend=true + ee_features_enabled=false means EE routes exist but features are locked (no license yet). This lets the frontend call billing/license endpoints to help the user obtain a license."_
> 
> Consider also exempting billing routes when `ee_features_enabled` is false and `running_ee_backend` is true (fresh EE case):
> 
> ```ts
> if (settings.ee_features_enabled === false) {
>   if (
>     settings.application_status === "gated_access" ||
>     pathname?.startsWith("/admin/billing")
>   ) {
>     return children;
>   }
>   return <EEFeatureRedirect />;
> }
> ```

@greptile there is no circular dependency. I tested manually. You are hallucinating

web/src/app/ee/EEFeatureRedirect.tsx
```diff
@@ -0,0 +1,19 @@
+"use client";
+
+import { useEffect } from "react";
+import { usePathname, useRouter } from "next/navigation";
+import { toast } from "@/hooks/useToast";
+
+export default function EEFeatureRedirect() {
```
wenxi-onyx, line 7: Is there a clean way to use `useToastFromQuery from "@/hooks/useToast"` here?

backend/ee/onyx/server/settings/api.py
```diff
@@ -80,6 +80,12 @@ def apply_license_status_to_settings(settings: Settings) -> Settings:
     If LICENSE_ENFORCEMENT_ENABLED is false, ee_features_enabled is set to True
     (since EE code was loaded via ENABLE_PAID_ENTERPRISE_EDITION_FEATURES).
     """
+    # This function only runs when the backend is EE, so EE routes are
+    # guaranteed to be registered. The frontend uses this to know it can
+    # safely call EE-only endpoints (e.g. /license, /admin/billing) without
+    # getting 404s. This is always True here — CE keeps the default (False).
+    settings.running_ee_backend = True
```
wenxi-onyx: Mirror of Slack question - now that license enforcement is implemented, does it make sense to always register EE backend and just control via `ee_features_enabled`?

web/src/hooks/useBillingInformation.ts
```diff
@@ -6,48 +6,46 @@ import {
   BillingInformation,
   SubscriptionStatus,
 } from "@/lib/billing/interfaces";
+import { useSettingsContext } from "@/providers/SettingsProvider";
+
+const EMPTY_STATE = {
```
wenxi-onyx: Should this be an interface? Would avoid having to create this empty obj

backend/ee/onyx/server/settings/api.py
wenxi-onyx: It seems there are 2 booleans `running_ee_backend` and `ee_features_enabled` that touch the availability of EE features

The first one shouldn't need to exist (we should always run EE backend) and the second one is the only one we really need

web/src/app/ee/EEFeatureRedirect.tsx
nmgarza5: Good call! Looked into this — `useToastFromQuery` is designed for when a toast needs to survive a full page load (it encodes the message in URL query params). Since `router.replace("/chat")` here is a client-side navigation, the imperative `toast()` works fine — the toast provider stays mounted across route changes in Next.js App Router.

Using `useToastFromQuery` would require adding a handler on the `/chat` page to read the query param, which couples `/chat` to the EE redirect logic. The current approach keeps it self-contained. But happy to switch if you feel strongly!

backend/ee/onyx/server/settings/api.py
nmgarza5: Done! Just pushed a commit that removes `running_ee_backend` entirely. Now `ee_features_enabled` is the sole flag for gating EE feature visibility. The EE backend always runs (routes always registered), and `ee_features_enabled` controls whether features are actually unlocked based on license status.

web/src/hooks/useBillingInformation.ts
nmgarza5: Resolved — latest commit removes the `EMPTY_STATE` object entirely. `useBillingInformation` now always fetches (URL based on cloud vs self-hosted), so there's no conditional early return anymore.

nmgarza5:
@wenxi-onyx Re: the FOSS repo concern — great catch. Looked into it:

**What's problematic:** `useLicense.ts`, `useBillingInformation.ts`, and `lib/billing/svc.ts` sit outside `ee/` dirs and will be copied to `onyx-foss`. They fetch EE-only endpoints (`/api/license`, `/api/admin/billing/*`) that don't exist in FOSS, causing 404s. In practice FOSS users won't hit these code paths (the pages that use them are gated), but it's dead code that shouldn't be there.

**What's safe:** `web/src/app/ee/` layout/redirect won't be copied. Settings fields (`ee_features_enabled`) just default to `false` — correct behavior.

**Follow-up PR:** Move the billing hooks and service functions into `ee/` directories so they get stripped from the FOSS build. This is the cleanest separation.

wenxi-onyx:
> @wenxi-onyx Re: the FOSS repo concern — great catch. Looked into it:
> 
> **What's problematic:** `useLicense.ts`, `useBillingInformation.ts`, and `lib/billing/svc.ts` sit outside `ee/` dirs and will be copied to `onyx-foss`. They fetch EE-only endpoints (`/api/license`, `/api/admin/billing/*`) that don't exist in FOSS, causing 404s. In practice FOSS users won't hit these code paths (the pages that use them are gated), but it's dead code that shouldn't be there.
> 
> **What's safe:** `web/src/app/ee/` layout/redirect won't be copied. Settings fields (`ee_features_enabled`) just default to `false` — correct behavior.
> 
> **Follow-up PR:** Move the billing hooks and service functions into `ee/` directories so they get stripped from the FOSS build. This is the cleanest separation.

The direct copy paste of claude is crazy :sadge:

nmgarza5:
@greptile please review and only consider the updated pr description and logic

nmgarza5:
@greptile

PR #9046, opened 2026-03-04
fix: doc to hierarchynode connection in pruning

evan-onyx:
## Description

We were creating hierarchynodes during pruning, but not linking documents to their parent nodes. Now we do that.

## How Has This Been Tested?

added some tests to cover this case

## Additional Options

---
## Summary by cubic
Fixes pruning so documents correctly link to their parent hierarchy nodes and links nodes that are also documents. Prevents orphaned docs and keeps pruning accurate across connectors.

- **Bug Fixes**
  - Extraction now returns raw_id_to_parent: document_id → parent_hierarchy_raw_node_id (or None), and includes hierarchy node raw IDs with None to avoid pruning.
  - Pruning resolves parent raw IDs to hierarchy node IDs via Redis, falls back to the source node, and bulk-updates Document.parent_hierarchy_node_id only when changed.
  - Links hierarchy nodes to documents for sources where nodes are also documents (e.g., Confluence); skips non-supported sources.
  - Connectors now emit parent_hierarchy_raw_node_id: Confluence, Google Drive, Jira, SharePoint, Slack.
  - Expanded tests for parent mapping preservation, inclusion of hierarchy nodes in extraction, idempotent parent updates, Confluence linkage, and skipping unsupported sources.

<sup>Written for commit 8bd35933a5ce0b9c8581802812041af7b6c34557. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/background/celery/celery_utils.py
+++ b/backend/onyx/background/celery/celery_utils.py
@@ -115,7 +115,8 @@ def _extract_from_batch(
     for item in doc_list:
         if isinstance(item, HierarchyNode):
             hierarchy_nodes.append(item)
-            ids[item.raw_node_id] = None
+            if item.raw_node_id not in ids:
+                ids[item.raw_node_id] = None
         elif isinstance(item, ConnectorFailure):
             failed_id = _get_failure_id(item)
             if failed_id:
@@ -191,7 +192,9 @@ def extract_ids_from_runnable_connector(
         batch_ids = batch_result.raw_id_to_parent
         batch_nodes = batch_result.hierarchy_nodes
         doc_batch_processing_func(batch_ids)
-        all_raw_id_to_parent.update(batch_ids)
+        for k, v in batch_ids.items():
+            if v is not None or k not in all_raw_id_to_parent:
+                all_raw_id_to_parent[k] = v
         all_hierarchy_nodes.extend(batch_nodes)
 
         if callback:


backend/onyx/background/celery/tasks/pruning/tasks.py
```diff
@@ -113,6 +118,38 @@ def progress(self, tag: str, amount: int) -> None:
         super().progress(tag, amount)
 
 
+def _resolve_and_update_document_parents(
+    db_session: Session,
+    redis_client: Redis,
+    source: DocumentSource,
+    doc_ids: dict[str, str | None],
```
wenxi-onyx: maybe rename all of these instances to `raw_id_to_parent`? It's the same issue here where if I see doc_ids passed around with `dict[str, str | None]` I don't know what the dict is mapping without some explanatory comment

wenxi-onyx: I actually do prefer CC's rec of NewType --> it's just an alias, but you can use it everywhere doc_ids is passed around
```
Best practice: use a NewType or a small typed wrapper. Here are the two good options:
Option 1: NewType aliases (lightweight, zero runtime cost)
# In connectors/models.py or a shared types module                                                                                                             

from typing import NewType

DocumentId = NewType("DocumentId", str)
RawParentId = NewType("RawParentId", str)

# The dict becomes:                                                                                                                                            DocParentMap = dict[DocumentId, RawParentId | None]
```

evan-onyx: imo this is best handled by good variable names. I don't love the thought of having to click an extra time in every dictionary to find out its yet another string alias

PR #9095, opened 2026-03-05
feat: Tool call arg streaming

Danelegend:
## Description
We want to be able to stream tool call arguments to the frontend. This is particularly useful for code interpreter where currently all the code is batched before it is sent to the frontend. This enables it to stream the tool arguments as they come in.

## How Has This Been Tested?
Manual + Unit

## Additional Options
closes: https://linear.app/onyx-app/issue/ENG-3798/tool-call-delta-streaming

---
## Summary by cubic
Streams live tool call argument deltas to the UI so code and other args appear as they’re generated. Closes ENG-3798.

- New Features
  - Emit ToolCallArgumentDelta during LLM streaming for opt-in tools (Python enabled) using per-call jsonriver.Parser diffs (handles fragmented escapes; strings only).
  - Add tool_call_argument_delta to streaming union and replay an initial code delta before PythonToolStart.
  - Frontend processes these packets: groups them with the Python tool, streams code into the renderer with a “Writing code…” state, and supports collapsed streaming.

- Migration
  - Frontend: on tool_call_argument_delta, append argument_deltas[key] to the current value.
  - Tools that want streaming should implement should_emit_argument_deltas() and return True.

<sup>Written for commit bb22039616751214386336f7a0ee506f3f097d53. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/app/app/message/messageComponents/renderMessageComponent.tsx
+++ b/web/src/app/app/message/messageComponents/renderMessageComponent.tsx
@@ -1,11 +1,14 @@
 import React, { JSX, memo } from "react";
 import {
   ChatPacket,
+  CODE_INTERPRETER_TOOL_TYPES,
   ImageGenerationToolPacket,
   Packet,
   PacketType,
   ReasoningPacket,
+  SearchToolStart,
   StopReason,
+  ToolCallArgumentDelta,
 } from "../../services/streamingModels";
 import {
   FullChatState,
@@ -26,7 +29,6 @@ import { DeepResearchPlanRenderer } from "./timeline/renderers/deepresearch/Deep
 import { ResearchAgentRenderer } from "./timeline/renderers/deepresearch/ResearchAgentRenderer";
 import { WebSearchToolRenderer } from "./timeline/renderers/search/WebSearchToolRenderer";
 import { InternalSearchToolRenderer } from "./timeline/renderers/search/InternalSearchToolRenderer";
-import { SearchToolStart } from "../../services/streamingModels";
 
 // Different types of chat packets using discriminated unions
 interface GroupedPackets {
@@ -56,7 +58,12 @@ function isImageToolPacket(packet: Packet) {
 }
 
 function isPythonToolPacket(packet: Packet) {
-  return packet.obj.type === PacketType.PYTHON_TOOL_START;
+  return (
+    packet.obj.type === PacketType.PYTHON_TOOL_START ||
+    (packet.obj.type === PacketType.TOOL_CALL_ARGUMENT_DELTA &&
+      (packet.obj as ToolCallArgumentDelta).tool_type ===
+        CODE_INTERPRETER_TOOL_TYPES.PYTHON)
+  );
 }
 
 function isCustomToolPacket(packet: Packet) {
--- a/web/src/app/app/message/messageComponents/timeline/hooks/packetProcessor.ts
+++ b/web/src/app/app/message/messageComponents/timeline/hooks/packetProcessor.ts
@@ -10,6 +10,8 @@ import {
   Stop,
   ImageGenerationToolDelta,
   MessageStart,
+  ToolCallArgumentDelta,
+  CODE_INTERPRETER_TOOL_TYPES,
 } from "@/app/app/services/streamingModels";
 import { CitationMap } from "@/app/app/interfaces";
 import { OnyxDocument } from "@/lib/search/interfaces";
@@ -138,6 +140,7 @@ const CONTENT_PACKET_TYPES_SET = new Set<PacketType>([
   PacketType.SEARCH_TOOL_START,
   PacketType.IMAGE_GENERATION_TOOL_START,
   PacketType.PYTHON_TOOL_START,
+  PacketType.TOOL_CALL_ARGUMENT_DELTA,
   PacketType.CUSTOM_TOOL_START,
   PacketType.FILE_READER_START,
   PacketType.FETCH_TOOL_START,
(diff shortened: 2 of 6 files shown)


backend/onyx/server/query_and_chat/streaming_models.py
```diff
@@ -259,6 +260,16 @@ class CustomToolDelta(BaseObj):
...
 
 
+class ToolCallArgumentDelta(BaseObj):
+    type: Literal["tool_call_argument_delta"] = (
+        StreamingType.TOOL_CALL_ARGUMENT_DELTA.value
+    )
+
+    tool_type: str
+    tool_id: str
+    argument_deltas: dict[str, Any]
```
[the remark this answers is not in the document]
Danelegend: Any is technically more correct since this is general for all arguments, and we don't know the type. But in python tool they are string. In line with ToolCallDebug

backend/onyx/chat/tool_call_args_streaming.py
```diff
@@ -0,0 +1,235 @@
...
+    tool_call_delta: ChatCompletionDeltaToolCall,
+) -> Type[Tool] | None:
+    """Look up the Tool subclass for a streaming tool call delta."""
+    tool_name = tool_calls_in_progress.get(tool_call_delta.index, {}).get("name")
+    if not tool_name:
+        return None
+    return TOOL_NAME_TO_CLASS.get(tool_name)
+
+
+class _Token(NamedTuple):
```
evan-onyx: can this be a BaseModel?

backend/onyx/chat/tool_call_args_streaming.py
```diff
@@ -0,0 +1,235 @@
...
+    """A parsed JSON string with position info."""
+
+    value: str  # raw content between the quotes
+    start: int  # index of first char inside the quotes
+    end: int  # index of closing quote, or len(text) if incomplete
+    complete: bool  # whether the closing quote was found
+
+
+def _parse_json_string(text: str, pos: int) -> _Token:
```
evan-onyx: is there no existing utility to do this? Parsing stuff can be pretty finicky

backend/onyx/chat/tool_call_args_streaming.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+def _skip(text: str, pos: int, chars: str = " \t\n\r,") -> int:
+    """Advance ``pos`` past any characters in ``chars``."""
+    while pos < len(text) and text[pos] in chars:
+        pos += 1
+    return pos
+
+
+def _decode_partial_json_string(raw: str) -> str:
```
evan-onyx: yeah all of this feels like it should be implemented already in some library, and I don't think we want to have to maintain this kind of thing

backend/onyx/tools/tool_implementations/python/python_tool.py
```diff
@@ -376,3 +376,8 @@ def run(
                     rich_response=None,
                     llm_facing_response=llm_response,
                 )
+
+    @classmethod
+    @override
+    def do_emit_argument_deltas(cls) -> bool:
```
evan-onyx: Imo `should_emit_argument_deltas`

backend/onyx/chat/tool_call_args_streaming.py
```diff
@@ -0,0 +1,235 @@
...
+    tool_cls = _get_tool_class(tool_calls_in_progress, tool_call_delta)
+    if not tool_cls or not tool_cls.do_emit_argument_deltas():
+        return
+
+    fn = tool_call_delta.function
+    delta_fragment = fn.arguments if fn else None
+    if not delta_fragment:
+        return
+
+    tc_data = tool_calls_in_progress[tool_call_delta.index]
```
evan-onyx: this map somewhat concerns me, is it getting updated concurrently? If so we probably want to use a ThreadSafeDict

backend/onyx/chat/tool_call_args_streaming.py
```diff
@@ -0,0 +1,235 @@
...
+    accumulated_args = tc_data["arguments"]
+    prev_args = accumulated_args[: -len(delta_fragment)]
```
evan-onyx: nit: empty delta_fragment is covered above so semantically this is fine but in case someone later adapts this code or copies somewhere else imo we should prefer `accumulated_args[: len(accumulated_args)-len(delta_fragment)]`

backend/onyx/tools/built_in_tools.py
```diff
@@ -56,3 +56,23 @@ def get_built_in_tool_ids() -> list[str]:
 
 def get_built_in_tool_by_id(in_code_tool_id: str) -> Type[BUILT_IN_TOOL_TYPES]:
     return BUILT_IN_TOOL_MAP[in_code_tool_id]
+
+
+def _build_tool_name_to_class() -> dict[str, Type[BUILT_IN_TOOL_TYPES]]:
```
evan-onyx, line 61: is there a way to avoid building this map? It feels weird to me but if we have to it's fine

backend/onyx/chat/tool_call_args_streaming.py
Danelegend: Shouldn't be. This is defined in llm_step.py::run_llm_step_pkt_generator, and consumes deltas sequentially and synchronously

backend/onyx/chat/tool_call_args_streaming.py
Danelegend: It was originally, but got the suggestion (altho ai suggestion) to make a NamedTuple

PR #9131, opened 2026-03-06
fix: Prevent the removal and hiding of default model

Danelegend:
## Description
Currently you are able to remove and hide the default model. This throws an error to prevent that

## How Has This Been Tested?
Manual + Tests

## Additional Options

---
## Summary by cubic
Prevents removing or hiding the default LLM model and keeps defaults stable during auto‑mode transitions. Default detection ignores visibility; sync only switches the default when this provider owns it and the recommendation changed, with atomic commits.

- **Bug Fixes**
  - `upsert_llm_provider` rejects deleting or hiding the current default; switch the default first.
  - Auto‑mode transition preserves existing `model_configurations` to avoid deletions; `sync_auto_mode_models` adds/hides models and updates the default atomically, no‑ops when already matching, and switches to the new recommendation while keeping both models visible.
  - Default lookup no longer filters by visibility.
  - Setup and seeding catch `ValueError` from protection checks and log warnings instead of failing.

<sup>Written for commit c0a0a3eea7ed6137b751b8a81a650b587621b810. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/db/llm.py
+++ b/backend/onyx/db/llm.py
@@ -853,21 +853,13 @@ def sync_auto_mode_models(
             and current_default.llm_provider_id == provider.id
             and current_default.name != recommended_default.name
         ):
-            try:
-                _update_default_model__no_commit(
-                    db_session=db_session,
-                    provider_id=provider.id,
-                    model=recommended_default.name,
-                    flow_type=LLMModelFlowType.CHAT,
-                )
-                changes += 1
-            except ValueError:
-                logger.warning(
-                    "Recommended default model '%s' not found "
-                    "for provider_id=%s; skipping default update.",
-                    recommended_default.name,
-                    provider.id,
-                )
+            _update_default_model__no_commit(
+                db_session=db_session,
+                provider_id=provider.id,
+                model=recommended_default.name,
+                flow_type=LLMModelFlowType.CHAT,
+            )
+            changes += 1
 
     db_session.commit()
     return changes


Danelegend:
@greptile

backend/onyx/server/manage/llm/api.py
```diff
@@ -478,20 +479,6 @@ def put_llm_provider(
...
                         config,
                     )
-
-                    # If this provider was the default before the transition,
-                    # restore the default using the recommended model.
-                    if was_default_provider:
-                        recommended = config.get_default_model(
-                            llm_provider_upsert_request.provider
-                        )
-                        if recommended:
```
Danelegend, line 488: sync_auto_mode_models actually takes care of this

backend/onyx/db/llm.py
```diff
@@ -538,7 +562,6 @@ def fetch_default_model(
         .options(selectinload(ModelConfiguration.llm_provider))
         .join(LLMModelFlow)
         .where(
-            ModelConfiguration.is_visible == True,  # noqa: E712
```
Danelegend, line 541: This may not actually be valid with contextual llms. The restriction should be enforced at the LLMProvider layer rather than the LLMModelFlow layer

Danelegend:
@cubic

backend/onyx/db/llm.py
```diff
@@ -814,30 +838,23 @@ def sync_auto_mode_models(
...
+    # Update the default if this provider currently holds the global CHAT default.
+    # We flush (but don't commit) so that _update_default_model can see the new
+    # model rows, then commit everything atomically to avoid a window where the
+    # old default is invisible but still pointed-to.
+    db_session.flush()
 
-    # Update the default if this provider currently holds the global CHAT default
     recommended_default = llm_recommendations.get_default_model(provider.provider)
     if recommended_default:
-        current_default_name = db_session.scalar(
```
Danelegend, line 822: A bit unrelated to this PR, but I did want to clean up this logic. Initially implemented as a fix removal of default model due to sync

justin-tahara (approved):

PR #9152, opened 2026-03-06
docs(cli): improve onyx-cli SKILL.md and fix README default server URL

rohoswagger:
## Description

- Fix incorrect default `ONYX_SERVER_URL` in `cli/README.md` (`http://localhost:3000` → `https://cloud.onyx.app`, matching the actual code default)
- Add `go install github.com/onyx-dot-app/onyx/cli@latest` as an install option in SKILL.md
- Document `ask` command statelessness — each call starts a fresh session with no cross-call context
- Show full JSON envelope shape for `citation_info` event including field names and `placement` metadata

## How Has This Been Tested?

Docs-only change, no code modified.

### DIFF
--- a/.cursor/skills/onyx-cli/SKILL.md
+++ b/.cursor/skills/onyx-cli/SKILL.md
@@ -23,12 +23,6 @@ which onyx-cli
 pip install onyx-cli
 ```
 
-**Via `go install`:**
-
-```bash
-go install github.com/onyx-dot-app/onyx/cli@latest
-```
-
 **From source (Go):**
 
 ```bash


.cursor/skills/onyx-cli/SKILL.md
```diff
@@ -23,6 +23,12 @@ which onyx-cli
 pip install onyx-cli
 ```
 
+**Via `go install`:**
```
jmelahman: I would just ignore this tbh. Many go packages have alternative packaging systems, `bazel`, `make`, in our case `uv`. `go install` is typically only used for basic apps and might not be worth the effort to maintain as a monorepo.

rohoswagger: gotcha, sounds good

jmelahman: `uv pip install` has a way to build packages from source given a github URL if you are interested, `uv pip install git+https://github.com/onyx-dot-app/onyx#cli`, but a main feature of the pip package is that it's prebuilt

PR #9177, opened 2026-03-08
fix: set event hook for wrapping values into SensitiveValue

wenxi-onyx:
## Description

- Fixes 'str' object has no attribute 'get_value' errors when creating or updating Slack bots (and any other model with encrypted columns)
- The root cause: assigning a raw string to an encrypted column (e.g. slack_bot.bot_token = "xoxb-...") left it as a plain str in memory. Any code reading the attribute in the same session expected a SensitiveValue and crashed.
- Adds a SQLAlchemy set event hook that automatically wraps raw values in SensitiveValue at assignment time, so the Python-side type is always SensitiveValue regardless of whether the value came from the DB or was assigned in application code.

Changes:
- _EncryptedBase base class: Extracts shared logic (wrap_raw, compare_values) from EncryptedString and EncryptedJson into a common base, with an _is_json flag to differentiate behavior
- mapper_configured event listener: Registers a set event on every _EncryptedBase column across all ORM models. On assignment, raw values are encrypted and wrapped in SensitiveValue immediately.
- SensitiveValue.__eq__: Changed from comparing encrypted bytes (which breaks when the same plaintext is encrypted twice with different IVs) to comparing decrypted content.
    - Returns NotImplemented for non-SensitiveValue comparisons instead of raising. This is required for SQLAlchemy's attribute dirty-tracking, which compares old vs new values with ==. During this check one side can be a SensitiveValue and the other a raw str — raising here crashed SQLAlchemy internals. NotImplemented tells Python "I don't know how to compare with this type," which falls through to identity comparison and correctly marks the attribute as dirty.
- SensitiveValue.__hash__: Updated to hash decrypted content for consistency with __eq__.
- Slack bot CRUD tests: The existing integration tests were passing mock_sensitive_value objects that bypassed the actual SensitiveValue wrapping path, so they never caught this bug. Added external dependency unit tests that exercise the real insert_slack_bot / update_slack_bot functions against Postgres and assert the returned tokens are proper SensitiveValue instances.

## Additional Options

---
## Summary by cubic
Auto-wrap values assigned to encrypted columns into `SensitiveValue` via a `sqlalchemy` attribute set hook. Fixes raw string tokens in SlackBot and ensures nested credential JSON round-trips cleanly without marking unchanged values dirty.

- **Bug Fixes**
  - Registered `mapper_configured` "set" event on all `_EncryptedBase` columns to wrap raw values for `EncryptedString` and `EncryptedJson` (with a double-registration guard), so assigned attributes are always `SensitiveValue`.
  - SlackBot CRUD now returns `SensitiveValue` for `bot_token`, `app_token`, and `user_token`; added regression tests for insert/update and `SlackBot.from_model`.
  - Added `Credential` tests for nested `credential_json` round-trip and `sqlalchemy` dirty-tracking (identical reassignment not modified). Updated OAuth config, user OAuth token, `oauth_token_manager`, and Discord bot tests to expect `SensitiveValue` and read secrets via `.get_value(...)`.

- **Refactors**
  - Introduced `_EncryptedBase` shared by `EncryptedString` and `EncryptedJson` with `wrap_raw`, strict input type checks (str vs dict), and common comparison logic.
  - Updated `SensitiveValue` equality to compare decrypted content and return `NotImplemented` for non-sensitive comparisons; hashing now uses decrypted content with stable, sorted JSON for nested dicts.

<sup>Written for commit eea87d33ca32e2f1dd8746cf597275cf93af7b71. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/db/models.py
+++ b/backend/onyx/db/models.py
@@ -133,7 +133,18 @@ class _EncryptedBase(TypeDecorator):
         SensitiveValue, regardless of whether the value was loaded from the DB
         or assigned in application code.
         """
-        raw_str = json.dumps(value) if self._is_json else value
+        if self._is_json:
+            if not isinstance(value, dict):
+                raise TypeError(
+                    f"EncryptedJson column expected dict, got {type(value).__name__}"
+                )
+            raw_str = json.dumps(value)
+        else:
+            if not isinstance(value, str):
+                raise TypeError(
+                    f"EncryptedString column expected str, got {type(value).__name__}"
+                )
+            raw_str = value
         return SensitiveValue(
             encrypted_bytes=encrypt_string_to_bytes(raw_str),
             decrypt_fn=decrypt_bytes_to_string,
@@ -151,7 +162,7 @@ class _EncryptedBase(TypeDecorator):
 
 
 class EncryptedString(_EncryptedBase):
-    _is_json = False
+    _is_json: bool = False
 
     def process_bind_param(
         self, value: str | SensitiveValue[str] | None, dialect: Dialect  # noqa: ARG002
@@ -177,7 +188,7 @@ class EncryptedString(_EncryptedBase):
 
 
 class EncryptedJson(_EncryptedBase):
-    _is_json = True
+    _is_json: bool = True
 
     def process_bind_param(
         self,
@@ -203,6 +214,9 @@ class EncryptedJson(_EncryptedBase):
         return None
 
 
+_REGISTERED_ATTRS: set[str] = set()
+
+
 @event.listens_for(Mapper, "mapper_configured")
 def _register_sensitive_value_set_events(
     mapper: Mapper,
@@ -214,6 +228,13 @@ def _register_sensitive_value_set_events(
             if isinstance(col.type, _EncryptedBase):
                 col_type = col.type
                 attr = getattr(class_, prop.key)
+
+                # Guard against double-registration (e.g. if mapper is
+                # re-configured in test setups)
+                attr_key = f"{class_.__qualname__}.{prop.key}"
+                if attr_key in _REGISTERED_ATTRS:
+                    continue
+                _REGISTERED_ATTRS.add(attr_key)
 
                 @event.listens_for(attr, "set", retval=True)
                 def _wrap_value(
--- a/backend/onyx/utils/sensitive.py
+++ b/backend/onyx/utils/sensitive.py
@@ -128,6 +128,8 @@ class SensitiveValue(Generic[T]):
         value = self._decrypt()
 
         if not apply_mask:
+            # Callers must not mutate the returned dict — doing so would
+            # desync the cache from the encrypted bytes and the DB.
             return value
 
         # Apply masking


backend/onyx/utils/sensitive.py
```diff
@@ -174,18 +174,17 @@ def __getitem__(self, key: Any) -> NoReturn:
...
-            # Compare encrypted bytes for equality check
-            return self._encrypted_bytes == other._encrypted_bytes
-        raise SensitiveAccessError(
-            "Cannot compare SensitiveValue with non-SensitiveValue. "
-            "Use .get_value(apply_mask=True/False) to access the value for comparison."
-        )
+        """Compare SensitiveValues by their decrypted content."""
+        if not isinstance(other, SensitiveValue):
+            return NotImplemented
+        return self._decrypt() == other._decrypt()
```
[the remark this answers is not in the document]
wenxi-onyx: mypy does not flag this, so although technically correct, it is not necessary to add these type ignores

backend/tests/external_dependency_unit/slack_bot/test_slack_bot_crud.py
```diff
@@ -0,0 +1,69 @@
...
+
+    assert isinstance(slack_bot.bot_token, SensitiveValue)
+    assert isinstance(slack_bot.app_token, SensitiveValue)
+    assert isinstance(slack_bot.user_token, SensitiveValue)
+
+    # Verify from_model works without error
+    pydantic_bot = SlackBot.from_model(slack_bot)
+    assert pydantic_bot.bot_token  # masked, but not empty
+    assert pydantic_bot.app_token
+
```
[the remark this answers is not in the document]
wenxi-onyx: added

backend/onyx/utils/sensitive.py
```diff
@@ -174,18 +174,17 @@ def __getitem__(self, key: Any) -> NoReturn:
...
-        if isinstance(other, SensitiveValue):
-            # Compare encrypted bytes for equality check
-            return self._encrypted_bytes == other._encrypted_bytes
-        raise SensitiveAccessError(
-            "Cannot compare SensitiveValue with non-SensitiveValue. "
-            "Use .get_value(apply_mask=True/False) to access the value for comparison."
-        )
+        """Compare SensitiveValues by their decrypted content."""
+        if not isinstance(other, SensitiveValue):
+            return NotImplemented
```
wenxi-onyx, line 184: When SQLAlchemy checks if an attribute is "dirty" (changed), it compares the old value to the new value using ==. During this comparison, one side might be a SensitiveValue and the other a raw str — for example, right before the set event wraps the new value.

The old code raised SensitiveAccessError on SensitiveValue == str, which crashed SQLAlchemy's internal dirty-tracking machinery.

Returning NotImplemented is the Python convention for "I don't know how to compare myself to this type." Python then tries the other operand's __eq__, and if that also returns NotImplemented, the result is False (not equal). No exception, no crash — SQLAlchemy just sees "old != new" and marks the attribute as dirty, which is the correct behavior.

nmgarza5: can you still add a helpful error/warning message stating the passed value is not a SensitiveValue? In the event this error is actually hit, not implemented doesnt really tell me much...even though that is the convention.

backend/onyx/db/models.py
```diff
@@ -117,10 +119,39 @@ class Base(DeclarativeBase):
...
+        Called by the attribute set event so the Python-side type is always
+        SensitiveValue, regardless of whether the value was loaded from the DB
+        or assigned in application code.
+        """
+        raw_str = json.dumps(value) if self._is_json else value
+        return SensitiveValue(
+            encrypted_bytes=encrypt_string_to_bytes(raw_str),
+            decrypt_fn=decrypt_bytes_to_string,
+            is_json=self._is_json,
+        )
```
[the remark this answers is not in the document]
nmgarza5: i dont think this is a big deal

backend/onyx/db/models.py
```diff
@@ -184,14 +202,30 @@ def process_result_value(
...
-        if x is None or y is None:
-            return x == y
-        if isinstance(x, SensitiveValue):
-            x = x.get_value(apply_mask=False)
-        if isinstance(y, SensitiveValue):
-            y = y.get_value(apply_mask=False)
-        return x == y
+
+@event.listens_for(Mapper, "mapper_configured")
+def _register_sensitive_value_set_events(
```
nmgarza5, line 221: any concern this will hinder db read performance? this basically means any column read needs to check this (i think?) might be negligible which is fine but something to consider

wenxi-onyx: The mapper_configured event fires once per model class at import time when SQLAlchemy configures the mapper. It registers set events on only the columns that use _EncryptedBase.

The set event itself only fires on attribute assignment (writes), not reads. Reading credential.credential_json goes through process_result_value on the TypeDecorator, which is unchanged.

wenxi-onyx: RE: NotImplemented

This isn't an error path. NotImplemented is a signal to the Python runtime, not to the user. When __eq__ returns NotImplemented, Python silently tries the other operand's __eq__, and if that also returns NotImplemented, the result is just False (not equal). No exception is raised, no message is displayed anywhere.

Adding a log message would be noisy. this comparison happens on every attribute assignment as part of normal SQLAlchemy internals.

If developer ends up running into this, they see something like str A != SensitiveValue[str] B, which, hopefully the type hints indicate that something might be wrong/they will investigate the SensitiveValue class. 

Will add an explanatory comment --> unfortunately, this might just non-ideal

nmgarza5: ^^ lol your claude is replying to the wrong comment

wenxi-onyx: It was me LMAO! I can't reply to your other comment for some reason (no text box shows)

PR #9196, opened 2026-03-09
fix(api memory): replace glibc with jemalloc for memory allocating

Bo-Onyx:
## Description

During a DR, the API server allocates large buffers — LLM streaming chunks, 100K+ token conversation history strings, and URL fetch content — peaking at ~400–500 MB above baseline. Python frees all of it, but glibc does not return those pages to the OS due to poor fragmentation handling. This PR replacing glibc with jemalloc for memory allocation.

linear: https://linear.app/onyx-app/project/memory-leaks-and-api-pods-crashing-88fcbf1510b0/overview
doc: https://docs.google.com/document/d/1_RfH2WzME73Bw4PaRcbEIPkNtdpVrvJpz6uaZ4XViyE/edit?tab=t.4564xw4az5tj

## How Has This Been Tested?

Before the fix:
Screenshot 2026-03-06 at 5 20 25 PM
After the fix:
Screenshot 2026-03-07 at 5 56 35 PM

## Additional Options

---
## Summary by cubic
Switch API server and workers to `jemalloc` so freed memory returns to the OS and RSS drops back to baseline after large requests. Addresses the Linear issue on memory leaks and pod crashes under DR-scale loads.

- **Bug Fixes**
  - Install `libjemalloc2` and set `LD_PRELOAD=libjemalloc.so.2` in `backend/Dockerfile`, placed after all RUN steps so only runtime processes use it.
  - Load tests show peak memory is released after requests; no sustained 400–500 MB RSS growth.

<sup>Written for commit 8a1461739dc0293cbeb03ab77f3e3fee67109b02. Summary will update on new commits.</sup>

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -52,12 +52,6 @@ RUN apt-get update && \
     rm -rf /var/lib/apt/lists/* && \
     apt-get clean
 
-# Use jemalloc instead of glibc malloc to reduce memory fragmentation
-# in long-running Python processes (API server, Celery workers).
-# The soname is architecture-independent; the dynamic linker resolves
-# the correct path from standard library directories.
-ENV LD_PRELOAD=libjemalloc.so.2
-
 # Conditionally install Node.js 20 for Craft (required for Next.js)
 # Only installed when ENABLE_CRAFT=true
 RUN if [ "$ENABLE_CRAFT" = "true" ]; then \
@@ -172,6 +166,13 @@ ENV PYTHONPATH=/app
 # Default ONYX_VERSION, typically overriden during builds by GitHub Actions.
 ARG ONYX_VERSION=0.0.0-dev
 ENV ONYX_VERSION=${ONYX_VERSION}
+
+# Use jemalloc instead of glibc malloc to reduce memory fragmentation
+# in long-running Python processes (API server, Celery workers).
+# The soname is architecture-independent; the dynamic linker resolves
+# the correct path from standard library directories.
+# Placed after all RUN steps so build-time processes are unaffected.
+ENV LD_PRELOAD=libjemalloc.so.2
 
 # Default command which does nothing
 # This container is used by api server and background which specify their own CMD


deployment/docker_compose/docker-compose.dev.yml
```diff
@@ -12,6 +12,8 @@ services:
   api_server:
     ports:
       - "8080:8080"
+    environment:
+      - LD_PRELOAD=/usr/lib/aarch64-linux-gnu/libjemalloc.so.2
```
jmelahman: Is there a better path for this? One which doesn't include the host architecture. On my machine, I have a `/usr/lib/libjemalloc.so.2` path

backend/Dockerfile
```diff
@@ -46,7 +46,8 @@ RUN apt-get update && \
         pkg-config \
         gcc \
         nano \
-        vim && \
+        vim \
+        libjemalloc2 && \
```
jmelahman:
```suggestion
        libjemalloc2 \
        && \
```
nit: reduces future deltas

deployment/docker_compose/docker-compose.dev.yml
Bo-Onyx: resolved by putting the logic in docker-compose.yml

deployment/docker_compose/docker-compose.yml
```diff
@@ -45,7 +45,10 @@ services:
       args:
         - ENABLE_CRAFT=${ENABLE_CRAFT:-false}
     command: >
-      /bin/sh -c "alembic upgrade head &&
+      /bin/sh -c "
+      JEMALLOC=$(find /usr/lib -name 'libjemalloc.so.2' | head -1) &&
```
jmelahman: I would only do this if we don't know where this is located, but we should always know since we're building the image ourselves. If we're concerned about this breaking in the future, we should catch that at build time with something like a container-structure-test, https://github.com/GoogleContainerTools/container-structure-test

Bo-Onyx: updated, tried a different approach, which should work for both local docker and awd pod

PR #9224, opened 2026-03-10
fix(llm): don't send tool_choice when no tools are provided

### LINKED ISSUE #9010: Fireworks provider fails when adding/testing models: tool_choice sent as null → Fireworks rejects request
I'm using a Self-hosted Onyx via Docker Compose
* `onyxdotapp/onyx-backend:craft-latest`
* Fireworks provider configured in Admin → LLM (API base: `https://api.fireworks.ai/inference/v1`)

Scenario

1. Admin → Configuration → LLM → Fireworks (existing provider config)
2. Add an additional model (e.g. `accounts/fireworks/models/kimi-k2p5` or any other)
3. Click Test/Save

Onyx returns 400 during provider/model test. Logs show Fireworks rejecting `tool_choice: null`:

```
litellm.BadRequestError: Fireworks_aiException - {"error":{"object":"error","type":"invalid_request_error","code":"invalid_request_error","message":"2 request validation errors: Input should be 'auto', 'none', 'any' or 'required', field: 'tool_choice.literal[\\'auto\\',\\'none\\',\\'any\\',\\'required\\']', value: None; Input should be a valid dictionary or object to extract fields from, field: 'tool_choice.FunctionSelection', value: None"}}
```

**Notes**
Grepping the container shows Onyx sets `tool_choice = None` when no tools are provided in `/app/onyx/llm/multi_llm.py` (“If no tools are provided, tool_choice should be None”), but then the request still passes `tool_choice` through, causing Fireworks to error because it does not accept null.

Fizza-Mukhtar:
## Description
Some LLM providers (e.g. Fireworks) reject requests where `tool_choice` is 
explicitly `null`, causing a 400 error when adding/testing models.

`tool_choice` was always passed to `litellm.completion()` even when set to 
`None` (no tools provided). Fixed by moving `tool_choice` into `optional_kwargs` 
so it's only included in the request when tools are actually present.

## How Has This Been Tested?
Verified by code inspection — `tool_choice` is now only passed when 
`tools` is non-empty and `tool_choice` is not None.

Fixes #9010

## Additional Options
- [X] [Optional] Override Linear Check

---
## Summary by cubic
Stop sending `tool_choice` when no tools are provided to avoid provider 400s (e.g., Fireworks) caused by null `tool_choice` values (fixes #9010). Updates tests and adds a regression check to match the new request shape.

- **Bug Fixes**
  - Move `tool_choice` into `optional_kwargs` and include it only when `tools` are present and `tool_choice` is not None; stop passing it unconditionally to `litellm.completion()`.
  - Tests: remove `tool_choice=None` assertions, add a regression test ensuring no `tool_choice` is sent when `tools` is None, fix the test to use the correct `UserMessage` field and a streaming mock, and apply Black formatting.
  - Normalize line endings to LF to avoid noisy diffs across OSes.

<sup>Written for commit 25a082a36e2f4366dd2af31f2c72f40c830cccae. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/unit/onyx/llm/test_multi_llm.py
+++ b/backend/tests/unit/onyx/llm/test_multi_llm.py
@@ -1460,4 +1460,3 @@ def test_no_tool_choice_sent_when_no_tools(default_multi_llm: LitellmLLM) -> Non
         assert (
             "tool_choice" not in kwargs
         ), "tool_choice must not be sent to providers when no tools are provided"
-               


backend/tests/unit/onyx/llm/test_multi_llm.py
```diff
@@ -1429,3 +1427,37 @@ def test_strip_tool_content_merges_consecutive_tool_results() -> None:
...
+    with patch("litellm.completion") as mock_completion:
+        mock_completion.return_value = mock_stream_chunks
+
+        default_multi_llm.invoke(messages, tools=None)
+
+        _, kwargs = mock_completion.call_args
+        assert (
+            "tool_choice" not in kwargs
+        ), "tool_choice must not be sent to providers when no tools are provided"
+
```
justin-tahara: Can we make sure to fix the newline?

Fizza-Mukhtar: Done

PR #9244, opened 2026-03-10
fix(storybook): case-sensitivity, icon rename, and story fixes

nmgarza5:
## Description

Fixes several issues preventing Storybook from building correctly on case-sensitive Linux CI filesystems and cleans up minor story bugs:

- **Case-sensitivity fix**: Story files were tracked under PascalCase directories (`Content/`, `ContentAction/`, `IllustrationContent/`) while source files live in lowercase dirs (`content/`, `content-action/`, `illustration-content/`). On macOS (case-insensitive) this works fine, but on Linux CI these become separate directories and imports break. Moved all story files to the correct lowercase paths.
- **Icon rename**: `import.tsx` → `import-icon.tsx` — `import` is a JS reserved word that causes Vite bundling issues.
- **Interactive "none" variant**: Fixed invisible text in the VariantMatrix story by using inline color instead of the `interactive-foreground` CSS class (which relies on a CSS variable that `variant="none"` doesn't set).
- **ButtonRenaming console noise**: Replaced `console.log` callbacks with silent no-ops to stop `"Closed"` spam on page load (caused by `autoFocus` + `onBlur` firing during Storybook render).

## How Has This Been Tested?

- `npm run storybook:build` passes cleanly
- All 60+ component docs pages validated via Playwright on the deployed Storybook
- ran the gh deployment locally and it worked `gh workflow run "Storybook Deploy" --repo onyx-dot-app/onyx --ref nik/fix-storybook-case-sensitivity`

## Additional Options

### DIFF
--- a/.github/workflows/storybook-deploy.yml
+++ b/.github/workflows/storybook-deploy.yml
@@ -48,7 +48,7 @@ jobs:
 
       - name: Deploy to Vercel (Production)
         working-directory: web
-        run: npx --yes "$VERCEL_CLI" deploy storybook-static/ --prod --yes --token=${{ secrets.VERCEL_TOKEN }}
+        run: npx --yes "$VERCEL_CLI" deploy storybook-static/ --prod --yes --token="$VERCEL_TOKEN"
 
   notify-slack-on-failure:
     needs: Deploy-Storybook


nmgarza5:
**Addressing review comments:**

**Cubic (Disabled prop):** Valid — fixed in 9e6cf3c. Added the missing `disabled` prop to `<Disabled>`.

**Greptile (redundant --token):** Disagree. The Vercel CLI does **not** auto-read `VERCEL_TOKEN` from env vars. The previous workflow run (without `--token`) failed with `Error: No existing credentials found. Please run vercel login or pass "--token"`. Adding the explicit `--token` flag is what fixed the deploy. The `preview.yml` workflow in this same repo also passes `--token` explicitly for the same reason.

PR #9245, opened 2026-03-10
fix(fe): correctly parse comma literals in CSVs

jmelahman:
## Description

The CSV parser in the frontend currently indiscriminately splits on all commas, but CSVs support comma literals when they're enclosed in quotes, so handle that case.

## How Has This Been Tested?

Unit-tests included

**before**
(note the presence of literal quotes and unexpected split in the `company` column)

20260310_12h10m54s_grim

**after**
20260310_12h11m36s_grim

## Additional Options

---
## Summary by cubic
Fixes CSV parsing in the frontend to handle quoted commas and strip surrounding quotes, preserving full field values. Also supports escaped quotes, embedded newlines, and CRLF.

- **Bug Fixes**
  - Added a `parseCSV` that supports quoted fields, escaped quotes, embedded newlines, CRLF, and errors on unterminated quotes.
  - Replaced naive splitting in `CSVContent` with `parseCSV`, mapping headers and rows by field index.
  - Added parser unit tests and updated `jest.config.js` to include `src/components` tests.

<sup>Written for commit 690a6ffd1ebb9ae2a104b39a10c9ee253d2209e1. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/components/tools/CSVContent.tsx
+++ b/web/src/components/tools/CSVContent.tsx
@@ -62,7 +62,7 @@ const CsvContent: React.FC<ContentComponentProps> = ({
       const csvData = await response.text();
       const rows = parseCSV(csvData.trim());
       const firstRow = rows[0];
-      if (!firstRow || firstRow.length === 0) {
+      if (!firstRow) {
         throw new Error("CSV file is empty");
       }
       const parsedHeaders = firstRow;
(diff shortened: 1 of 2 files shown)


web/src/components/tools/CSVContent.tsx
```diff
@@ -173,3 +174,49 @@ const csvCache = new Map<
   string,
   { headers: string[]; data: Record<string, string>[] }
 >();
+
+export function parseCSV(text: string): string[][] {
```
Danelegend, line 178: Is there a library/util function that's easy and worth using instead?

jmelahman: Eh. Thought about it. Don't think it's worth a dep tbh. It's a simple and well-defined spec

PR #9255, opened 2026-03-10
fix(craft): stop proxied webapp asset and HMR reload leaks

rohoswagger:
The fix does two things in the Craft webapp proxy:
  - rewrites leaked /_next/... asset URLs so they stay under /api/build/sessions/<id>/webapp/...
  - injects a small HMR websocket shim into proxied HTML so the page does not open a root-scoped /_next/webpack-hmr socket

  Why it works:

  - The font/static 404s were caused by preload and asset references escaping to root craft-dev.onyx.app/_next/... instead of going through the session proxy. Rewriting both response bodies and Link headers fixes that.
  - The periodic refresh was caused by Next’s dev client repeatedly trying a root HMR websocket, getting failures, and eventually reloading. The injected shim intercepts that specific HMR websocket path before the browser
    opens the real socket, so the reconnect/reload loop never starts.
    
 ### The Injection Script
The injection script patches window.WebSocket in the page.

  What it does:

  - If page code tries to open a normal websocket, it leaves it alone.
  - If page code tries to open a Next HMR websocket like /_next/webpack-hmr or /_next/hmr, it intercepts that call.
  - Instead of letting the browser create the real socket, it returns a tiny fake websocket object that looks “open” enough for the client not to enter its reconnect/failure loop.

  Why this matters:

  - The refresh loop was not coming from normal asset requests.
  - It was coming from the Next dev client repeatedly trying to connect to root HMR, getting failures, and eventually triggering a reload.
  - Rewriting static URLs fixed the 404s, but did not stop that runtime HMR behavior.
  - The shim fixes it at the exact point where the browser would otherwise create the bad websocket.

  Why inject into HTML:

  - This behavior happens in browser runtime code, after the page starts executing.
  - Backend URL rewriting can only affect response text and headers.
  - To change what the Next dev client does at runtime, the fix has to run in the page before that client initializes.
  - Injecting the script into <head> guarantees it runs early enough.

## How Has This Been Tested?

- `uv run pytest -xv backend/tests/unit/build/test_rewrite_asset_paths.py`
- Playwright validation against `craft-dev` for an active session confirmed:
  - proxied font preload requests return `200`
  - no root `/_next/static/media/...` 404 leak on initial load
  - no root `wss://.../_next/webpack-hmr` reconnect loop
  - no periodic preview refresh after redeploying tag `experimental-cc4a.87`

## Additional Options

### DIFF
--- a/backend/onyx/server/features/build/api/api.py
+++ b/backend/onyx/server/features/build/api/api.py
@@ -290,13 +290,14 @@ def _rewrite_asset_paths(content: bytes, session_id: str) -> bytes:
         text,
     )
     for hmr_path in hmr_paths:
+        escaped_hmr_path = hmr_path.replace("/", r"\/")
         text = text.replace(
             f"{webapp_base_path}{hmr_path}",
             hmr_path,
         )
         text = text.replace(
-            f"{escaped_webapp_base_path}{hmr_path.replace('/', r'\/')}",
-            hmr_path.replace("/", r"\/"),
+            f"{escaped_webapp_base_path}{escaped_hmr_path}",
+            escaped_hmr_path,
         )
     text = re.sub(
         r'"(/(?:[a-zA-Z0-9_-]+/)*[a-zA-Z0-9_-]+\.json)"',


rohoswagger:
@greptile @cubic-dev review this

wenxi-onyx (approved):

PR #9275, opened 2026-03-11
fix(indexing): Stop deep-copy during indexing

Danelegend:
## Description
Currently we use model_dump() to copy the elements into the new chunk element. This results in a deep copy of all the data, most notably the Document model (attribute of DocAwareChunk). As such each deep copy duplicates the document being processed.

In the pipeline, we currently do 2 deep copies per chunk (when creating IndexChunks and DocMetadataIndexAwareChunk) . For files that explode in the number of chunks (>1000), this results in many document copies 2 * num_chunks, which can cause massive memory usage.

This fix makes use of references instead.

Indexing pipeline currently does not mutate the document

## How Has This Been Tested?
Tested with ~1.6MB worth of google drive files

Screenshot 2026-03-11 at 11 44 52 AM

This brings us a 20% win with current test data

## Additional Options

---
## Summary by cubic
Stop deep-copying chunks and documents during indexing by constructing new chunks from references instead of `model_dump`. This reduces memory and speeds up ingest (~20% in tests).

- **Bug Fixes**
  - Build `IndexChunk` and `DocMetadataAwareIndexChunk` with `model_construct` + a new `shallow_model_dump` helper to preserve `Document` references and avoid deep copies across indexing and embedding paths.
  - Replace per-document chunk counting with a single pass over chunks in `document_indexing_adapter`.

<sup>Written for commit 4415c453e330626fc8661ba1e24cbbcd0313ea2e. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/indexing/embedder.py
+++ b/backend/onyx/indexing/embedder.py
@@ -16,6 +16,7 @@ from onyx.indexing.models import DocAwareChunk
 from onyx.indexing.models import IndexChunk
 from onyx.natural_language_processing.search_nlp_models import EmbeddingModel
 from onyx.utils.logger import setup_logger
+from onyx.utils.pydantic_util import shallow_model_dump
 from onyx.utils.timing import log_function_time
 from shared_configs.configs import INDEXING_MODEL_SERVER_HOST
 from shared_configs.configs import INDEXING_MODEL_SERVER_PORT
@@ -211,10 +212,7 @@ class DefaultIndexingEmbedder(IndexingEmbedder):
                     title_embed_dict[title] = title_embedding
 
             new_embedded_chunk = IndexChunk.model_construct(
-                **{
-                    field_name: getattr(chunk, field_name)
-                    for field_name in DocAwareChunk.model_fields
-                },
+                **shallow_model_dump(chunk),
                 embeddings=ChunkEmbedding(
                     full_embedding=chunk_embeddings[0],
                     mini_chunk_embeddings=chunk_embeddings[1:],
--- a/backend/onyx/indexing/models.py
+++ b/backend/onyx/indexing/models.py
@@ -12,6 +12,7 @@ from onyx.connectors.models import Document
 from onyx.db.enums import EmbeddingPrecision
 from onyx.db.enums import SwitchoverType
 from onyx.utils.logger import setup_logger
+from onyx.utils.pydantic_util import shallow_model_dump
 from shared_configs.enums import EmbeddingProvider
 from shared_configs.model_server_models import Embedding
 
@@ -134,10 +135,7 @@ class DocMetadataAwareIndexChunk(IndexChunk):
         ancestor_hierarchy_node_ids: list[int] | None = None,
     ) -> "DocMetadataAwareIndexChunk":
         return cls.model_construct(
-            **{
-                field_name: getattr(index_chunk, field_name)
-                for field_name in IndexChunk.model_fields
-            },
+            **shallow_model_dump(index_chunk),
             access=access,
             document_sets=document_sets,
             user_project=user_project,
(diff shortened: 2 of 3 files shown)


backend/onyx/indexing/embedder.py
```diff
@@ -210,8 +210,11 @@ def embed_chunks(
                     )[0]
                     title_embed_dict[title] = title_embedding
 
-            new_embedded_chunk = IndexChunk(
-                **chunk.model_dump(),
+            new_embedded_chunk = IndexChunk.model_construct(
+                **{
+                    field_name: getattr(chunk, field_name)
+                    for field_name in chunk.__class__.model_fields
```
Danelegend: https://docs.pydantic.dev/latest/examples/dynamic_models/

backend/onyx/indexing/adapters/document_indexing_adapter.py
```diff
@@ -122,16 +123,9 @@ def build_metadata_aware_chunks(
...
-        doc_id_to_new_chunk_cnt: dict[str, int] = {
-            document_id: len(
-                [
-                    chunk
-                    for chunk in chunks_with_embeddings
-                    if chunk.source_document.id == document_id
-                ]
-            )
-            for document_id in updatable_ids
-        }
```
Danelegend, line 134: Unnecessary copies

backend/onyx/indexing/models.py
```diff
@@ -133,9 +133,11 @@ def from_index_chunk(
...
         ancestor_hierarchy_node_ids: list[int] | None = None,
     ) -> "DocMetadataAwareIndexChunk":
-        index_chunk_data = index_chunk.model_dump()
-        return cls(
-            **index_chunk_data,
+        return cls.model_construct(
+            **{
+                field_name: getattr(index_chunk, field_name)
+                for field_name in index_chunk.__class__.model_fields
+            },
```
Danelegend: Probably also worth noting, the indexing pipeline could probably get upheaved to use a 'DocumentMetadata' class which contains data about the Document instead of the actual Document. This is because despite the chunks having reference to the document, they don't actual access the sections. They only use other various fields.

This change results in about the same memory win but is a larger refactor.

backend/onyx/indexing/embedder.py
```diff
@@ -210,8 +210,11 @@ def embed_chunks(
...
+                },
```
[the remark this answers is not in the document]
Danelegend: Is this still the right approach even though it was previously using model_dump?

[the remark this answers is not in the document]
Danelegend: Is this still the right approach even though it was previously using model_dump? @greptile

backend/onyx/indexing/embedder.py
```diff
@@ -210,8 +210,11 @@ def embed_chunks(
                     )[0]
                     title_embed_dict[title] = title_embedding
 
-            new_embedded_chunk = IndexChunk(
-                **chunk.model_dump(),
+            new_embedded_chunk = IndexChunk.model_construct(
+                **{
+                    field_name: getattr(chunk, field_name)
+                    for field_name in DocAwareChunk.model_fields
+                },
```
Danelegend: https://docs.pydantic.dev/latest/examples/dynamic_models/

Danelegend:
@greptile @cubic

backend/onyx/indexing/embedder.py
evan-onyx: this is so gross of them but ig we have to

evan-onyx: optional: have a helper function `shallow_model_dump` to do this ickiness

backend/onyx/indexing/models.py
evan-onyx: yeah this could be alright, I'm happy with the non deep copying approach though

PR #9288, opened 2026-03-12
fix(xlsx): Improve empty row/col handling

Danelegend:
## Description
Currently we skip an xlsx if it has >100 empty lines.

This means that we can still have xlsx with 100 empty rows of commas still go into the indexing pipeline. Similarly, we don't handle columns. This implements better handling for empty rows and empty columns to remove if we have a run greater than some defined number.

## How Has This Been Tested?
Unit test & manual

## Additional Options

---
## Summary by cubic
Improves .xlsx text extraction by capping consecutive empty rows/columns and switching to `csv.writer`, reducing noise from sparse sheets and fixing commas/quotes. Also reduces memory use by scanning columns in place without transposing.

- **Bug Fixes**
  - Cap runs longer than 2 empty rows/cols; keep shorter runs; drop trailing empty rows.
  - Replace “skip after >100 empty rows” with controlled capping; handle empty workbooks.
  - Serialize with `csv.writer` for correct quoting, "\n" line endings, and one line per row.
  - Return empty string for invalid files and temporary `~$` Excel files.
  - Add unit tests for capping, quoting/escaping, multiple sheets, per-row lines, and bad/empty files.

- **Refactors**
  - Remove empty-column runs without transposing to cut memory on large sheets.

<sup>Written for commit f69704734d6917a3ffe5750230e2f76abfaba2e9. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/unit/onyx/file_processing/test_xlsx_to_text.py
+++ b/backend/tests/unit/onyx/file_processing/test_xlsx_to_text.py
@@ -1,6 +1,8 @@
 import io
+from typing import cast
 
 import openpyxl
+from openpyxl.worksheet.worksheet import Worksheet
 
 from onyx.file_processing.extract_file_text import xlsx_to_text
 
@@ -9,7 +11,7 @@ def _make_xlsx(sheets: dict[str, list[list[str]]]) -> io.BytesIO:
     """Create an in-memory xlsx file from a dict of sheet_name -> matrix of strings."""
     wb = openpyxl.Workbook()
     if wb.active is not None:
-        wb.remove(wb.active)
+        wb.remove(cast(Worksheet, wb.active))
     for sheet_name, rows in sheets.items():
         ws = wb.create_sheet(title=sheet_name)
         for row in rows:


backend/onyx/file_processing/extract_file_text.py
```diff
@@ -352,6 +354,61 @@ def pptx_to_text(file: IO[Any], file_name: str = "") -> str:
...
+def _worksheet_to_matrix(worksheet: Worksheet) -> list[list[str]]:
+    """
+    Converts a singular worksheet to a matrix of values
+    """
+    rows: list[list[str]] = []
+    for worksheet_row in worksheet.iter_rows(min_row=1, values_only=True):
+        row = [str(cell or "") for cell in worksheet_row]
+        rows.append(row)
+
+    return rows
```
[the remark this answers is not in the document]
Danelegend: Like kinda valid, but this is also transitory so gonna ignore

backend/onyx/file_processing/extract_file_text.py
```diff
@@ -352,6 +354,65 @@ def pptx_to_text(file: IO[Any], file_name: str = "") -> str:
...
+    rows and removing cols if there are M consecutive empty columns
+    """
+    MAX_EMPTY_ROWS = 2  # Runs longer than this are capped to max_empty; shorter runs are preserved as-is
+    MAX_EMPTY_COLS = 2
+
+    # Row cleanup
+    matrix = _remove_empty_runs(matrix, max_empty=MAX_EMPTY_ROWS)
+
+    # Column cleanup (transpose, clean, transpose back)
+    transposed = list(map(list, zip(*matrix))) if matrix else []
```
evan-onyx: I'm somewhat worried about performance here for large csvs, python just isn't built to be good at this afaik

backend/onyx/file_processing/extract_file_text.py
```diff
@@ -352,6 +354,65 @@ def pptx_to_text(file: IO[Any], file_name: str = "") -> str:
...
+    Leading and trailing empty rows are always dropped regardless of run length,
+    since there is no adjacent non-empty row to bound the run.
+    """
+    result: list[list[str]] = []
+    empty_buffer: list[list[str]] = []
+
+    for row in rows:
+        # Check if empty
+        if not any(row):
+            empty_buffer.append(row)
```
evan-onyx: might as well not append when we reach max buffer size

PR #9299, opened 2026-03-12
feat(chat): support attaching more file types to chats

jmelahman:
## Description

Currently only a handful of file types are allowed to be attached to a chat session. This does not include many common file types such as `.py`, 

My understanding is that this `OnyxFileExtensions.ALL_ALLOWED_EXTENSIONS` is mostly relevant for connectors which are ingesting large amounts of potentially noisy data. In this case, the user is explicitly uploading the file to be processed, so I think it's fine to allow additional file types. Moreover, file types are mostly vanity anyways and and renaming any file to `.txt` allows it to be uploaded -- this simply allows them to be uploaded more conveniently + enables nicities such as syntax highlighting in the preview modals.

## How Has This Been Tested?

Unit tests included

## Additional Options

---
## Summary by cubic
Allow attaching many more file types to chats by accepting any file where text can be extracted, not just the old allowlist. This enables common code files (e.g., `.py`, `.js`, `.rs`) while still rejecting binaries and oversized content.

- **New Features**
  - Treat non-image uploads as text/doc: extract text and accept if present; reject if empty or over token threshold.
  - Keep image handling with patch-based token estimation.
  - Clearer rejections: “Unsupported file type: {ext}” for unreadable text, or “Unsupported file contents” when contents can’t be processed.
  - Added unit tests for accepting `.py` and rejecting binary `.bin`, including size limit checks.

<sup>Written for commit 717f9c24caa949fe61c8557383b163eb655a63af. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/server/features/projects/projects_file_utils.py
+++ b/backend/onyx/server/features/projects/projects_file_utils.py
@@ -220,7 +220,7 @@ def categorize_uploaded_files(
                     )
                     results.rejected.append(
                         RejectedFile(
-                            filename=filename, reason="Unsupported file contents."
+                            filename=filename, reason="Unsupported file contents"
                         )
                     )
                     continue


backend/onyx/server/features/projects/projects_file_utils.py
```diff
@@ -259,7 +264,10 @@ def categorize_uploaded_files(
                 if not text_content:
                     logger.warning(f"No text content extracted from '{filename}'")
                     results.rejected.append(
-                        RejectedFile(filename=filename, reason="Could not read file")
+                        RejectedFile(
+                            filename=filename,
+                            reason=f"Unsupported file type: {extension}",
```
Danelegend, line 268: Nit: I feel like there's more reasons for this failure. For example a file could be corrupted?

Danelegend (approved):

backend/onyx/server/features/projects/projects_file_utils.py
jmelahman: Any suggestions on copy? `Unsupported file contents.` maybe?

jmelahman: Ugh idk I feel like more generic errors is actually worse. Even with this change, we don't support extensionless files, so a plain-text `echo "hello" > my_file` will raise this error and I can find the being quite frustrating to try and reason about as a user...

PR #9362, opened 2026-03-15
feat: add MiniMax as a well-known LLM and voice provider (M3 default)

octo-patch:
## Summary
Upgrade MiniMax model configuration to use **MiniMax-M3** as the default model, replacing M2.5.

- Add MiniMax as a well-known LLM provider with **MiniMax-M3** as the default model
- Add MiniMax-M2.7 and MiniMax-M2.7-highspeed as additional model options
- Add MiniMax Cloud TTS voice provider with speech-2.8-hd and speech-2.8-turbo models
- Frontend modal for MiniMax API key configuration

## Changes
- `backend/onyx/llm/constants.py` — Register MiniMax as a known LLM provider; update MiniMax context length to 512K
- `backend/onyx/llm/well_known_providers/` — MiniMax provider config + recommended models (M3 default)
- `backend/onyx/voice/factory.py` + `backend/onyx/voice/providers/minimax.py` — MiniMax Cloud TTS provider
- `web/src/sections/modals/llmConfig/MiniMaxModal.tsx` — Frontend config modal (M3 default)
- `web/src/lib/llmConfig/providers.ts` + `web/src/interfaces/llm.ts` — Provider display name + icon mapping

## Why
MiniMax-M3 is the latest flagship model with a 512K context window, up to 128K output, and image input support. Setting it as the default ensures users get the best experience out of the box. M2.7 and M2.7-highspeed are retained as alternatives.

## Testing
- MiniMax-M3 verified via API (chat completions with image input)
- M2.7 and M2.7-highspeed verified as fallback options
- TTS models (speech-2.8-hd/turbo) tested separately

### DIFF
--- a/backend/onyx/llm/constants.py
+++ b/backend/onyx/llm/constants.py
@@ -24,6 +24,7 @@ class LlmProviderNames(str, Enum):
     OLLAMA_CHAT = "ollama_chat"
     LM_STUDIO = "lm_studio"
     MISTRAL = "mistral"
+    MINIMAX = "minimax"
     LITELLM_PROXY = "litellm_proxy"
 
     def __str__(self) -> str:
@@ -43,6 +44,7 @@ WELL_KNOWN_PROVIDER_NAMES = [
     LlmProviderNames.AZURE,
     LlmProviderNames.OLLAMA_CHAT,
     LlmProviderNames.LM_STUDIO,
+    LlmProviderNames.MINIMAX,
     LlmProviderNames.LITELLM_PROXY,
 ]
 
@@ -61,6 +63,7 @@ PROVIDER_DISPLAY_NAMES: dict[str, str] = {
     LlmProviderNames.OLLAMA_CHAT: "Ollama",
     LlmProviderNames.LM_STUDIO: "LM Studio",
     LlmProviderNames.LITELLM_PROXY: "LiteLLM Proxy",
+    LlmProviderNames.MINIMAX: "MiniMax",
     "groq": "Groq",
     "anyscale": "Anyscale",
     "deepseek": "DeepSeek",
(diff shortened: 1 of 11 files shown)


octo-patch:
Thanks for the thorough review! I've addressed all the feedback and also rebased on latest main to resolve merge conflicts.

**P1 fixes:**
- Added `getattr(litellm, "minimax_models", [])` guard to avoid AttributeError if litellm doesn't have the attribute
- Added `status != 2` filter in both `synthesize_stream` and `_synthesize_text` to prevent duplicate audio on the final completion event
- Added `try/except ValueError` around `bytes.fromhex()` with warning logging in both TTS methods
- Refactored MiniMaxModal to match the upstream shared component pattern (was using removed `FormWrapper`/`formUtils` components)

**P2 fixes:**
- Changed `receive_audio` to return `None` (not `b""`) to match the `StreamingSynthesizerProtocol` contract
- `get_minimax_model_names()` was already lazy (called inside `_get_provider_to_models_map()` function body, not at module level) -- no change needed
- Replaced duplicate `MINIMAX_PROVIDER_NAME` constant with shared `LLMProviderName.MINIMAX` enum
- `recommended-models.json` already had a trailing newline -- no change needed
- Added `minimax` to `PROVIDER_DISPLAY_ORDER` for proper UI card ordering

**Additional fixes from rebase:**
- Updated all imports to use default imports (matching upstream's refactor from named to default exports)
- Fully rewrote MiniMaxModal to use the new `LLMConfigurationModalWrapper`, `APIKeyField`, `ModelsField`, `ModelsAccessField` shared components (matching AnthropicModal pattern)
- Added onboarding variant support to MiniMaxModal

PR #9266, opened 2026-03-11
test(admin): add E2E Playwright tests for Users page

nmgarza5:
## Description

Adds comprehensive Playwright E2E tests for the admin Users page (`/admin/users`). This is a standalone test PR that sits on top of the 9-PR admin users stack.

**Coverage (26 test cases):**
- Page layout, stats bar, column headers, pagination controls
- Search (filter by email, no results, clearing search)
- Filter dropdowns (account type, status, reset)
- Column sorting (name, account type)
- Pagination navigation
- Invite users flow (single, multiple, invalid email)
- Deactivate/activate user lifecycle
- Delete user flow
- Cancel pending invite
- Inline role editing (Basic → Admin → Basic)
- Group management modal (add/remove user from group)
- Stats bar updates after invite

**Test infrastructure:**
- `UsersAdminPage.ts` — Page Object Model with robust Radix locator strategies
- `fixtures.ts` — Test fixtures with authenticated admin context
- `onyxApiClient.ts` — 6 API methods for test setup/teardown (register, invite, deactivate, delete, cancel invite, create/delete groups)

**Key locator decisions:**
- Popover getter filters `[data-radix-popper-content-wrapper]` for wrappers containing `[role="dialog"][data-state="open"]` — excludes tooltip wrappers and stale/closing popovers
- Dialog getter uses `.last()` to target confirmation modals over closing row-action popovers
- Uses `pressSequentially()` for React-controlled inputs (not `fill()`)

## How Has This Been Tested?

CI Playwright E2E suite on every push. Trace files from CI failures analyzed to diagnose and fix locator issues across multiple iterations.

## Additional Options

- [ ] [Optional] Please cherry-pick this PR to the latest release version.
- [x] [Optional] Override Linear Check

PR #9366, opened 2026-03-15
fix(admin): users page UI/UX polish

nmgarza5:
## Description

UI/UX polish pass on the admin Users page. Key changes:

- **Edit User modal rewrite**: Renamed `EditGroupsModal` → `EditUserModal`, replaced manual dropdown with Radix `Popover` for group search, added inline role editing via `InputSelect`, set `modal={false}` on Dialog to fix scroll-blocking inside portalled popovers, guarded `onClose` with `isSubmitting`, tinted group items with `bg-background-tint-01`, added `SimpleTooltip` on leave-group icon, removed dead commented-out code
- **User row actions overhaul**: Status-aware action menus (activate, deactivate, delete, cancel invite, reset password), SCIM user handling (groups/roles editable, deactivate disabled with explanation text), confirmation modals with danger styling and `isSubmitting` close guards on all modals, password reset flow with one-time display, string literals replaced with `Modal` enum, simplified inactive user menu to activate + delete only
- **Role cell cleanup**: Removed wrapper div, added `roundingVariant="compact"` to `OpenButton`, extracted role items into helper variable to keep logic out of JSX
- **Invite modal improvement**: Inline validation warning with `SvgAlertTriangle` icon for invalid emails, updated placeholder text
- **Filter test IDs**: Added `data-testid` attributes to role, group, and status filter buttons
- **Shared component enhancements**:
  - `Popover`: Added `sm` width option
  - `ShadowDiv`: `min-h-0` fix for flex containers, transition on shadow overlays, `shadowHeight` prop usage
  - `Separator`: Added `noPadding` and `paddingXRem` props
  - `Chip`: `error` prop for warning-colored right icon, fixed JSDoc
  - `InputChipField`: Simplified error icon via Chip's `error` prop, always-visible placeholder
  - `OpenButton`: `roundingVariant` override prop
  - `Section` (general-layouts): Numeric `Length` for arbitrary rem-based width/height, `className` prop
- **svc.ts**: Added `resetPassword` API function
- **UsersTable**: `aria-label` on CSV download button

## How Has This Been Tested?

manual testing. can see on the preview everything is adhering to mocks a bit more 

## Additional Options

### DIFF
--- a/web/src/refresh-pages/admin/UsersPage/EditUserModal.tsx
+++ b/web/src/refresh-pages/admin/UsersPage/EditUserModal.tsx
@@ -16,6 +16,7 @@ import SimpleTooltip from "@/refresh-components/SimpleTooltip";
 import { Section } from "@/layouts/general-layouts";
 import { toast } from "@/hooks/useToast";
 import { UserRole, USER_ROLE_LABELS } from "@/lib/types";
+import { usePaidEnterpriseFeaturesEnabled } from "@/components/settings/usePaidEnterpriseFeaturesEnabled";
 import useGroups from "@/hooks/useGroups";
 import { addUserToGroup, removeUserFromGroup, setUserRole } from "./svc";
 import type { UserRow } from "./interfaces";
@@ -50,6 +51,7 @@ export default function EditUserModal({
   onClose,
   onMutate,
 }: EditUserModalProps) {
+  const isPaidEnterpriseFeaturesEnabled = usePaidEnterpriseFeaturesEnabled();
   const { data: allGroups, isLoading: groupsLoading } = useGroups();
   const [searchTerm, setSearchTerm] = useState("");
   const [isSubmitting, setIsSubmitting] = useState(false);
@@ -86,6 +88,10 @@ export default function EditUserModal({
       (id) => !initialMemberGroupIds.has(id)
     );
   }, [memberGroupIds, initialMemberGroupIds]);
+
+  const visibleRoles = isPaidEnterpriseFeaturesEnabled
+    ? ASSIGNABLE_ROLES
+    : ASSIGNABLE_ROLES.filter((r) => r !== UserRole.GLOBAL_CURATOR);
 
   const hasRoleChange =
     user.role !== null && selectedRole !== "" && selectedRole !== user.role;
@@ -301,7 +307,7 @@ export default function EditUserModal({
                     >
                       <InputSelect.Trigger />
                       <InputSelect.Content>
-                        {user.role && !ASSIGNABLE_ROLES.includes(user.role) && (
+                        {user.role && !visibleRoles.includes(user.role) && (
                           <InputSelect.Item
                             key={user.role}
                             value={user.role}
@@ -310,7 +316,7 @@ export default function EditUserModal({
                             {USER_ROLE_LABELS[user.role]}
                           </InputSelect.Item>
                         )}
-                        {ASSIGNABLE_ROLES.map((role) => (
+                        {visibleRoles.map((role) => (
                           <InputSelect.Item
                             key={role}
                             value={role}


nmgarza5:
* **#9266** <a href="https://app.graphite.com/github/pr/onyx-dot-app/onyx/9266?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9366** <a href="https://app.graphite.com/github/pr/onyx-dot-app/onyx/9366?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/onyx-dot-app/onyx/9366?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

web/src/refresh-pages/admin/UsersPage/UserRowActions.tsx
```diff
@@ -71,21 +78,53 @@ export default function UserRowActions({
...
+          <Disabled disabled>
+            <Button
+              prominence="tertiary"
+              variant="danger"
+              icon={SvgUserX}
+              onClick={() => openModal(Modal.DEACTIVATE)}
+            >
+              Deactivate User
+            </Button>
+          </Disabled>
```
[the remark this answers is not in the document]
nmgarza5: Fixed — removed the dead `onClick`. (273bcec)

web/src/refresh-pages/admin/UsersPage/EditUserModal.tsx
```diff
@@ -162,7 +152,7 @@ export default function EditGroupsModal({
   const displayName = user.personal_name ?? user.email;
 
   return (
-    <Modal open onOpenChange={(isOpen) => !isOpen && onClose()}>
+    <Modal open onOpenChange={(isOpen) => !isOpen && onClose()} modal={false}>
```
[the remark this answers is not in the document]
nmgarza5: Not a concern — our `Modal` component renders its own overlay/backdrop independently of Radix Dialog's `modal` prop. Focus trapping is also handled by the Modal wrapper. The `modal={false}` only disables Radix's `react-remove-scroll` which was blocking wheel events on portalled Popovers outside the Dialog DOM tree. The backdrop and focus trap are intact.

web/src/refresh-pages/admin/UsersPage/EditUserModal.tsx
```diff
@@ -181,99 +171,84 @@ export default function EditGroupsModal({
...
-                                onClick={() => toggleGroup(group.id)}
-                              >
-                                {group.name}
-                              </LineItem>
-                            );
-                          })}
-                        </ShadowDiv>
-                      )}
-                    </div>
+                </Popover.Trigger>
```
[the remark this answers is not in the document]
nmgarza5: Good catch — added controlled `open` state with `onFocus` and `onChange` triggers so the popover opens on both click and keyboard interaction. (273bcec)

nmgarza5:
Responding to Greptile's "outside diff" comments:

**SCIM users can edit role inline** — Intentional. SCIM does not manage role assignments in our system — only group membership is synced from the IdP. Admins need to be able to set roles for SCIM users.

**`error` prop doesn't affect chip background** — Intentional. The `error` prop is scoped to icon coloring only. The chip represents a valid email that happens to have a validation warning — a full red background would be visually misleading. The current behavior (warning icon color only) matches the design.

web/src/refresh-pages/admin/UsersPage/UserRoleCell.tsx
```diff
@@ -89,47 +79,46 @@ export default function UserRoleCell({ user, onMutate }: UserRoleCellProps) {
...
+          </OpenButton>
+        </Popover.Trigger>
+        <Popover.Content align="start">
+          <div className="flex flex-col gap-1 p-1 min-w-[160px]">
+            {roleItems}
+          </div>
+        </Popover.Content>
+      </Popover>
+    </Disabled>
   );
```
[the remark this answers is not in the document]
nmgarza5: this is intended

web/src/layouts/general-layouts.tsx
```diff
@@ -90,11 +90,12 @@ export const heightClassmap: Record<Length, string> = {
  * @remarks
  * - The component defaults to column layout when no direction is specified
  * - Full width and height by default
- * - Prevents style overrides (className and style props are not available)
+ * - Accepts className for additional styling; style prop is not available
  * - Import using namespace import for consistent usage: `import * as GeneralLayouts from "@/layouts/general-layouts"`
  */
 export interface SectionProps
   extends WithoutStyles<React.HtmlHTMLAttributes<HTMLDivElement>> {
+  className?: string;
```
Subash-Mohan, line 98: I don’t think @raunakab would allow adding a className here. As of I know the primary idea of having a primitive here is to achieve things without className. If you want more custom handling, I recommend directly using a div.

web/src/refresh-components/Separator.tsx
```diff
@@ -7,6 +7,8 @@ import { cn } from "@/lib/utils";
 export interface SeparatorProps
   extends React.ComponentPropsWithoutRef<typeof SeparatorPrimitive.Root> {
   noPadding?: boolean;
+  /** Custom horizontal padding in rem. Overrides the default padding. */
+  paddingXRem?: number;
```
Subash-Mohan, line 11: Can't we able to use margin here instead of adding `paddingXRem ` ?

web/src/refresh-pages/admin/UsersPage/InviteUsersModal.tsx
```diff
@@ -145,9 +146,20 @@ export default function InviteUsersModal({
...
+            <div className="flex items-center gap-1 pt-1">
+              <SvgAlertTriangle
+                size={14}
+                className="text-status-warning-05 shrink-0"
+              />
+              <Text secondaryBody text03>
+                Some email addresses are invalid and will be skipped.
+              </Text>
+            </div>
+          )}
```
Subash-Mohan, line 162: Shall we wrap this inside either `input-layouts.tsx` if any of it component supports it or in `FormField.tsx` if any of these 2 supports this layout we can pass the errors in `description` or `FormFieldMessage` ? Just see if it possible use these for accessibility reasons otherwise it fine.

web/src/refresh-pages/admin/UsersPage/UserRowActions.tsx
```diff
@@ -313,6 +355,80 @@ export default function UserRowActions({
           </Text>
         </ConfirmationModalLayout>
       )}
+
+      {modal === Modal.RESET_PASSWORD && (
```
Subash-Mohan, line 359: Shall we move the modals into separate child components? It would make the code cleaner and easier to understand when we revisit this file later, and we can focus only on the modal we need to modify.

web/src/refresh-pages/admin/UsersPage/UserFilters.tsx
```diff
@@ -166,6 +166,7 @@ export default function UserFilters({
       <Popover>
         <Popover.Trigger asChild>
           <FilterButton
+            data-testid="filter-role"
```
Subash-Mohan, line 169: I remember we agreed on using aria-label for tests. Did we change that decision?

Subash-Mohan (approved):
Approving with a few comments. Everything else LGTM!

web/src/refresh-pages/admin/UsersPage/UserRowActions.tsx
nmgarza5: ahh yeah will do in a follow up. sorry saw this comment after merging

web/src/refresh-pages/admin/UsersPage/InviteUsersModal.tsx
nmgarza5: ooo yeah i'll take a look

web/src/refresh-components/Separator.tsx
nmgarza5: we dont apply margin anywhere so did this to be consisten

web/src/layouts/general-layouts.tsx
nmgarza5: Section is used enough in many random places that i think classname is warranted. And I personally don't agree with the no classname thing. Overrides are fine as long as we dont abuse them

web/src/refresh-pages/admin/UsersPage/UserFilters.tsx
nmgarza5: eh little but of both. this was a claude code thing that i glossed over but we should prefere aria-label for most cases. in this case i should have used aria label so i will update

nmgarza5: just to add on here. if the component is not needed by a screen reader we should use data-testid

PR #9384, opened 2026-03-16
refactor(indexing): Vespa & Opensearch index function use Iterable

Danelegend:
## Description
Changes the interface of index for Indexable to use an Iterable instead of a list.

The open-search implementor was also changed to use the iterable in a one-pass

## Additional Options

---
## Summary by cubic
Refactored the indexing API to accept an iterable of chunks and updated OpenSearch and Vespa to stream chunks in a single pass for lower memory use and better throughput. Added tests to confirm generators work in both old and new interfaces.

- **Refactors**
  - `Indexable.index` now takes `Iterable[DocMetadataAwareIndexChunk]` in both old and new interfaces; docstrings and the `disabled` index updated.
  - OpenSearch streams per document (buffers until doc ID changes), deletes once per doc, logs totals from `IndexingMetadata`, and records one `DocumentInsertionRecord` per doc before bulk indexing.
  - Vespa streams ID cleaning via a generator, builds the original→cleaned ID map on the fly, and batches inserts without materializing all cleaned chunks.

- **Migration**
  - Lists still work; any iterable is consumed lazily.
  - When streaming, keep chunks grouped contiguously by document.

<sup>Written for commit 925c7f7f3b068c9dc7f39f248abb687c49a2c0b6. Summary will update on new commits.</sup>


backend/onyx/document_index/opensearch/opensearch_document_index.py
```diff
@@ -668,29 +668,32 @@ def index(
...
         document_indexing_results: list[DocumentInsertionRecord] = []
-        # Try to index per-document.
-        for _, chunks in doc_id_to_chunks.items():
+        deleted_doc_ids: set[str] = set()
+        # Buffer chunks per document as they arrive from the iterable.
+        # When the document ID changes flush the buffered chunks.
+        current_doc_id: str | None = None
+        current_chunks: list[DocMetadataAwareIndexChunk] = []
+
+        def _flush_chunks(doc_chunks: list[DocMetadataAwareIndexChunk]) -> None:
```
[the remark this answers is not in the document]
Danelegend: Bro, you're tripping

backend/onyx/document_index/opensearch/opensearch_document_index.py
```diff
@@ -702,22 +705,40 @@ def index(
...
+            if doc_id != current_doc_id:
+                if current_chunks:
+                    _flush_chunks(current_chunks)
+                current_doc_id = doc_id
+                current_chunks = [chunk]
+            else:
+                current_chunks.append(chunk)
+
+        if current_chunks:
+            _flush_chunks(current_chunks)
```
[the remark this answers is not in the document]
Danelegend: Bad greptile - there are no bad side-effects if this happens

Danelegend:
@greptile @cubic

backend/onyx/document_index/opensearch/opensearch_document_index.py
```diff
@@ -671,29 +671,32 @@ def index(
...
         document_indexing_results: list[DocumentInsertionRecord] = []
-        # Try to index per-document.
-        for _, chunks in doc_id_to_chunks.items():
+        deleted_doc_ids: set[str] = set()
+        # Buffer chunks per document as they arrive from the iterable.
+        # When the document ID changes flush the buffered chunks.
+        current_doc_id: str | None = None
+        current_chunks: list[DocMetadataAwareIndexChunk] = []
+
+        def _flush_chunks(doc_chunks: list[DocMetadataAwareIndexChunk]) -> None:
```
acaprau, line 692: nit but i would add an assert or value error here if doc chunks has len 0

backend/onyx/document_index/opensearch/opensearch_document_index.py
```diff
@@ -645,10 +645,10 @@ def verify_and_create_index_if_necessary(
 
     def index(
         self,
-        chunks: list[DocMetadataAwareIndexChunk],
-        indexing_metadata: IndexingMetadata,  # noqa: ARG002
+        chunks: Iterable[DocMetadataAwareIndexChunk],
```
acaprau, line 650: so what happens if .index() is called multiple times on groups of chunks that span the same document? as i understand the current logic we're actually doing deletes every invocation which is not good.

Danelegend: If I'm understanding you correctly you're talking about how .index() deletes everything from that document then repopulates? And there is a worry about splitting doc A's chunks over multiple .index calls

We had a discussion earlier where you explained this logic to me about how this is the 'abstraction' and that there is a contract that a document's chunks cannot be spread out. All that's to point out that the ideas behind that were introduced in some other PR. This PR is maintaining what was previously done, just refactoring a little bit

acaprau (approved):

PR #9385, opened 2026-03-16
feat(canvas): add API client, data models, and connector scaffold 1/4

benwu408:
Add CanvasApiClient, data models (CanvasCourse, CanvasPage, CanvasAssignment, CanvasAnnouncement), checkpoint class, and register the Canvas connector source.

## Description

Add the Canvas LMS API client (CanvasApiClient) with HTTPS-only URL validation, safe Link-header pagination with host/scheme verification to prevent bearer token leakage, and comprehensive error mapping (401/403/404/429/5xx → OnyxErrorCode). 

Define Pydantic data models (CanvasCourse, CanvasPage, CanvasAssignment, CanvasAnnouncement) with from_api factory methods, and the CanvasConnectorCheckpoint for staged indexing. 

Includes unit tests for all client methods, error codepaths, and pagination security checks.

## How Has This Been Tested?

Unit tests covering all client methods: URL validation (success + rejection), URL building, header construction, get() success/error paths (all 3 Canvas error shapes, invalid JSON, pagination param skipping), and _parse_next_link (found, not found, empty, multiple rels, host mismatch rejection, non-HTTPS rejection).

## PR Stack

**Stacked PRs on my fork (each targets the previous branch for isolated review):**
1. [#1 — Scaffold: API client, data models, connector registration](https://github.com/benwu408/onyx/pull/1)
2. [#2 — Connector data fetching and conversion](https://github.com/benwu408/onyx/pull/2)
3. [#3 — Checkpoint-based indexing](https://github.com/benwu408/onyx/pull/3)
4. [#4 — Permission sync and frontend registration](https://github.com/benwu408/onyx/pull/4)

**PRs on upstream onyx (all target main):**
1. [#9385 — Scaffold](https://github.com/onyx-dot-app/onyx/pull/9385)
2. [#9386 — Connector data](https://github.com/onyx-dot-app/onyx/pull/9386)
3. [#9387 — Checkpoint](https://github.com/onyx-dot-app/onyx/pull/9387)
4. [#9388 — Perm sync + frontend](https://github.com/onyx-dot-app/onyx/pull/9388)

## Additional Options

- [ ] [Optional] Please cherry-pick this PR to the latest release version.
- [x] [Optional] Override Linear Check

---
## Summary by cubic
Scaffolds the Canvas LMS connector with a secure API client, core models, and a resumable checkpoint. Adds thorough client tests and tightens pagination, URL validation, and error handling.

- **New Features**
[...]

PR #9386, opened 2026-03-16
feat(canvas): Canvas Connector data fetching 2/4

benwu408:
## Description

Add the CanvasConnector class with methods for listing courses, pages, assignments, and announcements from the Canvas API. Convert each resource type into Onyx Document objects with HTML stripping, human-readable due dates, and nullable timestamp handling. Register DocumentSource.CANVAS in the connector registry, add the CE permission stub (access.py) with EE dynamic dispatch via fetch_versioned_implementation, and implement load_credentials / validate_connector_settings with full error mapping. 

Includes unit tests for credentials, document conversion, and settings validation.

## How Has This Been Tested?

Unit tests covering credential loading (success, missing credentials, invalid token), document conversion for all 3 resource types (pages, assignments, announcements — including nullable fields and HTML stripping), and validate_connector_settings (success, expired credential, rate limiting, unexpected errors).

## PR Stack

**Stacked PRs on my fork (each targets the previous branch for isolated review):**
1. [#1 — Scaffold: API client, data models, connector registration](https://github.com/benwu408/onyx/pull/1)
2. [#2 — Connector data fetching and conversion](https://github.com/benwu408/onyx/pull/2)
3. [#3 — Checkpoint-based indexing](https://github.com/benwu408/onyx/pull/3)
4. [#4 — Permission sync and frontend registration](https://github.com/benwu408/onyx/pull/4)

**PRs on upstream onyx (all target main):**
1. [#9385 — Scaffold](https://github.com/onyx-dot-app/onyx/pull/9385)
2. [#9386 — Connector data](https://github.com/onyx-dot-app/onyx/pull/9386)
3. [#9387 — Checkpoint](https://github.com/onyx-dot-app/onyx/pull/9387)
4. [#9388 — Perm sync + frontend](https://github.com/onyx-dot-app/onyx/pull/9388)

### DIFF
--- a/backend/onyx/connectors/canvas/connector.py
+++ b/backend/onyx/connectors/canvas/connector.py
@@ -37,7 +37,6 @@ from onyx.utils.logger import setup_logger
 logger = setup_logger()
 
 
-
 def _handle_canvas_api_error(e: OnyxError) -> NoReturn:
     """Map Canvas API errors to connector framework exceptions."""
     if e.status_code == 401:
@@ -51,6 +50,10 @@ def _handle_canvas_api_error(e: OnyxError) -> NoReturn:
     elif e.status_code == 429:
         raise ConnectorValidationError(
             "Canvas rate-limit exceeded (HTTP 429). Please try again later."
+        )
+    elif e.status_code >= 500:
+        raise UnexpectedValidationError(
+            f"Unexpected Canvas HTTP error (status={e.status_code}): {e}"
         )
     else:
         raise ConnectorValidationError(
@@ -86,9 +89,7 @@ class CanvasPage(BaseModel):
     course_id: int
 
     @classmethod
-    def from_api(
-        cls, payload: dict[str, Any], course_id: int
-    ) -> "CanvasPage":
+    def from_api(cls, payload: dict[str, Any], course_id: int) -> "CanvasPage":
         return cls(
             page_id=payload["page_id"],
             url=payload["url"],
@@ -111,9 +112,7 @@ class CanvasAssignment(BaseModel):
     due_at: str | None = None
 
     @classmethod
-    def from_api(
-        cls, payload: dict[str, Any], course_id: int
-    ) -> "CanvasAssignment":
+    def from_api(cls, payload: dict[str, Any], course_id: int) -> "CanvasAssignment":
         return cls(
             id=payload["id"],
             name=payload["name"],
@@ -135,9 +134,7 @@ class CanvasAnnouncement(BaseModel):
     course_id: int
 
     @classmethod
-    def from_api(
-        cls, payload: dict[str, Any], course_id: int
-    ) -> "CanvasAnnouncement":
+    def from_api(cls, payload: dict[str, Any], course_id: int) -> "CanvasAnnouncement":
         return cls(
             id=payload["id"],
             title=payload["title"],
@@ -299,8 +296,9 @@ class CanvasConnector(
             text_parts.append(body_text)
 
         doc_updated_at = (
-            datetime.fromisoformat(page.updated_at.replace("Z", "+00:00"))
-            .astimezone(timezone.utc)
+            datetime.fromisoformat(page.updated_at.replace("Z", "+00:00")).astimezone(
+                timezone.utc
+            )
             if page.updated_at
             else None
         )
@@ -316,9 +314,7 @@ class CanvasConnector(
         )
         return document
 
-    def _convert_assignment_to_document(
-        self, assignment: CanvasAssignment
-    ) -> Document:
+    def _convert_assignment_to_document(self, assignment: CanvasAssignment) -> Document:
         """Convert a Canvas assignment to a Document."""
         text_parts = [assignment.name]
         desc_text = (
@@ -335,8 +331,9 @@ class CanvasConnector(
             text_parts.append(f"Due: {due_dt.strftime('%B %d, %Y %H:%M UTC')}")
 
         doc_updated_at = (
-            datetime.fromisoformat(assignment.updated_at.replace("Z", "+00:00"))
-            .astimezone(timezone.utc)
+            datetime.fromisoformat(
+                assignment.updated_at.replace("Z", "+00:00")
+            ).astimezone(timezone.utc)
             if assignment.updated_at
             else None
         )
@@ -358,16 +355,15 @@ class CanvasConnector(
         """Convert a Canvas announcement to a Document."""
         text_parts = [announcement.title]
         msg_text = (
-            parse_html_page_basic(announcement.message)
-            if announcement.message
-            else ""
+            parse_html_page_basic(announcement.message) if announcement.message else ""
         )
         if msg_text:
             text_parts.append(msg_text)
 
         doc_updated_at = (
-            datetime.fromisoformat(announcement.posted_at.replace("Z", "+00:00"))
-            .astimezone(timezone.utc)
+            datetime.fromisoformat(
+                announcement.posted_at.replace("Z", "+00:00")
+            ).astimezone(timezone.utc)
             if announcement.posted_at
             else None
         )
@@ -376,8 +372,7 @@ class CanvasConnector(
             doc_id=f"canvas-announcement-{announcement.course_id}-{announcement.id}",
             link=announcement.html_url,
             text="\n\n".join(text_parts),
-            semantic_identifier=announcement.title
-            or f"Announcement {announcement.id}",
+            semantic_identifier=announcement.title or f"Announcement {announcement.id}",
             doc_updated_at=doc_updated_at,
             course_id=announcement.course_id,
             doc_type="announcement",
@@ -385,9 +380,7 @@ class CanvasConnector(
         return document
 
     @override
-    def load_credentials(
-        self, credentials: dict[str, Any]
-    ) -> dict[str, Any] | None:
+    def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None:
         """Load and validate Canvas credentials."""
         access_token = credentials.get("canvas_access_token")
         if not access_token:
@@ -400,9 +393,7 @@ class CanvasConnector(
             )
             client.get("courses", params={"per_page": "1"})
         except ValueError as e:
-            raise ConnectorValidationError(
-                f"Invalid Canvas base URL: {e}"
-            )
+            raise ConnectorValidationError(f"Invalid Canvas base URL: {e}")
         except OnyxError as e:
             _handle_canvas_api_error(e)
 
--- a/backend/tests/unit/onyx/connectors/canvas/test_canvas_connector.py
+++ b/backend/tests/unit/onyx/connectors/canvas/test_canvas_connector.py
@@ -16,9 +16,7 @@ from onyx.connectors.exceptions import CredentialExpiredError
 from onyx.connectors.exceptions import InsufficientPermissionsError
 from onyx.connectors.exceptions import UnexpectedValidationError
 from onyx.connectors.models import ConnectorMissingCredentialError
-from onyx.connectors.models import Document
 from onyx.error_handling.exceptions import OnyxError
-
 
 # ---------------------------------------------------------------------------
 # Helpers
@@ -555,7 +553,9 @@ class TestLoadCredentials:
     def test_load_credentials_insufficient_permissions(
         self, mock_requests: MagicMock
     ) -> None:
-        self._assert_load_credentials_raises(403, InsufficientPermissionsError, mock_requests)
+        self._assert_load_credentials_raises(
+            403, InsufficientPermissionsError, mock_requests
+        )
 
 
 # ---------------------------------------------------------------------------
@@ -621,6 +621,7 @@ class TestDocumentConversion:
         assert doc.source == DocumentSource.CANVAS
         assert doc.semantic_identifier == "Syllabus"
         assert doc.metadata == expected_metadata
+        assert doc.sections[0].link is not None
         assert f"{FAKE_BASE_URL}/courses/1/pages/syllabus" in doc.sections[0].link
         assert doc.doc_updated_at == expected_updated_at
 
@@ -639,6 +640,7 @@ class TestDocumentConversion:
 
         doc = self.connector._convert_page_to_document(page)
         section_text = doc.sections[0].text
+        assert section_text is not None
 
         assert "Empty Page" in section_text
         assert "<p>" not in section_text
@@ -665,6 +667,7 @@ class TestDocumentConversion:
         assert doc.id == expected_id
         assert doc.source == DocumentSource.CANVAS
         assert doc.semantic_identifier == "Homework 1"
+        assert doc.sections[0].text is not None
         assert expected_due_text in doc.sections[0].text
 
     def test_convert_assignment_without_description(self) -> None:
@@ -683,6 +686,7 @@ class TestDocumentConversion:
 
         doc = self.connector._convert_assignment_to_document(assignment)
         section_text = doc.sections[0].text
+        assert section_text is not None
 
         assert "Quiz 1" in section_text
         assert "Due:" not in section_text
@@ -759,9 +763,7 @@ class TestValidateConnectorSettings:
         self._assert_validate_raises(401, CredentialExpiredError, mock_requests)
 
     @patch("onyx.connectors.canvas.client.rl_requests")
-    def test_validate_insufficient_permissions(
-        self, mock_requests: MagicMock
-    ) -> None:
+    def test_validate_insufficient_permissions(self, mock_requests: MagicMock) -> None:
         self._assert_validate_raises(403, InsufficientPermissionsError, mock_requests)
 
     @patch("onyx.connectors.canvas.client.rl_requests")
@@ -776,6 +778,7 @@ class TestValidateConnectorSettings:
 # ---------------------------------------------------------------------------
 # _list_* pagination tests
 # ---------------------------------------------------------------------------
+
 
 class TestListCourses:
     @patch("onyx.connectors.canvas.client.rl_requests")


backend/onyx/connectors/canvas/connector.py
```diff
@@ -0,0 +1,421 @@
...
+                    title=p["title"],
+                    body=p.get("body"),
+                    created_at=p["created_at"],
+                    updated_at=p["updated_at"],
+                    course_id=course_id,
+                )
+                for p in response
+            )
+            if not next_url:
+                break
```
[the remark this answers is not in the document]
benwu408: not true, params are in pagination url, adding again would duplicate

backend/onyx/connectors/canvas/connector.py
```diff
@@ -0,0 +1,413 @@
...
+        try:
+            self._canvas_client.get("courses", params={"per_page": "1"})
+        except CanvasClientRequestFailedError as e:
+            if e.status_code == 401:
+                raise ConnectorMissingCredentialError(
+                    "Invalid Canvas API token"
+                )
+            raise
+
+        return None
```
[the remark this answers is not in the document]
benwu408: load_credentials validates the token works at credential-load time. validate_connector_settings is called independently from the UI settings page. may not always be called together

backend/onyx/connectors/canvas/connector.py
```diff
@@ -0,0 +1,414 @@
...
+    def _get_course_permissions(self, course_id: int) -> ExternalAccess | None:
+        """Get course permissions with caching."""
+        if course_id not in self._course_permissions_cache:
+            self._course_permissions_cache[course_id] = get_course_permissions(
+                canvas_client=self.canvas_client,
+                course_id=course_id,
+            )
+        return self._course_permissions_cache[course_id]
+
+    @retry(tries=3, delay=1, backoff=2)
```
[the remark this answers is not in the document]
benwu408: codebase wide pattern. other connectors use same format of retries regardless of error.

backend/onyx/connectors/canvas/connector.py
```diff
@@ -0,0 +1,414 @@
...
+                    params={
+                        "per_page": "100",
+                        "enrollment_state": "active",
+                    },
+                )
+                first_request = False
+            else:
+                response, next_url = self.canvas_client.get(
+                    "", full_url=next_url
+                )
```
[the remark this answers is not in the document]
benwu408: The first_request flag pattern separates "build URL from endpoint + params" (first request) from "follow pagination link" (subsequent requests), which is the correct semantic distinction. The suggested alternative hardcodes query strings and bypasses the client's URL building, which is more fragile.

backend/onyx/connectors/canvas/connector.py
```diff
@@ -0,0 +1,458 @@
...
+
+    Invariant:
+        If current_course_index is incremented, stage must be reset to
+        "pages" and next_url must be reset to None.
+    """
+
+    course_ids: list[int] = []
+    current_course_index: int = 0
+    stage: CanvasStage = "pages"
+    next_url: str | None = None
```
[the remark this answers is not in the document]
benwu408: no other checkpoint classes have a has_more default

backend/onyx/connectors/canvas/connector.py
```diff
@@ -0,0 +1,491 @@
...
+            )
+
+    @override
+    def load_from_checkpoint(
+        self,
+        start: SecondsSinceUnixEpoch,
+        end: SecondsSinceUnixEpoch,
+        checkpoint: CanvasConnectorCheckpoint,
+    ) -> CheckpointOutput[CanvasConnectorCheckpoint]:
+        raise NotImplementedError("Implemented in follow-up PR")
```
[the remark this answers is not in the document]
benwu408: kept as structure for following PR

backend/onyx/connectors/canvas/connector.py
```diff
@@ -0,0 +1,524 @@
...
+
+    Invariant:
+        If current_course_index is incremented, stage must be reset to
+        "pages" and next_url must be reset to None.
+    """
+
+    course_ids: list[int] = []
+    current_course_index: int = 0
+    stage: CanvasStage = "pages"
+    next_url: str | None = None
```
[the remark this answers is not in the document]
benwu408: not true, that notion has_more is for another class. no other classes have has_more default

backend/onyx/connectors/canvas/connector.py
```diff
@@ -0,0 +1,524 @@
...
+        courses: list[CanvasCourse] = []
+        next_url: str | None = None
+        first_request = True
+        while True:
+            if first_request:
+                response, next_url = self.canvas_client.get(
+                    "courses",
+                    params={
+                        "per_page": "100",
+                        "state[]": "available",
```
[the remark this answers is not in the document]
benwu408: only want to get info from current courses. search would be filled with meaningless data from old irrelevant courses if state wasn't limited to just available.

backend/onyx/connectors/canvas/connector.py
```diff
@@ -0,0 +1,527 @@
...
+
+    @classmethod
+    def from_api(cls, payload: dict[str, Any]) -> "CanvasCourse":
+        return cls(
+            id=payload["id"],
+            name=payload.get("name", ""),
+            course_code=payload.get("course_code", ""),
+            created_at=payload.get("created_at", ""),
+            workflow_state=payload.get("workflow_state", ""),
+        )
```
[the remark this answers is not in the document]
benwu408: Primary identifiers (id, page_id) and required fields (name, html_url, url, title) are guaranteed by the Canvas API for well-formed responses. Using .get() with sentinels would add complexity for a scenario that indicates a
 fundamentally broken API response — which should fail loudly, not silently produce garbage documents with id=0 or name="".

backend/onyx/connectors/canvas/connector.py
```diff
@@ -0,0 +1,527 @@
...
+            self._canvas_client = CanvasApiClient(
+                bearer_token=access_token,
+                canvas_base_url=self.canvas_base_url,
+            )
+            self._canvas_client.get("courses", params={"per_page": "1"})
+        except ValueError as e:
+            raise ConnectorValidationError(
+                f"Invalid Canvas base URL: {e}"
+            )
+        except OnyxError as e:
```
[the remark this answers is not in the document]
benwu408: Not fixing — invalid concern. ConnectorMissingCredentialError inherits from PermissionError, not OnyxError. The except OnyxError block in load_credentials will never catch it. No guard is needed

backend/onyx/connectors/canvas/connector.py
```diff
@@ -0,0 +1,424 @@
...
+
+class CanvasAssignment(BaseModel):
+    id: int
+    name: str
+    description: str | None = None
+    html_url: str
+    course_id: int
+    created_at: str
+    updated_at: str
+    due_at: str | None = None
```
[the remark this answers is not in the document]
benwu408: isnt used directly, all other connectors don't use either but keep for possible future features or debugging.

backend/onyx/connectors/canvas/connector.py
```diff
@@ -0,0 +1,444 @@
...
+                )
+
+            if not response:
+                break
+            pages.extend(
+                CanvasPage.from_api(p, course_id=course_id)
+                for p in response
+            )
+            if not next_url:
+                break
```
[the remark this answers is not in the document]
benwu408: next url link in header alr includes parameters

backend/onyx/connectors/canvas/connector.py
```diff
@@ -0,0 +1,421 @@
...
+                    html_url=a["html_url"],
+                    posted_at=a.get("posted_at"),
+                    course_id=course_id,
+                )
+                for a in response
+            )
+            if not next_url:
+                break
+
+        return announcements
```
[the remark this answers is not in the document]
benwu408: wrong, canvas does return params in next link

backend/onyx/connectors/canvas/connector.py
```diff
@@ -0,0 +1,527 @@
...
+            if e.status_code == 401:
+                raise CredentialExpiredError(
+                    "Canvas API token is invalid or expired (HTTP 401)."
+                )
+            elif e.status_code == 403:
+                raise InsufficientPermissionsError(
+                    "Canvas API token does not have sufficient permissions (HTTP 403)."
+                )
+            elif e.status_code == 429:
+                raise ConnectorValidationError(
```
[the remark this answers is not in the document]
benwu408: no one else does this. unnecessary

backend/onyx/connectors/canvas/connector.py
```diff
@@ -13,6 +45,16 @@ class CanvasCourse(BaseModel):
     created_at: str
     workflow_state: str
 
+    @classmethod
+    def from_api(cls, payload: dict[str, Any]) -> "CanvasCourse":
+        return cls(
+            id=payload["id"],
```
justin-tahara, line 74: Another place where we are implicitly accepting the fact that ID exists.

backend/onyx/connectors/canvas/connector.py
```diff
@@ -23,6 +65,20 @@ class CanvasPage(BaseModel):
...
 
+    @classmethod
+    def from_api(
+        cls, payload: dict[str, Any], course_id: int
+    ) -> "CanvasPage":
+        return cls(
+            page_id=payload["page_id"],
+            url=payload["url"],
+            title=payload["title"],
+            body=payload.get("body"),
```
justin-tahara, line 97: Should have a proper default value like the other ones.

backend/onyx/connectors/canvas/connector.py
```diff
@@ -34,6 +90,21 @@ class CanvasAssignment(BaseModel):
...
     due_at: str | None = None
 
+    @classmethod
+    def from_api(
+        cls, payload: dict[str, Any], course_id: int
+    ) -> "CanvasAssignment":
+        return cls(
+            id=payload["id"],
+            name=payload["name"],
+            description=payload.get("description"),
```
justin-tahara, line 119: Another instance where the class being constructed is not uniform but more importantly should we have a default?

backend/onyx/connectors/canvas/connector.py
```diff
@@ -72,3 +156,377 @@ def advance_course(self) -> None:
...
+class CanvasConnector(
+    CheckpointedConnectorWithPermSync[CanvasConnectorCheckpoint],
+    SlimConnectorWithPermSync,
+):
+    def __init__(
+        self,
+        canvas_base_url: str,
+        batch_size: int = INDEX_BATCH_SIZE,
+    ) -> None:
+        self.canvas_base_url = canvas_base_url.rstrip("/").removesuffix("/api/v1")
```
justin-tahara, line 187: Any reason this is being done here rather than when we setup the Canvas Connector? Should we sanitize the base url before passing it into this class initialization?

backend/onyx/connectors/canvas/connector.py
```diff
@@ -72,3 +156,377 @@ def advance_course(self) -> None:
...
+    @property
+    def canvas_client(self) -> CanvasApiClient:
+        if self._canvas_client is None:
+            raise ConnectorMissingCredentialError("Canvas")
+        return self._canvas_client
+
+    def _get_course_permissions(self, course_id: int) -> ExternalAccess | None:
+        """Get course permissions with caching."""
+        if course_id not in self._course_permissions_cache:
+            self._course_permissions_cache[course_id] = get_course_permissions(
```
justin-tahara, line 201: Should we be checking if this is properly returning something before we set it within the ExternalAccess

justin-tahara: I don't think there is any exception handling in the function you are calling either

backend/onyx/connectors/canvas/connector.py
```diff
@@ -72,3 +156,377 @@ def advance_course(self) -> None:
...
+
+    @retry(tries=3, delay=1, backoff=2)
+    def _list_courses(self) -> list[CanvasCourse]:
+        """Fetch all courses accessible to the authenticated user."""
+        logger.debug("Fetching Canvas courses")
+
+        courses: list[CanvasCourse] = []
+        next_url: str | None = None
+        first_request = True
+        while True:
```
justin-tahara: Is there a way to not have an infinite loop that we exit from?

backend/onyx/connectors/canvas/connector.py
```diff
@@ -72,3 +156,377 @@ def advance_course(self) -> None:
...
+            if first_request:
```
justin-tahara: Just take this out and then have a loop through the next pages.

backend/onyx/connectors/canvas/connector.py
```diff
@@ -72,3 +156,377 @@ def advance_course(self) -> None:
...
+                response, next_url = self.canvas_client.get(
+                    f"courses/{course_id}/pages",
+                    params={
+                        "per_page": "100",
+                        "include[]": "body",
+                        "published": "true",
+                    },
+                )
+                first_request = False
```
justin-tahara: Same pattern here. Probably worth taking this out and then just looping on page 2 and beyond

backend/onyx/connectors/canvas/connector.py
```diff
@@ -72,3 +156,377 @@ def advance_course(self) -> None:
...
+            else:
+                assert next_url is not None
+                response, next_url = self.canvas_client.get(
+                    "", full_url=next_url
+                )
+
+            if not response:
+                break
+            courses.extend(
+                CanvasCourse.from_api(course)
```
justin-tahara: Make a giant helper that does the following: 

Create the first endpoint and starts that process and then the second part of the helper function takes in a different arg like CanvasCourse, CanvasPage and other ones to get the responses for that Canvas resource. That way you can reduce a ton of the code in this file. It should simplify the tests as well.

backend/onyx/connectors/canvas/connector.py
```diff
@@ -72,3 +156,377 @@ def advance_course(self) -> None:
...
+            id=f"canvas-page-{page.course_id}-{page.page_id}",
+            sections=cast(list[TextSection | ImageSection], sections),
+            source=DocumentSource.CANVAS,
+            semantic_identifier=page.title or f"Page {page.page_id}",
+            doc_updated_at=datetime.fromisoformat(
+                page.updated_at.replace("Z", "+00:00")
+            ).astimezone(timezone.utc)
+            if page.updated_at
+            else None,
+            metadata={"course_id": str(page.course_id), "type": "page"},
```
justin-tahara: Let's make this cleaner

justin-tahara: Like we can get these values before the return statement to make this more readable

backend/onyx/connectors/canvas/connector.py
```diff
@@ -72,3 +156,377 @@ def advance_course(self) -> None:
...
+
+        return Document(
+            id=f"canvas-announcement-{announcement.course_id}-{announcement.id}",
+            sections=cast(list[TextSection | ImageSection], sections),
+            source=DocumentSource.CANVAS,
+            semantic_identifier=announcement.title
+            or f"Announcement {announcement.id}",
+            doc_updated_at=doc_updated_at,
+            metadata={"course_id": str(announcement.course_id), "type": "announcement"},
+        )
```
justin-tahara: I think the Document creation can be made into a helper function as well

backend/onyx/connectors/canvas/connector.py
```diff
@@ -72,3 +156,377 @@ def advance_course(self) -> None:
...
+        access_token = credentials.get("canvas_access_token")
+        if not access_token:
+            raise ConnectorMissingCredentialError("Canvas")
+
+        try:
+            client = CanvasApiClient(
+                bearer_token=access_token,
+                canvas_base_url=self.canvas_base_url,
+            )
+            client.get("courses", params={"per_page": "1"})
```
justin-tahara, line 394: Is this your attempt to check if the the permission works? What if there are no courses? is that a possibility?

backend/onyx/connectors/canvas/connector.py
```diff
@@ -72,3 +156,377 @@ def advance_course(self) -> None:
...
+                )
+
+        self._canvas_client = client
+        return None
+
+    @override
+    def validate_connector_settings(self) -> None:
+        """Validate Canvas connector settings by testing API access."""
+        try:
+            self.canvas_client.get("courses", params={"per_page": "1"})
```
justin-tahara, line 407: You have the same check as the other one. Is this enough? Is this equivalent to sending a chat message to the OpenAI chat endpoint?

backend/onyx/connectors/canvas/connector.py
```diff
@@ -72,3 +156,377 @@ def advance_course(self) -> None:
...
+                )
+            elif e.status_code == 429:
+                raise ConnectorValidationError(
+                    "Validation failed due to Canvas rate-limits being exceeded (HTTP 429). "
+                    "Please try again later."
+                )
+            else:
+                raise UnexpectedValidationError(
+                    f"Unexpected Canvas HTTP error (status={e.status_code}): {e}"
+                )
```
justin-tahara: Let's simplify this. I think it should be put in a helper function

backend/tests/unit/onyx/connectors/canvas/test_canvas_connector.py
```diff
@@ -379,3 +460,460 @@ def test_rejects_non_https_link(self) -> None:
...
+        self, mock_requests: MagicMock
+    ) -> None:
+        success_resp = _mock_response(json_data=[_mock_course()])
+        fail_resp = _mock_response(403, {})
+        mock_requests.get.side_effect = [success_resp, fail_resp]
+        connector = CanvasConnector(canvas_base_url=FAKE_BASE_URL)
+        connector.load_credentials({"canvas_access_token": FAKE_TOKEN})
+
+        with pytest.raises(InsufficientPermissionsError):
+            connector.validate_connector_settings()
```
justin-tahara: I think you should be able to make a helper function where you pass in the error code and the error that should be raised and you can reduce the lines here

backend/onyx/connectors/canvas/connector.py
benwu408: Ok so for the from api methods, the ones that use [] to get it are required fields, and then the ones that use get() are optional. but I guess when I was making the classes I made some optional and some not, so the ones that are just get() with no default are ones that the class sets up as str | none, and the rest the default gets set to "" cause the class can't handle none. but that was really dumb so I made all the optional fields str | none and just used get() so that if doesn't exist it'll just return none and the class can handle that now. the ones that are [] are required, so we'll get errors if for example we try to get data from a document with no title which should never happen.

backend/onyx/connectors/canvas/connector.py
benwu408: see comment on line 51

backend/onyx/connectors/canvas/connector.py
benwu408: see reply on line 51

backend/onyx/connectors/canvas/connector.py
benwu408: All connector classes get initialized with the values that are provided by admins when setting up a connector in the frontend. this is just normalizing the url given by frontend. No step in-between passing url into class init.

backend/onyx/connectors/canvas/connector.py
benwu408: For first question if ur asking if it's returning the right things, in get course permissions it returns none if not enterprise edition, so thats why none is an option. if it is EE, we build the external access, which validates while its built since its a pydantic model. 

for the second question, get course permissions is in PR2 because it is part of a perm synced connector, but all it really does is call get course permissions, which is actually another wrapper that checks if EE or CE, and then if EE it uses the real perm sync logic that is built in PR4. that does include exception handling and returns an empty external access on failure and logs errors.

backend/onyx/connectors/canvas/connector.py
benwu408: added a paginate function in client.py that takes in permissions, endpoint, and then returns it paginated for all the list ___ functions and the get course permission function!

backend/onyx/connectors/canvas/connector.py
benwu408: see reply on 199

backend/onyx/connectors/canvas/connector.py
benwu408: see comment on 199

backend/onyx/connectors/canvas/connector.py
benwu408: checks if link, token are correct and canvas is reachable. if no courses it returns empty list, still 200 response.

backend/onyx/connectors/canvas/connector.py
benwu408: load credentials is for setting up the client when canvas connector is initialized. validate connector settings checks again later, in situations like if a token was revoked or something. verifies canvas is reachable with link and token. more so equivalent to openai /models endpoint that lists models. lightest possible call, will always succeed if link and token are valid, no weird edge cases.

backend/onyx/connectors/canvas/connector.py
benwu408: done

backend/tests/unit/onyx/connectors/canvas/test_canvas_connector.py
benwu408: done

justin-tahara (approved):

PR #9421, opened 2026-03-17
chore: Scripts for search quality eval

yuhongsun96:
## Description
Add two benchmark utility scripts + a citation stripping fix

## How Has This Been Tested?
Ran it locally via API

## Additional Options

---
## Summary by cubic
Adds two utility scripts for search quality benchmarking and bulk file ingestion, makes citation stripping robust for all markdown `[[n]](...)` patterns, and removes legacy agentic citation handling from streaming.

- **New Features**
  - `backend/scripts/run_industryrag_bench_questions.py`: Runs JSONL questions against `/api/chat/send-chat-message` with forced internal search; supports retries/parallelism; writes JSONL answers with document IDs.
  - `backend/scripts/upload_files_as_connectors.py`: Uploads directory files as individual file connectors, creates credentials, links them, and triggers indexing.

- **Bug Fixes**
  - Improved `remove_answer_citations` to strip all markdown `[[#]](…)` citations, including empty links and URLs with parentheses, while preserving non-citation links; added unit tests.

<sup>Written for commit d48e7f368e6b831a407148355d6fd7a1fad78a34. Summary will update on new commits.</sup>


backend/onyx/server/query_and_chat/streaming_utils.py
```diff
@@ -26,37 +24,10 @@
 _CANNOT_SHOW_STEP_RESULTS_STR = "[Cannot display step results]"
 
 
-def _adjust_message_text_for_agent_search_results(
```
yuhongsun96, line 29: dead code

backend/onyx/chat/process_message.py
```diff
@@ -1047,7 +1047,7 @@ def llm_loop_completion_handle(
 
 
 def remove_answer_citations(answer: str) -> str:
-    pattern = r"\s*\[\[\d+\]\]\(http[s]?://[^\s]+\)"
+    pattern = r"\s*\[\[\d+\]\]\([^)]*\)"
```
evan-onyx: why this change? the fact that we truncate on right paren does seem like it could cause an issue for urls with right parens yah?

yuhongsun96: File connectors don't have a URL, so they don't get stripped

PR #9428, opened 2026-03-17
fix(opensearch): Fix env var mismatch issue with configuring subquery results; set default to 100

acaprau:
## Description
Fixes an issue where the env var name was not what we expected to configure for the value of `k` / number of results from each hybrid subquery.

Also sets the default value to 100, which we have to tolerate for now as we work on query performance.

## How Has This Been Tested?
Existing automated tests cover this case well. Also in my manual benchmarking I saw a significant improvement improvement by reducing this value from 750 to 100.

## Additional Options

---
## Summary by cubic
Fixes the env var used to configure hybrid subquery result count in OpenSearch and lowers the default from 750 to 100 to improve query performance.

- **Bug Fixes**
  - Use `DEFAULT_NUM_HYBRID_SUBQUERY_CANDIDATES` to set candidates per subquery; removed `OPENSEARCH_OVERRIDE_DEFAULT_NUM_HYBRID_SEARCH_CANDIDATES`.
  - `EF_SEARCH` now mirrors `DEFAULT_NUM_HYBRID_SUBQUERY_CANDIDATES`; updated all references.

- **Migration**
  - If you override this value, set `DEFAULT_NUM_HYBRID_SUBQUERY_CANDIDATES` (int) in the environment.
  - Remove any `OPENSEARCH_DEFAULT_NUM_HYBRID_SEARCH_CANDIDATES` env var; it’s no longer read.

<sup>Written for commit 740d65ca4499d763480f6c78bcf56e465a656536. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/document_index/opensearch/constants.py
+++ b/backend/onyx/document_index/opensearch/constants.py
@@ -37,8 +37,10 @@ M = 32  # Set relatively high for better accuracy.
 # we have a much higher chance of all 10 of the final desired docs showing up
 # and getting scored. In worse situations, the final 10 docs don't even show up
 # as the final 10 (worse than just a miss at the reranking step).
-DEFAULT_NUM_HYBRID_SUBSEARCH_CANDIDATES = int(
-    os.environ.get("DEFAULT_NUM_HYBRID_SUBSEARCH_CANDIDATES", 100)
+# Defaults to 100 for now. Initially this defaulted to 750 but we were seeing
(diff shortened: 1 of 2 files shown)


acaprau:
force merging, the only failing test is `test_update_single_can_clear_user_projects_and_personas` which cannot be failing due to this change, and which i therefore suspect is flaky and i will disable in another pr and look into.

PR #9454, opened 2026-03-18
fix(celery): add dedup guardrails to user file delete queue

justin-tahara:
## Description

The delete beat (`check_for_user_file_delete`) was enqueuing tasks with no dedup guard, no task expiry, and no queue depth backpressure. When files got stuck in DELETING status, the beat re-enqueued them every 20s indefinitely, leading to unbounded queue growth (hundreds of thousands of phantom tasks observed in production).

This PR adds the same three protections that already exist on the processing and project-sync paths:

1. **Queue depth backpressure** -- `USER_FILE_DELETE_MAX_QUEUE_DEPTH=500` stops the beat from enqueuing more tasks when workers are falling behind
2. **Per-file Redis guard key** -- `NX+EX` key with TTL = `CELERY_USER_FILE_DELETE_TASK_EXPIRES` prevents re-enqueuing a file that already has a live task in the queue
3. **Task expiry** -- `expires=CELERY_USER_FILE_DELETE_TASK_EXPIRES` on every `send_task()` call so stale tasks are automatically discarded

Also clears the guard key in `delete_user_file_impl` when the worker picks up the task, matching the processing path pattern.

### Changes
- `constants.py`: Added `CELERY_USER_FILE_DELETE_TASK_EXPIRES`, `USER_FILE_DELETE_MAX_QUEUE_DEPTH`, `USER_FILE_DELETE_QUEUED_PREFIX`
- `tasks.py`: Added `_user_file_delete_queued_key()` helper, updated `check_for_user_file_delete()` with all 3 protections, updated `delete_user_file_impl()` to clear guard key on pickup

## How Has This Been Tested?

Added external dependency unit tests in `test_user_file_delete_queue.py` mirroring the existing `test_user_file_processing_queue.py` structure:
- `TestDeleteQueueDepthBackpressure` -- verifies no tasks enqueued when queue exceeds limit
- `TestDeletePerFileGuardKey` -- verifies guarded files are skipped and guard key is set after enqueue
- `TestDeleteTaskExpiry` -- verifies `send_task` is called with correct `expires` value
- `TestDeleteWorkerClearsGuardKey` -- verifies guard key is cleared when worker picks up task
## Additional Options

### DIFF
--- a/backend/onyx/background/celery/tasks/user_file_processing/tasks.py
+++ b/backend/onyx/background/celery/tasks/user_file_processing/tasks.py
@@ -589,6 +589,8 @@ def check_for_user_file_delete(self: Task, *, tenant_id: str) -> None:
     skipped_guard = 0
     try:
         # --- Protection 1: queue depth backpressure ---
+        # NOTE: must use the broker's Redis client (not redis_client) because
+        # Celery queues live on a separate Redis DB with CELERY_SEPARATOR keys.
         r_celery: Redis = self.app.broker_connection().channel().client  # type: ignore
         queue_len = celery_get_queue_length(OnyxCeleryQueues.USER_FILE_DELETE, r_celery)
         if queue_len > USER_FILE_DELETE_MAX_QUEUE_DEPTH:
@@ -643,7 +645,7 @@ def check_for_user_file_delete(self: Task, *, tenant_id: str) -> None:
             lock.release()
 
     task_logger.info(
-        f"check_for_user_file_delete - Enqueued {enqueued} skipped_guard={skipped_guard} tasks for tenant={tenant_id}"
+        f"check_for_user_file_delete - Enqueued {enqueued} tasks, skipped_guard={skipped_guard} for tenant={tenant_id}"
     )
     return None
 


backend/onyx/background/celery/tasks/user_file_processing/tasks.py
```diff
@@ -555,8 +584,21 @@ def check_for_user_file_delete(self: Task, *, tenant_id: str) -> None:
     )
     if not lock.acquire(blocking=False):
         return None
+
     enqueued = 0
+    skipped_guard = 0
     try:
+        # --- Protection 1: queue depth backpressure ---
+        r_celery: Redis = self.app.broker_connection().channel().client  # type: ignore
```
evan-onyx, line 594: why isn't this using redis_client?

acaprau: +!

backend/onyx/background/celery/tasks/user_file_processing/tasks.py
```diff
@@ -568,23 +610,40 @@ def check_for_user_file_delete(self: Task, *, tenant_id: str) -> None:
...
-            f"check_for_user_file_delete - Error enqueuing deletes - {e.__class__.__name__}"
-        )
-        return None
     finally:
         if lock.owned():
             lock.release()
+
     task_logger.info(
-        f"check_for_user_file_delete - Enqueued {enqueued} tasks for tenant={tenant_id}"
+        f"check_for_user_file_delete - Enqueued {enqueued} skipped_guard={skipped_guard} tasks for tenant={tenant_id}"
```
acaprau: lol nit again but `check_for_user_file_delete` is redundant

acaprau (approved):

PR #9463, opened 2026-03-18
fix(fe): make indexing attempt error rows click to show trace

jmelahman:
## Description

Removes the `View Full Trace` button for errored indexing attempts and instead makes the entire table row a clickable to view the trace. This is similar behavior to the `Query History` table.

## How Has This Been Tested?

**before**
20260318_13h04m53s_grim

**after**
20260318_13h04m05s_grim

## Additional Options

---
## Summary by cubic
Make indexing attempt rows with a full exception trace clickable to open the trace, replacing the “View Full Trace” link. Add an invisible full-row button (with aria-label) and hover/cursor styles on those rows, and center pagination to match Query History.

<sup>Written for commit 62cc5d742be0de497ba2a56c3a8421529e9f8566. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/app/admin/connector/[ccPairId]/IndexAttemptsTable.tsx
+++ b/web/src/app/admin/connector/[ccPairId]/IndexAttemptsTable.tsx
@@ -169,6 +169,7 @@ export function IndexAttemptsTable({
                 <td className="w-0 p-0">
                   {indexAttempt.full_exception_trace && (
                     <button
+                      type="button"
                       aria-label="View full trace"
                       onClick={() =>
                         setIndexAttemptTracePopupId(indexAttempt.id)


raunakab:
I presume this page will get updated in the future, but if you want, you could transition this over to using @Subash-Mohan's new `Table` component.

Not a requirement for this PR (would rather have it be a follow-up, if at all), so approved.

web/src/app/admin/connector/[ccPairId]/IndexAttemptsTable.tsx
```diff
@@ -159,33 +166,31 @@ export function IndexAttemptsTable({
...
+                  <td className="w-0 p-0">
+                    <button
+                      aria-label="View full trace"
+                      onClick={() =>
+                        setIndexAttemptTracePopupId(indexAttempt.id)
+                      }
+                      className="absolute w-full h-full left-0 top-0"
+                    />
+                  </td>
+                )}
```
Comment (jmelahman) [inline web/src/app/admin/connector/[ccPairId]/IndexAttemptsTable.tsx]:
```suggestion
                <td className="w-0 p-0">
                  {indexAttempt.full_exception_trace && (
                    <button
                      aria-label="View full trace"
                      onClick={() =>
                        setIndexAttemptTracePopupId(indexAttempt.id)
                      }
                      className="absolute w-full h-full left-0 top-0"
                    /> 
                  )}
                </td>
```

PR #9464, opened 2026-03-18
feat(fe): "Full Exception Trace" modal uses CodePreview rendering

jmelahman:
## How Has This Been Tested?

**before**
20260318_13h35m02s_grim

**after**
20260318_13h35m21s_grim

## Additional Options

---
## Summary by cubic
Switches the “Full Exception Trace” modal to the unified `CodePreview` renderer for syntax-highlighted, scrollable traces. Extracts a reusable `FloatingFooter` with a copy action and removes the old modal.

- **New Features**
  - Renders the trace with `CodePreview` (normalized, highlighted; default language "python").
  - Floating footer with `CopyButton` for quick copying.

- **Refactors**
  - Replaced the old modal with `web/src/sections/modals/PreviewModal/ExceptionTraceModal.tsx` and updated the import in `IndexAttemptsTable`.
  - Extracted `FloatingFooter` and used it in `PreviewModal` and the exception modal; supports a code-themed gradient via `codeBackground`.

<sup>Written for commit 82899abfaa88903d785a84926f8f3e510071753c. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/sections/modals/PreviewModal/FloatingFooter.tsx
+++ b/web/src/sections/modals/PreviewModal/FloatingFooter.tsx
@@ -29,9 +29,11 @@ export default function FloatingFooter({
       <div className="pointer-events-auto">{left}</div>
 
       {/* Right slot */}
-      <div className="pointer-events-auto rounded-12 bg-background-tint-00 p-1 shadow-lg">
-        {right}
-      </div>
+      {right ? (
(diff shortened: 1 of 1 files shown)


jmelahman:
> Nice! One thing; maybe for rendering the code block, we could use `refresh-components/Code.tsx`? That comes with a default "copy-icon", I believe.

I think it's fine for now to keep it consistent with the existing preview modals

PR #9468, opened 2026-03-18
feat(indexing): Batch chunks during doc processing

Danelegend:
## Description
Introduces batching chunks during doc processing such that we don't have more than the MAX number of chunks when they are embedded. This includes the downstream variants of the embedding model.

Profiling pre batching vs post batching
Screenshot 2026-03-18 at 3 38 14 PM
Screenshot 2026-03-18 at 3 29 52 PM

Note in the pre-batching era, the massive rise during embedding. In the post-batching profiling, we no longer see this, and the memory is more consistant.

## How Has This Been Tested?
Manual

## Additional Options

---
## Summary by cubic
Embeds chunks in disk-backed batches and streams them to each document index to reduce memory and avoid re-embedding. Adds `embed_and_stream` with `ChunkBatchStore`, enforces per-index completeness, and guarantees all-or-nothing per document with precise success/failure accounting.

- **New Features**
  - Batch embedding with `MAX_CHUNKS_PER_DOC_BATCH`; spill to temp files via `ChunkBatchStore` and stream via `embed_and_stream`. Scrubs failed docs across all batches to keep documents all-or-nothing. Returns `ChunkEmbeddingResult.successful_chunk_ids` and tracks `connector_failures`. Enrichment runs only for embedded docs.
  - Vector DB writes now consume an enriched generator over `store.stream()`, enabling multi-pass streaming per index. Adds `_verify_indexing_completeness` to include embedding and write failures per index.

- **Migration**
  - Replace `build_metadata_aware_chunks(...)` with `prepare_enrichment(context, tenant_id, chunks)` and stream-enrich with `enricher.enrich_chunk(chunk, 1.0)`; update `IndexingBatchAdapter.post_index(...)` to accept the `enrichment` context.
  - Update `write_chunks_to_vector_db_with_backoff(...)` call sites to pass `make_chunks` (a generator) instead of a chunk list.

<sup>Written for commit 7604fa2d7691b5ad63ec2a883bd34cf73524f48b. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/indexing/chunk_batch_store.py
+++ b/backend/onyx/indexing/chunk_batch_store.py
@@ -1,8 +1,8 @@
 import pickle
+import shutil
 import tempfile
 from collections.abc import Iterator
 from pathlib import Path
-from types import TracebackType
 
 from onyx.indexing.models import IndexChunk
 
@@ -24,23 +24,18 @@ class ChunkBatchStore:
     _EXT = ".pkl"
 
     def __init__(self) -> None:
-        self._tmpdir_ctx = tempfile.TemporaryDirectory(prefix="onyx_embeddings_")
         self._tmpdir: Path | None = None
 
     # -- context manager -----------------------------------------------------
 
     def __enter__(self) -> "ChunkBatchStore":
-        self._tmpdir = Path(self._tmpdir_ctx.__enter__())
+        self._tmpdir = Path(tempfile.mkdtemp(prefix="onyx_embeddings_"))
         return self
 
-    def __exit__(
-        self,
-        exc_type: type[BaseException] | None,
-        exc_val: BaseException | None,
-        exc_tb: TracebackType | None,
-    ) -> None:
-        self._tmpdir_ctx.__exit__(exc_type, exc_val, exc_tb)
-        self._tmpdir = None
+    def __exit__(self, *_exc: object) -> None:
+        if self._tmpdir is not None:
+            shutil.rmtree(self._tmpdir, ignore_errors=True)
+            self._tmpdir = None
 
     @property
     def _dir(self) -> Path:


backend/onyx/indexing/indexing_pipeline.py
```diff
@@ -149,6 +164,105 @@ def _upsert_documents_in_db(
...
+            if c.source_document.id not in failed_doc_ids
+        )
+
+        # Spill embeddings to disk
+        batch_file = tmpdir / f"batch_{batch_idx}.pkl"
+        with open(batch_file, "wb") as f:
+            pickle.dump(chunks_with_embeddings, f)
+
+        # Free memory
+        del chunks_with_embeddings
```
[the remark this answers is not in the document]
Danelegend: Resolved

backend/onyx/indexing/indexing_pipeline.py
```diff
@@ -151,6 +163,164 @@ def _upsert_documents_in_db(
...
+    chunks: list[IndexChunk],
+    batch_file: Path,
+) -> None:
+    """Pickle embedded chunks to disk and free memory."""
+    with open(batch_file, "wb") as f:
+        pickle.dump(chunks, f)
+
+
+def _scrub_failed_docs_from_disk(
+    tmpdir: Path,
```
evan-onyx: nit: I'd prefer like chunk_batch_tmpdir to explain what the directory is used for

backend/onyx/indexing/indexing_pipeline.py
```diff
@@ -151,6 +163,164 @@ def _upsert_documents_in_db(
...
+        for batch_file in sorted(
+            self._tmpdir.glob("batch_*.pkl"),
+            key=lambda p: int(p.stem.removeprefix("batch_")),
+        ):
+            with open(batch_file, "rb") as f:
+                batch: list[IndexChunk] = pickle.load(f)
+            yield from batch
+
+
+def _embed_chunks_to_disk(
```
evan-onyx: why are we sending them to disk?

backend/onyx/indexing/indexing_pipeline.py
```diff
@@ -151,6 +163,164 @@ def _upsert_documents_in_db(
...
+        tmpdir.glob("batch_*.pkl"),
+        key=lambda p: int(p.stem.removeprefix("batch_")),
+    ):
+        with open(batch_file, "rb") as f:
+            batch_chunks: list[IndexChunk] = pickle.load(f)
+        cleaned = [
+            c for c in batch_chunks if c.source_document.id not in failed_doc_ids
+        ]
+        if len(cleaned) != len(batch_chunks):
+            with open(batch_file, "wb") as f:
```
evan-onyx: even though it's pretty simple, I think it would be nice to have load/store functions to abstract away what's happening here. it's kind of sus that we're storing on disk so if we need to go to filestore in the future it would make it easy

backend/onyx/indexing/indexing_pipeline.py
```diff
@@ -723,127 +916,99 @@ def index_doc_batch(
...
-            chunk.to_short_descriptor() for chunk in metadata_aware_chunks
-        ]
-        short_descriptor_log = str(short_descriptor_list)[:1024]
-        logger.debug(f"Indexing the following chunks: {short_descriptor_log}")
-
-        primary_doc_idx_insertion_records: list[DocumentInsertionRecord] | None = None
-        primary_doc_idx_vector_db_write_failures: list[ConnectorFailure] | None = None
+        # Acquires a lock on the documents so that no other process can modify
+        # them.  Not needed until here, since this is when the actual race
+        # condition with Vespa can occur.
```
evan-onyx: nit: Opensearch or just "vector db"

backend/onyx/indexing/indexing_pipeline.py
Danelegend: Resolved via sync.
  - We want to embed first then upsert to vector db, but can't hold all the embeddings in memory. So temp directory on disk is best place to go

PR #9481, opened 2026-03-19
chore(fe): opal button supports `revealOnHover`

jmelahman:
## Description

Adds support to the opal button for revealing (going from 0 to 100 opacity) on hover and tabbing.

Might help reduce the need to support an `iconClassName` prop, etc.

## How Has This Been Tested?

**before**
20260318_21h19m20s_grim

**after**
20260318_21h18m45s_grim

## Additional Options

---
## Summary by cubic
Adds reveal-on-hover support to `@opal` `Interactive` and `Button`, so controls fade in on row hover and keyboard focus. Updates the Admin Users page to reveal the Edit button without custom classes or layout hacks.

- **New Features**
  - `revealOnHover` prop on `@opal/components` `Button` (via `Interactive`): starts at opacity 0; fades to 1 on `[data-reveal-group]:hover` or `:focus-visible`.
  - CSS for `[data-reveal-group]` and `[data-revealed]` with 150ms transitions for opacity, background color, `--interactive-foreground`, and `--interactive-foreground-icon`.
  - Admin UsersPage `GroupsCell`: switches to `Button` with `revealOnHover`, adds `data-reveal-group` on the row, and replaces absolute positioning/padding with inline layout using `justify-between`.

<sup>Written for commit a5fe1d54865751c7b8e4aa3ef76bd95dbbcebd21. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/refresh-pages/admin/UsersPage/GroupsCell.tsx
+++ b/web/src/refresh-pages/admin/UsersPage/GroupsCell.tsx
@@ -146,7 +146,7 @@ export default function GroupsCell({
         {groups.length === 0 ? (
           <div
             ref={containerRef}
-            className="flex items-center gap-1 overflow-hidden flex-nowrap min-w-0 pr-7"
+            className="flex items-center gap-1 overflow-hidden flex-nowrap min-w-0"
           >
             <Text as="span" secondaryBody text03>
               —
@@ -163,7 +163,7 @@ export default function GroupsCell({
           >
             <div
               ref={containerRef}
-              className="flex items-center gap-1 overflow-hidden flex-nowrap min-w-0 pr-7"
+              className="flex items-center gap-1 overflow-hidden flex-nowrap min-w-0"
             >
               {tagsContent}
             </div>


raunakab:
Hm, this pattern of "needing an item that disappears on hover" is such a common pattern, I've created a specific component for it inside of `opal/core`, `Hoverable`.

The general API usage is as this:

```tsx
import { Hoverable } from "@opal/core";

<Hoverable.Item variant="opacity-on-hover">
  <MyComponent />
</Hoverable.Item>
```

This API also supports hover on parental elements as well:

```tsx
import { Hoverable } from "@opal/core";

<Hoverable.Root group="container-for-my-component">
  <div id="container">
    <OtherStuff />

    <Hoverable.Item group="container-for-my-component" variant="opacity-on-hover">
      <MyComponent />
    </Hoverable.Item>
  </div>
</Hoverable.Item>
```

With the above, hovering *anywhere* within the container `div` will cause the `MyComponent` item to appear.

Specifically, the nice thing about this API is that it can support *anything* that needs "opacity-100 on hover". `Text`s, `div`s, custom components, you name it. So ideally, we use that API instead.

It is possible that we keep `revealOnHover` and have it internally invoke this component; but I think we should regardless default to this component for use-cases like this.

PR #9484, opened 2026-03-19
fix(code interpreter): Caching files

justin-tahara:
## Description

Code interpreter or any session with multiple cycles will continue to stage the file and the embedding cpu overhead can cause API pods to get into a locked state where they are not necessarily unhealthy but are not able to accept more traffic. 

This specifically helps prevent those issues to ensure that we used the cached version of the file if it exists beforehand

## How Has This Been Tested?

Added unit tests

## Additional Options

---
## Summary by cubic
Cache Code Interpreter uploads in `PythonTool` per session to reuse file IDs across iterations, cutting duplicate uploads and CPU in multi-cycle runs. Helps prevent API pods from stalling under heavy embedding/upload load.

- **Bug Fixes**
  - Added a per-session cache keyed by (filename + SHA-256 content hash) -> CI `file_id`; identical bytes with different filenames upload separately.
  - Stopped deleting staged input files so cached IDs remain valid; rely on Code Interpreter TTL for cleanup. Updated tests to expect no DELETE for staged inputs and added unit tests for cache hits, content changes, same-bytes/different-filename, multiple files, per-instance isolation, and failed uploads not cached.

<sup>Written for commit 51f15dc7f940de350a926019c3d469eb8a993a1d. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/external_dependency_unit/tools/test_python_tool.py
+++ b/backend/tests/external_dependency_unit/tools/test_python_tool.py
@@ -1219,15 +1219,16 @@ def test_code_interpreter_receives_chat_files(
         finally:
             ci_mod.CodeInterpreterClient.__init__.__defaults__ = original_defaults
 
-    # Verify: file uploaded, code executed via streaming, staged file cleaned up
+    # Verify: file uploaded and code executed via streaming.
     assert len(mock_ci_server.get_requests(method="POST", path="/v1/files")) == 1
     assert (
         len(mock_ci_server.get_requests(method="POST", path="/v1/execute/stream")) == 1
     )
 
-    delete_requests = mock_ci_server.get_requests(method="DELETE")
-    assert len(delete_requests) == 1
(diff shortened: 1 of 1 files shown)


backend/onyx/tools/tool_implementations/python/python_tool.py
```diff
@@ -299,14 +313,10 @@ def run(
                             f"Failed to delete Code Interpreter generated file {ci_file_id}: {e}"
                         )
 
-                # Cleanup staged input files
-                for file_mapping in files_to_stage:
-                    try:
-                        client.delete_file(file_mapping["file_id"])
```
Danelegend, line 305: Probs worth deleting on instance destruction (__del__ or weakref.finalise)

justin-tahara: Decision here is to let the Code Interpreter TTL handle the clean up

PR #9488, opened 2026-03-19
chore(fe): reduce padding on elements at the bottom of modal headers

jmelahman:
## Description

We want the search icon to be aligned with the main header icon rather than the input target box/border, so allow content at the bottom of the header with less padding to achieve this.

Part of https://linear.app/onyx-app/issue/ENG-3635/memory-popup

## How Has This Been Tested?

20260319_11h00m59s_grim
20260319_11h01m24s_grim
20260319_11h01m33s_grim

## Additional Options

---
## Summary by cubic
Reduced padding in `Modal.Header` so bottom content aligns with the header icon and spacing is tighter. The title row keeps its visual padding; bottom content uses the reduced outer padding.

- **Refactors**
  - Set outer padding to 0.5 and inner header row padding to 0.5; bottom content (`children`) sits outside the padded row and inherits only the reduced padding.

<sup>Written for commit 89755ab9d4e7bdff899fbee3ae89db5c4ca4b1a0. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/refresh-components/Modal.tsx
+++ b/web/src/refresh-components/Modal.tsx
@@ -458,40 +458,38 @@ const ModalHeader = React.forwardRef<HTMLDivElement, ModalHeaderProps>(
         height="fit"
         {...props}
       >
-        <Section padding={0.5} alignItems="start">
-          <Section
-            flexDirection="row"
-            justifyContent="between"
-            alignItems="start"
-            gap={0}
-            padding={0}
-          >
-            <div className="relative w-full">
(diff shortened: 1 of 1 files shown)


raunakab:
Discussed some more offline. `bottomContent` actually renders "additional content" correctly. `children`'s padding is incorrect, as per the mocks.

The addition of `bottomContent` is a strict benefit. @jmelahman, since `children` is incorrect, we should consider (in a follow-up PR, if needed) deleting `children` and migrating everything to `bottomContent`.

PR #9491, opened 2026-03-19
feat(admin): inline group renaming

nmgarza5:
## Description

Add inline rename to group cards on the admin Groups page. Custom groups show an edit button on hover; clicking it swaps the title into an auto-sizing input for inline editing. Built-in groups (Admin, Basic) are not editable.

**Backend:** New `PATCH /admin/user-group/rename` endpoint with dedicated `rename_user_group` DB function and `UserGroupRename` Pydantic model. Uses `OnyxError` for duplicate name and not-found cases.

**Frontend:** `ContentMd` restructured to wrap icon + title in a `.opal-content-md-header` div with a transparent border that becomes visible during editing. Description moved outside the header as a sibling with `descriptionIndent` for alignment. `GroupCard` handles rename self-contained with SWR revalidation and toast feedback.

**Linear:** [ENG-3843](https://linear.app/onyx-app/issue/ENG-3843/group-card-hover-interactions-inline-rename)

## How Has This Been Tested?

- Unit tests for `rename_user_group`: happy path, not found, syncing raises
- Manual QA: inline rename flow (click → edit → Enter/Escape/blur), toast success/error, built-in groups non-editable

## Additional Options

### DIFF
--- a/backend/ee/onyx/db/user_group.py
+++ b/backend/ee/onyx/db/user_group.py
@@ -41,8 +41,6 @@ from onyx.utils.logger import setup_logger
 
 logger = setup_logger()
 
-BUILT_IN_GROUP_NAMES = {"Admin", "Basic"}
-
 
 def _cleanup_user__user_group_relationships__no_commit(
     db_session: Session,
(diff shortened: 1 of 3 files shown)


nmgarza5:
also verified the language models page still renders correctly

backend/ee/onyx/db/user_group.py
```diff
@@ -800,6 +802,30 @@ def update_user_group(
...
+    db_session: Session,
+    user_group_id: int,
+    new_name: str,
+) -> UserGroup:
+    stmt = select(UserGroup).where(UserGroup.id == user_group_id)
+    db_user_group = db_session.scalar(stmt)
+    if db_user_group is None:
+        raise ValueError(f"UserGroup with id '{user_group_id}' not found")
+
+    if db_user_group.name in BUILT_IN_GROUP_NAMES:
```
Subash-Mohan: For now, we should avoid special handling of default groups here. Handling it by group name feels brittle, so I’ll add the logic when implementing default groups.

backend/ee/onyx/db/user_group.py
```diff
@@ -800,6 +802,30 @@ def update_user_group(
     return db_user_group
 
 
+def rename_user_group(
+    db_session: Session,
+    user_group_id: int,
+    new_name: str,
+) -> UserGroup:
+    stmt = select(UserGroup).where(UserGroup.id == user_group_id)
+    db_user_group = db_session.scalar(stmt)
```
Subash-Mohan, line 809: I think once the rename is done, we should also trigger the vector DB sync, because the CC pairs related to the groups have the group name in their documents. If we change the group name, those documents may be missed.

nmgarza5: ah yes you are correct. syncing to vespa takes a long time so i figured this would be ok. I will fix and we will just pay the cost of waiting

backend/ee/onyx/db/user_group.py
nmgarza5: ok sounds good

PR #9502, opened 2026-03-19
fix(fe): editing an LLM provider uses the global default model

jmelahman:
## Description

Currently, all of my Ollama models are configured incorrectly in Onyx -- their context windows are too big and fail to load into memory -- so the only model that works is the `qwen3:0.6b` model.

Currently, the first model is always made default when editing an LLM provider. When I update my base URL (I switch between `localhost` and `host.docker.internal` depending on if I'm running the backend in a container -- I just realized I should probably have two separate providers but w.e.), the default model updates to a non-`qwen3:0.6b` model and my updates fail because the validation check before updating the provider fails and I have to switch the default model back to `qwen3:0.6b`.

I think generally this is currently unexpected behavior -- users should assume the same default model is still default unless they choose to make another model default explicitly.

## How Has This Been Tested?

Tested manually

## Additional Options

---
## Summary by cubic
Keep the current global default model when editing the default LLM provider instead of resetting to the first model. Prevents unintended default changes and validation errors.

- **Bug Fixes**
  - Added `defaultModelName` to `LLMProviderFormProps` and threaded it from `LLMConfigurationPage` → `ExistingProviderCard` → `getModalForExistingProvider` → all provider modals (only when editing the global default).
  - Updated `buildDefaultInitialValues` to use the passed default if it exists in the provider’s models; otherwise fall back to the first available model.
  - Ensured correct default model initialization across modals for OpenAI, Anthropic, Vertex AI, Ollama, Azure, Bedrock, LM Studio, LiteLLM Proxy, OpenRouter, and Custom.

<sup>Written for commit d5accfac3044789b035220360d3b914a9741a6ab. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/sections/modals/llmConfig/AnthropicModal.tsx
+++ b/web/src/sections/modals/llmConfig/AnthropicModal.tsx
@@ -73,7 +73,10 @@ export default function AnthropicModal({
         api_key: existingLlmProvider?.api_key ?? "",
         api_base: existingLlmProvider?.api_base ?? undefined,
         default_model_name:
-          defaultModelName ??
+          (defaultModelName &&
+          modelConfigurations.some((m) => m.name === defaultModelName)
+            ? defaultModelName
+            : undefined) ??
           wellKnownLLMProvider?.recommended_default_model?.name ??
           DEFAULT_DEFAULT_MODEL_NAME,
         is_auto_mode: existingLlmProvider?.is_auto_mode ?? true,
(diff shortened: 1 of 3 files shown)


web/src/sections/modals/llmConfig/utils.ts
```diff
@@ -12,9 +12,11 @@ export const LLM_FORM_CLASS_NAME = "flex flex-col gap-y-4 items-stretch mt-6";
 
 export const buildDefaultInitialValues = (
   existingLlmProvider?: LLMProviderView,
-  modelConfigurations?: ModelConfiguration[]
+  modelConfigurations?: ModelConfiguration[],
+  currentDefaultModelName?: string
 ) => {
   const defaultModelName =
+    currentDefaultModelName ??
```
jmelahman:
```suggestion
    (currentDefaultModelName &&
      existingLlmProvider?.model_configurations?.some(
        (m) => m.name === currentDefaultModelName
      )
      ? currentDefaultModelName
      : undefined) ??
```

PR #9514, opened 2026-03-20
fix(fe): scroll stack trace exception into view

jmelahman:
## Description

98% of the time, the exception is all I need to see, so don't make me scroll. Also, some users may not know that's the interesting part, so this helps improve the likelihood the exception is included in bug reports.

## How Has This Been Tested?

Captured by playwright

## Additional Options

---
## Summary by cubic
Automatically scrolls the exception into view when the stack trace is expanded in the error banner. Adds E2E coverage to ensure the error is visible without manual scrolling.

- **Bug Fixes**
  - When the stack trace is expanded, `ErrorBanner` scrolls the exception `<pre>` into view using `scrollIntoView` with `behavior: "instant"` and `block: "end"`.

<sup>Written for commit c48dd49d1ad7cedd5754a611483e18f19424db33. Summary will update on new commits.</sup>


jmelahman:
Too hard :/

PR #9515, opened 2026-03-20
feat(groups): add create group page

nmgarza5:
## Description

Add the Create Group page with header (Cancel/Create buttons), group name input, and a searchable members table that includes both active users and service accounts (API keys). The table supports multi-select with checkboxes, column visibility toggling, and displays masked API keys for service account rows.

Also adds a `disabled` convenience prop to opal's `Button` component — it wraps the button in a `Disabled` context internally, so consumers don't need to import and wrap with `Disabled` separately.

**Linear:** [ENG-3844](https://linear.app/onyx-app/issue/ENG-3844/create-group-page-layout-name-and-members-section)

## Additional Options

### DIFF
--- a/web/src/refresh-pages/admin/GroupsPage/shared.tsx
+++ b/web/src/refresh-pages/admin/GroupsPage/shared.tsx
@@ -104,5 +104,3 @@ export const memberTableColumns = [
   ...baseColumns,
   tc.actions({ showSorting: false }),
 ];
-
-export { tc };


web/lib/opal/src/components/buttons/button/components.tsx
```diff
@@ -43,6 +47,9 @@ type ButtonProps = InteractiveStatelessProps &
 
     /** Which side the tooltip appears on. */
     tooltipSide?: TooltipSide;
+
+    /** Convenience prop — wraps the button in a Disabled context. */
+    disabled?: boolean;
```
nmgarza5: Doing this so I don't always have to import the disabled provider when using an opal button. I would prefer any component that needs a disabled state follow this pattern. If you need to wrap multiple components in the disabled provider, we should add it as a prop to the Section/Card/etc component that wraps it.

cc: @raunakab

web/src/refresh-pages/admin/GroupsPage/CreateGroupPage.tsx
```diff
@@ -0,0 +1,274 @@
...
+    } finally {
+      setIsSubmitting(false);
+    }
+  }
+
+  const headerActions = (
+    <div className="flex flex-row items-center gap-2 shrink-0">
+      <Button
+        prominence="tertiary"
+        onClick={() => router.push("/admin/groups")}
```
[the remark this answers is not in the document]
nmgarza5: This is an invalid comment

web/src/refresh-pages/admin/GroupsPage/CreateGroupPage.tsx
```diff
@@ -0,0 +1,273 @@
...
+  api_key_display: string;
+  api_key_name: string | null;
+  api_key_role: UserRole;
+  user_id: string;
+}
+
+/** Extends UserRow with an optional API key display for service accounts. */
+interface MemberRow extends UserRow {
+  api_key_display?: string;
+}
```
nmgarza5: interfaces.ts

web/src/refresh-pages/admin/GroupsPage/CreateGroupPage.tsx
```diff
@@ -0,0 +1,273 @@
...
+
+// ---------------------------------------------------------------------------
+// Role icon mapping (mirrors UsersPage/UserRoleCell)
+// ---------------------------------------------------------------------------
+
+const ROLE_ICONS: Partial<Record<UserRole, IconFunctionComponent>> = {
+  [UserRole.ADMIN]: SvgUserManage,
+  [UserRole.GLOBAL_CURATOR]: SvgGlobe,
+  [UserRole.SLACK_USER]: SvgSlack,
+};
```
nmgarza5: interfaces.ts 

maybe consume from users interfaces.ts file

PR #9518, opened 2026-03-20
fix(a11y): hidden buttons appear on tabbing

jmelahman:
## Description

Second attempt at having buttons transition to `100` opacity on tabbing. Implements the styling in `Hoverable` per https://github.com/onyx-dot-app/onyx/pull/9481#issuecomment-4092902058.

**before**
20260318_21h19m20s_grim

**after**
20260318_21h18m45s_grim

## Additional Options

---
## Summary by cubic
Make hover-only action buttons reveal on keyboard focus (Tab) and add a clear focus ring for better accessibility. The Users table Edit button now appears on Tab, not just on hover.

- **Bug Fixes**
  - `@opal/core` `Hoverable.Root`: add `onFocusCapture`/`onBlurCapture` to track focus within; include focus in group state so items reveal on Tab.
  - `opacity-on-hover`: reveal when the item or a descendant has `:focus-visible` (via `:has()`); add a visible focus ring.
  - Shared timing tokens `--interactive-duration` and `--interactive-easing` for smoother `.interactive` transitions.
  - Admin Users page: wrap row with `Hoverable.Root`; move Edit into `Hoverable.Item` (group "tags"); replace `IconButton` with `@opal/components` `Button` so Edit reveals on hover or Tab.

<sup>Written for commit 63a0ce953a8246baf8d2337117b65e06650ea37e. Summary will update on new commits.</sup>

### DIFF
--- a/web/lib/opal/src/core/animations/styles.css
+++ b/web/lib/opal/src/core/animations/styles.css
@@ -16,3 +16,15 @@
 .hoverable-item[data-hoverable-variant="opacity-on-hover"][data-hoverable-local="true"]:hover {
   opacity: 1;
 }
+
+/* Focus — item (or a focusable descendant) receives keyboard focus */
+.hoverable-item[data-hoverable-variant="opacity-on-hover"]:has(:focus-visible) {
+  opacity: 1;
+}
+
+/* Focus ring on keyboard focus */
+.hoverable-item:focus-visible {
+  outline: 2px solid var(--border-04);
+  outline-offset: 2px;
+  border-radius: 0.25rem;
+}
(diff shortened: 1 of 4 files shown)


web/lib/opal/src/core/animations/styles.css
```diff
@@ -16,3 +16,16 @@
 .hoverable-item[data-hoverable-variant="opacity-on-hover"][data-hoverable-local="true"]:hover {
   opacity: 1;
 }
+
+/* Focus — item (or a focusable descendant) receives keyboard focus */
+/* Focus — item (or a focusable descendant) receives keyboard focus */
```
jmelahman, line 20:
```suggestion
```

PR #9528, opened 2026-03-20
chore(fe): clean up table qualifier API and internals

raunakab:
## Description

- Consolidate qualifier types to `"simple" | "icon" | "image"`
- Remove redundant `qualifier` prop from `<Table>` — derive from column definitions
- Replace `getInitials`/`getIcon` with `getContent` (`IconFunctionComponent`) and `getImageSrc` (string), add optional `background` prop
- Remove dead fields: `headerContentType`, `header`, `selectable`
- Remove `size` prop from all internal table components — use `useTableSize()` context only
- Remove `className` from `TableQualifier` and `Footer`
- Remove `minWidth` from public column API — auto-derive from header text length
- Remove all sticky positioning from table
- Right-align actions column cells
- Update `UsersTable` to use new qualifier API with `UserAvatar`
- Fix stale footer/story/README references

## Screenshots + Videos

This PR is *largely* internal, affecting very little styling. The largest styling updates include:

- Aligning the buttons in the "action column" to the right side, thus making the items snug.
- The "action column" is no longer sticky. Although this is required from @duo-onyx , it was causing some clipping issues. For now, we will remove it until we can come up with a proper solution.

As per the "action column" button alignment, take a look at the following:

Before:

image

After:

image

Notice how, in the second image, the "kebab button" is properly flush with the right side.

## Additional Options

---
## Summary by cubic
Cleaned up the table qualifier API and internals for a simpler, column‑driven setup. Size props and sticky header styles are removed, actions cells are right‑aligned, and Admin `UsersTable` now uses `UserAvatar` via `getContent`.

- **Refactors**
  - Simplified qualifier types to "simple" | "icon" | "image"; added optional `background`; replaced `getInitials`/`getIcon` with `getContent` and `getImageSrc`/`getImageAlt`.
  - Removed `qualifier` from `Table`; qualifier column controls behavior (simple shows only for multi‑select).
  - Dropped all `size` props across internals/popovers; use `useTableSize()`; removed `className` from `TableQualifier` and `Footer`.
  - Auto‑derived column `minWidth` from header text; removed from the public API.
  - Removed sticky positioning and the `headerBackground` prop; right‑aligned actions column body cells. Added `"use client"` to `TableElement`, `QualifierContainer`, and `ActionsContainer`.
  - Removed dead qualifier fields; updated stories/README; footer mode derives from `selectionBehavior`.

- **Migration**
  - Update `tc.qualifier`:
    - Use `content: "simple" | "icon" | "image"`.
    - Replace `getInitials`/`getIcon` with `getContent`, or use `getImageSrc`/`getImageAlt`; set `background: true` for a tinted container.
    - Remove `headerContentType`, `header`, and `selectable`.
  - Remove `qualifier` and `headerBackground` from `Table` props.
  - Stop passing `size` to subcomponents/popovers; sizing comes from context.
  - Remove `minWidth` from column configs; keep `weight` only.
  - Let footer mode be inferred by `selectionBehavior` (use `footer={{}}` or omit).

<sup>Written for commit 0651ecb20c8df6d842887b0d4755977ef1362316. Summary will update on new commits.</sup>

### DIFF
--- a/web/lib/opal/src/components/table/components.tsx
+++ b/web/lib/opal/src/components/table/components.tsx
@@ -154,7 +154,6 @@ export function Table<TData>(props: DataTableProps<TData>) {
     onRowClick,
     searchTerm,
     height,
-    headerBackground,
     serverSide,
     emptyState,
   } = props;
@@ -348,11 +347,6 @@ export function Table<TData>(props: DataTableProps<TData>) {
                     typeof height === "number" ? `${height}px` : height,
                   overflowY: "auto" as const,
                 }
-              : undefined),
-            ...(headerBackground
-              ? ({
-                  "--table-header-bg": headerBackground,
-                } as React.CSSProperties)
               : undefined),
           }}
         >
--- a/web/lib/opal/src/components/table/README.md
+++ b/web/lib/opal/src/components/table/README.md
@@ -63,7 +63,6 @@ function UsersTable({ users }: { users: User[] }) {
 | `onRowClick` | `(row: TData) => void` | — | Row click handler |
 | `searchTerm` | `string` | — | Global text filter |
 | `height` | `number \| string` | — | Max scrollable height |
-| `headerBackground` | `string` | — | Sticky header background |
 | `serverSide` | `ServerSideConfig` | — | Server-side pagination/sorting/filtering |
 | `emptyState` | `ReactNode` | — | Empty state content |
 
--- a/web/lib/opal/src/components/table/hooks/useColumnWidths.ts
+++ b/web/lib/opal/src/components/table/hooks/useColumnWidths.ts
@@ -277,7 +277,7 @@ function createSplitterResizeHandler(
  * const { containerRef, columnWidths, createResizeHandler } = useColumnWidths({
  *   headers: table.getHeaderGroups()[0].headers,
  *   fixedColumnIds: new Set(["actions"]),
- *   columnMinWidths: { name: 72, status: 80 }, // auto-derived by columns.ts from header text
+ *   columnMinWidths: { name: 72, status: 80 },
  * });
  * ```
  */
--- a/web/lib/opal/src/components/table/types.ts
+++ b/web/lib/opal/src/components/table/types.ts
@@ -165,9 +165,6 @@ export interface DataTableProps<TData> {
    * Accepts a pixel number (e.g. `300`) or a CSS value string (e.g. `"50vh"`).
    */
   height?: number | string;
-  /** Background color for the sticky header row, preventing rows from showing
-   *  through when scrolling. Accepts any CSS color value. */
-  headerBackground?: string;
   /**
    * Enable server-side mode. When provided:
    * - TanStack uses manualPagination/manualSorting/manualFiltering


web/lib/opal/src/components/table/components.tsx
```diff
@@ -152,25 +149,27 @@ export function Table<TData>(props: DataTableProps<TData>) {
     footer,
     size = "lg",
     variant = "cards",
-    qualifier = "simple",
     selectionBehavior = "no-select",
     onSelectionChange,
     onRowClick,
     searchTerm,
     height,
-    headerBackground,
```
Subash-Mohan, line 161: Why did we remove the header background? One of the table variant also needs an explicit background for the header.

web/lib/opal/src/components/table/QualifierContainer.tsx
```diff
@@ -1,22 +1,20 @@
...
   onClick?: (e: React.MouseEvent) => void;
 }
 
 export default function QualifierContainer({
   type,
   children,
-  size,
   onClick,
 }: QualifierContainerProps) {
-  const contextSize = useTableSize();
```
Subash-Mohan, line 18: The main idea behind this is that we should be able to use this qualifier in other places as well. If we limit it only to table size, we won’t be able to use this qualifier elsewhere.

web/lib/opal/src/components/table/Table.stories.tsx
```diff
@@ -108,17 +109,14 @@ const tc = createTableColumns<User>();
...
-    getInitials: (r) =>
-      r.name
-        .split(" ")
-        .map((n) => n[0])
-        .join(""),
+    content: "icon",
+    getContent: () => SvgUser,
+    background: true,
   }),
-  tc.column("name", { header: "Name", weight: 25, minWidth: 120 }),
```
Subash-Mohan, line 118: Why did we remove the minWidth? Without a minWidth, the user may have a poor experience when resizing the column.

web/lib/opal/src/components/table/TableQualifier.tsx
```diff
@@ -24,155 +18,101 @@ interface TableQualifierProps {
...
-    };
+    return "flex bg-action-link-00";
   }
-
-  return {
-    container: "bg-background-tint-01",
-    icon: "stroke-text-03",
-    overlay:
-      "flex opacity-0 group-hover/row:opacity-100 group-focus-within/row:opacity-100 bg-background-tint-01",
-    overlayImage:
```
Subash-Mohan, line 66: I just want you to verify that all variants and states of the qualifier match the mock.

raunakab: Yup! Duo and I walked through this together, and he mentioned that this is correct.

web/lib/opal/src/components/table/Table.stories.tsx
raunakab: `minWidth` still works — it's just no longer part of the public API. It's auto-derived in `columns.ts` from the header text length: `Math.max(header.length * 8 + 40, 80)`, which accounts for the header label plus the sort button. The floor is 80px minimum regardless.

The reason for removing it from the consumer API: previously the end-user could pass a value like `minWidth: 2`, which is clearly too small for any header text + sort button, causing overlapping. Now the floor is always safe and consumers don't need to guess pixel values for every column.

web/lib/opal/src/components/table/QualifierContainer.tsx
raunakab: `QualifierContainer` renders a `<th>` or `<td>` — it can only be used inside a `<table>`. It's a table-internal wrapper, not a reusable component.

If the qualifier *visual* (icon/avatar with checkbox overlay) needs to be reused outside a table, that's `TableQualifier` — which is a `<div>` and works anywhere. But `QualifierContainer` itself is just the table cell wrapper for it.

web/lib/opal/src/components/table/components.tsx
raunakab: I believe the mocks don't contain backgrounds for headers anywhere, but if they do, we can add them back in a follow-up PR.

web/src/refresh-pages/admin/UsersPage/UsersTable.tsx
```diff
@@ -98,19 +103,16 @@ function buildColumns(onMutate: () => void) {
...
       header: "Status",
       weight: 14,
-      minWidth: 100,
       cell: renderStatusColumn,
     }),
     tc.column("updated_at", {
       header: "Last Updated",
       weight: 14,
-      minWidth: 100,
       cell: renderLastUpdatedColumn,
```
nmgarza5, line 116: @raunakab removing these min widths broke the users table. was this accidental?

nmgarza5: `Remove minWidth from public column API — auto-derive from header text length`

doesnt seem to work. columns grow based on the row contents


Screenshot 2026-03-24 at 11 13 39 PM

nmgarza5: confused why we did this.... this is how the tanstack table works: https://tanstack.com/table/v8/docs/guide/column-sizing

PR #9529, opened 2026-03-20
feat(fe): migrate root layout from SSR to CSR with SWR data fetching

nmgarza5:
## Description

Migrates the Next.js root layout from server-side rendering to client-side rendering with SWR for all global data fetching (settings, enterprise settings, auth metadata, user). This eliminates RSC prefetch requests that were triggering WAF rate limiting on page navigations.

**Key changes:**
- New SWR hooks (`useSettings`, `useEnterpriseSettings`, `useCustomAnalyticsScript`, `useAuthTypeMetadata`) replace server-side fetch calls in `layout.tsx`
- `SettingsProvider` now fetches its own data client-side instead of receiving props from the layout
- Logo component switched from Next.js `Image` to plain `` with cache-buster query param (Next.js `Image` has internal caching that ignores `Cache-Control` headers)
- Backend logo endpoint now returns `Cache-Control: no-cache` so browsers revalidate
- Backend settings endpoint now includes `version` field (used by frontend to detect version changes)
- `DynamicMetadata`, `CustomAnalyticsScript`, and `ProductGatingWrapper` extracted into dedicated provider components
- Auth errors (401/403) are expected for unauthenticated users and handled gracefully — not treated as fatal

## Additional Options


web/src/hooks/useGroups.ts
```diff
@@ -1,6 +1,6 @@
 "use client";
 
-import useSWR from "swr";
+import useSWR, { mutate as globalMutate } from "swr";
```
raunakab: A super minor nit, but do we need this `mutate as globalMutate` alias?

web/src/hooks/useShareableGroups.ts
```diff
@@ -1,6 +1,6 @@
 "use client";
 
-import useSWR from "swr";
+import useSWR, { mutate as globalMutate } from "swr";
```
raunakab: Same alias comment here.

web/src/providers/ProductGatingWrapper.tsx
```diff
@@ -0,0 +1,25 @@
...
+  if (settingsLoading) return null;
+
+  if (
+    status === ApplicationStatus.GATED_ACCESS ||
+    status === ApplicationStatus.SEAT_LIMIT_EXCEEDED
+  ) {
+    return <GatedContentWrapper>{children}</GatedContentWrapper>;
+  }
+
+  return <>{children}</>;
```
raunakab: Minor nit here as well, but you can maybe just return `children`?

raunakab (approved):
Nice!

web/src/providers/ProductGatingWrapper.tsx
nmgarza5: claude being dumb dumb

web/src/hooks/useGroups.ts
nmgarza5: claude being dumb dumb

cotamatcotam:
what is the point of using Next.js if we don't use SSR? :))
BTW, nice work, I am in progress of forking onyx but using Tanstack Start instead :)))

PR #8715, opened 2026-02-24
feat(voice mode): stt and tts

jessicasingh7:
## Description
first STT

image
TTS:
while onyx is speaking:
image
after it's done speaking:
Screenshot 2026-03-05 at 2 28 23 PM
second STT:
Screenshot 2026-03-05 at 2 32 57 PM

admin panel:
image
specific provider setup (TTS):
Screenshot 2026-03-05 at 2 35 51 PM
user settings:
image

## Additional Options

- [x] [Optional] Please cherry-pick this PR to the latest release version.
- [x] [Optional] Override Linear Check

---
## Summary by cubic
Adds voice mode with real-time speech-to-text and gapless, sentence-level streaming text-to-speech synced to message generation. Includes admin setup, Redis-backed WebSocket auth, SSRF protections, and a DB migration for voice providers plus per-user voice preferences.

- **New Features**
  - STT/TTS: HTTP + WebSocket endpoints for transcribe/synthesize with gapless streaming, plus GET /api/voice/status for conditional UI.
  - Providers + security: Unified factory with OpenAI and ElevenLabs streaming STT/TTS, Azure streaming TTS (Azure STT off until resampling); SSML escaping; SSRF-safe outbound URL validation; adds `azure-cognitiveservices-speech`.
  - Admin + UI: Provider setup and defaults; per-message “Read aloud”; mic recording with live transcription and waveform; user voice settings (auto-send, auto-playback, speed); controls hide when unconfigured; mic append mode; 5-minute TTS session safety timeout; new icons and a `VoiceModeProvider` for playback/session state.

- **Bug Fixes**
  - Reliability/UX: Fixed OpenAI transcript wait exiting early; cancel TTS on flush timeout; recorder start cleanup; ObjectURL leak fixes in recorder/playback; moved ref mutations to useEffect and added messageNodeId to renderer memoization; prevent double-send on VAD+stop with time-windowed duplicate detection.
  - Auth/limits: Atomic, rate-limited WS token issuance/lookup in Redis; narrowed auth bypass to only voice routes; correct synthesize request params; added 413 PAYLOAD_TOO_LARGE; validated TTS query params and clamped playback speed.
[...]

PR #9558, opened 2026-03-23
fix(voice): add WebSocket upgrade headers to nginx configs

jessicasingh7:
## Description

The voice feature (PR #8715) added the first WebSocket endpoints in Onyx (`/voice/transcribe/stream` and `/voice/synthesize/stream`). All shipped nginx configs (docker compose, prod/TLS, no-letsencrypt, and Helm) were missing the `Upgrade` and `Connection` headers required to proxy WebSocket handshakes. Self-hosted customers hitting these endpoints get a **404 from nginx** because the WebSocket upgrade is silently dropped.

Adds `proxy_set_header Upgrade` and `proxy_set_header Connection` (using the standard nginx `map` pattern) to the `/api` location blocks in all 4 nginx config files. For the two-hop TLS configs (prod and no-letsencrypt), the headers are also added to the port-443 passthrough block so they survive both proxy hops.

## How Has This Been Tested?

Ran the backend and frontend locally, then started a standalone nginx:1.25.5-alpine Docker container with the updated config proxying to the local services. Verified:

1. **Regular API requests** — `GET /api/health` through nginx returns 200, confirming no regression for normal HTTP traffic.
2. **WebSocket upgrade with wrong origin** — Returns 403 "Invalid origin" from the backend's WebSocket auth. Before the fix, this was a 404 from nginx (request never reached the backend).
3. **WebSocket upgrade with correct origin + fake token** — Returns 403 "Invalid or expired authentication token", confirming the request passes through nginx and hits the full WebSocket auth pipeline.

## Additional Options

### DIFF
--- a/deployment/data/nginx/app.conf.template
+++ b/deployment/data/nginx/app.conf.template
@@ -23,6 +23,12 @@ upstream web_server {
 # Conditionally include MCP upstream configuration
 include /etc/nginx/conf.d/mcp_upstream.conf.inc;
 
+# WebSocket support: only set Connection "upgrade" for actual upgrade requests
+map $http_upgrade $connection_upgrade {
+    default upgrade;
+    ''      close;
+}
+
 server {
     listen 80 default_server;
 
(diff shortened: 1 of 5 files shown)


jessicasingh7:
@greptile

jessicasingh7:
@greptile

PR #9616, opened 2026-03-25
feat(llm): add Bifrost gateway as LLM provider (backend)

justin-tahara:
## Description

Registers Bifrost as a first-class LLM provider in the backend. Bifrost is an OpenAI-compatible AI gateway that routes requests to multiple upstream LLM providers (OpenAI, Anthropic, etc.) using `provider/model` naming conventions.

**Key changes:**
- Register Bifrost in `LlmProviderNames`, `WELL_KNOWN_PROVIDER_NAMES`, `AGGREGATOR_PROVIDERS`, and `DYNAMIC_LLM_PROVIDERS`
- Add `/bifrost/available-models` API endpoint that fetches models from a Bifrost gateway's `/v1/models` endpoint
- Handle runtime routing: set `custom_llm_provider="openai"` and pass model names directly (e.g. `anthropic/claude-sonnet-4-6`) without LiteLLM double-prefixing
- Add vendor extraction for Bifrost's `provider/model` format (shares logic with OpenRouter)

The key runtime challenge is that LiteLLM normally prepends the provider name to the model (e.g. `openai/model`), which would result in `bifrost/anthropic/claude-sonnet-4-6` — breaking Bifrost's routing. This PR sets `custom_llm_provider="openai"` and bypasses the prefix logic for Bifrost.

## How Has This Been Tested?

- Verified backend starts without import errors
- Verified Python model imports work correctly
- Endpoint designed to match existing patterns (LiteLLM Proxy, OpenRouter)
- Manual curl testing against a live Bifrost instance confirmed `/v1/models` returns expected format

## Additional Options

---
## Summary by cubic
Add Bifrost as a first-class LLM provider with OpenAI-compatible routing and a models discovery endpoint. Requests use `provider/model` names and route via `/v1` without LiteLLM double-prefixing.

- New Features
  - Register `BIFROST` across provider enums, display names, aggregators, and dynamic providers.
  - Runtime routing: set `custom_llm_provider="openai"`, normalize API base to `<base>/v1`, and send `provider/model` (e.g., `anthropic/claude-sonnet-4-6`) directly.
  - `POST /bifrost/available-models`: fetch from `<api_base>/v1/models` with optional Bearer auth, skip embedding models, return sorted list with `name`, `display_name`, `max_input_tokens`, `supports_image_input`; optionally sync to DB when `provider_name` is set.
  - Behavior tweaks: preserve tool selection by sending `allowed_openai_params=["tool_choice"]` for Bifrost; infer vision support from known model families; vendor extraction supports `provider/model` format.

<sup>Written for commit 4a160bdaa3e04f74f39775242d88c15cb3654efd. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/server/manage/llm/utils.py
+++ b/backend/onyx/server/manage/llm/utils.py
@@ -40,9 +40,20 @@ class ModelMetadata(TypedDict):
 # Non-LLM model patterns to filter out (image gen, embeddings, etc.)
 NON_LLM_PATTERNS = frozenset({"embed", "stable-", "titan-image", "titan-embed"})
 
-# Known vision-capable model families for providers where the source API
-# does not expose this metadata directly.
-VISION_MODEL_FAMILIES = frozenset(
+# Known Bedrock vision-capable models (for fallback when base model not in region)
+BEDROCK_VISION_MODELS = frozenset(
+    {
+        "anthropic.claude-3",
+        "anthropic.claude-4",
+        "amazon.nova-pro",
+        "amazon.nova-lite",
+        "amazon.nova-premier",
+    }
+)
+
+# Known Bifrost/OpenAI-compatible vision-capable model families where the
+# source API does not expose this metadata directly.
+BIFROST_VISION_MODEL_FAMILIES = frozenset(
     {
         "anthropic/claude-3",
         "anthropic/claude-4",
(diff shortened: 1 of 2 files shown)


backend/onyx/server/manage/llm/utils.py
```diff
@@ -76,11 +96,18 @@ def is_valid_bedrock_model(
...
     model_id_lower = model_id.lower()
-    return any(vision_model in model_id_lower for vision_model in BEDROCK_VISION_MODELS)
+    if any(vision_model in model_id_lower for vision_model in BEDROCK_VISION_MODELS):
+        return True
+
+    normalized_model_id = model_id_lower.replace(".", "/")
+    return any(
+        vision_model in normalized_model_id
+        for vision_model in BIFROST_VISION_MODEL_FAMILIES
+    )
```
Danelegend, line 110: Just want to check that we don't think there'll be problems with this being a check every model goes through (ie. Model not on bifrost hits this, resulting in true).

Also, if this is more general, maybe worth changing the name to something like 'KNOWN_VISION_MODELS'?

PR #9632, opened 2026-03-25
chore(agents): Match Mocks & Add Date Validation

justin-tahara:
## Description

After the refresh the page was not aligned with what we had in the mocks and the date validation was not properly handled. 

This is something that was reported to us and we validated and updated. 

The date picker is locked to the previous days and only dates in the past. Updated the text to make sure that it's less confusing. 

## How Has This Been Tested?

Tested it locally and validated that it matches the mock now. Validated that you are not able to select a date in the future.

Before:
Screenshot 2026-03-25 at 12 51 42 PM

After:
Screenshot 2026-03-25 at 12 45 02 PM

## Additional Options
Following the mocks from the Linear ticket
Closes https://linear.app/onyx-app/issue/ENG-3877/agent-advanced-options-not-up-to-spec

---
## Summary by cubic
Aligns Agent Advanced Settings with the mocks and blocks future Knowledge Cutoff dates. Adds clearer copy and a muted title suffix option; also makes admin group routing type-safe.

- New Features
  - `InputDatePicker`: adds `maxDate`, disables future days, normalizes dates, clamps the initial month, and makes “Today” respect `maxDate`; Yup rule blocks future `knowledge_cutoff_date`.
  - UI: `titleSuffix` in `ContentMd` (plumbed through input layouts); updated copy for Default Model and Knowledge Cutoff (optional), and adds “(Not Recommended)” to Overwrite System Prompt; marks Reminders as optional.

- Bug Fixes
  - Matches labels and behavior to the mocks from Linear ENG-3877.
  - Admin Groups: type-safe `router.push` via `Route`.

<sup>Written for commit 5257121eb5b357badada836a4ed1c6043385517d. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/refresh-components/inputs/InputDatePicker.tsx
+++ b/web/src/refresh-components/inputs/InputDatePicker.tsx
@@ -2,11 +2,11 @@
 
 import { Button } from "@opal/components";
 import { Disabled } from "@opal/core";
-import { toast } from "@/hooks/useToast";
+import { isAfterDate, normalizeDate } from "@/lib/dateUtils";
 import Calendar from "@/refresh-components/Calendar";
 import Popover from "@/refresh-components/Popover";
 import InputSelect from "@/refresh-components/inputs/InputSelect";
-import { useState } from "react";
+import { useMemo, useState } from "react";
 import { SvgCalendar } from "@opal/icons";
 import { Section } from "@/layouts/general-layouts";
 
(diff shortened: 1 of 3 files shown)


web/lib/opal/src/layouts/content/components.tsx
raunakab, line 1: I think a better API here instead of

```tsx
optional?: boolean;
titleSuffix?: string;
```

would be

```tsx
suffix?: string | keyof typeof TitleSuffixCommonOptions;
const TitleSuffixCommonOptions = { optional: "Optional", ... };
```

That way, end-consumers can either pass in a string OR one of those enum strings.

raunakab: Not a blocker for this PR, though.

web/src/refresh-components/inputs/InputDatePicker.tsx
raunakab, line 1: Hmm.. Not sure if I'm a fan of components throwing toasts... it couples a component (which should be extracted into a library) to application-specific knowledge (how errors should be surfaced via toasts).

raunakab: Not a blocker for this PR, though.

PR #9647, opened 2026-03-25
feat(chat): add multi-model parallel streaming backend

nmgarza5:
## Description

Implements the server-side execution path for multi-model comparison — running 2–3 LLMs in parallel on a single chat request and streaming their interleaved responses to the client. Stacked on #9803.

**New in this PR:**

- **`handle_multi_model_stream`** — new entry point; validates 2–3 overrides, rejects deep research, delegates to `_stream_chat_turn`
- **`build_chat_turn`** — extended to accept `llm_overrides`; reserves N `ChatMessage` IDs and yields `MultiModelMessageResponseIDInfo` for multi-model turns; single-model path unchanged
- **`_run_models`** — parallel execution engine with `ThreadPoolExecutor`; one worker thread per model, shared `queue.Queue`, drain loop with per-model error isolation; includes three concurrency correctness fixes:
  - **B1** (`completion_locks`): prevents double-completion race when a model finishes just before `GeneratorExit` fires on HTTP disconnect
  - **B2** (`model_errored` flag): stop-button loop skips errored models instead of saving "stopped by user" for a model that actually threw an exception
  - **P1** (`_delete_orphaned_message`): normal completion, stop-button, and disconnect paths delete reserved `ChatMessage` rows for models that errored
- **`reserve_multi_model_message_ids` / `set_preferred_response`** — DB helpers in `chat.py`
- **`chat_backend.py`** — new multi-model API endpoints
- **`test_multi_model_streaming.py`** — 26 unit tests covering emitter routing, drain-loop behaviour, cancellation, B1/B2/P1 races

**Backwards compatibility:** `handle_stream_message_objects` signature is unchanged; all existing single-model callers continue working through the same `_stream_chat_turn` path with `llm_overrides=None`.

## Additional Options

### DIFF
--- a/backend/onyx/chat/process_message.py
+++ b/backend/onyx/chat/process_message.py
@@ -13,7 +13,6 @@ from collections.abc import Callable
 from collections.abc import Generator
 from concurrent.futures import ThreadPoolExecutor
 from contextvars import Token
-from dataclasses import dataclass
 from typing import Final
 from uuid import UUID
 
@@ -71,7 +70,6 @@ from onyx.db.chat import reserve_multi_model_message_ids
 from onyx.db.engine.sql_engine import get_session_with_current_tenant
 from onyx.db.enums import HookPoint
 from onyx.db.memory import get_memories
-from onyx.db.memory import UserMemoryContext
 from onyx.db.models import ChatMessage
 from onyx.db.models import Persona
 from onyx.db.models import User
@@ -124,12 +122,12 @@ from onyx.tools.tool_constructor import FileReaderToolConfig
 from onyx.tools.tool_constructor import SearchToolConfig
 from onyx.utils.logger import setup_logger
 from onyx.utils.telemetry import mt_cloud_telemetry
-from onyx.utils.threadpool_concurrency import run_multiple_in_background
 from onyx.utils.timing import log_function_time
 from shared_configs.contextvars import get_current_tenant_id
 
 logger = setup_logger()
 ERROR_TYPE_CANCELLED = "cancelled"
+APPROX_CHARS_PER_TOKEN = 4
 
 
 def _collect_available_file_ids(
@@ -309,9 +307,13 @@ def extract_context_files(
     # Aggregate tokens for the file content that will be added
     # Skip tokens for those with metadata only
     aggregate_tokens = sum(
-        uf.token_count or 0 for uf in user_files if not mime_type_to_chat_file_type(uf.file_type).use_metadata_only()
+        uf.token_count or 0
+        for uf in user_files
+        if not mime_type_to_chat_file_type(uf.file_type).use_metadata_only()
     )
-    max_actual_tokens = (llm_max_context_window - reserved_token_count) * max_llm_context_percentage
+    max_actual_tokens = (
+        llm_max_context_window - reserved_token_count
+    ) * max_llm_context_percentage
 
     if aggregate_tokens >= max_actual_tokens:
         use_as_search_filter = not DISABLE_VECTOR_DB
@@ -319,7 +321,9 @@ def extract_context_files(
             overflow_tool_metadata = [_build_tool_metadata(uf) for uf in user_files]
         else:
             overflow_tool_metadata = [
-                _build_tool_metadata(uf) for uf in user_files if mime_type_to_chat_file_type(uf.file_type).use_metadata_only()
+                _build_tool_metadata(uf)
+                for uf in user_files
+                if mime_type_to_chat_file_type(uf.file_type).use_metadata_only()
             ]
         return ExtractedContextFiles(
             file_texts=[],
@@ -352,7 +356,9 @@ def extract_context_files(
             # Metadata-only files are not injected as full text.
             # Only the metadata is provided, with LLM using tools
             if not uf:
-                logger.error(f"File with id={f.file_id} in metadata-only path with no associated user file")
+                logger.error(
+                    f"File with id={f.file_id} in metadata-only path with no associated user file"
+                )
                 continue
             tool_metadata.append(_build_tool_metadata(uf))
         elif f.file_type.is_text_file():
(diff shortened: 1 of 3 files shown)


nmgarza5:
@greptile

backend/onyx/chat/process_message.py
```diff
@@ -1069,6 +1079,583 @@ def llm_loop_completion_callback(
             logger.exception("Error in setting processing status")
 
 
+def _build_model_display_name(override: LLMOverride) -> str:
+    """Build a human-readable display name from an LLM override."""
+    if override.display_name:
+        return override.display_name
+    if override.model_version:
```
evan-onyx: I feel like this shouldn't be the way we get display name: if we ever go past the first case then the user will be confused when their model is just "OpenAI" or wtvr

backend/onyx/chat/process_message.py
```diff
@@ -1069,6 +1079,583 @@ def llm_loop_completion_callback(
...
+    """Emitter that tags packets with model_index and forwards directly to a shared queue.
+
+    Unlike the standard Emitter (which accumulates in a local bus), this puts
+    packets into the shared merged_queue in real-time as they're emitted. This
+    enables true parallel streaming — packets from multiple models interleave
+    on the wire instead of arriving in bursts after each model completes.
+    """
+
+    def __init__(self, model_idx: int, merged_queue: queue.Queue) -> None:
+        super().__init__(queue.Queue())  # bus exists for compat, unused
```
evan-onyx: this breaks the main thing that the Emitter class is supposed to do afaik

backend/onyx/chat/process_message.py
```diff
@@ -412,6 +414,9 @@ def extract_context_files(
     )
 
 
+APPROX_CHARS_PER_TOKEN = 4
```
evan-onyx: nit: this should probably stay top of file

backend/onyx/chat/process_message.py
```diff
@@ -611,21 +623,42 @@ def build_chat_turn(
...
-    check_llm_cost_limit_for_provider(
-        db_session=db_session,
-        tenant_id=tenant_id,
-        llm_provider_api_key=primary_llm.config.api_key,
-    )
-    llms = [primary_llm]
-    model_display_names = [""]
-    token_counter = get_llm_token_counter(primary_llm)
+    # then build the LLM instance(s).
+    if is_multi:
```
evan-onyx: nit: I think this is cleaner (replacing the whole block):

```
llms: list[LLM] = []
model_display_names: list[str] = []
selected_overrides = llm_overrides if is_multi else [new_msg_req.llm_override or chat_session.llm_override]
for override in selected_overrides:
    llm = get_llm_for_persona(
        persona=persona,
        user=user,
        llm_override=override,
        additional_headers=litellm_additional_headers,
    )
    check_llm_cost_limit_for_provider(
        db_session=db_session,
        tenant_id=tenant_id,
        llm_provider_api_key=llm.config.api_key,
    )
    llms.append(llm)
    model_display_names.append(_build_model_display_name(override))
token_counter = get_llm_token_counter(llms[0])

# not sure why we do this, but to maintain parity with previous code:
if not is_multi:
    model_display_names = [""]
```

backend/onyx/chat/process_message.py
```diff
@@ -786,7 +819,12 @@ def build_chat_turn(
         db_session=db_session,
     )
 
-    llm_max_context_window = llms[0].config.max_input_tokens
+    # For multi-model: use the smallest context window across all models for safety
+    llm_max_context_window = (
+        min(llm.config.max_input_tokens for llm in llms)
+        if is_multi
```
evan-onyx: nit: don't need the if/else, just use min

backend/onyx/chat/process_message.py
```diff
@@ -1115,9 +1187,7 @@ def _run_model(model_idx: int) -> None:
                     if assistant_message is not None:
                         llm_loop_completion_handle(
                             state_container=state_containers[model_idx],
-                            # model_succeeded[model_idx] is True here (guard above),
-                            # so treat as connected — avoids "stopped by user" suffix.
-                            is_connected=lambda: model_succeeded[model_idx],
+                            is_connected=lambda: True,
```
evan-onyx: leaving a shorter version of the previously existing comment would probs be helpful for someone

backend/onyx/chat/process_message.py
```diff
@@ -991,11 +1048,20 @@ def _run_models(
...
 
     # Set when the drain loop exits early (HTTP disconnect / GeneratorExit).
     # Signals emitters to skip future puts and workers to self-complete.
     drain_done = threading.Event()
 
+    # One lock per model. acquire(blocking=False) claims completion responsibility.
+    # Whichever caller (worker self-complete or disconnect handler) acquires first
+    # is the one that calls llm_loop_completion_handle for that model.
+    # Locks are intentionally never released — they serve as permanent "claimed" markers.
+    completion_locks: list[threading.Lock] = [threading.Lock() for _ in range(n_models)]
```
evan-onyx: I'd ideally like to not need locks here by i.e. having all LLM completion (errors and normal completion) be handled by the main thread. But if this isn't doable or makes things extra gross then we can leave as is.

backend/onyx/db/chat.py
```diff
@@ -131,47 +132,32 @@ def get_chat_sessions_by_user(
...
+            .where(ChatMessage.chat_session_id == ChatSession.id)
+            .where(ChatMessage.message_type != MessageType.SYSTEM)
+            .correlate(ChatSession)
+        )
 
-    result = db_session.execute(stmt)
-    chat_sessions = list(result.scalars().all())
-
-    if not include_failed_chats and chat_sessions:
-        # Filter out "failed" sessions (those with only SYSTEM messages)
```
evan-onyx, line 148: I'm very nervous that we're switching to the approach that this comment explicitly recommends against 😅

PR #9668, opened 2026-03-26
chore(greptile): split greptile.json into .greptile/ directory

nmgarza5:
## Description

Restructures the monolithic `greptile.json` into three focused files under `.greptile/`:

- **`config.json`** — review settings + 4 scoped rules (rules that target specific file globs like `web/**` or `backend/**/*.py`, so Greptile enforces them only on matching files)
- **`rules.md`** — 7 global rules as prose markdown (guidelines that apply everywhere and don't need per-rule scoping or disable support)
- **`files.json`** — context file references the reviewer reads before reviewing

The split follows Greptile's own distinction: `config.json` rules support per-rule glob scoping, severity levels, and can be disabled by ID — making them the right home for precise, file-targeted rules. `rules.md` is free-form markdown suited to prose guidelines that apply to the whole codebase.

Also bumps `strictness` from `2` → `3` and clears the stale `ignorePatterns` entry.

## Additional Options

### DIFF
--- a/.greptile/files.json
+++ b/.greptile/files.json
@@ -10,6 +10,31 @@
     "description": "Frontend coding standards for the web directory"
   },
   {
+    "scope": ["web/**"],
+    "path": "web/tests/README.md",
+    "description": "Frontend testing guide and conventions"
+  },
+  {
+    "scope": ["web/**"],
+    "path": "web/CLAUDE.md",
+    "description": "Single source of truth for frontend coding standards"
+  },
+  {
+    "scope": ["web/**"],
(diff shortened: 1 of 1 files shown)


.greptile/config.json
```diff
@@ -0,0 +1,68 @@
+{
+    "labels": [],
+    "comment": "",
+    "fixWithAI": true,
+    "hideFooter": false,
+    "strictness": 3,
```
nmgarza5, line 6: should help with the thrash but if this ends up being too strong we can look into other options

PR #9678, opened 2026-03-26
fix(api-key): clarify upgrade message for trial accounts

nmgarza5:
## Description

Updates the admin API key page to display a clear warning message to trial users that API key creation requires a paid subscription, with an upgrade CTA linking to billing. Also fixes TypeScript errors in the `isTrialing` check — `billingData`'s narrowing from `!= null` wasn't propagating inside the `true || (...)` dead-code branch, causing TS18048/TS2339 errors. Fixed by extracting the billing check to its own const and using the existing `hasActiveSubscription` type predicate.

## How Has This Been Tested?
image

## Additional Options

### DIFF
--- a/web/src/app/admin/api-key/page.tsx
+++ b/web/src/app/admin/api-key/page.tsx
@@ -46,15 +46,12 @@ function Main() {
     error,
   } = useSWR<APIKey[]>("/api/admin/api-key", errorHandlingFetcher);
 
-  const canCreateKeys = false; // MOCK: hardcoded for screenshot
+  const canCreateKeys = useCloudSubscription();
   const { data: billingData } = useBillingInformation();
-  const billingIsTrialing =
+  const isTrialing =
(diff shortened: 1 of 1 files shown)


web/src/app/admin/api-key/page.tsx
```diff
@@ -85,23 +103,9 @@ function Main() {
...
-              variant="none"
-              prominence="tertiary"
-              size="2xs"
-              icon={SvgInfo}
-              tooltip="API keys enable programmatic access to Onyx for service accounts and integrations. Trial accounts do not include API key access — purchase a paid subscription to unlock this feature."
-            />
-          </div>
-          <Button href="/admin/billing">Upgrade to Paid Plan</Button>
-        </div>
-      )}
```
nmgarza5, line 104: i goofed yesterday and forgot to merge the updated version of this. this was my first iteration that looked bad

PR #9693, opened 2026-03-27
perf(swr): add sessionStorage cache provider to persist SWR cache across reloads

nmgarza5:
## Description

Every hard page reload was firing ~30 API requests from scratch because SWR's cache is in-memory only — wiped on every browser refresh. This adds a `sessionStorage`-backed cache provider to `SWRConfigProvider` so SWR can hydrate from cached responses immediately on reload, then revalidate in the background.

**What changed:**
- Added `sessionStorageProvider` function in `SWRConfigProvider.tsx` that reads the SWR cache from `sessionStorage` on init and writes it back on `beforeunload`
- Wired it into the global `SWRConfig` via the `provider` prop

**Behavior after this change:**
- First page load: all ~30 requests fire as usual (cache is empty)
- Subsequent reloads (F5): cached responses hydrate immediately → UI renders with stale data → SWR revalidates in background per each hook's existing settings (`revalidateOnMount: true` default)
- Tab close / new tab: `sessionStorage` clears → fresh fetch on next visit

`sessionStorage` (not `localStorage`) is used intentionally so the cache is scoped to the browser session and auto-clears when the tab closes.

## Additional Options

### DIFF
--- a/web/src/providers/SWRConfigProvider.tsx
+++ b/web/src/providers/SWRConfigProvider.tsx
@@ -1,7 +1,51 @@
 "use client";
 
-import { SWRConfig } from "swr";
+import { SWRConfig, type State } from "swr";
 import { skipRetryOnAuthError } from "@/lib/fetcher";
+
+const CACHE_KEY = "onyx-swr-v1";
+
+/**
+ * Backs the SWR in-memory cache with sessionStorage so cached responses
+ * survive page reloads. SWR still revalidates on mount (stale-while-revalidate),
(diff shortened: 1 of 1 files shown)


nmgarza5:
Closing in favor of a simpler approach: revalidateIfStale: false (PR #9695) suppresses re-fetches during client-side navigation using the in-memory SWR cache. Hard refreshes intentionally fire fresh requests. No sessionStorage layer needed.

PR #9695, opened 2026-03-27
perf(swr): add SWR_KEYS registry and skip revalidation for stable hooks

nmgarza5:
## Description

Two related changes to improve SWR cache hygiene:

**1. Centralized SWR_KEYS registry (`web/src/lib/swr-keys.ts`)**

Adds a single `as const` object containing all SWR endpoint strings used by the hooks in this PR. Previously, endpoint strings were inline literals, making it easy for a `mutate()` call to silently target a different string than the `useSWR()` call it was meant to invalidate.

A full migration of all remaining inline strings across the codebase will follow in a separate PR.

**2. `revalidateIfStale: false` on stable data hooks**

Adds `revalidateIfStale: false` to 18 hooks whose data only changes through explicit user actions. Without this flag, SWR fires a background re-fetch on every mount even when the cache is fresh, causing a burst of redundant requests on every page load.

Affected hooks: `useAgentPreferences`, `useAgents` / `useAgent`, `useAuthTypeMetadata`, `useAvailableTools`, `useBillingInformation`, `useLicense`, `useChatSessions`, `useCurrentUser`, `useDocumentSets`, `useLLMProviders` / `useAdminLLMProviders` / `useLLMProviderOptions`, `useProjects`, `usePromptShortcuts`, `useSettings` / `useEnterpriseSettings` / `useCustomAnalyticsScript`, `useTags`, `useVoiceProviders` / `useVoiceStatus`.

Data that changes independently of user action (connector sync status, embedding progress, notifications, health checks) is deliberately excluded.

**3. Mutation fix**

Adds `mutate(SWR_KEYS.documentSets)` and `mutate(SWR_KEYS.documentSetsEditable)` in `DocumentSetCreationForm` after create/update so the document sets list stays fresh without relying on the polling interval.

## How Has This Been Tested?

Manual testing + existing CI

## Additional Options

### DIFF
--- a/web/src/app/admin/documents/sets/hooks.tsx
+++ b/web/src/app/admin/documents/sets/hooks.tsx
@@ -1,17 +1,16 @@
 import { errorHandlingFetcher } from "@/lib/fetcher";
 import { DocumentSetSummary } from "@/lib/types";
 import useSWR, { mutate } from "swr";
-
-const DOCUMENT_SETS_URL = "/api/manage/document-set";
-const GET_EDITABLE_DOCUMENT_SETS_URL =
-  "/api/manage/document-set?get_editable=true";
+import { SWR_KEYS } from "@/lib/swr-keys";
 
 export function refreshDocumentSets() {
-  mutate(DOCUMENT_SETS_URL);
+  mutate(SWR_KEYS.documentSets);
 }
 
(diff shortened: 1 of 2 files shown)


nmgarza5:
will follow up and add all swr keys to this registry if you like the idea

PR #9710, opened 2026-03-27
chore(release): build all CLI wheels before publishing

jmelahman:
## Description

Builds all of the CLI wheels before publishing so that publishing is atomic.

## Additional Options

---
## Summary by cubic
Build all CLI wheels for linux, windows, and darwin (amd64 and arm64) before publishing to make releases atomic and prevent partial releases. Replaces the matrix strategy with an in-job loop so `uv publish` runs only after all wheels build successfully.

<sup>Written for commit dc1547be3e1ca005f29fc70e3f624a5b171a4721. Summary will update on new commits.</sup>

### DIFF
--- a/.github/workflows/release-cli.yml
+++ b/.github/workflows/release-cli.yml
@@ -13,15 +13,6 @@ jobs:
     permissions:
       id-token: write
     timeout-minutes: 10
-    strategy:
-      matrix:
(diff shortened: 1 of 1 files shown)


bekelnanti:
@CodiumAI-Agent /improve

PR #9734, opened 2026-03-28
Fix persona is_visible field reset on update

### LINKED ISSUE #9517: Bug: Persona `is_visible` field gets reset to `True` when updating persona through AgentEditor page
**I'm happy to submit a PR to fix this issue if you'd like.** 

### Description

When updating a persona through the normal agent editor flow, the `is_visible` field is inadvertently reset to `True`, even if it was previously set to `False` by an admin. This happens because the `is_visible` parameter is not passed through the update flow, causing it to use the default value.

### Root Cause

The `is_visible` field is missing from the entire persona update chain:

1. **Frontend (`web/src/refresh-pages/AgentEditorPage.tsx`)**:
   - `initialValues` doesn't include `is_visible`
   - The form doesn't capture or send `is_visible` to the API

2. **Frontend API layer (`web/src/app/admin/agents/lib.ts`)**:
   - `PersonaUpsertParameters` interface doesn't include `is_visible`
   - `buildPersonaUpsertRequest` doesn't handle `is_visible`

3. **Backend API (`backend/onyx/server/features/persona/models.py`)**:
   - `PersonaUpsertRequest` doesn't include `is_visible` field

4. **Backend DB (`backend/onyx/db/persona.py`)**:
   - `create_update_persona()` calls `upsert_persona()` but never passes `is_visible`
   - `upsert_persona()`  has `is_visible: bool = True` as a parameter
   - When updating an existing persona, it always sets: `existing_persona.is_visible = is_visible`

### Steps to Reproduce

1. As an admin, create an agent and set it to `is_visible=False` using the admin endpoint:
   ```
   PATCH /api/admin/persona/{persona_id}/visible
   {"is_visible": false}
   ```

2. Verify the persona is now hidden (not visible to non-admin users)

3. Edit that same persona through the normal agent editor (e.g., change the description)

4. Save the persona

5. Check the `is_visible` field in the database

### Expected Behavior

The `is_visible` field should be preserved when updating a persona through the editor.

### Actual Behavior

The `is_visible` field gets reset to `True` (the default value in `upsert_persona()`), making the persona visible again even though an admin explicitly hid it.

### Impact

- Admins cannot reliably hide personas from users
- Hidden personas become visible again after any edit

### Environment

- Onyx version: 50dde0be1
- Affected files:
  - `backend/onyx/db/persona.py`
  - `backend/onyx/server/features/persona/api.py`
  - `backend/onyx/server/features/persona/models.py`
  - `web/src/refresh-pages/AgentEditorPage.tsx`
  - `web/src/app/admin/agents/lib.ts`

jnMetaCode:
## Summary

- Fixes #9517: The `is_visible` (`is_listed`) field was being reset to `True` whenever a persona was updated through the AgentEditor page, because the field was never passed through the update chain.
- Added `is_visible` to the frontend form state (`AgentEditorPage.tsx`), the frontend API interfaces and request builder (`lib.ts`), the backend Pydantic model (`PersonaUpsertRequest`), and the `create_update_persona` DB function call to `upsert_persona`.
- On edit, the form now initializes `is_visible` from the existing persona's `is_listed` value and passes it through to the backend, preserving the admin's hidden/visible setting.

## Test plan

- [ ] Create a persona and set it to hidden via `PATCH /api/admin/persona/{id}/visible` with `{"is_visible": false}`
- [ ] Edit that persona through the AgentEditor (e.g., change description) and save
- [ ] Verify that `is_listed` remains `false` in the database after the edit
- [ ] Create a new persona through the editor and verify it defaults to `is_listed=true`

---
## Summary by cubic
Fixes persona visibility resetting on edit and when the field is omitted. `is_visible` now flows end-to-end and preserves existing `is_listed` on update; new personas still default to visible. Fixes #9517.

- **Bug Fixes**
  - Added `is_visible` to AgentEditor form state and request builder.
  - Made backend `PersonaUpsertRequest.is_visible` optional; omitting it keeps the current value.
  - Passed `is_visible` to DB upsert as `is_listed`; on update uses existing when `None`, on create defaults to `True`.

<sup>Written for commit 31b0462b5987f5c4912579beb0ad8e0e2f38524b. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/db/persona.py
+++ b/backend/onyx/db/persona.py
@@ -311,6 +311,7 @@ def create_update_persona(
             search_start_date=create_persona_request.search_start_date,
             label_ids=create_persona_request.label_ids,
             is_featured=create_persona_request.is_featured,
+            is_listed=create_persona_request.is_visible,
             user_file_ids=converted_user_file_ids,
             commit=False,
             hierarchy_node_ids=create_persona_request.hierarchy_node_ids,
--- a/backend/onyx/db/persona.py
+++ b/backend/onyx/db/persona.py
@@ -921,7 +921,7 @@ def upsert_persona(
     uploaded_image_id: str | None = None,
     icon_name: str | None = None,
     display_priority: int | None = None,
-    is_listed: bool = True,
+    is_listed: bool | None = None,
     remove_image: bool | None = None,
     search_start_date: datetime | None = None,
     builtin_persona: bool = False,
@@ -1048,7 +1048,9 @@ def upsert_persona(
         if remove_image or uploaded_image_id:
             existing_persona.uploaded_image_id = uploaded_image_id
         existing_persona.icon_name = icon_name
-        existing_persona.is_listed = is_listed
+        existing_persona.is_listed = (
+            is_listed if is_listed is not None else existing_persona.is_listed
+        )
         existing_persona.search_start_date = search_start_date
         if label_ids is not None:
             existing_persona.labels.clear()
@@ -1120,7 +1122,7 @@ def upsert_persona(
             uploaded_image_id=uploaded_image_id,
             icon_name=icon_name,
             display_priority=display_priority,
-            is_listed=is_listed,
+            is_listed=is_listed if is_listed is not None else True,
             search_start_date=search_start_date,
             is_featured=(is_featured if is_featured is not None else False),
             user_files=user_files or [],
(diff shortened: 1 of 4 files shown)


kamaiel:
Thanks for taking my issue into consideration. Could we use the name "`is_listed`" in the frontend instead of "`is_visible`"?

justin-tahara:
Closing as obsolete: the is_visible clobber was resolved by the #9569 rename plus the preserve-on-None handling in #12011, so this no longer applies to main.

PR #9751, opened 2026-03-30
feat(file-upload): Upload files exceeding tokens but skip indexing

Danelegend:
## Description
This allows files to be uploaded that exceed the maximum number of tokens. The token count is necessary to index the documents. When a user uploads a file that exceeds the max number of tokens (and is allowed to be uploaded), the file is uploaded but not indexed.

This scenario is currently restricted to Tabular data files. The current workflow is that the file is uploaded, to be used by code-interpreter.

## How Has This Been Tested?
Manual + tests

## Additional Options

---
## Summary by cubic
Allow uploading tabular files that exceed the token limit; they are stored for code interpreter use, marked `SKIPPED`, and never indexed or queued. Non-exempt types over the limit are still rejected.

- **New Features**
  - Exempt `.csv`, `.tsv`, `.xlsx` from token-limit rejection; accept and mark `SKIPPED` (not queued). Non-exempt types (e.g., `.pdf`, `.txt`) over the limit are rejected.
  - Backend: `CategorizedFiles.skip_indexing`; `CategorizedFilesResult.skip_indexing_filenames` + `indexable_files`; create `UserFile` with `SKIPPED` when flagged and queue only `indexable_files` in both in-process and Celery flows. Web: add `UserFileStatus.SKIPPED` and stop tracking skipped uploads like completed.

- **Migration**
  - Add Alembic migration to include `SKIPPED` in `user_file.status` check constraint; run DB migrations.

<sup>Written for commit b085deac3b298562f4de304b3e128da5f5a63031. Summary will update on new commits.</sup>


backend/onyx/server/features/projects/projects_file_utils.py
```diff
@@ -76,11 +76,26 @@ class CategorizedFiles(BaseModel):
...
     # Allow FastAPI UploadFile instances
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
 
+# Extensions that bypass the token-count threshold on upload.
+_TOKEN_THRESHOLD_EXEMPT_EXTENSIONS: set[str] = {
+    ".csv",
+    ".tsv",
+    ".xlsx",
+}
```
justin-tahara, line 91: Hopefully this is standard but it seems like there are others as well: https://support.microsoft.com/en-us/office/file-formats-that-are-supported-in-excel-0943ff2c-6014-4e8d-aaea-b83d51d46247

justin-tahara: Just double check which file types Onyx supports and if they are on this list maybe we can add them.

backend/onyx/db/projects.py
```diff
@@ -123,6 +136,14 @@ def upload_files_to_user_files_with_indexing(
     user_files = categorized_files_result.user_files
     rejected_files = categorized_files_result.rejected_files
     id_to_temp_id = categorized_files_result.id_to_temp_id
+    indexable_files = categorized_files_result.indexable_files
+
+    # Skip-indexed files should not be processed — mark them SKIPPED immediately
+    for uf in user_files:
+        if (uf.name or "") in categorized_files_result.skip_indexing_filenames:
+            uf.status = UserFileStatus.SKIPPED
+    db_session.commit()
```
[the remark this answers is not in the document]
Subash-Mohan: This seems valid

PR #9803, opened 2026-03-31
refactor(emitter): replace bus-polling with merge-queue

nmgarza5:
## Description

Replaces the per-model event bus and polling loop in `Emitter` with a single shared `queue.Queue`, eliminating the 50ms polling sleep and giving clean backpressure.

- **`emitter.py`** — removes `Bus` / `get_default_emitter()`; adds `Emitter(merged_queue, model_idx)` that tags every packet with `model_index` via `Placement`
- **`chat_state.py`** — removes `run_chat_loop_with_state_containers` (the 113-line bus-polling loop it replaced)
- **`process_message.py`** — introduces `ChatTurnSetup` dataclass and `build_chat_turn()` setup generator; rewrites `_stream_chat_turn` + `_run_models` around the merged queue; single-model (N=1) path is fully backwards-compatible
- **`placement.py`, `override_models.py`** — docstrings; `LLMOverride` gains optional `display_name` field
- **`research_agent.py`, `custom_tool.py`** — update `Emitter` call sites to new API
- **`test_emitter.py`** — unit tests for queue routing, `model_index` tagging, placement preservation
- **`lib.tsx`, `useChatController.ts`** — parse human-readable `detail` field on non-2xx responses; fix UI hang on pre-stream 429 errors

## Additional Options

### DIFF
--- a/backend/onyx/chat/emitter.py
+++ b/backend/onyx/chat/emitter.py
@@ -30,7 +30,7 @@ class Emitter:
         self._drain_done = drain_done
 
     def emit(self, packet: Packet) -> None:
-        if self._drain_done is not None and self._drain_done.is_set():
+        if self._drain_done and self._drain_done.is_set():
             return
         base = packet.placement or Placement(turn_index=0)
         tagged = Packet(
--- a/backend/onyx/chat/process_message.py
+++ b/backend/onyx/chat/process_message.py
@@ -3,7 +3,6 @@ IMPORTANT: familiarize yourself with the design concepts prior to contributing t
 An overview can be found in the README.md file in this directory.
 """
 
-import contextvars
 import functools
 import io
 import queue
@@ -12,18 +11,16 @@ import threading
 import traceback
 from collections.abc import Callable
 from collections.abc import Generator
-from concurrent.futures import ThreadPoolExecutor
 from contextvars import Token
-from dataclasses import dataclass
 from uuid import UUID
 
-from pydantic import BaseModel
 from sqlalchemy.orm import Session
 
 from onyx.cache.factory import get_cache_backend
-from onyx.cache.interface import CacheBackend
 from onyx.chat.chat_processing_checker import set_processing_status
+from onyx.chat.chat_state import AvailableFiles
 from onyx.chat.chat_state import ChatStateContainer
+from onyx.chat.chat_state import ChatTurnSetup
 from onyx.chat.chat_utils import build_file_context
 from onyx.chat.chat_utils import convert_chat_history
 from onyx.chat.chat_utils import create_chat_history_chain
@@ -70,9 +67,7 @@ from onyx.db.chat import reserve_message_id
 from onyx.db.engine.sql_engine import get_session_with_current_tenant
 from onyx.db.enums import HookPoint
 from onyx.db.memory import get_memories
-from onyx.db.memory import UserMemoryContext
 from onyx.db.models import ChatMessage
-from onyx.db.models import ChatSession
 from onyx.db.models import Persona
 from onyx.db.models import User
 from onyx.db.models import UserFile
@@ -121,6 +116,7 @@ from onyx.tools.tool_constructor import FileReaderToolConfig
 from onyx.tools.tool_constructor import SearchToolConfig
 from onyx.utils.logger import setup_logger
 from onyx.utils.telemetry import mt_cloud_telemetry
+from onyx.utils.threadpool_concurrency import run_multiple_in_background
 from onyx.utils.timing import log_function_time
 from shared_configs.contextvars import get_current_tenant_id
 
@@ -130,21 +126,12 @@ ERROR_TYPE_CANCELLED = "cancelled"
 APPROX_CHARS_PER_TOKEN = 4
 
 
-class _AvailableFiles(BaseModel):
-    """Separated file IDs for the FileReaderTool so it knows which loader to use."""
-
-    # IDs from the ``user_file`` table (project / persona-attached files).
-    user_file_ids: list[UUID] = []
-    # IDs from the ``file_record`` table (chat-attached files).
-    chat_file_ids: list[UUID] = []
-
-
 def _collect_available_file_ids(
     chat_history: list[ChatMessage],
     project_id: int | None,
     user_id: UUID | None,
     db_session: Session,
-) -> _AvailableFiles:
+) -> AvailableFiles:
     """Collect all file IDs the FileReaderTool should be allowed to access.
 
     Returns *separate* lists for chat-attached files (``file_record`` IDs) and
@@ -171,7 +158,7 @@ def _collect_available_file_ids(
         for uf in user_files:
             user_file_ids.add(uf.id)
 
-    return _AvailableFiles(
+    return AvailableFiles(
         user_file_ids=list(user_file_ids),
         chat_file_ids=list(chat_file_ids),
     )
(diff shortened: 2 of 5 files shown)


backend/onyx/chat/emitter.py
```diff
@@ -1,19 +1,40 @@
...
+        self._model_idx = model_idx
+        self._merged_queue = merged_queue
+        self._drain_done = drain_done
 
-def get_default_emitter() -> Emitter:
-    bus: Queue[Packet] = Queue()
-    emitter = Emitter(bus)
-    return emitter
+    def emit(self, packet: Packet) -> None:
+        if self._drain_done is not None and self._drain_done.is_set():
```
evan-onyx: nit: can leave off the `is not None`

backend/onyx/chat/process_message.py
```diff
@@ -483,525 +491,843 @@ def _resolve_query_processing_hook_result(
...
+    if not new_msg_req.chat_session_id:
+        if not new_msg_req.chat_session_info:
+            raise RuntimeError("Must specify a chat session id or chat session info")
+        chat_session = create_chat_session_from_request(
+            chat_session_request=new_msg_req.chat_session_info,
+            user_id=user_id,
+            db_session=db_session,
+        )
+        yield CreateChatSessionID(chat_session_id=chat_session.id)
+        chat_session = get_chat_session_by_id(
```
evan-onyx: I see you just moved this code, my comment(s) below are about a previously existing inefficiency... would be v happy if u could clean it up 🙇

nit: it's a bit cleaner imo to set a variable like `chat_session_id` here and in the else case, then call get_chat_session_id() after the if/else with that value since the other parameters are the same

evan-onyx: actually on second thought is there a reason we need to make this call in this branch? I would think the session already has the chat session object loaded since we just created it. If not then my above comment applies but it might just be valid to remove this call entirely

backend/onyx/chat/process_message.py
```diff
@@ -483,525 +491,843 @@ def _resolve_query_processing_hook_result(
...
+    check_is_connected: Callable[[], bool]
+    cache: CacheBackend
+    # Execution params forwarded to per-model tool construction
+    bypass_acl: bool
+    slack_context: SlackContext | None
+    custom_tool_additional_headers: dict[str, str] | None
+    mcp_headers: dict[str, str] | None
+
+
+def build_chat_turn(
```
evan-onyx, line 481: I like this refactoring, but do you think it would be reasonable to put this in chat_state.py to reduce the size of this file? Just a suggestion, I'm fine with either (imo the tradeoff is just co-location of related code vs having to scroll through large files)

backend/onyx/chat/process_message.py
```diff
@@ -483,525 +491,843 @@ def _resolve_query_processing_hook_result(
...
-    # Optional external state container for non-streaming access to accumulated state
-    external_state_container: ChatStateContainer | None = None,
-) -> AnswerStream:
-    tenant_id = get_current_tenant_id()
-    mock_response_token: Token[str | None] | None = None
+    # Additional context to include in the chat history, e.g. Slack threads where the
+    # conversation cannot be represented by a chain of User/Assistant messages.
+    # NOTE: not stored in the database, only passed in to the LLM as context
+    additional_context: str | None = None,
+) -> Generator[AnswerStreamPart, None, ChatTurnSetup]:
```
evan-onyx, line 496: no need to change this (definitely good enough for now) but it's probably nicer to break this up even further into various pieces of setup logic separated by the answerpart yields. effectively my ideal world looks like

```
# inside stream_chat_turn
setup1 = setup_part1()
yield setup1.first_thing()
setup2 = setup_part2(setup1)
yield setup2.second_thing()
# ...
```
effectively moving all the yields to stream_chat_turn. Reason being this pattern (generator returning stuff) isn't super common and standard functions are typically just easier to think about. Leave a TODO if you agree! or if not we can yap about it

backend/onyx/chat/process_message.py
```diff
@@ -483,525 +491,843 @@ def _resolve_query_processing_hook_result(
...
-        tool_id_to_name_map = {tool.id: tool.name for tool in all_tools}
+    merged_queue: queue.Queue[tuple[int, Packet | Exception | object]] = queue.Queue()
 
-        search_tool_id = next(
-            (tool.id for tool in all_tools if tool.in_code_tool_id == SEARCH_TOOL_ID),
-            None,
+    state_containers: list[ChatStateContainer] = [
+        (
+            external_state_container
+            if (external_state_container is not None and i == 0)
```
evan-onyx, line 942: why does only the first model get the state container that was passed in?

backend/onyx/chat/process_message.py
```diff
@@ -483,525 +491,843 @@ def _resolve_query_processing_hook_result(
...
+                if setup.forced_tool_id and setup.forced_tool_id not in {
+                    tool.id for tool in model_tools
+                }:
+                    raise ValueError(
+                        f"Forced tool {setup.forced_tool_id} not found in tools"
+                    )
 
-        emitter = get_default_emitter()
+                # Per-thread copy: run_llm_loop mutates simple_chat_history in-place.
+                if n_models == 1 and setup.new_msg_req.deep_research:
```
evan-onyx, line 1011: no deep research with multi-llm?

backend/onyx/chat/process_message.py
```diff
@@ -483,525 +483,782 @@ def _resolve_query_processing_hook_result(
...
+                                db_session=db_session,
+                                assistant_message=setup.reserved_messages[i],
+                                llm=setup.llms[i],
+                                reserved_tokens=setup.reserved_token_count,
+                            )
+                        except Exception:
+                            logger.exception(f"Failed completion for model {i} on disconnect ({setup.model_display_names[i]})")
+                    yield Packet(
+                        placement=Placement(turn_index=0),
+                        obj=OverallStop(type="stop", stop_reason="user_cancelled"),
```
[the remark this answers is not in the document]
evan-onyx: seems valid?

backend/onyx/chat/process_message.py
```diff
@@ -483,525 +491,843 @@ def _resolve_query_processing_hook_result(
...
+                logger.exception(
+                    "model %d (%s): self-completion after disconnect failed",
+                    model_idx,
+                    setup.model_display_names[model_idx],
+                )
 
-        # TODO Need to think of some way to support selected docs from the sidebar
+    # Copy contextvars before submitting futures — ThreadPoolExecutor does NOT
+    # auto-propagate contextvars in Python 3.11; threads would inherit a blank context.
+    ctx = contextvars.copy_context()
```
evan-onyx: ideally we should add a utility to the concurrency utils in threadpool_concurrency.py for this case, something like `run_multiple_in_background`.

backend/onyx/chat/process_message.py
```diff
@@ -483,525 +491,843 @@ def _resolve_query_processing_hook_result(
...
-            message_type=MessageType.ASSISTANT,
-        )
+        # ── Main thread: merge and yield packets ────────────────────────────
+        models_remaining = n_models
+        while models_remaining > 0:
+            try:
+                model_idx, item = merged_queue.get(timeout=0.05)
+            except queue.Empty:
+                # Check for user-initiated cancellation every 50 ms.
+                if not setup.check_is_connected():
```
evan-onyx: nit: prefer
```
if setup.check_is_connected():
    continue
```
to reduce indentation

backend/onyx/chat/process_message.py
```diff
@@ -483,525 +491,843 @@ def _resolve_query_processing_hook_result(
...
+                        error_msg = error_msg.replace(
+                            model_llm.config.api_key, "[REDACTED_API_KEY]"
+                        )
+                        stack_trace = stack_trace.replace(
+                            model_llm.config.api_key, "[REDACTED_API_KEY]"
+                        )
+                    yield StreamingError(
+                        error=error_msg,
+                        stack_trace=stack_trace,
+                        error_code="MODEL_ERROR",
```
evan-onyx, line 1152: ideally this comes from an enum but iiwii

backend/onyx/chat/process_message.py
```diff
@@ -1073,30 +1395,55 @@ def llm_loop_completion_callback(
...
+    litellm_additional_headers: dict[str, str] | None = None,
+    custom_tool_additional_headers: dict[str, str] | None = None,
+    mcp_headers: dict[str, str] | None = None,
+    bypass_acl: bool = False,
+    additional_context: str | None = None,
+    slack_context: SlackContext | None = None,
+    external_state_container: ChatStateContainer | None = None,
+) -> AnswerStream:
+    """Single-model streaming entrypoint. Delegates to ``_stream_chat_turn``."""
+    yield from _stream_chat_turn(
```
evan-onyx: is _stream_chat_turn used elsewhere? if not then we can just put it's body here I imagine? since it seems like this just wraps it

backend/onyx/chat/process_message.py
nmgarza5: im a fan

backend/onyx/chat/process_message.py
nmgarza5: nah. was explicitly listed as something we wont support

PR #9807, opened 2026-03-31
feat(canvas): checkpoint logic (3/4)

benwu408:
## Description

Add _load_from_checkpoint with staged processing (pages → assignments → announcements) per course, time-window filtering, per-document failure isolation via ConnectorFailure, and proper checkpoint state advancement. Security-critical pagination errors (host/scheme mismatch) propagate while recoverable API errors (404, 429, 5xx) trigger retries. Implements load_from_checkpoint, load_from_checkpoint_with_perm_sync, build_dummy_checkpoint, and validate_checkpoint_json.

Includes unit tests for checkpoint lifecycle, stage advancement, time filtering, failure handling, and perm sync attachment.

## How Has This Been Tested?

Unit tests covering checkpoint lifecycle: first-call course materialization, stage processing (pages → assignments → announcements), stage advancement across all 3 stages, time-window filtering, announcement skip on missing posted_at, stage failure retry (no advancement), per-document conversion failure yielding ConnectorFailure, terminal state (has_more=False), invalid stage rejection, and load_from_checkpoint_with_perm_sync attaching ExternalAccess. 

Also manually tested against a live Canvas instance.

## PR Stack

**Stacked PRs on my fork (each targets the previous branch for isolated review):**
1. [#1 — Scaffold: API client, data models, connector registration](https://github.com/benwu408/onyx/pull/1)
2. [#2 — Connector data fetching and conversion](https://github.com/benwu408/onyx/pull/2)
3. [#3 — Checkpoint-based indexing](https://github.com/benwu408/onyx/pull/3)
4. [#4 — Permission sync and frontend registration](https://github.com/benwu408/onyx/pull/4)

**PRs on upstream onyx (all target main):**
1. [#9385 — Scaffold](https://github.com/onyx-dot-app/onyx/pull/9385)
2. [#9386 — Connector data](https://github.com/onyx-dot-app/onyx/pull/9386)
3. [#9387 — Checkpoint](https://github.com/onyx-dot-app/onyx/pull/9387)
4. [#9388 — Perm sync + frontend](https://github.com/onyx-dot-app/onyx/pull/9388)

---
## Summary by cubic
Adds checkpoint-based indexing to the Canvas connector with staged per-course pagination, strict time-window filtering, and resumable progress. Improves error handling so auth/permission failures surface, stage fetch errors yield failures and auto-advance (404 skips the course), and per-item conversion errors are isolated.

- **New Features**
  - Staged per-course processing (pages → assignments → announcements) with `next_url` resume and automatic advancement via `advance_stage()`; shared `_load_from_checkpoint` powers both load paths.
  - Time filtering: pages sorted by `updated_at` with early exit; assignments filtered by `updated_at`; announcements filtered server-side using `start_date`/`end_date`.
  - Permission sync attaches course-level `ExternalAccess`; adds helpers for time parsing/formatting and window checks; implements `build_dummy_checkpoint` and `validate_checkpoint_json`.

- **Bug Fixes**
  - Stage fetch errors now yield a `ConnectorFailure` and auto-advance; 404 skips the whole course; unknown errors skip the stage to ensure progress.
  - 401/403 surface during course discovery and stage fetches; security-critical pagination URL mismatches still raise; invalid stage now raises a `ValueError` listing valid stages.
  - Removed unreachable 429 handling (handled by `rl_requests`).

<sup>Written for commit c881a400256780b9a5b22ec40329a16f732dca44. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/connectors/canvas/connector.py
+++ b/backend/onyx/connectors/canvas/connector.py
@@ -231,7 +231,7 @@ class CanvasConnectorCheckpoint(ConnectorCheckpoint):
         the next call starts fresh on the new stage.
         """
         self.next_url = None
-        stages = list(CanvasStage)
+        stages: list[CanvasStage] = list(CanvasStage)
         next_idx = stages.index(self.stage) + 1
         if next_idx < len(stages):
             self.stage = stages[next_idx]
--- a/backend/tests/unit/onyx/connectors/canvas/test_canvas_connector.py
+++ b/backend/tests/unit/onyx/connectors/canvas/test_canvas_connector.py
@@ -23,6 +23,7 @@ from onyx.connectors.exceptions import UnexpectedValidationError
 from onyx.connectors.models import ConnectorFailure
 from onyx.connectors.models import ConnectorMissingCredentialError
 from onyx.connectors.models import Document
+from onyx.connectors.models import HierarchyNode
 from onyx.error_handling.error_codes import OnyxErrorCode
 from onyx.error_handling.exceptions import OnyxError
 
@@ -153,10 +154,12 @@ def _run_checkpoint(
     checkpoint: CanvasConnectorCheckpoint,
     start: float = 0.0,
     end: float = datetime(2099, 1, 1, tzinfo=timezone.utc).timestamp(),
-) -> tuple[list[Document | ConnectorFailure], CanvasConnectorCheckpoint]:
+) -> tuple[
+    list[Document | HierarchyNode | ConnectorFailure], CanvasConnectorCheckpoint
+]:
     """Run load_from_checkpoint once and collect yielded items + returned checkpoint."""
     gen = connector.load_from_checkpoint(start, end, checkpoint)
-    items: list[Document | ConnectorFailure] = []
+    items: list[Document | HierarchyNode | ConnectorFailure] = []
     new_checkpoint: CanvasConnectorCheckpoint | None = None
     try:
         while True:
@@ -1426,7 +1429,7 @@ class TestLoadFromCheckpointWithPermSync:
         end = datetime(2025, 6, 30, tzinfo=timezone.utc).timestamp()
 
         gen = connector.load_from_checkpoint_with_perm_sync(start, end, cp)
-        items: list[Document | ConnectorFailure] = []
+        items: list[Document | HierarchyNode | ConnectorFailure] = []
         new_cp: CanvasConnectorCheckpoint | None = None
         try:
             while True:


backend/onyx/connectors/canvas/connector.py
```diff
@@ -60,6 +62,25 @@ def _handle_canvas_api_error(e: OnyxError) -> NoReturn:
             f"Canvas API error (status={e.status_code}): {e}"
         )
 
+def _handle_canvas_api_error(e: OnyxError) -> NoReturn:
```
jasmine-wu-ai: duplicate of function above

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +421,189 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
-        """Validate Canvas connector settings by testing API access."""
+    def _load_from_checkpoint(
+        self,
+        start: SecondsSinceUnixEpoch,
+        end: SecondsSinceUnixEpoch,
+        checkpoint: CanvasConnectorCheckpoint,
+        include_permissions: bool = False,
+    ) -> CheckpointOutput[CanvasConnectorCheckpoint]:
+        """Shared implementation for load_from_checkpoint and load_from_checkpoint_with_perm_sync."""
+        new_checkpoint = checkpoint.model_copy(deep=True)
```
jasmine-wu-ai, line 585: is model_copy just making a copy of the object?

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +421,189 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+
+        # First call: materialize the list of course IDs
+        if not new_checkpoint.course_ids:
+            try:
+                courses = self._list_courses()
```
jasmine-wu-ai: maybe i don't have a good understanding of what exactly "checkpoint" means - but from reading this i'd be concerned that we're just using the old course_ids if they exist - what if someone joins / drops a course since we last checkpoitned?

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +421,189 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+            except Exception as e:
+                logger.warning(f"Failed to list Canvas courses: {e}")
+                new_checkpoint.has_more = True
```
jasmine-wu-ai: what does has_more mean?

backend/onyx/connectors/canvas/connector.py
benwu408: lol

backend/onyx/connectors/canvas/connector.py
benwu408: yep, makes a copy that we modify then return. its a pydantic method

backend/onyx/connectors/canvas/connector.py
benwu408: This lowkey confused me and I had to spend some time understanding it, but basically indexing is like every minute or whatever when I sorts thru everything and finds new stuff. every indexing run makes a new empty checkpoint, calls list courses, then finds every single document and yields the ones in the window since the last indexing run. checkpoints are made at every single stage in the indexing process, each load from checkpoint run processes a page(at most 100) of things, and then returns a new checkpoint that is then called. So basically every new index you'd get the full course list again.

backend/onyx/connectors/canvas/connector.py
benwu408: the current indexing run still has more docs to go thru

jasmine-wu-ai: why is it only true if we fail to list course ids?

backend/onyx/connectors/canvas/connector.py
jasmine-wu-ai: oh so the checkpoint doesn't persist across calls for indexing? if it is completely reinitialized everytime we index, whats the purpose of it?

jasmine-wu-ai: maybe we're getting to the part where justin needs to review the code 🫠

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +402,223 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+                return new_checkpoint
+            new_checkpoint.course_ids = [c.id for c in courses]
+            new_checkpoint.current_course_index = 0
+            new_checkpoint.stage = "pages"
```
jasmine-wu-ai: aren't these the defaults? shouldn't need to set again here

backend/onyx/connectors/canvas/connector.py
jasmine-wu-ai: oh i see its also set below

backend/onyx/connectors/canvas/connector.py
jasmine-wu-ai: honestly - it would be cleaner for `CanvasConnectorCheckpoint` to have a real init method, generally you shouldn't be directly setting the member variables

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +402,223 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+            logger.info(
+                f"Found {len(courses)} Canvas courses to process"
+            )
+            new_checkpoint.has_more = len(new_checkpoint.course_ids) > 0
+            return new_checkpoint
```
jasmine-wu-ai, line 601: seems like on the first call all we do is get the course ids? why not also proceed below?

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +402,223 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+        ):
+            new_checkpoint.has_more = False
+            return new_checkpoint
+
+        course_id = new_checkpoint.course_ids[
+            new_checkpoint.current_course_index
+        ]
+        stage = new_checkpoint.stage
+
+        if stage not in ("pages", "assignments", "announcements"):
```
jasmine-wu-ai: this shouldn't be hardcoded. you should create an enum for all stages and reference them. here's a basic example i pulled from claude. i think enum values can be strings too

```
class Color(Enum):
    RED = 1
    GREEN = 2
    BLUE = 3

    @classmethod
    def values(cls):
        return [e.value for e in cls]

Color.values()  # [1, 2, 3]

print(Color.RED)
```

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +402,223 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+            raise ValueError(
+                f"Invalid checkpoint stage: {stage!r}"
+            )
+
+        def _in_time_window(timestamp_str: str) -> bool:
```
jasmine-wu-ai: i dont know why AI likes to create nested functions but its weird - just put it outside the class if it doesn't need to reference the class state

jasmine-wu-ai: this should take in timestamp_str, start, end

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +402,223 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+            ts = (
+                datetime.fromisoformat(timestamp_str.replace("Z", "+00:00"))
+                .astimezone(timezone.utc)
+                .timestamp()
+            )
+            return start < ts <= end
+
+        def _maybe_attach_permissions(document: Document) -> Document:
```
jasmine-wu-ai: same here - but this looks like it could be a member of the class. just doesn't need to be nested

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +402,223 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+            if include_permissions:
+                document.external_access = self._get_course_permissions(
+                    course_id
+                )
+            return document
+
+        # Fetch one page of API results for the current stage.
```
jasmine-wu-ai: the entire part of "fetching api results" should be pulled into a helper

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +402,223 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+        # If next_url is set, we're resuming mid-pagination.
+        start_dt = datetime.fromtimestamp(start, tz=timezone.utc)
+        end_dt = datetime.fromtimestamp(end, tz=timezone.utc)
+        stage_config: dict[str, dict[str, Any]] = {
```
jasmine-wu-ai: i think you should be able to pull this out into a constant (put it at the top of the file) and format the template here. example from claude

```
template = "Hello, {name}! You have {count} messages."

result = template.format(name="Alice", count=5)
print(result)  # "Hello, Alice! You have 5 messages."
```

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +402,223 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
-            raise UnexpectedValidationError(
-                f"Unexpected error during Canvas settings validation: {exc}"
+            if new_checkpoint.next_url:
+                response, result_next_url = self.canvas_client.get(
+                    full_url=new_checkpoint.next_url
+                )
+            else:
+                response, result_next_url = self.canvas_client.get(
+                    config["endpoint"], params=config["params"]
+                )
```
jasmine-wu-ai: am i understanding right - so if this is the first call, we need to pass config endpoint & params? otherwise the next url will have all that info?

we should add code comment explaining the difference here

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +421,189 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+            # Security errors have no HTTP status code override (they are
+            # raised locally, not from an API response).
+            is_api_error = oe._status_code_override is not None
+            if not is_api_error:
+                raise
+            logger.warning(
+                f"Failed to fetch {stage} for course {course_id}: {oe}"
             )
+            new_checkpoint.has_more = True
+            return new_checkpoint
```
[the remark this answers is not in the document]
jasmine-wu-ai: yeah this is weird. also - why don't we want to unconditionally throw error for OnyxError and regular error? seems kinda weird

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +402,223 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+                f"Failed to fetch {stage} for course {course_id}: {e}"
             )
+            new_checkpoint.has_more = True
+            return new_checkpoint
+
+        # Process fetched items
+        early_exit = False
+        for item in response or []:
+            try:
+                if stage == "pages":
```
jasmine-wu-ai: already mentioned above - but we should not have hardcoded strings - replace w enum

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +402,223 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+        except Exception as e:
+            logger.warning(
+                f"Failed to fetch {stage} for course {course_id}: {e}"
             )
+            new_checkpoint.has_more = True
+            return new_checkpoint
+
+        # Process fetched items
+        early_exit = False
+        for item in response or []:
```
jasmine-wu-ai: this line of code is a little weird 😆 
whats do u actually want to happen if you get an empty response?

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +402,223 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+                    # an item at or before `start`, all remaining items
+                    # on this and subsequent pages are older too.
+                    if not _in_time_window(page.updated_at):
+                        ts = (
+                            datetime.fromisoformat(
+                                page.updated_at.replace("Z", "+00:00")
+                            )
+                            .astimezone(timezone.utc)
+                            .timestamp()
+                        )
```
jasmine-wu-ai: these lines of code are kinda weird. at minimum lets pull into a helper and explain why we need to do so much on the timestamp (ex. what is canvas returning to use? also - can we do this processing within CanvasPage)

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +402,223 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+            except Exception as e:
+                item_id = item.get("id") or item.get("page_id", "unknown")
+                if stage == "pages":
+                    doc_link = (
+                        f"{self.canvas_base_url}/courses/{course_id}"
+                        f"/pages/{item.get('url', '')}"
+                    )
+                else:
+                    doc_link = item.get("html_url", "")
+                yield ConnectorFailure(
```
jasmine-wu-ai, line 654: what does it mean to yield a failure? why not just raise an error?

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +402,223 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+        if early_exit:
+            result_next_url = None
+
+        # If there are more pages, save the cursor and return
+        if result_next_url:
+            new_checkpoint.next_url = result_next_url
+        else:
+            # Stage complete — advance to next stage
+            new_checkpoint.next_url = None
+            next_stages: dict[str, CanvasStage | None] = {
```
jasmine-wu-ai: lol - there should be a cleaner way to do this. just have the order in a list and find the element and do idx + 1, and if that is >= length then exit.

its not necessarily less code but less error prone if/when the code changes

backend/onyx/connectors/canvas/connector.py
benwu408: got rid of the first part, but no one else uses an init method and they just use defaults and then set vars.

backend/onyx/connectors/canvas/connector.py
benwu408: just adds another checkpoint. I mirrored from others and they all return early

backend/onyx/connectors/canvas/connector.py
benwu408: done

backend/onyx/connectors/canvas/connector.py
benwu408: ok I will add. u are understanding right yes

backend/onyx/connectors/canvas/connector.py
benwu408: done

backend/onyx/connectors/canvas/connector.py
benwu408: replaced with that.

backend/onyx/connectors/canvas/connector.py
benwu408: it was because response could've returned None and then I'd get a type error so if it was none I set to []. but I made the helper func and then I made it return [].

backend/onyx/connectors/canvas/connector.py
benwu408: they return epoch seconds which is like seconds since 1970 lmao. I put it in a helper. I could do it within each canvas___ but then it'd be kinda redundant, I feel like its fine as a helper.

backend/onyx/connectors/canvas/connector.py
benwu408: all connectors just raise cause one bad document shouldn't raise an error and mess up an entire indexing batch. the yielded error logs which doc failed and adds to a list of failed docs that don't get processed for embedding and stuff

backend/onyx/connectors/canvas/connector.py
benwu408: wait why did I do it like that thats so weird 😂

backend/onyx/connectors/canvas/connector.py
```diff
@@ -61,6 +62,51 @@ def _handle_canvas_api_error(e: OnyxError) -> NoReturn:
...
+    },
+    CanvasStage.ANNOUNCEMENTS: {
+        "endpoint": "announcements",
+        "params": {
+            "per_page": "100",
+            "context_codes[]": "course_{course_id}",
+            "active_only": "true",
+        },
+    },
+}
```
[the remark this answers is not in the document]
jasmine-wu-ai: this is resolved right?

backend/onyx/connectors/canvas/connector.py
```diff
@@ -61,6 +62,57 @@ def _handle_canvas_api_error(e: OnyxError) -> NoReturn:
         )
 
 
+class CanvasStage(StrEnum):
```
jasmine-wu-ai, line 62: since its hard to merge code from the fork i wont make you do this - but ideally, the changes to switch to CanvasStage enum & creating `_parse_canvas_dt` helper changed enough lines they should go in their own PR

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +439,231 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+            if next_url:
+                # Resuming mid-pagination: the next_url from Canvas's
+                # Link header already contains endpoint + query params.
+                response, result_next_url = self.canvas_client.get(
+                    full_url=next_url
+                )
+            else:
+                # First request for this stage: build from endpoint + params.
+                response, result_next_url = self.canvas_client.get(
+                    endpoint, params=params
```
jasmine-wu-ai: nit - i'd add this, since all the args to this function are technically optional
```suggestion
                    endpoint=endpoint, params=params
```

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +439,231 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+                )
+        except OnyxError as oe:
+            # Re-raise security errors from _parse_next_link (host/scheme
+            # mismatch on pagination URLs) — these must not be silenced.
+            is_api_error = oe._status_code_override is not None
+            if not is_api_error:
+                raise
+            if oe.status_code in (401, 403):
```
jasmine-wu-ai: is it only exactly 401/403? can we filter on anything in 400s?

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +439,231 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+                )
+            else:
+                # First request for this stage: build from endpoint + params.
+                response, result_next_url = self.canvas_client.get(
+                    endpoint, params=params
+                )
+        except OnyxError as oe:
+            # Re-raise security errors from _parse_next_link (host/scheme
+            # mismatch on pagination URLs) — these must not be silenced.
+            is_api_error = oe._status_code_override is not None
```
jasmine-wu-ai: i don't fully understand why you need to do this, but we really shouldn't be reaching into the internal variables here. i did a code search for `_status_code_override` and its only used in the internal exception class ([link](https://github.com/search?q=repo%3Abenwu408%2Fonyx%20_status_code_override&type=code))

backend/onyx/connectors/canvas/connector.py
jasmine-wu-ai: i still don't really follow - if we fail to list course IDs why are we continuing?? what is there for us to do??

backend/onyx/connectors/canvas/connector.py
jasmine-wu-ai: is this a correct understanding - every time you start a new indexing run you create a brand new series of checkpoints? it seems like checkpoints are more for progress tracking - whats the functional purpose of them?

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +439,231 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+                new_checkpoint.has_more = True
+                return new_checkpoint
+            new_checkpoint.course_ids = [c.id for c in courses]
+            logger.info(
+                f"Found {len(courses)} Canvas courses to process"
+            )
+            new_checkpoint.has_more = len(new_checkpoint.course_ids) > 0
+            return new_checkpoint
+
+        # All courses done
```
jasmine-wu-ai:
```suggestion
        # All courses done.
```

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +439,231 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+            return new_checkpoint
+
+        course_id = new_checkpoint.course_ids[
+            new_checkpoint.current_course_index
+        ]
+        stage = new_checkpoint.stage
+
+        if stage not in CanvasStage:
+            raise ValueError(
+                f"Invalid checkpoint stage: {stage!r}"
```
jasmine-wu-ai: its easy to find, but still helpful to log the valid CanvasStage values here

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +439,231 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+
+        # Build endpoint + params from the static template.
+        config = _STAGE_CONFIG[stage]
+        endpoint = config["endpoint"].format(course_id=course_id)
+        params = {
+            k: v.format(course_id=course_id)
+            for k, v in config["params"].items()
+        }
+        if stage == CanvasStage.ANNOUNCEMENTS:
+            start_dt = datetime.fromtimestamp(start, tz=timezone.utc)
```
jasmine-wu-ai: i think this can also be a lightweight helper to create datetime & reformat str - something like `convertUnixTimeToCanvasTime` (if thats correct). then you can just do
```
params["start_date"] = convertUnixTimeToCanvasTime(start)
params["end_date"] = convertUnixTimeToCanvasTime(end)
```

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +439,231 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+            )
+
+        # Build endpoint + params from the static template.
+        config = _STAGE_CONFIG[stage]
+        endpoint = config["endpoint"].format(course_id=course_id)
+        params = {
+            k: v.format(course_id=course_id)
+            for k, v in config["params"].items()
+        }
+        if stage == CanvasStage.ANNOUNCEMENTS:
```
jasmine-wu-ai, line 626: why do you only populate these for announcements?

jasmine-wu-ai: (leave answer as code comment)

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +439,231 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+            return new_checkpoint
+        except Exception:
+            new_checkpoint.has_more = True
+            return new_checkpoint
+
+        # Process fetched items
+        early_exit = False
+        for item in response:
+            try:
+                if stage == CanvasStage.PAGES:
```
jasmine-wu-ai, line 493: i think all this can be a helper (convertItemToDocument(item)) which just replaces this bug if elif else block.

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +439,231 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+                    page = CanvasPage.from_api(item, course_id=course_id)
+                    if not page.updated_at:
+                        continue
+                    # Pages are sorted by updated_at desc — once we see
+                    # an item at or before `start`, all remaining items
+                    # on this and subsequent pages are older too.
+                    if not _in_time_window(page.updated_at, start, end):
+                        if _parse_canvas_dt(page.updated_at).timestamp() <= start:
+                            early_exit = True
```
jasmine-wu-ai, line 502: why is early_exit only set for pages? why not assignments / announcements

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +439,231 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+        # If there are more pages, save the cursor and return
+        if result_next_url:
+            new_checkpoint.next_url = result_next_url
+        else:
+            # Stage complete — advance to next stage
+            new_checkpoint.next_url = None
+            stages = list(CanvasStage)
+            idx = stages.index(stage)
+            if idx + 1 < len(stages):
+                new_checkpoint.stage = stages[idx + 1]
```
jasmine-wu-ai: nit
```suggestion
            next_idx = stages.index(stage) + 1
            if next_idx < len(stages):
                new_checkpoint.stage = stages[next_idx]
```

jasmine-wu-ai: and we should either move this get next stage stuff to a helper, or add a comment about it

backend/onyx/connectors/canvas/connector.py
benwu408: ok

backend/onyx/connectors/canvas/connector.py
benwu408: added. technically optional but not really for first request

backend/onyx/connectors/canvas/connector.py
benwu408: anything else we should either retry or just log for, but 401 and 403 are like bad credentials or token and those are the only ones we should raise error.

benwu408: non http errors we raise for though

backend/onyx/connectors/canvas/connector.py
benwu408: ok so for this status code override is basically when I get an actual http error from canvas. the other times are like invalid json returned, improper link, etc. but for onyx error I set those cases to error codes so status code over ride tells me whether it was an api error or a different sort of error. I have an overlap in an actual api error and a set api error(502), but I could either change that(but I feel like bad gateway is a good description of my error - the non api one), or just keep this with the status code override. the issue is that onyx error class doesn't have a method to be like get_status_code_override or something and there's not much precedent since no other connectors use onyx error and Justin thought we should implement it here since they are trying to shift to using it.

backend/onyx/connectors/canvas/connector.py
benwu408: okie doke

backend/onyx/connectors/canvas/connector.py
benwu408: added

backend/onyx/connectors/canvas/connector.py
benwu408: done

backend/onyx/connectors/canvas/connector.py
benwu408: the announcements api call lets u filter what it returns by start and end, pages lets u sort, assignments lets u do neither. commented

backend/onyx/connectors/canvas/connector.py
benwu408: pages is sorted so we leave early for it. added comment. other two cant be sorted

backend/onyx/connectors/canvas/connector.py
benwu408: ok I did it as a helper that returns all docs and then I go thru and yield all of them.

backend/onyx/connectors/canvas/connector.py
benwu408: added comment

backend/onyx/connectors/canvas/connector.py
benwu408: we still wanna try again for transient errors. we throw error for 401/3 but if its something like the network is slow or whatever it tries again. and then the place where the actual checkpoints are called handles for if a certain amount of calls don't get anywhere.

backend/onyx/connectors/canvas/connector.py
benwu408: yes. to track progress. a checkpoint is a single new like fetch during a bigger indexing run. a brand new empty checkpoint is initialized every indexing run.

backend/onyx/connectors/canvas/connector.py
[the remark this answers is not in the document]
benwu408: raise on 401/403

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +444,261 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+        course_id: int,
+        start: float,
+        end: float,
+        include_permissions: bool,
+    ) -> tuple[list[Document | ConnectorFailure], bool]:
+        """Process a page of API results into documents.
+
+        Returns (docs, early_exit). early_exit is True when pages
+        (sorted desc by updated_at) hit an item older than start,
+        signalling that pagination should stop.
```
justin-tahara: Are you British @benwu408?

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +444,261 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+            except OnyxError as e:
+                if e.status_code in (401, 403):
+                    _handle_canvas_api_error(e)
+                logger.warning(f"Failed to list Canvas courses: {e}")
+                new_checkpoint.has_more = True
+                return new_checkpoint
+            except Exception as e:
+                logger.warning(f"Failed to list Canvas courses: {e}")
+                new_checkpoint.has_more = True
+                return new_checkpoint
```
justin-tahara: When you do hit an exception, you try to continue and you may continuously fall into an infinite loop of failed attempts here. 

Ideally you are updating the `stage`, `next_url`, or `current_course_index` so that the next run is different. 

I believe in the current implementation, you may hit an issue and you would keep retrying it infinitely.

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +444,261 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+                f"Failed to fetch {stage} for course {course_id}: {oe}"
+            )
+            new_checkpoint.has_more = True
+            return new_checkpoint
+        except Exception as e:
+            logger.warning(
+                f"Failed to fetch {stage} for course {course_id}: {e}"
+            )
+            new_checkpoint.has_more = True
+            return new_checkpoint
```
justin-tahara, line 700: Similar to the other claim I think we need to do some updates on the fields I mentioned so that you don't enter an infinite loop of retries

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +444,261 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+            is_api_error = oe._status_code_override is not None
+            if not is_api_error:
+                raise
+            if oe.status_code in (401, 403):
+                _handle_canvas_api_error(oe)
+            logger.warning(
+                f"Failed to fetch {stage} for course {course_id}: {oe}"
+            )
+            new_checkpoint.has_more = True
+            return new_checkpoint
```
justin-tahara, line 680: I think if you can't find a course on the stage and see for example a 404, we might want to handle that differently so that you move onto the next course.

backend/onyx/connectors/canvas/connector.py
justin-tahara, line 328: Do we not use these functions anymore?

justin-tahara: A bit odd we are doing ConnectorValidationError. I think we need like a rate limit error since right now we are hitting a rate limit and then serving a credential error

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +444,261 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
+                    doc_link = (
+                        f"{self.canvas_base_url}/courses/{course_id}"
+                        f"/pages/{item.get('url', '')}"
+                    )
+                else:
+                    doc_link = item.get("html_url", "")
+                results.append(
+                    ConnectorFailure(
+                        failed_document=DocumentFailure(
+                            document_id=f"canvas-{stage.removesuffix('s')}-{course_id}-{item_id}",
```
justin-tahara, line 557: Any reason for the removal of the suffix here? Maybe I'm missing something

backend/onyx/connectors/canvas/connector.py
```diff
@@ -400,20 +444,261 @@ def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None
...
-        except ConnectorMissingCredentialError:
-            raise
-        except Exception as exc:
-            raise UnexpectedValidationError(
-                f"Unexpected error during Canvas settings validation: {exc}"
+            response, result_next_url = self._fetch_stage_page(
+                next_url=new_checkpoint.next_url,
+                endpoint=endpoint,
+                params=params,
             )
```
justin-tahara, line 635: Is it ever possible that the `result_next_url` returns a page that it has already seen before? I don't want to get into a cycle

backend/onyx/connectors/canvas/connector.py
benwu408: 😂

backend/onyx/connectors/canvas/connector.py
benwu408: ok for this one I changed to just let exceptions propagate.

backend/onyx/connectors/canvas/connector.py
benwu408: for this one I just have it go to the next stage, so like from pages to assignments in a course, if it hits an error since we can't get the next url without the current one. not too big of a problem since itll retry in like 15 secs or whatever when the next indexing run happens.

backend/onyx/connectors/canvas/connector.py
benwu408: used in branch 4 for perm sync

benwu408: so actually we can never even get a 429 error code because in the get request we use rl_requests.get which waits and retries on 429. I deleted the line and even if somehow we did get a 429 it'd hit the else block and still raise the same connector validation error

backend/onyx/connectors/canvas/connector.py
benwu408: canvas stage in the url uses page/announcement/assignment, we have it as plural in the stages, so remove s at end

backend/onyx/connectors/canvas/connector.py
benwu408: next url comes from canvas so we are trusting that canvas wouldn't return a next url that is the same. would u want to add maybe like an if cur url == next_url?

backend/onyx/connectors/canvas/connector.py
justin-tahara: I guess it's fine but it's just a bit confusing doing some naming updates in the setting of the document id. 

I think it's fine if it has the s in my opinion but minor nit

backend/onyx/connectors/canvas/connector.py
justin-tahara: I would try to understand the API and what it returns but ideally you could maintain a set and ensure that you are not going and creating a cycle

PR #9830, opened 2026-04-01
refactor(opal): split `SelectCard`'s `sizeVariant` prop into `paddingVariant` + `roundingVariant`

raunakab:
## Description

Splits the `SelectCard` `sizeVariant` prop into independent `paddingVariant` and `roundingVariant` props, matching the `Card` API introduced in #9823. Also hardcodes the `Interactive.Stateful` variant to `select-card` internally and removes it from the public props (it was always `select-card` at every callsite).

- `paddingVariant`: `PaddingVariants` (default `"sm"`) — uses the same scale as `Card` (`p-6` → `p-0`)
- `roundingVariant`: `RoundingVariants` (default `"lg"`) — uses the same scale as `Card` (`rounded-04` → `rounded-16`)
- All callsites updated to remove redundant `variant="select-card"` and `sizeVariant="lg"` (matches new defaults)

## Additional Options

---
## Summary by cubic
Split `SelectCard` `sizeVariant` into `padding` and `rounding`, hardcoded the interactive variant to `select-card`, extracted shared padding/rounding scales to `@opal/shared`, renamed card props to drop the `Variant` suffix, and standardized defaults to `"md"`; updated all callsites and docs. Added explicit `padding="sm"` and `rounding="lg"` to all `SelectCard` usages to preserve the previous spacing and shape after the default change.

- **Refactors**
  - `SelectCard`: adds `padding` (default `"md"`) and `rounding` (default `"md"`); removes public `variant` (now always `"select-card"`).
  - `Card`, `SelectCard` use shared `cardPaddingVariants` and `cardRoundingVariants` from `@opal/shared`.
  - Prop rename across cards: `paddingVariant` → `padding`, `roundingVariant` → `rounding`, `backgroundVariant` → `background`, `borderVariant` → `border`; `EmptyMessageCard` now `padding` (default `"md"`) and sets `rounding="md"`.
  - Removed redundant `variant="select-card"`; stories/READMEs updated.

- **Migration**
  - Replace `sizeVariant` on `SelectCard` with `padding` of the same value; set `rounding` if you relied on non-default rounding:
    - `lg` → `lg`; `md`/`sm` → `md`; `xs`/`2xs` → `sm`; `fit` → `lg`.
  - Defaults changed: `SelectCard` now defaults to `padding="md"` and `rounding="md"`; set `padding="sm"` and `rounding="lg"` to preserve the old look.
  - `Card`/`EmptyMessageCard` now default to `padding="md"`; set `padding="sm"` if you want the previous compact default.

<sup>Written for commit 41dceb691a0aa297e4ce9624dec362033c2aeea4. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/refresh-pages/admin/CodeInterpreterPage/index.tsx
+++ b/web/src/refresh-pages/admin/CodeInterpreterPage/index.tsx
@@ -112,7 +112,7 @@ export default function CodeInterpreterPage() {
       <SettingsLayouts.Body>
         {isEnabled || isLoading ? (
           <Hoverable.Root group="code-interpreter/Card">
-            <SelectCard state="filled">
+            <SelectCard state="filled" padding="sm" rounding="lg">
               <CardHeaderLayout
                 sizePreset="main-ui"
                 variant="section"
@@ -156,7 +156,12 @@ export default function CodeInterpreterPage() {
             </SelectCard>
           </Hoverable.Root>
         ) : (
-          <SelectCard state="empty" onClick={() => handleToggle(true)}>
+          <SelectCard
+            state="empty"
+            padding="sm"
+            rounding="lg"
+            onClick={() => handleToggle(true)}
+          >
             <CardHeaderLayout
               sizePreset="main-ui"
               variant="section"
--- a/web/src/refresh-pages/admin/ImageGenerationPage/ImageGenerationContent.tsx
+++ b/web/src/refresh-pages/admin/ImageGenerationPage/ImageGenerationContent.tsx
@@ -248,6 +248,8 @@ export default function ImageGenerationContent() {
                 >
                   <SelectCard
                     state={STATUS_TO_STATE[status]}
+                    padding="sm"
+                    rounding="lg"
                     aria-label={`image-gen-provider-${provider.image_provider_id}`}
                     onClick={
                       isDisconnected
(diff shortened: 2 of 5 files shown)


web/lib/opal/src/components/cards/select-card/components.tsx
```diff
@@ -1,50 +1,50 @@
...
    *
-   * | Size       | Rounding     |
-   * |------------|--------------|
-   * | `lg`       | `rounded-16` |
-   * | `md`–`sm`  | `rounded-12` |
-   * | `xs`–`2xs` | `rounded-08` |
-   * | `fit`      | `rounded-16` |
+   * @default "sm"
+   */
+  paddingVariant?: PaddingVariants;
```
jmelahman: Consider just naming these `padding` and `rounding` (or `borderRadius`).

"Variant" seems unnecessary and I find it a bit intimidating as a consumer. Feels like I need to understand this complex system of variations when I just want to adjust the padding, etc.

jmelahman (approved):

web/lib/opal/src/components/cards/select-card/components.tsx
raunakab: Sure. If I do this, I would then also like to update the API to `Card` as well.

raunakab: Updated.

PR #9842, opened 2026-04-01
feat: google drive error resolution

evan-onyx:
## Description

Add a new interface for connectors to implement, intended to be used for individual indexing error resolution. Implemented the interface for google drive.

## How Has This Been Tested?

added connector tests

## Additional Options

---
## Summary by cubic
Adds a new `Resolver` interface and Google Drive error resolution to re-fetch failed documents by `webViewLink` using batched Drive API calls. Emits ancestor folders and can sync permissions to reduce API calls and speed up recovery.

- **New Features**
  - Added `Resolver.resolve_errors(errors, include_permissions=False)` to re-process failures and emit `Document` and `HierarchyNode`.
  - Implemented `GoogleDriveConnector.resolve_errors` with batched `files().get` (100-item chunks), yields per-link failures when fetches fail, emits ancestors before documents, and optionally syncs permissions.
  - Added `get_files_by_web_view_links_batch` using Drive `BatchHttpRequest`; skips invalid links and continues; daily tests cover single/multiple files, invalid links, empty inputs, entity-failure skips, and hierarchy validation.

- **Bug Fixes**
  - Fixed Drive field selection for `files().get` by deriving single-file fields from list fields, preventing missing or incorrect metadata.

<sup>Written for commit 043df22f43fc0975465dbbce6aee379913d3d347. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/connectors/google_drive/file_retrieval.py
+++ b/backend/onyx/connectors/google_drive/file_retrieval.py
@@ -560,35 +560,46 @@ def get_file_by_web_view_link(
     )
 
 
+class BatchRetrievalResult:
+    """Result of a batch file retrieval, separating successes from errors."""
+
+    def __init__(self) -> None:
+        self.files: dict[str, GoogleDriveFileType] = {}
+        self.errors: dict[str, Exception] = {}
+
+
 def get_files_by_web_view_links_batch(
     service: GoogleDriveService,
     web_view_links: list[str],
     field_type: DriveFileFieldType,
-) -> dict[str, GoogleDriveFileType]:
+) -> BatchRetrievalResult:
     """Retrieve multiple Google Drive files by webViewLink using the batch API.
 
-    Returns a dict mapping web_view_link to file metadata.
+    Returns a BatchRetrievalResult containing successful file retrievals
+    and errors for any files that could not be fetched.
     Automatically splits into chunks of MAX_BATCH_SIZE.
     """
     fields = _get_single_file_fields(field_type)
     if len(web_view_links) <= MAX_BATCH_SIZE:
         return _get_files_by_web_view_links_batch(service, web_view_links, fields)
 
-    result: dict[str, GoogleDriveFileType] = {}
+    combined = BatchRetrievalResult()
     for i in range(0, len(web_view_links), MAX_BATCH_SIZE):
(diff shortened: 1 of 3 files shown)


backend/onyx/connectors/interfaces.py
```diff
@@ -298,6 +298,22 @@ def load_from_checkpoint_with_perm_sync(
...
+class Resolver(BaseConnector):
+    @abc.abstractmethod
+    def resolve_errors(
+        self,
+        errors: list[ConnectorFailure],
+        include_permissions: bool = False,
+    ) -> Generator[Document | ConnectorFailure | HierarchyNode, None, None]:
+        """Attempts to yield back ALL the documents described by the errors, no checkpointing.
+
+        Caller's responsibility is to delete the old ConnectorFailures and replace with the new ones.
```
acaprau, line 310: this comment doesn't make 100% sense to me. it seems to imply you are meant to return ConnectorFailure, but the typing indicates you can return Document and HierarchyNode. i think it just might need some more detail.

evan-onyx: u can return connector failures too!

acaprau: ofc, i just meant something like ending the sentence with "replace with the new ones if still necessary." or something like that.

acaprau:
@cubic

backend/onyx/connectors/google_drive/file_retrieval.py
```diff
@@ -536,3 +558,74 @@ def get_file_by_web_view_link(
...
+    Returns a BatchRetrievalResult containing successful file retrievals
+    and errors for any files that could not be fetched.
+    Automatically splits into chunks of MAX_BATCH_SIZE.
+    """
+    fields = _get_single_file_fields(field_type)
+    if len(web_view_links) <= MAX_BATCH_SIZE:
+        return _get_files_by_web_view_links_batch(service, web_view_links, fields)
+
+    combined = BatchRetrievalResult()
+    for i in range(0, len(web_view_links), MAX_BATCH_SIZE):
```
acaprau: nit i don't rly mind but can't this range logic also encompass the above case where the size of the links is less than the batch size for free?

PR #9847, opened 2026-04-02
fix(opensearch): Doc IDs whose length would exceed OpenSearch's ID length are hashed

acaprau:
## Description
When we generate an OpenSearch chunk ID, we have to respect OpenSearch's max ID length, which is 512 bytes. Since document IDs are a component of chunk IDs, an exceedingly long doc ID can be invalid and prevent indexing/migration of a document.

This PR checks whether a given doc ID would generate a chunk ID that is too long, and if so, uses a shorter hashed version of the doc ID for the chunk ID.

IMPORTANT: This PR **MUST NOT** change the ways existing document chunk IDs are represented, otherwise existing OpenSearch indices will break. Since the logic of this PR only modifies the representations of doc chunks of docs whose names are too long, which were never valid docs to index into OpenSearch anyway, this is fine.

## How Has This Been Tested?
Added unit tests for `get_opensearch_doc_chunk_id`.

## Additional Options

---
## Summary by cubic
Prevents OpenSearch indexing failures by hashing document IDs that would push chunk IDs over the 512-byte limit. Keeps existing valid IDs unchanged and works in both single- and multi-tenant modes.

- **Bug Fixes**
  - Hashes oversized doc IDs with blake2b; digest size is set so the final ID stays under 512 bytes after the chunk suffix and tenant prefix.
  - Computes the allowable doc ID length per call and revalidates the final chunk ID; preserves current ID format in single- and multi-tenant modes.
  - Adds `MAX_DOCUMENT_ID_ENCODED_LENGTH` and `DocumentIDTooLongError`; `filter_and_validate_document_id` now accepts `max_encoded_length`.
  - Adds unit tests covering hashing, boundaries, determinism, tenant prefixes, different tenants, special-char filtering, and enforcing the 512-byte cap.

<sup>Written for commit 4b89c7b57db925412e1ba7b325c663365506d75a. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/unit/onyx/document_index/opensearch/test_get_doc_chunk_id.py
+++ b/backend/tests/unit/onyx/document_index/opensearch/test_get_doc_chunk_id.py
@@ -175,6 +175,22 @@ class TestGetOpensearchDocChunkIdEdgeCases:
         result = get_opensearch_doc_chunk_id(SINGLE_TENANT_STATE, doc_id, chunk_index=0)
         assert doc_id in result
 
+    def test_doc_id_at_boundary_length_multitenant(self) -> None:
+        """
+        Tests that a doc ID right at the boundary should not be hashed in
+        multitenant mode.
+        """
+        suffix = f"__{DEFAULT_MAX_CHUNK_SIZE}__0"
+        suffix_len = len(suffix.encode("utf-8"))
+        prefix = f"{EXPECTED_SHORT_TENANT}__"
+        prefix_len = len(prefix.encode("utf-8"))
+        # Max doc ID length that won't trigger hashing (must be <
+        # max_encoded_length).
+        max_doc_len = MAX_DOCUMENT_ID_ENCODED_LENGTH - suffix_len - prefix_len - 1
+        doc_id = "a" * max_doc_len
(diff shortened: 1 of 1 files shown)


backend/onyx/document_index/opensearch/schema.py
```diff
@@ -75,17 +80,49 @@ def get_opensearch_doc_chunk_id(
...
     """
-    sanitized_document_id = filter_and_validate_document_id(document_id)
-    opensearch_doc_chunk_id = (
-        f"{sanitized_document_id}__{max_chunk_size}__{chunk_index}"
+    opensearch_doc_chunk_id_suffix: str = f"__{max_chunk_size}__{chunk_index}"
+    encoded_suffix_length: int = len(opensearch_doc_chunk_id_suffix.encode("utf-8"))
+    max_encoded_permissible_doc_id_length: int = (
+        MAX_DOCUMENT_ID_ENCODED_LENGTH - encoded_suffix_length
+    )
+    if tenant_state.multitenant:
```
evan-onyx, line 92: we shouldn't need both ifs u got this i believe in u

evan-onyx (approved):

acaprau:
@greptile @cubic

PR #9854, opened 2026-04-02
feat(chat): add multi-model selector and chat hook

nmgarza5:
## Description

Adds `ModelSelector` component and `useMultiModelChat` hook for multi-model chat support.

- `ModelSelector.tsx` — popover-based model picker using opal `SelectButton`/`OpenButton`, reuses `LLMPopover` patterns (accordion grouping, `LineItem` rendering, search filtering)
- `useMultiModelChat.ts` — hook managing selected model state (add/remove/replace up to 3 models)
- `useMultiModelChat.test.tsx` — unit tests for the hook

Stacked on #9648 (PR3: frontend types).

## Additional Options

- [ ] [Optional] Please cherry-pick this PR to the latest release version.
- [x] [Optional] Override Linear Check

PR #9855, opened 2026-04-02
feat(chat): add multi-model response panels

nmgarza5:
## Description

Adds the response panel UI for multi-model chat — displays side-by-side model responses in generation and selection modes.

- `MultiModelResponseView.tsx` — orchestrates panel layout with generation/selection mode switching
- `MultiModelPanel.tsx` — individual model response panel (renders `AgentMessage` with model header)
- `interfaces.ts` — `MultiModelResponse` type definition
- `AgentMessage.tsx` — adds `hideFooter` prop to suppress footer in multi-model panels

Stacked on #9854 (PR4a: selector + hook).

## Additional Options

### DIFF
--- a/web/src/app/app/message/MultiModelPanel.tsx
+++ b/web/src/app/app/message/MultiModelPanel.tsx
@@ -58,8 +58,8 @@ export default function MultiModelPanel({
   const ProviderIcon = getProviderIcon(provider, modelName);
 
   const handlePanelClick = useCallback(() => {
-    if (!isHidden) onSelect();
-  }, [isHidden, onSelect]);
+    if (!isHidden && !isPreferred) onSelect();
+  }, [isHidden, isPreferred, onSelect]);
 
   const header = (
     <div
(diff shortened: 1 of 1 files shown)


web/src/app/app/message/MultiModelPanel.tsx
```diff
@@ -0,0 +1,126 @@
+"use client";
+
+import { useCallback } from "react";
+import { Button } from "@opal/components";
+import { Text } from "@opal/components";
```
Subash-Mohan, line 5: shall we just consildate this in single import ?

web/src/app/app/message/MultiModelResponseView.tsx
```diff
@@ -0,0 +1,372 @@
...
+      chatState,
+      llmManager,
+      otherMessagesCanSwitchTo,
+      onMessageSelection,
+      onRegenerate,
+      parentMessage,
+    ]
+  );
+
+  if (showSelectionMode) {
```
Subash-Mohan, line 224: shall we move this logic into a helper function and use it here?

PR #9860, opened 2026-04-02
fix(offline): fallback to system sans-serif font

jmelahman:
## Description

This `hankenGrotesk.variable` overrides the [globally](https://github.com/onyx-dot-app/onyx/blob/722f7de335715ff6cc9dacc2231a4983ad100920/web/src/app/globals.css#L193-L194) defined `--font-hanken-grotesk` which loses some of the potential fallback fonts. While offline running the web server in dev mode, this means the app falls back to a serif font which bothers me.

## How Has This Been Tested?

Tested locally while offline

## Additional Options

---
## Summary by cubic
Fixes offline font fallback by self-hosting `Hanken_Grotesk` and `DM_Mono` via `next/font/google`, adding system sans-serif/monospace fallbacks so text doesn’t render serif when offline. Removes Google Fonts `@import`s and uses `cn` to apply both font variables on the html element.

<sup>Written for commit 114fa2937f08fe94b2aabe03ad1a25d853ee65bb. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/app/layout.tsx
+++ b/web/src/app/layout.tsx
@@ -14,6 +14,7 @@ import { WebVitals } from "./web-vitals";
 import { ThemeProvider } from "next-themes";
 import { TooltipProvider } from "@/components/ui/tooltip";
 import StatsOverlayLoader from "@/components/dev/StatsOverlayLoader";
+import { cn } from "@/lib/utils";
 import AppHealthBanner from "@/sections/AppHealthBanner";
 import CustomAnalyticsScript from "@/providers/CustomAnalyticsScript";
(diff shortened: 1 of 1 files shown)


web/src/app/layout.tsx
```diff
@@ -44,7 +67,7 @@ export default function RootLayout({
   return (
     <html
       lang="en"
-      className={`${hankenGrotesk.variable}`}
+      className={`${hankenGrotesk.variable} ${dmMono.variable}`}
```
raunakab: Small nit, but could we update this to use `cn` instead?

PR #9861, opened 2026-04-02
feat(install): add GitHub star prompt at end of install script

rohoswagger:
## Description

Adds an interactive GitHub star prompt in two places:

1. **Install script** (`deployment/docker_compose/install.sh`) — after installation completes, asks users to star the repo. Uses `gh api -X PUT /user/starred` (matching oh-my-codex's approach). Respects `--no-prompt` for CI/automation.

2. **onyx-cli** (`cli/internal/starprompt/`) — one-time prompt shown before the TUI launches. State persisted to `~/.config/onyx-cli/star-prompt.json` so it only shows once per user. Marks as prompted before asking (so Ctrl+C won't cause re-prompt). Silently skips if not a TTY or if `gh` CLI is not installed.

Inspired by [oh-my-codex](https://github.com/Yeachan-Heo/oh-my-codex/blob/main/src/cli/star-prompt.ts).

## How Has This Been Tested?

https://github.com/user-attachments/assets/c7ff2d5f-149a-4bdb-b20d-f97cdf8e3d67

https://github.com/user-attachments/assets/e816b399-a048-47e9-9e39-29963ed2df5d

- Go build verified (`go build ./...` passes)
- Manually reviewed install script logic against existing interactive prompt helpers
- Star prompt package follows same patterns as oh-my-codex (one-time state file, TTY check, gh CLI check, mark-before-ask)

## Additional Options

### DIFF
--- a/cli/internal/starprompt/starprompt.go
+++ b/cli/internal/starprompt/starprompt.go
@@ -1,0 +1,83 @@
+// Package starprompt implements a one-time GitHub star prompt shown before the TUI.
+// Skipped when stdin/stdout is not a TTY, when gh CLI is not installed,
+// or when the user has already been prompted. State is stored in the
+// config directory so it shows at most once per user.
+package starprompt
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/onyx-dot-app/onyx/cli/internal/config"
+	"golang.org/x/term"
+)
+
+const repo = "onyx-dot-app/onyx"
+
+func statePath() string {
+	return filepath.Join(config.ConfigDir(), ".star-prompted")
+}
+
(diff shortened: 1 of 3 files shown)


cli/internal/starprompt/starprompt.go
```diff
@@ -0,0 +1,93 @@
...
+	"time"
+
+	"github.com/onyx-dot-app/onyx/cli/internal/config"
+	"golang.org/x/term"
+)
+
+const repo = "onyx-dot-app/onyx"
+
+type state struct {
+	PromptedAt string `json:"prompted_at"`
```
jmelahman: super nit, but the fastest, most POSIX way to do this would be just write an empty file and then your `hasBeenPrompted` just returns whether that file exists since that's much faster than actually loading and parsing it.

rohoswagger: big brain

PR #9868, opened 2026-04-02
fix(federated): prevent masked credentials from corrupting stored secrets

nmgarza5:
## Description

When an admin edits a federated connector, the `GET /api/federated/{id}` endpoint returns credentials with `apply_mask=True`, replacing secrets with `••••••••••••` (U+2022 BULLET). The edit form loaded these masked values into state. On save, `PUT /api/federated/{id}` stored the masked string as the real credentials — permanently corrupting both `client_id` and `client_secret`. This caused Slack OAuth to fail with "Invalid client_id parameter" because the URL contained masked bullets instead of the real client ID.

### Frontend fix

Edit form no longer pre-fills credential fields with masked values in edit mode. A `credentialsModified` flag tracks whether the user has actually typed into any credential field:

- **If credentials are not modified:** `null` is passed to `updateFederatedConnector`, which serializes as `undefined` (omitted from the JSON body). The Pydantic model defaults `credentials` to `None`. The API endpoint passes `None` to the DB function. The DB function's `if credentials is not None:` guard skips the entire credential update block — existing credentials are untouched.
- **If credentials are modified:** The new values are sent and validated normally.

### Backend fix (defense-in-depth)

Both `create_federated_connector` and `update_federated_connector` now reject any credential value containing the mask character (U+2022 BULLET), preventing masked placeholders from ever being persisted.

**Linear:** [CS-55](https://linear.app/onyx-app/issue/CS-55/slack-oauth-fails-with-invalid-client-id-masked-credentials-corrupted)

## Additional Options

### DIFF
--- a/backend/onyx/db/federated.py
+++ b/backend/onyx/db/federated.py
@@ -8,6 +8,8 @@ from sqlalchemy.orm import selectinload
 from sqlalchemy.orm import Session
 
 from onyx.configs.constants import FederatedConnectorSource
+from onyx.configs.constants import MASK_CREDENTIAL_CHAR
+from onyx.configs.constants import MASK_CREDENTIAL_LONG_RE
 from onyx.db.engine.sql_engine import get_session_with_current_tenant
 from onyx.db.models import DocumentSet
 from onyx.db.models import FederatedConnector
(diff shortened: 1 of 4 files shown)


backend/onyx/db/federated.py
```diff
@@ -45,6 +45,21 @@ def fetch_all_federated_connectors_parallel() -> list[FederatedConnector]:
         return fetch_all_federated_connectors(db_session)
 
 
+# The mask_string() function in encryption.py uses "•" (U+2022 BULLET) to mask secrets.
+# If any credential value contains this character, it's a masked placeholder that must
+# never be persisted — doing so permanently corrupts the real secret.
+MASK_CHAR = "\u2022"
```
nmgarza5: move to constants file

PR #9870, opened 2026-04-02
fix(mobile): sidebar overlaps content on medium-sized screens

jmelahman:
## Description

Closes https://linear.app/onyx-app/issue/ENG-3890/sidebar-breakpoints

## How Has This Been Tested?

Expecting no/minimal change to large screen cases.

20260402_14h09m58s_grim

## Additional Options

---
## Summary by cubic
Fixes sidebar overlap on medium screens by using a foldable rail that stays in flow and expands as a fixed overlay with a blur backdrop. Aligns breakpoints and centralizes sidebar state to satisfy Linear `ENG-3890`.

- **Bug Fixes**
  - Medium screens: folded rail stays visible with reserved width; expanded sidebar overlays content with a blur-only backdrop; backdrop click closes.
  - Mobile/medium use a fixed overlay; page scroll remains intact; toggle button tooltip reflects open/close and adjusts side.

- **Refactors**
  - Centralized sidebar state in `@/layouts/sidebar-layouts` as `StateProvider` + `useSidebarState`; removed `@/providers/AppSidebarProvider`, wired into `AppProvider`, and migrated consumers (`NRFChrome`, `app-layouts`, `ClientLayout`, `AppSidebar`).
  - `SidebarBody` auto-hides when folded; `AppSidebar` relies on layout-managed folding; `AdminSidebar` adds a folded “Search” that expands and auto-focuses, and hides the footer when folded.
  - Breakpoints unified: Tailwind `sm` 724px, `md` 912px, `lg` 1232px; `mobile` max 724px; `MOBILE_SIDEBAR_BREAKPOINT_PX` set to 724px; updated admin document-processing page to `sm:*`.

<sup>Written for commit 89fcb1b44c3f7bfd0b6ecec0f7bf564b21569c69. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/components/admin/ClientLayout.tsx
+++ b/web/src/components/admin/ClientLayout.tsx
@@ -1,6 +1,5 @@
 "use client";
 
-import { useState } from "react";
 import AdminSidebar from "@/sections/sidebar/AdminSidebar";
 import { usePathname } from "next/navigation";
 import { useSettingsContext } from "@/providers/SettingsProvider";
@@ -10,6 +9,7 @@ import { cn } from "@/lib/utils";
 import { ADMIN_ROUTES } from "@/lib/admin-routes";
 import useScreenSize from "@/hooks/useScreenSize";
 import { SvgSidebar } from "@opal/icons";
+import { useSidebarState } from "@/layouts/sidebar-layouts";
 
 export interface ClientLayoutProps {
   children: React.ReactNode;
@@ -52,7 +52,8 @@ const SETTINGS_LAYOUT_PREFIXES = [
 ];
 
 export function ClientLayout({ children, enableCloud }: ClientLayoutProps) {
-  const [sidebarFolded, setSidebarFolded] = useState(true);
+  const { folded: sidebarFolded, setFolded: setSidebarFolded } =
+    useSidebarState();
   const { isMobile } = useScreenSize();
   const pathname = usePathname();
   const settings = useSettingsContext();


web/tailwind-themes/tailwind.config.js
```diff
@@ -65,11 +65,13 @@ module.exports = {
         "neutral-10": "var(--neutral-10) 5%",
       },
       screens: {
+        sm: "724px",
+        md: "912px",
+        lg: "1232px",
         "2xl": "1420px",
```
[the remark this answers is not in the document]
jmelahman: This was intentional

web/src/components/admin/ClientLayout.tsx
```diff
@@ -49,6 +52,8 @@ const SETTINGS_LAYOUT_PREFIXES = [
 ];
 
 export function ClientLayout({ children, enableCloud }: ClientLayoutProps) {
+  const [sidebarFolded, setSidebarFolded] = useState(true);
```
raunakab: There is a context that I've been using for "sidebar-closed"-ability: `useAppSidebarContext`. I think it could be useful here, because it would tie the closed-ness of the app-sidebar and the admin-sidebar together.

raunakab: We may even want to consider renaming it from `useAppSidebarContext` to `useSidebarState`? Idk, not too picky on the actual semantics, but the name should reflect that it's for *both* the app and admin sidebars.

web/tailwind-themes/tailwind.config.js
```diff
@@ -65,11 +65,13 @@ module.exports = {
         "neutral-10": "var(--neutral-10) 5%",
       },
       screens: {
+        sm: "724px",
+        md: "912px",
+        lg: "1232px",
```
raunakab, line 70: Should we have these sizes be available in the CSS and then *pulled into* Tailwind (via `var(--sm-screen-size-variable)`)?

raunakab (approved):
Is `SidebarBody` stale now? If so, we should delete it.

web/tailwind-themes/tailwind.config.js
jmelahman: I'll see if this can be improved, but I think we'll always need two declarations of this for CSS and another for JS

jmelahman:
> Is `SidebarBody` stale now? If so, we should delete it.

Still used by Craft. I can clean up later, but I don't think should block here.

web/src/components/admin/ClientLayout.tsx
jmelahman: I think `useSidebarState` is more apt. I'll look into addressing this with https://github.com/onyx-dot-app/onyx/pull/9870 as that further complicates the sidebar state, so better consolidation could be more useful.

web/tailwind-themes/tailwind.config.js
jmelahman: Ah, your concern is more for CSS compatibility. In general, I am not a fan of using CSS (since we already use tailwind extensively). IMO, this is currently a feature not a bug. If we expand the usage of Opal namely in the web extensions or desktop, this might make sense though.

PR #9874, opened 2026-04-02
feat(cli): make onyx-cli agent-friendly

rohoswagger:
## Description

Makes the onyx-cli significantly more useful when invoked by AI agents and scripts, based on patterns from CLI-for-agents design articles.

**Stdin support for `ask`:**
- Pipe content via stdin as context: `cat error.log | onyx-cli ask --prompt "find the root cause"`
- Stdin-only mode: `echo "what is onyx?" | onyx-cli ask`
- Positional arg + stdin combined: `cat log | onyx-cli ask "summarize this"`
- 10MB stdin cap to prevent OOM

**Auto-truncate for non-TTY callers:**
- When stdout is not a TTY (agent/script calling), output auto-truncates at 4KB
- Full response saved to temp file with exploration hints (`grep`, `tail`)
- `--max-output N` to override threshold, `--max-output 0` to disable
- Human terminal use is unchanged

**Semantic exit codes:**
- `0` success, `2` not configured, `3` auth failure, `4` unreachable, `5` bad request
- New `exitcodes` package with `ExitError` type, extracted in `main.go`

**Non-interactive configure:**
- `onyx-cli configure --server-url X --api-key Y` saves config without prompts
- `--dry-run` tests connection without saving

**Other improvements:**
- `--quiet`/`-q` flag on `ask` to buffer output and print once (no streaming)
- Actionable error messages with remediation hints on all commands
- Help examples added to every subcommand
- Consistent error message formatting

## How Has This Been Tested?

- 12 new unit tests: `resolveQuestion` (3 cases), `overflowWriter` (5 cases including quiet mode), `exitcodes` (4 cases)
- All existing tests pass (`go test ./...`)
- `go vet` and `golangci-lint` clean
- Manual verification of `--help` output for all commands

## Additional Options

---
## Summary by cubic
Make the CLI easier for agents and scripts with stdin-aware `ask`, disk-backed auto-truncate, non-interactive `configure` (implicit or explicit stdin via `--api-key-stdin`, `--dry-run`), quiet mode, semantic exit codes, and TTY-only streaming status lines (suppressed in `--quiet`).

- **New Features**
  - `ask`: accept piped stdin (works with arg or `--prompt`); `--quiet` for non-streaming; auto-truncate non-TTY output to 4KB by default (`--max-output`, 0 disables); full response saved to a temp file.
  - `ask`: show dim search/reasoning/tool status on stderr while streaming, only when stdout is a TTY; suppressed with `--quiet`.
  - `configure`: non-interactive via `--server-url` and `--api-key`, or read the API key from stdin when omitted; `--api-key-stdin` for explicit piping; validates connection before saving; `--dry-run` to test without saving.
  - Semantic exit codes via `exitcodes` (2 bad request, 3 not configured, 4 auth failure, 5 unreachable) with clearer help and remediation hints.

- **Bug Fixes**
  - `ask`: cap stdin at 10MB; stream overflow directly to disk; reject `--json` + `--quiet`; reject conflicting arg + `--prompt`; avoid trailing newline in JSON output; fall back to stdout if temp file write fails.
  - `configure`/`validate-config`: distinguish auth failures from unreachable via `*api.AuthError` and return codes 4 vs 5.
  - `configure`: don’t persist env var overrides when saving.

<sup>Written for commit 2c28b5799236f25a82c26efa31d37a6a9e3c793d. Summary will update on new commits.</sup>

### DIFF
--- a/cli/cmd/ask.go
+++ b/cli/cmd/ask.go
@@ -130,7 +130,7 @@ to a temp file. Set --max-output 0 to disable truncation.`,
 				case models.MessageDeltaEvent:
 					ow.Write(e.Content)
 				case models.SearchStartEvent:
-					if isTTY {
+					if isTTY && !askQuiet {
 						if e.IsInternetSearch {
 							fmt.Fprintf(os.Stderr, "\033[2mSearching the web...\033[0m\n")
 						} else {
@@ -138,21 +138,21 @@ to a temp file. Set --max-output 0 to disable truncation.`,
 						}
 					}
 				case models.SearchQueriesEvent:
-					if isTTY {
+					if isTTY && !askQuiet {
 						for _, q := range e.Queries {
 							fmt.Fprintf(os.Stderr, "\033[2m  → %s\033[0m\n", q)
 						}
 					}
 				case models.SearchDocumentsEvent:
-					if isTTY && len(e.Documents) > 0 {
+					if isTTY && !askQuiet && len(e.Documents) > 0 {
 						fmt.Fprintf(os.Stderr, "\033[2mFound %d documents\033[0m\n", len(e.Documents))
 					}
 				case models.ReasoningStartEvent:
-					if isTTY {
+					if isTTY && !askQuiet {
 						fmt.Fprintf(os.Stderr, "\033[2mThinking...\033[0m\n")
 					}
 				case models.ToolStartEvent:
-					if isTTY && e.ToolName != "" {
+					if isTTY && !askQuiet && e.ToolName != "" {
 						fmt.Fprintf(os.Stderr, "\033[2mUsing %s...\033[0m\n", e.ToolName)
 					}
 				case models.ErrorEvent:


cli/internal/exitcodes/codes.go
```diff
@@ -0,0 +1,46 @@
+// Package exitcodes defines semantic exit codes for the Onyx CLI.
+package exitcodes
+
+import "fmt"
+
+const (
+	Success        = 0
+	General        = 1
+	NotConfigured  = 2
```
jmelahman: I would make `BadRequest` two. Convention is typically to use `2` for invalid args/command line errors

cli/internal/exitcodes/codes.go
```diff
@@ -0,0 +1,46 @@
...
+type ExitError struct {
+	Code int
+	Err  error
+}
+
+func (e *ExitError) Error() string {
+	return e.Err.Error()
+}
+
+func (e *ExitError) Unwrap() error {
```
jmelahman: `Wrap` and `Unwrap` are only used in tests. Maybe we can remove?

cli/cmd/configure.go
```diff
@@ -1,19 +1,98 @@
...
+	"github.com/onyx-dot-app/onyx/cli/internal/exitcodes"
 	"github.com/onyx-dot-app/onyx/cli/internal/onboarding"
 	"github.com/spf13/cobra"
 )
 
 func newConfigureCmd() *cobra.Command {
-	return &cobra.Command{
+	var (
+		serverURL string
+		apiKey    string
```
jmelahman: Does this support reading from stdin? It's fairly common to use files/env to avoid leaking api keys in shell history (I would recommend this/put it in the docs as it is more secure). For example, docker has both a `--password` and `--password-stdin`, https://docs.docker.com/reference/cli/docker/login/#options

Also, this should probably respect `ONYX_API_KEY` if it doesn't already

cli/cmd/ask.go
```diff
@@ -82,43 +126,174 @@ func newAskCmd() *cobra.Command {
...
+	default:
+		return "", exitcodes.New(exitcodes.BadRequest, "no question provided\n  Usage: onyx-cli ask \"your question\"\n  Or:    echo \"context\" | onyx-cli ask --prompt \"your question\"")
+	}
+}
+
+// overflowWriter handles streaming output with optional truncation.
+// When limit > 0, it streams to a temp file on disk (not memory) and stops
+// writing to stdout after limit bytes. When limit == 0, it writes directly
+// to stdout. In quiet mode, it buffers in memory and prints once at the end.
+type overflowWriter struct {
```
jmelahman: nit: could maybe be moved to a dedicated package. Could be use for future commands and I find it better to refactor before LLMs duplicate it everywhere because they don't know to look in here -- they're much more willing to look at the file tree and especially utils than they are to read whole files.

cli/cmd/ask.go
```diff
@@ -82,43 +126,174 @@ func newAskCmd() *cobra.Command {
...
+			_ = w.tmpFile.Close()
+			_ = os.Remove(w.tmpFile.Name())
+		}
+		fmt.Println()
+		return
+	}
+
+	// Close the temp file so it's readable
+	tmpPath := w.tmpFile.Name()
+	_ = w.tmpFile.Close()
```
jmelahman: It's probably not a big deal, but we probably should be 1. `defer`ing closing files (so it closes even if this function is exited prematurely) and 2. at least logging when files fail to close. I don't think it's likely, but in an agentic world, this may manifest as `too many files open` errors eventually.

jmelahman (approved):

cli/cmd/configure.go
rohoswagger: yep added support for pipe, ONYX_API_KEY, and --api-key-stdin

PR #9891, opened 2026-04-03
fix(notion): extract people properties and inline table content

jessicasingh7:
## Summary
- **[ENG-3970](https://linear.app/onyx-app/issue/ENG-3970)**: Notion databases with "people" type columns (user mentions like @User) were not indexed. The `_recurse_properties` method followed `type: "person"` into the inner sub-dict, losing the `name` field. Now captures the name before descending.
- **[ENG-3971](https://linear.app/onyx-app/issue/ENG-3971)**: Inline table blocks (`table`/`table_row`) were not indexed. The connector only extracted text from `rich_text` fields, but `table_row` blocks store content in `cells`. Now extracts `plain_text` from each cell.

## Test plan
- Created test database with  with people columns and test page with inline table, confirmed indexed content is now visible to Onyx locally and new Notion connector tests pass

## Additional Options

### DIFF
--- a/backend/onyx/connectors/notion/connector.py
+++ b/backend/onyx/connectors/notion/connector.py
@@ -44,7 +44,7 @@ _NOTION_CALL_TIMEOUT = 30  # 30 seconds
 _MAX_PAGES = 1000
 
 
-# TODO: Tables need to be ingested, Pages need to have their metadata ingested
+# TODO: Pages need to have their metadata ingested
 
 
 class NotionPage(BaseModel):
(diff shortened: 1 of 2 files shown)


backend/onyx/connectors/notion/connector.py
```diff
@@ -452,6 +452,19 @@ def _recurse_properties(inner_dict: dict[str, Any]) -> str | None:
...
+
+                # Notion user objects (people properties, created_by, etc.) have
+                # "name" at the same level as "type": "person"/"bot". If we drill
+                # into the person/bot sub-dict we lose the name. Capture it here
+                # before descending, but skip "title"-type properties where "name"
+                # is not the display value we want.
+                if (
+                    "name" in sub_inner_dict
+                    and isinstance(sub_inner_dict["name"], str)
+                    and type_name not in ("title",)
```
evan-onyx, line 464: is this the right way to identify this case? I'm concerned there will be some types that have a "name" that we don't want to extract the name field like this? If we strongly believe this is likely to work long term I'm fine with it, we can just do type_name != "title"

jessicasingh7: I believe this is robust. In the Notion API, whenever an object has both `"type"` and `"name"` at the same dict level, `"name"` is always the human-readable display value we want:

- **Person users**: `{type: "person", name: "Jess", person: {email: "..."}}` — the main fix
- **Bot users**: `{type: "bot", name: "Onyx Integration", bot: {}}` — same pattern
- **File items**: `{type: "file", name: "document.pdf", file: {url: "..."}}` — actually an improvement, extracts filename instead of drilling into the temporary S3 URL

The cases where `"name"` means something else (select option labels, status names) have it nested one level deeper — e.g. `{type: "select", select: {name: "Option A"}}`. The outer dict has `type` but no `name`, and the inner dict has `name` but no `type`, so the while loop exits naturally before this check runs.

The only exception is `title` properties which can have a `name` field at the same level, hence the explicit exclusion.

jessicasingh7: I think only person, bot, file, and title have name and type at the same level and i think title is the only edge case which we exclude.

PR #9907, opened 2026-04-04
refactor: flatten opal card layouts, add `children` to `CardHeaderLayout`

raunakab:
## Description

- Flatten `cards/header-layout/` into `cards/components.tsx` — removes the unnecessary subdirectory nesting
- Add `children` prop to `CardHeaderLayout` — renders full-width below the header row, useful for expandable sections (search bars, tool lists, etc.)
- Update barrel re-export in `layouts/index.ts` — all consumers import via `@opal/layouts` so no callsite changes needed

## Additional Options

---
## Summary by cubic
Flattened the Opal card layouts and replaced `CardHeaderLayout` with the namespaced `Card.Header`. Added a full‑width `children` slot for content below the header row and updated docs, exports, and all usages.

- **Refactors**
  - Moved `cards/header-layout/*` into `cards/components.tsx` and removed the subdirectory.
  - Exported `Card` and `CardHeaderProps` from `@opal/layouts`; consumers now use `Card.Header`.
  - Updated internal callsites and Storybook to `Card.Header` (Admin pages and `ProviderCard`).
  - Added `cards/README.md` documenting `Card.Header`; removed the old `header-layout/README.md`.

<sup>Written for commit 467ff981ce77ec399be4b0b8e13749165b1d9ca0. Summary will update on new commits.</sup>


web/src/refresh-pages/admin/LLMConfigurationPage.tsx
```diff
@@ -424,7 +424,7 @@ export default function LLMConfigurationPage() {
 
       <SettingsLayouts.Body>
         {hasProviders ? (
-          <Card>
+          <LegacyCard>
```
jmelahman, line 427: I like this. I almost think we should use a convention like `DEPRECRATEDCard` and add a note to the `web/AGENTS.md` to never introduce a new usage/import with the `DEPRECATED` prefix.

PR #9920, opened 2026-04-05
fix(mt): Revamp Preprovision

justin-tahara:
## Description

We are currently hitting some bottlenecks with the pre-provisioning workflow. We are trying to pre-provision the tenants but due to the large influx of new users, the DB operations cannot keep up. This is partially due to the fact that the amount of time to spawn a new tenant is quite high due to the amount of schema migrations that is being run.

This change introduces a `create_all` workflow that will create the current table state. The problem with this is that we are disregarding all of the insertions that are happening with the schema migrations so this also adds a seeding operation that pre-seeds all of the default values that have been added historically.

This should hopefully help us see a 40~80x speed up in the cloud for pre-provisioning new blank tenants.

## How Has This Been Tested?

This has been tested by utilizing the old provisioning workflow where we are running an entire schema migration and then comparing the final DB state to ensure there is parity. 

This also ensures that if others add insertions that they are caught prior to the schema migration PR's get in.

## Additional Options

---
## Summary by cubic
Replace slow Alembic-based tenant pre-provisioning with a fast create_all + seed workflow to cut setup time from ~80s to ~2s and reduce DB load.

- **New Features**
  - Added `fast_provision_tenant_schema` to create tables from models, deduplicate conflicting index/constraint names, serialize metadata changes with a lock, exclude public-only tables, stamp `alembic_version`, and seed defaults.
  - Introduced `seed_tenant_defaults` to insert default tools, Assistant persona + tool links, code interpreter server, hierarchy nodes for all `DocumentSource`, anonymous user, KG config, search settings, and add FTS tsvector columns + GIN indexes via SQL.
  - Switched `setup_tenant` to the fast path instead of running full Alembic migrations.
  - Added a parity test that compares tables/columns, seed data, and Alembic head between the fast path and full migrations (ignores nondeterministic fields and known orphaned columns).

- **Migration**
  - No deploy action required; future Alembic migrations will continue from the stamped head.
  - When adding migrations that insert default data, also update `seed_tenant_data.py` (the parity test will catch any drift).

<sup>Written for commit 83cc13d32a114a6a1dce4595eaf710035a4fb93b. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/external_dependency_unit/test_fast_provision_parity.py
+++ b/backend/tests/external_dependency_unit/test_fast_provision_parity.py
@@ -1,0 +1,304 @@
+"""
+Drift-detection test: verifies that fast tenant provisioning
+(metadata.create_all + seed data) produces the same result as
+running full Alembic migrations.
+
+If this test fails, it means a migration was added that either:
+  1. Creates/alters a table not reflected in models.py (schema drift)
+  2. Inserts seed data not replicated in seed_tenant_data.py (data drift)
+
+Fix by updating the relevant source of truth.
+"""
+
+import uuid
+from collections.abc import Generator
+
+import pytest
+from sqlalchemy import text
+from sqlalchemy.engine import Engine
+
+from ee.onyx.server.tenants.schema_management import fast_provision_tenant_schema
+from ee.onyx.server.tenants.schema_management import run_alembic_migrations
+from onyx.db.engine.sql_engine import SqlEngine
+
+
+# Tables whose seed data we compare between the two provisioning paths.
+# These are the tables that migrations insert default/seed rows into.
+SEED_TABLES = [
+    "tool",
+    "persona",
+    "persona__tool",
+    "code_interpreter_server",
+    "hierarchy_node",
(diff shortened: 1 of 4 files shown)


backend/tests/external_dependency_unit/test_fast_provision_parity.py
```diff
@@ -0,0 +1,304 @@
...
+            "Column differences — models.py is MISSING columns that "
+            "migrations create:\n" + "\n".join(errors)
+        )
+
+    def test_same_seed_data(self, migration_schema: str, fast_schema: str) -> None:
+        """Seed tables should contain the same data in both schemas."""
+        engine = _get_engine()
+        differences: list[str] = []
+
+        for table in SEED_TABLES:
```
jmelahman, line 256: Do we need to look specifically over `SEED_TABLES` or can we do all tables? Or maybe skip specific, known non-deterministic tables. I worry about `SEED_TABLES` drifting

jmelahman (approved):

PR #9952, opened 2026-04-06
fix: make default model global instead of per-provider

raunakab:
## Description

Make "default model" a global concept instead of per-provider, per [ENG-3934](https://linear.app/onyx-app/issue/ENG-3934).

- Remove "Default" tag from provider cards — no more default provider concept
- Remove provider sort that placed default provider first
- Replace per-provider `default_model_name` tracking in `ModelsField` with global default comparison (`globalDefault` prop)
- **Existing provider modals**: "Set as default" fires the API immediately and shows a success toast (no form submission needed)
- **New provider modals**: "Set as default" choice is deferred until after successful form submission via `pendingDefaultModelName`
- First provider auto-default still works via `shouldMarkAsDefault`
- Update `LLMProviderFormProps` interface: `defaultModelName` → `globalDefault`
- Update `SubmitLLMProviderParams`: `shouldMarkAsDefault` → `pendingDefaultModelName`
- Remove `currentDefaultModelName` param from `buildDefaultInitialValues`

Closes: https://linear.app/onyx-app/issue/ENG-3934/default-model-is-global-not-per-provider.

### DIFF
--- a/web/src/interfaces/llm.ts
+++ b/web/src/interfaces/llm.ts
@@ -124,8 +124,8 @@ export interface LLMProviderFormProps {
   shouldMarkAsDefault?: boolean;
   onOpenChange?: (open: boolean) => void;
 
-  /** The current default model name for this provider (from the global default). */
-  defaultModelName?: string;
+  /** The current global default model (provider_id + model_name). */
+  globalDefault?: DefaultModel | null;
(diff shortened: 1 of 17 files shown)


raunakab:
Stale.

PR #10048, opened 2026-04-10
fix(scim): add advisory lock to prevent seat limit race condition

nmgarza5:
## Description

IdPs like Okta send SCIM provisioning requests in parallel batches during user sync. The seat-availability check in `_check_seat_availability` was vulnerable to a TOCTOU race condition: N concurrent requests could each see "seats available", all insert, and push the tenant over its seat limit.

Adds a `pg_advisory_xact_lock` in `_check_seat_availability` that serializes concurrent seat checks within the same DB transaction. The lock is automatically released on COMMIT/ROLLBACK, so it only holds for the duration of the check + insert — no manual cleanup needed.

## How Has This Been Tested?

unit tests

## Additional Options

### DIFF
--- a/backend/ee/onyx/server/scim/api.py
+++ b/backend/ee/onyx/server/scim/api.py
@@ -11,6 +11,7 @@ require a valid SCIM bearer token.
 
 from __future__ import annotations
 
+import hashlib
 from uuid import UUID
 
 from fastapi import APIRouter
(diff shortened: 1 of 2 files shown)


backend/ee/onyx/server/scim/api.py
```diff
@@ -65,12 +67,19 @@
...
 logger = setup_logger()
 
 # Group names reserved for system default groups (seeded by migration).
 _RESERVED_GROUP_NAMES = frozenset({"Admin", "Basic"})
 
+# First key for the two-argument pg_advisory_xact_lock that serializes
+# seat-allocation checks during concurrent SCIM provisioning requests.
+# The second key is derived from the tenant ID so that different tenants
+# do not block each other.
+_SEAT_LOCK_KEY1 = 294837
```
justin-tahara: What is this random number?

nmgarza5: just an arbitrary number

nmgarza5: i'll update

justin-tahara (approved):

PR #10066, opened 2026-04-10
fix(mcp): prevent masked OAuth credentials from being stored on re-auth

jmelahman:
## Description

Closes https://discord.com/channels/1119886360506023946/1491443877092135044

## How Has This Been Tested?

Testing locally. ~~Maybe playwright? TBD~~

## Additional Options

---
## Summary by cubic
Prevents masked OAuth credentials from overwriting real secrets during MCP re-auth and server updates. Restores stored values when masks are detected to avoid false config changes.

- **Bug Fixes**
  - Added _restore_masked_oauth_credentials and used it in _connect_oauth and _upsert_mcp_server to swap masked client_id/client_secret with stored values before save/compare, preventing masked writes and unnecessary updates.

<sup>Written for commit 15093dba44940e53d1e3acf0adb897f9d3f34a69. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/server/features/mcp/api.py
+++ b/backend/onyx/server/features/mcp/api.py
@@ -392,6 +392,29 @@ async def _connect_oauth(
             detail=f"Server was configured with authentication type {auth_type_str}",
         )
 
+    # If the frontend sent back masked credentials (unchanged by the user),
+    # restore the real stored values so we don't overwrite them with masks.
+    if mcp_server.admin_connection_config:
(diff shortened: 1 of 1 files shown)


evan-onyx (approved):
I have strong reservations about this approach but this is important/valuable enough that let's just merge it and follow up later (imo we should be passing "unset" or similar to avoid this pattern of running a masking check and then overwriting values fromt he request)

PR #10087, opened 2026-04-12
refactor(indexing): Split document chunking into section-based strategies

Danelegend:
## Description
The current document chunking pipeline treats documents in a monolithic block, handling every document, each section and the different types of sections in roughly the same pipeline.

This breaks it down to create more containable responsibilities for what is responsbile for each of those components. The improvement is now that we can write different implementations based on Sections and chunk Sections in a particular way while staying abstracted from the broader pipeline.

We start with an implementation of the TextSectionChunker and ImageSectionChunker as currently exists

## How Has This Been Tested?
Previous unit tests ensure it is consistent.

## Additional Options

---
## Summary by cubic
Refactors document chunking into section-based strategies keyed by `SectionType`. `chunker.py` now delegates to `DocumentChunker` and uses shared `extract_blurb`, keeping behavior the same with tests pinning it.

- **Refactors**
  - Added `onyx.indexing.chunking` package exporting `DocumentChunker` and `extract_blurb`; internals include `SectionChunker` (ABC), `TextChunker`, `ImageChunker`, and shared models. Dispatcher routes TEXT → `TextChunker`, IMAGE → `ImageChunker`, and errors on unknown types.
  - `chunker.py` builds a `DocumentChunker` with existing splitters and calls `extract_blurb(title, blurb_splitter)`; removed inline helpers. Link fallback and mini-chunks are handled in `ChunkPayload.to_doc_aware_chunk`.
  - Tests now target `DocumentChunker` with `chonkie.SentenceChunker` and a char tokenizer; strict fallback is pinned via `text_section_chunker.STRICT_CHUNK_TOKEN_LIMIT`. Verified continuation flags, image handling (including no-link), ID ordering, metadata propagation, `clean_text`, None-text handling, and no trailing chunk after image/oversized sections.

<sup>Written for commit e7b0522ac349ee4e5e2083ea67a3b9e348611c73. Summary will update on new commits.</sup>


backend/onyx/indexing/chunking/section_chunker.py
```diff
@@ -0,0 +1,99 @@
...
+        return ""
+    return texts[0]
+
+
+def get_mini_chunk_texts(
+    chunk_text: str,
+    mini_chunk_splitter: SentenceChunker | None,
+) -> list[str] | None:
+    if mini_chunk_splitter and chunk_text.strip():
+        return cast(list[str], mini_chunk_splitter.chunk(chunk_text))
```
evan-onyx: why do we need to cast() here?

backend/onyx/indexing/chunking/image_section_chunker.py
```diff
@@ -0,0 +1,34 @@
...
+    ) -> SectionChunkerOutput:
+        assert section.image_file_id is not None
+
+        section_text = clean_text(str(section.text or ""))
+        section_link = section.link or ""
+
+        payloads = accumulator.flush_to_list()
+        payloads.append(
+            ChunkPayload(
+                text=section_text,
```
evan-onyx, line 25: probably good to keep the comments that explained why we do this from the original code

evan-onyx (approved):
nits

backend/onyx/indexing/chunking/section_chunker.py
Danelegend: Still need a cast since mini_chunk_splitter.chunk has the function def `def chunk(self, text: str) -> Union[Sequence[Chunk], Sequence[str]]:`. Changed to cast convert to Sequence[str], which forces Chunk to go to str (custom __str__ method), and then wrap that cast in a list to bring it all together.

PR #10092, opened 2026-04-12
chore(indexing): Add kind enum to Section model

Danelegend:
## Description
Adds a new kind attribute to the section model.
Allows for differentiation between sections without relying on base-class.

Will be used for section type specific chunk dispatching.

## How Has This Been Tested?
Unit

## Additional Options

---
## Summary by cubic
Add a required `type` enum to `Section` (`text` or `image`) and propagate it through indexing and tool conversions for type-aware chunking. Normalize `text` to "" so downstream code never sees `None`.

- **Refactors**
  - Added `SectionType` and required `type` on `Section`; `TextSection`/`ImageSection` set it via `Literal`.
  - Propagated `type` when constructing base `Section`s in indexing (`process_image_sections`) and in search-like tool conversion.
  - Ensured `text` is always a string for both text and image sections; updated unit tests to cover new typing and image chunk cases.

- **Migration**
  - When creating a `Section`, set `type` (`SectionType.TEXT`/`SectionType.IMAGE`) or use `TextSection`/`ImageSection` which set it by default.

<sup>Written for commit d0d22bc115c3abc5c34e94c8aa44ca15fd12461f. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/connectors/models.py
+++ b/backend/onyx/connectors/models.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from enum import Enum
 from typing import Any
 from typing import cast
+from typing import Literal
 
 from pydantic import BaseModel
 from pydantic import Field
@@ -34,8 +35,7 @@ class ConnectorMissingCredentialError(PermissionError):
 
 
 class SectionKind(str, Enum):
-    """Discriminator for Section subclasses.
-    """
+    """Discriminator for Section subclasses."""
 
     TEXT = "text"
     IMAGE = "image"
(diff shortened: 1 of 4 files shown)


backend/onyx/connectors/models.py
```diff
@@ -33,9 +34,17 @@ def __init__(self, connector_name: str) -> None:
         )
 
 
+class SectionKind(str, Enum):
```
evan-onyx: slightly prefer SectionType or SectionVariant just bc my brain isn't used to seeing "Kind" in code

evan-onyx (approved):
left a nit, up to you

PR #10095, opened 2026-04-12
feat(indexing): Add TabularSection

Danelegend:
## Description
Adds a new TabularSection model and util functions to convert spreadsheets and CSV into the new Section.

## How Has This Been Tested?
Unit

## Additional Options

---
## Summary by cubic
Adds a `TabularSection` model and tooling to index `.xlsx`, `.csv`, and `.tsv` as structured sections. Enables per-sheet Excel indexing, TSV-to-CSV normalization, and fixes length counting, chunk flushing, and ClickUp comment handling.

- **New Features**
  - Added `TabularSection` (`type: tabular`) with `text` (CSV) and `link`; `DocumentBase.sections` now accepts any `Sequence`.
  - Added `tabular_file_to_sections(...)`: `.xlsx` → one section per non-empty sheet (`link=sheet:{title}`); `.csv`/`.tsv` → one section with decoded text and `link` (fallback to `file_name`), with TSV re-serialized to CSV; includes `is_tabular_file(...)`. Empty files are skipped; decode errors are logged and raised.
  - Added `xlsx_sheet_extraction(...)` returning `(csv_text, sheet_title)` per non-empty sheet; `xlsx_to_text(...)` now composes from this.

- **Bug Fixes**
  - Document chunker flushes leftover buffered text into a final payload.
  - `DocumentBase.get_total_char_length` sums `text` for `TextSection` and `TabularSection` (including processed).
  - ClickUp connector safely appends comment sections when `sections` is a `Sequence`.

<sup>Written for commit c108d0ce98db62430d1b28959dbe2ddd92c48f7c. Summary will update on new commits.</sup>


backend/onyx/connectors/models.py
```diff
@@ -172,7 +186,7 @@ class DocumentBase(BaseModel):
     """Used for Onyx ingestion api, the ID is inferred before use if not provided"""
 
     id: str | None = None
-    sections: list[TextSection | ImageSection]
+    sections: Sequence[TextSection | ImageSection | TabularSection]
```
Danelegend, line 189: Adding TabularSection to the sections union broke every connector that builds a narrower list locally (list[TextSection | ImageSection] → can't pass to a field typed list[TextSection | ImageSection | TabularSection] because list is invariant). 

Instead of widening each connector (which would be required for every new Section also), make use of Sequence type

backend/onyx/connectors/cross_connector_utils/tabular_section_utils.py
```diff
@@ -0,0 +1,50 @@
+from typing import IO
+
+from onyx.connectors.models import TabularSection
+from onyx.file_processing.extract_file_text import file_io_to_text
+from onyx.file_processing.extract_file_text import xlsx_sheet_extraction
+from onyx.file_processing.file_types import OnyxFileExtensions
+from onyx.utils.logger import setup_logger
+
+logger = setup_logger()
+
```
Danelegend, line 12: To be used by connectors to convert TSV, CSV and XLSX to TabularSections

backend/onyx/connectors/clickup/connector.py
```diff
@@ -171,7 +171,10 @@ def _get_all_tasks_filtered(
                         document.metadata[extra_field] = task[extra_field]
 
                 if self.retrieve_task_comments:
-                    document.sections.extend(self._get_task_comments(task["id"]))
+                    document.sections = [
```
evan-onyx, line 174: seems unrelated, remove?

backend/onyx/connectors/cross_connector_utils/tabular_section_utils.py
```diff
@@ -0,0 +1,64 @@
...
+    if lowered.endswith(".xlsx"):
+        return [
+            TabularSection(link=f"sheet:{sheet_title}", text=csv_text)
+            for csv_text, sheet_title in xlsx_sheet_extraction(
+                file, file_name=file_name
+            )
+        ]
+
+    if lowered.endswith((".csv", ".tsv")):
```
evan-onyx: nit: prefer inverting this to reduce indentation

backend/onyx/connectors/cross_connector_utils/tabular_section_utils.py
```diff
@@ -0,0 +1,64 @@
...
+        try:
+            text = file_io_to_text(file).strip()
+        except Exception as e:
+            logger.warning(f"Failed to decode {file_name}: {e}")
```
evan-onyx: we might want to actually start raising here, in the past we just didn't prioritize CSVs but since we're trying to make processing them robust now we probably want to know when this breaks (and potentially gracefully handle some more narrow cases)

backend/onyx/connectors/models.py
evan-onyx: this is fine, but longer term we should just have a `AnySectionType` type alias for this and all connectors can return that

evan-onyx (approved):
some nits

backend/onyx/connectors/clickup/connector.py
Danelegend: Necessary for the Sequence type change since Sequence does not have a .extend

backend/onyx/connectors/models.py
Danelegend: I would disagree. For some connector that only needs to care about TextSection, does it make sense for it to need to create a list[Everything] and thus need to know about more than it really cares about?

PR #10097, opened 2026-04-12
feat(chat): scrollable tables with overflow fade

nmgarza5:
## Description

Tables in chat messages now stay within their panel width instead of breaking out via container queries (`100cqw`). When a table's content exceeds the available width, a subtle fade appears on the right edge and a scrollbar reveals on hover. Columns size to their widest content (`whitespace-nowrap` on `th`/`td`) so nothing truncates — the user scrolls horizontally instead.

**Changes:**
- `custom-code-styles.css`: Replaced `100cqw` breakout with contained `overflow-x: auto`, hover-reveal scrollbar, and a `::after` fade pseudo-element driven by `data-overflows`
- `markdownUtils.tsx`: New `ScrollableTable` component wraps tables, detects overflow via `ResizeObserver` + scroll listener, and toggles the fade when content extends past the right edge (hidden when scrolled to end)

## How Has This Been Tested?

https://github.com/user-attachments/assets/8c37a5e1-b47c-4d3b-a2d4-e119a72a4ef2

Dark/light modes + custom backgrounds

## Additional Options

### DIFF
--- a/web/src/app/app/message/custom-code-styles.css
+++ b/web/src/app/app/message/custom-code-styles.css
@@ -258,23 +258,102 @@ pre[class*="language-"] {
   scrollbar-color: #4b5563 #1f2937;
 }
 
+/* Card wrapper — holds the background, border-radius, padding, and fade overlay.
+   Does NOT scroll — the inner .markdown-table-breakout handles that. */
+.markdown-table-card {
+  position: relative;
+  background: var(--background-neutral-01);
+  border-radius: 0.5rem;
+  padding: 0.5rem 0;
+}
+
 /*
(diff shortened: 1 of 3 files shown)


web/src/app/app/message/custom-code-styles.css
raunakab, line 1: Should we potentially move this file to `web/app/css` instead?

raunakab (approved):

web/src/app/app/message/custom-code-styles.css
nmgarza5: potentially. this css is specific for chat message rendering so i kinda prefer to keep here and isolated

PR #10112, opened 2026-04-13
chore: remove top-level Formik from ChatPreferencesPage

raunakab:
## Description

Remove the top-level `<Formik>` wrapper from `ChatPreferencesPage`. The page never submits as a form — every field saves independently via `onBlur` or `onCheckedChange` callbacks. Formik was only being used as a state container (`onSubmit: () => {}`), which is not its intended purpose.

## Screenshots + Videos

No UI changes; screenshots / videos not required.

## Additional Options

---
## Summary by cubic
Removed the top-level `<Formik>` from `ChatPreferencesPage`; fields now read from `settings` and save on blur/change. Also re-sync team text fields with external updates and restore the `disable_default_assistant` switch id to keep E2E tests stable.

- **Refactors**
  - Replaced `SwitchField`/`InputSelectField` with `Switch`/`InputSelect` reading directly from `settings`; removed page-level `<Formik>/<Form>` (system-prompt modal keeps its own).
  - Moved `company_name`/`company_description` to local `useState` with save-on-blur; swapped `InputTypeInField`/`InputTextAreaField` for `InputTypeIn`/`InputTextArea`.
  - Rewrote `NumericLimitField` with local state and an `initialValue` prop; `FileSizeLimitFields` now pass initial values and call `saveSettings` directly.

- **Bug Fixes**
  - Re-sync `company_name`/`company_description` when the settings context changes, unless the user is mid-edit.
  - Restore `id="disable_default_assistant"` on the `Switch` to keep Playwright E2E selector working.

<sup>Written for commit 433dc0176535955a53f0842db11fd178ce1fa18e. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/refresh-pages/admin/ChatPreferencesPage.tsx
+++ b/web/src/refresh-pages/admin/ChatPreferencesPage.tsx
@@ -965,6 +965,7 @@ function ChatPreferencesForm() {
                     nonInteractive
                   >
                     <Switch
+                      id="disable_default_assistant"
                       checked={s.disable_default_assistant ?? false}
                       onCheckedChange={(checked) => {
                         void saveSettings({
--- a/web/src/refresh-pages/admin/ChatPreferencesPage.tsx
+++ b/web/src/refresh-pages/admin/ChatPreferencesPage.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { markdown } from "@opal/utils";
-import React, { useCallback, useRef, useState } from "react";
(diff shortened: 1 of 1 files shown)


nmgarza5 (approved):

raunakab:
Addressed the review comment about stale local state in 8b5903eeb. Added `useEffect` hooks that re-sync `companyName`/`companyDescription` when the settings context changes externally, but only when the user isn't mid-edit (local value still matches the last-saved value).

PR #10115, opened 2026-04-13
fix(google): handle JSON credential payloads in KV storage

### LINKED ISSUE #10114: fix Google credential uploads storing strings in encrypted JSON fields
## Bug
Uploading Google app credentials or service-account keys for Google connectors can fail because the code writes serialized JSON strings into fields that are later treated as encrypted JSON objects.

In practice this surfaces during Google Drive admin setup on the credential upload endpoints and results in errors like:

```text
TypeError: EncryptedJson column expected dict, got str
```

## Affected code paths
The issue appears in `backend/onyx/connectors/google_utils/google_kv.py`:

- `upsert_google_app_cred()` stores `app_credentials.json()`
- `upsert_service_account_key()` stores `service_account_key.json()`
- `build_service_account_creds()` also embeds `service_account_key.json()` into the credential payload

Those values are JSON strings, but the backing storage / encrypted JSON path expects dict-like JSON payloads.

There is a second compatibility issue on the read path:

- `get_auth_url()`
- `get_google_app_cred()`
- `get_service_account_key()`
- `get_google_creds()` in `google_auth.py`

currently assume stored payloads are strings and call `json.loads(...)` unconditionally.

## Suggested fix
1. Store Google credential payloads with `model_dump(mode=\"json\")` instead of `.json()`.
2. Make the read path accept both legacy string payloads and new dict payloads so existing installs do not break.
3. Add a small regression test covering:
   - dict writes
   - legacy string reads
   - auth URL creation from either storage format
   - service-account credential loading from either storage format

## Why I’m confident this is real
I reproduced this against a live self-hosted deployment while setting up Google Drive auth, traced it back to these exact helper methods, patched the code locally, and verified that the credential flow succeeded after the change.

If helpful, I already have a minimal patch + regression tests prepared and can open a PR referencing this issue.

jack-larch:
## Summary
- store Google app and service-account credential payloads as JSON objects instead of serialized strings when writing to KV / credential storage
- keep the read path backward-compatible by accepting both legacy string payloads and new dict payloads in the Google auth helpers
- add targeted regression tests for dict writes, legacy reads, auth URL generation, and service-account credential loading

Closes #10114.

## Test plan
- [x] `python3 -m uv run --group backend --group dev ruff check backend/onyx/connectors/google_utils/google_kv.py backend/onyx/connectors/google_utils/google_auth.py backend/tests/unit/onyx/connectors/google_utils/test_google_credential_storage.py`
- [x] `python3 -m uv run --group backend --group dev black --check backend/onyx/connectors/google_utils/google_kv.py backend/onyx/connectors/google_utils/google_auth.py backend/tests/unit/onyx/connectors/google_utils/test_google_credential_storage.py`
- [x] `python3 -m uv run --group backend --group dev pytest backend/tests/unit/onyx/connectors/google_utils/test_google_credential_storage.py`

Made with [Cursor](https://cursor.com)

---
## Summary by cubic
Switch KV storage of Google app and service-account credentials to JSON objects and keep reads backward-compatible to prevent auth failures with mixed payload types. Closes #10114.

- **Bug Fixes**
  - Store credentials as dicts using `model_dump(mode="json")`; load via a `_load_google_json` helper.
  - Make `get_google_creds`, `get_google_app_cred`, `get_service_account_key`, and `get_auth_url` accept both dict and legacy string payloads.
  - Add unit tests for dict writes, legacy reads, auth URL state handling, and service-account credential loading.

<sup>Written for commit b41c4830615010a312f92c8a0cc5ef2e327868a0. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/connectors/google_utils/google_auth.py
+++ b/backend/onyx/connectors/google_utils/google_auth.py
@@ -103,8 +103,12 @@ def get_google_creds(
             GoogleOAuthAuthenticationMethod.UPLOADED.value,
         )
 
-        credentials_dict_str = credentials[DB_CREDENTIALS_DICT_TOKEN_KEY]
-        credentials_dict = json.loads(credentials_dict_str)
+        credentials_payload = credentials[DB_CREDENTIALS_DICT_TOKEN_KEY]
+        credentials_dict = (
+            json.loads(credentials_payload)
+            if isinstance(credentials_payload, str)
+            else credentials_payload
+        )
 
         # only send what get_google_oauth_creds needs
         authorized_user_info = {}
@@ -151,10 +155,14 @@ def get_google_creds(
                 }
     elif DB_CREDENTIALS_DICT_SERVICE_ACCOUNT_KEY in credentials:
         # SERVICE ACCOUNT
-        service_account_key_json_str = credentials[
+        service_account_key_payload = credentials[
             DB_CREDENTIALS_DICT_SERVICE_ACCOUNT_KEY
         ]
-        service_account_key = json.loads(service_account_key_json_str)
+        service_account_key = (
+            json.loads(service_account_key_payload)
+            if isinstance(service_account_key_payload, str)
+            else service_account_key_payload
+        )
 
         service_creds = ServiceAccountCredentials.from_service_account_info(
             service_account_key, scopes=GOOGLE_SCOPES[source]
--- a/backend/onyx/connectors/google_utils/google_kv.py
+++ b/backend/onyx/connectors/google_utils/google_kv.py
@@ -1,4 +1,5 @@
 import json
+from typing import Any
 from typing import cast
 from urllib.parse import parse_qs
 from urllib.parse import ParseResult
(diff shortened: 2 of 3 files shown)


wenxi-onyx:
Thanks for digging into this and putting up the fix. The analysis in #10114 is spot-on. I'm going to push a narrower version on top of what you have (#10160), wanted to explain why.

Scope change:
- I'm keeping the google_kv.py writer/reader changes (they fix the actual TypeError on first-time upload), but reverting the build_service_account_creds change and the google_auth.py changes.
- The TypeError: EncryptedJson column expected dict, got str only fires on two code paths, upsert_google_app_cred and upsert_service_account_key, which write directly into the KVStore.encrypted_value column (an EncryptedJson). 
- build_service_account_creds doesn't actually hit that validator: it puts a JSON string as a value inside a dict, and EncryptedJson only checks the top-level dict-ness. So build_service_account_creds → Credential.credential_json → get_google_creds was never broken; changing it was a stylistic cleanup bundled in with the bug fix.

Unbundling that cleanup buys a few things:
- get_google_creds keeps its credentials: dict[str, str] annotation. With your version, values can now be dicts, so that annotation becomes a lie and would need to widen to dict[str, Any]
- No behavior change for the Credential table path. Existing installs keep storing/reading the same shape they always have, so we only have one "legacy format" migration concern (the KV store) instead of two.
- Smaller diff, smaller review surface, smaller blast radius.

The tradeoff is that the nested-JSON-string-inside-dict pattern in credential_json survives. It's a wart, but it's a stable wart, and it doesn't block the upload path. If we want to clean it up later, it'll be a separate PR

jack-larch:
thanks!

PR #10141, opened 2026-04-13
fix(widget): surface descriptive error for trial account 429

nmgarza5:
## Description

Trial accounts have an API call limit of 0, so the chat widget would display a cryptic "Failed to send message: 429". Two changes:

1. **Backend**: When a trial account with limit=0 hits the API call check, return a clear message: "API access is not available on trial accounts. Please upgrade to a paid plan to use the API and chat widget."
2. **Widget**: Both `createSession` and `streamMessage` error handlers now read the response body JSON `detail` field instead of showing the generic HTTP status text.

## Additional Options

### DIFF
--- a/backend/onyx/server/usage_limits.py
+++ b/backend/onyx/server/usage_limits.py
@@ -2,7 +2,6 @@
 
 from collections.abc import Callable
 
-from fastapi import HTTPException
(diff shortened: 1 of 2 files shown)


evan-onyx (approved):
thanks!

PR #10159, opened 2026-04-14
fix(chat): eliminate long-lived DB session in multi-model worker threads

nmgarza5:
## Description

Multi-model LLM workers held a single DB session open for the entire streaming loop (minutes). Cloud infrastructure (load balancers, PgBouncer, idle-in-transaction timeouts) drops idle connections, causing `OperationalError` during session cleanup — even when the LLM response completed successfully. This turned successful responses into errors.

Audit showed only 4 callsites used the passed-in session, none requiring transactional consistency. Each now opens short-lived sessions on demand:
- `construct_tools()` opens its own session via `get_session_with_current_tenant_if_none`
- `add_memory()` / `update_memory_at_index()` open their own sessions (still accept explicit session for tests)
- `run_llm_loop` / `run_deep_research_llm_loop` no longer accept `db_session`
- `_run_model` no longer wraps everything in `get_session_with_current_tenant()`

Also adds `_safe_close_session()` in `sql_engine.py` as a safety net to catch disconnect errors during session cleanup.

### Before

```
_run_model(model_idx):
    with get_session_with_current_tenant() as thread_db_session:  # OPENS HERE
        construct_tools(..., db_session=thread_db_session)         # ~100ms
        run_llm_loop(..., db_session=thread_db_session)            # 30s - 5min+
            → streams tokens (no DB)
            → tool calls (SearchTool opens its own session anyway)
            → add_memory(db_session=thread_db_session)             # rare, ~10ms
    # CLOSES HERE — connection held entire time, idle 99%
```

### After

```
_run_model(model_idx):
    construct_tools(...)                    # opens session, reads, closes (~100ms)
    run_llm_loop(...)                       # NO session held during streaming
        → streams tokens (no DB, no connection)
        → tool calls (SearchTool/OpenURL open their own as before)
        → add_memory()                      # opens session, commits, closes (~10ms)
    # No long-lived connection — each DB op is a brief checkout-use-return cycle
```

**Linear:** [ENG-4018](https://linear.app/onyx-app/issue/ENG-4018/eliminate-long-lived-db-session-in-multi-model-worker-threads)

## Additional Options

### DIFF
--- a/backend/onyx/db/engine/sql_engine.py
+++ b/backend/onyx/db/engine/sql_engine.py
@@ -11,7 +11,7 @@ from sqlalchemy import event
 from sqlalchemy import pool
 from sqlalchemy.engine import create_engine
 from sqlalchemy.engine import Engine
-from sqlalchemy.exc import OperationalError
+from sqlalchemy.exc import DBAPIError
 from sqlalchemy.orm import Session
 
 from onyx.configs.app_configs import DB_READONLY_PASSWORD
@@ -354,12 +354,13 @@ def _safe_close_session(session: Session) -> None:
     open for minutes.  If the underlying connection is dropped by cloud
     infrastructure (load-balancer timeouts, PgBouncer, idle-in-transaction
     timeouts, etc.), the implicit rollback in Session.close() raises
-    OperationalError.  Since the work is already complete, we log and move
-    on — SQLAlchemy internally invalidates the connection for pool recycling.
+    OperationalError or InterfaceError.  Since the work is already complete,
+    we log and move on — SQLAlchemy internally invalidates the connection
+    for pool recycling.
     """
     try:
         session.close()
-    except OperationalError:
+    except DBAPIError:
         logger.warning(
             "DB connection lost during session cleanup — the connection will be invalidated and recycled by the pool."
         )
--- a/backend/onyx/db/engine/sql_engine.py
+++ b/backend/onyx/db/engine/sql_engine.py
@@ -11,6 +11,7 @@ from sqlalchemy import event
 from sqlalchemy import pool
 from sqlalchemy.engine import create_engine
 from sqlalchemy.engine import Engine
+from sqlalchemy.exc import OperationalError
 from sqlalchemy.orm import Session
 
 from onyx.configs.app_configs import DB_READONLY_PASSWORD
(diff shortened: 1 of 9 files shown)


nmgarza5:
@greptile

backend/onyx/db/engine/sql_engine.py
```diff
@@ -346,6 +347,25 @@ def get_session_with_shared_schema() -> Generator[Session, None, None]:
...
+    open for minutes.  If the underlying connection is dropped by cloud
+    infrastructure (load-balancer timeouts, PgBouncer, idle-in-transaction
+    timeouts, etc.), the implicit rollback in Session.close() raises
+    OperationalError or InterfaceError.  Since the work is already complete,
+    we log and move on — SQLAlchemy internally invalidates the connection
+    for pool recycling.
+    """
+    try:
+        session.close()
+    except DBAPIError:
```
wenxi-onyx, line 363: Is this the only possible error when attempting to close a connection that is closed already?

backend/onyx/db/memory.py
```diff
@@ -83,47 +84,49 @@ def get_memories(user: User, db_session: Session) -> UserMemoryContext:
...
-    db_session.add(memory)
-    db_session.commit()
-    return memory
+        memory = Memory(
+            user_id=user_id,
+            memory_text=memory_text,
+        )
+        db_session.add(memory)
+        db_session.commit()
+        return memory
```
wenxi-onyx: Nit/smell: since we now open/close a session for this function, returning memory results in a detached ORM. This works fine because `memory.id` is a native attribute of the object (and is kept in system memory). However, if _someone else_ attempts to access a relationship on the object (like memory.user), it will result in a `DetachedInstnaceError`.

**tl;dr I think this function should just return memory.id instead of passing back the full object with session closed.. Alternatively, can return pydantic model MemorySnapshot (which is what the caller does with the id anyways**

wenxi-onyx (commented):
Two small comments - will defer to @evan-onyx for final pass though.

backend/onyx/db/memory.py
wenxi-onyx: This same, problematic, pattern exists in `insert_slack_bot`. I don't recall off the top of my head, but this caused me some headache a while back 😅

evan-onyx (approved):
should test breakages partway through chats to see if they look okay

backend/onyx/db/engine/sql_engine.py
nmgarza5: this is a wrapper/catch all exception for things like connection issues, incorrect SQL syntax, or database constraint violations. It should be fine here

backend/onyx/db/memory.py
nmgarza5: good comments. i'll look into them. thanks!

PR #10166, opened 2026-04-14
fix: use static provider list instead of querying be

wenxi-onyx:
## Description

- Provider cards would not render until `/api/admin/llm/built-in/options` returned wellknownproviders --> this endpoint is a bit troll and takes ~1s to respond on our cloud
  - This would cause the Custom LLM Provider card to render and then shift after ~1s once all other providers were found
- We have a hardcoded list of providers on the FE that determine the sorting order. We can just use this to render
- What isn't great is that new provider implementations have an extra place they need to define their provider, but maybe that's good as it forces devs to think about the final order on the page?
## Additional Options

---
## Summary by cubic
Render LLM provider cards from a static list instead of querying the backend, removing the load-time layout shift on the configuration page. Provider order is now fixed by `PROVIDER_DISPLAY_ORDER`.

- **Bug Fixes**
  - Replaced backend fetch with static `PROVIDER_DISPLAY_ORDER`; removed the `useWellKnownLLMProviders` hook and related descriptor types.
  - Updated `NewProviderCard` to take `providerName` and resolve details via `getProvider`.

<sup>Written for commit 8525c1e18c166de35b1dfa579dcefe0a7d2d01ef. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/hooks/useLLMProviders.ts
+++ b/web/src/hooks/useLLMProviders.ts
@@ -114,26 +114,6 @@ export function useAdminLLMProviders() {
 }
 
 /**
- * Fetches the catalog of well-known (built-in) LLM providers.
- *
- * Hits `GET /api/admin/llm/built-in/options` which returns the static
- * list of provider descriptors that Onyx ships with out of the box
- * (OpenAI, Anthropic, Vertex AI, Bedrock, Azure, Ollama, OpenRouter,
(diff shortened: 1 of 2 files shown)


raunakab (approved):

wenxi-onyx:
@greptileai re-review this

PR #10167, opened 2026-04-14
fix: uploaded files as knowledge source

evan-onyx:
## Description

When uploading and attaching large files to an assistant and also selecting some subset of uploaded knowledge to attach to an agent, we were previously unable to actually retrieve info from the attached user files (we were not passing user_file as a source filter through to the vector db). Now we do so, and make explicit the uploaded files as a searchable source type in the actions popover.

One note: this toggle will now exist whether the uploaded files in question fit in the context or not. It's still meaningful (decides whether the search tool can see the files), but there's a potential weird UX thing where a user might expect that toggling it off means the agent can't see the uploaded files at all. That isn't the case, based on our assumptions about what people want when they attach uploaded files to assistants (we directly inject small uploaded files into the context).

Screenshot 2026-04-14 at 2 50 25 PM

## How Has This Been Tested?

tested in UI and added frontend unit and backend integration tests

## Additional Options

---
## Summary by cubic
Fixes knowledge search so attached uploaded files are actually queried and shown as an “Uploaded Files” source. Also scopes and resets source toggles per agent to avoid stale selections when switching agents.

- **Bug Fixes**
  - Add `user_file` to persona `knowledge_sources` when user files are attached; map federated connectors to their non‑federated sources.
  - Scope toggles and filter building to the agent’s `knowledge_sources` (default agent shows all); propagate selected sources to the `source_type` filter.
  - Re-initialize source preferences when the available source set changes. Added tests: backend unit + integration for `knowledge_sources`; frontend tests for preferences and filter payloads.

- **New Features**
  - Show “Uploaded Files” with a new icon and updated docs link in the Actions popover. UX: the toggle controls search visibility; small files may still be injected into context even if toggled off.

<sup>Written for commit 1a9386e34364a040797cf01c5a862191c00f9839. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/server/features/persona/models.py
+++ b/backend/onyx/server/features/persona/models.py
@@ -185,6 +185,10 @@ class MinimalPersonaSnapshot(BaseModel):
         for doc_set in persona.document_sets:
             for cc_pair in doc_set.connector_credential_pairs:
                 sources.add(cc_pair.connector.source)
+            for fed_ds in doc_set.federated_connectors:
+                non_fed = fed_ds.federated_connector.source.to_non_federated_source()
+                if non_fed is not None:
+                    sources.add(non_fed)
 
         # Sources from hierarchy nodes
         for node in persona.hierarchy_nodes:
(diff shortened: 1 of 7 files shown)


Bo-Onyx (approved):
Maybe good add a screen shot to show the toggle change for uploaded file

PR #10168, opened 2026-04-14
feat: add GCS backend with Workload Identity support for onyx backend

Mberga14:
## Description

Added support for google storage client with support work workload identity - this is a security improvemenet as admins hosting onyx on Google compute engine (or GKE) do not need to generate long lived HMAC keys anymore to operate the s3-like google storage endpoint, and leave the authentication to IAM + metadata server (just like AWS 😄 )

## How Has This Been Tested?

Set up onyx locally with ADC and linked it to a GCS bucket - mocked LLM response, started a chat and uploaded some images/documents. Checked the GCS bucket contained the new elements

## Additional Options

---
## Summary by cubic
Adds a native Google Cloud Storage backend to the Onyx file store with ADC/Workload Identity, so GCP deployments can use IAM instead of HMAC keys. `s3` remains the default; `s3` and `postgres` backends are unchanged.

- **New Features**
  - Added `GCSBackedFileStore`; enable with `FILE_STORE_BACKEND=gcs`.
  - Auth: ADC/Workload Identity (default), service account key file, or inline JSON.
  - Env vars: `GCS_FILE_STORE_BUCKET_NAME`, `GCS_FILE_STORE_PREFIX`, `GCS_PROJECT_ID`, `GCS_SERVICE_ACCOUNT_KEY_PATH`, `GCS_SERVICE_ACCOUNT_KEY_JSON`.
  - Parity with S3: init, save/read/delete, size, rename, list; wired into factory; unit tests; env template; docs for GCS setup, IAM, and auth.
  - Added `google-cloud-storage`.

- **Migration**
  - Set `FILE_STORE_BACKEND=gcs` and `GCS_FILE_STORE_BUCKET_NAME` (optionally `GCS_PROJECT_ID`); prefer ADC/Workload Identity or provide a service account key.
  - IAM needed: `roles/storage.objectAdmin` (object ops); `roles/storage.admin` if creating buckets.
  - No changes for existing `s3` or `postgres` users.

<sup>Written for commit 4f7c72394e4cfc90f40e333472e2d0c8a060a8b7. Summary will update on new commits. <a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/10168?utm_source=github">Review in cubic</a></sup>


Mberga14:
@cubic-dev-ai re-review please

uv.lock
justin-tahara, line 1: I don't think we should be changing this right? It might just be a rebasing issue.

backend/onyx/file_store/file_store.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
justin-tahara: Again not needed I believe

backend/onyx/file_store/file_store.py
justin-tahara, line 1: Feels like there is a lot of reformatting. Hopefully this can be just simplified

backend/onyx/file_store/gcs_file_store.py
```diff
@@ -0,0 +1,331 @@
...
+
+        Authentication priority:
+        1. Service account key file (GCS_SERVICE_ACCOUNT_KEY_PATH)
+        2. Inline service account JSON (GCS_SERVICE_ACCOUNT_KEY_JSON)
+        3. Application Default Credentials (Workload Identity, metadata server,
+           gcloud CLI). Project ID is auto-resolved from the environment.
+        """
+        if self._gcs_client is None:
+            try:
+                from google.cloud import storage
```
justin-tahara: import errors would be brought up s runtime errors here

backend/onyx/file_store/gcs_file_store.py
```diff
@@ -0,0 +1,331 @@
...
+        )
+        if len(key) == 1024:
+            logger.info(f"File name was too long and was truncated: {file_name}")
+        return key
+    
+    def initialize(self) -> None:
+        """Initialize the GCS file store by ensuring the bucket exists."""
+        from google.api_core.exceptions import Forbidden, NotFound
+
+        client = self._get_gcs_client()
```
justin-tahara: Is there any other error cases we should be adding here?

backend/onyx/file_store/gcs_file_store.py
```diff
@@ -0,0 +1,331 @@
...
+    def _get_gcs_client(self) -> GCSClient:
+        """Initialize GCS client if not already done.
+
+        Authentication priority:
+        1. Service account key file (GCS_SERVICE_ACCOUNT_KEY_PATH)
+        2. Inline service account JSON (GCS_SERVICE_ACCOUNT_KEY_JSON)
+        3. Application Default Credentials (Workload Identity, metadata server,
+           gcloud CLI). Project ID is auto-resolved from the environment.
+        """
+        if self._gcs_client is None:
```
justin-tahara: Might require a lock if two celery workers are creating clients. I think one will get garbage collected but might be work a comment

backend/onyx/file_store/gcs_file_store.py
```diff
@@ -0,0 +1,331 @@
...
+                    from google.api_core.exceptions import NotFound
+
+                    if isinstance(e, NotFound):
+                        logger.warning(
+                            f"delete_file: File {file_id} not found in GCS "
+                            f"(key: {file_record.object_key}), "
+                            "cleaning up database record."
+                        )
+                    else:
+                        raise
```
justin-tahara: Just feels a bit odd to import then check if the error is part of NotFound

backend/tests/unit/file_store/test_file_store.py
justin-tahara, line 1: Feels like we should add tests for read_file, delete_file, change_file_id, get_file_size, or initialize

backend/onyx/file_store/gcs_file_store.py
```diff
@@ -0,0 +1,331 @@
...
+                    display_name=old_file_record.display_name,
+                    file_origin=old_file_record.file_origin,
+                    file_type=old_file_record.file_type,
+                    bucket_name=self._bucket_name,
+                    object_key=new_object_key,
+                    db_session=db_session,
+                    file_metadata=file_metadata,
+                )
+
+                source_blob.delete()
```
justin-tahara: I feel like the source_blob.delete() should happen after the DB is updated so that we are 100% confident that we are safe to delete the object

uv.lock
Mberga14: I did add `google-cloud-storage` to the packages, so the new entry for backend should be present in the lock?

backend/onyx/file_store/gcs_file_store.py
Mberga14: I have followed the pattern of the S3BackendFileStore class - would be wise to do the same pattern there as well

backend/onyx/file_store/gcs_file_store.py
Mberga14: I do not think so - I will change the code so it explicitly catches the import error and propagates, all other cases are cought by `RuntimeErrors` as configuration issues

backend/onyx/file_store/gcs_file_store.py
Mberga14: You are right, that was a bit akward, fixed.

backend/onyx/file_store/gcs_file_store.py
Mberga14: Agreed 👍

PR #10199, opened 2026-04-14
fix(chat): fix fade gradient missing on last multi-model panel

nmgarza5:
## Description

Two fixes for non-preferred panel display in multi-model selection mode:

1. **Overflow detection:** The previous implementation measured overflow (`scrollHeight > clientHeight`) inside ref callbacks. Ref callbacks fire during React's commit phase — before the browser has applied the `maxHeight` style to the DOM. This means the measurement reflected the unconstrained panel height, so `scrollHeight === clientHeight` and the panel was never marked as overflowing. The last panel was most reliably affected because it rendered last in the `.map()` and had the least time for layout to settle. Fixed by moving the measurement to `useLayoutEffect`, which fires after React commits all DOM mutations (including style props), so `maxHeight` is applied and overflow is detectable.

2. **Fade style:** Replaced the opaque gradient overlay div with a CSS `mask-image` fade on the panel wrapper. Content now fades out smoothly to transparent instead of having a jarring solid-color overlay on top.

**Linear:** [ENG-4022](https://linear.app/onyx-app/issue/ENG-4022/multi-model-last-panel-footer-cut-off-and-missing-fade-on-long)

## How Has This Been Tested?

Screenshot 2026-04-14 at 3 58 40 PM

## Additional Options

### DIFF
--- a/web/src/app/app/message/MultiModelResponseView.tsx
+++ b/web/src/app/app/message/MultiModelResponseView.tsx
@@ -1,6 +1,13 @@
 "use client";
 
-import { useState, useCallback, useMemo, useEffect, useRef } from "react";
+import {
+  useState,
+  useCallback,
+  useMemo,
+  useEffect,
+  useLayoutEffect,
+  useRef,
+} from "react";
 import { FullChatState } from "@/app/app/message/messageComponents/interfaces";
 import { Message } from "@/app/app/interfaces";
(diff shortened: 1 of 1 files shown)


nmgarza5:
@greptile

Subash-Mohan (approved):

PR #10203, opened 2026-04-14
fix(chat): keep model selector popover open until max models reached

nmgarza5:
## Description

Improvements to the multi-model selector UX:

1. **Popover stays open:** Previously closed after each model addition. Now stays open until `MAX_MODELS` (3) is reached.

2. **Surviving model preserved:** When removing models from A/B/C down to just C, the active model reverted to the user's default. Now `removeModel` updates `llmManager.currentLlm` to the surviving model.

3. **Wider popover:** Changed width from `"lg"` to `"xl"` to match the `LLMPopover` in the message footer.

4. **Consistent pill style:** Both single and multi-model pills now use `SelectButton`. Single-model just omits the X icon.

5. **New `select-input` variant:** Created a new interactive variant that rests at `bg-background-neutral-00` (matches input bar) and shows `bg-background-neutral-03` on hover/open. Pills now blend seamlessly with the input bar at rest. This is to be mock compliant.

**Linear:** [ENG-4019](https://linear.app/onyx-app/issue/ENG-4019/model-selector-keep-popover-open-and-preserve-order-when-adding-models)

## Additional Options

### DIFF
--- a/web/lib/opal/src/core/interactive/stateful/components.tsx
+++ b/web/lib/opal/src/core/interactive/stateful/components.tsx
@@ -15,6 +15,7 @@ type InteractiveStatefulVariant =
   | "select-heavy"
   | "select-card"
   | "select-tinted"
+  | "select-input"
   | "select-filter"
   | "sidebar-heavy"
   | "sidebar-light";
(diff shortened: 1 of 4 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

Subash-Mohan (approved):

PR #10207, opened 2026-04-15
feat(indexing): notify admins when connector enters repeated error state

nmgarza5:
## Description

Connectors currently die silently. When a connector fails repeatedly and enters the "repeated error" state, admins only discover it when they happen to check the connector list in the UI, or when a user complains about stale search results. There's no proactive notification.

This adds an in-app admin notification at the exact point where a connector transitions to repeated error state. The notification includes:
- Connector name and source type (e.g. "Google Drive", "Confluence")
- CC pair ID for linking to the connector's detail page

**How it works:**

The transition point is in `check_for_indexing` (runs every 15s). When `is_in_repeated_error_state()` detects consecutive failures and the connector wasn't already in that state, three things now happen:
1. `set_cc_pair_repeated_error_state(True)` — existing behavior
2. `create_notification()` — **new**: creates an admin notification with `user_id=None` (visible to all admins)
3. `task_logger.error()` — **new**: structured ERROR log for cloud monitoring

The notification uses `additional_data={"cc_pair_id": cc_pair.id}` for deduplication — if the same connector re-enters error state after recovery, a new notification is created. If it's already in error state and hasn't been dismissed, the timestamp is updated (existing `create_notification` behavior).

Uses the new `NotificationType.CONNECTOR_REPEATED_ERRORS` enum value.

**Linear:** [ENG-3983](https://linear.app/onyx-app/issue/ENG-3983/connector-invalid-admin-notification)

## Additional Options

### DIFF
--- a/backend/onyx/background/celery/tasks/docprocessing/tasks.py
+++ b/backend/onyx/background/celery/tasks/docprocessing/tasks.py
@@ -50,6 +50,7 @@ from onyx.configs.constants import AuthType
 from onyx.configs.constants import CELERY_GENERIC_BEAT_LOCK_TIMEOUT
 from onyx.configs.constants import CELERY_INDEXING_LOCK_TIMEOUT
 from onyx.configs.constants import MilestoneRecordType
+from onyx.configs.constants import NotificationType
 from onyx.configs.constants import OnyxCeleryPriority
 from onyx.configs.constants import OnyxCeleryQueues
 from onyx.configs.constants import OnyxCeleryTask
@@ -85,6 +86,7 @@ from onyx.db.indexing_coordination import INDEXING_PROGRESS_TIMEOUT_HOURS
 from onyx.db.indexing_coordination import IndexingCoordination
 from onyx.db.models import IndexAttempt
 from onyx.db.models import SearchSettings
+from onyx.db.notification import create_notification
 from onyx.db.search_settings import get_current_search_settings
 from onyx.db.search_settings import get_secondary_search_settings
 from onyx.db.swap_index import check_and_perform_index_swap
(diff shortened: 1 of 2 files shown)


backend/onyx/background/celery/tasks/docprocessing/tasks.py
```diff
@@ -848,6 +849,33 @@ def check_for_indexing(self: Task, *, tenant_id: str) -> int | None:
...
+                    source = cc_pair.connector.source.value
+                    create_notification(
+                        user_id=None,
+                        notif_type=NotificationType.CONNECTOR_REPEATED_ERRORS,
+                        db_session=db_session,
+                        title=f"Connector '{connector_name}' has entered repeated error state",
+                        description=(
+                            f"The {source} connector has failed repeatedly and "
+                            f"has been flagged. Check the connector's indexing "
+                            f"history for details."
```
justin-tahara: Might be good to clarify it is in the advanced section when they click on the connector. Would be even better if we could just have a link to the connector. I think since we have the connector we in theory could build the URL but that's more of a nice to have.

backend/onyx/background/celery/tasks/docprocessing/tasks.py
```diff
@@ -848,6 +849,33 @@ def check_for_indexing(self: Task, *, tenant_id: str) -> int | None:
...
+                        ),
+                        additional_data={"cc_pair_id": cc_pair.id},
+                    )
```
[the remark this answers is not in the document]
justin-tahara: Maybe we can attach something unique like the date to ensure that the notifications are unique?

justin-tahara (commented):
minor comments

backend/onyx/background/celery/tasks/docprocessing/tasks.py
nmgarza5: makes sense, i'll get a URL included

nmgarza5:
@greptile

justin-tahara (approved):

PR #10209, opened 2026-04-15
feat(mt): Infra setup for Redis Set (1/3)

justin-tahara:
## Description

Beginning a stack of PRs to introduce a redis set to the multi-tenant architecture. 

This PR focuses on introducing the framework required to work towards a redis set implementation where the set only contains the tenants that require work to be done.

This change is meant to help with the load we are seeing on the cloud to better utilize the database for better operations.

## How Has This Been Tested?

Adding a new set of tests

## Additional Options

---
## Summary by cubic
Sets up infrastructure for tenant work-gating using a Redis sorted set to track recently active tenants, with runtime flags to compute and optionally enforce gating. No behavior change by default; enforcement is off and requires an explicit runtime flag.

- **New Features**
  - Added `active_tenants` Redis ZSET helpers: `mark_tenant_active`, `get_active_tenants(ttl_seconds)`, and `cleanup_expired(ttl_seconds)`; fail-open on Redis errors, no-op in single-tenant mode, cloud-prefixed key.
  - Introduced runtime controls:
    - `get_tenant_work_gating_enabled()` defaults to `ENABLE_TENANT_WORK_GATING` (shadow-only).
    - `get_tenant_work_gating_enforce()` is Redis-only with a hard default of False.
    - `get_tenant_work_gating_ttl_seconds()` and `get_tenant_work_gating_full_fanout_interval_seconds()` have env defaults with Redis runtime overrides.
  - Extended Redis client allowlist for ZSET ops (`zadd`, `zrangebyscore`, `zremrangebyscore`, `zscore`, `zcard`).
  - Added tests for membership writes/refresh, TTL filtering, cleanup, fail-open vs empty set, single-tenant no-op, and cloud key prefixing.

<sup>Written for commit 7463fe1f4944008f61498816e75a38f22ab17ff5. Summary will update on new commits.</sup>


backend/onyx/configs/app_configs.py
```diff
@@ -1125,6 +1125,30 @@ def get_current_tz_offset() -> int:
...
+ENABLE_TENANT_WORK_GATING = (
+    os.environ.get("ENABLE_TENANT_WORK_GATING", "").lower() == "true"
+)
+
+# Membership TTL for the `active_tenants` sorted set. Members older than this
+# are treated as inactive by the gate read path. Must be > (full-fanout
+# cadence × base task schedule) so self-healing re-adds a genuinely-working
+# tenant before their membership expires. Default 30 min.
+TENANT_WORK_GATING_TTL_SECONDS = int(
+    os.environ.get("TENANT_WORK_GATING_TTL_SECONDS", str(30 * 60))
```
acaprau: super nit but don't think you need to wrap this product in a str().

backend/onyx/configs/app_configs.py
```diff
@@ -1125,6 +1125,30 @@ def get_current_tz_offset() -> int:
...
+)
+
+# Every Nth fanout cycle the generator bypasses the gate and dispatches to
+# every non-gated tenant, letting consumers re-populate the active set. This
+# is the self-healing path that replaces the earlier hourly reconcile task.
+TENANT_WORK_GATING_FULL_FANOUT_EVERY_N = int(
```
acaprau: i don't feel too strongly about this but i think it might be a bit more robust if instead of counting every N iterations of a task, for which maybe the scheduling gets messed up and it's unclear how often N happens, we use time instead, so if S seconds have elapsed since the last time we did a full fanout, independent of how many times a task has run, we do another full fanout. something to think about.

backend/tests/external_dependency_unit/tenant_work_gating/__init__.py
acaprau, line 1: nit but don't think init files like this are needed for our version of python.

backend/onyx/redis/redis_tenant_work_gating.py
```diff
@@ -0,0 +1,104 @@
...
+        logger.exception("get_active_tenants failed")
+        return None
+
+    return {m.decode() if isinstance(m, bytes) else m for m in raw}
+
+
+def cleanup_expired(ttl_seconds: int) -> int:
+    """Remove members older than `ttl_seconds` from the set. Optional
+    memory-hygiene helper — correctness does not depend on calling this, but
+    without it the set grows unboundedly as old tenants accumulate. Returns
```
acaprau, line 90: no way for redis to do this for us?

backend/onyx/server/runtime/onyx_runtime.py
```diff
@@ -139,6 +142,84 @@ def get_doc_permission_sync_multiplier() -> float:
...
+            return default
+
+        try:
+            value = int(cast(bytes, raw).decode())
+            return value if value > 0 else default
+        except ValueError:
+            return default
+
+    @staticmethod
+    def get_tenant_work_gating_full_fanout_every_n() -> int:
```
acaprau: just to be explicit this pr does not define set_* methods to go with these gets?

acaprau (approved):

backend/onyx/redis/redis_tenant_work_gating.py
justin-tahara: Redis has `EXPIRE` on whole keys and `HEXPIRE` on hash fields but nothing that is per member on the sorted set.

backend/onyx/server/runtime/onyx_runtime.py
justin-tahara: This PR does not.

PR #10213, opened 2026-04-15
fix(voice): send TTS text in POST body instead of query params

nmgarza5:
## Description

Three fixes for TTS playback:

1. **403 from nginx:** Text was sent as a URL query parameter. Long responses exceeded nginx's URL length limit → 403 Forbidden. Moved text to a JSON POST body via `SynthesizeRequest` Pydantic model.

2. **Provider errors not surfacing:** Provider errors (e.g. ElevenLabs `text_too_long`) were thrown inside the streaming response generator — after 200 headers were already sent — so the frontend got a broken audio stream with no error message. Now the first audio chunk is pulled *before* returning the `StreamingResponse`. If the provider rejects immediately, the error surfaces as a proper HTTP error with a JSON body. A generic `_extract_provider_error()` helper parses the human-readable message from any provider's JSON error response. Frontend parses `detail` from the response and displays it in the toast.

3. **Duplicate console logging:** The error was logged twice — once in `useVoicePlayback` (`onError` callback) and again in `TTSButton` (`useEffect`). Removed the duplicate `console.error` from `TTSButton`; it now only fires the toast.

**Linear:** [ENG-4023](https://linear.app/onyx-app/issue/ENG-4023/multi-model-tts-playback-intermittently-fails-with-403)

## How Has This Been Tested?

Screenshot 2026-04-14 at 9 08 11 PM

## Additional Options

### DIFF
--- a/backend/onyx/server/manage/voice/user_api.py
+++ b/backend/onyx/server/manage/voice/user_api.py
@@ -8,6 +8,7 @@ from fastapi import File
 from fastapi import UploadFile
 from fastapi.responses import StreamingResponse
 from pydantic import BaseModel
+from pydantic import Field
 from sqlalchemy.orm import Session
 
 from onyx.auth.permissions import require_permission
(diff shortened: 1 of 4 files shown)


nmgarza5:
@greptile

jessicasingh7 (approved):

PR #10259, opened 2026-04-16
feat(indexing): Total descriptor chunk emitted during Tabular Indexing

Danelegend:
## Description
This PR adds a chunk that answers questions about the total, average, sum, etc around a TabularSection.
This improves hits on common aggregate questions by putting them in more semantically meaningful sentences rather than just values.

In chunk should look like this:
```
Totals and overall aggregates across all rows. This sheet can answer
whole-dataset questions about total, overall, grand total, sum across
all, average, combined, mean, minimum, maximum, and count of values.
Column <name>: total (sum across all rows) = X, average = Y,
minimum = Z, maximum = W, count = N.
[repeat for each numeric column]
Column <cat_name> most frequent value: <val> (<count> occurrences).
[repeat for each categorical column]
```

## How Has This Been Tested?
Unit Tests + Manual

## Additional Options

---
## Summary by cubic
Adds a totals descriptor to tabular indexing that writes whole‑sheet aggregates in plain sentences to improve retrieval for sum/avg/min/max/count and most‑frequent queries. Also derives headers for header‑only CSVs and appends totals after the sheet descriptor.

- **New Features**
  - Emits a totals chunk with per‑numeric‑column sum/avg/min/max/count, top categorical value, and a total row count; prefixed by the section heading plus a standard totals header (kept on split chunks).
  - Formats underscored headers with a friendly alias (e.g., MTTR_hours → “MTTR_hours (MTTR hours)”) and only emits totals when data rows exist.

- **Refactors**
  - Added `analysis.py` with `analyze_sheet`/`SheetAnalysis` to infer types and compute numeric values and categorical value counts for reuse.
  - Introduced `util.pack_lines` and `util.label`; `sheet_descriptor` now takes `headers`/`analysis`/`heading`. The chunker reads headers via `read_csv_header`, computes analysis once, and emits the sheet descriptor and totals as continuations after content. Expanded tests cover formatting, prefix behavior, split packing, continuation ordering, and header‑only CSV handling.

<sup>Written for commit 136094e72e9eb92f3497d988728897feace27a7f. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/indexing/chunking/tabular_section_chunker/analysis.py
+++ b/backend/onyx/indexing/chunking/tabular_section_chunker/analysis.py
@@ -54,10 +54,10 @@ def analyze_sheet(headers: list[str], parsed_rows: list[ParsedRow]) -> SheetAnal
             continue
 
         # Date: every value parses as a date — fold into the sheet-wide range.
-        dates = [_try_date(v) for v in values]
-        if all(d is not None for d in dates):
-            dmin = min(filter(None, dates))
-            dmax = max(filter(None, dates))
+        dates = _try_all_dates(values)
+        if dates:
+            dmin = min(dates)
+            dmax = max(dates)
             a.date_min = dmin if a.date_min is None else min(a.date_min, dmin)
             a.date_max = dmax if a.date_max is None else max(a.date_max, dmax)
             continue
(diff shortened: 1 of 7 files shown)


backend/onyx/indexing/chunking/tabular_section_chunker/analysis.py
```diff
@@ -0,0 +1,102 @@
...
+
+        # Numeric: every value parses as a number.
+        nums = _try_all_numeric(values)
+        if nums is not None:
+            a.numeric_cols.append(idx)
+            a.numeric_values[idx] = nums
+            continue
+
+        # Date: every value parses as a date — fold into the sheet-wide range.
+        dates = [_try_date(v) for v in values]
```
evan-onyx: nit: can we do a try_all_dates pattern here too so we don't have to waste work if the first one isn't a date

evan-onyx (approved):

PR #10261, opened 2026-04-16
fix(openpyxl): Adding support for xlsm

justin-tahara:
## Description

Adding support for xlsm

## How Has This Been Tested?

Adding in tests

## Additional Options

---
## Summary by cubic
Add support for `.xlsm` so macro‑enabled Excel files are parsed like `.xlsx` and produce tabular sections. Uses `openpyxl` with a shared spreadsheet extension list.

- **New Features**
  - Recognize `.xlsm` as a spreadsheet/tabular file in `OnyxFileExtensions` and parse it with `openpyxl` in the file connector and tabular utils.
  - Added tests for `.xlsm` detection and section parsing; unknown extensions still raise.

- **Refactors**
  - Introduced `SPREADSHEET_EXTENSIONS` (`.xlsx`, `.xlsm`) and updated checks to use it; `TABULAR_EXTENSIONS` and `DOCUMENT_EXTENSIONS` now include this set.

<sup>Written for commit f2cffccc818419551c8fdf2508dfd496a89fef8c. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/connectors/cross_connector_utils/tabular_section_utils.py
+++ b/backend/onyx/connectors/cross_connector_utils/tabular_section_utils.py
@@ -41,7 +41,7 @@ def tabular_file_to_sections(
     """
     lowered = file_name.lower()
 
-    if lowered.endswith(".xlsx"):
+    if lowered.endswith((".xlsx", ".xlsm")):
         return [
             TabularSection(
                 link=link or file_name,
--- a/backend/onyx/connectors/cross_connector_utils/tabular_section_utils.py
+++ b/backend/onyx/connectors/cross_connector_utils/tabular_section_utils.py
@@ -41,7 +41,7 @@ def tabular_file_to_sections(
     """
     lowered = file_name.lower()
 
-    if lowered.endswith((".xlsx", ".xlsm")):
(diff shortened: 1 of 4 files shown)


backend/onyx/file_processing/file_types.py
```diff
@@ -57,6 +57,7 @@ class OnyxFileExtensions:
         ".csv",
         ".tsv",
         ".xlsx",
```
Danelegend, line 57: Could we turn this into a SPREADSHEET_EXTENSIONS set here. I think there are a couple others like .xls that can be added down the line, but would need to be tested

backend/onyx/connectors/cross_connector_utils/tabular_section_utils.py
```diff
@@ -41,7 +41,7 @@ def tabular_file_to_sections(
     """
     lowered = file_name.lower()
 
-    if lowered.endswith(".xlsx"):
+    if lowered.endswith((".xlsx", ".xlsm")):
```
Danelegend: With the comment I left below, we can then use OnyxFileExtensions.SPREADSHEET instead

Danelegend (approved):

PR #10264, opened 2026-04-16
feat(img): Editing User Uploaded Images

Danelegend:
## Description
Currently the only reference image provided is that that the llm generates. This changes the flow to let the llm choose any image within the context window as the reference image. This allows it to edit / reference images that it generated, and images that the user uploaded. 

## How Has This Been Tested?
Tests + Manual

## Additional Options

---
## Summary by cubic
Let the LLM choose any image in the conversation as the edit reference by exposing image file_ids and updating `generate_image` to use them. Removes the “use the last image” fallback and supports edits from user uploads and prior generations with clear limits.

- **New Features**
  - Add `[attached image — file_id: <id>]` before each image so the LLM can copy ids.
  - `generate_image` accepts `reference_image_file_ids`; preserves order, de-duplicates, and keeps the first N per provider limits (primary edit source stays first).
  - Tool schema and guidance explain where ids come from, not to invent them, and to leave the field unset for fresh generations.
  - Clear errors for invalid ids or when the provider doesn’t support image edits.

- **Refactors**
  - Removed the automatic “most recent image” default, `ImageGenerationToolOverrideKwargs`, and all runner image-history logic.
  - Simplified the runner by dropping id extraction/overrides; added focused tests for reference resolution and removed obsolete tests.

<sup>Written for commit 5f738e3c240c6b12c24dccde0ccda5e1ae40fbd7. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/chat/llm_step.py
+++ b/backend/onyx/chat/llm_step.py
@@ -818,7 +818,10 @@ def translate_history_to_llm_format(
                     )
                 ]
 
-                # Add image parts
+                # Add image parts. Each image is preceded by a text tag
+                # carrying its file_id so the LLM can reference the image by
+                # ID when calling tools like generate_image (which expects
+                # reference_image_file_ids to edit a specific image).
                 for img_file in msg.image_files:
                     if img_file.file_type == ChatFileType.IMAGE:
                         try:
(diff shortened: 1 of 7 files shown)


justin-tahara (approved):
Maybe update the title of the PR to be Editing User Uploaded Images

PR #10272, opened 2026-04-16
feat(web): Refreshed `Index Settings Page`

raunakab:
## Description

Reskins the admin **Index Settings** page using Opal components and consolidates the prior multi-page configuration flow into a single page under `web/src/refresh-pages/admin/IndexSettingsPage/`.

### Workflow change

Previously, configuring indexing required hopping across multiple admin pages — embedding-model selection in one place, retrieval optimization in another, image-processing toggles in a third. Admins had to context-switch between pages and remember which knob lived where. This pulls the entire surface into one page with three logical sections:

- **Embedding Model** — pick a cloud-based or self-hosted model, or add a custom one. The current model is shown collapsed by default; "View All Models" expands a tabbed picker (Cloud-based / Self-hosted) with search.
- **Retrieval Optimization** — Multipass Indexing and Contextual Retrieval. The Contextual Retrieval toggle gates an LLM picker for the model used to generate per-chunk context.
- **Image Processing** — Extract & Caption Images. The toggle gates a Captioning LLM picker (vision-capable models only) and a Max Image Size selector.

### Persistence: form submission vs. auto-save

Two persistence patterns are used on this page, chosen based on whether the change requires re-indexing:

- **Embedding Model + Retrieval Optimization → Formik form, batched submit.** Changing the embedding model or any retrieval-optimization field requires a full re-index of all existing documents, which can take hours or days. To avoid accidentally triggering this, dirty fields in these two sections are staged in Formik form state and only persisted on explicit submit. The submit flow surfaces a switchover-strategy selector (re-index all connectors first, re-index active only, switch immediately, or apply going forward) and an "Apply & Re-index" CTA. A dirty banner at the top of the page warns that staged changes require re-indexing.

- **Image Processing → auto-save on change.** These fields take effect on documents indexed going forward — existing captions are baked into prior embeddings — so they have no re-indexing implications. Each control persists immediately on change. No batching, no submit step, no warning banner.

This split keeps the UX honest: the dirty banner and switchover prompt only appear for changes that actually require re-indexing. Image-processing tweaks feel like ordinary settings toggles because, behaviorally, they are.

### Section-level disabling

Every section is wrapped in a unified `CloudDisabled` guard that handles three disable triggers and surfaces a tooltip explaining *why* the section is unavailable:

- **Onyx Cloud** manages the setting upstream.
- **Prerequisite missing** — e.g. Contextual Retrieval needs at least one configured LLM; Image Processing needs a vision-capable LLM. Tooltips link to the LLM configuration page so the admin can resolve the prerequisite without leaving context.
- **Re-index in progress** — embedding-model and retrieval-optimization edits are blocked until the in-flight re-index completes or is cancelled. A dedicated banner replaces the dirty banner during this window.

### Provider credential modals

Cloud embedding providers each have a credentials modal (API key, endpoint, etc.) wrapped in Formik+Yup for consistent validation. The same modal is reused for initial connection, editing existing credentials, and on-demand connection when the user picks a model from a not-yet-connected provider.

Addresses: https://linear.app/onyx-app/project/admin-panel-llm-provider-configuration-page-187cb469820c.

### Choosing a Self-Hosted Model

https://github.com/user-attachments/assets/1d4bdb4e-8a57-45ae-b81d-4b7419a133c5

### Choosing a *Custom* Self-Hosted Model

TODO

### Updating "Contextual Retrieval" Options

https://github.com/user-attachments/assets/5ca7490c-ce41-40b7-add6-5b400c04a807

### Choosing a Cloud-Based Model

https://github.com/user-attachments/assets/0b91997d-6817-4dee-9bb2-d119313ed07d

### Updating Image Processing

TODO

### Cancelling Re-indexes

https://github.com/user-attachments/assets/eca16925-a6a8-404e-9d56-ad0518c2dde3

### Reverting selections after they've been chosen

https://github.com/user-attachments/assets/8cf8a783-660c-4b0c-b9ca-b3ab2d1e2e86


web/lib/opal/src/core/disabled/components.tsx
```diff
@@ -43,50 +43,50 @@ interface DisabledProps extends React.HTMLAttributes<HTMLElement> {
...
+  side = "right",
   ref,
   ...rest
 }: DisabledProps) {
   const showTooltip = disabled && tooltip;
   const enableClick = allowClick || showTooltip;
 
   const wrapper = (
-    <Slot
+    <div
```
jmelahman, line 76: Please make this a separate change.

backend/onyx/server/manage/search_settings.py
```diff
@@ -27,9 +27,6 @@
 from onyx.db.search_settings import update_search_settings_status
 from onyx.document_index.factory import get_all_document_indices
 from onyx.document_index.factory import get_default_document_index
-from onyx.file_processing.unstructured import delete_unstructured_api_key
```
jmelahman, line 30: Let's drop this since it's already merged at HEAD, https://github.com/onyx-dot-app/onyx/pull/10283. Best to keep API breaking changes (even if benign) separate from these large PRs

web/src/hooks/useSearchSettings.ts
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Fetches the list of LLM models available for contextual RAG, including
+ * per-model token cost.
+ */
+export function useLLMContextualCosts() {
+  return useSWR<LLMContextualCost[]>(
+    LLM_CONTEXTUAL_COST_ADMIN_URL,
+    errorHandlingFetcher
+  );
```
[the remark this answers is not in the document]
jmelahman: Seems valid?

web/lib/opal/src/core/disabled/components.tsx
```diff
@@ -1,15 +1,15 @@
 import "@opal/core/disabled/styles.css";
```
jmelahman, line 1: I do think these changes are causing a visual regression with at least these buttons, https://onyx-playwright-artifacts.s3.us-east-2.amazonaws.com/reports/pr-10272/24861786323/admin/index.html,

Image

jmelahman:
Some high-level feedback:

1. I feel somewhat strongly this change should come with _some_ tests. We don't need 100% coverage, but something basic that serves as a starting point for future tests to reference. Just prove it works a little bit. The description claims,

> No new automated tests — this is a UI-only reskin with stubbed click handlers on the future picker. A Playwright test will land alongside the real provider/model switch-over wire-up.

but I am not sure that is true. If there were tests and the spec did change, it would be apparent. The only changes to tests are seemingly unrelated changes to onboarding tests (and they seem related to the `Disabled` changes which ideally would be made separately, if at all).

2. Just QA'ing this blindly, this `Contextual Retrieval` is 1 of 2 non-disabled switches. Clicking it does nothing for me. The other non-disabled switch immediately errors with a generic `Failed to update settings` (the backend reads `OnyxError INSUFFICIENT_PERMISSIONS: You do not have the required permissions for this action.`, but I'm the only admin and running the EE version, so idk.. I dropped my DB and things are working a little better. The `Contextual Retrieval` still does nothing except toast that it succeeded,

20260423_21h05m20s_grim

  a. I also think all disabled items should have an associated tooltip for why they are disabled. It's quite re-assuring and informative to new users.

  b. this is probably an less realistic and nitpicky issue, but I'm not sure if we should allow Image Processing to be enabled if the user doesn't have any vision models configured? My dropdown is just empty,

20260423_21h02m16s_grim

4. I would really, really, really appreciate if any intentional, functional changes are called out clearly in the PR description. The description is stale and only has updates from Cubic even after most of what is documented has been pulled out into separate changes and are no longer relevant.

    a. FWIW, I had Codex review this specifically looking for functional regressions and this was the report (I haven't looked at it very thoroughly for validity),

> Index Settings reskin — review findings
> Most of these are functional regressions from the old admin/embeddings + admin/configuration/search pages rather than type or lint errors. tsc --noEmit passes.
>
> High
> 1. “Apply without Re-index” is a silent no-op.
> In web/src/refresh-pages/admin/IndexSettingsPage/index.tsx handleApply, the SWITCHOVER_NONE branch only calls toast.success("Settings applied") and clears local state — no API call. Users see the toast but the model selection is never persisted. Either remove the option or wire it to an endpoint that persists the new model without triggering re-index.
>
> 2. LiteLLM / Azure provider setup is incomplete.
>
> - connectEmbeddingProvider (web/src/lib/indexing/svc.ts) only sends provider_type, api_key, api_url. Azure requires api_version and deployment_name too — the Azure modal doesn’t collect them, so Azure embeddings can’t actually be configured from this page.
> - The LiteLLM modal collects modelName, modelDim, queryPrefix, passagePrefix, normalize, but none of those state values are ever sent anywhere. The custom-model form is purely decorative.
> - The PUT payload also includes is_default_provider / is_configured, which don’t exist on CloudEmbeddingProviderCreationRequest and are silently dropped.
>
> Medium
> 3. No re-index progress or cancel UI.
> The old UpgradingPage handled secondarySearchSettings, POST /api/search-settings/cancel-new-embedding, and the reindexing progress table. None of that exists on the new page, yet the page copy still promises “You can monitor the re-indexing progress on this page.” Users mid-switchover have no way to see status or cancel.
>
> 4. Editing existing provider credentials is broken by masking.
> GET /api/admin/embedding/embedding-provider returns api_key via get_value(apply_mask=True) (e.g. "my-v...here"). The edit modals initialize form state from existingCredentials.api_key ?? "" and submit that value through test-embedding, which will fail. Users can’t tweak api_url / api_version without also retyping the key. Initialize the field empty (or with a masked placeholder) and only send api_key when the user actually changes it.
>
> Low
> 5. Disconnect modal uses the DOM global name.
> In ProviderGroup, name is referenced inside the disconnect modal copy and a useCallback deps array. TypeScript compiles because window.name: string is globally typed, but at runtime it resolves to "", producing “Disconnect ” and “…provider .”. Should be provider.displayName.
>
> 6. Captioning LLM dropdown is inert.
Marked TODO(@raunakab): wire up but visible to end users with hard-coded value="" / onValueChange={() => {}}. Either wire it up or hide it until ready.
>
> Additional nits
> - Pre-connect test sends model_name: "" for non-OpenAI providers. connectEmbeddingProvider calls testEmbedding({ modelName: "" }); only provider_type === "openai" gets a fallback model name, so Connect for Cohere/Voyage/Google/Nomic may fail at the test-embedding step with a confusing error before the key is persisted.
> - Dead code. providerName is declared in StandardProviderModal, AzureProviderModal, and LiteLLMProviderModal but never read. configOnlyProviders is computed in IndexSettingsPage and never used.

web/lib/opal/src/core/disabled/styles.css
```diff
@@ -9,8 +10,15 @@
...
+  @apply self-stretch;
+}
+
 [data-opal-disabled] {
   @apply cursor-not-allowed select-none;
+}
+
+[data-opal-disabled] > * {
   pointer-events: none;
 }
```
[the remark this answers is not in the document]
raunakab: The `Disabled` component changes were extracted into separate PRs and have since landed on `main`; this branch no longer touches `web/lib/opal/src/core/disabled/` (`git diff origin/main...HEAD -- web/lib/opal/src/core/disabled` is empty). Resolving as no longer relevant to this PR.

web/src/lib/indexing/svc.ts
```diff
@@ -0,0 +1,129 @@
...
+ */
+export async function setNewSearchSettings(
+  model: EmbeddingModelDescriptor
+): Promise<Response> {
+  return await fetch("/api/search-settings/set-new-search-settings", {
+    method: "POST",
+    headers: { "Content-Type": "application/json" },
+    body: JSON.stringify({
+      ...model,
+      multipass_indexing: false,
```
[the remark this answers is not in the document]
raunakab: Addressed in 9a372dd55a:
- `enable_contextual_rag` is now plumbed through from the form values (no longer hardcoded).
- `embedding_precision` is dropped from the request body — the backend defaults it to `EmbeddingPrecision.FLOAT` (`backend/onyx/indexing/models.py:196`), so behavior is unchanged.
- `multipass_indexing: false` is intentionally kept hardcoded for this PR; the UI surfaces Multipass Indexing as `temporarily unavailable`.

web/src/hooks/useSearchSettings.ts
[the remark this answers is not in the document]
raunakab: Fixed in 8288b19. Added `llmContextualCost` to `SWR_KEYS` and migrated `useLLMContextualCosts` over; also removed the now-dead `LLM_CONTEXTUAL_COST_ADMIN_URL` constant. (`EMBEDDING_PROVIDERS_ADMIN_URL` does not exist anywhere in the codebase — `embeddingProviders` is already registered in `SWR_KEYS`.)

raunakab:
Thanks for the thorough review! A few updates:

**Tests (#1):** Playwright tests are coming — they'll land alongside the provider/model switchover wire-up. Agreed this needs coverage.

**Re-index progress (#3):** There is intentionally no full progress chart on the new page. We do have a small in-progress UI (a warning banner with a cancel button and a link to the connectors page to monitor indexing status). The stale copy promising a progress table has been removed.

**PR description (#4):** Updated — the description now reflects the current state of the PR.

web/src/refresh-pages/admin/IndexSettingsPage/index.tsx
```diff
@@ -0,0 +1,1681 @@
...
+      onSelectModel,
+      onDeselectModel,
+      connectModal,
+      provider.deprecated,
+      isCloud,
+      setPendingConnectModel,
+    ]
+  );
+
+  console.log("pendingConnectModel", pendingConnectModel);
```
jmelahman: rm?

jmelahman (approved):

web/src/refresh-pages/admin/IndexSettingsPage/index.tsx
raunakab: Removed!

PR #10285, opened 2026-04-16
chore(indexing): default embedding_precision to FLOAT

jessicasingh7:
## Description

Default \`embedding_precision\` to \`FLOAT\` (float32) instead of \`BFLOAT16\` on newly created \`SearchSettings\` rows.

OpenSearch ignores \`embedding_precision\` entirely. In \`backend/onyx/document_index/opensearch/opensearch_document_index.py\` the parameter is explicitly annotated \`# noqa: ARG002\` and vectors are stored as float32 regardless of the configured precision. Defaulting new rows to BFLOAT16 would create a lie in the DB (claimed BFLOAT16, actually stored as float32). Defaulting to FLOAT makes the DB match reality.

**Changes:**
- \`IndexingSetting.embedding_precision\` in \`backend/onyx/indexing/models.py\` — add default \`EmbeddingPrecision.FLOAT\`
- \`SearchSettings.embedding_precision\` in \`backend/onyx/db/models.py\` — add ORM default \`EmbeddingPrecision.FLOAT\`
- Update comments in \`backend/onyx/configs/embedding_configs.py\` and reorder generated variants so FLOAT is listed first

Linear: [ENG-4026](https://linear.app/onyx-app/issue/ENG-4026)

## How Has This Been Tested?

- [ ] Newly-created \`SearchSettings\` row without explicit \`embedding_precision\` defaults to \`FLOAT\` in the DB
- [ ] Existing rows with \`BFLOAT16\` remain unchanged after deploy
- [ ] Existing search_settings tests pass (fixtures set precision explicitly)
- [ ] Vespa-based tests pass

## Additional Options

### DIFF
--- a/backend/onyx/configs/embedding_configs.py
+++ b/backend/onyx/configs/embedding_configs.py
@@ -126,25 +126,28 @@ _BASE_EMBEDDING_MODELS = [
 
 # Automatically generate both FLOAT and BFLOAT16 versions of all models
 SUPPORTED_EMBEDDING_MODELS = [
-    # BFLOAT16 precision versions
-    *[
-        SupportedEmbeddingModel(
-            name=model.name,
-            dim=model.dim,
-            index_name=f"{model.index_name}_bfloat16",
-            embedding_precision=EmbeddingPrecision.BFLOAT16,
-        )
-        for model in _BASE_EMBEDDING_MODELS
-    ],
     # FLOAT precision versions
-    # NOTE: need to keep this one for backwards compatibility. We now default to
-    # BFLOAT16.
+    # NOTE: FLOAT (float32) is the default. OpenSearch (the target document
+    # index) ignores embedding_precision and stores vectors as float32
+    # regardless.
     *[
         SupportedEmbeddingModel(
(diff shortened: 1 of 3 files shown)


backend/onyx/configs/embedding_configs.py
```diff
@@ -126,25 +126,28 @@ class SupportedEmbeddingModel(BaseModel):
...
-    # BFLOAT16 precision versions
+    # FLOAT precision versions
+    # NOTE: FLOAT (float32) is the default. OpenSearch (the target document
+    # index) ignores embedding_precision and stores vectors as float32
+    # regardless.
     *[
         SupportedEmbeddingModel(
             name=model.name,
             dim=model.dim,
-            index_name=f"{model.index_name}_bfloat16",
```
acaprau, line 134: doesn't seem like the code changes in this file are needed at all, we already add both kinds of precision to the `_BASE_EMBEDDING_MODELS` list, all that needs changing are the comments. pls remove the diff from the code to reduce noise and risk.

acaprau (approved):

PR #10298, opened 2026-04-17
fix(image): Cap Uploaded File Image Count

justin-tahara:
## Description

Currently hitting an issue where we are seeing files that are uploaded be much smaller than the size threshold but are causing issues due to the number of images that are in the file.

We are introducing a per file limit of images which is set to 500 and a total combined limit of 1000 images in a chat in order to prevent any over loading the chat with too many images.

## How Has This Been Tested?

Added tests and tested locally.

Image of what the error looks like:
Screenshot 2026-04-16 at 6 12 22 PM

## Additional Options
Need to add to v3.0 v3.1 v3.2

---
## Summary by cubic
Add hard caps on embedded images per PDF and per upload batch to prevent worker OOMs. Enforced during uploads (single and ZIP) and in extraction, with clearer UI copy.

- **Bug Fixes**
  - Introduced `MAX_EMBEDDED_IMAGES_PER_FILE` (default 500) and `MAX_EMBEDDED_IMAGES_PER_UPLOAD` (default 1000); read from env and clamped to >= 0.
  - Added fast preflight PDF image counting to enforce caps early: single-file uploads reject over-cap PDFs/batches; ZIP uploads skip offending PDFs with warnings; project uploads reject with clear reasons while tracking batch totals.
  - PDF extraction short-circuits when the cap is reached (defense-in-depth).
  - Updated modal copy (“PDFs with many embedded images may be rejected.”) and added unit tests for caps (including the streaming callback) and the counting helper.

<sup>Written for commit aed41cf1747fbd13f41669849adb38f0a33e2d4c. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/file_processing/extract_file_text.py
+++ b/backend/onyx/file_processing/extract_file_text.py
@@ -23,6 +23,7 @@ import openpyxl
 from openpyxl.worksheet.worksheet import Worksheet
 from PIL import Image
 
+from onyx.configs.app_configs import MAX_EMBEDDED_IMAGES_PER_FILE
 from onyx.configs.constants import ONYX_METADATA_FILENAME
 from onyx.configs.llm_configs import get_image_extraction_and_analysis_enabled
 from onyx.file_processing.file_types import OnyxFileExtensions
@@ -254,8 +255,27 @@ def read_pdf_file(
         )
 
         if extract_images:
+            image_cap = MAX_EMBEDDED_IMAGES_PER_FILE
+            images_processed = 0
+            cap_reached = False
             for page_num, page in enumerate(pdf_reader.pages):
+                if cap_reached:
+                    break
                 for image_file_object in page.images:
+                    if images_processed >= image_cap:
+                        # Defense-in-depth backstop. Upload-time validation
+                        # should have rejected files exceeding the cap, but
+                        # we also break here so a single oversized file can
+                        # never pin a worker.
+                        logger.warning(
+                            "PDF embedded image cap reached (%d). "
+                            "Skipping remaining images on page %d and beyond.",
+                            image_cap,
+                            page_num + 1,
+                        )
+                        cap_reached = True
+                        break
+
                     image = Image.open(io.BytesIO(image_file_object.data))
                     img_byte_arr = io.BytesIO()
                     image.save(img_byte_arr, format=image.format)
@@ -268,6 +288,7 @@ def read_pdf_file(
                         image_callback(img_bytes, image_name)
                     else:
                         extracted_images.append((img_bytes, image_name))
+                    images_processed += 1
 
         return text, metadata, extracted_images
 
--- a/backend/onyx/file_processing/extract_file_text.py
+++ b/backend/onyx/file_processing/extract_file_text.py
@@ -192,6 +192,56 @@ def read_text_file(
     return file_content_raw, metadata
 
 
+def count_pdf_embedded_images(file: IO[Any], cap: int) -> int:
+    """Return the number of embedded images in a PDF, short-circuiting at cap+1.
+
+    Used to reject PDFs whose image count would OOM the user-file-processing
+    worker during indexing. Returns a value > cap as a sentinel once the count
+    exceeds the cap, so callers do not iterate thousands of image objects just
+    to report a number. Returns 0 if the PDF cannot be parsed.
+
+    Owner-password-only PDFs (permission restrictions but no open password) are
+    counted normally — they decrypt with an empty string. Truly password-locked
+    PDFs are skipped (return 0) since we can't inspect them; the caller should
+    ensure the password-protected check runs first.
+
+    Always restores the file pointer to its original position before returning.
+    """
+    from pypdf import PdfReader
+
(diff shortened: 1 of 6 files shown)


backend/onyx/file_processing/extract_file_text.py
```diff
@@ -254,8 +255,27 @@ def read_pdf_file(
         )
 
         if extract_images:
+            image_cap = MAX_EMBEDDED_IMAGES_PER_FILE
+            images_processed = 0
+            cap_reached = False
```
Danelegend, line 310: Don't think cap_reached is necssary here. Could replace with images_processed >= image_cap

backend/onyx/server/features/build/api/user_library.py
```diff
@@ -128,6 +132,66 @@ class DeleteFileResponse(BaseModel):
...
+    worker during indexing. Returns ``cap + 1`` as a sentinel if the count
+    exceeds the cap, so callers do not iterate thousands of image objects just
+    to report a number. Returns 0 if the PDF cannot be parsed — the downstream
+    indexing pipeline handles malformed PDFs separately.
+    """
+    try:
+        from pypdf import PdfReader
+
+        reader = PdfReader(BytesIO(content))
+        if reader.is_encrypted:
```
Danelegend: Use password_validation::is_pdf_protected instead

backend/onyx/server/features/build/api/user_library.py
```diff
@@ -128,6 +132,66 @@ class DeleteFileResponse(BaseModel):
...
+            # Encrypted PDFs cannot be inspected without the password. Let the
+            # indexing pipeline handle them (it has decrypt-with-empty-password
+            # fallback) and rely on the extraction-time cap as the backstop.
+            return 0
+        count = 0
+        for page in reader.pages:
+            for _ in page.images:
+                count += 1
+                if count > cap:
+                    return count
```
Danelegend: This feels a bit unexpected given the function name. Probs best to just return the entire count. 

The case where we run out of compute or something seems very unlikely, so I don't think early exit is necessary. But I'll leave to your judgement

backend/onyx/server/features/build/api/user_library.py
```diff
@@ -128,6 +132,66 @@ class DeleteFileResponse(BaseModel):
...
+
+        reader = PdfReader(BytesIO(content))
+        if reader.is_encrypted:
+            # Encrypted PDFs cannot be inspected without the password. Let the
+            # indexing pipeline handle them (it has decrypt-with-empty-password
+            # fallback) and rely on the extraction-time cap as the backstop.
+            return 0
+        count = 0
+        for page in reader.pages:
+            for _ in page.images:
```
Danelegend: Can we just do count += len(page.images)?

backend/onyx/server/features/build/api/user_library.py
```diff
@@ -128,6 +132,66 @@ class DeleteFileResponse(BaseModel):
 # =============================================================================
 
 
+def _count_pdf_embedded_images(content: bytes, cap: int) -> int:
```
Danelegend: Feels somewhat troll to put this pdf specific logic here. In the utils folder I created csv_utils. Maybe this could go in a pdf_utils as well?

Nicer way to centralise file specific logic

backend/onyx/server/features/projects/projects_file_utils.py
```diff
@@ -52,6 +54,48 @@ def get_upload_size_bytes(upload: UploadFile) -> int | None:
         return None
 
 
+def count_pdf_embedded_images(upload: UploadFile, cap: int) -> int:
```
Danelegend: Repeat of _count_pdf_embedded_images?

Danelegend (approved):
Interested to know if doc_processing faces a similar problem.

I think it would be good to get the differences between docprocessing and user_file_processing jotted down so we can consolidate the pipelines.

But for a quick fix, this looks good. A few nits and improvements though

justin-tahara:
Good question on doc_processing — I traced it and the connector paths are actually covered indirectly.

Grepping for callers of `extract_text_and_images` / `read_pdf_file` outside the user-file path:

- `connectors/file/connector.py`
- `connectors/sharepoint/connector.py`
- `connectors/blob/connector.py`
- `connectors/drupal_wiki/connector.py`
- `connectors/google_drive/doc_conversion.py`

All of those eventually hit `read_pdf_file`, which now short-circuits its image loop at `MAX_EMBEDDED_IMAGES_PER_FILE`. So a malicious/pathological PDF pulled by a connector can't pin the doc_processing worker either — it's the same backstop.

What's missing on the connector side:
- No **upload-time** pre-check (connectors aren't user-initiated, so there's nothing to reject against — files just arrive). The per-upload aggregate cap is therefore user_library/projects-only.
- Admins can't see _why_ a connector-ingested PDF only got partially indexed except via worker logs.

Happy to file a follow-up to unify user-file-processing + doc_processing pipelines — your instinct is right, there's meaningful duplication. Want me to open a Linear ticket for the consolidation work?

PR #10362, opened 2026-04-20
fix(db): skip no-op stale updates in external group sync

justin-tahara:
## Description

- Adds `WHERE stale = false` to `mark_old_external_groups_as_stale` so it only touches rows that actually need updating
- Previously every sync cycle rewrote ALL rows for a cc_pair — A single tenant was updating 141K rows (832MB) every cycle even though 96% were already `stale=true`
- This was the  top contributor to `celery_worker_primary_sync` AAS in Aurora Performance Insights

## How Has This Been Tested?

Verify external group sync still marks fresh rows as stale correctly

## Additional Options

---
## Summary by cubic
Skip no-op stale updates during external group sync by only updating rows where `stale = false`. This reduces redundant writes and lowers Aurora AAS.

Adds the `stale.is_(False)` filter in `mark_old_external_groups_as_stale` for `User__ExternalUserGroupId` and `PublicExternalUserGroup`.

<sup>Written for commit c9fe5acfe6448b4a27c3eb43ec20c1be4cc0ab6a. Summary will update on new commits.</sup>

### DIFF
--- a/backend/ee/onyx/db/external_perm.py
+++ b/backend/ee/onyx/db/external_perm.py
@@ -70,11 +70,13 @@ def mark_old_external_groups_as_stale(
     db_session.execute(
         update(User__ExternalUserGroupId)
         .where(User__ExternalUserGroupId.cc_pair_id == cc_pair_id)
+        .where(User__ExternalUserGroupId.stale.is_(False))
         .values(stale=True)
     )
     db_session.execute(
         update(PublicExternalUserGroup)
         .where(PublicExternalUserGroup.cc_pair_id == cc_pair_id)
+        .where(PublicExternalUserGroup.stale.is_(False))
         .values(stale=True)
     )
 


evan-onyx (approved):
approving in case I'm missing something, but when this runs I believe the intent is truly to make everything as stale and there should be few if any already marked as stale (since they would be cleaned up from past runs if they were stale).

justin-tahara:
Good call. After investigating further, the reviewer is correct that stale rows should already be cleaned up from previous cycles by `remove_stale_external_groups`. 

The 96% stale rate on savvyadvisors indicates that `remove_stale_external_groups` (line 583 of `external_group_syncing/tasks.py`) is NOT running — likely the sync is timing out or failing before reaching cleanup. The heavy worker logs show 4 minutes of "Folder has no permissions. Skipping." DEBUG lines with no completion/cleanup message.

The `WHERE stale = false` optimization is technically correct (same end state, fewer row rewrites) but it masks this underlying issue. The real fix is investigating why cleanup isn't completing for these tenants.

Closing this PR. The root cause is tracked in the external group sync optimization plan.

PR #10364, opened 2026-04-20
fix(external-group-sync): batch upsert + commit before API calls

justin-tahara:
## Description

- External group sync was silently failing for tenants with large Google Drive workspaces, causing `user__external_user_group_id` to accumulate ~20 GB of stale rows across 8 tenants (worst: 2.2M rows / 13 GB for one tenant)
- Root cause: `mark_old_external_groups_as_stale()` left a DB transaction open during Google API folder iteration (10+ min). `idle_in_transaction_session_timeout=60s` killed the connection, causing `remove_stale_external_groups()` to never run
- Three fixes: (1) commit after mark-stale so transaction closes before API calls, (2) batch upsert with `ON CONFLICT` replacing per-row SELECT+INSERT (400K queries/cycle → ~42), (3) run stale cleanup at the START of each sync so it runs regardless of current sync success

### Bug 1: DB connection death during long API calls
`_perform_external_group_sync` opens a single DB session and holds it for the entire sync. `mark_old_external_groups_as_stale()` runs an UPDATE without committing, then the Google Drive generator starts making API calls. The DB connection sits idle-in-transaction, and after 60s Postgres kills it. The sync exceptions with `InvalidTransactionError: Can't reconnect until invalid transaction is rolled back`, and `remove_stale_external_groups()` never runs.

**Fix:** Add `db_session.commit()` in `mark_old_external_groups_as_stale` so the transaction closes before API calls begin.

### Bug 2: N+1 query pattern
`upsert_external_groups` does one SELECT per user-group mapping to check existence, then INSERT or UPDATE. For 200K mappings, that's ~400K queries per cycle.

**Fix:** Replace with batch `INSERT ... ON CONFLICT DO UPDATE` in chunks of 5000.

### Bug 3: Cleanup only runs on success
`remove_stale_external_groups` was only called after the sync loop completes successfully. If the sync fails (timeout, API error, DB error), stale rows accumulate forever.

**Fix:** Also run `remove_stale_external_groups` at the START of each sync, cleaning up stale rows from the prior cycle before marking new ones.

## How Has This Been Tested?

Verify external group sync completes for a Google Drive tenant with SYNC access
Verify stale rows are cleaned up after a successful sync cycle

## Additional Options

---
## Summary by cubic
Fixes external group sync for large Google Drive tenants by committing before API calls, batching upserts, and running cleanup at the start. Prevents stale row buildup and slashes query load during syncs.

- **Bug Fixes**
  - Commit right after marking old groups as stale to avoid idle-in-transaction timeouts during long API calls.
  - Run stale cleanup at the start of each sync so it runs even if the prior or current sync fails.

- **Refactors**
  - Replace per-row checks with batch INSERT ... ON CONFLICT DO UPDATE (5k chunks) for user-group and public-group mappings. Deduplicate both mapping types before upsert to avoid conflict errors. Reduces ~400k queries per cycle to ~42.

<sup>Written for commit ec33e6489ec728c3bf83fdd05dc2a359e0540af9. Summary will update on new commits.</sup>

### DIFF
--- a/backend/ee/onyx/db/external_perm.py
+++ b/backend/ee/onyx/db/external_perm.py
@@ -5,6 +5,7 @@ from pydantic import BaseModel
 from sqlalchemy import delete
 from sqlalchemy import select
 from sqlalchemy import update
+from sqlalchemy.dialects.postgresql import insert as pg_insert
 from sqlalchemy.orm import Session
 
 from onyx.access.utils import build_ext_group_name_for_onyx
@@ -77,6 +78,15 @@ def mark_old_external_groups_as_stale(
         .where(PublicExternalUserGroup.cc_pair_id == cc_pair_id)
         .values(stale=True)
     )
+    # Commit immediately so the transaction closes before potentially long
+    # external API calls (e.g. Google Drive folder iteration). Without this,
+    # the DB connection sits idle-in-transaction during API calls and gets
+    # killed by idle_in_transaction_session_timeout, causing the entire sync
+    # to fail and stale cleanup to never run.
+    db_session.commit()
+
+
+_UPSERT_BATCH_SIZE = 5000
 
 
 def upsert_external_groups(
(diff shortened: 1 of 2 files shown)


backend/ee/onyx/db/external_perm.py
```diff
@@ -77,6 +78,15 @@ def mark_old_external_groups_as_stale(
         .where(PublicExternalUserGroup.cc_pair_id == cc_pair_id)
         .values(stale=True)
     )
+    # Commit immediately so the transaction closes before potentially long
+    # external API calls (e.g. Google Drive folder iteration). Without this,
+    # the DB connection sits idle-in-transaction during API calls and gets
+    # killed by idle_in_transaction_session_timeout, causing the entire sync
+    # to fail and stale cleanup to never run.
+    db_session.commit()
```
evan-onyx, line 86: is there a race condition where the external group stale cleanup task can start purging these while this task is running? And I'm assuming this is the case based on the comment, but does committing reset the connection timeout?

evan-onyx (approved):
mostly lgtm

backend/ee/onyx/db/external_perm.py
justin-tahara: Only one sync runs per cc-pair so no concurrent purging is possible. 

And yes the commit will only apply to the connections in transaction so they will be reset

PR #10366, opened 2026-04-20
fix: remove minio profile restriction to fix default docker deployment

### LINKED ISSUE #9793: docker compose up -d fails: minio not started but FILE_STORE_BACKEND defaults to s3
## Description

Following the official local Docker deployment docs, running `docker compose up -d` causes the API server to crash immediately with:

```
botocore.exceptions.EndpointConnectionError: Could not connect to the endpoint URL: "http://minio:9000/onyx-file-store-bucket"
ERROR: Application startup failed. Exiting.
```

## Root Cause

`docker-compose.yml` has a contradiction:
- `api_server` defaults to `FILE_STORE_BACKEND=s3` (line 79), which requires minio
- The `minio` service is gated behind `profiles: ["s3-filestore"]` (line 523), so it is **never started** by the default `docker compose up -d` command

Even though minio is listed as `required: false` in the api_server depends_on, the API server still crashes when it cannot connect to the S3 endpoint at startup.

## Steps to Reproduce

```bash
git clone --depth 1 https://github.com/onyx-dot-app/onyx.git
cd onyx/deployment/docker_compose
docker compose up -d
docker logs onyx-api_server-1
```

## Expected Behavior

The default `docker compose up -d` command should result in a working deployment as documented.

## Workaround

```bash
docker compose --profile s3-filestore up -d
```

## Suggested Fix

Either:
1. Remove `profiles: ["s3-filestore"]` from the minio service so it starts by default, or
2. Change the default `FILE_STORE_BACKEND` to a local/non-S3 option when minio is not running

## Environment

- OS: Ubuntu 24.04.4 (VMware)
- Docker version: latest

Khanz9664:
## Description

Following the official deployment documentation, running `docker compose up -d` currently causes the API server to crash immediately with a `botocore.exceptions.EndpointConnectionError`. This happens because the `api_server` defaults to `FILE_STORE_BACKEND=s3` (which requires Minio), but the `minio` service was gated behind `profiles: ["s3-filestore"]` and never started.

This PR removes the `profiles: ["s3-filestore"]` line from the `minio` service in `deployment/docker_compose/docker-compose.yml`. Minio will now start by default alongside the rest of the stack, satisfying the S3 backend requirement and resolving the API server crash.

Closes #9793

## How Has This Been Tested?

1. Cloned the repository and navigated to `deployment/docker_compose/`
2. Ran `docker compose up -d`
3. Verified the `minio` container is successfully pulled and running via `docker compose ps`
4. Checked the API server logs (`docker logs onyx-api_server-1`) and confirmed the `EndpointConnectionError` for the S3 bucket is no longer present and the server continues its startup sequence.

## Additional Options

### DIFF
--- a/deployment/docker_compose/docker-compose.yml
+++ b/deployment/docker_compose/docker-compose.yml
@@ -520,7 +520,6 @@ services:
 
   minio:
     image: minio/minio:RELEASE.2025-07-23T15-54-02Z-cpuv1
-    profiles: ["s3-filestore"]
     restart: unless-stopped
     # DEV: To expose ports, either:
     # 1. Use docker-compose.dev.yml: docker compose -f docker-compose.yml -f docker-compose.dev.yml up -d --wait
--- a/deployment/docker_compose/docker-compose.yml
+++ b/deployment/docker_compose/docker-compose.yml
@@ -156,6 +156,9 @@ services:
         condition: service_started
       indexing_model_server:
         condition: service_started
+      minio:
+        condition: service_started
+        required: false
     restart: unless-stopped
     environment:
       - FILE_STORE_BACKEND=${FILE_STORE_BACKEND:-s3}


Khanz9664:
Closing in favor of a new PR from a feature branch: https://github.com/onyx-dot-app/onyx/pull/10429

PR #10369, opened 2026-04-20
fix(gong): move call-details retry into checkpoint state instead of blocking

nmgarza5:
## Description

Addresses the Greptile P2 raised on #10359 (and present on `main`): `_fetch_call_details_with_retry` could block a single `load_from_checkpoint` call for up to ~930s (30 + 60 + 120 + 240 + 480) while waiting out the Gong propagation race. That fights the CheckpointedConnector contract — the checkpoint can't be persisted while we sleep, and long blocks risk worker timeouts.

This PR moves the retry into checkpoint state:

- `GongConnectorCheckpoint` gains `pending_transcripts: dict[str, dict]` and `pending_call_details_attempts: int`. Transcripts whose `/v2/calls/extensive` details are missing get stashed under their `call_id`.
- On each subsequent `load_from_checkpoint` invocation, we retry only the missing IDs *before* fetching a new page. Worker cadence between invocations provides the effective spacing — no in-call `time.sleep()` beyond the existing sub-second per-request throttle.
- After `MAX_CALL_DETAILS_ATTEMPTS` (6) invocations still missing details, we emit `ConnectorFailure` for the unresolved call_ids and clear pending state so the run progresses.
- `_process_transcripts` now splits into a pure `_build_document` helper plus a page-processing generator that mutates the checkpoint to stash deferred work.

Behavior preserved:
- Gong-race tolerance: still up to 6 retry attempts.
- Cursor-expiry handling unchanged.
- Workspace iteration unchanged.

## Additional Options

### DIFF
--- a/backend/onyx/connectors/gong/connector.py
+++ b/backend/onyx/connectors/gong/connector.py
@@ -42,11 +42,20 @@ class GongConnectorCheckpoint(ConnectorCheckpoint):
     time_range: tuple[str, str] | None = None
     # Transcripts whose call details were not yet available from /v2/calls/extensive
     # (Gong has a known race where transcript call IDs take time to propagate).
-    # Keyed by call_id. Retried on subsequent checkpoint invocations — the worker
-    # cadence between calls provides natural backoff without blocking this call.
+    # Keyed by call_id. Retried on subsequent checkpoint invocations.
+    #
+    # Invariant: all entries share one resolution session — they're stashed
+    # together from a single page and share the attempt counter and retry
+    # deadline. load_from_checkpoint only fetches a new page when this dict
+    # is empty, so entries from different pages can't mix.
     pending_transcripts: dict[str, dict[str, Any]] = {}
     # Number of resolution attempts made for pending_transcripts so far.
     pending_call_details_attempts: int = 0
+    # Unix timestamp before which we should not retry pending_transcripts.
+    # Enforces exponential backoff independent of worker cadence — Gong's
+    # transcript-ID propagation race can take tens of seconds to minutes,
+    # longer than typical worker reinvocation intervals.
+    pending_retry_after: float | None = None
 
 
 class _TranscriptPage(BaseModel):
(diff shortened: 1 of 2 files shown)


backend/onyx/connectors/gong/connector.py
```diff
@@ -40,6 +40,13 @@ class GongConnectorCheckpoint(ConnectorCheckpoint):
     cursor: str | None = None
     # Cached time range — computed once, reused across checkpoint calls
     time_range: tuple[str, str] | None = None
+    # Transcripts whose call details were not yet available from /v2/calls/extensive
+    # (Gong has a known race where transcript call IDs take time to propagate).
+    # Keyed by call_id. Retried on subsequent checkpoint invocations — the worker
+    # cadence between calls provides natural backoff without blocking this call.
+    pending_transcripts: dict[str, dict[str, Any]] = {}
+    # Number of resolution attempts made for pending_transcripts so far.
+    pending_call_details_attempts: int = 0
```
evan-onyx, line 53: can't this be different for different entries of the pending dict? i.e. new transcript gets added on a future run. Not a huge deal but we probably want to document if so?

backend/onyx/connectors/gong/connector.py
```diff
@@ -473,23 +481,80 @@ def load_from_checkpoint(
...
+                    ),
+                )
+            checkpoint.pending_transcripts = {}
+            checkpoint.pending_call_details_attempts = 0
+            # has_more is recomputed by the workspace iteration that follows;
+            # reset to False here so a stale True from a prior invocation
+            # can't leak out via any future early-return path.
+            checkpoint.has_more = False
+        else:
+            # Retry again on a future invocation; worker cadence provides gap.
```
evan-onyx: most of the time the next batch is happening almost immediately after the previous one; I don't have the full context here but if this retrying is due to i.e. rate limits then I don't think this assumption is sufficient (i.e. we'll end up just quickly retrying 6 times and emitting connectorfailures)

evan-onyx (approved):
some thoughts but u got this

backend/onyx/connectors/gong/connector.py
nmgarza5: hmmm fair point. let me think about this one a bit more

PR #10373, opened 2026-04-20
feat(invites): add tenant_invite_event table to cap trial tenant lifetime invites

nmgarza5:
## Description

Closes the trial invite-cap bypass where an admin could loop `invite 10 → remove 10 → invite 10` against the KV-backed `get_invited_users()` counter and escape the `NUM_FREE_TRIAL_USER_INVITES` lifetime cap.

**Design:** new `public.tenant_invite_counter(tenant_id PK, total_invites_sent INT, updated_at)` — one row per tenant. The counter is **monotonically incremented only**. `remove_invited_user` never touches it; the UI cleanup path still pops the KV list, but the cap enforcement state persists.

**Enforcement:** single atomic `INSERT ... ON CONFLICT DO UPDATE ... RETURNING total_invites_sent`. The UPSERT takes a row-level lock on `tenant_id` during UPDATE, so concurrent bulk-invite flows for the same tenant are serialized without needing an advisory lock or `FOR UPDATE` pre-check. Cap exceed → `ROLLBACK` so the reservation does not stick; reserved slots are discarded. Cost is O(1) PK lookup inside the UPSERT (no `COUNT(*)` scan) and ~1 row per tenant of storage (vs 1 row per invite).

**Scope:** trial tenants only. Paid tenants bypass via the existing `is_tenant_on_trial_fn(tenant_id)` gate.

Attack walkthrough (cap = 10):

| Step | Counter | Decision |
| --- | --- | --- |
| Invite 10 | 0 → 10 | sent |
| Remove 10 | 10 (unchanged) | KV cleared |
| Invite 10 more | 10 → 20 | 403 |
| Invite 1 | 10 → 11 | 403 |

## Additional Options

### DIFF
--- a/backend/alembic_tenants/versions/d4e7a92c1b38_add_tenant_invite_counter.py
+++ b/backend/alembic_tenants/versions/d4e7a92c1b38_add_tenant_invite_counter.py
@@ -1,0 +1,79 @@
+"""add_tenant_invite_counter_table
+
+Revision ID: d4e7a92c1b38
+Revises: 3b9f09038764
+Create Date: 2026-04-20 18:00:00.000000
+
+Adds `public.tenant_invite_counter`, the lifetime invite-quota counter used by
+the trial-tenant cap in `bulk_invite_users`. One row per tenant; holds a
+monotonically-incremented total of invites ever reserved by that tenant.
+
+Why we need it:
+    Trial tenants are capped at NUM_FREE_TRIAL_USER_INVITES per lifetime.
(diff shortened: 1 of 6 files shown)


evan-onyx (approved):
thoughts on having this be a count of emails sent per tenant rather than a row per email sent?

nmgarza5:
> thoughts on having this be a count of emails sent per tenant rather than a row per email sent?

yeah this makes sense. i figured having the audit trail would be nice but probably overkill. the counter works just fine

PR #10389, opened 2026-04-20
fix(desktop): check if backend server is available on start

jmelahman:
## Description

First checks if the configured `server_url` is reachable and if not re-directs the user to the configuration page rather than just returning `Could not connect to localhost: Connection refused` page (and from which the configuration page to update the `server_url` is inaccessible). 

I hit this in two ways:

1. when I am offline and want to connect to a local Onyx backend, but I was previously connecting to a cloud backend
2. when I am connected to a local Onyx backend I no longer have running and want to connect to a Cloud backend

## How Has This Been Tested?

20260420_13h26m24s_grim

## Additional Options

---
## Summary by cubic
Check the configured backend `server_url` on app start and route to Settings with a clear error if unreachable, instead of a native “connection refused” page. Hardened the probe so it only blocks on connect/timeout and uses the URL from config.

- **Bug Fixes**
  - Added Tauri command `check_server_reachable` using `reqwest` (5s timeout, HTTP/HTTPS only, HEAD) that reads `server_url` from `ConfigState` to prevent arbitrary native probes.
  - `index.html` checks reachability before redirect; on failure, opens Settings and shows a clear error. Only connect/timeout errors block; TLS/redirect errors pass so self‑signed backends still load.

- **Dependencies**
  - Added `reqwest` with `rustls`; updated `Cargo.lock`.

<sup>Written for commit c3ba232bcf92a6c52738cad8d2bf20ae2cafd61a. Summary will update on new commits.</sup>

### DIFF
--- a/desktop/src-tauri/Cargo.lock
+++ b/desktop/src-tauri/Cargo.lock
@@ -83,6 +83,28 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
+name = "aws-lc-rs"
+version = "1.16.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a054912289d18629dc78375ba2c3726a3afe3ff71b4edba9dedfca0e3446d1fc"
+dependencies = [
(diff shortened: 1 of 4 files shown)


jmelahman:
@greptile please review again

justin-tahara (approved):

PR #10395, opened 2026-04-20
fix(pdf): Allowing Image Extraction from PDF files

justin-tahara:
## Description

Historically if we had a PDF with images that was a user file upload, it would not parse the image and it would not extract any information from the file. This now has changed and we will try to get semantic meaning from the images if there is a vision model that is available. 

## How Has This Been Tested?

Tested locally. Validated that the toggle exists, is set by default and the fake text was extracted from the file. 

Screenshot 2026-04-20 at 2 12 22 PM

## Additional Options

---
## Summary by cubic
Enables image extraction and vision-LLM captioning for PDFs and DOCX so image-only uploads become searchable and can answer questions. Defaults on with an admin toggle; chat reads cached captions from the index and avoids re-running vision models.

- **New Features**
  - Indexes embedded images and stores captions; chat pulls cached captions from the document index and merges them with any file text.
  - Accepts scanned/image-only PDFs and DOCX for indexing when enabled, even if text extraction is empty.
  - Enforces embedded-image caps for DOCX like PDF and improves rejection messages.
  - Loads the document index/vision LLM only when needed; adds “Image Extraction & Analysis” toggle in Chat Preferences (default on).

<sup>Written for commit 495669b3d250f662e0098f8d07f0ab22145cdca5. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/chat/process_message.py
+++ b/backend/onyx/chat/process_message.py
@@ -60,6 +60,7 @@ from onyx.configs.constants import DEFAULT_PERSONA_ID
 from onyx.configs.constants import DocumentSource
 from onyx.configs.constants import MessageType
 from onyx.configs.constants import MilestoneRecordType
+from onyx.configs.llm_configs import get_image_extraction_and_analysis_enabled
 from onyx.context.search.models import BaseFilters
 from onyx.context.search.models import SearchDoc
 from onyx.db.chat import create_new_chat_message
@@ -80,6 +81,10 @@ from onyx.error_handling.error_codes import OnyxErrorCode
 from onyx.error_handling.exceptions import log_onyx_error
 from onyx.error_handling.exceptions import OnyxError
 from onyx.file_processing.extract_file_text import extract_file_text
+from onyx.file_processing.extract_file_text import extract_text_and_images
+from onyx.file_processing.image_summarization import (
+    summarize_image_with_error_handling,
+)
 from onyx.file_store.models import ChatFileType
 from onyx.file_store.models import InMemoryChatFile
 from onyx.file_store.utils import load_in_memory_chat_files
@@ -89,6 +94,7 @@ from onyx.hooks.executor import HookSkipped
 from onyx.hooks.executor import HookSoftFailed
 from onyx.hooks.points.query_processing import QueryProcessingPayload
 from onyx.hooks.points.query_processing import QueryProcessingResponse
+from onyx.llm.factory import get_default_llm_with_vision
 from onyx.llm.factory import get_llm_for_persona
 from onyx.llm.factory import get_llm_token_counter
 from onyx.llm.interfaces import LLM
@@ -127,6 +133,11 @@ from shared_configs.contextvars import get_current_tenant_id
 logger = setup_logger()
 ERROR_TYPE_CANCELLED = "cancelled"
 APPROX_CHARS_PER_TOKEN = 4
+# Cap for inline vision-LLM summarization of image-only files at chat time.
+# Each summary is a serial vision-LLM call (~seconds), so we bound latency for
+# files with many embedded images. Upload-time extraction/indexing summarizes
+# every embedded image (up to MAX_EMBEDDED_IMAGES_PER_FILE).
+_CHAT_TIME_IMAGE_SUMMARIZATION_CAP = 20
 
 
 def _collect_available_file_ids(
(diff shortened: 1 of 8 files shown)


backend/onyx/chat/process_message.py
```diff
@@ -244,22 +260,123 @@ def _empty_extracted_context_files() -> ExtractedContextFiles:
...
+    those summaries are stored as chunks whose `image_file_id` is set. Reading
+    them back at chat time avoids re-running vision-LLM calls per turn.
+    Returns an empty list if the index has no chunks yet (e.g. indexing is
+    still in flight) or on any fetch failure.
+    """
+    if user_file is None or document_index is None:
+        return []
+    try:
+        chunks = document_index.id_based_retrieval(
+            chunk_requests=[VespaChunkRequest(document_id=str(user_file.id))],
```
Danelegend, line 271: This hasn't been ported over?

backend/onyx/indexing/indexing_pipeline.py
```diff
@@ -531,7 +531,12 @@ def process_image_sections(documents: list[Document]) -> list[IndexingDocument]:
         List of IndexingDocument objects with processed_sections as list[Section]
     """
     # Check if image extraction and analysis is enabled before trying to get a vision LLM
-    if not get_image_extraction_and_analysis_enabled():
+    has_image_section = any(
+        isinstance(section, ImageSection)
```
Danelegend: Use section.type -> I believe in the doc processing task everything is a Section type, so this would fail

backend/onyx/chat/process_message.py
```diff
@@ -244,22 +260,123 @@ def _empty_extracted_context_files() -> ExtractedContextFiles:
...
+
+        parts: list[str] = []
+        if has_text:
+            parts.append(text)
+        parts.extend(cached_captions)
+
+        if parts:
+            return "\n\n".join(parts).strip() or None
+
+        # Nothing cached and no text — file is likely mid-indexing. Fall
```
Danelegend: I thought a document's chunks are uploaded together, so we shouldn't end up in this case? Could be wrong though.

backend/onyx/chat/process_message.py
```diff
@@ -244,22 +260,123 @@ def _empty_extracted_context_files() -> ExtractedContextFiles:
...
+        # back to inline vision summarization for a bounded number of
+        # images so the chat turn still has something meaningful.
+        images = extraction.embedded_images[:_CHAT_TIME_IMAGE_SUMMARIZATION_CAP]
+        truncated = len(extraction.embedded_images) > len(images)
+        vision_llm = get_default_llm_with_vision()
```
Danelegend: I have some latency concerns here tbh. In my mind, this path should be quite hot and we're better off dropping some image summarizations. Thoughts?

web/src/refresh-pages/admin/ChatPreferencesPage.tsx
```diff
@@ -996,6 +996,23 @@ export default function ChatPreferencesPage() {
                   </InputVertical>
                 </Card>
 
+                <Card border="solid" rounding="lg">
```
Danelegend, line 999: Screenshot pls :)

backend/onyx/server/settings/models.py
```diff
@@ -80,7 +80,7 @@ class Settings(BaseModel):
     query_history_type: QueryHistoryType | None = None
 
     # Image processing settings
-    image_extraction_and_analysis_enabled: bool | None = False
+    image_extraction_and_analysis_enabled: bool | None = True
```
Danelegend, line 83: Default on now?

backend/onyx/server/features/projects/projects_file_utils.py
```diff
@@ -308,6 +313,20 @@ def categorize_uploaded_files(
                     extension=extension,
                 )
                 if not text_content:
+                    # PDFs with embedded images (e.g. scans) have no extractable
+                    # text but can still be indexed via the vision-LLM
+                    # captioning path when image analysis is enabled.
+                    if extension == ".pdf" and count > 0 and image_extraction_enabled:
```
Danelegend: Can this PR handle other document extensions as well or would that require more work?

backend/onyx/chat/process_message.py
```diff
@@ -244,22 +260,123 @@ def _empty_extracted_context_files() -> ExtractedContextFiles:
...
+            filters=IndexFilters(access_control_list=None),
+        )
+    except Exception:
+        logger.warning(
+            f"Failed to fetch cached captions for user_file {user_file.id}",
+            exc_info=True,
+        )
+        return []
+    return [
+        f"[Image — {chunk.image_file_id}]\n{chunk.content}"
```
Danelegend: Although unlikely, might be worth combining chunks that have the same image_file_id

Danelegend (approved):

web/src/refresh-pages/admin/ChatPreferencesPage.tsx
Danelegend: I'm blind

PR #10399, opened 2026-04-20
feat(auth): add `USER_DIRECTORY_ADMIN_ONLY` flag to gate `/users`

jmelahman:
## Description

Adds a flag to disallow non-admin from calling `/users`. With this set, non-admin become unable to share agents with individual users without the assistance of an administrator.

## How Has This Been Tested?

20260420_16h54m39s_grim

## Additional Options

---
## Summary by cubic
Adds a `USER_DIRECTORY_ADMIN_ONLY` flag to restrict the `/users` directory to admins, and updates the Share Agent modal to gracefully handle restricted directories. The flag is off by default to preserve current sharing behavior.

- **New Features**
  - Backend: when `USER_DIRECTORY_ADMIN_ONLY` is true, `GET /users` requires `READ_USERS`; otherwise returns 403.
  - Frontend: Share Agent modal detects 403 on users fetch, disables user autocomplete with a tooltip; groups remain available.
  - Config: added `USER_DIRECTORY_ADMIN_ONLY` to `docker-compose.multitenant-dev.yml`, `env.template`, and `env.prod.template`.
  - Tests: added unit tests for restricted and unrestricted directory access and bot-account filtering.

- **Migration**
  - To enable restriction, set `USER_DIRECTORY_ADMIN_ONLY=true` and restart the service.
  - Confirm admins have `READ_USERS` so they can list users; inform non-admins they’ll need an admin to share with individual users.

<sup>Written for commit cc9f849cbd05fc39ce1b17f45c2ab324e0e5a381. Summary will update on new commits.</sup>


web/src/sections/modals/ShareAgentModal.tsx
```diff
@@ -214,14 +225,31 @@ function ShareAgentFormContent({ agentId }: ShareAgentFormContentProps) {
...
               <Section gap={0.5} alignItems="start">
-                <InputComboBox
-                  placeholder="Add users and groups"
-                  value=""
-                  onChange={() => {}}
-                  onValueChange={handleComboBoxSelect}
-                  options={comboBoxOptions}
-                  strict
-                />
+                <Disabled
```
Subash-Mohan, line 228: Are we still using the disabled component? I thought we were getting rid of it.

deployment/docker_compose/docker-compose.multitenant-dev.yml
```diff
@@ -40,6 +40,7 @@ services:
       - SMTP_USER=${SMTP_USER:-}
       - SMTP_PASS=${SMTP_PASS:-}
       - ENABLE_EMAIL_INVITES=${ENABLE_EMAIL_INVITES:-}
+      - USER_DIRECTORY_ADMIN_ONLY=${USER_DIRECTORY_ADMIN_ONLY:-}
```
Subash-Mohan, line 43: Just to confirm, does it disable the `/user` directory across all tenants in the cloud?

Subash-Mohan (approved):
LGTM!

deployment/docker_compose/docker-compose.multitenant-dev.yml
jmelahman: This flag would affect all tenants, but is opt-in and not on by default

PR #10400, opened 2026-04-21
fix(llm): Model reasoning returning None

Danelegend:
## Description
There is possibly an error where supports_reasoning does not exist and is returning None. This explicitly handles that case where supports_reasoning is None.

## Additional Options

---
## Summary by cubic
Fixes `model_is_reasoning_model` to treat `supports_reasoning: None` as `False` and log the model and provider. Fallback to `litellm.supports_reasoning()` stays the same.

- **Bug Fixes**
  - Log an error and return `False` when `supports_reasoning` is `None`.

<sup>Written for commit 551a3ed42aaeab1cefa3727653271baae275d74c. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/llm/utils.py
+++ b/backend/onyx/llm/utils.py
@@ -743,7 +743,12 @@ def model_is_reasoning_model(model_name: str, model_provider: str) -> bool:
             model_name,
         )
         if model_obj and "supports_reasoning" in model_obj:
-            return model_obj.get("supports_reasoning", False)
+            reasoning = model_obj.get("supports_reasoning")
+            if reasoning is None:
+                logger.error(
+                    f"Cannot find reasoning for name={model_name} and provider={model_provider}"
+                )
+            return reasoning or False
 
         # Fallback: try using litellm.supports_reasoning() for newer models
         try:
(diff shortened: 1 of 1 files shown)


backend/onyx/llm/utils.py
```diff
@@ -743,7 +743,12 @@ def model_is_reasoning_model(model_name: str, model_provider: str) -> bool:
             model_name,
         )
         if model_obj and "supports_reasoning" in model_obj:
-            return model_obj["supports_reasoning"]
+            reasoning = model_obj["supports_reasoning"]
+            if reasoning is None:
+                logger.error(
+                    f"Cannot find reasoning for name={model_name} and provider={model_provider}"
+                )
+            return reasoning or False
```
jmelahman: Should we either coerce this too a bool or update the def to return `Any`? I'm surprised the type-checker is not complaining since `reasoning` can be `Any` or am I missing something?

jmelahman: Also, sorry was in a bit of a rush getting off the train. The `return reasoning or False` stood out to me because if we know `reasoning` is a bool we should just `return reasoning`. Also given the error we're seeing should be caught by the type-checker, I'm a bit curious what is going on around there.

Danelegend: Classic litellm travesity :(((

jmelahman (approved):

PR #10381, opened 2026-04-20
feat(auth): reCAPTCHA cookie flow gating Google OAuth callback

nmgarza5:
## Description

L2-D from post-mortem [Invite-Spam Abuse & Control-Plane Crash — 2026-04-17](https://linear.app/onyx-app/document/post-mortem-invite-spam-abuse-and-control-plane-crash-2026-04-17-db5442c84f56).

### Defect this closes

Google OAuth signups are not gated by reCAPTCHA today. Concretely: `UserManager.create()` at `backend/onyx/auth/users.py:388` runs `verify_captcha_token(...)` when `is_captcha_enabled()` is true — that covers `POST /auth/register` (email/password). `UserManager.oauth_callback()` at `backend/onyx/auth/users.py:754` calls `self.user_db.create(user_dict)` directly, skipping `UserManager.create()` and therefore skipping the captcha check. Any account created via the Google OAuth callback bypasses reCAPTCHA regardless of enforcement config.

This PR adds a pre-redirect challenge on the Google OAuth button so the OAuth path is gated too.

### Flow

```
signup page ──> grecaptcha.execute("oauth") ──> POST /auth/captcha/oauth-verify
                                                          │
                                        Google siteverify
                                                          │
                                       ok → signed HttpOnly cookie set
                                                          │
                                  Continue-with-Google redirect fires
                                                          │
                                      /auth/oauth/callback
                                                          │
                               CaptchaCookieMiddleware validates cookie
                               valid → fastapi-users flow    invalid → 403
```

The cookie mechanism is the trick: Google's OAuth callback is a browser redirect we cannot attach a header to, but browsers send cookies automatically on same-origin requests. Verify once on click, cookie rides along on the callback.

Email/password signup is unchanged — its existing `captcha_token` body-field check inside `UserManager.create()` already covers that path.

### Enforcement switch

Fully gated on `is_captcha_enabled()` (driven by `NEXT_PUBLIC_RECAPTCHA_SITE_KEY` + `RECAPTCHA_SECRET_KEY`). Unset on self-hosted → the endpoint and middleware are no-ops.

### Files

**Backend**
- `onyx/auth/captcha.py` — adds HMAC-signed cookie issue/validate helpers alongside the existing `verify_captcha_token`
- `onyx/server/auth/captcha_api.py` — `POST /auth/captcha/oauth-verify` endpoint + `CaptchaCookieMiddleware` gating `/auth/oauth/callback`
- `onyx/main.py` — register router + middleware
[...]

PR #10401, opened 2026-04-21
feat(auth): harden /auth/register against captcha-score bots

nmgarza5:
## Description

Follow-up to PR #10381 (reCAPTCHA on OAuth callback). The post-mortem investigation showed the 2026-04-17 bot tenants all used email/password signup with disposable domains (zero Google-OAuth emails in the 1,677-tenant dataset). On cloud, `CAPTCHA_ENABLED=true` and `RECAPTCHA_SCORE_THRESHOLD=0.8` were active during the attack — meaning the bots passed Google v3 at score ≥ 0.8. v3 alone, even at 0.8, is not enough against sophisticated bots with residential proxies + real-browser automation.

This PR stacks four orthogonal defenses on top of v3:

### 1. Token replay cache (`backend/onyx/auth/captcha.py`)

Before calling Google siteverify, `verify_captcha_token` SETNX-claims `captcha:replay:<sha256(token)>` in Redis with a 120s TTL (Google's own token expiry). A second attempt to redeem the same token fails fast:

```
Captcha verification failed: token already used
```

Closes the "solve once, spray N registrations" attack where a single human-solved token is re-submitted across many registration attempts. Fails open on Redis errors — a Redis blip cannot hard-fail legitimate signups.

### 2. Per-IP signup rate limit (`backend/onyx/auth/signup_rate_limit.py`, new)

Hour-bucketed Redis `INCR` on `signup_rate:<ip>:<hour>`. Hard cap **5 registrations per IP per hour**. Only active on `MULTI_TENANT` deployments — self-hosted keeps today's behavior. Over-limit requests get `OnyxError(RATE_LIMITED)` → HTTP 429.

Closes the vector where a bot farm with real v3 scores tries to register hundreds of accounts from the same IP (or narrow IP pool). Raises attacker cost by forcing IP rotation.

Uses `X-Forwarded-For` for the real client IP (AWS LB / nginx pass-through); falls back to `request.client.host`.

### 3. Require v3 score field (`captcha.py`)

Google's public test secret returns `{success:true}` with **no score field**. Old code:

```python
if result.score is not None:      # skipped entirely when score missing
    if result.score < RECAPTCHA_SCORE_THRESHOLD:
        raise ...
```

That's an accidental bypass if a test secret ever leaks into prod config. New code rejects outright when the score is absent:

```python
if result.score is None:
    raise CaptchaVerificationError("Captcha verification failed: missing score")
```

### 4. Reorder `UserManager.create` — cheap checks first

Before: `captcha → validate_password → verify_email_domain`. Bots on disposable domains burn a Google API call before being rejected.

After: `verify_email_domain → enforce_signup_rate_limit → captcha → validate_password`. Disposable domains and rate-limited IPs get a fast 400/429 with no Google round-trip. Reduces siteverify load and tightens the failure fast-path.

### Item 5 deferred

The original plan included "v3 score between threshold and 1.0 → step up to v2 visible checkbox." That's larger — needs:

- Frontend: new `Recaptchav2Challenge` component + signup-page wiring
- Backend: accept v2 token path on `/auth/register` and call siteverify without action/score constraints
- UX decision: when exactly to show the checkbox, how to phrase it

Scoped for a follow-up PR so it can be reviewed on its own UX merits.

### No new env vars

Everything beyond the existing `CAPTCHA_ENABLED` / `RECAPTCHA_*` is hardcoded inside the modules:

- Replay TTL: `_REPLAY_CACHE_TTL_SECONDS = 120` (matches Google)
- Signup cap: `_PER_IP_PER_HOUR = 5`
- Bucket size: `_BUCKET_SECONDS = 3600`

Signup rate limit is gated on `MULTI_TENANT` so self-hosted is untouched.

### 31 unit tests pass

**15 new tests:**

```bash
$ pytest -xv backend/tests/unit/onyx/auth/test_captcha_replay.py \
              backend/tests/unit/onyx/auth/test_signup_rate_limit.py
# ============================== 15 passed in 0.06s ==============================
```

Covers:
- Replay cache: first-use success, second-use rejection, fail-open on Redis error, SHA-256-prefixed key (raw token never stored)
- Require-score: missing-score rejection, passing with score present
- Signup rate-limit IP extraction: XFF first-entry, client.host fallback, `"unknown"` when no client
- Rate-limit behavior: disabled when not multi-tenant, under-cap allowed, over-cap raises 429, TTL only on first hit, fail-open, cross-IP isolation

**16 existing tests still pass** after the `UserManager.create` reorder:

```bash
$ pytest -xv backend/tests/unit/onyx/auth/test_user_registration.py
# ============================== 16 passed in 0.29s ==============================
```

### Type checks

```bash
$ mypy onyx/auth/captcha.py onyx/auth/signup_rate_limit.py \
        tests/unit/onyx/auth/test_captcha_replay.py \
        tests/unit/onyx/auth/test_signup_rate_limit.py
# Success: no issues found in 4 source files
```

`onyx/auth/users.py` has 7 pre-existing mypy errors unrelated to this PR (verified by stashing my changes and re-running — same 7 errors on clean main).

### Manual smoke — replay cache

With a real local reCAPTCHA token reused twice via curl to `/auth/captcha/oauth-verify`: first call returns 200, second returns 403 with detail `"Captcha verification failed: token already used"`. Same mechanism applies to `/auth/register` since both paths go through `verify_captcha_token`.

### Manual smoke — rate limit

Simulated 6 POSTs to `/auth/register` within the hour bucket from the same X-Forwarded-For. Hit 6 returned 429 with detail `"Too many signup attempts from this network..."`.

## Additional Options

### DIFF
--- a/backend/onyx/auth/captcha.py
+++ b/backend/onyx/auth/captcha.py
@@ -1,5 +1,7 @@
 """Captcha verification for user registration."""
 
+import hashlib
+
 import httpx
 from pydantic import BaseModel
 from pydantic import Field
@@ -7,11 +9,17 @@ from pydantic import Field
 from onyx.configs.app_configs import CAPTCHA_ENABLED
 from onyx.configs.app_configs import RECAPTCHA_SCORE_THRESHOLD
 from onyx.configs.app_configs import RECAPTCHA_SECRET_KEY
+from onyx.redis.redis_pool import get_async_redis_connection
 from onyx.utils.logger import setup_logger
 
 logger = setup_logger()
 
 RECAPTCHA_VERIFY_URL = "https://www.google.com/recaptcha/api/siteverify"
+
+# Google v3 tokens expire server-side at ~2 minutes, so 120s is the max useful
+# replay window — after that Google would reject the token anyway.
+_REPLAY_CACHE_TTL_SECONDS = 120
+_REPLAY_KEY_PREFIX = "captcha:replay:"
 
 
 class CaptchaVerificationError(Exception):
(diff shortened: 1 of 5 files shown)


nmgarza5:
Splitting into separate PRs per item for independent review.

PR #10402, opened 2026-04-21
feat(auth): Redis replay cache on captcha tokens

nmgarza5:
## Description

Closes the "solve one reCAPTCHA token, spray N registrations" pattern. Before calling Google siteverify, `verify_captcha_token` SETNX-claims a SHA-256 fingerprint of the token in Redis with a 120s TTL (Google's own token lifetime). A second redeem of the same token fails fast with `Captcha verification failed: token already used`.

Raw token is never stored — only the hash. Fails open on Redis errors so a Redis blip cannot hard-fail legitimate signups.

TTL is a hardcoded module constant (`_REPLAY_CACHE_TTL_SECONDS = 120`). No new env vars.

Applies automatically to both existing callers: email/password `/auth/register` via `UserManager.create`, and `/auth/captcha/oauth-verify` from #10381.

## How Has This Been Tested?

4 unit tests — first-use success, replay rejection, fail-open on Redis error, SHA-256 key shape (raw token never leaked).

```bash
$ pytest -xv backend/tests/unit/onyx/auth/test_captcha_replay.py
# 4 passed
$ mypy onyx/auth/captcha.py tests/unit/onyx/auth/test_captcha_replay.py
# Success
```

## Additional Options

### DIFF
--- a/backend/onyx/auth/captcha.py
+++ b/backend/onyx/auth/captcha.py
@@ -27,12 +27,18 @@ from onyx.configs.app_configs import CAPTCHA_ENABLED
 from onyx.configs.app_configs import RECAPTCHA_SCORE_THRESHOLD
 from onyx.configs.app_configs import RECAPTCHA_SECRET_KEY
 from onyx.configs.app_configs import USER_AUTH_SECRET
+from onyx.redis.redis_pool import get_async_redis_connection
 from onyx.utils.logger import setup_logger
 
 logger = setup_logger()
 
 RECAPTCHA_VERIFY_URL = "https://www.google.com/recaptcha/api/siteverify"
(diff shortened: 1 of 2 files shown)


backend/onyx/auth/captcha.py
jessicasingh7: we might want to explicitly raise if no score:
```
if result.score is None: 
    raise CaptchaVerificationError("...missing score")
```

jessicasingh7 (approved):

backend/onyx/auth/captcha.py
nmgarza5: it was already done here: https://github.com/onyx-dot-app/onyx/pull/10403

PR #10413, opened 2026-04-21
fix(security): enforce chat session ownership on stop endpoint

jmelahman:
## Description
- The `POST /chat/stop-chat-session/{chat_session_id}` handler now resolves the target session through the same ownership-scoped lookup (`get_chat_session_by_id` filtered by `user.id`) used by the other session-scoped endpoints before setting the stop signal.
- Callers who do not own (or admin-access) the session receive a 404 and cannot affect its state.

## How Has This Been Tested?
- [x] New integration regression test (`test_stop_chat_session_rejects_non_owner`) verifies owner can stop, non-owner gets 404, unknown IDs get 404.
- [x] `pre-commit` clean on the changed files (black, ty, ruff, autoflake, reorder-python-imports).
- [x] Security review performed on the diff — no new findings.

- [x] Override Linear Check


Danelegend (approved):
Nice

PR #10437, opened 2026-04-21
feat(auth): gate /auth/login with captcha

nmgarza5:
## Description

Stacks on #10428. Adds `LoginCaptchaMiddleware` which verifies an `X-Captcha-Token` header against the reCAPTCHA Enterprise Assessment API before the fastapi-users login handler runs. Credential-stuffing attempts now cost the attacker a fresh captcha token per try.

New `CaptchaAction.LOGIN` value is checked for strict action equality against the Enterprise response, so a token minted for signup or OAuth cannot be replayed against login.

Frontend `basicLogin` gains an optional `captchaToken` parameter and sends it as `X-Captcha-Token`. `EmailPasswordForm` always calls `getCaptchaToken("login")` before `basicLogin` — including the auto-login that follows a successful signup.

No-op when `is_captcha_enabled()` is false, so self-hosted and dev deployments without the Enterprise envs set pass through unchanged.

### Unit
5/5 passing in `test_login_captcha_middleware.py`: captcha-disabled pass-through, missing header rejected, bad token rejected, valid token passes with correct `CaptchaAction.LOGIN` argument, other endpoints (`POST /auth/register`, `GET /auth/login`) not touched.

### End-to-end against real Google Enterprise API (dev1 worktree, localhost:3010)

**Login happy path (UI):** logged in via `/auth/login`; backend log shows one successful assessment `action=login`.

**Login — no captcha header:**
```bash
$ curl -i -X POST http://localhost:3010/api/auth/login \
    -H 'Content-Type: application/x-www-form-urlencoded' \
    --data-urlencode 'username=you@example.com' --data-urlencode 'password=whatever'
HTTP/1.1 403 Forbidden
{"error_code":"UNAUTHORIZED","detail":"Captcha token is required"}
```
Fires **before** the password is checked — credential stuffing is blocked at the captcha gate.

**Login — malformed token:**
```bash
$ curl -i -X POST http://localhost:3010/api/auth/login \
    -H 'Content-Type: application/x-www-form-urlencoded' \
    -H 'X-Captcha-Token: bogus' \
    --data-urlencode 'username=you@example.com' --data-urlencode 'password=whatever'
HTTP/1.1 403 Forbidden
{"error_code":"UNAUTHORIZED","detail":"Captcha verification failed: MALFORMED"}
```

**Signup happy path (UI):** proved the `basicLogin` auto-login now gets its own `action=login` token and succeeds (initial discovery bug — the signup form was previously calling `basicLogin` with `undefined` captcha, tripping the new middleware).

### Skipped e2e paths (covered by unit tests)

- Action mismatch (signup token redeemed at `/auth/login`) → replay cache fires first in practice; action check covered by `test_captcha_enterprise.py::test_action_mismatch_rejects_strictly`
- Middleware bypass on disabled / non-POST / wrong path → `test_login_captcha_middleware.py::test_passes_through_when_captcha_disabled` + `test_does_not_gate_other_endpoints`

## Additional Options

### DIFF
--- a/backend/onyx/auth/captcha.py
+++ b/backend/onyx/auth/captcha.py
@@ -59,6 +59,7 @@ class CaptchaAction(StrEnum):
     one endpoint cannot be replayed against another."""
 
     SIGNUP = "signup"
+    LOGIN = "login"
     OAUTH = "oauth"
 
 
--- a/backend/onyx/main.py
+++ b/backend/onyx/main.py
@@ -65,6 +65,7 @@ from onyx.file_store.file_store import get_default_file_store
 from onyx.hooks.registry import validate_registry
 from onyx.server.api_key.api import router as api_key_router
 from onyx.server.auth.captcha_api import CaptchaCookieMiddleware
+from onyx.server.auth.captcha_api import LoginCaptchaMiddleware
 from onyx.server.auth.captcha_api import router as captcha_router
 from onyx.server.auth_check import check_router_auth
 from onyx.server.documents.cc_pair import router as cc_pair_router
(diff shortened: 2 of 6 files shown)


nmgarza5:
@greptile

jessicasingh7 (approved):

PR #10477, opened 2026-04-22
feat(telemetry): capture real client IP on every request

nmgarza5:
## Description

Server-side PostHog captures don't set `\$ip`, so GeoIP enrichment is a no-op and every event looks like it came from the api-server pod. The per-IP signup rate limiter read the leftmost `X-Forwarded-For` hop, which a client can trivially spoof to reset its bucket.

**Changes:**

- `onyx/utils/client_ip.py` — new module. `get_client_ip(request)` walks `X-Forwarded-For` right-to-left and returns the first globally-routable hop (skips private / reserved / loopback proxy addresses). `ClientIPMiddleware` extracts it once per request into a contextvar; `current_client_ip()` reads it.
- `onyx/main.py` — register `ClientIPMiddleware` as outermost.
- `ee/onyx/utils/telemetry.py` — `event_telemetry` / `identify_user` merge `\$ip` from the contextvar into properties before `posthog.capture` / `posthog.identify`. Logs pre-enrichment properties so real IPs never hit log aggregators.
- `onyx/auth/signup_rate_limit.py` + `onyx/server/auth/captcha_api.py` — `_client_ip` and `_client_ip_for_log` delegate to `get_client_ip`, closing the leftmost-XFF spoof vector.

No env vars, no config, no migration. Contextvar is `None` outside a request (Celery, startup) → `\$ip` is omitted and payload shape is unchanged.

## Additional Options

- [x] Override Linear Check
- [ ] This PR is ready to cherry-pick to a release branch

### DIFF
--- a/backend/onyx/utils/client_ip.py
+++ b/backend/onyx/utils/client_ip.py
@@ -1,0 +1,32 @@
+"""Extract the real client IP from a FastAPI request.
+
+nginx-ingress forwards the real client address in ``X-Forwarded-For`` while
+``request.client.host`` is the in-cluster proxy address. Callers use this for
+rate limiting, PostHog ``$ip`` enrichment (drives GeoIP), and abuse attribution.
+
+Only globally-routable addresses are returned so private/loopback/link-local
+values (pod-to-pod hops, localhost) never leak into downstream systems as
+though they were the client.
+"""
+
+import ipaddress
+
+from fastapi import Request
+
+
+def _is_globally_routable(ip_str: str) -> bool:
+    try:
+        return ipaddress.ip_address(ip_str).is_global
+    except ValueError:
+        return False
+
+
+def get_client_ip(request: Request) -> str | None:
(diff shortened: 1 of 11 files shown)


backend/onyx/utils/client_ip.py
```diff
@@ -0,0 +1,80 @@
...
+threadpool handoffs that use ``contextvars.copy_context()``.
+"""
+
+import ipaddress
+from contextvars import ContextVar
+
+from fastapi import Request
+from starlette.middleware.base import BaseHTTPMiddleware
+from starlette.middleware.base import RequestResponseEndpoint
+from starlette.requests import Request as StarletteRequest
```
Bo-Onyx, line 24: fastapi.Request is starlette.requests.Request — FastAPI re-exports it verbatim. These two imports resolve to the same class object, so StarletteRequest in the dispatch type hint is just an alias for the same thing as Request. One import is sufficient.

Bo-Onyx (approved):

PR #10478, opened 2026-04-22
feat(tracing): auto-wrap every LLM subclass with braintrust generation_span

nmgarza5:
## Description

Auto-wrap `invoke` and `stream` on every concrete subclass of the abstract `LLM` base (`backend/onyx/llm/interfaces.py`) with a fallback braintrust `generation_span`. Applied at class-creation time via `LLM.__init_subclass__`, so every LLM call — from any call site, including future `LLM` subclasses — is captured in Braintrust without per-callsite or per-subclass instrumentation.

### Why

Onyx's existing Braintrust tracing is opt-in via per-callsite `llm_generation_span` wraps. Paths that don't wrap explicitly bypass Braintrust entirely, producing incomplete observability and no way to attribute LLM usage across the codebase. A previously-existing `@traced` decorator on the abstract `LLM.invoke` never actually ran, because subclass overrides bypass base-class decorators. This PR replaces that dead decorator with a working class-level wrap.

### How

- **New module** `backend/onyx/llm/tracing_wrap.py`: `wrap_invoke`, `wrap_stream`, `_outer_generation_span_active`, and `_extract_prompt` helpers. Tracing imports are lazy inside the wrappers to break a circular import with `onyx.tracing.llm_utils` (which imports `LLM`).
- **`LLM.__init_subclass__`** in `interfaces.py`: inspects `cls.__dict__` for `invoke`/`stream`. If either is defined on this specific subclass (not inherited), it's replaced with the wrapped version via `setattr`. Inherited wraps are not re-wrapped, preventing double-wrap.
- **Idempotent wrappers**: each sets a `_onyx_tracing_wrapped` attribute so accidental double-application is a no-op.
- **Outer-span guard** (`_outer_generation_span_active`): when a caller has already opened a `generation_span` via `llm_generation_span`, the wrap short-circuits so Braintrust never double-counts. Includes an `ended_at is None` check — `SpanImpl.__exit__` intentionally skips `Scope.reset_current_span` on `GeneratorExit`, so a stale finished span can otherwise linger in the contextvar and suppress subsequent fallback tracing.
- **Prompt forwarding**: `wrap_invoke` and `wrap_stream` pull the `prompt` argument from `args[0]` / `kwargs['prompt']` and pass it as `input_messages` to `llm_generation_span`, so the fallback span records the prompt alongside cost/usage.
- **Stream wrapper** accumulates yielded content + final usage and records on clean stream completion. Tool-call deltas are intentionally NOT accumulated — streaming deltas are partial fragments keyed on `index` that need `litellm.stream_chunk_builder`-style reassembly before being safe to log. Can be added as a follow-up.
- **Removed** dead `@traced(name="invoke llm", type="llm")` on the abstract `LLM.invoke`.

### Coverage

| Call-site pattern | Before | After |
|---|---|---|
| Explicit `llm_generation_span` wrap | traced | unchanged (guard short-circuits fallback) |
| `llm.invoke` / `llm.stream` called directly without explicit wrap | not traced | traced via fallback |
| Future `LLM` subclasses that override `invoke` / `stream` | would require manual instrumentation | automatically wrapped at class creation |

### Fallback-span tagging

Spans created by the fallback are tagged `flow=llm_invoke_fallback` or `flow=llm_stream_fallback`, so Braintrust queries can distinguish explicit callsite wraps from the catch-all.

## Additional Options

### DIFF
--- a/backend/onyx/llm/interfaces.py
+++ b/backend/onyx/llm/interfaces.py
@@ -1,7 +1,7 @@
 import abc
 from collections.abc import Iterator
+from typing import Any
 
-from braintrust import traced
 from pydantic import BaseModel
 
 from onyx.llm.model_response import ModelResponse
@@ -9,6 +9,8 @@ from onyx.llm.model_response import ModelResponseStream
 from onyx.llm.models import LanguageModelInput
 from onyx.llm.models import ReasoningEffort
 from onyx.llm.models import ToolChoiceOptions
+from onyx.llm.tracing_wrap import wrap_invoke
+from onyx.llm.tracing_wrap import wrap_stream
 from onyx.utils.logger import setup_logger
 
 logger = setup_logger()
@@ -34,12 +36,31 @@ class LLMConfig(BaseModel):
 
 
 class LLM(abc.ABC):
+    """Abstract base for every LLM backend used by Onyx.
+
+    Concrete subclasses have their ``invoke`` and ``stream`` methods
+    auto-wrapped (via ``__init_subclass__`` below) with a fallback braintrust
+    ``generation_span``. This guarantees that every LLM call — from any call
+    site, including future subclasses — is captured in braintrust without
+    per-callsite instrumentation. Callers that explicitly wrap their calls
+    with ``llm_generation_span`` are unaffected: the fallback detects the
+    outer span and no-ops.
+    """
+
+    def __init_subclass__(cls, **kwargs: Any) -> None:
+        super().__init_subclass__(**kwargs)
+        invoke_fn = cls.__dict__.get("invoke")
+        if invoke_fn is not None:
+            setattr(cls, "invoke", wrap_invoke(invoke_fn))
+        stream_fn = cls.__dict__.get("stream")
+        if stream_fn is not None:
+            setattr(cls, "stream", wrap_stream(stream_fn))
+
     @property
     @abc.abstractmethod
     def config(self) -> LLMConfig:
         raise NotImplementedError
 
-    @traced(name="invoke llm", type="llm")
     def invoke(
         self,
         prompt: LanguageModelInput,
--- a/backend/onyx/llm/interfaces.py
+++ b/backend/onyx/llm/interfaces.py
@@ -1,4 +1,5 @@
 import abc
+from collections.abc import Callable
 from collections.abc import Iterator
 from typing import Any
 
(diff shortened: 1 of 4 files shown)


backend/onyx/llm/tracing_wrap.py
```diff
@@ -0,0 +1,174 @@
...
+        and isinstance(current.span_data, GenerationSpanData)
+        and current.started_at is not None
+        and current.ended_at is None
+    )
+
+
+def _extract_prompt(args: tuple[Any, ...], kwargs: dict[str, Any]) -> Any | None:
+    """Return the ``prompt`` argument passed to ``invoke``/``stream``, if any.
+
+    ``LLM.invoke`` and ``LLM.stream`` take ``prompt`` as the first positional
```
Danelegend: This seems like a fragile assumption. I would recommend some validation that errors:
```
sig = inspect.signature(fn)
params = sig.parameters
if "prompt" not in params
```

OR/PLUS using the bindings to get the parameter
```
bound = sig.bind(self, *args, **kwargs)
prompt = bound.arguments.get("prompt")
```

backend/onyx/llm/interfaces.py
```diff
@@ -34,12 +36,31 @@ class LLMConfig(BaseModel):
...
+    auto-wrapped (via ``__init_subclass__`` below) with a fallback braintrust
+    ``generation_span``. This guarantees that every LLM call — from any call
+    site, including future subclasses — is captured in braintrust without
+    per-callsite instrumentation. Callers that explicitly wrap their calls
+    with ``llm_generation_span`` are unaffected: the fallback detects the
+    outer span and no-ops.
+    """
+
+    def __init_subclass__(cls, **kwargs: Any) -> None:
+        super().__init_subclass__(**kwargs)
```
Danelegend, line 52: Perhaps something a bit some explicit:
```
    def __init_subclass__(cls, **kwargs: Any) -> None:
        super().__init_subclass__(**kwargs)
        cls._wrap_method_if_defined("invoke", wrap_invoke)
        cls._wrap_method_if_defined("stream", wrap_stream)

    @classmethod
    def _wrap_method_if_defined(
        cls,
        name: str,
        wrapper_fn: Callable[[Callable[..., Any]], Callable[..., Any]],
    ) -> None:
        fn = cls.__dict__.get(name)
        if fn is not None:
```

Danelegend (approved):
CRAZY STUFF

backend/onyx/llm/interfaces.py
nmgarza5: i like it

PR #10486, opened 2026-04-22
feat(tracing): reassemble streaming tool-call deltas on fallback span

nmgarza5:
## Description

Follow-up to #10478. Adds tool-call-delta reassembly to the streaming fallback wrap in `onyx/llm/tracing_wrap.py` so Braintrust captures complete tool invocations on streamed calls.

**Stacked on `nikg/trace-all-llm-invocations`** — merge that PR first.

### The problem

LiteLLM streaming delivers tool calls as partial fragments per chunk, keyed on ``index``. Typical pattern:

```
chunk 1: tool_calls=[{id:"call_123", index:0, function:{name:"search", arguments:"{\"qu"}}]
chunk 2: tool_calls=[{id:None,       index:0, function:{name:None,     arguments:"ery\":"}}]
chunk 3: tool_calls=[{id:None,       index:0, function:{name:None,     arguments:" \"hi\"}"}}]
```

The original stream wrap in #10478 intentionally passed ``tool_calls=None`` to ``record_llm_span_output`` because blindly extending a list with every delta would produce `N` malformed partial entries per tool call — each with a different subset of `id` / `name` / `arguments`. `record_llm_span_output` then calls `.model_dump()` on each, writing fragmented duplicate records.

### The fix

Two new helpers in `tracing_wrap.py`:

- **`_merge_tool_call_delta(buffer, delta)`** — merges a single streaming delta into a per-`index` buffer. Preserves the first-seen `id` and `function.name`, concatenates `function.arguments` fragments.
- **`_finalize_tool_calls(buffer)`** — converts the reassembled buffer into a list of complete `onyx.llm.models.ToolCall` objects, sorted by `index`. Entries missing required fields (`id` or `function.name`) are skipped rather than fabricated.

The stream wrap now runs these during the chunk loop and passes the finalized list to `record_llm_span_output`. Braintrust displays one complete tool-call entry per invocation, matching the metadata level of the non-streaming `invoke` path (which already gets full reassembly via `litellm.stream_chunk_builder` before `record_llm_response` runs).

### Tests

Six new test cases in `backend/tests/unit/onyx/llm/test_tracing_wrap.py`:

- `test_merge_tool_call_delta_single_call_across_chunks`
- `test_merge_tool_call_delta_multiple_calls_by_index` — verifies per-`index` keying with interleaved deltas
- `test_merge_tool_call_delta_does_not_overwrite_first_id_or_name`
- `test_finalize_tool_calls_skips_entries_missing_required_fields`
- `test_finalize_tool_calls_returns_none_for_empty_buffer`
- `test_stream_forwards_every_chunk_unchanged_when_tool_calls_present` — regression guard: downstream consumers still receive the raw delta fragments; reassembly is internal to the span

All 24 tests pass (18 from the base PR + 6 new).

## How Has This Been Tested?

a bunch of local testing
https://www.braintrust.dev/app/Onyx%202/p/Nik%20Local/logs?r=trace_20aefa5118da4c83bcac554b15c8dd61&s=caa1a8b3-be68-42bf-a8bd-0443c3e68fb0

## Additional Options

### DIFF
--- a/backend/onyx/llm/tracing_wrap.py
+++ b/backend/onyx/llm/tracing_wrap.py
@@ -7,9 +7,10 @@ no-op when an outer `generation_span` is already active — callers that
 explicitly wrap their calls (via `llm_generation_span`) continue to work and
 are not double-counted.
 
-Imports from `onyx.tracing.*` are performed lazily inside the wrappers to
-avoid an import cycle between `onyx.llm.interfaces` and
-`onyx.tracing.llm_utils` (which itself imports `LLM`).
+Imports from `onyx.tracing.llm_utils` stay lazy (inside the wrappers) because
+it imports `onyx.llm.interfaces`, which imports this module — loading it at
+module level would deadlock the import graph. Everything else is imported
+at the top of the file.
 """
 
 from __future__ import annotations
@@ -21,9 +22,14 @@ from collections.abc import Iterator
 from typing import Any
 from typing import TYPE_CHECKING
 
+from onyx.llm.model_response import ChatCompletionDeltaToolCall
+from onyx.llm.model_response import FunctionCall as DeltaFunctionCall
+from onyx.llm.model_response import Usage
+from onyx.tracing.framework.create import get_current_span
+from onyx.tracing.framework.span_data import GenerationSpanData
+
 if TYPE_CHECKING:
     from onyx.llm.interfaces import LLM
-    from onyx.llm.model_response import ChatCompletionDeltaToolCall
     from onyx.llm.model_response import ModelResponse
     from onyx.llm.model_response import ModelResponseStream
     from onyx.llm.models import ToolCall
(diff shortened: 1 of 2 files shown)


backend/onyx/llm/tracing_wrap.py
```diff
@@ -217,8 +224,99 @@ def wrapper(
...
+    - Subsequent chunks for the same ``index`` carry additional
+      ``function.arguments`` fragments with ``id`` / ``name`` set to ``None``.
+
+    This helper merges them in place: it preserves the first seen ``id`` and
+    ``function.name`` and concatenates ``function.arguments`` fragments. The
+    result is a dict of complete ``ChatCompletionDeltaToolCall`` objects
+    keyed by ``index`` that can be converted to fully-formed ``ToolCall``
+    objects via :func:`_finalize_tool_calls`.
+    """
+    from onyx.llm.model_response import ChatCompletionDeltaToolCall
```
Danelegend: Why in here?

Danelegend (approved):

Issue #6401, opened 2025-11-23
[Bug] Incorrect encoding detection causes Mojibake (corrupted Cyrillic text) in File Connectors


viacheslav-diachenko:
### 📋 Environment

- **Onyx Version:** 2.4.2 and under
- **Deployment:** Self-hosted via Docker Compose
- **Affected Components:** File Connectors (Confluence, Jira, File Upload)
- **Languages Affected:** Ukrainian, Russian, and other Cyrillic-based languages
- **Severity:** 🔴 **Critical** for non-Latin deployments

---

### 📝 Description

When indexing documents containing Cyrillic text from Confluence or Jira connectors, the text becomes corrupted during encoding detection.

The `detect_encoding()` function in `backend/onyx/file_processing/extract_file_text.py` blindly trusts the `chardet` library without first attempting standard UTF-8 decoding. `chardet` frequently misidentifies UTF-8 encoded Cyrillic text as legacy 8-bit codecs like `MacRoman`, `ISO-8859-1`, or `windows-1252`.

Since there is no initial validation, valid UTF-8 documents are forcefully decoded as legacy formats, resulting in **mojibake** (completely corrupted text) in the search index.

---

### 🔍 Root Cause

Current implementation (`backend/onyx/file_processing/extract_file_text.py`, lines ~130-135) has critical flaws:

- ❌ **No UTF-8 fast-path:** Does not attempt UTF-8 decoding first (would satisfy 99% of modern web inputs)
- ❌ **Blind trust:** Accepts low-confidence guesses from `chardet`
- ❌ **No blocklist:** Allows unlikely legacy encodings for modern content

---

### 🔄 Steps to Reproduce

1. Configure Confluence or Jira connector pointing to instance with Cyrillic content
2. Create document with Ukrainian text: `Привіт світ`
3. Start indexing via connector
4. Query the indexed document

**Expected Result:** `Привіт світ`  
**Actual Result:** `ÐŸÑ€Ð¸Ð²Ñ–Ñ‚ ÑÐ²Ñ–Ñ‚` (MacRoman interpretation of UTF-8 bytes)

---

### 💡 Proposed Solution

Multi-stage detection strategy that prioritizes UTF-8, checks BOM, and uses `chardet` as last resort with validation:

**File:** `backend/onyx/file_processing/extract_file_text.py`
```python
def detect_encoding(file: IO[bytes]) -> str:
    raw_data = file.read(50000)
    file.seek(0)
    
[...]

Issue #4862, opened 2025-06-10
Confluence connector missing dates added with /Date macro


tomgallagher:
Beautiful Soup parser in Confluence connector is failing to collect text from nested spans around dates inserted with a macro.

`<span class="date-lozenger-container" data-ssr-placeholder="UW2mx:EfLS5:URVzo:4y5Pz:qz-Pe:RPrst-3"><span class="date-node" data-node-type="date" data-timestamp="1750377600000">20 Jun 2025</span></span>`

Missing out dates in this way can make a Document redundant for useful search.

Not sure if this is happening because the dates are being added by JS after the document has rendered, or if it is as simple as adding span tags in `elif isinstance(e, bs4.element.Tag)` branch to  `format_document_soup` function in: 

> backend/onyx/file_processing/html_utils.py

Hopefully the latter.

PR #10488, opened 2026-04-23
fix(ConfluenceConnector): preserve /Date macro text and fix Cyrillic encoding

Bo-Onyx:
## Description

This PR fixed two issue in confluence connector

1. /Date macro text dropped from index
Confluence's /Date macro wraps dates in <span class="date-lozenger-container">. BeautifulSoup's format_document_soup was stripping these spans, silently dropping the date text from indexed content. Fixed by replacing each such span with its plain text before formatting.

Issue #4862 — Confluence connector missing dates added with /Date macro

2. Cyrillic / multi-byte attachment text corrupted (mojibake)
detect_encoding was passing all files through chardet first. chardet sometimes misidentifies valid UTF-8 (Cyrillic, CJK, accented Latin) as a legacy encoding (e.g. windows-1251), causing garbled text in the index. Fixed by attempting UTF-8 decode first — falling back to chardet only on UnicodeDecodeError. Since UTF-8 is self-validating, this is unambiguous. Affects all connectors using detect_encoding (Confluence, Jira, File Upload).

Issue #6401 — [Bug] Incorrect encoding detection causes Mojibake (corrupted Cyrillic text) in File Connectors (Confluence, Jira, File Upload)

https://linear.app/onyx-app/issue/ENG-4072/improve-confluence-connector

## How Has This Been Tested?
Added unit tests

## Additional Options

---
## Summary by cubic
Preserves Confluence /Date macro text and fixes corrupted Cyrillic/multi‑byte attachment text by preferring UTF‑8 in `detect_encoding`. Improves indexing for Confluence and any connector using `detect_encoding` (Linear ENG‑4072).

- **Bug Fixes**
  - Keep /Date macro text by replacing `span.date-lozenger-container` with its plain text before formatting.
  - Decode as UTF‑8 first; fall back to `chardet` only on UnicodeDecodeError, and reset the file cursor to 0 after sampling to prevent mojibake in attachments/uploads.

<sup>Written for commit 027ba4dbab7c161cc386737f54be4c77549b52ab. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/file_processing/extract_file_text.py
+++ b/backend/onyx/file_processing/extract_file_text.py
@@ -93,10 +93,24 @@ def is_text_file(file: IO[bytes]) -> bool:
 
 
 def detect_encoding(file: IO[bytes]) -> str:
+    """Detect the character encoding of a binary file.
+
+    Tries UTF-8 first — if the bytes decode cleanly, they are definitively UTF-8
+    (UTF-8 is self-validating). Falls back to chardet only when UTF-8 fails, since
+    chardet can misidentify valid UTF-8 text (e.g. Cyrillic) as a legacy encoding
+    like windows-1251, producing mojibake. Defaults to utf-8 if chardet gives up.
+
+    Resets the file cursor to 0 after sampling so callers can still read the full file.
+    """
     raw_data = file.read(50000)
     file.seek(0)
-    encoding = chardet.detect(raw_data)["encoding"] or "utf-8"
-    return encoding
+    try:
+        raw_data.decode("utf-8")
+        return "utf-8"
+    except UnicodeDecodeError:
+        # utf-8 failed — bytes are genuinely non-UTF-8, let chardet guess
+        encoding = chardet.detect(raw_data)["encoding"] or "utf-8"
+        return encoding
 
 
 def is_macos_resource_fork_file(file_name: str) -> bool:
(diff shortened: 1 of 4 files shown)


backend/onyx/file_processing/extract_file_text.py
```diff
@@ -93,10 +93,24 @@ def is_text_file(file: IO[bytes]) -> bool:
...
+    like windows-1251, producing mojibake. Defaults to utf-8 if chardet gives up.
+
+    Resets the file cursor to 0 after sampling so callers can still read the full file.
+    """
     raw_data = file.read(50000)
     file.seek(0)
-    encoding = chardet.detect(raw_data)["encoding"] or "utf-8"
-    return encoding
+    try:
+        raw_data.decode("utf-8")
```
evan-onyx, line 108: can calling decode() here consume the input bytes if a stream is passed in? if so I don't think we can safely do this (need to seek 0 in the except branch perhaps)

evan-onyx (approved):
1 nit

backend/onyx/file_processing/extract_file_text.py
Bo-Onyx: The concern is not applicable here. raw_data is a bytes object — already fully loaded into memory by file.read(50000). Calling .decode() on a bytes object creates a new string but doesn't modify or consume raw_data itself. It's still intact for chardet.detect(raw_data) in the except branch.

The file.seek(0) also already runs unconditionally before the try/except, so the file cursor is reset regardless of which branch is taken.

You might be thinking of a case where decode() is called directly on the stream — that's not what's happening here. raw_data is just bytes in memory.

PR #10507, opened 2026-04-23
feat(admin-agents): Agent ordering

raunakab:
## Description

Ensures the agent display order set by admins (via drag-and-drop in the admin agents table) is reflected in the user-facing explore agents page (`AgentsNavigationPage`).

Previously, agents were sorted by `b.id - a.id` (newest first). Now they use `personaComparator` which respects the `display_priority` field — agents with lower priority values appear first, and agents without a priority appear at the end sorted by ID.

Also includes fixes to the Opal Table's drag-and-drop implementation:
- Move `DndContext` outside `<table>` to fix invalid HTML nesting (`<div>` inside `<table>`)
- Make entire row the drag target instead of a hidden drag handle button
- Remove dnd-kit attributes from `<tr>` that interfered with pointer events

**Stack:** 3/3 — stacked on #10518.

## How Has This Been Tested?

- Type check passes (`bun run types:check`)

## Additional Options

### DIFF
--- a/web/lib/opal/src/components/table/TableBody.tsx
+++ b/web/lib/opal/src/components/table/TableBody.tsx
@@ -1,16 +1,5 @@
 "use client";
 
-import type { ReactNode } from "react";
-import {
-  DndContext,
-  DragOverlay,
-  type DragStartEvent,
-  type DragEndEvent,
-  type CollisionDetection,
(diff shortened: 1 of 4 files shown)


raunakab:
Not working; need a re-implementation.

PR #10513, opened 2026-04-23
feat: google oauth scope overrides

evan-onyx:
## Description

allow users to change what scopes they request from google during the login flow. useful mostly for pass-through OAuth so users don't have to authenticate twice (on login and when using an MCP server that uses google as Idp)

## How Has This Been Tested?

n/a pretty basic

## Additional Options

---
## Summary by cubic
Adds configurable Google OAuth login scopes via `GOOGLE_OAUTH_SCOPE_OVERRIDE`, with safe defaults in `GOOGLE_LOGIN_BASE_SCOPES`. Enables pass-through OAuth so users don’t re-consent when tools need extra Google scopes.

- **New Features**
  - Default login scopes: `["openid", "email", "profile"]` via `GOOGLE_LOGIN_BASE_SCOPES`.
  - Override scopes with `GOOGLE_OAUTH_SCOPE_OVERRIDE` (comma-separated).
  - OSS and EE apps now pass computed `google_login_scopes` to Google OAuth.
  - On bad overrides, logs an error and falls back to base scopes.

- **Migration**
  - To request extra scopes, set `GOOGLE_OAUTH_SCOPE_OVERRIDE="openid,email,profile,https://www.googleapis.com/auth/calendar"`.
  - Leave unset to use defaults.

<sup>Written for commit 9bcd5493bfa812806c10e0103a9a04938f136fa8. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/configs/app_configs.py
+++ b/backend/onyx/configs/app_configs.py
@@ -200,6 +200,28 @@ OAUTH_CLIENT_SECRET = (
 # Whether Google OAuth is enabled (requires both client ID and secret)
 OAUTH_ENABLED = bool(OAUTH_CLIENT_ID and OAUTH_CLIENT_SECRET)
 
+# Default scopes requested when signing in with Google (AUTH_TYPE=google_oauth
+# or AUTH_TYPE=cloud, and the BASIC + OAuth fallback path). These are the
+# minimum required to identify the user via OpenID Connect.
+GOOGLE_LOGIN_BASE_SCOPES = ["openid", "email", "profile"]
+
+# Applicable for Google OAuth login, allows you to override the scopes that
+# are requested from Google. Mirrors OIDC_SCOPE_OVERRIDE; useful when the
+# access token needs to be passed through to tool calls that require
+# additional Google API scopes.
+GOOGLE_OAUTH_SCOPE_OVERRIDE: list[str] | None = None
+_GOOGLE_OAUTH_SCOPE_OVERRIDE = os.environ.get("GOOGLE_OAUTH_SCOPE_OVERRIDE")
+
+if _GOOGLE_OAUTH_SCOPE_OVERRIDE:
+    try:
+        GOOGLE_OAUTH_SCOPE_OVERRIDE = [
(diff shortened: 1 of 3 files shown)


backend/onyx/configs/app_configs.py
```diff
@@ -200,6 +200,28 @@
...
+# access token needs to be passed through to tool calls that require
+# additional Google API scopes.
+GOOGLE_OAUTH_SCOPE_OVERRIDE: list[str] | None = None
+_GOOGLE_OAUTH_SCOPE_OVERRIDE = os.environ.get("GOOGLE_OAUTH_SCOPE_OVERRIDE")
+
+if _GOOGLE_OAUTH_SCOPE_OVERRIDE:
+    try:
+        GOOGLE_OAUTH_SCOPE_OVERRIDE = [
+            scope.strip() for scope in _GOOGLE_OAUTH_SCOPE_OVERRIDE.split(",")
+        ]
```
Danelegend, line 219: Is this a trust the user sort of deal, or are there scopes that are mandatory that we should error on?

backend/onyx/main.py
```diff
@@ -556,10 +558,18 @@ def get_application(lifespan_override: Lifespan | None = None) -> FastAPI:
     if AUTH_TYPE == AuthType.GOOGLE_OAUTH or (
         AUTH_TYPE == AuthType.BASIC and OAUTH_ENABLED
     ):
+        try:
+            google_login_scopes = list(
+                GOOGLE_OAUTH_SCOPE_OVERRIDE or GOOGLE_LOGIN_BASE_SCOPES
+            )
+        except Exception as e:
```
Danelegend: why would this error?

Danelegend (approved):

backend/onyx/configs/app_configs.py
evan-onyx: trust the user; anyone configuring this should know what they're doing and might appreciate the added flexibility.

backend/onyx/main.py
evan-onyx: I thought they could pass an empty string but yeah the or case makes this troll

PR #10521, opened 2026-04-23
feat(billing): redis cache for trial + billing info lookups 

nmgarza5:
## Description

One legitimately-indexing tenant generated **301 control-plane `/billing-information` calls in 30 min** — over half of control-plane load during that window. Root cause: `_check_chunk_usage_limit` calls `check_usage_and_raise` → `is_tenant_on_trial` → `fetch_billing_information` once per document batch. Uncached. Any large indexing run fans out.

**Fix:** cache `BillingInformation | SubscriptionStatusResponse` per tenant in shared Redis with a 24h TTL. `cached_is_tenant_on_trial` derives the boolean from the same cache entry so the two views can't drift.

**Fallback order** (every call):
1. Redis hit → return cached
2. Redis miss → control-plane fetch + write-through + return
3. Redis error (read or write) → log + uncached control-plane fetch
4. Corrupt cache entry → log + refetch + overwrite

**One call-site swap** covers the hot path: `ee/onyx/server/usage_limits.py` → `cached_fetch_billing_information`. `check_usage_and_raise`, `is_tenant_on_trial_fn`, and `bulk_invite_users` all reach this via the versioned-implementation dispatcher.

**New:** `BILLING_CACHE_TTL_SECONDS` env var (default 86400). `invalidate_billing_cache(tenant_id)` exposed for write paths that just mutated a subscription and want the next read to be fresh.

**Expected impact:** CP billing rate `~10/min/tenant → ~1/24h/tenant` on hot tenants. No tenant-side change.

**Out of scope** (per plan): CP-side webhook invalidation (follow-up), caching `fetch_tenant_stripe_information` (~1/day/tenant already), DB-vs-Stripe drift reconciliation.

## Additional Options

- [x] Override Linear Check
- [ ] This PR is ready to cherry-pick to a release branch

### DIFF
--- a/backend/ee/onyx/server/billing/api.py
+++ b/backend/ee/onyx/server/billing/api.py
@@ -33,6 +33,9 @@ from sqlalchemy.orm import Session
 
 from ee.onyx.db.license import get_license
 from ee.onyx.db.license import get_used_seats
+from ee.onyx.server.billing.billing_cache import (
+    invalidate_billing_cache as invalidate_indexing_trial_cache,
+)
 from ee.onyx.server.billing.models import BillingInformationResponse
 from ee.onyx.server.billing.models import CreateCheckoutSessionRequest
 from ee.onyx.server.billing.models import CreateCheckoutSessionResponse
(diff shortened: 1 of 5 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

justin-tahara (approved):

evan-onyx (approved):

PR #10567, opened 2026-04-24
fix(google-auth): convert RefreshError to PermissionError on revoked creds

nmgarza5:
## Description

Service-account credentials that have been deleted or revoked upstream surface a `google.auth.exceptions.RefreshError` at refresh time. Previously this bubbled up as an unhandled Celery exception in `connector_external_group_sync_generator_task`, flagging Sentry ~7 events/hr — Sentry: `ONYX-BACKEND-H6GF`.

- `get_google_creds` now catches `RefreshError` around `service_creds.refresh(...)` and converts it to a `PermissionError` with a clear "credentials likely revoked" message (chained via `from e`).
- `_timed_perform_external_group_sync` gains a `PermissionError` handler that marks the attempt as failed, increments the error metric, and logs at warning. Admin still sees the failed attempt + needs to reconnect, but the event is no longer an unhandled exception.

The existing `except Exception` fallback is unchanged so genuinely unexpected errors still surface as before.

## Additional Options


backend/onyx/connectors/google_utils/google_auth.py
```diff
@@ -157,7 +158,16 @@ def get_google_creds(
         )
 
         if not service_creds.valid or not service_creds.expired:
-            service_creds.refresh(Request())
+            try:
+                service_creds.refresh(Request())
+            except RefreshError as e:
```
evan-onyx: this breaks the error handling and retry that we use for this upstream

nmgarza5:
closing for now

PR #10594, opened 2026-04-24
test: verify knowledge toggle persists SearchTool to agent

wenxi-onyx:
## Description
Original PR description:
```
- Some agents (on cloud and likely on self-hosted instances) got in a bad state where they had document sets, connectors, or other knowledge attached, but did not explicitly have the SearchTool. As a result, using this agent would result in no search ever being performed

I think there were 2 ways to get in this state
- There was a bug on the agent editing page that would toggle Use Knowledge off no matter what (didn't check the db to see if knowledge was enabled) --> I think if you create an agent, enable knowledge, attach sources, save, then edit the agent and save again, it would create an endstate where knowledge is off, but there are sources attached. This was fixed in #10151 
- Creating agent via API directly and specifying knowledge sources, but not explicitly adding search tool
```

Update:
- we decided it's actually valid to have search tool disabled but sources configured (i.e. explicitly turning it off to do some tests, then turning it back on and having your configs saved)
- the desired end state is just informing the user that the agent config has search disabled (if the agent is compelled to do a search)
  - this won't be implemented until we do a bigger change to inject available sources into prompt

## Additional Options

---
## Summary by cubic
Personas with any attached knowledge sources now always include the `SearchTool` so they can search their knowledge. If the tool is missing from the DB, we now fail fast with a clear error; otherwise missing tools are auto-repaired at build time.

- **Bug Fixes**
  - In `upsert_persona`, auto-add `SearchTool` when document sets, hierarchy nodes, attached documents, or user files exist and it’s not already in `tools`, using `SEARCH_TOOL_ID` via `get_builtin_tool_by_id`; if not found, raise an error.
  - In tool construction, keep the safety net that detects knowledge without `SearchTool`, appends it, flushes, and builds it; logs a warning.

<sup>Written for commit f6aac62e67d4f232068662d521236e06851bbccf. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/db/persona.py
+++ b/backend/onyx/db/persona.py
@@ -44,7 +44,6 @@ from onyx.server.features.persona.models import PersonaSharedNotificationData
 from onyx.server.features.persona.models import PersonaSnapshot
 from onyx.server.features.persona.models import PersonaUpsertRequest
 from onyx.server.features.tool.tool_visibility import should_expose_tool_to_fe
-from onyx.tools.constants import SEARCH_TOOL_ID
 from onyx.utils.logger import setup_logger
 from onyx.utils.variable_functionality import fetch_versioned_implementation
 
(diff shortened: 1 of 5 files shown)


justin-tahara (approved):

wenxi-onyx:
@greptileai review this

wenxi-onyx:
@greptileai re-review

evan-onyx (approved):

PR #10599, opened 2026-04-24
fix(embedding): Self-blocking async code, other cleanups

acaprau:
## Description
When `EmbeddingModel.encode` is called from an async event loop (the Discord connector is the only current known source for this) an exception is raised because `_batch_encode_texts` delegates to async functions, but those invocations can never be evaluated because the caller event loop is already waiting for a result from the very same function.

Additionally, this function had some inefficiencies in its use of `partial` and the way it was using indices that was cleaned up.

## How Has This Been Tested?
Added unit tests for this function being called from both sync and async contexts, and for serial and batched implementations.

## Additional Options

---
## Summary by cubic
Fixes embedding deadlocks and event-loop leaks by detecting running loops and offloading direct API calls to a worker thread; sync paths reuse a thread‑local loop. Adds tests for sync/async and batch paths, and fixes a small OpenSearch comment typo.

- **Bug Fixes**
  - In `EmbeddingModel._batch_encode_texts`, detect a running event loop; for API providers, run `_make_direct_api_call` in a worker thread via `asyncio.run` when called from async, and reuse a thread‑local loop for sync callers to prevent leaks and deadlocks.
  - Unit tests cover async vs sync callers, single vs multi‑batch ordering, error propagation, and local model paths (async single‑batch uses `asyncio.run` once; no `asyncio.run` in multi‑batch async cases).

- **Refactors**
  - Removed unused `tenant_id`/`request_id` from `_make_direct_api_call`; avoided closure capture by passing batch args explicitly; simplified futures handling and timing logs; fixed "propogate" → "propagate" in `opensearch_document_index.py`.

<sup>Written for commit 4e146718ba056dc69cbd8881d9a2746cf09c7781. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/natural_language_processing/search_nlp_models.py
+++ b/backend/onyx/natural_language_processing/search_nlp_models.py
@@ -890,7 +890,7 @@ class EmbeddingModel:
                     asyncio.get_running_loop()
                 except RuntimeError:
                     # This code is being called synchronously, safe to use
-                    # asyncio.run.
+                    # run_until_complete.
                     # Use thread-local event loop to prevent memory leaks
                     # from creating thousands of event loops during batch
                     # processing.
--- a/backend/onyx/natural_language_processing/search_nlp_models.py
+++ b/backend/onyx/natural_language_processing/search_nlp_models.py
@@ -889,7 +889,7 @@ class EmbeddingModel:
                     # or not.
                     asyncio.get_running_loop()
                 except RuntimeError:
-                    # This code is being called synchonously, safe to use
+                    # This code is being called synchronously, safe to use
                     # asyncio.run.
                     # Use thread-local event loop to prevent memory leaks
                     # from creating thousands of event loops during batch
@@ -902,9 +902,7 @@ class EmbeddingModel:
                     # This code is being called from an event loop, can't
                     # block on it from the same thread without deadlocking.
                     # Run in a separate thread with its own loop.
-                    import concurrent.futures
-
-                    with concurrent.futures.ThreadPoolExecutor(max_workers=1) as pool:
+                    with ThreadPoolExecutor(max_workers=1) as pool:
                         response = cast(
                             EmbedResponse,
                             pool.submit(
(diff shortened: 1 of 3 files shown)


acaprau:
@greptile

backend/onyx/natural_language_processing/search_nlp_models.py
```diff
@@ -882,74 +880,91 @@ def process_batch(
...
+                            text_batch=batch,
                             tenant_id=tenant_id,
                             request_id=request_id,
                         )
-                    ): idx
-                    for idx, batch in enumerate(text_batches, start=1)
-                }
+                    )
+                    for idx, batch in enumerate(text_batches)
+                ]
```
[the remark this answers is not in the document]
acaprau: dammit good feedback, i should've caught that immediately.

acaprau:
@greptile

acaprau:
@greptile @cubic

backend/onyx/natural_language_processing/search_nlp_models.py
```diff
@@ -882,74 +880,92 @@ def process_batch(
...
         if num_threads >= 1 and self.provider_type and len(text_batches) > 1:
             with ThreadPoolExecutor(max_workers=num_threads) as executor:
-                future_to_batch = {
+                # NOTE: We explicitly do not use a lambda here because a lambda
+                # would close idx and batch by reference not value. By
+                # submitting directly to .submit we lose typing but the correct
+                # values are captured for each iteration.
+                futures = [
                     executor.submit(
-                        partial(
```
acaprau, line 920: using `partial` here adds nothing.

backend/onyx/natural_language_processing/search_nlp_models.py
```diff
@@ -882,74 +880,92 @@ def process_batch(
...
-                            process_batch,
-                            idx,
-                            len(text_batches),
-                            batch,
-                            tenant_id=tenant_id,
-                            request_id=request_id,
-                        )
-                    ): idx
```
acaprau, line 928: this dict did nothing.

backend/onyx/natural_language_processing/search_nlp_models.py
```diff
@@ -882,74 +880,92 @@ def process_batch(
...
-                    for idx, batch in enumerate(text_batches, start=1)
-                }
+                        process_batch,
```
evan-onyx: since we're changing this anyways, wanna make it a 0-arg lambda for type safety :D

backend/onyx/natural_language_processing/search_nlp_models.py
```diff
@@ -882,74 +880,92 @@ def process_batch(
...
-                # For API providers, make direct API call
-                # Use thread-local event loop to prevent memory leaks from creating
-                # thousands of event loops during batch processing
-                loop = _get_or_create_event_loop()
-                response = loop.run_until_complete(
-                    self._make_direct_api_call(
-                        embed_request, tenant_id=tenant_id, request_id=request_id
+                # For API providers, make direct API call.
+                try:
+                    # Detect if this code is being called from an event loop
```
evan-onyx, line 888: this pains me but if there is no better way I will simply avert my eyes

evan-onyx (approved):
ew

backend/onyx/natural_language_processing/search_nlp_models.py
acaprau: i had a similar thought, gotta read the comment i left about lambdas and closures in python.

acaprau:
@greptile @cubic

PR #10611, opened 2026-04-24
fix(llm): deliver bedrock bearer token via per-session provider, not global env

justin-tahara:
## Description

`POST /admin/llm/bedrock/available-models` fails with `AccessDeniedException` on `ListFoundationModels` whenever the caller uses the **Long-term API Key** auth method, even when the supplied bearer token is valid. The error names the managed-cluster pod IRSA role (e.g. `AmazonEKSTFWorkloadAccessRole-<cluster>`), which is the default-credential-chain fallback, not the bearer token the customer configured.

### Root cause

The endpoint set `AWS_BEARER_TOKEN_BEDROCK`, built a `boto3.Session`, and then popped the env var inside a `try/finally` — before `session.client("bedrock")` or `list_foundation_models()` actually ran:

```python
if request.aws_bearer_token_bedrock:
    try:
        os.environ["AWS_BEARER_TOKEN_BEDROCK"] = request.aws_bearer_token_bedrock
        session = boto3.Session(region_name=request.aws_region_name)
    finally:
        os.environ.pop("AWS_BEARER_TOKEN_BEDROCK", None)   # <-- popped too early

bedrock = session.client("bedrock")
model_summaries = bedrock.list_foundation_models()...      # <-- boto3 reads env var HERE
```

boto3 resolves `AWS_BEARER_TOKEN_BEDROCK` lazily at call time (not at `Session()` construction). By the time the call went out, the env var was gone, so boto3 fell through to SigV4 — landing on the pod's IRSA role on managed clusters, which has no `bedrock:*` permissions.

Reproduced live on a customer's managed cluster with their real bearer token:
- Same pattern as the endpoint → `AccessDeniedException: User: arn:aws:sts::...:assumed-role/AmazonEKSTFWorkloadAccessRole-<cluster>/... is not authorized to perform: bedrock:ListFoundationModels` (matches the customer's Sentry event).
- With the token held for the duration of the call → returns 125 models.

### Fix

Rather than mutating the process-global `AWS_BEARER_TOKEN_BEDROCK` env var (which the original code did, and which races across the threaded API server), the bearer token is now delivered through a **per-session botocore token provider**:

- `_build_bedrock_bearer_token_session` builds a `botocore.session.Session`, registers a `TokenProviderChain` containing a `_StaticBedrockBearerTokenProvider`, and wraps it in a `boto3.Session`. The provider is scoped to `signing_name == "bedrock"`, so it only ever supplies the token for Bedrock calls.
- The client is created with `Config(signature_version="bearer")`, which selects botocore's bearer-auth signer. This mirrors how botocore's own `ScopedEnvTokenProvider` resolves `AWS_BEARER_TOKEN_BEDROCK` internally — minus the global env state.
- **Masked-token round-trip**: when editing an existing provider without retyping the key, the frontend sends the masked placeholder (`abcd****wxyz`). The token is resolved back to the provider's stored `custom_config["AWS_BEARER_TOKEN_BEDROCK"]` (via `provider_name`) before the AWS call, matching the `update`/`test` endpoints' `_restore_masked_custom_config_values` behavior. Without this, re-fetching models on an existing provider would still hit AWS with the masked placeholder and fail.

This also removes the `os.environ` mutation and `try/finally` restore entirely.

### Why per-session instead of the env var

- **No cross-request race.** The API server is threaded; concurrent admin requests fetching Bedrock models would clobber / read each other's `AWS_BEARER_TOKEN_BEDROCK`. The token is now bound to a stack-local session object.
- **Tighter blast radius.** The provider only resolves for `signing_name == "bedrock"`; it can't leak into incidental STS/SSO calls on the same session.
- **Better secret hygiene.** The token never lands in `os.environ` — a surface that's inherited by subprocesses and captured wholesale by some error reporters.

## How Has This Been Tested?

New/updated unit tests in `test_fetch_models_api.py::TestGetBedrockAvailableModels`:
- `test_bearer_token_delivered_via_per_session_provider` — asserts the session is built from a `botocore_session` whose `token_provider` returns the request token for `signing_name="bedrock"` (and `None` for other services), that the client is created with `signature_version="bearer"`, and that `AWS_BEARER_TOKEN_BEDROCK` is never touched in the environment.
- `test_no_bearer_token_means_no_signature_override` — key/IAM auth path must not force bearer signing (`config is None`), so SigV4 still works.
- `test_masked_bearer_token_resolved_from_existing_provider` — sends the masked placeholder with a `provider_name`; asserts the per-session provider carries the real stored token, not the mask.
- `test_real_client_signs_with_bearer_token` — end-to-end with no mocks: builds a real client via the session helper and asserts the outbound request carries `Authorization: Bearer <token>`. Guards the botocore-internals coupling against silent breakage on upgrades.

Live reproducer against the customer's managed cluster confirmed the original failure; the per-session-provider approach was verified to produce a `Bearer` Authorization header against the pinned botocore 1.39.11.

## Additional Options


mhaslinsky:
Thanks for this, ran into the issue myself just now, and luckily found this before creating my own fix.

justin-tahara:
@greptile can you review one more time?

justin-tahara:
@cubic-dev-ai can you take a look as well?

jmelahman (approved):

PR #10634, opened 2026-04-26
fix: opensearch redis lock contention

evan-onyx:
## Description

We were seeing some contention for the lock which was being acquired on every index __init__, so to resolve that and save round trips to the index we can cache whether it looked fine when the process started. 

At the moment I have it set so we still check the index whenever a DocumentIndex is created, but there's some argument that it should actually be only a single request per process start OR maybe even a full validation whenever a DocumentIndex object is created. lmk ur thoughts

## How Has This Been Tested?

TBD based on the approach we take

## Additional Options

---
## Summary by cubic
Reduces Redis lock contention during OpenSearch index verification by caching verified index names per process and pinging the index on each construction. Skips repeated locks and round-trips while surfacing outages or deleted indices; prevents `RedisSharedLockAcquisitionError` under load.

- **Bug Fixes**
  - Added per-process `_verified_index_names` cache with early return.
  - Check index existence on every call; drop from cache if missing to recreate under the lock.
  - Re-check after acquiring the shared lock; extracted `_verify_and_create_index` for the locked path.
  - Keep schema current: create index if missing, otherwise `put_mapping`; log and raise on incompatible changes.

<sup>Written for commit 776d334b7aa60004f3732a7056fc3840ee6aee40. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/document_index/opensearch/opensearch_document_index.py
+++ b/backend/onyx/document_index/opensearch/opensearch_document_index.py
@@ -83,12 +83,11 @@ logger = setup_logger(__name__)
 VERIFY_INDEX_LOCK_TTL_S = 60
 VERIFY_INDEX_LOCK_BLOCKING_TIMEOUT_S = 60
 
-# Per-process cache of indices that have already been verified/created. Verify
-# is idempotent and the schema only changes on deploys (which restart the
-# process), so once we've verified an index once we can skip the redis lock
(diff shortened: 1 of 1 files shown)


evan-onyx:
We're going with a different approach: https://github.com/onyx-dot-app/onyx/pull/10737#top

PR #10647, opened 2026-04-27
fix(discord): scale to large servers + add forum channel support

### LINKED ISSUE #10642: Discord connector: fails on large channels + missing forum channel support
We've been running into two problems with the Discord connector and would like
to get sign-off on a fix before sending PRs.

## Problem 1: ingestion fails on large channels

We've been trying to ingest a Discord server with roughly 500k messages and
it keeps failing. Reading through `backend/onyx/connectors/discord/connector.py`,
a few things stand out as likely contributors.

The connector implements `PollConnector` and `LoadConnector` only (line 267),
so there's no checkpointing. Whenever a sync times out or crashes partway
through, the next run starts again from scratch. Slack already handles this by
using `CheckpointedConnector` with a per-channel watermark, and Discord could
follow the same pattern.

The `run_and_yield()` helper (line 244) wraps the async discord.py iterator in
a sync generator by calling `loop.run_until_complete(anext(...))` once per
document (line 250). At 500k messages that becomes 500k event-loop entries with
no progress signal and no heartbeat back to the indexer, which makes
worker-watchdog timeouts hard to avoid.

Channels are also drained sequentially with no intermediate yield
(`for channel in filtered_channels:` at line 236), so a single very large
channel has to fully complete before anything moves downstream. There's also no
explicit handling for Discord rate limits (429s); the code relies on whatever
discord.py decides to do internally.

This pushes against `CONTRIBUTING.md` §Performance and Correctness ("any
in-memory structure that can grow without bound based on input must be
periodically size-checked"; "make existing async code synchronous when
possible").

## Problem 2: forum channels are not ingested

Discord forum channels (`ChannelType.forum`) are silently skipped at
`connector.py:104`:

```python
if not isinstance(channel, TextChannel):
    continue
```

There's no support for them anywhere in the codebase (no references to `forum`,
`ForumChannel`, or `GUILD_FORUM`). Forum channels don't expose `.messages`
directly; they expose `.threads` and `.archived_threads()`, and each thread is
a forum post with its own message history. Adding support means recognising
the channel type and iterating threads rather than messages.

## Proposed approach

Two PRs, each kept under the 500-LOC ceiling from `CONTRIBUTING.md`
§Trunk-based development:

1. **Production fix, no user-visible change.** Convert `DiscordConnector` to
   `CheckpointedConnector`, mirroring the Slack `channel_completion_map`
   pattern. Replace the per-document `run_until_complete` wrapper with a single
   async run that streams batches. Add an `IndexingHeartbeatInterface` ping
   between batches and explicit 429 backoff.

2. **Forum support, behind a feature flag.** Branch on `ChannelType`; for
   `ForumChannel`, enumerate active and archived threads and treat each thread
   as a unit (post title plus messages). Reuse the checkpoint mechanism keyed
   by thread id. Add an optional "include forums" toggle in the connector
   config UI.

cc @yuhongsun96, would appreciate a quick sanity-check on the direction before
we start.

loispostula:
Resolves #10642.

## Why one PR

The original plan was two PRs (a fix + a feat). After implementation
the diff for both pieces ended up small and the two parts are tightly
coupled — the forum-support state machine reuses the checkpoint
primitives that PR1 introduces, and shipping the fix without forum
support would force a near-immediate follow-up. Two commits in one PR
keeps each piece atomic and reviewable while landing the full user
value at once.

### `fix(discord): use CheckpointedConnector for large-volume ingestion`

The Discord connector failed on servers with hundreds of thousands of
messages. Three correctness problems compounded: no checkpointing, an
async generator that wrapped the entire run with one
`loop.run_until_complete` per document, and channels (plus active +
archived threads inside them) drained sequentially in one async pass
with no intermediate yield.

Convert `DiscordConnector` to `CheckpointedConnector[DiscordCheckpoint]`,
mirroring Slack. The new state machine processes one channel-page or
thread-page per `load_from_checkpoint` cycle, persists per-channel and
per-thread cursors that drop completed entries to stay bounded, and
yields documents continuously so the indexer heartbeats naturally. The
async surface shrinks to a single page-fetch coroutine called via a
fresh event loop per page; the outer connector class is pure sync.

Threads inside text channels (active + archived) are now first-class
checkpoint state, fixing the same unbounded-fetch bug at thread
granularity.

### `feat(discord): add forum channel support and enumeration toggles`

Forum channels (`ChannelType.forum`) were silently dropped at channel
discovery. Add two opt-in fields on the connector spec:

- `include_forum_channels` (default `False`) — when enabled, forum
  channels are routed through the existing state machine. Forum posts
  are threads with no main-channel history, so the dispatcher pre-sets
  `current_channel_main_exhausted=True` on rollover into a forum and
  phases 3, 4, 5 handle the rest unchanged.
- `include_archived_threads` (default `True`, preserves prior
  behavior) — when disabled, `_enumerate_threads` skips the archived
  REST call.

The checkpoint gains one optional field, `forum_channel_ids: list[str]`,
populated once at cold start so rollover can pre-set the exhaustion
flag without an extra REST call. Existing in-flight checkpoints default
to an empty list, preserving prior behavior.

The connector form gains two checkboxes in the advanced values block.
For `include_archived_threads` we set `default: true` on the form field
to prevent newly-created connectors silently submitting `false` and
overriding the backend default; same idiom as the gitlab connector's
`include_mrs` / `include_issues`.

## Validation

- `uv run ty check` clean
- 13 tests pass: 11 ext-dep unit (six lifecycle/state-machine scenarios
  plus four PR2 forum scenarios in
  `backend/tests/external_dependency_unit/connectors/discord/test_discord_connector.py`),
  2 unit (validation)
- Tested end-to-end against a production server (~140 channels, mix of
  text and forum, indexed continuously past the 20k-doc mark with
  bounded checkpoint size and flat memory)

## Backwards compatibility

- Existing connector specs lack the new fields → Pydantic defaults
  preserve prior behavior
- Existing in-flight checkpoints lack `forum_channel_ids` → defaults to
  `[]`, never pre-sets the exhaustion flag → identical to prior
  behavior
- No DB migration, no registry change, no `pyproject.toml` change (the
  `discord-py==2.4.0` dep stays — it's also used by
  `backend/onyx/onyxbot/discord/`)
- Old `daily/connectors/discord/test_discord_connector.py` deleted
  (token-fragile; replaced by the ext-dep tests)

---
## Summary by cubic
Switches the Discord connector to checkpointed, page-based ingestion so it scales to very large servers and stays reliable, and adds opt‑in support for forum channels with a toggle for archived threads.

- **Refactors**
  - Converted to `CheckpointedConnector[DiscordCheckpoint]` with a 5‑phase state machine; fetches one channel/thread page per cycle and yields continuously.
  - Persists per‑channel/thread cursors; drops completed entries to keep checkpoints bounded and memory flat.
  - Narrows async to single‑page REST calls; the connector is now sync and heartbeats naturally.
  - Retries transient REST errors (rate limits, 5xx) with backoff; surfaces `Forbidden`/`NotFound` as `ConnectorFailure`; if retries exhaust, the error propagates for a clean checkpointed retry. Per‑message failures are isolated.
  - Enumerates archived threads one page per cycle to stay within the heartbeat budget.
  - Backwards compatible defaults; no DB or dependency changes.

- **New Features**
  - Forum channel support via `include_forum_channels` (off by default). Forums skip main history and drain posts as threads.
  - `include_archived_threads` toggle (default true) to skip archived threads on large servers.
  - Checkpoint adds `forum_channel_ids` to pre‑skip main history for forums.
  - UI: added “Include Forum Channels” and “Include Archived Threads” checkboxes (`default: true` for archived).

<sup>Written for commit f7bee6938a4716533ed8cf1b99952798a77728e3. Summary will update on new commits. <a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/10647?utm_source=github">Review in cubic</a></sup>


loispostula:
Addressed findings 1 and 2: `server_ids` / `channel_names` now use `None` sentinel defaults, and the dead `batch_size` parameter and assignment have been dropped. Leaving finding 3 (`guild.active_threads()` per-channel) as-is for this PR — happy to address it as a follow-up if you'd prefer.

loispostula:
Addressed @cubic-dev-ai's P1 finding plus a related heartbeat bug surfaced during production testing.

**Cubic finding (transient errors):** narrowed `_HISTORY_FAILURE_ERRORS` to `(Forbidden, NotFound)` only — those are genuinely permanent. Transient errors (`HTTPException` 5xx, `RateLimited`) are now retried inside the connector via a bounded backoff helper (`_retry_transient`, max 3 attempts, capped at ~30s total backoff). After exhaustion the exception propagates so the framework retries the attempt from the persisted checkpoint instead of silently advancing past missed data.

**Heartbeat bug:** in production we hit a 30-min heartbeat timeout because `channel.archived_threads(limit=None)` drains all archived threads in one cycle; on a chatty channel with rate-limit pressure that single cycle blows past the watchdog budget. Phase 3 (thread enumeration) is now split into 3a (active threads, one shot) and 3b (archived threads, **one page per cycle**). `DiscordCheckpoint` gains an additive optional `archived_thread_cursor: str | None = None`. Each cycle now returns within seconds in the worst case, keeping the heartbeat ticking.

Two new tests cover these: `test_transient_error_retries_then_propagates` and `test_archived_thread_enumeration_paginates_across_cycles`. Existing `test_main_channel_forbidden_skips_channel` was using `HTTPException(429)` despite the name; now uses an actual `Forbidden` so it tests what it says.

15 tests pass on the second commit, ty check clean. Backwards-compatible: existing checkpoints get `archived_thread_cursor=None` by default and resume from a cold archived enumeration.

PR #10635, opened 2026-04-27
feat: data models for index attempt metrics

evan-onyx:
## Description

data models for the new index attempt metrics display

## How Has This Been Tested?

tests in future PRs

## Additional Options

- [ ] [Optional] Please cherry-pick this PR to the latest release version.
- [x] [Optional] Override Linear Check

---
## Summary by cubic
Adds canonical stage enums and a DB model for index attempt metrics to power the new metrics UI. Creates the `index_attempt_stage_metric` table and links it to `IndexAttempt`.

- New Features
  - Introduced `IndexAttemptStage` (pipeline-ordered) and `StageScope` mapping for UI grouping and sorting.
  - Added `IndexAttemptStageMetric` with aggregates: count, total/ms, min/max, M2 (for std dev), and first/last event timestamps.
  - Linked via `IndexAttempt.stage_metrics`; unique per `(index_attempt_id, stage)` with cascade on attempt deletion.
  - Alembic migration creates `index_attempt_stage_metric` using non-native enum storage to allow adding stages without enum migrations.

- Migration
  - Run DB migrations (alembic upgrade).
  - No backfill required; table starts empty.

<sup>Written for commit 019d956ecc8f300be5b44865f623448813a826af. Summary will update on new commits.</sup>

PR #10689, opened 2026-04-28
fix(metrics): pass tenant_id explicitly to indexing-task connector resolver [ENG-4085]

justin-tahara:
## Description

The indexing-task celery signal callbacks (`on_indexing_task_prerun`, `on_indexing_task_postrun`) fire **outside** the `TenantAwareTask.__call__` body — which is where `CURRENT_TENANT_ID_CONTEXTVAR` actually gets set:

```
task_prerun signal       <-- contextvar NOT set
TenantAwareTask.__call__ <-- contextvar set from kwargs["tenant_id"]
task body                <-- contextvar available
__call__ finally         <-- contextvar cleared
task_postrun signal      <-- contextvar NOT set
```

`_resolve_connector` was reading the contextvar (always empty in this context) and calling `get_session_with_current_tenant()`, which then fell back to the `public` schema. In multi-tenant deployments the `connector_credential_pair` table only exists in `tenant_*` schemas, so every indexing-task callback was logging:

```
psycopg2.errors.UndefinedTable: relation "public.connector_credential_pair" does not exist
```

Caught at `DEBUG` level so tasks didn't fail, but every connector reported as `source=unknown` in metrics — defeating the purpose of the new per-connector labels added in #10635. Observed firing ~30/min/pod in the managed_services cluster.

**Fix**: thread `tenant_id` from `kwargs` into `_resolve_connector` and use `get_session_with_tenant(tenant_id=...)` directly. Also bail out early when `tenant_id` is missing or `"unknown"` so we never touch the DB without an explicit tenant.

## How Has This Been Tested?

- Updated all existing unit tests to use the new `_resolve_connector(cc_pair_id, tenant_id)` signature
- Added a regression test (`test_unknown_tenant_skips_db_lookup`) that asserts no DB call is made when `tenant_id` is `""` or `"unknown"` (the exact path that was hitting `public.connector_credential_pair`)
- All 16 tests pass: `pytest -xv backend/tests/unit/server/metrics/test_indexing_task_metrics.py`
- Verified post-deploy: error log rate for `UndefinedTable: public.connector_credential_pair` should drop to 0 in the managed_services cluster

## Additional Options

---
## Summary by cubic
Fixes multi-tenant connector resolution in indexing-task metrics by passing `tenant_id` explicitly, preventing public-schema lookups and restoring accurate per-connector labels. Reduces noisy `UndefinedTable` logs and improves monitoring accuracy for ENG-4085.

- **Bug Fixes**
  - Pass `tenant_id` into `_resolve_connector(cc_pair_id, tenant_id)` and read it from task kwargs.
  - Switch to `get_session_with_tenant(tenant_id=...)` (replaces `get_session_with_current_tenant()`).
  - Return early when `tenant_id` is missing or `"unknown"` to avoid DB access.
  - Updated tests to new signature; added `test_unknown_tenant_skips_db_lookup`.

<sup>Written for commit 8965995d7ed3b9213a9af818b83efce924d23bfb. Summary will update on new commits. <a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/10689?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/backend/onyx/server/metrics/indexing_task_metrics.py
+++ b/backend/onyx/server/metrics/indexing_task_metrics.py
@@ -32,7 +32,6 @@ from prometheus_client import Histogram
 from onyx.configs.constants import OnyxCeleryTask
 from onyx.server.metrics.celery_task_metrics import _MAX_START_TIME_AGE_SECONDS
 from onyx.utils.logger import setup_logger
-from shared_configs.contextvars import CURRENT_TENANT_ID_CONTEXTVAR
 
 logger = setup_logger()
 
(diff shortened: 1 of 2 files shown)


evan-onyx (approved):

backend/onyx/server/metrics/indexing_task_metrics.py
evan-onyx, line 188: any reason not to set the tenant id contextvar here anyways?

PR #10703, opened 2026-04-29
feat(onyx/backend): add ability to use GCS with Workload identity

Mberga14:
## Description

Added support for google storage client with support work workload identity - this is a security improvemenet as admins hosting onyx on Google compute engine (or GKE) do not need to generate long lived HMAC keys anymore to operate the s3-like google storage endpoint, and leave the authentication to IAM + metadata server (just like AWS 😄 )

## How Has This Been Tested?

Set up onyx locally with ADC and linked it to a GCS bucket - mocked LLM response, started a chat and uploaded some images/documents. Checked the GCS bucket contained the new elements

## Additional Options

---
## Summary by cubic
Add a native Google Cloud Storage backend for the file store with ADC/Workload Identity or service account auth. Set `FILE_STORE_BACKEND=gcs` to use IAM-based access without long‑lived HMAC keys; the default remains `s3`.

- **New Features**
  - `GCSBackedFileStore` with init, CRUD, size, rename, and prefix listing; wired via `FileStoreType.GCS` and `get_default_file_store()`.
  - Config: `GCS_FILE_STORE_BUCKET_NAME` (required), `GCS_FILE_STORE_PREFIX`, `GCS_PROJECT_ID`, `GCS_SERVICE_ACCOUNT_KEY_PATH`, `GCS_SERVICE_ACCOUNT_KEY_JSON`.
  - Docs/templates: README covers `s3`/`gcs`/`postgres`, GCS auth (ADC, key file, inline JSON), required IAM perms; Docker env adds a GCS block; blob connector docs add GCS auth with a security note.
  - Tests/deps: add `google-cloud-storage`; unit tests for ADC/key auth, bucket init/create, read/write/delete, rename, size, and backend selection.

- **Migration**
  - Set `FILE_STORE_BACKEND=gcs` and `GCS_FILE_STORE_BUCKET_NAME`; prefer ADC/Workload Identity or use a service account key (file or inline JSON).
  - In Docker Compose, remove the `s3-filestore` profile; no sidecar is needed for GCS.

<sup>Written for commit 9aaa2226a6449c3db29a04508a3126f3db3b5d6f. Summary will update on new commits.</sup>


uv.lock
```diff
@@ -5140,15 +5142,15 @@ wheels = [
 
 [[package]]
 name = "pdfminer-six"
```
justin-tahara, line 5144: I would just revert these changes

pyproject.toml
```diff
@@ -53,6 +52,7 @@ backend = [
     "google-api-python-client==2.86.0",
     "google-auth-httplib2==0.1.0",
     "google-auth-oauthlib==1.0.0",
+    "google-cloud-storage>=3.0.0,<4.0.0",
```
justin-tahara, line 55: I think this is the only thing that needs to be added. We should be able to revert the rest of the changes in the PR

justin-tahara (commented):
Some small nits. Running CI now so we should be able to uncover some other small issues as well

justin-tahara (approved):

PR #10729, opened 2026-04-30
fix(doc-processing): Cap num threads for contextual rag

Danelegend:
## Description
In the contextual rag part of the doc processing pipeline, we run assign_context on each chunk. assign_context is an I/O bound task hence we we want the threads. A thread in python I believe is allocated ~8mb of memory. This means that documents w/ large number of chunks can take up massive amounts of memory (2000 chunks = ~16gb).

This change caps the number of threads so that we don't reach our memory limits and OOM

## Additional Options


evan-onyx (approved):
so free

PR #10730, opened 2026-04-30
fix(google-connector): catch impersonation failures across all Google Drive API calls

Bo-Onyx:
## Description

Translates RefreshError from impersonated Google Drive API calls into two typed outcomes: silent skip (user confirmed deleted/suspended) or ImpersonationError (impersonation misconfiguration). The original proxy-based approach was replaced with a simpler guard after it caused silent data loss and confusing crashes at call sites that consume API objects as values rather than chaining to .execute().

Key changes
resources.py: Removes RefreshableDriveObject proxy and _ImpersonationState. Adds ImpersonationGuard — a simple stateful object that fires an Admin SDK callback at most once to check if a user was removed. get_drive_service returns a plain service object with no wrapping.
connector.py: Creates an ImpersonationGuard per user in _impersonate_user_for_retrieval, catches RefreshError directly at the validation gate and mid-run, and calls guard.is_user_removed() to determine which path to take.
file_retrieval.py / doc_conversion.py: Reverts the _unwrap() escape hatches that were needed to work around proxy limitations in BatchHttpRequest and MediaIoBaseDownload call sites — no longer needed.
Tests: test_refreshable_drive_object.py replaced by test_impersonation_guard.py; test_slim_retrieval.py updated to mock RefreshError + _get_all_user_emails directly.

https://linear.app/onyx-app/issue/ENG-4091/improving-google-connector-to-handle-user-removal-during-task-refresh

## How Has This Been Tested?
Updated unit tests
local test
Screenshot 2026-04-30 at 11 26 57 AM

## Additional Options

### DIFF
--- a/backend/onyx/connectors/google_drive/connector.py
+++ b/backend/onyx/connectors/google_drive/connector.py
@@ -60,6 +60,8 @@ from onyx.connectors.google_utils.google_utils import GoogleFields
 from onyx.connectors.google_utils.resources import get_admin_service
 from onyx.connectors.google_utils.resources import get_drive_service
 from onyx.connectors.google_utils.resources import GoogleDriveService
+from onyx.connectors.google_utils.resources import ImpersonationError
+from onyx.connectors.google_utils.resources import make_user_removal_checker
 from onyx.connectors.google_utils.shared_constants import (
     DB_CREDENTIALS_PRIMARY_ADMIN_KEY,
 )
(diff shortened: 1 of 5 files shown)


backend/onyx/connectors/google_drive/connector.py
```diff
@@ -862,196 +877,213 @@ def _impersonate_user_for_retrieval(
...
-                        include_shared_with_me=self.include_files_shared_with_me,
-                        max_num_pages=MY_DRIVE_PAGES_PER_CHECKPOINT,
-                        start=curr_stage.completed_until if resuming else start,
-                        end=end,
-                        cache_folders=not bool(curr_stage.completed_until),
-                        page_token=curr_stage.next_page_token,
-                    ),
-                    user_email,
-                    DriveRetrievalStage.MY_DRIVE_FILES,
+        def _post_validation_iteration() -> Iterator[RetrievedDriveFile]:
```
evan-onyx: I don't love how everything ends up indented as a result of this change; I'd actually prefer explicitly passing in all the arguments to the stage (this should make the PR's diff a lot smaller)

evan-onyx (approved):
approach is good, I have a nit about the code structure that I'd like to see changed

PR #10733, opened 2026-04-30
chore(tests): migrate connectors test to pytest_secrets

jmelahman:
## Description

Makes these tests easier to run locally so we can start squashing these flakes like https://onyx-company.slack.com/archives/C07K8KBMGKF/p1777567464769549 and https://onyx-company.slack.com/archives/C07K8KBMGKF/p1777481156995889

## How Has This Been Tested?

Captured by existing

## Additional Options


backend/tests/utils/secret_names.py
```diff
@@ -29,6 +29,61 @@ class TestSecret(StrEnum):
     OLLAMA_API_KEY = "OLLAMA_API_KEY"
     BEDROCK_API_KEY = "bedrock-api-key"
 
+    # Connector test secrets (mirrors .github/workflows/pr-python-connector-tests.yml).
```
Danelegend: Given that (I think) we're removing the references to `.github/workflows/pr-python-connector-tests.yml`, is this comment necessary?

backend/tests/utils/secret_names.py
```diff
@@ -29,6 +29,61 @@ class TestSecret(StrEnum):
...
+    # Member names match the CI env var; values match the AWS Secrets Manager
+    # key (the suffix after the ``test/`` prefix).
+    AWS_ACCESS_KEY_ID_DAILY_CONNECTOR_TESTS = "aws-access-key-id"
```
Danelegend, line 34: So to understand the flow more:
It used to be CI populates these values that we can request from aws?
Now they are defined here, so running tests locally can request them from aws?

Does this mean anyone running this locally will pull the creds, or is that some gate-keeping to only ci / us?

Danelegend (approved):
Approved, but quick security understanding on that last question

backend/tests/utils/secret_names.py
jmelahman: Yep. We're still fetching the same data (I moved the data from Github to AWS a few weeks ago). Here we're moving the fetch machinery to this `backend/tests/utils/` so that CI and developers fetch in a similar manner. This is gated by AWS managed permissions, so anyone with access to these in AWS can now,

```
aws login sso
uv run pytest backend/tests/daily/connectors/
```

This library also supports environment variables and `.env` files, so if folks were running these tests locally before, that way should still work.

PR #10735, opened 2026-04-30
feat(tracing): LLMFlow registry, wrap every LLM call site

nmgarza5:
## Description

Adds `onyx.tracing.flows.LLMFlow` — a single `StrEnum` registry of every operation that opens a generation span — and a `traced_llm_call` helper for sites that don't go through the `LLM` abstraction. All existing magic-string flows now use enum values, and every previously-untraced LLM/AI call has an explicit span.

New coverage:
- Indexing contextual-RAG (doc summary + chunk context)
- Image generation/edit across OpenAI, Azure, and Vertex providers
- Voice STT/TTS across OpenAI, Azure, and ElevenLabs providers
- Embedding (query + passage), reranking, and intent classification at the model_server boundary

The auto-wrap fallback in `tracing_wrap.py` now emits `LLMFlow.UNTAGGED_INVOKE` / `UNTAGGED_STREAM`, so any future untagged site is discoverable in dashboards instead of hiding under a generic label.

Provider tag is read from `EmbeddingProvider.value` / `RerankerProvider.value` (those enums are `(str, Enum)` not `StrEnum`, so `str(member)` would have rendered as `"EmbeddingProvider.OPENAI"`).

## Additional Options

### DIFF
--- a/backend/onyx/tracing/flows.py
+++ b/backend/onyx/tracing/flows.py
@@ -10,8 +10,6 @@ and makes it possible to enforce instrumentation coverage by searching for
 ``LLMFlow.UNTAGGED_*`` (the sentinel used by the auto-wrap fallback).
 """
 
-from __future__ import annotations
-
 from enum import StrEnum
 
 
--- a/backend/onyx/tracing/flows.py
+++ b/backend/onyx/tracing/flows.py
@@ -1,0 +1,69 @@
+"""Registry of flow tags applied to LLM-related generation spans.
+
+Each value identifies the *operation* a span represents — e.g.
+``contextual_rag_chunk_context`` or ``image_generation``. The provider, model,
+and deployment are captured separately on the span's ``model_config``, so the
+flow tag should describe **what the call does**, not **who serves it**.
+
+A single source of truth here lets dashboards group/filter without typo drift,
+and makes it possible to enforce instrumentation coverage by searching for
+``LLMFlow.UNTAGGED_*`` (the sentinel used by the auto-wrap fallback).
+"""
+
+from __future__ import annotations
+
+from enum import StrEnum
+
+
+class LLMFlow(StrEnum):
+    # Chat / agent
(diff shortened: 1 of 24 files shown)


Weves (approved):

backend/onyx/tracing/flows.py
```diff
@@ -0,0 +1,66 @@
...
+``contextual_rag_chunk_context`` or ``image_generation``. The provider, model,
+and deployment are captured separately on the span's ``model_config``, so the
+flow tag should describe **what the call does**, not **who serves it**.
+
+A single source of truth here lets dashboards group/filter without typo drift,
+and makes it possible to enforce instrumentation coverage by searching for
+``LLMFlow.UNTAGGED_*`` (the sentinel used by the auto-wrap fallback).
+"""
+
+from __future__ import annotations
```
Danelegend: Why this?

backend/onyx/tracing/llm_utils.py
```diff
@@ -10,25 +10,26 @@
...
 
-def build_llm_model_config(llm: LLM, flow: str | None = None) -> dict[str, str]:
+def build_llm_model_config(llm: LLM, flow: LLMFlow | None = None) -> dict[str, str]:
     model_config: dict[str, str] = {
         "base_url": str(llm.config.api_base or ""),
         "model_provider": llm.config.model_provider,
     }
-    if flow:
-        model_config["flow"] = flow
+    if flow is not None:
```
Danelegend: `if flow` should still work no?

Danelegend (approved):
Not a massive fan explicitly wrapping each call with the `with ...` but I can't think of a nicer way that doesn't involve a decent refactor

nmgarza5:
> Not a massive fan explicitly wrapping each call with the `with ...` but I can't think of a nicer way that doesn't involve a decent refactor

eh yeah the alternatives aren't great. I think its fine to be verbose about this

PR #10738, opened 2026-04-30
chore(ci): enable license enforcement in test infra via dev license

nmgarza5:
## Description

Replaces every `LICENSE_ENFORCEMENT_ENABLED=false` bypass on full-stack CI workflows with the new default (`true`). For workflows that boot the real backend (integration, playwright, nightly LLM provider chat), pulls a signed dev license from AWS Secrets Manager (`test/onyx-dev-license`) and seeds it into the DB before tests run via a new `scripts/seed_dev_license.py`. AWS auth + secret pull live in a new `./.github/actions/setup-test-license` composite action used by all five call sites.

Pure-pytest contexts (`pr-python-tests`, `pr-database-tests` OpenAPI step, `tests/integration/Dockerfile` OpenAPI step, `daily/conftest.py`) keep `LICENSE_ENFORCEMENT_ENABLED=false`. With enforcement on, `is_ee_version=True` activates EE versioned implementations that depend on Redis/Vespa, which unit tests don't run; `openapi-generator-cli` also rejects some EE-only schema shapes. Full-stack jobs are where the coverage gain matters.

Fork PR behavior: `pull_request` from forks does not propagate secrets, so `aws-actions/configure-aws-credentials` fails fast at OIDC. There is no silent enforcement bypass — that would let EE-validating tests pass without actually exercising enforcement.

Closes ENG-1: https://linear.app/onyx-app/issue/ENG-1/update-test-infra-to-use-test-license

## Additional Options

### DIFF
--- a/.github/workflows/pr-playwright-tests.yml
+++ b/.github/workflows/pr-playwright-tests.yml
@@ -682,26 +682,8 @@ jobs:
       - name: Start Docker containers (lite)
         run: |
           cd deployment/docker_compose
-          docker compose -f docker-compose.yml -f docker-compose.onyx-lite.yml -f docker-compose.dev.yml up -d
+          docker compose -f docker-compose.yml -f docker-compose.onyx-lite.yml -f docker-compose.dev.yml up -d --wait
         id: start_docker
-
-      - name: Wait for API server to be ready
-        run: |
-          start_time=$(date +%s)
-          timeout=300
-          while true; do
-            elapsed=$(( $(date +%s) - start_time ))
-            if [ $elapsed -ge $timeout ]; then
-              echo "Timeout reached. API server did not become ready in $timeout seconds."
-              exit 1
-            fi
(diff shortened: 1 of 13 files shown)


backend/scripts/seed_dev_license.py
nmgarza5, line 1: its more work than I wanna do to write this in GO so leaving as is

.github/workflows/pr-playwright-tests.yml
```diff
@@ -670,6 +685,29 @@ jobs:
           docker compose -f docker-compose.yml -f docker-compose.onyx-lite.yml -f docker-compose.dev.yml up -d
         id: start_docker
 
+      - name: Wait for API server to be ready
```
jmelahman: Let's prefer docker compose's `--wait`

jmelahman (approved):

PR #10746, opened 2026-04-30
feat: crawler playwright retry +improved error msg

evan-onyx:
## Description

While the web connector used a playwright fallback, we didn't previously have it hooked up to the Onyx web crawler (now we do). there are many cases where it just isn't feasible to get past cloudflare bot protection even with this, but in those cases we want the user to know that Onyx couldn't retrieve the info and why (which this PR does)

## How Has This Been Tested?

added tests for the discrete new stuff and checked E2E behavior in UI

## Additional Options

### DIFF
--- a/backend/onyx/tools/tool_implementations/open_url/onyx_web_crawler.py
+++ b/backend/onyx/tools/tool_implementations/open_url/onyx_web_crawler.py
@@ -46,6 +46,13 @@ class FailureReason:
         "cannot solve — try a different URL or configure Firecrawl as the "
         "web content provider"
     )
+    # Generic 403 with no Cloudflare evidence. Kept distinct from
+    # CLOUDFLARE_CHALLENGE so we don't tell the LLM to "configure Firecrawl"
+    # for what's actually an auth wall, expired presigned URL, private repo, etc.
+    HTTP_403_BLOCKED = (
+        "upstream returned HTTP 403 — the URL likely requires authentication "
+        "or is otherwise restricted from the built-in crawler"
+    )
     SSRF_BLOCKED = "blocked by SSRF protection (URL resolves to an internal address)"
     NETWORK_ERROR = "network error while fetching the URL"
     OVERSIZED_HTML = "HTML response exceeded the configured maximum size"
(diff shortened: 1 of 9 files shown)


backend/onyx/tools/tool_implementations/open_url/onyx_web_crawler.py
```diff
@@ -94,44 +178,50 @@ def _fetch_url(self, url: str) -> WebContent:
...
                 exc.__class__.__name__,
             )
-            return _failed_result(url)
+            return _failed_result(url, FailureReason.NETWORK_ERROR)
 
         if response.status_code >= 400:
+            cf_challenge_suspected = _looks_like_bot_challenge_response(response)
+            if cf_challenge_suspected and self._playwright_fallback_enabled:
+                logger.info(
+                    "Onyx crawler got %s for %s with bot-challenge signals; "
```
Danelegend: f"...{var}..." formatting pls

backend/onyx/tools/tool_implementations/open_url/onyx_web_crawler.py
```diff
@@ -25,14 +31,119 @@
...
+    as "not Cloudflare" here.
+    """
+    headers = response.headers
+    if any(name in headers for name in _CLOUDFLARE_HEADER_NAMES):
+        return True
+    server = headers.get("Server", "").lower()
+    return server.startswith("cloudflare")
+
+
+def _should_try_playwright_fallback(response: requests.Response) -> bool:
```
Danelegend, line 95: Can we add a pre-condition assertion of the status code's that should pass.

Or at the very least assert status_code != 200 (or 2XX)

backend/onyx/tools/tool_implementations/open_url/onyx_web_crawler.py
```diff
@@ -94,44 +212,57 @@ def _fetch_url(self, url: str) -> WebContent:
...
+            #     S3 URL gets the generic-403 reason instead).
+            has_cf_signals = _has_cloudflare_signals(response)
+            try_fallback = (
+                self._playwright_fallback_enabled
+                and _should_try_playwright_fallback(response)
+            )
+
+            if try_fallback:
+                logger.info(
+                    "Onyx crawler got %s for %s; retrying via Playwright "
```
Danelegend, line 244: f-string formatting pls

backend/onyx/tools/tool_implementations/open_url/onyx_web_crawler.py
```diff
@@ -143,28 +274,90 @@ def _fetch_url(self, url: str) -> WebContent:
...
+        """
+        rendered: RenderedPage | None = fetch_rendered_html(url)
+        if rendered is None:
+            return None
+
+        if (
+            self._max_html_size_bytes is not None
+            and len(rendered.html) > self._max_html_size_bytes
+        ):
+            logger.warning(
```
Danelegend, line 341: f-string again pls

backend/onyx/utils/playwright_fetch.py
```diff
@@ -0,0 +1,304 @@
...
+        validate_outbound_http_url(url)
+    except (SSRFException, ValueError) as exc:
+        logger.warning(
+            "Refusing Playwright fallback for %s (%s)", url, exc.__class__.__name__
+        )
+        return None
+
+    playwright: Playwright | None = None
+    context: BrowserContext | None = None
+    try:
```
Danelegend, line 270: Reasoning about the flow due to try-except-finally logic might be a bit difficult (or skill issue on my end).

For example, we have to know not to return in the finally's because then that will override whatever was returned in the try. I think having multiple try-except-finally makes this a little bit non-intuitive to reason about.

Is there a clearer way to write this?

Danelegend (approved):

PR #10749, opened 2026-05-01
fix: paginate role assignments

evan-onyx:
## Description

add pagination to role assignments in the sharepoint connector. We were seeing some failures that might just have been due to making too large of a request.

## How Has This Been Tested?

tested in UI

## Additional Options

### DIFF
--- a/backend/onyx/connectors/sharepoint/connector.py
+++ b/backend/onyx/connectors/sharepoint/connector.py
@@ -325,7 +325,7 @@ def sleep_and_retry(
             # into ClientRequestException with response=None (e.g.
             # ChunkedEncodingError).
             wrapped_transport_error = e.response is None and isinstance(
-                e.__cause__, TRANSIENT_TRANSPORT_EXCEPTIONS
+                e.__cause__ or e.__context__, TRANSIENT_TRANSPORT_EXCEPTIONS
             )
 
             is_retryable = status in RETRYABLE_HTTP_STATUSES or wrapped_transport_error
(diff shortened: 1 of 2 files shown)


backend/onyx/connectors/sharepoint/connector.py
```diff
@@ -252,35 +252,82 @@ def _site_page_in_time_window(
...
 
-            # 429 / 503 — rate limit or transient error.  Back off and retry.
-            if status in (429, 503) and attempt < max_retries:
-                logger.warning(
-                    f"Rate limit exceeded on {method_name}, attempt {attempt + 1}/{max_retries + 1}, sleeping and retrying"
-                )
+            # Some office365 SDK versions wrap transport errors (e.g.
+            # ChunkedEncodingError) into ClientRequestException with
+            # response=None. Treat that as transient too.
+            wrapped_transport_error = e.response is None and isinstance(
```
Bo-Onyx, line 327: The comment says the SDK wraps transport errors inside execute_query's except requests.exceptions.RequestException block. If the SDK uses implicit re-raise (raise ClientRequestException(...) inside an except block, without from e), the original exception lands in e.__context__, not e.__cause__.  Checking only __cause__ would silently miss those cases. Consider:
```
  wrapped_transport_error = e.response is None and isinstance(
      e.__cause__ or e.__context__, TRANSIENT_TRANSPORT_EXCEPTIONS
  )
```

Bo-Onyx (approved):

PR #10753, opened 2026-05-01
fix(chat): defer mainline-pointer PUTs until session is idle

nmgarza5:
## Description

Customers continue to hit `"The new message sent is not on the latest mainline of messages"` (500) on v3.2.11+. #10075/#10078 patched only the multi-model preferred-response path. Same race fires on every other place the client mutates the local message tree before its backend mainline pointer (`latest_child_message_id`) lands:

- branch-switcher arrows on a prior assistant response
- editing a prior user message
- regenerating mid-stream and sending a follow-up
- agent / persona switch on an existing session
- multi-model preferred-response selection

The repro that broke earlier attempts: while an agent stream is in flight, click a branch arrow on a parent message, then send a new message → 500.

The deepest cause: backend `create_new_chat_message` and the stream-completion writeback both unconditionally set `parent.latest_child_message_id = streaming_msg.id`. A user PUT issued while a stream is in flight gets clobbered by the stream's own commits, so a simple "wait for the PUT before sending" gate on the FE isn't enough.

This PR:

1. Defers `set-message-as-latest` and `set-preferred-response` PUTs until the session's `chatState` transitions to `"input"` (all streams finished saving). PUT then lands AFTER the streaming writeback and is not overwritten.
2. Tracks the deferred PUT in a module-level `pendingMainlineSync` promise; `sendMessage` awaits it before issuing the POST. A queued send (auto-flushed by `AppInputBar` when the stream ends) blocks on the PUT before posting.
3. If `chatState` is already `"input"`, `awaitChatStateInput` resolves immediately — common path adds no latency.

## Additional Options

### DIFF
--- a/web/src/app/app/message/MultiModelResponseView.tsx
+++ b/web/src/app/app/message/MultiModelResponseView.tsx
@@ -238,8 +238,12 @@ export default function MultiModelResponseView({
       // frontend chain walk disagrees, the next follow-up fails with
       // "not on the latest mainline".
       if (parentMessage?.messageId && response.messageId && currentSessionId) {
-        setPreferredResponse(parentMessage.messageId, response.messageId).catch(
-          (err) => console.error("Failed to persist preferred response:", err)
+        setPreferredResponse(
+          parentMessage.messageId,
+          response.messageId,
+          currentSessionId
+        ).catch((err) =>
+          console.error("Failed to persist preferred response:", err)
         );
 
         const tree = useChatSessionStore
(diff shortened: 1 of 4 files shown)


nmgarza5:
Replacing with backend fix — FE gate doesn't cover the streaming-overwrite race (backend itself rewrites latest_child_message_id at message creation and stream completion, racing the user's PUT independent of any FE work).

nmgarza5:
Closing — switching to BE approach. FE gate / defer-PUT both fail when backend's own stream-completion writeback clobbers latest_child_message_id. BE walks parent_message_id chain instead, immune to all races.

PR #10757, opened 2026-05-01
fix(chat): resolve send parent via parent_message_id chain

nmgarza5:
## Description

**Bug**: Customers hit `"The new message sent is not on the latest mainline of messages"` (500) on v3.2.11+. #10075/#10078 patched only the multi-model preferred-response path. Same race fires on branch-switching, edits, regenerate-mid-stream, agent switches, multi-tab clients.

**Why**: Backend identified the parent message by walking DOWN from root via `latest_child_message_id` (mainline pointer). That pointer is rewritten by 3 concurrent writers — client PUTs, `create_new_chat_message`, stream-completion writeback — so even a synchronized client could be clobbered by its own backend.

**Fix**: Walk UP from `parent_message_id` via the immutable `parent_message_id` chain. Tree structure never mutates after insert, so the lookup can't lose the race. `latest_child_message_id` becomes a UI bookmark, not a correctness gate.

**Perf**: Net 1 fewer O(N) SELECT per send.
- Bug-fix path uses the same `get_chat_messages_by_session` call as before — same query, same rows, opposite walk direction.
- Removed the post-save compression-check fetch in `llm_loop_completion_handle` — chain is already in memory at completion time, just append the assistant message instead of re-fetching the whole session.
- Skipped the redundant permission re-check on the bug-fix-path fetch (session ownership already validated upstream).

## Additional Options

### DIFF
--- a/backend/onyx/chat/chat_state.py
+++ b/backend/onyx/chat/chat_state.py
@@ -191,6 +191,10 @@ class ChatTurnSetup:
     llms: list[LLM]  # length 1 for single-model, N for multi-model
     model_display_names: list[str]  # parallel to llms
     simple_chat_history: list[ChatMessageSimple]
+    # Full chain of ChatMessage rows from the first non-root message up to and
+    # including the just-saved user message. Reused by `llm_loop_completion_handle`
+    # so the post-save compression check doesn't re-fetch every message in the session.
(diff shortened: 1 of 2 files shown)


nmgarza5:
Switching back to FE-only approach per request — defer mainline PUTs until stream completes.

PR #10758, opened 2026-05-01
feat(license): tiered expiry warnings + 14d grace period

nmgarza5:
## Description

Self-hosted licenses currently insta-expire with no warning, no grace period, and no admin email — surfaced when a customer's instance went down with no notice.

This adds tiered notifications driven by a daily Celery task (self-hosted only):

- **Pre-expiry banner + email + in-app notification** at 30d, 14d, and 1d before `expires_at`.
- **14-day post-expiry grace period** — instance stays up, persistent banner, one admin email per UTC date until renewal.

### Idempotency

Reuses the existing `notification` table's `(user_id, notif_type, COALESCE(additional_data, '{}'::jsonb))` unique index — no new tables, no new License columns. Email is sent only to admins whose row was actually inserted (`INSERT ... ON CONFLICT DO NOTHING RETURNING user_id`), so concurrent beat ticks can't double-email. `additional_data` keys: `{stage, expires_at}` for pre-expiry, `{stage, expires_at, sent_date}` for grace.

### API

`LicenseMetadata` (cached in Redis) gains `expiry_warning_stage` and `grace_days_remaining` fields, surfaced via the `/license` endpoint.

### Frontend

`LicenseExpiryBanner` renders via Opal `MessageCard` (matches billing-page style) — `warning` variant for T_30D / T_14D, `error` variant for T_1D / GRACE. Centered over `[data-main-container]` (sidebar excluded), fixed-positioned, overlays content. Re-measures on route change + DOM mutations so chat and admin layouts both center correctly.

Dismissal persists per stage in `localStorage`. Grace-stage key includes UTC date so each daily fire is independently dismissable.

Dev affordance: `?mock_license_stage=t_30d|t_14d|t_1d|grace` renders the banner against synthetic data on any page.

### Tests

- 21 unit tests for stage derivation across all boundaries.
- 8 external-dependency unit tests for `notify_admins_for_stage` covering the race-safe `INSERT...RETURNING` path: first call inserts + emails, second is a no-op, stage transitions and grace daily cadence each fire exactly once, basic users are never targeted, multiple admins all get notified.
- 5 external-dependency unit tests for the celery beat task covering multi-tenant skip, missing license row, signature failure, stage NONE short-circuit, and dispatch when in-window.

Multi-tenant cloud licensing is unaffected.

## How Has This Been Tested?

Screenshot 2026-05-01 at 3 11 18 PM

Screenshot 2026-05-01 at 1 37 24 PM

Screenshot 2026-05-01 at 1 37 14 PM

Screenshot 2026-05-01 at 1 37 05 PM

## Additional Options

### DIFF
--- a/backend/ee/onyx/background/celery/tasks/beat_schedule.py
+++ b/backend/ee/onyx/background/celery/tasks/beat_schedule.py
@@ -65,6 +65,15 @@ if not MULTI_TENANT:
             },
         },
         {
+            "name": "check-license-expiry-notifications",
+            "task": OnyxCeleryTask.CHECK_LICENSE_EXPIRY_NOTIFICATIONS,
+            "schedule": timedelta(hours=1),
+            "options": {
+                "priority": OnyxCeleryPriority.MEDIUM,
+                "expires": BEAT_EXPIRES_DEFAULT,
+            },
+        },
+        {
             "name": "autogenerate-usage-report",
             "task": OnyxCeleryTask.GENERATE_USAGE_REPORT_TASK,
             "schedule": timedelta(days=30),  # TODO: change this to config flag
--- a/backend/ee/onyx/background/celery/tasks/beat_schedule.py
+++ b/backend/ee/onyx/background/celery/tasks/beat_schedule.py
@@ -67,7 +67,7 @@ if not MULTI_TENANT:
         {
             "name": "check-license-expiry-notifications",
             "task": OnyxCeleryTask.CHECK_LICENSE_EXPIRY_NOTIFICATIONS,
-            "schedule": timedelta(hours=1),
+            "schedule": timedelta(days=1),
             "options": {
                 "priority": OnyxCeleryPriority.MEDIUM,
                 "expires": BEAT_EXPIRES_DEFAULT,
--- a/backend/ee/onyx/utils/license_expiry.py
+++ b/backend/ee/onyx/utils/license_expiry.py
@@ -49,8 +49,6 @@ def get_grace_period_end(expires_at: datetime) -> datetime:
 
 
 def get_grace_days_remaining(expires_at: datetime) -> int:
-    grace_end = get_grace_period_end(expires_at)
-    seconds_left = (grace_end - datetime.now(timezone.utc)).total_seconds()
-    if seconds_left <= 0:
-        return 0
-    return max(1, int(seconds_left // 86400) + (1 if seconds_left % 86400 > 0 else 0))
+    grace_end_date = get_grace_period_end(expires_at).date()
+    today = datetime.now(timezone.utc).date()
+    return max(0, (grace_end_date - today).days)
--- a/backend/ee/onyx/utils/license_expiry.py
+++ b/backend/ee/onyx/utils/license_expiry.py
@@ -27,14 +27,8 @@ class ExpiryWarningStage(str, Enum):
     GRACE = "grace"
 
 
-def get_expiry_warning_stage(
-    expires_at: datetime,
-    now: datetime | None = None,
-) -> ExpiryWarningStage:
-    if now is None:
-        now = datetime.now(timezone.utc)
-
-    seconds_remaining = (expires_at - now).total_seconds()
+def get_expiry_warning_stage(expires_at: datetime) -> ExpiryWarningStage:
+    seconds_remaining = (expires_at - datetime.now(timezone.utc)).total_seconds()
     days_remaining = seconds_remaining / 86400.0
 
     if days_remaining > 30:
@@ -54,11 +48,9 @@ def get_grace_period_end(expires_at: datetime) -> datetime:
     return expires_at + timedelta(days=LICENSE_GRACE_PERIOD_DAYS)
 
 
-def get_grace_days_remaining(expires_at: datetime, now: datetime | None = None) -> int:
-    if now is None:
-        now = datetime.now(timezone.utc)
+def get_grace_days_remaining(expires_at: datetime) -> int:
     grace_end = get_grace_period_end(expires_at)
-    seconds_left = (grace_end - now).total_seconds()
+    seconds_left = (grace_end - datetime.now(timezone.utc)).total_seconds()
     if seconds_left <= 0:
         return 0
     return max(1, int(seconds_left // 86400) + (1 if seconds_left % 86400 > 0 else 0))
--- a/backend/ee/onyx/utils/license_expiry.py
+++ b/backend/ee/onyx/utils/license_expiry.py
@@ -12,6 +12,7 @@ Stages:
 """
 
 from datetime import datetime
+from datetime import timedelta
 from datetime import timezone
 from enum import Enum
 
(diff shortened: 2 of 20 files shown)


backend/ee/onyx/utils/license_expiry.py
```diff
@@ -0,0 +1,64 @@
...
+    T_14D = "t_14d"
+    T_1D = "t_1d"
+    GRACE = "grace"
+
+
+def get_expiry_warning_stage(
+    expires_at: datetime,
+    now: datetime | None = None,
+) -> ExpiryWarningStage:
+    if now is None:
```
wenxi-onyx: It seems like the only place we explicitly pass `now` is in tests. Is it unnecessary boilerplate? Can we assume that `now is always datetime.now` and just patch it if necessary in the tests? Makes the API cleaner, unless I'm not understanding when `now` may not actually mean `datetime.now` in prod.

backend/ee/onyx/utils/license_expiry.py
```diff
@@ -0,0 +1,64 @@
...
+    if days_remaining > -LICENSE_GRACE_PERIOD_DAYS:
+        return ExpiryWarningStage.GRACE
+    return ExpiryWarningStage.NONE
+
+
+def get_grace_period_end(expires_at: datetime) -> datetime:
+    return expires_at + timedelta(days=LICENSE_GRACE_PERIOD_DAYS)
+
+
+def get_grace_days_remaining(expires_at: datetime, now: datetime | None = None) -> int:
```
wenxi-onyx: Same comment as above about `now: datetime | None = None` --> is it only used in tests?

backend/ee/onyx/utils/license_notifications.py
```diff
@@ -0,0 +1,235 @@
...
+from onyx.configs.constants import NotificationType
+from onyx.configs.constants import ONYX_DEFAULT_APPLICATION_NAME
+from onyx.db.models import Notification
+from onyx.db.models import User
+from onyx.utils.logger import setup_logger
+
+logger = setup_logger()
+
+
+def _get_admin_users(db_session: Session) -> list[User]:
```
wenxi-onyx: This seems like it would be a shared util. If it doesn't exist, maybe should be placed in a shared util file for future use?

backend/ee/onyx/utils/license_notifications.py
```diff
@@ -0,0 +1,235 @@
...
+    return list(
+        db_session.execute(
+            select(User).where(
+                User.is_active,  # ty: ignore[invalid-argument-type]
+                User.role == UserRole.ADMIN,
```
wenxi-onyx: Does this return API keys

backend/ee/onyx/utils/license_notifications.py
```diff
@@ -0,0 +1,235 @@
...
+    title: str,
+    description: str,
+    additional_data: dict[str, Any],
+) -> set[UUID]:
+    if not admin_ids:
+        return set()
+
+    now = datetime.now(timezone.utc)
+
+    stmt = (
```
wenxi-onyx: Is a raw DB statement here necessary? Any reason why `onyx/db/notification.py:create_notification` doesn't work?

backend/ee/onyx/utils/license_notifications.py
```diff
@@ -0,0 +1,235 @@
...
+    db_session.commit()
+    return inserted_ids
+
+
+def notify_admins_for_stage(
+    db_session: Session,
+    stage: ExpiryWarningStage,
+    expires_at: datetime,
+    today: date | None = None,
+) -> int:
```
wenxi-onyx: I don't see how this return value is used except for in tests. Is all of the logic around calculating this value necessary?

backend/ee/onyx/background/celery/tasks/beat_schedule.py
```diff
@@ -64,6 +64,15 @@
                 "expires": BEAT_EXPIRES_DEFAULT,
             },
         },
+        {
+            "name": "check-license-expiry-notifications",
+            "task": OnyxCeleryTask.CHECK_LICENSE_EXPIRY_NOTIFICATIONS,
+            "schedule": timedelta(days=1),
```
wenxi-onyx, line 70: 1 - does this also run on our cloud? Could it cause a crazy number of notifications/db load if so?

2 - Cloud has a time dilation factor of 8x or so, meaning this would only run once every 8 days I think. If this is expected to run on cloud, should verify behavior

wenxi-onyx (changes requested):

backend/ee/onyx/utils/license_expiry.py
nmgarza5: makes sense. i glossed over this

backend/ee/onyx/utils/license_notifications.py
nmgarza5: yeah definitely. adding to db/users.py

backend/ee/onyx/utils/license_notifications.py
nmgarza5: yeah.. i'll add filters

backend/ee/onyx/utils/license_notifications.py
nmgarza5: nah will remove

backend/ee/onyx/background/celery/tasks/beat_schedule.py
nmgarza5: no to both - licensing doesnt apply for the cloud

backend/ee/onyx/utils/license_expiry.py
```diff
@@ -0,0 +1,56 @@
...
+    return ExpiryWarningStage.NONE
+
+
+def get_grace_period_end(expires_at: datetime) -> datetime:
+    return expires_at + timedelta(days=LICENSE_GRACE_PERIOD_DAYS)
+
+
+def get_grace_days_remaining(expires_at: datetime) -> int:
+    grace_end = get_grace_period_end(expires_at)
+    seconds_left = (grace_end - datetime.now(timezone.utc)).total_seconds()
```
wenxi-onyx: nit: i dont think all of this finagling with seconds is necessary. You should be able use grace_end and datetime.now as day values only with no hour/minute/seconds values and the expiration happens on the day values.

backend/ee/onyx/utils/license_notifications.py
```diff
@@ -0,0 +1,152 @@
...
+            "the instance is gated. Renew now.",
+            f"Onyx license expired — {grace_days_remaining} grace days remaining",
+        )
+    raise ValueError(f"Unsupported stage for notification copy: {stage}")
+
+
+def _send_email_for_stage(
+    user_email: str, subject: str, heading: str, message: str
+) -> None:
+    if not EMAIL_CONFIGURED:
```
wenxi-onyx, line 73: The majority of self-hosted customers probably don't have an email send configured. 

Idea: would it be better to manage these emails via control plane and not have this in-app at all? We can send to license registrant email probably?

wenxi-onyx (approved):

backend/ee/onyx/utils/license_notifications.py
nmgarza5: eh yeah i agree with you but also... there is no guarantee they can communicate with the control plane (firewall/air gapped). I agree with your change but just calling out that the emails might still not get send due to this limitation. 

if no `EMAIL_CONFIGURED`, we will fallback to control plane and if that fails, we throw a warning in logs that email could not be sent

backend/ee/onyx/utils/license_expiry.py
nmgarza5: fair point. will simplify

backend/ee/onyx/utils/license_notifications.py
nmgarza5: actually this is a bigger change than i want to include in this pr. will follow up

PR #10782, opened 2026-05-02
fix: search ignores structured sources when multiple connectors are connected

### LINKED ISSUE #10781: Fix: Search ignores structured sources when multiple connectors are connected
## Problem

When an agent has both Jira and Confluence connectors (via document sets), queries about Jira always return Confluence results only.

Asking *"what's the status of ticket PROJ-123?"* returns Confluence pages instead of the Jira issue — even when the ticket clearly exists.

### Root cause:
The search tool runs a semantic vector search across all indexed documents with no source filtering. Confluence pages are long, text-rich documents that score higher in semantic similarity. Jira issues (short title + description + metadata) get buried and never surface in the top results.

**Jira alone works fine** — the problem only occurs when both connectors are present together.

## Solution

Expose an optional `source_type` parameter in the search tool definition so the LLM can explicitly filter by source when the intent is clear.

**Examples:**
- User asks *"what's the status of ticket PROJ-123?"* → LLM sets `source_type: ["jira"]`
- User asks *"find the onboarding docs"* → LLM sets `source_type: ["confluence"]`
- User asks something general → `source_type` is omitted, all sources are searched (no behavior change)

cc @yuhongsun96

mo7amed7assan1911:
## Problem
When an agent has both Jira and Confluence connectors (via document sets), queries about Jira always return Confluence results only.

**Example:** Asking *"what's the status of ticket PROJ-123?"* returns Confluence pages instead of the Jira issue — even when the ticket clearly exists.

**Root cause:** The search tool runs a semantic vector search across all indexed documents with no source filtering. Confluence pages are long, text-rich documents that score higher in semantic similarity. Jira issues (short title + description + metadata) get buried and never surface in the top results.

**Single connector works fine** — the problem only occurs when both are present together.

## Solution
Expose an optional `source_type` parameter in the search tool so the LLM can explicitly filter by source when the intent is clear.

- User asks *"what's the status of ticket PROJ-123?"* → LLM sets `source_type: ["jira"]`
- User asks *"find the onboarding docs"* → LLM sets `source_type: ["confluence"]`
- User asks something general → `source_type` omitted, all sources searched (no behavior change)

## Changes
- `backend/onyx/tools/tool_implementations/search/search_tool.py`
  - Added `SOURCE_FILTER_FIELD = "source_type"` constant
  - Exposed `source_type` as an optional array parameter in `tool_definition()`
  - In `run()`: if the LLM provides `source_type`, convert to `DocumentSource` enums and merge into active filters before search executes

## Testing
- [x] Agent with Jira + Confluence: ask about a ticket → only Jira results
- [x] Agent with Jira + Confluence: ask about a doc/page → only Confluence results
- [x] Agent with Jira + Confluence: ask a general question → both sources searched
- [x] Agent with single connector → no behavior change

Closes #10781

---
## Summary by cubic
Fixes multi-connector search so Jira- or Confluence-specific queries return results from the right source. Adds an optional `source_type` filter applied to both vector search and federated retrievers; omitted means no change.

- **Bug Fixes**
  - Exposes and validates `source_type` (array); coerces bare strings and rejects invalid types.
  - Uses call-local `effective_filters` and intersects with user-selected sources so the LLM can only narrow; prevents leakage between calls.
  - Applies `source_type` to federated retrievers so only matching sources run (e.g., skip Slack when filtering to Jira).

<sup>Written for commit 861de14c8a5ac800a5f95bd5b84efbf745daaf2a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/10782?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mo7amed7assan1911:
Hi Mr. @yuhongsun96, please, if you have time to take a moment to review my PR.

Thanks

PR #10794, opened 2026-05-04
fix(auth): refresh OAuth tokens for any OIDC provider, not just Google

beraterkanelcelik:
## Description

`oauth_refresher.refresh_oauth_token` previously only knew about Google because `REFRESH_ENDPOINTS` was a hardcoded dict keyed on provider name. For users signed in via the generic OIDC client (`httpx_oauth.clients.openid.OpenID`, which sets `oauth_name = "openid"` by default), refresh silently no-op'd with `"Refresh endpoint not configured for provider: openid"` — even when `offline_access` was in scope and a `refresh_token` was sitting on the account row.

Result: stored access tokens rotted at the IdP's lifetime (~1 h on Microsoft Entra default) and any flow that forwarded `user.oauth_accounts[0].access_token` (PT_OAUTH MCP tools, custom HTTP tools with bearer pass-through) broke until the user logged out and back in.

This change adds an async `_resolve_token_endpoint(provider)` helper that:
- Returns the static URL for hardcoded providers (Google preserved verbatim).
- For `"openid"`, reads `token_endpoint` from the configured `OPENID_CONFIG_URL` discovery document, caching the result after first successful fetch so the `.well-known/openid-configuration` doc is only retrieved once per process.
- Returns `None` for unknown providers (preserves the existing warn-and-skip behaviour).

`refresh_oauth_token` calls this helper in place of the dict lookup. The fix is intentionally generic — Microsoft Entra is the motivating provider, but Okta, Keycloak, Auth0, AAD-B2C, and any other OIDC-compliant IdP now work without provider-specific code.

Confirmed with @evan-onyx in Slack — "Definitely a bug, no particular reason we only have Google in the refresh endpoints."

### Out of scope (potential follow-up)

The web client at `web/src/hooks/useTokenRefresh.ts:31-38` deliberately skips `/api/auth/refresh` for `AuthType.OIDC` and `AuthType.SAML`, so today nothing on the request path proactively triggers `check_and_refresh_oauth_tokens` for OIDC users. Once this PR lands, a follow-up can wire the refresher into either the OIDC frontend tick or a server-side hook (e.g. `tool_constructor.py`'s PT_OAUTH branch) — depending on which call site you'd prefer. Happy to open that as a separate PR after this one is merged.

## How Has This Been Tested?

5 new unit tests in `backend/tests/unit/onyx/auth/test_oauth_refresher.py`:

- `test_resolve_token_endpoint_google_static` — Google bypasses discovery (preserves existing behaviour).
- `test_resolve_token_endpoint_openid_via_discovery` — OIDC reads `token_endpoint` from a mocked discovery document and caches it; second call does not re-fetch.
- `test_resolve_token_endpoint_openid_no_config_url` — without `OPENID_CONFIG_URL`, returns `None`.
- `test_resolve_token_endpoint_unknown_provider` — unknown provider returns `None` and triggers no network calls.
- `test_refresh_oauth_token_openid_provider` — end-to-end: `refresh_oauth_token` for an `oauth_name="openid"` account hits the resolved `token_endpoint` and writes the new tokens back.

All 12 tests in the file pass:
```
pytest backend/tests/unit/onyx/auth/test_oauth_refresher.py -xvs
============================== 12 passed in 0.24s ==============================
```

`uv run ty check` clean. `uv run ruff check --fix` clean.

## Additional Options

---
## Summary by cubic
Fixes OAuth token refresh for OIDC users by resolving the token endpoint from the discovery document, so tokens refresh for any OIDC provider. Also hardens discovery by catching invalid JSON and using a lock to avoid concurrent fetch storms.

- **Bug Fixes**
  - Added `_resolve_token_endpoint`: static URL for Google; for `openid`, read `token_endpoint` from `OPENID_CONFIG_URL`, cache it, and coalesce first-time fetches with an `asyncio.Lock`; warn/skip if unknown or not configured.
  - Updated `refresh_oauth_token` to use the resolved endpoint.
  - Hardened discovery: catch `httpx` errors and `ValueError` from invalid JSON; keep cache empty on failure for clean retries.
  - Tests updated: added cases for invalid discovery JSON and concurrent-fetch coalescing; coverage includes Google/static, OIDC discovery + cache, missing config, unknown provider, and end-to-end `openid` refresh.

<sup>Written for commit 61b92eacee7c235f663051b0255438bb30fb45ee. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/auth/oauth_refresher.py
+++ b/backend/onyx/auth/oauth_refresher.py
@@ -1,3 +1,4 @@
+import asyncio
 from datetime import datetime
 from datetime import timezone
 from typing import Any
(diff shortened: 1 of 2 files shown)


beraterkanelcelik:
Pushed `61b92eace` addressing both bot findings on the previous commit:

1. **`json.JSONDecodeError` escape (greptile + cubic, P2):** `response.json()` is now caught by `except (httpx.HTTPError, ValueError)`, so a non-JSON discovery body degrades to `None` with the correct log line and the cache stays empty for retry — instead of propagating up to `refresh_oauth_token`'s generic `except Exception`.

2. **Concurrent first-fetch coalescing (greptile, P2):** Added an `asyncio.Lock` (lazily instantiated so it binds to the running event loop, not at import time) with double-checked locking around the discovery fetch + cache write. A burst of expired tokens after a server restart now produces a single `.well-known/openid-configuration` request instead of N parallel ones.

Tests: 2 new — invalid-JSON degrades to None, and 3 concurrent callers trigger only one discovery request. All 14 tests in `tests/unit/onyx/auth/test_oauth_refresher.py` pass; `ruff` / `black` / `ty` clean.

@cubic-dev-ai please re-review.

beraterkanelcelik:
@greptile-apps please re-review — fixup `61b92eace` addresses both P2 findings from your earlier pass (broadened exception to catch `json.JSONDecodeError` and added a lazy `asyncio.Lock` with double-checked locking around the discovery fetch).

evan-onyx (approved):
looks good, I don't think auth/refresh is called for oidc auth atm but that can be addressed in a follow up if I'm right

PR #10806, opened 2026-05-04
fix(auth): refresh OIDC access tokens on every authenticated request (follow-up to #10794)

beraterkanelcelik:
## Description

Follow-up to #10794. That PR taught `oauth_refresher` how to refresh tokens for any OIDC provider via the discovery doc, but nothing on the request path actually invoked it for OIDC users — `web/src/hooks/useTokenRefresh.ts:31-38` returns early for `AuthType.OIDC` because `/api/auth/refresh` isn't registered for OIDC strategies. So the stored `oauth_accounts.access_token` rotted at the IdP's lifetime (~1 h on Microsoft Entra default) and any flow that forwarded `user.oauth_accounts[0].access_token` (PT_OAUTH MCP tools, custom HTTP tools with bearer pass-through) would 401 until the user signed out and back in.

This PR adds a small async helper `_maybe_refresh_oidc_tokens` and calls it from `optional_user` after the user is resolved. `check_and_refresh_oauth_tokens` already short-circuits when no `oauth_account` is within the 5-minute renewal buffer, so per-request overhead is one `expires_at` comparison in the common case and an HTTPS roundtrip to the IdP at most every ~55 min per user when the access_token is actually about to expire. The call is wrapped in `try/except` so a misconfigured IdP can never break authentication of an otherwise-valid request — failures log and the request proceeds with the stored token (same worst-case as before this PR; never worse).

Two additional commits address the bot review nits surfaced on the mirror of #10794:

- **TTL on the OIDC discovery cache** (greptile P2 from #10798): the cache now records `fetched_at` alongside the URL and re-fetches when an entry is older than `OIDC_DISCOVERY_CACHE_TTL_SECONDS` (default 1 h, env-overridable). Worst-case self-heal after a `token_endpoint` rotation is therefore one TTL window.
- **Lock reset in `test_resolve_token_endpoint_openid_via_discovery`** (greptile + cubic P2 from #10798): aligns this test with its siblings so `_OIDC_TOKEN_ENDPOINT_LOCK` is created in the current event loop, eliminating a "lock bound to a different event loop" flake under pytest loop reuse.

### Unit tests

17 tests across `backend/tests/unit/onyx/auth/test_oauth_refresher.py` (15) and `backend/tests/unit/onyx/auth/test_optional_user_oauth_refresh.py` (2 new):

```
17 passed in 0.49s
```

`uv run ty check` clean. `uv run ruff check` clean. The full `backend/tests/unit/onyx/auth/` suite (190 tests) passes with no regressions.

### Live verification on a real Onyx-on-Entra deployment

Deployed both files as a volume-mount overlay on a private production Onyx instance (Microsoft Entra OIDC, ~30 s recreate, fully reversible). Within seconds of the api_server coming back up, the api_server logs showed the refresh chain firing successfully for multiple distinct users whose `expires_at` was negative-by-millions-of-seconds, with the expected log shape:

```
INFO oauth_refresher.py:276  [API:xxxxxxxx] OAuth token for <user> is about to expire - refreshing
INFO oauth_refresher.py:182  [API:xxxxxxxx] Refreshing OAuth token for <user>'s openid account
INFO oauth_refresher.py:242  [API:xxxxxxxx] Successfully refreshed OAuth token for <user>
```

Across the first ~5 minutes after the recreate: **5+ distinct users refreshed end-to-end, zero refresh failures, zero auth-related tracebacks, zero nginx 502s**. The discovery document was fetched once (the lock + double-checked locking from #10794 worked as designed under real-world concurrent load) and every subsequent refresh used the cached endpoint.

DB confirmation: previously-stale `oauth_accounts` rows now have `expires_at ≈ now + 58 min` instead of being weeks in the past, and they continue to roll forward as users keep using the app.

## Additional Options

### DIFF
--- a/backend/onyx/auth/oauth_refresher.py
+++ b/backend/onyx/auth/oauth_refresher.py
@@ -1,6 +1,7 @@
 import asyncio
 import os
 import time
+import uuid
 from datetime import datetime
 from datetime import timezone
 from typing import Any
@@ -54,6 +55,34 @@ def _get_oidc_lock() -> asyncio.Lock:
     if _OIDC_TOKEN_ENDPOINT_LOCK is None:
         _OIDC_TOKEN_ENDPOINT_LOCK = asyncio.Lock()
     return _OIDC_TOKEN_ENDPOINT_LOCK
+
+
+# Per-user locks coalescing concurrent token-refresh attempts. Without this,
+# two requests for the same user near expiry could both POST a refresh, and
+# IdPs that rotate refresh tokens (e.g. Microsoft Entra) would invalidate one
+# of them with `400 invalid_grant`. The lock pairs with a re-read inside
+# `check_and_refresh_oauth_tokens` so the second coroutine skips the redundant
+# request entirely once the first has succeeded.
+_USER_REFRESH_LOCKS: Dict[uuid.UUID, asyncio.Lock] = {}
+_USER_REFRESH_LOCKS_GUARD: Optional[asyncio.Lock] = None
+
+
+def _get_user_refresh_locks_guard() -> asyncio.Lock:
+    """Lazy-init the meta-lock that protects the per-user lock dict itself."""
+    global _USER_REFRESH_LOCKS_GUARD
+    if _USER_REFRESH_LOCKS_GUARD is None:
+        _USER_REFRESH_LOCKS_GUARD = asyncio.Lock()
+    return _USER_REFRESH_LOCKS_GUARD
+
+
+async def _get_user_refresh_lock(user_id: uuid.UUID) -> asyncio.Lock:
+    """Get-or-create a per-user lock keyed by `user.id`."""
+    async with _get_user_refresh_locks_guard():
+        lock = _USER_REFRESH_LOCKS.get(user_id)
+        if lock is None:
+            lock = asyncio.Lock()
+            _USER_REFRESH_LOCKS[user_id] = lock
+        return lock
 
 
 def _cached_token_endpoint() -> Optional[str]:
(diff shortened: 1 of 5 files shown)


beraterkanelcelik:
@evan-onyx thanks for cutting #10794. End-to-end verified the chain on a private Microsoft Entra OIDC deployment via volume-mount overlay — refresh fires successfully across many concurrent users with zero failures and zero auth-related regressions. PR description has the redacted log shape + tests.

backend/onyx/auth/oauth_refresher.py
```diff
@@ -27,9 +29,18 @@
...
 # Token endpoint resolved from the configured OIDC discovery document.
-# Populated lazily on first successful fetch and re-used thereafter, so the
-# .well-known/openid-configuration document is only retrieved once per process.
-_OIDC_TOKEN_ENDPOINT_CACHE: Dict[str, str] = {}
+# Populated lazily on first successful fetch and re-used until the entry's
+# `fetched_at` exceeds OIDC_DISCOVERY_CACHE_TTL_SECONDS, at which point it is
+# re-fetched. The TTL guards against the IdP rotating its `token_endpoint`
+# without us noticing (rare for major IdPs but possible across tenant moves
+# or app-reg migrations).
+_OIDC_TOKEN_ENDPOINT_CACHE: Dict[str, Any] = {}
```
evan-onyx: prefer str | float instead of Any

beraterkanelcelik:
Pushed `d1c9c110a` addressing both P2s from the previous review:

1. Renamed `_maybe_refresh_oidc_tokens` → `_maybe_refresh_oauth_tokens` so the name reflects that it covers any provider (Google + OIDC), not just `openid`.
2. Added a per-user `asyncio.Lock` (mirroring the existing `_get_oidc_lock` shape) around the refresh inside `check_and_refresh_oauth_tokens`, plus a `db_session.refresh(oauth_account)` re-read inside the critical section so the second coroutine observes the post-refresh `expires_at` and skips its redundant POST. Detached `oauth_account` (pre-loaded by the caller) is handled by a try/except around the SQLAlchemy refresh so the lock degrades gracefully to old behaviour rather than blocking.

Tests: 1 new exercising the concurrent-coalescing invariant (2 simultaneous calls → 1 actual refresh). All 18 auth tests pass; ruff / black / ty clean.

@cubic-dev-ai @greptile-apps please re-review.

evan-onyx (approved):
seems good, just one nit

PR #10810, opened 2026-05-04
feat: paste tiles in contentEditable input bars

wenxi-onyx:
## Description

https://github.com/user-attachments/assets/ea0f7633-fcc0-4064-9c1e-699ebc5b1688

When paste-as-tile is enabled (user preference from #10803) and the user pastes text longer than 3 lines or 200 characters, the input bar collapses it into a compact "rich tile" instead of inserting it inline.

**Key features:**
- Tiles show a truncated preview, line/char count, and clipboard icon
- Click a tile to open an editable popover for the full text
- Arrow keys navigate into/out of tiles, Enter opens the popover
- Backspace/Delete on a highlighted tile removes it
- Ctrl+A highlights tiles within the selection range
- Copy/cut extracts the full tile text
- Disabled state prevents tile interactions
- Empty tile text in the popover auto-removes the tile

**Files added:**
- `web/src/lib/contentEditable.ts` — DOM utilities, text extraction, tile creation with `RichTileConfig` interface for future tile types
- `web/src/hooks/useContentEditable.ts` — Hook managing contentEditable state + tile interactions
- `web/src/sections/input/PasteTilePopover.tsx` — Popover for editing tile text, repositions on scroll/resize
- `web/src/app/css/content-editable.css` — Shared tile styles
- `web/src/lib/__tests__/contentEditable.test.ts` — 28 Jest unit tests for pure functions

**Files modified:**
- `web/src/sections/input/AppInputBar.tsx` — Consumer wiring
- `web/src/app/craft/components/InputBar.tsx` — Consumer wiring

**Architecture notes:**
- Data attributes use generic `data-rich-tile` + `data-tile-type="paste"` to support future tile types
- `createRichInputTileNode` accepts a config object (`{type, text, preview, meta}`) so future tile types reuse the same builder
- Tiles inside contentEditable are raw DOM nodes (React can't manage contentEditable children); shared CSS classes ensure visual consistency

Depends on #10803.

## How Has This Been Tested?

- 28 Jest unit tests for `shouldCreatePasteTile`, `getPasteTilePreview`, `getPasteTileMeta`, `getTextContent`
- Manual testing of tile creation, editing, keyboard navigation, copy/cut, disabled state
- E2E tests in follow-up PR

## Additional Options

---
## Summary by cubic
Adds “paste-as-tile” to contentEditable inputs: pasting >3 lines or >200 chars collapses into a compact, editable tile. Improves caret behavior around tiles and ensures both App and Craft input bars respect the `paste_as_tile` preference.

- **New Features**
  - Collapse long pastes into a tile with preview, line/char count, and clipboard icon.
  - Click or Enter opens a popover to edit; clearing text removes the tile and keeps the caret at the tile position.
  - Arrow keys select/exit tiles; Backspace/Delete removes; typing with a selected tile inserts after it; tiles highlight when selected.
  - Copy/cut returns the full tile text and modifier combos (Ctrl/Cmd) keep the tile selected; disabled inputs block tile interactions.

- **Refactors**
  - Moved tile logic to `web/src/lib/richInputTile.ts`; `web/src/lib/contentEditable.ts` now provides generic cursor/insertion/text extraction helpers (e.g., `insertNodeAtCursor`, `setCursorAfterNode`).
  - Shared styles in `web/src/app/css/content-editable.css` (imported in `globals.css`); `PasteTilePopover` renders inside `Disabled`.
  - Only register `selectionchange` when paste tiles are enabled; added 28 Jest tests for tile rules and text extraction.
  - Both App and Craft `InputBar` read `paste_as_tile` from user preferences.

<sup>Written for commit 1a016bbee91b4c667c75515bd3d084d46b03bd43. Summary will update on new commits.</sup>


web/src/hooks/useContentEditable.ts
```diff
@@ -68,6 +92,41 @@ export function useContentEditable({
...
+          sel &&
+          sel.rangeCount > 0 &&
+          !sel.isCollapsed &&
+          sel.getRangeAt(0).intersectsNode(tile)
+        ) {
+          htmlTile.classList.add("rich-input-tile-in-selection");
+        } else {
+          htmlTile.classList.remove("rich-input-tile-in-selection");
+        }
+      });
```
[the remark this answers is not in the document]
wenxi-onyx: fixed

web/src/sections/input/AppInputBar.tsx
```diff
@@ -151,17 +153,25 @@ const AppInputBar = React.memo(
...
+      handleTileMouseDown,
+      handleTileClick,
+      handleTileKeyDown,
+      tilePopover,
+      dismissTilePopover,
+      updateTileText,
     } = useContentEditable({
       initialContent: initialMessage,
       wrapperRef: inputWrapperRef,
+      pasteTilesEnabled: user?.preferences?.paste_as_tile ?? false,
```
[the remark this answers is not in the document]
wenxi-onyx: i dont think this is possible

web/src/hooks/useContentEditable.ts
```diff
@@ -172,6 +236,256 @@ export function useContentEditable({
...
+      if (!ref.current?.contains(range.commonAncestorContainer)) return;
+
+      event.preventDefault();
+      const fragment = range.cloneContents();
+      const temp = document.createElement("div");
+      temp.appendChild(fragment);
+      event.clipboardData.setData("text/plain", getTextContent(temp));
+
+      range.deleteContents();
+      syncFromDOM();
```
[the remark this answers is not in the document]
wenxi-onyx: i don't think this is possible

wenxi-onyx:
@greptileai re-review

web/src/lib/richInputTile.ts
```diff
@@ -0,0 +1,144 @@
...
+  return `${text.length} chars`;
+}
+
+// Path data mirrored from @opal/icons — keep in sync with:
+//   clipboard.tsx (SvgClipboard, viewBox 0 0 16 16)
+//   x.tsx (SvgX, viewBox 0 0 28 28)
+// We can't import the React icon components here because paste tiles are
+// created as raw DOM nodes inside a contentEditable div (React doesn't
+// manage the div's children), so we build the SVGs imperatively.
+const CLIPBOARD_PATH =
```
nmgarza5, line 33: Can we instead convert the svg on the fly to a raw DOM node? Extract the path and then call createSvgIcon

web/src/sections/input/PasteTilePopover.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { createPortal } from "react-dom";
+
+interface PasteTilePopoverProps {
+  text: string;
+  tileElement: HTMLElement;
+  onDismiss: () => void;
+  onTextChange: (newText: string) => void;
+}
+
+function PasteTilePopover({
```
nmgarza5, line 17: Can you document that this component cannot import any opal/refresh components? Or if it can, please do so lol 

And add a doc string for what this is for

nmgarza5 (approved):
A few comments. I looked into using some richtexteditor lib but didnt seem worth it. This approach seems to be the way

web/src/hooks/useContentEditable.ts
```diff
@@ -172,6 +240,268 @@ export function useContentEditable({
...
+      if (!newText.trim()) {
+        const marker = document.createTextNode("");
+        tile.replaceWith(marker);
+        selectedTileRef.current = null;
+        syncFromDOM();
+        resize();
+        setTilePopover(null);
+        ref.current?.focus();
+        setCursorAfterNode(marker);
+        ref.current?.normalize();
```
nmgarza5, line 353: **Cursor position bug on empty-tile path.** `setCursorAfterNode(marker)` runs before `normalize()`. `normalize()` removes the empty text node, invalidating the range — final cursor lands wherever the browser falls back to (often start of element).

Suggest capturing siblings before remove:

```ts
const next = tile.nextSibling;
const prev = tile.previousSibling;
tile.remove();
ref.current?.normalize();
if (next) setCursorBeforeNode(next);
else if (prev) setCursorAfterNode(prev);
else setCursorToEndUtil(ref.current);
```

web/src/sections/input/PasteTilePopover.tsx
```diff
@@ -0,0 +1,90 @@
...
+        onDismiss();
+      }
+    }
+    document.addEventListener("keydown", handleEscape);
+    return () => document.removeEventListener("keydown", handleEscape);
+  }, [onDismiss]);
+
+  useEffect(() => {
+    window.addEventListener("resize", updateRect);
+    document.addEventListener("scroll", updateRect, true);
```
nmgarza5, line 52: **No throttle on scroll/resize repositioning.** `setRect(getBoundingClientRect())` runs every event. Capture-mode catches every nested scroll container — scrolling the chat history triggers popover state updates per frame. Wrap in `requestAnimationFrame`:

```ts
const rafId = useRef<number | null>(null);
const updateRect = useCallback(() => {
  if (rafId.current !== null) return;
  rafId.current = requestAnimationFrame(() => {
    rafId.current = null;
    setRect(tileElement.getBoundingClientRect());
  });
}, [tileElement]);
```

web/src/hooks/useContentEditable.ts
```diff
@@ -172,6 +240,268 @@ export function useContentEditable({
...
+      const preview = tile.querySelector(".rich-input-tile-preview");
+      if (preview) {
+        preview.textContent = getPasteTilePreview(newText);
+      }
+      const meta = tile.querySelector(".rich-input-tile-meta");
+      if (meta) {
+        meta.textContent = getPasteTileMeta(newText);
+      }
+
+      syncFromDOM();
```
nmgarza5: **Per-keystroke full-tree walk.** Each character typed in the popover textarea calls `syncFromDOM` → `getTextContent(el)` walks the entire input subtree. The tile holds the full pasted text in `data-text` (can be 50KB+), so the walk is O(N) per keystroke. Defer until popover closes, or debounce ~100ms.

web/src/hooks/useContentEditable.ts
```diff
@@ -172,6 +240,268 @@ export function useContentEditable({
...
+
+  const handleCopy = useCallback(
+    (event: React.ClipboardEvent<HTMLDivElement>) => {
+      const sel = window.getSelection();
+      if (!sel || sel.rangeCount === 0 || sel.isCollapsed) return;
+
+      const range = sel.getRangeAt(0);
+      if (!ref.current?.contains(range.commonAncestorContainer)) return;
+
+      event.preventDefault();
```
nmgarza5: **Copy/cut blanket override drops native `text/html`.** Handlers always `preventDefault` even when the selection contains no tiles. Pasting into rich-text targets (Slack, Notion, Gmail) loses formatting the browser would have included. Same concern applies to `handleCut` below.

Suggest gating on actual tile presence:

```ts
const hasTile = temp.querySelector("[data-rich-tile]") !== null;
if (!hasTile) return;  // let browser default proceed
event.preventDefault();
event.clipboardData.setData("text/plain", getTextContent(temp));
```

web/src/lib/richInputTile.ts
wenxi-onyx: The icon files (clipboard.tsx, x.tsx) are React components that can't be introspected at runtime without rendering them. We'd need ReactDOM.renderToString or similar, which is overkill for two static path strings. Did you have something else in mind?

web/src/sections/input/PasteTilePopover.tsx
wenxi-onyx: Documented: uses raw HTML elements because it sits outside the opal/refresh component (the tile itself is a raw DOM node inside contentEditable)

web/src/lib/richInputTile.ts
nmgarza5: meh ok if we can't inspect at runtime then i guess we can leave it as is for now. i dont want to block.

wenxi-onyx:
@greptileai review

PR #10814, opened 2026-05-04
feat(license): control-plane email fallback for expiry notifications

nmgarza5:
## Description

Pairs with [danswer-control-plane#134](https://github.com/onyx-dot-app/danswer-control-plane/pull/134). Most self-hosted customers don't have local email transport configured, so the per-stage license-expiry email from the previous PR fired into a `logger.warning` and never reached the registrant.

This wires a cloud-plane fallback. When `EMAIL_CONFIGURED` is False, `notify_admins_for_stage` POSTs to `CLOUD_DATA_PLANE_URL/proxy/send-license-expiry-email` with the local license as Bearer auth. The cloud DP forwards to the control plane, which resolves the Stripe customer email and delivers via SendGrid.

- New cloud DP proxy endpoint `POST /proxy/send-license-expiry-email` (license-bearer auth) — forwards to CP `/send-license-expiry-email`.
- `notify_admins_for_stage`: when `EMAIL_CONFIGURED` is False, fires the proxy once per stage (recipient is per-tenant via Stripe, not per-admin).
- Renamed `_send_email_for_stage` to `_send_local_email` to make the two paths obvious.
- Air-gapped deployments hit the request-error path and log a warning; they continue to rely on the in-app banner.
- Tests cover the fallback path, the per-fire fan-out semantic, and the missing-license short-circuit.

## Additional Options

### DIFF
--- a/backend/ee/onyx/server/auth_check.py
+++ b/backend/ee/onyx/server/auth_check.py
@@ -24,6 +24,7 @@ EE_PUBLIC_ENDPOINT_SPECS = PUBLIC_ENDPOINT_SPECS + [
     ("/proxy/billing-information", {"GET"}),
     ("/proxy/license/{tenant_id}", {"GET"}),
     ("/proxy/seats/update", {"POST"}),
+    ("/proxy/send-license-expiry-email", {"POST"}),
 ]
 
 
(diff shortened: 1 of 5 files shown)


nmgarza5:
clsoing. will send these emails directly from control plane

PR #10826, opened 2026-05-04
fix(targeted-reindex): keep staged files of concurrent in-progress attempts

nmgarza5:
## Description

\`reap_prior_attempt_staged_files\` runs at the start of every docfetching attempt and deletes orphan staged files left behind by previous attempts on the same cc_pair. Without a status filter, a concurrent in-progress attempt (e.g. a targeted reindex on the same cc_pair, landing soon) would have its in-flight binaries wiped from under it.

The fix excludes files whose owning attempt is in a non-terminal state (\`NOT_STARTED\` or \`IN_PROGRESS\`). The subquery is scoped to the current cc_pair so it doesn't scan the whole \`index_attempt\` table. The exclusion is expressed as \`NOT IN (non-terminal attempt ids)\`, which keeps orphans whose owning attempt no longer exists in the DB at all (deleted by retention, synthetic ids in test fixtures, etc.) reapable — nothing is going to consume them.

Part of the targeted-reindex hardening called out in the design doc under "Risks worth calling out — Staging-files sweep."

## How Has This Been Tested?

\`backend/tests/external_dependency_unit/file_store/test_staging_concurrent_attempt_skip.py\` covers four cases against real Postgres + file store:
- Concurrent \`IN_PROGRESS\` attempt's file is kept; sibling terminal-old file is reaped
- \`NOT_STARTED\` attempt's file is kept (worker may have crashed before status advanced)
- Orphan with no owning \`IndexAttempt\` row at all is reaped
- Current attempt's own file is kept (excluded by \`current_attempt_id\`)

## Additional Options

### DIFF
--- a/backend/onyx/db/file_record.py
+++ b/backend/onyx/db/file_record.py
@@ -1,12 +1,16 @@
 from sqlalchemy import and_
+from sqlalchemy import cast
 from sqlalchemy import select
+from sqlalchemy import String
 from sqlalchemy.dialects.postgresql import insert
 from sqlalchemy.orm import Session
 
 from onyx.background.task_utils import QUERY_REPORT_NAME_PREFIX
 from onyx.configs.constants import FileOrigin
 from onyx.configs.constants import FileType
+from onyx.db.enums import IndexingStatus
 from onyx.db.models import FileRecord
+from onyx.db.models import IndexAttempt
 
 
 def get_query_history_export_files(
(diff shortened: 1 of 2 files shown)


nmgarza5:
@greptileai

nmgarza5:
@greptileai retrigger

evan-onyx (approved):

nmgarza5:
@greptileai retrigger

PR #10828, opened 2026-05-04
feat(targeted-reindex): API endpoints + DB writes + stub celery task

nmgarza5:
## Description

Adds the request-receiving half of the targeted-reindex flow.

`POST /manage/admin/indexing/targeted-reindex`
- Accepts `error_ids` (failure-driven retry) and/or arbitrary `(cc_pair_id, document_id)` pairs.
- Validates, dedups, caps at 100, rejects entity-level errors and already-resolved errors.
- Writes the job row + per-doc target rows + one synthetic `IndexAttempt` per `(cc_pair_id, search_settings_id)` the targets span (so PRESENT+FUTURE indexes both get the reindex during a model swap).
- Enqueues the celery task with the pre-allocated UUID at `OnyxCeleryPriority.HIGHEST` on the existing `PRIMARY` queue. The trigger task itself owns the indexing scheduler; per-cc-pair fan-out hands off to the existing docfetching/docprocessing queues from inside the task body.
- Propagates `tenant_id` through `kwargs` so `TenantAwareTask` runs against the right schema.

`GET /manage/admin/indexing/targeted-reindex/{job_id}`
- Returns current status + counts for the FE to poll.

The celery task is a **stub** for now: transitions the job to `IN_PROGRESS` then `SUCCESS` without doing real reindex work. The real lifecycle + resolution tracking land in #10829; connector + pipeline plumbing lands in a follow-up. The stub exists so the task name resolves at the worker and the API path's `apply_async` doesn't fail with "unknown task."

The flow bypasses `try_create_index_attempt` — full crawls are allowed to overlap with retries by design (per-doc row-locks handle write conflicts; consumer queries already filter retry attempts via `ignore_targeted_reindex=True`).

Builds on the merged schema PR (#10747) and consumer-filters PR (#10764).

## How Has This Been Tested?

7 tests in \`backend/tests/external_dependency_unit/db/test_targeted_reindex_create.py\` against real Postgres:
- Row creation: job + targets + synthetic IndexAttempt(s)
- Dedup of duplicate \`(cc_pair_id, document_id)\` pairs
- Max-targets cap enforcement
- Unknown cc_pair rejection
- Empty-input rejection
- Error-id resolution (already-resolved + entity-level rows skipped)
- \`source_error_id\` propagation to target rows

All green.

## Additional Options

### DIFF
--- a/backend/onyx/db/targeted_reindex.py
+++ b/backend/onyx/db/targeted_reindex.py
@@ -140,6 +140,7 @@ def create_targeted_reindex_job(
     db_session: Session,
     requested_by_user_id: UUID | None,
     targets: Sequence[TargetSpec],
+    upstream_skipped_count: int = 0,
 ) -> CreateTargetedReindexJobResult:
     """Persist a targeted reindex request.
 
@@ -147,6 +148,12 @@ def create_targeted_reindex_job(
     `(cc_pair_id, search_settings_id)` tuple. Pre-allocates the celery
     task UUID so the orphan-detector can clean up if `apply_async` fails
     after this returns.
+
+    `upstream_skipped_count` is added to the dedup-skipped count and
+    persisted on the job row so the GET status endpoint can return the
+    full at-create-time skip count (e.g. error_ids that resolved to
+    already-resolved or entity-level rows in the API layer). The task
+    later folds in any runtime skips.
 
     The caller (API endpoint) is responsible for enqueueing the celery
     task with the returned `celery_task_id` after this commits.
@@ -174,14 +181,6 @@ def create_targeted_reindex_job(
 
     celery_task_id = str(uuid4())
 
-    job = TargetedReindexJob(
-        requested_by_user_id=requested_by_user_id,
-        celery_task_id=celery_task_id,
-        status=IndexingStatus.NOT_STARTED,
-    )
-    db_session.add(job)
-    db_session.flush()
-
     # Dedup at the (cc_pair_id, document_id) level — composite PK on the
     # target table would catch this anyway, but better to error early.
     seen: set[tuple[int, str]] = set()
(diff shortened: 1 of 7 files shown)


nmgarza5:
@greptileai retrigger

nmgarza5:
@greptileai retrigger

nmgarza5:
@greptileai retrigger

nmgarza5:
@greptileai retrigger

nmgarza5:
@greptileai retrigger

nmgarza5:
@greptileai retrigger

backend/onyx/db/targeted_reindex.py
```diff
@@ -0,0 +1,261 @@
...
+    `IndexAttemptError`; NULL when the request came in as an arbitrary
+    `(cc_pair_id, document_id)` pair.
+    """
+
+    cc_pair_id: int
+    document_id: str
+    source_error_id: int | None = None
+
+
+class CreateTargetedReindexJobResult(BaseModel):
```
evan-onyx, line 56: nit: brief comment explaining how this is used would help me

backend/onyx/db/targeted_reindex.py
```diff
@@ -0,0 +1,261 @@
...
+                source_error_id=err.id,
+            )
+        )
+    # Errors not found in DB (caller passed invalid IDs) are also skipped.
+    found_ids = {err.id for err in rows}
+    skipped += len(set(error_ids) - found_ids)
+    return targets, skipped
+
+
+def _group_targets_by_search_settings(
```
evan-onyx: this feels somewhat unnecessary; it's just retrieving the search settings (we should already have a db func for doing this that we use in the main indexing flow) and making an explicit map from each cc pair to all the relevant search settings. imo we should be able to just call whatever search settings getter function is already being used for normal indexing and then double for loop there

backend/onyx/db/targeted_reindex.py
```diff
@@ -0,0 +1,261 @@
...
+    }
+    missing = cc_pair_ids - existing_pairs
+    if missing:
+        raise ValueError("unknown cc_pair_ids: %s" % sorted(missing))
+
+    celery_task_id = str(uuid4())
+
+    # Dedup at the (cc_pair_id, document_id) level — composite PK on the
+    # target table would catch this anyway, but better to error early.
+    seen: set[tuple[int, str]] = set()
```
evan-onyx:
```
seen = {(t.cc_pair_id, t.document_id) for t in targets}
deduped = list(seen)
```

evan-onyx (approved):

nmgarza5:
@greptileai retrigger

backend/onyx/db/targeted_reindex.py
nmgarza5: unable to do this actually. I added a comment explaining in more detail why but basically if the request includes target with `source_error_id`, we always need to use that so we update the errored index attempt and mark the document as resolved. 

this is relevant once we have have UI for viewing all docs per connector

PR #10846, opened 2026-05-05
refactor(persona): replace llm override strings with model_configuration FK

raunakab:
## Description

Drops the deprecated `llm_model_provider_override` / `llm_model_version_override` string pair
from the `persona` table and replaces them with the pre-existing `default_model_configuration_id`
FK (`→ model_configuration.id`).

**Why:** The string pair was a weak reference with no FK guarantees — renaming or deleting a
provider silently orphaned persona overrides. `default_model_configuration_id` already encodes
both the provider *and* the model in a single FK with `ON DELETE SET NULL`, making it the correct
single source of truth.

**Migration (`a5370af8f8a0`):**
- Backfills `default_model_configuration_id` from the string pair via `UPDATE…FROM` joining on
  `llm_provider.name` + `model_configuration.name`.
- Rows where the string pair doesn't resolve to a valid `(provider, model)` combination are left
  as `NULL` (the override was already broken; falling back to the system default is correct).
- Drops both legacy string columns.

**Code changes:**
- `factory.py` (`get_llm_for_persona`): reads `default_model_configuration_id` exclusively;
  the legacy string-lookup branch is removed.
- `db/llm.py` (`get_personas_using_provider`): now queries via `JOIN model_configuration`
  instead of matching on the provider name string.
- `db/persona.py`, `server/features/persona/models.py`, API endpoints, seeding, and all tests
  updated to use `default_model_configuration_id`.

## Additional Options

---
## Summary by cubic
Replaced persona LLM override strings with a `default_model_configuration_id` FK to `model_configuration`. Tightened access checks and fallbacks, and fixed a bug that ignored `temperature=0.0`.

- **Refactors**
  - Replaced `llm_model_provider_override`/`llm_model_version_override` with `default_model_configuration_id` across DB helpers, factory, APIs, seeding, and tests; provider lookups now join via `model_configuration` and `get_personas_using_provider(provider_id)`; merged `remove_llm_provider` variants into one that clears the FK on delete.
  - `get_llm_for_persona` resolves via the FK with runtime overrides (reuses the persona’s model name when only a provider is overridden), honors `temperature=0.0`, and falls back to the global default when the FK is missing/invalid; persona GET clears the FK when the referenced model is inaccessible; added `get_valid_model_configuration_ids_for_persona` for unambiguous ID-based access checks.

- **Migration**
  - Backfills `default_model_configuration_id` by joining on provider and model names; leaves unresolved rows as NULL to use system defaults; legacy string columns are retained for now.
  - Downgrade now only clears backfilled FKs (scoped to rows with both legacy columns set).

<sup>Written for commit dc9c2129335730eaeda025dd4be7354fe81ea92a. Summary will update on new commits.</sup>


backend/tests/integration/common_utils/managers/llm_provider.py
```diff
@@ -73,6 +73,11 @@ def create(
             personas=response_data.get("personas", []),
             api_base=response_data["api_base"],
             api_version=response_data["api_version"],
+            model_configuration_ids=[
```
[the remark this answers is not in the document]
raunakab: The IDs are present in the response. `ModelConfigurationView.from_model` always sets `id=model_configuration_model.id` (both code paths in the method). `upsert_llm_provider` calls `db_session.flush()` before `db_session.refresh(existing_llm_provider)`, so newly inserted configurations have their auto-increment PKs assigned before `LLMProviderView.from_model` is called. The `if mc.get("id") is not None` guard is just defensive — not evidence that IDs are missing.

backend/alembic/versions/a5370af8f8a0_persona_default_model_fk.py
```diff
@@ -0,0 +1,96 @@
...
+            == model_configuration_table.c.name,
+            model_configuration_table.c.llm_provider_id == llm_provider_table.c.id,
+            persona_table.c.llm_model_provider_override.is_not(None),
+            persona_table.c.llm_model_version_override.is_not(None),
+            persona_table.c.default_model_configuration_id.is_(None),
+        )
+    )
+
+    # 2. Drop the now-redundant string columns.
+    op.drop_column("persona", "llm_model_provider_override")
```
Danelegend: I would recommend a 2-stage approach in-case something goes wrong.

PR 1 / this one: Do the migration / port everything over
PR 2: Drop the unused column and all the associated data

backend/onyx/db/llm.py
```diff
@@ -595,18 +598,18 @@ def remove_llm_provider(db_session: Session, provider_id: int) -> None:
     if not provider:
         raise ValueError("LLM Provider not found")
 
-    # Clear the provider override from any personas using it
-    # This causes them to fall back to the default provider
-    personas_using_provider = get_personas_using_provider(db_session, provider.name)
-    for persona in personas_using_provider:
-        persona.llm_model_provider_override = None
+    # Clear the model config FK from any personas using this provider so they
```
Danelegend: ai comment bloat :(

backend/onyx/db/llm.py
Danelegend: Hmm, might be worth merging the duplicate logic of `remove_llm_provider` and `remove_llm_provider__no_commit` into a single function

backend/onyx/llm/factory.py
```diff
@@ -95,19 +96,61 @@ def get_llm_for_persona(
     model_version_override = llm_override.model_version if llm_override else None
     temperature_override = llm_override.temperature if llm_override else None
 
-    provider_name = provider_name_override or persona.llm_model_provider_override
-    if not provider_name:
+    # Resolve the provider and model.
```
Danelegend: Instead of function doc-string would be great:
"""Get the appropriate LLM for a persona, with the following priority:
    1. LLM override (provider + model version)
    2. Persona's model configuration override
    3. Default LLM
    """

backend/onyx/llm/factory.py
```diff
@@ -95,19 +96,61 @@ def get_llm_for_persona(
...
+    if not provider_name_override and not persona.default_model_configuration_id:
         return get_default_llm(
             temperature=temperature_override or GEN_AI_TEMPERATURE,
             additional_headers=additional_headers,
         )
 
     with get_session_with_current_tenant() as db_session:
-        provider_model = fetch_existing_llm_provider(provider_name, db_session)
+        model: str | None
+        if provider_name_override:
```
Danelegend: I think this can be cleaned up a bit into a helper function.
I took a crack at this change a bit ago but forgot to merge 😨  (and was a bit too scared to lol).

This is roughly what I thought up for this path: https://github.com/onyx-dot-app/onyx/pull/7907/changes#diff-449063813b94cd14fb7e94a9ecb34a5ced05cada0f57c68d2bfa2b63372a79ecR88

Be mindful that the flow might've changed a bit since I did this lol

backend/onyx/server/features/persona/api.py
```diff
@@ -531,15 +531,14 @@ def get_persona(
...
-                valid_models[0] if valid_models else None
-            )
+    # Validate and clear the model override if the referenced model is no longer
+    # accessible to this persona (e.g. provider was restricted after the persona was saved).
+    if persona.default_model_configuration_id:
+        valid_ids = get_valid_model_configuration_ids_for_persona(
+            persona_id, user, db_session
+        )
+        if persona.default_model_configuration_id not in valid_ids:
+            persona.default_model_configuration_id = None
```
Danelegend, line 541: This makes more sense than updating to a random model and seems more correct, but might technically be a regression :P

I'd say keep it this way though

backend/onyx/server/kg/api.py
```diff
@@ -140,8 +140,6 @@ def enable_or_disable_kg(
         is_public=False,
         document_set_ids=[],
         tool_ids=[search_tool.id, kg_tool.id],
```
Danelegend, line 142: Explicit `default_model_configuration_id=None,` pls

backend/onyx/server/manage/llm/api.py
```diff
@@ -770,6 +771,37 @@ def get_valid_model_names_for_persona(
     return valid_models
 
 
+def get_valid_model_configuration_ids_for_persona(
+    persona_id: int,
```
Danelegend: Can we pass Persona instead? Might not be possible because of `fetch_persona_with_groups`, but if we can, it would save a db lookup

backend/onyx/server/manage/llm/api.py
```diff
@@ -825,45 +857,25 @@ def list_llm_providers_for_persona(
...
+    # available, falling back to the global defaults.
     default_text = DefaultModel.from_model_config(default_text_model)
     default_vision = DefaultModel.from_model_config(default_vision_model)
 
-    if persona_default_provider:
-        provider = fetch_existing_llm_provider(persona_default_provider, db_session)
-        if provider and can_user_access_llm_provider(
-            provider, user_group_ids, persona, is_admin=is_admin
+    if persona.default_model_configuration_id:
+        model_config = db_session.get(
```
Danelegend: Would prefer to make a call to a db helper function so that we maintain the abstraction

backend/tests/external_dependency_unit/answer/test_stream_chat_message_objects.py
```diff
@@ -73,8 +73,6 @@ def mock_post(
         user=None,  # System persona
         name=f"Test Persona {uuid.uuid4()}",
         description="Test persona with no tools for web search test",
-        llm_model_provider_override=None,
```
Danelegend, line 76: `default_model_configuration_id=None,` pls

Danelegend (approved):
Nice. I'm a bit more opinionated on this one cause I gave it a shot lol.

Please make sure the migration occurs nicely.
I'd also recommend writing a few external-dep tests in a PR before this going through various flows, then making sure that these changes abide by those tests to check for regressions

backend/onyx/server/features/persona/api.py
raunakab: Ya, I think there is a possibility of this "regressing". However, if there is no valid ID in the first place, then the `llm_model_override` is already invalid.

backend/alembic/versions/a5370af8f8a0_persona_default_model_fk.py
raunakab: Done in 8b60377. Removed the op.drop_column calls from upgrade(). The string columns are kept; downgrade just clears the backfilled FKs.

backend/onyx/db/llm.py
raunakab: Fixed in 8b60377. Removed the multi-line comment.

backend/onyx/db/llm.py
raunakab: Done in 8b60377. Merged into a single remove_llm_provider(db_session, provider_id, commit: bool = True). Updated callers in image_generation/api.py to pass commit=False.

backend/onyx/llm/factory.py
raunakab: Done in 8b60377. Replaced the inline comment with the docstring you suggested.

backend/onyx/llm/factory.py
raunakab: Done in 8b60377. Extracted _resolve_provider_and_model(persona, provider_name_override, model_version_override, db_session) helper; also added a fetch_model_configuration_by_id DB helper used inside it.

backend/onyx/server/kg/api.py
raunakab: Added in 8b60377.

backend/onyx/server/manage/llm/api.py
raunakab: Done in 8b60377. Changed signature to (persona: Persona, user, db_session) and updated the caller in persona/api.py to pass the already-loaded object directly.

backend/onyx/server/manage/llm/api.py
raunakab: Done in 8b60377. Added fetch_model_configuration_by_id to db/llm.py and replaced the inline db_session.get call with it.

backend/tests/external_dependency_unit/answer/test_stream_chat_message_objects.py
raunakab: Added in 8b60377.

backend/alembic/versions/a5370af8f8a0_persona_default_model_fk.py
```diff
@@ -0,0 +1,67 @@
...
+            persona_table.c.llm_model_provider_override.is_not(None),
+            persona_table.c.llm_model_version_override.is_not(None),
+            persona_table.c.default_model_configuration_id.is_(None),
+        )
+    )
+
+
+def downgrade() -> None:
+    # Clear the backfilled FK values; the string columns still hold the original data.
+    op.execute(sa.update(persona_table).values(default_model_configuration_id=None))
```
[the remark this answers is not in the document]
raunakab: Fixed in dc9c212. Scoped the downgrade to rows where both string columns are non-null — those are the only rows upgrade() could have touched.

PR #10847, opened 2026-05-06
feat(llm): make provider display name optional and allow renaming

raunakab:
## Description

Makes `llm_provider.name` a nullable, non-unique cosmetic label. The integer PK (`id`) is the stable identity; name is now purely for display.

**Migration (`c1d2e3f4a5b6`):**
- Drops the `llm_provider_name_key` unique constraint.
- Makes `name` nullable.
- Downgrade fills NULLs with `<provider>-<id>` before re-adding the constraint.

**Code changes:**
- `LLMProvider.name` DB column: `Mapped[str | None]`.
- `LLMProvider` / `LLMProviderDescriptor` Pydantic models: `name: str | None = None`.
- `upsert_llm_provider` (`db/llm.py`): removed rename guard and duplicate-name check — identity is now the integer `id`.
- `put_llm_provider` API (`server/manage/llm/api.py`): same blocks removed.
- All call-sites that required `provider.name: str` now fall back to `provider.provider` (the type slug e.g. `"openai"`) when `name` is `None`.

## Additional Options

---
## Summary by cubic
Make LLM provider display names optional and non-unique, and allow renaming. Providers are identified by integer `id`; when `name` is missing we fall back to the `provider` slug.

- **Migration**
  - Adds `f0db5f1c6370`: drops `llm_provider_name_key`; makes `llm_provider.name` nullable. Downgrade fills NULLs with `__unnamed_<id>` before restoring NOT NULL + UNIQUE.

- **Refactors**
  - Models/API: `LLMProvider.name` → `str | None`; `LLMProviderDescriptor.name` optional; expose `ModelConfigurationView.id`; rename `LLMCost.provider` → `provider_name` (frontend `LLMContextualCost` updated).
  - Upsert/API: remove rename/duplicate-name checks; update `name` only when present using Pydantic `model_fields_set` (absent = no change, `null` = clear, string = set).
  - Lookups: add `fetch_existing_llm_provider_by_name_and_type` (returns None if ambiguous) and `fetch_existing_llm_provider_by_type_nameless` (warns if multiple).
  - Seeding/provisioning: seeding skips nameless lookups; provisioning matches by `(name, type)` or falls back to nameless-by-type.
  - Fallbacks/logging: display/log use `name or provider`; max-token lookup uses provider slug; image gen uses `name or ""`; auto-update maps results by `provider.id` and logs `name or provider`.

<sup>Written for commit ccb90af421d1ae8d557680f17c136f89e789694e. Summary will update on new commits.</sup>


raunakab:
Fixed in 44fbfe5ded — logger now uses `provider.name or provider.provider` and seeding falls back to `fetch_existing_llm_provider_by_type_nameless` when name is None.

backend/ee/onyx/server/seeding.py
```diff
@@ -121,7 +122,16 @@ def _seed_llms(
...
     for request in llm_upsert_requests:
-        existing = fetch_existing_llm_provider(name=request.name, db_session=db_session)
+        if request.name:
+            existing = fetch_existing_llm_provider_by_name_and_type(
+                name=request.name,
+                provider_type=request.provider,
+                db_session=db_session,
+            )
+        else:
+            existing = fetch_existing_llm_provider_by_type_nameless(
```
Danelegend: I think there is potential for risk here.

Imagine the case:
 -> No LLMs Seeded
 -> Create LLM w/ No Name
 -> Seed LLMs (override occurs)

In particular, this could occur in cloud I think. I think we should skip over nameless requests

backend/onyx/db/llm.py
```diff
@@ -524,6 +519,49 @@ def fetch_existing_llm_provider_by_id(
...
 
+def fetch_existing_llm_provider_by_name_and_type(
+    name: str, provider_type: str, db_session: Session
+) -> LLMProviderModel | None:
+    """Return a provider matching both display name and provider type.
+
+    More specific than fetch_existing_llm_provider (name only) — avoids matching
+    the wrong record now that names are non-unique.
+    """
+    return db_session.scalar(
```
Danelegend: More than one can be returned I believe.
We should probably be specific about returning the first one (and logging a warning if there are more than one). Or return all of them and let the consumer decide to take the first and log warning

backend/ee/onyx/server/seeding.py
```diff
@@ -121,7 +122,16 @@ def _seed_llms(
 
     logger.notice("Seeding LLMs")
     for request in llm_upsert_requests:
-        existing = fetch_existing_llm_provider(name=request.name, db_session=db_session)
+        if request.name:
+            existing = fetch_existing_llm_provider_by_name_and_type(
+                name=request.name,
+                provider_type=request.provider,
+                db_session=db_session,
```
Danelegend: One potential error case to consider:
We want to update cloud keys so we seed the llms.
Someone sets the name and provider to the same as ours.
We end up selecting their LLM and setting the credentials.

Idk if this flow is possible or not, but something worth checking / considering.

backend/onyx/db/models.py
```diff
@@ -3152,7 +3152,7 @@ class LLMProvider(Base):
     __tablename__ = "llm_provider"
 
     id: Mapped[int] = mapped_column(Integer, primary_key=True)
-    name: Mapped[str] = mapped_column(String, unique=True)
+    name: Mapped[str | None] = mapped_column(String, nullable=True)
```
Danelegend, line 3155: Thoughts on empty string instead of nullable?

backend/onyx/llm/well_known_providers/auto_update_service.py
```diff
@@ -132,9 +132,11 @@ def sync_llm_models_from_github(
         )
 
         if changes > 0:
-            results[provider.name] = changes
+            results[f"{provider.provider}-{provider.id}"] = changes
```
Danelegend: Why not just provider.id?

backend/onyx/server/manage/llm/api.py
```diff
@@ -923,7 +906,7 @@ def get_provider_contextual_cost(
             cost = get_llm_contextual_cost(llm)
             costs.append(
                 LLMCost(
-                    provider=provider.name,
+                    provider=provider.name or provider.provider,
```
Danelegend: Worth changing this attribute to provider_name? When I saw provider, I immediately thought it meant OpenAi, Anthropic, ... but it seems like it might actually just be the name

Danelegend (approved):

backend/ee/onyx/server/seeding.py
raunakab: Good catch. Fixed by skipping the lookup entirely for nameless requests — if the seed config has no name, we create fresh rather than risking a match against a user-created provider.

backend/onyx/db/llm.py
raunakab: Fixed. Now fetches all matches, logs a warning with count+id if multiple found, and returns the first.

backend/ee/onyx/server/seeding.py
raunakab: Fixed. `fetch_existing_llm_provider_by_name_and_type` now returns `None` when multiple rows match the same name+type rather than silently returning the first. The seeder then creates a new provider rather than overwriting an ambiguous match.

backend/onyx/db/models.py
raunakab: Nullable is the right choice here — it clearly distinguishes "no name set" from "name is empty string". An empty string would also require extra checks throughout (e.g. treating `""` as absent) and would leak into the UI as blank display names rather than triggering the fallback. `NULL` is the standard SQL signal for "not present".

backend/onyx/llm/well_known_providers/auto_update_service.py
raunakab: Good call. Simplified to `str(provider.id)` since `id` is already the unique identifier.

backend/onyx/server/manage/llm/api.py
raunakab: Done — renamed to `provider_name` in the model, the API, and the frontend interface.

PR #10851, opened 2026-05-06
fix(ui): correct indexing errors modal pagination

Bo-Onyx:
## Description

The Indexing Errors modal was only showing ~10-20 errors regardless of how many existed. The bug was that `total_items` was set to `indexAttemptErrorsPage.length` (the current page slice, ≤10) instead of the real total from the server. The modal then did client-side pagination over those ≤10 items, so users only ever saw 1–2 pages.

Fix: thread `totalItems`, `currentPage`, and `goToPage` from `usePaginatedFetch` into the modal so pagination is driven by server-side data. The modal's internal ResizeObserver + local page state were removed since page management now happens at the parent level.

https://linear.app/onyx-app/issue/ENG-4105/google-drive-for-savvy-wealth

## How Has This Been Tested?

Manually verified against a connector with more than 10 errors.

Before the change
Screenshot 2026-05-05 at 6 21 50 PM

After the change
Screenshot 2026-05-05 at 6 46 32 PM

## Additional Options

---
## Summary by cubic
Fixes the Indexing Errors modal so it paginates across all server-side errors and stops reading/writing the `?page` URL param. Supports Linear ENG-4105.

- **Bug Fixes**
  - Drive pagination from `usePaginatedFetch` (`totalPages`, `totalItems`, `currentPage`, `goToPage`); remove local pagination and the `ITEMS_PER_PAGE` constant.
  - Restore dynamic page sizing: measure table height and call `onPageSizeChange` so the hook fetches the right page size.
  - Add and respect `disableUrlSync` in `usePaginatedFetch` (including on init) so modal page changes don’t read/write the index-attempts table `?page`.

<sup>Written for commit 7660a5ee80aa6b3206d654d8095a5c5264601eab. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/app/admin/connector/[ccPairId]/IndexAttemptErrorsModal.tsx
+++ b/web/src/app/admin/connector/[ccPairId]/IndexAttemptErrorsModal.tsx
@@ -15,13 +15,11 @@ import { PageSelector } from "@/components/PageSelector";
 import { useMemo } from "react";
 import { SvgAlertTriangle } from "@opal/icons";
 
-const ITEMS_PER_PAGE = 10;
-
 export interface IndexAttemptErrorsModalProps {
   errors: {
     items: IndexAttemptError[];
-    total_items: number;
   };
+  totalPages: number;
   currentPage: number;
   onPageChange: (page: number) => void;
   onClose: () => void;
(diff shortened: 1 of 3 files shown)


web/src/app/admin/connector/[ccPairId]/IndexAttemptErrorsModal.tsx
```diff
@@ -125,10 +76,7 @@ export default function IndexAttemptErrorsModal({
             </div>
           )}
 
-          <div
-            ref={tableContainerRef}
-            className="flex-1 w-full overflow-hidden min-h-0"
-          >
+          <div className="flex-1 w-full overflow-hidden min-h-0">
```
Comment (evan-onyx) [inline web/src/app/admin/connector/[ccPairId]/IndexAttemptErrorsModal.tsx]: since we're rewriting this anyways we should probably use opal components and layouts from generalLayouts

evan-onyx (commented):
LGTM but you should probably get a FE person's approval

raunakab (approved):

PR #10856, opened 2026-05-06
fix: handle already-deleted batch files in cleanup_all_batches

onematchfox:
## Summary

`cleanup_all_batches` iterates over batch files and calls `delete_batch_by_name` for each. If a file is already gone — either from a partial cleanup that previously ran, or because it was never written due to an earlier failure in the same indexing attempt — `delete_file` raises a `RuntimeError` (the default `error_on_missing=True`), aborting the entire cleanup loop.

The fix passes `error_on_missing=False` so cleanup is idempotent and tolerates files that no longer exist.

## Reproducing

Observed during Jira connector indexing: a batch file was deleted or never written, then `cleanup_all_batches` raised:

```
RuntimeError: File by id iab/3/116/14353.json does not exist or was deleted
```

## Fix

One-line change in `delete_batch_by_name`:

```python
# Before
self.file_store.delete_file(batch_file_name)

# After
self.file_store.delete_file(batch_file_name, error_on_missing=False)
```

`FileStore.delete_file` already accepts `error_on_missing: bool = True` — this just opts in to the tolerant behaviour that cleanup should have.

---
## Summary by cubic
Make batch cleanup idempotent by tolerating missing batch files. `delete_batch_by_name` now calls `file_store.delete_file(..., error_on_missing=False)`, so `cleanup_all_batches` completes even if files were already deleted or never written; added a unit test to cover this case.

<sup>Written for commit 7bd7356c36ad6265debd9f5acfeb9cd0b686a756. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/file_store/document_batch_storage.py
+++ b/backend/onyx/file_store/document_batch_storage.py
@@ -207,7 +207,7 @@ class FileStoreDocumentBatchStorage(DocumentBatchStorage):
 
     def delete_batch_by_name(self, batch_file_name: str) -> None:
         """Delete a specific batch from FileStore."""
-        self.file_store.delete_file(batch_file_name)
+        self.file_store.delete_file(batch_file_name, error_on_missing=False)
         logger.debug("Deleted batch %s from FileStore", batch_file_name)
 
     def delete_batch_by_num(self, batch_num: int) -> None:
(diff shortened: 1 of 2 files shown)


evan-onyx (approved):
good catch, thanks!

evan-onyx:
merged #10899 instead to have the tests pass, thanks for contributing!

PR #10860, opened 2026-05-06
fix(indexing): tabular chunker crash on Google Sheets with multi-line cells

Bo-Onyx:
## Description

Google Drive CSV files with old Mac-style (\r-only) line endings caused csv.reader to raise csv.Error: new-line character seen in unquoted field on every row after the header, resulting in 240 document indexing failures for a customer.

parse_csv_string and read_csv_header in csv_utils.py now transparently retry with normalized line endings when they encounter this specific error. Normalization (\r\n → \n, \r → \n) only triggers as a fallback — well-formed CSVs are untouched. All other csv.Error types are re-raised with the originating function name for easier diagnosis. tabular_section_chunker.py is unchanged.
## How Has This Been Tested?
unit test
local test
before fix
Screenshot 2026-05-06 at 11 59 11 AM

after fix
Screenshot 2026-05-06 at 12 14 31 PM

https://linear.app/onyx-app/issue/ENG-4105/google-drive-for-savvy-wealth

## Additional Options

---
## Summary by cubic
Fixes tabular chunker crashes on Google Sheets/Drive CSVs with multi‑line cells by detecting the “new-line character…” error, normalizing CRLF/CR to LF, and retrying header and row parsing. When parsing still fails, clearer errors help callers fall back to plain‑text indexing; Sheets with multi‑line cells now index reliably with structured rows (ENG-4105).

- **Bug Fixes**
  - Normalize Windows and old‑Mac line endings only when `csv.reader` raises the specific error; retry in `read_csv_header` and `parse_csv_string`.
  - Prefix other `csv.Error`s with function names for easier diagnosis; callers can fall back to plain‑text indexing.
  - Tests cover bare‑CR CSVs producing row and descriptor chunks (with and without metadata).

<sup>Written for commit 9e4b46d05aaa70cf03281a6c8609345d3937b864. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/indexing/chunking/tabular_section_chunker/tabular_section_chunker.py
+++ b/backend/onyx/indexing/chunking/tabular_section_chunker/tabular_section_chunker.py
@@ -1,3 +1,4 @@
+import csv
 from collections.abc import Iterable
 
 from pydantic import BaseModel
@@ -236,8 +237,22 @@ class TabularChunker(SectionChunker):
         payloads = accumulator.flush_to_list()
 
         text = section.text or ""
-        parsed_rows = list(parse_csv_string(text))
-        headers = parsed_rows[0].header if parsed_rows else read_csv_header(text)
+        try:
+            parsed_rows = list(parse_csv_string(text))
+        except csv.Error as e:
+            logger.warning(
+                "TabularChunker: failed to parse CSV for section (link=%s): %s",
+                section.link,
+                e,
+            )
+            parsed_rows = []
+        if parsed_rows:
+            headers = parsed_rows[0].header
+        else:
+            try:
+                headers = read_csv_header(text)
+            except csv.Error:
+                headers = []
         heading = section.heading or ""
 
         chunk_texts: list[str] = []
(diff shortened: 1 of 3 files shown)


backend/onyx/indexing/chunking/tabular_section_chunker/tabular_section_chunker.py
```diff
@@ -236,8 +237,27 @@ def chunk_section(
         payloads = accumulator.flush_to_list()
 
         text = section.text or ""
-        parsed_rows = list(parse_csv_string(text))
-        headers = parsed_rows[0].header if parsed_rows else read_csv_header(text)
+        try:
+            parsed_rows = list(parse_csv_string(text))
+        except csv.Error as e:
+            logger.warning(
```
Danelegend: This should be an error/exception level. This is a failure case that we want to be alarmed about

Danelegend (changes requested):
Not a big fan of this approach. It's essentially covering up any errors that we would want to catch and fix.

Left actionable items in slack but echoing them here:
1. Solve the underlying multi-line cell issue -> Parse it into a correct CSV that we can handle
2. In the event of a failure case, we handle as regular chunks. We still want some way of knowing that errors occurred though, otherwise these will go unnoticed

backend/onyx/indexing/chunking/tabular_section_chunker/tabular_section_chunker.py
Bo-Onyx: Added normalization as fall back and changed logger level from warning to error

Danelegend (approved):

PR #10829, opened 2026-05-04
feat(targeted-reindex): celery task lifecycle + resolution tracking

nmgarza5:
## Description

Stacks on #10828 (API + DB writes). Replaces the no-op stub task body with a real implementation that handles the lifecycle and resolution-tracking layers.

What this task does now:
- transitions the job + every linked synthetic \`IndexAttempt\` \`NOT_STARTED\` → \`IN_PROGRESS\` → \`SUCCESS\`
- marks \`IndexAttemptError.is_resolved=True\` for every target whose \`source_error_id\` is set
- snapshots \`resolved_summary\` so the audit survives later retention cleanup of \`index_attempt_errors\`
- updates \`resolved_count\` / \`skipped_count\` / \`still_failing_count\` on the job row
- on mid-task exception, marks job + non-terminal synthetic attempts \`FAILED\`, then re-raises so celery records the root cause; cleanup commit is wrapped so a secondary commit failure cannot mask the original exception

The task body is a plain function (\`run_targeted_reindex\`) wrapped by \`@shared_task\` to keep tests free of celery binding machinery.

What's still missing (next PR):
- the \`connector.reindex(refs, include_permissions=True)\` call inside the per-cc-pair loop
- piping yielded \`Document\` objects through \`index_doc_batch_with_handler\`
- \`still_failing_count\` increment when the connector yields \`ConnectorFailure\` instead of \`Document\`
- stale-write mitigation under \`prepare_to_modify_documents\`

The lifecycle and resolution-tracking layer landing here is what the FE polls and what the design's resolution-tracking contract depends on. Wiring the connector + pipeline is purely additive on top.

Design: https://linear.app/onyx-app/document/targeted-doc-re-indexing-design-4f9f9080760e

## How Has This Been Tested?

7 tests in \`backend/tests/external_dependency_unit/db/test_targeted_reindex_task.py\` against real Postgres:
- Job transitions to terminal state
- Synthetic IndexAttempts transition to terminal state
- Failure-derived targets mark the matching \`IndexAttemptError\` rows as resolved
- Arbitrary targets are no-ops on the error path
- Already-terminal job is idempotent (task drops early on re-delivery)
- Mixed jobs: failure-derived + arbitrary in the same request
- Mid-task exception → job + synthetic attempts transition to \`FAILED\` (not stuck \`IN_PROGRESS\`)

All green.

## Additional Options

- [ ] [Optional] Please cherry-pick this PR to the latest release version.
- [x] [Optional] Override Linear Check

PR #10864, opened 2026-05-06
feat(targeted-reindex): connector.reindex + indexing pipeline wiring

nmgarza5:
## Description

PR D in the targeted-reindex series. Stacks on #10829 (celery task lifecycle + resolution tracking).

Replaces the per-cc-pair stub with a real fetch + index path:

- Per-cc-pair processor (\`onyx/background/indexing/run_targeted_reindex.py\`) instantiates the connector via the standard factory, checks \`Resolver\` capability, converts target rows into \`ConnectorFailure\` inputs (rehydrating original failure context for failure-derived targets, synthesizing for arbitrary ones), and streams \`connector.reindex()\` output through the standard indexing pipeline once per active search_settings.
- Connectors that don't subclass \`Resolver\` short-circuit: their targets are marked \`still_failing\` so the admin sees a clear signal that targeted reindex isn't supported yet for that source. Today only Google Drive implements \`Resolver\` — adding it to other sources is a follow-up.
- Pipeline failures (handler caught) and connector-side \`ConnectorFailure\` yields both flow into \`still_failing_count\`.
- Resolution tracking is now gated on \`landed_doc_ids\`: an \`IndexAttemptError\` is marked resolved only if its target's doc actually made it through the pipeline. Errors whose docs failed to land stay open so the admin can retry.
- Stale-write mitigation rides for free — \`DocumentIndexingBatchAdapter\` already wraps writes in \`prepare_to_modify_documents\`, serializing concurrent full-crawl writes on the same docs.
- Module split keeps the celery task focused on lifecycle + counter aggregation; the heavy lifting lives in a helper module that mirrors \`run_docfetching.py\`.

## How Has This Been Tested?

\`backend/tests/external_dependency_unit/db/test_targeted_reindex_task.py\` — existing lifecycle tests now mock \`process_targets_for_cc_pair\`; added one new test (\`test_task_does_not_resolve_errors_when_doc_failed_to_land\`) covering the resolution-tracking gate.

\`backend/tests/external_dependency_unit/db/test_run_targeted_reindex.py\` — new file, 5 tests against the processor itself: unsupported connector short-circuit, full landing, \`ConnectorFailure\` routing, silent drop, \`ConnectorFailure\` rehydration from \`IndexAttemptError\`.

End-to-end against a real Drive cc_pair will land in the integration test suite as a follow-up.

## Additional Options

### DIFF
--- a/backend/onyx/background/indexing/run_targeted_reindex.py
+++ b/backend/onyx/background/indexing/run_targeted_reindex.py
@@ -1,0 +1,283 @@
+"""Per-cc-pair fetch + index for the targeted-reindex flow.
+
+The targeted-reindex celery task delegates the actual connector
+invocation + pipeline plumbing to this module so the task body stays
+focused on lifecycle and resolution-tracking. One call to
+`process_targets_for_cc_pair` covers a single cc_pair: it instantiates
+the connector, converts the per-doc target rows into
+`ConnectorFailure` inputs that the `Resolver.reindex` interface
+expects, and streams yielded `Document` objects through the standard
+indexing pipeline once per active search_settings (so PRESENT and
+FUTURE indexes both receive the reindex during a model swap).
+
+Connectors that don't subclass `Resolver` short-circuit: their targets
+are reported as still-failing so the admin sees clearly that targeted
+reindex isn't supported yet for that source.
+"""
+
+from collections import defaultdict
+from collections.abc import Iterable
+from collections.abc import Sequence
+
+from more_itertools import chunked
+from sqlalchemy.orm import Session
+
+from onyx.configs.app_configs import INDEX_BATCH_SIZE
+from onyx.connectors.factory import instantiate_connector
+from onyx.connectors.interfaces import Resolver
+from onyx.connectors.models import ConnectorFailure
+from onyx.connectors.models import Document
+from onyx.connectors.models import IndexAttemptMetadata
+from onyx.db.connector_credential_pair import get_connector_credential_pair_from_id
+from onyx.db.enums import AccessType
+from onyx.db.models import IndexAttempt
+from onyx.db.models import TargetedReindexJobTarget
+from onyx.db.targeted_reindex import targets_to_connector_failures
+from onyx.document_index.factory import get_all_document_indices
+from onyx.httpx.httpx_pool import HttpxPool
+from onyx.indexing.adapters.document_indexing_adapter import (
+    DocumentIndexingBatchAdapter,
+)
+from onyx.indexing.embedder import DefaultIndexingEmbedder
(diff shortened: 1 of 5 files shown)


nmgarza5:
@greptileai retrigger

backend/onyx/background/indexing/run_targeted_reindex.py
```diff
@@ -0,0 +1,323 @@
...
+    failure context (message, link) it failed on. Arbitrary targets
+    (no source error) are synthesized with a generic
+    `failure_message` — `Resolver.reindex` only consumes
+    `failed_document.document_id`, so the message body is informational.
+    """
+    error_ids = [t.source_error_id for t in targets if t.source_error_id is not None]
+    error_rows: dict[int, IndexAttemptError] = {}
+    if error_ids:
+        rows = (
+            db_session.query(IndexAttemptError)
```
evan-onyx: pls put in a db file

backend/onyx/background/indexing/run_targeted_reindex.py
```diff
@@ -0,0 +1,323 @@
...
+    ):
+        if isinstance(item, ConnectorFailure):
+            if item.failed_document is not None:
+                failed_from_connector.add(item.failed_document.document_id)
+            continue
+        if isinstance(item, Document):
+            docs.append(item)
+            continue
+        # HierarchyNode and other yields aren't actionable for
+        # targeted reindex (no hierarchy state to update mid-flow); skip.
```
evan-onyx: technically hierarchynodes is doable here; it's a bigger effort so I'm happy with just a TODO for now (potentially a followup later). We would need to mirror what happens in the indexing pipeline. it _does_ add value, since for example maybe there's a specific folder where indexing used to fail -> admin changes folder permissions and reindexes-> this flow runs and then we don't have that folder's hierarchynode. That's fine for now since by design a "null" parent is just the source type which is okay, but ideally we'd like to make the knowledge easier to select.

backend/onyx/background/indexing/run_targeted_reindex.py
```diff
@@ -0,0 +1,323 @@
...
+
+    # Per-attempt pipeline run. Each attempt commits to its own
+    # search_settings's document_indices.
+    landed_overall: set[str] = set()
+    failed_pipeline_overall: set[str] = set()
+    for attempt in cc_pair_attempts:
+        for i in range(0, len(docs), INDEX_BATCH_SIZE):
```
evan-onyx: nit: I think there might already be a batching utility for this, but if not or if it just isn't convenient here dwai

evan-onyx (approved):
some thoughts

nmgarza5:
@greptileai retrigger

nmgarza5:
@greptileai retrigger

PR #10866, opened 2026-05-06
refactor(search): replace contextual RAG string columns with model_configuration FK

raunakab:
## Description

Replaces the `contextual_rag_llm_provider` + `contextual_rag_llm_name` bare string pair on the `search_settings` table with a proper `contextual_rag_model_configuration_id` integer FK — the same pattern we already used for persona LLM provider overrides.

**Why:** String-based lookups are fragile. If a provider's display name changes or if two providers share the same name, the lookup silently resolves to the wrong model (or fails). An integer FK to `model_configuration.id` is always stable.

**Backend:**
- New Alembic migration: adds `contextual_rag_model_configuration_id` (FK → `model_configuration.id ON DELETE SET NULL`), back-fills from existing string columns, drops the strings
- `SearchSettings` SQLAlchemy model updated to use FK column
- `create_search_settings` / `update_search_settings` thread the new ID through
- `validate_contextual_rag_model` now validates by PK instead of name strings
- `update_default_contextual_model` / `get_llm_for_contextual_rag` use the ID directly
- `swap_index.py` passes the ID when syncing the default contextual model on switchover

**Frontend:**
- `AdvancedSearchConfiguration` interface: replaces string pair with `contextual_rag_model_configuration_id: number | null`
- `setNewSearchSettings` svc function threads `contextual_rag_model_configuration_id`
- `IndexSettingsPage`: new `LlmPickerById` component selects models by `model_configuration_id` instead of provider name + model name strings
- `ModelConfiguration` interface gets an `id` field (backend already returned it); `LLMOption` gets a `modelConfigurationId` field so `LlmPickerById` can resolve selections cleanly

**Backward compat:** Deprecated `contextual_rag_llm_name` / `contextual_rag_llm_provider` kept as optional fields in `IndexingSetting` Pydantic model for any callers that may still send them; they are ignored on write.

## Additional Options

---
## Summary by cubic
Replaced the contextual RAG provider/name strings with a `contextual_rag_model_configuration_id` FK to `model_configuration.id` for stable, ID-based model selection across backend and frontend. Added a safe default fallback when unset and stricter, ID-based validation.

- **Refactors**
  - Backend: `SearchSettings` stores `contextual_rag_model_configuration_id`; `update_default_contextual_model` and `get_llm_for_contextual_rag` accept IDs; indexing falls back to the global default contextual model when unset; validation checks the `ModelConfiguration` by ID and raises `OnyxError` on invalid input; update logic skips unmapped fields via SQLAlchemy inspection; session scoping fixes prevent `DetachedInstanceError`.
  - Frontend: `AdvancedSearchConfiguration` and form state use `contextual_rag_model_configuration_id`; `IndexSettingsPage` uses a unified `LlmPicker` supporting ID or name modes; `LLMOption` includes `modelConfigurationId`; `svc.ts` threads the ID.

- **Migration**
  - Adds the FK, back-fills from the removed string columns using SQLAlchemy expressions, then drops those columns (`ON DELETE SET NULL`).
  - Updated `down_revision` to the latest main head to resolve a two-head conflict.

<sup>Written for commit 35d4edbb8012efb918c604ec0f1689852d7e7b24. Summary will update on new commits.</sup>


raunakab:
Fixed in 354f2bfe2c7a13bd656fe89a52f2e84ddad9982e: rewrote migration using SQLAlchemy expressions (also fixes the invalid FROM-clause JOIN), moved `llm_from_provider` inside the session context, added `contextual_rag_model_configuration_id` to `from_db_model`, and renamed the inner `db_session` to avoid shadowing.

backend/onyx/db/search_settings.py
```diff
@@ -180,8 +180,9 @@ def update_search_settings(
     updated_settings: SavedSearchSettings,
     preserved_fields: list[str],
 ) -> None:
+    mapped_columns = {c.key for c in sa_inspect(SearchSettings).mapper.columns}
```
Danelegend, line 183: It looks like previously we ignored piping through `contextual_rag_llm_name` and `contextual_rag_llm_provider`, but now we're wanting to pipe through. Why?

backend/alembic/versions/4ff2545411ad_contextual_rag_model_configuration_fk.py
```diff
@@ -0,0 +1,101 @@
...
+            _llm_provider.c.id == _model_configuration.c.llm_provider_id,
+            _model_configuration.c.name == _search_settings.c.contextual_rag_llm_name,
+            _llm_provider.c.name == _search_settings.c.contextual_rag_llm_provider,
+            _search_settings.c.contextual_rag_llm_name.isnot(None),
+            _search_settings.c.contextual_rag_llm_provider.isnot(None),
+        )
+    )
+
+    # 3. Drop the string columns
+    op.drop_column("search_settings", "contextual_rag_llm_name")
```
Danelegend, line 66: Drop in a different PR about 1 week from now

Danelegend (approved):

backend/onyx/db/search_settings.py
raunakab: Nothing changed here — `update_search_settings` writes a field only when it exists in `mapped_columns` (the actual DB column set). The deprecated `contextual_rag_llm_name`/`contextual_rag_llm_provider` fields still exist on the Pydantic model for backward-compat acceptance of incoming requests, but the corresponding DB columns were dropped in this migration, so they simply won't appear in `mapped_columns` and are silently discarded at that boundary. Same behaviour as before.

PR #10886, opened 2026-05-06
feat(users): re-add invite-only toggle

nmgarza5:
## Description

Restores the workspace-level invite-only switch dropped during the admin Users page revamp. Backend (`invite_only_enabled`) was unchanged.

- New `InviteOnlyCard` with copy `"Restrict Open Sign-Up"` / `"New users must be invited to join this workspace."`
- Right slot of stats row: SCIM | InviteOnly | nothing
- Hidden on SAML, OIDC, CLOUD (backend bypasses the flag for SSO; CLOUD uses a different flow)
- Stats card always combined; drops the per-stat split branch

**Linear:** [ENG-4102](https://linear.app/onyx-app/issue/ENG-4102/add-invite-only-switch)

## Additional Options

### DIFF
--- a/web/src/refresh-pages/admin/UsersPage/InviteOnlyCard.tsx
+++ b/web/src/refresh-pages/admin/UsersPage/InviteOnlyCard.tsx
@@ -40,7 +40,11 @@ export default function InviteOnlyCard() {
         toast.success("Settings updated");
       } catch (err) {
         console.error("Failed to update invite_only_enabled", err);
-        toast.error("Failed to update settings");
+        const message =
+          err instanceof Error && err.message
+            ? err.message
+            : "Failed to update settings";
+        toast.error(message);
       }
     },
     [settings]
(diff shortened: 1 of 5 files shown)


web/src/refresh-pages/admin/UsersPage/InviteOnlyCard.tsx
```diff
@@ -0,0 +1,72 @@
...
+import Switch from "@/refresh-components/inputs/Switch";
+import { useSettingsContext } from "@/providers/SettingsProvider";
+import { SWR_KEYS } from "@/lib/swr-keys";
+import { toast } from "@/hooks/useToast";
+import { Settings } from "@/interfaces/settings";
+
+export default function InviteOnlyCard() {
+  const { settings } = useSettingsContext();
+
+  const saveSettings = useCallback(
```
wenxi-onyx, line 17: Should this live inside InviteOnlyCard? There's no existing util that does this?

wenxi-onyx (approved):

PR #10906, opened 2026-05-07
fix(chat): release DB session before LLM stream begins

nmgarza5:
## Description

Removes the request-scoped DB session that wrapped the entire chat streaming generator in `chat_backend.py` (lines 590, 621, 660). The session committed at `build_chat_turn()` then sat idle for the full LLM inference window — load balancers / PgBouncer dropped the connection, surfacing as `OperationalError` on cleanup.

Companion to #10159, which fixed the per-worker-thread session inside `_run_model`. This eliminates the outer session that wrapped both setup and streaming.

- `chat_backend.py`: drop outer `with`-blocks. Non-stream usage check runs in its own short session.
- `_stream_chat_turn`: opens a short session for ACL check + setup, closes it before `_run_models` begins. `expunge_all()` on the way out so `ChatTurnSetup` ORM members are detached but still readable (persona is already eager-loaded per `construct_tools` docstring).
- `_run_models`, `handle_stream_message_objects`, `handle_multi_model_stream`: drop `db_session` parameter.
- `_save_errored_message` + `llm_loop_completion_handle`: open their own short-lived sessions internally; the latter `merge()`s the detached `ChatMessage` so `save_chat_turn`'s mutations are tracked.
- Update existing eval, slack, and test callers to match.

## How Has This Been Tested?

- `mypy` clean on touched files.
- `ruff` clean on touched files.
- `pytest backend/tests/unit/onyx/chat/test_multi_model_streaming.py backend/tests/unit/onyx/chat/test_process_message_mock_llm.py` — 28 passed.

## Additional Options

### DIFF
--- a/backend/tests/external_dependency_unit/answer/test_stream_chat_message.py
+++ b/backend/tests/external_dependency_unit/answer/test_stream_chat_message.py
@@ -91,7 +91,6 @@ def test_stream_chat_with_answer(
         answer_stream = submit_query(
             query=query,
             chat_session_id=chat_session.id,
-            db_session=db_session,
             user=test_user,
         )
 
@@ -135,7 +134,6 @@ def test_stream_chat_with_answer_create_chat(
         answer_stream = submit_query(
             query=query,
             chat_session_id=None,
-            db_session=db_session,
             user=test_user,
         )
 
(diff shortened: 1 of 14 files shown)


backend/tests/external_dependency_unit/answer/test_stream_chat_message.py
```diff
@@ -222,7 +220,7 @@ def test_stream_chat_with_search_and_openurl_tools(
     ]
 
     REASONING_RESPONSE_3 = (
-        "I now know everything that I need to know. " "I can now answer the question."
+        "I now know everything that I need to know. I can now answer the question."
     )
```
nmgarza5, line 224: hmmmm maybe revert this. idk why this update was pushed

Bo-Onyx (approved):
thanks for the fix

PR #10926, opened 2026-05-07
fix(docfetching): Gracefully handle restarts

justin-tahara:
## Description

Updating the workflow for docfetching to ensure that we are able to handle proper cases when the pods that are running tasks are restarted or killed for any given reason. This is to harden the codepath to ensure that we are fault tolerant for our tasks that we enqueue

## How Has This Been Tested?

Locally + unit tests

## Additional Options

---
## Summary by cubic
Add `INTERRUPTED` indexing status with auto-resume from checkpoints when docfetching pods restart. Retries start immediately (including one-time-sync connectors) and the UI shows an “Interrupted” badge.

- **Bug Fixes**
  - Watchdog marks `INTERRUPTED` during docfetching when a heartbeat is lost and a checkpoint exists; otherwise falls back to `FAILED` with a clear “cannot auto-resume” reason. Docprocessing logic is unchanged and still consults Redis counters before invalidation.
  - Prevent race conditions: `mark_attempt_interrupted` no-ops if the attempt is no longer `IN_PROGRESS` so we don’t clobber a terminal status; it also clears `celery_task_id` and cleans counters.
  - Scheduler retries immediately when the last attempt was `INTERRUPTED`; the check runs before the `refresh_freq=None` gate so one-time-sync connectors recover. `INTERRUPTED` is ignored by repeated-error auto-pause.
  - Checkpoint recovery accepts `INTERRUPTED`; the docfetching runner preserves terminal states (including `INTERRUPTED`) and uses them to set the next poll window.
  - Monitoring treats `INTERRUPTED` as terminal. UI shows an “Interrupted” badge with a helpful tooltip; `ValidStatuses` updated.

<sup>Written for commit db62a1f8251cd2447ea60b27beba866170b59278. Summary will update on new commits.</sup>


backend/onyx/db/index_attempt.py
```diff
@@ -431,6 +431,51 @@ def mark_attempt_failed(
         raise
 
 
+def mark_attempt_interrupted(
```
Bo-Onyx, line 452: There's a narrow race: the watchdog fetches a fresh copy of the attempt (no lock), confirms it's IN_PROGRESS, then races to call mark_attempt_interrupted. If the docfetching task completes and calls mark_attempt_succeeded in that window, the watchdog can overwrite SUCCESS → INTERRUPTED. Both use with_for_update() but neither re-checks the status after acquiring the lock.

mark_attempt_failed has the same gap — it's a pre-existing pattern — but mark_attempt_interrupted makes it more dangerous because it triggers an  immediate scheduler retry on what was actually a successful attempt.

Fix is one line inside mark_attempt_interrupted after acquiring the lock:
  if attempt.status != IndexingStatus.IN_PROGRESS:
      return  # already terminal, don't overwrite

Bo-Onyx (commented):
1.
If a connector pod is killed early in its run before saving the first checkpoint, checkpoint_pointer is None → falls through to FAILED. No auto-recovery. This could be the common case for connectors that:
  - Checkpoint infrequently (e.g. only at the end of each page/cursor)
  - Were killed within the first checkpoint window
The PR has no logging or UI distinction between "FAILED because pod died with no checkpoint" vs "FAILED because of a logical connector error". Users can't tell the difference.

2.
We need some follow up changes 
https://docs.google.com/document/d/1OdxmtobFPEO1Y0kUHqj6mUWcZ3dRM1qBeJSuHaQM8ps/edit?tab=t.d7fd51vmcfno#heading=h.fpeg8rcwro2n

web/src/components/Status.tsx
```diff
@@ -70,6 +70,19 @@ export function IndexAttemptStatus({
         Canceled
       </Badge>
     );
+  } else if (status === "interrupted") {
+    const icon = (
+      <Badge variant="in_progress" icon={FiClock}>
+        Auto-retrying
```
Bo-Onyx: "Auto-retrying" label is inaccurate (nit, but visible to users)

interrupted renders as "Auto-retrying" in Status.tsx, but the INTERRUPTED attempt is a terminal state — it's done. The scheduler creates a new IndexAttempt row for the retry; the interrupted one is never resumed. The badge on the old attempt should describe what happened to it, not what happens next. Something like "Checkpointed" (or "Interrupted") would be more accurate. The new attempt already shows as "In Progress" naturally.

backend/onyx/db/enums.py
```diff
@@ -42,13 +42,19 @@ class IndexingStatus(str, PyEnum):
     CANCELED = "canceled"
     FAILED = "failed"
     COMPLETED_WITH_ERRORS = "completed_with_errors"
+    # Worker died mid-run (pod OOM, rolling deploy SIGKILL). Distinguished from
+    # FAILED so the scheduler retries immediately instead of waiting for the
+    # connector's refresh cadence, and so it isn't counted toward the 5-strike
+    # auto-pause for repeated logical failures.
+    INTERRUPTED = "interrupted"
```
evan-onyx, line 49: don't we need an alembic migration for this?

web/src/components/Status.tsx
evan-onyx: Imo the actualy status to be communicated requires like a tooltip or smth

web/src/components/Status.tsx
```diff
@@ -70,6 +70,19 @@ export function IndexAttemptStatus({
...
+      </Badge>
+    );
+    badge = errorMsg ? (
+      <Tooltip tooltip={errorMsg}>
+        <div className="cursor-pointer">{icon}</div>
```
evan-onyx, line 84: plain divs bad

evan-onyx (approved):
this is a big blast radius change so make sure u test it throughly

backend/onyx/db/index_attempt.py
justin-tahara: Fixed in db62a1f825. After acquiring the row lock with `with_for_update()` we now re-check status and bail if it's no longer IN_PROGRESS. Regression test in `test_indexing_interrupted_recovery.py::TestMarkAttemptInterruptedRace` covers all four terminal predecessor states.

You're right that `mark_attempt_failed` has the same gap pre-existing — I left it alone here since it's out of scope and FAILED doesn't trigger the immediate-retry behaviour that makes this dangerous for INTERRUPTED. Happy to file a follow-up if you'd like.

web/src/components/Status.tsx
justin-tahara: Renamed to "Interrupted" in db62a1f825 (both `IndexAttemptStatus` and `CCPairStatus` branches). Also made the tooltip unconditional so users always get an explanation of what happened — fallback copy is "Worker restarted mid-run. Onyx will automatically resume indexing from the last saved checkpoint." when there's no errorMsg.

backend/onyx/db/enums.py
justin-tahara: Confirmed no migration needed — `IndexingStatus` is declared with `Enum(IndexingStatus, native_enum=False, ...)` everywhere it's used (see `models.py` lines 2212, 2332, 2487). That stores the value as VARCHAR rather than as a Postgres native enum type, so adding a new Python enum value doesn't change the schema.

web/src/components/Status.tsx
justin-tahara: Tooltip is now unconditional in db62a1f825. If there's an errorMsg we surface it; otherwise we fall back to an explanation of what "Interrupted" means and what Onyx will do next.

web/src/components/Status.tsx
justin-tahara: Fair, but the same `<div className="cursor-pointer">` pattern already wraps the "failed" badge above (line 36-38), so I'm leaving it as-is in this PR to stay scoped — happy to take a follow-up PR to swap both for `<span>` or pass `className` through to `Badge`.

PR #10932, opened 2026-05-08
fix(ingestion): preserve external_access in Document.from_base

qasimsalam:


qasimsalam:
Wrong base — meant to target qasimsalam/alkaline3, not upstream. Closing.

Issue #2281, opened 2024-08-30
Jira Service Management Connector


Weves:
Pull in all tickets from a specified [Jira Service Management](https://www.atlassian.com/software/jira/service-management) project. 

Checkout the connector creation README here for more details on the best way to add new connectors: https://github.com/danswer-ai/danswer/blob/main/backend/danswer/connectors/README.md.

Weves:
Note on the above: please check the `Before opening PR` section for additional requirements (testing + style)!

Mubashirshariq:
@Weves I had one doubt i found jira-danswer in backend/danswer/connectors/danswer-jira, what needs to be implemented if it is already there?

Weves:
@Mubashirshariq we have a Jira connector already, but it does not work for Jira Service Management specifically (only traditional Jira). We'd like a separate JSM connector. It likely will be similar to the existing Jira connector.

Weves:
@jaykayudo, thanks for the Pull Request! Added a few comments

0Anshu1:
@Weves I need to submit only one type of connector(any one out of load,poll or event) or can I submit all three ?

harshavardhangsv:
/attempt [#2281](https://github.com/danswer-ai/danswer/issues/2281)

harshavardhangsv:
@Weves I have raised a pr for this. can you please review it.

voarsh2:
I was hoping you'd implement for DC as well - esp given Atlassian Cloud already has Rovo

harshavardhangsv:
Hey @Weves , did you get a chance to review the pr I have raised for this

serunkuma:
/attempt [#2281](https://github.com/onyx-dot-app/onyx/issues/2281)

stdthoth:
@Weves is help still needed on this ??

techsouvik:
is this issue open?

aditi-dsi:
@yuhongsun96 I'd like to work on this issue, could you please assign this to me?

Excellencedev:
unstale

Excellencedev:
@Weves Please assign me to this

nifski:
Hi @Weves!

I noticed the previous attempt (PR #5392) has been stale for months with merge conflicts and failing tests.

I have picked this up and:

Ported the logic to a fresh branch to resolve the conflicts.

Fixed the Icon format issue (SVG vs PNG) that was causing the build bot to fail.
[...]

PR #10938, opened 2026-05-08
feat(connectors): Jira Service Management connector (#2281)

MilosM348:
## Summary

Adds a new `backend/onyx/connectors/jira_service_management/` package that talks to Atlassian's Service Desk REST API (`/rest/servicedeskapi/`) instead of core Jira's `/rest/api/3`. The maintainer asked for a [separate connector](https://github.com/onyx-dot-app/onyx/issues/2281#issuecomment-2323158706) rather than extending the existing `jira/` module, so this PR adds one without touching `connectors/jira/*`.

`/claim #2281`

Closes #2281.

## What changed

| File | Purpose | LOC |
| --- | --- | --- |
| `connectors/jira_service_management/models.py` | Pydantic models (`JsmCustomer`, `JsmRequestType`, `JsmRequest`) normalised from the raw JSM API shape | +56 |
| `connectors/jira_service_management/utils.py` | `build_jsm_session` (Basic Auth: email + API token), paged `jsm_get` with typed exception mapping for 401 / 403 / 404, ISO-8601 timestamp parsing, raw -> `JsmRequest` normalisation | +143 |
| `connectors/jira_service_management/connector.py` | `JiraServiceManagementConnector` implementing `LoadConnector` (full sync) and `PollConnector` (incremental, client-side filtered on `currentStatus.statusDate`) | +201 |
| `configs/constants.py` | `+ DocumentSource.JIRA_SERVICE_MANAGEMENT` | +1 |
| `connectors/registry.py` | `+ ConnectorMapping` registration for the new source | +4 |

Total: 6 files, +405 / -0.

## API choices

* **`/rest/servicedeskapi/request`** for listing customer requests, paginated with `start` / `limit` (matches JSM's documented pagination shape, not core Jira's).
* **`requestStatus` filter** is supported but optional - leaving it `None` returns all requests in the configured Service Desk.
* **Reporter, participants, request type, organization IDs** are projected onto the `Document` so they're searchable through Onyx's standard metadata facets.
* **Document ID format**: `jsm:{issue_key}` to keep JSM and core-Jira documents disambiguated even when both connectors are enabled in the same workspace.

## Polling semantics

JSM does not expose a server-side "updated since" filter on `/request`, so `poll_source` paginates the project and filters client-side on the Pydantic-parsed `updated_at`. For small-to-medium service desks this is cheap; for large desks an explicit JQL-backed connector against core Jira is the right escape hatch and is already supported by the existing `jira/` module.

## Validation

`validate_connector_settings` hits `/servicedesk/{service_desk_id}` with the supplied credentials, so misconfigured `service_desk_id` or revoked API tokens fail at config time rather than during the first sync.

## Caveats / what I'm specifically asking maintainers to look at

1. **No new unit tests in this PR.** The test pattern in `backend/tests/unit/onyx/connectors/` for the existing Jira connector uses live-style fixtures rather than mocked HTTP, so I'd like guidance on whether to add `responses`-mocked tests under `backend/tests/unit/onyx/connectors/jira_service_management/` or expand the existing pattern. Happy to add either - asking before doing.
2. **Frontend connector card** under `web/src/lib/connectors/connectors.tsx` is intentionally not in this PR. The shape of that file has changed across recent releases and I'd rather sync the card to the latest convention as a follow-up than guess.
3. **`SlimConnectorWithPermSync`** is also out of scope here - JSM's permission model is request-by-request and the slim variant deserves a dedicated PR with explicit alignment on how org-scoped requests should map to Onyx's document-permission shape.

If any of this should be folded into this PR before review rather than as follow-ups, happy to do so. Just trying not to balloon the diff.

## Test plan

- [x] Code compiles in the local Onyx tree (`backend/onyx/connectors/jira_service_management/` imports cleanly).
- [ ] `responses`-mocked unit tests (pending guidance per caveat 1 above).
- [ ] Real-API smoke against an Atlassian Cloud free trial (will run before final review).
- [ ] Frontend connector card (follow-up PR per caveat 2).

---

Co-authored-by: Cursor <cursoragent@cursor.com>

---
## Summary by cubic
Adds a Jira Service Management connector that indexes customer requests via `/rest/servicedeskapi` with full and incremental syncs. Improves timestamp parsing and polling to avoid re-index loops and better capture key updates. 

- **New Features**
  - New `backend/onyx/connectors/jira_service_management` with `JiraServiceManagementConnector`, Pydantic models, and helpers.
  - Basic Auth (email + API token), paginated fetch, and typed errors for 401/403/404.
  - Normalizes requests to `Document` with reporter, participants, request type, org IDs, status, and priority; IDs `jsm:{issue_key}`; source `DocumentSource.JIRA_SERVICE_MANAGEMENT`.
  - Incremental sync now filters on a derived `updated_at = max(createdDate, currentStatus.statusDate, resolutionDate)`.
  - Validates settings via `/servicedesk/{service_desk_id}` and registers the connector in `connectors/registry.py`.

- **Bug Fixes**
  - Fixed datetime parsing to handle both positive and negative UTC offsets (prevents false “just updated” re-index loops).
  - Minor cleanups: explicit return type on `_session`; removed unused method.

<sup>Written for commit 585d495cf542f9e96a302e12631845bd4afa118d. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/configs/constants.py
+++ b/backend/onyx/configs/constants.py
@@ -216,6 +216,7 @@ class DocumentSource(str, Enum):
     OUTLINE = "outline"
     CONFLUENCE = "confluence"
     JIRA = "jira"
+    JIRA_SERVICE_MANAGEMENT = "jira_service_management"
     SLAB = "slab"
     PRODUCTBOARD = "productboard"
     FILE = "file"
--- a/backend/onyx/connectors/jira_service_management/connector.py
+++ b/backend/onyx/connectors/jira_service_management/connector.py
@@ -13,6 +13,18 @@ Why a separate connector instead of extending ``JiraConnector``:
     (https://github.com/onyx-dot-app/onyx/issues/2281#issuecomment-2322316167)
     explicitly asked for a separate connector.
 
+Incremental sync caveat:
+    The JSM ``/request`` endpoint does not expose a true "last modified"
+    timestamp. ``poll_source`` therefore filters on ``updated_at``, which is
+    the most-recent of ``createdDate``, ``currentStatus.statusDate`` and
+    ``resolutionDate`` (see ``utils._derive_updated_at``). That captures
+    creation, status moves and resolutions — the events most workflows care
+    about — but pure description / comment / custom-field edits on
+    already-stable tickets are **not** seen by incremental polls. Run
+    ``load_from_state`` (full sync) periodically to pick those up, or open
+    a follow-up to round-trip to ``/rest/api/3/issue/{key}.fields.updated``
+    when the maintainer signals appetite for the extra request volume.
+
 Resolves: https://github.com/onyx-dot-app/onyx/issues/2281
 """
 
@@ -23,6 +35,7 @@ from datetime import datetime
 from datetime import timezone
 from typing import Any
 
+import requests
 from typing_extensions import override
 
 from onyx.configs.app_configs import INDEX_BATCH_SIZE
@@ -81,7 +94,7 @@ class JiraServiceManagementConnector(LoadConnector, PollConnector):
         self._credentials = credentials
         return None
 
-    def _session(self):
+    def _session(self) -> requests.Session:
         if self._credentials is None:
             raise ConnectorMissingCredentialError("JSM credentials not loaded")
         return build_jsm_session(
@@ -153,12 +166,6 @@ class JiraServiceManagementConnector(LoadConnector, PollConnector):
                 "organization_ids": ",".join(request.organization_ids),
             },
         )
-
-    def _flush(
-        self, batch: list[Document]
-    ) -> Iterable[list[Document]]:
-        if batch:
-            yield batch
 
     def _yield_in_batches(
         self, requests: Iterable[JsmRequest]
--- a/backend/onyx/connectors/jira_service_management/connector.py
+++ b/backend/onyx/connectors/jira_service_management/connector.py
@@ -1,0 +1,201 @@
+"""Jira Service Management connector for Onyx.
+
+Pulls customer requests from a specified JSM (Service Desk) project and yields
+them as ``Document`` objects to be indexed by Onyx. Mirrors the structure of
+``onyx.connectors.jira.connector`` but talks to the Service Desk REST API
+(``/rest/servicedeskapi``) instead of core Jira's ``/rest/api/3``.
+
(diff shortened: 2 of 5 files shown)


MilosM348:
Thanks @greptile-apps and @cubic-dev-ai for the review — addressed all 4 findings in commit [`585d495`](https://github.com/onyx-dot-app/onyx/pull/10938/commits/585d4951).

**P1 — `parse_jsm_datetime` silently fell back to `now()` for negative UTC offsets** (`utils.py:81-85`)
The `"+" in cleaned[10:]` guard fired only for positive offsets, so `"…-0500"` raised `ValueError` on Python <3.11, the `except` returned `None`, and `to_jsm_request` fell back to `datetime.now(UTC)` — re-indexing the request on every poll. Replaced with a single `re.sub(r"([+-])(\d{2})(\d{2})$", r"\1\2:\3", cleaned)` that handles both signs (and is a no-op for already-colonised offsets like `+02:00`). Verified manually:

```
'-0500' -> 2026-05-06 18:45:00+00:00   # was: None -> now()
'+0530' -> 2026-05-06 08:15:00+00:00
'+02:00' -> 2026-05-06 11:45:00+00:00
```

**P1 — `updated_at` from `currentStatus.statusDate` misses non-status edits** (`utils.py:138`)
Right call. The JSM `/request` endpoint doesn't return a `fields.updated`-style "last modified" timestamp at all, so I can't fully fix this without round-tripping to `/rest/api/3/issue/{key}` (extra request volume per ticket). The minimal-risk fix I shipped: `_derive_updated_at` now takes `max(createdDate, statusDate, resolutionDate)`, which captures creation, status moves and resolutions — the events most support workflows care about — and never falls back below `createdDate`. Pure description / comment / custom-field edits on already-stable tickets are still missed by incremental polls; documented prominently in the module docstring with the two mitigations (periodic full sync, or a follow-up that does the per-issue round-trip). Happy to take that follow-up if a maintainer signals appetite for the extra request budget.

**P2 — `_session` missing return-type annotation** (`connector.py:84`)
Added `-> requests.Session`.

**P2 — `_flush` is dead code** (`connector.py:157-161`)
Removed. `_yield_in_batches` already handles the final partial batch.

Diff: 2 files, +68 / -14. No public-API changes; the regex normalisation is strictly more permissive than the old guard.

The CLA / unit-test guidance question from the original PR body is still open whenever the human reviewers have time.

PR #10891, opened 2026-05-07
feat(targeted-reindex): wire Resolve-All button + harden synthetic IndexAttempt lifecycle

nmgarza5:
## Description

FE Resolve-All on the indexing-errors modal POSTs unresolved error ids to `/targeted-reindex` (chunked at 100) and fires-and-forgets. The errors table self-refreshes via existing SWR auto-refresh; rows flip to Resolved as the backend lands docs. No client-side job-status polling.

BE fixes uncovered while testing the flow:
- Stamp `celery_task_id` on synthetic `IndexAttempt` at creation so the orphan-sweep heartbeat doesn't false-flag it.
- Increment `total_docs_indexed` / `total_chunks` / `completed_batches` / `last_progress_time` per batch in `run_targeted_reindex`.
- Filter `targeted_reindex_job_id IS NULL` in three docprocessing monitors (active-attempt validator, inconsistent-attempt sweeper, `monitor_indexing_attempt_progress`).
- Use configured `client_app` (broker=redis) instead of unconfigured `celery.current_app` (defaults to AMQP) when sending the targeted reindex task from the API.

## How Has This Been Tested?

Live Drive cc_pair (cc_pair=194, ~125 unresolved errors) on `dev1` with `DEV_LOGGING_ENABLED=true`. Two parallel jobs.

### POST → Celery dispatch

Click Resolve-All. FE chunks unresolved error ids at 100 and fires `Promise.all` of POSTs.

```
POST /manage/admin/indexing/targeted-reindex HTTP/1.1" 200
POST /manage/admin/indexing/targeted-reindex HTTP/1.1" 200
```

### Synthetic IndexAttempt

`create_targeted_reindex_job` inserts one synthetic `IndexAttempt` per active `search_settings` with `celery_task_id` stamped at creation:

```sql
SELECT id, status, targeted_reindex_job_id AS job, celery_task_id IS NOT NULL AS has_task
FROM index_attempt WHERE targeted_reindex_job_id IN (95, 96);

 id  | status  | job | has_task
-----+---------+-----+----------
 341 | SUCCESS |  95 | t
 340 | SUCCESS |  96 | t
```

### Worker executes

`process_targets_for_cc_pair` iterates `connector.reindex(errors=...)` and streams `Document` objects through the indexing pipeline. Per-batch counter updates land:

```
 id  | status  | job | docs | chunks | batches | has_err
-----+---------+-----+------+--------+---------+--------
 341 | SUCCESS |  95 |   83 |    416 |       6 | f
[...]

PR #10942, opened 2026-05-08
feat(targeted-reindex): Confluence Resolver.reindex

nmgarza5:
## Description

Implements `Resolver.reindex` for `ConfluenceConnector`. Inputs are doc URLs (Confluence's `document_id` format). Page ids are parsed out of `/pages/<id>/...` and `pageId=<id>` URL shapes, batched into one `type=page and id IN (...)` CQL query. Yields space + ancestor `HierarchyNode`s, converts each returned page to a `Document`, emits `ConnectorFailure` for unparseable URLs and for page ids Confluence doesn't return.

### Permissions

`include_permissions=True` is honored. When the cc_pair is `access_type=SYNC` and EE is enabled:

- expand fields gain `restrictions` + `ancestors.restrictions` on the page CQL fetch
- `get_all_space_permissions(..., add_prefix=True)` runs once per call
- per page, `get_page_restrictions(..., add_prefix=True)` is combined with the space-level fallback and stamped onto `Document.external_access`

`add_prefix` is a new defaulted param on both wrappers in `backend/onyx/connectors/confluence/access.py`. Slim/perm-sync callers keep their unprefixed default; the indexing path passes `True` so group ids carry the source-type prefix the search ACL filter expects (matches Drive and the rest of the indexing path).

### Implications

| Area | Effect |
| -- | -- |
| Non-EE | Helpers short-circuit, no-op |
| EE + SYNC cc_pair | Reindexed doc lands with ACL stamped → searchable under correct perms without waiting for next slim sync |
| EE + non-SYNC cc_pair | `include_permissions=False`, helpers not called |
| API cost | +1 HTTP for space perms, +1 per page for restrictions. Bounded at MAX_TARGETS_PER_REQUEST=100 |
| CQL URL length | ~1500–2000 chars at 100 ids + restrictions expand. Within ~5KB Confluence limit |
| Concurrent perm sync race | Per-doc `prepare_to_modify_documents` lock, last-writer-wins (same as full-crawl ↔ slim-sync today) |
| Slim/perm-sync callers | Unchanged — `add_prefix` defaults to `False` on both wrappers |

> Stacked on #10891. Base auto-retargets to `main` once that merges.

## How Has This Been Tested?

`backend/tests/daily/connectors/confluence/test_resolver.py` covers six scenarios against the real `DailyConne` test space:

- single page reindex → 1 Document, 0 failures, ≥1 hierarchy node
- multi-page reindex → all pages returned, 0 failures
- unknown page id → 1 failure, 0 documents
- unparseable URL (`/display/SPACE/Title`) → 1 failure, 0 documents
- empty input → empty output
- `EntityFailure` (not `DocumentFailure`) input → empty output

## Additional Options


backend/onyx/connectors/confluence/connector.py
```diff
@@ -861,6 +881,107 @@ def build_dummy_checkpoint(self) -> ConfluenceCheckpoint:
...
+            if include_permissions
+            else {}
+        )
+
+        expand_fields = list(_PAGE_EXPANSION_FIELDS)
+        if include_permissions:
+            expand_fields.extend(_RESTRICTIONS_EXPANSION_FIELDS)
+
+        quoted_ids = ",".join("'%s'" % pid for pid in url_to_page_id.values())
+        cql = "type=page and id IN (%s)" % quoted_ids
```
evan-onyx, line 931: somewhat concerned about how well this scales, for now I'd just say leave a TODO to paginate this

evan-onyx (approved):
lgtm just one lil sugg

PR #10949, opened 2026-05-08
feat(targeted-reindex): plumb HierarchyNode yields through reindex

nmgarza5:
## Description

`process_targets_for_cc_pair` was discarding `HierarchyNode` items the Resolver yields (TODO since the connector wiring landed in #10864). Mirror the docfetching path: collect ancestor nodes from `connector.reindex(...)`, sanitize, upsert via `upsert_hierarchy_nodes_batch`, link to the cc_pair, and cache in redis for fast ancestor resolution.

Without this, an admin restoring folder permissions and reindexing the affected docs would leave the new ancestor relationship stuck on "source-type root" until the next full crawl.

Single new helper `_persist_hierarchy_nodes` in `run_targeted_reindex.py`. Reuses every helper docfetching uses — same sanitization, same upsert, same redis cache call. Bounded: `MAX_TARGETS_PER_REQUEST=100` caps the universe of yielded nodes.

## How Has This Been Tested?

Manual code path verification + ty/ruff/black green. End-to-end coverage will land via the existing daily Drive integration suite (`backend/tests/daily/targeted_reindex/test_drive_targeted_reindex_integration.py`) on the next nightly — Drive's Resolver already yields hierarchy nodes for ancestors via `_get_new_ancestors_for_files`.

## Additional Options

### DIFF
--- a/backend/onyx/background/indexing/run_targeted_reindex.py
+++ b/backend/onyx/background/indexing/run_targeted_reindex.py
@@ -290,6 +290,11 @@ def process_targets_for_cc_pair(
             tenant_id=tenant_id,
             db_session=db_session,
         )
+        logger.debug(
+            "Persisted and cached %s hierarchy nodes for cc_pair_id=%s",
+            len(hierarchy_nodes),
+            cc_pair_id,
+        )
 
     # Per-attempt pipeline run. Each attempt commits to its own
     # search_settings's document_indices.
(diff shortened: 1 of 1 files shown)


backend/onyx/background/indexing/run_targeted_reindex.py
```diff
@@ -164,6 +172,41 @@ def _flush_batch(
     return landed_ids, failed_ids
 
 
+def _persist_hierarchy_nodes(
```
evan-onyx, line 175: ideally I'd like to have this and docfetching actually call the same function (this one) but not critical

evan-onyx (approved):

backend/onyx/background/indexing/run_targeted_reindex.py
nmgarza5: ok yeah will follow up on this. there is more work with timers to get things hooked up

PR #10967, opened 2026-05-09
chore: unify integration tests to use shared fixtures

wenxi-onyx:
Adds shared `api_key` fixture to the root integration conftest and refactors 7 test files to use it (along with the existing `admin_user` fixture) instead of creating them inline. Reduces boilerplate and makes the test setup pattern consistent across connector, document_set, usergroup, MCP, and query_history tests.

## Additional Options


acaprau (approved):
delicious. the comments about test dependency ordering sound legit tho w.r.t. flakiness perhaps.

PR #10989, opened 2026-05-11
Add Jira Service Management connector (#2281)

s300169140:
Closes #2281.

## What

A new connector for Jira Service Management (JSM) projects, implemented as a thin subclass of \`JiraConnector\`.

JSM is built on Jira's data model — JSM \"requests\" are stored as Jira issues with specific issue types (Service Request, Incident, Problem, Change, etc.), surfaced through a different UI but fetched via the same Atlassian Cloud REST API. So the right abstraction here is to **delegate to the proven Jira connector** rather than duplicate 1,000+ lines.

## What this connector adds on top of JiraConnector

1. **Auto-injected JQL filter** scoped to the standard JSM issue types:
   \`Service Request\`, \`Service Request with Approvals\`, \`Incident\`, \`Problem\`, \`Change\`, \`Emergency Change\`, \`Task\`.
   Configurable via the new \`jsm_issue_types\` argument; pass an empty tuple to disable the filter entirely.

2. **\`JIRA_SERVICE_MANAGEMENT\` source tag** on every yielded \`Document\` so JSM tickets are distinguishable from regular Jira tickets in search, filters, and the UI.

3. **\`jql_query\` overrides the auto-filter** — if a user provides their own JQL, the issue-type clause is *not* prepended. Power users with custom workflows retain full control.

## What it inherits unchanged

Authentication, project scoping, scoped tokens, checkpointing, permission sync, bulk-fetch retry semantics — all delegate to \`JiraConnector\`. JSM Cloud uses the same auth tokens and the same REST endpoints as Jira Cloud, so there's nothing JSM-specific to change here.

## Files

| File | What |
|---|---|
| \`backend/onyx/configs/constants.py\` | New \`DocumentSource.JIRA_SERVICE_MANAGEMENT\` |
| \`backend/onyx/connectors/jira_service_management/connector.py\` | New connector (~160 LOC) |
| \`backend/onyx/connectors/registry.py\` | Register \`JsmConnector\` |
| \`backend/tests/unit/onyx/connectors/jira_service_management/test_jsm_connector.py\` | 11 unit tests |

## Tests

```
$ pytest backend/tests/unit/onyx/connectors/jira_service_management/
======================== 11 passed ========================

$ pytest backend/tests/unit/onyx/connectors/jira/
======================== 32 passed ========================  ← existing Jira tests unchanged
```

Unit tests cover:
- JQL clause building (default types / single type / empty / multi-word quoting)
- Constructor variants: default issue types, explicit \`jql_query\` override, custom \`jsm_issue_types\`, empty filter, base-class attrs preserved
- \`process_jsm_issue\` correctly tags the source as \`JIRA_SERVICE_MANAGEMENT\` and respects the labels-to-skip logic

## What I haven't tested (honest)

End-to-end against a live JSM Cloud project. I don't have a JSM Cloud trial environment set up. The connector reuses the proven Jira fetch path, so the risk is concentrated in:

- The JQL clause (unit-tested)
- The source tagging (unit-tested)
- The JsmConnector constructor wiring (unit-tested)

Happy to collaborate on integration tests once a JSM test project is available, or to defer to a maintainer with a JSM Cloud connection to run validation. If you'd like me to add an external-dependency-unit test using a mocked Atlassian server (like \`responses\`/\`requests-mock\`), let me know and I'll add one.

## Frontend / connector config UI

This PR is backend-only. If the JSM connector needs additional fields in the frontend config form (it shouldn't — auth/project setup is identical to Jira), I can follow up with a separate PR once you confirm the desired UX.

---
## Summary by cubic
Adds a Jira Service Management connector that reuses the Jira API path to index JSM requests with scoped JQL and JSM-specific tagging. Also fixes a `_load_from_checkpoint` signature mismatch to prevent runtime errors, and removes a no-op settings override.

- New Features
  - Introduces `JsmConnector` subclassing `JiraConnector` with inherited auth, scoping, checkpointing, and retries.
  - Auto-injects a JQL `issuetype IN (...)` filter for standard JSM types; configurable via `jsm_issue_types` (empty tuple disables).
  - Respects a user-provided `jql_query` and does not prepend the auto-filter.
  - Tags documents with `JIRA_SERVICE_MANAGEMENT`; adds `DocumentSource.JIRA_SERVICE_MANAGEMENT` and registry mapping.

- Bug Fixes
  - Corrects `_load_from_checkpoint(jql, checkpoint, include_permissions)` signature and adds regression tests for signature alignment and JSM source retagging.
  - Removes a no-op `validate_connector_settings` override.
  - Removes unused imports.

<sup>Written for commit dd629c5ee845c63ff8eef6f04bd7a08bdbba8f79. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/configs/constants.py
+++ b/backend/onyx/configs/constants.py
@@ -216,6 +216,7 @@ class DocumentSource(str, Enum):
     OUTLINE = "outline"
     CONFLUENCE = "confluence"
     JIRA = "jira"
+    JIRA_SERVICE_MANAGEMENT = "jira_service_management"
     SLAB = "slab"
     PRODUCTBOARD = "productboard"
     FILE = "file"
--- a/backend/onyx/connectors/jira_service_management/connector.py
+++ b/backend/onyx/connectors/jira_service_management/connector.py
@@ -1,0 +1,185 @@
+"""Connector for Jira Service Management (JSM) projects.
+
+JSM is built on Jira's data model — JSM "requests" are stored as Jira issues
+with a specific set of issue types (Service Request, Incident, Problem, Change,
+etc.). The underlying REST API for fetching them is the same Jira REST API
+used by the standard Jira connector, so this connector reuses
+`JiraConnector` and adds two things on top:
+
+  1. A default JQL filter scoped to JSM issue types. Users can override this
+     by setting `jsm_issue_types` or by providing their own `jql_query` (which
+     replaces the default filter entirely).
+  2. A document `source` tag of `JIRA_SERVICE_MANAGEMENT` instead of `JIRA`,
+     so JSM tickets show up distinctly in Onyx.
+
+Authentication and credential handling are identical to the standard Jira
+connector — JSM is part of the same Atlassian Cloud tenant and uses the same
+API tokens / scoped tokens.
+"""
+
+from collections.abc import Callable
+from collections.abc import Generator
+from typing import Any
+
+from jira.resources import Issue
+from typing_extensions import override
+
+from onyx.configs.app_configs import INDEX_BATCH_SIZE
+from onyx.configs.app_configs import JIRA_CONNECTOR_LABELS_TO_SKIP
+from onyx.configs.constants import DocumentSource
+from onyx.connectors.cross_connector_utils.miscellaneous_utils import time_str_to_utc
+from onyx.connectors.exceptions import ConnectorValidationError
+from onyx.connectors.interfaces import CheckpointOutput
+from onyx.connectors.interfaces import GenerateSlimDocumentOutput
+from onyx.connectors.interfaces import SecondsSinceUnixEpoch
+from onyx.connectors.jira.connector import bulk_fetch_issues
+from onyx.connectors.jira.connector import enhanced_search_ids
+from onyx.connectors.jira.connector import JiraConnector
+from onyx.connectors.jira.connector import JiraConnectorCheckpoint
+from onyx.connectors.jira.connector import process_jira_issue
+from onyx.connectors.jira.utils import build_jira_url
+from onyx.connectors.jira.utils import extract_text_from_adf
+from onyx.connectors.models import ConnectorMissingCredentialError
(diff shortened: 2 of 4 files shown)


s300169140:
Thanks @greptile-apps — both issues fixed in 763c3c4:

**1. Signature mismatch on `_load_from_checkpoint`**

You were right — the override was `(start, end, checkpoint, include_permissions)` but the parent's signature (and the dispatch from `load_from_checkpoint` / `load_from_checkpoint_with_perm_sync` at `jira/connector.py:741,760`) is `(jql, checkpoint, include_permissions)`. Fixed to match parent exactly.

To prevent regression I added two new unit tests:
- `test_load_from_checkpoint_signature_matches_parent` — introspects both signatures and asserts parameter names align, so any future drift fails the unit suite rather than production indexing.
- `test_load_from_checkpoint_retags_source` — monkeypatches the parent's load to yield a `Document` with `source=JIRA`, then verifies the override correctly re-tags it as `JIRA_SERVICE_MANAGEMENT`.

**2. Unused imports**

Trimmed down to only what the module actually references (11 removed).

```
$ pytest backend/tests/unit/onyx/connectors/jira_service_management/
======================== 13 passed ========================
$ pytest backend/tests/unit/onyx/connectors/jira/
======================== 32 passed ========================
```

s300169140:
Cleaned up the P2 in dd629c5 — dropped the no-op `validate_connector_settings` override since MRO already routes to the parent's implementation. Left a comment naming the inherited public-API surface so it's still discoverable.

All three Greptile P0/P2s now addressed:
1. ✅ `_load_from_checkpoint` signature corrected to match parent
2. ✅ Unused imports removed (11 down to 0)  
3. ✅ No-op `validate_connector_settings` removed

```
$ pytest backend/tests/unit/onyx/connectors/jira_service_management/ backend/tests/unit/onyx/connectors/jira/
======================== 45 passed ========================
```

The 15 "awaiting approval" workflows are blocked on the first-time-contributor security gate — a maintainer needs to click "Approve and run" to release them. Happy to wait.

PR #10995, opened 2026-05-11
fix(mcp): per user multiple header

evan-onyx:
## Description

We were previously not storing/sending the actual keys needed per-user for multi-header MCP servers. This was working fine since we defaulted to just asking for API key which covered 99% of use cases. Now we cover the rest!

## How Has This Been Tested?

added new playwright tests and tested in UI

## Additional Options

---
## Summary by cubic
Fixes the per-user MCP API key flow to support servers that require multiple headers. Users now see prompts for all needed fields, and backend-managed placeholders like `{user_email}` are auto-filled.

- **Bug Fixes**
  - Persist and derive `required_fields` on `MCPAuthTemplate` (excluding auto-filled placeholders) to drive multi-field user prompts; older servers fall back to deriving from header templates.
  - Centralize substitutions with `apply_auto_substitutions` and use it when building headers; persist `header_substitutions` on upsert.

- **New Features**
  - Added e2e tests (`mcp_per_user_key.spec.ts`) using a FastAPI mock server that enforces required headers via `--require-header`; covers admin setup, attaching to the default agent, multi-field prompts, and re-auth.
  - CI spins up the mock server via `docker-compose.mcp-per-user-key-test.yml` and waits on `/healthz`; local tests can launch it with `startMcpPerUserKeyServer`.

<sup>Written for commit 985b1c959d2622d8123c71e6df9f4e3df4a32bc5. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/server/features/mcp/models.py
+++ b/backend/onyx/server/features/mcp/models.py
@@ -103,9 +103,8 @@ class MCPAuthTemplate(BaseModel):
     @staticmethod
     def derive_required_fields(headers: dict[str, str]) -> list[str]:
         """Extract the set of `{placeholder}` field names referenced by
-        ``headers`` values, preserving the order they're first encountered
-        and excluding placeholders the backend fills in automatically (see
-        ``AUTO_SUBSTITUTED_PLACEHOLDER_KEYS``).
+        ``headers`` values, excluding placeholders the backend fills in
+        automatically (see ``AUTO_SUBSTITUTED_PLACEHOLDER_KEYS``).
         """
         seen: set[str] = set()
         for value in headers.values():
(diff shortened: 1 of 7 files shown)


backend/onyx/server/features/mcp/models.py
```diff
@@ -61,6 +100,21 @@ class MCPAuthTemplate(BaseModel):
...
     )
 
+    @staticmethod
+    def derive_required_fields(headers: dict[str, str]) -> list[str]:
+        """Extract the set of `{placeholder}` field names referenced by
+        ``headers`` values, preserving the order they're first encountered
+        and excluding placeholders the backend fills in automatically (see
+        ``AUTO_SUBSTITUTED_PLACEHOLDER_KEYS``).
+        """
+        seen: dict[str, None] = {}
```
Danelegend: set?

Danelegend (approved):

backend/onyx/server/features/mcp/models.py
evan-onyx: wtf was claude cooking lol

PR #11005, opened 2026-05-12
refactor: better copy PAT button in creation modal

wenxi-onyx:
## Additional Options

---
## Summary by cubic
Improved the PAT creation flow by moving the generated token into a focused modal with a clear “Copy Token” action. This makes copying explicit and clarifies the token won’t be shown again.

- **New Features**
  - Show the token in a dedicated `Modal` with a warning that it won’t be shown again; hide the inline code block copy control.
  - Add a footer “Copy Token” button using `CopyIconButton` with clipboard write and toast feedback.

- **Refactors**
  - Simplified the creation modal to only include name/expiry and a “Create Token” button; removed “Done”. Close via Escape or the header close icon.
  - Updated e2e tests to click “Copy Token” and close the modal with Escape.

<sup>Written for commit 6cac56ab1bc525303b411e308c8904a34e5fda4f. Summary will update on new commits.</sup>

### DIFF
--- a/web/src/refresh-pages/SettingsPage.tsx
+++ b/web/src/refresh-pages/SettingsPage.tsx
@@ -14,6 +14,7 @@ import { Formik, Form } from "formik";
 import * as Yup from "yup";
 import {
   SvgArrowExchange,
+  SvgCopy,
   SvgKey,
   SvgLock,
   SvgMinusCircle,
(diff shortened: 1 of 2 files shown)


raunakab (approved):
We have a dedicated `CopyButton`. Maybe we could use that one here instead?

PR #11016, opened 2026-05-12
fix(asana): stop dropping teamless projects from explicit allowlist

justin-tahara:
## Description

The previous fix in #6689 added an unconditional `if not team_gid: return` skip in `_get_tasks_for_project`. Asana legitimately returns `team: null` for workspace-level projects (and apparently in other circumstances we still need to characterize), so this silently dropped every such project — **even when the user had explicitly listed it in `asana_project_ids`**.

Symptom in the field: self-hosted customers on v2.5.19+ saw their Asana syncs go to 0 documents with "Succeeded" status and no errors in logs. The only signal was the `Skipping project without a team: ...` info line — easy to miss.

Concrete change in `_get_tasks_for_project`:
- Removes the unconditional teamless-project skip.
- Keeps team-based filtering only for **private** projects whose team doesn't match the configured `team_gid` (matching the spirit of the prior behavior but no longer crashing on `None`).
- **Bypasses the team filter entirely when the user supplied an explicit `asana_project_ids` allowlist** — those projects were deliberately opted in, regardless of which team they sit in.

## How Has This Been Tested?

New unit tests in `test_asana_connector.py`:
- `test_get_tasks_processes_teamless_project_when_no_team_filter` — teamless project must sync when no team filter is configured.
- `test_get_tasks_processes_teamless_projects_when_explicitly_allowlisted` — the customer's exact production scenario: 3 explicitly-listed projects, all returning `team=None`, `team_gid` configured. Before this PR all 3 were dropped; now they sync.
- `test_get_tasks_skips_private_project_in_wrong_team` — preserves the team-narrowing behavior for the non-allowlist case.
- `test_get_tasks_processes_private_wrong_team_project_when_allowlisted` — explicit allowlist overrides the team filter.
- `test_get_tasks_skips_archived_project_even_when_allowlisted` — archived skip remains unconditional.

Local repro against a real Asana account is the next step.

## Additional Options

---
## Summary by cubic
Fixes Asana sync dropping workspace-level (teamless) projects and honors explicit project allowlists, restoring expected indexing and preventing empty “Succeeded” runs.

- **Bug Fixes**
  - Removed the unconditional skip for teamless projects; archived skip remains.
  - Team filter applies only to private projects and is bypassed for projects in `asana_project_ids`.
  - Added tests to pin teamless + team-filter behaviors and allowlist overrides; updated test mocks to accept keyword args.

- **Refactors**
  - Passed `project_gids` through to per-project task fetching and removed the derived “allowlist” boolean; no behavior change.

<sup>Written for commit 66e4f39d395676ad2e8ee876ab0d855b1d1f2254. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/connectors/asana/asana_api.py
+++ b/backend/onyx/connectors/asana/asana_api.py
@@ -82,10 +82,6 @@ class AsanaAPI:
                 logger.info("Processed %s projects", project_count)
 
         logger.info("Found %s projects to process", len(projects_list))
-        # When the user supplied an explicit project allowlist, bypass the team
-        # filter for those projects — they were deliberately opted in even if
-        # they sit outside the configured team or have no team association.
-        from_explicit_allowlist = project_gids is not None
         # Asana tasks can belong to multiple projects and thus tasks
         # can get reported multiple times
         seen_task_gids: set[str] = set()
@@ -95,7 +91,7 @@ class AsanaAPI:
                 start_date,
                 start_seconds,
                 seen_task_gids,
-                from_explicit_allowlist,
+                project_gids,
             ):
                 yield task
         logger.info("Completed fetching %s tasks from Asana", self.task_count)
@@ -110,7 +106,7 @@ class AsanaAPI:
         start_date: str,
         start_seconds: int,
         seen_task_gids: set[str],
-        from_explicit_allowlist: bool,
+        project_gids: list[str] | None,
     ) -> Iterator[AsanaTask]:
         project = self.project_api.get_project(project_gid, opts={})
         project_name = project.get("name", project_gid)
@@ -120,11 +116,15 @@ class AsanaAPI:
         if project.get("archived"):
             logger.info("Skipping archived project: %s (%s)", project_name, project_gid)
             return
+        # The team filter narrows the workspace-wide sync. When the user has
+        # picked specific projects via `project_gids`, respect that choice
+        # regardless of team membership — those projects were explicitly
+        # opted in upstream by the project_gid filter in get_tasks.
         if (
             project.get("privacy_setting") == "private"
             and self.team_gid
             and team_gid != self.team_gid
-            and not from_explicit_allowlist
+            and project_gids is None
         ):
             logger.info(
                 "Skipping private project not in configured team: %s (%s)",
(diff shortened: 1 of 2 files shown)


justin-tahara:
**Live repro + validation against a real Asana workspace**

Reproduced the customer's exact symptom against a real Asana project that returns `team: null` from `GET /projects/<gid>`. Same connector config (workspace + team_id + explicit project_ids allowlist), only the source branch differs:

| | main (buggy) | jtahara/asana-team-filter-fix |
|---|---|---|
| log line | `Skipping project without a team: Justin's first project` | `Fetching tasks modified since ... for project: Justin's first project` |
| tasks emitted | 0 | 3 |

The skipped-project log message and the 0-doc outcome match the customer's screenshots line-for-line. Also confirmed: the project in question has `privacy_setting: "private"` and `team: null` simultaneously — so the only filter that matters end-to-end here is the new explicit-allowlist bypass.

backend/onyx/connectors/asana/asana_api.py
```diff
@@ -82,12 +82,20 @@ def get_tasks(
                 logger.info("Processed %s projects", project_count)
 
         logger.info("Found %s projects to process", len(projects_list))
+        # When the user supplied an explicit project allowlist, bypass the team
+        # filter for those projects — they were deliberately opted in even if
+        # they sit outside the configured team or have no team association.
+        from_explicit_allowlist = project_gids is not None
```
Danelegend: A bit confused here - Not too sure why having a list of gids implies it belongs in the explicit allowlist

Danelegend (approved):

backend/onyx/connectors/asana/asana_api.py
Danelegend: Actually worked it out

justin-tahara: Fair callout — the boolean conflated 'user provided a list' with 'this project is allowlisted'. The link only works because of the upstream filter at line 74-79 (`if project_gids is None or project_gid in project_gids`), so by the time `_get_tasks_for_project` runs, every project that reaches it is one the user named.

Refactored in 66e4f39 to drop the intermediate flag and pass `project_gids: list[str] | None` straight through. The skip now reads `project_gids is None` directly, which mirrors the upstream filter and reads as 'apply the team filter only when the user is asking for the whole workspace'. No behavior change; all 18 tests still green.

PR #11025, opened 2026-05-12
feat(indexing): add Cohere embed-v4.0 with response-format compat

justin-tahara:
## Description

Adds Cohere's [embed-v4.0](https://docs.cohere.com/docs/cohere-embed) (1536-dim) as a first-class embedding model for [ENG-4039](https://linear.app/onyx-app/issue/ENG-4039).

**Backend:**
- Register \`cohere/embed-v4.0\` in \`SUPPORTED_EMBEDDING_MODELS\` (index name \`danswer_chunk_cohere_embed_v4_0\`, dim 1536) so the Vespa setup creates the index.
- Add \`_extract_cohere_embeddings\` helper in \`search_nlp_models.py\` and route \`_embed_cohere\` through it. Cohere v3 returns \`response.embeddings\` as a flat \`list[list[float]]\`; v4 returns an \`EmbedByTypeResponseEmbeddings\` object that carries the embeddings under \`.float_\`. The helper handles both shapes so existing v3 users aren't broken while new v4 users work correctly.

**Frontend:**
- Add a \`embed-v4.0\` entry under the Cohere provider in \`web/src/lib/indexing/index.ts:CLOUD_BASED_PROVIDERS\` so the admin can select it from the embedding picker.

Note: deliberately added only one \`_BaseEmbeddingModel\` entry for v4 (no legacy \`danswer_chunk_embed_v4_0\` alias) — that two-entry pattern only exists for older models that went through a Vespa naming migration. Newer additions like \`gemini-embedding-001\` and \`text-embedding-005\` also use a single entry.

Re-opens the change-set from the now-stale #10523.

## How Has This Been Tested?

- Three new unit tests pass locally:
  - \`test_supported_embedding_models_include_cohere_embed_v4\` — config registers correctly.
  - \`test_cohere_embed_supports_v3_response_format\` — v3 flat-list response unchanged.
  - \`test_cohere_embed_supports_v4_response_format\` — v4 \`.float_\` response decoded correctly.
- Pre-commit (ty, ruff, black, prettier, tsgo) passes.

## Additional Options

---
## Summary by cubic
Adds Cohere `cohere/embed-v4.0` as a supported embedding model and makes response handling compatible with both v3 and v4. This registers a Vespa index and exposes the model in the admin UI, addressing ENG-4039.

- **New Features**
  - Add `cohere/embed-v4.0` (1536-dim) to `SUPPORTED_EMBEDDING_MODELS` with index `danswer_chunk_cohere_embed_v4_0`.
  - Expose the model in the embedding picker under the Cohere provider (`web/src/lib/indexing/index.ts`).

- **Bug Fixes**
  - Normalize Cohere embed responses across SDK versions via `_extract_cohere_embeddings` and explicitly request `embedding_types=['float']` in `_embed_cohere` to guarantee v4 `.float_` data (ENG-4039).
  - Add unit tests for v3/v4 response formats and model registration.

<sup>Written for commit cd46a314742e7a72e832dec5a11cdb2b0d4c83f0. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/configs/embedding_configs.py
+++ b/backend/onyx/configs/embedding_configs.py
@@ -42,6 +42,11 @@ _BASE_EMBEDDING_MODELS = [
         index_name="danswer_chunk_embed_english_light_v3_0",
     ),
     _BaseEmbeddingModel(
+        name="cohere/embed-v4.0",
+        dim=1536,
+        index_name="danswer_chunk_cohere_embed_v4_0",
+    ),
+    _BaseEmbeddingModel(
         name="openai/text-embedding-3-large",
         dim=3072,
         index_name="danswer_chunk_openai_text_embedding_3_large",
--- a/backend/onyx/natural_language_processing/search_nlp_models.py
+++ b/backend/onyx/natural_language_processing/search_nlp_models.py
@@ -250,6 +250,25 @@ def format_embedding_error(
     )
 
 
+def _extract_cohere_embeddings(response_embeddings: Any) -> list[Embedding]:
+    """Normalize Cohere's embed response across SDK versions.
+
+    v3 models return ``response.embeddings`` as a flat ``list[list[float]]``.
+    v4 models return an ``EmbedByTypeResponseEmbeddings`` object that carries
+    the embeddings under ``float_`` (and possibly other per-type buckets when
+    ``embedding_types`` is set). We always read the float bucket here since
+    we never request alternate types.
+    """
+    if isinstance(response_embeddings, list):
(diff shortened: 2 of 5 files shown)


justin-tahara:
Addressed greptile's P2 comment in cd46a31474 — `_embed_cohere` now passes `embedding_types=["float"]` explicitly so v4's `.float_` field is guaranteed populated rather than depending on Cohere's default. The helper is left intact (still handles both shapes) so v3 callers that don't request a type continue to work.

acaprau (approved):

PR #10490, opened 2026-04-23
refactor(voice): restructure Voice config page and migrate modals to Formik

raunakab:
## Description

Restructure the Voice Configuration page into a directory with shared modal components, and migrate the setup modal to Formik:

- **Directory restructure**: Move `VoiceConfigurationPage.tsx` → `VoiceConfigurationPage/index.tsx`, co-locate `VoiceProviderSetupModal` as `shared.tsx`
- **Formik migration**: Replace manual `useState` form management with Formik + Yup validation schema, using proper `-Field` components (`PasswordInputTypeInField`, `InputTypeInField`, `InputSelectField`, `InputComboBoxField`)
- **New `InputComboBoxField`**: Add Formik-bound combo box field component following the existing `-Field` pattern
- **Modal infrastructure**: Use `useCreateModal`/`useModalClose` from `ModalContext` instead of manual open/close state — removes `onClose` props from modal components
- **Modal width**: Bump modals from `sm` (30rem) to `md` (40rem)
- **Header icons**: Use `moreIcon1`/`moreIcon2` props on `Modal.Header` instead of custom `LogoArrangement` component
- **Error handling**: Replace inline `Callout` error displays with `toast.error` notifications
- **Shared helpers**: Extract `getProviderIcon`, `getProviderLabel`, `ProviderMode` to shared module

## Screenshots + Videos

https://github.com/user-attachments/assets/25074c3d-a8e9-4814-b822-683332433d46

## Additional Options

- [ ] [Optional] Please cherry-pick this PR to the latest release version.
- [x] [Optional] Override Linear Check

---
## Summary by cubic
Restructured the Voice Configuration admin page and migrated provider modals to `formik` with shared components and per-card modal context. Adds masked API key support end-to-end, preserves defaults on edit, fixes model ID mapping, and resolves disconnect icon and cross-card modal bugs.

- **Refactors**
  - Restructured `VoiceConfigurationPage.tsx` → `VoiceConfigurationPage/index.tsx`; extracted shared helpers/modals to `shared.tsx`; created `lib/voice/{interfaces,svc,utils}` and moved API calls from `lib/admin/voice/svc.ts` to `lib/voice/svc.ts`.
  - Migrated forms to `formik` + `yup`; adopted `InputTypeInField`, `PasswordInputTypeInField`, `InputSelectField`, and new `InputComboBoxField` (voice picker); added docstring for `InputComboBoxField`; introduced `VoiceProviderType` literal type for provider params; used masked `api_key` on edit with change detection.
[...]

PR #11029, opened 2026-05-12
test(voice): verify STT works without a TTS provider (ENG-3927)

justin-tahara:
## Description

[ENG-3927](https://linear.app/onyx-app/issue/ENG-3927) asks for users to be able to use voice/dictation (speech-to-text) without configuring a text-to-speech provider.

**Audit finding:** STT and TTS are already decoupled across the stack:

- Backend: `GET /api/voice/status` returns `stt_enabled` and `tts_enabled` independently (`backend/onyx/server/manage/voice/user_api.py`).
- Admin voice page: the setup modal's `mode === "stt"` path only configures STT fields and saves the provider with `activate_tts=false` (`web/src/refresh-pages/admin/VoiceConfigurationPage/shared.tsx`).
- Chat surface: `AppInputBar.tsx:188-190` renders the mic button when `sttEnabled || isAdmin`; `MessageToolbar.tsx:287` only renders the TTS button when `ttsEnabled`; `VoiceModeProvider.tsx:164` only auto-plays when `ttsEnabled`.

Most of this decoupling landed via PR #10490 (refactor(voice): restructure Voice config page and migrate modals to Formik). The original ticket was filed 2026-04-01, before that refactor.

This PR adds a Playwright spec at `web/tests/e2e/admin/voice/stt-only.spec.ts` that locks the behaviour in:

1. **Admin voice page** — mocks `/api/admin/voice/providers` with a single OpenAI provider that has `is_default_stt=true, is_default_tts=false`. Confirms the STT card exposes a Disconnect button (proves the provider was accepted as STT-only) while the TTS card for the same provider does not (proves no TTS default was activated).
2. **Chat surface** — mocks `/api/voice/status` to return `{stt_enabled: true, tts_enabled: false}`. Sends a message and confirms the mic button is visible and the agent-message TTS playback button is absent.

If this PR's CI passes the spec, ENG-3927 can be closed with no production code change required.

## How Has This Been Tested?

- New spec compiles cleanly under `tsgo`.
- Could not run Playwright locally (the dev environment is configured with `auth_type=oidc`, which doesn't expose the `/api/auth/register` endpoint Playwright's global-setup uses to provision test users). CI runs with basic auth and will exercise the spec.

## Additional Options

---
## Summary by cubic
Add a `@playwright/test` e2e spec to confirm STT works without a TTS provider, fulfilling ENG-3927. It mocks `/api/voice/status` with `stt_enabled=true` and `tts_enabled=false`, verifies the chat mic is visible, and ensures the agent-message TTS button does not render; the admin-page check was removed as redundant.

<sup>Written for commit 6b83563af9d59b54118b7e5420026d8b5ef84090. Summary will update on new commits.</sup>

### DIFF
--- a/web/tests/e2e/admin/voice/stt-only.spec.ts
+++ b/web/tests/e2e/admin/voice/stt-only.spec.ts
@@ -1,41 +1,18 @@
 import { test, expect, Page } from "@playwright/test";
-import { loginAs, loginAsWorkerUser } from "@tests/e2e/utils/auth";
+import { loginAsWorkerUser } from "@tests/e2e/utils/auth";
 import { sendMessage } from "@tests/e2e/utils/chatActions";
 
-const ADMIN_VOICE_URL = "/admin/configuration/voice";
-const ADMIN_PROVIDERS_API = "**/api/admin/voice/providers";
 const USER_VOICE_STATUS_API = "**/api/voice/status";
 
-const OPENAI_STT_ONLY_PROVIDER = {
-  id: 1,
-  name: "openai",
-  provider_type: "openai",
-  is_default_stt: true,
-  is_default_tts: false,
-  stt_model: "whisper",
-  tts_model: null,
-  default_voice: null,
-  has_api_key: true,
-  target_uri: null,
-};
-
-async function mockAdminProviders(
-  page: Page,
-  providers: (typeof OPENAI_STT_ONLY_PROVIDER)[]
-) {
-  await page.route(ADMIN_PROVIDERS_API, async (route) => {
-    if (route.request().method() === "GET") {
-      await route.fulfill({ status: 200, json: providers });
-    } else {
-      await route.continue();
-    }
-  });
+interface VoiceStatusPayload {
+  stt_enabled: boolean;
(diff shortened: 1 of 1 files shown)


justin-tahara:
Updated in 6b83563af9 — dropped the admin-page test (it was redundant with the chat assertion and was failing on a strict-mode locator violation against the section heading vs. the provider descriptions). The chat test, which is what actually proves ENG-3927, passed on the prior run.

For greptile's feedback:
- **P2** (explicit interface for the mock helper): addressed — replaced the literal-inference type with an explicit `VoiceStatusPayload` interface.
- **P1** (Page Object Model): the failing admin test is gone, so the VoiceAdminPage POM is no longer needed. The chat test still uses a couple of raw locators (mic button `aria-label` and `AgentMessage/tts-button` testid). That's consistent with the existing `disconnect-provider.spec.ts` in the same folder, and these two locators don't have a home on `ChatPage` today. Happy to wrap them into a follow-up POM if we want to enforce the rule retroactively across the voice specs, but I'd rather not bolt that onto this regression-test PR.

raunakab (approved):

PR #11030, opened 2026-05-12
test(ingestion): assert ingestion-API docs are searchable (ENG-3387)

justin-tahara:
## Description

[ENG-3387](https://linear.app/onyx-app/issue/ENG-3387) is a customer report from January 2026: docs pushed through the ingestion API show up in storage but not in search results. The ticket says "Requires investigation" and points to a Discord thread.

**Code-audit finding:** The default cc_pair has been marked \`access_type=AccessType.PUBLIC\` since [hagen-danswer's 2024-09 commit](https://github.com/onyx-dot-app/onyx/commit/2274cab5549), and the ingestion endpoint correctly routes docs through the full indexing pipeline (\`DocumentIndexingBatchAdapter\` → \`get_access_for_documents\` → PUBLIC ACL → Vespa write). On paper the bug should be fixed.

**Why this PR is just a test:** the existing \`test_ingestion_api_crud\` covers the create/list/delete path but only verifies that documents reach Postgres and Vespa. It never issues a search query. That's exactly the gap the customer hit: a doc can be indexed perfectly and still get dropped at query time by ACL filtering or a misconfigured cc_pair.

The new \`test_ingestion_api_doc_is_searchable\` closes that gap:

1. Creates a PUBLIC cc_pair and an API key.
2. POSTs a doc containing a deterministic UUID-style phrase through \`/onyx-api/ingestion\`.
3. Sanity-checks Vespa actually received the doc.
4. Queries that same phrase through \`/search/send-search-message\` as the admin and asserts the doc surfaces in results.

If the bug is gone, this test passes and ENG-3387 can be closed as fixed-and-now-covered-by-regression. If it fails, the assertion message points at whether the breakage is in indexing (Vespa empty) or in search-time ACL/filter logic (Vespa has it, but search dropped it).

Could not run the integration suite locally (the generated OpenAPI client used by \`CCPairManager\` is built by openapi-generator-cli during the integration Docker image build; my dev env doesn't have it). The spec passes \`ty\` / \`ruff\` / \`black\` / lazy-import checks and CI will exercise it.

## How Has This Been Tested?

- Pre-commit (\`ty\`, \`ruff\`, \`black\`, lazy-import check) passes.
- CI's integration-tests (\`tests/ingestion\`) job will run the spec end-to-end against Vespa, Postgres, Redis, and the model server.

## Additional Options

---
## Summary by cubic
Add a regression test to ensure docs ingested via the API show up in search (ENG-3387). It seeds a doc with a unique phrase on a PUBLIC cc_pair, confirms Vespa wrote it, then queries the phrase and asserts the doc appears—catching ACL/cc_pair misconfigurations at query time.

<sup>Written for commit 4655aaf3aae4e03b895869b3f6f0a394bab81f57. Summary will update on new commits.</sup>

### DIFF
--- a/backend/tests/integration/tests/ingestion/test_ingestion_api.py
+++ b/backend/tests/integration/tests/ingestion/test_ingestion_api.py
@@ -5,6 +5,9 @@ from onyx.db.models import Document
 from tests.integration.common_utils.managers.api_key import APIKeyManager
 from tests.integration.common_utils.managers.cc_pair import CCPairManager
 from tests.integration.common_utils.managers.document import IngestionManager
+from tests.integration.common_utils.managers.document_search import (
+    DocumentSearchManager,
+)
 from tests.integration.common_utils.managers.user import UserManager
 from tests.integration.common_utils.test_models import DATestUser
 from tests.integration.common_utils.vespa import vespa_fixture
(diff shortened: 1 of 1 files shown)


justin-tahara:
Closing — customer confirms the original report is no longer reproducible after upgrading to v3.0. ENG-3387 is being closed as fixed-by-upgrade rather than fixed-here. If we want a permanent regression test we can re-open this PR later, but for now there's no active bug to guard against.

PR #11035, opened 2026-05-12
chore(deps): upgrade tailwindcss v3 → v4

jmelahman:
## Summary

Full migration of the web frontend from Tailwind CSS v3.4 → v4.3 using the official `@tailwindcss/upgrade` codemod plus manual cleanup.

### What changed
- **`tailwindcss` ^3.4.17 → ^4.3.0**, plus `@tailwindcss/postcss` added and `autoprefixer` removed (bundled in v4).
- **PostCSS config** swapped to use `@tailwindcss/postcss`.
- **`globals.css`** migrated from `@tailwind base/components/utilities` to `@import "tailwindcss"`, custom utilities converted from `@layer utilities` to `@utility`, default-border-color compatibility shim added.
- **~233 `.tsx` files** got mechanical class renames from the codemod (`shadow` → `shadow-sm`, `z-[20]` → `z-20`, `outline-none` → `outline-hidden`, etc.).

### Whitelabel theming preserved
`NEXT_PUBLIC_THEME` is used by the Dockerfile and all 6 docker-compose files to load a custom theme from `web/tailwind-themes/custom/<theme>/tailwind.config.js` at build time. The JS-based loader is preserved by wiring `tailwind.config.js` into v4 via `@config '../../tailwind.config.js';` in `globals.css`. No change to the customer-facing whitelabel flow.

### Plugin changes
- **Dropped** `@tailwindcss/container-queries` — native in v4.
- **Swapped** `tailwindcss-animate` (unmaintained, pre-v4) → `tw-animate-css` (v4-native drop-in). Imported via `@import "tw-animate-css"` in `globals.css` so `animate-in`/`fade-*`/`slide-*`/`zoom-*` utilities (used heavily by Radix components) keep working.
- **Kept** `@tailwindcss/typography`, `@headlessui/tailwindcss` (both declare v4 compat).
- **Bumped** `tailwind-merge` ^2.5.4 → ^3.6.0 (required for v4 class awareness).

### Opal kept publishable as a standalone package
v4 processes each CSS file independently, so every opal stylesheet using `@apply` needs an `@reference` to resolve custom theme tokens (`rounded-08`, `text-text-04`, etc.). Rather than reach into the host app's `globals.css` (which would break `@onyx-ai/opal` for external consumers), opal now has its own internal reference file:
- New `web/lib/opal/src/_reference.css` imports `tailwindcss`, `tw-animate-css`, and the opal tailwind preset.
- All 19 opal CSS files `@reference` it via relative paths.
- `bundle-css.mjs` places `_reference.css` first in the concatenated `dist/styles.css` and strips redundant per-file `@reference` lines so the published artifact is self-contained.
- Replaced `@import "@opal/core/interactive/shared.css";` with a relative path in `table/styles.css` — PostCSS doesn't resolve TypeScript path aliases.

### Other manual fixes the codemod missed
- Removed unused custom screens (`mobile`/`tall`/`short`/`very-short`) — they had no references anywhere in the codebase and v4's `@config` doesn't translate the v3 `raw: "..."` syntax correctly, producing invalid container CSS like `@media (width >= (min-height: 800px))`.

### Codemod over-reaches reverted
The codemod's string-literal renames caught two non-CSS strings by accident:
- `ValidSources.Outline = "outline"` (the Outline note-taking connector ID) → wrongly became `"outline-solid"`.
- `ScrollIndicatorDiv`'s `variant?: "gradient" | "shadow"` prop type → wrongly became `"shadow-sm"`, leaving 4 callers type-broken.

Both reverted; `npm run types:check` is green.

## Test plan
- [x] `npm run types:check` passes
- [x] `npm run build` clean (only sandbox-side Google Fonts network errors; no Tailwind/PostCSS errors)
- [x] `node lib/opal/scripts/bundle-css.mjs` produces a self-contained `dist/styles.css`
- [ ] Manual UI smoke: light + dark mode, sidebar/dropdowns/popovers (Radix animations), settings page layout, scroll indicators, custom shadows
- [ ] Storybook still builds (`npm run storybook:build`)
- [ ] Verify a custom whitelabel theme still loads via `NEXT_PUBLIC_THEME=...`


raunakab (approved):
Looks good. Ignored a lot of the actual classname changes (assuming you used a compiler there for the transformation).

Interesting that we now need to append `@reference '../../../_reference.css'; to each CSS file now. What is the purpose of that?

PR #11044, opened 2026-05-13
feat(craft): add user library sync to sandbox via S3

wenxi-onyx:
## Description

Add a new delivery mechanism for user-uploaded raw files (spreadsheets, PDFs, etc.) that the sandbox agent opens directly with Python libraries. With the file-sync sidecar removed in PR 3, user library files need a new path into the sandbox.

- Add `sync_user_library()` to `SandboxManager`: K8s implementation runs `aws s3 sync --delete` via kubectl exec, local is a no-op
- Add `sync_user_library_files_task` Celery task dispatched after user library uploads/deletions (routed to SANDBOX queue, `expires=600` on dispatch)
- Symlink `session/user_library` → `/workspace/user_library/` so the agent accesses files via relative path (same pattern as skills — no opencode external_directory allowlist changes needed)
- Call sync on session create and restore paths (wrapped in try/except so sync failures don't block session creation)
- Honor sync-disabled files: query `sync_disabled` from document metadata and remove those files from the sandbox after sync
- Local dev: symlink `sandbox/user_library` to persistent storage path so local uploads are visible
- Update `AGENTS.template.md` to document `user_library/` directory
- Add `get_disabled_user_library_paths()` to `db/user_library.py`
- Add `/workspace/user_library` to Dockerfile mkdir

## How Has This Been Tested?

- Verified Celery task routes to SANDBOX queue (matches existing `cleanup_idle_sandboxes` routing)
- Verified `aws s3 sync --delete` command correctly mirrors S3 path used by `PersistentDocumentWriter.write_raw_file()`
- Verified symlink approach works (agent accesses `user_library/` as relative path in session)
- Verified sync dispatch fires after upload_files, upload_zip, and delete_file endpoints
- Verified disabled paths are removed from sandbox after sync

## Additional Options


wenxi-onyx:
@greptileai re-review

wenxi-onyx:
@greptileai re-review

Weves:
how has this been tested?

PR #10996, opened 2026-05-11
feat(skills): Establish Skills V1 foundation schema

rohoswagger:
## Summary
- Claims the Phase 1 foundation DB/skeleton slice in docs/craft/features/TODOS.md
- Adds Skill and Skill__UserGroup ORM models plus FileOrigin.SKILL_BUNDLE
- Adds the skills Alembic revision and empty backend/onyx/skills + backend/onyx/db/skill.py skeletons

## Validation
- python -m py_compile targeted files
- ruff check targeted files
- black --check targeted files
- ty check targeted files
- imported Skill/Skill__UserGroup metadata
- alembic -x upgrade_all_tenants=true upgrade 4ff2545411ad:head --sql

## Not tested
- alembic -n schema_private upgrade head on a fresh EE tenant

---
## Summary by cubic
Establishes the Skills V1 foundation: adds the `Skill` model with group grants and soft-delete, the `Skill__UserGroup` join table, and string-backed `FileOrigin.SKILL_BUNDLE`. Adds empty `backend/onyx/skills/*` entrypoints and a full PRD with HTML mockups; uses `author_user_id` and a partial unique slug index.

- **New Features**
  - `Skill` schema with bundle metadata, `author_user_id`, soft-delete, and group-only grants; `Skill__UserGroup` join plus `Skill.groups` and `UserGroup.skills`.
  - Empty `backend/onyx/skills/*` and `backend/onyx/db/skill.py`.
  - Docs: Skills V1 PRD + HTML mockups and sandbox file sync design; reorganizes skills docs under `docs/craft/features/skills/`.

- **Migration**
  - Single Alembic migration creates `skill` and `skill__user_group`; adds partial unique `ux_skill_slug` on active rows (`WHERE deleted_at IS NULL`).
  - Regenerated from current main to keep one head; no enum DDL (string-backed `FileOrigin.SKILL_BUNDLE`).
  - Supports multi-tenant upgrades via `-x upgrade_all_tenants=true`.

<sup>Written for commit faad0acddecbf18c13bbaef37d518f85753d816a. Summary will update on new commits.</sup>

## Additional Options

- [ ] [Optional] Please cherry-pick this PR to the latest release version.
- [x] [Optional] Override Linear Check

PR #11059, opened 2026-05-13
feat(skills): bundle validator for custom skill uploads

rohoswagger:
## Description

Implements Phase 1.4 of Skills V1 (spec §5 in `docs/craft/features/skills/skills_plan.md`): the synchronous validator for admin-uploaded custom skill bundles.

New module `backend/onyx/skills/bundle.py`:

- `InvalidBundleError(OnyxError)` — clean propagation through the global FastAPI exception handler.
- `validate_custom_bundle(zip_bytes, slug, *, reserved_slugs=None, …) -> ManifestMetadata` — zip parse, root `SKILL.md` check, no `*.template`, path-traversal + symlink rejection, streaming per-file (25 MiB) + total (100 MiB) uncompressed caps, slug regex, reserved-slug check, optional YAML frontmatter capture.
- `_safe_unzip(zip_bytes, dest)` — defense-in-depth re-check for the materializer at extraction time.
- `compute_bundle_sha256(zip_bytes)` — raw-bytes hash to detect re-uploads.
- `ManifestMetadata` / `ManifestFileEntry` / `ManifestFrontmatter` Pydantic models.

Covers TODOs **P1.030-P1.041** in `docs/craft/features/skills/TODOS.md`, except **P1.035** (reserved-slug check via `BuiltinSkillRegistry.reserved_slugs()`) which is held back pending the in-flight registry work. The validator already accepts a `reserved_slugs` parameter the API layer will populate once the registry lands.

## How Has This Been Tested?

`backend/tests/unit/onyx/skills/test_bundle.py` — 26 unit tests, one per validation rule plus happy paths, sha256 determinism, and `_safe_unzip` defense-in-depth checks:

```
uv run pytest backend/tests/unit/onyx/skills/test_bundle.py
============================== 26 passed in 0.09s ==============================
```

## Additional Options

### DIFF
--- a/backend/onyx/skills/bundle.py
+++ b/backend/onyx/skills/bundle.py
@@ -11,18 +11,14 @@ import re
 import stat
 import zipfile
 from pathlib import Path
-from typing import Any
 from typing import Final
 
-import yaml
 from pydantic import BaseModel
 from pydantic import ConfigDict
 from pydantic import Field
 
 from onyx.error_handling.error_codes import OnyxErrorCode
 from onyx.error_handling.exceptions import OnyxError
-
-VALIDATOR_VERSION: Final[int] = 1
 
 DEFAULT_PER_FILE_MAX_BYTES: Final[int] = 25 * 1024 * 1024
 DEFAULT_TOTAL_MAX_BYTES: Final[int] = 100 * 1024 * 1024
@@ -35,18 +31,6 @@ SLUG_REGEX: Final[re.Pattern[str]] = re.compile(r"^[a-z][a-z0-9-]{0,63}$")
 _ZIP_UNIX_CREATE_SYSTEM: Final[int] = 3
 
 
-class InvalidBundleError(OnyxError):
-    """Raised when a custom skill bundle fails validation."""
-
-    def __init__(
-        self,
-        detail: str,
-        *,
-        error_code: OnyxErrorCode = OnyxErrorCode.INVALID_INPUT,
-    ) -> None:
-        super().__init__(error_code, detail)
-
-
 class ManifestFileEntry(BaseModel):
     model_config = ConfigDict(frozen=True)
 
@@ -55,37 +39,41 @@ class ManifestFileEntry(BaseModel):
 
 
 class ManifestMetadata(BaseModel):
-    """Bundle metadata captured at validation time. Persisted as JSONB on
-    the ``Skill`` row.
+    """Bundle-content inventory captured at validation time, persisted as
+    JSONB on the ``Skill`` row.
 
-    ``name`` and ``description`` come from the bundle's optional YAML
-    frontmatter and are pre-fill hints only — the admin-typed values on the
-    ``Skill`` row are authoritative.
+    Only carries bundle facts that the ``Skill`` row itself doesn't already
+    have — the file list and total size, for admin-UI surfacing. Frontmatter
+    name/description live on the row; admins type them into the upload form
+    (with client-side pre-fill via jszip per P4.021).
     """
 
     model_config = ConfigDict(frozen=True)
 
-    name: str | None = None
-    description: str | None = None
     files: list[ManifestFileEntry] = Field(default_factory=list)
     total_uncompressed_bytes: int = 0
-    validator_version: int = VALIDATOR_VERSION
 
 
 def _check_slug(slug: str) -> None:
     if not SLUG_REGEX.match(slug):
-        raise InvalidBundleError(f"invalid slug '{slug}'")
+        raise OnyxError(OnyxErrorCode.INVALID_INPUT, f"invalid slug '{slug}'")
 
 
 def _is_symlink(info: zipfile.ZipInfo) -> bool:
-    """True if the zip entry was archived as a Unix symlink."""
+    """True if the zip entry was archived as a Unix symlink.
+
+    We inspect the zip-entry metadata (``external_attr`` mode bits) rather
+    than ``Path.is_symlink()`` because at validation time nothing has been
+    extracted to disk yet — and the whole point of the check is to refuse
+    to extract.
+    """
     if info.create_system != _ZIP_UNIX_CREATE_SYSTEM:
         return False
     unix_mode = (info.external_attr >> 16) & 0xFFFF
     return stat.S_ISLNK(unix_mode)
 
 
-def _normalize_zip_path(name: str) -> str:
+def _check_zip_entry_path(name: str) -> str:
     """Reject path-traversal entries; return a clean relative posix path.
 
     A zip-bomb-style entry like ``../../etc/passwd`` or ``/etc/passwd`` must
@@ -93,53 +81,22 @@ def _normalize_zip_path(name: str) -> str:
     """
     trimmed = name.rstrip("/")
     if not trimmed:
-        raise InvalidBundleError(f"bundle entry has empty path: '{name}'")
+        raise OnyxError(
+            OnyxErrorCode.INVALID_INPUT,
+            f"bundle entry has empty path: '{name}'",
+        )
     if trimmed.startswith("/") or "\\" in trimmed:
-        raise InvalidBundleError(f"bundle entry escapes root: '{name}'")
+        raise OnyxError(
+            OnyxErrorCode.INVALID_INPUT,
+            f"bundle entry escapes root: '{name}'",
+        )
     parts = trimmed.split("/")
     if any(p in ("", ".", "..") for p in parts):
-        raise InvalidBundleError(f"bundle entry escapes root: '{name}'")
+        raise OnyxError(
+            OnyxErrorCode.INVALID_INPUT,
+            f"bundle entry escapes root: '{name}'",
+        )
     return trimmed
-
-
-def _parse_frontmatter(skill_md_bytes: bytes) -> tuple[str | None, str | None]:
-    """Best-effort YAML frontmatter extraction.
-
-    Returns ``(name, description)``; either or both may be ``None`` when the
-    frontmatter is missing, malformed, or unparseable.
-    """
-    empty: tuple[str | None, str | None] = (None, None)
-    try:
-        text = skill_md_bytes.decode("utf-8")
-    except UnicodeDecodeError:
-        return empty
-
-    if not text.startswith("---"):
-        return empty
-
-    lines = text.split("\n")
-    end_idx: int | None = None
-    for i in range(1, len(lines)):
-        if lines[i].strip() == "---":
-            end_idx = i
-            break
-    if end_idx is None:
-        return empty
-
-    yaml_text = "\n".join(lines[1:end_idx])
-    try:
-        parsed: Any = yaml.safe_load(yaml_text)
-    except yaml.YAMLError:
-        return empty
-    if not isinstance(parsed, dict):
-        return empty
-
-    name = parsed.get("name")
-    description = parsed.get("description")
-    return (
-        str(name) if name is not None else None,
-        str(description) if description is not None else None,
-    )
 
 
 def validate_custom_bundle(
@@ -155,9 +112,11 @@ def validate_custom_bundle(
     Args:
         zip_bytes: Raw zip bytes uploaded by an admin.
         slug: Caller-supplied slug for this skill.
-        reserved_slugs: Slugs registered as built-ins (rejected here).
-            Pass ``BuiltinSkillRegistry.instance().reserved_slugs()`` from
-            the API layer.
+        reserved_slugs: Slugs registered as built-ins (rejected here). Caller
+            threads in ``BuiltinSkillRegistry.instance().reserved_slugs()``;
+            we don't call it directly because the registry is a separate
+            in-flight task on this stack and we don't want the import edge.
+            Consolidate once the registry lands on skills-phase-1.
         per_file_max_bytes: Per-entry uncompressed cap.
         total_max_bytes: Total uncompressed cap.
 
@@ -165,32 +124,40 @@ def validate_custom_bundle(
         ManifestMetadata to persist on the ``Skill`` row.
 
     Raises:
-        InvalidBundleError: any rule from spec §5 fails.
+        OnyxError(INVALID_INPUT): structural violations (bad slug, missing
+            SKILL.md, traversal, symlink, template, unreadable entry).
+        OnyxError(PAYLOAD_TOO_LARGE): per-file or total size cap exceeded.
     """
     _check_slug(slug)
     if reserved_slugs and slug in reserved_slugs:
-        raise InvalidBundleError(f"slug '{slug}' is reserved")
+        raise OnyxError(OnyxErrorCode.INVALID_INPUT, f"slug '{slug}' is reserved")
 
     try:
         zf = zipfile.ZipFile(io.BytesIO(zip_bytes))
     except zipfile.BadZipFile:
-        raise InvalidBundleError("bundle is not a valid zip")
+        raise OnyxError(OnyxErrorCode.INVALID_INPUT, "bundle is not a valid zip")
 
     with zf:
         files: list[ManifestFileEntry] = []
         total = 0
-        skill_md_info: zipfile.ZipInfo | None = None
+        saw_skill_md = False
 
         for info in zf.infolist():
             if info.is_dir():
-                _normalize_zip_path(info.filename)
+                _check_zip_entry_path(info.filename)
                 continue
 
-            normalized = _normalize_zip_path(info.filename)
+            normalized = _check_zip_entry_path(info.filename)
             if _is_symlink(info):
-                raise InvalidBundleError(f"bundle contains a symlink: '{normalized}'")
+                raise OnyxError(
+                    OnyxErrorCode.INVALID_INPUT,
+                    f"bundle contains a symlink: '{normalized}'",
+                )
             if normalized.endswith(TEMPLATE_SUFFIX):
-                raise InvalidBundleError("custom skills cannot ship templates")
+                raise OnyxError(
+                    OnyxErrorCode.INVALID_INPUT,
+                    "custom skills cannot ship templates",
+                )
 
             size = 0
             try:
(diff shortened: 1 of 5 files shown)


backend/onyx/skills/bundle.py
```diff
@@ -1 +1,267 @@
...
+                        )
+
+            files.append(ManifestFileEntry(path=normalized, size=size))
+            if normalized == SKILL_MD_NAME:
+                skill_md_info = info
+
+        if skill_md_info is None:
+            raise InvalidBundleError("SKILL.md missing at bundle root")
+
+        frontmatter = _parse_frontmatter(zf.read(skill_md_info))
```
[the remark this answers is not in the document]
rohoswagger: Fixed in 1121d12. Both `zf.open()` (per-entry read loop, line ~190) and `zf.read(skill_md_info)` (frontmatter read, line ~222) are now wrapped: `except InvalidBundleError: raise` then `except Exception as exc: raise InvalidBundleError(f"cannot read '{normalized}': {exc}") from exc`. `_safe_unzip` got the same treatment for its extraction `zf.open()`. New test `test_rejects_unsupported_compression` constructs a zip with a patched compression-method byte that triggers `NotImplementedError` from `zf.open()` and verifies we surface `InvalidBundleError` (HTTP 400), not a 500. Parallel test for `_safe_unzip`.

backend/onyx/skills/bundle.py
```diff
@@ -1 +1,267 @@
...
+            if info.is_dir():
+                target.mkdir(parents=True, exist_ok=True)
+                continue
+            target.parent.mkdir(parents=True, exist_ok=True)
+            with zf.open(info, mode="r") as src, open(target, "wb") as out:
+                while True:
+                    chunk = src.read(64 * 1024)
+                    if not chunk:
+                        break
+                    out.write(chunk)
```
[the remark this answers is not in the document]
rohoswagger: Fixed in 1121d12. `_safe_unzip` now runs the same streaming per-file (default 25 MiB) and total (default 100 MiB) counters as `validate_custom_bundle`, raising `InvalidBundleError(..., error_code=PAYLOAD_TOO_LARGE)` mid-write before any oversized data lands on disk. Signature gained `per_file_max_bytes` and `total_max_bytes` keyword args matching the validator. New tests `test_safe_unzip_enforces_per_file_cap` and `test_safe_unzip_enforces_total_cap` cover both paths.

backend/onyx/skills/bundle.py
```diff
@@ -1 +1,267 @@
...
+SLUG_REGEX: Final[re.Pattern[str]] = re.compile(r"^[a-z][a-z0-9-]{0,63}$")
+
+_ZIP_UNIX_CREATE_SYSTEM: Final[int] = 3
+
+
+class InvalidBundleError(OnyxError):
+    """Raised when a custom skill bundle fails validation."""
+
+    def __init__(self, detail: str) -> None:
+        super().__init__(OnyxErrorCode.INVALID_INPUT, detail)
```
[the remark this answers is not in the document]
rohoswagger: Adopted in 1121d12. `InvalidBundleError.__init__` now takes an optional `error_code` kwarg (default still `INVALID_INPUT`). The four size-cap raise sites in `validate_custom_bundle` and `_safe_unzip` pass `error_code=OnyxErrorCode.PAYLOAD_TOO_LARGE`, which maps to HTTP 413. New tests `test_size_violation_returns_413` and `test_validation_violation_returns_400` pin the mapping.

backend/onyx/skills/bundle.py
```diff
@@ -1 +1,267 @@
...
+                target.relative_to(dest_resolved)
+            except ValueError:
+                raise InvalidBundleError(
+                    f"bundle entry escapes root: '{info.filename}'"
+                )
+            if info.is_dir():
+                target.mkdir(parents=True, exist_ok=True)
+                continue
+            target.parent.mkdir(parents=True, exist_ok=True)
+            with zf.open(info, mode="r") as src, open(target, "wb") as out:
```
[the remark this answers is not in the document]
rohoswagger: Fixed in 1121d12. `_safe_unzip` now enforces per-file (25 MiB) and total (100 MiB) uncompressed caps inside the write loop — the streaming counter aborts mid-write and the partial output is left for the caller to clean up. Size violations raise `InvalidBundleError(..., error_code=PAYLOAD_TOO_LARGE)`. Defense-in-depth claim in the docstring now matches reality. New tests `test_safe_unzip_enforces_per_file_cap` and `test_safe_unzip_enforces_total_cap`.

backend/onyx/skills/bundle.py
```diff
@@ -1 +1,267 @@
+"""Custom skill bundle validation and helpers.
```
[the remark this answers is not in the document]
rohoswagger: Fixed in 1121d12. All `zf.open()`/`zf.read()` call sites in both `validate_custom_bundle` and `_safe_unzip` are wrapped to translate `NotImplementedError` (unsupported compression methods like Zstandard/Implode), `zlib.error`/`lzma.LZMAError` (corrupt streams), and any other read-time failure into `InvalidBundleError` (HTTP 400) — no more 500s on malformed bundles. Tests `test_rejects_unsupported_compression` and `test_safe_unzip_rejects_unsupported_compression` exercise the path with a zip whose compression-method byte is patched to a stdlib-unknown value.

backend/onyx/skills/bundle.py
```diff
@@ -1 +1,313 @@
...
+
+SKILL_MD_NAME: Final[str] = "SKILL.md"
+TEMPLATE_SUFFIX: Final[str] = ".template"
+
+SLUG_REGEX: Final[re.Pattern[str]] = re.compile(r"^[a-z][a-z0-9-]{0,63}$")
+
+_ZIP_UNIX_CREATE_SYSTEM: Final[int] = 3
+
+
+class InvalidBundleError(OnyxError):
```
wenxi-onyx: Should be organized with other OnyxErrors I think. Should this just be whatever Onyx equivalent of value error? Do we need a custom Onyx Error for this?

backend/onyx/skills/bundle.py
```diff
@@ -1 +1,321 @@
...
+    total_uncompressed_bytes: int = 0
+    validator_version: int = VALIDATOR_VERSION
+
+
+def _check_slug(slug: str) -> None:
+    if not SLUG_REGEX.match(slug):
+        raise InvalidBundleError(f"invalid slug '{slug}'")
+
+
+def _is_symlink(info: zipfile.ZipInfo) -> bool:
```
wenxi-onyx, line 37: does pathlib:Path.is_symlink() not work?

backend/onyx/skills/bundle.py
```diff
@@ -1 +1,321 @@
...
+        raise InvalidBundleError(f"bundle entry has empty path: '{name}'")
+    if trimmed.startswith("/") or "\\" in trimmed:
+        raise InvalidBundleError(f"bundle entry escapes root: '{name}'")
+    parts = trimmed.split("/")
+    if any(p in ("", ".", "..") for p in parts):
+        raise InvalidBundleError(f"bundle entry escapes root: '{name}'")
+    return trimmed
+
+
+def _parse_frontmatter(skill_md_bytes: bytes) -> tuple[str | None, str | None]:
```
wenxi-onyx: This should be shared with the other front matter parsing util for built in skills

backend/onyx/skills/bundle.py
```diff
@@ -1 +1,321 @@
...
+        str(name) if name is not None else None,
+        str(description) if description is not None else None,
+    )
+
+
+def validate_custom_bundle(
+    zip_bytes: bytes,
+    slug: str,
+    *,
+    reserved_slugs: set[str] | None = None,
```
wenxi-onyx: This probably shouldn't need to be passed as an arg? cant we just hit the registry inside this func?

backend/onyx/skills/bundle.py
```diff
@@ -1 +1,321 @@
...
+        raise InvalidBundleError("bundle is not a valid zip")
+
+    with zf:
+        total = 0
+        for info in zf.infolist():
+            if _is_symlink(info):
+                raise InvalidBundleError(
+                    f"bundle contains a symlink: '{info.filename}'"
+                )
+            normalized = _normalize_zip_path(info.filename)
```
wenxi-onyx: nit: if this is normalizing file paths, it should be renamed for clarity. Also could be good to have it as a shared util with other file processing like file connector

backend/onyx/skills/bundle.py
```diff
@@ -1 +1,321 @@
...
+
+
+class ManifestFileEntry(BaseModel):
+    model_config = ConfigDict(frozen=True)
+
+    path: str
+    size: int
+
+
+class ManifestMetadata(BaseModel):
```
wenxi-onyx: I don't understand why this exists/why it's persisted to the DB. Is any of this info accessed later? Especially name and description, since we already store those

backend/onyx/skills/bundle.py
```diff
@@ -1 +1,321 @@
...
+    ``Skill`` row are authoritative.
+    """
+
+    model_config = ConfigDict(frozen=True)
+
+    name: str | None = None
+    description: str | None = None
+    files: list[ManifestFileEntry] = Field(default_factory=list)
+    total_uncompressed_bytes: int = 0
+    validator_version: int = VALIDATOR_VERSION
```
wenxi-onyx: Why store validator version? Do we increment this?

wenxi-onyx (changes requested):

backend/onyx/skills/bundle.py
rohoswagger: Agreed — there are zero other `OnyxError` subclasses in the codebase and `CLAUDE.md` explicitly says to raise `OnyxError` directly. Dropped the subclass in 3214b2f. Validator now raises `OnyxError(OnyxErrorCode.INVALID_INPUT, ...)` for structural problems and `OnyxError(OnyxErrorCode.PAYLOAD_TOO_LARGE, ...)` for size-cap violations. Tests catch `OnyxError` directly.

backend/onyx/skills/bundle.py
rohoswagger: `Path.is_symlink()` checks a filesystem path — but at validation time nothing has been extracted to disk yet, and the whole point of this check is to refuse to extract. The check has to run against the zip entry's metadata (`external_attr` Unix mode bits) before any IO. Added a docstring clarifying this in 3214b2f. If we used `Path.is_symlink()` we'd have already created the symlink and (worst case) lost the security guarantee.

backend/onyx/skills/bundle.py
rohoswagger: Sidestepped this in 3214b2f by deleting `_parse_frontmatter` outright — see the reply on the `ManifestMetadata` comment below. Frontmatter pre-fill happens client-side via `jszip` per P4.021, so there's no server-side consumer left.

(For the record: the existing `extract_skill_description` in `agent_instructions.py:120` is also slated for deletion in Phase 3 — see P3.080–P3.085, which rip out the whole `{{AVAILABLE_SKILLS_SECTION}}` inlining flow. Wasn't a great consolidation target.)

backend/onyx/skills/bundle.py
rohoswagger: Agreed in principle, deferring for now: the registry (`BuiltinSkillRegistry`) is a parallel in-flight task on this stack (P1.020–P1.027, `@codex-charged-perovskite`) and `backend/onyx/skills/registry.py` is still empty on `origin/skills-phase-1`. If I import it now my unit tests fail at import time on this branch. Will consolidate to a direct `BuiltinSkillRegistry.instance().reserved_slugs()` call once the registry lands; in the meantime the parameter is documented as "caller threads in the registry's reserved set."

backend/onyx/skills/bundle.py
rohoswagger: Renamed to `_check_zip_entry_path` in 3214b2f — you're right that it's an assertion, not normalization (it raises on unsafe entries and returns a cleaned name only on success).

Leaving "share with file connector" as a follow-up — I didn't survey the file-connector ingestion paths to know what its current zip-handling looks like, and don't want to ship a shared util without first looking at whether the rules are actually the same. Happy to file a refactor ticket if you have a pointer to the closest analog there.

backend/onyx/skills/bundle.py
rohoswagger: Trimmed hard in 3214b2f. After the trim, `ManifestMetadata` is just:

```python
class ManifestMetadata(BaseModel):
    files: list[ManifestFileEntry]  # path + size per entry
    total_uncompressed_bytes: int
```

- **`name` / `description`**: dropped. They duplicated the authoritative copies on the `Skill` row, and the upload form's pre-fill happens client-side via `jszip` (P4.021), so the server never needed them.
- **`files` / `total_uncompressed_bytes`**: kept. These are bundle-content facts the `Skill` row doesn't have — they let the admin UI render "this bundle contains 5 files totaling 234 KiB" without re-reading the blob from S3 on every page load.

Open to dropping the rest too if you don't think that admin-UI surface is worth the JSONB column — I can swap it for `Skill.bundle_file_count` + `Skill.bundle_total_bytes` columns instead, which would be cheaper still.

backend/onyx/skills/bundle.py
rohoswagger: Accepted — dropped in 3214b2f. The spec called for it as forward-compat, but it was speculative until we actually have a v2 validator. If we ever need to distinguish, we'll add the field at that point with a migration that defaults pre-v2 rows to 1.

backend/onyx/skills/bundle.py
rohoswagger: Done in 861ac31 — dropped `ManifestMetadata` entirely, removed `manifest_metadata` from the `Skill` model, and (rather than ship an add-then-drop migration pair) edited the original `b6d184cfdaf3_skills.py` to not create the column in the first place since #10996 has not deployed yet. `validate_custom_bundle` now returns `None` — it either succeeds or raises `OnyxError`. If we ever need file count / bundle bytes for the admin UI we will add concrete columns on `Skill` at that point.

backend/onyx/skills/bundle.py
rohoswagger: Done in 861ac31. Moved this PR base from `skills-phase-1` onto `skills-1` so the registry is on the same stack, dropped the `reserved_slugs` kwarg from `validate_custom_bundle`, and call `BuiltinSkillRegistry.instance().reserved_slugs()` directly inside the function. Test uses `monkeypatch` on the registry singleton.

rohoswagger:
@greptile-ai @cubic-dev-ai review this

wenxi-onyx (approved):

backend/onyx/skills/bundle.py
```diff
@@ -1 +1,246 @@
...
+                        if total > total_max_bytes:
+                            raise OnyxError(
+                                OnyxErrorCode.PAYLOAD_TOO_LARGE,
+                                f"bundle exceeds "
+                                f"{total_max_bytes // (1024 * 1024)} MiB uncompressed",
+                            )
+                        out.write(chunk)
+            except OnyxError:
+                raise
+            except Exception as exc:
```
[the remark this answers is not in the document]
rohoswagger: Fixed in 2e2cc09. On any mid-extraction failure in `_safe_unzip` (size cap, unreadable entry, mkdir failure, anything), the outer `try` wraps the whole extraction loop and `shutil.rmtree(dest, ignore_errors=True)` wipes the half-extracted tree before re-raising. New test `test_safe_unzip_cleans_dest_on_size_cap_failure` constructs a bundle whose second file tips us past the per-file cap, asserts `OnyxError`, and asserts `out.exists() is False`. Parallel test `test_safe_unzip_cleans_dest_on_unreadable_entry` for the corrupt-compression path.

backend/onyx/skills/bundle.py
```diff
@@ -1 +1,246 @@
...
+                                f"{per_file_max_bytes // (1024 * 1024)} MiB",
+                            )
+                        total += len(chunk)
+                        if total > total_max_bytes:
+                            raise OnyxError(
+                                OnyxErrorCode.PAYLOAD_TOO_LARGE,
+                                f"bundle exceeds "
+                                f"{total_max_bytes // (1024 * 1024)} MiB uncompressed",
+                            )
+            except OnyxError:
```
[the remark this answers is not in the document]
rohoswagger: Fixed in 2e2cc09 — same fix as the greptile P1 above. `_safe_unzip` now wraps the entire extraction in a try-block and `shutil.rmtree(dest, ignore_errors=True)`s on any failure, so a failed extraction never leaves truncated files behind. Tests cover both the size-cap-mid-write and corrupt-entry paths.

backend/onyx/skills/bundle.py
```diff
@@ -1 +1,246 @@
...
+        zf = zipfile.ZipFile(io.BytesIO(zip_bytes))
+    except zipfile.BadZipFile:
+        raise OnyxError(OnyxErrorCode.INVALID_INPUT, "bundle is not a valid zip")
+
+    with zf:
+        total = 0
+        saw_skill_md = False
+
+        for info in zf.infolist():
+            if info.is_dir():
```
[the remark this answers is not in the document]
rohoswagger: Fixed in 2e2cc09. Added `_mkdir_or_raise(path)` that wraps `path.mkdir(parents=True, exist_ok=True)` and translates `OSError` to `OnyxError(INVALID_INPUT, "cannot create ...")`. All three mkdir call sites in `_safe_unzip` (dest root, intermediate parent dirs, explicit directory entries) now go through it. New test `test_safe_unzip_wraps_mkdir_oserror` monkeypatches `Path.mkdir` to raise `PermissionError` and asserts the validator surfaces `OnyxError` instead of a raw 500.

backend/onyx/skills/bundle.py
```diff
@@ -1 +1,246 @@
...
+    return trimmed
+
+
+def validate_custom_bundle(
+    zip_bytes: bytes,
+    slug: str,
+    *,
+    per_file_max_bytes: int = DEFAULT_PER_FILE_MAX_BYTES,
+    total_max_bytes: int = DEFAULT_TOTAL_MAX_BYTES,
+) -> None:
```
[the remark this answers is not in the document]
rohoswagger: Reading stale context — the PR description and TODOS line that say `frontmatter parse` + `-> ManifestMetadata` were both revd during review. Per wenxi-onyx (#11059), `ManifestMetadata` was dropped entirely (commit `d44f46a065`), the `manifest_metadata` column was removed from the `Skill` model, and the original migration (#10996) was edited to never create it.

`validate_custom_bundle` intentionally does **not** populate `Skill.name` / `Skill.description`. Those are admin-typed values from the upload form; client-side `jszip` parses frontmatter and pre-fills the form fields per P4.021. The validators job is just "does this zip pass our safety rules" — passing/failing, no metadata return.

So no NULL violations at insert: the API endpoint takes `(zip_bytes, slug, name, description, …)` as multipart fields and inserts `Skill` rows directly. Will tidy the PR description.

PR #11060, opened 2026-05-13
feat(skills): DB ops for custom skills (P1.060-P1.068)

rohoswagger:
## Description

Implements the Phase 1.6 DB ops layer for the Skills V1 universal primitive in `backend/onyx/db/skill.py`. This is the persistence-helper slice claimed on the shared task board ([`docs/craft/features/skills/TODOS.md`](../blob/skills-2/docs/craft/features/skills/TODOS.md)) as `P1.060-P1.068`.

Helpers added:

- `list_skills_for_user` / `fetch_skill_for_user` — `enabled = True AND deleted_at IS NULL`, plus `is_public` OR group-grant visibility via `_add_user_visibility_filter` (mirrors `onyx.db.persona._add_user_filters` minus the V1.5 direct-user-grant branch).
- `list_skills_for_admin` / `fetch_skill_for_admin` — `deleted_at IS NULL` only, so admins can still surface and re-enable disabled skills.
- `create_skill` and `patch_skill` — pre-check slug collisions against non-deleted rows and raise `OnyxError(DUPLICATE_RESOURCE)` so callers get a structured error instead of an IntegrityError from the partial unique index.
- `replace_skill_bundle` — returns `(skill, old_bundle_file_id)` so the API layer can delete the old blob AFTER commit.
- `replace_skill_grants` — dedups and re-issues join rows in a single transaction.
- `delete_skill` — soft-delete via `deleted_at = func.now()`. The sweep (`P5.031-P5.033`) handles blob cleanup + hard delete after the retention window.

None of the helpers commit — callers own the transaction so multi-step admin flows (create row + replace grants) can roll back atomically.

Stacked on `skills-base` (#11058).

## How Has This Been Tested?

- Smoke import (`uv run python -c "from onyx.db.skill import …"`) — passes against the venv.
- Pre-commit (`ruff`, `black`, `ty`, `lazy-imports`) — all pass on the new file.
- Functional coverage deferred to the Phase 2 API external-dependency tests (`P2.032-P2.036`), which exercise these helpers end-to-end through the admin and user routers. Phase 1.6 itself does not list test tasks on the board.

## Additional Options

---

**Stack:**
1. [skills-base #11058](https://github.com/onyx-dot-app/onyx/pull/11058)

---
## Summary by cubic
Implements Phase 1.6 DB helpers for custom skills with user/admin visibility, user reads filtered to enabled only, and hard-delete. Removes `deleted_at`, replaces the partial slug index with `UniqueConstraint` (`uq_skill_slug`), and marks P1.060–P1.068 as REVIEW in docs.

- **New Features**
  - Read helpers: users see enabled + public or group-granted; admins see all.
  - `create_skill` / `patch_skill`: pre-check slug collisions; `patch_skill` uses `UnsetType` from `onyx.db.utils`.
  - `replace_skill_bundle`: swaps bundle and returns old `bundle_file_id` for post-commit cleanup.
  - `replace_skill_grants`: dedups and replaces group grants; raises `NOT_FOUND` when the skill is missing.
  - Return `CustomSkill` views; helpers never commit—callers own the transaction.

- **Bug Fixes**
  - Translate only slug-unique `IntegrityError` (`uq_skill_slug`) into `DUPLICATE_RESOURCE` via `is_unique_violation`; others bubble up.
  - Dropped SAVEPOINT usage in create/patch paths.

<sup>Written for commit cfdc5c932bd52be3578afdd6cadc714d842eee91. Summary will update on new commits.</sup>

### DIFF
--- a/backend/onyx/db/skill.py
+++ b/backend/onyx/db/skill.py
@@ -1,1 +1,268 @@
+"""DB operations for custom (admin-uploaded) skills.
 
+Access model:
+- Admin reads: filter `deleted_at IS NULL` only. Disabled skills stay visible
+  so admins can re-enable them.
+- User reads: filter `enabled = True AND deleted_at IS NULL`, plus `is_public`
+  OR the user is in a group that has been granted access.
+
+These helpers never commit — callers control the transaction boundary so a
+multi-step admin flow (e.g. create row + replace grants) can roll back atomically.
+"""
+
+from collections.abc import Sequence
+from typing import Any
+from typing import Final
+from uuid import UUID
+
+from sqlalchemy import delete
+from sqlalchemy import func
+from sqlalchemy import or_
+from sqlalchemy import Select
+from sqlalchemy import select
+from sqlalchemy.orm import Session
+
+from onyx.auth.schemas import UserRole
+from onyx.db.models import Skill
+from onyx.db.models import Skill__UserGroup
+from onyx.db.models import User
+from onyx.db.models import User__UserGroup
+from onyx.error_handling.error_codes import OnyxErrorCode
+from onyx.error_handling.exceptions import OnyxError
+
+_UNSET: Final[Any] = object()
+
+
+def _add_user_visibility_filter(
+    stmt: Select[tuple[Skill]], user: User
+) -> Select[tuple[Skill]]:
+    """Restrict a `select(Skill)` to rows the given user can see.
+
+    Mirrors `onyx.db.persona._add_user_filters` minus the direct user-grant
+    branch (no `Skill__User` table in V1). Admins bypass the filter; everyone
+    else sees `is_public=True` OR membership in a group that holds a grant.
+    """
+    if user.role == UserRole.ADMIN:
+        return stmt
+
+    group_grant_exists = (
+        select(Skill__UserGroup.skill_id)
+        .join(
+            User__UserGroup,
+            User__UserGroup.user_group_id == Skill__UserGroup.user_group_id,
+        )
+        .where(Skill__UserGroup.skill_id == Skill.id)
+        .where(User__UserGroup.user_id == user.id)
+        .exists()
+    )
+
+    return stmt.where(or_(Skill.is_public.is_(True), group_grant_exists))
+
+
+def list_skills_for_user(user: User, db_session: Session) -> Sequence[Skill]:
+    """Skills the user can use in a session.
+
+    Filtered to `enabled = True AND deleted_at IS NULL`: disabled or soft-deleted
+    skills never reach the materializer.
+    """
+    stmt = (
+        select(Skill)
+        .where(Skill.enabled.is_(True))
+        .where(Skill.deleted_at.is_(None))
+        .order_by(Skill.name)
+    )
+    stmt = _add_user_visibility_filter(stmt, user)
+    return db_session.scalars(stmt).all()
+
+
+def fetch_skill_for_user(
+    skill_id: UUID, user: User, db_session: Session
+) -> Skill | None:
+    """Single-skill lookup with the same filter as `list_skills_for_user`.
+
+    Returns None when the skill does not exist, is disabled, soft-deleted, or
+    the user has no grant — callers translate to 404 as needed.
+    """
+    stmt = (
+        select(Skill)
+        .where(Skill.id == skill_id)
+        .where(Skill.enabled.is_(True))
+        .where(Skill.deleted_at.is_(None))
+    )
+    stmt = _add_user_visibility_filter(stmt, user)
+    return db_session.scalars(stmt).one_or_none()
+
+
+def fetch_skill_for_admin(skill_id: UUID, db_session: Session) -> Skill | None:
+    """Admin lookup: `deleted_at IS NULL` only (no `enabled` filter)."""
+    stmt = select(Skill).where(Skill.id == skill_id).where(Skill.deleted_at.is_(None))
+    return db_session.scalars(stmt).one_or_none()
+
+
+def list_skills_for_admin(db_session: Session) -> Sequence[Skill]:
+    """All non-soft-deleted skills, for the admin UI.
+
+    Disabled skills are included so the admin can re-enable them; soft-deleted
+    rows are hidden by default (engineer-only undelete bypasses this helper).
+    """
+    stmt = select(Skill).where(Skill.deleted_at.is_(None)).order_by(Skill.name)
+    return db_session.scalars(stmt).all()
+
+
+def create_skill(
+    *,
+    slug: str,
+    name: str,
+    description: str,
+    bundle_file_id: str,
+    bundle_sha256: str,
+    manifest_metadata: dict[str, Any],
+    is_public: bool,
+    author_user_id: UUID | None,
+    db_session: Session,
+) -> Skill:
+    """Insert a new Skill row.
+
+    Raises `OnyxError(DUPLICATE_RESOURCE)` if a non-deleted skill with the
+    same slug already exists. The partial unique index on `slug WHERE
+    deleted_at IS NULL` enforces this at the DB layer as a backstop; this
+    helper checks first so the caller gets a clean structured error rather
+    than an IntegrityError.
+    """
+    existing = db_session.scalars(
+        select(Skill.id).where(Skill.slug == slug).where(Skill.deleted_at.is_(None))
+    ).first()
+    if existing is not None:
+        raise OnyxError(
+            OnyxErrorCode.DUPLICATE_RESOURCE,
+            f"A skill with slug '{slug}' already exists.",
+        )
+
+    skill = Skill(
+        slug=slug,
+        name=name,
+        description=description,
+        bundle_file_id=bundle_file_id,
+        bundle_sha256=bundle_sha256,
+        manifest_metadata=manifest_metadata,
+        is_public=is_public,
+        author_user_id=author_user_id,
+        enabled=True,
+    )
+    db_session.add(skill)
+    db_session.flush()
+    return skill
+
+
+def replace_skill_bundle(
+    *,
+    skill_id: UUID,
+    new_bundle_file_id: str,
+    new_bundle_sha256: str,
+    new_manifest_metadata: dict[str, Any],
+    db_session: Session,
+) -> tuple[Skill, str]:
+    """Swap a skill's bundle blob.
+
+    Returns `(skill, old_bundle_file_id)` so the caller can delete the old
+    blob from FileStore AFTER the transaction commits — never inline.
+    """
+    skill = fetch_skill_for_admin(skill_id, db_session)
+    if skill is None:
+        raise OnyxError(
+            OnyxErrorCode.NOT_FOUND,
+            f"Skill {skill_id} not found.",
+        )
+
+    old_bundle_file_id = skill.bundle_file_id
+    skill.bundle_file_id = new_bundle_file_id
+    skill.bundle_sha256 = new_bundle_sha256
+    skill.manifest_metadata = new_manifest_metadata
+    db_session.flush()
+    return skill, old_bundle_file_id
+
+
+def patch_skill(
+    *,
+    skill_id: UUID,
+    slug: str | Any = _UNSET,
+    name: str | Any = _UNSET,
+    description: str | Any = _UNSET,
+    is_public: bool | Any = _UNSET,
+    enabled: bool | Any = _UNSET,
+    db_session: Session,
+) -> Skill:
+    """Partial update of admin-controlled metadata.
+
+    `_UNSET` distinguishes "leave alone" from "set to None/falsy". Slug
+    uniqueness is re-checked when the slug changes (the partial unique index
+    is the DB backstop; this raises `DUPLICATE_RESOURCE` first for a clean
+    structured error).
+    """
+    skill = fetch_skill_for_admin(skill_id, db_session)
+    if skill is None:
+        raise OnyxError(
+            OnyxErrorCode.NOT_FOUND,
+            f"Skill {skill_id} not found.",
+        )
+
+    if slug is not _UNSET and slug != skill.slug:
+        clashing = db_session.scalars(
+            select(Skill.id)
+            .where(Skill.slug == slug)
+            .where(Skill.deleted_at.is_(None))
+            .where(Skill.id != skill_id)
+        ).first()
+        if clashing is not None:
+            raise OnyxError(
+                OnyxErrorCode.DUPLICATE_RESOURCE,
+                f"A skill with slug '{slug}' already exists.",
+            )
+        skill.slug = slug
+
+    if name is not _UNSET:
+        skill.name = name
+    if description is not _UNSET:
+        skill.description = description
+    if is_public is not _UNSET:
+        skill.is_public = is_public
+    if enabled is not _UNSET:
+        skill.enabled = enabled
+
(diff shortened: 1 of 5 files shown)


backend/onyx/db/skill.py
```diff
@@ -1 +1,268 @@
...
+    db_session.execute(
+        delete(Skill__UserGroup).where(Skill__UserGroup.skill_id == skill_id)
+    )
+    seen: set[int] = set()
+    for group_id in group_ids:
+        if group_id in seen:
+            continue
+        seen.add(group_id)
+        db_session.add(Skill__UserGroup(skill_id=skill_id, user_group_id=group_id))
+    db_session.flush()
```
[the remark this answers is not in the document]
rohoswagger: Already addressed in [`5024027`](https://github.com/onyx-dot-app/onyx/commit/50240276908fb1e59c193be868dc9a3765a183e0): `replace_skill_grants` now calls `fetch_skill_for_admin` up front and raises `OnyxError(OnyxErrorCode.NOT_FOUND)` for missing skills, matching the pattern in `replace_skill_bundle` / `patch_skill` / `delete_skill`. No more FK-violation 500 — callers get a structured `NOT_FOUND`.

backend/onyx/db/skill.py
```diff
@@ -1 +1,268 @@
...
+
+from onyx.auth.schemas import UserRole
+from onyx.db.models import Skill
+from onyx.db.models import Skill__UserGroup
+from onyx.db.models import User
+from onyx.db.models import User__UserGroup
+from onyx.error_handling.error_codes import OnyxErrorCode
+from onyx.error_handling.exceptions import OnyxError
+
+_UNSET: Final[Any] = object()
```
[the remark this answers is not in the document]
rohoswagger: Fixed in [`12a50c8`](https://github.com/onyx-dot-app/onyx/commit/12a50c83d6). Replaced `_UNSET: Final[Any] = object()` with a dedicated `_UnsetType` class and `_UNSET: Final[_UnsetType] = _UnsetType()`, so `patch_skill` parameters are typed as `str | _UnsetType` / `bool | _UnsetType` — ty/mypy/pyright now enforce the concrete type at the call site. Narrowing inside the function uses `isinstance(v, _UnsetType)` (rather than `is _UNSET`) because SQLAlchemy `Mapped[str]` descriptors only accept the post-narrow `str`. Went with the simpler module-private class — no `__new__` guard — since `_UnsetType` is unexported and the singleton is identity-stable.

backend/onyx/db/skill.py
```diff
@@ -1 +1,268 @@
...
+    stmt: Select[tuple[Skill]], user: User
+) -> Select[tuple[Skill]]:
+    """Restrict a `select(Skill)` to rows the given user can see.
+
+    Mirrors `onyx.db.persona._add_user_filters` minus the direct user-grant
+    branch (no `Skill__User` table in V1). Admins bypass the filter; everyone
+    else sees `is_public=True` OR membership in a group that holds a grant.
+    """
+    if user.role == UserRole.ADMIN:
+        return stmt
```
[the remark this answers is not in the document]
rohoswagger: Clarified in the docstring in [`6b3c1aa`](https://github.com/onyx-dot-app/onyx/commit/6b3c1aacb9) rather than changing behavior. The curator-specific elevated path in `persona._add_user_filters` exists to gate editability of curator-owned personas; skills have no editability dimension at the user-read layer in V1 (only admins mutate skills), so global-curators are intentionally treated the same as regular users for visibility (is_public OR group-grant). The docstring now states that explicitly so the question doesn't re-surface.

backend/onyx/db/skill.py
```diff
@@ -1 +1,268 @@
...
+) -> Skill:
+    """Insert a new Skill row.
+
+    Raises `OnyxError(DUPLICATE_RESOURCE)` if a non-deleted skill with the
+    same slug already exists. The partial unique index on `slug WHERE
+    deleted_at IS NULL` enforces this at the DB layer as a backstop; this
+    helper checks first so the caller gets a clean structured error rather
+    than an IntegrityError.
+    """
+    existing = db_session.scalars(
```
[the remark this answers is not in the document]
rohoswagger: Fixed in [`83cefbd`](https://github.com/onyx-dot-app/onyx/commit/83cefbdae3). Both `create_skill` and `patch_skill` now wrap the flush in `db_session.begin_nested()` (SAVEPOINT) and translate `IntegrityError` from the partial unique index into `OnyxError(DUPLICATE_RESOURCE)`. The pre-check is still there for the fast happy path; the SAVEPOINT-protected flush closes the concurrent-writer race so no caller ever sees a raw `IntegrityError` for slug collisions. In `patch_skill` the translation only fires when the slug actually changed — any other constraint failure still bubbles up unchanged.

backend/onyx/db/skill.py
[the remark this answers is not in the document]
rohoswagger: Fixed in [`12a50c8`](https://github.com/onyx-dot-app/onyx/commit/12a50c83d6). Introduced a module-private `_UnsetType` class and `_UNSET: Final[_UnsetType] = _UnsetType()`. `patch_skill` parameters are now `str | _UnsetType` / `bool | _UnsetType` so type checkers enforce the concrete type at the call site. Internal narrowing uses `isinstance(v, _UnsetType)` because SQLAlchemy `Mapped[str]` descriptors reject the union after a value-identity check.

backend/onyx/db/skill.py
```diff
@@ -1 +1,307 @@
...
+        manifest_metadata=manifest_metadata,
+        is_public=is_public,
+        author_user_id=author_user_id,
+        enabled=True,
+    )
+    try:
+        with db_session.begin_nested():
+            db_session.add(skill)
+            db_session.flush()
+    except IntegrityError as e:
```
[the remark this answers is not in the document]
rohoswagger: Fixed in [`b1ac871`](https://github.com/onyx-dot-app/onyx/commit/b1ac871664). The `IntegrityError` handler in `create_skill` now calls a small `_is_slug_unique_violation(e)` helper that checks `e.orig.diag.constraint_name == "ux_skill_slug"` (the partial unique index from the V1 migration). Only that specific violation translates to `DUPLICATE_RESOURCE`; FK violations on `author_user_id`, NOT NULL failures, etc. now bubble up unchanged so the real cause isn't hidden behind a duplicate-slug message.

backend/onyx/db/skill.py
```diff
@@ -1 +1,307 @@
...
+    """
+    skill = fetch_skill_for_admin(skill_id, db_session)
+    if skill is None:
+        raise OnyxError(
+            OnyxErrorCode.NOT_FOUND,
+            f"Skill {skill_id} not found.",
+        )
+
+    slug_changed = not isinstance(slug, _UnsetType) and slug != skill.slug
+    if slug_changed:
```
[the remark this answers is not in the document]
rohoswagger: Fixed in the same commit [`b1ac871`](https://github.com/onyx-dot-app/onyx/commit/b1ac871664). `patch_skill` now gates the translation on `slug_changed AND _is_slug_unique_violation(e)`, so an `IntegrityError` from anything other than the slug partial unique index — or from a patch where the slug wasn't even changed — re-raises unchanged. The constraint-name match is via `e.orig.diag.constraint_name == "ux_skill_slug"`.

backend/onyx/db/skill.py
```diff
@@ -1 +1,328 @@
...
+    seen: set[int] = set()
+    for group_id in group_ids:
+        if group_id in seen:
+            continue
+        seen.add(group_id)
+        db_session.add(Skill__UserGroup(skill_id=skill_id, user_group_id=group_id))
+    db_session.flush()
+
+
+def delete_skill(skill_id: UUID, db_session: Session) -> None:
```
wenxi-onyx: Since skills will be synced via S3 files, i think this should be hard-delete immediately. Also allows for reducing complexity of sweep system

backend/onyx/db/skill.py
```diff
@@ -1 +1,328 @@
...
+from onyx.auth.schemas import UserRole
+from onyx.db.models import Skill
+from onyx.db.models import Skill__UserGroup
+from onyx.db.models import User
+from onyx.db.models import User__UserGroup
+from onyx.error_handling.error_codes import OnyxErrorCode
+from onyx.error_handling.exceptions import OnyxError
+
+
+class _UnsetType:
```
wenxi-onyx: TBH might be good to put this in a higher-level file to share with other patch/upsert functions.

backend/onyx/db/skill.py
```diff
@@ -1 +1,328 @@
...
+_UNSET: Final[_UnsetType] = _UnsetType()
+
+# Name of the partial unique index on `skill.slug` (created in the Skills V1
+# migration). We match on this when translating `IntegrityError` so we only
+# remap the actual slug-collision violation, not FK or other constraint
+# failures.
+_SKILL_SLUG_UNIQUE_INDEX = "ux_skill_slug"
+
+
+def _is_slug_unique_violation(exc: IntegrityError) -> bool:
```
wenxi-onyx: If this is necessary + the only way to do this, might want to make this utility more generic and store constraint strings together. 

Generally, do we even use db/sql-level constraints like this? I wanna say we let sqlalchemy handle these kinds of things with Unique(), but not 100% sure

```
from psycopg.errors import UniqueViolation
from sqlalchemy.exc import IntegrityError


def is_unique_violation(
    exc: IntegrityError,
    constraint: str,
) -> bool:
    orig = exc.orig

    return (
        isinstance(orig, UniqueViolation)
        and orig.diag.constraint_name == constraint
    )
```

backend/onyx/db/skill.py
```diff
@@ -1 +1,328 @@
...
+        description=description,
+        bundle_file_id=bundle_file_id,
+        bundle_sha256=bundle_sha256,
+        manifest_metadata=manifest_metadata,
+        is_public=is_public,
+        author_user_id=author_user_id,
+        enabled=True,
+    )
+    try:
+        with db_session.begin_nested():
```
wenxi-onyx: why is this necessary? does it auto-rollback() when you use `with`?

backend/onyx/db/skill.py
```diff
@@ -1 +1,328 @@
...
+
+    old_bundle_file_id = skill.bundle_file_id
+    skill.bundle_file_id = new_bundle_file_id
+    skill.bundle_sha256 = new_bundle_sha256
+    skill.manifest_metadata = new_manifest_metadata
+    db_session.flush()
+    return skill, old_bundle_file_id
+
+
+def patch_skill(
```
wenxi-onyx, line 216: Should this _UNSET validation stuff happen at the db layer or the API layer?

wenxi-onyx (changes requested):

backend/onyx/db/skill.py
rohoswagger: Switched to hard-delete in [`9fa4617`](https://github.com/onyx-dot-app/onyx/commit/9fa4617d8d). `delete_skill` now does `db_session.delete(skill)` (the `skill__user_group` FK cascades automatically) and returns the `bundle_file_id` so the API caller can drop the blob from the file store immediately — same return-for-cleanup shape as `replace_skill_bundle`. The sweep tasks (P5.030-P5.038) no longer have a "blob retained from a soft-deleted row" code path to handle. Read filters that referenced `deleted_at IS NULL` were dead code and have been removed; the column itself stays on the model from the V1 migration as a no-op (drop in a follow-up migration).

backend/onyx/db/skill.py
rohoswagger: Hoisted in [`9d11ef1`](https://github.com/onyx-dot-app/onyx/commit/9d11ef16fc). `UnsetType` + `UNSET` now live in `backend/onyx/db/utils.py` and are imported here. The sentinel pattern is general-purpose — any patch helper can adopt the same `field: T | UnsetType = UNSET` idiom and narrow internally via `isinstance(v, UnsetType)`.

backend/onyx/db/skill.py
rohoswagger: Generalized in the same commit [`9d11ef1`](https://github.com/onyx-dot-app/onyx/commit/9d11ef16fc). `db/utils.py` now exposes `is_unique_violation(exc: IntegrityError, constraint: str) -> bool`, matching the shape you sketched (`psycopg2.errors.UniqueViolation` + `diag.constraint_name`). The constraint name itself stays in `skill.py` as `SKILL_SLUG_UNIQUE_INDEX` — kept next to the migration that creates the index so the coupling is local.

On the SQLAlchemy `UniqueConstraint`/`Unique()` question: SQLAlchemy's declarative `UniqueConstraint` covers plain unique, but this column needs a **partial** unique index (`WHERE deleted_at IS NULL`) which SQLAlchemy doesn't render natively — the V1 migration uses a raw `CREATE UNIQUE INDEX ... WHERE ...`. With hard-delete now in play the partial clause is unused, but the index is still real, so the constraint-name match remains the right place to translate.

backend/onyx/db/skill.py
rohoswagger: `with db_session.begin_nested():` auto-rolls back the SAVEPOINT on exception, but the exception itself still propagates out of the `with` block. The outer `try` / `except IntegrityError` exists to **translate** the propagated `IntegrityError` into `OnyxError(DUPLICATE_RESOURCE)` — different job from rolling back the savepoint. Without the `with`, an `IntegrityError` would poison the outer transaction; without the `try`, callers would see a raw `IntegrityError` instead of the structured error. Both layers are doing different work.

backend/onyx/db/skill.py
rohoswagger: Kept at the DB layer intentionally. The API layer maps incoming JSON (where a missing field genuinely means "leave alone") onto `patch_skill(...)`, and the DB helper is the single source of truth for which columns get touched. The alternative — API layer reads current values, fills in unchanged fields, then calls a full-record update — adds a round-trip and duplicates the "which columns mutate" knowledge in two places. The `UnsetType` sentinel keeps the partial-update intent expressible in the function signature without any of that.

Happy to revisit if the API ends up needing materially different semantics (e.g. conditional update gated on optimistic locking), but right now the helper is doing exactly one job.

backend/onyx/db/skill.py
wenxi-onyx: This operation only inserts a single skill. Are skills batch created? Otherwise, callers would just commit after this right? There's no transaction to poison if this is the only operation in the transaction

rohoswagger: yeah you're right, dropped the savepoints in c0e8fec. caller does one create_skill then commits (or rolls back on error), nothing in between cares about the broken session state. same applies to patch_skill since the api handler doesn't keep using the session after an OnyxError fires.

kept the pre-check for the clean error message in the common case and the try/except for the race translation, but no more begin_nested.

backend/onyx/db/skill.py
rohoswagger: ok cleaned this up in 431f388. since the schema hasn't been deployed anywhere yet, edited the v1 migration in place rather than adding a new one. swapped the partial `Index(..., postgresql_where=text("deleted_at IS NULL"))` for a plain `UniqueConstraint("slug", name="uq_skill_slug")` and dropped the `deleted_at` column entirely. `is_unique_violation` still does the same constraint-name match, just against `uq_skill_slug` now.

wenxi-onyx (approved):

PR #11061, opened 2026-05-13
feat(skills): Add builtin skill registry primitive

rohoswagger:
## Summary
- add the universal BuiltinSkillRegistry models and singleton
- load built-in skill metadata from SKILL.md or SKILL.md.template frontmatter
- add unit coverage for duplicate registration, missing SKILL.md, template detection, and requirement filtering/admin status
- update the skills task board entries for this branch to REVIEW

## Tests
- uv run pytest -q backend/tests/unit/onyx/skills/test_registry.py
- uv run ruff check backend/onyx/skills/registry.py backend/tests/unit/onyx/skills/test_registry.py
- uv run ruff format --check backend/onyx/skills/registry.py backend/tests/unit/onyx/skills/test_registry.py
- uv run ty check backend/onyx/skills/registry.py backend/tests/unit/onyx/skills/test_registry.py

---
## Summary by cubic
Adds a process-wide `BuiltinSkillRegistry`, DB ops for custom skills, and a custom bundle validator with safe unzip and SHA-256 hashing. Built-ins are filtered by availability; admins get CRUD + grants for uploaded skills with strict slug uniqueness.

- **New Features**
  - Registry loads `SKILL.md` or `.template`, validates slugs, tracks template presence, and filters via `list_available(db)`. APIs: `register`, `list_all`, `get`, `reserved_slugs`; test-only `_reset_for_testing()`. `BuiltinSkill` is frozen with optional `unavailable_reason`.
  - Custom skills: `CustomSkill` view; DB ops for visibility and CRUD — `list_skills_for_user/admin`, `fetch_skill_for_user/admin`, `create_skill` (pre-check + unique-constraint translation on `uq_skill_slug`), `replace_skill_bundle` (returns old `bundle_file_id`), `patch_skill` (uses `UNSET`; rechecks slug), `replace_skill_grants`, `delete_skill` (hard delete; returns `bundle_file_id`).
  - Bundle handling: `validate_custom_bundle` (slug + reserved-slug check, no `*.template`, path traversal and symlink rejection, streaming per-file/total size caps with 413s); `_safe_unzip` (defensive re-checks, cleans dest on failure); `compute_bundle_sha256`.

- **Refactors**
  - Renamed `list_satisfied` to `list_available`; default availability uses a lambda.
  - Slug validation moved to a Pydantic field validator; frontmatter parsing uses a single delimiter regex; registry singleton init/reset is lock-protected.
  - Schema: dropped `skill.deleted_at` and `skill.manifest_metadata`; replaced partial unique index with `uq_skill_slug`; models updated accordingly.
  - Removed built-in `configure_url`; frontend derives setup routes from known built-in slugs.

<sup>Written for commit df99b93e73c1ba79afd949d4fc7da9020b2851b7. Summary will update on new commits.</sup>

## Additional Options

---

### DIFF
--- a/backend/onyx/skills/registry.py
+++ b/backend/onyx/skills/registry.py
@@ -9,10 +9,7 @@ from pydantic import ConfigDict
 from sqlalchemy.orm import Session
 
 _SLUG_REGEX = re.compile(r"^[a-z][a-z0-9-]{0,63}$")
-
-
-def _always_available(_: Session) -> bool:
-    return True
+_DEFAULT_IS_AVAILABLE: Callable[[Session], bool] = lambda _: True  # noqa: E731
 
 
 class BuiltinSkill(BaseModel):
@@ -25,7 +22,7 @@ class BuiltinSkill(BaseModel):
     name: str
     description: str
     has_template: bool
-    is_available: Callable[[Session], bool] = _always_available
+    is_available: Callable[[Session], bool] = _DEFAULT_IS_AVAILABLE
     unavailable_reason: str | None = None
     configure_url: str | None = None
 
@@ -109,7 +106,7 @@ class BuiltinSkillRegistry:
         self,
         slug: str,
         source_dir: Path,
-        is_available: Callable[[Session], bool] = _always_available,
+        is_available: Callable[[Session], bool] = _DEFAULT_IS_AVAILABLE,
         unavailable_reason: str | None = None,
         configure_url: str | None = None,
     ) -> None:
--- a/backend/onyx/skills/registry.py
+++ b/backend/onyx/skills/registry.py
@@ -139,7 +139,7 @@ class BuiltinSkillRegistry:
     def list_all(self) -> list[BuiltinSkill]:
         return list(self._skills.values())
 
-    def list_satisfied(self, db: Session) -> list[BuiltinSkill]:
+    def list_available(self, db: Session) -> list[BuiltinSkill]:
         return [skill for skill in self.list_all() if skill.is_available(db)]
 
     def get(self, slug: str) -> BuiltinSkill | None:
--- a/backend/onyx/skills/registry.py
+++ b/backend/onyx/skills/registry.py
@@ -1,40 +1,55 @@
 import re
 from collections.abc import Callable
 from pathlib import Path
+from threading import Lock
 from typing import ClassVar
 
 import yaml
 from pydantic import BaseModel
 from pydantic import ConfigDict
+from pydantic import field_validator
 from sqlalchemy.orm import Session
 
 _SLUG_REGEX = re.compile(r"^[a-z][a-z0-9-]{0,63}$")
+_FRONTMATTER_REGEX = re.compile(
+    r"\A---[ \t]*\r?\n(?P<frontmatter>.*?)(?:\r?\n)---[ \t]*(?:\r?\n|\Z)",
+    re.DOTALL,
+)
 _DEFAULT_IS_AVAILABLE: Callable[[Session], bool] = lambda _: True  # noqa: E731
+_SLUG_ERROR = (
+    "Skill slug must start with a lowercase letter, contain only lowercase "
+    "letters, numbers, and hyphens, and be at most 64 characters."
+)
 
 
-class BuiltinSkill(BaseModel):
-    """In-memory entry for an on-disk built-in skill."""
+class Skill(BaseModel):
+    """Common skill metadata shared by built-in and custom skill views."""
 
     model_config = ConfigDict(frozen=True, arbitrary_types_allowed=True)
 
     slug: str
-    source_dir: Path
     name: str
     description: str
+
+    @field_validator("slug")
+    @classmethod
+    def validate_slug(cls, slug: str) -> str:
+        if not _SLUG_REGEX.fullmatch(slug):
+            raise ValueError(_SLUG_ERROR)
+        return slug
+
+
+class BuiltinSkill(Skill):
+    """In-memory entry for an on-disk built-in skill."""
+
+    source_dir: Path
     has_template: bool
     is_available: Callable[[Session], bool] = _DEFAULT_IS_AVAILABLE
     unavailable_reason: str | None = None
     configure_url: str | None = None
 
 
-def _validate_slug(slug: str) -> None:
-    if not _SLUG_REGEX.fullmatch(slug):
-        raise ValueError(
-            f"Skill slug must match ^[a-z][a-z0-9-]{{0,63}}$; got {slug!r}"
-        )
-
-
-def _skill_metadata_path(source_dir: Path) -> tuple[Path, bool]:
+def _select_skill_definition_path(source_dir: Path) -> tuple[Path, bool]:
     skill_md_path = source_dir / "SKILL.md"
     template_path = source_dir / "SKILL.md.template"
 
@@ -46,31 +61,26 @@ def _skill_metadata_path(source_dir: Path) -> tuple[Path, bool]:
 
     raise ValueError(
         f"Built-in skill source directory {source_dir} must contain "
-        "SKILL.md or SKILL.md.template"
+        "either SKILL.md or SKILL.md.template"
     )
 
 
 def _read_frontmatter(path: Path) -> dict[str, object]:
     content = path.read_text(encoding="utf-8")
-    lines = content.splitlines()
+    match = _FRONTMATTER_REGEX.match(content)
+    if match is None:
+        raise ValueError(
+            f"{path} must start with YAML frontmatter delimited by two --- lines"
+        )
 
-    if not lines or lines[0].strip() != "---":
-        raise ValueError(f"{path} must start with YAML frontmatter")
-
-    frontmatter_lines: list[str] = []
-    for line in lines[1:]:
-        if line.strip() == "---":
-            parsed = yaml.safe_load("\n".join(frontmatter_lines)) or {}
-            if not isinstance(parsed, dict):
-                raise ValueError(f"{path} frontmatter must be a mapping")
-            return parsed
-        frontmatter_lines.append(line)
-
-    raise ValueError(f"{path} frontmatter is missing closing --- delimiter")
+    parsed = yaml.safe_load(match.group("frontmatter")) or {}
+    if not isinstance(parsed, dict):
+        raise ValueError(f"{path} frontmatter must be a mapping")
+    return parsed
 
 
 def _read_metadata(source_dir: Path) -> tuple[str, str, bool]:
-    metadata_path, has_template = _skill_metadata_path(source_dir)
+    metadata_path, has_template = _select_skill_definition_path(source_dir)
     frontmatter = _read_frontmatter(metadata_path)
 
     name = frontmatter.get("name")
@@ -88,6 +98,7 @@ class BuiltinSkillRegistry:
     """Process-wide registry populated with on-disk built-in skills at boot."""
 
     _instance: ClassVar["BuiltinSkillRegistry | None"] = None
+    _instance_lock: ClassVar[Lock] = Lock()
 
     def __init__(self) -> None:
         self._skills: dict[str, BuiltinSkill] = {}
@@ -95,12 +106,15 @@ class BuiltinSkillRegistry:
     @classmethod
     def instance(cls) -> "BuiltinSkillRegistry":
         if cls._instance is None:
-            cls._instance = cls()
+            with cls._instance_lock:
+                if cls._instance is None:
+                    cls._instance = cls()
         return cls._instance
 
     @classmethod
     def _reset_for_testing(cls) -> None:
-        cls._instance = None
+        with cls._instance_lock:
+            cls._instance = None
 
     def register(
         self,
@@ -110,8 +124,6 @@ class BuiltinSkillRegistry:
         unavailable_reason: str | None = None,
         configure_url: str | None = None,
     ) -> None:
-        _validate_slug(slug)
-
         if slug in self._skills:
             raise ValueError(f"Built-in skill {slug!r} is already registered")
 
--- a/backend/onyx/skills/registry.py
+++ b/backend/onyx/skills/registry.py
@@ -1,1 +1,193 @@
+import re
+from collections.abc import Callable
+from collections.abc import Sequence
+from pathlib import Path
+from typing import ClassVar
 
+import yaml
+from pydantic import BaseModel
+from pydantic import ConfigDict
+from sqlalchemy.orm import Session
+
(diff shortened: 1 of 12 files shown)


backend/onyx/skills/registry.py
```diff
@@ -1 +1,193 @@
...
+    _instance: ClassVar["BuiltinSkillRegistry | None"] = None
+
+    def __init__(self) -> None:
+        self._skills: dict[str, BuiltinSkill] = {}
+
+    @classmethod
+    def instance(cls) -> "BuiltinSkillRegistry":
+        if cls._instance is None:
+            cls._instance = cls()
+        return cls._instance
```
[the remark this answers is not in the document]
rohoswagger: Fixed in `6b4dfdbd6f`.

Added `BuiltinSkillRegistry._reset_for_testing()` in `backend/onyx/skills/registry.py` so tests using the singleton can clear process-local registrations. Added `test_reset_for_testing_clears_singleton` in `backend/tests/unit/onyx/skills/test_registry.py` to cover the isolation path through `.instance()`.

Verified with:
- `uv run pytest -q backend/tests/unit/onyx/skills/test_registry.py`
- `uv run ruff check backend/onyx/skills/registry.py backend/tests/unit/onyx/skills/test_registry.py`
- `uv run ruff format --check backend/onyx/skills/registry.py backend/tests/unit/onyx/skills/test_registry.py`
- `uv run ty check backend/onyx/skills/registry.py backend/tests/unit/onyx/skills/test_registry.py`

backend/onyx/skills/registry.py
```diff
@@ -1 +1,149 @@
...
+from sqlalchemy.orm import Session
+
+_SLUG_REGEX = re.compile(r"^[a-z][a-z0-9-]{0,63}$")
+
+
+def _always_available(_: Session) -> bool:
+    return True
+
+
+class BuiltinSkill(BaseModel):
```
wenxi-onyx: Why not have a Skill class that this extends?

backend/onyx/skills/registry.py
```diff
@@ -1 +1,149 @@
...
 
+import yaml
+from pydantic import BaseModel
+from pydantic import ConfigDict
+from sqlalchemy.orm import Session
+
+_SLUG_REGEX = re.compile(r"^[a-z][a-z0-9-]{0,63}$")
+
+
+def _always_available(_: Session) -> bool:
```
wenxi-onyx: What is this bruh

backend/onyx/skills/registry.py
```diff
@@ -1 +1,149 @@
...
+    source_dir: Path
+    name: str
+    description: str
+    has_template: bool
+    is_available: Callable[[Session], bool] = _always_available
+    unavailable_reason: str | None = None
+    configure_url: str | None = None
+
+
+def _validate_slug(slug: str) -> None:
```
wenxi-onyx: Pydantic validator --> should be a part of the class

backend/onyx/skills/registry.py
```diff
@@ -1 +1,149 @@
...
+            if not isinstance(parsed, dict):
+                raise ValueError(f"{path} frontmatter must be a mapping")
+            return parsed
+        frontmatter_lines.append(line)
+
+    raise ValueError(f"{path} frontmatter is missing closing --- delimiter")
+
+
+def _read_metadata(source_dir: Path) -> tuple[str, str, bool]:
+    metadata_path, has_template = _skill_metadata_path(source_dir)
```
wenxi-onyx: Confusing function name / what is the intended flow here? Check if destination has template or skill? I think the function name and behavior can both be made more clear.

backend/onyx/skills/registry.py
```diff
@@ -1 +1,149 @@
...
+    frontmatter_lines: list[str] = []
+    for line in lines[1:]:
+        if line.strip() == "---":
+            parsed = yaml.safe_load("\n".join(frontmatter_lines)) or {}
+            if not isinstance(parsed, dict):
+                raise ValueError(f"{path} frontmatter must be a mapping")
+            return parsed
+        frontmatter_lines.append(line)
+
+    raise ValueError(f"{path} frontmatter is missing closing --- delimiter")
```
wenxi-onyx: Maybe we should start using OnyxErrors

wenxi-onyx: Also, you can definitely build a regex that immediately validates front matter and reads the entire content between the dashes (in a single regex). Should fail immediately if the 2 triple dash lines don't exist and then just return what's between them

backend/onyx/skills/registry.py
```diff
@@ -1 +1,149 @@
...
+    has_template: bool
+    is_available: Callable[[Session], bool] = _always_available
+    unavailable_reason: str | None = None
+    configure_url: str | None = None
+
+
+def _validate_slug(slug: str) -> None:
+    if not _SLUG_REGEX.fullmatch(slug):
+        raise ValueError(
+            f"Skill slug must match ^[a-z][a-z0-9-]{{0,63}}$; got {slug!r}"
```
wenxi-onyx: Not a helpful error. Should just explain what the slug format should be

backend/onyx/skills/registry.py
```diff
@@ -1 +1,149 @@
...
+
+
+def _always_available(_: Session) -> bool:
+    return True
+
+
+class BuiltinSkill(BaseModel):
+    """In-memory entry for an on-disk built-in skill."""
+
+    model_config = ConfigDict(frozen=True, arbitrary_types_allowed=True)
```
wenxi-onyx, line 46: Because this is loaded in memory and frozen, if an admin adds something that makes this skill technically available, it can never be updated right?

backend/onyx/skills/registry.py
```diff
@@ -1 +1,149 @@
...
+
+    slug: str
+    source_dir: Path
+    name: str
+    description: str
+    has_template: bool
```
wenxi-onyx, line 50: Is there a reason we need templates? Can we just have SKILL.md with placeholder text that we substitute directly + some field on the class that denotes it is dynamic?

backend/onyx/skills/registry.py
```diff
@@ -1 +1,149 @@
...
+    is_available: Callable[[Session], bool] = _always_available
```
wenxi-onyx: Does it make sense that built-in skills are always available? What if no connectors exist or no image gen model exist?

backend/onyx/skills/registry.py
```diff
@@ -1 +1,149 @@
...
+
+    if not isinstance(name, str) or not name.strip():
+        raise ValueError(f"{metadata_path} frontmatter must include name")
+    if not isinstance(description, str) or not description.strip():
+        raise ValueError(f"{metadata_path} frontmatter must include description")
+
+    return name, description, has_template
+
+
+class BuiltinSkillRegistry:
```
wenxi-onyx, line 112: If this is a singleton, does it need a thread lock

backend/onyx/skills/registry.py
```diff
@@ -1 +1,149 @@
...
+    model_config = ConfigDict(frozen=True, arbitrary_types_allowed=True)
+
+    slug: str
+    source_dir: Path
+    name: str
+    description: str
+    has_template: bool
+    is_available: Callable[[Session], bool] = _always_available
+    unavailable_reason: str | None = None
+    configure_url: str | None = None
```
wenxi-onyx: It feels weird to store the FE url on this BE skill model. The route should ideally be built off the name (which, also ideally should be unique) + some static route that is configured on the FE.

wenxi-onyx (changes requested):

backend/onyx/skills/registry.py
rohoswagger: Skill is a DB model, thats what we use for Custom skills. BuiltInSkills are the only type that are defined in code and need to be registered like this

backend/onyx/skills/registry.py
rohoswagger: not actually always available, this is callable we implement for each BuiltInSkill that defines how we can validate (check db for api key, etc.)

backend/onyx/skills/registry.py
rohoswagger: is_available is a Callable so this the default value

backend/onyx/skills/registry.py
rohoswagger: Updated this to keep `is_available` callable-only while removing the named `_always_available` helper. The default is now a lambda-backed callable constant:

`_DEFAULT_IS_AVAILABLE: Callable[[Session], bool] = lambda _: True`

Conditional built-ins still pass their own callable, so availability is evaluated when `list_available(db)` runs rather than being frozen at registration time.

backend/onyx/skills/registry.py
rohoswagger: Added the small shared abstraction here: `Skill` now carries the common immutable metadata (`slug`, `name`, `description`), and `BuiltinSkill` extends it with registry-only fields (`source_dir`, `has_template`, availability metadata). I kept the DB `Skill` model separate because that is persistence for custom skills, not the in-memory registry view.

backend/onyx/skills/registry.py
rohoswagger: Kept `is_available` callable-only, but removed the named placeholder function. The default is now `_DEFAULT_IS_AVAILABLE: Callable[[Session], bool] = lambda _: True`, so `list_available(db)` still has a single code path and conditional built-ins can pass their own DB-backed callable.

backend/onyx/skills/registry.py
rohoswagger: Moved slug validation into the Pydantic model. `Skill` now has a `field_validator("slug")`, and `BuiltinSkill` inherits it. Added `test_register_rejects_invalid_slug` to cover the validation path.

backend/onyx/skills/registry.py
rohoswagger: Renamed the helper to `_select_skill_definition_path` to make the flow clearer: it chooses the source definition file (`SKILL.md.template` if present, otherwise `SKILL.md`) and returns whether that source is templated. It is not checking the materialized destination.

backend/onyx/skills/registry.py
rohoswagger: Changed the frontmatter parser to a single delimiter regex so it fails immediately unless the file starts with two `---` delimiters and then parses only the content between them.

I intentionally kept these as `ValueError`s rather than `OnyxError`s: this registry is loaded from repo-owned built-in files at startup, so failures here are developer/deploy configuration errors. The upload/API path for custom skill bundles should use `OnyxError` because those errors are request-facing.

backend/onyx/skills/registry.py
rohoswagger: Replaced the regex-shaped error with a human-readable format description: slugs must start with a lowercase letter, contain only lowercase letters/numbers/hyphens, and be at most 64 characters.

backend/onyx/skills/registry.py
rohoswagger: updated to not freeze, BuiltInSkill is ok freezing bc its defined in code

backend/onyx/skills/registry.py
rohoswagger: I think keeping `SKILL.md.template` is the better V1 boundary. A plain `SKILL.md` should be a valid literal skill file. A separate template file makes dynamic built-ins explicit and lets the materializer write a normal `SKILL.md` into the runtime directory after substitution. The `has_template` field is that dynamic marker; using placeholders directly in `SKILL.md` would make the source file both template and runtime artifact.

backend/onyx/skills/registry.py
rohoswagger: Resolved in code by keeping `is_available` callable-only and replacing the named always-available helper with a lambda-backed default. Built-ins with dependencies still pass their own callable, so no-connectors/no-image-provider cases can be checked against the current DB/session state when `list_available(db)` runs.

backend/onyx/skills/registry.py
rohoswagger: Added a lock around singleton creation and reset. `instance()` now uses double-checked locking, and `_reset_for_testing()` takes the same lock before clearing the process-wide instance.

backend/onyx/skills/registry.py
rohoswagger: The registry already owns the availability check and unavailable reason, so it is the only layer that knows which setup page is relevant for a built-in. Removing the field would force either a hardcoded frontend slug map or a new symbolic indirection key.

docs/craft/features/skills/TODOS.md
```diff
@@ -51,47 +55,47 @@ _(Update this section as you claim things. Keep it short — just the active `WI
...
+- `[REVIEW @codex-charged-perovskite #11061]` `P1.020` Define `BuiltinSkill` in `registry.py` as a Pydantic `BaseModel` with frozen + arbitrary-types config, including optional `is_available`, `unavailable_reason`, and `configure_url` fields  (deps: P1.011)
+- `[REVIEW @codex-charged-perovskite #11061]` `P1.021` Keep built-in availability as a single optional callable on `BuiltinSkill`; do not introduce a separate `SkillRequirement` abstraction in V1  (deps: P1.020)
+- `[REVIEW @codex-charged-perovskite #11061]` `P1.022` Implement `BuiltinSkillRegistry` singleton accessor (`.instance()`)  (deps: P1.021)
+- `[REVIEW @codex-charged-perovskite #11061]` `P1.023` Implement `register(slug, source_dir, is_available=..., unavailable_reason=None, configure_url=None)` — read frontmatter, detect `SKILL.md.template` presence, slug regex validation, raise on duplicate or missing SKILL.md  (deps: P1.022)
+- `[REVIEW @codex-charged-perovskite #11061]` `P1.024` Implement `list_all() -> list[BuiltinSkill]`  (deps: P1.022)
+- `[REVIEW @codex-charged-perovskite #11061]` `P1.025` Implement `list_available(db) -> list[BuiltinSkill]` — filter by `skill.is_available(db) == True`  (deps: P1.020, P1.024)
```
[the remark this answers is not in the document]
rohoswagger: This is no longer applicable. `is_available` is not optional in the implementation; it is always a callable and defaults to `_DEFAULT_IS_AVAILABLE`, so `list_available(db)` cannot hit a `NoneType` call. Conditional built-ins still override that default with their own callable.

backend/onyx/skills/registry.py
rohoswagger: also its not even stored in db so p inexpensive to keep this

wenxi-onyx (approved):

backend/onyx/skills/registry.py
rohoswagger: removed configure_url

PR #11111, opened 2026-05-14
fix(helm): provision redis CRD and postgres cluster; drop -master from REDIS_HOST

wenxi-onyx:
## Description

A fresh `helm install onyx onyx/onyx` on a clean cluster previously failed: the redis CRD was missing, postgres was never provisioned, and REDIS_HOST pointed at a non-existent service. This PR makes the chart self-contained so a first-time install works without manual steps.

**Fixes:**

- **Redis operator** — added `redis-operator@0.24.0` as a subchart dependency (provides the CRD the `redis` subchart's CR needs). Removed the standalone pre-install step from the CI workflow since it now conflicts.
- **PostgreSQL Cluster CR** — new `templates/postgres-cluster.yaml` provisions postgres via the CNPG operator. Named `<release>-pg` to avoid a ServiceAccount collision with the operator Deployment. Resolves the superuser secret via `onyx.secretName` (honors `existingSecret` / `secretName` overrides) and uses `hasKey` for `enableSuperuserAccess` so `false` isn't coerced to `true`.
- **REDIS_HOST** — dropped the `-master` suffix from `configmap.yaml`. The chart provisions standalone Redis (no `-master` service).
- **CNPG CRDs** — copied into `crds/cnpg-crds.yaml` so Helm pre-installs them before template validation. The subchart's own CRD template is disabled (`postgresql.crds.create: false`) to avoid ownership conflicts. `scripts/check-cnpg-crds.sh` + a CI step verify the copy stays in sync on subchart version bumps.
- **Pre-delete hook** — `templates/pre-delete-cleanup.yaml` deletes the Cluster and Redis CRs while the operators are still running, so finalizers are processed and `helm uninstall` / namespace cleanup completes cleanly.

Chart version bumped to 0.5.0.

**For users with external databases (RDS, Elasticache):** set `postgresql.enabled: false` and `redis.enabled: false` as before — none of these changes render when the subcharts are disabled.

## How Has This Been Tested?

- Fresh `helm install` on a local kind cluster: redis operator + pod come up, CNPG Cluster provisions postgres, `POSTGRES_HOST` and `REDIS_HOST` resolve, api-server starts successfully.
- `helm template` verified for: default values, custom `auth.postgresql.secretName`, `existingSecret` override, `enableSuperuserAccess: false`.
- `ct install` CI passes (install + test + uninstall + namespace cleanup).
- `scripts/check-cnpg-crds.sh` passes against the current subchart.

## Additional Options


deployment/helm/charts/onyx/templates/configmap.yaml
```diff
@@ -19,7 +19,9 @@ data:
   ENABLE_OPENSEARCH_RETRIEVAL_FOR_ONYX: "false"
   {{- end }}
   {{- if .Values.redis.enabled }}
-  REDIS_HOST: {{ .Values.redis.redisStandalone.name | default .Release.Name }}-master
+  # Standalone Redis CR produces a Service named after the CR (no `-master`
+  # suffix — that suffix is only for RedisReplication CRs).
+  REDIS_HOST: {{ .Values.redis.redisStandalone.name | default .Release.Name }}
```
justin-tahara, line 24: I feel like there was history around why we did this but if we are adding the dependencies this could be reverted I guess

deployment/helm/charts/onyx/Chart.yaml
```diff
@@ -5,7 +5,7 @@ home: https://www.onyx.app/
 sources:
   - "https://github.com/onyx-dot-app/onyx"
 type: application
-version: 0.4.50
+version: 0.4.52
```
justin-tahara: preference would be to go to 0.5.0 for this

justin-tahara (commented):
I think there might be some helm test stuff that you might need to touch as well

wenxi-onyx:
@greptileai re-review

justin-tahara (approved):

PR #11155, opened 2026-05-18
fix(deletion): prevent OOMKill of monitoring pod during large deletions

justin-tahara:
## Description

Fix OOMKill of the monitoring pod during large connector deletions by replacing expensive Redis SCARD operations with fence metadata. The previous approach called `redis_connector.delete.get_remaining()` which would scan the entire taskset (150k+ items) and cause memory exhaustion on the monitoring pod. Now we estimate remaining tasks from the initial fence count (`num_tasks`), which is sufficient for progress reporting without expensive Redis operations.

## How Has This Been Tested?

- Verified that monitoring pod stays stable during large deletions (147k+ tasks)
- Confirmed fence metadata is correctly populated with initial task count
- Tested that progress reporting works with estimated remaining count
- Verified no degradation in monitoring visibility for normal-sized deletions

## Additional Options

### DIFF
--- a/backend/onyx/background/celery/tasks/connector_deletion/tasks.py
+++ b/backend/onyx/background/celery/tasks/connector_deletion/tasks.py
@@ -410,18 +410,16 @@ def monitor_connector_deletion_taskset(
         # the fence is setting up but isn't ready yet
         return
 
-    remaining = redis_connector.delete.get_remaining()
-    task_logger.info(
-        f"Connector deletion progress: cc_pair={cc_pair_id} remaining={remaining} initial={fence_data.num_tasks}"
-    )
-    if remaining > 0:
+    # Check if the taskset still exists in Redis without reading its size.
+    # redis.exists() is O(1) and very cheap, whereas scard() on 150k+ items OOMKills.
+    if r.exists(redis_connector.delete.taskset_key):
         with get_session_with_current_tenant() as db_session:
             update_sync_record_status(
                 db_session=db_session,
                 entity_id=cc_pair_id,
                 sync_type=SyncType.CONNECTOR_DELETION,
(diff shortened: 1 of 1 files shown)


backend/onyx/file_processing/extract_file_text.py
nmgarza5: intended?

backend/onyx/server/manage/administrative.py
```diff
@@ -168,6 +168,31 @@ def create_deletion_attempt_for_connector_id(
         cc_pair_id=cc_pair.id, db_session=db_session, include_secondary_index=True
     )
 
+    # Pre-flight size check to prevent accidental mass deletions
+    from sqlalchemy import func
+
+    from onyx.db.models import DocumentByConnectorCredentialPair
```
nmgarza5: top of file?

backend/onyx/server/manage/administrative.py
```diff
@@ -168,6 +168,31 @@ def create_deletion_attempt_for_connector_id(
...
+            DocumentByConnectorCredentialPair.connector_id == cc_pair.connector_id,
+            DocumentByConnectorCredentialPair.credential_id == cc_pair.credential_id,
+        )
+        .scalar()
+        or 0
+    )
+
+    if doc_count > 100_000:
+        from onyx.error_handling.error_codes import OnyxErrorCode
+        from onyx.error_handling.exceptions import OnyxError
```
nmgarza5: same comment as previous

nmgarza5 (approved):
approving but the non tasks.py changes seem like a mistake maybe? O rshould be in a diff pr? approving nonetheless and will let you make the call

PR #11175, opened 2026-05-18
fix(deletion): release PG connection before document-index I/O in per-doc cleanup

justin-tahara:
## Description

The per-document cleanup task (`document_by_cc_pair_cleanup_task`) held a SQLAlchemy session — and therefore a pgbouncer slot — open across the document-index HTTP delete/update call (OpenSearch on cloud where `ONYX_DISABLE_VESPA=true`, Vespa on legacy deployments). Under bulk-deletion fan-out across the light worker fleet (potentially 1000+ concurrent tasks for a large tenant), this saturated the pgbouncer pool with idle-in-transaction connections.

Observed today on the production managed_services cluster: ~947 of ~1051 pgbouncer server connections were idle-in-tx, all from `celery_worker_light_sync`, all running per-document SELECTs against the same tenant's schema. Pool was 97% saturated, average transaction time 17.4s, average client wait 5.9s. New API pods could not complete `warm_up_connections()` within the startup probe window and crash-looped.

This PR restructures the task into three explicit phases:

1. **Read** DB state inside a session, then exit the `with` block to release the connection.
2. **Document-index I/O** with no DB connection held. The HTTP round-trip no longer pins a pgbouncer slot.
3. **Write** back to PG in a fresh transaction.

Behavior is preserved on both branches:

- `count == 1`: SELECT chunk_count → document-index delete → KG cleanup + doc delete + commit (via `delete_documents_complete`).
- `count > 1`: SELECT doc/access/sets → document-index update → cc_pair detach + mark synced + commit.

Failure-mode notes:

- If the document-index call fails (phase 2): nothing has been written yet. Existing retry / max-retries-mark-dirty path handles this exactly as before.
- If phase 3 fails after phase 2 succeeded: document index is already updated/cleared. On retry, phase 1 sees the same state (count unchanged because phase 3 didn't run), phase 2 re-runs against the index (idempotent under the existing 404-tolerant handling), phase 3 finishes. Eventually consistent; same semantics as the existing partial-failure behavior.

## How Has This Been Tested?

- `pytest backend/tests/external_dependency_unit/indexing/test_document_deletion_file_cleanup.py` — all 5 tests pass, including the two cleanup-task tests (`test_count_1_branch_cleans_up_file`, `test_count_gt_1_branch_preserves_file`) which cover both branches.
- `pre-commit run` clean (ty, ruff, ruff format).

## Additional Options

### DIFF
--- a/backend/onyx/background/celery/tasks/shared/tasks.py
+++ b/backend/onyx/background/celery/tasks/shared/tasks.py
@@ -90,11 +90,16 @@ def document_by_cc_pair_cleanup_task(
     start = time.monotonic()
 
     completion_status = OnyxCeleryTaskCompletionStatus.UNDEFINED
+    action = "skip"
 
     try:
+        # Phase 1: read DB state, then release the connection before the
+        # Vespa HTTP call. Holding a pg transaction across the round-trip
+        # pins a pgbouncer slot for the duration and saturates the pool under
+        # bulk-deletion fan-out across the light worker fleet.
+        chunk_count: int | None = None
+        update_request: MetadataUpdateRequest | None = None
         with get_session_with_current_tenant() as db_session:
-            action = "skip"
-
             active_search_settings = get_active_search_settings(db_session)
             # This flow is for updates and deletion so we get all indices.
             document_indices = get_all_document_indices(
@@ -103,40 +108,10 @@ def document_by_cc_pair_cleanup_task(
                 httpx_client=HttpxPool.get("vespa"),
             )
 
-            retry_document_indices: list[RetryDocumentIndex] = [
-                RetryDocumentIndex(document_index)
-                for document_index in document_indices
-            ]
-
             count = get_document_connector_count(db_session, document_id)
             if count == 1:
-                # count == 1 means this is the only remaining cc_pair reference to the doc
-                # delete it from vespa and the db
-                action = "delete"
-
                 chunk_count = fetch_chunk_count_for_document(document_id, db_session)
-
-                for retry_document_index in retry_document_indices:
-                    _ = retry_document_index.delete(
-                        document_id,
-                        chunk_count=chunk_count,
-                    )
-
-                delete_document_references_from_kg(
-                    db_session=db_session,
-                    document_id=document_id,
-                )
-
-                delete_documents_complete(
-                    db_session=db_session,
-                    document_ids=[document_id],
-                )
-
-                completion_status = OnyxCeleryTaskCompletionStatus.SUCCEEDED
             elif count > 1:
-                action = "update"
-
-                # count > 1 means the document still has cc_pair references
                 doc = get_document(document_id, db_session)
                 if not doc:
                     return False
@@ -148,7 +123,6 @@ def document_by_cc_pair_cleanup_task(
                 )
 
                 doc_sets = fetch_document_sets_for_document(document_id, db_session)
-                update_doc_sets: set[str] = set(doc_sets)
 
                 update_request = MetadataUpdateRequest(
                     document_ids=[document_id],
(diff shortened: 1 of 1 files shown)


backend/onyx/background/celery/tasks/shared/tasks.py
```diff
@@ -90,11 +90,17 @@ def document_by_cc_pair_cleanup_task(
...
+        # a pgbouncer slot for the duration and saturates the pool under
+        # bulk-deletion fan-out across the light worker fleet.
+        chunk_count: int | None = None
+        update_request: MetadataUpdateRequest | None = None
         with get_session_with_current_tenant() as db_session:
-            action = "skip"
-
             active_search_settings = get_active_search_settings(db_session)
             # This flow is for updates and deletion so we get all indices.
             document_indices = get_all_document_indices(
```
acaprau: for even more savings you can also move this function call outside of the db session if you want, we just need to save the `active_search_settings` data and then we can create our doc index objects later.

why you might want to do this is because it can sometimes take a few seconds to connect to the document index server when you create the doc index classes.

backend/onyx/background/celery/tasks/shared/tasks.py
```diff
@@ -158,19 +133,50 @@ def document_by_cc_pair_cleanup_task(
...
-                    # it was ok if a doc did not exist in the document index. I
-                    # don't agree with that claim, so keep an eye on this task
-                    # to see if this raises.
-                    retry_document_index.update([update_request])
+        retry_document_indices: list[RetryDocumentIndex] = [
+            RetryDocumentIndex(document_index) for document_index in document_indices
+        ]
+
+        # Phase 2: document-index I/O — no DB connection held.
+        if count == 1:
```
acaprau: if it makes sense to you i would prefer if `action` is the source of truth variable here rather than `count`. inside of phase 1 we can set `action` to the appropriate value and then reference it here. i don't think `count` is defined until we get to line 111, so this way seems cleaner to me.

backend/onyx/background/celery/tasks/shared/tasks.py
```diff
@@ -90,11 +90,17 @@ def document_by_cc_pair_cleanup_task(
     start = time.monotonic()
 
     completion_status = OnyxCeleryTaskCompletionStatus.UNDEFINED
+    action: str = "skip"
```
acaprau: gigachads would make this an enum to eliminate a class of bugs at build time.

acaprau (approved):

backend/onyx/background/celery/tasks/shared/tasks.py
justin-tahara: Done in 56c0103. Lifted out of the session — only `active_search_settings.primary` / `.secondary` are captured inside the `with`, and `get_all_document_indices` + `RetryDocumentIndex(...)` construction now happen after the session closes. (`expire_on_commit=False` is set on the session, so accessing the detached `SearchSettings` attributes is safe.)

backend/onyx/background/celery/tasks/shared/tasks.py
justin-tahara: Done in 56c0103. Phase 1 now sets `action` (`DELETE`/`UPDATE`/`SKIP`) alongside the other read state, and phases 2/3 branch on `action` rather than `count`. `count` is kept only for the trailing log line.

backend/onyx/background/celery/tasks/shared/tasks.py
justin-tahara: Gigachad upgrade applied in 56c0103 — `action` is now a `DocumentCleanupAction(str, Enum)` with `SKIP`/`DELETE`/`UPDATE` members.

PR #11181, opened 2026-05-19
chore(opensearch, prometheus): Emit telemetry on search latency overhead

acaprau:
## Description
Search latency overhead is defined as the difference between wall clock time observed within Onyx for a search request, and the internal time OpenSearch reports for that request. This diff is meaningful because it includes certain pre- and post-search operations within OpenSearch that it does not report as time took, and it also includes network overhead.

Previously we were computing this in Grafana by finding the difference between the two relevant histograms, but this is flawed because these data sources are not directly comparable. They reflect aggregates, and additionally OpenSearch might not report time took. Explicitly computing these values in memory when both are present is more accurate.

## How Has This Been Tested?
Added automated tests.

## Additional Options

---
## Summary by cubic
Adds a Prometheus histogram to measure per-request search latency overhead (client wall-clock minus OpenSearch “took”) for accurate quantiles and dashboards. Computes the diff in-process instead of subtracting aggregate histograms in Grafana.

- New Features
  - Added `onyx_opensearch_search_overhead_seconds` with fine sub-second buckets for network/serialization overhead.
  - Observes overhead only when both client duration and server “took” are known; drops negative diffs with a warning, while still recording client and server histograms.
  - Added unit tests for overhead observation, missing server “took”, and negative-diff handling.

<sup>Written for commit 11d4b3940e8f860b07fc21942fd431adbba1e2ce. Summary will update on new commits. <a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11181?utm_source=github">Review in cubic</a></sup>


evan-onyx (approved):

acaprau:
@greptile @cubic

PR #11187, opened 2026-05-19
fix: jira perm sync dc v10

evan-onyx:
## Description

Adjust perm sync implementation to account for newer things in the Jira Data Center v10 SDK

## How Has This Been Tested?

added tests. All code should be backwards compatible, i.e. if perm sync worked for your Jira instance it should still work.

## Additional Options

---
## Summary by cubic
Fix Jira permission sync to support Jira Data Center v10 SDK and mixed Cloud/DC shapes. Improves user/group resolution and logging while keeping behavior the same.

- **Bug Fixes**
  - Normalize Cloud/DC holder and role shapes across attributes, raw dicts, and nested raw payloads; support both value/parameter forms.
  - Resolve user emails from direct holders and role actors: prefer embedded email; else look up by id (Cloud: accountId/name/key; DC: name/key/accountId); skip non-Atlassian users.
  - Read project role group names from actorGroup (with raw fallback) and include direct group holders; treat "anyone" and "applicationRole" as public.
  - Tweak logging: warn (not error) for dynamic-only/unsupported holders that produce empty private access; clearer messages for missing identifiers/emails. Added unit tests for Cloud/DC direct users, direct groups, raw actorGroup, and dynamic-only cases.

<sup>Written for commit 708807ec623cedaf9ea83afd5209de87149155ba. Summary will update on new commits. <a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11187?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/backend/ee/onyx/external_permissions/jira/page_access.py
+++ b/backend/ee/onyx/external_permissions/jira/page_access.py
@@ -1,5 +1,4 @@
 from collections import defaultdict
-from typing import cast
 
 from jira import JIRA
 from jira.resources import PermissionScheme
@@ -43,17 +42,13 @@ def _get_role_id(holder: Holder) -> str | None:
 # depending on Jira version and endpoint.
 def _get_obj_value(obj: object, field: str) -> object | None:
     if isinstance(obj, dict):
-        obj_dict = cast(dict[str, object], obj)
-        return obj_dict.get(field)
+        return obj.get(field)  # ty: ignore[invalid-argument-type]
     return getattr(obj, field, None)
 
 
 def _get_raw_value(obj: object, field: str) -> object | None:
     raw = _get_obj_value(obj, "raw")
-    if not isinstance(raw, dict):
-        return None
-    raw_dict = cast(dict[str, object], raw)
-    return raw_dict.get(field)
+    return _get_obj_value(raw, field)
 
 
 def _get_first_str_value(obj: object, fields: tuple[str, ...]) -> str | None:
@@ -222,15 +217,13 @@ def _get_user_emails(jira_project: str, user_holders: list[Holder]) -> list[str]
 
 
 def _get_actor_group_name(actor: object) -> str | None:
-    actor_group = _get_obj_value(actor, "actorGroup")
-    if actor_group is not None:
+    for actor_group in [
+        _get_obj_value(actor, "actorGroup"),
+        _get_raw_value(actor, "actorGroup"),
+    ]:
+        if actor_group is None:
+            continue
         group_name = _get_first_str_value(actor_group, ("name", "displayName"))
-        if group_name:
-            return group_name
-
-    raw_actor_group = _get_raw_value(actor, "actorGroup")
-    if raw_actor_group is not None:
-        group_name = _get_first_str_value(raw_actor_group, ("name", "displayName"))
         if group_name:
             return group_name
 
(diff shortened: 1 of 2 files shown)


backend/ee/onyx/external_permissions/jira/page_access.py
```diff
@@ -10,18 +11,80 @@
...
 
 def _get_role_id(holder: Holder) -> str | None:
     return holder.get("value") or holder.get("parameter")
 
 
+# Jira SDK resources expose fields as attributes, raw dicts, or nested raw payloads
+# depending on Jira version and endpoint.
+def _get_obj_value(obj: object, field: str) -> object | None:
+    if isinstance(obj, dict):
+        obj_dict = cast(dict[str, object], obj)
```
Bo-Onyx: nit: the key may not be a string, so the cast is a lie to the type checker, which has zero runtime effect, , so if the dict has non-string keys, obj_dict.get(field) is called on an untyped dict anyway. Python's dict.get() won't raise — it just returns None if the key isn't found — so there's no runtime crash. But the cast is misleading.

The simpler and more honest version:
```
  def _get_obj_value(obj: object, field: str) -> object | None:
      if isinstance(obj, dict):
          return obj.get(field)  # type: ignore[union-attr]
      return getattr(obj, field, None)
```

backend/ee/onyx/external_permissions/jira/page_access.py
```diff
@@ -10,18 +11,80 @@
...
+        return obj_dict.get(field)
+    return getattr(obj, field, None)
+
+
+def _get_raw_value(obj: object, field: str) -> object | None:
+    raw = _get_obj_value(obj, "raw")
+    if not isinstance(raw, dict):
+        return None
+    raw_dict = cast(dict[str, object], raw)
+    return raw_dict.get(field)
```
Bo-Onyx: This can call `_get_obj_value` again like 
```
  def _get_raw_value(obj: object, field: str) -> object | None:
      raw = _get_obj_value(obj, "raw")
      return _get_obj_value(raw, field)
```

backend/ee/onyx/external_permissions/jira/page_access.py
```diff
@@ -10,18 +11,80 @@
...
+
+
+def _get_first_str_value(obj: object, fields: tuple[str, ...]) -> str | None:
+    for field in fields:
+        value = _get_obj_value(obj, field) or _get_raw_value(obj, field)
```
Bo-Onyx, line 56: nit: this is like getting a field recursively

```
  def _get_nested_value(obj: object, *fields: str) -> object | None:
      current: object = obj
      for field in fields:
          if isinstance(current, dict):
              current = current.get(field)
          else:
              current = getattr(current, field, None)
          if current is None:
              return None
      return current
```

Maybe we can put into a util, could be helpful for other use case.

backend/ee/onyx/external_permissions/jira/page_access.py
```diff
@@ -98,28 +161,135 @@ def _build_holder_map(permissions: list[dict]) -> dict[str, list[Holder]]:
...
+    if actor_group is not None:
+        group_name = _get_first_str_value(actor_group, ("name", "displayName"))
+        if group_name:
+            return group_name
+
+    raw_actor_group = _get_raw_value(actor, "actorGroup")
+    if raw_actor_group is not None:
+        group_name = _get_first_str_value(raw_actor_group, ("name", "displayName"))
+        if group_name:
+            return group_name
```
Bo-Onyx: same patterns here https://github.com/onyx-dot-app/onyx/pull/11187/changes#r3267892057

Bo-Onyx (approved):

PR #11190, opened 2026-05-19
chore(logging, celery): Single source of truth for logging in celery workers

acaprau:
## Description
This PR makes it such that the Celery log level precedence looks as follows: Celery default -(clobbered by)-> `LOG_LEVEL` env var set (same env var as the rest of the application) -(clobbered by)-> Celery task given explicit log level CLI command. As a result, by default we no longer pass in CLI args to Celery workers, although customers can still opt to do this in their self-hosted deploys.

## How Has This Been Tested?
Added automated tests for the log level logic.

## Additional Options

---
## Summary by cubic
Unifies Celery worker logging with the rest of the app. `LOG_LEVEL` now controls all workers (and beat) by default, while an explicit `--loglevel` remains a per-worker override.

- **Refactors**
  - Precedence: CLI `--loglevel` > `LOG_LEVEL` env var > Celery default; at startup we log the effective level and its source. Invalid or empty `LOG_LEVEL` falls back to INFO and is called out in the message.
  - Detect explicit CLI flags (`--loglevel`, `-l`) via `sys.argv` to treat them as overrides.
  - Config: removed hardcoded `--loglevel` from `backend/supervisord.conf`; Helm renders `--loglevel` only when `.logLevel` is set; defaults in `values.yaml` are `""` to defer to `LOG_LEVEL`.
  - Tests and docs: added unit tests for precedence, CLI forms, case-insensitive parsing, and empty/invalid fallback; documented `LOG_LEVEL` in `deployment/docker_compose/env.template` and Helm `values.yaml`.

- **Migration**
  - Set global verbosity with `LOG_LEVEL` (e.g., `info`, `debug`).
  - Per-worker override: Helm — set `.Values.celery_worker_*.logLevel` or `celery_beat.logLevel`; Supervisor — add `--loglevel=<value>` to that program’s command.

<sup>Written for commit 25c5321c1be3bc8f98f58f613dbb11252e27bce0. Summary will update on new commits. <a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11190?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/backend/onyx/background/celery/apps/app_base.py
+++ b/backend/onyx/background/celery/apps/app_base.py
@@ -432,7 +432,7 @@ def _cli_loglevel_explicitly_set() -> bool:
     """
     Returns True iff --loglevel or -l was explicitly passed on the celery CLI.
 
-    The setup_logging signal handler receives a `loglevel` int regardless of
+    The setup_logging signal handler receives a ``loglevel`` int regardless of
     whether the operator actually passed --loglevel — when the flag is absent,
     Celery substitutes its own default. To distinguish "operator passed it" (CLI
     should win) from "Celery defaulted it" (LOG_LEVEL env should win) we have to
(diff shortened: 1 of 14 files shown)


justin-tahara (approved):

acaprau:
@greptile @cubic

acaprau:
@greptile @cubic

PR #11222, opened 2026-05-20
feat(craft): docker-compose sandbox backend

rohoswagger:
## Description

Adds a Docker backend for Craft sandboxes (`SANDBOX_BACKEND=docker`) so self-hosted docker-compose deployments can run isolated per-user sandbox containers instead of falling back to in-process local mode. `api_server` mounts the host Docker socket and drives sandbox container lifecycle directly via the Docker Engine API, mirroring the K8s manager architecture.

**Stacked on #11218** (PR 1: backend-agnostic cleanup + snapshot stream helpers).

### Components
- `DockerSandboxManager` — per-user containers on a dedicated `onyx_craft_sandbox` bridge with `cap_drop=ALL`, `no-new-privileges`, non-root user, no Docker socket mount, fixed env allowlist (`ONYX_PAT` + `ONYX_SERVER_URL` only), K8s-aligned resource defaults (1 CPU / 2Gi). Auto-detects `com.docker.compose.project` so Docker Desktop groups sandboxes under the same stack as api_server.
- `DockerACPExecClient` — JSON-RPC over a multiplexed `docker exec` socket. Ephemeral process per `send_message` so flat-file session storage isn't corrupted by concurrent replicas.
- `stream_stdin_to_container` / `stream_stdout_from_container` exec helpers — pipe tar bytes for snapshots through api_server-owned `FileStore` (no S3/MinIO creds in the sandbox).
- Compose wiring: `api_server` and `background` mount the Docker socket and join the craft bridge; `SANDBOX_BACKEND` defaults to `docker`.
- `install.sh --include-craft` creates the bridge, warns about the socket trust boundary, and on EC2 installs a `DOCKER-USER` iptables rule to block sandbox→IMDS traffic (or fails unless `CRAFT_ALLOW_UNBLOCKED_IMDS=true`).

### Bug fixes uncovered during end-to-end testing
- `sessions_api.restore_session` now passes `onyx_pat` via `ensure_sandbox_pat` (was missing — provision failed every restore on K8s too).
- Restore rolls the sandbox back to `SLEEPING` on failure rather than leaving it stuck in `PROVISIONING`.
- Snapshot gate broadened from `backend == KUBERNETES` to `backend != LOCAL` so Docker restores actually pull snapshots.
- `_regenerate_session_config` recreates the `.opencode/skills` symlink after restore (the snapshot tar excludes it; restored sessions would otherwise have no skill access).
- `SSEKeepalive` moved to `sandbox/base` so `isinstance` checks in `session/manager` match across backends.
- `_recv_exact` retries through `socket.timeout` rather than discarding partial header/payload bytes mid-frame.

### Out of scope (separate PR)
ACP packet-loss investigation where opencode generates ~30 deltas but only ~10 reach the client in some prompts.

## How Has This Been Tested?

- Unit tests cover config / env allowlist / security options (no Docker socket mount, no S3/MinIO creds in env), naming + labels, snapshot stream helpers, factory dispatch, exec-helper log redaction, and ACP exec client framing/initialize. 121 tests pass.
- Manual smoke script at \`backend/scripts/manual_test_docker_sandbox.py\` drives provision → setup → file ops → terminate against the local Docker daemon end-to-end.
- End-to-end run against a live docker-compose stack: provisioned a sandbox, sent agent messages (full ACP pipeline through SSE), uploaded files, listed directories, viewed the Next.js preview iframe via api_server's proxy, exercised \`/sandbox/reset\` (explicit terminate), and watched the natural celery beat \`cleanup_idle_sandboxes_task\` snapshot the session into MinIO and tear down the container/volume. Restored a session from snapshot (FileStore → \`tar -xzf -\` in a fresh sandbox via stdin stream); verified the prior outputs came back.

## Additional Options


wenxi-onyx (approved):

rohoswagger:
@cubic-dev-ai @greptile-ai re-review this

PR #11234, opened 2026-05-20
chore(deployment): wait for healthy services via docker compose --wait

### LINKED ISSUE #9794: nginx startup stuck in loop: curl returns 000 resolving 'api_server' on Alpine (wget works)
## Description

After starting Onyx with Docker Compose, nginx remains stuck in its startup health-check loop indefinitely, making `localhost:3000` unreachable:

```
Waiting for API server to boot up; this may take a minute or two...
API server responded with 000, retrying in 5 seconds...
API server responded with 000, retrying in 5 seconds...
(repeats forever)
```

## Root Cause

The nginx startup script (`run-nginx.sh`) uses `curl` to health-check the API server:

```sh
status_code=$(curl -o /dev/null -s -w "%{http_code}\n" "http://${ONYX_BACKEND_API_HOST}:8080/health")
```

On Alpine Linux, `curl` fails to resolve `api_server` via Docker's internal DNS (`127.0.0.11`), always returning `000` (connection failure). However, `wget` resolves the same hostname correctly:

```sh
# Inside nginx container:
curl -s http://api_server:8080/health   # → 000 (fails)
wget -qO- http://api_server:8080/health # → {"success":true} (works)
```

The issue appears to be a DNS resolution incompatibility between Alpine's musl-based `curl` and Docker's embedded DNS resolver (`127.0.0.11`) with `search localdomain` in `/etc/resolv.conf`.

## Steps to Reproduce

```bash
docker compose --profile s3-filestore up -d
docker logs onyx-nginx-1 --follow
# Observe "API server responded with 000" repeating indefinitely
```

## Workaround

Manually add the API server IP to `/etc/hosts` inside the nginx container:

```bash
# Get api_server IP
API_IP=$(docker inspect onyx-api_server-1 --format '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}')
docker exec onyx-nginx-1 sh -c "echo '$API_IP api_server' >> /etc/hosts"
```

## Suggested Fix

Replace `curl` with `wget` in the health-check loop in `run-nginx.sh`, or use the IP address directly, or switch to a curl invocation that bypasses DNS (e.g. `--resolve`).

## Environment

- OS: Ubuntu 24.04.4 (VMware)
- nginx image: `nginx:1.25.5-alpine`
- Docker version: latest

jmelahman:
## Summary

Replaces install.sh's hand-rolled "wait 10s then inspect restart counts" with `docker compose up --wait`, and removes nginx's homegrown curl-polling loop in favor of Compose's native `depends_on: condition: service_healthy`. Together this closes #9794 — the curl loop fails on some Alpine + Docker DNS combos, and we no longer rely on it.

### Changes

- **Healthchecks added** to `nginx` (via a dedicated `/nginx-health` location served by nginx itself) and `web_server` (via node's built-in `http` module).
- **`nginx.depends_on`** switched to long-form `condition: service_healthy` on `api_server` and `web_server`. nginx now only starts once its upstreams are actually serving — same guarantee the curl loop was trying to provide.
- **`run-nginx.sh`** loses its curl polling loop (root cause of #9794) since Compose now enforces upstream readiness.
- **`install.sh`** passes `--wait --wait-timeout 600` to `docker compose up` by default. On failure it dumps `docker compose ps` so the user can see exactly which service is unhealthy. A new `--no-wait` flag falls back to the legacy 10s sleep + restart-count inspection for anyone who needs it.
- **`start_period` bumped to 600s** on `api_server` (alembic on fresh DB), `inference_model_server`, and `indexing_model_server` (HuggingFace model downloads). Healthy is still reported the moment the underlying check passes — long `start_period` only matters for slow boots.

### Notes for reviewers

- `--wait-timeout 600` matches the existing `check_onyx_health` 10-minute ceiling, so the worst-case wall-clock is unchanged.
- In lite mode the model servers and other heavy services are gated behind profiles, so `--wait` doesn't try to wait on them.
- The prod compose files (`docker-compose.prod*.yml`) aren't used by `install.sh` and are left untouched; they can be updated in a follow-up if desired.

## Test plan

20260520_11h51m29s_grim

- [ ] Fresh install in **lite mode**: `./install.sh --no-prompt --lite` reaches the success message without errors; `docker compose ps` shows `nginx`, `web_server`, `api_server`, `relational_db` all `(healthy)`.
- [ ] Fresh install in **standard mode**: `./install.sh --no-prompt` succeeds within 10 minutes; model servers reach healthy after HF downloads complete.
- [ ] Reinstall (re-running on an existing `onyx_data/`) succeeds.
- [ ] **Issue #9794 reproducer** (Ubuntu 24.04 + Docker on VMware): nginx starts and serves `localhost:3000` without the previous infinite `responded with 000` loop.
- [ ] **`--no-wait` escape hatch** still produces the legacy output (10s progress bar + per-container restart-count check).
- [ ] **Failure path**: kill a container after `up` starts (or break a healthcheck) and confirm `--wait` exits non-zero and `docker compose ps` is dumped naming the unhealthy service.
- [ ] Manual probes inside running containers: `docker compose exec nginx wget -qO- http://localhost/nginx-health` returns `ok`; `docker compose exec web_server node -e "require('http').get('http://localhost:3000/', r => console.log(r.statusCode))"` prints a status code.


deployment/docker_compose/install.sh
```diff
@@ -242,7 +250,7 @@ NC='\033[0m' # No Color
 
 # Step counter variables
 CURRENT_STEP=0
-TOTAL_STEPS=10
+TOTAL_STEPS=9
```
justin-tahara, line 272: I would not catch things like this

deployment/helm/charts/onyx/templates/nginx-conf.yaml
justin-tahara, line 1: Can we also ensure we are bumping chart.yaml version?

justin-tahara (approved):

PR #11237, opened 2026-05-20
feat(helm): optional External Secrets Operator support

justin-tahara:
## Description

Adds a single `ExternalSecret` CR to the Onyx Helm chart, gated on `externalSecret.enabled`. When enabled, the chart materializes one shared Kubernetes Secret from an upstream secret manager (AWS Secrets Manager, GCP Secret Manager, Vault, etc.) via [External Secrets Operator](https://external-secrets.io/). Each `auth.*` section can then read its specific keys from that materialized Secret by setting `existingSecret`, and a new top-level `extraEnvFromSecret` knob projects the entire materialized Secret into backend pods via `envFrom: secretRef` — covering the loose env-var secrets that today live as plaintext in customers' `configMap:` blocks (e.g. `LINEAR_CLIENT_SECRET`, `BRAINTRUST_API_KEY`, `OAUTH_CLIENT_SECRET`).

Together these let cloud-deployment-yamls customer values stop holding any secret material in git: every secret is sourced from AWS Secrets Manager, and `helm upgrade` works from a developer laptop without any secret material on disk.

**Off by default.** When `externalSecret.enabled=false` (the default) the chart renders functionally identical YAML to 0.5.3 — `helm template` against existing values produces zero diff outside the chart version label and its checksum cascade. Existing self-hosters of the public chart are unaffected.

Chart bumped to 0.5.4. Docs added under `deployment/helm/README.md`.

[ENG-4120](https://linear.app/onyx-app/issue/ENG-4120/helm-chart-optional-external-secrets-operator-support)

### Files touched
- New: `templates/external-secret.yaml`
- `values.yaml`: new `externalSecret:` block + `extraEnvFromSecret`
- 14 backend deployment templates: one `{{- with .Values.extraEnvFromSecret }} ... {{- end }}` block added inside each existing `envFrom:` list. **Model-server deployments (inference + indexing) deliberately excluded** — they only need model-config env vars from the chart's configMap, not app-level secrets.
- `Chart.yaml`: version bump + changelog entry
- `deployment/helm/README.md`: new "Sourcing secrets from an external secret manager (ESO)" section, including notes on the first-install reconciliation window and why `secretRef` is non-optional.

## How Has This Been Tested?

- `helm lint .` clean (both default and with `externalSecret.enabled=true --set externalSecret.refPath=... --set extraEnvFromSecret=...`).
- `helm template` with defaults diffed against 0.5.3 baseline: only differences are the `helm.sh/chart` label and the resulting `checksum/config` annotation cascade (expected for any version bump). No functional changes.
- `helm template` with `externalSecret.enabled=true`: renders exactly one `ExternalSecret` CR (now `apiVersion: external-secrets.io/v1`) and adds `- secretRef:` to the `envFrom` of every default-enabled backend deployment (12 total, model servers excluded).
- Fail guards verified for empty `refPath`, empty `secretStoreRef.name`, and empty `secretStoreRef.kind`.

End-to-end cluster validation (ESO actually pulling from AWS Secrets Manager) will happen in the cloud-deployment-yamls cutover PR, which depends on this chart release.

## Additional Options

### DIFF
--- a/deployment/helm/README.md
+++ b/deployment/helm/README.md
@@ -71,6 +71,56 @@ auth:
       REDIS_PASSWORD: redis_password
 ```
 
+## Sourcing secrets from an external secret manager (ESO)
+
+The chart can render an `ExternalSecret` CR that has [External Secrets
+Operator](https://external-secrets.io/) materialize a single Kubernetes Secret
+from an upstream provider (AWS Secrets Manager, GCP Secret Manager, Vault,
+etc.). All `auth.*` sections then read their per-section keys from that
+materialized Secret, and any loose env-var secrets are projected into pods
+via `envFrom`.
+
+**Prerequisite:** the chart does *not* install ESO. Install it separately and
+create a `SecretStore` or `ClusterSecretStore` (e.g. `aws-secrets-manager`)
+before enabling `externalSecret`. See https://external-secrets.io/ for setup.
+
+The upstream secret should be a JSON object whose top-level keys map 1:1 to
+the keys the chart expects in the materialized Secret (the union of every
+`auth.<section>.secretKeys` value plus any loose env-var names you project via
+`extraEnvFromSecret`).
+
+```yaml
+externalSecret:
+  enabled: true
+  # Identifier of the upstream secret (e.g. AWS Secrets Manager secret name).
+  refPath: "onyx/<customer>/app-secrets"
+  # Name of the materialized k8s Secret. Point auth.*.existingSecret here.
+  secretName: onyx-app-secrets
+  secretStoreRef:
+    name: aws-secrets-manager
+    kind: ClusterSecretStore
+  refreshInterval: 1h
+
+# Inject every key from onyx-app-secrets as an env var on backend pods.
+# Use for secrets that today live as plaintext in `configMap:` blocks.
+extraEnvFromSecret: onyx-app-secrets
+
+auth:
+  postgresql:
+    existingSecret: onyx-app-secrets
+    secretKeys:
+      POSTGRES_USER: postgres_user
+      POSTGRES_PASSWORD: postgres_password
+  redis:
+    existingSecret: onyx-app-secrets
+    secretKeys:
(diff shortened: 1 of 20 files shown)


justin-tahara:
Thanks @cubic-dev-ai @greptile-apps — addressed in 4ab7c93f12. Summary:

**Fixed**
- 🔧 `external-secrets.io/v1beta1` → `external-secrets.io/v1` (cubic P1). Agreed: v1beta1 is no longer served on ESO v0.17+.
- 🔧 Added fail guards for empty `externalSecret.secretStoreRef.name` and `secretStoreRef.kind` (greptile issue 1). Now any misconfig surfaces at `helm template` time with a clear message rather than as a cryptic ESO reconcile error.
- 🔧 Removed the `extraEnvFromSecret` `envFrom` block from `indexing-model-deployment.yaml` and `inference-model-deployment.yaml` (greptile issue 3). Model servers don't need app-level secrets and the chart's existing `configMap` already covers their config. Now 14 → 12 deployments inject the secret. README updated to call this out.

**Documented, intentionally not coded**
- 📝 First-install reconciliation window (cubic P2, greptile issue 2). I considered marking `secretRef` as `optional: true` but settled on leaving it required and documenting the brief `CreateContainerConfigError` window instead. Rationale (now in README): `optional: true` would let pods boot with empty env vars and crash-loop later in a way that's harder to diagnose. With required, the kubelet retries pod creation every few seconds and the pod transitions to Running automatically once ESO reconciles (~5-15s on a healthy install). The clearly-broken-then-healed failure mode is a better signal than the start-then-fail one if the SM blob is misconfigured (wrong path, missing keys, IRSA permissions wrong). Happy to revisit if you'd prefer the optional path.

Re-validated: `helm lint` clean, default render functionally unchanged, enabled render now uses `external-secrets.io/v1`, fail guards fire on each empty field, 12 backend deployments get the `secretRef`, model servers do not.

jmelahman (approved):

PR #11250, opened 2026-05-20
fix(install): default to latest GitHub release for configs and image tag

### LINKED ISSUE #11227: Onyx edge and latest tags require Vespa but docker-compose lacks Vespa service
Problem:
- The edge image tag defaults to using Vespa as the search backend (tries to connect to localhost:19071)
- The provided docker-compose.yml only includes OpenSearch - no Vespa service
- This causes the api_server to crash during initialization with:
    requests.exceptions.ConnectionError: HTTPConnectionPool(host='localhost', port=19071): 
  Failed to establish a new connection: [Errno 111] Connection refused"

Expected:
- Either the docker-compose should include a working Vespa configuration, OR
- There should be an env var to switch back to OpenSearch (e.g., SEARCH_BACKEND=opensearch)

jmelahman:
## Summary
- `install.sh` now calls the GitHub releases API to discover the most recent published Onyx release tag.
- That tag becomes the git ref used to download `docker-compose.yml`, the lite/craft overlays, and the nginx templates, so the config files match the images that will actually be pulled.
- The version prompt's default switches from the rolling `edge` tag to the resolved release tag (so non-interactive / `--no-prompt` runs land on a tested version too).
- If the API is unreachable, falls back to the previous behavior (`main` for configs, `edge` for the image tag) with a warning.
- Floating tags (`edge`, `latest`, `craft-*`) still pull configs from `main`; pinned release tags pull configs from their own ref. A new `CONFIG_REF` variable drives the re-download so we only re-fetch when the chosen ref differs from the initial download ref.

Closes https://github.com/onyx-dot-app/onyx/issues/11227

## Test plan

20260520_15h43m32s_grim

- [ ] `bash -n install.sh` (syntax check) — done locally
- [ ] `./install.sh --dry-run` shows the resolved default tag + config ref
- [ ] `./install.sh --dry-run` with network blocked falls back to `main` / `edge`
- [ ] Fresh install in a clean dir: accept default → ends up on the latest release tag, configs come from that tag
- [ ] Fresh install, type `edge` at the prompt → configs re-fetched from `main`, images pull `edge`
- [ ] Fresh install, type a known pinned tag (e.g. previous release) → configs re-fetched from that tag

---
## Summary by cubic
Default the installer to the latest GitHub release for both the image tag and config downloads, and ensure Docker Compose uses the selected tag even if the shell has a conflicting `IMAGE_TAG`. Adds bounded API timeouts and offline `--local` behavior; falls back to `main`/`edge` with a warning.

- **New Features**
  - Resolve and display the latest release via the GitHub Releases API.
  - Use that tag for config URLs via `RELEASE_REF`; prompt defaults via `DEFAULT_IMAGE_TAG`.
  - Floating tags (`edge`, `latest`, `craft-*`) keep configs from `main`; pinned tags use their own ref.
  - Re-fetch configs when needed via `CONFIG_REF` so files match the selected tag; `--dry-run` shows the resolved defaults.

- **Bug Fixes**
  - Pass `IMAGE_TAG` inline to Compose (pull/up/ps) so the user’s choice overrides any shell export without leaking env vars.
  - Bound GitHub API timeouts, skip the release lookup in `--local`, and print the fallback warning after the banner.

<sup>Written for commit c0a189a5328b2b88e178118d8e7376ba7c00c879. Summary will update on new commits. <a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11250?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/deployment/docker_compose/install.sh
+++ b/deployment/docker_compose/install.sh
@@ -161,6 +161,35 @@ download_file() {
     fi
 }
 
+# Fetches the most recent published release tag from the Onyx GitHub repo.
+# Used to default the installer to a pinned, tested release (instead of the
+# rolling "edge" tag) and to pull compose/nginx files that match it.
+LATEST_RELEASE_TAG=""
+fetch_latest_release_tag() {
+    local api_url="https://api.github.com/repos/onyx-dot-app/onyx/releases/latest"
+    local response=""
+    if [[ "$DOWNLOADER" == "curl" ]]; then
+        response=$(curl -fsSL --retry 2 --retry-delay 1 \
+            --connect-timeout 5 --max-time 10 \
+            -H "Accept: application/vnd.github+json" \
+            "$api_url" 2>/dev/null) || return 1
+    else
+        response=$(wget -q --tries=2 --timeout=10 \
+            --header="Accept: application/vnd.github+json" \
+            -O- "$api_url" 2>/dev/null) || return 1
+    fi
+    local tag
+    tag=$(echo "$response" \
+        | grep -oE '"tag_name"[[:space:]]*:[[:space:]]*"[^"]+"' \
+        | head -1 \
+        | sed -E 's/.*"tag_name"[[:space:]]*:[[:space:]]*"([^"]+)".*/\1/')
+    if [[ -z "$tag" ]]; then
+        return 1
+    fi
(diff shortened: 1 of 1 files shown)


acaprau (approved):
my guy

PR #11289, opened 2026-05-21
chore(ci): drop backend image build from integration tests

jmelahman:
## Summary

- Embed api_server (uvicorn) and the celery worker set inside the test-runner container instead of running them as separate `docker-compose` services. This deletes the `build-backend-image` job and removes `ONYX_BACKEND_IMAGE` from the workflow entirely, shaving the full build-backend-image wall-clock off every PR.
- Adds `supervisor` and `opencode-ai` (via bun) to the devcontainer so it can host both the FastAPI server and the celery workers that the suite already exercises. Postgres / OpenSearch / Redis / Minio / model servers stay in compose unchanged.
- New `backend/tests/integration/scripts/run-with-backend.sh` is the docker entrypoint for all three test jobs: it runs `alembic upgrade head`, sets up Craft templates when `ENABLE_CRAFT=true`, starts supervisord (`full`) or direct uvicorn (`api_only` for onyx-lite, where redis/celery are unavailable), waits for `/health`, then `exec`s pytest with whatever args were passed.

Includes a small follow-up to `setup_craft_templates.sh` so it falls back to `bun install` when npm is not on PATH (the devcontainer ships bun, not Node).

The previous remote-supervisord healthcheck step is replaced by the equivalent local `supervisorctl status` inside the new script. `slack_bot` / `discord_bot` / `watchdog` are still allowed to flap (no creds in CI), matching the old behavior.

Plan: `plans/nifty-yawning-rainbow.md`.

## Test plan

- [ ] `integration-tests` matrix is green (in particular `tests/craft/*`, which exercises the Craft prerequisites inside the test runner).
- [ ] `onyx-lite-tests` is green via the `api_only` path (uvicorn only, no celery, no redis).
- [ ] `multitenant-tests` is green under the `full` path with `MULTI_TENANT=true` / `AUTH_TYPE=cloud`.
- [ ] Workflow graph no longer contains `build-backend-image`; `api_server` / `background` no longer appear in the docker-compose log artifacts.
- [ ] Wall-clock for the integration suite drops by the full duration of today's `build-backend-image` job.

---
## Summary by cubic
Run the backend inside the test-runner devcontainer and drop the compose `api_server`/`background` services and backend image build. This simplifies CI and speeds up the integration suites.

- **Refactors**
  - Add `backend/tests/integration/scripts/run-with-backend.sh` to run migrations, optional Craft setup, install Playwright Chromium, start `uvicorn` (and `supervisor` in `full` mode), wait for `/health`, then exec pytest.
  - Add `supervisor`, `opencode-ai` (via `bun`, pinned via `OPENCODE_VERSION`), and Playwright Chromium deps to the devcontainer; Craft template setup falls back to `bun install` when `npm` isn’t present.
  - Replace the bespoke integration-test image with the devcontainer; CI bind-mounts the repo and runs the script via `--entrypoint bash`; remove the `build-backend-image` job and stop composing `api_server`/`background`.

- **Bug Fixes**
  - Set `API_SERVER_HOST=127.0.0.1` and pass `MODEL_SERVER_HOST`/`INDEXING_MODEL_SERVER_HOST`; for Onyx Lite, skip model server, Redis, S3, and MinIO and use Postgres-backed stores.
  - In `MULTI_TENANT=true`, run Alembic with the `schema_private` config and forward `ENABLE_PAID_ENTERPRISE_EDITION_FEATURES`, `LICENSE_ENFORCEMENT_ENABLED`, and `OPENAI_DEFAULT_API_KEY` so tenant seeding runs and related tests pass.
  - Point Craft template paths to `/workspace/...`; make the Playwright platform override arch-aware and preinstall Chromium deps so it works on Ubuntu 26.04 and x64/arm64 hosts.
  - Ensure the `supervisord` cleanup can’t hang by sending SIGTERM after `supervisorctl shutdown` when the socket never appears.

<sup>Written for commit 1d4739d93c87dc69d9d0b3cd9c9b79e1d2a6c97e. Summary will update on new commits. <a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11289?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/backend/tests/integration/scripts/run-with-backend.sh
+++ b/backend/tests/integration/scripts/run-with-backend.sh
@@ -1,0 +1,108 @@
+#!/usr/bin/env bash
+# Boots the in-container backend (alembic migrations, optional Craft templates,
+# uvicorn api_server, and — in `full` mode — supervisord-managed celery workers)
+# before invoking pytest with whatever args were passed through.
+#
+# Modes (RUN_WITH_BACKEND_MODE):
+#   full     — integration / multitenant suites: uvicorn for api_server plus
+#              supervisord for the full celery worker set (the workers
+#              autostart from backend/supervisord.conf).
+#   api_only — onyx-lite suite: redis/celery aren't available, so we only
+#              start uvicorn.
+set -euo pipefail
+
+MODE="${RUN_WITH_BACKEND_MODE:-full}"
+BACKEND_DIR="/workspace/backend"
+SUPERVISORD_CONF="${BACKEND_DIR}/supervisord.conf"
+API_HEALTH_URL="http://127.0.0.1:8080/health"
+HEALTH_TIMEOUT_SECONDS="${API_SERVER_HEALTH_TIMEOUT_SECONDS:-600}"
+
+cd "$BACKEND_DIR"
+
+echo "==> Running alembic upgrade head"
+uv run --no-sync alembic upgrade head
+
+if [[ "${ENABLE_CRAFT:-false}" == "true" ]]; then
+    echo "==> Setting up Craft templates"
+    bash "${BACKEND_DIR}/scripts/setup_craft_templates.sh"
+fi
+
+SUPERVISORD_PID=""
+UVICORN_PID=""
+
+cleanup() {
+    local exit_code=$?
+    if [[ -n "$UVICORN_PID" ]] && kill -0 "$UVICORN_PID" 2>/dev/null; then
+        kill "$UVICORN_PID" 2>/dev/null || true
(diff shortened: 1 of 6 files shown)


backend/tests/integration/scripts/run-with-backend.sh
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env bash
```
jmelahman, line 1: FYI, I think this entire file can be removed in favor of using FastAPI's `TestClient` in-process, but that's a high-touch change while this is the minimum needed to drop depending on the `api_server` container, so will leave for a follow up.

PR #11293, opened 2026-05-21
feat(connector): Index documents on github connector

Danelegend:
## Description
Github repos can contain documents valuable for indexing. This change creates an additional toggle on the github connector that can allow for documents from repos to be indexed.

This is done by iterating through the repo and finding documents with the following extensions:
.md, .mdx, .markdown, .rst, .txt

## How Has This Been Tested?
Manually + some tests

## Additional Options

---
## Summary by cubic
Adds optional document indexing to the GitHub connector. When enabled, we crawl the default branch and index text docs (.md, .mdx, .markdown, .rst, .txt) and common extensionless docs (README, LICENSE, etc.).

- **New Features**
  - New `include_files` setting and UI checkbox (“Include Documents?”) to index repository documents from the default branch.
  - Indexes only human-readable files; recognizes extensionless docs; skips denylisted paths, files >1MB, and non-text/undecodable content. Produces documents with blob links, file-based identifiers, and hierarchy metadata; surfaces truncated trees as failures.
  - Adds a FILES stage with checkpointed pagination (100 files per batch) and a `pushed_at` gate to skip unchanged repos. Validation now requires at least one of PRs, issues, or files to be enabled.

- **Bug Fixes**
  - Skip PRs/issues stages when GitHub returns 404 (feature disabled) instead of failing.
  - Prevent stage regression when resuming FILES pagination so each file is indexed exactly once.

<sup>Written for commit c566ebd8c5d78e02fe7015b69387f70bf2b07799. Summary will update on new commits. <a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11293?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/backend/onyx/connectors/github/connector.py
+++ b/backend/onyx/connectors/github/connector.py
@@ -1,10 +1,12 @@
 import copy
+import os
 from collections.abc import Callable
 from collections.abc import Generator
 from datetime import datetime
 from datetime import timedelta
 from datetime import timezone
 from enum import Enum
+from io import BytesIO
 from typing import Any
 from typing import cast
 
@@ -47,6 +49,8 @@ from onyx.connectors.models import DocumentFailure
 from onyx.connectors.models import HierarchyNode
 from onyx.connectors.models import SlimDocument
 from onyx.connectors.models import TextSection
+from onyx.file_processing.extract_file_text import detect_encoding
+from onyx.file_processing.extract_file_text import is_text_file
 from onyx.utils.logger import setup_logger
 
 logger = setup_logger()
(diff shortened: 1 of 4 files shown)


backend/onyx/connectors/github/connector.py
```diff
@@ -58,6 +64,77 @@
...
+    if segments & GITHUB_PATH_DENYLIST:
+        return False
+
+    basename = path.rsplit("/", 1)[-1]
+    _, extension = os.path.splitext(basename)
+    if extension.lower() in GITHUB_INDEXABLE_FILE_EXTENSIONS:
+        return True
+
+    # Extensionless docs like README / LICENSE (basename has no extension).
+    if not extension and basename.lower() in GITHUB_INDEXABLE_FILENAMES:
```
evan-onyx, line 132: you can return the condition here, but if you think its cleaner the current way i don't mind

backend/onyx/connectors/github/connector.py
```diff
@@ -397,10 +484,71 @@ def _convert_issue_to_document(
     )
 
 
+def _decode_file_content(raw: bytes) -> str | None:
```
evan-onyx, line 487: I'm suspicious that we don't already have a helper for this, can you check?

backend/onyx/connectors/github/connector.py
```diff
@@ -569,6 +725,65 @@ def _issues_func(
...
+        try:
+            git_tree = repo.get_git_tree(repo.default_branch, recursive=True)
+            truncated = bool(git_tree.raw_data.get("truncated"))
+            if truncated:
+                logger.error(
+                    "Git tree for repo %s was truncated by GitHub; "
+                    "some files will not be indexed",
+                    repo.full_name,
+                )
+            paths = [
```
evan-onyx, line 748: is there no way to filter server side? unlucky if so

backend/onyx/connectors/github/connector.py
```diff
@@ -765,10 +983,108 @@ def _fetch_from_github(
...
+        # checkpoint (mid file-pagination) — that would null out file_paths and
+        # re-index from page 0.
+        if checkpoint.stage in (
+            GithubConnectorStage.PRS,
+            GithubConnectorStage.ISSUES,
+        ):
+            checkpoint.stage = GithubConnectorStage.FILES
+
+        if self.include_files and checkpoint.stage == GithubConnectorStage.FILES:
+            if checkpoint.file_paths is None:
```
evan-onyx: can we put this inner if block in a helper func

evan-onyx (approved):
I have some thoughts but broadly looks good!

backend/onyx/connectors/github/connector.py
Danelegend: Might keep as is, since returning condition is a bit less clean

backend/onyx/connectors/github/connector.py
Danelegend: no :(

PR #11323, opened 2026-05-22
feat(voice): migrate OpenAI streaming STT to GA Realtime API

justin-tahara:
## Description

Long-term followup to #11322 (the small chunked-fallback fix that's keeping OpenAI STT working in production today). This PR migrates `OpenAIStreamingTranscriber` to OpenAI's GA Realtime API so customers get native streaming transcripts again.

The GA shape is non-obvious — the docs were ambiguous and contradictory across pages, so this PR was iterated against the live API with `backend/tests/external_dependency_unit/voice/test_openai_streaming.py` until both the test and a developer running it locally end-to-end confirmed transcription works.

## Final wire shape

**WebSocket URL** — `wss://api.openai.com/v1/realtime?model=gpt-realtime-1.5`

- `?model=` must specify a **realtime session model** (speech-to-speech), not a transcription model. `OPENAI_REALTIME_SESSION_MODEL` constant.
- The `OpenAI-Beta: realtime=v1` header is no longer required (GA endpoint).
- The Beta `?intent=transcription` form now returns `beta_api_shape_disabled`.

**First message (`session.update`)** — `session.type` is **`"realtime"`**, NOT `"transcription"`:

```json
{
  "type": "session.update",
  "session": {
    "type": "realtime",
    "audio": {
      "input": {
        "format": {"type": "audio/pcm", "rate": 24000},
        "transcription": {"model": "gpt-realtime-whisper"},
        "turn_detection": null
      }
    }
  }
}
```

GA does NOT have a dedicated transcription-only session type on the WebSocket. Per OpenAI's own cookbook (`realtime_out_of_band_transcription`), transcription is configured as a side-effect of a standard realtime session via `audio.input.transcription.model`. Sending `session.type: "transcription"` yields:

> `invalid_parameter` — Passing a transcription session update event to a realtime session is not allowed.

**Model handling.** Two separate model slots, two separate constants:

| Constant | Value | Where it goes |
|---|---|---|
| `OPENAI_REALTIME_SESSION_MODEL` | `gpt-realtime-1.5` | URL `?model=` query param (realtime session model) |
| `OPENAI_REALTIME_STT_MODEL` | `gpt-realtime-whisper` | `audio.input.transcription.model` (transcription model) |

The provider's configured `stt_model` (`whisper-1` / `gpt-4o-transcribe` / etc.) continues to govern the chunked HTTP path. The streaming WS always uses the two constants above — existing customers don't need to change settings.

**Event names that did NOT change in GA.** The `_receive_loop` code path survives intact:
- `conversation.item.input_audio_transcription.delta`
- `conversation.item.input_audio_transcription.completed`
- `input_audio_buffer.committed`
- `error`

Session lifecycle events dropped the `transcription_` prefix:
- `session.created` (was `transcription_session.created`)
- `session.updated` (was `transcription_session.updated`)

## UX trade-off — auto-send on silence is lost

`gpt-realtime-whisper` does not support server-side VAD. The Beta code drove auto-send via `speech_stopped` events; the GA path now relies on the user manually stopping the recording, which triggers `input_audio_buffer.commit` from `close()`. The final transcript still arrives; intermediate VAD-driven turn boundaries do not.

Two options for restoring auto-send as a followup: (a) client-side VAD in the browser (e.g. `@ricky0123/vad`), or (b) wait for OpenAI to add VAD support to `gpt-realtime-whisper`. I'd recommend (a) — keeps the UX consistent across providers.

## How Has This Been Tested?

**Live regression tests** at `backend/tests/external_dependency_unit/voice/test_openai_streaming.py`:

1. **`test_streaming_connect_uses_ga_realtime_shape`** — handshake completes without an OpenAI `error` event. Asserts `_last_error is None` after sleeping 1.5s post-connect.
2. **`test_streaming_accepts_pcm16_audio_chunks`** — sends 2s of PCM16 audio in 100ms chunks, calls `close()`, asserts the protocol round-trip works AND `_last_error is None`.
3. **`test_chunked_transcribe_accepts_pcm16`** — `OpenAIVoiceProvider.transcribe()` wraps PCM16 in WAV before posting to `/v1/audio/transcriptions`. Guards the chunked fallback that #11322 introduced.

Tests use `OPENAI_API_KEY` via `pytest.mark.secrets` and run on every PR (not gated by `nightly`). All 3 pass in CI.

**End-to-end manual validation:** a developer ran the full Onyx stack locally with this branch, recorded audio in the chat UI, and got transcripts back from OpenAI. Voice is `Whisper (Current Default)` in the admin Voice config; the WS opens, sends `session.update`, receives `session.created` / `session.updated`, and produces `conversation.item.input_audio_transcription.completed` events.

### Why the test was previously broken — and what changed

The test passed in CI even when the live API was rejecting our `session.update` with `session.type: "transcription"`. Root cause: OpenAI does NOT close the WebSocket when it rejects a session.update — it sends an `error` event over the open socket and leaves the connection alive. The original assertion (`not _ws.closed`) stayed True against a poisoned session that would never produce transcripts.

Commit `d9341742dc` captures the error event on `OpenAIStreamingTranscriber._last_error` and the tests now assert `_last_error is None`. This is the generalizable lesson — when an external service signals errors out-of-band, "is the connection alive?" is not a sufficient health check.

## Iteration history

For posterity, the live API surfaced each layer of the GA shape in turn:

| Attempt | Change | OpenAI response |
|---|---|---|
| Initial (Beta shape, untouched) | `transcription_session.update` + `OpenAI-Beta` header | `beta_api_shape_disabled` |
| First GA guess | `?model=gpt-realtime-whisper` | `invalid_model` — transcription model can't be session model |
| Strip model param | bare `/v1/realtime` | `missing_model` — must provide a model parameter |
| Use realtime model in URL | `?model=gpt-realtime-1.5` + `session.type: "transcription"` | `invalid_parameter` — transcription session update event not allowed on realtime session |
| ✅ Final | `?model=gpt-realtime-1.5` + `session.type: "realtime"` | Works |

## Additional Options

## Notes for reviewers

- Builds on / supersedes parts of #11322. Already rebased on top of main; the small fix's PCM16→WAV wrap remains as defense-in-depth for the chunked path.
- `supports_streaming_stt()` returns `True` again in this PR. #11322 flipped it to `False` as a workaround; this PR's working streaming path makes that workaround unnecessary.

### DIFF
--- a/backend/onyx/voice/providers/openai.py
+++ b/backend/onyx/voice/providers/openai.py
@@ -22,6 +22,7 @@ import io
 import json
 from collections.abc import AsyncIterator
 from enum import StrEnum
+from typing import Any
 from typing import TYPE_CHECKING
 
 import aiohttp
@@ -103,6 +104,10 @@ class OpenAIStreamingTranscriber(StreamingTranscriberProtocol):
         self._accumulated_transcript = ""  # Accumulated across all turns
         self._receive_task: asyncio.Task | None = None
         self._closed = False
+        # Captures the most recent `error` event from OpenAI so callers
+        # (e.g. tests, the WS handler) can detect a poisoned session even
+        # though OpenAI does not close the socket on protocol errors.
+        self._last_error: dict[str, Any] | None = None
 
     async def connect(self) -> None:
         """Establish WebSocket connection to OpenAI Realtime API (GA shape)."""
@@ -127,15 +132,21 @@ class OpenAIStreamingTranscriber(StreamingTranscriberProtocol):
             self._logger.error("Failed to connect to OpenAI Realtime API: %s", e)
             raise
 
-        # GA shape: `session.update` (was `transcription_session.update`), with
-        # the session typed as `transcription` and audio config nested under
-        # `audio.input`. `gpt-realtime-whisper` does not support server-side
-        # VAD — turn boundaries are driven by `input_audio_buffer.commit`
-        # which is sent from `close()` when the user stops recording.
+        # GA shape: `session.update` (was `transcription_session.update`),
+        # with the session typed as `realtime` — *not* `transcription`.
+        # OpenAI's GA does not have a dedicated transcription-only session
+        # type on the WS; transcription is configured as a side-effect of
+        # a standard realtime session via `audio.input.transcription.model`.
+        # Sending `session.type: "transcription"` here yields:
+        #   "Passing a transcription session update event to a realtime
+        #    session is not allowed."
+        # `gpt-realtime-whisper` does not emit server-side VAD events; turn
+        # boundaries are driven by an explicit `input_audio_buffer.commit`
+        # from `close()` when the user stops recording.
         config_message = {
             "type": "session.update",
             "session": {
-                "type": "transcription",
+                "type": "realtime",
                 "audio": {
                     "input": {
                         "format": {"type": "audio/pcm", "rate": 24000},
@@ -163,9 +174,15 @@ class OpenAIStreamingTranscriber(StreamingTranscriberProtocol):
                     msg_type = data.get("type", "")
                     self._logger.debug("Received message type: %s", msg_type)
 
-                    # Handle errors
+                    # Handle errors. OpenAI does NOT close the WebSocket on
+                    # protocol errors (e.g. a malformed `session.update`),
+                    # so we capture the most recent error here for callers
+                    # to inspect — the test/handler can fail fast instead
+                    # of waiting on a session that will never produce a
+                    # transcript.
                     if msg_type == OpenAIRealtimeMessageType.ERROR:
                         error = data.get("error", {})
+                        self._last_error = error
                         self._logger.error("OpenAI error: %s", error)
                         continue
 
--- a/backend/onyx/voice/providers/openai.py
+++ b/backend/onyx/voice/providers/openai.py
@@ -35,8 +35,15 @@ if TYPE_CHECKING:
 DEFAULT_OPENAI_API_BASE = "https://api.openai.com"
 
 
+# OpenAI Realtime API only streams transcription for `gpt-realtime-whisper`.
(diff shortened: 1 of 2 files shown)


justin-tahara:
Rebased on top of main now that #11322 (small fix) has merged, and removed the `pytestmark = pytest.mark.nightly` gate so CI actually runs the new tests.

**Why CI was failing:** The `external-dependency-unit-tests (voice)` job runs `pytest -m "not nightly"`. With my 3 tests all marked `nightly`, the job collected 3 / deselected 3 / selected 0, and pytest exits with code 5 (no tests collected) → job marked failed. The `@pytest.mark.secrets(TestSecret.OPENAI_API_KEY)` decorator already handles skip-if-missing-key, so the extra `nightly` gate was redundant for these short single-API-call tests.

**Why the merge conflict happened:** #11322 updated the module docstring and the `OPENAI_STT_MODELS` comment with "streaming disabled" language. This PR re-enables streaming via GA, so I took the post-GA descriptions from this branch (already addressed in the doc-cleanup commit from Greptile review).

Diff against main is now the GA migration only — the small fix's changes are baseline.

backend/onyx/voice/providers/openai.py
```diff
@@ -85,18 +104,26 @@ def __init__(
...
+        # though OpenAI does not close the socket on protocol errors.
+        self._last_error: dict[str, Any] | None = None
 
     async def connect(self) -> None:
-        """Establish WebSocket connection to OpenAI Realtime API."""
+        """Establish WebSocket connection to OpenAI Realtime API (GA shape)."""
         self._session = aiohttp.ClientSession()
 
-        # OpenAI Realtime transcription endpoint
+        # GA Realtime API: `?model=` must be a realtime *session* model
```
Danelegend: Can we keep comments concise pls

Danelegend (approved):

PR #11327, opened 2026-05-22
fix(ods): Forward SSH agent via Docker Desktop helper on macOS

acaprau:
## Description

`ods dev up` currently fails on macOS with current Docker Desktop:

docker: Error response from daemon: invalid mount config for type "bind":
bind source path does not exist:
/socket_mnt/private/var/run/com.apple.launchd./Listeners

`sshAgentMount()` in `tools/ods/cmd/dev_up.go` passes `$SSH_AUTH_SOCK`
straight to `docker run --mount` as the bind source. On macOS that
variable is a launchd socket (`/var/run/com.apple.launchd.*/Listeners`).
The host-side `os.Stat` succeeds because the path exists on the Mac, but
Docker Desktop's host-to-VM path translation (the `/socket_mnt/...` remap
shown in the error) is currently broken for those paths, so the whole
`docker run` aborts before the container ever starts.

This kills `ods dev up` for every Mac contributor and doesn't have a
self-evident workaround — the only escape was to unset `SSH_AUTH_SOCK`
before invoking, which also disables in-container SSH-based git.

On macOS, when `$SSH_AUTH_SOCK` is unset or points to the default launchd
socket (`com.apple.launchd`), use Docker Desktop's purpose-built
ssh-agent forwarding socket at `/run/host-services/ssh-auth.sock`. That
path lives inside the Docker Desktop VM and is wired by Docker Desktop
directly to the host's ssh-agent — no path translation needed. This is
the officially documented Docker Desktop method:
https://docs.docker.com/desktop/networking/#ssh-agent-forwarding

Users with custom SSH agents (1Password, GPG, Secretive, etc.) at normal
filesystem paths are unaffected — their `$SSH_AUTH_SOCK` is bind-mounted
directly. Linux behavior is also unchanged.

## How Has This Been Tested?

- `go build ./...` clean.
- `ods dev up` on macOS / Docker Desktop 28.5.2 (engine 29.4.3): the
  emitted `docker run` line now shows
  `--mount type=bind,src=/run/host-services/ssh-auth.sock,dst=/tmp/ssh-agent.sock`
  and the container actually starts.

## Additional Options

---
## Summary by cubic
Fixes `ods dev up` on macOS by using Docker Desktop’s SSH agent helper when the default launchd socket can’t be bind-mounted, so containers start and SSH-based git works. Linux and custom agent setups remain unchanged.

- **Bug Fixes**
  - On macOS, when `$SSH_AUTH_SOCK` is unset or a `com.apple.launchd*` path, bind `/run/host-services/ssh-auth.sock` to `/tmp/ssh-agent.sock` in `docker run` (with a debug log when used).
  - Keep direct bind-mount of `$SSH_AUTH_SOCK` for other paths; retain warnings when unset or inaccessible.

<sup>Written for commit a2c584a3967d696a8f36a00647070f2c8b5acf80. Summary will update on new commits. <a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11327?utm_source=github">Review in cubic</a></sup>


jmelahman (approved):

acaprau:
@greptile @cubic

PR #11344, opened 2026-05-23
fix(images): deduplicate project images and add opt-in Azure cap

Subash-Mohan:
## Description
  - Fix: project-attached images were being sent to the LLM twice per request when used in a chat under a Project, doubling image-token cost and surfacing the Azure 50-image-per-request ceiling at half the expected count. Each project image now reaches the model exactly once per turn.
  - Add: opt-in `ENABLE_AZURE_IMAGE_CAP` env var that caps outgoing requests at 50 images for Azure-family providers (matching the documented Azure OpenAI ceiling, [Microsoft Learn — Quotas and Limits (https://learn.microsoft.com/en-us/azure/foundry/openai/quotas-limits)). When engaged, the most recent attachments are kept, older ones are dropped, and a `<system-reminder>` notice is appended so the model can acknowledge missing context. Off by default; non-Azure providers are never capped.

## How Has This Been Tested?
  - New unit tests in `test_llm_step.py`, `test_chat_utils.py`, and updated tests in `test_llm_loop.py`. Run with `pytest
  backend/tests/unit/onyx/chat/`.

## Additional Options

---
## Summary by cubic
Stops sending project images twice and ensures each image is included once per turn. Adds an opt-in 50-image cap for Azure providers with a clear reminder when images are dropped.

- **Bug Fixes**
  - Project images are no longer duplicated; they appear exactly once on the last user message per turn.

- **New Features**
  - Opt-in Azure image cap: when `ENABLE_AZURE_IMAGE_CAP=true`, cap Azure-family providers to 50 images per request; keep the most recent images across turns and prefer earlier attachments within a message; non-Azure providers are never capped.
  - If images are dropped, append a trailing `<system-reminder>` user message so the model knows some context was omitted.

<sup>Written for commit feee087e60971f2badcbb6a01e266aa9f5fd9cdf. Summary will update on new commits. <a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11344?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/backend/onyx/chat/llm_loop.py
+++ b/backend/onyx/chat/llm_loop.py
@@ -485,16 +485,6 @@ def construct_message_history(
                         forgotten_meta, token_counter
                     )
 
-    # Attach project images to the last user message
-    if context_files and context_files.image_files:
-        existing_images = last_user_message.image_files or []
-        last_user_message = ChatMessageSimple(
-            message=last_user_message.message,
-            token_count=last_user_message.token_count,
-            message_type=last_user_message.message_type,
-            image_files=existing_images + context_files.image_files,
-        )
-
     # Build the final message list according to README ordering:
     # [system], [history_before_last_user], [custom_agent], [context_files],
     # [forgotten_files], [last_user_message], [messages_after_last_user], [reminder]
--- a/backend/onyx/configs/app_configs.py
+++ b/backend/onyx/configs/app_configs.py
@@ -1040,6 +1040,13 @@ LOG_ONYX_MODEL_INTERACTIONS = (
 PROMPT_CACHE_CHAT_HISTORY = (
     os.environ.get("PROMPT_CACHE_CHAT_HISTORY", "").lower() == "true"
 )
+
+# Opt-in cap on outgoing image-count for Azure providers (any model_provider
+# starting with "azure"). When enabled, requests are capped at 50 images each
+# (matching the documented Azure OpenAI ceiling) to avoid raw 400s from the
+# gateway. Off by default.
+ENABLE_AZURE_IMAGE_CAP = os.environ.get("ENABLE_AZURE_IMAGE_CAP", "").lower() == "true"
+
 # If set to `true` will enable additional logs about Vespa query performance
 # (time spent on finding the right docs + time spent fetching summaries from disk)
 LOG_VESPA_TIMING_INFORMATION = (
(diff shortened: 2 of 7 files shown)


backend/onyx/chat/llm_loop.py
```diff
@@ -485,16 +485,6 @@ def construct_message_history(
...
 
-    # Attach project images to the last user message
-    if context_files and context_files.image_files:
-        existing_images = last_user_message.image_files or []
-        last_user_message = ChatMessageSimple(
-            message=last_user_message.message,
-            token_count=last_user_message.token_count,
-            message_type=last_user_message.message_type,
-            image_files=existing_images + context_files.image_files,
-        )
```
justin-tahara, line 496: Are we sure we want to be doing this? Does this cause issues with subsequent messages within projects?

backend/onyx/configs/app_configs.py
```diff
@@ -1040,6 +1040,13 @@ def get_current_tz_offset() -> int:
 PROMPT_CACHE_CHAT_HISTORY = (
     os.environ.get("PROMPT_CACHE_CHAT_HISTORY", "").lower() == "true"
 )
+
+# Opt-in cap on outgoing image-count for Azure providers (any model_provider
+# starting with "azure"). When enabled, requests are capped at 50 images each
+# (matching the documented Azure OpenAI ceiling) to avoid raw 400s from the
+# gateway. Off by default.
+ENABLE_AZURE_IMAGE_CAP = os.environ.get("ENABLE_AZURE_IMAGE_CAP", "").lower() == "true"
```
justin-tahara, line 1048: Is the idea to leave this as not on by default because not all azure providers will hit this by default?

backend/onyx/chat/llm_loop.py
Subash-Mohan: We're safe — the block I removed was a double-attachment. The same images are already attached upstream in convert_chat_history (`chat_utils.py:658-660`), so removing this just drops the duplicate. No issue with subsequent messages either: project images live only in ExtractedContextFiles (rebuilt per turn) and never get persisted to ChatMessage.files — `process_message.py:743 `only writes user-attached descriptors. So on later turns, the previous message has no image in its DB row, and the current user message just gets a fresh single attachment. No accumulation.

backend/onyx/configs/app_configs.py
Subash-Mohan: Yeah exactly. The 50 limit is only documented for Azure OpenAI vision models (GPT-4o, GPT-4.1 in the Quotas reference). Other Azure-hosted providers — Foundry partner models (Claude/Llama/Mistral), don't have a documented per-request image limit anywhere I could find. Since the cap matches by `model_provider.startswith("azure")`, turning it on by default would over-apply to providers where the limit isn't proven, and silently start truncating images on deployments that work fine today. Keeping it opt-in so admins flip it on when they actually see the 400 from their specific deployment.

justin-tahara (approved):

PR #11356, opened 2026-05-25
fix(ui): match SearchTool by in_code_tool_id when picking default temperature

ppomes:
Default temperature for a persona is 0 when it has a SearchTool attached, 0.5 otherwise (more creative). The check in `useChatSessionTemperature` compares `tool.name === SEARCH_TOOL_ID`, where `SEARCH_TOOL_ID = "SearchTool"`.

But the default seed for the search tool inserts `tool.name = "internal_search"`. So the check fails, the persona falls back to 0.5, and RAG responses come out more creative than they should be (a custom Rémi-style persona was hallucinating steps because of this).

The rest of the codebase (`AgentEditorPage.tsx` and friends) already compares against `tool.in_code_tool_id`, which is the stable identifier. Aligning these two spots with that.

Workaround we used locally:
`UPDATE tool SET name='SearchTool' WHERE in_code_tool_id='SearchTool'` — which is exactly what this PR makes unnecessary.

---
## Summary by cubic
Match `SearchTool` by `tool.in_code_tool_id` instead of `tool.name` when setting the default temperature, so personas with the tool default to 0 and avoid accidental 0.5 with `internal_search`. Ran `oxfmt` on `web/src/lib/hooks.ts` to wrap long else-if lines in `useLlmManager` and fix quality checks; no behavior change.

<sup>Written for commit cbfdb4f7d37132a6be53ba25bae7663972a7de44. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11356?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ppomes:
👋 Friendly bump — tiny 1-file fix, all review feedback addressed. Looks like it just needs a maintainer to kick off CI + take a look whenever you get a sec 🙏 cc @justin-tahara

acaprau (approved):

acaprau (approved):

PR #11362, opened 2026-05-25
fix(craft): remove scheduled task timezones

Weves:
## Description

Removes the scheduled-task timezone model from the Craft scheduled task feature + makes the form look more similar to existing agent creation form.

Screenshot 2026-05-28 at 1 20 14 PM

Removes the scheduled-task timezone model from the Craft scheduled task feature.

- Removes timezone fields from the scheduled-task API, DB model, frontend types, and editor UI.
- Evaluates scheduled-task cron expressions directly in UTC and removes IANA timezone validation from the schedule helpers.
- Adds an Alembic migration to drop `scheduled_task.timezone` and updates scheduled-task docs/tests.
- Adds unit coverage for the UTC-only schedule helper behavior.

## How Has This Been Tested?

- `python -m compileall backend/onyx/server/features/build/scheduled_tasks backend/onyx/db/scheduled_task.py backend/alembic/versions/d0e1f2a3b4c5_remove_scheduled_task_timezone.py`
- `ruff check backend/onyx/server/features/build/scheduled_tasks backend/onyx/db/scheduled_task.py backend/onyx/db/models.py backend/tests/unit/onyx/server/features/build/scheduled_tasks/test_schedule.py backend/tests/integration/tests/craft/test_scheduled_tasks_api.py backend/tests/external_dependency_unit/craft/test_scheduled_task_executor.py backend/alembic/versions/d0e1f2a3b4c5_remove_scheduled_task_timezone.py`
- `pytest -q backend/tests/unit/onyx/server/features/build/scheduled_tasks/test_schedule.py`
- `alembic heads`

Not fully run locally:

- `pytest -q backend/tests/integration/tests/craft/test_scheduled_tasks_api.py` and `pytest -q backend/tests/external_dependency_unit/craft/test_scheduled_task_executor.py` were blocked by sandbox denial of localhost Postgres access (`Operation not permitted`).
- `npm run types:check` could not run because `web/node_modules` is absent in this checkout; `npx --no-install tsc --noEmit --project tsconfig.types.json` attempted npm registry access and failed under restricted network.
- The commit hook `typescript-check` was skipped for the same `tsgo` resolution issue; all other commit hooks passed.

## Additional Options

---
## Summary by cubic
Remove timezone support from Craft scheduled tasks. Cron is stored and evaluated in UTC; the web app compiles browser‑local schedules to UTC on save and decodes UTC cron back to local for display. Also shows why disabled Save buttons are blocked.

- **Refactors**
  - Dropped `scheduled_task.timezone` via Alembic; removed from ORM, API schemas, and web types.
  - UTC-only helpers: `compute_next_run_at`, `next_n_fires`, and `human_readable` no longer accept a timezone; naive times are treated as UTC and aware times normalize to UTC; `human_readable` no longer appends a timezone.
  - API: create/patch no longer accept `timezone`; list/detail derive human-readable schedules from cron.
  - Frontend: editor reduced to interval (minutes/hours) and daily/weekly; removed Advanced mode, timezone picker, and the “Next runs” preview; added local⇄UTC cron helpers (`compileLocalPayloadToUtcCron`, `decodeUtcCronToLocalPayload`) and `humanReadableScheduleFromCron`; restyled create/edit to match the agent editor with header actions and deferred field errors; Save buttons use a `Disabled` wrapper to surface the blocking reason.
  - Fixed daily/weekly local↔UTC conversion by applying a single, uniform UTC day‑shift when local time crosses midnight to prevent weekday drift across DST; added tests and made the midnight-crossing test timezone‑portable.

- **Migration**
  - Run `alembic upgrade head` to drop the `timezone` column.
  - Note: Tasks previously saved for a local-time schedule now fire at that time in UTC. Adjust cron expressions if you need a specific local-time schedule.

<sup>Written for commit 1d08899634f773619093f8be88e2b8840abdc8be. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11362?utm_source=github">Review in cubic</a>


backend/alembic/versions/d0e1f2a3b4c5_remove_scheduled_task_timezone.py
```diff
@@ -0,0 +1,28 @@
...
+down_revision = "39287906b97a"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.drop_column("scheduled_task", "timezone")
+
+
+def downgrade() -> None:
```
wenxi-onyx, line 23: Probably fine to just pass here right?

wenxi-onyx (approved):
I think the body/content in the Schedule tab should span to fill the entire width of the container. I.e. the small day buttons are quite weird/miss-sized compared to the rest of the form.

Side note: are tabs the right component for this? The two massive tab buttons with relatively sparse body feels weird.

backend/alembic/versions/d0e1f2a3b4c5_remove_scheduled_task_timezone.py
Weves: you should have an up/down of the migration give you the same state (as much as possible). 

For example, if you just passed, the upgrade would fail (since column doesn't exist, so you can't drop it)

PR #11366, opened 2026-05-25
feat(craft): Set policy for external app actions

Danelegend:
## Description
Adds per-action policies for connected external apps: admins can set `ALWAYS` / `ASK` / `DENY` per action (built-in Slack, Google Calendar, Linear catalogs), instead of all-or-nothing access. Ships the policy layer only — storage, the admin API/UI, and the catalog→override merge contract; egress-proxy enforcement is a separate workstream and nothing consumes these policies yet. Supplying `action_policies` full-replaces stored overrides (empty clears, omitted leaves untouched), backed by a new external-dependency-unit suite covering the create/edit/read lifecycle.

Screenshot 2026-05-27 at 5 54 20 PM

## How Has This Been Tested?
Manual + Tests

## Additional Options

---
## Summary by cubic
Adds per-action policies for external apps so admins can set ALWAYS, ASK, or DENY for Slack, Linear, and Google Calendar; ships the policy layer (storage, API/UI, catalogs) only — enforcement via the egress proxy is separate. Also fixes a bad rebase by restoring gate-related files to main’s content.

- **New Features**
  - Database: new `external_app_policy` table and `EndpointPolicy` enum; `ExternalApp.policies` relationship; helpers to read/replace; eager-load on fetch.
  - Providers: declarative `EndpointSpec` with `RestRoute`/`GraphQLOp`; typed `ExternalAppAction` enums and action catalogs for Slack, Linear, and Google Calendar; path-matching utility.
  - Registry: new `providers/registry.py` for provider wiring; adds `validate_action_policies`, `build_action_policies`, and `action_policy_views`; built-in descriptors now include catalog `actions`.
  - API: `UpsertExternalAppRequest.action_policies` (full-replace; `None` leaves stored policies); validates action ids; persists a complete per-catalog set; admin response returns merged `actions` with effective state (unset → `ASK`).
  - UI: Configure modal adds bulk “Auto-approve/Ask” and Advanced per-action toggles (incl. Deny); seeds actions to Ask; preserves policies on enable/disable; TS types for `EndpointPolicy`, descriptors, and views.
  - Docs: action-policy design and egress-proxy enforcement guides.
  - Tests: integration coverage for the admin lifecycle; unit tests for REST path matching.

- **Migration**
  - Run Alembic migration `39287906b97a` to create `external_app_policy`. No manual steps; actions without an override default to `ASK`.

<sup>Written for commit 6ba152bb2d35453b263e2e6c438c34022553db7b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11366?utm_source=github">Review in cubic</a>


Danelegend:
@greptile @cubic

backend/onyx/db/models.py
```diff
@@ -5933,3 +5939,56 @@ class ExternalAppUserCredential(Base):
...
+        nullable=False,
+    )
+    # Hierarchical action id, e.g. "slack.messages.read".
+    action_id: Mapped[str] = mapped_column(Text, nullable=False)
+    policy: Mapped[EndpointPolicy] = mapped_column(
+        Enum(EndpointPolicy, native_enum=False),
+        nullable=False,
+    )
+    # NULL for built-in (display from the code catalog); set for custom apps.
+    name: Mapped[str | None] = mapped_column(Text, nullable=True)
```
wenxi-onyx: If built-in apps are the only reason `name` is nullable, might pref making it non-nullable so the contract here is more clear. I don't think built-in app names will change often either. I think it's okay for desc to be nullable

backend/onyx/external_apps/providers/__init__.py
```diff
@@ -1,13 +1,21 @@
+from onyx.db.enums import EndpointPolicy
```
wenxi-onyx: I think we have a pretty strong pattern of "no __init__.py unless absolutely necessary" i.e. why is there application code in this file? Pref deleting and moving those with this PR. Additionally, make imports are not importing the `providers` module and actually importing directly from the source files

backend/onyx/external_apps/providers/google_calendar.py
```diff
@@ -3,13 +3,82 @@
...
+
+    CALENDARS_READ = "gcal.calendars.read"
+    EVENTS_READ = "gcal.events.read"
+    FREEBUSY_READ = "gcal.freebusy.read"
+    EVENTS_CREATE = "gcal.events.create"
+    EVENTS_UPDATE = "gcal.events.update"
+    EVENTS_DELETE = "gcal.events.delete"
+
+
+_EVENTS_COLLECTION = r"^/calendar/v3/calendars/[^/]+/events/?$"
```
wenxi-onyx: Highly pref using path templates with named placeholders instead of regex. Regex is very hard to read and immediately understand what is being matched and if it's matching correctly. `/calendar/v3/users/{userId}/calendarList` for example should accomplish what you're doing here without an issue. Please also update Slack regex as well.

backend/onyx/external_apps/providers/actions.py
```diff
@@ -0,0 +1,70 @@
...
+
+    # Stable, namespaced id as a provider-specific enum member (e.g.
+    # ``SlackAction.MESSAGES_READ`` → "slack.messages.read"). ``InstanceOf``
+    # accepts any ``ExternalAppAction`` subclass member while rejecting bare
+    # strings, so catalogs can't drift onto untyped ids.
+    id: InstanceOf[ExternalAppAction]
+    normalised_name: str
+    description: str
+    matches: tuple[MatchRule, ...]
+    # Prior ids this action has had, so renames don't break stored admin rows.
```
wenxi-onyx: Why is this needed?

backend/onyx/server/features/build/api/external_apps_api.py
```diff
@@ -52,6 +55,7 @@
 
 def _to_admin_response(app: ExternalApp) -> ExternalAppAdminResponse:
     # Display + lifecycle fields live on the linked Skill row.
+    stored = {policy.action_id: policy.policy for policy in app.policies}
```
wenxi-onyx, line 58: Instead of storing policy overrides, can we just store a policy for every action (including default policies)? This way you don't have to do this "catalog + overrides" merge logic. Whatever is in the db, for all policies, is the source of truth.

backend/tests/external_dependency_unit/craft/test_external_app_action_policies.py
```diff
@@ -0,0 +1,245 @@
+"""Ext-dep tests for the admin action-policy lifecycle on built-in external
```
wenxi-onyx: I believe this set of ext dep tests all rely on a real kubernetes cluster being present. If your tests don't need the cluster, it's much lighter weight/faster if you just create a tests/ext-dep-unit/exteranl_apps/

wenxi-onyx: There might also be a justification that core happy flows should be integration tests (since you have APIs implemented).

backend/tests/external_dependency_unit/craft/test_external_app_action_policies.py
```diff
@@ -0,0 +1,245 @@
...
+from onyx.error_handling.exceptions import OnyxError
+from onyx.external_apps.providers.slack import SlackAction
+from onyx.server.features.build.api.models import ExternalAppAdminResponse
+from onyx.server.features.build.api.models import UpsertExternalAppRequest
+
+_AUTH_TEMPLATE = {"Authorization": "Bearer {token}"}
+
+
+@pytest.fixture(autouse=True)
+def _clean_slack_rows(db_session: Session) -> Generator[None, None, None]:
```
wenxi-onyx: I think this is handled for you if you, per above comment, move this out of the craft/ test lane.

backend/tests/external_dependency_unit/craft/test_external_app_action_policies.py
```diff
@@ -0,0 +1,245 @@
...
+
+def _view(resp: ExternalAppAdminResponse) -> dict[str, EndpointPolicy]:
+    return {action.action_id: action.state for action in resp.actions}
+
+
+def test_create_persists_overrides_and_merges_unset_to_ask(
+    db_session: Session,
+    test_user: User,
+) -> None:
+    resp = _upsert(
```
wenxi-onyx: Pref minimizing raw db calls in tests like this I think. I.e. you should maybe have a test manager that hits APIs (in an integration test) rather than directly upserting into db in a test --> this way, the test follow the actual application code/flow

wenxi-onyx (approved):

web/src/app/craft/v1/apps/admin/ConfigureProviderModal.tsx
```diff
@@ -131,6 +211,70 @@ export default function ConfigureProviderModal({
...
+                  open={advancedOpen}
+                  onOpenChange={setAdvancedOpen}
+                >
+                  <SimpleCollapsible.Header
+                    title="Advanced"
+                    description="Set a policy for each action individually."
+                  />
+                  <SimpleCollapsible.Content>
+                    <div className="flex flex-col gap-2">
+                      {descriptor.actions.map((action) => (
```
wenxi-onyx, line 249: Might be a move to have each App editor be a real page instead of a modal --> as soon as you get ~5+ 10+ endpoints, the modal will be quite cumbersome (long)

Danelegend: Might come round to this a bit later

Danelegend:
@greptile @cubic

PR #11384, opened 2026-05-26
fix(craft): opencode-serve translator — termination, status shape, delta race

rohoswagger:
## Description

Three independent bugs in serve_client.py caused empty/truncated turns:

  - Per-step  no longer terminates the turn; only session.idle / session.status:{type:'idle'} do. Captures the LLM finish reason as messages stream by and surfaces it on the real terminator.
  - session.status payload is an object with a  discriminator, not a string. Fixed the equality check.
  - message.part.delta races ahead of message.updated by 1–300ms; the filter now REST-fetches the message via GET /session/{id}/message/{id} when the messageID is unknown, caching role + part types.

Adds AGENT_TRANSPORT to .vscode/.env.k8s.template, renames the prior deploy postmortem to match doc naming convention, and publishes a new technical postmortem at docs/craft/issues/opencode-serve-event-stream-pitfalls.md.

## How Has This Been Tested?

unit tests, local kind cluster

## Additional Options

---
## Summary by cubic
Fixes the opencode serve translator to stop early terminations and dropped content. Uses real turn-level signals, hydrates raced deltas with negative caching, preserves the model’s finish reason, and injects a `fetch_message` callback to keep translation I/O-free.

- **Bug Fixes**
  - End turn only on `session.idle` or `session.status:{type:"idle"}`; `message.updated` no longer terminates.
  - Capture LLM finish reason from `message.updated` and emit it on the real terminator.
  - Correct `session.status` handling (object with `type`, not a string).
  - If `message.part.delta` arrives before `message.updated`, REST-fetch `GET /session/{id}/message/{id}`; cache role/part types and negative results to prevent retry storms; filter assistant-only content.
  - Tests cover REST hydrate outcomes (assistant/user/failure/no-callback) and assert a single fetch per message.

- **Refactors**
  - Inject `fetch_message` as a parameter to `translate_opencode_event` and thread it through the consumer; `_TurnState` no longer carries I/O.

<sup>Written for commit 32b44037b18d266302cdebd9eb8e0e6c40d4e66c. Summary will update on new commits. <a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11384?utm_source=github">Review in cubic</a></sup>


backend/onyx/server/features/build/sandbox/opencode/serve_client.py
```diff
@@ -261,8 +266,54 @@ def _wrap_raw_output(state: dict[str, Any]) -> dict[str, Any] | None:
...
+    parts = body.get("parts") or []
+    if isinstance(parts, list):
+        for part in parts:
+            if not isinstance(part, dict):
+                continue
+            part_id = part.get("id")
+            part_type = part.get("type")
+            if isinstance(part_id, str) and isinstance(part_type, str):
+                state.part_types[part_id] = part_type
+    return role if isinstance(role, str) else None
```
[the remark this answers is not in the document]
rohoswagger: Fixed in 59391187d4. `_hydrate_message` now adds `msg_id` to `state.user_message_ids` in all three failure branches (no `fetch_message`, empty/failed fetch, and unknown role), so `_is_assistant_message` short-circuits as "non-assistant" on subsequent deltas instead of re-issuing REST calls. Regression coverage in `test_translate_opencode_event.py::test_hydrate_failure_cached_so_subsequent_deltas_skip_fetch` and `::test_hydrate_unknown_role_cached_negatively` asserts a single fetch call across multiple deltas for the same problematic `messageID`.

backend/onyx/server/features/build/sandbox/opencode/serve_client.py
```diff
@@ -112,13 +113,17 @@ class _TurnState:
...
     assistant_message_ids: set[str] = field(default_factory=set)
-    # PartID → part type ("text", "reasoning", "tool", "step-start",
-    # "step-finish", ...). Populated from ``message.part.updated`` events.
-    # Read by the delta handler to decide whether a delta is a text chunk
-    # (AgentMessageChunk) or a reasoning chunk (AgentThoughtChunk) —
-    # opencode emits BOTH on ``field=text`` deltas (the field refers to
-    # the part's text attribute, not whether the content is reasoning).
+    # PartID → "text" | "reasoning" | "tool" | ... (the delta routes on this).
     part_types: dict[str, str] = field(default_factory=dict)
+    # MessageIDs confirmed non-assistant — caches the negative hydrate result.
```
Weves: what is `negative hydrate`?

backend/onyx/server/features/build/sandbox/opencode/serve_client.py
```diff
@@ -112,13 +113,17 @@ class _TurnState:
...
+    user_message_ids: set[str] = field(default_factory=set)
+    # GET /session/{id}/message/{id} — hydrates role + parts when a delta
+    # arrives before message.updated (races by up to ~300ms in practice).
+    fetch_message: Callable[[str], dict[str, Any] | None] | None = None
```
Weves: hmm, why does this need to be part of the obj? Feels a little weird

rohoswagger: yea true, will pass it in as a function param instead

backend/onyx/server/features/build/sandbox/opencode/serve_client.py
rohoswagger: ai slop LOL, we have to hydrate messages from opencode assistant, but negative hydrate is just all user messages

Weves (approved):

PR #11388, opened 2026-05-26
feat(api): cache /auth/type response in browsers

nmgarza5:
## Description

**Bug:** `/auth/type` is hit on every page mount by `useAuthTypeMetadata` in the root provider tree. Prod shows ~65 RPS mean / 176 RPS peak against api-server for a response that essentially never changes.

**Fix:** Set `Cache-Control: public, max-age=60` so browsers cache the response. Guarded with `if has_users:` so the bootstrap (zero users) signup flow stays live.

**Notes:** Response is computed pre-tenant-context, so `public` is safe. In CLOUD mode the response is fully static.

## Additional Options

### DIFF
--- a/backend/onyx/server/manage/get_state.py
+++ b/backend/onyx/server/manage/get_state.py
@@ -42,13 +42,12 @@ async def get_auth_type(response: Response) -> AuthTypeResponse:
         has_users = user_count > 0
 
     # Cache only after bootstrap; the first user flow depends on a live
-    # has_users flag so avoid serving a stale redirect. Set an explicit
-    # no-store in that case so an intermediate CDN with a default TTL
-    # doesn't pin has_users=false past the first signup.
-    if has_users:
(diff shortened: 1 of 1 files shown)


backend/onyx/server/manage/get_state.py
```diff
@@ -40,6 +41,15 @@ async def get_auth_type() -> AuthTypeResponse:
         user_count = await get_user_count()
         has_users = user_count > 0
 
+    # Cache only after bootstrap; the first user flow depends on a live
+    # has_users flag so avoid serving a stale redirect. Set an explicit
+    # no-store in that case so an intermediate CDN with a default TTL
+    # doesn't pin has_users=false past the first signup.
+    if has_users:
+        response.headers["Cache-Control"] = "public, max-age=60"
```
evan-onyx: if an inline conditional fits in one line here I'd prefer that

evan-onyx (approved):

PR #11398, opened 2026-05-26
chore(db): drop unused KG indexes + redundant chunk_stats.id index

justin-tahara:
## Description

Removes `index=True` from 47 columns across the KG models (KGEntity, KGEntityType, KGEntityExtractionStaging, KGRelationship, KGRelationshipType, KGRelationshipTypeExtractionStaging, KGTerm) plus the redundant `index=True` on `ChunkStats.id` (already covered by its primary key). Removes 6 explicit `Index(...)` declarations in `__table_args__` for the same indexes. Adds an idempotent alembic migration `c7bc8cc2921d` that runs `DROP INDEX IF EXISTS` for the 51 corresponding indexes per-tenant.

### Why

KG indexes are unused: `pg_stat_user_indexes.idx_scan = 0` across every KG index on every tenant schema observed over a 3+ day window. The Knowledge Graph feature isn't in active use, so its secondary indexes are pure overhead — schema bloat, planner cost, and per-tenant file-count pressure in multi-tenant deployments.

This PR is one wave of a broader index cleanup. A prior wave (run separately) cleaned up 45 alembic-only "removed feature" indexes. Wave 3.A in this PR removes the unused KG indexes plus the PK-duplicate `ix_chunk_stats_id`.

### Migration semantics

- `upgrade()` iterates 51 names and calls `op.execute("DROP INDEX IF EXISTS <name>;")`. Alembic runs per-tenant via `SET search_path` in `env.py`, so unqualified names resolve to the current tenant's schema.
- Idempotent: `IF EXISTS` makes each DROP safe to re-run.
- `downgrade()` is intentionally a no-op. These indexes were determined unused; rollback should not recreate them. If KG is re-enabled later, recreation should happen via a forward migration tailored to the new workload.

### Trade-offs surfaced in review

A few dropped `*_document_id` / `*_source_document` indexes are still referenced by code paths that run regardless of whether KG has data:
- The `update_kg_entity_name_from_doc` PG trigger fires on every document upsert that touches `semantic_id` and does `UPDATE kg_entity WHERE document_id = NEW.id`.
- `delete_documents_complete__no_commit` (`backend/onyx/db/document.py`) unconditionally clears KG tables on every document deletion.

Cost is effectively zero on empty KG tables (cheap empty-table scan). Cost grows linearly if any tenant ever repopulates KG. If KG is re-enabled, the four `*_document_id` / `*_source_document` indexes can be reintroduced via a follow-up migration.

## How Has This Been Tested?

- `from onyx.db import models` imports cleanly.
- `alembic heads` returns a single head (`c7bc8cc2921d`) — no graph divergence.
- The 51-entry `INDEXES_TO_DROP` list was generated from live `pg_class` (every name corresponds to an index that actually exists in tenant schemas).
- The DB-side DROP was executed ahead of this PR shipping; the migration is the persistent codification (handles future tenant provisioning and brings model state into sync).
- No Python code outside `alembic/versions/` hardcodes any of the dropped index names (grep verified across `backend/` and `web/`).

## Additional Options

### DIFF
--- a/backend/alembic/versions/c7bc8cc2921d_drop_unused_kg_indexes.py
+++ b/backend/alembic/versions/c7bc8cc2921d_drop_unused_kg_indexes.py
@@ -1,0 +1,131 @@
+"""drop_unused_kg_indexes
+
+Revision ID: c7bc8cc2921d
+Revises: b02d7b35e48b
+Create Date: 2026-05-26 13:00:00.000000
+
+Drop a set of unused database indexes from each tenant schema.
+
+Motivation
+----------
+The ``managed_services`` cluster's primary RDS instance hit the EXT4
+htree directory limit on its data volume because the multi-tenant
+deployment (>6,000 tenant schemas) creates the same set of indexes per
+schema, multiplying every redundant index by the tenant count. The
+Knowledge Graph (KG) feature is not in active use on
+``managed_services``: across 6,134 tenants we observed zero index
+scans on the KG indexes (``pg_stat_user_indexes.idx_scan = 0``) over a
+3+ day observation window. We are dropping their non-PK, non-unique
+indexes here.
+
+Also included: ``ix_chunk_stats_id``, a redundant b-tree on
+``chunk_stats.id`` that duplicates the primary key index.
+
+Out of scope: the underlying ``kg_*`` tables and their unique
+constraints (``uq_kg_*``) remain — only secondary indexes are dropped.
+If KG is re-enabled later, the corresponding ``CREATE INDEX``
+statements will need to be reintroduced via a follow-up migration.
+
+Manual cleanup already performed
+--------------------------------
+The DB-side ``DROP INDEX`` was run manually across all existing tenant
+schemas on 2026-05-26, before this migration shipped. This migration's
+job is to:
+
+  (a) catch any new tenants that received these indexes during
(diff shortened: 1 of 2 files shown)


backend/alembic/versions/c7bc8cc2921d_drop_unused_kg_indexes.py
```diff
@@ -0,0 +1,97 @@
...
+    "ix_search_term_term",
+]
+
+
+def upgrade() -> None:
+    for index_name in INDEXES_TO_DROP:
+        op.execute(sa.text(f"DROP INDEX IF EXISTS {index_name};"))
+
+
+def downgrade() -> None:
```
evan-onyx, line 300: this should have a downgrade

evan-onyx (approved):
approving to unblock but pls write the downgrade

backend/alembic/versions/c7bc8cc2921d_drop_unused_kg_indexes.py
justin-tahara: Done in 33fa80f. `downgrade()` now recreates all 51 indexes from a shared `INDEXES` list (single source of truth with `upgrade()`). Both directions idempotent — `IF EXISTS`/`IF NOT EXISTS`.

PR #11406, opened 2026-05-26
fix(smtp): allow unauthenticated and non-TLS relays (#10682)

### LINKED ISSUE #10682: SMTP without authentication is broken — EMAIL_CONFIGURED requires credentials and login() is called unconditionally
### Symptom

  When using an SMTP server that accepts connections from whitelisted IPs without authentication (e.g. an internal relay, Mailhog, Postfix with IP whitelist), email sending is completely non-functional.
  Setting `SMTP_SERVER` and `SMTP_PORT` without credentials causes Onyx to report email as "not configured" and never attempt delivery. Even if the check is bypassed, the SMTP connection fails because
  `login()` is called unconditionally.

  ### Steps to reproduce

  1. Configure an SMTP server that does not require authentication.
  2. Set `SMTP_SERVER` and `SMTP_PORT` in the environment; leave `SMTP_USER` and `SMTP_PASS` empty/unset.
  3. Trigger any action that sends an email (user invitation, password reset, email verification).
  4. Observe that the action silently fails with `"Email is not configured."` — even though a working SMTP server is present.
  5. If `SMTP_USER`/`SMTP_PASS` are set to dummy values to pass the `EMAIL_CONFIGURED` check, the SMTP `login()` call fails with an authentication error and the email is never sent.

  ### Root cause — two separate issues

  **Issue 1** — `app_configs.py`:

```python
  # Current (broken):
  EMAIL_CONFIGURED = all([SMTP_SERVER, SMTP_USER, SMTP_PASS]) or SENDGRID_API_KEY
 ```

  Credentials are required for EMAIL_CONFIGURED to be True. An operator with a credential-free relay has no way to enable email without supplying dummy values.

  Issue 2 — email_utils.py, send_email_with_smtplib():
```python
  with smtplib.SMTP(SMTP_SERVER, SMTP_PORT) as s:
      s.starttls()
      s.login(SMTP_USER, SMTP_PASS)   # ← always called, even if credentials are empty
      s.send_message(msg)
```

  login() is called unconditionally regardless of whether credentials were provided. If SMTP_USER/SMTP_PASS are empty strings, this raises SMTPException. If they are dummy values, it raises
  SMTPAuthenticationError.

  ▎ Note: starttls() is also called unconditionally, which may cause additional failures on servers that do not advertise or support STARTTLS.

  Effect

  - Email sending is completely broken for any deployment using an internal/unauthenticated SMTP relay.
  - The only workaround is to manually comment out s.login(...) (and optionally s.starttls()) inside the running container after every docker compose up --force-recreate.
  - There is no environment variable to configure credential-free or TLS-free operation.

  Recommended fix

  app_configs.py — only require a server address:

  EMAIL_CONFIGURED = bool(SMTP_SERVER) or bool(SENDGRID_API_KEY)

  email_utils.py — guard both starttls() and login(), and introduce an opt-out env var for STARTTLS:

```python
  # app_configs.py
  SMTP_STARTTLS = os.environ.get("SMTP_STARTTLS", "true").lower() not in ("false", "0", "no")
```

```python
  # email_utils.py
  with smtplib.SMTP(SMTP_SERVER, SMTP_PORT) as s:
      if SMTP_STARTTLS:
          s.starttls()
      if SMTP_USER and SMTP_PASS:
          try:
              s.login(SMTP_USER, SMTP_PASS)
          except smtplib.SMTPAuthenticationError:
              pass  # credentials set but server doesn't require auth
      s.send_message(msg)
```

  This preserves the current behavior for all existing deployments (credentials + STARTTLS remain the default) while allowing operators with unauthenticated or plaintext relays to opt out via environment
  variables.

justin-tahara:
## Description

Fixes #10682.

Deployments using an internal / IP-whitelisted SMTP relay (e.g. Postfix without auth, Mailhog) couldn't send email at all:

1. **`EMAIL_CONFIGURED`** in `backend/onyx/configs/app_configs.py:289` required `SMTP_USER` and `SMTP_PASS`, so server-only configurations were reported as "Email is not configured."
2. **`send_email_with_smtplib`** in `backend/onyx/auth/email_utils.py:291-294` called `starttls()` and `login()` unconditionally — so even if users worked around the gate by setting dummy credentials, the SMTP exchange crashed inside `starttls()` (servers that don't advertise it) or `login()` (servers that don't accept auth).

### Fix

- `EMAIL_CONFIGURED` now requires only `SMTP_SERVER` (or `SENDGRID_API_KEY`).
- `starttls()` is gated on a new `SMTP_STARTTLS` env var (default `true`, so Gmail / SES / SendGrid on 587 keep working).
- `login()` runs only when both `SMTP_USER` and `SMTP_PASS` are set.

Includes env.template and Helm chart entries for the new `SMTP_STARTTLS` knob.

All five `EMAIL_CONFIGURED` consumers (`auth/users.py`, `server/manage/users.py`, `server/documents/connector.py`, `ee/onyx/utils/license_notifications.py`, plus this module) treat it purely as a gate to decide whether to call `send_email`; none assume `SMTP_USER`/`SMTP_PASS` are populated, so loosening the predicate is safe.

## How Has This Been Tested?

- New unit test `backend/tests/unit/onyx/auth/test_smtp_unauthenticated.py` mocks `smtplib.SMTP` and covers the four combinations of `SMTP_STARTTLS` × creds: default-config, STARTTLS-off, no-creds, and unauthenticated-relay (both off). All pass locally.
- Pre-commit (ruff, ty) clean.

## Additional Options

---
## Summary by cubic
Allows sending email through unauthenticated or non-STARTTLS SMTP relays by making auth optional and STARTTLS configurable. Keeps secure defaults for hosted providers on port 587 and treats `SMTP_STARTTLS` as fail-safe (only explicit false/0/no disables).

- **Bug Fixes**
  - `EMAIL_CONFIGURED` now requires only `SMTP_SERVER` (or `SENDGRID_API_KEY`).
  - Gated `starttls()` behind `SMTP_STARTTLS`; parsed as not in (`false`, `0`, `no`) so empty values don’t disable STARTTLS.
  - Call `login()` only when both `SMTP_USER` and `SMTP_PASS` are set.
  - Updated Docker Compose and Helm values; template shows `SMTP_STARTTLS=false` explicitly.
  - Added unit tests for all `SMTP_STARTTLS` × credentials combinations.

- **Migration**
  - For internal/non-STARTTLS relays: set `SMTP_STARTTLS=false`.
  - No changes needed for Gmail/SES/SendGrid on 587.

<sup>Written for commit ac2b87e6e177cb9169c1bea58d6944aa849ab83c. Summary will update on new commits. <a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11406?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/backend/onyx/auth/email_utils.py
+++ b/backend/onyx/auth/email_utils.py
@@ -25,6 +25,7 @@ from onyx.configs.app_configs import SENDGRID_API_KEY
 from onyx.configs.app_configs import SMTP_PASS
 from onyx.configs.app_configs import SMTP_PORT
 from onyx.configs.app_configs import SMTP_SERVER
+from onyx.configs.app_configs import SMTP_STARTTLS
 from onyx.configs.app_configs import SMTP_USER
 from onyx.configs.app_configs import WEB_DOMAIN
 from onyx.configs.constants import AuthType
@@ -289,8 +290,10 @@ def send_email_with_smtplib(
         msg.attach(html_part)
 
     with smtplib.SMTP(SMTP_SERVER, SMTP_PORT) as s:
-        s.starttls()
-        s.login(SMTP_USER, SMTP_PASS)
+        if SMTP_STARTTLS:
+            s.starttls()
+        if SMTP_USER and SMTP_PASS:
+            s.login(SMTP_USER, SMTP_PASS)
         s.send_message(msg)
 
 
--- a/backend/onyx/configs/app_configs.py
+++ b/backend/onyx/configs/app_configs.py
@@ -284,9 +284,15 @@ SMTP_PORT = int(os.environ.get("SMTP_PORT") or "587")
 SMTP_USER = os.environ.get("SMTP_USER") or ""
 SMTP_PASS = os.environ.get("SMTP_PASS") or ""
 # Whether to issue STARTTLS on the SMTP connection. Default true preserves
-# behavior for hosted providers (Gmail / SendGrid / SES on port 587). Set to
-# false for internal/whitelisted relays that don't advertise STARTTLS.
-SMTP_STARTTLS = os.environ.get("SMTP_STARTTLS", "true").lower() == "true"
+# behavior for hosted providers (Gmail / SendGrid / SES on port 587). Only an
+# explicit opt-out disables it, so an empty value (e.g. an uncommented
+# `SMTP_STARTTLS=` in a .env file) doesn't silently drop credentials onto a
+# plaintext connection.
+SMTP_STARTTLS = os.environ.get("SMTP_STARTTLS", "true").lower() not in (
+    "false",
+    "0",
+    "no",
+)
 EMAIL_FROM = os.environ.get("EMAIL_FROM") or SMTP_USER
 
 SENDGRID_API_KEY = os.environ.get("SENDGRID_API_KEY") or ""
--- a/backend/onyx/configs/app_configs.py
+++ b/backend/onyx/configs/app_configs.py
@@ -283,10 +283,16 @@ SMTP_SERVER = os.environ.get("SMTP_SERVER") or ""
 SMTP_PORT = int(os.environ.get("SMTP_PORT") or "587")
 SMTP_USER = os.environ.get("SMTP_USER") or ""
 SMTP_PASS = os.environ.get("SMTP_PASS") or ""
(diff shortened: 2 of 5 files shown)


justin-tahara:
Closing in favor of https://github.com/onyx-dot-app/onyx/pull/10931

PR #11417, opened 2026-05-27
fix(clickup): use relative API endpoints

dimanjet:
## Description

Fixes a ClickUp connector URL construction bug where task and comment endpoints were passed with a leading slash even though the request helper already inserts the separator after the API base URL.

That could produce URLs like:

`https://api.clickup.com/api/v2//team/<team_id>/task`

ClickUp can return a 404 for that shape even when the token and team ID are valid. This PR changes those call sites to use relative endpoints, so the connector now builds URLs like:

`https://api.clickup.com/api/v2/team/<team_id>/task`

I also added focused unit coverage for both affected request paths.

## How Has This Been Tested?

- `uv run pytest -q backend/tests/unit/onyx/connectors/clickup/test_clickup_connector.py`

---
## Summary by cubic
Fix ClickUp connector URL construction by using relative endpoints for tasks and comments, preventing double slashes in API URLs that returned 404s. Added unit tests to cover both request paths and guard against regressions.

<sup>Written for commit 9844b43a2e1da3c3dbe39efa80a287d791b13267. Summary will update on new commits. <a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11417?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/backend/onyx/connectors/clickup/connector.py
+++ b/backend/onyx/connectors/clickup/connector.py
@@ -65,7 +65,7 @@ class ClickupConnector(LoadConnector, PollConnector):
         return response.json()
 
     def _get_task_comments(self, task_id: str) -> list[TextSection]:
-        url_endpoint = f"/task/{task_id}/comment"
+        url_endpoint = f"task/{task_id}/comment"
         response = self._make_request(url_endpoint)
         comments = [
             TextSection(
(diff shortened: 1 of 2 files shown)


wenxi-onyx (approved):

wenxi-onyx:
Thanks for the contribution!

PR #11424, opened 2026-05-27
fix(install): guard DOCKER_SUDO[@] expansion for bash 3.2

rohoswagger:
## Description

Fixes a regression in the curl-pipe installer (`curl -fsSL https://onyx.app/install_onyx.sh | bash`) that breaks on macOS before reaching the Docker daemon check:

```
/bin/bash: DOCKER_SUDO[@]: unbound variable
```

**Cause:** macOS ships bash 3.2.57 by default, where `set -u` errors on `"${arr[@]}"` if the array is empty — even though `arr=()` was explicitly declared. `DOCKER_SUDO=()` is the no-sudo default; the empty-array expansion at `install.sh:688` (and 9 other call sites) trips `set -euo pipefail` and aborts before any user-visible error.

**Fix (defense in depth, both layers per review):**

1. Switch every `"${DOCKER_SUDO[@]}"` to the standard `${DOCKER_SUDO[@]+"${DOCKER_SUDO[@]}"}` idiom — only expands if the array is set, which bash 3.2 needs. No behavior change when the array is populated (sudo path); no behavior change on bash 4+.
2. Drop `-u` from `set -euo pipefail`. `set -u` is widely considered more trouble than it's worth (empty-array expansion, parameter-default interactions, version-specific surprises); keeping `-e` and `pipefail`.

Also adds a header comment documenting that this script must stay bash 3.2 compatible — macOS still ships 3.2.57 by default.

Regression came in via c2be1e18a5 (#11326 "fix(install): drop sg dependency, use sudo for docker commands"), which introduced the `DOCKER_SUDO=()` array but didn't account for the bash 3.2 empty-array trap.

## How Has This Been Tested?

- Confirmed `/bin/bash --version` is `3.2.57(1)-release` on macOS.
- Reproduced the original `unbound variable` error on bash 3.2 with `set -euo pipefail` + empty `arr=()` + `"${arr[@]}"`.
- Verified the `${arr[@]+"${arr[@]}"}` idiom expands to nothing on empty arrays and to the full array when populated, on both bash 3.2 and bash 5.x.
- `bash -n install.sh` passes syntax check.
- `bash install.sh --help` and `bash install.sh --dry-run --no-prompt` run end-to-end clean on bash 3.2.

## Additional Options

### DIFF
--- a/deployment/docker_compose/install.sh
+++ b/deployment/docker_compose/install.sh
@@ -1,6 +1,11 @@
 #!/bin/bash
 
-set -euo pipefail
+# This script must remain compatible with bash 3.2 — macOS still ships
+# 3.2.57 by default and the curl-pipe installer is invoked with /bin/bash.
+# Avoid bash 4+ features (associative arrays, ${var,,}, etc.). Notably,
+# do not re-enable `set -u`: on bash 3.2, `"${arr[@]}"` errors as
+# "unbound variable" when arr is an explicitly-declared empty array.
+set -eo pipefail
 
 # Expected resource requirements (overridden below if --lite)
 EXPECTED_DOCKER_RAM_GB=10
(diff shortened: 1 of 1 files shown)


jmelahman:
let's just remove the `set -u`. A lot of bash devs consider it dangerous probably for these very reasons

jmelahman (approved):

jmelahman:
@greptile please review

rohoswagger:
@greptile @cubic-dev-ai please review

jmelahman (approved):

Issue #3616, opened 2025-01-07
Web Connector: Inadequate Mimics in requests.get and Ineffective Playwright Mimics for Cloudflare-Protected Hosts


makoronius:
### Analysis and Attempt to Resolve Website Scraping Errors

Out of curiosity and professional interest, I attempted to scrape the entire company's support website (customer knowledgebase) but encountered the following error:

```
RuntimeError: Failed to fetch 'https://xxxxx/': Error accessing https://xxxxx/: Forbidden (403) for https://xxxx/
```

I can provide the full URL to any developer interested in addressing this issue, as it pertains to my company's public customer support website hosted on ZenDesk behind Cloudflare as I suppose based on headers). However, I prefer not to share it publicly here.

Response headers from curl (obfuscated a bit):
```
< x-zendesk-origin-server: app-server-xxxx-xxxx
< x-zendesk-processed-host-header: xxxx
< X-Zendesk-Zorg: yes
< Set-Cookie: __cf_bm=xxxx; path=/; expires=Mon, 06-Jan-25 23:14:31 GMT; domain=xxxx; HttpOnly; Secure; SameSite=None
< Report-To: {"endpoints":[{"url":"https:\/\/xxxx.cloudflare.com\/report\/v4?s=xxxx"}],"group":"xxxx","max_age":604800}
< NEL: {"success_fraction":0.01,"report_to":"xxxx","max_age":604800}
< Set-Cookie: __cfruid=xxxx; path=/; domain=xxxx; HttpOnly; Secure; SameSite=None
< Set-Cookie: _cfuvid=xxxx; path=/; domain=xxxx; HttpOnly; Secure; SameSite=None
< Server: cloudflare
```

### Initial Observations

Interestingly, a simple `curl` request worked without issue:

```
curl -L -v https://xxxxx/ -A "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36"
```

Even with basic headers, the request succeeded:

```
> User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36
> Accept: */*
```

### Debugging and Modifications

[...]

PR #11433, opened 2026-05-27
docs(craft): Invisible Firefox fetcher (proposal)

feder-cr:
opening as draft to check interest before building the env-gated branch.

would an optional Firefox-based stealth fetcher path in `backend/onyx/utils/playwright_fetch.py` be in scope, parallel to the current Chromium-based `start_playwright()` and `fetch_rendered_html()`? selected via env (`ONYX_WEB_FETCHER=invisible_firefox`), no change to defaults.

motivation: the current module handles bot-detection-aware navigation, but the underlying browser is vanilla Chromium. a growing share of customer doc sites and SaaS targets (ZenDesk article centers, Cloudflare-protected portals) return empty content or 403 even with the existing tuning. related signal: closed-stale #3616 (reporter switched connectors but underlying Web Connector pain remains). the lazy `OnyxWebCrawler` Cloudflare-fallback path in the module docstring is exactly the case where stealth helps.

the backend wraps feder-cr/invisible_playwright (on PyPI as `invisible-playwright`), a Playwright wrapper around a Firefox build patched at the source level for a realistic fingerprint. the wrapper resolves and pins its own patched binary through feder-cr/invisible_core (on PyPI as `invisible-core`), which downloads it from the firefox_antidetect_patch release assets - no separate app, GUI or profile manager involved, just `pip install invisible-playwright` plus a launch call. optional dependency.

this PR adds the proposal doc in `docs/craft/`. tracking discussion: #11432

if not in scope i'll close without noise.

### DIFF
--- a/docs/craft/2026-05-27-invisible-firefox-fetcher-proposal.md
+++ b/docs/craft/2026-05-27-invisible-firefox-fetcher-proposal.md
@@ -1,0 +1,32 @@
+# Invisible Firefox fetcher (proposal)
+
+> Status: Draft proposal
+> Created: 2026-05-27
+> Tracking discussion: TBD
+
+## Goal
+
+Optional Firefox-based stealth fetcher path in `backend/onyx/utils/playwright_fetch.py`, parallel to the current Chromium-based `start_playwright()` and `fetch_rendered_html()`. Selected via env, no change to defaults.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


feder-cr:
@Weves would an optional Firefox-stealth fetcher path help for connectors that hit 403? proposal doc only at this stage, glad to close if out of scope.

feder-cr:
@Weves circling back on this, still an optional fetcher path for connectors hitting 403s. Let me know if it's worth pursuing or if I should close it.

feder-cr:
updating this one: invisible_firefox was a separate GUI app for managing profiles, never part of this path, and it has been retired. the patched binary this PR wants already comes from invisible_playwright, which downloads it through invisible_core, so invisible_firefox was never needed here. the proposal itself is unchanged, I only corrected the line that attributed the binary to the wrong repo.

PR #11435, opened 2026-05-27
fix: clarify S3 filestore region config

### LINKED ISSUE #10971: Cryptic HTTP 400 while trying to use Cloudflare R2 as filestore
Hey! I am trying to setup Onyx on a Kubernetes K3S single node. 

The main problem is:
```
RuntimeError: Failed to check S3 bucket 'onyx-data': An error occurred (400) when calling the HeadBucket operation: Bad Request
```

I have found a similar issue, but the fix is irrelevant here because it's Cloudflare and the port is standard. I verified 3 or 4 times the API creds, but it didn't do anything. I suspected it comes from the app itself because I tried a `EU` bucket with an `s3EndpointURL` pointing to the standard non-`EU` URL, and it delivered me a `403`, expected (the same happens if using an EU string against a non-EU bucket). 
#6548

In `file_store.py`, you default `AWS_REGION_NAME` to "us-east-2". The cryptic error message does not help, especially when nothing in the values tells you that you can set the region. 

I suggest that :
-  `AWS_REGION_NAME`  is added to filestore relative options in `values.yaml` (with something equivalent for the other deployment methods)
- A small text is added in https://docs.onyx.app/deployment/configuration/configuration#file-storage-configuration, to warn about region setup
- The `AWS_REGION_URL` environment variable be mentionned in File Store details there too.

minnhnh:
## Summary
- add AWS_REGION_NAME to Helm filestore config and document external S3-compatible storage
- surface S3_ENDPOINT_URL/AWS_REGION_NAME hints when HeadBucket fails with opaque errors
- document Cloudflare R2 region setup in file store and docker compose templates

Closes #10971

## Testing
- python -m py_compile backend\\onyx\\file_store\\file_store.py backend\\tests\\unit\\file_store\\test_file_store.py
- git diff --check

Not run: pytest/ruff in this local environment because fastapi_users and ruff are not installed.

---
## Summary by cubic
Clarifies S3/S3-compatible region configuration across docs, Helm, and Docker templates, and improves S3 init errors with clear hints. Adds Cloudflare R2 guidance and tests to prevent opaque `HeadBucket` 400/403 failures.

- **New Features**
  - Added `AWS_REGION_NAME` to Helm `values.yaml` and docker-compose templates; documented signing region requirements (R2 uses "auto").
  - Documented external S3-compatible setup with Cloudflare R2 examples; updated file store README and Helm docs with 400/403 troubleshooting.

- **Bug Fixes**
  - `HeadBucket` 400 and 403 errors now include `S3_ENDPOINT_URL`/`AWS_REGION_NAME` hints in logs and exceptions.
  - Added unit tests to ensure error messages surface region/endpoint guidance for 400 and 403 cases.

<sup>Written for commit 6887178a289e7fa4f712ed066d8829a7cd07c91e. Summary will update on new commits. <a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11435?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/backend/onyx/file_store/README.md
+++ b/backend/onyx/file_store/README.md
@@ -1,6 +1,6 @@
 # Onyx File Store
 
-The Onyx file store provides a unified interface for storing files and large binary objects. It supports three storage backends: S3-compatible storage (AWS S3, MinIO, Digital Ocean Spaces, etc.), Google Cloud Storage (GCS), and PostgreSQL Large Objects.
+The Onyx file store provides a unified interface for storing files and large binary objects. It supports three storage backends: S3-compatible storage (AWS S3, MinIO, Cloudflare R2, Digital Ocean Spaces, etc.), Google Cloud Storage (GCS), and PostgreSQL Large Objects.
 
 ## Architecture
 
@@ -26,7 +26,7 @@ The backend is selected via the `FILE_STORE_BACKEND` environment variable:
 
 | Value | Backend | Description |
 |---|---|---|
-| `s3` (default) | S3-compatible | AWS S3, MinIO, Digital Ocean Spaces, etc. |
+| `s3` (default) | S3-compatible | AWS S3, MinIO, Cloudflare R2, Digital Ocean Spaces, etc. |
 | `gcs` | Google Cloud Storage | Native GCS with ADC/Workload Identity support |
 | `postgres` | PostgreSQL Large Objects | No external storage service required |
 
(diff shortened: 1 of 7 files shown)


minnhnh:
Addressed the Greptile 403-path note in 6887178 by sharing the S3_ENDPOINT_URL/AWS_REGION_NAME hint across both 403 and generic HeadBucket failures, with a unit test for the 403 path.

PR #11439, opened 2026-05-27
fix(chat): lazy-load chat file bytes to stop api-server OOM

justin-tahara:
## Description

Some api-server pods OOMKill every few hours (exit 137, sub-15s memory spikes from ~1.1 GiB to >5 GiB). Root cause traced to `backend/onyx/chat/process_message.py:862`:

```python
files = load_all_chat_files(chat_history, db_session)
chat_files_for_tools = _convert_loaded_files_to_chat_files(files)
```

`load_all_chat_files` does an **unbounded parallel S3 fetch + full-bytes-into-memory** for every file ever attached to **any** message in the chat session. In some deployments there are chat sessions with **500+ accumulated files**. With `run_functions_tuples_in_parallel` defaulting to one thread per file, a single new message spawns 500+ threads each pulling a file from S3 and holding the bytes in RAM. Round-robin LB spreads the killer evenly across replicas — all pods restart at similar rates.

**Critical waste**: only the ~5–20 files attached to messages that *survive* chat-history truncation inside `convert_chat_history` are actually used. The other ~95% of loaded bytes are immediately garbage-collected.

There is an existing TODO at the call site acknowledging this is wrong.

### How this regressed

The eager `load_all_chat_files` shape predates current main, but the **memory impact was significantly amplified by PR #8489 "fix: Code interpreter takes files"** (merged 2026-02-16). Before that PR, the loaded files only fed `convert_chat_history` — which uses `.content_text` (cheap, plaintext-cached) and rarely touches raw `.content`. PR #8489 added `_convert_loaded_files_to_chat_files`, which shallow-copies every loaded file's bytes into a new `list[ChatFile]` that is then passed into `PythonTool` for every chat turn, regardless of whether the LLM intends to invoke the code interpreter. The combination — eager parallel byte-load × every-turn × every-history-file × the python_tool staging path — is what produces the sub-15s spikes.

Two compounding factors that surfaced this in production now:
1. Chat sessions doing long-running file-analysis workloads have organically accumulated hundreds of files (largest observed: ~530 unique files in one chat).
2. Round-robin load balancing across api replicas means a single heavy user's message-send spreads the OOM evenly across the entire fleet — no single "hot pod" to blame; all pods restart at similar rates.

### What changed

- `InMemoryChatFile` and `ChatFile`: added `.lazy_from_*` classmethods plus a `__getattribute__` override scoped to `"content"` only. Eager constructor call sites (uploads, generated files) are 100% unchanged.
- `ChatLoadedFile`: added `.lazy_loaded(...)` (separate name from the base method to avoid an LSP-violating override since the subclass requires `content_text` + `token_count`).
- `load_chat_file` now constructs lazy `ChatLoadedFile`s. The cached-plaintext fast path already avoids touching original bytes on the steady state, so `convert_chat_history`'s per-message text injection runs without ever hitting S3.
- `load_all_chat_files` caps `max_workers=16` as defense-in-depth.
- `_load_context_user_files_for_tools` had the same eager `read_file().read()` pattern for project/persona tabular files — fixed alongside.
- `_convert_loaded_files_to_chat_files` (the function introduced by #8489) previously gated on `len(loaded_file.content) > 0`, which would force every lazy file to materialize. Guard removed; empty-content files now pass through (sha256 of \`b""\` and an empty upload are both fine for PythonTool).

### Why this is a "real fix" and not just a band-aid

PythonTool already dedupes via a per-content sha256 cache, so most of its `.content` accesses on a heavy chat are wasted reads. With lazy `ChatFile`, only the files the LLM *actually* asks PythonTool to stage materialize their bytes — typically 0 (LLM never calls python_tool) or 1–5 (single-file analysis). The other 525+ files in a heavy chat never touch S3 on a given turn.

### Out of scope

- nginx sticky sessions: sticky would *concentrate* deaths on hot pods, not stop them — fix is correctness, not LB.
- The full summarization TODO at `process_message.py:860` — larger architectural change.
- The pre-#8489 design decision to fan-load `content_text` for all history files. Lazy loading neutralizes its cost; replacing it with on-demand history summarization is a separate effort.

## How Has This Been Tested?

10 new external-dependency-unit tests in `backend/tests/external_dependency_unit/chat/test_lazy_chat_file_loading.py`:

**Shim contract (no infra needed):**
- \`test_no_load_on_construction\` — \`.lazy_loaded(...)\` doesn't invoke the loader
- \`test_first_access_materializes_once_only\` — first \`.content\` triggers; second is memoized
- \`test_non_content_attrs_do_not_trigger\` — \`file_id\`, \`filename\`, \`file_type\`, \`content_text\`, \`token_count\` access does not load
- \`test_to_file_descriptor_does_not_materialize\` — descriptor extraction is cheap
- \`test_to_base64_materializes_image\` — image serialization correctly forces load
- \`test_chat_file_lazy_content\` — same invariants for \`ChatFile\`

**End-to-end with real Postgres + MinIO:**
- \`test_load_chat_file_does_not_read_bytes\` — wraps \`S3BackedFileStore.read_file\` with a counter, asserts zero raw reads at construction
- \`test_returns_lazy_files_for_history\` — 10-message chat with attached files, asserts zero raw reads after \`load_all_chat_files\` returns
- \`test_max_workers_capped_at_16\` — patches \`run_functions_tuples_in_parallel\`, asserts \`max_workers=16\` even with 200 files passed in
- \`test_wrapping_does_not_materialize\` — \`_convert_loaded_files_to_chat_files\` produces lazy \`ChatFile\`s

All 10 new tests pass; full \`backend/tests/external_dependency_unit/chat\` suite (31 tests) passes with no regressions.

## Additional Options

### DIFF
--- a/backend/onyx/chat/chat_utils.py
+++ b/backend/onyx/chat/chat_utils.py
@@ -402,22 +402,38 @@ def _get_or_extract_plaintext(
 def load_chat_file(
     file_descriptor: FileDescriptor, db_session: Session
 ) -> ChatLoadedFile:
-    file_io = get_default_file_store().read_file(file_descriptor["id"], mode="b")
-    content = file_io.read()
+    """Build a ChatLoadedFile whose raw ``content`` bytes are loaded lazily.
 
-    # Extract text content if it's a text file type (not an image)
-    content_text = None
+    Chat sessions accumulate hundreds of files over time, and a new message
+    sent in such a session previously triggered an unbounded parallel fan-out
+    of full-bytes-into-memory reads, the vast majority of which were
+    immediately discarded by chat-history truncation. We now defer the raw
+    bytes read until something downstream actually accesses ``.content`` —
+    typically only a handful of files survive truncation per turn.
+
+    ``content_text`` (used for LLM context injection) and ``token_count``
+    remain eager because they're cheap: the cached-plaintext store hit avoids
+    reading the original bytes entirely on the common path, and token_count
+    is a single DB lookup.
+    """
+    file_id = file_descriptor["id"]
     # `FileDescriptor` is often JSON-roundtripped (e.g. JSONB / API), so `type`
     # may arrive as a raw string value instead of a `ChatFileType`.
     file_type = ChatFileType(file_descriptor["type"])
+    filename = file_descriptor.get("name")
 
+    # Extract text content if it's a text file type (not an image). The
+    # cached-plaintext path avoids reading the original bytes on the steady
+    # state; only the cache miss branch opens the binary stream.
+    content_text: str | None = None
     if file_type.is_text_file():
-        file_id = file_descriptor["id"]
 
         def _extract() -> str:
+            # Only invoked on cache miss; bytes-read happens here, not upfront.
+            file_io = get_default_file_store().read_file(file_id, mode="b")
             return extract_file_text(
                 file=file_io,
-                file_name=file_descriptor.get("name") or "",
+                file_name=filename or "",
                 break_on_unprocessable=False,
             )
 
@@ -432,7 +448,7 @@ def load_chat_file(
         except Exception as e:
             logger.warning(
                 "Failed to retrieve content for file %s: %s",
-                file_descriptor["id"],
+                file_id,
                 str(e),
             )
 
@@ -448,25 +464,32 @@ def load_chat_file(
             if user_file and user_file.token_count:
                 token_count = user_file.token_count
         except (ValueError, TypeError) as e:
-            logger.warning(
-                "Failed to get token count for file %s: %s", file_descriptor["id"], e
-            )
+            logger.warning("Failed to get token count for file %s: %s", file_id, e)
 
-    return ChatLoadedFile(
-        file_id=file_descriptor["id"],
-        content=content,
+    def _load_content() -> bytes:
+        return get_default_file_store().read_file(file_id, mode="b").read()
+
+    return ChatLoadedFile.lazy_loaded(
+        file_id=file_id,
         file_type=file_type,
-        filename=file_descriptor.get("name"),
+        filename=filename,
         content_text=content_text,
         token_count=token_count,
+        loader=_load_content,
     )
+
+
+_MAX_PARALLEL_CHAT_FILE_LOADS = 16
 
 
 def load_all_chat_files(
     chat_messages: list[ChatMessage],
     db_session: Session,
 ) -> list[ChatLoadedFile]:
-    # TODO There is likely a more efficient/standard way to load the files here.
+    # Returns lazy ChatLoadedFile instances — raw bytes are not read here.
+    # Defense-in-depth: even though per-file work is now cheap (DB lookup +
+    # optional cached-plaintext fetch), cap fan-out so no future regression
+    # can re-introduce a 500-thread storm.
     file_descriptors_for_history: list[FileDescriptor] = []
     for chat_message in chat_messages:
         if chat_message.files:
@@ -478,7 +501,8 @@ def load_all_chat_files(
             [
                 (load_chat_file, (file, db_session))
                 for file in file_descriptors_for_history
-            ]
+            ],
+            max_workers=_MAX_PARALLEL_CHAT_FILE_LOADS,
         ),
     )
     return files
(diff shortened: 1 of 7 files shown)


backend/onyx/chat/chat_utils.py
```diff
@@ -478,7 +501,8 @@ def load_all_chat_files(
             [
                 (load_chat_file, (file, db_session))
                 for file in file_descriptors_for_history
-            ]
+            ],
+            max_workers=_MAX_PARALLEL_CHAT_FILE_LOADS,
```
Danelegend, line 505: This function has been problematic in a few occasions where we spin up many threads. In the event that max_workers is passed in as None, should we have the function assign a max rather than be unbounded?

backend/onyx/tools/models.py
```diff
@@ -197,6 +205,46 @@ class ChatFile(BaseModel):
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
+    @classmethod
+    def lazy_from_filename(
```
Danelegend, line 204: Seems very copied from the above. There's probably some consolidation that can be done. Couple of suggestions:
1. Stop using ChatFile / delete this class and use InMemoryChatFile instead
2. Use some sort of inheritance
3. Have a common method that these classes can call

backend/onyx/file_store/models.py
```diff
@@ -45,6 +54,59 @@ class InMemoryChatFile(BaseModel):
...
+            file_type=file_type,
+            filename=filename,
+        )
+        object.__setattr__(inst, _LAZY_LOADER_ATTR, loader)
+        object.__setattr__(inst, _LAZY_DONE_ATTR, False)
+        object.__setattr__(inst, _LAZY_LOCK_ATTR, threading.Lock())
+        return inst
+
+    def __getattribute__(self, name: str):  # type: ignore[no-untyped-def]
+        if name == "content":
```
Danelegend, line 132: I think a function might be better called get_or_load_content() (instead of directly accessing content). Feels weird to have this as a side-effect that occurs.

backend/onyx/chat/chat_utils.py
Danelegend: In an ideal world, these classes know have a function for loading the file from the file store or whatever. That file is only loaded when it is needed via a context manager or something so we know it will be cleaned up shortly.

So the flow looks something like:
 -> Go to FS / Database to get a list of 'references' to file
 -> Pass these references into the python tool
 -> Python tool selects files it wants to upload -> We pull the files in then and only then
Downside is that this introduces more latency during the tool call so idk

Danelegend (approved):
Few comments, some probs a bit larger. If you agree it should possibly be reworked, lmk and I can take this off your plate

backend/onyx/tools/models.py
justin-tahara: Done in 37dd20e — extracted `install_lazy_content_loader` and `maybe_materialize_lazy_content` helpers into `file_store/models.py`. Both `InMemoryChatFile` and `ChatFile` (and `ChatLoadedFile.lazy_loaded`) now delegate to those. ~40 lines deleted, identical behavior. Tests still pass.

backend/onyx/file_store/models.py
justin-tahara: Fair point — `get_or_load_content()` is more explicit than the `__getattribute__` magic. The trade-off here is that explicit means updating every `.content` caller (PythonTool, `to_base64`, the wrapping closure in `_convert_loaded_files_to_chat_files`, tests) which would expand the cherry-pick surface. I went with the magic-but-drop-in path so the change stays minimal for v3.2.x. Happy to revisit as a follow-up once this lands and we have headroom for the bigger refactor.

backend/onyx/chat/chat_utils.py
justin-tahara: Good call — making `run_functions_tuples_in_parallel` itself have a default cap (rather than each caller passing `max_workers`) is the right direction. Capped our specific call site at 16 for now; the broader util-level fix wants its own PR + grep audit of all callers, so I'd rather not bundle it onto the cherry-pick. Will file as a follow-up.

justin-tahara: Agreed, that's the cleaner long-term shape. The current lazy ChatFile is the surgical interim — same memory result with no signature changes to PythonTool, so it cherry-picks back to v3.2.x cleanly. The reference-passing + load-at-tool-time architecture you describe is the proper follow-up (and yes, latency is the right thing to measure against). Calling it out in the PR description's "Out of scope" section.

PR #11453, opened 2026-05-27
feat(security): runtime-configurable security settings via KV store

jmelahman:
## Summary

- Adds `GET`/`PUT /admin/security` so admins can toggle the 10 existing security knobs (`PASSWORD_*` ×6, `USER_DIRECTORY_ADMIN_ONLY`, `TRACK_EXTERNAL_IDP_EXPIRY`, `VALID_EMAIL_DOMAINS`, `MASK_CREDENTIAL_PREFIX`) at runtime, unblocking the admin UI in #11219. KV blob is per-tenant; env values remain the fallback so an existing deployment with no override blob behaves byte-identically.
- Hot-path reads go through a process-local `TTLCache(ttl=10s)` under an `RLock`. Pre-tenant call sites (`/auth/type`, `UserManager.create`, JWT pre-resolution) short-circuit to env defaults without touching KV or raising. KV outages fall back to env, never bricking auth.
- Writes hold a per-tenant Redis lock (`OnyxRedisLocks.SECURITY_SETTINGS`), preserve unset fields via raw-dict merge, validate cross-field invariants against the merged effective state with explicit `is None` checks, and invalidate the local cache. Operator-locked fields (`password_*`, `valid_email_domains`, `mask_credential_prefix`) are rejected with `INSUFFICIENT_PERMISSIONS` server-side in multi-tenant, with belt-and-braces stripping in `store_overrides`.
- Every consumer reads `get_security_settings()` once at the route boundary and passes scalars into `from_model` / `from_credential_db_model` as required keyword args — Pydantic classmethods no longer do IO. Env constants no longer imported outside `app_configs.py`, the security store, and the documented `oauth_refresher.py` fallback.
- Promotes `cachetools==6.2.2` to a direct backend dep.

## Test plan

- [ ] `python -m dotenv -f .vscode/.env run -- pytest backend/tests/external_dependency_unit/server/security` — store/cache + PUT race tests
- [ ] `python -m dotenv -f .vscode/.env run -- pytest backend/tests/integration/tests/security` — runtime toggle integration test
- [ ] `pytest backend/tests/unit/onyx/auth/test_verify_email_domain.py backend/tests/unit/onyx/auth/test_jwt_provisioning.py backend/tests/unit/onyx/server/manage/test_users_directory_visibility.py` — refactor-driven signature updates
- [ ] Manual: as admin, `PUT /admin/security {"user_directory_admin_only": true}`; confirm non-admin `GET /users` is rejected within ~15s; flip back to `false` and confirm access restored
- [ ] Multi-tenant smoke: `PUT /admin/security {"password_min_length": 12}` returns `INSUFFICIENT_PERMISSIONS`
- [ ] Restart `api_server` and confirm the override survives the restart (KV persistence)

---
## Summary by cubic
Adds runtime‑configurable security settings with per‑tenant overrides and admin endpoints, backed by a typed `security_settings` table. Admins can toggle password policy, directory visibility, IDP expiry tracking, email domain allowlist, and credential masking at runtime; env values remain the fallback so default behavior is unchanged.

- **New Features**
  - Added GET/PUT `/admin/security` backed by a singleton‑per‑tenant `security_settings` table, a process‑local 10s TTL cache, and a cache‑backend lock via `get_cache_backend()`; cold misses are serialized under an RLock. Validation enforces password_min ≤ password_max, caps max at 256, and floors max at 4.
  - Multi‑tenant enforcement: operator‑locked fields (`password_*`, `valid_email_domains`, `mask_credential_prefix`) are rejected at the API and stripped to null in storage; PATCH semantics with explicit null clears; pre‑tenant and DB outages fall back to env.
  - Callers now read once via `get_security_settings()`: password rules in auth flows, `/users` honors `user_directory_admin_only`, OAuth/JWT syncs `track_external_idp_expiry`, `mask_credential_prefix` is applied across snapshots (credentials, connectors, document sets, groups), and `/manage/get_state` returns the current `password_min_length`.

- **Refactors**
  - Replaced the per‑tenant KV JSONB blob with the typed `security_settings` table and centralized writes in `apply_patch(patch, present_keys)` which acquires `OnyxRedisLocks.SECURITY_SETTINGS` and raises `OnyxError` for CONFLICT/INVALID_INPUT; unknown keys are forbidden and error messages hide input.
  - DB upsert derives columns from `SecuritySettingsOverrides.model_fields`; env constants are no longer read at call sites (only in the security store). Tests cover store, locking/caching, validators, and end‑to‑end toggling; `cachetools==6.2.2` remains a direct backend dependency.

<sup>Written for commit 77677db888ca685a2a35c18fc62d9477ee60f69d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11453?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jmelahman:
@greptile @cubic-dev-ai please review

backend/onyx/server/security/store.py
```diff
@@ -0,0 +1,273 @@
...
+    get_current_tenant_id()) to avoid the stack-traced RuntimeError that
+    function raises in multi-tenant when the contextvar is unset. Hot
+    paths like /auth/type call into the loader before tenant resolution.
+    """
+    tid = CURRENT_TENANT_ID_CONTEXTVAR.get()
+    return tid if tid is not None else POSTGRES_DEFAULT_SCHEMA
+
+
+def invalidate_security_cache(tenant_id: str) -> None:
+    with _CACHE_LOCK:
```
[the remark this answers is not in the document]
jmelahman: Acknowledged — this is the cold-miss thundering herd. Documented trade-off in `plans/wild-hopping-gray.md`: cache misses are rare (≤1 per process per ~10s per tenant) and the brief sync KV roundtrip matches what `load_settings()` / `load_enterprise_settings()` already do today from the same async routes. Proper fix is a per-tenant in-flight-load future map (so N concurrent cold-miss readers share one KV read) plus a double-checked write — meaningful complexity for a cold-start-only edge. Worth revisiting if cache-miss latency becomes a problem in practice; we did just fix the related staleness race in `30bdd5959e` with the generation counter.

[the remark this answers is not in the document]
jmelahman: Acknowledged — `get_security_settings()` is sync and the cold-miss path does one blocking KV read inside async handlers. This matches the existing pattern of every other settings consumer in the codebase: `load_settings()`, `load_enterprise_settings()`, etc. are all called the same way from async routes today. Wrapping in `run_in_threadpool` only on miss is doable but pairs naturally with an in-flight-load future map (see sibling thread on the thundering herd) to avoid N threadpool roundtrips on stampede — same documented trade-off in `plans/wild-hopping-gray.md`. Acceptable given the once-per-10s-per-process miss frequency.

backend/onyx/server/security/store.py
```diff
@@ -0,0 +1,299 @@
...
+        "valid_email_domains",
+        "mask_credential_prefix",
+    }
+)
+
+
+# 10s TTL: short enough that admin saves propagate across processes within
+# the worst-case window the brief allows; long enough that hot paths see
+# nearly-pure cache hits in steady state.
+_CACHE_TTL_SECONDS = 10.0
```
evan-onyx, line 30: this is somewhat surprising to me; I would imagine that admins aren't going to be setting these very often? I'd expect a TTL of between 4 and 24 hours and normally cache invalidation is done when the endpoint that writes settings is hit

backend/onyx/server/security/store.py
```diff
@@ -0,0 +1,299 @@
...
+            # salvaged set, fall back to env defaults rather than 500.
+            logger.exception(
+                "Salvaged security overrides still invalid; using env defaults",
+            )
+            return SecuritySettingsOverrides()
+
+
+def store_overrides(overrides: SecuritySettingsOverrides) -> None:
+    """Persist overrides to KV (absent fields stay absent via exclude_none)
+    and invalidate the current tenant's local cache entry. Caller must
```
evan-onyx: would be good to directly enforce this condition (caller must hold the lock) in code)

backend/onyx/server/security/api.py
```diff
@@ -0,0 +1,185 @@
...
+
+    try:
+        overrides = SecuritySettingsOverrides.model_validate(payload_dict)
+    except ValidationError as e:
+        raise OnyxError(OnyxErrorCode.INVALID_INPUT, str(e))
+
+    return overrides, payload_dict
+
+
+def _validate_effective(effective: SecuritySettings) -> None:
```
evan-onyx: this can probably be moved to a pydantic "after" type validator

backend/onyx/server/security/api.py
```diff
@@ -0,0 +1,185 @@
...
+        raise OnyxError(
+            OnyxErrorCode.INVALID_INPUT,
+            "password_min_length must be <= password_max_length",
+        )
+
+    # A password must be able to contain one character from each required
+    # class. If max_length is smaller than the number of required classes,
+    # NO valid password exists and every signup/reset would fail — locking
+    # out the entire tenant with no clear admin signal.
+    required_classes = sum(
```
evan-onyx: Claude cooked with this one but imo we should just make the minimum password_max_length 4 haha

backend/onyx/server/security/models.py
```diff
@@ -0,0 +1,57 @@
...
+    password_require_lowercase: bool | None = None
+    password_require_digit: bool | None = None
+    password_require_special_char: bool | None = None
+
+    @field_validator("valid_email_domains")
+    @classmethod
+    def _normalize_domains(cls, v: list[str] | None) -> list[str] | None:
+        if v is None:
+            return None
+        # Match backend/onyx/configs/app_configs.py:174-185 env-parse behavior
```
evan-onyx: comment should not reference line numbers

backend/onyx/server/security/store.py
```diff
@@ -0,0 +1,299 @@
...
+from shared_configs.contextvars import CURRENT_TENANT_ID_CONTEXTVAR
+
+logger = setup_logger()
+
+
+# Fields that, in multi-tenant deployments, can only be set by the operator
+# via environment variables. Tenant admins cannot override them at runtime.
+OPERATOR_LOCKED_FIELDS: frozenset[str] = frozenset(
+    {
+        "password_min_length",
```
evan-onyx: these strs should be in an enum, I'm also somewhat uncomfortable with this approach on the whole since it requires someone who's adding a new field to the security settings to remember to add to this list. Ideally I'd want it to be like adding to this list requires constructing an object where the creator has to explicitly specify both the name of the field and whether it's locked in multi tenant

backend/onyx/server/security/store.py
```diff
@@ -0,0 +1,299 @@
...
+    locked = OPERATOR_LOCKED_FIELDS if MULTI_TENANT else frozenset()
+
+    def pick(field: str, override_value: Any, env_value: Any) -> Any:
+        if field in locked:
+            return env_value
+        return env_value if override_value is None else override_value
+
+    return SecuritySettings(
+        user_directory_admin_only=pick(
+            "user_directory_admin_only",
```
evan-onyx: these should come from the enum I mentioned earlier rather than raw strings

backend/onyx/server/security/store.py
```diff
@@ -0,0 +1,299 @@
...
+    """
+    tenant_id = CURRENT_TENANT_ID_CONTEXTVAR.get()
+    if tenant_id is None:
+        # MULTI_TENANT=true and contextvar not yet set (pre-tenant). Single-
+        # tenant deployments default the contextvar to POSTGRES_DEFAULT_SCHEMA
+        # at module import, so this branch only fires in multi-tenant.
+        return _build_env_defaults()
+
+    with _CACHE_LOCK:
+        cached = _CACHE.get(tenant_id)
```
evan-onyx, line 197: imo better to 
```
with lock:
    # try getting from cache
    # if failed, do the load, merge, and cache write
```

this should hardly ever error and should take like 100ms at most I think?

evan-onyx (approved):
left some thoughts

PR #11469, opened 2026-05-28
fix(hierarchy): iterative upsert_parents to avoid RecursionError on deep chains

justin-tahara:
## Description

`upsert_parents` in `backend/onyx/db/hierarchy.py` recursed once per ancestor and only added to `done_ids` *after* the recursive call returned. Two failure modes were observed in production:

1. **Deep parent chain.** Notion lets users nest pages arbitrarily; chains >~1000 ancestors blow the Python recursion limit and raise `RecursionError`, killing the whole docfetching batch for the connector.
2. **Cycle / self-parent.** If the source ever returns a circular parent reference, `done_ids` is never populated for the cycle so recursion is unbounded.

A customer on v3.3.x hit (1) on a deeply nested Notion workspace. The code is byte-identical on v3.3.x, v4.0.1, and `main` — only one unrelated commit (`cbed2c21f5`, tier-management) touched `hierarchy.py` in that span — so upgrading without this fix would not have resolved it.

Rewrites `upsert_parents` as an iterative walk seeded with a `visiting` set. Depth is bounded by ancestor count rather than the call stack, and cycles break cleanly with a warning and the existing SOURCE-node fallback that `resolve_parent_hierarchy_node_id` already implements.

Linear: ENG-4121

## How Has This Been Tested?

New external-dependency-unit tests in `backend/tests/external_dependency_unit/hierarchy/test_upsert_parents_deep_chain.py`:

- 1500-deep parent chain (fails today with `RecursionError`, passes after the fix; verified by reproducing the recursion error against the old recursive logic in-memory)
- Two-node parent cycle (A → B → A)
- Self-parent (A → A)

Run with:

```
python -m dotenv -f .vscode/.env run -- pytest backend/tests/external_dependency_unit/hierarchy/test_upsert_parents_deep_chain.py
```

All three pass.

## Additional Options

---
## Summary by cubic
Rewrote parent upsert to be iterative to avoid RecursionError on deep hierarchies and to safely handle cycles/self-parent cases. This stabilizes doc fetching for Notion and other deeply nested sources. Addresses Linear ENG-4121.

- **Bug Fixes**
  - Replaced recursion in upsert_parents with an iterative walk using a visiting set; processes ancestors oldest-to-newest and avoids stack overflows.
  - Detects cycles/self-parent and stops, falling back to the SOURCE node; added regression tests for a 1500-deep chain, a two-node cycle, and a self-parent.

<sup>Written for commit 917b61f941f6db5aa068608ec413d279e92cd7f1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11469?utm_source=github">Review in cubic</a>

### DIFF
--- a/backend/onyx/db/hierarchy.py
+++ b/backend/onyx/db/hierarchy.py
@@ -197,8 +197,7 @@ def upsert_parents(
             break
         if parent_raw_id in visiting:
             # Cycle in the parent chain (e.g. A -> B -> A or A -> A). Stop
-            # walking; remaining ancestors in the cycle will fall back to the
-            # SOURCE node via resolve_parent_hierarchy_node_id.
+            # walking.
             logger.warning(
                 "Cycle detected in hierarchy parent chain for source=%s "
                 "at raw_node_id=%s (raw_parent_id=%s); falling back to SOURCE "
--- a/backend/onyx/db/hierarchy.py
+++ b/backend/onyx/db/hierarchy.py
@@ -174,31 +174,55 @@ def upsert_parents(
     is_connector_public: bool = False,
 ) -> None:
     """
-    Upsert the parents of a hierarchy node.
+    Upsert the in-batch ancestors of ``node`` in oldest-to-newest order.
+
+    Walks ``raw_parent_id`` links iteratively (not recursively) so deeply nested
+    hierarchies — Notion page chains can exceed Python's recursion limit — and
+    cyclic parent references do not raise ``RecursionError``.
     """
-    if (
-        node.node_type == HierarchyNodeType.SOURCE
-        or (node.raw_parent_id not in node_by_id)
-        or (node.raw_parent_id in done_ids)
-    ):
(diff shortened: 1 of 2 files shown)


backend/onyx/db/hierarchy.py
```diff
@@ -174,31 +174,55 @@ def upsert_parents(
...
+        parent_raw_id = current.raw_parent_id
+        if (
+            current.node_type == HierarchyNodeType.SOURCE
+            or parent_raw_id is None
+            or parent_raw_id not in node_by_id
+            or parent_raw_id in done_ids
+        ):
+            break
+        if parent_raw_id in visiting:
+            # Cycle in the parent chain (e.g. A -> B -> A or A -> A). Stop
```
evan-onyx, line 199: comment should end after "stop walking"

evan-onyx (approved):

PR #11503, opened 2026-05-28
fix(craft): bump opencode-serve readiness timeout to 60s for CI cold path

wenxi-onyx:
## Description

Bumps `OPENCODE_SERVE_READY_TIMEOUT_SECONDS` from 30s to 60s and adds periodic progress logging during the wait.

### Why

The Craft K8s test suite intermittently fails with `opencode-serve never became ready in sandbox pod` even though the pod itself reaches K8s-Ready. The 30s constant was calibrated for local dev (M-series Mac, warm pod) where opencode binds `:4096` in <1s. CI runs on a shared 4-vCPU GitHub runner with workflow-configured `SANDBOX_POD_CPU_REQUEST: 100m` to fit 4 concurrent test pods — under that throttling, bun cold start + first-ever sqlite migration can spill past 30s for a fresh pod. Bumping to 60s covers the CI cold path while still failing fast on a real crash.

### Why not exponential backoff

Backoff is the right pattern when probe cost matters (rate-limited APIs, fragile services, distributed coordination). For "wait for a local process to bind a port," the probe is a 1-second-timeout HTTP `GET /doc` against an in-cluster IP — failure is an instant TCP-RST. Polling every 0.5s for 60s = 120 probes, cost negligible. Backoff would actively hurt first-bind detection latency for the common-case warm path.

### Other change

Added a periodic progress log every 10s during the wait. 60s of silence in CI logs makes "wedged" indistinguishable from "still loading"; the progress line shows last_err so we know what the probe is seeing.

## How Has This Been Tested?

- Syntax-checked the edit locally.
- The original 30s value triggered the failure in #11403 CI run https://github.com/onyx-dot-app/onyx/actions/runs/26602203127 ; this PR re-runs `craft-k8s-tests` against the bumped value.

## Additional Options

---
## Summary by cubic
Increases the opencode-serve readiness timeout from 30s to 60s to reduce flaky CI failures on cold pods, and adds periodic progress logs so waits are visible in CI.

- **Bug Fixes**
  - Set `OPENCODE_SERVE_READY_TIMEOUT_SECONDS` to 60; polling stays at 0.5s.
  - Log a progress line every 10s with elapsed time and last error; log time-to-ready on success.

<sup>Written for commit f602ab98be992c882dbf1609810d0d4bd2bb07e3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11503?utm_source=github">Review in cubic</a>

### DIFF
--- a/backend/onyx/server/features/build/sandbox/serve_transport.py
+++ b/backend/onyx/server/features/build/sandbox/serve_transport.py
@@ -43,9 +43,10 @@ SandboxEvent = Any
 # Tags serve-transport logs with the api_server replica handling the prompt.
 _API_SERVER_HOSTNAME = os.environ.get("HOSTNAME", "unknown")
 
-# opencode-serve boot lags backend Ready by ~1–3s warm, up to ~15s cold.
-OPENCODE_SERVE_READY_TIMEOUT_SECONDS = 30
+# 60s covers cold-start in CI under shared-CPU pressure; 30s was too tight.
+OPENCODE_SERVE_READY_TIMEOUT_SECONDS = 60
 OPENCODE_SERVE_READY_POLL_INTERVAL_SECONDS = 0.5
+_OPENCODE_SERVE_READY_PROGRESS_LOG_INTERVAL_SECONDS = 10.0
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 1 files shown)


wenxi-onyx:
Closing — separate `craft-ci` branch already moves sandbox tests off the deprecated `gpt-4o-mini`, which is the actual root cause of the CI failure (not the timeout). Bumping to 60s only moved the symptom from 'never bound' to 'binding but unhealthy', which fits LLM-validation-on-startup failing for a deprecated model.

PR #11508, opened 2026-05-28
feat(craft): Docker compose proxy path phase 1

acaprau:
## Description

Second of N in the docker-compose proxy stack. Wires the dispatch layer that #11507 landed into the proxy boot path, plus a small opencode-config change so the docker sandbox can register the session-tag plugin.

**`server.py` boot wire-up.** Replaces direct `K8sSecretCAStore()` / `K8sInformerLookup()` calls with `build_ca_store()` / `build_ip_lookup()` from `sandbox_proxy.backend`. Extracts five helpers (`_bootstrap_ca`, `_build_lookup`, `_build_cache_factory`, `_build_mitm_options`, `_run_master`) so the boot path reads top-down and the helpers are unit-testable in isolation. Net effect: `SANDBOX_BACKEND=docker` now actually selects the docker stores at startup. `SANDBOX_BACKEND=kubernetes` behavior is byte-identical (same calls, same stores, just one layer of indirection).

**`build_opencode_config` learns `plugins=`.** The multi-provider builder already accepts a `plugins` list; the single-provider wrapper (which the docker backend uses) didn't forward it. Without this fix, docker sandboxes have no way to register the session-tagging plugin and the proxy can't route approval cards back to the originating session.

Zero runtime impact on K8s. Two new unit tests pin the plugin-forwarding behavior.

## How Has This Been Tested?

- [x] `pytest backend/tests/unit/onyx/server/features/build/sandbox/test_opencode_config.py` -- all green; +2 new tests for plugin forwarding (with-plugins and default-omitted).
- [x] `pre-commit run --all-files` on the touched files -- clean.
- [x] Smoke-tested end-to-end as part of the full stack (proxy container on the sandbox bridge actually using `build_ca_store()` / `build_ip_lookup()` to instantiate docker stores; sandbox identity resolution through to a real 200 against example.com). Details in
`docs/craft/features/approvals/phase-5-docker.md` Testing section.

## Additional Options

- [ ] [Optional] Please cherry-pick this PR to the latest release version.
- [x] [Optional] Override Linear Check

---
## Summary by cubic
Make the sandbox proxy work with Docker Compose by decoupling backend selection from Kubernetes and wiring the dispatch layer into startup. Restores session routing via plugin forwarding; phase 1 of the Docker proxy path.

- **Refactors**
[...]

PR #11510, opened 2026-05-28
feat(craft): Docker compose proxy path phase 2

acaprau:
## Description

Follows #11508. Tightens the sandbox security posture so the running agent process can never acquire `NET_ADMIN`, even if a setuid-NET_ADMIN binary somehow ended up in its filesystem.

The docker-compose sandbox needs `cap_add=[NET_ADMIN]` at startup so `firewall-init.sh` can install iptables rules. Before this PR, the script's tail used `gosu` to setuid to UID 1000 -- but `gosu` `setuid()`s directly without an execve, so the bounding set kept `NET_ADMIN`. Not an active grant, but a weaker posture than the K8s init-container model where `NET_ADMIN` is only on the init container's security context.

This PR switches the tail to `capsh --drop=all --user=sandbox`, which clears the bounding set (via `PR_CAPBSET_DROP`) before setuid + execve. After this the agent process has zero caps in any set and an empty bounding set, matching the K8s posture.

**Changes:**
- `Dockerfile`: add `libcap2-bin` (provides `capsh`).
- `firewall-init.sh`: swap `exec gosu 1000:1000 "$@"` for `exec capsh --drop=all --user=sandbox -- "$@"`. Updated the precondition check + comment.
- `docker_sandbox_manager` (Phase D) already grants `SETPCAP` alongside `NET_ADMIN` so the `PR_CAPBSET_DROP` syscall is authorised. Both get dropped by capsh before exec.

The K8s init path is unaffected; `initcontainer` mode exits before the capability-bound section.

## How Has This Been Tested?

- [x] Smoke-tested locally as part of the full stack: started a docker sandbox under the proxy posture; confirmed the firewall-init log line `entrypoint mode: clearing bounding set, dropping to UID 1000`; verified inside the running container that the bounding set is empty (`capsh --print | grep "Bounding 
set"` shows no entries) and the effective/permitted/inheritable sets are also empty.
- [x] `pre-commit run` on touched files — clean (shellcheck passes `firewall-init.sh`).
- [x] K8s sandbox lane unaffected: `initcontainer` mode early-returns before the capsh tail; no change to existing K8s init-container behaviour.

## Additional Options

---
## Summary by cubic
Hardens the Docker Compose proxy bootstrap by switching entrypoint privilege drop from `gosu` to `capsh` and granting `SETPCAP` with `NET_ADMIN` so we clear the capability bounding set before dropping to UID 1000. The agent now runs with zero capabilities while still allowing iptables during init.

- **Refactors**
  - Entrypoint mode now uses `capsh --drop=all --user=sandbox` (replaces `gosu`) to clear the bounding set, drop privileges, then exec; entrypoint mode now requires `capsh`.
  - Compose-only change; the Kubernetes init path exits before the main process and never runs `capsh`.
  - Docs updated: `cap_add=["NET_ADMIN","SETPCAP"]` when proxy is enabled and the new `capsh` command; clarifies zero caps after exec.

- **Dependencies**
  - Add `libcap2-bin` to provide `capsh` in the sandbox image (keeps `gosu` installed but unused).

<sup>Written for commit 035a778405db0cbefc7ea02291969094a3859b13. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11510?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


wenxi-onyx (approved):

acaprau:
@greptile @cubic

PR #11514, opened 2026-05-28
feat(craft): Encrypt external app credentials

Danelegend:
## Description
This encrypts external app credentials and prevents organisation wide credentials from being visible on the frontend. User level credentials should still be visible to the user

## How Has This Been Tested?
Manual + Tests

## Additional Options

---
## Summary by cubic
Encrypt external app credentials at rest and mask org-wide secrets in the admin UI and API. Secrets are preserved on edits via masked placeholders, and all runtime reads decrypt safely.

- New Features
  - Store `organization_credentials` and `user_credentials` encrypted via `EncryptedJson` wrapped by `SensitiveValue`.
  - Admin responses mask org secrets; edits can resubmit masks without overwriting real values; masked values on create or without stored values are rejected.
  - All reads use `.get_value(apply_mask=False)` in header injection, OAuth, and auth checks.
  - Admin API expects `organization_credentials` as `dict[str, str]`.

- Migration
  - Move both credential columns from JSONB to encrypted binary. Delete per-user rows; reset org credentials to empty. Users reconnect; admins re-enter secrets.
  - Downgrade recreates empty JSONB columns (values are not restored).

<sup>Written for commit e41b193e1f5e1d7bedcd8a71424006620369cf75. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11514?utm_source=github">Review in cubic</a>


Danelegend:
@greptile @cubic

backend/alembic/versions/b4950827c0dd_encrypt_external_app_credentials.py
```diff
@@ -0,0 +1,80 @@
...
+        "external_app_user_credential",
+        sa.Column("user_credentials", sa.LargeBinary(), nullable=False),
+    )
+
+    # Keep app config rows (name, url patterns, policies) but reset their org
+    # credentials; admins re-enter client_id/secret. b"{}" is the empty JSON
+    # object — no secret to protect — and is replaced by an encrypted value on
+    # the next save.
+    op.drop_column("external_app", "organization_credentials")
+    op.add_column(
```
wenxi-onyx, line 47: This dance here is a little bit awkward, but I guess it's because we need to clear external_app_user_credential but not external_app?

My claude gave me a slightly cleaner version
```
op.add_column(
      "external_app",
      sa.Column(
          "organization_credentials",
          sa.LargeBinary(),
          nullable=False,
          server_default=sa.text(r"'\x7b7d'::bytea"),  # b"{}" — backfills existing
  rows
      ),
  )
  op.alter_column("external_app", "organization_credentials", server_default=None)
```

backend/onyx/db/external_app.py
```diff
@@ -74,6 +76,29 @@ def validate_auth_template(
             )
 
 
+def resolve_masked_credentials(
+    incoming: dict[str, Any],
```
wenxi-onyx: Can these types get any stricter?

web/src/refresh-components/inputs/PasswordInputTypeIn.tsx
```diff
@@ -140,6 +140,12 @@ export interface PasswordInputTypeInProps extends Omit<
    * The input remains editable so users can type a new value.
    */
   isNonRevealable?: boolean;
+  /**
+   * When true, the reveal toggle is never rendered at all. Use this for secrets
+   * that must never be shown in plaintext in the UI (e.g. an external app's
+   * client secret). The value stays masked and remains editable.
+   */
+  hideRevealToggle?: boolean;
```
wenxi-onyx: What is the difference between `isNonRevealable` and `hideRevealToggle` --> if the value is masked, I think it's fine to leave the toggle? tl;dr is it possible to not edit the refresh component for this? I don't think it's a super special case we're dealing with here.

wenxi-onyx (approved):

PR #11547, opened 2026-05-29
perf(build): replace pypandoc-binary with a mistune + python-docx converter

jmelahman:
## Summary

Replace `pypandoc-binary` with an in-process `mistune` + `python-docx` Markdown → DOCX converter for the build-session **"export as DOCX"** feature. `pypandoc-binary` bundled the full **~150 MB pandoc binary** into the production backend image purely for this one conversion, making it the cheapest large image-size win available.

The conversion now lives in `backend/onyx/server/features/build/session/md_to_docx.py` (`markdown_to_docx_bytes`), called from `manager.py`. It parses Markdown to an AST with `mistune` and writes the `.docx` with `python-docx` — no subprocess, no temp files, no external binary.

## Why this is safe to switch to

The export feature was pypandoc's only user, so the bar was "don't make the DOCX look worse." I tuned the converter against the previous pandoc output until it matched, using a structural diff harness (see Tooling). On a recovered real report (`foundation_report.md`: 11 tables, 88 footnotes, headings/lists/blockquotes/links), **full paragraph text+style alignment with the pandoc export reached 100%**.

What was matched to pandoc (each with a regression test):
- **Prose styles** — `Body Text`/`First Paragraph` (first prose after any non-paragraph block, except after a table → `Body Text`), `Compact` tight lists, `Block Text` blockquotes, `Image Caption`.
- **Document defaults** — Aptos 12 pt body / Aptos Display headings, non-bold colored headings, single line spacing, 1" margins. (Clean-room reproductions of pandoc's `reference.docx` values — no GPL asset is bundled.)
- **Lists** — tight vs loose handling, preserved ordered-list start values, **numbering restarts per list**, loose-list continuation paragraphs, 0.5"/level indent with a 0.25" hanging marker.
- **Tables** — per-column alignment from the separator row, borderless with a header-row rule + pandoc cell padding, auto-fit columns.
- **Hyperlinks** — `Hyperlink` char style (muted blue, no underline), inline formatting preserved **inside** the label, whitespace + balanced-parentheses preserved (e.g. Wikipedia `..._(novel)` URLs), in both target and visible text.
- **Footnotes** — `[^n]` citations become **real Word footnotes** (a `word/footnotes.xml` part built by hand, since python-docx has no footnote API): superscript references Word links to page-bottom notes, one footnote per reference (matching pandoc). This alone took corpus alignment from ~34% → 100%.
- **Inline** — HTML entities decoded; `<br>` → line break; other raw HTML dropped (not shown as literal markup).

## Performance

In-process conversion vs the old `pypandoc` path (subprocess spawn + temp-file round-trip), same documents:

| Document | This PR | pandoc subprocess |
|---|---|---|
| small | ~36 ms | ~69 ms |
| moderate (list + table) | ~61 ms | ~85 ms |
| real report (29 KB, 11 tables, 88 footnotes) | ~183 ms | ~191 ms |
| pathological (~290 KB) | ~1390 ms | ~1170 ms |

We're **faster for the small-to-moderate documents this feature actually produces** (pandoc pays a ~50–65 ms fixed spawn/RTS/temp-file cost on every call), roughly **tie around ~30 KB**, and pandoc's optimized core **wins on very large documents**. So this is a size/robustness win that is also a latency win for realistic inputs — not a universal speedup. A by-id style-application fix (avoiding python-docx's per-paragraph by-name style scan) made the converter ~2.1–2.3× faster than its first cut on table/list-heavy docs.

## Robustness

Fuzzed with malformed LLM output; the converter always returns a valid `.docx` rather than raising. Hardening: strip XML-invalid control characters and unpaired UTF-16 surrogates before parsing (python-docx otherwise raises); graceful handling of unclosed fences/links, ragged tables, orphan footnotes, raw HTML, and deeply nested structures.

## Dependencies

- **Removed** `pypandoc_binary` from the backend group (drops the ~150 MB binary); regenerated `uv.lock` + `backend/requirements/default.txt`.
- **(Re)added** `python-docx` to the backend group — it had been pruned as transitive-only (#11533) but is now a direct production import. `mistune` was already a direct backend dep.

## Tooling

`backend/scripts/build_docx_preview.py` — a dev-only harness to `generate` a `.docx` from Markdown (via this converter, and pandoc when `pypandoc` is installed transiently) and structurally `compare` two `.docx` files (style histogram, per-paragraph alignment, similarity score). `pypandoc` is **not** a project dependency.

## Testing

- `backend/tests/unit/onyx/server/features/build/session/test_md_to_docx.py` — 43 unit tests covering every construct above plus the pandoc-matching rules, footnote structure, XML-escaping/control-char/surrogate handling, malformed-input robustness, and the link-paren/label-formatting edge cases.
- `ty`, `ruff`, and `uv lock --check` pass; the converter resolves and runs end-to-end from the backend dependency group alone.

## Known non-goals / minor gaps

- Images are rendered as alt-text captions, not embedded (pandoc didn't fetch remote images either).
- Fenced code blocks use a monospace `Normal` paragraph rather than pandoc's `Source Code` style; a fenced code block nested *inside a footnote* is dropped. Both are rare and can be follow-ups.

---
## Summary by cubic
Replaced `pypandoc-binary` with a `mistune` + `python-docx` Markdown→DOCX converter for build exports, cutting ~150 MB from the backend image and running ~2x faster. Output closely matches Pandoc (styles, fonts, margins, spacing, lists, tables, hyperlinks, images, real footnotes).

- **Refactors**
  - Added `markdown_to_docx_bytes` in `backend/onyx/server/features/build/session/md_to_docx.py` and wired `manager.py` to use it; added `backend/scripts/build_docx_preview.py` plus a realistic fixture and targeted tests.
  - Styles/layout: Pandoc-like prose styles (Body Text, First Paragraph, Compact, Block Text, Image Caption), non-bold colored headings, Aptos 12pt body + Aptos Display headings, 1" margins, single spacing.
  - Lists: tight/loose handling, preserve start values and restart numbering per list, continuation paragraphs, 0.5" per-level indent with 0.25" hanging marker.
  - Tables/media/links: per-column alignment; borderless tables with a header-row rule and padding; auto-fit columns; images as alt-text captions; links use a `Hyperlink` char style (muted blue, no underline), keep inline formatting and whitespace, and preserve balanced parentheses.
  - Footnotes/perf/robustness: real Word footnotes (one per reference); decode HTML entities; drop raw HTML; strip XML-invalid control chars and invalid surrogates; set styles by id and use `str.translate`; ~2.1–2.3x faster on list/table-heavy docs.

- **Dependencies**
  - Removed `pypandoc-binary`; added `python-docx`. Updated `uv.lock`, `backend/requirements/default.txt`, and `pyproject.toml`.

<sup>Written for commit 7c2fb7ec6e655f7b6ad262572113a385655d181d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11547?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jmelahman:
Addressed in 797d8549ff: `_add_hyperlink` now carries the inherited `fmt` flags (bold/italic/strikethrough/code) into the link's `w:rPr` in CT_RPr schema order, and sets `xml:space="preserve"` on the `w:t` when the text has leading/trailing whitespace. Added unit tests `test_hyperlink_inherits_surrounding_formatting` and `test_hyperlink_preserves_whitespace_in_text`.

jmelahman:
@greptile please review

wenxi-onyx (approved):

jmelahman:
@greptile review please :pray:

PR #11555, opened 2026-05-29
perf(sandbox): tighten pod-ready feedback loop

wenxi-onyx:
## Description

Speed up the Craft K8s test lane by cutting the per-pod ``Ready`` latency floor. Touches production K8s manager code — separate from any test-only changes.

- **Tune the sidecar readiness probe.** Was ``initialDelaySeconds=3, periodSeconds=10``, so a sidecar that bound :8731 at t=2.5s still got marked Ready around t=13. Drop to ``initialDelaySeconds=1, periodSeconds=2`` and raise ``failureThreshold`` to 10 so a slow-binding sidecar still gets ~21s of grace.

- **Replace the pod-ready poll loop with a kubernetes ``watch``.** The apiserver streams status transitions as they happen, so Ready is observed within ~100ms instead of waiting for the next poll tick (which had crept up to 2s by the back half of the schedule). Bounded retry on ``410 Gone`` preserves correctness when resource_versions age out.

- **Halve the opencode-serve ready-poll interval** (0.5s → 0.25s). Same wait runs once per provision; the win compounds with the probe tuning.

Combined: pod-ready floor drops from ~12–18s to ~3–5s per pod. Across the ~44 pod provisions in the Craft K8s lane, expect ~3–5 min off the test phase.

## How Has This Been Tested?

Pre-commit (ruff, ty) clean. K8s tests aren't runnable locally; the ``pr-craft-k8s-tests.yml`` lane will exercise the new probe timing and watch path under load (~44 pod provisions).

## Additional Options

### DIFF
--- a/backend/onyx/server/features/build/sandbox/kubernetes/kubernetes_sandbox_manager.py
+++ b/backend/onyx/server/features/build/sandbox/kubernetes/kubernetes_sandbox_manager.py
@@ -57,6 +57,7 @@ from cryptography.hazmat.primitives.asymmetric.ed25519 import Ed25519PrivateKey
 from cryptography.hazmat.primitives.serialization import Encoding
 from cryptography.hazmat.primitives.serialization import PublicFormat
 from kubernetes import client
+from kubernetes import watch
 from kubernetes.client.rest import ApiException
 from kubernetes.stream import stream as k8s_stream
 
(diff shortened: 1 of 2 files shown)


wenxi-onyx:
@greptileai re-review

Weves (approved):

PR #11566, opened 2026-05-30
perf(craft): async webapp proxy with connection pooling and per-asset DB caching

rohoswagger:
## Description

Speeds up the Craft webapp preview proxy, which forwards browser requests (HTML + JS/CSS/font/image fan-out) from the user's iframe to the Next.js dev server inside the sandbox pod.

**Changes:**

- **Async endpoint** (`get_webapp`, `_proxy_request`, helpers): converted from sync `def` (ran in the 40-thread shared pool) to `async def` with `httpx.AsyncClient`. Slow sandbox upstreams now suspend a coroutine rather than parking a threadpool worker, so preview load can no longer starve unrelated API endpoints. Binary assets stream via `aiter_bytes` with a `finally`-guarded generator so client disconnects can't leak pooled connections.

- **Connection pooling**: replaced the per-request `httpx.Client(...)` with a module-level `httpx.AsyncClient` (keepalive, 50/200 limits). Eliminates TCP setup/teardown on every proxied asset.

- **Per-asset DB caching**: `_SANDBOX_URL_CACHE` (session_id → upstream URL, 60s TTL) and `_WEBAPP_ACCESS_CACHE` ((session_id, viewer_id) → grant, 30s TTL) — both bounded `TTLCache`s. On the warm path (every asset after the first), zero DB connections are opened. Two new async DB helpers (`get_webapp_target_async`, `get_webapp_access_async`) collapse what was 3 sync round-trips per asset into a single async join on cache miss.

- **Immutable cache headers**: `/_next/static/` assets get `Cache-Control: public, max-age=31536000, immutable`, overriding the dev server's `no-store` so the browser skips re-proxying on reload.

- **assetPrefix on the sandbox dev server**: `next.config.ts` reads `WEBAPP_ASSET_PREFIX` (exported by the dev-server start scripts at launch), so Next.js emits already-proxied URLs. The proxy then only rewrites HTML (small, single doc); JS/CSS bundles pass through as-is.

## How Has This Been Tested?

- **Unit tests** (`tests/unit/build/test_rewrite_asset_paths.py`): 34 pass. Added async wiring tests (HTML rewrite, HMR injection, header stripping, immutable cache override, JS pass-through, upstream connection closed on full drain and on early client disconnect, auth grant cached / denial not cached).
- **Live benchmark** against the running local dev environment (telepresence → cluster sandbox):
  - Sync baseline (main): **p50 = 10,214 ms** per page-load (HTML + 20 assets)
  - Async (this PR): **p50 = 183 ms** — ~56× faster on loopback (amplified by telepresence round-trips; production gains are smaller in absolute terms but the same overheads are removed)
  - Static `Cache-Control` confirmed `public, max-age=31536000, immutable` on live assets

## Additional Options

---
## Summary by cubic
Makes the Craft webapp preview proxy async with pooled connections and shared Redis caching to speed up previews and prevent API thread starvation. Adds immutable caching for content‑hashed media and asset prefixing to cut rewrites and round trips; local p50 page load dropped from ~10s to ~183ms.

- **Refactors**
  - Converted `get_webapp` and proxy helpers to `async` with `httpx.AsyncClient` streaming and a lazily initialized pooled client (keepalive; 50/200 limits).
  - Replaced per-process caches with a shared Redis backend via `get_cache_backend()` for sandbox URL (session → URL, 60s) and access grants ((session, viewer) → grant, 30s; grants only). New async DB helpers (`get_webapp_target_async`, `get_webapp_access_async`) collapse lookups; evict cached sandbox URL on 502/503/504.
  - Always rewrite HTML/CSS/JS bodies (idempotent with asset prefix). Stream binary assets and ensure upstream connections are closed on drain or early client disconnect using a single guarded finally; `_aiter_and_close` is typed as `AsyncGenerator[bytes, None]`.
  - Hoisted repeated `str(session_id)`, relative path, and cache-key computation in the proxy (no behavior change).

- **New Features**
  - Set `Cache-Control: public, max-age=31536000, immutable` for `/_next/static/media/` only; preserve the dev server’s no-cache for chunks/CSS; strip `pragma`/`expires` on media responses.
  - `next.config.ts` reads `WEBAPP_ASSET_PREFIX` (set by dev‑server start scripts) so JS/CSS URLs are pre‑prefixed; proxy rewrites remain safe for old and new sandboxes.

<sup>Written for commit f560a5d7c467e88a6b8ecd0b52e10acac1b3f904. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11566?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/server/features/build/api/api.py
+++ b/backend/onyx/server/features/build/api/api.py
@@ -4,7 +4,6 @@ from pathlib import Path
 from uuid import UUID
 
 import httpx
-from cachetools import TTLCache
 from fastapi import APIRouter
 from fastapi import Depends
 from fastapi import HTTPException
@@ -16,6 +15,7 @@ from sqlalchemy.orm import Session
 
 from onyx.auth.permissions import require_permission
 from onyx.auth.users import optional_user
+from onyx.cache.factory import get_cache_backend
 from onyx.db.engine.async_sql_engine import get_async_session_context_manager
 from onyx.db.engine.sql_engine import get_session
 from onyx.db.enums import Permission
@@ -65,12 +65,19 @@ def _get_proxy_client() -> httpx.AsyncClient:
     return _ASYNC_PROXY_CLIENT
 
 
-# Per-asset DB-skip caches. Only GRANTs cached; 30s TTL = max stale access
-# after un-sharing. Concurrent misses hit the DB redundantly (not a bug).
-_SANDBOX_URL_CACHE: TTLCache[UUID, str] = TTLCache(maxsize=10_000, ttl=60.0)
-_WEBAPP_ACCESS_CACHE: TTLCache[tuple[UUID, UUID], bool] = TTLCache(
-    maxsize=50_000, ttl=30.0
-)
+# Per-asset DB-skip caches, backed by the shared cache (Redis) so the entry is
+# shared across api_server pods rather than per-process. Only GRANTs are cached;
+# the access TTL bounds how long a grant outlives an un-share.
+_SANDBOX_URL_TTL = 60
+_WEBAPP_ACCESS_TTL = 30
+
+
+def _sandbox_url_cache_key(session_id: UUID) -> str:
+    return f"craft:webapp:url:{session_id}"
+
+
+def _webapp_access_cache_key(session_id: UUID, user_id: UUID) -> str:
+    return f"craft:webapp:access:{session_id}:{user_id}"
 
 
 def require_onyx_craft_enabled(
@@ -275,9 +282,10 @@ REWRITABLE_CONTENT_TYPES = {
 
 async def _get_sandbox_url(session_id: UUID) -> str:
     """Resolve a session's Next.js server URL; cache hits open no DB connection."""
-    cached = _SANDBOX_URL_CACHE.get(session_id)
+    cache = get_cache_backend()
+    cached = cache.get(_sandbox_url_cache_key(session_id))
     if cached is not None:
-        return cached
+        return cached.decode()
 
     async with get_async_session_context_manager() as db_session:
         target = await get_webapp_target_async(db_session, session_id)
@@ -291,7 +299,7 @@ async def _get_sandbox_url(session_id: UUID) -> str:
         raise HTTPException(status_code=404, detail="Sandbox not found")
 
     url = get_sandbox_manager().get_webapp_url(sandbox_id, nextjs_port)
-    _SANDBOX_URL_CACHE[session_id] = url
+    cache.set(_sandbox_url_cache_key(session_id), url, ex=_SANDBOX_URL_TTL)
     return url
 
 
@@ -398,7 +406,8 @@ async def _proxy_request(
 async def _check_webapp_access(session_id: UUID, user: User | None) -> None:
     # Cached grants short-circuit before the DB fetch; 404 (missing session)
     # must still beat 401 (unauthenticated), so only grants are cached.
-    if user is not None and _WEBAPP_ACCESS_CACHE.get((session_id, user.id)):
+    cache = get_cache_backend()
+    if user is not None and cache.get(_webapp_access_cache_key(session_id, user.id)):
         return
 
     async with get_async_session_context_manager() as db_session:
@@ -412,7 +421,9 @@ async def _check_webapp_access(session_id: UUID, user: User | None) -> None:
     if sharing_scope == SharingScope.PRIVATE and owner_id != user.id:
         raise HTTPException(status_code=404, detail="Session not found")
 
-    _WEBAPP_ACCESS_CACHE[(session_id, user.id)] = True
+    cache.set(
+        _webapp_access_cache_key(session_id, user.id), b"1", ex=_WEBAPP_ACCESS_TTL
+    )
 
 
 _OFFLINE_HTML = (_TEMPLATES_DIR / "webapp_offline.html").read_text()
@@ -452,7 +463,7 @@ async def get_webapp(
         if e.status_code in (502, 503, 504):
             # The cached URL may point at a dead/recreated pod; drop it so the
             # next request re-resolves from the DB.
-            _SANDBOX_URL_CACHE.pop(session_id, None)
+            get_cache_backend().delete(_sandbox_url_cache_key(session_id))
             return _offline_html_response()
         raise
 
--- a/backend/onyx/server/features/build/api/api.py
+++ b/backend/onyx/server/features/build/api/api.py
@@ -49,8 +49,7 @@ logger = setup_logger()
 _TEMPLATES_DIR = Path(__file__).parent / "templates"
 _WEBAPP_HMR_FIXER_TEMPLATE = (_TEMPLATES_DIR / "webapp_hmr_fixer.js").read_text()
 
-# Shared pooled async client. Lazy init so importing this module in tests
-# doesn't leave an unclosed client.
+# Lazy-init so importing this module (e.g. in tests) doesn't leak an open client.
 _ASYNC_PROXY_CLIENT: httpx.AsyncClient | None = None
 
 
@@ -65,9 +64,7 @@ def _get_proxy_client() -> httpx.AsyncClient:
     return _ASYNC_PROXY_CLIENT
 
 
-# Per-asset DB-skip caches, backed by the shared cache (Redis) so the entry is
-# shared across api_server pods rather than per-process. Only GRANTs are cached;
-# the access TTL bounds how long a grant outlives an un-share.
+# Redis-backed so cache entries are shared across pods. Only grants are cached.
 _SANDBOX_URL_TTL = 60
 _WEBAPP_ACCESS_TTL = 30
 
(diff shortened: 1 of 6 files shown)


backend/onyx/server/features/build/api/api.py
```diff
@@ -317,80 +322,107 @@ def _proxy_request(
...
+                raise HTTPException(status_code=502, detail="Bad gateway")
+            finally:
+                await response.aclose()
+            if "text/html" in content_type:
+                content = _inject_hmr_fixer(
+                    _rewrite_asset_paths(raw, str(session_id)), str(session_id)
+                )
+            else:
+                # JS/CSS: sandbox emits assetPrefix-prefixed URLs, no rewrite needed.
+                content = raw
```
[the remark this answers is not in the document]
rohoswagger: Fixed in a5cae1b. Restored `_rewrite_asset_paths` for all JS/CSS responses. The rewrite is idempotent — already-prefixed URLs (from `assetPrefix` in new sandboxes) are not double-rewritten, so both old sessions (bare `/_next/` refs) and new ones work correctly. The removed `else: content = raw` path was the regression.

backend/onyx/server/features/build/api/api.py
```diff
@@ -47,14 +49,27 @@
...
+    limits=httpx.Limits(max_keepalive_connections=50, max_connections=200),
+)
+
+# Per-asset DB-skip caches. Lock-free: only touched from the event loop with no
+# await between read and write. Only GRANTS are cached; the 30s access TTL is
+# the max window a viewer keeps access after a session is un-shared.
+_SANDBOX_URL_CACHE: TTLCache[UUID, str] = TTLCache(maxsize=10_000, ttl=60.0)
+_WEBAPP_ACCESS_CACHE: TTLCache[tuple[UUID, UUID], bool] = TTLCache(
+    maxsize=50_000, ttl=30.0
+)
```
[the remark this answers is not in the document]
rohoswagger: Fixed in a5cae1b. Updated the comment to accurately describe the situation: there are `await`s between the cache miss check and the write, so concurrent misses on the same key will each hit the DB (minor thundering-herd on cold start, not a correctness issue). The old 'no await between read and write' wording was wrong.

backend/onyx/server/features/build/api/api.py
```diff
@@ -47,14 +49,27 @@
...
 _WEBAPP_HMR_FIXER_TEMPLATE = (_TEMPLATES_DIR / "webapp_hmr_fixer.js").read_text()
 
+# Shared pooled client: async so a slow sandbox suspends its coroutine rather
+# than holding one of the ~40 shared sync threads. Lives for the process
+# lifetime (no shared-lifespan close hook, to keep Craft out of all images).
+_ASYNC_PROXY_CLIENT = httpx.AsyncClient(
+    timeout=30.0,
+    follow_redirects=True,
+    limits=httpx.Limits(max_keepalive_connections=50, max_connections=200),
+)
```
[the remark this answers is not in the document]
rohoswagger: Fixed in a5cae1b. Replaced the module-level `httpx.AsyncClient` instance with a `_get_proxy_client()` lazy initialiser that creates the client on first call. Tests that import the module no longer get an unclosed client as a side effect.

rohoswagger:
@greptile-ai @cubic-dev-ai review this

rohoswagger:
@greptile review this

backend/onyx/server/features/build/api/api.py
```diff
@@ -46,14 +49,33 @@
...
+            timeout=30.0,
+            follow_redirects=True,
+            limits=httpx.Limits(max_keepalive_connections=50, max_connections=200),
+        )
+    return _ASYNC_PROXY_CLIENT
+
+
+# Per-asset DB-skip caches. Only GRANTs cached; 30s TTL = max stale access
+# after un-sharing. Concurrent misses hit the DB redundantly (not a bug).
+_SANDBOX_URL_CACHE: TTLCache[UUID, str] = TTLCache(maxsize=10_000, ttl=60.0)
```
wenxi-onyx: Any reason why we can't use Redis here (`get_cache_backend()`)? A few thoughts:
- I think the in-memory cache here is fine as implemented, but using Redis would resolve the 30s stale grant + stale URL issues (since each API pod would have it's own in-memory cache that would need to be populated/drained on request)
- Redis would add a small amount of latency, but should be quick enough?
- Would avoid introducing `TTLCache`
- Would allow stricter typing possibly

backend/onyx/server/features/build/api/api.py
```diff
@@ -308,80 +326,99 @@ def _proxy_request(
...
-                )
-
-            return StreamingResponse(
-                content=response.iter_bytes(chunk_size=8192),
-                status_code=response.status_code,
-                headers=response_headers,
-                media_type=content_type or None,
-            )
-
+        response = await client.send(req, stream=True)
```
wenxi-onyx, line 339: Any reason why we need `stream=True`? 

Claude is telling me this is the source of the lifecycle complexity below. 
```
 If you await client.send(req) without stream=True, httpx manages the connection itself and the entire cleanup apparatus (_aiter_and_close, the flag, the disconnect tests) disappears.
```

Alternative suggestion (to keep stream)
```
A. Ownership-transfer flag (no behavior change). Collapse the three cleanup paths into one finally. aclose() is idempotent, so a single guard covers buffered success, buffered failure, and header-building failure; only a successful handoff to StreamingResponse skips it:

response = await client.send(req, stream=True)   # 502/504 handling as today
handed_off = False
try:
    response_headers = _build_headers(response, session_id, path)
    if any(ct in content_type for ct in REWRITABLE_CONTENT_TYPES):
        raw = await response.aread()              # errors → 502 via outer handler
        ...
        return Response(content=content, ...)
    stream = _aiter_and_close(response)
    handed_off = True
    return StreamingResponse(stream, ...)
finally:
    if not handed_off:
        await response.aclose()
```

backend/onyx/server/features/build/sandbox/kubernetes/kubernetes_sandbox_manager.py
```diff
@@ -354,6 +354,7 @@ def _build_nextjs_start_script(
 set -e
 cd {session_path}/outputs/web
 {install_check}
+export WEBAPP_ASSET_PREFIX="/api/build/sessions/$(basename {session_path})/webapp"
```
wenxi-onyx, line 357: If this fixes the thing Evan sent us, you should respond on that gist + make sure we comm. in Slack as to not waste their time.

wenxi-onyx (approved):

backend/onyx/server/features/build/api/api.py
rohoswagger: Done — swapped the two per-process `TTLCache`s for `get_cache_backend()` (Redis) in `api/api.py` (commit 5f241e5).

- Sandbox-URL and access-grant entries now live in the shared cache, so an entry populated by one api_server pod is visible to all → the per-pod stale window you flagged is gone.
- Keys are namespaced + tenant-scoped: `craft:webapp:url:{session_id}` and `craft:webapp:access:{session_id}:{user_id}`.
- Behaviour preserved: grant-only caching (so 404-missing still beats 401-unauth), URL TTL 60s / access TTL 30s, and the 502/503/504 path now `delete`s the URL key so a dead/recreated pod re-resolves.
- Unit tests updated to drive a fake `CacheBackend` (`TestWebappAccessCache`).

backend/onyx/server/features/build/api/api.py
rohoswagger: Kept `stream=True` and applied your ownership-transfer simplification (commit 5068f65).

The three cleanup paths (buffered `finally` + `except HTTPException` + `except BaseException`) are now a single `handed_off` flag with one `finally: if not handed_off: await response.aclose()`. `aclose()` is idempotent, so the guard covers buffered success/failure and header-building errors; only a successful `StreamingResponse` handoff sets `handed_off=True`, transferring connection ownership to `_aiter_and_close`.

We considered dropping `stream=True` entirely (simplest), but kept it so large binary assets stream instead of being fully buffered in memory. `test_streaming_closes_upstream_on_early_disconnect` and `test_buffered_response_closes_upstream` both still pass.

backend/onyx/server/features/build/api/api.py
rohoswagger: switched to use redis

backend/onyx/server/features/build/api/api.py
rohoswagger: stream=True is more efficient for serving large files/videos/images/etc, rather than storing in-memory. implemented your solution to keep stream

backend/onyx/server/features/build/sandbox/kubernetes/kubernetes_sandbox_manager.py
rohoswagger: nope, just sets groundwork for it

wenxi-onyx (approved):

PR #11594, opened 2026-06-01
fix(chat): recover raw answer when content/citation processing strips it empty

justin-tahara:
## Description

A model's answer can be entirely consumed by content/citation processing, leaving an empty answer even though the provider responded. The common trigger is an answer dominated by bracketed-numeric text that matches the citation pattern but isn't in the citation mapping (e.g. `[123456789012345]`, an array, a list of ids) — `DynamicCitationProcessor` strips such unmapped markers (existing, intended behavior).

The empty answer then raises `EmptyLLMResponseError` ("provider returned no text or tool calls") — misleading (it did respond) and unlogged. This became user-visible after the citation-regex ReDoS fix (#11527): the same long bracketed-digit content that used to hang the pod now streams to completion, where the stripping empties the answer. Reproduces across providers; works via Claude CLI (no citation post-processing).

**Fix:** in `run_llm_step_pkt_generator`, if the model emitted text but the answer is empty after processing (and there are no tool calls), surface the raw output instead of erroring, and log a warning. Gated to non-`REQUIRED` tool choice so deep-research / forced-tool flows keep using fallback tool extraction. Does **not** change the intended unmapped-citation stripping.

## How Has This Been Tested?

`TestEmptyAnswerRecovery` in `test_llm_step.py` (real generator + real `DynamicCitationProcessor`): bracketed-number / numeric-list recovery, normal prose unaffected, genuinely-empty stream still empty, `REQUIRED` no-recovery, recovered answer recorded in the tracing span, and pre-answer timing persisted. 229 chat tests pass; pre-commit clean.

## Additional Options

### DIFF
--- a/backend/onyx/chat/llm_loop.py
+++ b/backend/onyx/chat/llm_loop.py
@@ -12,6 +12,7 @@ from onyx.chat.citation_processor import CitationMode
 from onyx.chat.citation_processor import DynamicCitationProcessor
 from onyx.chat.citation_utils import update_citation_processor_from_tool_response
 from onyx.chat.emitter import Emitter
+from onyx.chat.llm_step import _looks_like_xml_tool_call_payload
 from onyx.chat.llm_step import extract_tool_calls_from_response_text
 from onyx.chat.llm_step import run_llm_step
 from onyx.chat.models import ChatMessageSimple
@@ -131,18 +132,6 @@ def _build_empty_llm_response_error(
             "completed. No text or tool calls were received from the upstream "
             "provider."
         ),
-    )
-
-
-def _looks_like_xml_tool_call_payload(text: str | None) -> bool:
-    """Detect XML-style marshaled tool calls emitted as plain text."""
-    if not text:
-        return False
-    lowered = text.lower()
-    return (
-        "<function_calls" in lowered
-        and "<invoke" in lowered
-        and "<parameter" in lowered
     )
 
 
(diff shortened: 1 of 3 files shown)


justin-tahara:
Both findings valid — fixed in 3a455dfc8c.

1. **Span recorded empty output** — the span closed before recovery ran (processors snapshot data on `__exit__`). Moved the citation flush + recovery inside the span so the trace reflects the answer the user got.
2. **Missing `set_pre_answer_processing_time`** in recovery's start branch (only when `answer_start` is still False) — mirrored the normal path.

Added regression tests; 229 chat tests green, pre-commit clean.

nmgarza5 (approved):
Approving to unblock but something to consider:

The empty-answer recovery fires in AUTO mode whenever there's raw output but no parsed tool calls, including when a model emits a `<function_calls>...` tool call as plain text. In that case it streams the raw XML markup to the client and sets it as answer, even though `llm_loop._try_fallback_tool_extraction` is about to parse that same XML into a real tool call (previously answer stayed None and nothing leaked). Suggest gating the recovery with not `_looks_like_xml_tool_call_payload(accumulated_raw_answer)` so tool-call markup isn't surfaced as an answer.

This is mainly to prevent context pollution and weird UX where the user would then see a bunch of XML gabagool

justin-tahara:
Valid — fixed in 05ae9d309c. `_extract_xml_tool_calls_from_response_text` keys on `<invoke name="...">` and treats `<parameter>` as optional, so it already handles zero-arg invocations. Requiring `<parameter>` in the detector meant a zero-arg tool call emitted as text was neither extracted (fallback skipped in AUTO) nor gated out of the empty-answer recovery — so the raw markup would leak. Relaxed to require only `<function_calls>` + `<invoke>`, which keeps the recovery gate and fallback detection unified. Added a zero-arg regression test. 231 chat tests green.

PR #11603, opened 2026-06-01
feat(craft): Seed external apps on Cloud

Danelegend:
## Description
Provisions built-in apps with from the credentials that have been populated in the .env. All built-in apps will be provisioned and managed with this change. If a built-in app does not have associated credentials, it will still be created but credentials will be set to empty.

Screenshot 2026-06-02 at 1 14 14 PM

## How Has This Been Tested?
Tests + Manual (Cloud MT / Non-Cloud / ST)

## Additional Options

---
## Summary by cubic
Seeds Onyx-managed built-in apps per Cloud tenant with Onyx-owned OAuth credentials and locks them so admins can only enable/disable and set action policies. Provisioning is idempotent, enforces one app per type, refreshes creds without wiping, and runs during tenant setup.

- New Features
  - Cloud provisioning: gated by `AUTO_PROVISION_DEFAULT_EXTERNAL_APPS` (default false; set true on Cloud); creates each built-in (disabled) and populates creds from `EXT_APP_<APP>_CLIENT_ID/SECRET` (now includes `GITHUB`); safe to re-run (rotates creds only when configured; preserves enablement/policies); seeds default action policies.
  - Managed provider model: added `OnyxManagedExtApp`; `SLACK`/`GMAIL`/`GOOGLE_CALENDAR`/`LINEAR`/`GITHUB` expose managed creds from env; `ExternalAppType.is_built_in`; registry `get_onyx_managed_provider`; DB helpers for fetching the single built-in per type and rotating org creds; patch helpers `UNSET`/`none_as_unset`/`is_set`.
  - Admin API: built-ins via `POST /admin/apps/built-in`; custom via `POST /admin/apps/custom` (multipart); updates via `PATCH /admin/apps/{id}` to toggle enablement and merge policies; Onyx-managed built-ins block create, credential/config edits, and deletion on Cloud; responses include `is_onyx_managed` and hide credentials/config.
  - UI: hides already-configured built-in types (`availableBuiltInDescriptors` with tests); managed apps use a policies-only modal and disable delete; switched to the new `POST`/`PATCH` endpoints; Jest now picks up `__tests__/`; added docs for cloud-managed credentials.

- Migration
  - Set env vars for managed creds: `EXT_APP_SLACK_CLIENT_ID/SECRET`, `EXT_APP_GMAIL_CLIENT_ID/SECRET`, `EXT_APP_GOOGLE_CALENDAR_CLIENT_ID/SECRET`, `EXT_APP_LINEAR_CLIENT_ID/SECRET`, `EXT_APP_GITHUB_CLIENT_ID/SECRET`. Set `AUTO_PROVISION_DEFAULT_EXTERNAL_APPS=true` on Cloud.
  - New tenants: no action — apps seed during provisioning. Existing tenants: no backfill; provisioning is safe to run operationally.

<sup>Written for commit e3d5c99c52e138114fc744042498ec25d834185d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11603?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Danelegend:
@cubic @greptile

backend/onyx/server/features/build/api/external_apps_api.py
```diff
@@ -166,12 +206,6 @@ def upsert_external_app(
             action_policies=action_policies,
         )
 
-    # create/update wrote the rows out-of-band (bulk delete + insert) within
-    # their own commit, so the app's loaded ``policies`` collection is stale.
-    # With ``expire_on_commit=False`` the commit won't refresh it; expire so the
-    # response reflects what was just persisted.
-    db_session.expire(app, ["policies"])
```
Danelegend, line 174: policies is now set on the app when writing policies, so we no longer need this

backend/ee/onyx/server/tenants/provisioning.py
```diff
@@ -539,6 +546,72 @@ def _upsert(request: LLMProviderUpsertRequest, default_model: str) -> None:
...
+def provision_built_in_external_apps(db_session: Session) -> None:
+    """Provision every Onyx-managed built-in external app into the current tenant
+    (disabled), populating Onyx-owned credentials from operator config where
+    available. ``onyx_managed=False`` built-ins are not seeded.
+
+    - **New app:** created disabled, with the operator's credentials (or empty if
+      none configured — still provisioned, just not enableable until creds exist).
+    - **Existing app:** credentials are refreshed in place; enabled state and
+      action policies are left untouched. Credentials are only overwritten when
+      operator config has an entry for that type, so a re-run never wipes the
```
wenxi-onyx: Do we want this behavior? If we remove an external app, say "we no longer want to support google at all," wouldn't we want to remove the oauth client creds from the tenants?

backend/onyx/server/features/build/api/external_apps_api.py
```diff
@@ -53,23 +56,35 @@
...
+    toggle enablement + set policies, never edit credentials/config or delete.
+    Custom, self-hosted built-ins, and ``onyx_managed=False`` built-ins are all
+    tenant/admin-owned, so none is "managed"."""
+    return MULTI_TENANT and is_onyx_managed_app_type(app_type)
+
+
 def _to_admin_response(app: ExternalApp) -> ExternalAppAdminResponse:
     # Display + lifecycle fields live on the linked Skill row.
     stored = {policy.action_id: policy.policy for policy in app.policies}
+    managed = _is_onyx_managed(app.app_type)
```
wenxi-onyx: nit: would personally perfer in-lining this. The logic is straight forward and I don't think the doc string adds antyhing.
```
managed = MUTLI_TENANT and is_onyx_managed_app_type(app_type)
```

backend/onyx/configs/app_configs.py
```diff
@@ -1398,6 +1398,23 @@ def get_current_tz_offset() -> int:
...
+AUTO_PROVISION_DEFAULT_EXTERNAL_APPS = (
+    os.environ.get("AUTO_PROVISION_DEFAULT_EXTERNAL_APPS", "true").lower() == "true"
+)
+
+# Onyx-owned OAuth credentials for built-in external apps, seeded per tenant at
+# provisioning time so cloud admins never register their own OAuth app. JSON
+# object mapping a built-in app_type to its credential fields, e.g.
+# {"gmail": {"client_id": "...", "client_secret": "..."}}. Parsed in
+# onyx/external_apps/managed_credentials.py; unset/empty is a valid no-op.
+MANAGED_EXTERNAL_APP_CREDENTIALS = os.environ.get(
```
wenxi-onyx: Thoughts on just flattening this and having a single config var for every item we need to read? I imagine is slightly easier to 1) insert into secrets and 2) validate on read from env vars. Don't feel super strongly about this, but do pref flattening.

Something like:
```
EXT_APP_GOOGLE_OAUTH_CLIENT_ID
EXT_APP_GOOGLE_OAUTH_CLIENT_SECRET
etc.
```
If you do this, you'll need to make the distinction clear since we already have "oauth google" vars for auth.

backend/onyx/db/external_app.py
```diff
@@ -327,12 +338,57 @@ def update_external_app(
...
+    app: ExternalApp,
+    organization_credentials: dict[str, str],
+) -> None:
+    """Replace an app's organization credentials and commit. Used by the
+    Onyx-managed provisioning/rotation path — deliberately touches nothing else
+    (enabled state, policies, gateway config are left untouched)."""
+    # EncryptedJson column accepts a plain dict and encrypts on write (same
+    # assignment shape as update_external_app's masked-credential restore).
+    app.organization_credentials = organization_credentials  # ty: ignore[invalid-assignment]
+    db_session.commit()
```
wenxi-onyx: For future, I think we pref some kind of consistency of "all db methods flush" or "all db methods commit." I know it's currently quite scattered across the codebase, but I've personally been opting for "callers commit" for atomic transactions (multiple related db operations flush then commit altogether).

backend/onyx/db/external_app.py
```diff
@@ -157,6 +146,29 @@ def get_external_apps(
     return list(db_session.scalars(stmt).all())
 
 
+def get_external_app_by_app_type(
```
wenxi-onyx: I think this is a bit confusing. Perhaps it's just the function name that's confusing? I.e. it's not immediately clear that this is only applicable to built-ins (without reading the doc string). Ideally, there's some typing or run-time guard that restricts `app_type` to only the non-CUSTOM values. Could do this on the Enum I believe.

backend/onyx/server/features/build/api/external_apps_api.py
wenxi-onyx: Also `is_onyx_managed_app_type` can be defined on the enum via 
```
class ExternalAppType(str, PyEnum)
   ...
   @property
       def is_onyx_managed(self) -> bool:
           return self is not ExternalAppType.CUSTOM
```

backend/onyx/db/external_app.py
```diff
@@ -327,12 +338,57 @@ def update_external_app(
...
+    """Replace an app's organization credentials and commit. Used by the
+    Onyx-managed provisioning/rotation path — deliberately touches nothing else
+    (enabled state, policies, gateway config are left untouched)."""
+    # EncryptedJson column accepts a plain dict and encrypts on write (same
+    # assignment shape as update_external_app's masked-credential restore).
+    app.organization_credentials = organization_credentials  # ty: ignore[invalid-assignment]
+    db_session.commit()
+
+
+def set_external_app_enablement_and_policies(
```
wenxi-onyx: The two checks in this function are quite awkward because the request contains two identifiers for the same built-in app `id` and `app_type`. By allowing (or requiring) callers to specify both, we're forced to validate that they agree.

Generally, I think ExternalApp APIs should be simplified significantly by consolidating on a single identifier (the id). AND by separating out the "is enabled + policies" update from the general upsert endpoint.

backend/onyx/server/features/build/api/external_apps_api.py
```diff
@@ -122,6 +137,32 @@ def upsert_external_app(
             OnyxErrorCode.INVALID_INPUT,
             "Custom apps must be managed via POST /admin/apps/custom.",
         )
+
+    if _is_onyx_managed(request.app_type):
```
wenxi-onyx: On cloud, I think this entire endpoint should be disabled. Instead, there should be a separate `PATCH /admin/apps/{id}` that only mutates `{enabled, action_policies}`.

backend/onyx/server/features/build/api/external_apps_api.py
```diff
@@ -406,6 +441,12 @@ def delete_external_app_admin(
             OnyxErrorCode.NOT_FOUND,
             f"External app with id {external_app_id} not found.",
         )
+    if _is_onyx_managed(app.app_type):
```
wenxi-onyx: See message about defining onyx managed on the enum itself

docs/craft/features/external-apps/cloud-managed-app-credentials.md
```diff
@@ -0,0 +1,261 @@
+# Cloud-Managed External-App Credentials — Plan
```
wenxi-onyx: nit: I've been having my claude update these docs specifically with the instruction that they're "dual-purpose implementation and post-implementation documentation" --> as a result, it revises out all of the design decisions and random, specific implementation details that are in the code. 

I would pref if we start doing this as it makes the docs significantly more readable/usable as actual documentation after the PR lands.

wenxi-onyx (changes requested):

Danelegend:
@greptile @cubic

backend/ee/onyx/server/tenants/provisioning.py
Danelegend: Yes but this only runs on tenant provisioning, so this existing-app path occurring is more defense in depth. Done in a way so that it matches how the llm provisioning handles conflicts

backend/ee/onyx/server/tenants/provisioning.py
```diff
@@ -539,6 +546,74 @@ def _upsert(request: LLMProviderUpsertRequest, default_model: str) -> None:
         )
 
 
+def provision_built_in_external_apps(db_session: Session) -> None:
+    """Provision every Onyx-managed built-in external app into the current tenant
```
wenxi-onyx: nit (not worth addressing): this doc string is so hard to understand 😅

backend/onyx/db/external_app.py
```diff
@@ -157,6 +146,35 @@ def get_external_apps(
...
+    stmt = (
+        select(ExternalApp)
+        .options(
+            selectinload(ExternalApp.skill),
+            selectinload(ExternalApp.policies),
+        )
+        .where(ExternalApp.app_type == app_type)
+        .order_by(ExternalApp.id)
+    )
+    return db_session.scalars(stmt).first()
```
wenxi-onyx: You might want to use `.scalar_one_or_none()` here --> it asserts that only one entry exists and raises if multiple of the same type are found. Pro is that it would hard fail and not accidentally use an unintended credential. Con is that it will hard fail for the user and will need to be debugged.

backend/onyx/db/external_app.py
```diff
@@ -256,40 +274,37 @@ def create_external_app(
...
     db_session: Session,
     external_app_id: int,
-    name: str,
-    description: str,
-    enabled: bool,
     app_type: ExternalAppType,
-    upstream_url_patterns: list[str],
-    auth_template: dict[str, Any],
-    organization_credentials: dict[str, str],
+    name: str | None = None,
```
wenxi-onyx: Checkout the UNSET sentinel value. Should be useful here I think

backend/onyx/db/external_app.py
```diff
@@ -348,38 +384,42 @@ def get_policies(
...
-    db_session: Session,
-    external_app_id: int,
+    app: ExternalApp,
     policies: dict[str, EndpointPolicy],
 ) -> None:
-    """Replace the app's per-action policy rows with exactly ``policies`` (full
-    delete + insert). No commit — runs inside the create/update transaction so
-    the app and its policies persist atomically. ``action_id`` validation
-    against the provider catalog is the caller's responsibility.
+    """Replace ``app``'s per-action policy rows with exactly ``policies``.
```
wenxi-onyx, line 404: Can we not just drop all existing and insert all the desired ones?

backend/onyx/external_apps/providers/base.py
```diff
@@ -143,6 +146,65 @@ def __init_subclass__(cls, *, abstract: bool = False, **kwargs: Any) -> None:
             )
 
 
+class OnyxManagedProvider(ExternalAppProvider, abstract=True):
```
wenxi-onyx: nit: would do something like `OnyxManagedExtApp` or `OnyxManagedApp` etc. --> provider is an overloaded term in our codebase 😓

backend/onyx/external_apps/providers/base.py
```diff
@@ -143,6 +146,65 @@ def __init_subclass__(cls, *, abstract: bool = False, **kwargs: Any) -> None:
...
+        """This provider's Onyx-owned credentials if fully configured, else None."""
+        present = {
+            field_key: stripped
+            for field_key, raw in self.managed_org_credentials.items()
+            if (stripped := raw.strip())
+        }
+        if not present:
+            return None
+        if len(present) != len(self.managed_org_credentials):
+            missing = [
```
wenxi-onyx: How is this possible if creation will fail if there are missing components to the credential?

backend/onyx/external_apps/providers/registry.py
```diff
@@ -158,6 +166,24 @@ def fetch_available_built_in_apps() -> list[BuiltInExternalAppDescriptor]:
     return [_descriptor_for(cls) for cls in _PROVIDER_CLASSES]
 
 
+def is_onyx_managed_app_type(app_type: ExternalAppType) -> bool:
```
wenxi-onyx: Can all instances of this be replaced with the new `is_built_in` field on the enum?

backend/onyx/external_apps/providers/registry.py
```diff
@@ -158,6 +166,24 @@ def fetch_available_built_in_apps() -> list[BuiltInExternalAppDescriptor]:
...
+    """Whether a built-in ``app_type`` has Onyx-owned credentials — i.e. its
+    provider is an :class:`OnyxManagedProvider`. False for ``CUSTOM`` (no
+    provider) and for non-managed built-ins. Not gated on ``MULTI_TENANT`` —
+    callers add that for the cloud-only lockdown."""
+    return isinstance(PROVIDERS.get(app_type), OnyxManagedProvider)
+
+
+def fetch_onyx_managed_built_in_apps() -> list[BuiltInExternalAppDescriptor]:
```
wenxi-onyx, line 173: I feel like this and `get_onyx_managed_provider` can somehow be consolidated (might be wrong, just spidey sense).

wenxi-onyx (approved):
I would recommend setting up local multi-tenant to test this end-to-end before merging

backend/onyx/db/external_app.py
Danelegend: No, sqlalchemy flushes inserts before deletes, so re-inserting any surviving action_iid collides with its not-yet-deleted row on the (external_app_id, action_id) unique constraint. The in-place reconcile sidesteps that.

backend/onyx/external_apps/providers/base.py
Danelegend: The class check only validates the credential keys; the values come from EXT_APP_* env vars (default ""), so an operator setting one but not the other yields all-keys-present-but-a-blank-value — and even then creation doesn't fail, it just provisions the app disabled with empty creds.

backend/onyx/external_apps/providers/registry.py
Danelegend: The idea is to leave this open for extension if we have built-in apps that aren't onyx managed. Ie, hubspot with api key drop-in rather than oauth

PR #11630, opened 2026-06-02
feat(auth): scope Personal Access Tokens via Permission intersection

wenxi-onyx:
## Description

Adds per-PAT scopes (PR 2 of the [PAT scopes plan](docs/craft/features/pat-scopes/pat-scopes.md)). Effective authority becomes `user permissions ∩ token scopes`, reusing the existing `Permission` enum and implication closure — no parallel system.

- Nullable `scopes` column on `PersonalAccessToken` (`NULL` = unrestricted, so existing tokens are unchanged; migration just adds the column).
- PAT auth surfaces the token's scopes on `request.state.token_scopes`; `require_permission` denies if the required permission isn't in both the user's permissions and the token's scopes.
- **Fail-closed for unguarded routes:** `optional_user` denies a scoped PAT on any route that declares no `require_permission` (reusing the `_is_require_permission` sentinel `check_router_auth` introspects). So a scoped token reaches only routes guarded by a permission it covers; unrestricted PATs / sessions / API keys / websocket auth are untouched.
- `create_pat(scopes=...)` to mint scoped tokens (internal for now); unscoped auth stays unrestricted.

Scope enforcement is now complete (guarded + unguarded routes). Re-guarding routes with finer permissions and scoping the Craft PAT are follow-up PRs. No behavior change until a token is explicitly scoped (none are yet).

## How Has This Been Tested?

Unit tests for the closure + the intersection gate (within-user passes, out-of-scope denied, scoped token caps an admin, token can't exceed user, empty scopes fail-closed) and the fail-closed route gate (scoped token denied on a guard-less route). Integration tests assert a scoped PAT is denied on a `require_permission` route it doesn't cover and on an unguarded route (`/me`), while an unrestricted PAT passes both.

## Additional Options

### DIFF
--- a/backend/onyx/auth/users.py
+++ b/backend/onyx/auth/users.py
@@ -35,6 +35,7 @@ from fastapi import status
 from fastapi import WebSocket
 from fastapi.responses import JSONResponse
 from fastapi.responses import RedirectResponse
+from fastapi.routing import APIRoute
 from fastapi.security import OAuth2PasswordRequestForm
 from fastapi_users import BaseUserManager
 from fastapi_users import exceptions
@@ -70,6 +71,7 @@ from sqlalchemy import select
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.orm import Session
+from starlette.routing import BaseRoute
 
 from onyx.auth.api_key import get_hashed_api_key_from_request
 from onyx.auth.disposable_email_validator import is_disposable_email
@@ -123,11 +125,12 @@ from onyx.db.engine.async_sql_engine import get_async_session_context_manager
 from onyx.db.engine.sql_engine import get_session_with_current_tenant
 from onyx.db.engine.sql_engine import get_session_with_tenant
 from onyx.db.enums import AccountType
+from onyx.db.enums import Permission
 from onyx.db.models import AccessToken
 from onyx.db.models import OAuthAccount
 from onyx.db.models import Persona
 from onyx.db.models import User
-from onyx.db.pat import fetch_user_for_pat
+from onyx.db.pat import resolve_pat
 from onyx.db.users import assign_user_to_default_groups__no_commit
 from onyx.db.users import get_user_by_email
 from onyx.db.users import is_limited_user
@@ -1794,6 +1797,32 @@ async def _maybe_refresh_oauth_tokens(
         )
 
 
+def _scoped_pat_permitted_on_route(
+    token_scopes: list[Permission] | None, route: BaseRoute | None
+) -> bool:
+    """Fail-closed coverage check for scoped PATs.
+
+    Unrestricted PAT / session / API-key auth (token_scopes is None) is
+    unaffected. A scoped PAT may only reach routes that declare a
+    require_permission dependency; require_permission then adjudicates whether
+    the scopes actually cover it. A route with no require_permission is denied,
+    so a narrowly-scoped token can't reach arbitrary endpoints that merely lack
+    a permission guard.
+    """
+    # No scopes (None) == unrestricted: session / unrestricted PAT / API-key
+    # auth carry no token cap, so every route is allowed.
+    if token_scopes is None:
+        return True
+    # Only an APIRoute carries a dependant; anything else (no match, a mounted
+    # or websocket route) declares no permission, so deny the scoped PAT.
+    if not isinstance(route, APIRoute):
+        return False
+    return any(
+        getattr(dependency.cache_key[0], "_is_require_permission", False)
+        for dependency in route.dependant.dependencies
+    )
+
+
 async def optional_user(
     request: Request,
     async_db_session: AsyncSession = Depends(get_async_session),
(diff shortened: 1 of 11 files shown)


wenxi-onyx:
@greptileai review this

wenxi-onyx:
@greptileai re-review

backend/onyx/auth/users.py
```diff
@@ -1807,7 +1807,11 @@ async def optional_user(
 
     try:
         if hashed_pat := get_hashed_pat_from_request(request):
-            user = await fetch_user_for_pat(hashed_pat, async_db_session)
+            pat_result = await fetch_user_for_pat(hashed_pat, async_db_session)
+            if pat_result is not None:
+                # Expose the token's scopes so require_permission can cap the
+                # request to them.
+                user, request.state.token_scopes = pat_result
```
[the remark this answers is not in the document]
wenxi-onyx: I don't think this is a vulnerability. Session/cookie and PAT are *alternative* credentials, and a valid session is full, unrestricted auth. A scoped PAT only caps a request when it **is** the authenticating credential. When both are present the request is authenticated by the session — i.e. the caller is already that fully-authenticated user — so ignoring the token grants nothing beyond what the session already grants. There's no privilege escalation: you can't reach the session path without already holding a valid session. Scoped PATs exist for credential-only clients with no cookie (e.g. the Craft sandbox), which never hit the session path. You can't downgrade your own authenticated session by attaching a weaker token, and the server doesn't promise that. Leaving session-first precedence as-is.

backend/onyx/db/enums.py
```diff
@@ -532,3 +533,18 @@ class Permission(str, PyEnum):
...
+    Stored values are validated on write, so an unknown one means stale/corrupt
+    data; dropping it fails safe (it only ever narrows access).
+    """
+    parsed: list[Permission] = []
+    for value in values:
+        try:
+            parsed.append(Permission(value))
+        except ValueError:
+            continue
+    return parsed
```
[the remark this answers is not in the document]
wenxi-onyx: Deliberate trade-off. The logging was dropped to keep `enums.py` a pure leaf — it's imported by ~160 modules, and our `setup_logger` would wire `shared_configs` into all of them. Unknown values are near-impossible in practice (permission/scope strings are validated on write) and dropping them is fail-safe, so the lost signal is low value. The suggested `warn_on_unknown` / `context` params are opt-in and neither caller would pass them, so they'd be dead code. Leaving as-is.

backend/onyx/db/pat.py
```diff
@@ -24,34 +27,46 @@
...
+                | (PersonalAccessToken.expires_at > now)
+            )
         )
-    )
-    if not user:
+    ).first()
+    if result is None:
         return None
 
+    user, scopes = result
```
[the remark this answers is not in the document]
wenxi-onyx: Good catch — real bug, and it's fixed in the latest revision. `resolve_pat` (renamed from `fetch_user_for_pat`) now selects the `PersonalAccessToken` entity with `contains_eager(PersonalAccessToken.user)` and uses `.scalars().unique().one_or_none()`, so the joined `oauth_accounts` collection is fully populated — matching the `.scalars().unique()` convention in saml.py / the api_key list fetch. Single query, scopes read as an attribute.

backend/onyx/auth/users.py
```diff
@@ -1794,6 +1795,29 @@ async def _maybe_refresh_oauth_tokens(
         )
 
 
+def _scoped_pat_permitted_on_route(
+    token_scopes: list[Permission] | None, route: object
```
evan-onyx: why is this typed as `object`?

backend/onyx/auth/users.py
```diff
@@ -1794,6 +1795,29 @@ async def _maybe_refresh_oauth_tokens(
...
+    Unrestricted PAT / session / API-key auth (token_scopes is None) is
+    unaffected. A scoped PAT may only reach routes that declare a
+    require_permission dependency; require_permission then adjudicates whether
+    the scopes actually cover it. A route with no require_permission is denied,
+    so a narrowly-scoped token can't reach arbitrary endpoints that merely lack
+    a permission guard.
+    """
+    if token_scopes is None:
+        return True
+    dependant = getattr(route, "dependant", None)
```
evan-onyx: I think this check needs an explanation

backend/onyx/db/enums.py
```diff
@@ -532,3 +533,18 @@ class Permission(str, PyEnum):
...
+    """Parse stored permission strings into Permissions, dropping unknown values.
+
+    Stored values are validated on write, so an unknown one means stale/corrupt
+    data; dropping it fails safe (it only ever narrows access).
+    """
+    parsed: list[Permission] = []
+    for value in values:
+        try:
+            parsed.append(Permission(value))
+        except ValueError:
```
evan-onyx: probably should log these

evan-onyx (approved):
fine in principle but the implementation needs a few nits addressed

backend/onyx/db/enums.py
wenxi-onyx: Claude wanted to put this func in the enums file, but I didn't want to pollute the import with `import logging`. I thought it was unlikely that anyone would ever really look for this log, but yea should probably do anyways. Will move the func to a different file though.

backend/onyx/auth/users.py
wenxi-onyx: Explained in a few other comments, but no scopes = unrestricted while empty list = no permissions. Will add a comment here too. Do you think it's too confusing that null = unrestricted? Originally, there was a wildcard sentinel value `*`, but the behavior is the same if we just use null

backend/onyx/auth/users.py
wenxi-onyx: :claude-intensifies:

PR #11631, opened 2026-06-02
feat(chat): use RE2 (google-re2) for citation-processor regexes

justin-tahara:
## Description

Trials Google **RE2** (`google-re2`) for the streaming citation processor's regexes — a follow-up to the ReDoS fix (#11527) and the empty-answer recovery work. RE2 matches in **linear time with no backtracking**, so these patterns — which run over LLM-generated text on every streamed token and were the original CPU-pinning ReDoS site — become *structurally* immune to catastrophic backtracking regardless of model output, instead of relying on hand-authored backtracking-safe patterns.

Kept intentionally small so we can run it on `main` and see how it behaves before expanding:

- New shim `onyx/utils/regex_engine.compile_linear()` — compiles with RE2 when available, **falls back to stdlib `re`** (with a one-time warning) if the wheel is missing, so no environment breaks. Exposes `USING_RE2` so the active engine is observable.
- `citation_processor.py`: the two citation patterns now compile via `compile_linear`; the one `re.search(pattern, seg)` call became `pattern.search(seg)` (RE2 patterns are method-style).
- `google-re2==1.1.20251105` added to `pyproject.toml` (backend group); lock + `default.txt` regenerated. Official Google package with prebuilt manylinux/macOS/Windows wheels for py3.9–3.14, so no source build in the Docker image.

**Not changed (deliberately, to keep this small):** `citation_utils.py` citation patterns and the XML tool-call extraction in `llm_step.py` — the latter uses a backreference, which RE2 doesn't support and would need a rewrite. Easy follow-ups if we keep this.

RE2's feature gaps (no backreferences / lookaround / atomic groups) don't affect the citation patterns — they use only character classes, anchors, groups, and quantifiers.

## How Has This Been Tested?

- The full existing `test_citation_processor.py` suite (129 tests) passes with RE2 active — it's the behavioral-equivalence oracle, and it covers the cases most likely to diverge (e.g. `match.lastindex` distinguishing `[[1]]` vs `[1]`, unicode brackets, code-block handling, comma-separated citations). No behavior change.
- New `test_regex_engine.py`: asserts RE2 is the active engine (so a missing wheel surfaces loudly rather than silently falling back), basic API, and that a textbook catastrophic-backtracking pattern (`(a+)+$`) resolves in <1s under RE2.
- 234 chat/utils unit tests green; pre-commit (ruff, ty, uv-lock/export) clean.

## Additional Options

---
## Summary by cubic
Switches the streaming citation regexes to a RE2-backed shim for linear-time matching, removing ReDoS/CPU pinning risk while keeping behavior the same. Adds `google-re2` with a safe fallback to stdlib `re` (warning now fires on first use, not import).

- **New Features**
  - Added `onyx/utils/regex_engine.compile_linear()` using `google-re2`; falls back to `re` with a one-time, deferred warning and exposes `USING_RE2`.
  - Updated `backend/onyx/chat/citation_processor.py` to compile citation patterns via the shim; behavior unchanged.
  - Added `backend/tests/unit/onyx/utils/test_regex_engine.py` to confirm RE2 is active, pathological patterns are linear, and fallback warns once.

- **Dependencies**
  - Added `google-re2==1.1.20251105` to backend; updated `uv.lock` and `backend/requirements/default.txt`.

<sup>Written for commit 9c68763bff295628a2e27f6f52ad37ce3205a003. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11631?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


justin-tahara:
Valid — fixed in 9c68763bff (thanks @cubic-dev-ai / greptile). Moved the stdlib-fallback warning out of the module-level `except ImportError` into a cached `_warn_fallback_once()` that fires on first fallback use in `compile_linear`, so there is no import-time logging side effect (matches `best_practices.md`). One-time semantics preserved; `USING_RE2` is still set at import (a plain bool, no I/O) so the test that asserts RE2 is active is unaffected. Added a warn-once regression test. 133 unit tests green, pre-commit clean.

PR #11632, opened 2026-06-02
fix(search-settings): coerce null embedding prefixes to empty string

justin-tahara:
## Description

Creating new search settings for a registry-less cloud embedding provider (e.g. Bedrock / LiteLLM / Azure) with the optional prefix fields left blank sent `query_prefix` / `passage_prefix` as `None`. The `search_settings` DB columns are `NOT NULL`, so the insert failed with:

```
psycopg2.errors.NotNullViolation: null value in column "query_prefix" of relation "search_settings" violates not-null constraint
```

The SQLAlchemy model declares these columns `nullable=True`, but no migration ever dropped the original `NOT NULL` constraint, so the DB still rejects NULLs. This latent bug surfaced once #11606 unblocked the UI flow for connecting these providers.

Fix is **pure code, no schema migration** (so it can be cherry-picked cleanly):
- Coerce `None -> ""` (the empty-string "no prefix" sentinel, matching the original embedding-models migration semantics) in `create_search_settings`.
- Default the Pydantic request fields (`EmbeddingModelDetail.query_prefix` / `passage_prefix`) to `""` so `None` never arrives when the client omits them.

The model's `nullable=True` annotation is left as-is to keep the hotfix surface minimal.

## How Has This Been Tested?

Added two external-dependency-unit regression tests in `test_search_settings.py`:
- `test_create_search_settings_coerces_none_prefixes` — drives the exact `None` path through the DB layer and asserts it persists `""` instead of raising `NotNullViolation`.
- `test_creation_request_defaults_blank_prefixes` — asserts the request model defaults blank prefixes to `""`.

Both pass. Pre-commit (ty type-check, ruff) passes.

## Additional Options

---
## Summary by cubic
Fixes DB failures when creating or updating search settings with blank embedding prefixes by coercing null prefixes to empty strings and defaulting request fields to avoid NOT NULL errors.

- **Bug Fixes**
  - Coerce `None` to `""` for `query_prefix` and `passage_prefix` in both `create_search_settings` and `update_search_settings` to satisfy NOT NULL DB columns.
  - Default `EmbeddingModelDetail.query_prefix` and `EmbeddingModelDetail.passage_prefix` to `""` so omitted fields never send `None`.
  - Added regression tests for create/update paths and request defaults. No schema migration required.

<sup>Written for commit 8bbaaf4c6290d4b86e28d05d6355578d47a1e79c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11632?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/db/search_settings.py
+++ b/backend/onyx/db/search_settings.py
@@ -42,8 +42,11 @@ def create_search_settings(
         model_name=search_settings.model_name,
         model_dim=search_settings.model_dim,
         normalize=search_settings.normalize,
-        query_prefix=search_settings.query_prefix,
-        passage_prefix=search_settings.passage_prefix,
+        # These DB columns are NOT NULL. Registry-less cloud providers (e.g.
+        # Bedrock/LiteLLM/Azure) can arrive without prefixes, so coerce None to
+        # the empty-string "no prefix" sentinel to avoid a NotNullViolation.
+        query_prefix=search_settings.query_prefix or "",
+        passage_prefix=search_settings.passage_prefix or "",
         status=status,
         index_name=search_settings.index_name,
         provider_type=search_settings.provider_type,
--- a/backend/onyx/db/search_settings.py
+++ b/backend/onyx/db/search_settings.py
@@ -21,6 +21,11 @@ from shared_configs.enums import EmbeddingProvider
 
 logger = setup_logger()
 
+# search_settings columns that are NOT NULL but whose Pydantic source fields are
+# typed str | None. Registry-less cloud providers (e.g. Bedrock/LiteLLM/Azure)
+# can arrive without prefixes; coerce None -> "" to avoid a NotNullViolation.
+_NOT_NULL_STR_FIELDS = {"query_prefix", "passage_prefix"}
+
 
 class ActiveSearchSettings:
     primary: SearchSettings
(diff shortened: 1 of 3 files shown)


justin-tahara:
Good catch — addressed in 8bbaaf4c62.

Worth noting the nuance: this path isn't reachable through current callers. Both `update_current_search_settings` and `update_secondary_search_settings` pass `PRESERVED_SEARCH_FIELDS`, which already includes `query_prefix` and `passage_prefix` (and `update_secondary_search_settings` even hard-reassigns it), so the prefixes are always preserved and never overwritten on update today.

Still, I've added the same `None -> ""` coercion to `update_search_settings` as defense-in-depth so a future caller passing a `preserved_fields` list without the prefixes can't reintroduce the `NotNullViolation`. The coerced-field set is now a shared `_NOT_NULL_STR_FIELDS` constant used by both the create and update paths, plus a regression test that exercises the update path with the prefixes un-preserved.

jmelahman (approved):

PR #11671, opened 2026-06-03
feat(craft): clean up the Craft sidebar to four tabs and rehome settings

rohoswagger:
## Description

Cleans up the Onyx Craft sidebar and rehomes the settings that lived on the (now-deleted) Configure page.

**Sidebar → four tabs:** `Start Crafting · Scheduled Tasks · Skills · Apps`.

**Apps** is now one unified page instead of the split *My Apps* / *Manage Apps* tabs:
- The Craft `/craft/v1/apps` page shows the user's own connections (Connected list + a Browse grid with search), plus an admin/curator-only **Manage apps** button.
- Org-wide app configuration now lives **inside the Craft UI** at `/craft/v1/apps/manage` (admin/curator-gated), reached via that button; `/craft/v1/apps/admin` redirects there.
- Skills management works the same way: the `/craft/v1/skills` page has a **Manage skills** button → `/craft/v1/skills/manage`.
- Both management surfaces were **removed from the main app admin panel** (Skills + External Apps routes/nav), keeping Craft administration within Craft. The manage pages baseline-align with their personal pages (shared `md` width, inline header action) so navigating in/out doesn't shift the layout.

**Default LLM → composer model picker.** The model is now a picker in the chat composer (matching the main app's `ModelSelector` placement/pill), shown only before the first message of a session. It's **session-scoped**: each session stores its model (`agent_provider`/`agent_model`), seeded into the picker on load (resetting when a session has none, so a prior session's pick can't leak). Sending a message threads an optional `model`/`provider` override on `MessageRequest`, persisted to the session row — no sandbox re-provision needed (opencode-serve resolves the per-prompt override across the pre-registered providers). The old per-user selection cookie is removed.

**User Library → composer `+` menu.** Added a Library flyout (files + "Manage library…") via the shared `FlyoutRow`; the `UserLibraryModal` moved out of the deleted Configure page.

**Misc UI:**
- Deleted the Configure page, `CRAFT_CONFIGURE_PATH`, and the "Connect your data" row (the welcome-screen `ConnectDataBanner` still nudges admins).
- Welcome screen mirrors the main AppPage layout: an "onyx Craft" lockup (typed Onyx wordmark + baseline-aligned "Craft") on the left, model picker on the right, AppPage spacing/width.
- Real Google Calendar brand logo (`@opal/logos/google-calendar`) for the `GOOGLE_CALENDAR` app type.
- Fixed admin app-card alignment (equal-height, centered icons/buttons) and the `+`-menu popover/tooltip flashing over the Library modal.
- All Craft pages (Tasks/Skills/Apps) use the standard `md` page width.
- Added a `ModelPickerButton` Storybook story.

Design notes: `docs/craft/ui/sidebar-cleanup.md`.

## Screenshots

**Welcome** — "onyx Craft" lockup, composer model picker, 4-tab sidebar
Welcome

**Apps** — personal connections + admin "Manage apps" button
Apps

**Manage apps** — in-Craft org-wide app configuration (admin/curator), reached via "Manage apps"
Manage apps

**Skills** — available skills + admin "Manage skills" button
Skills

**Manage skills** — in-Craft org-wide skill management (admin/curator), reached via "Manage skills"
Manage skills

## How Has This Been Tested?

Manual verification against the local dev app (logged in, walked the welcome, Apps, Skills, Scheduled Tasks, and the in-Craft Manage Apps/Skills pages — confirmed layout, the model picker, the Library flyout, app-card alignment, the personal↔manage transition, and the redirect). `tsc --noEmit`, `oxlint`, and `oxfmt` all clean; backend `ruff`/`ty` pass via pre-commit. Automated Playwright (Apps scopes + `/apps/admin` redirect) and an integration test for the per-message model override are noted as follow-ups in the spec.

## Additional Options


web/src/app/craft/components/ModelPickerButton.tsx
```diff
@@ -0,0 +1,78 @@
+"use client";
+
+import { useMemo } from "react";
+import { SelectButton } from "@opal/components";
+import { BuildLLMPopover } from "@/app/craft/components/BuildLLMPopover";
```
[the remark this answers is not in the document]
rohoswagger: Fixed in 455254cfdb — added `web/src/app/craft/components/ModelPickerButton.stories.tsx` (title `Apps/Craft/Model Picker Button`) covering the **Default** (no selection → recommended default), **Selected**, and **Disabled** states. It mirrors `CraftInputBar.stories.tsx`'s setup: wraps `SWRConfig` (with an `llmProviders` fallback), `UserProvider`, and `BuildOnboardingProvider` so the component's `useLLMProviders`/`useOnboarding` hooks resolve.

web/src/app/craft/components/ChatPanel.tsx
```diff
@@ -77,6 +79,20 @@ export default function BuildChatPanel({
...
+  const [selectedModel, setSelectedModel] = useState<BuildLlmSelection | null>(
+    null
+  );
+  useEffect(() => {
+    if (session?.agentProvider && session?.agentModel) {
+      setSelectedModel({
+        provider: session.agentProvider,
+        providerName: session.agentProvider,
+        modelName: session.agentModel,
+      });
```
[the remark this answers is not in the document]
rohoswagger: Fixed in 978a47d2ed (`ChatPanel.tsx`). The seeding effect now resolves `providerName` from `llmProviders` instead of the raw provider key — but using the matched provider's `.name` (`match?.name ?? session.agentProvider`), not `provider_display_name`. That matches the canonical contract: `getDefaultLlmSelection` seeds `providerName: match.name`, the `BuildLlmSelection` type documents `providerName` as `LLMProviderDescriptor.name`, and `BuildLLMPopover` consumes `providerName` as a lookup key (`isProviderConfigured` / `getProviderDescriptor`). So `.name` is the consistent value here rather than the display label.

web/src/app/craft/components/buildEntryMenuItems.tsx
```diff
@@ -61,5 +85,27 @@ export function buildEntryMenuItems(
...
+        ...libraryFiles.map((file) => ({
+          key: file.id,
+          icon: SvgFileText,
+          label: file.name,
+          onSelect: onManageLibrary,
+        })),
+        {
+          key: "manage",
+          icon: SvgFolder,
+          label: "Manage library…",
```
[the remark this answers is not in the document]
rohoswagger: Addressed in f733f5dd22 with a documented `TODO`. There's no per-file attach handler in this flyout — `buildEntryMenuItems` only receives `onSelectEntry`/`onManageLibrary` from the inherited craft-input-bar feature — so wiring real per-file attach is a separate change. Added `TODO(craft-library)` on the file-row `onSelect` noting the intended attach behavior and that rows currently open the manage modal, so the gap is tracked rather than drifting.

web/src/app/craft/components/ChatPanel.tsx
```diff
@@ -77,6 +79,20 @@ export default function BuildChatPanel({
...
+    }
```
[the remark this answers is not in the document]
rohoswagger: Fixed in 978a47d2ed (`ChatPanel.tsx`). The seeding effect now calls `setSelectedModel(null)` when the active session has no `agentProvider`/`agentModel`, so a prior session's model can't leak into a new session's first message. (Same effect also addresses the adjacent `providerName` thread.)

wenxi-onyx (approved):

backend/onyx/server/features/build/db/build_session.py
```diff
@@ -196,6 +196,25 @@ def update_session_status(
...
+    agent_provider: str,
+    agent_model: str,
+    db_session: Session,
+) -> None:
+    """Persist the agent provider/model on the session row so a composer
+    model override sticks across reloads."""
+    session = (
+        db_session.query(BuildSession)
+        .filter(BuildSession.id == session_id)
+        .one_or_none()
```
wenxi-onyx: Can you just do `.one()` here? Any reason why this would be called on a non-existent session_id?

backend/onyx/server/features/build/session/manager.py
```diff
@@ -684,9 +685,32 @@ def send_message(
...
+            session_id: The session UUID
+            user_id: The user ID
+            content: The message content
+            agent_provider: Optional per-message model provider override
+            agent_model: Optional per-message model name override
+
+        Yields:
+            SSE formatted event strings
+        """
+        # Persist the picker's override up front; the turn reads it back from
```
wenxi-onyx: comment vomit :claude-intensifies:

web/src/app/craft/components/buildEntryMenuItems.tsx
```diff
@@ -23,7 +43,11 @@ export function buildEntryMenuItems(
     },
   ];
 
-  if (sections.skills.length > 0 || sections.apps.length > 0) {
+  if (
+    sections.skills.length > 0 ||
+    sections.apps.length > 0 ||
```
wenxi-onyx: Should double check that apps still exist when there are no org creds (i.e. should expose to users that you can connect these apps)

web/src/app/craft/components/ChatPanel.tsx
```diff
@@ -77,6 +80,30 @@ export default function BuildChatPanel({
   const { isMobile } = useScreenSize();
   const toggleOutputPanel = useToggleOutputPanel();
 
+  // Model the next message will use, seeded from the loaded session's row.
```
wenxi-onyx: This whole dance feels unnecessary. If we already know the provider/model in the BE, why do we need to pass it to the FE to then pass it back to the BE.

web/src/app/craft/components/CraftInputBar.tsx
```diff
@@ -185,8 +202,13 @@ const CraftInputBar = memo(
           buildEntryMenuItems(pickerSections, {
             onAttachFiles: () => fileInputRef.current?.click(),
             onSelectEntry: addEntry,
+            libraryFiles,
+            // Let the + menu finish its close animation before the modal
+            // mounts, otherwise the closing popover paints over it.
+            onManageLibrary: () =>
+              window.setTimeout(() => setLibraryModalOpen(true), 200),
```
wenxi-onyx, line 212: There must be a better way to do this...

web/src/app/craft/onboarding/constants.ts
```diff
@@ -91,6 +91,16 @@ const RECOMMENDED_MODEL_NAMES = new Set(
   )
 );
 
+// Display label of the top recommended Craft model (the recommended provider's
+// recommended model). Derived so UI copy never hardcodes a stale version.
+export const RECOMMENDED_CRAFT_MODEL_LABEL: string = (() => {
+  const provider =
+    BUILD_MODE_PROVIDERS.find((p) => p.recommended) ?? BUILD_MODE_PROVIDERS[0]!;
```
wenxi-onyx, line 97: Maybe not in scope for this PR, but we should probably migrate to use the same recommended LLMs JSON that the main chat uses --> it refreshes the list from GitHub ~daily and falls back to hardcoded list in image if fetch fails

web/src/app/craft/v1/apps/admin/page.tsx
```diff
@@ -1,321 +1,14 @@
...
-          Create
-        </Button>
-      </div>
-    </Card>
-  );
+import { useEffect } from "react";
+import { useRouter } from "next/navigation";
+
+// App management now lives at /craft/v1/apps/manage. Keep this route as a
+// redirect so existing links don't break.
```
wenxi-onyx: What existing links?

web/src/app/craft/v1/apps/manage/page.tsx
```diff
@@ -0,0 +1,23 @@
+"use client";
```
wenxi-onyx, line 1: This should be gated server side in a layout like normal admin pages. 

Also, all of the APIs requires full admin access, which excludes curators.

web/src/app/craft/v1/skills/manage/page.tsx
```diff
@@ -0,0 +1,23 @@
+"use client";
```
wenxi-onyx, line 1: Ditto above, use shared layout for craft management pages

wenxi-onyx (changes requested):

backend/onyx/server/features/build/db/build_session.py
rohoswagger: Done in f9650a9091 — switched to `.one()`. `update_session_agent_selection` is only called from `send_message` for an already-loaded session, so the `.one_or_none()` + silent `if session:` guard was effectively dead. `.one()` now surfaces the invariant if a non-existent id ever reaches it.

backend/onyx/server/features/build/session/manager.py
rohoswagger: Trimmed in f9650a9091 to a single line:
`# Persist the per-message model override; the turn reads it off the session row.`

web/src/app/craft/components/buildEntryMenuItems.tsx
rohoswagger: Verified — no change needed. `buildPickerSections` (`web/src/lib/skills/picker.ts`) maps **every** entry from `externalApps` regardless of `authenticated`, and `buildEntryMenuItems` renders unauthenticated ones with a `Connect` affordance (`rightContent` when `!app.authenticated`). So org-enabled apps the user hasn't connected still surface in the `+` menu as connectable; the Apps section only hides when the org has zero enabled apps (i.e. nothing to connect).

web/src/app/craft/components/ChatPanel.tsx
rohoswagger: The BE doesn't already know it for a **new** session — `agent_provider`/`agent_model` start null and are only set once the user picks (or accepts the default) and sends the first message. The seed (`session.agentProvider → picker`) is purely to *display* the current value when one exists (e.g. a reloaded/pre-provisioned session); the picker is hidden after the first message. So it's BE→FE to display, FE→BE to set on first send. A deeper refactor (BE defaults the model, FE only sends on an explicit change) could drop the seed-and-resend, but that's a larger change I kept out of this PR.

web/src/app/craft/components/CraftInputBar.tsx
rohoswagger: Agreed it's a workaround. The 200ms defers mounting `UserLibraryModal` until the `+` popover finishes its close animation — otherwise the closing popover paints over the modal (same layering issue as the earlier tooltip-flash fix). The clean fix is an `onExitComplete`/animation-end callback from the Popover so we open the modal exactly when it's gone, but the `@opal` Popover doesn't expose one today. Left as-is to avoid a shared-component change in this PR; can file a follow-up to add that callback.

web/src/app/craft/onboarding/constants.ts
rohoswagger: Agreed, and out of scope here as you note. `BUILD_MODE_PROVIDERS` is currently a hardcoded list; migrating Craft to the main chat's daily-refreshed recommended-LLMs JSON (with the in-image hardcoded fallback) is a good follow-up. Leaving this PR as-is and we can track that separately.

web/src/app/craft/v1/apps/admin/page.tsx
rohoswagger: Good catch — clarified the comment in 1639c594f8. `/craft/v1/apps/admin` was a real shipped route on `main` (the old admin apps management page — confirmed via `git cat-file -e origin/main:web/src/app/craft/v1/apps/admin/page.tsx`), so the redirect preserves bookmarks/links to that path now that management moved to `/craft/v1/apps/manage`.

web/src/app/craft/v1/apps/manage/page.tsx
rohoswagger: Fixed in 3b811328ea. Added a shared server-side `CraftManageLayout` (`web/src/layouts/craft/CraftManageLayout.tsx`), used via `apps/manage/layout.tsx` and `skills/manage/layout.tsx`, mirroring the admin panel's layout gating.

It gates **full-admin only** — `requireAuth()` + `role === ADMIN` — rather than `requireAdminAuth`, which allows `CURATOR`/`GLOBAL_CURATOR`. As you noted, the `/api/admin/*` endpoints these pages call require full admin, so the gate is stricter to match. Dropped the per-page client-side `useUser` redirect, and reverted the "Manage apps/skills" buttons to admin-only so curators aren't pointed at a page they can't use.

web/src/app/craft/components/CraftInputBar.tsx
wenxi-onyx: Could this be lowkey resolved just by having a higher z-index? Then it doesn't matter whether or not/when you unmount

web/src/app/craft/v1/skills/manage/page.tsx
rohoswagger: eh i guess i can create this, but don't think ill be adding an entrypoint to this rn, think its a bit overkill/don't really know a good place to put this esp given how little "admin config" there is atp

web/src/app/craft/components/ChatPanel.tsx
wenxi-onyx: I think it's not a huge refactor to do something like BE always defaults to value derived from DB, but falls back to request values if null. Your call

web/src/app/craft/v1/skills/manage/page.tsx
wenxi-onyx: Is the https://github.com/onyx-dot-app/onyx/commit/3b811328eabd4165af4cc6af352f65550fcc3424 layout not usable here?

web/src/app/craft/components/ChatPanel.tsx
rohoswagger: its bc DB has a default / suggested model but if user overrides that we have to respect that, its bc we pre-provision sessions that we have to fetch the pre-provisioned model

rohoswagger: Simplified the FE side in 45c1061ee3 — dropped the `useState` + `useEffect` that synced `selectedModel` from the session row. It's now derived: a memoized `sessionModel` (from the row) plus a per-session override map, where the picker writes the user's pick keyed by session id. No sync effect, and the per-session keying structurally prevents the cross-session leak.

On the round-trip itself, I kept it as-is: the row is already the source of truth (the stream reads `session.agent_provider`/`agent_model`); the request only *sets* it, and only on the first message (the picker is hidden afterward). The value rides on the session-load response and the send-message request, both of which happen regardless — so there's no dedicated FE↔BE hop to remove. The "DB defaults, request fallback if null" variant would actually drop a user's pick on an already-provisioned session (row would win over the picker), so request-sets-row → stream-reads-row is the correct shape.

web/src/app/craft/components/CraftInputBar.tsx
rohoswagger: nah pref not messing around with z-indexes, has a lot of downstream effects across the app. specifically, popovers have higher z-index than modal bc we have popovers inside modals 😓

rohoswagger: Looked into the z-index route — it's not safe here, so I'm keeping the deferred mount.

Our z-scale (`src/app/css/z-index.css`) intentionally puts popover **above** modal:
```
--z-modal-overlay: 900
--z-modal:        1000
--z-popover:      1200
```
That ordering is what lets a dropdown/select opened *inside* a modal sit above the modal. Bumping `UserLibraryModal` above `z-popover` to win against the closing `+` popover would invert that — and this modal opens a nested `ConfirmEntityModal` (delete) and a New Folder modal, both at `z-modal` (1000), so they'd then render *behind* the library modal. So the higher-z-index fix trades one paint-over for a worse one.

The clean alternative is opening the modal on the popover's close-complete instead of a fixed delay, but opal's `Popover`/`PlusMenuButton` don't expose an `onExitComplete` today, so that's a shared-component change I'd rather not fold into this PR. Keeping the 200ms defer (it just waits out the popover's exit animation; self-contained). Can file a follow-up for the callback if we want to kill the magic number later.

web/src/app/craft/v1/skills/manage/page.tsx
rohoswagger: Yep — it's usable and already wired up. `web/src/app/craft/v1/skills/manage/layout.tsx` re-exports `CraftManageLayout` (the 3b811328ea layout), so skills/manage is gated server-side (full-admin only) by the same shared layout as apps/manage. (My earlier "overkill" note was before I created/applied it.)

wenxi-onyx (approved):

web/src/app/craft/v1/apps/admin/page.tsx
wenxi-onyx: nit: I would just nuke this. There's no need to keep this backwards compat around

PR #11677, opened 2026-06-03
fix(craft): grant scheduled-tasks worker sandbox push access

Weves:
## Description

The scheduled-tasks Celery worker provisions and drives sandboxes headlessly (`run_scheduled_task` → `ensure_sandbox_running` → `provision()`), so it needs the same sandbox access as the api_server. Two pieces were missing in the Helm chart:

1. **Push private key secret** — `provision()` calls `_create_sandbox_pod` → `_get_push_key_pair()`, which raises `RuntimeError` if `ONYX_SANDBOX_PUSH_PRIVATE_KEY` is unset. That secret (`sandboxPushSecret`, `allPods: false`) was only injected into api-server pods via `onyx.envSecretsRestricted`; the worker used `onyx.envSecrets`, which skips `allPods: false` entries.
2. **NetworkPolicy** — `network-policy-sandbox-push` only allowed `app: api-server` pods to reach sandbox pods on ports 8731 (push daemon), 4096 (opencode-serve), and the Next.js dev range. Even with the secret, the worker would be blocked.

Changes:
- `celery-worker-scheduled-tasks.yaml`: add `onyx.envSecretsRestricted` so the worker receives `ONYX_SANDBOX_PUSH_PRIVATE_KEY` (the only `allPods: false` secret today).
- `network-policy-sandbox-push.yaml`: add `app: celery-worker-scheduled-tasks` as an allowed ingress source in the release namespace.
- `values.yaml`: update the `sandboxPushSecret` comment — no longer api-server-only.

## How Has This Been Tested?

`helm template` renders both changed templates correctly:
- worker now mounts `ONYX_SANDBOX_PUSH_PRIVATE_KEY` from `onyx-sandbox-push-secret`
- NetworkPolicy now lists both `app: api-server` and `app: celery-worker-scheduled-tasks`

## Additional Options

### DIFF
--- a/deployment/helm/charts/onyx/Chart.yaml
+++ b/deployment/helm/charts/onyx/Chart.yaml
@@ -5,7 +5,7 @@ home: https://www.onyx.app/
 sources:
   - "https://github.com/onyx-dot-app/onyx"
 type: application
-version: 0.5.16
+version: 0.5.17
 appVersion: latest
 annotations:
   category: Productivity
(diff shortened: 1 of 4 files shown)


justin-tahara (commented):
Let's bump Chart.yaml version

justin-tahara (approved):

PR #11678, opened 2026-06-03
feat(ods): add `celery` command to run all workers locally

Weves:
## Description

Adds a new `ods celery` command that launches the Onyx celery workers in a single process, mirroring the **"Run All Onyx Services"** debug compound in `.vscode/launch.json`.

- **No args** → runs the full dev worker set: `primary, light, heavy, docfetching, docprocessing, user_file_processing, scheduled_tasks, beat` (the same set, and same queues/concurrency/prefetch, as the launch compound). `monitoring` is excluded from the default to match the compound.
- **Named args** → run only a subset, e.g. `ods celery primary beat`.
- **`--all`** → additionally include the `monitoring` worker.
- **Hot-reload by default** via `backend/scripts/dev_celery_reload.py` (watches `backend/onyx` + `backend/ee`); `--no-reload` runs celery directly.
- All workers stream to one terminal with a per-worker colored prefix, and a single `Ctrl-C` tears down every worker's process group cleanly.
- Reuses `backend`'s env handling — loads `.vscode/.env`, EE-on with license enforcement off, shell vars take precedence. Flags: `--no-ee`, `--no-reload`, `--all`, `--loglevel`.

This completes the host-side full-stack dev loop alongside `ods compose dev --infra`, `ods backend api`, `ods backend model_server`, and `ods web dev` — no devcontainer or VS Code launch compound required.

Files:
- `tools/ods/cmd/celery.go` — new command (worker table kept in sync with launch.json, arg builder, multi-process runner with prefixed output + signal handling)
- `tools/ods/cmd/root.go` — register `NewCeleryCommand()`
- `tools/ods/README.md` — new `celery` section

## How Has This Been Tested?

- `go build ./...`, `go vet`, `gofmt`, and `golangci-lint` (via pre-commit) all clean.
- Verified `--help` output and the unknown-worker error path.
- Live smoke test against running Redis: started `beat` — env loaded from `.vscode/.env`, EE enabled, Redis readiness probe succeeded, scheduler dispatching due tasks, per-worker prefix rendered, and `Ctrl-C` confirmed-stopped the process group.

## Additional Options

---
## Summary by cubic
Adds `ods background` to run all Onyx background (Celery) workers locally with hot-reload in one terminal, mirroring the "Run All Onyx Services" compound. This completes the host-side dev loop without the devcontainer or VS Code launch compound.

- **New Features**
  - Default worker set matches the launch compound (excludes `monitoring`); select workers by name or use `--all`.
  - Hot-reload via `backend/scripts/dev_celery_reload.py`; disable with `--no-reload`. Use `ods background` (renamed from `ods celery`) to match the `background` compose service.
  - Unified, color-prefixed logs; graceful shutdown on Ctrl-C.
  - Env handling: loads `.vscode/.env`, EE enabled by default. Flags: `--no-ee`, `--no-reload`, `--all`, `--loglevel`. Docs in `tools/ods/README.md`.

- **Bug Fixes**
  - Reject `--all` when combined with named workers.
  - Fail fast if any worker fails to start; tear down started workers and exit non-zero.
  - Flush buffered partial lines on worker exit.
  - Second Ctrl-C escalates to SIGKILL for stuck process groups.

<sup>Written for commit a13cdeaf2a346532adab0ff1e70f851e22f07210. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11678?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/tools/ods/README.md
+++ b/tools/ods/README.md
@@ -220,6 +220,57 @@ ods backend model_server
 ods backend model_server --port 9001
 ```
 
+### `celery` - Run Celery Workers
+
+Run Onyx celery workers with environment loaded from `.vscode/.env` (same env
+handling and EE defaults as `backend`). With no arguments, starts every worker
+in the "Run All Onyx Services" debug compound; pass worker names to run a
+subset, or `--all` to also include the `monitoring` worker.
+
+Each worker runs through `backend/scripts/dev_celery_reload.py` for hot-reload
+on changes under `backend/onyx` and `backend/ee`. All workers stream to one
+terminal with a per-worker prefix, and a single `Ctrl-C` stops them all.
+
+```shell
+ods celery [worker...]
+```
+
+**Available workers:**
+
+`primary`, `light`, `heavy`, `docfetching`, `docprocessing`,
+`user_file_processing`, `scheduled_tasks`, `beat`, `monitoring`
+
+(`monitoring` is excluded from the default set, matching the launch compound.)
+
+**Flags:**
+
+| Flag | Default | Description |
+|------|---------|-------------|
+| `--all` | `false` | Run every worker, including `monitoring` |
+| `--no-reload` | `false` | Disable hot-reload (run celery directly) |
+| `--no-ee` | `false` | Disable Enterprise Edition features (enabled by default) |
+| `--loglevel` | `INFO` | Celery log level |
+
+**Examples:**
+
+```shell
+# Start the full dev worker set
+ods celery
+
+# Start only specific workers
+ods celery primary beat
+
+# Include the monitoring worker
+ods celery --all
+
+# Run without hot-reload
+ods celery docfetching docprocessing --no-reload
+```
+
+This pairs with `ods compose dev --infra`, `ods backend api`,
+`ods backend model_server`, and `ods web dev` to run the full stack on the host
+with hot-reload — no devcontainer or VS Code launch compound required.
+
 ### `web` - Run Frontend Scripts
 
 Run bun scripts from `web/package.json` without manually changing directories.
--- a/tools/ods/README.md
+++ b/tools/ods/README.md
@@ -268,10 +268,6 @@ ods background --all
 ods background docfetching docprocessing --no-reload
 ```
 
-This pairs with `ods compose dev --infra`, `ods backend api`,
-`ods backend model_server`, and `ods web dev` to run the full stack on the host
-with hot-reload — no devcontainer or VS Code launch compound required.
-
 ### `web` - Run Frontend Scripts
 
 Run bun scripts from `web/package.json` without manually changing directories.
--- a/tools/ods/README.md
+++ b/tools/ods/README.md
@@ -220,19 +220,20 @@ ods backend model_server
 ods backend model_server --port 9001
 ```
 
-### `celery` - Run Celery Workers
+### `background` - Run Background (Celery) Workers
 
-Run Onyx celery workers with environment loaded from `.vscode/.env` (same env
-handling and EE defaults as `backend`). With no arguments, starts every worker
-in the "Run All Onyx Services" debug compound; pass worker names to run a
-subset, or `--all` to also include the `monitoring` worker.
+Run Onyx background celery workers with environment loaded from `.vscode/.env`
+(same env handling and EE defaults as `backend`). With no arguments, starts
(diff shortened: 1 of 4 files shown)


acaprau:
nit why not name this command `ods backend celery`?

tools/ods/cmd/celery.go
```diff
@@ -0,0 +1,331 @@
...
+
+// CeleryOptions holds options for the celery command.
+type CeleryOptions struct {
+	NoEE     bool
+	NoReload bool
+	LogLevel string
+	All      bool
+}
+
+func NewCeleryCommand() *cobra.Command {
```
jmelahman: thoughts on `ods background` (can even alias to `ods bg`), just so the interface isn't tied to `celery` which is an implementation detail (and one I've heard talks recently about replacing)

tools/ods/cmd/background.go
```diff
@@ -0,0 +1,331 @@
...
+		Use:   "celery [worker...]",
+		Short: "Run Onyx celery workers (with hot-reload)",
+		Long: `Run Onyx celery workers with environment from .vscode/.env.
+
+With no arguments, starts every worker in the "Run All Onyx Services" debug
+compound (primary, light, heavy, docfetching, docprocessing,
+user_file_processing, scheduled_tasks, beat). Pass worker names to run only a
+subset, or --all to additionally include the monitoring worker.
+
+Each worker runs through scripts/dev_celery_reload.py for hot-reload on changes
```
jmelahman, line 70: Is this file load-bearing? Can we use just a native file watcher instead? Probably, https://github.com/fsnotify/fsnotify. IMO, if you wanted to use this (we should try not to if possible), you should at least embed it, https://pkg.go.dev/embed

tools/ods/README.md
```diff
@@ -220,6 +220,57 @@ ods backend model_server
...
+ods celery --all
+
+# Run without hot-reload
+ods celery docfetching docprocessing --no-reload
+```
+
+This pairs with `ods compose dev --infra`, `ods backend api`,
+`ods backend model_server`, and `ods web dev` to run the full stack on the host
+with hot-reload — no devcontainer or VS Code launch compound required.
+
```
jmelahman:
```suggestion
```
Also, no promoting not using the devcontainer. I still think everyone should and all of this should be compatible with it anyways

tools/ods/cmd/celery.go
Weves: yea, that makes sense to me. Will change

tools/ods/cmd/background.go
Weves: this is load-bearing. Comment from the top of `scripts/dev_celery_reload.py`:

```
Run a celery worker with watchfiles hot-reload, preserving VSCode breakpoints.

LOAD-BEARING: referenced by every "Celery <name>" configuration in
.vscode/launch.json (both local and k8s variants). Deleting or renaming this
file will break the vscode debugger for celery. See also CONTRIBUTING.md
("VSCode Debugger") and docs/dev/local-kubernetes.md.
```

Weves: we can switch, but I think that's a discussion for a separate PR. @wenxi-onyx added this a couple weeks ago and it's the default in launch.json afaik

tools/ods/README.md
Weves: sure, will remove

tools/ods/cmd/background.go
jmelahman: We don't necessarily need to remove it from the `launch.json` path, but I don't think `ods` should go through it just because `launch.json` does. I think that is in-scope for this change

Weves: we should probably only have one way to spin up the celery tasks locally. And this afaik is our standard way, so it feels reasonable to use it

jmelahman: I don't think the stub is that doing much that it's not worth implementing correctly here IMO

PR #11680, opened 2026-06-03
fix(craft): fast-fail bus creation for terminated sandboxes across replicas

rohoswagger:
## Description

`_ServeMixin._terminated_sandboxes` is an in-memory per-process tombstone that blocks a late `subscribe` from racing a `terminate` and spinning up a fresh `PodEventBus` against a backend that's being deleted. Across multiple `api_server` replicas the tombstone is local to the pod that ran `terminate`, so a late subscribe on a **peer** replica has no entry and builds a doomed bus against the already-deleted Service — wasting a reader thread + httpx connection through its full reconnect budget (~30s of backoff) before self-closing, plus log noise and a delayed terminal signal to the subscriber.

The fix consults the authoritative signal that already exists: the DB `Sandbox.status` column (strongly consistent, indexed via `ix_sandbox_status`). `_get_or_create_event_bus` now checks the local tombstone first (no-DB fast path), then consults `Sandbox.status` and refuses to build a bus for a **terminal** sandbox (`TERMINATED` / `FAILED`) — so the refusal fires uniformly on every replica. No Redis: the authority is the DB, not a cache.

Key guard: `PROVISIONING` / not-yet-ready sandboxes are **not** terminal and are deliberately allowed through to the readiness wait — the check detects terminal states specifically, not absence-of-readiness. A missing row (`status is None`) is left to the existing connection-info handling.

### Changes
- `db/sandbox.py`: add `get_sandbox_status(db_session, sandbox_id)` — reads only the indexed `status` column.
- `serve_transport.py`: add `_assert_sandbox_not_terminal`, invoked in `_get_or_create_event_bus` after the local tombstone check.

## How Has This Been Tested?

Added an External Dependency Unit test (`test_terminated_sandbox_fast_fail.py`) that drives bus creation on a **fresh** `StubSandboxManager` (empty local tombstone, standing in for a peer replica):
- `TERMINATED` and `FAILED` rows → `_get_or_create_event_bus` raises (refuses to build a bus).
- `RUNNING` row → builds a `PodEventBus`.
- `PROVISIONING` row → still builds a bus (not fast-failed).

Locally verified: `ruff format`, `ruff check`, and `ty check` pass on all changed files; the suite collects cleanly. The DB-backed assertions run in CI (no host Postgres in this worktree).

## Additional Options

---
## Summary by cubic
Fast-fail event bus creation across replicas by checking DB `Sandbox.status` in `_get_or_create_event_bus`. Reject `TERMINATED`/`FAILED`/`SLEEPING` so we don’t build buses for sandboxes with no live backend.

- **Bug Fixes**
  - Inline no-live-backend guard in `_get_or_create_event_bus` using `get_sandbox_by_id`; reject `TERMINATED`/`FAILED`/`SLEEPING`; keep `_terminated_sandboxes` as a no-DB fast path and hold `_event_buses_lock` across cache/status/insert. Preserve behavior for `RUNNING`/`PROVISIONING`. Tests cover refusal and pass-through.

<sup>Written for commit 81d9adae126f20180e1b4ccad139257b424264ac. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11680?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/server/features/build/db/sandbox.py
+++ b/backend/onyx/server/features/build/db/sandbox.py
@@ -120,13 +120,6 @@ def get_sandbox_by_id(db_session: Session, sandbox_id: UUID) -> Sandbox | None:
     return db_session.execute(stmt).scalar_one_or_none()
 
 
-def get_sandbox_status(db_session: Session, sandbox_id: UUID) -> SandboxStatus | None:
-    """Return the sandbox's status, or None if no such row. Reads only the
-    indexed ``status`` column (``ix_sandbox_status``)."""
-    stmt = select(Sandbox.status).where(Sandbox.id == sandbox_id)
-    return db_session.execute(stmt).scalar_one_or_none()
-
-
 def update_sandbox_status__no_commit(
     db_session: Session,
     sandbox_id: UUID,
--- a/backend/onyx/server/features/build/db/sandbox.py
+++ b/backend/onyx/server/features/build/db/sandbox.py
@@ -121,12 +121,8 @@ def get_sandbox_by_id(db_session: Session, sandbox_id: UUID) -> Sandbox | None:
 
 
 def get_sandbox_status(db_session: Session, sandbox_id: UUID) -> SandboxStatus | None:
-    """Return the sandbox's current status, or None if no such row exists.
-
-    Reads only the indexed ``status`` column (``ix_sandbox_status``). Used by
-    the serve transport to fast-fail event-bus creation for terminal sandboxes
-    uniformly across all api_server replicas.
-    """
+    """Return the sandbox's status, or None if no such row. Reads only the
+    indexed ``status`` column (``ix_sandbox_status``)."""
     stmt = select(Sandbox.status).where(Sandbox.id == sandbox_id)
     return db_session.execute(stmt).scalar_one_or_none()
 
--- a/backend/onyx/server/features/build/db/sandbox.py
+++ b/backend/onyx/server/features/build/db/sandbox.py
@@ -120,6 +120,17 @@ def get_sandbox_by_id(db_session: Session, sandbox_id: UUID) -> Sandbox | None:
     return db_session.execute(stmt).scalar_one_or_none()
 
 
+def get_sandbox_status(db_session: Session, sandbox_id: UUID) -> SandboxStatus | None:
+    """Return the sandbox's current status, or None if no such row exists.
+
+    Reads only the indexed ``status`` column (``ix_sandbox_status``). Used by
+    the serve transport to fast-fail event-bus creation for terminal sandboxes
+    uniformly across all api_server replicas.
+    """
+    stmt = select(Sandbox.status).where(Sandbox.id == sandbox_id)
+    return db_session.execute(stmt).scalar_one_or_none()
+
+
 def update_sandbox_status__no_commit(
     db_session: Session,
     sandbox_id: UUID,
--- a/backend/onyx/server/features/build/sandbox/serve_transport.py
+++ b/backend/onyx/server/features/build/sandbox/serve_transport.py
@@ -338,12 +338,6 @@ class _ServeMixin:
                     f"Sandbox {sandbox_id} has been terminated; refusing to "
                     "create a new event bus against its (deleted) backend"
                 )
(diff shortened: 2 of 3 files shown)


backend/onyx/server/features/build/sandbox/serve_transport.py
```diff
@@ -304,6 +310,27 @@ def _get_or_create_event_bus(self, sandbox_id: UUID, directory: str) -> PodEvent
...
+        on to the readiness wait. A missing row (status ``None``) is left to
+        the existing connection-info handling rather than treated as terminal.
+        """
+        with get_session_with_current_tenant() as db_session:
+            status = get_sandbox_status(db_session, sandbox_id)
+        if status is not None and status.is_terminal():
+            raise RuntimeError(
+                f"Sandbox {sandbox_id} is {status.value} (terminal per DB); "
+                "refusing to create a new event bus against its (deleted) backend"
+            )
```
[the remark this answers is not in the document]
rohoswagger: Fixed in d7e23ce4ec.

Good catch — `SLEEPING` (pod torn down, snapshot saved to S3) has no live backend pod, so a bus built against it burns the full ~30s reconnect budget exactly like a terminal sandbox.

Rather than widen the shared `is_terminal()` (SLEEPING is *restorable*, not terminal, and that helper is used with permanent-termination semantics elsewhere), I renamed `_assert_sandbox_not_terminal` → `_assert_sandbox_backend_live` and the guard now rejects `TERMINATED` / `FAILED` / `SLEEPING` via `status.is_terminal() or status.is_sleeping()`. `RUNNING` / `PROVISIONING` keep a (possibly still-coming-up) pod and continue to flow into the readiness wait. Added `test_sleeping_sandbox_refused_on_peer_replica` to lock this in.

backend/onyx/server/features/build/db/sandbox.py
```diff
@@ -120,6 +120,13 @@ def get_sandbox_by_id(db_session: Session, sandbox_id: UUID) -> Sandbox | None:
     return db_session.execute(stmt).scalar_one_or_none()
 
 
+def get_sandbox_status(db_session: Session, sandbox_id: UUID) -> SandboxStatus | None:
```
wenxi-onyx: I think you should just use `get_sandbox_by_id` and then grab the `.status` from the result

backend/onyx/server/features/build/sandbox/serve_transport.py
```diff
@@ -335,6 +338,9 @@ def _get_or_create_event_bus(self, sandbox_id: UUID, directory: str) -> PodEvent
                     f"Sandbox {sandbox_id} has been terminated; refusing to "
                     "create a new event bus against its (deleted) backend"
                 )
+            # Tombstone is local to the terminating replica; the DB status
+            # makes the refusal fire on every replica.
+            self._assert_sandbox_backend_live(sandbox_id)
```
wenxi-onyx: nit: would personally in-line since the assert func is only called once and has relatively simple logic.

wenxi-onyx (approved):

backend/onyx/server/features/build/db/sandbox.py
rohoswagger: Done in eba85a2e46 — removed `get_sandbox_status` and the inlined guard now uses `get_sandbox_by_id(db_session, sandbox_id)`, reading `.status` off the result. `db/sandbox.py` is no longer touched by this PR.

backend/onyx/server/features/build/sandbox/serve_transport.py
rohoswagger: Done in eba85a2e46 — inlined the check into `_get_or_create_event_bus` and dropped the standalone `_assert_sandbox_backend_live` method. Behavior is unchanged: it still refuses bus creation when the DB reports the pod gone (`TERMINATED` / `FAILED` / `SLEEPING`).

PR #11681, opened 2026-06-03
fix(craft): make prompt_slot a cross-replica distributed lock

rohoswagger:
## Description

`_ServeMixin.prompt_slot()` serialized turns per build session using an **in-process** `threading.Lock`, which only coordinates within one process. With multiple `api_server` replicas, two prompts for the same build session can land on different pods, each acquires its own local lock, and both POST `prompt_async` to opencode-serve. Since `prompt_async` is fire-and-forget and not concurrent-safe, the second POST is silently dropped, the second subscriber latches onto the first turn's terminator, and a phantom `user_message` is persisted with **no assistant reply** — corrupted, persisted conversation state. (`interrupt_signal.py` already exists because turns for one session are driven across replicas.)

This layers a `CacheBackend.lock()` — the existing distributed-lock abstraction, mirroring `interrupt_signal`'s fence — **under** the in-process `threading.Lock`:

- The local lock stays as a cheap same-pod fast path (avoids redundant cache round-trips and same-pod thundering); the **cache lock is the cross-pod correctness boundary**.
- Auto-expiry (`timeout`) = `SANDBOX_TURN_TIMEOUT_SECONDS` so a crashed holder can't pin the slot forever (mirrors `FENCE_TTL`); `blocking_timeout` = `PROMPT_SLOT_ACQUIRE_TIMEOUT_SECONDS`, budgeted against the time the local acquire already consumed.
- **Fail open** on `CACHE_TRANSIENT_ERRORS`: a cache outage must not brick every turn — the local guard still serializes same-pod and the cross-pod race is rare (same posture as the interrupt fence).
- Key stays `(sandbox_id, build_session_id)`, never `opencode_session_id` (which can rotate mid-turn via `on_opencode_session_resolved`).

Callers (`streaming.py`, `scheduled_tasks/executor.py`) are untouched — the `bool` + context-manager contract is unchanged.

## How Has This Been Tested?

- **Unit** (`test_prompt_slot.py`, hermetic with an in-memory cache fake): existing same-pod invariants still hold, plus a new fail-open-on-cache-error case. 7/7 pass.
- **External Dependency Unit** (`test_prompt_slot_distributed.py`, requires real Redis): two manager instances sharing one cache (simulating two replicas) — the second acquire is refused while the first holds the slot, and a third succeeds once the first releases. This proves the cross-pod fix an in-process test cannot. Runs in CI (real Redis); not runnable in the local k3d/telepresence env.

## Additional Options

---
## Summary by cubic
Make `prompt_slot` a cross-replica distributed lock so only one turn per build session runs across pods. Prevents double `prompt_async` posts that caused phantom user messages and missing assistant replies.

- **Bug Fixes**
  - Use a distributed cache lock with TTL `SANDBOX_TURN_TIMEOUT_SECONDS`; acquire waits up to `PROMPT_SLOT_ACQUIRE_TIMEOUT_SECONDS`; fail-open on `CACHE_TRANSIENT_ERRORS`; warn and rely on TTL if release fails. Key stays `(sandbox_id, build_session_id)`; callers unchanged.
  - Tests moved to the external-dependency suite using Redis: verifies cross-replica refusal/admission, distinct sessions don’t block, release on exception, and fail-open when the cache is unavailable.

- **Refactors**
  - Removed the in-process `threading.Lock` layer and its terminate-time cleanup; inlined the slot key and trimmed docstrings.

<sup>Written for commit 1425319a251475b63a35f92ee65c57cda0ebf53a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11681?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/tests/external_dependency_unit/craft/test_prompt_slot_distributed.py
```diff
@@ -0,0 +1,77 @@
...
+    monkeypatch.setattr(factory, "CACHE_BACKEND", CacheBackendType.REDIS)
+
+
+def test_second_replica_refused_then_admitted_after_release(
+    tenant_context: None,  # noqa: ARG001
+    _fast_acquire: None,  # noqa: ARG001
+    _redis_backend: None,  # noqa: ARG001
+) -> None:
+    sandbox_id = uuid4()
+    build_session_id = uuid4()
```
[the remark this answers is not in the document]
rohoswagger: Fixed in c1bf93f.

Replaced the manual `slot_a.__enter__()` + `try/finally: slot_a.__exit__(...)` in `test_second_replica_refused_then_admitted_after_release` with a nested `with` statement. The held slot is now torn down by the context manager on every exit path — including an unexpected non-`CACHE_TRANSIENT_ERRORS` exception raised during acquisition — so the local `threading.Lock` can no longer leak for the process lifetime.

Coverage is unchanged: replica A holds the slot in the outer `with` (asserts `True`), replica B is refused in the inner `with` (asserts `False`) while A holds it, A is released on outer-`with` exit, and the subsequent acquire then succeeds (asserts `True`).

backend/onyx/server/features/build/sandbox/serve_transport.py
```diff
@@ -52,6 +55,14 @@
 # (abort → session.idle → terminator) as the next queued message fires.
 PROMPT_SLOT_ACQUIRE_TIMEOUT_SECONDS = 10.0
 
+# Distributed prompt-slot lock key prefix. Mirrors interrupt_signal._fence_key;
+# tenant isolation is handled by the cache backend's key-prefixing.
+PROMPT_SLOT_PREFIX = "buildpromptslot"
+
+
+def _prompt_slot_key(sandbox_id: UUID, build_session_id: UUID) -> str:
```
wenxi-onyx: nit: pref in-lining this. It's only used once and it's an unnecessary re-direct to this one-liner.

backend/onyx/server/features/build/sandbox/serve_transport.py
```diff
@@ -148,42 +156,58 @@ def prompt_slot(
...
-        concurrent-safe: a second POST while a turn is in flight is
-        silently dropped, the second subscriber catches the *first* turn's
-        terminator, and a phantom user_message gets persisted with no
-        assistant reply. The short wait (rather than an instant refusal) lets
-        a queued message that fires while an interrupted turn is still unwinding
-        serialize cleanly behind it instead of erroring. Key on
-        ``build_session_id`` (not ``opencode_session_id``) so the lock
-        survives id rotation from ``on_opencode_session_resolved`` and bounds
-        the dict to one entry per build session.
+        concurrent-safe, so a second in-flight POST corrupts session state
```
wenxi-onyx: clean up docstrings + comments pls 🥲

backend/tests/unit/onyx/server/features/build/sandbox/test_prompt_slot.py
```diff
@@ -1,35 +1,90 @@
...
+from redis.exceptions import RedisError
 
+from onyx.cache.interface import CacheLock
+from onyx.server.features.build.sandbox import serve_transport
 from onyx.server.features.build.sandbox.kubernetes.kubernetes_sandbox_manager import (
     KubernetesSandboxManager,
 )
 
 
+class _FakeCacheLock(CacheLock):
```
wenxi-onyx: Didn't look too hard, but i think a bunch of your PRs have had stubbed caches. Should these tests be ext-dep unit tests with a real cache? If not, double check that you're using shared utils/managers for stubs like this.

wenxi-onyx (approved):

backend/onyx/server/features/build/sandbox/serve_transport.py
rohoswagger: Inlined in f8771a1 — removed the `_prompt_slot_key` helper (and the single-use `PROMPT_SLOT_PREFIX` constant); the key is now built inline at the `get_cache_backend().lock(...)` call in `prompt_slot`.

backend/onyx/server/features/build/sandbox/serve_transport.py
rohoswagger: Done in f8771a1 — cut the `prompt_slot` docstring down to the essentials (concurrent-unsafe → corruption, the False contract, fail-open, key choice) and dropped the surrounding block comments.

backend/tests/unit/onyx/server/features/build/sandbox/test_prompt_slot.py
rohoswagger: Good call — moved them to the external-dependency-unit suite against a real cache and deleted the hand-rolled stub (1425319).

`test_prompt_slot.py` (the unit test with `_FakeCacheBackend`/`_FakeCacheLock`) is removed. Its cases — distinct build sessions not blocking, release-on-exception, and fail-open on cache error — now live in `test_prompt_slot_distributed.py`, which forces the Redis backend so the lock under test is the real distributed one.

On the shared-stub point: there isn't a shared cache-stub util today (other suites hand-roll their own `_MemoryCacheBackend`, e.g. `test_stop_signal_checker.py`), and unit tests can't assume external services per CLAUDE.md — so using a real cache in the ext-dep suite was the cleaner of your two options rather than adding yet another stub.

PR #11693, opened 2026-06-03
fix(craft): Clear _synced on clean watch EOF for docker + k8s lookups

acaprau:
## Description

Follow-up to #11507 review: Both `DockerEventsLookup` and `K8sInformerLookup` only cleared `_synced` in the exception arms of `_run`. Clean returns from `_watch_loop` left `_synced=True` during the reconnect backoff window, so `/healthz` reported 200 while the proxy was not actively watching events. A sandbox starting in that 1-30s window would 403 with `gate.unidentified_sandbox` because the cache never saw its start event.

Clean returns are NOT a rare path: `CancellableStream` converts every daemon-side close (EOF, restart, network hiccup) to `StopIteration`, and the K8s API server closes the watch every `_WATCH_TIMEOUT_SECONDS` (300s). So the exception arms only fire on errors *before* the iterator starts, not on stream termination — the common case was running through the gap.

**Fix:** move `_synced.clear()` into a `finally` block so it fires on both clean return and exception. Idempotent. Same shape in both files.

## How Has This Been Tested?

- [x] `pytest backend/tests/unit/sandbox_proxy/test_identity_docker.py backend/tests/unit/sandbox_proxy/test_informer.py` -- 34 passed.
- [x] New regression test in each file (`test_synced_clears_after_watch_loop_returns_cleanly`): drives `_run` with an empty events/watch iterator (simulating a clean daemon-side close), sets stop after one iteration, asserts `_initial_sync_done` is set (proves the iteration ran), `_synced` is False (proves
the finally fired), and the iteration ran exactly once.
- [x] `pre-commit run` on touched files -- clean.

## Additional Options

---
## Summary by cubic
Fixes stale sync state in Docker and K8s identity lookups. `_synced` now clears after every watch iteration so `/healthz` is accurate and sandboxes don’t 403 during reconnect backoff.

- **Bug Fixes**
  - Move `_synced.clear()` into `finally` in `DockerEventsLookup` and `K8sInformerLookup` so it clears on clean EOF and exceptions.
  - Add regression tests for clean-return paths in Docker and K8s.
  - Tidy log and error messages for consistency.

<sup>Written for commit b5295ca9d021e76aabe945bc32422b6dcc458021. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11693?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


wenxi-onyx (approved):
lgtm, clean up claude comment vomit pls 🥲

PR #11700, opened 2026-06-03
fix(chat): make history compression resilient to message-tree forks

justin-tahara:
## Description

Chat-history compression breaks permanently for any session whose message tree forks (regenerate, edit a message, retry, or two messages sent concurrently — including frontend retries after a dropped stream).

**Root cause:** `find_summary_for_branch` matched summaries on `parent_message_id` membership in the mainline. `latest_child_message_id` is rewritten whenever a sibling is created, so any fork after a summary's parent orphans every existing summary. From then on the session:
- re-summarizes its **entire history** at the end of **every turn** (observed: 80–150s LLM call summarizing 336 messages, repeatedly, with the cutoff never advancing — 16 dead summary rows in a single session), and
- never truncates its prompt (observed: ~220k-token prompts on every agent-loop iteration).

These runaway turns can occupy threadpool workers/DB sessions for minutes and, under a CPU limit, throttle the api-server hard enough to fail liveness probes — the resulting kill triggers a frontend retry, which forks the tree again and feeds the loop.

**Changes** (all in CE `backend/onyx/chat/`):
1. **Fork-proof summary matching** — match on cutoff (`last_summarized_message_id`) membership in the branch. Every message has exactly one parent, so the root→cutoff path is unique: any branch containing the cutoff shares the entire summarized prefix. Highest applicable cutoff wins. Trade-off (documented in COMPRESSION.md): the summarizer sees post-cutoff messages "for context only", so a reused summary may carry faint context from a sibling branch — accepted versus losing compression entirely.
2. **Per-session compression lock + min-size guard** — concurrent turns no longer run duplicate summarization calls; tails under `MIN_TOKENS_TO_COMPRESS` skip the LLM round-trip.
3. **Prompt-effective trigger** — end-of-turn trigger now counts what the next prompt actually contains (summary + post-cutoff messages) instead of the raw chain total, which kept re-triggering compression every turn once a session ever crossed the threshold.
4. **Hard-trim fallback** — if no summary applies at prompt-build time and history exceeds the trigger threshold (compression in flight, failed, or not yet run), trim to a recent suffix so prompt size stays bounded; dropped messages' file metadata feeds the existing forgotten-file mechanism.

## How Has This Been Tested?

- New external-dependency-unit suite `backend/tests/external_dependency_unit/chat/test_compression_fork_resilience.py` (real Postgres + cache backend): summary survives a regeneration fork; cutoff advances instead of full recompression; trivial tails skip the LLM call; concurrent compression no-ops under the lock; effective-token trigger doesn't re-fire post-compression. 5/5 passing.
- Extended `backend/tests/unit/onyx/chat/test_compression.py` with fork-resilience matching, highest-cutoff selection, effective-token math, and hard-trim behavior (26/26 passing, existing tests unchanged).
- `pre-commit` (ty, ruff, ruff format) clean on all touched files.

## Additional Options

---
## Summary by cubic
Fixes chat-history compression so it works across message-tree forks and stops re-summarizing full histories, keeping prompts bounded and reducing latency. This prevents runaway turns that could throttle the API server.

- **Bug Fixes**
  - Match summaries by cutoff membership in the branch (highest applicable cutoff wins) so regenerate/edit/retry/concurrent sends don’t orphan summaries.
  - Trigger compression on prompt-effective tokens (summary + post-cutoff), not the raw chain total.
  - Hard-trim history when no summary applies to keep the prompt within budget; dropped files feed the forgotten-file flow.

- **Performance**
  - Per-session cache lock dedups concurrent summarization; fails open on cache outages and checks lock ownership before release.
  - Skip compression when the older portion is below a minimum token threshold.

<sup>Written for commit 08c7e7665b17c7abdf0dc9b091354aac1ec7b9c4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11700?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/chat/COMPRESSION.md
+++ b/backend/onyx/chat/COMPRESSION.md
@@ -9,8 +9,19 @@ Summaries are stored as `ChatMessage` records with two key fields:
 - `parent_message_id` → last message when compression triggered (places summary in the tree)
 - `last_summarized_message_id` → pointer to an older message up the chain (the cutoff). Messages after this are kept verbatim.
 
-**Why store summary as a separate message?** If we embedded the summary in the `last_summarized_message_id` message itself, that message would contain context from messages that came after it—context that doesn't exist in other branches. By creating the summary as a new message attached to the branch tip, it only applies to the specific branch where compression occurred. It's only back-pointed to by the
-branch which it applies to. All of this is necessary because we keep the last few messages verbatim and also to support branching logic.
+**Why store summary as a separate message?** If we embedded the summary in the `last_summarized_message_id` message itself, that message would contain context from messages that came after it—context that doesn't exist in other branches. By creating the summary as a new message hanging off the tree (never on the mainline — it doesn't update `latest_child_message_id`), the original history remains intact and branching keeps working.
+
+**Why match summaries by cutoff, not parent?** A summary applies to a branch when its
+cutoff (`last_summarized_message_id`) is on that branch. Every message has exactly one
+parent, so the path from the root to the cutoff is unique: any branch containing the
+cutoff shares the entire summarized prefix. Matching on the summary's
+`parent_message_id` instead (the original design) was fragile — `latest_child_message_id`
+is rewritten whenever a sibling is created, so any regenerate, edit, retry, or
+concurrent send after the summary's parent orphaned every existing summary. Affected
+sessions then re-summarized their full history at the end of every turn and their
+prompts were never truncated. Trade-off: the summarization prompt shows post-cutoff
+messages "for context only", so a summary reused across a fork may carry faint context
+from a sibling branch — accepted versus losing compression entirely.
 
 ### Progressive Summarization
 Subsequent compressions incorporate the existing summary text + new messages, preventing information loss in very long conversations.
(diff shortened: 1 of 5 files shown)


justin-tahara:
Addressed the review findings in 08c7e76:

- **P1 (lock acquisition outside exception handling)** — valid. Lock setup/acquire is now wrapped; on cache backend errors compression proceeds *without* the dedup lock (fail-open, degrades to pre-PR concurrency behavior) rather than propagating into the completion handler or silently dropping compression. Added a regression test that patches the cache backend to raise and asserts a summary is still produced.
- **P2 (release without ownership check)** — now `if lock.owned(): lock.release()` in the `finally`, so a compression that outlives the 300s lock timeout doesn't log spurious `LockNotOwnedError` warnings.
- **P2 (`UUID | None` annotation)** — tightened to `UUID`; the caller guarantees non-empty history.

PR #11732, opened 2026-06-04
Per-user token + cost usage tracking and budget enforcement

beraterkanelcelik:
## Description

Adds **per-user token + cost usage tracking and budget enforcement** to Onyx.

Today Onyx can rate-limit by summing `ChatMessage.token_count`, but there is no per-user record of *actual* model consumption (real provider token usage, cost, cache reads) and no cost-based budgets. This PR fills that gap by building on Onyx's existing generation-span tracing (`LLMFlow` / the `tracing_wrap` auto-wrap), so **every** LLM/embedding/rerank/image/voice call is captured without instrumenting individual call sites.

### How it works

- **Universal capture, off the hot path.** A new buffered `UserUsageTracingProcessor.on_span_end` records usage for every generation span. It drains on a background thread and never blocks or raises on the LLM path — if tracking fails, the user's request is unaffected. Because it consumes the existing span system (including the `UNTAGGED_INVOKE/STREAM` fallback in `tracing_wrap`), no LLM call can silently escape counting — tool-call and MCP-tool turns are counted because the tool result feeds the next LLM invoke, whose span is recorded.
- **User attribution** via a request-scoped contextvar (`CURRENT_USER_ID_CONTEXTVAR`), set by a small `set_current_user_id_dependency` FastAPI dependency reused across chat endpoints.
- **Per-user rollup, not a raw ledger.** `user_usage` holds one row per (user × model × window-day), matching Onyx's existing usage-reporting granularity and keeping the table bounded.
- **Cost resolution** (`compute_cost_cents`): admin per-model override → image pricing → litellm `cost_per_token` (the bundled `model_prices_and_context_window_backup.json`) → a configurable fallback rate. Cache-read tokens are priced at a separate cache rate (falling back to the input rate) and subtracted from the input count before pricing, so cache reads are never double-billed.
- **Budgets reuse `TokenRateLimit`.** It gains an optional `cost_budget_cents` alongside a now-nullable `token_budget`; a limit can be token-only, cost-only, or both. Group budgets keep the existing most-permissive-wins semantics. Over-budget returns a clean 429 with a structured reset time + `Retry-After`.

### Public surface

- `GET /user/usage` — self-service: per-day-by-model breakdown, window cost, budget + remaining + period, selected model price.
- `/admin/cost-overrides` — admin CRUD for per-model negotiated rates (incl. optional cache-read rate).
- Admin per-user usage view on the existing Usage Statistics page; a self-service **Usage** settings tab; cost-budget field on the token-rate-limit modal; a dedicated 429 banner.

### Config (all env; defaults preserve current behavior)

- `USER_USAGE_TRACKING_ENABLED` (default `true`)
- `USAGE_LIMIT_WINDOW_SECONDS` (default `604800` = 7d)
- `DEFAULT_LLM_INPUT_COST_PER_MTOK` / `DEFAULT_LLM_OUTPUT_COST_PER_MTOK` (default `0` → unknown models stay free unless an operator sets a rate)
- `DEFAULT_IMAGE_COST_CENTS`

### Migrations (5, all additive)

`model_cost_override` table · `cost_budget_cents` on `token_rate_limit` · `user_usage` table · `token_budget` made nullable · cache-read rate column. Single head.

### Note for review — one shared-infra change

`OnyxError` gains two optional fields: `extra` (machine-readable fields merged into the JSON body, e.g. `reset_at`) and `headers` (e.g. `Retry-After`). Both are additive and backward-compatible; existing raises are unaffected.

### Deferred to follow-up PRs (by design)

- **Context-window usage gauge** — a separate feature, intentionally not in this PR.
- **Onyx Craft / Build-Mode attribution** — deferred; the contextvar is intentionally not wired into the Build endpoints yet.

## How Tested

- **97 unit tests** covering: cost resolution (override / image / litellm / fallback / cache-rate), the enforcement gate (token-only, cost-only, both, nullable/zero-budget edge cases), `GET /user/usage`, the admin export, the recording processor, and the contextvar dependency.
- **Live end-to-end** on a local instance against a real local model:
  - Recorded input/output tokens matched the provider-reported usage exactly.
  - Cumulative cost matched override pricing to the cent (verified per-token).
  - **Tool calling:** with a custom OpenAPI tool, the tool's HTTP execution was confirmed (server logged the hit) *and* the tool-decision + result-reading LLM invokes were recorded into the user's usage — proving tool/MCP token counting is complete.
  - Over-budget produced a clean structured 429.

## Cherry-pick

- [ ] This PR does not need to be cherry-picked.

---
## Summary by cubic
Adds per-user token and cost tracking with budget enforcement across user, group, and global scopes, plus self-service usage views and admin cost overrides. Also improves enforcement accuracy and performance, includes group cost in the user budget view, skips token aggregation for cost‑only limits, and keeps the `user_usage` dedup index compatible with Postgres <15.

- **New Features**
  - Background usage recorder captures every LLM/embedding/rerank/image/voice span, prices with admin overrides (incl. cache‑read), and flushes on shutdown; never blocks requests.
  - Rollup: `user_usage` stores per‑day (user × model × flow × provider) with accurate costs; provider coalesced to '' for cross‑version unique dedup; cache reads priced separately.
  - Pricing: override → image flat rate → `litellm` price book (provider‑aware) → configurable fallback; 60s per‑tenant override cache; negative rates rejected.
  - Budgets: `TokenRateLimit` supports nullable `token_budget` and `cost_budget_cents` (token‑only, cost‑only, or both); token budgets use the “thousands” convention.
  - Enforcement: structured 429 with scope and `Retry-After`; longest reset window across token and cost gates; group cost lookups batched; token aggregation skipped when all limits are cost‑only.
  - APIs/UI
    - `GET /user/usage` with per‑day × model breakdown, window cost, budget (amount, remaining, period) including most‑permissive group cost, and selected model price.
    - `GET/PUT/DELETE /admin/cost-overrides` for per‑model rates (incl. cache‑read).
    - `GET /admin/usage/export` for tenant‑wide usage grouped by user email.
    - Usage tab (cost, daily breakdown, budget), Cost Overrides panel, Token Rate Limit modal with cost budgets and hour/day/week picker, chat usage‑limit banner with countdown; auto‑naming chat titles is budget‑gated; `TokenRateLimitDisplay.token_budget` typed as `number | null`.

- **Migration**
  - Additive migrations: `user_usage`; `model_cost_override` (+ `cache_read_cost_per_mtok`); `cost_budget_cents` on `token_rate_limit`; `token_budget` nullable. Migration lineage re‑pointed to current head (`99ecd56cb2ce`).
  - Postgres 13/14: `user_usage` stores provider as '' to support the unique dedup index across versions.
  - Config (defaults preserve behavior): `USER_USAGE_TRACKING_ENABLED` (true), `DEFAULT_LLM_INPUT_COST_PER_MTOK`, `DEFAULT_LLM_OUTPUT_COST_PER_MTOK` (clamped ≥ 0), `DEFAULT_IMAGE_COST_CENTS` (clamped ≥ 0).
  - Token budgets use the “thousands” convention in admin UI; cost budgets are cents at the API boundary. No call‑site instrumentation needed; tracing auto‑captures usage.

<sup>Written for commit 009a250f7afc47da307db831748ebf31f7a610d8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11732?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


beraterkanelcelik:
Thanks for the reviews — all four points addressed in the latest push:

- **`/rename-chat-session` budget bypass** (cubic): the auto-naming LLM call now runs the same `check_token_rate_limits(user)` gate as send-message. Manual renames (explicit name) return before the gate and stay free.
- **Contradictory token-limit units** (cubic): dropped the `(tokens)` suffix; the header now reads `Token Limit (thousands)` consistently.
- **SWR_KEYS registry** (greptile): `/user/usage` and the admin usage-export key now go through `SWR_KEYS.userUsage(days)` / `SWR_KEYS.adminUsageExport(...)`.
- **Dead `_is_rate_limited`** (greptile): removed from production; the boolean wrapper now lives in the test that used it, calling the real `_first_triggered_limit`.

@cubic-dev-ai review
@greptile-apps review

beraterkanelcelik:
Second-round fixes pushed (all from the latest cubic review):

- **Non-deterministic `Retry-After`/`reset_at`** (P1): when multiple limits are exceeded, enforcement now reports the **longest** window instead of the first DB-ordered one, so a client that waits it out can't immediately re-trip a still-exceeded longer limit. Renamed `_first_triggered_*` → `_worst_triggered_*` to match, + added a determinism test (order-independent).
- **Zero-priced image models** (P1): `_image_cost_cents` uses an explicit `is not None` check, so a genuinely free (0.0) model bills 0 rather than the flat fallback.
- **Group budget validation** (P1): a 0/negative token budget, or a cost that rounds to 0 cents, is treated as unset — it can no longer overwrite a real existing limit via positional matching.
- **Cost-override cache fail-open** (P2): on a transient DB load failure the last good snapshot is served (stale beats silently dropping negotiated rates); only with no prior snapshot do we fall back to no-overrides.

Also locked in the scope semantics from the earlier question: a generous group raises the group ceiling but never overrides a global or per-user cap (the three scopes are independent gates, ANDed) — covered by a new `TestEEScopeComposition`.

@cubic-dev-ai review
@greptile-apps review

beraterkanelcelik:
Round-3 fixes pushed (from the latest cubic review):

- **Negative fallback cost rates** (P2): `DEFAULT_IMAGE_COST_CENTS` and `DEFAULT_LLM_INPUT/OUTPUT_COST_PER_MTOK` are clamped to `>= 0`.
- **Negative override rates** (P2): the request model now enforces `Field(ge=0)` on all three rates (422 before any write), with a defensive non-negative guard in `upsert_override` for non-API callers. Added API + helper tests.
- **Token gate short-circuiting the cost gate** (P2): both gates are now evaluated before raising, and the 429 reports the **longest** window across token + cost — a short token window can no longer mask a longer cost reset. Unified via a shared `_raise_for_longest_window` used by the global, user, and group paths; added a cross-gate test.

All green: ruff/format/ty, 109 backend unit tests, FE tsc 0 / oxlint.

@cubic-dev-ai review
@greptile-apps review

beraterkanelcelik:
Round-4 fix pushed (cubic's remaining perf finding):

- **Group cost enforcement N+1** (P2): the group cost gate called `get_group_cost_cents_since` once per group/limit (O(groups × limits) round-trips). It now mirrors the token path — one batched `get_group_cost_cents_buckets_since` query fetches every group's cost buckets since the broadest cost window, then each limit's window is summed in Python. Tests updated to the batched path.

All green: ruff/format/ty, 109 backend unit tests, FE tsc 0 / oxlint.

@cubic-dev-ai review
@greptile-apps review

beraterkanelcelik:
Fixes pushed for the latest reviews:

**greptile — migration fails on Postgres ≤ 14** (`NULLS NOT DISTINCT` is PG15+): the `user_usage` dedup index no longer uses that clause. `provider` is now stored as `''` (not NULL) for "no provider", so a plain unique index dedups correctly on every Postgres version. Provider is coalesced to `''` at record time; cost is computed upstream from the real provider, so this only affects the ledger's dedup dimension. (greptile's two round-1 findings — SWR_KEYS and the dead `_is_rate_limited` — are already resolved.)

**cubic — group cost N+1** (previous round): the group cost gate now batches into one query (`get_group_cost_cents_buckets_since`) like the token path, instead of a query per group/limit.

All green: ruff/format/ty, single alembic head, 109 backend unit tests, FE tsc 0 / oxlint.

@cubic-dev-ai review
@greptile-apps review

beraterkanelcelik:
Both points from the latest reviews are fixed:

- **Budget display omitted group cost limits** (cubic + greptile): `/user/usage` now includes group-scoped cost limits, using the same most-permissive-group rule the gate enforces (a user capped by a group no longer sees an overstated remaining). Added `fetch_user_group_token_rate_limits` + a batched bucket query (tz-coerced for cross-DB safety) and a test.
- **Cost-only limits ran the token aggregation** (cubic): the gate now skips the token-usage query entirely when every limit is cost-only (global/user/group paths), via a small `_has_token_budget` guard. Added a test asserting the aggregation isn't called.

All green: ruff/format/ty, single alembic head, 111 backend unit tests, FE tsc 0 / oxlint.

@cubic-dev-ai review
@greptile-apps review

beraterkanelcelik:
Fixed the last item: `TokenRateLimitDisplay.token_budget` is now `number | null` to match the backend's nullable column (frontend typechecks clean). Thanks for the reviews!

@greptile-apps review

beraterkanelcelik:
Per the review feedback, I'm splitting this into smaller, reviewable, stacked PRs (schema → accounting → enforcement → frontend), and moving the `ee/` parts (per-user/group budget enforcement + the admin per-user usage report) out for the Onyx team to implement, since EE contributions can't be accepted right now.

First of the stack is up — **#11766** (schema + models). The remaining three open in order as each merges. I'll close this monolith once the stack is rolling.

beraterkanelcelik:
Superseded by the reviewable stack: #11766 (schema) → #11887 (accounting) → #11888 (enforcement) → #11889 (frontend). The ee/ enforcement parts are in a written spec. Closing this in favor of the split.

PR #11774, opened 2026-06-05
fix(craft): gate external-app domains by credential, not stored policy rows

wenxi-onyx:
## Description

A connected external app now gates its **whole upstream domain by credential availability**, rather than requiring a per-action policy row. Previously, an action with no stored `ExternalAppPolicy` row was treated as off-catalog → no credential injected → the upstream returned 401.

`match_action` now resolves each matched catalog action via a shared `effective_policy` helper (stored override → catalog `default_policy`, the same formula the FE view already uses), injects when a usable credential exists, blocks a recorded `DENY` regardless of credential, and gates any uncategorized request to a connected domain under a synthesized `ASK`. A request with no usable credential still forwards bare (no prompt). `user_id` is threaded through the matcher so the credential check runs before any prompt.

Storage semantics are unchanged in this PR (still full materialization); the override-only storage model is a follow-up that stacks on this.

Fixes: https://linear.app/onyx-app/issue/ENG-4141/craft-polish-external-app-policies-should-default-to-pass-through-for
## How Has This Been Tested?

- Rewrote `test_action_matching.py`: catalog-default fallback, synthesized `ASK`, no-credential bare-forward, `DENY`-survives-no-credential, and the `user_id`-threaded matcher.
- 321 unit tests + `ty` pass locally; the external-dependency-unit suite runs in CI (needs Postgres).

## Additional Options

---
## Summary by cubic
Gates each external app’s whole upstream domain by credential availability instead of stored policy rows. Recognition uses catalog defaults; off-catalog on a connected domain synthesizes an ASK when the app is available, while missing required credentials forwards bare; DENY still blocks.

- **Bug Fixes**
  - Gate by credential presence; empty `auth_template` apps still gate; disabled apps never serve.
  - Share catalog defaults via `effective_policy` (runtime + FE) to avoid drift.
  - Split recognition and gating: `recognize_actions` + `apply_credential_gate`, with availability from `app_is_available`.
  - Off-catalog on a connected domain → synthesized ASK when available; not available → keep explicit DENY only, else forward bare.
  - Resolve `app_name` via provider; for `CUSTOM` use `skill.name`, else fall back to enum for safety during catalog drift.
  - Whole-domain synthesized ASK now shows “Perform action” as the approval label.

- **Migration**
  - Replace `ActionMatcher` with `RequestEvaluator`; new signature `evaluate(request, tenant_id, user_id)`.
  - Type rename: `ActionMatch` → `MatchedAction`, `RequestMatch` → `AllMatchedActions`.
  - Property rename: `decisive` → `governing_action`.
  - `InjectionContext` field rename: `match` → `matched_actions`.
  - Update imports and call sites accordingly.

<sup>Written for commit a1df8f64c8e7b7690d51028d007dbe92df743140. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11774?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/external_apps/matching/engine.py
+++ b/backend/onyx/external_apps/matching/engine.py
@@ -17,11 +17,15 @@ from onyx.db.models import ExternalApp
 from onyx.external_apps.matching.request import MatchContext
 from onyx.external_apps.matching.request import ProxiedRequest
 from onyx.external_apps.matching.rules import rule_matches
+from onyx.external_apps.providers.registry import effective_policy
 from onyx.external_apps.providers.registry import get_endpoint_catalog
 from onyx.external_apps.providers.registry import get_provider_for_app
 
+# action_type for a domain-matched request that hit no catalog action.
+WHOLE_DOMAIN_ACTION_TYPE = "unspecified"
 
-class ActionMatch(BaseModel):
+
+class MatchedAction(BaseModel):
     """One catalog action a request invoked, with the display strings the
     FE renders. Carried verbatim from matcher through DB JSONB to API."""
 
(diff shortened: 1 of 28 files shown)


backend/onyx/external_apps/matching/engine.py
```diff
@@ -79,18 +94,31 @@ def match_action(
...
+        if any(rule_matches(rule, context) for rule in endpoint.matches)
     ]
-    if not matched:
-        return None
-    matched.sort(key=lambda a: POLICY_SEVERITY[a.policy], reverse=True)
 
     provider = get_provider_for_app(app)
     app_name = provider.spec.app_name if provider is not None else app.app_type.value
+
+    # No usable credential: return None forwards bare (DENY still blocks).
```
wenxi-onyx: In future, can extend this to prompt the user for credentials

backend/onyx/external_apps/matching/engine.py
```diff
@@ -79,18 +94,31 @@ def match_action(
...
+    if not resolve_injection_headers(db_session, app.id, user_id):
+        matched = [a for a in matched if a.policy is EndpointPolicy.DENY]
+        if not matched:
+            return None
```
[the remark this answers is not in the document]
wenxi-onyx: addresed

backend/onyx/sandbox_proxy/request_evaluator.py
```diff
@@ -76,16 +84,21 @@ def match(self, request: http.Request, tenant_id: str) -> RequestMatch | None:
                 path=(request.path or "").split("?", 1)[0],
                 body=request.raw_content,
             )
-            matched = match_action(db, app, proxied)
-            if matched is None:
+            matched_actions = apply_credential_gate(
+                app,
+                proxied,
+                recognize_actions(db, app, proxied),
+                can_serve=can_serve_app(db, app, user_id),
```
Weves: can we short circuit here rather than calling `recognize_actions` / `apply_credential_gate`?

Weves: I think it's fine, but maybe better name instead of `can_serve`?

Weves (approved):

PR #11780, opened 2026-06-05
fix(craft): keep Craft preview URLs contained and copyable

rohoswagger:
## Description

Fixes the Craft preview URL bar so long URLs remain contained, while still keeping the full URL accessible.

- Long preview URLs are clipped/ellipsized inside the URL pill.
- Hovering the clipped URL uses the standard Opal tooltip to show the full URL.
- Clicking the URL text copies the full URL to the clipboard for real web URLs only (`http:` / `https:`).
- Internal display values such as `no-sandbox://` and `artifacts://` remain display-only and are not copyable.
- After copy, the existing open-in-new-tab icon briefly changes to a success check with a light animation.
- Copy feedback no longer uses timeout/ref orchestration; it is tied to the copied URL and clears on the check animation end.
- The open-in-new-tab tooltip remains `open in a new tab`; the check icon carries the transient copy feedback.
- The URL text uses the normal pointer cursor; no `cursor-copy` override.
- No Opal tooltip sizing/behavior override is included.
- The URL bar regression test targets explicit URL pill/wrapper test IDs instead of fragile parent traversal.

Screenshot:

Screenshot 2026-06-06 at 16 55 24

## How Has This Been Tested?

- `bunx jest src/app/craft/components/output-panel/UrlBar.test.tsx --runInBand`
- `bun run types:check`
- `bun run lint`
- `bun run format:check`
- Repo pre-commit hooks during commit: `oxfmt`, `oxlint`, and TypeScript type check passed.
- Browser layout verification against this worktree on `localhost:3101`: long URL intrinsic text width was larger than the URL pill, while the rendered URL remained clipped inside the bar with ellipsis.

Test cases covered:

- Render a long unbroken preview URL with navigation, refresh, open-in-new-tab, and share controls present.
- Assert URL text remains a block truncation target inside a shrinkable overflow-hidden flex wrapper.
- Assert hover displays the full URL through the app tooltip.
- Assert clicking an `http:` / `https:` URL copies the full URL.
- Assert copy feedback changes the open-in-new-tab button to the copied/check state.
- Assert the copied/check state clears when its animation ends.
- Assert the open-in-new-tab tooltip still reads `open in a new tab` after copy feedback.
- Assert copy feedback resets immediately when `displayUrl` changes mid-animation.
- Assert internal display values (`no-sandbox://`, `artifacts://`) are not copyable.

## Additional Options

### DIFF
--- a/web/src/app/craft/components/output-panel/UrlBar.test.tsx
+++ b/web/src/app/craft/components/output-panel/UrlBar.test.tsx
@@ -1,0 +1,36 @@
+import React from "react";
+import { render, screen } from "@tests/setup/test-utils";
+import UrlBar from "./UrlBar";
+
+describe("UrlBar", () => {
+  test("contains long preview URLs inside the URL bar", () => {
+    const longPreviewUrl =
+      "https://craft-preview.onyx.app/sessions/session-with-a-very-long-id/apps/generated-webapp/routes/deeply/nested/path/with/an-unbroken-segment-that-would-otherwise-overflow-the-url-bar?query=another-unbroken-value-that-keeps-going";
+
+    render(
+      <UrlBar
+        displayUrl={longPreviewUrl}
+        previewUrl={longPreviewUrl}
+        showNavigation
+        canGoBack
+        canGoForward
+        onBack={jest.fn()}
+        onForward={jest.fn()}
+        onRefresh={jest.fn()}
+        sessionId="session-with-a-very-long-id"
+      />
+    );
+
+    const urlText = screen.getByText(longPreviewUrl);
+    const textWrapper = urlText.parentElement;
+    const urlPill = textWrapper?.parentElement;
+
+    expect(urlText.tagName).toBe("P");
+    expect(urlText).toHaveClass("truncate");
+    expect(textWrapper).toHaveClass("min-w-0", "flex-1", "overflow-hidden");
+    expect(urlPill).toHaveClass("min-w-0", "flex-1");
+    expect(
+      screen.getByRole("button", { name: "Share webapp" })
+    ).toBeInTheDocument();
+  });
+});
--- a/web/src/app/craft/components/output-panel/UrlBar.test.tsx
+++ b/web/src/app/craft/components/output-panel/UrlBar.test.tsx
@@ -27,6 +27,7 @@ describe("UrlBar", () => {
 
     expect(urlText.tagName).toBe("P");
     expect(urlText).toHaveClass("truncate");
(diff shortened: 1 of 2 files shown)


wenxi-onyx:
I’m in favor of keeping the URL text itself click-to-copy. The part that feels off to me is mostly that the copied state is a boolean-ish value for whatever the current `displayUrl` is, and that the copy wrapper is applied to every `displayUrl` that `UrlBar` renders.

One way to keep the interaction with less state smell is to track the value that was actually copied, and only make the text copyable when this is a real preview URL:

```tsx
const copyableUrl = previewUrl && displayUrl.startsWith("http") ? displayUrl : null;
const [copiedUrl, setCopiedUrl] = React.useState<string | null>(null);
const copyResetTimeoutRef = React.useRef<ReturnType<typeof setTimeout> | null>(
  null
);
const isUrlCopied = copiedUrl === copyableUrl;

React.useEffect(() => {
  return () => {
    if (copyResetTimeoutRef.current) {
      clearTimeout(copyResetTimeoutRef.current);
    }
  };
}, []);

const handleCopyUrl = async () => {
  if (!copyableUrl) {
    return;
  }

  if (copyResetTimeoutRef.current) {
    clearTimeout(copyResetTimeoutRef.current);
  }

  await copyText(copyableUrl);
  setCopiedUrl(copyableUrl);
  copyResetTimeoutRef.current = setTimeout(() => {
    setCopiedUrl(null);
    copyResetTimeoutRef.current = null;
  }, 1500);
};
```

Then render the clickable URL text only for `copyableUrl`; otherwise render the plain truncated `Text`:

```tsx
{copyableUrl ? (
  <Tooltip tooltip={copyableUrl} side="bottom" delayDuration={200}>
    <button
      type="button"
      onClick={handleCopyUrl}
      className="block w-full min-w-0 cursor-pointer text-left focus:outline-hidden"
      aria-label={`Copy URL: ${copyableUrl}`}
    >
      <Text as="p" font="secondary-body" color="text-03" maxLines={1}>
        {displayUrl}
      </Text>
    </button>
  </Tooltip>
) : (
  <Text font="secondary-body" color="text-03" maxLines={1}>
    {displayUrl}
  </Text>
)}
```

This preserves click-the-URL-to-copy, but avoids advertising copy for status/pseudo URLs. `UrlBar` also renders values like `Loading...`, `sandbox://...`, `artifacts://`, and `no-sandbox://`, which probably should not get the same copy affordance.

I would not block this on the async race by itself, but tying feedback to the copied value makes URL transitions less misleading essentially for free.

wenxi-onyx (approved):

PR #11804, opened 2026-06-08
fix(craft): prevent stuck opencode turns

rohoswagger:
## Description

Fixes the two user-visible Craft turn failure modes that make a session appear stuck after an opencode turn fails:

- Configure Redis-backed cache locks with `thread_local=False` so prompt-slot locks can be released when Starlette/anyio resumes cleanup on a different worker thread. This addresses ENG-4144.
- Fan out unscoped opencode terminal events (`session.error`, `session.idle`, idle `session.status`) to active directory subscribers instead of dropping them.
- Treat translated opencode `Error` events as terminal in `send_message`, so the stream stops on the real failure instead of continuing with keepalives until timeout.

Related: ENG-4149, ENG-4144, ENG-4135. This PR intentionally does not add `/event` reconnect/backoff hardening for ENG-4135; that belongs with the attach-live-stream work.

## How Has This Been Tested?

- `pytest -q backend/tests/unit/onyx/cache/test_redis_backend.py backend/tests/unit/onyx/server/features/build/sandbox/test_event_bus.py backend/tests/unit/onyx/server/features/build/sandbox/test_send_message_with_bus.py`
- `ruff check backend/onyx/cache/redis_backend.py backend/onyx/server/features/build/sandbox/opencode/event_bus.py backend/onyx/server/features/build/sandbox/opencode/serve_client.py backend/tests/unit/onyx/cache/test_redis_backend.py backend/tests/unit/onyx/server/features/build/sandbox/test_event_bus.py backend/tests/unit/onyx/server/features/build/sandbox/test_send_message_with_bus.py`
- `ruff format --check backend/onyx/cache/redis_backend.py backend/onyx/server/features/build/sandbox/opencode/event_bus.py backend/onyx/server/features/build/sandbox/opencode/serve_client.py backend/tests/unit/onyx/cache/test_redis_backend.py backend/tests/unit/onyx/server/features/build/sandbox/test_event_bus.py backend/tests/unit/onyx/server/features/build/sandbox/test_send_message_with_bus.py`
- `git diff --check origin/main...HEAD`

## Additional Options

---
## Summary by cubic
Fixes stuck opencode turns by allowing cross-thread Redis lock release and treating unscoped terminal events as stream terminators. Streams now end promptly on errors or idle, and dispatch logs clearly label dropped unscoped events.

- **Bug Fixes**
  - Set Redis cache locks with thread_local=False so prompt-slot locks can be released from a different thread. Addresses ENG-4144.
  - Fan out unscoped terminal events (`session.error`, `session.idle`, idle `session.status`) to all active directory subscribers to avoid dropping terminators when the event lacks `sessionID`. Addresses ENG-4149.
  - End `send_message` when a translated opencode `Error` is received so the stream stops on the real failure instead of continuing keepalives. Related to ENG-4135.
  - Clarified dispatch naming and drop logs; log "<unscoped>" when a subscriber queue drops an unscoped terminal event.

<sup>Written for commit 26ca4301d46f6cc1f19e14558b59458048e955b9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11804?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/server/features/build/sandbox/opencode/event_bus.py
+++ b/backend/onyx/server/features/build/sandbox/opencode/event_bus.py
@@ -303,25 +303,34 @@ class PodEventBus:
                             )
 
         sid = _extract_session_id(evt)
-        if sid is None:
-            return
 
         # Deliver to sid's own subscribers AND to every ancestor's subscribers
         # so a turn subscribed only to the parent session also sees descendant
         # (subagent) events. Walk _child_to_parent up to the root, deduping so
         # no subscriber receives the event twice.
         with self._lock:
-            target_sids = [sid]
-            ancestor = self._child_to_parent.get(sid)
-            seen_sids = {sid}
-            while ancestor is not None and ancestor not in seen_sids:
-                target_sids.append(ancestor)
-                seen_sids.add(ancestor)
-                ancestor = self._child_to_parent.get(ancestor)
             queues: list[_Subscription] = []
             seen_subs: set[int] = set()
-            for target_sid in target_sids:
-                for sub in self._subscribers.get(target_sid, ()):
+            if sid is None:
+                if not _is_unscoped_terminal_event(evt):
+                    return
+                # Some opencode terminal events are published without a
+                # sessionID. The bus is directory-scoped, so active subscribers
+                # are the only consumers that can observe that turn ending.
+                subscriber_groups = self._subscribers.values()
+            else:
+                target_sids = [sid]
+                ancestor = self._child_to_parent.get(sid)
+                seen_sids = {sid}
+                while ancestor is not None and ancestor not in seen_sids:
+                    target_sids.append(ancestor)
+                    seen_sids.add(ancestor)
+                    ancestor = self._child_to_parent.get(ancestor)
+                subscriber_groups = (
+                    self._subscribers.get(target_sid, ()) for target_sid in target_sids
+                )
+            for subs in subscriber_groups:
+                for sub in subs:
                     if id(sub) not in seen_subs:
                         seen_subs.add(id(sub))
                         queues.append(sub)
@@ -367,6 +376,19 @@ def _extract_session_id(evt: dict[str, Any]) -> str | None:
         if isinstance(nested, str) and nested:
             return nested
     return None
+
+
+def _is_unscoped_terminal_event(evt: dict[str, Any]) -> bool:
+    etype = evt.get("type")
+    if etype in ("session.error", "session.idle"):
+        return True
+    if etype != "session.status":
+        return False
+    props = evt.get("properties")
+    if not isinstance(props, dict):
+        return False
+    status = props.get("status")
+    return isinstance(status, dict) and status.get("type") == "idle"
 
 
 def _parse_sse_block(block: str) -> dict[str, Any] | None:
--- a/backend/onyx/server/features/build/sandbox/opencode/event_bus.py
+++ b/backend/onyx/server/features/build/sandbox/opencode/event_bus.py
@@ -275,11 +275,11 @@ class PodEventBus:
         self._auth = new_auth
         logger.info("PodEventBus reloaded auth after 401 on %s/event", self._base_url)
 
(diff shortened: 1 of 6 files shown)


backend/onyx/server/features/build/sandbox/opencode/event_bus.py
```diff
@@ -369,6 +378,19 @@ def _extract_session_id(evt: dict[str, Any]) -> str | None:
     return None
 
 
+def _is_unscoped_terminal_event(evt: dict[str, Any]) -> bool:
+    etype = evt.get("type")
```
wenxi-onyx: nits: 
- maybe cast events into some known type, also reduces magic strings
- personal pref, I really dislike abbreviations, pref `event` and `event_type` etc.

backend/tests/unit/onyx/cache/test_redis_backend.py
```diff
@@ -0,0 +1,79 @@
...
+        thread_local: bool = True,
+    ) -> object:
+        self.lock_calls.append(
+            {"name": name, "timeout": timeout, "thread_local": thread_local}
+        )
+        self.lock_obj = _RedisLikeLock(thread_local=thread_local)
+        return self.lock_obj
+
+
+def test_redis_cache_locks_can_release_from_a_different_thread() -> None:
```
wenxi-onyx, line 56: I don't think this should matter, but note that `get_cache_backend()` can return Postgres as the cache --> hopefully no one sets cache to pg and attempt to use the full app (would be troll, I think it's meant for lite mode). 

May want to verify that the stuff your doing is compatible with pg cache backend or loudly fails if the user has pg cache set

wenxi-onyx (approved):

backend/onyx/server/features/build/sandbox/opencode/event_bus.py
rohoswagger: Fixed in `26ca4301d4` (`fix(craft): clarify unscoped event dispatch logging`).

I renamed the event-bus dispatch variables to the explicit `event`, `event_type`, `session_id`, `target_subscriptions`, and `subscription` forms, and changed the full-queue log path to report `"<unscoped>"` instead of `None` for terminal events without a `sessionID`.

I did not add a new opencode event schema in this PR because this module and the adjacent `serve_client.py` currently consume opencode's upstream event strings directly; adding a typed adapter would be a broader cleanup than this lock/terminal-event fix needs.

Verified with:
- `pytest -q backend/tests/unit/onyx/cache/test_redis_backend.py backend/tests/unit/onyx/server/features/build/sandbox/test_event_bus.py backend/tests/unit/onyx/server/features/build/sandbox/test_send_message_with_bus.py`
- `ruff check backend/onyx/server/features/build/sandbox/opencode/event_bus.py backend/tests/unit/onyx/server/features/build/sandbox/test_event_bus.py`
- `ruff format --check backend/onyx/server/features/build/sandbox/opencode/event_bus.py backend/tests/unit/onyx/server/features/build/sandbox/test_event_bus.py`

backend/tests/unit/onyx/cache/test_redis_backend.py
rohoswagger: Rationale-only after checking the implementation.

`get_cache_backend()` can return Postgres, but this PR changes only `RedisCacheBackend.lock()`. The Postgres path uses `PostgresCacheLock` with PostgreSQL advisory-lock state on the lock object/session, and it does not use redis-py lock tokens or `threading.local()`, so the Redis `thread_local=False` issue does not apply there.

The existing external dependency cache parity suite already exercises the shared `CacheBackend.lock()` API for both Redis and Postgres. The new unit test is intentionally Redis-specific because the production failure came from redis-py's thread-local token storage.

PR #11812, opened 2026-06-08
fix(tools): disambiguate colliding MCP tool names for the LLM

harigovind-s-menon:
## Problem

Multiple MCP servers commonly expose tools with the same generic name (e.g. `search`). When two such tools are attached to a single assistant/persona, the LLM receives **duplicate function declarations**:

- **Gemini** rejects the request outright: `400 Duplicate function declaration found: search`.
- **OpenAI / Anthropic / Bedrock** tolerate the duplicate, but the model cannot reliably tell the two tools apart, and dispatch may resolve to the wrong server's tool.

Repro: connect two MCP servers that each expose a `search` tool (e.g. Atlassian Rovo + Smartsheet), attach both to one persona, set the model to Gemini, send any message → 400.

## Fix

Disambiguate tool names **only on collision**, and only for **MCP tools** (built-in tools keep their stable, well-known names). A colliding MCP tool is renamed to `<server>_<tool>`, sanitized via the existing `sanitize_tool_name`, truncated to the 64-char function-name limit, and de-duplicated with a numeric suffix if needed.

The rename is applied to the tool object through a new `Tool.set_llm_name_override(...)`. Because both the tool definition sent to the LLM (`tool.tool_definition()`) and every dispatch map (`{tool.name: tool}`) key off `tool.name`, overriding it in one place keeps the declaration and the reverse-dispatch automatically in sync. The user-facing `display_name` is intentionally left unchanged, so the UI still shows the original tool name.

This also removes the previously unused `MCPTool.llm_name` property, whose namespacing intent is now realized on the actual definition/dispatch path.

### Why prefix-on-collision (not always-prefix)

Keeps tool names stable and short in the common (no-collision) case, avoids churning names users and the model are used to, and limits the behavior change to exactly the broken scenario.

## Changes

- `tools/interface.py` — add `set_llm_name_override` + backing field on the base `Tool`.
- `tools/tool_implementations/mcp/mcp_tool.py` — `name` honors the override; `tool_definition()` uses `self.name`; remove dead `llm_name`. The real server-side tool name (`self._name`) is still used for the actual MCP call and UI packets.
- `tools/tool_name.py` — add `disambiguate_tool_names()` (+ 64-char truncation helper).
- `chat/llm_loop.py` — call `disambiguate_tool_names(tools)` once before the LLM cycle.
- `tests/unit/onyx/tools/test_tool_name.py` — unit tests.

## Tests

New unit tests cover: no-collision no-op, two MCP servers sharing a tool name, override flowing into `tool_definition()`, `display_name` unchanged, built-in names staying stable while MCP renames around them, server-name sanitization, same-server duplicate uniqueness, and 64-char truncation.

```
backend/tests/unit/onyx/tools/  445 passed
ruff check / ruff format / ty check  clean on changed files
```

## Notes / scope

- Applied in the main chat loop (`run_llm_loop`), which is where the reported failure occurs. Other tool-list assembly paths (deep research / research agent) could reuse the same helper as a follow-up if desired.

Made with [Cursor](https://cursor.com)

---
## Summary by cubic
Fixes LLM tool name collisions from MCP servers by renaming only the colliding MCP tools, unblocking Gemini and improving tool dispatch accuracy. Names change only for the LLM; the UI keeps the original labels.

- **Bug Fixes**
  - Added `disambiguate_tool_names()` to prefix colliding MCP tools with their server, sanitize names, truncate to 64 chars, and add numeric suffixes if needed.
  - Introduced `Tool.set_llm_name_override()`; `MCPTool.name` and `tool_definition()` honor it so the LLM declaration and dispatch stay in sync.
  - Disambiguation runs once in the chat loop; built-in tool names and `display_name` remain unchanged. Tests cover no-op, collisions, sanitization, truncation, and uniqueness.

<sup>Written for commit 6a470c478642b28eca849cbf9ec43b5c6ea4d745. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11812?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/chat/llm_loop.py
+++ b/backend/onyx/chat/llm_loop.py
@@ -63,6 +63,7 @@ from onyx.tools.tool_implementations.python.python_tool import PythonTool
 from onyx.tools.tool_implementations.search.search_tool import SearchTool
 from onyx.tools.tool_implementations.web_search.utils import extract_url_snippet_map
 from onyx.tools.tool_implementations.web_search.web_search_tool import WebSearchTool
+from onyx.tools.tool_name import disambiguate_tool_names
 from onyx.tools.tool_runner import run_tool_calls
 from onyx.tracing.framework.create import trace
 from onyx.utils.logger import setup_logger
@@ -636,6 +637,13 @@ def run_llm_loop(
         )  # Here for lazy load LiteLLM
 
         initialize_litellm()
+
+        # Ensure tool names sent to the LLM are unique. Multiple MCP servers can
+        # expose tools with the same name (e.g. "search"), which produces
+        # duplicate function declarations that some providers (e.g. Gemini)
+        # reject. This renames only the colliding MCP tools and leaves their
+        # user-facing display names untouched.
+        disambiguate_tool_names(tools)
 
         # Normalize chat_files to a mutable list so we can extend it mid-loop
         # when a search hit carries an attached file the Python tool should
--- a/backend/onyx/tools/interface.py
+++ b/backend/onyx/tools/interface.py
@@ -18,6 +18,8 @@ class Tool(abc.ABC, Generic[TOverride]):
     def __init__(self, emitter: Emitter | None = None):
         """Initialize tool with optional emitter. Emitter can be set later via set_emitter()."""
         self._emitter = emitter
+        # When set, overrides the name sent to the LLM (see set_llm_name_override).
+        self._llm_name_override: str | None = None
 
     @property
     def emitter(self) -> Emitter:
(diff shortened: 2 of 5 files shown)


evan-onyx:
thanks for the contribution! I had a different preferred implementation, so closing this in favor of #12548

PR #11814, opened 2026-06-08
fix(scim): promote shadow EXT_PERM_USER on SCIM adoption

nmgarza5:
## Description

**Bug:** A SCIM-synced group shows its full member count (e.g. "90 Members") but the member detail and Users page list only a couple — and most synced users never count as active.

**Cause:** When SCIM provisions a user whose email already exists as an `EXT_PERM_USER` shadow (created by external permission sync, which doesn't count toward seats), `create_user` / `replace_user` / `patch_user` adopt the row and link a SCIM mapping but never promote the role. The user stays `EXT_PERM_USER`, so it's filtered out of the Users page and group member lists while still counted in the raw `user__user_group` total.

**Fix:** Promote adopted shadow users to `BASIC` / `STANDARD` and seat-check the promotion — an `EXT_PERM_USER` → `STANDARD` transition consumes a seat even when the user is already active. Real `BASIC` / `ADMIN` users are left untouched, so admins are never demoted.

## Additional Options

### DIFF
--- a/backend/ee/onyx/db/scim.py
+++ b/backend/ee/onyx/db/scim.py
@@ -46,6 +46,7 @@ from onyx.db.models import ScimUserMapping
 from onyx.db.models import User
 from onyx.db.models import User__UserGroup
 from onyx.db.models import UserGroup
+from onyx.db.models import UserRole
 from onyx.utils.logger import setup_logger
 
 logger = setup_logger()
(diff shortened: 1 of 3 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

justin-tahara (approved):

nmgarza5:
@greptile

PR #11826, opened 2026-06-08
fix(users): revalidate users table + counts after invite

nmgarza5:
## Description

**Bug:** `InviteUsersModal` awaited the invite request but never revalidated any SWR key, so a successfully-invited user did not appear in the table until the next focus revalidation — making a working invite look like a no-op.

**Fix:** Revalidate `invitedUsers` / `acceptedUsers` / `userCounts` on success so the row and the toast land together.

## Additional Options


nmgarza5:
@greptile

acaprau (approved):

PR #11828, opened 2026-06-08
ci: add python-version input to setup-python action

nmgarza5:
## Description

The shared `setup-python-and-install-dependencies` action hardcodes `python-version: "3.13"` (set in #11721), with no override. Consumers still on an older runtime — e.g. `danswer-control-plane`, which deploys on Python 3.11 with 3.11-pinned deps — can't opt out, so their pytest jobs break on 3.13-incompatible deps (asyncpg, SQLAlchemy).

Add an optional `python-version` input, **defaulting to `"3.13"`** (all onyx usages unchanged), so consumers can pass e.g. `"3.11"`.

## Additional Options

### DIFF
--- a/.github/actions/setup-python-and-install-dependencies/action.yml
+++ b/.github/actions/setup-python-and-install-dependencies/action.yml
@@ -4,6 +4,10 @@ inputs:
   requirements:
     description: "Newline-separated list of requirement files to install (relative to repo root)"
     required: true
+  python-version:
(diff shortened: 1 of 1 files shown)


jmelahman (approved):
As an aside, we should also just freeze Onyx actions downstream. I.e. like,

```
        uses: onyx-dot-app/onyx/.github/actions/slack-notify@6ca2bb1ec4ffd879bb1c7d67a7603cad238cf602
```
etc.

PR #11860, opened 2026-06-09
fix(craft): always register all supported providers in opencode config

rohoswagger:
## Description

Per-turn cross-provider model overrides in Craft failed with **"Model not found: `<provider>/<model>`"** when the requested provider wasn't baked into the sandbox's opencode config at provision time.

Root cause: opencode reads its provider config from `OPENCODE_CONFIG_CONTENT` (K8s) / the container env (Docker) **once at pod/container start**, and the set we baked was derived from the org's *configured* providers at that moment. So the config drifted — a sandbox provisioned when only OpenAI was configured could never serve a later Anthropic override, and K8s can't refresh a running container's env.

Fix: make the provider set **independent of org state**. `get_all_build_mode_llm_configs` now always registers every `BUILD_MODE_ALLOWED_PROVIDER_TYPES` — real config when the org has it, else a dummy-key placeholder (`BUILD_MODE_NOT_CONFIGURED_API_KEY`). A cross-provider override can therefore never hit "model not found", and a turn targeting an unconfigured provider **fails closed** — egress-proxy 403 on K8s (no real key to inject), upstream 401 on Docker — surfacing as a real terminating error instead of confusing failures or indefinite keepalives.

- **K8s**: already builds the config from the full `all_llm_configs` list, so it picks up the expanded set with no manager changes.
- **Docker**: previously *ignored* `all_llm_configs` and baked a single-provider config (with a warning that overrides "will fail until Docker grows multi-provider support"). Switched to the multi-provider builder via a new testable `_container_llm_configs` helper (the Docker analog of K8s's `_placeholder_llm_configs`), which also fixes the single-provider regression there.

No new provider/model parsing scheme — provider and model are still passed separately; the config `model` stays a single `provider/model` (no double-prefix).

### Not in scope
Sandboxes provisioned **before** this change keep their frozen env until they recycle (sleep/terminate/re-provision); new sandboxes are immune. Auto-recreation of legacy pods (snapshot→recreate→restore) was discussed and intentionally deferred.

## How Has This Been Tested?

Unit tests (`uv run pytest`, all green):
- `test_get_all_build_mode_llm_configs.py` — rewritten for the new contract: all supported types always present, real key for configured, dummy key for unconfigured, default preserved first, no duplicate provider blocks.
- `test_docker_multi_provider_config.py` (new) — `_container_llm_configs` registers every provider; proxy swaps real keys for the placeholder while leaving keyless providers untouched; no-proxy keeps keys (incl. the dummy) verbatim; rendered config enables all providers, leaks no real key, single `provider/model` prefix.
- Existing `test_opencode_config.py` / `test_opencode_key_ripout.py` / `test_docker_manager_config.py` / proxy `test_llm_provider_key_resolver.py` (fail-closed → 403) still pass.

No live Docker craft stack was available in this environment to exercise an end-to-end cross-provider turn.

## Additional Options

---
## Summary by cubic
Always register all supported LLM providers in sandbox `opencode.json` so cross‑provider overrides never hit “Model not found”. Unconfigured providers are included with a dummy key and fail closed (proxy 403; Docker 401).

- **Bug Fixes**
  - `get_all_build_mode_llm_configs` now returns the default plus every supported provider type: configured types keep real keys; others use the dummy `BUILD_MODE_NOT_CONFIGURED_API_KEY`. Default stays first; no duplicate provider blocks.
  - Docker now bakes a multi‑provider config via `build_multi_provider_opencode_config`. `_container_llm_configs` preloads all providers and, when the proxy is on, swaps any non‑empty key for `SANDBOX_PROXY_INJECTED_PLACEHOLDER` (keyless providers stay keyless); without proxy, keys are kept verbatim.

- **Refactors**
  - Added `BUILD_MODE_NOT_CONFIGURED_API_KEY` and the `_container_llm_configs` helper. New `test_docker_multi_provider_config` and updated `test_get_all_build_mode_llm_configs`.

<sup>Written for commit 6e96cbffec44a48d6574c9f3fdffca1613d9e44a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11860?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/server/features/build/configs.py
+++ b/backend/onyx/server/features/build/configs.py
@@ -67,6 +67,11 @@ ENABLE_CRAFT = os.environ.get("ENABLE_CRAFT", "false").lower() == "true"
 # shared recommended-models config (served via /build/recommended-models).
 BUILD_MODE_ALLOWED_PROVIDER_TYPES = ["anthropic", "openai", "openrouter"]
 
+# apiKey sentinel for a supported provider the org hasn't configured. We register
+# every supported provider so a cross-provider override never hits "model not
+# found"; an unconfigured one fails closed instead (proxy 403 / upstream 401).
+BUILD_MODE_NOT_CONFIGURED_API_KEY = "onyx-provider-not-configured"
+
 # Dev/debug-only: exposes an SSE endpoint that tails the sandbox pod's
 # opencode-serve container logs. Never enable in prod — the logs include LLM I/O
 # and tool invocations that may contain sensitive data. When false, the endpoint
--- a/backend/onyx/server/features/build/sandbox/docker/docker_sandbox_manager.py
+++ b/backend/onyx/server/features/build/sandbox/docker/docker_sandbox_manager.py
@@ -133,7 +133,7 @@ from onyx.server.features.build.sandbox.util.agent_instructions import (
     generate_agent_instructions,
 )
 from onyx.server.features.build.sandbox.util.opencode_config import (
-    build_opencode_config,
+    build_multi_provider_opencode_config,
 )
 from onyx.utils.logger import setup_logger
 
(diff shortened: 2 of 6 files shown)


wenxi-onyx (approved):

rohoswagger:
@greptile-ai @cubic-dev-ai re-review this

PR #11878, opened 2026-06-10
fix(chat): runs survive client disconnect — workers own persistence

nmgarza5:
## Description

**Bug:** When the client disconnects mid-run (closed tab, proxy timeout), the run dies with the connection: the reserved assistant message keeps its placeholder text forever, and the session can stay stuck in "processing". There is disconnect-cleanup code in `_run_models` that is supposed to save finished work, but it runs inside the dying request's generator teardown, and the ASGI stack can abandon that teardown entirely — in a production incident it never executed (no logs, placeholder untouched). Background: [Deep Research instability tracker](https://dev-wiki.onyx.app/app/wiki/Engineering%20Projects/Onyx%20Projects/Deep%20Research/dr_instability_fixes.md), follow-up A′.

**Fix:** saving results no longer depends on the dying request. The worker threads are unaffected by a disconnect and keep running, so they now do the saving themselves:

- each worker saves its own model's outcome when its loop exits (lock-guarded, exactly once)
- after a disconnect, the last worker to finish runs the wrap-up: a final save sweep, resetting the session's "processing" flag, and draining the packet queue
- the disconnect cleanup no longer blocks or saves anything
- the stop button still responds immediately: already-finished answers are saved as complete, in-flight ones as a partial snapshot marked "stopped by the user"

Applies to regular chat, multi-model chat, and deep research — all three run through `_run_models`.

### In plain terms

A run has three parts: your browser holding a connection, a coordinator on the server feeding output down it, and worker threads doing the actual LLM work. The coordinator was also the one who saved the finished answer.

**Before:** saving depended on the coordinator, which dies with the connection. Its "last gasp" cleanup was supposed to save finished work, but the ASGI stack can skip that teardown entirely — so workers could finish a perfectly good answer and nobody would ever write it down. The chat kept the placeholder forever and could stay stuck on "processing".

**After:** each worker saves its own answer the moment it finishes — workers aren't tied to the connection, so they keep going through a disconnect. A lock makes every save exactly-once, and the last worker to finish does the housekeeping (clear "processing", drop undelivered output). Close your laptop mid-deep-research, come back, and the finished report is in the chat.

## Additional Options

### DIFF
--- a/backend/onyx/chat/process_message.py
+++ b/backend/onyx/chat/process_message.py
@@ -14,6 +14,7 @@ from collections.abc import Callable
 from collections.abc import Generator
 from concurrent.futures import ThreadPoolExecutor
 from contextvars import Token
+from enum import Enum
 from typing import Final
 from uuid import UUID
 
@@ -1018,6 +1019,15 @@ def build_chat_turn(
 # Sentinel placed on the merged queue when a model thread finishes.
 _MODEL_DONE = object()
 
+
+# Which exit path persisted a model's outcome — for log attribution.
+class _PersistContext(Enum):
+    WORKER = "worker"
+    STOP_BUTTON = "stop-button"
+    NORMAL = "normal"
+    POST_STEPS = "post-steps"
+
+
 # How often the drain loop polls for user-initiated cancellation (stop button).
 _CANCEL_POLL_INTERVAL_S: Final[float] = 0.05
 
@@ -1081,7 +1091,10 @@ def _run_models(
     drain_done = threading.Event()
 
     def _persist_model_outcome(
-        model_idx: int, context: str, *, stop_button: bool = False
+        model_idx: int,
+        context: _PersistContext,
+        *,
+        stop_button: bool = False,
     ) -> None:
         """Persist one model's outcome exactly once, from any thread.
 
@@ -1120,7 +1133,7 @@ def _run_models(
         except Exception:
             logger.exception(
                 "%s completion failed for model %d (%s)",
-                context,
+                context.value,
                 model_idx,
                 setup.model_display_names[model_idx],
             )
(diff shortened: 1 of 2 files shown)


nmgarza5:
@greptile

backend/onyx/chat/process_message.py
```diff
@@ -1071,11 +1071,65 @@ def _run_models(
...
+            return
+
+        try:
+            llm_loop_completion_handle(
+                state_container=state_containers[model_idx],
+                is_connected=setup.check_is_connected,
+                assistant_message=setup.reserved_messages[model_idx],
+                llm=setup.llms[model_idx],
+                reserved_tokens=setup.reserved_token_count,
+            )
```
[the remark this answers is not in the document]
nmgarza5: Closed completely rather than narrowing the window: run_llm_loop/run_deep_research_llm_loop never observe the stop signal (no is_connected/stop-signal reads in the loop code), so a non-errored worker return always holds a complete answer — worker self-persists now always pass completed_normally=True. The stop-button path force-claims every model at stop time (finished → complete; in-flight → partial snapshot + stopped-by-user, matching the pre-refactor behavior), so no persist decision depends on stop-signal timing anymore.

nmgarza5:
@greptile

backend/onyx/chat/process_message.py
```diff
@@ -1071,11 +1071,83 @@ def _run_models(
...
+    persisted: list[bool] = [False] * n_models
+    finished_count: list[int] = [0]
+    post_steps_done = threading.Event()
 
     # Set when the drain loop exits early (HTTP disconnect / GeneratorExit).
     # Signals emitters to skip future puts so workers exit promptly.
     drain_done = threading.Event()
 
+    def _persist_model_outcome(
+        model_idx: int, context: str, *, stop_button: bool = False
```
evan-onyx: the context seems like it should be an enum

evan-onyx (approved):

backend/onyx/chat/process_message.py
nmgarza5: Done — `_PersistContext` enum.

PR #11856, opened 2026-06-09
perf(llm): cache non-admin LLM provider listings per tenant

nmgarza5:
## Description

`GET /llm/provider` and `GET /llm/persona/{id}/providers` rebuild the same descriptor payload from Postgres on every chat page load — ~2.7s p90 in-handler in cloud (descriptor construction over providers × model configurations), stacking under peak concurrency until it trips the functional-frontend uptime monitor.

Fix: memoize the final response in the tenant-scoped cache backend (Redis; Postgres on Lite).

- Keyed by persona / admin status / group memberships — authz semantics unchanged (persona existence + access still checked per-request; groups are part of the key).
- Entries namespaced by a per-tenant version token; LLM provider mutation endpoints (PUT/DELETE provider, defaults, image-gen configs) rewrite it.
- 60s TTL bounds staleness for writes that bypass those endpoints (persona default-model edits, auto-mode sync, seeding).
- Cache failures fall through to Postgres.

## Additional Options

- [x] [Optional] Please cherry-pick this PR to the latest release version.
- [x] [Optional] Override Linear Check

PR #11879, opened 2026-06-10
perf(llm): memoize litellm reasoning-support probe per process

nmgarza5:
## Description

**Problem:** to label models as "reasoning" in the model picker (and pick reasoning settings in chat/deep research), the backend asks `litellm.supports_reasoning` per model. For models missing from litellm's local table (e.g. self-hosted Ollama), litellm answers by **calling the model's host over the network — on every request**. All these calls share litellm's sync httpx connection pool, so one slow/unreachable host queues everyone: py-spy during the 2026-06-09 incident showed 80/92 in-flight threads blocked on that pool lock, stalling chat bootstrap and completions.

**Fix:** ask once per process, remember the answer.
- Successful probes cache permanently (model metadata is static).
- Failures cache as `False` for 300s — an unreachable host costs one attempt per window, but recovers without a restart (a stuck `False` silently downgrades reasoning models).
- Per-model single-flight lock: concurrent cold misses produce one probe, not a stampede.

All callers heal via the shared helper. Complements #11856 (cached the listing response); this fixes the underlying per-call cost everywhere else.

## Additional Options

### DIFF
--- a/backend/onyx/llm/utils.py
+++ b/backend/onyx/llm/utils.py
@@ -843,13 +843,12 @@ def litellm_thinks_model_supports_image_input(
 
 _REASONING_PROBE_FAILURE_TTL_SECONDS = 300
 
-# value: (result, expires_at) — expires_at None means permanent (a real probe
-# result, which is static model metadata); a float means a failure placeholder
-# that re-probes once the TTL passes.
+# (result, expires_at): None expiry = permanent probe result (static metadata);
+# float = failure placeholder that re-probes once the TTL passes
 _LITELLM_SUPPORTS_REASONING_CACHE: dict[str, tuple[bool, float | None]] = {}
 
-# per-model locks so concurrent cold misses probe once instead of stampeding;
-# a single shared lock would serialize unrelated models behind one slow host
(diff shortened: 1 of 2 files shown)


nmgarza5:
@greptile

backend/onyx/llm/utils.py
```diff
@@ -839,9 +841,62 @@ def litellm_thinks_model_supports_image_input(
         return False
 
 
-def model_is_reasoning_model(model_name: str, model_provider: str) -> bool:
-    import litellm
+_REASONING_PROBE_FAILURE_TTL_SECONDS = 300
+
+# (result, expires_at): None expiry = permanent probe result (static metadata);
+# float = failure placeholder that re-probes once the TTL passes
+_LITELLM_SUPPORTS_REASONING_CACHE: dict[str, tuple[bool, float | None]] = {}
```
evan-onyx, line 851: I think this leaks info cross-tenant, should probably go in redis

evan-onyx (approved):

nmgarza5:
@greptile

PR #11880, opened 2026-06-10
feat(connectors): add Braintrust connector

nmgarza5:
## Description

New checkpointed connector indexing Braintrust LLM-eval content so experiment results are queryable through Onyx search/chat and computable in code interpreter.

**Doc model** (one doc per object)
- Prompts — template, model, description
- Datasets — prose header + `TabularSection` of all rows; CSV staged → `Document.file_id`
- Experiments — prose score summary (scores, baseline deltas, regressions via `/summarize`) + `TabularSection` of per-case rows with one flattened column per score; CSV staged → `file_id`

**Behavior**
- Rows fetched via BTQL with the `[start, end]` window pushed server-side (the plain fetch endpoint silently ignores a `filter` body field — verified) and keyset pagination on `_pagination_key`
- Datasets rebuilt only when a cheap BTQL probe finds rows created in the window; experiments (write-once) re-indexed only when created in the window
- `experiment_row_lookback_days` (default 30, 0 = unlimited): older in-window experiments get summary-only docs — nightly runs recreate every row under a new experiment, so table volume otherwise grows with time, not suite size
- Rate limiting: 4 req/s client throttle, 429 honored via `Retry-After`, transient 5xx retried; per-object failures yield `EntityFailure` instead of aborting the sweep
- Optional project-name scoping

## Additional Options


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

Bo-Onyx (approved):
Things to consider:

Row filtering by the poll's `[start, end]` window is done client-side. So we download the rows every poll and throw most away — only indexing is saved, not the fetch. 
Braintrust supports server-side filtering on `fetch` via a BTQL `filter` (e.g. `created >= start`), so we can push the window bound into the request and have Braintrust return only the rows we need.

nmgarza5:
Verified the server-side filtering suggestion empirically — one important catch: `POST /v1/{obj}/{id}/fetch` **silently ignores** a `filter` body field (even syntactically invalid filters return 200 with unfiltered events), so adding it there would look incremental while still downloading everything. The BTQL endpoint does filter correctly (`filter: created >= '...'` verified returning 0 rows for a future cutoff), with keyset pagination via `_pagination_key`. Reworking the row fetching to BTQL with the window pushed server-side — also moving rows into TabularSections per Evan's suggestion.

nmgarza5:
@greptile

PR #11892, opened 2026-06-10
fix(craft): make sandbox snapshots reliable + fail-closed idle cleanup

rohoswagger:
## Description

Craft sessions were silently restored as the **blank template** instead of the user's app. Root cause chain: idle-cleanup session snapshots were **failing**, but the pod was **terminated anyway** — so on the next open there was no snapshot and the restore endpoint fell back to `setup_session_workspace()` (fresh template), permanently losing the workspace.

This PR fixes the snapshot reliability bugs and makes idle-cleanup fail-closed so a transient failure can no longer destroy a user's workspace.

- **Fail-closed idle cleanup** (`tasks.py`): if a session snapshot fails and the pod is still healthy, leave the sandbox `RUNNING` and retry next cycle instead of terminating with unsnapshotted work. Only terminate-despite-failure when the pod is unreachable (can't recover, won't pin `RUNNING` forever).
- **K8s snapshot reliability** (`kubernetes_sandbox_manager.py`): re-sign the sidecar request **per host attempt**. A slow first attempt (e.g. a hung in-pod upload consuming the client timeout) handed the pod-IP failover a >60s-old timestamp, surfacing as a misleading `401 "Timestamp out of range"` instead of the real failure. Applies to `_post_to_sidecar` and `write_files_to_sandbox`.
- **Sidecar snapshot daemon** (`snapshot.py`): bound `tar | s5cmd` pipelines with a process-group-killing timeout so a hung `s5cmd` fails loudly instead of holding the HTTP request open; copy `PIPESTATUS` in one shot (reading the indices on separate lines resets it, so an `s5cmd` upload failure was silently read as exit 0 → a false `"created"` with no blob).
- **Dev env** (`.vscode/.env.k8s.template`): document the `AWS_*` / `SANDBOX_S3_BUCKET` the sidecar's `s5cmd` needs (the manager forwards `AWS_*`, not the `S3_*` file-store vars). Without them, in-pod uploads hang against real S3 behind the egress lockdown and snapshots never land.

**Docker backend:** unaffected by the K8s-specific bugs — it streams `tar` through the api-server's `FileStore` (no sidecar / `s5cmd` / signed-timestamp path). The fail-closed cleanup is backend-agnostic, and Docker already raises on a failed `tar`; `test_docker_snapshot_failure_propagation` pins that contract.

## How Has This Been Tested?

- **Unit:** Docker snapshot failure-propagation (`_GeneratorReader` surfaces a non-zero `tar` exit, doesn't swallow it as EOF); existing docker sandbox unit tests (57 passed).
- **External-dependency-unit (kind):** rewrote idle-cleanup tests for the fail-closed contract — healthy-pod snapshot failure stays `RUNNING`, unreachable-pod failure still terminates; assertions scoped to the test's own sandbox (the task is tenant-wide). `test_snapshot_restore.py` passes with only the env file (sidecar gets S3 creds from a fresh pod).
- **End-to-end on a kind cluster:** built a marker page in a fresh (fix-enabled) sandbox → ran the real idle-cleanup → confirmed a real snapshot blob in MinIO containing the page → re-provisioned + restored → the **actual app renders in the preview** (not the template).

## Additional Options

---
## Summary by cubic
Fixes unreliable sandbox snapshots and makes idle cleanup fail-closed so transient failures no longer wipe user work. K8s sidecar requests are re-signed per attempt and the `tar | s5cmd` pipeline now has a bounded 240s timeout that kills the process group.

- **Bug Fixes**
  - Idle cleanup: if any snapshot fails and the pod is healthy, leave the sandbox RUNNING and retry next cycle; only terminate if the pod is unreachable.
  - K8s manager: sign sidecar requests per host attempt in `_post_to_sidecar` and `write_files_to_sandbox` to avoid stale-timestamp 401s masking real errors.
  - Sidecar: add a 240s timeout that kills the entire `tar | s5cmd` process group; snapshot `PIPESTATUS` in one read so `s5cmd` failures are detected; propagate non-zero exits with error output (create and restore).
  - Docker backend: unchanged; added tests to confirm snapshot failures propagate and do not produce false “created” snapshots.

- **Migration**
  - Set `SANDBOX_S3_BUCKET` and `AWS_*` (including `AWS_ENDPOINT_URL`) in `.vscode/.env.k8s.template` for the sidecar’s `s5cmd`.
  - Create the bucket once per cluster before testing snapshot/restore.

<sup>Written for commit 769af2360667e1cb5408452754a6416b6915e088. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11892?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.vscode/.env.k8s.template
+++ b/.vscode/.env.k8s.template
@@ -138,6 +138,16 @@ SANDBOX_NAMESPACE=onyx-sandboxes
 # In-cluster Service hosting the egress proxy. Required.
 SANDBOX_PROXY_HOST=onyx-sandbox-proxy.onyx.svc.cluster.local
 
+# Snapshot storage for the sandbox sidecar's s5cmd. The manager forwards AWS_*
+# (NOT the S3_* file-store vars above) into the sidecar; without these,
+# idle-sleep snapshots never land and session workspaces are lost on restore.
+# Create the bucket once: aws --endpoint-url http://onyx-minio:9000 s3 mb s3://onyx-sandbox-files
+SANDBOX_S3_BUCKET=onyx-sandbox-files
+AWS_ACCESS_KEY_ID=minioadmin
+AWS_SECRET_ACCESS_KEY=minioadmin
+AWS_REGION=us-east-1
+AWS_ENDPOINT_URL=http://onyx-minio:9000
+
 # Ed25519 private key signing pushes to the sandbox push daemon. Must match
 # the public key the sandbox image expects. The dev value below is a
 # 32-byte all-zeros key — fine for kind. Rotate for any shared cluster.
--- a/backend/onyx/server/features/build/sandbox/image/sandbox_daemon/snapshot.py
+++ b/backend/onyx/server/features/build/sandbox/image/sandbox_daemon/snapshot.py
@@ -5,7 +5,9 @@ upload/download tar.gz archives to/from S3. Tarring/extraction happens
 via shell pipelines so we don't buffer large snapshots in memory.
 """
 
+import os
 import shlex
+import signal
 import subprocess
 from pathlib import Path
 from uuid import UUID
@@ -18,6 +20,10 @@ SESSIONS_ROOT = Path("/workspace/sessions")
 BUN_CACHE_DIR = SESSIONS_ROOT / ".bun-cache"
 BUN_IMAGE_CACHE_DIR = Path("/home/sandbox/.bun/install/cache")
 
+# Bounded under the manager's 300s client deadline so a hung s5cmd fails as a
+# SnapshotError instead of holding the request open.
+SNAPSHOT_PIPELINE_TIMEOUT_SECONDS = 240
+
 
 class SnapshotError(RuntimeError):
     """Raised when a snapshot subprocess fails. Carries stderr from the
@@ -25,7 +31,7 @@ class SnapshotError(RuntimeError):
     """
 
 
-def _run(script: str) -> None:
+def _run(script: str, timeout: float | None = None) -> None:
     """Run a shell script with stderr merged into stdout for the error.
 
     We deliberately merge stderr into stdout (rather than capturing them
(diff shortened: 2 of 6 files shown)


.vscode/.env.k8s.template
```diff
@@ -138,6 +138,16 @@ SANDBOX_NAMESPACE=onyx-sandboxes
 # In-cluster Service hosting the egress proxy. Required.
 SANDBOX_PROXY_HOST=onyx-sandbox-proxy.onyx.svc.cluster.local
 
+# Snapshot storage for the sandbox sidecar's s5cmd. The manager forwards AWS_*
+# (NOT the S3_* file-store vars above) into the sidecar; without these,
+# idle-sleep snapshots never land and session workspaces are lost on restore.
+# Create the bucket once: aws --endpoint-url http://onyx-minio:9000 s3 mb s3://onyx-sandbox-files
```
wenxi-onyx, line 147: Pref removing this line as the bucket should never be manually created (I am working on this)

wenxi-onyx: The below env vars should ideally be codified into the default dev values (also working on this), should be fine to merge for now

backend/onyx/server/features/build/sandbox/image/sandbox_daemon/snapshot.py
```diff
@@ -18,33 +20,44 @@
...
+
 
 class SnapshotError(RuntimeError):
     """Raised when a snapshot subprocess fails. Carries stderr from the
     underlying tool (s5cmd / tar) so the manager can see the cause.
     """
 
 
-def _run(script: str) -> None:
+def _run(script: str, timeout: float | None = None) -> None:
```
wenxi-onyx, line 34: Maybe follow up - we should really get rid of all of this shell script stuff and do it via the daemon API

wenxi-onyx (approved):

PR #11917, opened 2026-06-10
feat(load-tests): Add Locust load-test harness for the chat path + deterministic mock LLM server 0/n

justin-tahara:
## Description

Adds the foundation for load testing the critical chat path (streaming chat turns, tool calls, and later deep research), so we can find and fix product/infra failure modes — DB connections held per turn, per-stream memory, model-server saturation — before large-scale deployments hit them.

**Guiding principle: these tests measure Onyx's application code and infrastructure, never LLM answer quality.** The LLM is treated as a controllable dependency via a mock server, so call volume is unlimited/free and every measured regression is attributable to Onyx code.

What's included (new top-level `loadtest/` directory):

- **Standalone uv project** — intentionally *not* a member of the root uv workspace (Locust pins gevent/flask, which must not constrain the backend lockfile). Excluded from root workspace discovery and from the pre-commit `ty` hook, following the existing sandbox_daemon/skills pattern. Load-test code never imports `onyx.*` (gevent monkey-patching breaks backend deps).
- **`onyx_client/stream_parser.py`** — incremental NDJSON chat-stream parser (vendored/adapted from the integration-test `ChatSessionManager`, stdlib-json only) that surfaces milestones the moment packets arrive.
- **`onyx_client/chat_user.py` + `locustfile.py`** — Locust user driving `POST /api/chat/send-chat-message` with `stream=true`, firing per-milestone metrics as named pseudo-requests: `chat:first_packet`, `chat:first_answer_token` (TTFT), `chat:first_search_doc`, `chat:total_turn`. A turn fails on non-200, error packets, stream stalls, or streams ending without answer content. Configuration via env vars (`ONYX_API_KEY`, optional `llm_override`, deep-research flag, timeouts).
- **`mock_llm/app.py`** — deterministic OpenAI-compatible mock LLM server (`/v1/chat/completions` streaming + non-streaming, `/v1/models`). Timing knobs are encoded in the model name (`mock-ttft500-itl20-len400` = 500ms time-to-first-token, 20ms inter-token, 400-token answer) since litellm passes model names through verbatim. Optional tool-call scripting (`MOCK_FORCE_TOOL_CALL`) to drive the real tool-execution path. Register it as an `openai_compatible` provider with the server URL as `api_base` — that provider type sends plain chat-completions to the custom base URL (true-`openai` providers get routed through litellm's `/v1/responses` bridge, which the mock intentionally doesn't implement).

Roadmap (follow-up PRs): in-cluster Locust master/workers + mock provider on a dev cluster, weighted scenario suite (chat+search multi-turn, deep research), Prometheus export + Grafana correlation dashboards. See `loadtest/README.md`.

## How Has This Been Tested?

- **End-to-end against a locally running Onyx stack**: registered the mock server as an `openai_compatible` provider (custom `api_base`, default chat model) and confirmed a full chat turn flows through the real production path — api_server → litellm → mock LLM → NDJSON stream with `message_start` → `message_delta`s → `stop` — via curl before running load.
- **Locust headless run** (`-u 3 -r 1 -t 60s` against the local stack): 24 chat turns, **0 failures**, with sane milestone stats — `chat:first_packet` p50 39ms, `chat:first_answer_token` p50 420ms (mock TTFT 300ms + ~120ms Onyx pipeline overhead, i.e. the harness measures real app-side latency), `chat:total_turn` 2829–2889ms across 21 completed turns. The 60ms total-turn spread demonstrates the run-to-run determinism the mock exists to provide.
- **Failure paths exercised during development**: missing default LLM and provider auth errors both surfaced as failed turns with the error packet content attached (not silent successes).
- Mock server endpoints smoke-tested directly (streaming chunk format, knob parsing via model-name suffix).
- `uv sync` in `loadtest/` resolves into its own lockfile; root `uv.lock` untouched. Pre-commit (including root `ty`/ruff) passes with the hook exclude in place.

## Additional Options

---
## Summary by cubic
Adds a standalone `loadtest/` Locust harness for the chat streaming path and a deterministic OpenAI-compatible mock LLM to measure app/infra latency (TTFT, first packet, search docs, total turn) without using paid models. Stream success now requires a stop packet, and TTFT fires only on the first real content token to avoid false positives.

- **New Features**
  - Locust user posts to `/api/chat/send-chat-message` with `stream=true` and emits `chat:first_packet`, `chat:first_search_doc`, `chat:first_answer_token` (TTFT), and `chat:total_turn`. Config via `ONYX_API_KEY`, optional `ONYX_LLM_PROVIDER`/`ONYX_LLM_MODEL`, `ONYX_WAIT_SECONDS`, `ONYX_STREAM_READ_TIMEOUT`, `ONYX_DEEP_RESEARCH`.
  - Incremental NDJSON stream parser (`loadtest/onyx_client/stream_parser.py`) that fires milestones as chunks arrive; counts turn success only if a `stop` packet is seen and triggers TTFT only on non-empty answer content.
  - Deterministic mock LLM server (`loadtest/mock_llm/`) implementing `/v1/chat/completions` (streaming and non-streaming) and `/v1/models`. Timing knobs via model name (e.g. `mock-ttft500-itl20-len400`); optional tool-calls via `MOCK_FORCE_TOOL_CALL`. Register as an `openai_compatible` provider with `api_base`.

- **Dependencies**
  - `loadtest/` is an isolated uv project with `locust`, `fastapi`, and `uvicorn`; excluded from the root `uv` workspace and from pre-commit type checks. No changes to backend dependencies or the root lockfile.

<sup>Written for commit df4ef176182629b22bbcf531775718b553ae823a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11917?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


loadtest/mock_llm/app.py
```diff
@@ -0,0 +1,206 @@
...
+from fastapi import FastAPI
+from fastapi import Request
+from fastapi import Response
+from fastapi.responses import JSONResponse
+from fastapi.responses import StreamingResponse
+
+app = FastAPI()
+
+DEFAULT_TTFT_MS = int(os.environ.get("MOCK_TTFT_MS", "300"))
+DEFAULT_ITL_MS = int(os.environ.get("MOCK_ITL_MS", "15"))
```
nmgarza5, line 44: Do you plan on running variations of these? or what are your thoughts on tuning the knobs to represent different llm providers. I think increased latency is something we should test

loadtest/mock_llm/app.py
```diff
@@ -0,0 +1,206 @@
...
+def _chunk(
+    model: str, completion_id: str, delta: dict[str, Any], finish: str | None
+) -> str:
+    payload = {
+        "id": completion_id,
+        "object": "chat.completion.chunk",
+        "created": int(time.time()),
+        "model": model,
+        "choices": [{"index": 0, "delta": delta, "finish_reason": finish}],
+    }
```
nmgarza5, line 77: maybe overkill but can we add pydantic models? a few places in the pr where having typed request/response models would be nice

loadtest/mock_llm/app.py
```diff
@@ -0,0 +1,206 @@
...
+    return f"data: {json.dumps(payload)}\n\n"
```
nmgarza5, line 78: nit: i think we decided we weren't going to use fstrings in the codebase anymore. maybe there was nuance to that though

nmgarza5 (approved):

loadtest/mock_llm/app.py
justin-tahara: Yes — exactly the plan. The knobs ride in the model name, so each scenario picks a latency profile per run without redeploying the mock. The follow-up branch (`jtahara/chat-loadtest-phase1`) adds a "Provider profiles" table to the README with presets: fast gpt-class (`mock-ttft300-itl15-len150`), slow reasoning model with a long silent TTFT (`mock-ttft8000-itl40-len600`), and a degraded/overloaded provider (`mock-ttft20000-itl200-len300`). The slow profiles are honestly the most valuable runs: streams held open longer is precisely what stresses api-server connections, DB sessions, and per-stream memory.

loadtest/mock_llm/app.py
justin-tahara: Done on the phase-1 branch (`jtahara/chat-loadtest-phase1`, commit 220b1fd44a, where this file gets rewritten for tool-call/deep-research support anyway): new `mock_llm/models.py` with typed request models (`ChatCompletionRequest`/`ChatMessage`/`ToolDefinition`) and response/chunk models, FastAPI request validation, `extra="allow"` everywhere so litellm's extra fields can never 422. The refactor also tightened the wire format (delta omits unset fields, `finish_reason: null` always present, matching OpenAI).

loadtest/mock_llm/app.py
justin-tahara: I looked for the rule — couldn't find anything codified in CONTRIBUTING.md, the ruff config, or pre-commit (no flake8-logging-format/G004 either). My guess is this is the "no f-strings in *logging calls*" convention (lazy %-formatting so the string isn't built when the log level is off) — this file has no logging, the f-strings just build payloads. Happy to switch if there's a decision I'm missing — point me at it and I'll fold it into the phase-1 branch.

nmgarza5: this pr should have enforced it https://github.com/onyx-dot-app/onyx/pull/10767. separate pyproject.toml so that is why. if you could follow up and add this rule that would be appreciated

justin-tahara: Followed up in #11938 (commit c4f106d2c4) — and found something bigger while adding it: **G004 is currently missing from the root pyproject too.** #10767 added it to the root `select` list, but a later lint-config rewrite (`07eaa81298`, the GCS Workload Identity PR) replaced the whole select list and silently dropped it — so it hasn't been enforced anywhere for ~6 weeks. The separate loadtest pyproject wasn't actually the cause: it has no `[tool.ruff]` section, so ruff falls back to the root config for loadtest files.

The commit restores `G004` to the root select (with a comment so it doesn't get dropped again), converts the 15 f-string logging calls that crept in since (13 of them in `gcs_file_store.py`, fittingly from the same PR that dropped the rule), and adds a note in `loadtest/pyproject.toml` warning against ever adding a `[tool.ruff]` section there without replicating the root rules.

justin-tahara: Update: the G004 restore moved to its own PR — #11945 — to keep backend changes out of the load-test PR.

PR #11923, opened 2026-06-11
feat(skills): add user-managed personal skills

rohoswagger:
## Description

Adds user-managed **personal skills**. Basic users can create, replace the bundle of, enable/disable, and delete their own skill bundles, which are visible only to themselves. Admins retain oversight on the `/admin/skills` surface: they can widen a personal skill's visibility (org-wide or to specific groups) via the Share modal — which warns that the author loses self-serve management — disable it (a reversible mute the owner can re-enable), or hard-delete it.

Key implementation details:
- New user endpoints: `POST /skills/custom`, `PUT /skills/custom/{id}/bundle`, `PATCH /skills/custom/{id}`, `DELETE /skills/custom/{id}`.
- Ownership enforced via `_ensure_owned_personal` (404 for non-owners, 403 once promoted/grants-shared).
- Per-user cap (`MAX_PERSONAL_SKILLS_PER_USER`, env-overridable) made race-free with a `pg_advisory_xact_lock` keyed on the user, so concurrent uploads can't exceed it.
- Reserved-slug and duplicate-slug (shared namespace with admin skills) handling.
- Visibility filter feeds sandbox injection with no admin bypass, so untrusted personal skills never leak into other users' agent contexts.
- Frontend: create-skill modal, per-card replace/toggle/delete actions with in-flight guards, group-ordered skill lists, and admin visibility editing that warns when a personal skill is being widened.

## How Has This Been Tested?

- Added integration tests in `backend/tests/integration/tests/skills/test_skills_personal.py` covering visibility isolation, ownership guards, reserved/duplicate slugs, cross-tier slug collisions (both directions), promotion lockout, owner toggle, admin-disabled-but-owner-deletable, admin hard-delete, and the per-user cap.
- `ruff check` / `ruff format` clean; `tsc --noEmit` clean; pre-commit hooks (ty, oxfmt, oxlint, TypeScript) pass.

## Additional Options

## Screenshots

**User skills page** — skills grouped built-in -> custom -> personal (A-Z within each); personal cards show the owner toggle / replace / delete controls.

Screenshot 2026-06-16 at 15 50 21

**Create skill modal** — upload a `.zip` bundle; the filename becomes the slug and SKILL.md frontmatter supplies the name + description.

Screenshot 2026-06-16 at 15 50 35

**Admin · Manage skills** — custom skills table grouped org-wide -> group-granted -> personal (A-Z within each).

Screenshot 2026-06-16 at 15 50 49

**Edit visibility (Share) modal** — widening a personal skill surfaces a warning, below the controls, that the author loses self-serve management.

Screenshot 2026-06-16 at 15 50 59

---
## Summary by cubic
Adds user-managed personal skills so basic users can upload, replace, toggle, and delete their own skill bundles. Personal skills are private to the owner; admins can promote them org‑wide or hard‑delete, with safe visibility, limits, and a streamlined UI.

- **New Features**
  - API: `POST /skills/custom`, `PUT /skills/custom/{id}/bundle`, `PATCH /skills/custom/{id}` (enabled), `DELETE /skills/custom/{id}` with strict ownership checks.
  - Visibility/sandboxing: only owners see their personal skills (disabled ones stay listed for them); no admin bypass in the user visibility filter so personal skills never inject into other users’ sandboxes; admins review via `/admin/skills`.
  - Safety and limits: reserved built‑in/external‑app slugs are rejected before ingest for both admin and personal creation; 409 on duplicate slugs in the shared namespace; upload read capped; per‑user cap guarded by a transaction‑scoped advisory lock.
  - Responses/UI: responses include `is_personal`; “Create skill” modal and per‑card replace/toggle/delete for owned personal skills with in‑flight guards; “Personal” labeling; the admin Share modal warns when widening a personal skill’s visibility; user page orders skills built‑in → org custom → personal (A‑Z), admin table orders org‑wide → group‑granted → personal (A‑Z).
  - Docs: added `docs/craft/features/skills/personal-skills.md` (design, API, ownership/cap/promotion model, frontend, tests; clarifies that personal status is derived and reversible by design).

- **Refactors**
  - Moved `MAX_PERSONAL_SKILLS_PER_USER` to `backend/onyx/configs/app_configs.py` and guarded negative env values via `_non_negative_int_env`; applied the helper to `MAX_ALLOWED_UPLOAD_SIZE_MB` and `DEFAULT_USER_FILE_MAX_UPLOAD_SIZE_MB`.
  - Advisory lock: derive a signed int64 lock id via SHA‑256 with explicit big‑endian for cross‑platform stability; added debug logs around lock acquisition for observability.
  - Renamed `fetch_skill_for_admin` to `fetch_skill_by_id` and updated callers.

<sup>Written for commit d97a9b0ab8d9df79b2732e8b51b601051733f7c5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/11923?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/db/skill.py
+++ b/backend/onyx/db/skill.py
@@ -18,6 +18,8 @@ These helpers never commit — callers control the transaction boundary so a
 multi-step admin flow (e.g. create row + replace grants) can roll back atomically.
 """
 
+import hashlib
+import struct
 from collections.abc import Sequence
 from dataclasses import dataclass
 from uuid import UUID
@@ -29,6 +31,7 @@ from sqlalchemy import func
 from sqlalchemy import or_
 from sqlalchemy import Select
 from sqlalchemy import select
+from sqlalchemy import text
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import selectinload
 from sqlalchemy.orm import Session
@@ -246,7 +249,7 @@ def list_skills_for_sandbox_injection(
     return list(db_session.scalars(stmt))
 
 
-def fetch_skill_for_admin(skill_id: UUID, db_session: Session) -> Skill | None:
+def fetch_skill_by_id(skill_id: UUID, db_session: Session) -> Skill | None:
     stmt = select(Skill).where(Skill.id == skill_id).options(selectinload(Skill.author))
     return db_session.scalars(stmt).one_or_none()
 
@@ -370,7 +373,7 @@ def replace_skill_bundle(
 
     Rejects built-in rows — they have no bundle.
     """
-    skill = fetch_skill_for_admin(skill_id, db_session)
+    skill = fetch_skill_by_id(skill_id, db_session)
     if skill is None:
         raise OnyxError(
             OnyxErrorCode.NOT_FOUND,
@@ -405,7 +408,7 @@ def patch_skill(
     patch: SkillPatch,
     db_session: Session,
 ) -> Skill:
-    skill = fetch_skill_for_admin(skill_id, db_session)
+    skill = fetch_skill_by_id(skill_id, db_session)
     if skill is None:
         raise OnyxError(
             OnyxErrorCode.NOT_FOUND,
@@ -424,7 +427,7 @@ def patch_skill(
 def replace_skill_grants(
     skill_id: UUID, group_ids: Sequence[int], db_session: Session
 ) -> None:
-    if fetch_skill_for_admin(skill_id, db_session) is None:
+    if fetch_skill_by_id(skill_id, db_session) is None:
         raise OnyxError(
             OnyxErrorCode.NOT_FOUND,
             f"Skill {skill_id} not found.",
@@ -451,7 +454,7 @@ def replace_skill_grants(
 
 def delete_skill(skill_id: UUID, db_session: Session) -> str | None:
     """Hard-delete a skill and return its `bundle_file_id` for caller cleanup."""
-    skill = fetch_skill_for_admin(skill_id, db_session)
+    skill = fetch_skill_by_id(skill_id, db_session)
     if skill is None:
         return None
     bundle_file_id = skill.bundle_file_id
@@ -507,19 +510,25 @@ def count_personal_skills_for_user(user_id: UUID, db_session: Session) -> int:
     return db_session.scalar(stmt) or 0
 
 
-# Arbitrary namespace so this per-user lock can't collide with other advisory locks.
-_PERSONAL_SKILL_LOCK_NAMESPACE = 0x534B494C
+# Namespaced + user-hashed so unrelated users don't block each other and the
+# lock id can't collide with other advisory locks in the codebase.
+_PERSONAL_SKILL_LOCK_NAMESPACE = "onyx_personal_skill_lock"
+
+
+def _personal_skill_lock_id(user_id: UUID) -> int:
+    digest = hashlib.sha256(
+        f"{_PERSONAL_SKILL_LOCK_NAMESPACE}:{user_id}".encode()
+    ).digest()
+    # pg_advisory_xact_lock takes a signed 8-byte int.
+    return struct.unpack("q", digest[:8])[0]
 
 
 def lock_personal_skills_for_user(user_id: UUID, db_session: Session) -> None:
     """Serialize a user's concurrent personal-skill creates so the
     count-then-insert cap check is race-free; released at transaction end."""
     db_session.execute(
-        select(
-            func.pg_advisory_xact_lock(
-                _PERSONAL_SKILL_LOCK_NAMESPACE, func.hashtext(str(user_id))
-            )
-        )
+        text("SELECT pg_advisory_xact_lock(:lock_id)"),
+        {"lock_id": _personal_skill_lock_id(user_id)},
     )
 
 
--- a/backend/onyx/db/skill.py
+++ b/backend/onyx/db/skill.py
@@ -3,8 +3,11 @@
 Access model:
 - Admin reads: see every row. Disabled skills stay visible so admins can
   re-enable them.
-- User reads: filter `enabled = True`, plus `is_public` OR the user is in a
-  group that has been granted access.
+- User reads: filter `enabled = True` (except the user's own personal skills,
+  which stay listed while disabled so the owner can re-enable them), plus
+  `is_public` OR the user is in a group that has been granted access OR the
+  skill is the user's own personal skill (custom, non-public, zero group
+  grants, authored by the user).
 
 Delete is a hard delete — `delete_skill` removes the row and returns its
 `bundle_file_id` so the caller can drop the blob from the file store
@@ -20,7 +23,9 @@ from dataclasses import dataclass
 from uuid import UUID
 
 from sqlalchemy import and_
+from sqlalchemy import ColumnElement
 from sqlalchemy import delete
+from sqlalchemy import func
 from sqlalchemy import or_
 from sqlalchemy import Select
 from sqlalchemy import select
@@ -28,7 +33,6 @@ from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import selectinload
 from sqlalchemy.orm import Session
 
-from onyx.auth.schemas import UserRole
 from onyx.db.enums import SandboxStatus
 from onyx.db.external_app import is_user_authenticated_for_app
 from onyx.db.models import ExternalApp
@@ -55,17 +59,34 @@ class SkillPatch:
     enabled: bool | UnsetType = UNSET
 
 
+def _personal_skill_clause(user_id: UUID) -> ColumnElement[bool]:
+    """Predicate matching *user_id*'s personal skills: custom, non-public,
+    zero group grants, authored by the user. Shared by the visibility
+    filter and the sandbox-push fanout so the two can't drift."""
+    no_grants = ~(
+        select(Skill__UserGroup.skill_id)
+        .where(Skill__UserGroup.skill_id == Skill.id)
+        .exists()
+    )
+    return and_(
+        Skill.author_user_id == user_id,
+        Skill.built_in_skill_id.is_(None),
+        Skill.is_public.is_(False),
+        no_grants,
+    )
+
+
 def _add_user_visibility_filter(
     stmt: Select[tuple[Skill]], user: User
 ) -> Select[tuple[Skill]]:
-    """Restrict a `select(Skill)` to rows the given user can see.
+    """Restrict a `select(Skill)` to rows the given user can see:
+    is_public OR group-granted OR their own personal skill.
 
-    Admins bypass the filter; everyone else goes through the
-    is_public-or-group-grant path.
+    Admins get NO bypass here on purpose: this filter also feeds sandbox
+    injection, and a bypass would push every user's personal skill
+    (untrusted content) into admins' agent contexts. Admin oversight lives
+    on the /admin/skills surface instead.
     """
-    if user.role == UserRole.ADMIN:
-        return stmt
-
     group_grant_exists = (
         select(Skill__UserGroup.skill_id)
         .join(
@@ -77,7 +98,13 @@ def _add_user_visibility_filter(
         .exists()
     )
 
-    return stmt.where(or_(Skill.is_public.is_(True), group_grant_exists))
+    return stmt.where(
+        or_(
+            Skill.is_public.is_(True),
+            group_grant_exists,
+            _personal_skill_clause(user.id),
+        )
+    )
 
 
 def _exclude_unavailable_built_ins(
@@ -146,10 +173,14 @@ def list_skills_for_user(user: User, db_session: Session) -> Sequence[Skill]:
     ``list_skills_for_sandbox_injection`` to get the wider set the
     sandbox actually receives (which includes authenticated external
     apps).
+
+    Disabled skills are hidden EXCEPT the user's own personal skills,
+    which stay listed (greyed out in the UI) so the owner can re-enable
+    them. Sandbox injection never includes disabled skills.
     """
     stmt = (
         select(Skill)
-        .where(Skill.enabled.is_(True))
+        .where(or_(Skill.enabled.is_(True), _personal_skill_clause(user.id)))
         .where(Skill.id.notin_(_external_app_skill_ids_subquery()))
         .options(selectinload(Skill.author))
         .order_by(Skill.name)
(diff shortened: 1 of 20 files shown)


web/src/refresh-pages/UserSkillsPage.tsx
```diff
@@ -4,28 +4,110 @@ import { useMemo, useRef, useState } from "react";
...
+
+    try {
+      await deleteUserSkill(target.id);
+      toast.success(`Deleted "${target.name}"`);
+      refresh();
+    } catch (err) {
+      console.error("Failed to delete skill", err);
+      toast.error(err instanceof Error ? err.message : "Failed to delete");
+    }
+  }
```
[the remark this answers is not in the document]
rohoswagger: Fixed in be640a069e. `handleDeleteConfirmed` now calls `setPendingId(target.id)` before the `await deleteUserSkill(...)` and clears it in a `finally`, matching `handleToggleEnabled`/`handleReplaceBundleFile`. Since `SkillCard` receives `busy={pendingId !== null}` (which disables the toggle, replace, and delete controls on every card), the in-flight delete window is now guarded just like the toggle/replace windows.

backend/onyx/server/features/skill/api.py
```diff
@@ -301,7 +363,143 @@ def fetch_skill_for_current_user(
...
+            OnyxErrorCode.INVALID_INPUT,
+            f"You have reached the limit of {MAX_PERSONAL_SKILLS_PER_USER} "
+            "personal skills. Delete one before creating another.",
+        )
+
+    file_store = get_default_file_store()
+    ingested = ingest_skill_bundle(
+        _read_bundle_upload(bundle), bundle.filename, file_store
+    )
+
```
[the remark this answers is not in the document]
rohoswagger: Fixed in be640a069e. The reserved-slug guard now runs *before* `ingest_skill_bundle`, using `slug_from_filename(bundle.filename)` — so a reserved slug is rejected with no bundle blob ever written. On the broader "ingest itself raises → orphaned blob" point: in `ingest_skill_bundle`, `file_store.save_file(...)` is the last operation, after `slug_from_filename` → `validate_custom_bundle` → `parse_skill_md_metadata` → `compute_bundle_sha256`. A malformed-zip / validation / parse error therefore raises *before* any blob is written, so it can't orphan one. A failure *after* a successful ingest (e.g. the DB insert) is still cleaned up by the existing `except: delete_bundle_blob(...)`.

backend/onyx/db/skill.py
```diff
@@ -451,6 +486,61 @@ def affected_user_ids_for_skill(skill: Skill, db_session: Session) -> set[UUID]:
...
+# lock id can't collide with other advisory locks in the codebase.
+_PERSONAL_SKILL_LOCK_NAMESPACE = "onyx_personal_skill_lock"
+
+
+def _personal_skill_lock_id(user_id: UUID) -> int:
+    digest = hashlib.sha256(
+        f"{_PERSONAL_SKILL_LOCK_NAMESPACE}:{user_id}".encode()
+    ).digest()
+    # pg_advisory_xact_lock takes a signed 8-byte int.
+    return struct.unpack("q", digest[:8])[0]
```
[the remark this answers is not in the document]
rohoswagger: Fixed in be640a069e. `_personal_skill_lock_id` now unpacks the sha256 digest with explicit big-endian (`struct.unpack(">q", digest[:8])`) instead of native-endian `"q"`, so the advisory-lock id is identical regardless of the host platform's byte order.

backend/onyx/db/skill.py
```diff
@@ -451,6 +483,55 @@ def affected_user_ids_for_skill(skill: Skill, db_session: Session) -> set[UUID]:
...
+def lock_personal_skills_for_user(user_id: UUID, db_session: Session) -> None:
+    """Serialize a user's concurrent personal-skill creates so the
+    count-then-insert cap check is race-free; released at transaction end."""
+    db_session.execute(
+        select(
+            func.pg_advisory_xact_lock(
+                _PERSONAL_SKILL_LOCK_NAMESPACE, func.hashtext(str(user_id))
+            )
+        )
+    )
```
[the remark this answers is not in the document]
rohoswagger: Addressed. This comment was on the original `pg_advisory_xact_lock(namespace, hashtext(...))` form; the caa37af632 refactor already replaced `hashtext` (32-bit) with a full 64-bit key derived from `sha256("onyx_personal_skill_lock:" + user_id)[:8]`, so distinct users no longer share the 32-bit collision space. Follow-up be640a069e makes the unpack explicit big-endian (">q") for cross-platform stability. No `hashtext` remains in the lock path.

rohoswagger:
@greptile-ai @cubic-dev-ai @codex review this

backend/onyx/db/skill.py
```diff
@@ -55,17 +62,34 @@ class SkillPatch:
     enabled: bool | UnsetType = UNSET
 
 
+def _personal_skill_clause(user_id: UUID) -> ColumnElement[bool]:
```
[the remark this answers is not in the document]
rohoswagger: not valid, this is the intended behavior of custom skills

backend/onyx/server/features/skill/api.py
```diff
@@ -100,6 +129,35 @@ def _ensure_custom(skill: Skill) -> None:
         )
 
 
+def _read_bundle_upload(bundle: UploadFile) -> bytes:
```
Danelegend, line 132: Is there a general helper function that does this? If not I'd consider turning this into one due to how general and widely applicable it is

backend/onyx/server/features/skill/api.py
```diff
@@ -301,7 +361,143 @@ def fetch_skill_for_current_user(
...
+    ):
+        raise OnyxError(
+            OnyxErrorCode.INVALID_INPUT,
+            f"You have reached the limit of {MAX_PERSONAL_SKILLS_PER_USER} "
+            "personal skills. Delete one before creating another.",
+        )
+
+    # Reject reserved slugs up front so we never write a bundle blob for one.
+    slug = slug_from_filename(bundle.filename)
+    if slug in _RESERVED_SKILL_SLUGS:
```
Danelegend: Is this restriction shared with admin created skills? I'd try to generalise if so

backend/onyx/server/features/skill/api.py
```diff
@@ -301,7 +361,143 @@ def fetch_skill_for_current_user(
...
+) -> CustomSkillResponse:
+    # fetch_skill_by_id bypasses the enabled filter on purpose: an
+    # admin-disabled personal skill must stay mutable by its owner.
+    skill = fetch_skill_by_id(skill_id, db_session)
+    if skill is None:
+        raise OnyxError(OnyxErrorCode.NOT_FOUND, "Skill not found")
+    _ensure_owned_personal(skill, user, db_session)
+
+    file_store = get_default_file_store()
+    ingested = ingest_skill_bundle(
```
Danelegend, line 442: This sort of ingest -> try -> push pattern seems to be done quite a bit across skills and external apps. Is there any way to generalise to reduce code dupe

backend/onyx/db/skill.py
```diff
@@ -451,6 +486,63 @@ def affected_user_ids_for_skill(skill: Skill, db_session: Session) -> set[UUID]:
...
+    digest = hashlib.sha256(
+        f"{_PERSONAL_SKILL_LOCK_NAMESPACE}:{user_id}".encode()
+    ).digest()
+    # Full 64-bit key (not 32-bit hashtext) so distinct users don't collide;
+    # explicit big-endian so the id is stable across platforms.
+    # pg_advisory_xact_lock takes a signed 8-byte int.
+    return struct.unpack(">q", digest[:8])[0]
+
+
+def lock_personal_skills_for_user(user_id: UUID, db_session: Session) -> None:
```
Danelegend, line 531: I'd argue this should be a general database function that takes in a callable accepting user_id and produced the lock_id

Danelegend: It's also probs important to log when someone is trying to acquire this lock

Danelegend: Might be worth making a context manager also with it owning the session commit / transaction end. Makes it a bit easier to reason about when the lock is being used and when it isn't

backend/onyx/server/features/skill/api.py
```diff
@@ -301,7 +361,143 @@ def fetch_skill_for_current_user(
...
+    )
+
+
+@user_router.post("/custom")
+def create_personal_skill(
+    bundle: UploadFile = File(...),
+    user: User = Depends(require_permission(Permission.BASIC_ACCESS)),
+    db_session: Session = Depends(get_session),
+) -> CustomSkillResponse:
+    lock_personal_skills_for_user(user.id, db_session)
```
Danelegend, line 388: Is there a db constraint we can use instead of lock?

Danelegend (approved):

backend/onyx/db/skill.py
```diff
@@ -55,17 +62,34 @@ class SkillPatch:
...
+    no_grants = ~(
+        select(Skill__UserGroup.skill_id)
+        .where(Skill__UserGroup.skill_id == Skill.id)
+        .exists()
+    )
+    return and_(
+        Skill.author_user_id == user_id,
+        Skill.built_in_skill_id.is_(None),
+        Skill.is_public.is_(False),
+        no_grants,
```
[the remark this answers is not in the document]
rohoswagger: Rationale (no code change) — this is intended. "Personal" is a *derived view* of a custom skill's current state (custom, not public, no grants, authored by you), not a stored mode. If an admin clears `is_public` and removes all grants via the normal admin visibility controls, the row is once again a private author-owned custom skill, so the author legitimately regains self-serve control. We deliberately do **not** persist a `promoted_at`/flag to make promotion one-way — that's the confirmed design for custom-skill visibility (same call as @rohoswagger made on the sibling thread). The "author loses self-serve management" warning describes the state *while* the skill is shared, which is accurate. Clarified this explicitly in docs/craft/features/skills/personal-skills.md (§1 and §8) so it isn't re-flagged as an invariant violation.

backend/onyx/db/skill.py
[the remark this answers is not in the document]
rohoswagger: Resolving per @rohoswagger's note: the reversibility is intended behavior for custom skills. Documented in docs/craft/features/skills/personal-skills.md (§1, §8) so the derived-predicate design is explicit for future readers.

backend/onyx/server/features/skill/api.py
rohoswagger: Rationale — `_read_bundle_upload` is already the single shared reader for all four skill upload endpoints (create/replace × admin/personal) in this module, so it's DRY within the feature. There's no existing general util for capped uploads. Promoting it cross-module would mean importing FastAPI's `UploadFile` into the lower-level `onyx/skills/bundle.py` (a layering smell) and adopting it in `external_apps_api.py`, which today calls `bundle.file.read()` **without** a size cap — i.e. a real but separate, pre-existing gap. I'd rather not widen this feature PR into external-apps; tracking that as a focused follow-up (give external-app bundle uploads the same cap via a shared helper) is cleaner than bundling it here.

backend/onyx/server/features/skill/api.py
rohoswagger: Good catch — it should be shared, and it wasn't. Admin `create_custom_skill` had no reserved-slug guard, so an admin could claim a built-in/external-app slug (the non-pre-seeded external-app ones like `github` would even succeed rather than 409). Fixed in 64b2c80d04: extracted `_reject_reserved_slug(bundle)` and call it from **both** `create_custom_skill` and `create_personal_skill` (before ingest, so no blob is written for a rejected slug). This also makes the pre-existing `test_create_skill_rejects_reserved_slug` admin test pass via the intended 400 path. Verified live against the running backend: admin create of `company-search.zip` now returns 400 "slug 'company-search' is reserved"; a normal admin slug still 200s.

backend/onyx/server/features/skill/api.py
rohoswagger: Rationale — the shared primitives are already factored out (`_read_bundle_upload`, `ingest_skill_bundle`, `delete_bundle_blob`, `push_skill_to_affected_sandboxes`); what's left is the orchestration, and the four sites genuinely differ: create-personal does lock + cap + reserved-slug; create-admin does grants; both replace paths additionally delete the *old* blob and return `old_file_id`; the `except` cleanup target differs per path. A single wrapper would need enough hooks (pre-create validation, post-commit cleanup, push-target selection) that it'd be leakier than the current explicit blocks. The external-apps call sites the comment mentions also skip the size cap entirely, so they're not a clean fit either. I think a shared `ingest → transactional create → push` helper is worth doing, but as a focused refactor across skills + external-apps rather than inside this feature PR — leaving as-is here.

backend/onyx/db/skill.py
rohoswagger: All three are fair. Context: this mirrors the existing `acquire_seat_lock` in `ee/onyx/db/license.py` — also a plain function (no callable-keyed helper, no context manager, no per-acquire log). So there are now two near-identical advisory-lock call sites in the codebase. The right move for (a) generalized `advisory_xact_lock(key_fn)` helper and (c) context-manager-owning-the-txn is to unify *both* sites into one util, which is a cross-cutting refactor touching the EE license code — out of scope for this PR but a good follow-up. On (b) logging: I deliberately left it off to stay consistent with the unlogged seat-lock precedent and because the acquire is on a rare path (personal-skill create); if we add lock observability it should land in the shared helper so both sites get it uniformly. Keeping the current minimal function here to match precedent.

backend/onyx/server/features/skill/api.py
rohoswagger: Rationale — a per-user *row-count* cap can't be expressed as a declarative Postgres constraint: CHECK can't aggregate/subquery, and UNIQUE/exclusion constraints can't encode "at most N rows where author=U". The options would be a row-counting trigger or a separate per-user counter table, both heavier and more failure-prone than the advisory lock. The lock-around-count-then-insert is the same pattern the codebase already uses for the tenant seat limit (`acquire_seat_lock`), so I kept it consistent. (Worst case if the lock were ever dropped is a small transient cap overrun, not corruption — but the lock makes it exact.)

backend/onyx/db/skill.py
rohoswagger: Follow-up on the logging point (b): added in d97a9b0ab8. `lock_personal_skills_for_user` now emits a debug log before the blocking `pg_advisory_xact_lock` (the "trying to acquire" signal — the gap until the next line surfaces contention) and one after it's held. Still leaving the generalized helper (a) and context-manager (c) as a cross-cutting follow-up that should also absorb `acquire_seat_lock`.

PR #11962, opened 2026-06-11
fix(web): stop background tabs from polling the API forever

nmgarza5:
## Description

**Bug:** api-server request rate sawtooths between deploys — collapses to ~4 req/s right after each deploy, then climbs continuously to 600–800 req/s until the next one.

**Evidence that the climb is unattended/background tabs** (Prometheus `http_requests_total`, prod):
- The climb runs straight through nights with no diurnal dip (153 → 655 req/s across two consecutive nights, 06-07 → 06-08). Monotonic overnight growth = unattended clients, not users.
- The post-deploy floor is ~4 req/s during business hours — nearly all steady-state traffic is machine-generated.
- The growing handlers map exactly onto the client loops that ignore tab visibility: `/auth/refresh` +59 req/s (`useTokenRefresh`, raw 10-min `setInterval`), cc-pair `index-attempts` / `errors` +27 req/s each (`usePaginatedFetch`, raw 5s `setInterval`), `/me` +29 req/s **including 403s** — expired sessions whose loops kept firing with nobody there to re-login. SWR's own `refreshInterval` already pauses hidden tabs; the climbers are precisely the loops that don't.
- Deploys reset it because long-lived sessions break on the new build, then accumulation restarts.

**Fix:** gate background network loops on tab visibility:
- new `useVisibilityGatedInterval` hook — skips ticks while `document.hidden`, one catch-up tick on return; `usePaginatedFetch` now uses it
- `useTokenRefresh` / `AppHealthBanner` skip hidden ticks; visibilitychange handler + SWR `revalidateOnFocus` catch sessions/data up on return
- `skipRetryOnAuthError` defers scheduled retries while hidden and fires them on return

Visible tabs behave exactly as before. Known limit: clients that are *visible* but unattended (wall monitors, headless automation) are intentionally untouched here; #11963 removes the heaviest per-page polling for that class.

**Verification after deploy:** the post-deploy baseline should hold flat (organic diurnal only) instead of climbing day-over-day — directly visible on the api-server request-rate board.

## Additional Options

### DIFF
--- a/web/src/hooks/usePaginatedFetch.ts
+++ b/web/src/hooks/usePaginatedFetch.ts
@@ -5,6 +5,7 @@ import { usePathname, useRouter, useSearchParams } from "next/navigation";
 import type { Route } from "next";
 
 import { errorHandlingFetcher } from "@/lib/fetcher";
+import { useVisibilityGatedInterval } from "@/hooks/useVisibilityGatedInterval";
 
 // Any type that has an id property
 type PaginatedType = {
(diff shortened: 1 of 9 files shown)


nmgarza5:
@greptile

jmelahman (approved):

PR #11963, opened 2026-06-11
fix(web): poll search settings only while a re-embedding migration is active

nmgarza5:
## Description

**Bug:** the admin Index Settings page polls `/api/search-settings/get-current-search-settings` and `get-secondary-search-settings` every 5s per viewer, unconditionally — even when no migration is running. In prod this amplified into single-tenant storms of 30–45 req/s per endpoint (at peak ~22% of all api-server requests), from clients sitting on the page.

**Fix:**
- `useSecondarySearchSettings` self-throttles: 5s while a migration is in flight (non-null response), 60s idle heartbeat to catch migrations started elsewhere
- `useCurrentSearchSettings` / `useCurrentEmbeddingModel` no longer poll by default; the page passes a 5s interval only while reindexing
- when a migration finishes, the page revalidates the current settings once so the new model shows immediately
- explicit user actions still refresh instantly via the existing `mutate` calls

Idle viewer: 24 req/min → 1 req/min. Active migration UX unchanged.

## Additional Options

### DIFF
--- a/web/src/hooks/useSearchSettings.test.ts
+++ b/web/src/hooks/useSearchSettings.test.ts
@@ -1,0 +1,28 @@
+import { secondaryRefreshInterval } from "@/hooks/useSearchSettings";
+import type { EmbeddingModelResponse } from "@/lib/indexing/interfaces";
+
+const embeddingModelResponse: EmbeddingModelResponse = {
+  model_name: "text-embedding-3-small",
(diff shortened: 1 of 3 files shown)


nmgarza5:
@greptile

jmelahman (approved):

PR #11969, opened 2026-06-12
fix(indexing): cap spawned worker memory + Drive extracted-text size

nmgarza5:
## Description

**What happened**

A cloud prospect connected Google Drive. The first indexing run brought in 932 documents, then every run after failed with "No heartbeat received for 1800 seconds". Heartbeats were not the problem: their Drive is full of very large, image-heavy Docs/Slides, and the indexing worker process grew to 13.5GB of RAM while converting them. The kernel OOM-killed the whole worker pod. That kill also took out the heartbeat thread (producing the misleading error) and killed other tenants' indexing tasks running on the same pod. Every retry resumed into the same files and repeated the cycle.

**How file-size checking works, and the gap**

- When the Drive connector lists files, Google reports a `size` field for regular binary files (PDF, docx, pptx…). Files over `GOOGLE_DRIVE_CONNECTOR_SIZE_THRESHOLD` (10MB default) are skipped before download, and downloads stream in chunks and abort if they grow past the threshold mid-download.
- Google-native files (Docs/Slides/Sheets) have **no `size` field** — they aren't stored as bytes. The threshold check is `if size_str:`, so native files bypass it entirely and we fetch + parse however much content they contain. That is the hole this corpus went through.

**The fix, in three layers**

1. **Cap extracted content** — after converting a file, count the characters of extracted text and drop content past `CONNECTOR_MAX_EXTRACTED_TEXT_CHARS` (10M chars default; connector-agnostic — Drive is the first consumer), with a warning log. Bounds native files the same way the byte threshold bounds binary files.
2. **Memory watchdog** — the watchdog that babysits each spawned indexing worker (5s poll) now also reads the worker's actual RAM (RSS via psutil). Past `INDEXING_WORKER_MEMORY_LIMIT_MB` (opt-in, unset = off) it fails the attempt with a clear, user-visible error and terminates the worker — the pod, the heartbeat, and other tenants' work survive.
3. **Leak forensics** — at 75% of the limit, a one-shot thread inside the worker logs a tracemalloc report of the largest allocation sites (gated by `INDEXING_WORKER_TRACEMALLOC`), so the next incident names the exact code path eating memory.

Both knobs are exposed in the helm chart values. Activating on cloud requires setting `INDEXING_WORKER_MEMORY_LIMIT_MB` in the cloud env configmap (separate PR in cloud-deployment-yamls).

## Additional Options


nmgarza5:
@greptile

nmgarza5:
@greptile

backend/onyx/background/celery/memory_diagnostics.py
justin-tahara, line 1: I feel like we might have some memory files that already exist. I wonder if we can combine or at least consolidate

backend/onyx/configs/app_configs.py
```diff
@@ -852,6 +852,13 @@ def get_current_tz_offset() -> int:
     os.environ.get("GOOGLE_DRIVE_CONNECTOR_SIZE_THRESHOLD", 10 * 1024 * 1024)
 )
 
+# Google-native files (Docs/Slides/Sheets) report no `size` metadata, so they
+# bypass GOOGLE_DRIVE_CONNECTOR_SIZE_THRESHOLD. Cap the total text extracted per
+# file to bound connector memory; content past the cap is dropped with a warning.
+GOOGLE_DRIVE_MAX_EXTRACTED_TEXT_CHARS = int(
+    os.environ.get("GOOGLE_DRIVE_MAX_EXTRACTED_TEXT_CHARS") or 10_000_000
+)
```
justin-tahara: Should this be a google drive specific thing? Ideally we can reuse some values that we have or we make this connector agnostic

backend/onyx/connectors/google_drive/doc_conversion.py
```diff
@@ -681,6 +682,33 @@ def convert_drive_item_to_document(
...
 
 
+def _cap_extracted_text(
+    sections: list[TextSection | ImageSection | TabularSection],
+    file_name: str | None,
+) -> list[TextSection | ImageSection | TabularSection]:
+    """Bound the total text retained per file. Google-native files (Docs/Slides/
+    Sheets) report no `size` metadata, so they bypass the download size threshold
+    and can extract to arbitrarily large text."""
+    remaining = GOOGLE_DRIVE_MAX_EXTRACTED_TEXT_CHARS
```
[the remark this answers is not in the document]
nmgarza5: ugh fixing now

justin-tahara (commented):
Left some comments. Let me know what you think about them

nmgarza5:
@greptile

backend/onyx/configs/app_configs.py
nmgarza5: yeah makes sense

backend/onyx/background/celery/memory_diagnostics.py
nmgarza5: Consolidated — merged into the existing memory_monitoring.py (next to emit_process_memory) and deleted the new file. 813e5fa

backend/onyx/configs/app_configs.py
nmgarza5: Made it connector-agnostic: CONNECTOR_MAX_EXTRACTED_TEXT_CHARS + a shared cap_sections_text in cross_connector_utils/section_utils.py. Drive is the first consumer; other connectors can adopt the same call. 813e5fa

nmgarza5:
@greptile

nmgarza5:
@greptile

backend/onyx/background/celery/tasks/docfetching/tasks.py
```diff
@@ -12,6 +13,8 @@
 from onyx import __version__
 from onyx.background.celery.apps.app_base import task_logger
```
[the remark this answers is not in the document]
nmgarza5: Stale — this reviewed 813e5fa, where the consolidation edits were missing (bad staging on my side). Fixed in 85a4b2a: tasks.py imports memory_monitoring, mypy-check is green on HEAD.

justin-tahara (approved):

PR #11980, opened 2026-06-12
feat(reindex): OpenSearch PIT scan + create-only writes (4/11)

Subash-Mohan:
>
> #11977 — `01-schema`
> #11978 — `02-lifecycle`
> #11979 — `03-defer`
> #11980 — `04-opensearch`  :point_left: **this PR**
> #11981 — `05-reembed`
> #11982 — `06-copy`
> #11983 — `07-swap`
> #11984 — `08-sync`
> #11985 — `09-machinery`
> #11986 — `10-enablement`
> #11987 — `11-tests`

## Description

- Add point-in-time (PIT) scanning to the OpenSearch client (open / iterate / refresh-on-expiry) for consistent reads of live chunks during a port.
- Surface FUTURE-index 404s as `SecondaryIndexDocumentMissingError` and continue past missing docs in bulk updates instead of failing the whole batch.
- Support create-only writes so porting never overwrites live chunks.

## How Has This Been Tested?

**Automated**
- `test_opensearch_client.py` (PIT scan, external-versioning newest-wins, missing-doc surfacing) and `test_index_pair_secondary_missing.py` (FUTURE-404 conversion) against a real OpenSearch.
- `ty` and `ruff` pass.

- Verified a PIT scan pages through all live chunks consistently while concurrent writes occur, and that a missing FUTURE doc is reported rather than fatal.

## Additional Options

### DIFF
--- a/backend/onyx/document_index/chunk_content_enrichment.py
+++ b/backend/onyx/document_index/chunk_content_enrichment.py
@@ -3,10 +3,9 @@ from onyx.configs.constants import RETURN_SEPARATOR
 from onyx.context.search.models import InferenceChunk
 from onyx.context.search.models import InferenceChunkUncleaned
 from onyx.indexing.models import DocAwareChunk
-from onyx.indexing.models import DocMetadataAwareIndexChunk
 
 
-def generate_enriched_content_for_chunk_text(chunk: DocMetadataAwareIndexChunk) -> str:
+def generate_enriched_content_for_chunk_text(chunk: DocAwareChunk) -> str:
     return f"{chunk.title_prefix}{chunk.doc_summary}{chunk.content}{chunk.chunk_context}{chunk.metadata_suffix_keyword}"
 
 
--- a/backend/onyx/document_index/opensearch/client.py
+++ b/backend/onyx/document_index/opensearch/client.py
@@ -1,12 +1,15 @@
 import json
 import logging
 import time
+from collections.abc import Iterator
 from contextlib import AbstractContextManager
 from contextlib import nullcontext
+from http import HTTPStatus
 from typing import Any
 from typing import Generic
 from typing import TypeVar
 
+from opensearchpy import NotFoundError
 from opensearchpy import OpenSearch
 from opensearchpy import TransportError
 from opensearchpy.helpers import bulk
@@ -18,11 +21,18 @@ from onyx.configs.app_configs import OPENSEARCH_ADMIN_USERNAME
 from onyx.configs.app_configs import OPENSEARCH_HOST
 from onyx.configs.app_configs import OPENSEARCH_REST_API_PORT
 from onyx.configs.app_configs import OPENSEARCH_USE_SSL
+from onyx.configs.app_configs import PIT_KEEP_ALIVE
 from onyx.document_index.interfaces_new import TenantState
+from onyx.document_index.opensearch.constants import DEFAULT_MAX_CHUNK_SIZE
 from onyx.document_index.opensearch.constants import OpenSearchSearchType
+from onyx.document_index.opensearch.schema import CHUNK_INDEX_FIELD_NAME
+from onyx.document_index.opensearch.schema import CONTENT_VECTOR_FIELD_NAME
+from onyx.document_index.opensearch.schema import DOCUMENT_ID_FIELD_NAME
 from onyx.document_index.opensearch.schema import DocumentChunk
 from onyx.document_index.opensearch.schema import DocumentChunkWithoutVectors
 from onyx.document_index.opensearch.schema import get_opensearch_doc_chunk_id
+from onyx.document_index.opensearch.schema import MAX_CHUNK_SIZE_FIELD_NAME
+from onyx.document_index.opensearch.schema import TITLE_VECTOR_FIELD_NAME
 from onyx.document_index.opensearch.search import DEFAULT_OPENSEARCH_MAX_RESULT_WINDOW
 from onyx.server.metrics.opensearch_search import observe_opensearch_search
 from onyx.server.metrics.opensearch_search import record_opensearch_search_error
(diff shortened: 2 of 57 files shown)


backend/onyx/document_index/opensearch/client.py
```diff
@@ -108,6 +118,35 @@ class OpenSearchIndexError(Exception):
...
+
+
+# Server-side error.type strings (not exposed as enums by opensearch-py; cf.
+# _RETRYABLE_UPDATE_ERROR_TYPES above). Status codes use http.HTTPStatus.
+_DOCUMENT_MISSING_ERROR_TYPE = "document_missing_exception"
+_VERSION_CONFLICT_ERROR_TYPE = "version_conflict_engine_exception"
+# Raised by a search whose PIT has expired/been deleted; we re-open and retry.
+_SEARCH_CONTEXT_MISSING_ERROR_TYPE = "search_context_missing"
+# Chunks per PIT-scan page. A port doc-batch is small (INDEX_BATCH_SIZE docs), so
+# one page covers a batch; paging still protects against a pathological doc.
```
acaprau, line 147: what does this mean?

backend/onyx/document_index/opensearch/client.py
```diff
@@ -1135,7 +1246,14 @@ def bulk_update_documents(
...
                 err_type = err_obj.get("type", "") if isinstance(err_obj, dict) else ""
 
-                if status >= 500 and err_type in _RETRYABLE_UPDATE_ERROR_TYPES:
+                if (
+                    swallow_document_missing
+                    and status == HTTPStatus.NOT_FOUND
+                    and err_type == _DOCUMENT_MISSING_ERROR_TYPE
+                ):
+                    # doc not in this index yet; surface instead of failing
+                    missing_chunk_ids.append(info.get("_id", ""))
```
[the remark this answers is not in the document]
acaprau: yes, pls address this feedback.

backend/onyx/document_index/opensearch/client.py
```diff
@@ -1444,6 +1565,214 @@ def search_for_document_ids(
         )
         return document_chunk_ids
 
+    def open_pit(self, keep_alive: str = PIT_KEEP_ALIVE) -> str:
+        """Opens a point-in-time (PIT) over this index for a consistent scan.
+
+        The PIT pins the index across searches so concurrent writes don't shift
+        the result set. The caller threads the returned id into
```
acaprau: let's not use the word thread here 😅 maybe "passes"

backend/onyx/document_index/opensearch/client.py
```diff
@@ -1444,6 +1565,214 @@ def search_for_document_ids(
...
+        self,
+        pit_id: str,
+        doc_ids: list[str],
+        search_after: list[object] | None = None,
+        page_size: int = _PIT_SCAN_PAGE_SIZE,
+        keep_alive: str = PIT_KEEP_ALIVE,
+    ) -> tuple[list[DocumentChunkWithoutVectors], list[object] | None, str]:
+        """Fetches one page of regular chunks for a batch of documents from a PIT.
+
+        Filters to regular chunks (max_chunk_size == DEFAULT_MAX_CHUNK_SIZE) so
```
acaprau: this logic can't hurt anything but we don't have large or mini chunks in opensearch indices as of now.

backend/onyx/document_index/opensearch/client.py
```diff
@@ -1444,6 +1565,214 @@ def search_for_document_ids(
...
+            )
+        except NotFoundError as e:
+            if not self._is_pit_expired(e):
+                raise
+            logger.debug(
+                "PIT %s expired mid-scan for index %s; reopening.",
+                pit_id,
+                self._index_name,
+            )
+            pit_id = self.open_pit(keep_alive)
```
acaprau, line 1690: this seems sus, this could be a completely different snapshot of the index. my initial impression is a pit id should never expire and callers have to explicitly close it. also, have you considered what happens if a doc is deleted in the middle of a migration from one embedding model/index to another? it seems like the pid snapshot would capture that deleted doc, and that doc would be moved to the new index, which does not seem good. what are the tradeoffs between the pid approach and a new field in the index which tracks if a chunk was migrated? or better yet perhaps, just use postgres as the source of truth for which docs have yet to be migrated to the new index, and only search for those doc ids, taking the state of the chunk as-is.

backend/onyx/document_index/opensearch/client.py
```diff
@@ -1444,6 +1565,214 @@ def search_for_document_ids(
...
+        self,
+        pit_id: str,
+        doc_ids: list[str],
+        search_after: list[object] | None,
+        page_size: int,
+        keep_alive: str,
+    ) -> dict[str, Any]:
+        """Builds the PIT search body for one page.
+
+        No index= is sent — the PIT pins the index; keep_alive in the pit block
```
acaprau, line 1771: ah ok this partially negates my previous comment, if the pit lease is renewed every time we query then it makes it much less likely for the pit to expire.

backend/onyx/document_index/opensearch/client.py
```diff
@@ -1444,6 +1565,214 @@ def search_for_document_ids(
...
+            "query": {
+                "bool": {
+                    "filter": [
+                        {"terms": {DOCUMENT_ID_FIELD_NAME: doc_ids}},
+                        {"term": {MAX_CHUNK_SIZE_FIELD_NAME: DEFAULT_MAX_CHUNK_SIZE}},
+                    ]
+                }
+            },
+            "sort": [
+                {DOCUMENT_ID_FIELD_NAME: "asc"},
```
acaprau, line 1791: can we not just sort by chunk id? this seems cleaner. chunk id already encapsulates this information.

backend/tests/external_dependency_unit/opensearch/test_opensearch_client.py
```diff
@@ -747,6 +749,181 @@ def fake_bulk(
         with pytest.raises(OpenSearchIndexError, match="does not match"):
             test_client.bulk_index_documents(documents=docs, tenant_state=tenant_state)
 
+    def test_port_create_only_yields_to_forward_write(
```
acaprau, line 752: nit but can we not mention the port/migration here? this abstraction layer should have no knowledge of that, this is just testing a function which tries to create chunks of they don't exist, that's it.

backend/tests/external_dependency_unit/opensearch/test_opensearch_client.py
```diff
@@ -747,6 +749,181 @@ def fake_bulk(
...
+        # 3) The stale port write lands AFTER, now CREATE-ONLY: the chunk already
+        # exists (the forward path owns it), so this is a benign 409 -- no raise,
+        # no overwrite. The port yields.
+        test_client.bulk_index_documents(
+            documents=[stale_snapshot],
+            tenant_state=tenant_state,
+            use_create_only=True,
+        )
+
+        # FIX: the stale port write did not clobber. Fresh content and the
```
acaprau: why does it say FIX? don't think anything here is broken is it?

backend/onyx/document_index/chunk_content_enrichment.py
```diff
@@ -3,10 +3,9 @@
 from onyx.context.search.models import InferenceChunk
 from onyx.context.search.models import InferenceChunkUncleaned
 from onyx.indexing.models import DocAwareChunk
-from onyx.indexing.models import DocMetadataAwareIndexChunk
 
 
-def generate_enriched_content_for_chunk_text(chunk: DocMetadataAwareIndexChunk) -> str:
```
acaprau, line 9: is this right? we need metadata_suffix_keyword and just guessing from the name that sounds like DocMetadataAwareIndexChunk

backend/onyx/document_index/opensearch/client.py
Subash-Mohan: - _SEARCH_CONTEXT_MISSING_ERROR_TYPE — the error string OpenSearch returns when our PIT has expired/been deleted. We match on   it to re-open a fresh PIT and retry, instead of failing.
  - _PIT_SCAN_PAGE_SIZE = 1000 — chunks fetched per page. A port batch is small, so 1000 covers it in one page; we still page so a giant doc can't blow up memory.

backend/onyx/document_index/opensearch/client.py
Subash-Mohan: - Deleted docs can't be resurrected. Right before writing each batch we re-check Postgres via
  `filter_existing_cc_pair_document_ids` — any doc deleted mid-run is dropped, so it never lands in the new index. (That call lives in the write-path branch of this stack, not this diff.)
  - The PIT doesn't just expire on us. We pass keep_alive on every page query, so each scan extends the lease — it only lapses on a genuine stall, and `search_context_missing` re-opens and retries if that happens.
  ### On a per-chunk "migrated" flag vs. the PIT:                                                                                    
 - It'd mean writing a flag back into the index we're retiring (extra writes + a mapping change), and it only buys resumability — which we already get from the `last_processed_doc_id` cursor on the PortAttempt.

backend/onyx/document_index/opensearch/client.py
Subash-Mohan: The chunk id is OpenSearch's _id, which can't be sorted on (no doc_values). document_id + chunk_index are already sortable fields with the same values — so we sort on those.

PR #11981, opened 2026-06-12
feat(reindex): re-embedding pipeline for the port flow (5/11)

Subash-Mohan:
>
> #11977 — `01-schema`
> #11978 — `02-lifecycle`
> #11979 — `03-defer`
> #11980 — `04-opensearch`
> #11981 — `05-reembed`  :point_left: **this PR**
> #11982 — `06-copy`
> #11983 — `07-swap`
> #11984 — `08-sync`
> #11985 — `09-machinery`
> #11986 — `10-enablement`
> #11987 — `11-tests`

## Description

- Add the re-embedding pipeline that rebuilds embeddings for ported chunks under the new model settings (model-only and full augmentation / contextual-RAG strategies), reusing existing chunk text and metadata.
- Thread `index_to_secondary` through the indexing pipeline/adapters and switch contextual-RAG LLM selection to the search-settings-aware factory.

## How Has This Been Tested?

**Automated**
- `test_port_reembed.py` (strategy selection, identity-based embedding match) and `test_indexing_pipeline.py`.
- `ty` and `ruff` pass.

- Re-embedded ported chunks with a changed embedding model, and again with contextual-RAG enabled; confirmed embeddings are rebuilt and chunk identities line up.

## Additional Options

### DIFF
--- a/backend/onyx/indexing/port_reembed.py
+++ b/backend/onyx/indexing/port_reembed.py
@@ -53,7 +53,7 @@ from onyx.document_index.opensearch.schema import DocumentChunkWithoutVectors
 
 # Reused from the chunker (not replicated) so the rebuilt tail stays byte-identical
 # to indexing and the skip test mirrors it exactly; replicating either risks drift.
-from onyx.indexing.chunker import _get_metadata_suffix_for_document_index
+from onyx.indexing.chunker import get_metadata_suffix_for_document_index
 from onyx.indexing.chunker import MAX_METADATA_PERCENTAGE
 from onyx.indexing.embedder import IndexingEmbedder
 from onyx.indexing.models import DocAwareChunk
@@ -118,7 +118,7 @@ def rebuild_semantic_tail(chunk: DocumentChunkWithoutVectors) -> str:
     if not chunk.metadata_list:
         return ""
     metadata = convert_metadata_list_of_strings_to_dict(chunk.metadata_list)
-    semantic_suffix, _ = _get_metadata_suffix_for_document_index(
+    semantic_suffix, _ = get_metadata_suffix_for_document_index(
         metadata, include_separator=True
     )
     return semantic_suffix
--- a/backend/onyx/indexing/port_reembed.py
+++ b/backend/onyx/indexing/port_reembed.py
@@ -122,21 +122,23 @@ def rebuild_semantic_tail(chunk: DocumentChunkWithoutVectors) -> str:
 
 
 def _semantic_tail_was_embedded(
-    semantic_tail: str, max_chunk_size: int, tokenizer: BaseTokenizer
+    semantic_tail: str, max_chunk_size: int, present_tokenizer: BaseTokenizer
 ) -> bool:
     """Whether indexing embedded the semantic tail. The chunker drops it from the
     embedding when it exceeds MAX_METADATA_PERCENTAGE of the chunk budget, yet still
     stores the keyword tail on `content` — so recovery must reproduce the skip or it
-    re-embeds metadata the vector never saw. `tokenizer` is the FUTURE model's, a
-    coarse proxy for the PRESENT one that the 25% threshold tolerates."""
+    re-embeds metadata the vector never saw. Both inputs must be PRESENT-side to
+    match the index-time decision: `max_chunk_size` is the stored PRESENT budget and
+    `present_tokenizer` is the PRESENT model's — the FUTURE model's can count the
+    tail differently and flip the threshold."""
     if not semantic_tail:
         return False
-    metadata_tokens = len(tokenizer.encode(semantic_tail))
+    metadata_tokens = len(present_tokenizer.encode(semantic_tail))
     return metadata_tokens < max_chunk_size * MAX_METADATA_PERCENTAGE
 
 
 def recover_embedding_input(
-    chunk: DocumentChunkWithoutVectors, tokenizer: BaseTokenizer
+    chunk: DocumentChunkWithoutVectors, present_tokenizer: BaseTokenizer
 ) -> str:
     """Rebuild the text that was actually embedded when this chunk was indexed.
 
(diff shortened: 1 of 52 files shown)


backend/onyx/indexing/port_reembed.py
```diff
@@ -0,0 +1,390 @@
...
+  enriched text — the title prefix, doc summary and chunk context are kept,
+  because the stored vector encoded all of them and the new model must encode
+  the same text. The only catch is that the stored `content` ends with the
+  *keyword* metadata tail while indexing embedded the *semantic* tail, so we
+  swap just that tail back.
+- AUGMENTATION (contextual-RAG toggle or model changed): the enriched text
+  itself changes, so we strip the stored augmentation back to the bare chunk
+  text, re-glue under FUTURE settings, then re-embed. Two sub-cases keyed on the
+  FUTURE `enable_contextual_rag`:
+    * FUTURE RAG off — strip the doc summary / chunk context / title / metadata
```
acaprau, line 17: you mean contextual rag right? i'm writing this before i've looked at the code before but does this logic hold if PRESENT also does not use contextual rag? there would be nothing to strip then right?

backend/onyx/indexing/port_reembed.py
```diff
@@ -0,0 +1,390 @@
...
+from onyx.document_index.chunk_content_enrichment import (
+    generate_enriched_content_for_chunk_text,
+)
+from onyx.document_index.opensearch.schema import DocumentChunk
+from onyx.document_index.opensearch.schema import DocumentChunkWithoutVectors
+
+# The chunker's canonical metadata-suffix builder. Reused (rather than
+# replicated) so the rebuilt semantic tail is byte-identical to what indexing
+# produced; replicating it would risk silent drift in the embedding input.
+from onyx.indexing.chunker import _get_metadata_suffix_for_document_index
```
acaprau: since this is no longer constrained to just its module this function should not start with a _

backend/onyx/indexing/port_reembed.py
```diff
@@ -0,0 +1,390 @@
...
+from onyx.connectors.models import TextSection
+from onyx.db.models import SearchSettings
+from onyx.document_index.chunk_content_enrichment import cleanup_content_for_chunks
+from onyx.document_index.chunk_content_enrichment import (
+    generate_enriched_content_for_chunk_text,
+)
+from onyx.document_index.opensearch.schema import DocumentChunk
+from onyx.document_index.opensearch.schema import DocumentChunkWithoutVectors
+
+# The chunker's canonical metadata-suffix builder. Reused (rather than
```
acaprau: this comment reads a little bit like ai slop; opus 4.8 really likes writing comments where they're not needed, imo we can get rid of this.

backend/onyx/indexing/port_reembed.py
```diff
@@ -0,0 +1,390 @@
...
+    text changes), otherwise MODEL_ONLY. A change in
+    `contextual_rag_model_configuration_id` only matters when contextual RAG is
+    on in present or future — if it is off in both, no enrichment exists in
+    either index, so a stale model-id difference must not force AUGMENTATION.
+    Model/prefix/normalize/dimension and multipass changes only alter the vectors
+    (or large/mini chunks the port doesn't read), so they fall through to
+    MODEL_ONLY."""
+    rag_relevant = present_ss.enable_contextual_rag or future_ss.enable_contextual_rag
+    augmentation_changed = (
+        present_ss.enable_contextual_rag != future_ss.enable_contextual_rag
```
acaprau, line 98: similar question to my comment at the top of this file, what if we turn off contextual rag in FUTURE? is it still appropriate to return AUGMENTATION from this function?

backend/onyx/indexing/port_reembed.py
```diff
@@ -0,0 +1,390 @@
...
+    """
+    keyword_metadata = chunk.metadata_suffix or ""
+    if not keyword_metadata:
+        return chunk.content
+    without_metadata = chunk.content.removesuffix(keyword_metadata)
+    # If nothing was removed, content didn't end with the stored metadata, so
+    # return it unchanged rather than appending a second metadata block.
+    if without_metadata == chunk.content:
+        return chunk.content
+    return without_metadata + rebuild_semantic_tail(chunk)
```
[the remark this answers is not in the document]
acaprau: take a look at this comment, if true then this is not ideal.

backend/onyx/indexing/port_reembed.py
Subash-Mohan: This only reaches the strip path when the Contextual RAG toggle actually differs; if PRESENT and FUTURE are both off, select_reembed_strategy returns MODEL_ONLY (the rag_relevant gate at port_reembed.py:99-112), so we never strip in the both-off case. 
 
 And when PRESENT is off → FUTURE on, you're right that there's nothing Contextual RAG-wise to strip: doc_summary/chunk_context are stored empty, and _remove_contextual_rag is guarded on them being non-empty (chunk_content_enrichment.py:91,94), so it no-ops. The strip still removes the title prefix + keyword metadata suffix (which are present regardless of Contextual RAG) to get the bare chunk text, and _augmentation_reembed then generates the fresh doc summary / chunk context under FUTURE settings. So the off→on case rebuilds correctly.

backend/onyx/indexing/port_reembed.py
Subash-Mohan: It depends on PRESENT, and both outcomes are intended:
 - PRESENT contextual RAG on → FUTURE off: returns AUGMENTATION, and that's correct. "AUGMENTATION" here means "the embedded text changed," not "enrichment was added" — turning contextual RAG off still changes the text, because PRESENT's stored content has the doc summary + chunk context baked in. We have to strip those and re-embed the bare chunk, which is the future_rag_on=False sub-case in _augmentation_reembed (strip only, no LLM). A plain MODEL_ONLY re-embed would wrongly keep the stale contextual-RAG summaries in the vector.
 - PRESENT contextual RAG off → FUTURE off: returns MODEL_ONLY, since rag_relevant is False and nothing enrichment-wise differs (a stale contextual-RAG model-id diff is intentionally ignored). So the only case where turning contextual RAG off in FUTURE yields AUGMENTATION is when PRESENT had it on — which is exactly when there's stale contextual-RAG enrichment to strip.

PR #11986, opened 2026-06-12
feat(reindex): enable the port flow (queue, seeding, deploy) (10/11)

Subash-Mohan:
> **Stacked PR — review in order. Base of the stack: `main`.**
>
> #11977 — `01-schema`
> #11978 — `02-lifecycle`
> #11979 — `03-defer`
> #11980 — `04-opensearch`
> #11981 — `05-reembed`
> #11982 — `06-copy`
> #11983 — `07-swap`
> #11984 — `08-sync`
> #11985 — `09-machinery`
> #11986 — `10-enablement`  :point_left: **this PR**
> #11987 — `11-tests`

## Description

- Turn the flow on: seed a synthetic port attempt from the search-settings endpoint, register the `port` queue/workers (supervisord, helm, dev scripts, celery apps), and surface contextual-RAG validation in the admin re-embed UI.

## How Has This Been Tested?

**Automated**
- Wiring covered by `test_port_wiring.py` (PR 11); web type-check for the admin page.
- `ty` and `ruff` pass.

**Manual** (verified end-to-end on a local stack; this is a stacked feature, so manual scenarios exercise the full flow once the stack is assembled):
- Started a re-embedding migration from the admin UI with `use_port_flow` enabled and confirmed the port queue/worker pick it up and run end-to-end.

## Additional Options

- [ ] [Optional] Please cherry-pick this PR to the latest release version.
- [x] [Optional] Override Linear Check

---
## Summary by cubic
Enables the port reindex flow end-to-end. Adds a server‑controlled `use_port_flow`, atomic FUTURE create-and-seed, `port` queue/workers, stricter Contextual Retrieval checks, and comprehensive tests (E2E, lifecycle, and Celery wiring).

- **New Features**
  - Seed synthetic index attempts for FUTURE on save to resume from PRESENT’s poll cursor; skip `INSTANT`; respect `ACTIVE_ONLY`; ignore `DELETING` pairs; commit FUTURE and seeds together; `use_port_flow` is server-set (not user-editable) and enabled for new FUTURE settings.
  - Register `onyx.background.celery.tasks.port` and add `port` to light worker queues in VS Code, dev script, supervisord, and Helm; unit tests guard beat schedule, queue consumption, and task registration.
  - Cancel active port attempts when a FUTURE is superseded or reindex is canceled.
  - Contextual Retrieval enforcement: require a model when enabled (server validation and Admin UI error/disabled “Apply & Re-index”).
[...]

PR #11987, opened 2026-06-12
test(reindex): end-to-end port flow + lifecycle suites (11/11)

Subash-Mohan:
>
> #11977 — `01-schema`
> #11978 — `02-lifecycle`
> #11979 — `03-defer`
> #11980 — `04-opensearch`
> #11981 — `05-reembed`
> #11982 — `06-copy`
> #11983 — `07-swap`
> #11984 — `08-sync`
> #11985 — `09-machinery`
> #11986 — `10-enablement`
> #11987 — `11-tests`  :point_left: **this PR**

## Description

- Add the end-to-end and lifecycle test suites for the whole flow: `test_port_flow_e2e.py`, `test_port_wiring.py`, and the full `test_reindex_port.py` DB lifecycle suite.

## How Has This Been Tested?

**Automated**
- The suites run against real Postgres + OpenSearch (external-dependency-unit) and as unit tests.
- `ty` and `ruff` pass.

- Full end-to-end run on a local stack (Postgres + OpenSearch): enable -> synthetic seed -> PIT copy + re-embed -> deferred-sync drain -> gated swap.

## Additional Options

### DIFF
--- a/backend/tests/external_dependency_unit/db/test_port_flow_e2e.py
+++ b/backend/tests/external_dependency_unit/db/test_port_flow_e2e.py
@@ -1,0 +1,378 @@
+"""End-to-end composition test for the reindex port flow (MODEL_ONLY strategy).
+
+Proves the layers COMPOSE with NO PortCopier mocking: check_for_port kickoff ->
+run_port_attempt (real PIT scan of PRESENT + real re-embed under the FUTURE model
+via the local model server + create-only write to FUTURE) -> the port-aware
+swap promotion. Asserts the FUTURE chunks carry the seeded content with a freshly
+re-embedded 768-dim vector (different from the seeded placeholder), and that the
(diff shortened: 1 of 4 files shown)


acaprau (approved):
can you also please add tests for the cases where we're trying to index a doc at the same time that a port is happening, and also when we're trying to update metadata at the same time as a port?

Subash-Mohan:
Added the two tests requested:
- test_forward_write_wins_over_concurrent_port_copy — indexing a doc while a port is in flight: the live forward write of the doc overwrites the port's stale copy in FUTURE (fresh content + ACL win; the port's create-only yields).
- test_acl_update_during_port_applies_to_both_indices — updating permissions while a port is in flight: for a doc already copied to FUTURE, the ACL update lands on both PRESENT and FUTURE with no false defer.

PR #12002, opened 2026-06-12
feat(craft): prune session snapshots on write (keep only the latest)

rohoswagger:
## Description

Snapshot storage grew unbounded, for two reasons:

- Snapshot creation minted a **fresh path + new DB row every idle reap** (it never replaced the prior one), so each session accumulated a snapshot per sleep cycle.
- The `cleanup_old_snapshots` task that the Celery beat schedule referenced was **never actually defined**, so it silently no-op'd. The `delete_old_snapshots` helper also only deleted DB rows, never the file-store blobs — so even a working task would have leaked storage.

This PR makes the policy **exactly one snapshot per session** and enforces it at the source via **prune-on-write** (rather than a periodic sweeper):

- After a session's new snapshot is persisted in `cleanup_idle_sandboxes_task`, `_prune_prior_session_snapshots` deletes that session's prior snapshots (blob then row). The new snapshot is fully durable *before* any old one is removed, so a session is never left without a restorable snapshot, and storage never accumulates.
- Each reap prunes **all** of a session's priors, so any backlog (e.g. rows left by a past partial failure, or the pre-existing accumulation) self-heals on that session's next reap.
- `SnapshotManager.delete_snapshot` is now **idempotent** (`delete_file(error_on_missing=False)`): a row whose blob is already gone still gets dropped; a genuine delete failure (e.g. S3 unreachable) leaves the row in place to retry next reap, so a blob and its row never leak out of sync.
- **Removes** the never-defined `cleanup_old_snapshots` task, its beat-schedule entry, and the now-unused `CLEANUP_OLD_SNAPSHOTS` / `CLEANUP_OLD_SNAPSHOTS_BEAT_LOCK` constants.

No DB migration, no config knobs (keep-latest is a hard rule), no periodic task. Restart Celery workers to pick up the new reap behavior.

Docs: `docs/craft/infra/snapshot-retention.md`.

## How Has This Been Tested?

- **Unit test** (`test_prune_prior_snapshots.py`): verifies blob-then-row deletion per prior snapshot, that an empty prior list is a no-op, and that a failed blob delete keeps that row while still pruning the rest.
- Verified `ty` + `ruff` clean; the idle task imports, `cleanup_old_snapshots_task` is gone, `beat_schedule` imports cleanly.

## Additional Options

### DIFF
--- a/backend/onyx/server/features/build/sandbox/tasks/tasks.py
+++ b/backend/onyx/server/features/build/sandbox/tasks/tasks.py
@@ -1,7 +1,5 @@
 """Celery tasks for sandbox operations (cleanup, etc.)."""
 
-from typing import TYPE_CHECKING
-
 from celery import shared_task
 from celery import Task
 from redis.lock import Lock as RedisLock
@@ -11,6 +9,7 @@ from onyx.background.celery.apps.app_base import task_logger
 from onyx.configs.constants import OnyxCeleryTask
 from onyx.configs.constants import OnyxRedisLocks
 from onyx.db.engine.sql_engine import get_session_with_current_tenant
+from onyx.db.models import Snapshot
 from onyx.redis.redis_pool import get_redis_client
 from onyx.redis.redis_tenant_work_gating import maybe_mark_tenant_active
 from onyx.server.features.build.configs import SANDBOX_IDLE_TIMEOUT_SECONDS
@@ -19,12 +18,7 @@ from onyx.server.features.build.db.build_session import (
     mark_user_sessions_idle__no_commit,
 )
 from onyx.server.features.build.sandbox.base import get_sandbox_manager
-
-if TYPE_CHECKING:
-    from onyx.db.models import Snapshot
-    from onyx.server.features.build.sandbox.manager.snapshot_manager import (
-        SnapshotManager,
-    )
+from onyx.server.features.build.sandbox.manager.snapshot_manager import SnapshotManager
 
 # 100 minutes - snapshotting can take time
 TIMEOUT_SECONDS = 6000
@@ -32,8 +26,8 @@ TIMEOUT_SECONDS = 6000
 
 def _prune_prior_session_snapshots(
     db_session: Session,
-    snapshot_manager: "SnapshotManager",
-    prior_snapshots: list["Snapshot"],
+    snapshot_manager: SnapshotManager,
+    prior_snapshots: list[Snapshot],
 ) -> None:
     """Delete a session's now-superseded snapshots (blob then row).
 
@@ -94,9 +88,6 @@ def cleanup_idle_sandboxes_task(self: Task, *, tenant_id: str) -> None:  # noqa:
         from onyx.server.features.build.db.sandbox import get_snapshots_for_session
         from onyx.server.features.build.db.sandbox import (
             update_sandbox_status__no_commit,
-        )
-        from onyx.server.features.build.sandbox.manager.snapshot_manager import (
-            SnapshotManager,
         )
 
         sandbox_manager = get_sandbox_manager()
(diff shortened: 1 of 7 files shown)


backend/onyx/server/features/build/sandbox/tasks/tasks.py
```diff
@@ -203,3 +203,71 @@ def cleanup_idle_sandboxes_task(self: Task, *, tenant_id: str) -> None:  # noqa:
...
+                deleted += 1
+
+            db_session.commit()
+            task_logger.info(f"Pruned {deleted} snapshots")
+
+    except Exception:
+        task_logger.exception("Error in cleanup_old_snapshots_task")
+        raise
+
+    finally:
```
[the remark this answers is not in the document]
rohoswagger: Good catch — addressed in two parts:

1. `SnapshotManager.delete_snapshot` is now **idempotent** (`delete_file(error_on_missing=False)`), so a row whose blob was already removed by a prior partial run still gets its row dropped instead of re-raising forever.
2. The two-phase periodic cleanup this comment targeted has been **removed entirely**. We now keep exactly one snapshot per session and **prune on write**: right after a session's new snapshot is written, `_prune_prior_session_snapshots` deletes the prior ones (blob then row). Because each reap prunes *all* of a session's priors, any rows left by a partial failure are cleaned on that session's next reap — self-healing, with no periodic GC and no accumulation.

wenxi-onyx:
@greptileai

backend/onyx/server/features/build/sandbox/tasks/tasks.py
```diff
@@ -17,13 +20,39 @@
 )
 from onyx.server.features.build.sandbox.base import get_sandbox_manager
 
-# Snapshot retention period in days
-SNAPSHOT_RETENTION_DAYS = 30
+if TYPE_CHECKING:
+    from onyx.db.models import Snapshot
```
wenxi-onyx: Why can't this be imported normally?

backend/onyx/server/features/build/sandbox/tasks/tasks.py
```diff
@@ -59,13 +88,19 @@ def cleanup_idle_sandboxes_task(self: Task, *, tenant_id: str) -> None:  # noqa:
     try:
         # Import here to avoid circular imports
         from onyx.db.enums import SandboxStatus
+        from onyx.file_store.file_store import get_default_file_store
```
wenxi-onyx, line 85: nit: no need to fix in this PR, but the fact that we have so many circular import issues to me feels like we have not written/separated the modules correctly.

wenxi-onyx (approved):

backend/onyx/server/features/build/sandbox/tasks/tasks.py
rohoswagger: It can — fixed in cafb8ce. I verified there's no actual circular import (imported `onyx...tasks.tasks` and the `heavy` worker app fresh, both load cleanly), so I dropped the `TYPE_CHECKING` block and now import `Snapshot` and `SnapshotManager` normally at module top. The helper's annotations are de-quoted accordingly.

backend/onyx/server/features/build/sandbox/tasks/tasks.py
rohoswagger: Agreed — it's a real smell. I removed the need for the type-only block (see the sibling thread) and my newly-added imports are now at module top. The remaining inline block here is pre-existing (it predates this PR), so per your "no need to fix in this PR" note I've left it rather than expand scope — but you're right that the recurring need for inline imports points to module-boundary debt (the `tasks` ↔ `db.sandbox`/`manager` layering) worth untangling separately.

PR #12003, opened 2026-06-12
fix(security): harden default MinIO/S3 object-storage credentials

jmelahman:
## Description

Onyx shipped the well-known `minioadmin:minioadmin` default for MinIO/S3 object storage throughout the deployment layer. Any deployment that didn't explicitly override it ran on universally-known credentials with no warning, so anyone able to reach MinIO could list, download, or modify stored files (uploaded documents, file-store objects).

The exploitability comes from **predictability**, so this change randomizes credentials per-deployment on the standard install path, fails closed in production (compose **and** Helm), and warns at runtime to cover every other path. It implements the three recommended remediations:

**① Generate strong random MinIO credentials at install time**
- `install.sh` auto-generates a matched access-key/secret pair on a fresh install (mirroring the existing `USER_AUTH_SECRET` generation) and writes it to both `MINIO_ROOT_USER`/`MINIO_ROOT_PASSWORD` and the matching `S3_AWS_ACCESS_KEY_ID`/`S3_AWS_SECRET_ACCESS_KEY` (the app authenticates to MinIO with its root credentials, so the pairs stay identical).

**② Don't publish MinIO ports on `0.0.0.0` in production**
- Already satisfied: none of the `docker-compose.prod*.yml` files publish MinIO ports. Only the dev/test files do (`dev.yml`, `multitenant-dev.yml`, `search-testing.yml`, `restart_containers.sh`), which bind `0.0.0.0`. Those carry `TODO(security)` comments to bind `127.0.0.1` later — deferred to avoid changing remote-host dev workflows.

**③ Remove default passwords from Helm `values.yaml`**
- `values.yaml` now ships **empty** objectstorage credentials, so `helm install`/`upgrade` fails closed via the existing `auth-secrets.yaml` required-value check unless the operator provides values or an `existingSecret`/`externalSecret`. CI keeps working by supplying explicit values in `values-ci.yaml` (craft-k8s render) and `ci/ct-values.yaml` (chart-testing), plus placeholder `--set` flags in the `helm lint` step.

**Defense-in-depth (all paths):** `app_configs.py` logs a warning when an S3 endpoint is configured and either credential is still `minioadmin` — catching compose-up-direct, Helm, and manual deployments. Real AWS S3 (no endpoint URL) never trips the check. `env.template` and the `docker-compose.yml` production checklist also document the risk.

Dev/CI/test compose surfaces intentionally keep `minioadmin` (local-only, never network-exposed).

## Behavior changes (by scenario)

- **Local dev (base/dev/CI compose, `restart_containers.sh`):** unchanged — MinIO still comes up with `minioadmin:minioadmin`; only a new startup log warning when an S3 endpoint + default creds are detected.
- **Fresh `install.sh` install:** credentials are now random/matched — read them from the generated `.env`. Existing `.env` on upgrade is untouched.
- **Production compose (`docker-compose.prod*.yml`):** now **fails closed** — `up` errors immediately if the four credential vars aren't set. The only potentially-breaking change, and only if you relied on the implicit default.
- **Helm:** now **fails closed** too — a default `helm install` errors with a required-value message unless you set the objectstorage creds or an `existingSecret`/`externalSecret`. CI is unaffected (explicit values supplied).
- **Touching this code:** new testable helper `_uses_default_object_storage_credentials(endpoint, access_key, secret_key)` in `app_configs.py` with a unit test. The warning only fires when an endpoint URL is present (real-AWS path is exempt).

## How Has This Been Tested?

- New unit test `backend/tests/unit/onyx/configs/test_object_storage_credentials.py` (4 cases) — passes.
- Verified the backend warning fires for `minioadmin` + MinIO endpoint and stays **silent** for real AWS creds (no endpoint).
- Simulated the `install.sh` sed flow: access-key == root-user, secret == root-password, non-`minioadmin`, meets MinIO length minimums.
- Traced the Helm `auth-secrets.yaml` fail-closed logic; updated `values-ci.yaml`, `ci/ct-values.yaml`, and the `helm lint --set` so CI renders pass. (Live `helm` not available in the dev container — CI is the authoritative render check.)
- `pre-commit` on all changed files: `ty`, `ruff`/`ruff format`, `Check YAML`, `Lint GitHub Actions`, `shellcheck`, `zizmor`, `ripsecrets` all pass.

## Additional Options

### DIFF
--- a/deployment/helm/charts/onyx/Chart.yaml
+++ b/deployment/helm/charts/onyx/Chart.yaml
@@ -5,7 +5,7 @@ home: https://www.onyx.app/
 sources:
   - "https://github.com/onyx-dot-app/onyx"
 type: application
-version: 0.5.26
+version: 0.5.27
 appVersion: latest
 annotations:
   category: Productivity
--- a/deployment/helm/charts/onyx/Chart.yaml
+++ b/deployment/helm/charts/onyx/Chart.yaml
@@ -5,7 +5,7 @@ home: https://www.onyx.app/
 sources:
   - "https://github.com/onyx-dot-app/onyx"
 type: application
-version: 0.5.27
+version: 0.6.0
 appVersion: latest
 annotations:
   category: Productivity
--- a/.github/workflows/pr-helm-chart-testing.yml
+++ b/.github/workflows/pr-helm-chart-testing.yml
@@ -130,8 +130,7 @@ jobs:
           echo "=== Validating chart dependencies ==="
           cd deployment/helm/charts/onyx
           helm dependency update
-          # objectstorage ships no defaults (fail closed in values.yaml), so lint
-          # must supply placeholder credentials the same way userauth does.
+          # objectstorage ships no defaults (fail closed), so lint must set placeholders.
           helm lint . \
             --set auth.userauth.values.user_auth_secret=placeholder \
             --set auth.objectstorage.values.s3_aws_access_key_id=placeholder \
--- a/.github/workflows/pr-helm-chart-testing.yml
+++ b/.github/workflows/pr-helm-chart-testing.yml
@@ -130,7 +130,14 @@ jobs:
           echo "=== Validating chart dependencies ==="
           cd deployment/helm/charts/onyx
           helm dependency update
-          helm lint . --set auth.userauth.values.user_auth_secret=placeholder
+          # objectstorage ships no defaults (fail closed in values.yaml), so lint
+          # must supply placeholder credentials the same way userauth does.
+          helm lint . \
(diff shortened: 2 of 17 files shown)


justin-tahara (approved):
Let's also update Chart.yaml version

deployment/helm/charts/onyx/Chart.yaml
```diff
@@ -5,7 +5,7 @@ home: https://www.onyx.app/
 sources:
   - "https://github.com/onyx-dot-app/onyx"
 type: application
-version: 0.5.26
+version: 0.5.27
```
jmelahman:
```suggestion
version: 0.6.0
```

PR #12005, opened 2026-06-12
fix(users): bound invite request with a timeout so the modal can't hang

nmgarza5:
## Description

**Bug:** Clicking Invite in the admin Users modal can freeze the dialog indefinitely — every control (Invite, Cancel, X, backdrop) is disabled while submitting, and no error ever appears. Reported by a customer right after a seat purchase.

**Why:** Seat-limit enforcement on `PUT /manage/admin/users` can stall server-side (control-plane + Stripe round-trips under a per-tenant lock). The frontend fetch has no deadline, so a stalled request leaves `isSubmitting` stuck forever and the existing error toast never fires.

**Fix:** Abort the invite fetch after 30s (`AbortSignal.timeout`) and map the timeout to a clear message through the existing toast path. The modal unfreezes and tells the user the invites may still be processing.

The server-side lock-over-network-calls issue is the root cause and worth its own fix; this bounds the user-facing damage.

## Additional Options

### DIFF
--- a/web/src/refresh-pages/admin/UsersPage/InviteUsersModal.tsx
+++ b/web/src/refresh-pages/admin/UsersPage/InviteUsersModal.tsx
@@ -2,7 +2,7 @@
 
 import { useState, useCallback } from "react";
 import { Button } from "@opal/components";
-import { SvgUsers, SvgAlertTriangle } from "@opal/icons";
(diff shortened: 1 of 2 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

raunakab (approved):

PR #12009, opened 2026-06-12
ci(cli): scan release wheels with govulncheck

wenxi-onyx:
## Description

Adds a post-publish vulnerability scan to the CLI release workflow for the PyPI wheel artifacts that users install.

The PyPI release job now uploads the built wheels as a short-lived workflow artifact before publishing. After `uv publish` succeeds, a new `govulncheck` job downloads those wheels, extracts the bundled `onyx-cli` Go binaries, scans each binary with pinned `govulncheck@v1.3.0` in binary mode, merges the per-wheel SARIF into one upload with distinct automation IDs, and uploads the SARIF results to the GitHub Security tab.

The scan step preserves SARIF output even when vulnerabilities are found, so the release workflow can fail while still publishing findings for review.

This replaces the earlier Docker-image Trivy scan approach because the primary CLI distribution path is PyPI, not `docker.io/onyxdotapp/onyx-cli`.

## Additional Options

### DIFF
--- a/.github/workflows/release-cli.yml
+++ b/.github/workflows/release-cli.yml
@@ -29,6 +29,14 @@ jobs:
             done
           done
         working-directory: cli
+      - name: Upload built wheels
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a
+        with:
+          name: onyx-cli-wheels
+          path: cli/dist/*.whl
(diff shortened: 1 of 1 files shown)


wenxi-onyx:
@greptileai

wenxi-onyx:
@greptileai

rohoswagger (approved):

PR #12044, opened 2026-06-12
fix(sandbox): update initial dependencies

wenxi-onyx:
## Description

Update the sandbox image's preinstalled Python dependency set while keeping the existing Debian/glibc base and release workflow unchanged.

This includes `onyx-cli==1.1.1`, which carries the CLI-side dependency fixes, and keeps the wheel-only validation note for both supported manylinux targets.

## Additional Options


wenxi-onyx:
@greptileai re-review

rohoswagger (approved):

PR #12059, opened 2026-06-14
fix(google_drive): bound connector memory by streaming documents in sub-batches

nmgarza5:
## Description

**Bug:** Large Google Drive connectors OOM'd the indexing worker — it drained the whole drive into one list, converted every file 8-wide, and buffered all documents before yielding, so peak memory scaled with drive size. Real tenant: climbed to 9+ GB, worker killed, 0 docs indexed on every retry.

**Fix:** Convert and yield documents in bounded sub-batches of `DRIVE_CONVERSION_BATCH_SIZE` (50), so resident documents stay ~500 MB regardless of drive size (~10 MB/doc observed).

**Hierarchy ordering preserved.** A document must not emit before its ancestor folder node (downstream resolves the doc→folder link from a cache populated by earlier-yielded nodes). Sub-batching can split a folder's node from its files, so a file whose parent folder isn't resolved yet is parked — keyed by folder id, lightweight metadata, never documents — and released in O(1) the moment that node is emitted. Folders reachable by nobody fall back to the source root at stream end, matching the pre-batching single-pass behavior. The park queue is per-checkpoint, so it never grows with drive size.

Also removes a dead `DRIVE_BATCH_SIZE` constant.

## How Has This Been Tested?

- **Unit tests (7):** bounded sub-batch conversion (incl. force-flush), incremental streaming, ancestor-node-before-document ordering, folder-keyed deferral + release, and that skipped (None) conversions are dropped not counted.
- **Memory harness:** ran the real connector over up to 4,000 simulated documents — peak stayed flat (~250 MB normal, ~500 MB worst-case all-deferred) vs ~5–20 GB for the old code.
- Reviewed via 3 sequential recall passes (no surviving correctness findings). Not yet run against a live Google Drive end-to-end (none configured locally).

## Additional Options

### DIFF
--- a/backend/onyx/connectors/google_drive/connector.py
+++ b/backend/onyx/connectors/google_drive/connector.py
@@ -1749,13 +1749,17 @@ class GoogleDriveConnector(
         """Resolve ancestors for this sub-batch's NEW files and convert the files
         whose parent hierarchy node is now available, nodes first.
 
-        A file whose immediate parent folder has not been emitted yet (a folder
-        only a not-yet-processed user can reach) is parked in ``pending_by_folder``
-        under that folder id, so its document is never emitted before its ancestor
-        node. It is released the moment that folder's node is emitted by some other
-        file's walk, so the per-sub-batch cost is independent of the queue size.
-        ``force_flush`` drains everything still parked — those folders never
-        resolved, so the files fall back to the source root."""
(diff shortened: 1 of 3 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

justin-tahara (approved):

PR #12060, opened 2026-06-14
feat: add DISABLE_USER_IDENTITY_IN_PROMPT to keep user name/email out of LLM prompts

### LINKED ISSUE #11929: [Privacy/Bug] User name and email are injected into LLM system prompt without opt-out
### Summary

Onyx appears to inject user-identifying information, including the user's name and email address, into the system prompt that is sent to the configured LLM provider.

Specifically, the prompt template in:

https://github.com/onyx-dot-app/onyx/blob/main/backend/onyx/prompts/user_info.py

includes fields such as:

- `User name: {user_name}`
- `User email: {user_email}`

When Onyx is configured to use an external LLM provider, this means user PII may be transmitted to that provider as part of the prompt body.

### Why this is a privacy concern

This seems unnecessary for many deployments and creates privacy/compliance concerns, especially for organizations using external LLM providers.

Onyx already has an opt-in setting for forwarding user/session metadata to LLM providers:

- `SEND_USER_METADATA_TO_LLM_PROVIDER`

However, the same type of privacy control does not appear to exist for user PII embedded directly in the system prompt.

So there appears to be an inconsistency:

| Data channel | Default behavior | Admin control? |
|---|---:|---:|
| LLM request metadata / user/session IDs | Off by default | Yes |
| User name/email in prompt body | Sent by default | No apparent opt-out |

### Current behavior

User name and email can be included in the system prompt sent to the configured LLM provider.

This affects deployments using external/cloud LLM providers such as OpenAI, Anthropic, etc.

### Expected behavior

Please provide an admin-level setting to disable injecting user PII into the LLM prompt, for example:

- `INCLUDE_USER_INFO_IN_LLM_PROMPT=false`
- or `INCLUDE_USER_PII_IN_PROMPT=false`

Vyacheslav-Tomashevskiy:
## Summary

Closes #11929.

Onyx always injects the signed-in user's name and email (and role) into the system prompt via the `# User Information` section built in `_build_user_information_section` (`backend/onyx/chat/prompt_utils.py`). When a deployment uses an external LLM provider, this sends user PII to that provider on every chat, with no way to turn it off.

This adds an opt-out:

- New env var **`DISABLE_USER_IDENTITY_IN_PROMPT`** (default `false`, so existing behaviour is unchanged).
- - When `true`, the "Basic Information" block (name / email / role) is omitted from the prompt. Team information, user preferences, and memories are left untouched, since the issue is specifically about the identifying name/email fields.
I mirrored the existing `DISABLE_USER_KNOWLEDGE` toggle right above it for naming and style.

## Why default to `false`

To stay fully backward compatible — nothing changes unless an operator explicitly opts in. Privacy/compliance-sensitive deployments can set the flag.

## Testing

Verified locally that with the flag unset the name/email still render in the section, and with `DISABLE_USER_IDENTITY_IN_PROMPT=true` the basic-information block is dropped while preferences/memories still render.

Happy to add a unit test for `_build_user_information_section`, or to rework this as a workspace/admin setting instead of an env var if that fits the codebase better — just let me know your preference.

---
## Summary by cubic
Add an opt-out env flag to keep user PII (name, email, role) out of LLM system prompts. Default remains unchanged; tests verify the gating.

- **New Features**
  - Added `DISABLE_USER_IDENTITY_IN_PROMPT` (default false).
  - When true, omits the Basic Information block; team info, preferences, and memories are still included.

<sup>Written for commit 1d12a7444eb6b3594c21a93f69f27b16da16f09d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12060?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/chat/prompt_utils.py
+++ b/backend/onyx/chat/prompt_utils.py
@@ -4,6 +4,7 @@ from uuid import UUID
 
 from sqlalchemy.orm import Session
 
+from onyx.configs.app_configs import DISABLE_USER_IDENTITY_IN_PROMPT
 from onyx.db.memory import UserMemoryContext
 from onyx.db.persona import get_default_behavior_persona
 from onyx.db.user_file import calculate_user_files_token_count
@@ -146,7 +147,7 @@ def _build_user_information_section(
     in the correct order: Basic Info → Team Info → Preferences → Memories."""
     sections: list[str] = []
 
-    if user_memory_context:
+    if user_memory_context and not DISABLE_USER_IDENTITY_IN_PROMPT:
         ctx = user_memory_context
         has_basic_info = ctx.user_info.name or ctx.user_info.email or ctx.user_info.role
 
--- a/backend/onyx/configs/app_configs.py
+++ b/backend/onyx/configs/app_configs.py
@@ -76,6 +76,15 @@ GENERATIVE_MODEL_ACCESS_CHECK_FREQ = int(
 # Controls whether users can use User Knowledge (personal documents) in assistants
 DISABLE_USER_KNOWLEDGE = os.environ.get("DISABLE_USER_KNOWLEDGE", "").lower() == "true"
 
+# Controls whether the user's basic identity (name, email, role) is injected into the
+# LLM system prompt. When True, this identifying information is omitted from the prompt
+# sent to the (potentially external) LLM provider, which is useful for privacy/compliance
+# sensitive deployments. Other user context (team info, preferences, memories) is
+# unaffected. Defaults to False to preserve existing behavior.
+DISABLE_USER_IDENTITY_IN_PROMPT = (
+    os.environ.get("DISABLE_USER_IDENTITY_IN_PROMPT", "").lower() == "true"
+)
+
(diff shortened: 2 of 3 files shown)


Vyacheslav-Tomashevskiy:
Added unit tests in `backend/tests/unit/onyx/chat/test_prompt_utils.py` covering the new branch: default behaviour still renders name/email/role, and with `DISABLE_USER_IDENTITY_IN_PROMPT=true` the basic-info block is omitted while team info, preferences and memories are preserved. This closes the coverage gap noted in the review.

On the per-tenant point: I kept this as a process-level env var to stay consistent with the sibling `DISABLE_USER_KNOWLEDGE` flag. Happy to move it to a workspace/tenant-level setting instead if you'd prefer that for multi-tenant deployments.

PR #12077, opened 2026-06-15
fix(sentry): reduce captured content in error events

nmgarza5:
## Description

Sentry was storing things we dont want them store in plaintext in error events.

- **Bug:** with `include_local_variables` on (the default), a raised `litellm` call captured the outbound request `headers` dict — which holds the protected info — into the event. Sentry's default scrubber doesnt catch everything, so some headers slipped through. Confirmed on a real production event.
- **Fix:** centralize all Sentry setup in one `init_sentry()` helper so no entrypoint can reintroduce the leak. It turns off `include_local_variables` and installs a recursive `EventScrubber` whose denylist adds the hyphenated header names. All four init sites (api server, model server, celery app, docfetching task) route through it.
- **Test:** unit test for the nested-header redaction.

Stops new captures only

## Additional Options

### DIFF
--- a/backend/model_server/main.py
+++ b/backend/model_server/main.py
@@ -5,7 +5,6 @@ from collections.abc import AsyncGenerator
 from contextlib import asynccontextmanager
 from pathlib import Path
 
-import sentry_sdk
 import torch
 import uvicorn
 from fastapi import FastAPI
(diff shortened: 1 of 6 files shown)


nmgarza5:
@greptile review

nmgarza5:
@greptile review

jmelahman (approved):

PR #12079, opened 2026-06-15
fix(llm): return 400 not 502 when an Ollama server is unreachable

nmgarza5:
## Description

**Bug:** The admin "available models" endpoint returned `502 BAD_GATEWAY` for *any* failure reaching the user-supplied Ollama URL — including the common case where the URL simply isn't reachable from Onyx (a localhost/private address entered on cloud). That reads as a server fault in metrics and tells the admin nothing actionable.

**Fix:** Distinguish the two failure modes in `_get_ollama_available_model_names`:
- connection / timeout (`httpx.RequestError`) → `400 VALIDATION_ERROR` ("could not reach `<url>`") — a client misconfiguration
- an error *response* from Ollama (`httpx.HTTPStatusError`) → `502 BAD_GATEWAY` — a genuine upstream fault

Added two unit tests for the new mapping.

**Design note:** mirrors the existing `_get_openai_compatible_models_response` error-handling shape. Ollama maps unreachable→400 because it's overwhelmingly self-hosted (see `api.py:1213`). Litellm/bifrost via the shared helper still map unreachable→502; aligning them is a possible follow-up, left out to keep this scoped to the observed defect.

## Additional Options

### DIFF
--- a/backend/onyx/server/manage/llm/api.py
+++ b/backend/onyx/server/manage/llm/api.py
@@ -1179,12 +1179,28 @@ def get_bedrock_available_models(
 
 
 def _get_ollama_available_model_names(api_base: str) -> set[str]:
-    """Fetch available model names from Ollama server."""
+    """Fetch available model names from an Ollama server.
+
+    An unreachable address surfaces as 400, not 502: the admin supplied the
+    URL, so a server Onyx can't route to is a client misconfiguration.
+    """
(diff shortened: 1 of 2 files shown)


nmgarza5:
@greptile

justin-tahara (approved):

nmgarza5:
@greptile

PR #12085, opened 2026-06-15
fix(llm): map unreachable OpenAI-compatible model server to 400, not 502

nmgarza5:
## Description

Follow-up to #12079 (which fixed the Ollama handler). Applies the same fix to the shared `_get_openai_compatible_models_response` helper used by **LiteLLM, Bifrost, and LM Studio**.

- **Bug:** a connection/timeout error on the user-supplied model-server URL raised `502 BAD_GATEWAY` — but an unreachable address is a client misconfiguration, not a gateway fault, and it shows up as a server error in metrics.
- **Fix:** connection errors (`httpx.RequestError`) → `400 VALIDATION_ERROR` with a "could not reach `<url>`" message. An actual error *response* from the upstream (`httpx.HTTPStatusError`) still maps to 502.
- Updates the two existing connection-failure tests (LiteLLM, Bifrost) to assert the new 400.

## Additional Options

### DIFF
--- a/backend/onyx/server/manage/llm/api.py
+++ b/backend/onyx/server/manage/llm/api.py
@@ -1667,8 +1667,6 @@ def _get_openai_compatible_models_response(
             extra={"source": source_name, "url": url, "error": str(e)},
             exc_info=True,
         )
-        # Unreachable user-supplied URL is a client misconfiguration, not a
-        # gateway fault — mirror the Ollama handler and return 400, not 502.
         raise OnyxError(
             OnyxErrorCode.VALIDATION_ERROR,
             f"Could not reach {source_name} at {url}. Check that the URL is "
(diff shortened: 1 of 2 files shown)


backend/onyx/server/manage/llm/api.py
```diff
@@ -1663,13 +1663,16 @@ def _get_openai_compatible_models_response(
             )
     except httpx.RequestError as e:
         logger.warning(
-            "Failed to fetch models from OpenAI-compatible endpoint",
+            "Could not reach OpenAI-compatible models endpoint",
             extra={"source": source_name, "url": url, "error": str(e)},
             exc_info=True,
         )
+        # Unreachable user-supplied URL is a client misconfiguration, not a
```
Danelegend: Remove comment

Danelegend (approved):

PR #12087, opened 2026-06-15
fix(file-store): make change_file_id a metadata-only rename

nmgarza5:
## Description

**Bug:** On checkpoint resume, reissuing leftover batches renames each via `change_file_id`, which physically copied the underlying object (S3 `copy_object`+`delete_object`, GCS `copy_blob`+delete) before repointing the DB record. A connector with tens of thousands of un-processed batches (e.g. a large Gmail mailbox) spent ~an hour in serial object copies; the DB connection dropped mid-resume and the attempt failed during checkpoint load before indexing anything.

**Fix:** `change_file_id` is now a metadata-only rename — the new file_record points at the same `object_key` as the old; the object never moves. Reads resolve via the stored `object_key` (never re-derived from `file_id`), so this is safe, and the per-batch rename on resume becomes a cheap DB op. Batches still consolidate under the resuming attempt. Postgres already moved only the large-object pointer (unchanged). Added an `old == new` guard (the old copy path would self-destruct on it).

**Constraint:** the object keeps `old_file_id`'s storage key, so `old_file_id` must not be reused for a later `save_file` (it would overwrite the renamed object). The sole caller (batch reissue) never reuses a prior attempt's paths — documented in the docstring.

## Additional Options

### DIFF
--- a/backend/onyx/file_store/file_store.py
+++ b/backend/onyx/file_store/file_store.py
@@ -516,54 +516,36 @@ class S3BackedFileStore(FileStore):
     def change_file_id(
         self, old_file_id: str, new_file_id: str, db_session: Session | None = None
     ) -> None:
+        """Rename a file by repointing its DB record at the existing object.
+
+        The object is not moved — only file_id changes — and reads resolve via
+        the stored object_key, so they still find it. The object keeps its
+        original key, so a file_id must not be reused for a new save_file after
+        it has been renamed (the new write would overwrite the renamed object).
+        """
+        if old_file_id == new_file_id:
+            return
         with get_session_with_current_tenant_if_none(db_session) as db_session:
             try:
-                # Get the existing file record
                 old_file_record = get_filerecord_by_file_id(
                     file_id=old_file_id, db_session=db_session
                 )
-
-                # Generate new S3 key for the new file ID
-                new_s3_key = self._get_s3_key(new_file_id)
-
-                # Copy S3 object to new key
-                s3_client = self._get_s3_client()
-                bucket_name = self._get_bucket_name()
-
-                copy_source = (
-                    f"{old_file_record.bucket_name}/{old_file_record.object_key}"
-                )
-
-                s3_client.copy_object(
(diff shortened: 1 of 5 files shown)


backend/onyx/file_store/file_store.py
```diff
@@ -516,54 +516,40 @@ def delete_file(
     def change_file_id(
         self, old_file_id: str, new_file_id: str, db_session: Session | None = None
     ) -> None:
+        """Repoint file_id at the existing object — metadata only, no S3 copy.
+
+        Reads resolve the object via the stored object_key (never re-derived
+        from file_id), so the object stays put and only the DB pointer moves.
```
evan-onyx: could we clean this comment up to not reference previous code state (since it will become stale) and things like "sole caller" which can become very problematically stale if that changes

backend/onyx/file_store/gcs_file_store.py
```diff
@@ -326,31 +326,35 @@ def change_file_id(
         new_file_id: str,
         db_session: Session | None = None,
     ) -> None:
+        """Repoint file_id at the existing blob — metadata only, no GCS copy.
+
+        Reads resolve the blob via the stored object_key (never re-derived
+        from file_id), so the blob stays put and only the DB pointer moves.
```
evan-onyx: clean up, I think this should be 3-4 lines

evan-onyx (approved):
nice!! would like to see shorter and less stale-prone comments but otherwise LGTM

backend/onyx/file_store/file_store.py
nmgarza5: ugh yeah. it should be using the code comment skill you created so maybe i'll make an update to that. (or just ensure it is actually being called)

backend/onyx/file_store/gcs_file_store.py
nmgarza5: yeah yeah

nmgarza5:
@greptile

PR #12096, opened 2026-06-16
fix: flush hold buffer on stream end in DynamicCitationProcessor

### LINKED ISSUE #11932: Streaming answers truncated mid-token at end of stream: `DynamicCitationProcessor` final flush drops `self.hold` buffer
### Summary

When an LLM answer ends while the citation processor is holding trailing
characters in its stop-token buffer (`self.hold`), those characters are
permanently dropped. The user-visible answer is truncated mid-word at the
very end, even though the model finished normally (`finish_reason=stop`)
and emitted the full token count.

### Environment

- Onyx v4.0.5 (self-hosted, Docker Compose)
- Index: OpenSearch (Vespa removed)
- LLM provider: OpenAI-compatible gateway backed by vLLM
- Model: `qwen3-235b-a22b-instruct-2507`

### Symptom

A RAG answer renders fully, then cuts off mid-word on the last line, e.g.:

> ... In case of doubt about the splice quality, **always redo it**: it's faster than re

The text ends abruptly right after a region dense with citation markers.
The provider stream is complete and correct (verified independently — see below).

### Root cause

In `onyx/chat/citation_processor.py`, `DynamicCitationProcessor.process_token`,
the end-of-stream flush (called via `process_token(None)`) only emits
`self.curr_segment`:

```python
# None -> end of stream, flush remaining segment
if token is None:
    if self.curr_segment:
        yield self.curr_segment
    return
```

However, the stop-token logic earlier in the same method diverts trailing
characters into `self.hold` whenever they form a prefix of `stop_stream`,
returning without yielding:

```python
elif next_hold == self.stop_stream[: len(next_hold)]:
    self.hold = next_hold
    return
```

If the stream ends while `self.hold` is non-empty (i.e. the final tokens
happened to partially match the stop pattern), the `token is None` branch
never flushes `self.hold`, so those characters are lost.

This is most likely to trigger when an answer ends shortly after a citation,
which is common with well-cited RAG responses.

### Independent verification that the provider is not at fault

Three direct streaming calls to the provider all return complete, correct output:

1. Plain prompt, no tools → full prose in `delta.content`.
2. Tools + `tool_choice="auto"` → correct `tool_calls` stream.
3. Tools + `tool_choice="none"` → full prose in `delta.content`.

The truncation only appears after Onyx's citation processing.

### Proposed fix

Flush `self.hold` together with `self.curr_segment` at end of stream:

```python
if token is None:
    remaining = self.hold + self.curr_segment
    if remaining:
        yield remaining
    self.hold = ""
    return
```

(Held content is only ever a partial stop-pattern prefix, so it is very
unlikely to contain a complete citation; emitting it as plain text is
acceptable. A stricter fix could re-run it through the citation matcher.)

### Impact

Silent truncation of the tail of cited answers. No error is raised; the
answer simply looks incomplete to the user.

wengkit218-pixel:
## Summary

Fixes #11932

When the LLM stream ends (token=None), `self.hold` could still contain characters from a partial stop-stream match that turned out to be a false alarm. Previously, only `self.curr_segment` was flushed in the end-of-stream handler, causing the held characters to be silently dropped and the LLM output to be truncated mid-word at the very end.

This is most likely to trigger when an answer ends shortly after a citation, which is common with well-cited RAG responses.

## Changes

```python
# Before (bug):
if token is None:
    if self.curr_segment:
        yield self.curr_segment
    return

# After (fix):
if token is None:
    remaining = self.hold + self.curr_segment
    if remaining:
        yield remaining
    self.hold = ""
    return
```

## Why this fix is safe

- `self.hold` only ever contains characters that partially matched the `stop_stream` pattern prefix. If the stream ends without completing the match, these characters are legitimate output that should be emitted as plain text.
- Held content is very unlikely to contain a complete citation (it's at most a few characters of a partial stop-pattern match), so emitting it as plain text is acceptable.
- `self.hold` is reset to empty string after flushing to maintain clean state.

## Testing

Added 4 new test cases in `TestHoldBufferFlushOnStreamEnd`:
1. `test_hold_flushed_on_stream_end_with_stop_stream` — verifies held characters are emitted
2. `test_hold_and_curr_segment_both_flushed` — verifies both buffers are combined and flushed
3. `test_no_truncation_with_citation_near_end` — end-to-end scenario matching the original bug report
4. `test_hold_cleared_after_flush` — verifies `self.hold` is reset

All existing tests continue to pass.

---
## Summary by cubic
Flushes the hold buffer at stream end in `DynamicCitationProcessor`, emitting `curr_segment` + `hold` in order so partial stop-pattern characters are output instead of dropped. Prevents end-of-output truncation; fixes #11932.

- **Bug Fixes**
  - On `token=None`, yield `self.curr_segment + self.hold` once, then clear `self.hold` and `self.curr_segment`.
  - Adds and strengthens regression tests, including explicit assertions that both buffers are non-empty before stream end, plus checks for ordering, no truncation near citations, and state reset.

<sup>Written for commit 0576d5539b6ef339fc5f940570f105e9e266b7cc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12096?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/chat/citation_processor.py
+++ b/backend/onyx/chat/citation_processor.py
@@ -275,10 +275,12 @@ class DynamicCitationProcessor:
             str: Text chunks to display. Citation format depends on citation_mode.
             CitationInfo: Citation metadata (only when citation_mode=HYPERLINK)
         """
-        # None -> end of stream, flush remaining segment
+        # None -> end of stream, flush remaining segment and held tokens
         if token is None:
-            if self.curr_segment:
-                yield self.curr_segment
+            remaining = self.hold + self.curr_segment
+            if remaining:
+                yield remaining
+            self.hold = ""
             return
 
         # Handle stop stream token
--- a/backend/onyx/chat/citation_processor.py
+++ b/backend/onyx/chat/citation_processor.py
@@ -277,10 +277,11 @@ class DynamicCitationProcessor:
         """
         # None -> end of stream, flush remaining segment and held tokens
         if token is None:
-            remaining = self.hold + self.curr_segment
+            remaining = self.curr_segment + self.hold
             if remaining:
                 yield remaining
             self.hold = ""
+            self.curr_segment = ""
             return
 
         # Handle stop stream token
--- a/backend/tests/unit/onyx/chat/test_citation_processor.py
+++ b/backend/tests/unit/onyx/chat/test_citation_processor.py
@@ -2545,21 +2545,22 @@ class TestHoldBufferFlushOnStreamEnd:
 
     def test_hold_and_curr_segment_both_flushed(self) -> None:
         """Both `self.hold` and `self.curr_segment` should be flushed on
-        stream end."""
+        stream end, with correct ordering (curr_segment before hold)."""
         processor = DynamicCitationProcessor(stop_stream="END")
 
         results: list[str | CitationInfo] = []
-        # "EN" partially matches stop stream, held in self.hold
-        for token in ["Prefix ", "EN"]:
+        # "[" starts a citation -> sits in curr_segment
+        # "EN" partially matches stop stream -> held in self.hold
+        # At stream end, output should be "[EN" (curr_segment first, then hold)
+        for token in ["[", "EN"]:
             for result in processor.process_token(token):
                 results.append(result)
-        # End of stream — should flush held "EN"
+        # End of stream — should flush curr_segment + hold in order
         for result in processor.process_token(None):
             results.append(result)
 
         output = "".join(r for r in results if isinstance(r, str))
-        assert "Prefix " in output
-        assert "EN" in output
+        assert output == "[EN"
 
     def test_no_truncation_with_citation_near_end(self) -> None:
         """End-to-end: a well-cited RAG response whose final characters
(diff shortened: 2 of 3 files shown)


wengkit218-pixel:
Thanks @greptile-apps for the detailed review! 

You're right about the concatenation order. I've already fixed it in the latest commit (`bc70a95`) 鈥?the code now correctly uses `self.curr_segment + self.hold`.

I'll also update `test_hold_and_curr_segment_both_flushed` to properly exercise the dual-non-empty scenario as you described. Thanks for catching that!

wengkit218-pixel:
Hi! I submitted this PR 8 days ago to fix the hold buffer flush issue in DynamicCitationProcessor. I've already addressed the Greptile review feedback (commit bc70a95). Could you please review when you have a chance? Let me know if I should make any changes. Thanks!

wengkit218-pixel:
Hi! I submitted this PR 9 days ago to fix the hold buffer flush issue in DynamicCitationProcessor. I have already addressed the Greptile review feedback (commit bc70a95). Could you please review when you have a chance? Let me know if I should make any changes. Thanks!

wengkit218-pixel:
Closing due to no human review after 15+ days. Bug fix addressed Greptile feedback. Thanks!

PR #12100, opened 2026-06-16
refactor(oauth): shared onyx/oauth layer for OAuth refresh

Danelegend:
**Foundation PR** — extracts a shared OAuth layer from the code that exists today, so the MCP proactive-refresh fix (follow-up, stacked on this) is built clean on it rather than duplicating Craft's patterns and being de-duped afterward.

### What
New neutral `onyx/oauth/` package (depends only on `onyx.utils` + `onyx.server.security`):
- `errors.py` — refresh error taxonomy + `token_response_error` classifier
- `expiry.py` — clockless `needs_refresh` / `stamp_expires_at`
- `exchange.py` — the auth-code wire layer (`OAuthFlowParams`, SSRF endpoint guard, authorize-URL builder, code/refresh exchanges) + the shared `request_oauth_token` primitive
- `single_flight.py` — `SingleFlightTokenRefresher` Template Method

These primitives previously lived inside `onyx/external_apps/...` (forcing `auth/` and the MCP feature to import *up* into a feature package) and `onyx/auth/oauth_token_manager.py`. Now everything imports *down* into `onyx/oauth/`.

The two OAuth refreshers that exist today are refactored onto it:
- Craft external-app provider refresh delegates its POST/classify to `request_oauth_token`; `ensure_fresh_credentials` becomes a thin `SingleFlightTokenRefresher` strategy.
- The tool-OAuth manager reuses the shared `exchange_refresh_token`.

### Why this ordering
Both this and the MCP fix were unmerged, so rather than "ship the fix (which mirrors Craft) then dedupe," the abstraction lands first and the fix is built on it — no introduce-then-delete churn, and reviewers of the fix see clean code.

### Safety
No behavior change. Pure relocation + import repoints + routing the two existing refreshers through the shared primitives. `ty` + pre-commit clean; 223 affected unit tests pass. (`SingleFlightTokenRefresher` has one consumer here — Craft — by design; the MCP fix adds the second in the next PR.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Extracted a shared OAuth layer under `onyx/oauth` and refactored existing refresh flows to use it, with no behavior changes. This standardizes token exchange/refresh, SSRF guarding, and error handling across external apps, tools, and MCP.

- **Refactors**
[...]

PR #12101, opened 2026-06-16
fix(mcp): proactively refresh OAuth tokens so connections don't lapse

Danelegend:
**Feat PR** — built on the shared OAuth layer in #12100 (its base). Review/merge #12100 first; this retargets to `main` once it lands.

Supersedes #12052: same fix, but built *on* the shared `onyx/oauth` abstractions instead of mirroring Craft's refresh code and being de-duped afterward.

### Problem
OAuth-backed MCP servers (e.g. Asana) disconnect once the access token expires and require a manual admin reconnect. The MCP SDK's `OAuthClientProvider` is rebuilt per tool call and never restores token expiry on load, so its proactive-refresh branch never fires — an expired token escalates to a full re-auth ("Please Reconnect to the server") instead of a refresh-token grant, and the stored refresh token is effectively never used.

### Fix
Onyx drives refresh itself. `_MCPTokenRefresher` is a thin `SingleFlightTokenRefresher` strategy — the **second consumer** of the shared Template Method from #12100:
- Persists absolute `token_expires_at` + the resolved token endpoint (the SDK's `OAuthToken` drops `expires_at`).
- Proactive, Redis-single-flighted refresh: resolve endpoint (persisted → server row → RFC 8414 discovery) → `exchange_refresh_token` (shared) → persist → flip an **admin-owned** server to `AWAITING_AUTH` only on a genuinely dead grant. Transient failures keep the token; per-user grants don't disconnect the whole server.
- Called from `MCPTool.run` before each call; also covers SSE (where the SDK auth provider is dropped), since refresh runs header-based regardless of transport.

### Scope
Only the MCP-specific surface: `oauth_refresh.py` (the strategy), `MCPOAuthKeys` additions, `mcp_tool.py` wiring, the known-provider callback in `mcp/api.py`, and `test_oauth_refresh.py`. The wire layer, exchange primitive, error taxonomy, and single-flight skeleton all come from #12100.

`ty` + pre-commit clean; 237 affected unit tests pass; the shared wire layer is additionally covered by a live local-endpoint test.

---
## Summary by cubic
Proactively refresh OAuth tokens for MCP servers so connections don’t lapse and admins don’t need to reconnect. Onyx now refreshes via the shared `onyx/oauth` layer and applies new headers before each call.

- **Bug Fixes**
  - Persist absolute `token_expires_at` and `token_endpoint` in connection config on OAuth callback and token updates; heals legacy configs missing expiry.
  - Add `_MCPTokenRefresher` (Redis single-flighted): resolve endpoint (persisted → server row → RFC 8414 discovery), call `exchange_refresh_token`, persist tokens/headers; mark admin-owned servers `AWAITING_AUTH` only on terminal errors; keep existing token on transient/infra errors and for per-user failures.
  - Wire into `MCPTool.run` to update Authorization headers before each call (incl. SSE); safe fallback if refresh can’t run.
  - Add unit tests for staleness checks, double-check under lock, discovery, contention/Redis-down paths, and error handling.

<sup>Written for commit 63a9e0058a2b5162352ea3e10c73e58b9d131984. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12101?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/server/features/mcp/api.py
+++ b/backend/onyx/server/features/mcp/api.py
@@ -389,6 +389,13 @@ class OnyxTokenStorage(TokenStorage):
             config_data["headers"] = {
                 "Authorization": f"{tokens.token_type} {tokens.access_token}"
             }
+            # OAuthToken drops expires_at, so persist absolute expiry separately —
+            # proactive refresh has no other way to tell when this token lapses.
+            if tokens.expires_in is not None:
+                config_data[MCPOAuthKeys.EXPIRES_AT.value] = (
+                    datetime.datetime.now(datetime.timezone.utc)
+                    + datetime.timedelta(seconds=tokens.expires_in)
+                ).isoformat()
             update_connection_config(config.id, db_session, config_data)
 
         # The shared admin row is intentionally NOT written here: it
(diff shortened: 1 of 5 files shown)


Danelegend:
Closing in favor of the standalone approach: the MCP OAuth fix ships on its own via #12052, and combining MCP + external-apps OAuth onto a shared layer is deferred to a later effort. Branch kept (not deleted) so this rebuilt-on-foundation version is available when we pick up the combine work.

PR #12102, opened 2026-06-16
fix(file-processing): extract PDF text with pypdfium2 instead of pypdf

nmgarza5:
## Description

**Bug:** Connectors with many large PDFs (e.g. a Google Drive full of regulatory docs) never finish indexing — every attempt dies with `No heartbeat received for 1800s`.

**Cause:** `read_pdf_file` extracts text with pypdf, whose `extract_text` is pure Python and holds the GIL while parsing content streams. On big PDFs it pins a docfetching worker for minutes; because the Google Drive connector converts files in a thread pool, it starves the indexing heartbeat thread and the watchdog kills the whole attempt. (Diagnosed with py-spy on a live hung worker.)

**Fix:** Extract text via pypdfium2 instead of pypdf.

### Why pypdfium2 — in plain terms

pypdfium2 wraps **PDFium, the C++ PDF engine inside Google Chrome** — the same code billions of people use to open PDFs every day.

- **It stops one PDF from freezing everything.** pypdf is pure Python, so while it chews through a big PDF it holds Python's global lock and freezes every other task in that worker — including the "I'm still alive" heartbeat. The watchdog then assumes the worker is dead and kills the entire indexing run. PDFium does its heavy work in C and lets go of that lock, so the heartbeat and the other files keep flowing.
- **It's much faster.** A native C engine parses PDFs in a fraction of the time pure-Python extraction takes, so big libraries get through quicker.
- **It's proven.** PDFium is the most heavily used PDF engine in the world — it renders every PDF in Chrome — so the extracted text is at least as good.

**Why this is the right approach, not a band-aid:**

- Removes the root cause (the freeze) instead of hiding it. Bumping the 1800s timeout would just delay the same failure on the next big PDF.
- Surgical: only the slow text step changes. pypdf still handles metadata, encryption, and image extraction — what it's good at.
- One shared path: every PDF in Onyx (Google Drive, web content, file uploads) flows through `read_pdf_file`, so this fixes all of them at once.

PR 1 of 2 — PR 2 adds a subprocess-isolated extraction timeout as a backstop.

## Additional Options


nmgarza5:
@greptile

evan-onyx (approved):

PR #12104, opened 2026-06-16
feat(mcp): surface MCP tool auth failures as re-authenticatable errors

beraterkanelcelik:
## Description

MCP tool calls that fail authentication — a per-user MCP server with no credentials configured, or a 401/403 at call time — only embedded the auth message as a plain string in the tool result. The chat UI's re-authentication prompt keys off `CustomToolDelta.error.is_auth_error`, which the MCP tool never set, so an auth failure on an MCP tool surfaced no re-auth affordance (unlike `custom_tool.py`, which does set it).

This makes `mcp_tool.py` populate `CustomToolErrorInfo(is_auth_error=True, status_code, message)` on both auth-failure paths — the pre-call "requires auth but no credentials" branch and the call-time 401/403 exception branch — and attach it to both the streamed `CustomToolDelta` and the `CustomToolCallSummary`, mirroring `custom_tool.py`. The existing re-authentication UX then works for MCP tools too.

This is **PR 1 of a small stacked series** for surfacing MCP re-authentication needs in the chat UI. The follow-ups are pushed to my fork and build on this commit:
1. (this) backend — emit `is_auth_error` on MCP auth failures
2. frontend — a re-auth badge + "Needs re-authentication" section on the chat-input actions popover
3. backend — a derived per-user auth-status endpoint so the badge can show on page load
4. backend — persist observed 401s so the signal survives a reload

Opening them sequentially since cross-fork PRs can't be stacked against this repo; happy to have the branches pulled and stacked if you prefer.

## How Has This Been Tested?

- Unit test (`backend/tests/unit/onyx/tools/tool_implementations/mcp/test_mcp_tool_auth_error.py`): both auth paths set `error.is_auth_error` (401, and 403 for forbidden); a non-auth failure does not (guards against false positives).
- Manually verified against a local mock MCP server that returns 401 on tool calls — the streamed `CustomToolDelta` now carries `error.is_auth_error: true`.

## Additional Options

---
## Summary by cubic
Surfaces MCP tool auth failures as re-authenticatable errors so the chat UI can prompt re-auth for MCP servers. Covers missing credentials and call-time 401/403, matching `custom_tool.py`.

- **Bug Fixes**
  - Populate `CustomToolErrorInfo` (is_auth_error, status_code, message) on both `CustomToolDelta` and `CustomToolCallSummary` for the pre-call “requires auth” path and for auth exceptions.
  - Status codes: 401 for missing creds; 403 when the exception mentions 403/forbidden/access denied; otherwise 401. Added unit tests; non-auth failures are not tagged.

<sup>Written for commit 0b66e57c5ac6f8e951b38545414e333f63208a77. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12104?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/tools/tool_implementations/mcp/mcp_tool.py
+++ b/backend/onyx/tools/tool_implementations/mcp/mcp_tool.py
@@ -10,6 +10,7 @@ from onyx.db.models import MCPConnectionConfig
 from onyx.db.models import MCPServer
 from onyx.server.query_and_chat.placement import Placement
 from onyx.server.query_and_chat.streaming_models import CustomToolDelta
+from onyx.server.query_and_chat.streaming_models import CustomToolErrorInfo
 from onyx.server.query_and_chat.streaming_models import CustomToolStart
 from onyx.server.query_and_chat.streaming_models import Packet
 from onyx.tools.interface import Tool
@@ -195,6 +196,15 @@ class MCPTool(Tool[None]):
 
                 error_result = {"error": auth_error_msg}
                 llm_facing_response = json.dumps(error_result)
+                # Tag as an auth error so the chat UI can prompt re-auth (mirrors
+                # custom_tool.py); covers the never-authenticated case. message is
+                # the user-facing label the re-auth UI renders, not the verbose
+                # LLM instruction above.
+                auth_error_info = CustomToolErrorInfo(
+                    is_auth_error=True,
+                    status_code=401,
+                    message=f"Authentication required for {self.mcp_server.name}",
+                )
 
                 # Emit CustomToolDelta packet
                 self.emitter.emit(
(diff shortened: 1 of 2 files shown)


beraterkanelcelik:
Addressed greptile's P2: `CustomToolErrorInfo.message` now carries a short, server-scoped user-facing label (`Authentication required for <server>` / `Re-authentication required for <server>`) rather than the verbose LLM-facing instruction string, which stays in the tool result for the model to act on. Pushed.

beraterkanelcelik:
Also addressed greptile's second P2 (mcp_tool.py status mapping): added `"access denied"` to the 403 branch so an access-denied failure is tagged 403 (creds provided, access refused) rather than 401 — mirrors the existing `"forbidden"` handling. Pushed.

PR #12105, opened 2026-06-16
feat(skills): attach skills to agents, loaded on demand

beraterkanelcelik:
## Description

**Builds on #11923 (user-managed personal skills).** #11923 lets a user create/manage their own personal skills and delivers them to the **Craft sandbox**. This PR adds the missing half: **attach skills to a normal agent and load them on demand in regular chat** — no sandbox involved.

- **`Persona__Skill` association + `skill_ids`** on the agent upsert. Attached ids are validated against the caller's visible set (`filter_visible_skill_ids`, built on #11923's `_add_user_visibility_filter`/`_personal_skill_clause`) and silently dropped otherwise; fetch-time masking (`fetch_persona_skills_visible_to_user`) shows each viewer only the intersection, so a private skill attached to a shared agent never leaks.
- **On-demand consumption.** When an agent has attached skills, its system prompt carries only a compact skill *index* (name/slug/description). The full `SKILL.md` loads only when the model calls a new `load_skill` in-code tool — offered for a turn only when the agent has visible attached skills — so the load shows as a tool call. Skills never execute in normal chat; `load_skill` only ever renders the `SKILL.md` text, so attaching any personal skill to a normal agent is safe.

Two additive migrations: the `Persona__Skill` association, and seeding the `load_skill` tool row (chained after #11923's migrations).

A small frontend PR (#12146) adds the agent-editor skill selector; it follows this one.

## How Has This Been Tested?

- Unit: the `load_skill` tool (loads a visible attached skill by slug, rejects an unavailable slug).
- External-dependency-unit: index rendering + per-skill body load (`test_attached_skills_render`).
- Integration: persona attach + per-viewer masking (`test_skills_persona_assoc`).
- `ruff`/`ty` clean; migrations apply to a fresh DB with a single head.

## Additional Options

### DIFF
--- a/backend/alembic/versions/d1bba4465bb7_seed_load_skill_tool.py
+++ b/backend/alembic/versions/d1bba4465bb7_seed_load_skill_tool.py
@@ -1,0 +1,67 @@
+"""seed_load_skill_tool
+
+Revision ID: d1bba4465bb7
+Revises: dfe0f30fd0ce
+Create Date: 2026-06-16 00:00:00.000000
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
+revision = "d1bba4465bb7"
+down_revision = "dfe0f30fd0ce"
+branch_labels = None
+depends_on = None
+
+
+LOAD_SKILL_TOOL = {
+    "name": "LoadSkillTool",
+    "display_name": "Load Skill",
+    "description": (
+        "Load the full instructions for one of this assistant's attached "
+        "skills, identified by its slug."
+    ),
+    "in_code_tool_id": "LoadSkillTool",
+    "enabled": True,
+}
+
+
+def upgrade() -> None:
+    conn = op.get_bind()
+
+    existing = conn.execute(
+        sa.text(
+            "SELECT in_code_tool_id FROM tool WHERE in_code_tool_id = :in_code_tool_id"
+        ),
+        {"in_code_tool_id": LOAD_SKILL_TOOL["in_code_tool_id"]},
+    ).fetchone()
+
+    if existing:
+        conn.execute(
+            sa.text("""
+                UPDATE tool
+                SET name = :name,
+                    display_name = :display_name,
+                    description = :description
+                WHERE in_code_tool_id = :in_code_tool_id
+                """),
+            LOAD_SKILL_TOOL,
+        )
+    else:
+        conn.execute(
+            sa.text("""
+                INSERT INTO tool (name, display_name, description, in_code_tool_id, enabled)
+                VALUES (:name, :display_name, :description, :in_code_tool_id, :enabled)
+                """),
+            LOAD_SKILL_TOOL,
+        )
+
+
+def downgrade() -> None:
+    conn = op.get_bind()
+    conn.execute(
+        sa.text("DELETE FROM tool WHERE in_code_tool_id = :in_code_tool_id"),
+        {"in_code_tool_id": LOAD_SKILL_TOOL["in_code_tool_id"]},
+    )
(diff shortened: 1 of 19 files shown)


beraterkanelcelik:
Addressed both greptile P2s:
- `ingest.py`: `parse_markdown_skill` is now a bare validation-gate call (no misleading unpack into discarded `name, description`); the returned `IngestedBundle` fields still come from `ingest_skill_bundle` re-parsing the synthesized zip.
- `render.py`: dropped the unused `db_session`/`user` parameters from `render_attached_skills_section` (and the `# noqa: ARG001` suppressions) and updated the single call site in `process_message.py`. Pushed.

beraterkanelcelik:
Thanks — addressed cubic's findings:
- **ingest.py (`name`/`description` unused, F841):** `parse_markdown_skill` is now a bare validation-gate call; no unused unpack.
- **migration (`ix_persona__skill_persona_id` redundant):** dropped it — the composite PK `(persona_id, skill_id)` already serves persona_id-prefix lookups. Kept `ix_persona__skill_skill_id` since the reverse lookup (which personas reference a skill) can't use the PK's left-prefix.
- **test `finally` UnboundLocalError:** this one I believe is a false positive — `ingested` is assigned *before* the `try` (and `token` before its `try` in the fixture), so the `finally` can't reference an unbound name. Moving the assignment inside `try` is the pattern that would actually introduce that risk. Left as-is; happy to adjust if you'd prefer a different idiom.

beraterkanelcelik:
Marking this **draft** for now. #11923 (user-managed personal skills) just merged and covers user skill creation/management, which overlaps the create half of this PR. I'll refocus this onto the genuinely net-new parts that #11923 doesn't have — attaching skills to agents (`Persona__Skill`/`skill_ids`), the `load_skill` on-demand tool, and the chat-time render — built on top of the personal-skills feature rather than adding a parallel create path. Will un-draft once rebased + refocused.

beraterkanelcelik:
Refocused on top of #11923. Since #11923 now provides user skill create/manage (+ Craft-sandbox delivery), I dropped this PR's duplicate self-serve create endpoint and rebased onto current `main`. What remains is the part #11923 doesn't have: attaching skills to a **normal agent** + loading a skill's `SKILL.md` on demand via a new `load_skill` tool (no sandbox). Visibility/masking now reuses #11923's `_add_user_visibility_filter`/`_personal_skill_clause` rather than a parallel path. Migrations re-chained after #11923 (single head).

beraterkanelcelik:
Thanks — fixed both cubic findings:
- **P1 (persona.py): private-skill detach on co-editor save.** Persona update no longer wholesale-replaces `skills`. It now preserves attached skills the acting user can't see (another editor's private skill) and reconciles only within the caller's visible set — so saving a shared agent from a masked snapshot can't silently drop invisible attachments.
- **P2 (render.py): unguarded built-in read.** The built-in `SKILL.md` read is now wrapped (returns `None` on read/decode failure), matching the custom-skill loader's contract.
ruff/ty + load_skill unit + attached-render ext-dep all green.

beraterkanelcelik:
Rebased onto latest `main` (mergeable again).

Conflict resolved in `backend/onyx/chat/llm_loop.py`: upstream refactored the inline reminder-message `if/elif/else` into a new module-level `select_reminder_text(...)` helper. The resolution adopts upstream's `select_reminder_text(...)` call and preserves this PR's attached-skills block (the `system_prompt_additional_instructions` append that runs just before reminder selection). `py_compile` clean; no behavior change to the skills logic.

@greptile-apps @cubic-dev-ai please re-review.

beraterkanelcelik:
Addressed the review findings:
- Mask hidden skill IDs in the persona create/update response (matching the GET/list paths) — previously a persona update echoed skill IDs the caller can't see.
- Fix the self-serve test helper to upload a bundle zip under the `bundle` field (the real `POST /skills/custom` contract), not a bare `.md`.
- Cap `load_skill` output via the existing `truncate_output` so a large `SKILL.md` can't overflow the next turn's context.

(The `load_skill` tool-picker visibility was already fixed.) @greptile-apps @cubic-dev-ai please re-review.

PR #12130, opened 2026-06-16
fix(chat): disable model selector when there are no models to select

jmelahman:
20260616_14h16m19s_grim

## Description

Disables the chat model selector (`data-testid="model-selector"`) when there are no models available to choose from.

`MultiModelSelector` now derives whether any selectable models exist using the **same data source as the popover content** (`ModelSelectorContent`): `useCurrentAgent()` + `useLLMProviders(currentAgent?.id)` fed through `buildLlmOptions(...)`. So the selector disables in precisely the cases where opening the popover would otherwise show *"No models found"*. The SWR call dedupes against the popover's identical request (same key, 60s deduping), so there's no extra network cost.

When no models are selectable:
- the container is greyed out (`opacity-50`), marked `aria-disabled`, and made non-interactive (`pointer-events-none`), so the add button and model pills can't be clicked;
- `handleOpenChange` also guards against the popover opening, as defense-in-depth.

The selector stays **enabled while providers are still loading** (`!isLoading` guard) to avoid a flash of the disabled state.

Implementation note: I applied the disabled utilities directly rather than wrapping in Opal's `<Disabled>` component on purpose — `<Disabled>` always renders an extra `self-stretch` wrapper div, which would shift the selector's bottom-alignment in the welcome-screen row (`alignItems="end"`). The direct utilities match exactly what Opal's disabled styling uses while avoiding that layout regression.

## How Has This Been Tested?

- `tsc --noEmit` passes; `oxfmt` and `oxlint` clean on the changed file.
- Manual reasoning against the popover's empty-state logic (the disable condition mirrors the popover's `groupedOptions.length === 0` / "No models found" path).

No automated test added: exercising this requires a deployment with zero configured LLM providers, which is awkward to stand up, and per the repo's "don't overtest" guidance this is a small UI-state change. Happy to add a Playwright test that stubs `/api/llm/provider` (and the per-agent variant) to return an empty list and asserts the selector is `aria-disabled` if desired.

## Additional Options

---
## Summary by cubic
Disable the chat model selector when no models are available, matching the popover’s “No models found” state. Adds a tooltip and guards to block both mouse and keyboard activation, while keeping the selector enabled during provider loading.

- **Bug Fixes**
  - Prevent opening the popover and activating pills/add via early-return guards; the container stays hoverable to show “No models currently configured”.

- **Refactors**
  - Introduced `useCurrentAgentLLMProviders()` and adopted it in `MultiModelSelector`, `ModelSelector`, and `ModelSelectorContent`. Replaced `data-opal-disabled` with `aria-disabled` + `[&>*]:pointer-events-none`, and consolidated to a single container tooltip (“No models currently configured” or “Add Model”).

<sup>Written for commit df61510864d8dee7671039f0a64137b8308ab076. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12130?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/web/src/sections/model-selector/MultiModelSelector.tsx
+++ b/web/src/sections/model-selector/MultiModelSelector.tsx
@@ -1,6 +1,5 @@
 "use client";
 
-import "@opal/core/disabled/styles.css";
 import { useState, useMemo, useRef } from "react";
 import { getModelIcon } from "@/lib/languageModels";
 import {
@@ -11,6 +10,7 @@ import {
   Tooltip,
 } from "@opal/components";
 import { SvgPlusCircle, SvgX } from "@opal/icons";
+import { cn } from "@opal/utils";
 import { useSettingsContext } from "@/providers/SettingsProvider";
 import { LLMOption, buildLlmOptions } from "@/lib/languageModels/options";
 import { useCurrentAgent } from "@/lib/agents/hooks";
@@ -62,6 +62,15 @@ export default function MultiModelSelector({
 
   const isMultiModel = selectedModels.length > 1;
   const atMax = selectedModels.length >= MAX_MODELS || !multiModelAllowed;
+
+  // Single tooltip for the whole selector. The disabled reason takes
+  // precedence; otherwise it labels the add affordance. When at max there is
+  // no add action, so the tooltip is omitted.
+  const selectorTooltip = noModelsToSelect
+    ? "No models currently configured"
+    : atMax
+      ? undefined
+      : "Add Model";
 
   const selectedKeys = useMemo(
     () => new Set(selectedModels.map((m) => modelKey(m.provider, m.modelName))),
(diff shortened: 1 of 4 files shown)


web/src/sections/model-selector/MultiModelSelector.tsx
```diff
@@ -121,90 +149,104 @@ export default function MultiModelSelector({
...
-            onClick={(e: React.MouseEvent) => {
-              anchorRef.current = e.currentTarget as HTMLElement;
-              setReplacingIndex(null);
-              setOpen(true);
-            }}
+      {/*
+        When disabled, pointer events are blocked on the children (so the add
+        button / pills are inert) while the container itself stays hoverable, so
+        the Tooltip can still surface its message even in the disabled state.
+      */}
```
raunakab, line 156: Is this the reason `Disabled` from `@opal/core` is not being used?

web/src/sections/model-selector/MultiModelSelector.tsx
```diff
@@ -42,9 +51,27 @@ export default function MultiModelSelector({
...
     settings?.settings?.multi_model_chat_enabled ?? true;
 
+  // Mirror the data source used by `ModelSelectorContent` so the selector is
+  // disabled precisely when the popover would render "No models found".
+  const currentAgent = useCurrentAgent();
+  const { llmProviders, isLoading } = useLLMProviders(currentAgent?.id);
+  const noModelsToSelect = useMemo(
+    () => !isLoading && buildLlmOptions(llmProviders).length === 0,
+    [isLoading, llmProviders]
+  );
```
raunakab: Are some of these hook invocations duplicated from `ModelSelector`? If so, maybe we could create a shared hook inside of `lib/languageModels/hooks` and use it in both places.

(Of course, this is contingent on this specific logic being shared at all in the first place.)

raunakab (approved):

web/src/sections/model-selector/MultiModelSelector.tsx
jmelahman: Good call — done in df61510864. Extracted `useCurrentAgentLLMProviders()` into `lib/languageModels/hooks` (it just wraps `useCurrentAgent()` + `useLLMProviders(agent?.id)`) and swapped it in here, in `ModelSelector`, and in `ModelSelectorContent`. The `noModelsToSelect` derivation itself stays local since it's specific to this component.

web/src/sections/model-selector/MultiModelSelector.tsx
jmelahman: Partly, yes — two reasons I went with the disabled utilities directly instead of `Disabled`:

1. **Tooltip needs a hoverable trigger.** `Disabled` makes children inert; to surface the "No models currently configured" tooltip I need the container itself to stay hoverable while only its children are click-blocked (`[&>*]:pointer-events-none`).
2. **Layout.** `Disabled` always renders a `self-stretch` wrapper `<div>`, which would stretch the selector to the full row height and break its bottom-alignment in the welcome-screen row (`alignItems="end"`).

So the container gets `opacity-50 cursor-not-allowed [&>*]:pointer-events-none` + `aria-disabled`, plus the keyboard guards added in df61510864.

PR #12147, opened 2026-06-17
feat(mobile-auth): runtime instance URL, session manager + login flow

Subash-Mohan:
## Description

- Resolve the backend base URL from the user-entered Onyx instance URL at runtime; `EXPO_PUBLIC_API_URL` is now only a dev fallback.
- Add a SessionManager (login / logout / single-flight refresh) with a session-epoch guard so a late refresh can't resurrect a logged-out session, plus `useEmailLogin` / `useLogout` / `useSessionRefresh` hooks.
- Discover available sign-in methods from the connected backend via `/api/auth/type` (`useAuthConfig` + provider registry); ships email/password.
- Isolate state per instance: query keys and bearer tokens are scoped by server URL, and `/api/me` PII is kept out of the on-disk cache snapshot.
- Add a jest-expo test harness with centralized native-module mocks and unit tests for token storage, the session manager, and query keys.

## How Has This Been Tested?

- `bun run test` (14 unit tests pass) and `tsc --noEmit` clean.

## Additional Options

### DIFF
--- a/mobile/src/api/auth/tokenStore.ts
+++ b/mobile/src/api/auth/tokenStore.ts
@@ -1,35 +1,62 @@
 // Auth-token storage.
 //
-// `apiFetch` reads the access token via `getToken()`. The future login flow
-// calls `setToken(token)`; logout calls `setToken(null)`.
+// `apiFetch` reads the access token via `getToken()`. `SessionManager` calls
+// `setToken(token)` on login and `setToken(null)` on logout.
 //
 // Tokens are credentials, so they live in the device keychain via
 // expo-secure-store — NOT in MMKV (which is reserved for non-secret cache /
 // persisted UI state). Tests mock the `expo-secure-store` module.
 //
-// SECURITY TODO — must be handled when the login/logout flow lands (these are
-// dormant today only because no token exists, so /api/me always 403s and no
-// user data is ever cached):
-//   1. Logout must `queryClient.clear()` AND purge the persisted snapshot
-//      (`persister.removeClient()` / `queryStorage.clearAll()`) — otherwise the
-//      previous user's cached /api/me PII survives logout for up to 24h.
-//   2. Scope the cached-user query key to the authenticated identity (or bump a
-//      session epoch on login/logout) so user B can never read user A's cache.
-//   3. The query-cache MMKV instance (state/storage.ts) is unencrypted; once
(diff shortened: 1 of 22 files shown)


mobile/src/api/auth/tokenStore.ts
```diff
@@ -1,35 +1,62 @@
...
+import { getBaseUrl } from "@/api/config";
+
 // SecureStore keys allow alphanumerics plus ".", "-", "_".
-const ACCESS_TOKEN_KEY = "onyx.auth.access_token";
+const ACCESS_TOKEN_KEY_PREFIX = "onyx.auth.access_token";
+const SAFE_KEY_CHAR = /^[A-Za-z0-9.-]$/;
+
+const TOKEN_OPTIONS: SecureStore.SecureStoreOptions = {
+  keychainAccessible: SecureStore.WHEN_UNLOCKED_THIS_DEVICE_ONLY,
+};
```
nmgarza5: what does this do exactly?

nmgarza5 (approved):
lgtm but comments are pretty verbose. can you use the [concise-code-comments](https://github.com/onyx-dot-app/onyx-llm-context/blob/main/enforced/concise-code-comments/SKILL.md) skill?

mobile/src/api/auth/tokenStore.ts
nmgarza5: `WHEN_UNLOCKED_THIS_DEVICE_ONLY ` can we update this variable name a bit? This is just telling us the device is logged in? and only this device? Or can you add a code comment?

Subash-Mohan: Good call — renamed `TOKEN_OPTIONS` → `TOKEN_KEYCHAIN_OPTIONS` and added an inline comment (b54536c016).

To clarify: it's **not** a login signal. `keychainAccessible` controls *when the OS keychain will release the stored token*:

- **`WHEN_UNLOCKED`** — the entry is readable only while the device is unlocked (when locked, reads fail).
- **`THIS_DEVICE_ONLY`** — the entry is never synced to iCloud Keychain or restored onto a different device from an encrypted backup; it stays on this one physical device.

Combined: the token can be read only while *this* device is unlocked, and it can't leak to another device via backup/sync. (There's also a fuller writeup in SECURITY note #4 at the top of the file.)

PR #12165, opened 2026-06-17
fix(chat): surface context-window errors on long threads + overflow guard

justin-tahara:
## Description

Long chat threads could overflow a model's context window and fail with an **opaque** error ("There was an error with the response" / "model encountered an error during generation"), with no indication that the input was too long. This fixes it at three levels:

**1. Classifier ordering (`llm/utils.py`).** `ContextWindowExceededError` and `ContentPolicyViolationError` both subclass `BadRequestError`, but `BadRequestError` was checked first — so the `CONTEXT_TOO_LONG` / `CONTENT_POLICY` branches were **unreachable** and context overflow was mislabeled `BAD_REQUEST`. Check the subclasses first.

**2. Generation-path error surfacing (`chat/process_message.py`).** The per-model worker error was hardcoded to `error_code="MODEL_ERROR"` with a fixed "model encountered an error during generation" message, **bypassing the classifier** (a long-standing issue: the classifier is only wired on the outer/setup path, which generation errors never reach — present since at least v4.0, not a recent regression). Now the exception is classified once (in `_run_model`, where the llm is in scope) and the real code/message flows into **both** the streamed error packet and the persisted message. Context overflow now surfaces as `CONTEXT_TOO_LONG` + "Your input is too long…", `is_retryable=false`.

**3. Overflow guard (`chat/llm_loop.py` + `configs/model_configs.py`).** Hold back a safety margin (`GEN_AI_INPUT_TOKEN_SAFETY_MARGIN`, default 5%) below `max_input_tokens` when fitting history — headroom for tiktoken-vs-provider tokenizer drift, the common long-thread overflow cause.

## How Has This Been Tested?

- New unit tests (`tests/unit/onyx/llm/test_llm_exception_classification.py`) assert the classification order: `ContextWindowExceededError → CONTEXT_TOO_LONG` (was `BAD_REQUEST`), `ContentPolicyViolationError → CONTENT_POLICY`, plain `BadRequestError → BAD_REQUEST`. **3 pass.**
- Reproduced the original failure deterministically with a load-test mock that simulates a provider context-window rejection on long prompts; the error packet came back as `error_code=MODEL_ERROR` (the bug). With the classifier fix verified in isolation, that path now resolves to `CONTEXT_TOO_LONG`.
- Full end-to-end re-run against a main-based stack is pending (local dev stack is currently on an unrelated branch with a shared DB).

- [x] Override Linear Check

## Notes
- The margin mitigates tokenizer-drift overflow; a genuinely mis-configured `max_input_tokens` (advertised >> real) still needs correct model config. Proactive pre-call compression is a possible follow-up for the deepest robustness.

---
## Summary by cubic
Surfaces context-window overflows on long chats and adds a safety margin to prevent them. Worker errors now use classifier-derived codes/messages, with provider API keys redacted and margin config validated.

- **Bug Fixes**
  - Fixed exception classification order in `llm/utils.py` (`ContextWindowExceededError` → `CONTEXT_TOO_LONG`, `ContentPolicyViolationError` → `CONTENT_POLICY`; plain `BadRequestError` stays `BAD_REQUEST`).
  - Classify generation-path errors once in `_run_model` and reuse for streamed and persisted errors; generic exceptions now surface as `UNKNOWN_ERROR`.
  - Redact provider API keys at the source in `_run_model` so both streamed and persisted errors are safe.
  - Added `GEN_AI_INPUT_TOKEN_SAFETY_MARGIN` (default 0.05) and apply it in `chat/llm_loop.py` when fitting history; validate it in `[0, 1)`.

<sup>Written for commit 7fa50dc4a446275236138fa3248acb3c79c05287. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12165?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/chat/llm_loop.py
+++ b/backend/onyx/chat/llm_loop.py
@@ -28,6 +28,7 @@ from onyx.configs.app_configs import INTEGRATION_TESTS_MODE
 from onyx.configs.chat_configs import MAX_LLM_CYCLES
 from onyx.configs.constants import DocumentSource
 from onyx.configs.constants import MessageType
+from onyx.configs.model_configs import GEN_AI_INPUT_TOKEN_SAFETY_MARGIN
 from onyx.context.search.models import SearchDoc
 from onyx.context.search.models import SearchDocsResponse
 from onyx.db.engine.sql_engine import get_session_with_current_tenant
@@ -669,8 +670,14 @@ def run_llm_loop(
             raw_answer=None,
         )
 
-        # Pass the total budget to construct_message_history, which will handle token allocation
-        available_tokens = llm.config.max_input_tokens
+        # Pass the budget to construct_message_history, which handles token
+        # allocation. Hold back a safety margin below the advertised limit:
+        # our tiktoken estimate can undercount vs the provider's real tokenizer,
+        # so fitting right up to max_input_tokens risks a context-window
+        # rejection on long threads.
+        available_tokens = int(
+            llm.config.max_input_tokens * (1 - GEN_AI_INPUT_TOKEN_SAFETY_MARGIN)
+        )
         tool_choice: ToolChoiceOptions = ToolChoiceOptions.AUTO
         # Initialize gathered_documents with project files if present
         gathered_documents: list[SearchDoc] | None = (
--- a/backend/onyx/chat/llm_loop.py
+++ b/backend/onyx/chat/llm_loop.py
@@ -670,11 +670,8 @@ def run_llm_loop(
             raw_answer=None,
         )
 
-        # Pass the budget to construct_message_history, which handles token
-        # allocation. Hold back a safety margin below the advertised limit:
-        # our tiktoken estimate can undercount vs the provider's real tokenizer,
-        # so fitting right up to max_input_tokens risks a context-window
-        # rejection on long threads.
+        # Hold back a margin below max_input_tokens: our tiktoken estimate can
+        # undercount the provider's tokenizer and overflow the context window.
         available_tokens = int(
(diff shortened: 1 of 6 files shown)


justin-tahara:
Thanks — both points addressed in `4c8171c9c5` (flagged by greptile and cubic):

1. **Unredacted persisted error text (P1):** the API key is now redacted **at the source** in `_run_model`, so the tuple stored in `model_error_info` is safe for every consumer — both the streamed `StreamingError` and the message persisted by `_save_errored_message`. The `_drain_to_completion` redaction stays as an idempotent backstop that also covers the stack trace and the `str(item)` fallback path.

2. **`GEN_AI_INPUT_TOKEN_SAFETY_MARGIN` bounds (P2):** validated to `[0, 1)` at import — out-of-range (e.g. `>=1.0` or negative) now raises `ValueError` at startup instead of silently producing `<=0` or an inflated `available_tokens`. Verified default loads, `1.5` and `-0.1` both raise.

nmgarza5 (approved):
lgtm but comments are pretty verbose. can you use the [concise-code-comments](https://github.com/onyx-dot-app/onyx-llm-context/blob/main/enforced/concise-code-comments/SKILL.md) skill?

justin-tahara:
Done — ran the concise-code-comments skill over every touched file in `7fa50dc4a4`. Comments are now ≤2 lines and state present intent rather than change history. Thanks for the review!

PR #12170, opened 2026-06-17
feat(craft): Persist opencode session for docker compose path

acaprau:
## Description
Docker sandboxes lost all opencode chat history when a container was reaped -- only K8s persisted it. This brings the Docker backend to parity.

- Capture (`create_opencode_history_snapshot`): docker-execs the shared `sandbox_daemon` archive helper (sqlite-safe backup) and streams the tarball to the FileStore. The helper's data home is now `OPENCODE_DATA_HOME`-configurable, so K8s and Docker share one code path.
- Restore (`_maybe_restore_opencode_history`): on a fresh provision, `put_archive`s the snapshot into the still stopped container before start, so opencode-serve opens the restored DB instead of an empty one. Provision is now create -> restore -> start.
- Reliability: a `created`-state container (an incomplete prior provision) is removed and rebuilt rather than blindly started, so an interrupted/failed restore can't silently boot opencode against an empty data home. Restore failures are fatal and remove the half-built container.

## How Has This Been Tested?

Unit tests cover the provision/cleanup paths and the `created`-container handling. #12171 adds an integration test.

## Additional Options

---
## Summary by cubic
Persist opencode history (chat DB and sessions) for Docker Compose sandboxes. We snapshot from the running container and restore into new containers before start so sessions survive rebuilds.

- **New Features**
  - Added snapshot/restore via `sandbox_daemon.opencode_history`: build an in-container archive (Python exec), stream it to FileStore, and restore with Docker `put_archive` so it lands under `/workspace/.opencode-data`. The daemon now honors `OPENCODE_DATA_HOME` (Docker uses `/workspace/.opencode-data`; default stays `/workspace/opencode-data`).
  - Provision now does `docker create`, restores history into the stopped container, then starts it. If restore fails, the half-provisioned container is force-removed. Containers stuck in `'created'` are removed so we can re-create and restore; `'exited'` containers are started. Exposes `create_opencode_history_snapshot` and sets `supports_opencode_history_persistence = True`.

<sup>Written for commit d17549ca93c0399cf33f4ba40811b9001f50515e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12170?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/server/features/build/sandbox/docker/docker_sandbox_manager.py
+++ b/backend/onyx/server/features/build/sandbox/docker/docker_sandbox_manager.py
@@ -887,12 +887,20 @@ class DockerSandboxManager(SandboxManager):
             # starting the container, so entrypoint.sh's opencode-serve opens
             # the restored database instead of racing an extraction on top of a
             # live process. Mirrors the K8s init-sidecar restore handshake.
-            self._maybe_restore_opencode_history(container, sandbox_id, tenant_id)
+            #
+            # On any failure here, remove the half-provisioned container so it
+            # doesn't strand in 'created' state -- a retry's
+            # _reuse_existing_container would otherwise just start it and skip
+            # the restore, booting opencode against an empty data home while a
+            # good snapshot still sits in the FileStore. Restore failures are
+            # fatal (re-raised), matching the K8s fail-closed restore.
             try:
+                self._maybe_restore_opencode_history(container, sandbox_id, tenant_id)
                 container.start()
-            except APIError as e:
+            except Exception as e:
+                self._remove_incomplete_container(container)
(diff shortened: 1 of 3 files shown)


acaprau:
@greptile

wenxi-onyx (approved):

PR #12174, opened 2026-06-17
feat(service-accounts): add Groups & Roles modal

nmgarza5:
## Description

Adds a **Groups & Roles** action to each Service Accounts row, opening a modal (mirrors the Users page) that lets an admin:
- add/remove the service account from any user group
- change its role (Admin / Basic / Limited)

A service account is backed by a real user, so this reuses the existing user-group endpoints and the api-key role update — **no backend changes**. Current group membership is derived from the groups list (the service-account user already appears in each group's members). Role options are shared between the Account Type column and the modal via `SERVICE_ACCOUNT_ROLE_OPTIONS`.

## Additional Options


nmgarza5:
@greptile

Subash-Mohan (approved):

PR #12182, opened 2026-06-18
feat(indexing): docprocessing stage-metric coverage + residual

Subash-Mohan:
## Description

- Instruments the previously-untimed docprocessing regions — document & coordination lock acquisition, enrichment prep, finalization, and garbage collection — as new pipeline stages.
- Adds a computed `BATCH_UNACCOUNTED` residual so each batch's `BATCH_TOTAL` reconciles to its sub-stages, exposing lock-wait/contention in the admin per-connector stage-metrics view.
- Times lock acquisition separately from the held critical section and records it after release, so the instrumentation never lengthens the contended lock.

## How Has This Been Tested?

- Unit tests (residual math) + external-dependency tests (new-stage round-trip and acquire-only timing guard), all passing.

## Additional Options


evan-onyx (commented):
would like some of the comments to be a bit more concise but LGTM

evan-onyx (approved):

PR #12189, opened 2026-06-18
feat(settings): add custom chat retention duration option

justin-tahara:
## Description

Adds a **Custom…** option to the Chat History retention dropdown on the admin **Chat Preferences** page. Selecting it reveals a numeric **days** input, so admins can set any retention period to match their org policy — not just the presets.

This is a follow-up to #12186 (which added a 60-day preset); a customer asked for arbitrary durations. Putting it up so we can align on the UX before finalizing.

### Approach

Implemented as a self-contained `RetentionField` component that **drop-in replaces** the bare `<InputSelect>`. Frontend-only and backward compatible:

- The persisted shape (`maximum_chat_retention_days: number | null`) is **unchanged** — the backend already stores it as a free-form `float | None` with no preset validation, and the TTL cleanup task reads it raw. No backend/API/migration work.
- Any existing value round-trips correctly. A non-preset value (e.g. set via API) previously rendered **blank** in the trigger; it now displays under **Custom…**, so this also fixes a latent display gap.
- The `RetentionField` is the only place in the FE that interprets the value — every other reference treats it as `number | null`.
Screenshot 2026-06-18 at 10 09 23 AM

### Behavior / decisions to confirm

- **Presets retained** (Forever / 7 / 30 / 60 / 90 / 365) for one-click use, with **Custom…** added at the bottom. (Alternative: drop presets for a pure numeric field — happy to change.)
- **Save-on-blur** for the custom input; selecting "Custom…" does **not** persist until a valid number is entered, so the stored value isn't wiped mid-edit.
- **FE guard:** positive whole number, capped at **36500 days** (~100 years; the backend has no upper bound). Invalid/empty input reverts to the last saved value and shows an inline hint.
- **Enterprise-tier gating** is carried through to the numeric input.

## How Has This Been Tested?

Manual verification (the diff is FE-only and self-contained):
- `tsc --noEmit` — clean (exit 0, 0 errors).
- `oxfmt --check` and `oxlint` on the file — clean.
- Logic traced for the key paths: preset↔custom switching, blur with invalid/empty input (reverts), external value re-sync, and that selecting "Custom…" defers persistence.

No automated tests added (consistent with the existing untested presets). Happy to add a Playwright e2e if we want coverage on the save round-trip — flagging as an open question.

## Additional Options

---
## Summary by cubic
Adds a “Custom…” option to Chat History retention on the admin Chat Preferences page so admins can set any number of days. Presets remain; change is frontend-only and backward compatible.

- **New Features**
  - Introduced `RetentionField` that replaces the select and reveals a numeric days input when “Custom…” is chosen.
  - Persists `maximum_chat_retention_days` as-is (`number | null`); non-preset values now display under “Custom…”.
  - Save-on-blur with digits-only validation for positive whole numbers up to 36,500; invalid input reverts with an inline hint.
  - Enterprise-tier gating applies to the numeric input.

- **Refactors**
  - Rendered retention presets via a map over `RETENTION_PRESETS`.
  - Hoisted helper predicates to module scope for cleaner effects and consistent external value sync.

<sup>Written for commit 609e68303c35f6d7e68ed986670ef2198c0e6180. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12189?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/web/src/refresh-pages/admin/ChatPreferencesPage.tsx
+++ b/web/src/refresh-pages/admin/ChatPreferencesPage.tsx
@@ -469,11 +469,11 @@ function RetentionField({ value, disabled, onSave }: RetentionFieldProps) {
           <InputSelect.Item value={FOREVER_RETENTION_VALUE}>
             Forever
           </InputSelect.Item>
-          <InputSelect.Item value="7">7 days</InputSelect.Item>
-          <InputSelect.Item value="30">30 days</InputSelect.Item>
-          <InputSelect.Item value="60">60 days</InputSelect.Item>
-          <InputSelect.Item value="90">90 days</InputSelect.Item>
-          <InputSelect.Item value="365">365 days</InputSelect.Item>
+          {RETENTION_PRESETS.map((d) => (
+            <InputSelect.Item key={d} value={String(d)}>
+              {d} days
+            </InputSelect.Item>
+          ))}
           <InputSelect.Item value={CUSTOM_RETENTION_VALUE}>
             Custom…
           </InputSelect.Item>
--- a/web/src/refresh-pages/admin/ChatPreferencesPage.tsx
+++ b/web/src/refresh-pages/admin/ChatPreferencesPage.tsx
@@ -379,6 +379,19 @@ const MAX_RETENTION_DAYS = 36500; // ~100 years
 const CUSTOM_RETENTION_VALUE = "custom";
 const FOREVER_RETENTION_VALUE = "forever";
 
+// Pure predicate — lives at module scope so it can be referenced inside
+// useEffect without an exhaustive-deps suppression.
+const valueIsCustomRetention = (v: number | null): v is number =>
+  v !== null && !RETENTION_PRESETS.includes(v);
+
+// True only when the string is one or more digits within the allowed range.
+// parseInt alone would silently accept "1.5" → 1 or "7abc" → 7, so guard with
+// a digits-only check before persisting.
+const isValidCustomRetention = (raw: string): boolean =>
+  /^\d+$/.test(raw) &&
+  parseInt(raw, 10) > 0 &&
+  parseInt(raw, 10) <= MAX_RETENTION_DAYS;
+
 interface RetentionFieldProps {
   value: number | null;
   disabled: boolean;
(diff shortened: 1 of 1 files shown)


justin-tahara:
Thanks @cubic-dev-ai and @greptile-apps — both points were valid and are now fixed in 7f9651166b:

1. **Non-integer / trailing-garbage input** (`"1.5"` → `1`, `"7abc"` → `7`): added a digits-only `/^\d+$/` guard. Validation is centralized in a new `isValidCustomRetention(raw)` helper used by both the live error hint (`customInvalid`) and the blur handler, so such input is now rejected (shows the inline hint and reverts) instead of being silently truncated and persisted.
2. **`valueIsCustom` exhaustive-deps suppression**: hoisted the predicate to module scope as `valueIsCustomRetention`, so the external-sync `useEffect` no longer needs the `eslint-disable-line`.

Verified locally: `tsc --noEmit` clean, `oxfmt --check` and `oxlint` clean.

nmgarza5 (approved):

web/src/refresh-pages/admin/ChatPreferencesPage.tsx
```diff
@@ -370,6 +370,145 @@ function FileSizeLimitFields({
...
+        <InputSelect.Trigger />
+        <InputSelect.Content>
+          <InputSelect.Item value={FOREVER_RETENTION_VALUE}>
+            Forever
+          </InputSelect.Item>
+          <InputSelect.Item value="7">7 days</InputSelect.Item>
+          <InputSelect.Item value="30">30 days</InputSelect.Item>
+          <InputSelect.Item value="60">60 days</InputSelect.Item>
+          <InputSelect.Item value="90">90 days</InputSelect.Item>
+          <InputSelect.Item value="365">365 days</InputSelect.Item>
```
nmgarza5:
```suggestion
          {RETENTION_PRESETS.map((d) => (
		    <InputSelect.Item key={d} value={String(d)}>{d} days</InputSelect.Item>
		  ))}
```

nmgarza5 (approved):
sorry forgot to submit that last comment. approving but please update this

web/src/refresh-pages/admin/ChatPreferencesPage.tsx
justin-tahara: Good call — done in 609e68303c. Mapped over \`RETENTION_PRESETS\` so the preset list lives in one place. tsc/oxfmt/oxlint clean.

PR #12196, opened 2026-06-18
fix(craft): send tenant_id to PostHog so Craft flags can target one tenant

rohoswagger:
## Description

Onyx Craft is gated by the PostHog flag `onyx-craft-enabled`, and per-user usage limits by `craft-has-usage-limits`. Both checks evaluated their flags with **only the user's UUID and no person properties**, so PostHog never received a `tenant_id`. As a result a flag could not be targeted at a single tenant — enabling Craft (or unlimited usage) for `tenant_dev` would have required flipping the flag on for every individual user UUID. This blocks rolling Craft out to just one tenant.

This PR routes both checks through the existing (previously-unused) `FeatureFlagProvider.feature_enabled_for_user_tenant(flag, user, tenant_id)` helper, passing `get_current_tenant_id()`. That forwards `{tenant_id, email}` to PostHog as person properties, so a flag can now be scoped with a condition like `tenant_id = tenant_dev`.

- `backend/onyx/server/features/build/utils.py` — `is_onyx_craft_enabled`
- `backend/onyx/server/features/build/rate_limit.py` — `_should_skip_rate_limiting` (its docstring already states it "grants unlimited usage to dev tenant users (tenant_dev)", which this makes actually work)

The gate remains **fail-closed** (no client / `None` / exception → `False`), and the NoOp/env (`ENABLE_CRAFT`) fallback path is unchanged and never touches the tenant contextvar.

> Follow-up (config, not code): in the PostHog `onyx-craft-enabled` flag, add a release condition `tenant_id = tenant_dev`. The property is now sent on every evaluation, so the condition will match.

## How Has This Been Tested?

`backend/tests/unit/onyx/server/features/build/test_feature_gate.py` (4 tests, all passing):
- env/NoOp disable + enable paths (and a guard that the env path does **not** call `get_current_tenant_id`)
- PostHog flag overrides env, asserting `tenant_id` + email are forwarded as person properties
- PostHog flag returning `False` disables Craft even when `ENABLE_CRAFT=True`

Also ran `pre-commit` on all changed files: `ty` (type check), `ruff`, `ruff format`, secrets — all pass.

### Known minor items (for reviewer; not fixed here)
- `interface.feature_enabled_for_user_tenant` sends `email` as well as `tenant_id` to PostHog. Only `tenant_id` is needed for tenant targeting; dropping `email` would minimize PII egress. Left as-is since it's an existing helper-level behavior, out of this diff's scope.

## Additional Options

---
## Summary by cubic
Send `tenant_id` to PostHog when evaluating Craft flags so we can target a single tenant. Also gate Craft rate limits to Onyx Cloud only; self-hosted and managed deployments are unlimited.

- **Bug Fixes**
  - Evaluate `is_onyx_craft_enabled` via `feature_enabled_for_user_tenant(..., get_current_tenant_id())`.
  - Evaluate `_should_skip_rate_limiting` the same way; rate limits apply only when `MULTI_TENANT` and `AUTH_TYPE == CLOUD`.
  - Forward `tenant_id` (and email) as person properties to PostHog.
  - Per-tenant unlimited usage now works (e.g., `tenant_dev`).

- **Migration**
  - In PostHog, add a release condition to `onyx-craft-enabled` and/or `craft-has-usage-limits`: `tenant_id = tenant_dev` as needed.

<sup>Written for commit 57fcabd7bf8b64bf6683855b27ad36ec4aa05db7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12196?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/server/features/build/rate_limit.py
+++ b/backend/onyx/server/features/build/rate_limit.py
@@ -8,9 +8,10 @@ from typing import Literal
 from pydantic import BaseModel
 from sqlalchemy.orm import Session
 
+from onyx.configs.app_configs import AUTH_TYPE
+from onyx.configs.constants import AuthType
 from onyx.db.models import User
 from onyx.feature_flags.factory import get_default_feature_flag_provider
-from onyx.feature_flags.interface import NoOpFeatureFlagProvider
 from onyx.server.features.build.configs import CRAFT_PAID_USER_RATE_LIMIT
 from onyx.server.features.build.db.rate_limit import count_user_messages_in_window
 from onyx.server.features.build.db.rate_limit import count_user_messages_total
@@ -38,8 +39,9 @@ def _should_skip_rate_limiting(user: User) -> bool:
     """
     Check if rate limiting should be skipped for this user.
 
-    Currently grants unlimited usage to dev tenant users (tenant_dev).
-    Controlled via PostHog feature flag.
+    Grants unlimited usage to tenants/users whose `craft-has-usage-limits`
+    PostHog flag is off (e.g. tenant_dev). Only reached on Onyx Cloud, where
+    PostHog is configured.
 
     Returns:
         True to skip rate limiting (unlimited), False to apply normal limits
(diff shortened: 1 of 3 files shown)


wenxi-onyx (approved):

rohoswagger:
@greptile-ai @cubic-dev-ai @codex review this

PR #12208, opened 2026-06-18
fix(chat): keep processing indicator until user files finish indexing

justin-tahara:
## Description

When a user uploads a file in chat, the backend reports `token_count = null` until indexing finishes. The chat input treated a `null` token count as "0 tokens → fits in the context window → ready" and hid the **"Processing…"** indicator early. The file then looked ready while it was still being indexed, so a user could send a message referencing a not-yet-queryable file and get a degraded answer — it only worked correctly once indexing had finished (e.g. in a later session). This was especially visible for image-bearing PDFs, where image summarization can push processing to several minutes.

This keeps the processing state accurate and prevents premature sends:

- Don't hide the processing indicator until **every** still-processing attached file reports a real (non-null) token count.
- Disable the send button while any attached file is still uploading **or** processing, with a tooltip explaining why — mirroring the existing in-progress-upload gate.

Frontend-only change in `web/src/sections/input/AppInputBar.tsx`.

## How Has This Been Tested?

Manually against a local instance. Pinned the `/api/user/projects/file/statuses` poll to `PROCESSING` (`token_count: null`) and confirmed the file card keeps its spinner/"Processing…" label and the send button stays disabled with the explanatory tooltip. Releasing to `COMPLETED` clears the spinner and re-enables send. Also confirmed the inverse (`token_count` set while `PROCESSING`) reproduces the old premature-hide behavior. `tsc`, `oxlint`, and `oxfmt` all pass.

## Additional Options

---
## Summary by cubic
Fixes the chat input showing files as ready before indexing completes. Keeps the “Processing…” indicator and disables Send until all attached files finish indexing and return a non-null token count, with a tooltip explaining the wait.

- **Bug Fixes**
  - Use an explicit null check (`token_count !== null`) so not-yet-indexed files don’t look ready.
  - Stub `/api/user/projects/file/statuses` in e2e tests to return `COMPLETED` to avoid racing the real indexing pipeline.

<sup>Written for commit ea0cfb7d8f61d868c0750769e4e52fa29b908869. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12208?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/web/src/sections/input/AppInputBar.tsx
+++ b/web/src/sections/input/AppInputBar.tsx
@@ -280,6 +280,9 @@ const AppInputBar = React.memo(
       );
     }, [currentMessageFiles]);
 
+    // A file isn't queryable until indexing completes, so gate send on it.
+    const hasIndexingFiles = currentIndexingFiles.length > 0;
+
     // Convert ProjectFile to MinimalOnyxDocument format for viewing
     const handleFileClick = useCallback(
       (file: ProjectFile) => {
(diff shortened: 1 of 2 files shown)


justin-tahara:
Addressed the review feedback from cubic and greptile in 323cf8ce91: switched `(file.token_count ?? 0) > 0` to a direct `file.token_count !== null` check so an indexed-but-empty file (0 tokens) isn't conflated with a not-yet-indexed one. `token_count` is typed `number | null`, so this is exact.

nmgarza5 (approved):

PR #12212, opened 2026-06-18
feat(search): scope internal search to connectors relevant to conversation

Danelegend:
## What

Internal search now decides, per call, which connected source(s) to cover — driven by the persona/user routing instructions and the current request. This lets an agent be told e.g. *"check Zendesk first, then Confluence, then Google Drive"* and have the search walk those sources one at a time, falling back to the next (and finally everything) on repeat calls.

## How

- **`decide_search_scope`** (new secondary LLM flow): one cheap LLM call, fed **only the user-side turns** (persona instructions + the asks — assistant/tool turns excluded, so the request always ends on a user message) plus the **set of sources already searched this turn**. Returns `(scope_now, next_source)`. Fails open to "no scope" (search everything) on any error.
- The walk is **stateless in the flow** — "already tried" is supplied by the caller (`SearchTool._tried_scopes`), so the LLM advances the order deterministically without a cached plan.
- A persona/user **source restriction is the outer bound** the flow refines within, never exceeded.
- The resolved scope is threaded into retrieval filters + federated/Slack backends, and summarized in a **note back to the agent** so it knows what was searched and whether searching again changes the scope.
- Removes the unused legacy `extract_source_filter` / `SOURCE_FILTER_PROMPT`.

## Notes

- This PR is the **filtering implementation only**. Surfacing the applied filter in the UI (the `SearchToolFilterDelta` packet + frontend chip) is a follow-up PR stacked on this one.
- An extra secondary LLM call runs on the first search whenever 2+ sources are connected; it's parallelized with query expansion so latency is hidden. With 0–1 candidate sources it's skipped.

## Tests

- `test_source_filter.py` — `decide_search_scope` builds a user-only prompt that ends on a user message (regression for the Anthropic "must end with a user message" 400).
- `test_search_tool_run.py` — resolved scope is passed to retrieval; persona restriction is honored / refined.
- A gated real-LLM eval suite (`RUN_CONNECTOR_FILTER_EVAL`) exists locally and will land separately once de-flaked.

---
## Summary by cubic
Scopes each internal search call to the right connector(s) based on persona/user routing, with a per-call source walk that handles combined vs backoff directives across turns. Queries stay content-only, first-call expansion is reused when advancing to a new source, federated backends are filtered (Slack disabled when out of scope), and a scope note is returned in the LLM payload and empty-result messages.

- **New Features**
  - `decide_search_scope` uses only the last 5 user turns, this cycle’s queries, and previous cycles (queries + searched_sources); returns a bracketed list limited to connected sources, treats `[]` as unscoped, runs only with 2+ candidate sources, never infers from topic, and always stays within any persona/user `source_type` restriction. It distinguishes combined vs backoff directives and advances one source per call when ordered.
  - `SearchTool` tracks per-turn `_search_cycles`, reuses cached expansion when switching to a not-yet-searched source, applies the resolved scope to retrieval and federated backends, filters federated connectors to in-scope sources, disables Slack when out of scope, and appends a scope note (searched sources + queries run) to the LLM response and empty-result messages.
  - Query hygiene: tool schema and expansion prompts strip connector/app names (scoping is automatic). New `parse_bracketed_list` parses output like `[zendesk, asana]`, tolerates stray text, and uses the last list.

- **Bug Fixes**
  - Hardened model output parsing: `parse_llm_json_response` returns the first complete JSON object; the bracketed-list parser restricts to connected sources and uses the last list to avoid fail-open to unscoped search.
  - Scope decision prompt is a single user message and excludes assistant/tool turns.

<sup>Written for commit e0037a0feb507572711aa767601b6ba35179a2c2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12212?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Danelegend:
@greptile @cubic

backend/onyx/tools/tool_implementations/search/search_tool.py
```diff
@@ -656,11 +690,33 @@ def run(
...
+            allowed = set(user_source_restriction)
+            candidate_sources = [s for s in connected_sources if s in allowed]
+        else:
+            candidate_sources = connected_sources
+
+        decide_args = (
+            message_history,
+            self.llm,
+            candidate_sources,
+            list(self._search_cycles),
```
evan-onyx, line 709: is this intentionally a shallow copy? i think no need to copy here but if there is a need probably should copy() explicitly

evan-onyx (approved):

PR #12222, opened 2026-06-19
fix(file-processing): fall back to pypdf when PDFium text extraction fails

nmgarza5:
## Description

**Bug:** Since the switch to pypdfium2 (#12102), PDFs that pypdf can read but PDFium rejects (`FPDF_ERR_FORMAT`, "Data format error") get indexed with **no text** — their content is silently dropped.

**Why:** PDFium is stricter than pypdf. On a Google Drive connector, 100+ files per indexing window logged "PDFium: Data format error" while pypdf accepted the same files (0 pypdf-level rejections), so their text never reached the index.

**Fix:** When PDFium raises, fall back to pypdf's `extract_text()`. PDFium stays the primary path (keeps the GIL-release / heartbeat protection from #12102); only PDFium-rejected files take the slower pypdf path.

**Test:** unit test asserting the fallback extracts text when PDFium raises.

## Additional Options

### DIFF
--- a/backend/onyx/file_processing/extract_file_text.py
+++ b/backend/onyx/file_processing/extract_file_text.py
@@ -307,6 +307,7 @@ def read_pdf_file(
     """
     from pypdf import PdfReader
     from pypdf.errors import PdfStreamError
+    from pypdfium2 import PdfiumError
 
     metadata: dict[str, Any] = {}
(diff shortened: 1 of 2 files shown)


nmgarza5:
@greptile review

evan-onyx (approved):

PR #12237, opened 2026-06-19
fix(google-drive): bound advanced Doc parse to prevent indexing-worker OOM

nmgarza5:
## Description

**Bug:** Google Drive indexing OOMs and crash-loops on tenants with large Google Docs — the indexing worker exceeds its memory limit, the watchdog kills it, and the attempt auto-reschedules.

**Why:** A Google Doc's advanced parse (`get_document_sections`) loads the full Docs-API structural JSON, which is far heavier than the doc's text and isn't subject to any size limit (unlike exports). The prior text-export size gate (#12223) let structure-heavy / text-light Docs through, and they still OOM'd.

**Fix:** Stream the Docs-API response under a byte cap (`GOOGLE_DRIVE_ADVANCED_PARSE_MAX_BYTES`, default 50 MB). Docs over the cap fall back to the size-bounded basic text export — still indexed (full text), just whole-doc links instead of section-level links. Also cap the O(n²) `align_basic_advanced` pass by section count, and close the streamed session on every path. Adds a `get_impersonated_creds` helper (reused by the service builder).

Exports (Docs-text / Sheets-CSV / Slides) are already size-bounded; this closes the last unbounded native-Docs path.

## Additional Options

### DIFF
--- a/backend/onyx/connectors/google_drive/section_extraction.py
+++ b/backend/onyx/connectors/google_drive/section_extraction.py
@@ -96,10 +96,9 @@ def get_document_sections(
 ) -> list[TextSection] | None:
     """Extract heading-aware sections from a Google Doc.
 
-    Streams the Docs-API response under a byte cap. Returns None when the
-    structural payload exceeds `max_response_bytes`, so the caller can fall back
-    to the size-bounded basic text export instead of loading a structurally
-    unbounded doc into memory. `includeTabsContent` pulls every tab's content.
+    Streams the Docs-API response and returns None once it exceeds
+    `max_response_bytes`, so the caller can fall back to the basic text export
+    rather than load an unbounded structural payload into memory.
     """
     impersonated_creds = get_impersonated_creds(creds, user_email)
     with AuthorizedSession(impersonated_creds) as session:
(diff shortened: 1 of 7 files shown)


backend/onyx/connectors/google_drive/section_extraction.py
```diff
@@ -79,27 +83,39 @@ def _extract_text_from_table(table: dict[str, Any]) -> str:
...
-        documentId=doc_id
-    )
+    user_email: str,
+    max_response_bytes: int,
+) -> list[TextSection] | None:
+    """Extract heading-aware sections from a Google Doc.
+
+    Streams the Docs-API response; returns None if it exceeds `max_response_bytes`.
+    """
+    impersonated_creds = get_impersonated_creds(creds, user_email)
```
evan-onyx: nit: would like to see the creation of this session be done in google_utils/resources.py similar to how docs and drive services are created there. imo all google auth should be centralized there and this function should take the authorizedsession as a parameter like it used to take a Docs Service.

Ideally I wouldn't even want to add another means of talking to drive (just keep everything flowing through the SDK's Resource objects) but I think it's necessary to get this behavior so we can just treat this as a new method for contacting drive that resources.py exposes.

evan-onyx (approved):

nmgarza5:
@greptile

PR #12238, opened 2026-06-19
fix(agent-sharing): "Your Agents" tab lists owned + shared-edit agents

nmgarza5:
## Description

The "Your Agents" tab listed only agents you *own*, and its section header always read "All Agents". Filtering by edit-access alone doesn't work — admins are EDITOR on every agent, so that would list every agent for an admin.

Fix:
- Backend: new server-computed `user_is_owner_or_editor` flag on the agent-list payload — true for owners and explicit editor shares, with admins' blanket access excluded (`get_persona_access_level(..., treat_admin_as_editor=False)`).
- Frontend: the "your" tab filters by that flag; the section header reads "Your Agents" on that tab.

Note: a public agent with `public_permission=EDITOR` counts as "yours" for everyone (literal reading of "have EDIT access"). Trivial to drop if undesired.

## Additional Options

### DIFF
--- a/backend/onyx/db/persona.py
+++ b/backend/onyx/db/persona.py
@@ -45,6 +45,7 @@ from onyx.db.notification import create_notification
 from onyx.db.persona_sharing import get_persona_access_level
 from onyx.db.persona_sharing import get_user_group_ids_for_user
 from onyx.db.persona_sharing import persona_ownership_is_vacant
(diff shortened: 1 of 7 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

PR #12253, opened 2026-06-20
fix(agent-sharing): show display names in the share dialog

nmgarza5:
## Description

The share dialog and Add People picker rendered raw emails for every user. Now they show the user's display name when set, with the email as the secondary line.

- Backend: `MinimalUserSnapshot` gains `personal_name` (from `User.personal_name`), populated in the persona share snapshots and the `/users` picker source.
- Frontend: name-over-email in the owner row, user-share rows, picker suggestions, and staged tags; falls back to email when no name is set.

## Additional Options


nmgarza5:
@greptile

PR #12261, opened 2026-06-22
Add Keenable web search provider (keyless by default)

ilya-bogin-keenable:
Adds Keenable as a web search provider, alongside the existing
Exa/Serper/Brave/SearXNG/Google PSE providers.

The difference from the other providers is that it doesn't require an API key.
By default it uses Keenable's keyless public endpoint, so it works out of the
box (like SearXNG, but with no URL to configure). Passing an API key uses the
authenticated endpoint and lifts the rate limits.

### Changes

- `web_search/clients/keenable_client.py`: `KeenableClient(WebSearchProvider)`
  with `search()` + `test_connection()`, following the Serper/Brave clients.
- `providers.py`: keyless build branch (handled before the API-key check) and
  `provider_requires_api_key()` returns `False` for Keenable.
- `WebSearchProviderType.KEENABLE` enum value.
- Frontend: provider detail + capabilities (`requiresApiKey: false`) in
  `lib/webSearch`.

### Config

- API key: optional. Blank = keyless free tier; set it to lift rate limits.
- `base_url` config (optional): override the API base (HTTPS; defaults to
  `https://api.keenable.ai`).

Verified the client's search path against the public endpoint (keyless).

---
## Summary by cubic
Adds Keenable as a web search provider that works without an API key by default. Previously all non-`searxng` providers required a key; now Keenable uses a keyless public endpoint, and adding a key lifts rate limits.

- Backend: adds `WebSearchProviderType.KEENABLE`, `KeenableClient`, a keyless branch in the factory, and sets `provider_requires_api_key()` to false for Keenable.
- Endpoint selection: `/v1/search/public` when no key; `/v1/search` when a key is set.
- Resilience: retries only transient failures (429/5xx/network) via `RetryableKeenableSearchError`; `test_connection` reports invalid keys (401/403) and rate limits.
- Result mapping: read `snippet` (fallback to `description`), collapse whitespace, cap to 500 chars; unit tests cover mapping and edge cases.
- Config: default base URL `https://api.keenable.ai`; require HTTPS (allow HTTP only for localhost).
- Frontend: adds provider with `requiresApiKey: false`, subtitle "Keenable Search API", and ships the real Keenable brandmark.

<sup>Written for commit 524aebeff3b1dd7547af47a14335e0a47a730a9f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12261?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ilya-bogin-keenable:
@justin-tahara @jmelahman ping on this. It adds Keenable as a sixth web search provider, keyless by default, following the shape of the existing five. Still mergeable against main. Anything you want changed before it goes to review?

ilya-bogin-keenable:
Pushed a correctness fix (`2d6a935`) found while auditing our integrations.

The client read each result's snippet from the API's `description` field. Keenable returns **both** `description` and `snippet`: `description` is frequently empty and `snippet` carries the page text. Verified against the live keyless endpoint: **10/10 results came back with `len(description) == 0` and `len(snippet)` between 2000 and 2600**, so `WebSearchResult` was reaching the model with a title, a link and an empty snippet.

It now reads `snippet` first with a `description` fallback. One judgement call worth reviewing: Keenable returns whole-page text where Tavily/Serper/Brave return a short snippet, so the text is whitespace-collapsed (it arrives with newlines) and capped at 500 characters via `KEENABLE_MAX_SNIPPET_CHARS`, keeping `WebSearchResult.snippet` preview-sized. The full page is still reachable through the contents step, so nothing is lost.

Also adds `backend/tests/unit/.../websearch/test_keenable_client.py` — the mapping against a realistic fixture (empty `description`, populated `snippet`), the fallback, the collapse and the cap, and link-less rows. That fixture is the point: the bug was invisible to any mock that populates `description`, which no real response does. This also closes the earlier cubic-dev-ai note about the client lacking test coverage.

Pinned `ruff==0.12.0` (from `requirements/dev.txt`) `check` + `format --check` clean, and the new tests pass.

PR #12266, opened 2026-06-22
fix(llm): resolve vision support for dynamic-provider models + clearer 413

justin-tahara:
## Description

Fixes image input being blocked for models served through dynamic/aggregator LLM providers (Bifrost, OpenRouter, custom OpenAI-compatible gateways), reported on v4.1.4 with `vertex/gemini-3-pro-image-preview` via a Bifrost provider.

Two distinct failures were happening on the same model:

1. **"The current model does not support image input" toast on new uploads.** The model was stored with only a `CHAT` capability flow, so the model-config API reported `supports_image_input=false`. A prior fix (#12157) corrected the *fetch* path to consult the LiteLLM cost map, but it never reached models that were already configured.
2. **`413 Request Entity Too Large` on previously-uploaded images.** This originates from nginx in front of the customer's gateway rejecting the base64 image payload — it surfaced as a raw nginx HTML blob in the UI.

This PR addresses the Onyx side of both:

- **`ModelConfigurationView.from_model`** — fall back to the LiteLLM cost map for vision on *all* dynamic providers, not only `custom_config` ones. The previously-synced model now resolves `supports_image_input` correctly at read time, so an upgrade fixes the toast with **no admin action required** (no delete/recreate).
- **`sync_model_configurations`** — on re-fetch, upgrade an existing model's capability flows (`VISION`/`REASONING`) when the source newly reports support. Flags are only ever *added*, never removed, so admin overrides and user prefs are preserved. Previously re-fetch only inserted new models, so a stale row stayed broken forever — the reason #12157 didn't help already-configured models.
- **`litellm_exception_to_error_msg`** — map an upstream HTTP 413 to an actionable message ("increase the gateway's max request body size, e.g. nginx `client_max_body_size`") instead of dumping raw nginx HTML.

Note: the 413 itself is a gateway-side limit (the operator must raise `client_max_body_size`); this PR makes the failure self-explanatory rather than silently downscaling image-edit inputs.

## How Has This Been Tested?

Unit tests (all passing):
- `test_model_configuration.py` — dynamic provider with no stored `VISION` flow but cost-map vision support → `supports_image_input=true`; and stays `false` when the cost map is unaware.
- `test_llm_sync.py` — existing model gains a `VISION` flow on re-fetch when newly reported; flags are never removed; no commit when nothing changes.
- `test_llm_exception_classification.py` — upstream 413 (by status code and by body text) → `REQUEST_TOO_LARGE` with an actionable message, raw HTML not surfaced.
- `test_fetch_models_api.py` — updated existing-model sync expectation for flow reconciliation.

`pre-commit run --files` (ty, ruff, ruff format) passes on all changed files.

## Additional Options

---
## Summary by cubic
Fixes blocked image uploads for models behind dynamic providers (Bifrost, OpenRouter, OpenAI-compatible gateways) and shows a clear, actionable message for HTTP 413 errors. Existing models now resolve vision support correctly without admin action; re-syncs add missing capability flows.

- **Bug Fixes**
  - `ModelConfigurationView.from_model`: fall back to the LiteLLM cost map for vision on all dynamic/aggregator providers so `supports_image_input` is correct even when no `VISION` flow is stored.
  - `sync_model_configurations`: on re-fetch, add missing capability flows (`VISION`, `REASONING`) to existing models; never remove flags; commit when new or upgraded. Caveat: admin-removed flows are re-added on sync (Linear ENG-4233).
  - `litellm_exception_to_error_msg`: classify HTTP 413 as `REQUEST_TOO_LARGE` (including when only present in the body) and advise increasing the gateway’s max body size instead of surfacing raw HTML.

- **Migration**
  - No admin changes needed for vision support; upgrade and re-fetch will reconcile flows.
  - If you hit 413 errors, increase the gateway’s request body limit (e.g., `nginx` `client_max_body_size`).

<sup>Written for commit 642309066fbedf591988dbc343c835dbe20471f2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12266?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/db/llm.py
+++ b/backend/onyx/db/llm.py
@@ -367,8 +367,10 @@ def sync_model_configurations(
 ) -> int:
     """Sync model configurations for a dynamic provider (OpenRouter, Bedrock, Ollama, etc.).
 
-    This inserts NEW models from the source API without overwriting existing ones.
-    User preferences (is_visible, max_input_tokens) are preserved for existing models.
+    Inserts NEW models from the source API and, for models that already exist,
+    upgrades their capability flows (VISION/REASONING) when the source newly
+    reports support. Flags are only ever added, never removed, so admin overrides
+    and user preferences (is_visible, max_input_tokens) are preserved.
 
     Args:
         db_session: Database session
@@ -382,12 +384,13 @@ def sync_model_configurations(
     if not provider:
         raise ValueError(f"LLM Provider '{provider_name}' not found")
 
-    # Get existing model names to count new additions
-    existing_names = {mc.name for mc in provider.model_configurations}
+    existing_by_name = {mc.name: mc for mc in provider.model_configurations}
 
     new_count = 0
+    upgraded_flow_count = 0
     for model in models:
-        if model.name not in existing_names:
+        existing = existing_by_name.get(model.name)
+        if existing is None:
             # Insert new model with is_visible=False (user must explicitly enable)
             supported_flows = [LLMModelFlowType.CHAT]
             if model.supports_image_input:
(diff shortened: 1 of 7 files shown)


backend/onyx/db/llm.py
```diff
@@ -405,8 +408,29 @@ def sync_model_configurations(
                 display_name=model.display_name,
             )
             new_count += 1
+            continue
+
+        # Existing model: add any newly-reported capability flows so a re-fetch
+        # repairs rows that were synced before the source exposed the capability.
```
jmelahman: Do you know if this may adversely impact custom model configurations? For example, if a user wanted to _remove_ vision capability from a model which would otherwise support it, would this overwrite that configuration? That's probably fine, but maybe worth tracking separately if that is true.

jmelahman (approved):

backend/onyx/db/llm.py
justin-tahara: Good catch — you're right. The sync is additive-only, so a flow an admin deliberately removed (e.g. setting a custom-config model to Text-Only via CustomModal) gets silently re-added on the next "fetch models". `is_visible`/`max_input_tokens` overrides are preserved, but flow-type removals aren't durable.

I'd rather not expand the scope of this PR (it's a customer-facing fix we want in + cherry-picked), so I've tracked the durable fix as ENG-4233 and left a `TODO(ENG-4233)` plus a docstring caveat here. The ticket also folds in the per-model lazy-load N+1 that the bots flagged on the same loop.

PR #12274, opened 2026-06-22
feat(celery): self-heal wedged consumers via liveness probe

nmgarza5:
## Description

**What this fixes:**

Background workers pull small jobs off queues and run them. Kubernetes decides whether a worker pod is healthy by watching a "liveness" file the worker keeps touching — if that file goes stale, k8s restarts the pod.

The bug: a worker can end up **alive but not working**. The process keeps running, keeps touching the liveness file on a timer, and still gossips with the other workers — but it has silently **stopped pulling jobs off the queue**. Work piles up while every thread sits idle. Because the timer keeps the file fresh, k8s never notices, so the pod is never restarted.

This isn't hypothetical — a light-worker fleet got stuck exactly like this for **~51 hours**, with its queue climbing into the hundreds of thousands and nothing draining, until someone manually restarted it. The liveness check was effectively lying: it confirmed "the process is running," not "the worker is doing its job."

**The fix:** make liveness actually mean something. A worker now reports healthy only if it's genuinely consuming work — or has nothing to do. If it has gone too long without starting a job **while jobs are clearly waiting and nothing is running**, it stops touching the liveness file and k8s restarts it automatically (~10 minutes instead of 51 hours / a human noticing).

It is **opt-in per deployment** (off by default) and **fails open** — any uncertainty leaves the worker alive, so a healthy pod is never restarted.

### How it works (technical)

- `on_task_prerun` stamps a monotonic timestamp every time the worker *starts* a task (`mark_task_consumed`).
- The existing `LivenessProbe` bootstep (15s timer) withholds the file touch only when **all three** hold:
  1. no task has started within `CELERY_LIVENESS_WEDGE_STALE_THRESHOLD_S` (the opt-in knob; `0` = off),
  2. `celery.worker.state.active_requests == 0` — nothing is actually running, so a *busy/saturated* worker on long tasks is never mistaken for a wedged one,
  3. there is queued work — broker `LLEN` over the worker's own queues **+** `reserved_requests` (already prefetched).
- Watched queues are derived from the worker's runtime `-Q` (`amqp.queues.consume_from`) — no separate list to keep in sync.
- **Fails open** on every "can't tell" (broker unreachable, unreadable counts, malformed env) → keep the worker alive.
- The stale file then trips the existing k8s liveness probe (mtime > 60s, x5) → pod restart.

**Scope:** off by default, so other fleets are unaffected. The cloud light deployment sets the env var separately (deploy repo) — lands after this image builds.

**What this does NOT do:** it's a self-healing safety net for the *symptom* (a wedged consumer). It does not fix the root cause of the wedge, and is unrelated to the prune-queue / monitoring-OOM issues tracked elsewhere.

## Additional Options

### DIFF
--- a/backend/onyx/background/celery/apps/app_base.py
+++ b/backend/onyx/background/celery/apps/app_base.py
@@ -24,10 +24,21 @@ from sqlalchemy.orm import Session
 from onyx.background.celery.apps.task_formatters import CeleryTaskColoredFormatter
 from onyx.background.celery.apps.task_formatters import CeleryTaskJsonFormatter
 from onyx.background.celery.apps.task_formatters import CeleryTaskPlainFormatter
+from onyx.background.celery.celery_consumer_liveness import active_request_count
+from onyx.background.celery.celery_consumer_liveness import BACKLOG_UNKNOWN
+from onyx.background.celery.celery_consumer_liveness import mark_task_consumed
+from onyx.background.celery.celery_consumer_liveness import reserved_request_count
+from onyx.background.celery.celery_consumer_liveness import seconds_since_last_consumed
+from onyx.background.celery.celery_consumer_liveness import (
+    should_withhold_liveness_touch,
+)
+from onyx.background.celery.celery_redis import celery_get_broker_client
+from onyx.background.celery.celery_redis import celery_get_queue_length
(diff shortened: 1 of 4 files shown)


nmgarza5:
closing for now

PR #12277, opened 2026-06-22
feat(opal): add `titleMaxLines` to `Content` + refactor `@opal/layouts.inputs`'s error layout

raunakab:
## Description

Three related Opal improvements that unblock the auth layout work, plus a follow-up tightening of `InputErrorText` types.

**`titleMaxLines` on `Content`**
- Adds `titleMaxLines?: number` to `ContentBaseProps` (default `1`, preserving existing truncation behaviour)
- Threads through `ContentXl`, `ContentLg`, `ContentMd`, `ContentSm`
- Replaces the ad-hoc `titleWrap?: boolean` that existed on `ContentMd` and `MessageCard` with a uniform numeric prop

**`"warning"` `ColorType`**
- Adds `"warning"` to `ColorTypes` union in `types.ts`
- Adds `[data-content-color="warning"]` CSS rule mapping to `--status-warning-05`

**`InputErrorText` TODO resolved**
- Replaces the manual `Section + Icon + span + Text` layout with `Content sizePreset="secondary" variant="body"`
- Uses `color="danger" | "warning"`, `role="alert"`, `titleMaxLines={undefined}` (free wrapping)
- Narrows `children` from optional `ReactNode` to required `string | RichStr`; removes the `?? ""` null-coalescing fallback on `title`
- Adds `meta.error!` non-null assertion inside the `hasError` guard (TypeScript can't narrow through it otherwise)
- Also adds `role?: string` to `ContentSm` so ARIA roles can be forwarded to the title text element
- Migrates `BillingDetailsView` JSX children → `markdown()` RichStr
- Migrates `ToastProvider` `titleWrap` → `titleMaxLines={undefined}`

**Export/type cleanup in `inputs/components.tsx`**
- `VerticalProps`, `HorizontalProps`, and `InputErrorType` are no longer `export`-prefixed on their declarations; they are collected into a single explicit export block at the bottom of the file, consistent with the rest of Opal

## Additional Options

---
## Summary by cubic
Adds line clamping to titles across `Content` and `MessageCard` with a new `titleMaxLines` prop, defaulting to free wrapping. Adds a `warning` color and refactors input error UI to use `Content` for better accessibility and consistency, and removes a dead import.

- **New Features**
  - `titleMaxLines?: number` on `ContentXl`, `ContentLg`, `ContentMd`, `ContentSm`, and `MessageCard` (default free-wrap; pass a number to clamp).
  - Added `warning` to `ColorTypes` with `[data-content-color="warning"]` styles.
  - `ContentSm` now accepts `role` to forward ARIA roles to the title text.

- **Migration**
  - Replace `titleWrap` with `titleMaxLines` where used.
  - `InputErrorText` now renders via `Content` with `color="danger" | "warning"`, `role="alert"`, and free title wrapping; its `children` is now required and typed as `string | RichStr`.
  - Updated call sites: billing error copy uses `markdown()`, and `ToastProvider` relies on the default free-wrap behavior (no prop needed).

<sup>Written for commit a5a2774109da0219bd029a5b2131baab3fe18fd0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12277?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/web/lib/opal/src/components/cards/message-card/components.tsx
+++ b/web/lib/opal/src/components/cards/message-card/components.tsx
@@ -35,8 +35,8 @@ interface MessageCardBaseProps {
   /** Optional description below the title. */
   description?: string | RichStr;
 
-  /** Let the title wrap to multiple lines instead of truncating to one. */
-  titleWrap?: boolean;
+  /** Clamp the title to N lines with ellipsis. Default: `1`. Pass `undefined` to wrap freely. */
+  titleMaxLines?: number;
 
   /** Padding preset. @default "sm" */
   padding?: Extract<PaddingVariants, "sm" | "xs">;
@@ -127,7 +127,7 @@ function MessageCard({
   icon: iconOverride,
   title,
   description,
-  titleWrap,
+  titleMaxLines,
   padding = "sm",
   headerPadding = "fit",
   bottomChildren,
@@ -164,7 +164,7 @@ function MessageCard({
           )}
           title={title}
           description={description}
-          titleWrap={titleWrap}
+          titleMaxLines={titleMaxLines}
           sizePreset="main-ui"
           variant="section"
           padding="md"
--- a/web/src/app/admin/billing/BillingDetailsView.tsx
+++ b/web/src/app/admin/billing/BillingDetailsView.tsx
@@ -2,6 +2,7 @@
 
 import { useState } from "react";
 import Link from "next/link";
+import { markdown } from "@opal/utils";
 import { Section } from "@/layouts/general-layouts";
 import { Content, InputErrorText, InputVertical } from "@opal/layouts";
 import Card from "@/refresh-components/cards/Card";
@@ -484,16 +485,9 @@ function SeatsCard({
 
             {isBelowMinimum ? (
               <InputErrorText type="error">
-                You cannot set seats below current{" "}
-                <span className="font-semibold">{minRequiredSeats}</span> seats
-                in use/pending.{" "}
-                <Link
-                  href="/admin/users"
-                  className="underline hover:no-underline"
-                >
-                  Remove users
-                </Link>{" "}
-                first before adjusting seats.
+                {markdown(
+                  `You cannot set seats below current **${minRequiredSeats}** seats in use/pending. [Remove users](/admin/users) first before adjusting seats.`
+                )}
               </InputErrorText>
             ) : seatDifference !== 0 ? (
               <Text secondaryBody text03>
--- a/web/src/app/admin/billing/BillingDetailsView.tsx
+++ b/web/src/app/admin/billing/BillingDetailsView.tsx
@@ -1,7 +1,6 @@
 "use client";
 
 import { useState } from "react";
-import Link from "next/link";
 import { markdown } from "@opal/utils";
 import { Section } from "@/layouts/general-layouts";
 import { Content, InputErrorText, InputVertical } from "@opal/layouts";
--- a/web/src/providers/ToastProvider.tsx
+++ b/web/src/providers/ToastProvider.tsx
@@ -123,7 +123,7 @@ function ToastContainer() {
             <MessageCard
               variant={LEVEL_TO_VARIANT[t.level ?? "info"]}
               title={truncatedTitle}
-              titleWrap
+              titleMaxLines={undefined}
               description={buildDescription(t)}
               padding="xs"
               onClose={t.dismissible ? () => handleClose(t.id) : undefined}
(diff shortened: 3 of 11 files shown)


web/src/providers/ToastProvider.tsx
```diff
@@ -123,7 +123,6 @@ function ToastContainer() {
             <MessageCard
               variant={LEVEL_TO_VARIANT[t.level ?? "info"]}
               title={truncatedTitle}
-              titleWrap
```
nmgarza5, line 126: you're removing this but not adding in `titleMaxLines`. this will cause a regression. i added titleWrap for a reason because some toast notifications are way too long. most error handling toasts need this

web/src/app/admin/billing/BillingDetailsView.tsx
```diff
@@ -484,16 +485,9 @@ function SeatsCard({
...
-                <Link
-                  href="/admin/users"
-                  className="underline hover:no-underline"
-                >
-                  Remove users
-                </Link>{" "}
-                first before adjusting seats.
+                {markdown(
+                  `You cannot set seats below current **${minRequiredSeats}** seats in use/pending. [Remove users](/admin/users) first before adjusting seats.`
+                )}
```
nmgarza5, line 489: how does this redirect for `/admin/users` still work? the markdown util just renders markdown but doesnt have any routing stuff baked in iiuc

nmgarza5 (commented):
some comments. i think this will cause 2 small regressions

web/src/providers/ToastProvider.tsx
raunakab: `titleMaxLines` defaults to `undefined`, which then enforces no maximum number of lines on the component. This is effectively the same as `titleWrap`.

nmgarza5: is this comment stale?

`/** Clamp the title to N lines with ellipsis. Default: `1`. Pass `undefined` to wrap freely. */`

if we default to 1, we would need to explicitly pass `titleMaxLines={undefined}`.

web/src/app/admin/billing/BillingDetailsView.tsx
raunakab: Hmm, could you expand on what you mean here?

nmgarza5: you removed a link component. is the redirect still working?

the markdown utility resolves the link so a clickable link appears but does it still redirect me to the correct page. when i look at what it does, i dont see any page routing logic baked into it so it feels like a regression -> click the link and nothing happens

web/lib/opal/src/components/cards/message-card/components.tsx
```diff
@@ -35,8 +35,8 @@ interface MessageCardBaseProps {
   /** Optional description below the title. */
   description?: string | RichStr;
 
-  /** Let the title wrap to multiple lines instead of truncating to one. */
-  titleWrap?: boolean;
+  /** Clamp the title to N lines with ellipsis. Default: `1`. Pass `undefined` to wrap freely. */
```
nmgarza5, line 38: this comment is stale. please fix

web/src/app/admin/billing/BillingDetailsView.tsx
raunakab: The string contains `"... [Remove users](/admin/users) ..."`, which should perform the correct routing. Namely, the `InlineMarkdown` engine parses that syntax into links, and the end renderer renders that as either a `Link` or `a` tag.

The actual parsing is done internally by the library, and the hooks into how to render each item is configured by us inside of `opal/src/comonents/text/InlineMarkdown.tsx`.

web/src/providers/ToastProvider.tsx
raunakab: Oh shoot. Yes, it is stale. My apologies!

web/src/app/admin/billing/BillingDetailsView.tsx
nmgarza5: perfect, that was the gap in my udnerstanding. thanks for clarifying!

nmgarza5 (approved):

web/lib/opal/src/components/cards/message-card/components.tsx
raunakab: Updated!

web/src/app/admin/billing/BillingDetailsView.tsx
raunakab: Yessir 🚀🚀🚀

web/src/providers/ToastProvider.tsx
raunakab: Updated the docstrings 😤💪

PR #12284, opened 2026-06-22
fix(voice): backend markdown stripping for TTS + optional REST STT toggle

justin-tahara:
## Description

Two fixes for Azure-backed voice mode (both surfaced on a self-hosted deployment whose backend runs on aarch64).

**1. TTS narrated markdown aloud (`**`, `##`, backticks, links).**
Markdown stripping previously lived only in the web client, so any backend TTS entry point — and any web bundle older than that client change — synthesized raw markdown. This adds a shared `strip_markdown_for_tts()` and applies it at every backend synthesis entry point:
- `POST /voice/synthesize` (`user_api.py`) — the non-streaming REST endpoint
- both websocket handlers in `websocket_api.py` (`handle_streaming_synthesis`, `handle_chunked_synthesis`)

The helper is line-anchored for headers so inline text like `C#` or `#1` is left intact.

**2. `VOICE_DISABLE_STREAMING_STT` env flag to force STT onto the chunked/REST path.**
In some runtimes a provider's native streaming SDK transport cannot initialize (the recognizer connects, the session starts and stops, but no audio is ever recognized — empty transcripts with **no error**). The provider's REST transcription path works in the same environment. This flag (sibling of the existing `VOICE_DISABLE_STREAMING_FALLBACK`) lets an operator route STT through the working REST path per-deployment, without affecting environments where native streaming works.

## How Has This Been Tested?

- New unit tests for `strip_markdown_for_tts` (bold/italic/code/headers/links, inline-hash preservation, whitespace collapsing): `backend/tests/unit/onyx/server/manage/voice/test_text_utils.py` — all passing.
- Verified live on a staging deployment: with the flag set, STT returns correct transcripts via REST; TTS no longer speaks markdown markers. Provider REST round-trip (TTS→STT) confirmed end-to-end for the affected Azure resource.
- `ruff`, `ruff format`, and `ty` pass on the changed files.

## Additional Options

## Notes / follow-ups

- The chunked/REST STT path transcribes the full utterance on stop; it does not emit native VAD auto-stop, and `MIN_CHUNK_BYTES` is tuned for compressed (webm/opus) audio rather than raw PCM. A follow-up can make the interim threshold format-aware to reduce redundant REST calls.
- `AzureVoiceProvider.transcribe()` hardcodes `language=en-US`; making recognition language configurable is a sensible follow-up for non-English users.

---
## Summary by cubic
Strips markdown for TTS on the server using `markdown-it-py` across all synthesis paths so we don’t speak markup, and adds a toggle to force STT over REST when native streaming is unreliable. Also simplifies STT branching so disabling streaming cleanly falls back without errors.

- **Bug Fixes**
  - TTS: parse CommonMark with `markdown-it-py` and emit readable text only; drop fenced code, collapse whitespace, preserve inline hashes (e.g., C#, #1), strip only paired emphasis so `*` and `snake_case` survive, and turn `[label](url)` into `label` (handles balanced-paren URLs). Applied to `POST /voice/synthesize`, websocket streaming, and chunked synthesis.
  - STT: respect `VOICE_DISABLE_STREAMING_STT` even when `VOICE_DISABLE_STREAMING_FALLBACK` is true; flattened streaming/chunked logic routes to chunked/REST without the “provider doesn't support streaming STT” error.

- **New Features**
  - `VOICE_DISABLE_STREAMING_STT` env flag to force STT through the chunked/REST path instead of native streaming.

<sup>Written for commit 91508f60a84e6a84c08901981dc2a750941a7995. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12284?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/server/manage/voice/text_utils.py
+++ b/backend/onyx/server/manage/voice/text_utils.py
@@ -1,0 +1,32 @@
+import re
+
+# [label](url) -> label
+_MD_LINK = re.compile(r"\[([^\]]+)\]\([^)]+\)")
+# Leading header hashes at the start of a line. Line-anchored so inline text such
+# as "C#" or "#1" is left untouched.
+_MD_HEADER = re.compile(r"(?m)^[ \t]*#{1,6}[ \t]*")
+# Inline code / code-fence backticks.
+_MD_CODE = re.compile(r"`{1,3}")
+# Runs of whitespace (including newlines).
+_WS = re.compile(r"\s+")
+
+
+def strip_markdown_for_tts(text: str) -> str:
+    """Remove common markdown markers so TTS does not read them aloud.
+
+    Mirrors the frontend ``stripMarkdownForTTS`` and is applied server-side as a
+    version-independent safety net: the web bundle may lag the backend, and not
+    every TTS entry point strips on the client.
+
+    Removes bold/italic markers, inline-code and code-fence backticks, and leading
+    header hashes (line-anchored to avoid mangling text like ``C#`` or ``#1``),
+    and converts ``[label](url)`` to ``label``. Whitespace runs are collapsed so
+    the synthesized speech does not contain odd pauses.
+    """
+    text = text.replace("**", "").replace("__", "")
+    text = text.replace("*", "").replace("_", "")
+    text = _MD_CODE.sub("", text)
+    text = _MD_HEADER.sub("", text)
+    text = _MD_LINK.sub(r"\1", text)
+    text = _WS.sub(" ", text)
+    return text.strip()
--- a/backend/onyx/server/manage/voice/text_utils.py
+++ b/backend/onyx/server/manage/voice/text_utils.py
@@ -1,43 +1,31 @@
-import re
+from markdown_it import MarkdownIt
+from markdown_it.token import Token
 
-# [label](url) -> label. The url may contain one level of balanced parens
-# (e.g. Wikipedia "..._(programming_language)" links).
-_MD_LINK = re.compile(r"\[([^\]]+)\]\((?:[^()]|\([^()]*\))*\)")
-# Paired emphasis markers only, so literal "*" and intraword underscores
-# (snake_case) survive. Delimiters must hug non-whitespace (CommonMark flanking),
-# and underscore variants additionally require non-word boundaries.
-_MD_BOLD_STAR = re.compile(r"\*\*(?=\S)(.+?)(?<=\S)\*\*")
-_MD_ITALIC_STAR = re.compile(r"\*(?=\S)(.+?)(?<=\S)\*")
-_MD_BOLD_UNDERSCORE = re.compile(r"(?<!\w)__(?=\S)(.+?)(?<=\S)__(?!\w)")
-_MD_ITALIC_UNDERSCORE = re.compile(r"(?<!\w)_(?=\S)(.+?)(?<=\S)_(?!\w)")
-# Inline-code / code-fence backticks.
-_MD_CODE = re.compile(r"`{1,3}")
-# Leading header hashes at the start of a line. Line-anchored so inline text
-# such as "C#" or "#1" is left untouched.
-_MD_HEADER = re.compile(r"(?m)^[ \t]*#{1,6}[ \t]*")
-# Runs of whitespace (including newlines).
-_WS = re.compile(r"\s+")
+_md = MarkdownIt("commonmark")
+
+
+def _render(tokens: list[Token]) -> str:
+    out: list[str] = []
+    for tok in tokens:
+        if tok.type == "inline":
+            out.append(_render(tok.children or []))
+        elif tok.type in ("text", "code_inline"):
+            out.append(tok.content)
+        elif tok.type in ("softbreak", "hardbreak"):
+            out.append(" ")
+        elif tok.type.endswith("_close") and tok.block:
+            # Block boundary (paragraph, heading, list item, ...) -> a space.
+            out.append(" ")
+    return "".join(out)
 
 
 def strip_markdown_for_tts(text: str) -> str:
-    """Remove common markdown markers so TTS does not read them aloud.
+    """Reduce markdown to plain text so TTS does not read markup aloud.
 
-    Mirrors the frontend ``stripMarkdownForTTS`` and is applied server-side as a
-    version-independent safety net: the web bundle may lag the backend, and not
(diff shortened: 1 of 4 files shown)


backend/onyx/server/manage/voice/text_utils.py
```diff
@@ -0,0 +1,32 @@
...
+    version-independent safety net: the web bundle may lag the backend, and not
+    every TTS entry point strips on the client.
+
+    Removes bold/italic markers, inline-code and code-fence backticks, and leading
+    header hashes (line-anchored to avoid mangling text like ``C#`` or ``#1``),
+    and converts ``[label](url)`` to ``label``. Whitespace runs are collapsed so
+    the synthesized speech does not contain odd pauses.
+    """
+    text = text.replace("**", "").replace("__", "")
+    text = text.replace("*", "").replace("_", "")
```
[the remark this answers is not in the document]
justin-tahara: Addressed in 54e9490a38. Emphasis stripping now only removes *paired* delimiters using CommonMark flanking rules, so literal `*` and intraword underscores (e.g. `snake_case`, `API_KEY`) are preserved. Added unit tests covering these cases.

backend/onyx/server/manage/voice/text_utils.py
```diff
@@ -0,0 +1,32 @@
+import re
+
+# [label](url) -> label
+_MD_LINK = re.compile(r"\[([^\]]+)\]\([^)]+\)")
```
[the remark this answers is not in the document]
justin-tahara: Addressed in 54e9490a38. The link regex now matches one level of balanced parentheses in the URL, so links like `[C](https://en.wikipedia.org/wiki/C_(programming_language))` no longer leak a trailing `)`. Added a unit test for this.

justin-tahara:
Addressed the STT flag-interaction edge case in 57c443b1cc.

Previously, when both `VOICE_DISABLE_STREAMING_STT=true` and `VOICE_DISABLE_STREAMING_FALLBACK=true`, the non-streaming branch returned a misleading "provider doesn't support streaming STT" error and aborted — even though the operator deliberately chose the chunked/REST path. The no-fallback guard now only fires when the provider *genuinely* lacks streaming support (`VOICE_DISABLE_STREAMING_FALLBACK and not VOICE_DISABLE_STREAMING_STT`), so an explicit `VOICE_DISABLE_STREAMING_STT` always routes to chunked regardless of the fallback flag.

Resolved truth table:
| DISABLE_STT | DISABLE_FALLBACK | result |
|---|---|---|
| off | off | chunked |
| off | on | error (provider truly can't stream + no fallback) |
| on | off | chunked |
| on | on | chunked (was: error — now fixed) |

backend/onyx/server/manage/voice/text_utils.py
```diff
@@ -0,0 +1,43 @@
+import re
+
+# [label](url) -> label. The url may contain one level of balanced parens
+# (e.g. Wikipedia "..._(programming_language)" links).
+_MD_LINK = re.compile(r"\[([^\]]+)\]\((?:[^()]|\([^()]*\))*\)")
```
Danelegend: We could use markdown_it instead so we don't have to maintain these? `markdown-it-py` is already an installed package.

```
from markdown_it import MarkdownIt

_md = MarkdownIt("commonmark")

def strip_markdown_for_tts(text: str) -> str:
    tokens = _md.parse(text)

    def render(tokens):
        out = []
        for tok in tokens:
            if tok.type == "inline":
                out.append(render(tok.children))
            elif tok.type in ("text", "code_inline"):
                out.append(tok.content)
            elif tok.type == "softbreak" or tok.type == "hardbreak":
                out.append(" ")
            elif tok.type.endswith("_close") and tok.block:
                out.append(" ")  # paragraph/heading boundaries -> space
        return "".join(out)

    return " ".join(render(tokens).split())
```

backend/onyx/server/manage/voice/websocket_api.py
```diff
@@ -410,8 +416,8 @@ async def websocket_transcribe(
                 await websocket.send_json({"type": "error", "message": str(e)})
                 return
 
-        # Use native streaming if provider supports it
-        if provider.supports_streaming_stt():
+        # Use native streaming if provider supports it (unless forced off)
+        if provider.supports_streaming_stt() and not VOICE_DISABLE_STREAMING_STT:
```
Danelegend: I think we can improve the layout of these if statements

backend/onyx/server/manage/voice/websocket_api.py
```diff
@@ -434,18 +440,20 @@ async def websocket_transcribe(
                 chunked_transcriber = ChunkedTranscriber(provider, audio_format="pcm16")
                 await handle_chunked_transcription(websocket, chunked_transcriber)
         else:
-            # Fall back to chunked transcription
-            if VOICE_DISABLE_STREAMING_FALLBACK:
+            # Here either the provider lacks streaming support, or streaming was
```
Danelegend: Can the comments be more concise also

Danelegend (approved):

backend/onyx/server/manage/voice/text_utils.py
justin-tahara: Good call — switched to `markdown-it-py` in 91508f60a8 (parse to tokens, emit text only). It also handles the cases the bots flagged for free (balanced-paren link URLs, nested emphasis) and drops fenced code blocks so they aren't read aloud. All existing tests still pass; added tests for code-fence/list handling.

backend/onyx/server/manage/voice/websocket_api.py
justin-tahara: Flattened in 91508f60a8 — a single `use_streaming` guard, streaming returns on success, and one chunked/REST path at the bottom (removed the duplicated chunked block).

backend/onyx/server/manage/voice/websocket_api.py
justin-tahara: Trimmed the comments in 91508f60a8.

PR #12285, opened 2026-06-22
fix: notion orphaned dbs

evan-onyx:
## Description

We don't know a priori where dbs are parented in notion, so rather than try to carefully order the hierarchy node returns and ending up with no parents, we just reconcile parents at the end of each index attempt. We also give each db that ends up without a parent (due to fetching restrictions, for example) a single fake "unfiled" parent to make it easier to navigate the hierarchy if this is prevalent

## How Has This Been Tested?

added unit tests

## Additional Options

---
## Summary by cubic
Fixes orphaned Notion databases by reconciling parents after a full sync and moving unresolved ones under a per‑workspace "Unfiled databases" page. This fixes ordering issues when parent pages arrive later and avoids scattering under SOURCE.

- **Bug Fixes**
  - Re-emit database nodes at the end of `load_from_state`; attach to the real parent page when available, else to a synthetic "Unfiled databases" page (stable id `${workspace_id}__unfiled`).
  - Track only non‑workspace‑parented databases during the prepass and resolve against `seen_hierarchy_node_raw_ids`.
  - Skip reconciliation in `poll_source` to avoid false orphans; prepass still uses the real parent page id from prior full syncs.
  - Standardize Notion URLs via `_notion_link`; update workspace and database links.
  - Add unit tests for resolved and orphan cases, workspace‑scoped unfiled ids, and poll behavior.

<sup>Written for commit 2aa6d9fc94d12ad09aa57d11792ed8225765fe2d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12285?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/connectors/notion/connector.py
+++ b/backend/onyx/connectors/notion/connector.py
@@ -41,6 +41,21 @@ _NOTION_PAGE_SIZE = 100
 _NOTION_CALL_TIMEOUT = 30  # 30 seconds
 _MAX_PAGES = 1000
 
+# Per-workspace bucket for databases whose real parent page never became a node.
+# Raw id derives from the workspace id so re-syncs update one row per workspace.
+_UNFILED_NODE_SUFFIX = "__unfiled"
+_UNFILED_DATABASES_DISPLAY_NAME = "Unfiled databases"
+
+
+def _notion_link(raw_id: str) -> str:
+    """Notion web URL for an object; Notion URLs omit the UUID's dashes."""
+    return f"https://notion.so/{raw_id.replace('-', '')}"
+
+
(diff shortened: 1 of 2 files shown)


Bo-Onyx (commented):
discussed offline, maybe first we figure out why search pages (those parents) are not fetched.

PR #12290, opened 2026-06-22
feat(db): support TLS for Postgres with server certificate verification

justin-tahara:
## Description

Adds first-class TLS for non-IAM PostgreSQL connections so deployments can both **encrypt** the connection and **verify the server certificate** against a provider CA bundle (the capability a security/compliance review cares about — `require` alone encrypts but does not stop a MITM).

Two new env vars (libpq vocabulary):
- `POSTGRES_SSLMODE` — `disable | allow | prefer | require | verify-ca | verify-full` (validated at config import).
- `POSTGRES_SSLROOTCERT` — path to the CA bundle used by `verify-ca` / `verify-full` (required by most managed providers — RDS/Cloud SQL/Azure — whose server certs don't chain to a system-trusted CA).

Implementation notes:
- New `onyx/db/engine/pg_ssl.py` is the single source of truth, applied through `connect_args` for **both** drivers: libpq `sslmode`/`sslrootcert` keys for psycopg2, and a built `ssl.SSLContext` for asyncpg (which does not accept libpq sslmode keys — verifying a custom CA requires a context).
- SSL is intentionally **not** placed in the connection URL. asyncpg's `connect_args["ssl"]` takes precedence over URL query params, so a URL-based `ssl=` would be silently overridden and the connection would drop to plaintext.
- IAM auth keeps precedence (it already enforces its own TLS); the explicit settings only apply when IAM is off.
- Wired into all engine-creation sites: primary sync, readonly sync, async, and both Alembic envs.
- Mirrors the existing Redis TLS pattern in `redis_pool.py`.
- docker-compose passes the vars automatically via `env_file`; documented in `env.template`. On Helm, mount the CA via the existing `customCACerts` block and set the vars through per-service `extraEnv` (no chart change required).

This supersedes community PR #9442, which put SSL in the URL (hits the silent-plaintext issue above) and had no way to supply a CA bundle. Client-certificate mTLS (`SSLCERT`/`SSLKEY`) is intentionally out of scope here; the design leaves it as a small follow-up.

## How Has This Been Tested?

- New unit tests in `backend/tests/unit/onyx/db/engine/test_postgres_ssl.py` (10 cases, no live DB) covering: psycopg2 connect_args per mode, asyncpg `SSLContext` resolution (`require` → `CERT_NONE`; `verify-ca` → `CERT_REQUIRED` + hostname off; `verify-full` → `CERT_REQUIRED` + hostname on; CA bundle actually loaded), `allow`/`prefer` passthrough, IAM precedence, config validation, and a regression guard that explicitly-configured SSL never resolves to `None` (the silent-plaintext footgun).
- Full db unit suite green: `pytest backend/tests/unit/onyx/db` → **176 passed**.
- `ty`, `ruff`, `ruff format`, and the full `pre-commit` set pass on all changed files.

## Additional Options

---
## Summary by cubic
Adds TLS for non-IAM PostgreSQL connections with server certificate verification across sync/async engines and Alembic. Enforces safe config and prevents silent plaintext fallbacks.

- **New Features**
  - Adds `POSTGRES_SSLMODE` (`disable|allow|prefer|require|verify-ca|verify-full`) and `POSTGRES_SSLROOTCERT` (CA bundle path).
  - `verify-ca`/`verify-full` require `POSTGRES_SSLROOTCERT` and verify the file exists; `require` encrypts without CA or hostname checks.
  - Fails fast if `POSTGRES_SSLROOTCERT` is set without `POSTGRES_SSLMODE`.
  - Central `pg_ssl.py` applies TLS: `psycopg2` via `sslmode`/`sslrootcert`; `asyncpg` via an `ssl.SSLContext`; never via URL.
  - IAM auth remains authoritative; SSL validation is skipped when `USE_IAM_AUTH=true`, with warnings if IAM overrides explicit SSL or if `POSTGRES_SSLROOTCERT` is set with `require`.
  - Applied to primary/readonly sync engines, async engine, and both Alembic envs; Docker Compose env template updated.

- **Migration**
  - For managed providers, set `POSTGRES_SSLMODE=verify-full` (or `verify-ca`) and `POSTGRES_SSLROOTCERT` to the provider CA bundle path.
  - Helm: mount CA with `customCACerts` and set env via `extraEnv`; Docker Compose reads from `env_file`. No action when `USE_IAM_AUTH=true`.

<sup>Written for commit 5047320bf7e8dfc9d065f8159f0114b1968657d0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12290?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/alembic/env.py
+++ b/backend/alembic/env.py
@@ -1,5 +1,6 @@
 from typing import Any
 from onyx.db.engine.iam_auth import get_iam_auth_token
+from onyx.db.engine.pg_ssl import create_pg_ssl_context
 from onyx.configs.app_configs import USE_IAM_AUTH
 from onyx.configs.app_configs import POSTGRES_HOST
 from onyx.configs.app_configs import POSTGRES_PORT
@@ -270,6 +271,7 @@ async def run_async_migrations() -> None:
     engine = create_async_engine(
         build_connection_string(),
         poolclass=pool.NullPool,
+        connect_args={"ssl": create_pg_ssl_context()},
     )
 
     if USE_IAM_AUTH:
--- a/backend/alembic_tenants/env.py
+++ b/backend/alembic_tenants/env.py
@@ -6,6 +6,7 @@ from sqlalchemy.engine import Connection
 from sqlalchemy.ext.asyncio import create_async_engine
 
 from alembic import context
+from onyx.db.engine.pg_ssl import create_pg_ssl_context
 from onyx.db.engine.sql_engine import build_connection_string
 from onyx.db.models import PublicBase
 
@@ -77,6 +78,7 @@ async def run_async_migrations() -> None:
     connectable = create_async_engine(
         build_connection_string(),
         poolclass=pool.NullPool,
+        connect_args={"ssl": create_pg_ssl_context()},
     )
 
     async with connectable.connect() as connection:
--- a/backend/onyx/configs/app_configs.py
+++ b/backend/onyx/configs/app_configs.py
@@ -614,6 +614,15 @@ elif POSTGRES_SSLMODE:
             "verify the server certificate; use verify-ca or verify-full to "
             "verify against the CA bundle."
         )
+elif POSTGRES_SSLROOTCERT:
+    # A CA bundle without a mode is dead config: SSL stays off and the
+    # connection runs unverified despite the operator supplying a cert. Fail
+    # loudly so this can't masquerade as a verified connection.
+    raise ValueError(
+        "POSTGRES_SSLROOTCERT is set but POSTGRES_SSLMODE is not. The CA bundle "
+        "is only used by a verifying mode; set POSTGRES_SSLMODE=verify-ca or "
+        "verify-full (or unset POSTGRES_SSLROOTCERT)."
+    )
 
 # Redis IAM authentication - enables IAM-based authentication for Redis ElastiCache
 # Note: This is separate from RDS IAM auth as they use different authentication mechanisms
--- a/backend/onyx/configs/app_configs.py
+++ b/backend/onyx/configs/app_configs.py
@@ -559,29 +559,61 @@ POSTGRES_TCP_KEEPALIVES_COUNT = int(
     os.environ.get("POSTGRES_TCP_KEEPALIVES_COUNT") or 5
 )
 
+# RDS IAM authentication - enables IAM-based authentication for PostgreSQL
+USE_IAM_AUTH = os.getenv("USE_IAM_AUTH", "False").lower() == "true"
+
 # TLS for non-IAM PostgreSQL connections.
 #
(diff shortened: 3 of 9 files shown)


justin-tahara:
Thanks for the review — addressed in d9583da15c. Summary:

**verify-ca/verify-full without a CA bundle (cubic P1):** Valid, and it also masked a cross-driver inconsistency — without `POSTGRES_SSLROOTCERT`, libpq errors out while asyncpg silently falls back to the system trust store. Rather than warn, I went stronger and **raise at config validation** when a verify mode lacks `POSTGRES_SSLROOTCERT`, so the CA-pinned trust model is explicit and both drivers behave identically. (Users with a publicly-trusted server cert point it at the system bundle.)

**SSLMODE validation runs under IAM (cubic P2):** Fixed — the whole validation block is now gated behind the non-IAM path, so an ignored value can't fail startup on IAM deployments.

**IAM silently ignores TLS env vars (cubic P2):** Fixed — emit a warning at config import when `USE_IAM_AUTH` overrides `POSTGRES_SSLMODE`/`POSTGRES_SSLROOTCERT`.

**`require` + `POSTGRES_SSLROOTCERT` loads then discards the CA (greptile P2):** Fixed — the `require` path no longer loads a CA bundle, and we warn that `require` doesn't verify (suggesting verify-ca/verify-full).

**`POSTGRES_SSLROOTCERT` path not validated at startup (greptile P2):** Fixed — added an `os.path.exists` check at config import, mirroring the IAM `SSL_CERT_FILE` check, so a bad path fails loudly instead of producing a cryptic `ssl.SSLError` at first connect.

Added unit tests for all of the above (verify-mode-without-CA raises, nonexistent CA raises, invalid mode ignored under IAM); full `backend/tests/unit/onyx/db` suite is green.

justin-tahara:
Addressed cubic's new P1 in 5047320bf7.

**`POSTGRES_SSLROOTCERT` set without `POSTGRES_SSLMODE` (cubic P1):** Valid — a CA bundle with no mode is dead config (SSL stays off), so it could masquerade as a verified connection. Rather than infer a mode implicitly, I **raise at startup** for this combination, consistent with the rest of this PR's fail-loud validation. Added a unit test (`test_rootcert_without_sslmode_raises`).

With this, the SSL config now has no silent path: any cert-bearing or verify-mode misconfiguration fails at startup, `require` warns that it doesn't verify, and IAM warns when it overrides the explicit settings. Full `backend/tests/unit/onyx/db` suite green (180 passed).

Bo-Onyx (approved):

PR #12292, opened 2026-06-22
feat(indexing): stream large spreadsheets via file-backed TabularSection

nmgarza5:
### The bug
A single oversized Google Drive `.xlsx` OOMs the indexing worker (~8 GB / 8192 MB limit). Extraction renders the **whole** sheet into one in-memory CSV string, and docfetching converts 8 files at once → 8 large sheets ≈ 8 GB. The old `MAX_XLSX_CELLS_PER_SHEET` cap truncated data **and** (verified live on cc7) didn't even bound peak — multi-sheet workbooks and wide rows slip past a cell-count cap.

### The fix — never hold a whole sheet in memory
Like reading a huge book one page at a time instead of photocopying the whole stack. Each stage is bounded:

1. **Extraction** (`extract_file_text.stage_or_inline_xlsx_sheets`) — stream the worksheet row by row to a temp CSV. A sheet whose CSV is ≤ `XLSX_STREAM_SHEET_BYTES` (50 MB) is kept inline; a larger one is staged in the file store and referenced by `csv_file_id` (no giant text on the heap or in the doc batch).
2. **Chunking** (`TabularChunker`) — one CSV row-parsing core (`parse_csv_stream`). Inline sheets parse from memory; file-backed sheets stream rows back from the file store (`read_file(use_tempfile=True)`), a row at a time.
3. **Descriptors** (`analyze_sheet`) — a single **bounded streaming pass** (running numeric/date aggregates, a cardinality-capped Counter, capped id-uniqueness) instead of transposing the sheet. So column-stat / total chunks are produced for **every** sheet, huge ones included, at flat memory.

**Result:** the *entire* sheet is indexed (no truncation), with full descriptor metadata, and worker memory stays flat regardless of sheet size.

### Routing
`xlsx_has_large_sheet` does a cheap xlsx zip-directory size check (no parse) to decide whether a workbook takes the streaming path; small workbooks keep the existing inline `xlsx_sheet_extraction` path untouched.

### What to review
- **`models.py` `TabularSection`** — `text` optional + new `csv_file_id`; validator enforces exactly one.
- **`analysis.py`** — `analyze_sheet` rewritten as a bounded single pass; `SheetAnalysis` now carries `NumericAggregate` (total/count/min/max) instead of a full value list. Inline output is unchanged (existing descriptor tests pass verbatim); huge sheets drop only very-high-cardinality categoricals (> cap), where a "most frequent value" isn't useful anyway.
- **`indexing_pipeline.py`** — non-image sections are copied with `model_copy()` so a file-backed `TabularSection`'s `csv_file_id` survives the pre-chunk rebuild (regression-tested; without it the streamed sheet reached the chunker empty).
- **`document_chunker.py`** — the empty-section filter keyed off `section.text`, which is `None` for a file-backed `TabularSection`, so an untitled doc's sheet (Drive files carry no title) was dropped before `TabularChunker` ran → indexed with **0 chunks**. File-backed sections now bypass the empty check. Caught end-to-end testing a 52.9 MB sheet on a live Drive connector; regression-tested.
- **Rolling deploy** — a new docfetching pod can emit `text=None` + `csv_file_id`; an old docprocessing pod would reject that on deserialize, so the attempt fails and **self-heals to full data on retry** (chosen over a partial-preview shim). Only large-xlsx attempts caught mid-rollout are affected.

### Tests
Unit coverage for: streamed extraction (large → file-backed + no truncation; small → inline), the chunker streaming a file-backed section (rows **and** descriptor/total chunks), `process_image_sections` preserving `csv_file_id`, and the DocumentChunker keeping a file-backed section in an untitled doc.

## Additional Options

### DIFF
--- a/backend/onyx/file_processing/extract_file_text.py
+++ b/backend/onyx/file_processing/extract_file_text.py
@@ -751,11 +751,13 @@ def stage_or_inline_xlsx_sheets(
     max_inline_bytes: int,
     file_name: str = "",
 ) -> list[StreamedSheet]:
-    """Render each non-empty worksheet to a temp CSV, streaming rows so peak RAM
-    is one row rather than the whole sheet. A sheet whose CSV is at most
-    `max_inline_bytes` is returned inline; a larger one is staged via `stage` and
-    referenced by id. The temp handle never escapes this function. No column
-    trimming or empty-run collapsing — those need a full in-memory pass."""
+    """Render each non-empty worksheet to a temp CSV, writing rows one at a time
+    so no worksheet is fully held in memory during conversion. A worksheet whose
+    CSV is at most `max_inline_bytes` is read back inline; a larger one is staged
+    via `stage` and referenced by `csv_file_id` so it stays off the heap
+    downstream. The returned list thus holds full CSV text for inline sheets
+    (each bounded by `max_inline_bytes`) and only an id for file-backed ones.
+    Empty rows are dropped; columns are not trimmed (that needs a second pass)."""
     sheets: list[StreamedSheet] = []
     workbook = _load_readonly_workbook(file, file_name)
     if workbook is None:
--- a/backend/onyx/file_processing/extract_file_text.py
+++ b/backend/onyx/file_processing/extract_file_text.py
@@ -745,6 +745,14 @@ class StreamedSheet(NamedTuple):
     csv_file_id: str | None
 
 
+def _row_has_content(row: tuple[Any, ...]) -> bool:
+    return any(v is not None and v != "" for v in row)
+
+
+def _cell(value: Any) -> str:
+    return "" if value is None else str(value)
+
+
 def stage_or_inline_xlsx_sheets(
     file: IO[bytes],
     stage: Callable[[IO[bytes], str], str],
@@ -768,27 +776,23 @@ def stage_or_inline_xlsx_sheets(
             ro_sheet.reset_dimensions()
             with tempfile.TemporaryFile(mode="w+", encoding="utf-8", newline="") as tmp:
                 writer = csv.writer(tmp, lineterminator="\n")
-                wrote = False
                 for row in ro_sheet.iter_rows(values_only=True):
-                    if not any(v is not None and v != "" for v in row):
-                        continue
-                    writer.writerow(["" if v is None else str(v) for v in row])
-                    wrote = True
-                if not wrote:
-                    continue
+                    if _row_has_content(row):
+                        writer.writerow([_cell(v) for v in row])
                 tmp.flush()
                 binary = cast(IO[bytes], tmp.buffer)
                 size = binary.seek(0, io.SEEK_END)
-                binary.seek(0)
-                if size <= max_inline_bytes:
-                    text = binary.read().decode("utf-8").strip()
-                    if not text:
-                        continue
-                    sheets.append(StreamedSheet(ro_sheet.title, text, None))
-                else:
+                if size > max_inline_bytes:
+                    binary.seek(0)
                     sheets.append(
                         StreamedSheet(ro_sheet.title, None, stage(binary, "text/csv"))
                     )
+                    continue
+                binary.seek(0)
+                text = binary.read().decode("utf-8").strip()
+                if not text:
+                    continue
+                sheets.append(StreamedSheet(ro_sheet.title, text, None))
     finally:
         workbook.close()
     return sheets
--- a/backend/onyx/file_processing/extract_file_text.py
+++ b/backend/onyx/file_processing/extract_file_text.py
@@ -4,6 +4,7 @@ import io
 import json
 import os
 import re
+import tempfile
 import zipfile
 from collections.abc import Callable
 from collections.abc import Iterator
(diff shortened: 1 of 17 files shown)


backend/onyx/file_processing/extract_file_text.py
```diff
@@ -703,6 +714,84 @@ def xlsx_sheet_extraction(file: IO[Any], file_name: str = "") -> list[tuple[str,
...
+    file: IO[bytes],
+    stage: Callable[[IO[bytes], str], str],
+    max_inline_bytes: int,
+    file_name: str = "",
+) -> list[StreamedSheet]:
+    """Render each non-empty worksheet to a temp CSV, streaming rows so peak RAM
+    is one row rather than the whole sheet. A sheet whose CSV is at most
+    `max_inline_bytes` is returned inline; a larger one is staged via `stage` and
+    referenced by id. The temp handle never escapes this function. No column
+    trimming or empty-run collapsing — those need a full in-memory pass."""
```
evan-onyx: I don't think this is true? there are pretty natural places in the current function for trimming and run collapsing I think?

backend/onyx/file_processing/extract_file_text.py
```diff
@@ -703,6 +714,84 @@ def xlsx_sheet_extraction(file: IO[Any], file_name: str = "") -> list[tuple[str,
...
+                    continue
+                tmp.flush()
+                binary = cast(IO[bytes], tmp.buffer)
+                size = binary.seek(0, io.SEEK_END)
+                binary.seek(0)
+                if size <= max_inline_bytes:
+                    text = binary.read().decode("utf-8").strip()
+                    if not text:
+                        continue
+                    sheets.append(StreamedSheet(ro_sheet.title, text, None))
```
evan-onyx: I don't understand how we can make the claim that this is one row in memory at once, it seems like we're  accumulating entire sheets in a list?

backend/onyx/indexing/chunking/tabular_section_chunker/tabular_section_chunker.py
```diff
@@ -235,6 +256,12 @@ def chunk_section(
     ) -> SectionChunkerOutput:
         payloads = accumulator.flush_to_list()
 
+        csv_file_id = getattr(section, "csv_file_id", None)
```
evan-onyx: why getattr?

backend/onyx/indexing/chunking/tabular_section_chunker/tabular_section_chunker.py
```diff
@@ -289,3 +316,44 @@ def chunk_section(
...
+        self,
+        section: Section,
+        csv_file_id: str,
+        payloads: list[ChunkPayload],
+        content_token_limit: int,
+    ) -> SectionChunkerOutput:
+        """Stream the staged CSV straight into row chunks so a huge sheet is never
+        materialized. Descriptor/analysis chunks are skipped: ``analyze_sheet``
+        transposes the whole sheet into memory, which would defeat the streaming
+        bound. Small sheets keep the inline path and its descriptors."""
```
evan-onyx: imo we need to keep this somehow to make the spreadsheets useful for analysis

backend/onyx/indexing/indexing_pipeline.py
```diff
@@ -720,6 +721,19 @@ def process_image_sections(documents: list[Document]) -> list[IndexingDocument]:
         # Only get the vision LLM if image processing is enabled
         llm = get_default_llm_with_vision()
 
+    def _clone_non_image_section(section: Section) -> Section:
+        """Copy a non-image section, preserving a TabularSection (and its
+        csv_file_id) rather than flattening it to a base Section."""
+        if isinstance(section, TabularSection):
+            return section.model_copy()
```
evan-onyx: why not always use model_copy()?

evan-onyx (commented):
here are my thoughts, will let Dane give the check mark since this mostly touches his stuff

nmgarza5:
@greptile

backend/onyx/file_processing/extract_file_text.py
```diff
@@ -703,6 +714,86 @@ def xlsx_sheet_extraction(file: IO[Any], file_name: str = "") -> list[tuple[str,
...
+    via `stage` and referenced by `csv_file_id` so it stays off the heap
+    downstream. The returned list thus holds full CSV text for inline sheets
+    (each bounded by `max_inline_bytes`) and only an id for file-backed ones.
+    Empty rows are dropped; columns are not trimmed (that needs a second pass)."""
+    sheets: list[StreamedSheet] = []
+    workbook = _load_readonly_workbook(file, file_name)
+    if workbook is None:
+        return sheets
+    try:
+        for sheet in workbook.worksheets:
```
Danelegend, line 774: Lets try to remove the wrote flag. Here is a sample change

```
for sheet in workbook.worksheets:
      ro_sheet = cast(ReadOnlyWorksheet, sheet)
      ro_sheet.reset_dimensions()
      with tempfile.TemporaryFile(mode="w+", encoding="utf-8", newline="") as tmp:
          writer = csv.writer(tmp, lineterminator="\n")
          for row in ro_sheet.iter_rows(values_only=True):
              if _row_has_content(row):
                  writer.writerow([_cell(v) for v in row])

          tmp.flush()
          binary = cast(IO[bytes], tmp.buffer)
          size = binary.seek(0, io.SEEK_END)
          if size > max_inline_bytes:
              ...  # large-file path (your existing else)
              continue
          binary.seek(0)
          text = binary.read().decode("utf-8").strip()
          if not text:
              continue
          sheets.append(StreamedSheet(ro_sheet.title, text, None))
```

Danelegend (approved):

PR #12294, opened 2026-06-22
feat(db): mutual TLS (client certs) for Postgres + shared SSLContext builder

justin-tahara:
## Description

Follow-up to the Postgres server-cert-verification PR. Adds **client-certificate mutual TLS** for Postgres and extracts the SSLContext construction into a shared helper so Redis/OpenSearch can reuse it later.

**Mutual TLS** — new env vars `POSTGRES_SSLCERT` / `POSTGRES_SSLKEY` (+ optional `POSTGRES_SSLKEY_PASSWORD`):
- psycopg2: passed as libpq `sslcert`/`sslkey`/`sslpassword` `connect_args`.
- asyncpg: `SSLContext.load_cert_chain(...)`, which libpq-style keys can't express.
- Validated loudly at startup (consistent with the prior PR): cert and key must be set **together**, under a `POSTGRES_SSLMODE` that actually negotiates SSL (`require`/`verify-ca`/`verify-full`), and the files must exist. A cert/key with no mode is rejected as dead config.

**Shared builder** — `onyx/utils/ssl.build_ssl_context(verify_mode, check_hostname, ca_certs, certfile, keyfile, key_password)` centralizes context creation (including the check_hostname-before-CERT_NONE ordering gotcha). `pg_ssl.py` now uses it. Redis (`redis_pool.py`) and OpenSearch verification are intentionally **not** migrated here — that belongs with their own TLS work, not a Postgres PR — but the helper is ready for them.

Client-cert mTLS is mutually exclusive with IAM auth (IAM uses token auth and keeps precedence).

## How Has This Been Tested?

- `onyx/utils/test_ssl.py` — `build_ssl_context` loads a real (cryptography-generated) self-signed client cert; a mismatched key raises `ssl.SSLError`; verify-mode/hostname wiring.
- `test_postgres_ssl.py` — psycopg2 args include `sslcert`/`sslkey`/`sslpassword`; asyncpg builds a context with the client cert loaded; validation rejects cert-without-key, cert under a non-negotiating mode (`prefer`), and cert-without-mode.
- Full `backend/tests/unit/onyx/db` + `backend/tests/unit/onyx/utils` → **478 passed**; `ty` + `ruff` + `pre-commit` clean.

## Additional Options

---
## Summary by cubic
Adds mutual TLS (client certificates) for Postgres and introduces a shared TLS context builder in `onyx.utils.tls`. Startup validation is stricter; IAM auth still takes precedence. Docs for env vars were clarified.

- **New Features**
  - Client cert mTLS via `POSTGRES_SSLCERT` / `POSTGRES_SSLKEY` (+ optional `POSTGRES_SSLKEY_PASSWORD`).
  - `psycopg2`: passes libpq `sslcert`/`sslkey`/`sslpassword`. `asyncpg`: loads the chain with `SSLContext.load_cert_chain(...)`.
  - New helper `onyx.utils.tls.build_ssl_context(...)` centralizes CA bundle, verify mode, hostname checks, and client certs; `pg_ssl.py` uses it.
  - Startup validation: cert+key required together; allowed modes are `require`/`verify-ca`/`verify-full`; file paths must exist; reject any `POSTGRES_SSL*` (incl. lone `POSTGRES_SSLKEY_PASSWORD`) without a mode; warn and ignore all `POSTGRES_SSL*` when `USE_IAM_AUTH=true`.
  - `deployment/docker_compose/env.template` now explains `require` → `verify-ca` → `verify-full` and optional mTLS.

- **Migration**
  - No change unless enabling mTLS.
  - To enable: set `POSTGRES_SSLMODE` to `require`/`verify-ca`/`verify-full`, add `POSTGRES_SSLCERT` and `POSTGRES_SSLKEY` (and `POSTGRES_SSLROOTCERT` for verify modes).
  - When `USE_IAM_AUTH=true`, `POSTGRES_SSL*` is ignored.

<sup>Written for commit 0648161599b8c281f7541098b3e670c4bbd9cb80. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12294?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


justin-tahara:
Thanks @greptile — both addressed in 6caf610d1a.

**P1 — `POSTGRES_SSLKEY_PASSWORD` dead config:** Valid; it was the one combination slipping through (not a file path, so absent from `_POSTGRES_SSL_FILE_SETTINGS`). Now: a password without a client cert raises (it only decrypts `POSTGRES_SSLKEY`), and the no-mode dead-config branch keys off a new `_HAS_POSTGRES_SSL_SECONDARY` that includes the password. Added `test_key_password_without_cert_raises` and `test_key_password_without_mode_raises`.

**P2 — `require` comment:** Fixed — clarified that `create_default_context` does load the system CA bundle, it's just never consulted under `CERT_NONE`.

`backend/tests/unit/onyx/db` + `.../utils` → 480 passed; ty/ruff/pre-commit clean.

backend/onyx/utils/tls.py
Bo-Onyx, line 1: the file name shadows the stdlib ssl module. Safe today thanks to absolute imports, but it's a classic foot-gun and the name undersells what's  inside (a client-side SSLContext builder). Consider tls.py or ssl_context.py. build_ssl_context itself is a good name.

deployment/docker_compose/env.template
Bo-Onyx, line 1: The current block is accurate but expert-facing — it doesn't make clear that one-way SSL is the common case and mTLS is the uncommon add-on. Optional rewrite:
```
  # --- Database TLS (libpq sslmode vocabulary; ignored when USE_IAM_AUTH=true) ---
  # Pick the level you need; each builds on the one above. Default (unset) = no TLS.
  #
  #   require       encrypt only — does NOT detect a MITM (server is not authenticated)
  #   verify-ca     + verify the server cert chains to POSTGRES_SSLROOTCERT
  #   verify-full   + also verify the server hostname  ← recommended for managed DBs
  #
  # verify-ca / verify-full REQUIRE POSTGRES_SSLROOTCERT (CA bundle path; use the
  # system bundle for publicly-trusted server certs).
  # POSTGRES_SSLMODE=          # disable | allow | prefer | require | verify-ca | verify-full
  # POSTGRES_SSLROOTCERT=      # e.g. /etc/ssl/certs/rds-ca-bundle.pem
  #
  # Mutual TLS (optional, rarely needed): also present a CLIENT cert so the server
  # can authenticate us. Most deployments do NOT set these. Requires a POSTGRES_SSLMODE
  # that negotiates SSL (require / verify-ca / verify-full); cert and key must both be set.
  # POSTGRES_SSLCERT=          # client certificate path
  # POSTGRES_SSLKEY=           # client private key path
  # POSTGRES_SSLKEY_PASSWORD=  # optional, only if the key file is encrypted
```

Bo-Onyx (approved):

justin-tahara:
Thanks for the review @Bo-Onyx — both applied in 0648161599.

- **`utils/ssl.py` shadowing stdlib `ssl`:** renamed to `onyx/utils/tls.py` (and `test_ssl.py` → `test_tls.py`). Good catch on the foot-gun; `build_ssl_context` kept as-is.
- **env.template:** adopted your layered rewrite — makes clear one-way SSL (require → verify-ca → verify-full) is the common path and mTLS is the rare opt-in. Used `<-` instead of the unicode arrow to keep the file ASCII.

480 db+utils unit tests green; ty/ruff/pre-commit clean.

PR #12308, opened 2026-06-23
feat(web): add full-width chat mode toggle

nmgarza5:
## Description

Adds a per-browser toggle that expands the chat message area to the full window width. The fixed 720px reading-width cap is cramped for wide content; this lets messages flow to the window.

- Toggle: a maximize icon button above the input, next to the model selector.
- On: assistant/human/error messages drop the max-width and fill the window. The input bar stays at reading width.
- Off (default): unchanged 720px column. No table-rendering changes.
- Persisted per-browser via localStorage; frontend-only, no backend.

## Additional Options


web/src/providers/FullWidthChatProvider.tsx
nmgarza5, line 1: im not confident this needs to be a provider. open to discussion here

raunakab (approved):

PR #12321, opened 2026-06-23
feat(redis): mutual TLS (client certificate) for Redis connections

justin-tahara:
## Description

Onyx supports one-way Redis TLS (`REDIS_SSL`, `REDIS_SSL_CA_CERTS`, `REDIS_SSL_CERT_REQS`) but has no way to present a **client certificate**, so deployments whose managed Redis requires **mutual TLS** can't connect.

Adds `REDIS_SSL_CERTFILE` / `REDIS_SSL_KEYFILE`, threaded into **every** Redis connection path:
- **Sync pool** (`redis_pool.py` `RedisPool.create_pool`) — `ssl_certfile`/`ssl_keyfile` kwargs on `BlockingConnectionPool`.
- **Async client** (`_build_async_redis_connection`) — `SSLContext.load_cert_chain(...)`.
- **Celery broker + result backend** (`celery/configs/base.py`) — appended to the `rediss://` URL query so the **workers** get mTLS too, not just the API/app (easy to miss otherwise).

Validated loudly at startup (same style as the Postgres/OpenSearch TLS work): cert and key must be set together, require `REDIS_SSL=true`, and the files must exist. IAM auth keeps precedence.

Certs are referenced by **file path** (consistent with the existing `REDIS_SSL_CA_CERTS`); a managed Redis that hands out base64-encoded certs decodes them into a mounted secret and points these vars at the paths — matching the `redis.Redis(ssl_certfile=..., ssl_keyfile=...)` shape.

## How Has This Been Tested?

- `test_redis_mtls.py`: the sync pool forwards `ssl_certfile`/`ssl_keyfile`; the async client calls `load_cert_chain`; the Celery broker **and** result-backend URLs include the cert params; config validation rejects cert-without-key, client-cert-without-`REDIS_SSL`, and a nonexistent cert path.
- `backend/tests/unit/onyx/redis` + `db` + `utils` → 494 passed; `ty` + `ruff` + `pre-commit` clean.

## Additional Options

---
## Summary by cubic
Adds mutual TLS support for Redis so the app and Celery workers can present a client certificate. This enables connecting to managed Redis that require mTLS.

- **New Features**
  - Adds `REDIS_SSL_CERTFILE` and `REDIS_SSL_KEYFILE` (must be set together, require `REDIS_SSL=true`; paths are validated at startup).
  - Threads client cert/key through all Redis paths: sync via `redis.BlockingConnectionPool`, async via `ssl.SSLContext.load_cert_chain`, and `Celery` broker/result URLs via `rediss://` query with percent-encoded paths so special characters don’t break the URL.
  - IAM auth continues to take precedence when enabled.

- **Migration**
  - Provide file paths for the client cert and key via `REDIS_SSL_CERTFILE` and `REDIS_SSL_KEYFILE` and set `REDIS_SSL=true`.
  - If your provider gives base64 certs, decode to files (e.g., mounted secrets) and point the env vars to those paths.
  - Startup will fail if only one is set or the files do not exist.

<sup>Written for commit eafc24c1a6fc98f6f55118e7eac8a70fcdb3e395. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12321?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/background/celery/configs/base.py
+++ b/backend/onyx/background/celery/configs/base.py
@@ -12,6 +12,8 @@ from onyx.configs.app_configs import REDIS_PORT
 from onyx.configs.app_configs import REDIS_SSL
 from onyx.configs.app_configs import REDIS_SSL_CA_CERTS
 from onyx.configs.app_configs import REDIS_SSL_CERT_REQS
+from onyx.configs.app_configs import REDIS_SSL_CERTFILE
+from onyx.configs.app_configs import REDIS_SSL_KEYFILE
 from onyx.configs.app_configs import USE_REDIS_IAM_AUTH
 from onyx.configs.constants import OnyxCeleryPriority
 from onyx.configs.constants import REDIS_SOCKET_KEEPALIVE_OPTIONS
@@ -31,6 +33,12 @@ if REDIS_SSL and not USE_REDIS_IAM_AUTH:
     SSL_QUERY_PARAMS = f"?ssl_cert_reqs={REDIS_SSL_CERT_REQS}"
     if REDIS_SSL_CA_CERTS:
         SSL_QUERY_PARAMS += f"&ssl_ca_certs={REDIS_SSL_CA_CERTS}"
+    # Client certificate for mutual TLS — the broker URL is how the Celery
+    # workers get their Redis SSL config, so they need this too.
+    if REDIS_SSL_CERTFILE:
+        SSL_QUERY_PARAMS += (
+            f"&ssl_certfile={REDIS_SSL_CERTFILE}&ssl_keyfile={REDIS_SSL_KEYFILE}"
+        )
 
 # region Broker settings
 # example celery_broker_url: "redis://:password@localhost:6379/15"
--- a/backend/onyx/background/celery/configs/base.py
+++ b/backend/onyx/background/celery/configs/base.py
@@ -34,10 +34,13 @@ if REDIS_SSL and not USE_REDIS_IAM_AUTH:
     if REDIS_SSL_CA_CERTS:
         SSL_QUERY_PARAMS += f"&ssl_ca_certs={REDIS_SSL_CA_CERTS}"
     # Client certificate for mutual TLS — the broker URL is how the Celery
-    # workers get their Redis SSL config, so they need this too.
-    if REDIS_SSL_CERTFILE:
+    # workers get their Redis SSL config, so they need this too. Percent-encode
+    # the paths (like REDIS_PASSWORD above) so URL-special characters in a path
+    # can't corrupt the query string.
+    if REDIS_SSL_CERTFILE and REDIS_SSL_KEYFILE:
         SSL_QUERY_PARAMS += (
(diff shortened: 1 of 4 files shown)


justin-tahara:
Thanks @cubic-dev-ai @greptile-apps — all addressed in eafc24c1a6.

- **URL-encode cert/key paths in the broker URL (P1/P2):** Valid — `urllib.parse.quote(..., safe="")` on `REDIS_SSL_CERTFILE`/`REDIS_SSL_KEYFILE`, matching how `REDIS_PASSWORD` is already handled, so a path with `&`/`=`/`#`/space can't corrupt the query string. (Scoped to the new lines; left the pre-existing unchanged `ssl_ca_certs` line alone.)
- **Test leaks `celery_base` state (P2):** Valid — the autouse teardown now reloads `celery_base` (not just `app_configs`) so the reloaded `broker_url`/`result_backend` reset to defaults; updated the URL assertions to expect the encoded paths.
- **Fixture return type (P2):** changed `-> object` to `Generator[None, None, None]`.

`backend/tests/unit/onyx/redis` + `db` → 193 passed; ty/ruff/pre-commit clean.

nmgarza5 (approved):

justin-tahara:
⚠️ Pushed a follow-up fix (a21684e) after **live end-to-end testing** against a real TLS + mTLS Redis — worth a re-look since it changes behavior.

**Finding:** the **async** path was silently not applying TLS material. `_build_async_redis_connection` set `connection_kwargs["ssl"] = <prebuilt ssl_context>`, but redis-py's *async* client (5.0.8) has no `ssl_context` parameter (only the *sync* `SSLConnection` does). So the CA bundle and the mTLS client cert (`load_cert_chain`) were dropped — TLS negotiated but unverified and with no client cert. This was masked because `REDIS_SSL_CERT_REQS` defaults to `none` (no verification), so nobody hit it until verification/mTLS was actually required.

**Fix:** pass redis-py's native async `ssl_*` kwargs (`ssl_ca_certs`, `ssl_cert_reqs`, `ssl_certfile`, `ssl_keyfile`, `ssl_check_hostname`) — the same mechanism the sync pool already uses. Removed the now-dead hand-built context + `SSL_CERT_REQS_MAP` + `import ssl`.

**Verified end-to-end** (live TLS+mTLS Redis, `tls-auth-clients yes`):
- sync pool **and** async client → PING OK with CA + client cert
- wrong CA → rejected; missing client cert → server rejects

Updated the async unit test to assert the native kwargs. The sync path was already correct. (Aside: the IAM async path passes a context the same way and may have the same latent issue, but that's a separate provider/code path — out of scope here.)

PR #12325, opened 2026-06-23
feat(audit): emit admin-config and access-control audit events

justin-tahara:
## Description

Phase 2 (final MVP piece) of the SIEM-compatible audit-event subsystem: wires `emit_audit_event` into the **admin-config** and **access-control** endpoints. Builds directly on the core subsystem (#12219) and auth events (#12271).

Events emitted (on the success path only, after the DB op commits):

- **LLM provider** — create / update (upsert toggles on `is_creation`) / delete
- **Connector** — create / update / delete
- **Connector-credential pair** — create (associate) / update (status) / delete (dissociate)
- **API key** — create / regenerate / delete
- **Credential** — create / update / delete (all delete variants, incl. admin + force)
- **User access-control** — role change / deactivate / reactivate / delete

Each event carries the acting admin (`actor`), the affected `resource_type` / `resource_id`, OCSF class, and the usual tenant / request / source-IP context. All actions already exist in the `AuditAction` taxonomy shipped in #12219 — this PR is pure call-site wiring.

Adds a small `actor_from_user()` helper to `onyx.utils.audit` so each call site stays a single emit. The helper is best-effort and never raises; it's typed `Any` to keep `audit.py` free of an ORM import dependency. Audit emission remains fully fail-safe — it never raises into the request path.

No secret values are ever logged (only ids, names, and coarse source/status metadata).

## How Has This Been Tested?

- Unit tests for `actor_from_user` (None, populated, exception-safety) in `test_audit.py`.
- New `test_admin_audit_events.py` exercises representative wired endpoints (API key create/regenerate/delete + LLM provider delete) via `caplog`, asserting the right `action` / `outcome` / `actor` / `resource_id`.
- `ty` + `ruff` clean; all pre-commit hooks pass.

## Additional Options

---
## Summary by cubic
Adds SIEM-ready audit events to admin-config and access-control endpoints to track changes across providers, connectors, keys, and users. Also stabilizes audit payloads and stops chat from nudging `open_url` when the tool is disabled.

- **New Features**
  - Emit post-commit audit events for LLM providers, connectors, cc-pairs, API keys, credentials, and user access control; cc-pair create/delete only on actual changes and delete records the cc-pair id.
  - Events include actor/resource ids and OCSF class; fail-safe and no secrets. Added `actor_from_user()` with a diagnostic warning on failure; updated docs.

- **Bug Fixes**
  - Gate `open_url` reminders on `OpenURLTool` availability and centralize via `select_reminder_text`.
  - Serialize enum fields in audit `extra` using `.value` for stable SIEM strings.

<sup>Written for commit d90a422453ac926b26d2c24cecc9dbebf22f9ebb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12325?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/server/api_key/api.py
+++ b/backend/onyx/server/api_key/api.py
@@ -13,6 +13,10 @@ from onyx.db.engine.sql_engine import get_session
 from onyx.db.enums import Permission
 from onyx.db.models import User
 from onyx.server.api_key.models import APIKeyArgs
+from onyx.utils.audit import actor_from_user
+from onyx.utils.audit import AuditAction
+from onyx.utils.audit import AuditOutcome
+from onyx.utils.audit import emit_audit_event
 
 router = APIRouter(prefix="/admin/api-key")
 
@@ -31,16 +35,32 @@ def create_api_key(
     user: User = Depends(require_permission(Permission.FULL_ADMIN_PANEL_ACCESS)),
     db_session: Session = Depends(get_session),
 ) -> ApiKeyDescriptor:
-    return insert_api_key(db_session, api_key_args, user.id)
+    api_key = insert_api_key(db_session, api_key_args, user.id)
+    emit_audit_event(
+        AuditAction.API_KEY_CREATE,
+        AuditOutcome.SUCCESS,
+        actor=actor_from_user(user),
+        resource_type="api_key",
+        resource_id=api_key.api_key_id,
+    )
+    return api_key
 
 
 @router.post("/{api_key_id}/regenerate")
 def regenerate_existing_api_key(
     api_key_id: int,
-    _: User = Depends(require_permission(Permission.FULL_ADMIN_PANEL_ACCESS)),
+    user: User = Depends(require_permission(Permission.FULL_ADMIN_PANEL_ACCESS)),
     db_session: Session = Depends(get_session),
 ) -> ApiKeyDescriptor:
-    return regenerate_api_key(db_session, api_key_id)
+    api_key = regenerate_api_key(db_session, api_key_id)
+    emit_audit_event(
+        AuditAction.API_KEY_REGENERATE,
+        AuditOutcome.SUCCESS,
+        actor=actor_from_user(user),
+        resource_type="api_key",
+        resource_id=api_key_id,
+    )
+    return api_key
 
 
 @router.patch("/{api_key_id}")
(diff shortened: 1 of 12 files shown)


justin-tahara:
Addressed the review feedback in 38583ccb19:

**cc_pair false-positive SUCCESS (greptile P1):** Both `CC_PAIR_CREATE` and `CC_PAIR_DELETE` now emit only when the operation actually changed state. `add_credential_to_connector` / `remove_credential_from_connector` return `success=False` *without raising* on a no-op (credential already linked / not linked), so the emits are now guarded on `response.success` / `result.success`. I guarded CREATE too since it has the same no-op path.

**cc_pair_delete missing resource_id (greptile P1 / cubic P2):** Fixed, but not via `result.data` as suggested — `remove_credential_from_connector` returns the **connector_id** in `.data` on both paths, not the cc_pair id (unlike CREATE, where `add_credential_to_connector` returns the cc_pair PK). So I look up the cc_pair id via `get_connector_credential_pair_for_user` *before* removal and record that.

**actor_from_user silent swallow (cubic P2):** Now logs a warning (with `exc_info`) before degrading to `None`. The warning goes to a standard `onyx.utils.audit` logger, deliberately **not** the `onyx.audit` JSON stream, so it never pollutes the parsed audit output.

Also fixed the two CI failures (mypy + llm external-dependency tests): renaming the unused `_` dependency param to `user` in `put_llm_provider` broke existing direct callers in `test_llm_provider_api_base.py` that passed `_=` by keyword — updated those to `user=` (the `run_llm_config_test`/`test_llm_configuration` calls keep `_=` since that param is unchanged).

justin-tahara:
Addressed in d90a422453: the cc_pair status and the connector/credential `source` fields in `extra` now serialize via `.value` instead of `str(enum)`. Since these are `(str, Enum)` members, `str()` returns the repr-style `ClassName.MEMBER` rather than the canonical value — `.value` emits the stable string SIEM filters expect (consistent with the user-role events, which already used `.value`).

(The earlier cc_pair false-positive / missing-resource_id / actor_from_user comments were on the original commit 6760082ae and were already resolved in db2306d52a — guarded both create/delete emits on `.success`, look up the real cc_pair id before dissociation, and log a warning before `actor_from_user` degrades to None.)

jmelahman (approved):

PR #12333, opened 2026-06-23
fix(search): bound document_filter LLM calls with a timeout (ENG-4238)

justin-tahara:
## Description

`classify_section_relevance()` and `select_sections_for_expansion()` in `backend/onyx/secondary_llm_flows/document_filter.py` called `llm.invoke()` with **no timeout**. A stalled provider connection (`litellm.Timeout: Connection timed out after None seconds`) hangs the worker until the api-server liveness probe SIGKILLs the pod (exit 137, not OOM), dropping all in-flight chat/search streams — observed in prod as "Claude Internal Search errors." Present in v4.0.7 and v4.2.0; an upgrade does not fix it.

Changes:
- New config `SECONDARY_LLM_FLOW_TIMEOUT_S` (default **60s**, env-overridable) in `configs/chat_configs.py`.
- Pass `timeout_override=SECONDARY_LLM_FLOW_TIMEOUT_S` to **both** `llm.invoke()` calls. Both already sit in `try/except` that degrades to a safe default, so a timeout now **fails fast into the existing fallback** (search still returns, just without LLM-optimized section selection) instead of hanging a worker.
- Added `@log_function_time(print_only=True)` to both flows so per-call latency lands in Loki/Prometheus (it was previously only in Braintrust, tagged by flow) — lets us tune the timeout on data.

Why 60s is safe regardless of exact percentiles: the failure mode is graceful degradation, and 60s is already more lenient than the current ~30s liveness-kill window (3×10s), so it can only reduce blast radius. These are `ReasoningEffort.OFF` classification calls (single-digit-second p50), so 60s only ever catches the pathological hang.

Linear: ENG-4238 (related: ENG-4236 streaming-consumer bound).

## How Has This Been Tested?

New unit test `backend/tests/unit/onyx/secondary_llm_flows/test_document_filter.py` (2 cases, passing): a timed-out `invoke` (`LLMTimeoutError`) degrades to the default classification, and `timeout_override` is passed on both the timeout and success paths. `ruff check` / `ruff format` clean; mypy clean on changed files.

## Additional Options

---
## Summary by cubic
Add a 60s timeout to document-filter secondary LLM calls so stalled providers fail fast and search degrades gracefully instead of dropping streams. Addresses Linear ENG-4238.

- **Bug Fixes**
  - Introduced `SECONDARY_LLM_FLOW_TIMEOUT_S` (default 60s; env-overridable) and passed `timeout_override` to `llm.invoke()` in `classify_section_relevance` and `select_sections_for_expansion`.
  - Timeouts now fall into existing try/except fallbacks, preventing liveness-kill-induced stream drops; added `@log_function_time(print_only=True)` on both flows for per-call latency.
  - Added unit tests covering timeout fallback and success-path timeout propagation for both flows, including the `select_sections_for_expansion` timeout path.

<sup>Written for commit 560c46664423c5d2e33723f8368c6e61b9f74a18. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12333?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/configs/chat_configs.py
+++ b/backend/onyx/configs/chat_configs.py
@@ -41,6 +41,13 @@ LLM_FIRST_CHUNK_MAX_RETRIES = max(
 # Socket-read timeout for deep-research report calls — bounds inter-chunk gaps
 # (including a zero-chunk stall), not total generation time.
 DR_REPORT_LLM_TIMEOUT_S = int(os.environ.get("DR_REPORT_LLM_TIMEOUT_S") or "60")
+# Timeout for non-streaming secondary LLM flows (e.g. search section-relevance
+# classification and section-expansion selection). These are short, low-effort
+# calls; the bound exists so a stalled provider connection fails fast into the
+# existing graceful fallback instead of hanging a worker until liveness kills it.
+SECONDARY_LLM_FLOW_TIMEOUT_S = int(
+    os.environ.get("SECONDARY_LLM_FLOW_TIMEOUT_S") or "60"
+)
 # Live buffer TTL. Refreshed per write.
 CHAT_STREAM_BUFFER_TTL_S = int(os.environ.get("CHAT_STREAM_BUFFER_TTL_S") or 3600)
 # Retention after the run is done.
--- a/backend/onyx/secondary_llm_flows/document_filter.py
+++ b/backend/onyx/secondary_llm_flows/document_filter.py
@@ -1,6 +1,7 @@
 import json
 import re
 
+from onyx.configs.chat_configs import SECONDARY_LLM_FLOW_TIMEOUT_S
 from onyx.context.search.models import ContextExpansionType
 from onyx.context.search.models import InferenceChunk
 from onyx.context.search.models import InferenceSection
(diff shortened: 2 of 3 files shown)


justin-tahara:
Addressed the Greptile/Cubic P2: added `test_select_sections_for_expansion_timeout_falls_back` and `test_select_sections_for_expansion_passes_timeout_on_success` in `test_document_filter.py` (commit 560c466644). Both now assert the timeout fallback (returns the input sections, capped) and that `timeout_override` is propagated — symmetric with the `classify_section_relevance` coverage. Full file: 4 tests passing, ruff + mypy clean.

jmelahman (approved):

PR #12342, opened 2026-06-23
fix(redis): async TLS/IAM passed an unsupported ssl_context — use native ssl_* kwargs

justin-tahara:
## Description

Fixes two async Redis bugs with the **same root cause**: redis-py's async client (5.0.8) has **no `ssl_context` parameter** (only the sync `SSLConnection` does), so passing one is either silently dropped or raises a `TypeError`. Both surfaced via live end-to-end TLS testing.

**1. `REDIS_SSL` path (mTLS / CA silently dropped).** `_build_async_redis_connection` set `connection_kwargs["ssl"] = <prebuilt ssl_context>`. The context — including the CA bundle and the mTLS client cert (`load_cert_chain`) — was ignored: TLS negotiated but **unverified and with no client cert**. Masked because `REDIS_SSL_CERT_REQS` defaults to `none`. (Follow-up to #12321, which merged before this fix.)

**2. `USE_REDIS_IAM_AUTH` path (hard crash).** `configure_redis_iam_auth` set `connection_kwargs["ssl_context"]`, which the async client rejects with `TypeError: Redis.__init__() got an unexpected keyword argument 'ssl_context'` — **every async Redis connection under ElastiCache IAM auth crashed at construction.**

**Fix (both):** pass redis-py's native async `ssl_*` kwargs (`ssl`, `ssl_cert_reqs`, `ssl_ca_certs`, `ssl_certfile`, `ssl_keyfile`, `ssl_check_hostname`) — the same mechanism the sync pool already uses. The sync paths are unchanged (sync `SSLConnection` does support `ssl_context`).

## How Has This Been Tested?

- **Live e2e** against a TLS + mTLS Redis (`tls-auth-clients yes`): async client PINGs with CA + client cert; rejected on wrong CA / missing client cert.
- Direct repro confirming the IAM path no longer raises the `ssl_context` `TypeError` and now sets verifying native kwargs.
- Unit tests assert native kwargs reach the async client for both the `REDIS_SSL` and IAM paths; `backend/tests/unit/onyx/redis` green; `ty` + `ruff` + `pre-commit` clean.

## Additional Options

### DIFF
--- a/backend/onyx/redis/iam_auth.py
+++ b/backend/onyx/redis/iam_auth.py
@@ -14,19 +14,26 @@ from typing import Any
 
 def configure_redis_iam_auth(connection_kwargs: dict[str, Any]) -> None:
     """
-    Configure Redis connection parameters for IAM authentication.
+    Configure async Redis connection kwargs for IAM authentication.
     Modifies the connection_kwargs dict in-place to:
     1. Remove password (not needed with IAM)
-    2. Enable SSL with system CA certificates
-    3. Set proper SSL context for secure connections
+    2. Enable TLS with server-cert verification against the system trust store
+       (ElastiCache uses a publicly-trusted CA)
+
+    Uses redis-py's native async ssl_* kwargs. The async client has no
+    ``ssl_context`` parameter, so passing one raises
+    ``TypeError: ... unexpected keyword argument 'ssl_context'`` — i.e. every
+    async IAM connection would crash.
     """
     # Remove password as it's not needed with IAM authentication
     if "password" in connection_kwargs:
         del connection_kwargs["password"]
 
-    # Ensure SSL is enabled for IAM authentication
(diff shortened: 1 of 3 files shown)


justin-tahara:
Thanks @greptile-apps — added the `ssl_check_hostname is False` assertion (1e7aa2b). It's a real behavioral change worth locking in. cubic flagged no issues. 7 tests pass, ty/ruff/pre-commit clean.

backend/onyx/utils/audit.py
nmgarza5, line 1: did these changes sneak in? want to remove?

nmgarza5 (approved):

backend/onyx/utils/audit.py
justin-tahara: Yeah cleaning up

PR #12346, opened 2026-06-23
fix(audit): attach dedicated stdout handler so api-server audit events aren't dropped

justin-tahara:
## Description

**Fixes a bug where audit events emitted from the api-server are silently dropped and never reach stdout / a SIEM.** Found by live-testing the audit subsystem on st-dev.

The audit loggers (`onyx.audit.*`) used a bare `getLogger` with **no handler** and relied on propagating to the root logger. In processes that don't configure root — notably the **uvicorn-based api-server**, which is where *all* auth / admin-config / access-control events are emitted — records fall through to Python's `lastResort` handler (level `WARNING`), so the `INFO` audit lines are **discarded**. Only celery-emitted events surfaced, because the celery bootstrap configures the root logger.

### How it was found
On st-dev (running an `edge` build with the audit code), an OAuth login produced `GET /auth/oauth/callback … 302`, which runs `on_after_login` and its unconditional `auth.login` emit — yet the api-server had emitted **0** `onyx.audit` lines ever. Logger introspection in the running pod confirmed it:
```
<root>                     handlers=0  eff=WARNING
onyx.audit                 handlers=0  eff=WARNING
lastResort_level=WARNING
```
The existing unit tests missed this because pytest's `caplog` attaches its own root handler, so emission "works" under test but not under uvicorn.

### Fix
`onyx.audit` now owns a dedicated `INFO` `StreamHandler(stdout)` that emits the bare JSON message, with `propagate=False` so the line is byte-identical across processes (matching the documented schema) and never double-logged where root *is* configured (celery). Idempotent. The `credential_audit.py` seed is covered automatically via the shared `onyx.audit` parent — its api-server emissions were being dropped too and are now fixed (seed file untouched).

## How Has This Been Tested?

- New regression test (`test_audit_logger_self_contained_without_root_handler`) asserts the subsystem owns a handler, sets `propagate=False`, and delivers an event with **no root handler present** — it fails on the pre-fix code.
- Added an autouse conftest fixture so the existing `caplog`-based audit tests capture the now-non-propagating logger.
- `backend/tests/unit/onyx/utils/test_audit.py`, `test_auth_audit_events.py`, `test_admin_audit_events.py` all pass (20); broader `unit/onyx/{utils,auth}` slice green (549 passed). `ty` + `ruff` clean.
- Next: redeploy `edge` to st-dev and re-run the OAuth-login test to confirm `auth.login` lands in the api-server logs.

## Additional Options

---
## Summary by cubic
Fixes dropped api-server audit events by giving `onyx.audit` its own `INFO` stdout handler and disabling propagation so JSON audit lines always reach stdout/SIEM. Removes reliance on `uvicorn`’s root logger that was discarding `INFO` records.

- **Bug Fixes**
  - Attached `onyx_audit_stdout` `StreamHandler(sys.stdout)` to `onyx.audit` with `%(message)s`, `level=INFO`; set logger `level=INFO`, `propagate=False` (idempotent at import).
  - Added regression test to verify delivery with no root handler; added autouse `pytest` fixture to route `caplog` to `onyx.audit`.
  - Child loggers (including the credential audit seed) inherit the handler; no emitter changes needed.

<sup>Written for commit 39779513248761db5d4dcffc8513879e000c4b91. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12346?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/utils/audit.py
+++ b/backend/onyx/utils/audit.py
@@ -13,6 +13,7 @@ hot event classes via Redis (degrade to always-emit if Redis is down).
 
 import json
 import logging
+import sys
 import time
 from dataclasses import dataclass
 from enum import Enum
@@ -20,10 +21,10 @@ from typing import Any
 
 AUDIT_SCHEMA_VERSION = "1.0"
 
-# Root of the audit logger tree; per-class children propagate up to it. Plain
-# ``getLogger`` (not ``setup_logger``) keeps the record message a bare JSON
-# object with no human-readable prefix.
+# Root of the audit logger tree; children propagate up to the dedicated stdout
+# handler configured in ``_configure_audit_logging``.
 AUDIT_LOGGER_ROOT = "onyx.audit"
+AUDIT_HANDLER_NAME = "onyx_audit_stdout"
 
 # Diagnostics logger for the subsystem's own failures. Deliberately NOT under
 # ``onyx.audit`` so internal warnings never land in the parsed audit stream.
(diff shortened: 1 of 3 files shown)


justin-tahara:
Addressed both Greptile comments in `3977951324`:

1. **conftest fixture return type** — changed `-> object` to `-> Generator[None, None, None]` (added the `collections.abc.Generator` import).

2. **Import-time side effect** — kept the module-level configuration and documented it prominently with a deliberate-side-effect comment (Greptile's "at minimum" option). I intentionally did **not** move to an explicit `audit.configure()` startup hook: that reintroduces the exact bug class this PR fixes — a process that emits audit events without someone having remembered to call `configure()` would silently drop them again, which is precisely what happened with uvicorn vs. the root logger. Import-time configuration is what makes the subsystem self-contained and host-independent. The side effect is also idempotent and inert — it only produces output once an event is actually emitted, and is analogous to the stdlib's recommended import-time `NullHandler` pattern.

jmelahman (approved):

PR #12351, opened 2026-06-23
fix(model_server): cap torch threads to cgroup CPU quota

justin-tahara:
## Description

The indexing model server periodically pegs its CPU limit and gets heavily CFS-throttled. Root cause is PyTorch thread oversubscription: `torch.set_num_threads(max(MIN_THREADS_ML_MODELS, torch.get_num_threads()))` sizes the intra-op thread pool from the **host** core count, ignoring the container's cgroup CPU quota. On a large node (observed: 384-core host, 8-core pod limit) this launches ~192 BLAS/OMP threads against an 8-core quota — they thrash, peg the limit, and get throttled ~95% of scheduler periods, slowing indexing throughput.

This makes the model server cgroup-aware:

- New `get_cgroup_cpu_limit()` helper in `model_server/utils.py` reads the CFS quota (cgroup v2 `cpu.max`, falling back to v1 `cpu.cfs_quota_us`/`cpu.cfs_period_us`). Returns `None` when unlimited/unreadable.
- `main.py` now caps `torch.set_num_threads` to that quota while preserving the historical `MIN_THREADS_ML_MODELS` floor. Startup log now prints torch default + detected cgroup limit for diagnosability.

Fixes both the indexing and inference model servers (shared entrypoint), for every deployment, with no values/env changes required.

## How Has This Been Tested?

- **Unit tests** (`tests/unit/model_server/test_cgroup_cpu_limit.py`): cgroup v2/v1 parsing, unlimited (`max` / `-1`), sub-core floor-to-1, rounding, missing files, and malformed contents. 8 cases, all pass.
- **Real cgroup smoke test**: ran the read logic in a Linux container under `docker run --cpus=N`. Host reports 14 cores, but the helper correctly reads the quota — `--cpus=2` → `2`, `--cpus=6` → `6` — i.e. it caps to the container budget rather than the host count.

## Additional Options

---
## Summary by cubic
Cap `torch` intra‑op threads to the container’s cgroup CPU quota to stop CPU thrash and CFS throttling, improving throughput for indexing and inference with no config changes. cgroup v2 is authoritative; parse failures are debug‑logged.

- **Bug Fixes**
  - Added `get_cgroup_cpu_limit()` that reads v2 `cpu.max` as authoritative (return `None` on `max` or unreadable), falling back to v1 `cpu.cfs_quota_us`/`cpu.cfs_period_us` only if the v2 file is missing; fractional quotas are floored (never over‑allocate) and sub‑core quotas floor to 1.
  - Updated `model_server/main.py` to cap `torch.set_num_threads` to the detected limit with the `MIN_THREADS_ML_MODELS` floor; logs torch default, detected quota, and final value.
  - Added tests for v2/v1 parsing, v2 `max` not falling through to v1, unreadable v2 remaining authoritative, unlimited quotas, fractional‑quota flooring, floor‑to‑1, and missing/malformed files.

<sup>Written for commit effc49d43e1550e10a9e2fa3658da81c77a5aaaf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12351?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/model_server/main.py
+++ b/backend/model_server/main.py
@@ -15,6 +15,7 @@ from transformers import logging as transformer_logging
 
 from model_server.encoders import router as encoders_router
 from model_server.management_endpoints import router as management_router
+from model_server.utils import get_cgroup_cpu_limit
 from model_server.utils import get_gpu_type
 from onyx import __version__
 from onyx.utils.logger import setup_logger
@@ -85,8 +86,24 @@ async def lifespan(app: FastAPI) -> AsyncGenerator:
             e,
         )
 
-    torch.set_num_threads(max(MIN_THREADS_ML_MODELS, torch.get_num_threads()))
-    logger.notice("Torch Threads: %s", torch.get_num_threads())
+    # torch sizes its intra-op thread pool from the host core count, ignoring the
+    # container's cgroup CPU quota. On a large node this oversubscribes threads against
+    # a small quota, causing thread thrash and CFS throttling. Cap to the quota (while
+    # keeping the historical MIN_THREADS_ML_MODELS floor).
+    torch_default_threads = torch.get_num_threads()
+    cpu_limit = get_cgroup_cpu_limit()
+    num_threads = (
+        torch_default_threads
+        if cpu_limit is None
+        else min(torch_default_threads, cpu_limit)
+    )
+    torch.set_num_threads(max(MIN_THREADS_ML_MODELS, num_threads))
+    logger.notice(
+        "Torch Threads: %s (torch default: %s, cgroup cpu limit: %s)",
+        torch.get_num_threads(),
+        torch_default_threads,
+        cpu_limit,
+    )
 
     yield
 
--- a/backend/model_server/utils.py
+++ b/backend/model_server/utils.py
@@ -4,6 +4,7 @@ from collections.abc import Callable
 from collections.abc import Generator
 from collections.abc import Iterator
 from functools import wraps
+from pathlib import Path
 from typing import Any
 from typing import cast
 from typing import TypeVar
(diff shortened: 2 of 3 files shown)


justin-tahara:
Addressed both review comments in `ad76e9a5af`:

- **Greptile (v2 `max` fallthrough):** valid — when cgroup v2 `cpu.max` is present it's authoritative, so we now `return None` immediately on `max` rather than consulting v1 (a hybrid host can carry a stale non-zero v1 quota that doesn't reflect reality). Added a regression test (`test_cgroup_v2_max_does_not_fall_through_to_v1`).
- **cubic (swallowed exceptions):** now log at `debug` for genuine parse/permission failures, while keeping the expected file-absent case (`FileNotFoundError`) silent so we don't emit startup noise on hosts that simply aren't running that cgroup version. The resolved limit is also already logged at notice level in `main.py`.

justin-tahara:
Addressed both in `effc49d43e` (identified by cubic):

- **Fractional quota over-allocation:** switched to integer floor division so the cap never exceeds the quota — `round(1.5)=2` over-allocated, floor gives 1. Verified end-to-end against real cgroups: `docker run --cpus=1.5` → 1, `--cpus=2.9` → 2, `--cpus=8` → 8. Added `test_cgroup_v2_floors_fractional_quota` (7.5 → 7).
- **v2 read/parse failure falling through to v1:** a present-but-unparseable `cpu.max` now returns `None` rather than consulting v1, keeping v2 authoritative; only a *missing* v2 file falls back to v1. Added `test_cgroup_v2_malformed_does_not_fall_through_to_v1`.

10 unit tests + pre-commit green.

evan-onyx (approved):

PR #12365, opened 2026-06-24
fix: hierarchy node cross batch stubs

evan-onyx:
## Description

We were previously enforcing that hierarchynodes needed to be yielded parent-before-child, and then worst case picking up those relationships on the next indexing or pruning attempt. This worst case wasn't ideal, since if the child wasn't reindexed in a future run (i.e. it didn't get modified), users might have to wait until the pruning run kicked in for nodes to be parented correctly. Notion in particular was consistently hitting this worst case, as the connector structure makes it very inconvenient to actually yield parent-before-child.

Now we're adding a reasonably efficient mechanism for handling the case of "child before parent" in docfetching. Basically we make a stub node when a child is yielded without its parent, then update the node with its real info when it becomes available.

## How Has This Been Tested?

added tests
## Additional Options

---
## Summary by cubic
Fixes incorrect parenting when a child node is indexed before its parent by creating `STUB` placeholders and promoting them when the parent arrives. Adds concurrency-safe stub creation and hides stubs from user-facing queries, improving Notion and Google Drive syncs.

- **Bug Fixes**
  - Added `HierarchyNodeType.STUB` for missing parents; promote in place on parent upsert so child links stay correct.
  - Updated batch upsert to upsert in-batch ancestors first and create cross-batch stubs; returns all upserted nodes (including stubs); prevents duplicate stubs within a session and across concurrent workers.
  - Excluded `STUB` nodes from accessible/source listing queries in both `backend/onyx` and `backend/ee/onyx`.
  - Added tests for cross-batch stub creation/promotion and one-stub-for-many-children; updated Google Drive walker test to validate stub-based healing across yields.

<sup>Written for commit c7818a1447b49af647695bcdc850dc32f0003b60. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12365?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/ee/onyx/db/hierarchy.py
+++ b/backend/ee/onyx/db/hierarchy.py
@@ -14,6 +14,7 @@ from sqlalchemy.orm import Session
 from sqlalchemy.sql.elements import ColumnElement
 
 from onyx.configs.constants import DocumentSource
+from onyx.db.enums import HierarchyNodeType
 from onyx.db.models import HierarchyNode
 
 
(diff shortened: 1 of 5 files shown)


nmgarza5 (approved):
i like it 🔥

Weves (approved):

PR #12378, opened 2026-06-24
refactor(indexing): require csv_file_id for TabularSection

nmgarza5:
## Description

Builds on #12292. Removes the inline `text` option from `TabularSection`: a sheet's CSV is now **always** staged in the file store and streamed at chunk time — one mechanism instead of `text`-or-`csv_file_id` (per Dane's review). Non-docfetching connector runs lack the docfetching staging callback, so they're wired up or tabular files vanish — notably **pruning**, where non-slim tabular connectors (blob/file) would otherwise lose doc ids and pruning would **delete the indexed docs**. The batch-handoff deserialize is made tolerant so the model change survives a rolling deploy.

### Changes by location

**Model**
- `connectors/models.py` — `TabularSection`: drop `text`, `csv_file_id: str` required, remove the runtime "exactly one of" validator.

**Extraction & staging**
- `file_processing/extract_file_text.py` — `StreamedSheet` is now `(title, csv_file_id)`; `stage_or_inline_xlsx_sheets` → `stage_xlsx_sheets` (always stages, no inline branch); remove dead `xlsx_has_large_sheet`.
- `configs/app_configs.py` — remove dead `XLSX_STREAM_SHEET_BYTES`.
- `connectors/cross_connector_utils/tabular_section_utils.py` — `tabular_file_to_sections` takes a required `stage` callback and stages each sheet/file CSV → `csv_file_id`; `extract_and_stage_tabular_file` threads it through.
- `file_store/staging.py` — add `build_tracking_raw_file_callback` (returns the callback + the list of staged ids) for paths with no attempt-end reaper.

**Chunking**
- `indexing/chunking/tabular_section_chunker/tabular_section_chunker.py` — `chunk_section` file-backed only (inline branch removed); header-only sheets still get a descriptor.

**Connectors (docfetching producers)**
- `braintrust` — stage via `_stage_csv` → `csv_file_id`; skip only the tabular section (keep the TextSection) if no callback.
- `blob` — stage via `extract_and_stage_tabular_file`; drop the empty `text=""` fallback; skip+warn if no callback.
- `file` — `_process_file` takes `stage`; `LocalFileConnector` passes `self.raw_file_callback`; skip+warn if None.
- `sharepoint` — return a `ConnectorFailure` when no callback (and on tabular extraction error) instead of emitting an empty-section `Document`.
- `drupal_wiki`, `google_drive/doc_conversion` — drop the no-callback inline fallback (warn + skip).

**Non-docfetching connector runs (the "everywhere" fixes)**
- `background/celery/tasks/user_file_processing/tasks.py` — tracking callback on the file connector; reap staged CSVs in `finally`.
- `background/celery/celery_utils.py` (`extract_ids_from_runnable_connector`, the pruning enumeration) — tracking callback + reap in `finally`; without it non-slim tabular connectors lose doc ids → pruning deletes them.
- `background/indexing/run_targeted_reindex.py` — tracking callback + reap staged CSVs in `finally` (this path has no attempt-end reaper).

**Rolling-deploy safety**
- `file_store/document_batch_storage.py` — on deserialize, skip a doc carrying a legacy inline tabular section (no `csv_file_id`) instead of crashing the whole batch on `Document.model_validate`. During a rolling deploy an old docfetcher can stage a text-only section; the doc is skipped (logged) and re-indexes on the next attempt rather than failing the batch.

**Tests** — `test_tabular_section_chunker.py`, `test_tabular_section_utils.py`, `test_braintrust_connector.py` updated to file-backed (mock file store / fake stage).

## Additional Options

### DIFF
--- a/backend/onyx/background/celery/tasks/user_file_processing/tasks.py
+++ b/backend/onyx/background/celery/tasks/user_file_processing/tasks.py
@@ -37,6 +37,7 @@ from onyx.configs.constants import USER_FILE_PROJECT_SYNC_MAX_QUEUE_DEPTH
 from onyx.connectors.file.connector import LocalFileConnector
 from onyx.connectors.models import Document
 from onyx.connectors.models import HierarchyNode
+from onyx.connectors.models import TabularSection
 from onyx.db.engine.sql_engine import get_session_with_current_tenant
 from onyx.db.enums import UserFileStatus
 from onyx.db.models import UserFile
@@ -273,10 +274,22 @@ def _process_user_file_without_vector_db(
 
     user_file_uuid = _as_uuid(user_file_id)
 
-    # Combine section text from all document sections
-    combined_text = " ".join(
-        section.text for doc in documents for section in doc.sections if section.text
-    )
+    # Combine section text from all document sections. Tabular sections are
+    # file-backed (no inline text), so read their staged CSV from the file store.
+    file_store = get_default_file_store()
+    text_parts: list[str] = []
+    for doc in documents:
+        for section in doc.sections:
+            if isinstance(section, TabularSection):
+                with file_store.read_file(
+                    section.csv_file_id, use_tempfile=True
+                ) as raw:
+                    csv_text = raw.read().decode("utf-8", errors="replace").strip()
+                if csv_text:
+                    text_parts.append(csv_text)
+            elif section.text:
+                text_parts.append(section.text)
+    combined_text = " ".join(text_parts)
 
     # Compute token count using the user's default LLM tokenizer
     try:
--- a/backend/onyx/background/celery/tasks/user_file_processing/tasks.py
+++ b/backend/onyx/background/celery/tasks/user_file_processing/tasks.py
@@ -275,16 +275,12 @@ def _process_user_file_without_vector_db(
     user_file_uuid = _as_uuid(user_file_id)
 
     # Combine section text from all document sections. Tabular sections are
-    # file-backed (no inline text), so read their staged CSV from the file store.
-    file_store = get_default_file_store()
+    # file-backed; read their staged CSV on demand.
     text_parts: list[str] = []
     for doc in documents:
         for section in doc.sections:
             if isinstance(section, TabularSection):
-                with file_store.read_file(
-                    section.csv_file_id, use_tempfile=True
-                ) as raw:
-                    csv_text = raw.read().decode("utf-8", errors="replace").strip()
+                csv_text = section.read_csv_text().strip()
                 if csv_text:
                     text_parts.append(csv_text)
             elif section.text:
--- a/backend/onyx/background/celery/tasks/user_file_processing/tasks.py
+++ b/backend/onyx/background/celery/tasks/user_file_processing/tasks.py
@@ -46,6 +46,8 @@ from onyx.db.user_file import fetch_user_files_with_access_relationships
 from onyx.document_index.factory import get_all_document_indices
 from onyx.document_index.interfaces_new import MetadataUpdateRequest
 from onyx.file_store.file_store import get_default_file_store
+from onyx.file_store.staging import build_tracking_raw_file_callback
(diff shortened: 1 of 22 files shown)


nmgarza5:
@greptile

backend/onyx/background/celery/tasks/user_file_processing/tasks.py
```diff
@@ -321,10 +324,22 @@ def _process_user_file_without_vector_db(
...
-    combined_text = " ".join(
-        section.text for doc in documents for section in doc.sections if section.text
-    )
+    # Combine section text from all document sections. Tabular sections are
+    # file-backed (no inline text), so read their staged CSV from the file store.
+    file_store = get_default_file_store()
+    text_parts: list[str] = []
+    for doc in documents:
+        for section in doc.sections:
+            if isinstance(section, TabularSection):
```
Danelegend: I think we can override the .text property of TabularSection, so the logic here stays the same.

Instead .text is a property function that reaches the file_store and reads the file from the file store based on the csv_file_id.

Basically abstracting this logic away in TabularSection

Danelegend: This might cause some changes in other places but overall I think it'll be cleaner

backend/onyx/file_store/document_batch_storage.py
```diff
@@ -20,6 +20,20 @@
 logger = setup_logger()
 
 
+def _has_legacy_tabular_section(doc_dict: dict) -> bool:
```
Danelegend, line 23: With the proposed change above, this may be able to be removed since we technically still have the 'text' property?

Danelegend (approved):

backend/onyx/background/celery/tasks/user_file_processing/tasks.py
nmgarza5: Good call on abstracting it. I used read_csv_text() instead of a .text property though. get_text_content only joins sections whose text is set, so a sheet is skipped there today (TabularSection.text is None). A file backed .text would flip that and load the whole CSV into memory during contextual RAG (chunker fit check + doc summary), which is the OOM we're avoiding. It'd also break model_copy(update={"text"}) in section_utils. So kept it explicit off the DISABLE_VECTOR_DB path.

backend/onyx/file_store/document_batch_storage.py
nmgarza5: Kept this one. csv_file_id is required, so a legacy section without it still fails model_validate on a rolling deploy whether or not text is a property, and that crashes the batch. The guard skips those legacy docs so the deploy survives.

nmgarza5: can be removed after these changes have been live for a bit

backend/onyx/background/celery/tasks/user_file_processing/tasks.py
Danelegend: `if isinstance(section, TabularSection):` is kind of a code smell and I don't think it's the best idea to have instance specific code here. I thought my above idea would solve it but looking into it, it looks like it doesn't.

What we could do is make a `materialize_text` method of Section that returns self.text or "". The TabularSection overrides this will the read_csv_text logic.

Now this change here just becomes:
```
combined_text = " ".join(
        section.materialize_text() for doc in documents for section in doc.sections if section.materialize_text()
    )
```

backend/onyx/file_store/document_batch_storage.py
Danelegend: Yep, that checks out

backend/onyx/background/celery/tasks/user_file_processing/tasks.py
Danelegend: Also, worth noting that this is materializing all the CSVs again. Although it's in the user processing file pipeline so should be chill

PR #12402, opened 2026-06-24
fix(craft): keep sandbox-proxy from crash-looping on slow startup

rohoswagger:
## Description

Hardens the Craft `sandbox-proxy` so a stalled boot-time Kubernetes API call fails fast and observably instead of crash-looping. Backend-only — no Helm/chart changes.

> **Scope:** this is the **impact** fix (make the failure graceful + diagnosable), not the **root-cause** fix. The incident's actual trigger — `aws-network-policy-agent` leaving the pod's eBPF datapath in a deny-all state for ~75 min under churn — is tracked separately and is what prevents recurrence. This PR makes the proxy resilient to that *and any other* boot-time apiserver stall; it does not (and cannot) keep the proxy serving during a real datapath blackout.

### The incident (confirmed live in prod)
On `managed_services`, a `sandbox-proxy` pod restarted **13×** before stabilizing. The pod is still live and its scars confirm the failure exactly: `restarts=13`, last termination `exit=137` (SIGKILL from the liveness probe — **not** OOM), lifetime `21:24:09Z → 21:29:54Z = 345s`, logs frozen at `loaded in-cluster Kubernetes config`.

### Root cause (three layers)
1. **Amplifier — untimed boot calls.** The proxy's first boot action is a synchronous, *untimed* k8s API call: the CA Secret read (`read_namespaced_secret`), plus the CA ConfigMap create/replace and the informer's initial pod list. With no `_request_timeout`, a half-open apiserver socket blocks until the kernel's TCP retransmission limit (minutes) instead of failing fast.
2. **Multiplier — healthz gated behind the hang.** `/healthz` bound only *after* CA bootstrap + informer sync, so while a boot call hung, `:8081` was unbound and the kubelet got connection-refused (indistinguishable from "process dead") → SIGKILL mid-boot.
3. **Trigger (separate, not fixed here).** A node-wide reschedule placed 6 danswer pods on one node in ~75s. Only `sandbox-proxy` crash-looped; the co-tenant pods (api-server, all celery workers) booted with 0 restarts. The single differentiator: `sandbox-proxy` is the only pod under an **egress NetworkPolicy**, so only its apiserver egress depends on `aws-network-policy-agent` programming egress-allow eBPF rules. Under the churn burst the agent (standard mode, v1.3.5) failed to program those rules correctly for ~75 min, silently dropping the proxy's apiserver packets until it reconciled. Ruled out: apiserver contention (co-tenants fine), strict-mode deny-window (cluster is `standard`), fresh-node/agent-restart (node 1.8d old, agent 0 restarts), dual-stack ipBlock fail-closed (policy correctly splits v4/v6).

### What this PR changes (backend only)
- **Client-side timeouts on every boot call (the durable fix).** New shared `build_core_v1_api()` factory (connect-only retries — safe, a failed connect means the request never reached the server) and `K8S_BOOT_REQUEST_TIMEOUT_S = (5s, 15s)` passed to the CA Secret read/create, CA ConfigMap create/replace, and the informer's initial list. A stalled connection now fails fast and the existing informer retry/backoff + 60s boot guard engage as designed. The informer watch stream keeps its server-side 300s timeout (a short client read timeout would kill the watch prematurely).
- **Bind `/healthz` before the blocking boot steps.** The probe answers an honest 503 (not connection-refused) from t=0; `ca_ready`/`is_synced()` stay false until genuinely ready. Inlines the now-trivial `_build_lookup` helper.

No probe/Helm change: with the timeouts in place, no boot call can hang, and the worst-case *legitimate* boot (~CA + the 60s informer-sync guard ≈ 75s) already fits inside the existing liveness budget (`30 + 8×15 = 150s`). A startupProbe would not be load-bearing, so it's omitted.

### Scope — what this does and does NOT do
**This PR (impact / hardening):** makes the proxy robust to *any* boot-time apiserver stall (CNI reconcile, apiserver contention, control-plane failover, node blip), turning 13× 345s SIGKILL hangs into fast clean restarts + honest 503 + automatic recovery once the datapath returns. Trigger-agnostic and in our control to land now. Its biggest practical win is **diagnosability** — today the failure is a silent 345s hang with healthz unbound (looks like a dead pod; initially misdiagnosed as OOM); with this change it's an explicit `ReadTimeoutError` + honest 503 that points straight at the network.

**Not this PR (root cause / frequency):** the datapath blackout itself is an `aws-network-policy-agent` reconcile bug (deny-all under churn, standard mode, v1.3.5), tracked as a separate follow-up. Upgrading the agent is what prevents recurrence; this PR does not change availability during a genuine blackout (the proxy correctly fails closed — no apiserver, no identity backing, no serving). Merging this should **not** be read as closing the incident; the agent upgrade is the incident-closer.

## How Has This Been Tested?

- **Live A/B on craft-dev (EKS, real VPC CNI).** Overlaid the changed files onto the running image and pointed boot at a fake apiserver that accepts the TCP connection but never replies (deterministically reproduces the incident's half-open hang):
  - **OLD (no fix):** hung indefinitely on `read_namespaced_secret` for `/secrets/sandbox-proxy-ca`, 0 restarts, `:8081` connection-refused — the exact incident signature.
  - **NEW (this PR):** `ReadTimeoutError (read timeout=5.0) → MaxRetryError`, clean `exit 1`; logs show `healthz listening on 0.0.0.0:8081` *before* the CA bootstrap (early bind active).
- **Prod forensics (read-only).** Confirmed the live incident pod matches the documented failure to the second (exit 137 / 345s / 13 restarts / freeze point) and isolated the trigger to the egress-policy eBPF program (see root cause).
- **Static checks:** `ruff` (check + format) and `ty` pass on all changed files; verified `Configuration.retries` + the `(connect, read)` tuple are honored by the installed kubernetes client.

Details: `docs/craft/infra/sandbox-proxy-boot-resilience.md`.

## Additional Options

### DIFF
--- a/backend/onyx/sandbox_proxy/ca_k8s.py
+++ b/backend/onyx/sandbox_proxy/ca_k8s.py
@@ -19,7 +19,10 @@ from onyx.server.features.build.configs import SANDBOX_NAMESPACE
 from onyx.server.features.build.configs import SANDBOX_PROXY_CA_CONFIGMAP
 from onyx.server.features.build.configs import SANDBOX_PROXY_CA_SECRET
 from onyx.server.features.build.configs import SANDBOX_PROXY_NAMESPACE
-from onyx.server.features.build.sandbox.kubernetes.k8s_client import load_kube_config
+from onyx.server.features.build.sandbox.kubernetes.k8s_client import build_core_v1_api
+from onyx.server.features.build.sandbox.kubernetes.k8s_client import (
+    K8S_BOOT_REQUEST_TIMEOUT_S,
+)
 from onyx.server.features.build.sandbox.labels import LABEL_K8S_COMPONENT
 from onyx.server.features.build.sandbox.labels import LABEL_K8S_MANAGED_BY
 from onyx.server.features.build.sandbox.labels import LABEL_K8S_MANAGED_BY_ONYX
@@ -58,8 +61,7 @@ class K8sSecretCAStore(CAStore):
         configmap_name: str = SANDBOX_PROXY_CA_CONFIGMAP,
     ) -> None:
         if core_api is None:
-            load_kube_config()
-            core_api = client.CoreV1Api()
+            core_api = build_core_v1_api()
         self._core = core_api
         self._proxy_ns = proxy_namespace
         self._sandbox_ns = sandbox_namespace
@@ -69,7 +71,9 @@ class K8sSecretCAStore(CAStore):
     def load(self) -> tuple[bytes, bytes] | None:
         try:
             secret = self._core.read_namespaced_secret(
-                name=self._secret_name, namespace=self._proxy_ns
+                name=self._secret_name,
+                namespace=self._proxy_ns,
+                _request_timeout=K8S_BOOT_REQUEST_TIMEOUT_S,
             )
         except ApiException as e:
             if e.status == 404:
@@ -120,7 +124,11 @@ class K8sSecretCAStore(CAStore):
         )
 
         try:
-            self._core.create_namespaced_secret(namespace=self._proxy_ns, body=secret)
+            self._core.create_namespaced_secret(
+                namespace=self._proxy_ns,
+                body=secret,
+                _request_timeout=K8S_BOOT_REQUEST_TIMEOUT_S,
+            )
         except ApiException as e:
             # 409 on create = cold-cluster race lost.
             if e.status in (409, 422):
@@ -149,7 +157,9 @@ class K8sSecretCAStore(CAStore):
 
         try:
             self._core.create_namespaced_config_map(
-                namespace=self._sandbox_ns, body=body
+                namespace=self._sandbox_ns,
+                body=body,
+                _request_timeout=K8S_BOOT_REQUEST_TIMEOUT_S,
             )
             return
         except ApiException as e:
@@ -165,6 +175,7 @@ class K8sSecretCAStore(CAStore):
                     name=self._configmap_name,
                     namespace=self._sandbox_ns,
                     body=body,
+                    _request_timeout=K8S_BOOT_REQUEST_TIMEOUT_S,
                 )
                 return
             except ApiException as e:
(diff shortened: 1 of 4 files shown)


backend/onyx/sandbox_proxy/ca_k8s.py
```diff
@@ -19,7 +19,10 @@
 from onyx.server.features.build.configs import SANDBOX_PROXY_CA_CONFIGMAP
 from onyx.server.features.build.configs import SANDBOX_PROXY_CA_SECRET
 from onyx.server.features.build.configs import SANDBOX_PROXY_NAMESPACE
-from onyx.server.features.build.sandbox.kubernetes.k8s_client import load_kube_config
+from onyx.server.features.build.sandbox.kubernetes.k8s_client import build_core_v1_api
+from onyx.server.features.build.sandbox.kubernetes.k8s_client import (
+    K8S_BOOT_REQUEST_TIMEOUT_S,
```
wenxi-onyx: can this be configured once via some client config instead of re-invoked every time?

Also, the var name is a bit confusing as this timeout is for any client operation, not just boot

wenxi-onyx (approved):

backend/onyx/sandbox_proxy/ca_k8s.py
rohoswagger: Good call on both — fixed in 279394b0b2.

**Configure once, not per call:** The timeout is now baked into the client. `build_core_v1_api()` returns a small `client.ApiClient` subclass (`_DeadlinedApiClient`) whose `request()` injects the deadline whenever a call omits one. The kubernetes client has no Configuration-level request timeout (only `retries` lives on `Configuration`, which is why connect-retries were already set there), so a subclass is the single place this can live. All five boot call sites in `ca_k8s.py` / `identity_k8s.py` drop their `_request_timeout=` argument, and the exported `K8S_BOOT_REQUEST_TIMEOUT_S` constant is gone (now a private `_REQUEST_TIMEOUT_S` inside the builder).

**Naming:** Agreed it read as misleading — renamed to `_REQUEST_TIMEOUT_S` and made it private, since it's no longer something callers reference. One subtlety worth flagging: the same client is also used by the long-lived informer watch (`_watch_loop` in `identity_k8s.py`), which legitimately reads nothing for up to `_WATCH_TIMEOUT_SECONDS` (300s). A blanket 15s read deadline would make the idle watch reconnect-storm, so the watch explicitly opts out with a connect-only deadline `(15s, None)` — connect fail-fast preserved, no read deadline. Verified the informer unit tests still pass.

rohoswagger:

PR #12410, opened 2026-06-24
fix(pruning): stop a failed prune from re-firing and flooding the cleanup queue

nmgarza5:
## Description

A connector prune that fails *after* fanning out its per-doc cleanup tasks called `reset()` — which clears the fence/taskset but does **not** revoke the already-queued tasks — and because `last_pruned` only advances on full completion, the scheduled beat re-fired the prune every ~20s. Each re-fire re-enumerated and re-queued the whole set: one stuck prune produced ~1.8M cleanup tasks (~2.4x the doc count) on the `connector_deletion` queue, wedging the light fleet.

Fix:
- Only `reset()` when no cleanup tasks were fanned out (`generator_complete is None`). If they were, keep the fence so the monitor finalizes the in-flight taskset instead of orphaning them and re-queuing.
- Add a 30-min failure backoff (survives `reset()`) that the scheduled beat respects, so a persistently-failing prune can't hot-loop. Manual prunes bypass it.

## How Has This Been Tested?

- `ruff format --check`, `ruff check`, and `ty check` pass on both files.
- No automated test added: the re-fire path is integration-level (generator + monitor + beat); behavior verified by reading those paths.

## Additional Options

### DIFF
--- a/backend/onyx/background/celery/tasks/pruning/tasks.py
+++ b/backend/onyx/background/celery/tasks/pruning/tasks.py
@@ -246,6 +246,14 @@ def check_for_pruning(self: Task, *, tenant_id: str) -> bool | None:
                         logger.info("CC pair not due for pruning: %s", cc_pair_id)
                         continue
 
+                    # Skip auto-scheduling during a prune failure backoff; a manual
+                    # prune (API) bypasses this path and can still force a run.
+                    if RedisConnector(tenant_id, cc_pair_id).prune.in_failure_backoff:
+                        logger.info(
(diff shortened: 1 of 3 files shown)


Bo-Onyx (approved):
Shipped.

Had one question: what are those real failures. 
Did some research, fan-out is not the last step. After generate_tasks returns, the generator does a whole synchronous hierarchy-node pruning phase before it finishes.So "after fan-out" there are three Postgres mutations (each with its own commit=True) plus two Redis cache operations still to run.

PR #12411, opened 2026-06-24
feat(hierarchy): add hierarchy node display_name search endpoint

evan-onyx:
## Description

Adds `GET /hierarchy-nodes/search?query=<str>&source=<source>` — a new endpoint that searches hierarchy nodes by `display_name` substring (case-insensitive, LIKE-escaped). Supports optional multi-valued `source` filtering. Returns at most 30 results.

- MIT version returns all ACL-accessible nodes; EE version applies the existing email/group permission filter from `ee/onyx/db/hierarchy.py`.
- STUB and SOURCE-type root nodes are always excluded from results.
- Response shape is `HierarchyNodeSearchResponse { nodes: HierarchyNodeSearchSummary[] }` where each summary includes `id`, `title`, `link`, `parent_id`, and `source` (needed for source icons in the UI).
- Also fixes a worktree-specific `ty` false-positive: adds `extra-paths = ["backend"]` to `[tool.ty.environment]` in `pyproject.toml` so ty resolves `onyx.*` from the current project's backend first, even when `PYTHONPATH` contains another checkout of the repo.

This is the backend half of the knowledge pane search revamp ([plan](plans/knowledge-pane-search-revamp.md)). The frontend will call this endpoint in parallel with `POST /api/search/send-search-message` when the user submits a search query in the agent knowledge picker.

## How Has This Been Tested?

12 external-dependency-unit tests in `backend/tests/external_dependency_unit/hierarchy/test_hierarchy_node_search.py`:
- Case-insensitive substring matching (lower and upper case)
- No-match returns empty list
- STUB nodes excluded
- SOURCE-type root nodes excluded (non-vacuous: searches by the SOURCE node's display_name and asserts it's absent)
- Single-source filter narrows results
- Multi-source filter
- LIKE metacharacter escaping (`_` treated as literal, not wildcard)
- EE: public node visible to all
- EE: email-gated node visible only to matching user
- EE: group-gated node visible only to matching group
- EE: hidden node not visible to anyone

## Additional Options


pyproject.toml
```diff
@@ -224,6 +224,10 @@ override-dependencies = [
 [tool.ty.environment]
 python-version = "3.13"
 root = ["./backend"]
+# Ensure ty resolves onyx.* from this project's backend directory first, even
+# when PYTHONPATH contains another checkout (e.g. when working in a git worktree
+# alongside the main repo).
+extra-paths = ["backend"]
```
evan-onyx, line 230: @nmgarza5 how do you handle type checking in worktrees?

nmgarza5: I don't do anything special or export `PYTHONPATH` from my `.zshrc`.  Each worktree gets its own .venv from uv sync, and ty runs via uv run ty check in pre-commit, so onyx.* always resolves from that worktree's own backend/.

nmgarza5 (approved):

justin-tahara (approved):

PR #12412, opened 2026-06-24
feat(search): Skip source filter once no filter settled

Danelegend:
## Description
Currently the source filter is ran on every internal search tool. However, there are many scenarios where a filter is not mentioned in a prompt. As such, in these cases, we do not want to keep running the filter if it has been provided that we should not use a filter.

This PR checks if no filter has been provided, and saves this so that other internal searches in the same iteration don't use a filter.

## How Has This Been Tested?
Tests + Manual 

## Additional Options

---
## Summary by cubic
Skip source-scope decision after the first no-scope result within a user turn. Run query expansion and scope decision together (in parallel when both apply) and reuse cached expansions to cut latency and LLM calls.

- **Refactors**
  - Added `_expand_queries_and_decide_scope` and `QueryExpansionAndScope` to coordinate rephrase, keyword expansion, and scope decision; runs conditionally in parallel and logs one combined duration.
  - Cache semantic/keyword expansions across cycles; respect `skip_query_expansion`.
  - Latch scope decision off after an unscoped result; keep deciding while a directive exists. Added unit tests for both behaviors.

<sup>Written for commit 08a3b4a9fc8cd250c7c03c6643c53a3c7c59ef71. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12412?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/tools/tool_implementations/search/search_tool.py
+++ b/backend/onyx/tools/tool_implementations/search/search_tool.py
@@ -585,11 +585,13 @@ class SearchTool(Tool[SearchToolOverrideKwargs]):
         decide_scope = not self._scope_decision_settled
 
         jobs: list[tuple[Callable, tuple]] = []
+        scope_job_index: int | None = None
         if expand_queries:
             expansion_args = (message_history, self.llm, user_info, memories)
             jobs.append((semantic_query_rephrase, expansion_args))
             jobs.append((keyword_query_expansion, expansion_args))
         if decide_scope:
+            scope_job_index = len(jobs)
             jobs.append((decide_search_scope, decide_args))
 
         results = run_functions_tuples_in_parallel(jobs) if jobs else []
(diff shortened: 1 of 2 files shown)


backend/onyx/tools/tool_implementations/search/search_tool.py
```diff
@@ -565,6 +567,49 @@ def emit_start(self, placement: Placement) -> None:
...
         )
 
+    def _expand_queries_and_decide_scope(
+        self,
+        skip_query_expansion: bool,
+        message_history: list[ChatMinimalTextMessage],
+        user_info: str | None,
+        memories: list[str],
+        decide_args: tuple[Any, ...],
+    ) -> tuple[str | None, list[str], list[DocumentSource] | None]:
```
evan-onyx: probably should be BaseModel

evan-onyx (approved):

PR #12436, opened 2026-06-25
fix(chat): bound the LLM stream consumer (ENG-4236) to release v4.2

justin-tahara:
Cherry-pick of #12433 to **release/v4.2**.

A degenerate LLM stream that emits empty packets (no content/reasoning/tool
calls) without end holds an anyio worker thread blocked on the socket; enough
concurrent stuck streams saturate the pool, `/health` stops responding, and the
liveness probe SIGKILLs the api-server pod (exit 137) — an api-server restart
loop observed in production. Adds a consecutive-empty-packet cap (resets on any
non-empty packet) plus a generous wall-clock backstop, both raising
`LLMStreamError` so a runaway stream becomes a single failed request instead of
a pod-wide wedge. Limits are env-overridable in `chat_configs.py`.

See #12433 for full context and ENG-4236. Also being cherry-picked to v4.0 and v4.1.

> ⚠️ @yuhongsun96 — touches the body of `run_llm_step_pkt_generator`; additive
> safety guard only, normal streaming path unchanged.

## How Has This Been Tested?

Unit tests in `test_llm_step_stream_guards.py` pass on `main` (the change is
identical). Note: this repo's unit suite currently fails to *collect* on the
v4.2 branch in some local envs due to a pre-existing `UnicodeEncodeError` in
`text_processing.py` import — unrelated to this change; CI runs in the correct
environment.

## Additional Options


evan-onyx:
upstream was closed so closing this too

PR #12415, opened 2026-06-25
feat(craft): connect external apps on demand — backend request flow

Danelegend:
## What

Backend for connecting org external apps on demand. When the agent calls the no-op `connect_app` tool, opencode emits a permission request; the api-server turns it into a connect card on the user's screen and resumes the agent once the user connects (or declines).

## How

- **`connect_app`** — a Redis **announce** channel (request → the worker holding the user's live SSE stream) plus a single-shot **decision** latch (answer → the parked turn). No DB row, no `/live` poll.
- **`serve_client`** — intercepts the `connect_app` `permission.asked`, announces the request keyed by the build session, blocks on the decision, then answers opencode allow (connected) / reject (declined). A wait timeout declines.
- **`streaming`** — merges connect-app announces into the live SSE stream as a `ConnectAppRequestPacket`.
- **`external_apps`** — `POST /build/apps/connect/{request_id}/decision` to resolve a request, plus `supports_oauth` on the user app view.
- Drops the now-unused `CRAFT_REQUEST_APP_SETUP` permission and the `require_scoped_token` gate (replaced by this flow).

The agent tool (#12443) and the connect card UI (#12444) stack on this.

- [x] [Optional] Override Linear Check

PR #12443, opened 2026-06-25
feat(craft): connect_app sandbox tool + agent instructions

Danelegend:
## What

The agent-facing half of connect-on-demand: a no-op `connect_app` tool plus the AGENTS.md section that tells the agent which org apps a user can still connect.

## How

- **`connect-app.ts`** — the no-op opencode tool. Its `context.ask` pauses the turn and emits `permission.asked`; the backend (#12415) drives the connect card and answers allow/deny. On decline the tool returns a normal result so the agent keeps control and picks an alternative.
- **`opencode_config`** — `connect_app: "ask"` so the tool prompts rather than auto-allowing.
- **docker / kubernetes sandbox managers** — register the connect-app plugin.
- **`agent_instructions` + `skills/push`** — list the user's connectable apps in AGENTS.md.
- **`db`** — `get_connectable_apps_for_user`.

Stacks on #12415 (backend request flow).

### DIFF
--- a/backend/onyx/db/skill.py
+++ b/backend/onyx/db/skill.py
@@ -114,12 +114,17 @@ def _add_user_visibility_filter(
 
 
 def visible_skill_ids_for_user(user: User, db_session: Session) -> set[UUID]:
-    """Enabled skill ids the user can see (public / group-granted / personal) —
-    the visibility source of truth shared with sandbox injection."""
+    """Enabled skill ids the user can see (public / group-granted / personal).
+
+    A lightweight authorization primitive (ids only, for membership checks such
+    as which external apps a user may connect) — distinct from
+    ``list_skills_for_user``, which is the skills-endpoint listing and excludes
+    external-app-backed skills. The shared visibility predicate is
+    ``_add_user_visibility_filter``."""
     stmt = _add_user_visibility_filter(
         select(Skill).where(Skill.enabled.is_(True)), user
-    )
-    return {skill.id for skill in db_session.scalars(stmt)}
+    ).with_only_columns(Skill.id)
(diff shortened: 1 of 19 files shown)


Danelegend:
@greptile

Danelegend:
@greptile

rohoswagger (approved):

backend/onyx/db/skill.py
```diff
@@ -113,6 +113,15 @@ def _add_user_visibility_filter(
     )
 
 
+def visible_skill_ids_for_user(user: User, db_session: Session) -> set[UUID]:
```
wenxi-onyx, line 116: I believe this is redundant with `list_skills_for_user` which applies the filter already

backend/onyx/server/features/build/sandbox/util/agent_instructions.py
```diff
@@ -90,6 +91,31 @@ def build_skills_section_from_data(skills: Iterable[Skill]) -> str:
...
+    """Render the "Connectable apps" block: org apps the user hasn't set up yet.
+
+    These aren't usable until connected, so the agent can't call them directly —
+    but knowing they exist lets it plan around them and offer to connect one via
+    the ``connect_app`` tool. Empty input renders nothing (no heading)."""
+    entries = sorted((app.skill.slug, _truncate(app.skill.description)) for app in apps)
+    if not entries:
+        return ""
+
+    lines = [
```
wenxi-onyx: This static content should be in the AGENTS.template.md file. The func should just find the relevant value (like `{CONNECDTABLE_APPS}`) to replace with the list

PR #12450, opened 2026-06-25
fix(reporting): stop usage CSV export dropping rows at the 500-row page boundary (ONX-2293)

rohoswagger:
### Root cause
The admin **usage-report CSV export** paginates chat sessions in
`fetch_chat_sessions_eagerly_by_time` (`backend/ee/onyx/db/query_history.py`)
using a keyset cursor that was **`ChatSession.time_created > initial_time`**,
ordered **only** by `time_created`. But `time_created` is **not unique** —
sessions created together (bulk Slack/connector imports, sub-second bursts)
share a timestamp.

Once the dataset exceeds the **500-row page limit**, the next page's strict
`> time_created` filter **skips every session that shares the boundary page's
timestamp**, silently dropping the tail of the export. The sibling
`get_page_of_chat_sessions` already carried an `id` tiebreaker — this function
did not, which is the asymmetry that caused the bug.

A second, latent defect lived in `get_all_empty_chat_message_entries`
(`backend/ee/onyx/db/usage_export.py`): it stopped paginating as soon as a page
produced an empty *skeleton* list, so a full page of 500 sessions that happened
to contain no USER messages would end iteration early and drop everything after.

### Fix
- **Stable composite `(time_created, id)` keyset cursor** — order by
  `(time_created, id)` and resume with `time > t OR (time == t AND id > id)`, so
  ties at the page boundary are no longer skipped.
- **Terminate on a short page** (`len(sessions) < limit`) instead of on an empty
  skeleton list, so a full page with no USER messages keeps paginating.
- The first page now starts with no cursor (inclusive window start), fixing an
  off-by-one that dropped a session created exactly at the window start.

### Verification
- New regression test `test_pagination_does_not_drop_sessions_sharing_boundary_timestamp`:
  501 sessions sharing one `time_created` (one past the page limit) — fails
  before the fix, passes after; asserts no drops and no duplicates.
- New test `test_full_page_with_no_user_messages_keeps_paginating` locks the
  short-page termination contract (red on the old empty-skeleton logic).
- Full `test_usage_export_pairing.py` green (5 passed); `ruff` and `ty` clean.
- Existing integration test `test_usage_reports.py` already exercises 2048
  sessions across 90 days, validating the new keyset path end-to-end.

### Known minor items (not fixed here)
- `query_history.py`: the `initial_id is None` `else` branch is an unreachable
  defensive default in the only call path — harmless, kept for safety.
- `query_history.py:162` gates on `if initial_time:` (truthiness) rather than
  `is not None`; pre-existing and harmless (a set `datetime` is always truthy).

---
## Summary by cubic
Fixes dropped rows in the usage report `chat_messages.csv` when pagination crosses the 500-row boundary by switching to a stable composite cursor, validating it, and correcting the pagination stop condition. Addresses ONX-2293.

- **Bug Fixes**
  - Use a composite `(time_created, id)` keyset cursor and `ORDER BY time_created, id` in `fetch_chat_sessions_eagerly_by_time` to prevent skips on tied timestamps.
  - Require both `initial_time` and `initial_id`; raise if only one is provided to avoid falling back to a time-only cursor.
  - Thread the composite cursor through `get_empty_chat_messages_entries__paginated` and `get_all_empty_chat_message_entries`.
  - Stop paginating on a short page (`len(page) < limit`), not on an empty USER-message batch.
  - Added regression tests: 501 sessions sharing one timestamp (no drops) and a full page with no USER messages (pagination continues).

<sup>Written for commit fc44357172df48a921a9b3116f8601f0bbe068c0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12450?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/ee/onyx/db/usage_export.py
+++ b/backend/ee/onyx/db/usage_export.py
@@ -25,21 +25,24 @@ def get_empty_chat_messages_entries__paginated(
     db_session: Session,
     period: tuple[datetime, datetime],
     limit: int | None = 500,
-    initial_time: datetime | None = None,
-) -> tuple[Optional[datetime], list[ChatMessageSkeleton]]:
+    initial_cursor: tuple[datetime, uuid.UUID] | None = None,
+) -> tuple[Optional[tuple[datetime, uuid.UUID]], list[ChatMessageSkeleton]]:
     """Returns a tuple where:
-    first element is the most recent timestamp out of the sessions iterated
-    - this timestamp can be used to paginate forward in time
+    first element is the ``(time_created, id)`` cursor of the last session
+    iterated, or ``None`` once the final page has been reached - pass it back in
+    as ``initial_cursor`` to fetch the next page
     second element is a list of messages belonging to all the sessions iterated
 
     Only messages of type USER are returned
     """
+    initial_time, initial_id = initial_cursor if initial_cursor else (None, None)
     chat_sessions = fetch_chat_sessions_eagerly_by_time(
         start=period[0],
         end=period[1],
         db_session=db_session,
         limit=limit,
         initial_time=initial_time,
+        initial_id=initial_id,
     )
 
     message_skeletons: list[ChatMessageSkeleton] = []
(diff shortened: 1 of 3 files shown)


backend/ee/onyx/db/usage_export.py
```diff
@@ -25,21 +25,24 @@ def get_empty_chat_messages_entries__paginated(
...
-    first element is the most recent timestamp out of the sessions iterated
-    - this timestamp can be used to paginate forward in time
+    first element is the ``(time_created, id)`` cursor of the last session
+    iterated, or ``None`` once the final page has been reached - pass it back in
+    as ``initial_cursor`` to fetch the next page
     second element is a list of messages belonging to all the sessions iterated
 
     Only messages of type USER are returned
     """
+    initial_time, initial_id = initial_cursor if initial_cursor else (None, None)
```
evan-onyx, line 38: `initial_cursor or (None, None)`

evan-onyx (approved):

PR #12495, opened 2026-06-26
feat(craft): image generation via admin-configured provider + onyx-cli

rohoswagger:
## Description

Reworks the built-in **image-generation** Craft skill to generate via the **admin-configured default image-gen provider** (OpenAI / Gemini / Azure, from `/admin/configuration/image-generation`) instead of requiring a `GEMINI_API_KEY` in the sandbox (which was never injected, so the old skill was effectively inert).

- **New endpoint** `POST /image-generation/generate` — gated on the new `GENERATE_IMAGE` permission (implied by `BASIC_ACCESS` and `CRAFT_SANDBOX`, so it's reachable by authenticated users and the sandbox PAT alike) and mounted standalone under `/api`. Runs the existing server-side pipeline so **no provider keys ever enter the sandbox**, and returns base64 image(s).
- **New shared helper** `onyx/image_gen/generation.py` (`generate_images_with_default_config`, `resolve_image_size`). Resolves provider/model/credentials in a **short-lived DB session that's released before the slow provider call** (avoids pinning a pooled connection for the whole generation). The chat `ImageGenerationTool` now shares `resolve_image_size`.
- **`ImageShape`** moved into `onyx/image_gen/interfaces.py` (re-exported from the tool's `models.py` for back-compat).
- **Skill is now script-less** — `SKILL.md` documents `onyx-cli image` and deletes the old `scripts/generate.py` (google-genai). The skill is gated on `is_image_generation_configured`, so it only appears when a default provider is set up.

This is **3 of 3** PRs splitting the rework: 1 = onyx-cli command (#12493), 2 = sandbox image deps (#12494), 3 = this. The skill references `onyx-cli image`, so it's fully functional in sandboxes once the CLI is released and the sandbox image rebuilt; the endpoint itself works as soon as this deploys.

## How Has This Been Tested?

`ruff check` + `ruff format`, `ty`, and the full unit suite pass. Unit tests cover the helper (size mapping, not-configured/invalid-creds, reference-unsupported, b64 extraction, multi-image), the endpoint (error→`OnyxError` code mapping, mime detection incl. jpeg + fallback, reference forwarding, input validation), the `GENERATE_IMAGE` scope path through `require_permission`, malformed-credential handling, and a regression guard that the refactored chat tool still maps shape→size correctly. Also reviewed by a parallel multi-lens review (correctness/security/tests/regressions) and validated end-to-end against a local kind cluster: an agent in a real sandbox ran `onyx-cli image generate` and produced a valid PNG via this endpoint.

No new integration/Playwright tests — image generation makes real provider calls (the existing chat image-gen suite deliberately skips the real-call path for the same reason).

## Known minor items (for reviewer)

These were surfaced during review and intentionally **not** fixed here to keep the diff focused — flagging so they're a conscious choice:

- **Per-provider `max_reference_images` not enforced (medium).** The endpoint caps reference images at a hardcoded 16; it doesn't check `provider.max_reference_images` (Vertex 14, dall-e-2 1). Over-limit requests still validate locally and fail downstream as an opaque provider error mapped to `LLM_PROVIDER_ERROR` (500) rather than a clean `INVALID_INPUT` (400).
- **No request-body byte cap on base64 reference images (low).** Reference images are count-capped (16) but not size-capped; large payloads are decoded into memory. Upstream (nginx/FastAPI) body limits mitigate, but a per-image byte cap would be cleaner.
- **Model family is inferred from the model name string (`"gpt-image-" in model`) for both `size` and `response_format` (low/medium, pre-existing).** Carried over verbatim from the chat `ImageGenerationTool` (extracted into `resolve_image_size`), so the chat path behaves identically. Two consequences: dall-e-2 only supports square sizes (LANDSCAPE/PORTRAIT would error), and a **custom-named Azure gpt-image deployment** (e.g. `prod-image`) is misclassified as non-gpt-image, so it gets DALL-E sizes and an unsupported `response_format`. There's no reliable signal to detect the model family from a custom Azure deployment alias without adding an explicit model-family field to the image-gen provider config — a separate change that should touch both call sites and the admin config, so it's out of scope for this skill rework.

## Additional Options

---
## Summary by cubic
Adds a server-side image generation API that uses the admin-configured default provider, so any authenticated user or sandbox can generate images via `onyx-cli` without provider keys. The built-in skill only appears when a default provider is configured.

- **New Features**
  - Added POST /image-generation/generate, gated by GENERATE_IMAGE (implied by BASIC_ACCESS and CRAFT_SANDBOX). Returns base64 image(s); usable from `onyx-cli image`.
  - Supports shape, n (up to 4), quality, and reference images (up to 16). Auto-detects image MIME and maps errors to OnyxError codes with safe logging.
  - Uses the workspace’s default provider/model from /admin/configuration/image-generation (OpenAI/Gemini/Azure); clean “not configured” errors. Shared is_image_generation_configured hides the skill unless set up.

- **Refactors**
  - Added onyx/image_gen/generation.py with generate_images_with_default_config, generate_images_with_provider, response_format_for_model, and resolve_image_size. Endpoint and chat tool share this core.
  - Chat ImageGenerationTool reuses the shared size resolver and availability gate; no longer returns the raw provider response.
  - Moved ImageShape to onyx/image_gen/interfaces.py; removed the re-export and updated imports. Dropped the old script; SKILL.md now documents `onyx-cli image`.
  - Hardened credential validation (e.g., Vertex JSON) and tightened error disclosure. Updated permission specs/tests for GENERATE_IMAGE.

<sup>Written for commit 408f42a8731c084c975a16736325f2a95deff9f6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12495?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/onyx/server/features/build/image_generation/api.py
```diff
@@ -0,0 +1,129 @@
...
+from onyx.image_gen.generation import generate_images_with_default_config
+from onyx.image_gen.interfaces import ImageShape
+from onyx.image_gen.interfaces import ReferenceImage
+from onyx.utils.b64 import get_image_type
+from onyx.utils.logger import setup_logger
+
+logger = setup_logger()
+
+# Called by the sandbox itself (onyx-cli), not the frontend, so this is mounted
+# standalone rather than under the BASIC_ACCESS-gated /build router and is gated
```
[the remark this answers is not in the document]
rohoswagger: Outdated + not an issue. This endpoint was moved out of `/build` in the latest revision — it is now `POST /image-generation/generate` (generic, gated on the new `GENERATE_IMAGE` scope), so the `/build/image-generation` path this comment anchors to no longer exists. The routing concern also does not apply: `onyx-cli`'s `Client.newRequest` builds `config.APIURL(serverURL) + path`, which prepends `/api`, so requests hit `/api/image-generation/generate` — caught by the existing `^/(api|openapi\.json)` nginx block exactly like `/api/search` and `/api/persona`. No nginx location needed.

backend/onyx/image_gen/generation.py
```diff
@@ -0,0 +1,127 @@
...
+    provider = get_image_generation_provider(provider_name, credentials)
+
+    if reference_images and not provider.supports_reference_images:
+        raise ValueError(
+            f"Provider '{provider_name}' does not support image edits / reference images."
+        )
+
+    size = resolve_image_size(model, shape)
+    logger.debug(
+        "Generating %d image(s) with provider=%s model=%s size=%s",
```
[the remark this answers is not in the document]
rohoswagger: Not an issue — the `LLMFlow` span is opened one layer down, in the provider. `generate_images_with_default_config` calls `provider.generate_image(...)`, and every image provider wraps that call in `with traced_llm_call(flow=LLMFlow.IMAGE_GENERATION | IMAGE_EDIT)` (see `openai_img_gen.py`, `azure_img_gen.py`, `vertex_img_gen.py`, including the reference-image/edit paths). So these generations carry the correct flow tag; the helper deliberately does not double-wrap. This mirrors the existing chat `ImageGenerationTool`, which also calls `provider.generate_image` directly and relies on the provider span.

rohoswagger:
@greptile-ai @codex review this

backend/onyx/server/features/image_generation/api.py
```diff
@@ -0,0 +1,125 @@
...
+            quality=request.quality,
+            reference_images=reference_images or None,
+        )
+    except ImageGenerationNotConfiguredError as e:
+        raise OnyxError(OnyxErrorCode.NOT_FOUND, str(e))
+    except ValueError as e:
+        raise OnyxError(OnyxErrorCode.INVALID_INPUT, str(e))
+    except Exception as e:
+        logger.exception("Image generation failed")
+        raise OnyxError(OnyxErrorCode.LLM_PROVIDER_ERROR, str(e))
```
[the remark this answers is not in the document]
rohoswagger: Fixed in 852c3530ab. The catch-all now logs the full exception (`logger.exception`) and returns a generic `OnyxError(LLM_PROVIDER_ERROR, "Image generation failed.")` instead of the raw `str(e)`, so LiteLLM/provider error text — which can embed `api_base` URLs or auth tokens — no longer reaches the caller. I also dropped the credentials object from the `ImageProviderCredentialsError` message in `interfaces.py` (`build_from_credentials`) to close the related key-leak path, and added a unit test asserting malformed Vertex creds surface a clean error rather than a raw `JSONDecodeError`.

The `ValueError` branch still surfaces its message, but those are author-controlled validation strings (empty prompt, `n` out of range, unsupported reference images) — never provider/LiteLLM output, which is a non-`ValueError` exception caught by the sanitized catch-all.

backend/onyx/image_gen/generation.py
```diff
@@ -0,0 +1,127 @@
...
+    b64_data: str
+    revised_prompt: str
+
+
+def resolve_image_size(model: str, shape: ImageShape) -> str:
+    is_gpt_image = _GPT_IMAGE_PREFIX in model
+    if shape == ImageShape.LANDSCAPE:
+        return "1536x1024" if is_gpt_image else "1792x1024"
+    if shape == ImageShape.PORTRAIT:
+        return "1024x1536" if is_gpt_image else "1024x1792"
```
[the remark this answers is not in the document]
rohoswagger: Valid concern, but leaving it as-is in this PR — it's pre-existing behavior, not introduced here. The `"gpt-image-" in model` inference for both `size` and `response_format` was carried over verbatim from the chat `ImageGenerationTool` (origin/main lines 173/179/193); this PR only extracted it into the shared `resolve_image_size` helper, so the chat path has identical behavior today.

You're right that a custom-named Azure gpt-image deployment (e.g. `prod-image`) gets misclassified as non-gpt-image → DALL-E sizes (`1792x1024`/`1024x1792`) and an unsupported `response_format`. Fixing it correctly requires an explicit model-family signal, since a custom Azure deployment alias carries no reliable hint of the underlying model — and that change should touch both call sites and the admin image-gen config rather than this skill rework. Tracked as a known item in the PR description so it's a conscious deferral, not an oversight.

backend/onyx/server/features/image_generation/__init__.py
Danelegend, line 1: __init__ bad

backend/onyx/tools/tool_implementations/images/models.py
```diff
@@ -1,20 +1,19 @@
-from enum import Enum
-
 from pydantic import BaseModel
 
+from onyx.image_gen.interfaces import ImageShape
 from onyx.server.query_and_chat.streaming_models import GeneratedImage
 
+__all__ = [
```
Danelegend: why this?

backend/onyx/image_gen/generation.py
```diff
@@ -0,0 +1,137 @@
...
+        custom_config=llm_provider.custom_config,
+    )
+    if not validate_credentials(llm_provider.provider, credentials):
+        raise ImageGenerationNotConfiguredError(
+            "The configured image generation provider has invalid credentials."
+        )
+    return llm_provider.provider, config.model_configuration.name, credentials
+
+
+def is_image_generation_configured(db_session: Session) -> bool:
```
Danelegend, line 112: Can ImageGenerationTool::is_available also use this?

Danelegend (approved):

backend/onyx/server/features/image_generation/__init__.py
rohoswagger: Removed it in d85717e511. The sibling feature packages (`build/`, `skill/`, `mcp/`, `default_assistant/`, ...) ship no `__init__.py` and resolve fine as namespace packages, and `main.py` imports `...image_generation.api` the same way. Confirmed the import still resolves with the file gone.

backend/onyx/tools/tool_implementations/images/models.py
rohoswagger: Removed in 7ea1c80248. The `__all__` only existed to mark the `ImageShape` re-export (added when `ImageShape` moved to `onyx.image_gen.interfaces`) as intentional so ruff wouldn't flag it as unused — there are no `import *` of this module, so it did nothing else. Dropped both the re-export and `__all__`, and changed the single in-repo importer (`image_generation_tool.py`) to import `ImageShape` straight from `onyx.image_gen.interfaces`.

backend/onyx/image_gen/generation.py
rohoswagger: Good call — done in 7ea1c80248. `ImageGenerationTool.is_available` now just `return is_image_generation_configured(db_session)` instead of re-deriving the default-config + credential check. Both it and `_default_provider_and_model` were doing the same work, so the chat tool, the skill availability gate, and the endpoint now share one source of truth. Dropped the now-unused `get_default_image_generation_config` / `validate_credentials` imports from the tool; existing tests (`test_image_generation_reference_resolution.py`, image-gen unit suite) still pass.

rohoswagger:
@greptile-ai @codex review this

backend/onyx/image_gen/generation.py
```diff
@@ -0,0 +1,156 @@
...
+        return "1536x1024" if is_gpt_image else "1792x1024"
+    if shape == ImageShape.PORTRAIT:
+        return "1024x1536" if is_gpt_image else "1024x1792"
+    return "1024x1024"
+
+
+def response_format_for_model(model: str) -> str | None:
+    """gpt-image-* models reject the ``response_format`` param; every other
+    model needs ``b64_json`` to return base64 inline."""
+    return None if _GPT_IMAGE_PREFIX in model else "b64_json"
```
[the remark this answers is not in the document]
rohoswagger: You're correct about the failure mode, but I'm leaving this out of scope for this PR — it's pre-existing and there's no in-scope fix. Details:

**There is no config metadata to switch on.** I checked: `AzureImageGenForm.tsx` sets `modelName = deploymentName || imageProvider.model_name`, so for Azure the stored `model_configuration.name` is the deployment alias, and `ModelConfigurationUpsertRequest` / the model-configuration schema carry no base-model or model-family field. The model-name string is the only signal that exists today, which is exactly why a custom alias like `prod-image` can't be classified.

**It's pre-existing, not introduced here.** `"gpt-image-" in model` is a codebase-wide heuristic: the chat `ImageGenerationTool` has used it for both size and `response_format` on `main` (image_generation_tool.py:173/178/193), and `_get_test_quality_for_model` in the manage API uses it too. This PR only extracted that existing logic into `response_format_for_model` / `resolve_image_size` (one source of truth) — the behavior is unchanged from `main`, so this isn't a regression in this PR.

**A correct fix is cross-cutting.** Detecting GPT-image reliably for Azure aliases means capturing the model family in the admin config: a new field on the Azure form (`AzureImageGenForm.tsx`), `ModelConfigurationUpsertRequest` + the `ModelConfiguration` table (a migration), and then having *all three* call sites (this endpoint, the chat tool, and `_get_test_quality_for_model`) read it. That belongs in a focused follow-up in the admin-image-gen domain, not bundled into this skill/endpoint PR.

I've documented this as a known item in the PR description. Happy to take the schema/form change as a separate PR if we want to close the Azure-alias gap properly.

PR #12517, opened 2026-06-29
feat(file-processing): isolate PDF text extraction in a subprocess

nmgarza5:
## Description

**Bug:** A malformed PDF can make PDFium's C++ parser hard-abort (SIGABRT) or segfault deep inside `libpdfium` — confirmed from a core dump. Because that happens in native code reached through ctypes, it's uncatchable in-process: the existing `except PdfiumError` never sees it, so the whole indexing worker dies and fails the attempt (and every other document in the batch).

**Fix:** Add a reusable `onyx/utils/process_isolation.py` helper that runs a callable in a spawned subprocess and turns a signal kill, timeout, or runaway allocation into typed exceptions, re-raising ordinary exceptions unchanged. PDFium text extraction routes through it, so a crash or hang falls back to pypdf exactly like a `PdfiumError` — one bad file is skipped instead of killing the worker. Set `PDF_TEXT_EXTRACTION_TIMEOUT_SECONDS=0` to disable.

**Tests:** Lib unit tests (crash, segfault, timeout, large result, memory cap) and the PDF crash/error → pypdf fallback path.

## Additional Options

### DIFF
--- a/backend/onyx/file_processing/extract_file_text.py
+++ b/backend/onyx/file_processing/extract_file_text.py
@@ -27,6 +27,7 @@ from PIL import Image
 
 from onyx.configs.app_configs import MAX_EMBEDDED_IMAGES_PER_FILE
 from onyx.configs.app_configs import MAX_XLSX_CELLS_PER_SHEET
+from onyx.configs.app_configs import PDF_TEXT_EXTRACTION_TIMEOUT_SECONDS
 from onyx.configs.app_configs import XLSX_STREAM_SHEET_BYTES
 from onyx.configs.constants import ONYX_METADATA_FILENAME
 from onyx.configs.llm_configs import get_image_extraction_and_analysis_enabled
@@ -38,6 +39,8 @@ from onyx.file_processing.html_utils import parse_html_page_basic
 from onyx.file_processing.unstructured import get_unstructured_api_key
 from onyx.file_processing.unstructured import unstructured_to_text
 from onyx.utils.logger import setup_logger
+from onyx.utils.process_isolation import IsolatedProcessError
+from onyx.utils.process_isolation import run_in_isolated_process
 
 if TYPE_CHECKING:
     from markitdown import MarkItDown
@@ -352,12 +355,16 @@ def read_pdf_file(
                 ):
                     metadata[clean_key] = ", ".join(value)
 
-        # GIL-releasing extraction so a large PDF can't pin the worker — see helper.
+        # PDFium can hard-abort or hang on a malformed PDF (uncatchable in-process),
+        # so run it isolated; a crash, timeout, or PdfiumError falls back to pypdf.
         try:
-            text = _extract_pdf_text_pdfium(file_bytes, decrypt_password)
-        except PdfiumError as pdfium_err:
-            # PDFium rejects some malformed PDFs that pypdf can still read; fall
-            # back so those docs aren't dropped. Only failing files take this path.
+            text = run_in_isolated_process(
+                _extract_pdf_text_pdfium,
+                file_bytes,
+                decrypt_password,
+                timeout=PDF_TEXT_EXTRACTION_TIMEOUT_SECONDS,
+            )
+        except (PdfiumError, IsolatedProcessError) as pdfium_err:
             logger.warning(
                 "PDFium text extraction failed (%s); falling back to pypdf",
                 pdfium_err,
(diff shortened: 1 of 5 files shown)


backend/onyx/file_processing/extract_file_text.py
```diff
@@ -352,12 +355,17 @@ def read_pdf_file(
...
-        # GIL-releasing extraction so a large PDF can't pin the worker — see helper.
+        # PDFium's C++ parser can hard-abort (SIGABRT/SIGSEGV) or hang on a
+        # malformed PDF — uncatchable in-process — so it runs isolated; a crash,
+        # timeout, or PdfiumError all fall back to pypdf rather than drop the doc.
         try:
-            text = _extract_pdf_text_pdfium(file_bytes, decrypt_password)
-        except PdfiumError as pdfium_err:
-            # PDFium rejects some malformed PDFs that pypdf can still read; fall
-            # back so those docs aren't dropped. Only failing files take this path.
+            text = run_in_isolated_process(
```
evan-onyx, line 361: nit: for these cases where the function-calling function isn't called in a loop I prefer wrapping in a `lambda: ` to preserve type checking

backend/onyx/utils/process_isolation.py
```diff
@@ -0,0 +1,162 @@
+"""Run a callable in a disposable subprocess so a native crash (SIGSEGV/SIGABRT
+from a C/C++ extension on malformed input), an unbounded hang, or a runaway
+allocation can't take down the caller. The child absorbs the failure; the parent
+raises a typed error and the caller decides how to recover (skip, fall back).
+
+Intended for extraction/parse paths that hand untrusted bytes to native
+libraries (PDF/image/office parsing). Each call spawns a process — a guard for
```
evan-onyx: this could be more concise, the last sentence gave me a visceral LLM cringe reaction xD

backend/onyx/utils/process_isolation.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def __init__(self, timeout_seconds: float) -> None:
+        self.timeout_seconds = timeout_seconds
+        super().__init__(f"isolated process exceeded {timeout_seconds:g}s timeout")
+
+
+def _apply_memory_limit(memory_limit_bytes: int) -> None:
+    # Best effort: caps address space so a runaway allocation fails inside the
+    # child instead of letting the host OOM-kill the pod. Not enforced on every
+    # platform (e.g. macOS treats RLIMIT_AS loosely).
```
evan-onyx: it's time to cross our fingers

backend/onyx/utils/process_isolation.py
```diff
@@ -0,0 +1,162 @@
...
+    try:
+        import resource
+
+        resource.setrlimit(resource.RLIMIT_AS, (memory_limit_bytes, memory_limit_bytes))
+    except (ImportError, AttributeError, ValueError, OSError):
```
evan-onyx: probably should have a debug log (at least) here, I'm more inclined towards info or warning tbh

backend/onyx/utils/process_isolation.py
```diff
@@ -0,0 +1,162 @@
...
+    kwargs: dict[str, Any],
+    memory_limit_bytes: int | None,
+    result_path: str,
+) -> None:
+    if memory_limit_bytes is not None:
+        _apply_memory_limit(memory_limit_bytes)
+
+    try:
+        payload: tuple[str, Any] = (_STATUS_OK, fn(*args, **kwargs))
+    except BaseException as e:  # noqa: BLE001 - relay every failure mode to the parent
```
evan-onyx: I think this is a bit nonstandard, is the exception from the function in segfaults and sigaborts not an Exception subclass?

backend/onyx/utils/process_isolation.py
```diff
@@ -0,0 +1,162 @@
...
+            raise IsolatedProcessTimeout(timeout)
+
+        # A negative exit code means the child was killed by signal -exitcode.
+        if proc.exitcode is not None and proc.exitcode < 0:
+            raise IsolatedProcessCrashed(-proc.exitcode)
+
+        try:
+            with open(result_path, "rb") as f:
+                # Trusted input: this payload was written by our own child below.
+                status, value = pickle.loads(f.read())  # noqa: S301
```
evan-onyx: i think if an attacker can write to the filesystem they can inject their own pickle file after our process writes it and before the read. It might be just something we have to live with, but probably worth checking if there's a canonical way to do this for security

evan-onyx (approved):

backend/onyx/file_processing/extract_file_text.py
nmgarza5: Lambdas and closures can't be pickled, and the forkserver child receives the target by pickling it by reference, so run_in_isolated_process(lambda: _extract_pdf_text_pdfium(...)) would fail to start the subprocess. That's why it takes the function plus args. The args do lose static type checking across the process boundary, which is an inherent cost of the multiprocessing API; functools.partial has the same gap since ty types it as Callable[..., T].

PR #12529, opened 2026-06-29
feat(tracing): DB-backed tracing provider config with live env fallback (PR 1/3)

justin-tahara:
## Description

First of three PRs adding **UI-configurable tracing providers** (Braintrust, Langfuse). Today these are configured **only via environment variables**, read once at process startup. This PR lays the backend foundation so they can later be managed from an admin page, while staying **fully backwards compatible**.

This PR is intentionally a **no-op for existing env-configured deployments** — there is no API or UI yet (those land in PRs 2 and 3).

**What's included:**
- **`TracingProviderConfig` model + migration** — one row per provider, each enabled independently (both can be on at once). API keys stored in an `EncryptedString` column; non-secret settings (Braintrust `project`; Langfuse `public_key`, `host`) in a JSONB `config` column. Mirrors the existing `InternetSearchProvider` (web-search) precedent.
- **`db/tracing.py`** — `fetch`/`upsert` (with `api_key_changed` so the key isn't rewritten on edit)/`delete` helpers.
- **`resolve_effective_tracing_config()`** — per provider, a DB row is authoritative (enabled → on, disabled → off); with no row, falls back to the existing env vars. In `MULTI_TENANT` (cloud) it consults **env vars only** — the UI feature is unsupported there.
- **`DynamicTracingProcessor`** — registered once at startup; resolves the effective config on a short TTL (`TRACING_CONFIG_CACHE_TTL_SECONDS`, default 30s) and rebuilds the Braintrust/Langfuse delegates when it changes, so a future connect/disconnect takes effect **without a restart** across every process that emits traces (api_server, all Celery workers, MCP server, Slack bot, eval CLI). The delegate set is **pinned per-trace**, so a config change never disrupts an in-flight trace.
- **`setup_tracing()`** refactored to register the single dynamic processor; its signature is unchanged, so all five call sites are untouched.

`onyx/evals/` continues to read `BRAINTRUST_*` / `LANGFUSE_*` env vars directly — that's a separate concern and is deliberately left alone.

## How Has This Been Tested?

- **Unit** (`tests/unit/onyx/tracing/`): `DynamicTracingProcessor` rebuilds delegates on config change, no-ops when no provider is enabled, honors the TTL throttle, and keeps in-flight traces on their original (pinned) delegates while retiring/draining replaced ones.
- **External-dependency-unit** (`tests/external_dependency_unit/tracing/`): real Postgres-backed encrypt/decrypt round-trip + masking, DB-wins-over-env precedence, env fallback when no row exists, disabled-row overrides env, and `api_key_changed=false` preserves the stored key.
- All 22 tracing tests pass; `ruff`, `ruff format`, and `ty` (strict typing) clean. Migration applies and downgrades cleanly; verified the full DB round-trip (resolve from DB, masking, env fallback after delete) against a live database.

## Additional Options

---
## Summary by cubic
Adds a DB-backed tracing provider config for `braintrust` and `langfuse` with env fallback and live reload across all processes. Supports non-blocking config swaps, fail-open refresh at trace start, and automatic env sync to avoid stale `LANGFUSE_HOST`/`BRAINTRUST_API_URL`; env-only setups behave the same. This is the backend for a future admin UI.

- **New Features**
  - `TracingProviderConfig` table with encrypted `api_key` and JSONB `config`; one row per provider, enabled independently.
  - Effective config resolution: DB row wins; if none, fall back to env. In `MULTI_TENANT`, env-only.
  - `DynamicTracingProcessor`: TTL-based reads via `TRACING_CONFIG_CACHE_TTL_SECONDS` (default 30s), rebuilds delegates without restart, pins per-trace, builds/shuts down off-lock, retires old sets after in-flight traces drain, guards refresh at trace start (fail-open), and syncs/clears `LANGFUSE_HOST` and `BRAINTRUST_API_URL`.
  - `setup_tracing()` registers this single dynamic processor; signature unchanged and applies to all emitters (API server, Celery, MCP, Slack bot, eval CLI).

- **Migration**
  - Apply the migration creating `tracing_provider_config`; no config changes required.

<sup>Written for commit b01337c431a4643cae56842a26421eec37e98f8a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12529?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/onyx/tracing/dynamic_processor.py
```diff
@@ -0,0 +1,193 @@
...
+    for processor in delegates:
+        try:
+            getattr(processor, method)(*args)
+        except Exception as e:
+            logger.error(
+                "Error in trace processor %s during %s: %s", processor, method, e
+            )
+
+
+class DynamicTracingProcessor(TracingProcessor):
```
nmgarza5, line 102: not seeing much error handling in this class. for example, what if connection to the tracing provider is interrupted mid trace?

nmgarza5 (approved):
lgtm but one comment

backend/onyx/tracing/dynamic_processor.py
justin-tahara: Good call. Tracing is fail-open by design: every delegate callback (on_span_*/on_trace_end) is wrapped per-call in `_forward`, so a provider connection dropping mid-trace is caught + logged and never propagates into the traced operation. I also added a guard around the config `reconcile()` at `on_trace_start` (bd1f941) so a config-refresh failure (e.g. DB unreachable) is logged and we keep the current delegates rather than disrupting the trace.

PR #12535, opened 2026-06-29
feat(tracing): admin API for tracing providers (PR 2/3, stacked on #12529)

justin-tahara:
## Description

**PR 2 of 3** for UI-configurable tracing providers. **Stacked on #12529** (base branch is `jtahara/tracing-ui-config-backend`) — review that first; this PR's diff shows only the API layer.

Adds the admin API at **`/admin/tracing`** that drives the PR 1 data model. Admin-gated (`FULL_ADMIN_PANEL_ACCESS`) and rejected under `MULTI_TENANT` (single-tenant only), per the agreed scope:

- `GET /providers` — per-provider status: `connected`, `source` (`db` | `env` | `none`), `enabled`, non-secret `config`, and `masked_api_key` (plaintext key never returned).
- `POST /providers` — connect/update; honors `api_key_changed` so the stored key isn't overwritten on edit.
- `DELETE /providers/{provider_type}` — disconnect (reverts to env fallback if one exists).
- `POST /providers/test` — validates credentials against the provider (Braintrust login / Langfuse `auth_check`) before the admin commits.
- `POST /providers/{provider_type}/adopt-env` — the migration action: copies env-var config into a DB row so it can be managed in the UI.

Uses `OnyxError` + typed returns (no `HTTPException` / `response_model`) per repo conventions. Router registered in `main.py`.

## How Has This Been Tested?

External-dependency-unit tests driving the handlers in-process against real Postgres (`tests/external_dependency_unit/tracing/test_tracing_admin_api.py`): connect → list shows `db`/masked, disconnect removes the row, `/test` rejects a missing key, `adopt-env` copies env→DB (and errors when no env config exists), and the `MULTI_TENANT` gate raises. Full suite (PR 1 resolution + PR 2 API) = 27 tests green; `ruff`/`ruff format`/`ty` clean.

Note: a full HTTP/auth integration test will run in CI once this branch is deployed — it can't be exercised against the shared local api_server, which doesn't run this branch's code. The handler logic, DB round-trip, masking, and gating are covered in-process.

## Additional Options

---
## Summary by cubic
Adds an admin-only API at `/admin/tracing` to manage tracing providers (Braintrust, Langfuse): list status, connect/update, disconnect, test credentials, and adopt env-based configs. Requires `FULL_ADMIN_PANEL_ACCESS`; blocked in `MULTI_TENANT`.

- **New Features**
  - `GET /admin/tracing/providers` — returns per-provider status: `connected`, `source` (`db` | `env` | `none`), `enabled`, non-secret `config`, `masked_api_key`.
  - `POST /admin/tracing/providers` — upserts provider; honors `api_key_changed` so stored keys aren’t overwritten on edit.
  - `DELETE /admin/tracing/providers/{provider_type}` — disconnects; falls back to env config if present.
  - `POST /admin/tracing/providers/test` — validates credentials via direct HTTP calls; Braintrust validation honors `api_url` in `config`.
  - `POST /admin/tracing/providers/{provider_type}/adopt-env` — copies env var config into a DB row for UI management.
  - Admin-gated and single-tenant only; uses `OnyxError` with typed returns; router registered in `main.py`.

- **Migration**
  - For env-based configs, call `POST /admin/tracing/providers/{provider_type}/adopt-env` to migrate into the DB.
  - No migration on multi-tenant deployments (request is rejected).

<sup>Written for commit 663c601c809be5f9a82768864779c0d84c546e2a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12535?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/server/manage/tracing/api.py
+++ b/backend/onyx/server/manage/tracing/api.py
@@ -61,11 +61,12 @@ def _build_view(
     row: TracingProviderConfig | None,
     effective: EffectiveTracingConfig,
 ) -> TracingProviderView:
-    is_connected = (
-        effective.braintrust is not None
-        if provider_type == TracingProviderType.BRAINTRUST
-        else effective.langfuse is not None
-    )
+    if provider_type == TracingProviderType.BRAINTRUST:
+        is_connected = effective.braintrust is not None
+    elif provider_type == TracingProviderType.LANGFUSE:
+        is_connected = effective.langfuse is not None
+    else:
+        is_connected = False
     if row is not None:
         return TracingProviderView(
             provider_type=provider_type,
--- a/backend/onyx/server/manage/tracing/api.py
+++ b/backend/onyx/server/manage/tracing/api.py
@@ -1,0 +1,224 @@
+from __future__ import annotations
+
+from fastapi import APIRouter
+from fastapi import Depends
+from sqlalchemy.orm import Session
+
+from onyx.auth.permissions import require_permission
+from onyx.db.engine.sql_engine import get_session
+from onyx.db.enums import Permission
+from onyx.db.models import TracingProviderConfig
+from onyx.db.models import User
+from onyx.db.tracing import delete_tracing_provider
+from onyx.db.tracing import fetch_tracing_provider
+from onyx.db.tracing import upsert_tracing_provider
+from onyx.error_handling.error_codes import OnyxErrorCode
+from onyx.error_handling.exceptions import OnyxError
+from onyx.server.manage.tracing.models import TracingProviderTestRequest
+from onyx.server.manage.tracing.models import TracingProviderUpsertRequest
+from onyx.server.manage.tracing.models import TracingProviderView
+from onyx.tracing.provider_config import BraintrustConfig
+from onyx.tracing.provider_config import EffectiveTracingConfig
+from onyx.tracing.provider_config import LangfuseConfig
+from onyx.tracing.provider_config import resolve_effective_tracing_config
+from onyx.tracing.validation import validate_tracing_credentials
+from onyx.utils.logger import setup_logger
+from shared_configs.configs import MULTI_TENANT
+from shared_configs.enums import TracingProviderType
+
+logger = setup_logger()
+
+
+def _reject_if_multi_tenant() -> None:
+    if MULTI_TENANT:
+        raise OnyxError(
+            OnyxErrorCode.SINGLE_TENANT_ONLY,
+            "Tracing provider configuration is not available on this deployment.",
+        )
+
+
+admin_router = APIRouter(
+    prefix="/admin/tracing", dependencies=[Depends(_reject_if_multi_tenant)]
+)
+
+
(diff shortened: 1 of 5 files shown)


backend/onyx/server/manage/tracing/api.py
```diff
@@ -0,0 +1,225 @@
...
+
+logger = setup_logger()
+
+
+def _reject_if_multi_tenant() -> None:
+    if MULTI_TENANT:
+        raise OnyxError(
+            OnyxErrorCode.SINGLE_TENANT_ONLY,
+            "Tracing provider configuration is not available on this deployment.",
+        )
```
nmgarza5, line 37: sorry why does this need to be single tenant only? if we had to live by env vars, i understand but wont that be removed? seems like multi tenant can be supported imo

nmgarza5: i guess we can just remove this logic after everything is in

backend/onyx/tracing/validation.py
```diff
@@ -0,0 +1,77 @@
...
+import requests
+
+from onyx.utils.logger import setup_logger
+from shared_configs.enums import TracingProviderType
+
+logger = setup_logger()
+
+_VALIDATION_TIMEOUT_S = 10
+_BRAINTRUST_API_URL = "https://api.braintrust.dev"
+_LANGFUSE_DEFAULT_HOST = "https://cloud.langfuse.com"
```
nmgarza5, line 20: just checking that there are not region based urls for these. do we need to allow users to specify their region? similar to [this](https://github.com/onyx-dot-app/onyx/pull/12264)

nmgarza5: looks like langfuse is handled but maybe check on braintrust

nmgarza5 (approved):

backend/onyx/server/manage/tracing/api.py
justin-tahara: Intentional for now: this is a product decision to not support tracing config on multi-tenant cloud yet (per the feature owner). Agreed it's removable once everyone is off env-var config — I left a note to revisit the gate then. Keeping it for this PR.

backend/onyx/tracing/validation.py
justin-tahara: Added Braintrust parity: a custom Braintrust API URL is now supported via a BRAINTRUST_API_URL env var + an optional `api_url` config value (self-hosted / non-default regions), used both by the SDK logger and this credential validation. Langfuse already had the base-URL field. (5503465, plus PR 1 for the config/model + PR 3 for the UI field.)

PR #11766, opened 2026-06-05
feat: per-user usage + cost tracking — schema + models (1/4)

beraterkanelcelik:
**Part 1 of 4** — splitting the per-user usage + cost tracking work (originally #11732) into smaller, reviewable, stacked PRs per maintainer feedback. This first PR is **schema only** — no logic.

### Adds
- **`user_usage`** — per-user usage rollup, one row per (user × model × window-day): input/output/cache tokens + cost.
- **`model_cost_override`** — admin-set per-model negotiated rates (input / output / optional cache-read).
- **`token_rate_limit.cost_budget_cents`** (+ `token_budget` becomes nullable) so a limit can be token-only, cost-only, or both.
- One consolidated additive migration (single head).

### The stack (each builds on the previous, opened as the prior merges)
1. **Schema + models** ← this PR
2. Per-user usage + cost accounting (recording processor, `GET /user/usage`, cost resolution, admin cost-overrides + export)
3. Cost budgets on token rate limits — CE/global enforcement + structured 429
4. Frontend (Usage tab, cost-overrides panel, rate-limit cost-budget field, 429 banner)

The per-user/group budget **enforcement** and the admin per-user usage **report** live under `ee/` and are being implemented separately by the Onyx team.

Supersedes #11732.

---
## Summary by cubic
Add schema and models for per-user usage and cost tracking, plus optional cost budgets on token rate limits. Schema-only; one additive migration with a safe downgrade.

- **New Features**
  - user_usage: per-user × model × flow × provider × window rollup with input/output/cache tokens and cost_cents; unique on (user_id, window_start, model, flow, provider); provider defaults to '' for cross-PG support.
  - model_cost_override: admin-set USD/mtok rates (input/output, optional cache-read) per provider+model (unique); provider defaults to '' for provider-agnostic overrides; used before `litellm` defaults; includes updated_at.
  - token_rate_limit: adds cost_budget_cents; token_budget is nullable; DB check constraint requires at least one of token or cost budget.

- **Migration**
  - On downgrade, delete cost-only token limits before restoring NOT NULL to avoid token_budget=0 rows that block all requests.
  - Dropped redundant user_usage user_id index; the unique dims index covers it.

<sup>Written for commit dc6de975544b67812e77ae21f9ba67a87210b4be. Summary will update on new commits.</sup>

[...]

PR #12550, opened 2026-06-30
feat: per-user usage + cost tracking — schema + models (1/4) [modifying #11766]

evan-onyx:
mirror of PR #11766 so I can make edits

---
## Summary by cubic
Add per-user LLM usage and cost tracking with cost-aware rate limits. Adds `user_usage` and `model_cost_override`, and supports token-only, cost-only, or combined budgets on `token_rate_limit`.

- **New Features**
  - `user_usage`: per-user rollup by window/model/flow/provider with tokens and `cost_cents` (unique on `(user_id, window_start, model, flow, provider)`; `provider` stored as `""` for reliable uniqueness).
  - `model_cost_override`: per-model rates with provider-specific pricing (unique on `(provider, model)`; `provider` uses `""` when unset), optional `cache_read_cost_per_mtok`.
  - Rate limits: `token_rate_limit` adds `cost_budget_cents`; `token_budget` is nullable; DB check requires at least one budget. Pydantic models enforce this, API display includes both budgets, and token enforcement skips when `token_budget` is null.

- **Migration**
  - Single Alembic migration creates `user_usage` and `model_cost_override`, adds `cost_budget_cents`, makes `token_budget` nullable, and adds the budget check. Money values use `Numeric(18,6)` with `asdecimal=False`. Adds `created_at` to `model_cost_override`.
  - Downgrade deletes cost-only limits before restoring non-null `token_budget` to avoid blocking after rollback.

<sup>Written for commit 4163d9c97697501556e0122c90e73dd4ad13d693. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12550?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/alembic/versions/8c1d4f6a2e9b_add_usage_cost_tracking.py
+++ b/backend/alembic/versions/8c1d4f6a2e9b_add_usage_cost_tracking.py
@@ -29,10 +29,16 @@ def upgrade() -> None:
         # Empty string (not NULL) for a provider-agnostic override, so the unique
         # key works on every Postgres version (NULLS NOT DISTINCT is PG15+ only).
         sa.Column("provider", sa.String(), nullable=False, server_default=""),
-        sa.Column("input_cost_per_mtok", sa.Float(), nullable=False),
-        sa.Column("output_cost_per_mtok", sa.Float(), nullable=False),
+        sa.Column("input_cost_per_mtok", sa.Numeric(18, 6), nullable=False),
+        sa.Column("output_cost_per_mtok", sa.Numeric(18, 6), nullable=False),
         # null cache rate bills cache reads at the input rate (litellm default).
-        sa.Column("cache_read_cost_per_mtok", sa.Float(), nullable=True),
+        sa.Column("cache_read_cost_per_mtok", sa.Numeric(18, 6), nullable=True),
+        sa.Column(
+            "created_at",
+            sa.DateTime(timezone=True),
+            server_default=sa.func.now(),
+            nullable=False,
+        ),
         # The model's onupdate=func.now() is ORM-side only (not DDL), so
         # updated_at auto-bumps on ORM writes but not on raw-SQL UPDATEs.
         sa.Column(
@@ -69,7 +75,7 @@ def upgrade() -> None:
         sa.Column(
             "cache_read_tokens", sa.BigInteger(), nullable=False, server_default="0"
         ),
-        sa.Column("cost_cents", sa.Float(), nullable=False, server_default="0.0"),
+        sa.Column("cost_cents", sa.Numeric(18, 6), nullable=False, server_default="0"),
         sa.Column(
             "created_at",
             sa.DateTime(timezone=True),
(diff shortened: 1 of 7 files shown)


backend/alembic/versions/8c1d4f6a2e9b_add_usage_cost_tracking.py
```diff
@@ -0,0 +1,132 @@
...
+        sa.Column("output_cost_per_mtok", sa.Numeric(18, 6), nullable=False),
+        # null cache rate bills cache reads at the input rate (litellm default).
+        sa.Column("cache_read_cost_per_mtok", sa.Numeric(18, 6), nullable=True),
+        sa.Column(
+            "created_at",
+            sa.DateTime(timezone=True),
+            server_default=sa.func.now(),
+            nullable=False,
+        ),
+        # The model's onupdate=func.now() is ORM-side only (not DDL), so
```
Weves: hmm, this seems not optimal, since it could be postgres-server timezone to start and then api server timezone on update?

Weves: also ideally they're set the same way imo (either by ORM or by DB)

backend/onyx/db/models.py
```diff
@@ -4600,7 +4601,13 @@ class TokenRateLimit(Base):
 
     id: Mapped[int] = mapped_column(primary_key=True)
     enabled: Mapped[bool] = mapped_column(Boolean, nullable=False, default=True)
-    token_budget: Mapped[int] = mapped_column(Integer, nullable=False)
+    # A row carries a token budget, a cost budget, or both; a null side is exempt
+    # from that gate. At least one must be set — enforced by the check constraint
+    # below, so a backfill / direct write can't create an unenforceable row.
+    token_budget: Mapped[int | None] = mapped_column(Integer, nullable=True)
+    cost_budget_cents: Mapped[float | None] = mapped_column(
+        Numeric(18, 6, asdecimal=False), nullable=True
```
Weves, line 4676: why are we using Numeric rather than Float or Integer?

Weves: nit: it's not 100% clear to me if this is the budget in cents (e.g. 100 = 100 cents) or with granularity up to the cent level. I would expect the first, but the fact it's a float is throwing me off.

Weves: I see greptile's comment below. Would add that comment here to avoid confusion

Weves (approved):

backend/alembic/versions/8c1d4f6a2e9b_add_usage_cost_tracking.py
evan-onyx: I think this is standard for the codebase and the comment is a bit troll; it's always using the db clock

PR #12553, opened 2026-06-30
fix(file-processing): isolate PDF extraction in a subprocess (daemon-safe)

### LINKED ISSUE #12499: Drive connector: SIGSEGV (exit code -11) during PDF extraction on 4.2.0
### Summary

After upgrading to Onyx 4.2.0, indexing a Google Drive shared folder (~300 PDFs) consistently crashes the docfetching worker with a SIGSEGV
(exit code -11). This worked fine on 4.0. I wanted to share the traces and a few observations in case they're useful.

### Environment

- Onyx version: 4.2.0
- Python: 3.13 (from the image, `/usr/local/lib/python3.13/...`)
- Connector: Google Drive (shared folder, ~300 PDF files)
- Deployment: self-hosted (Docker image)

### Symptom

The `connector_doc_fetching_task` dies consistently after ~35s:

```
Indexing watchdog - finished: ... status=undefined exit_code=-11
exception="Docfetching returned exit code -11 with exception:
Job with ID '0' did not report an exception."
```

### Faulthandler trace

With faulthandler enabled, the crashing thread shows:

```
Fatal Python error: Segmentation fault
Thread 0x00007767e95ff6c0 (most recent call first):
  File ".../pypdfium2/_helpers/document.py", line 370 in get_page
  File ".../pypdfium2/_helpers/document.py", line 129 in getitem
  File ".../pypdfium2/_helpers/document.py", line 126 in iter
  File "/app/onyx/file_processing/extract_file_text.py", line 283 in _extract_pdf_text_pdfium
  File "/app/onyx/file_processing/extract_file_text.py", line 355 in read_pdf_file
  File "/app/onyx/connectors/google_drive/doc_conversion.py", line 461 in _download_and_extract_sections_basic
  File "/app/onyx/connectors/google_drive/doc_conversion.py", line 714 in _basic_extraction
  File "/app/onyx/connectors/google_drive/doc_conversion.py", line 818 in _convert_drive_item_to_document
  File "/app/onyx/connectors/google_drive/doc_conversion.py", line 643 in convert_drive_item_to_document
  File "/app/onyx/connectors/google_drive/connector.py", line 1825 in _convert_retrieved_file_to_document
  File ".../concurrent/futures/thread.py", line 59 in run
  File ".../concurrent/futures/thread.py", line 93 in _worker
  File ".../threading.py", line 995 in run
  File ".../threading.py", line 1044 in _bootstrap_inner
  File ".../threading.py", line 1015 in _bootstrap
```

### Observations

- The crash is in pypdfium2.get_page, called from _extract_pdf_text_pdfium, running inside a ThreadPoolExecutor (the _convert_retrieved_file_to_document conversion path).
- We tried MAX_DRIVE_WORKERS=1. The crash still occurs. Our reading is that this setting affects the download path rather than the conversion/extraction path.
- We noticed 4.2.0 switched PDF text extraction from pypdf to pypdfium2 (appears to be PR #12102; pypdfium2==5.9.0 added).
- pypdfium2/PDFium documents that it is not thread-safe, which might be relevant given the extraction runs in a thread pool.
- The pypdf fallback added in PR #12222 catches PdfiumError, but a SIGSEGV is a native crash rather than a Python exception, so it doesn't appear to engage here.

Is there a supported way to serialize or single-thread the extraction path as a workaround while this is investigated? We changed embedding models and now many of our search indices are useless until we can re-index.

nmgarza5:
## Description

**Bug:** PDF text extraction is supposed to run isolated, since a malformed PDF can hard-abort PDFium (SIGABRT/SIGSEGV) and take down the indexing worker. The merged isolation path spawned a `multiprocessing` child — but the indexing worker itself runs as a daemon (`job_client` sets `daemon=True`), and daemon processes aren't allowed multiprocessing children. So `run_in_isolated_process` raised `daemonic processes are not allowed to have children` on every call and silently fell back to pypdf. Isolation never actually ran.

**Fix:** Shell out with `subprocess.Popen` (OS-level fork+exec, not subject to the daemon restriction) via a new `isolated_runner` module. Adds a regression test that runs isolation from inside a daemon process.

**Verified:** Live on a dev instance — real Google Drive PDFs extract through subprocess isolation (`rc=0`, `_extract_pdf_text_pdfium`), with zero `daemonic` errors. Crash / timeout / exception fallbacks covered by unit tests.

## Additional Options

### DIFF
--- a/backend/onyx/utils/isolated_runner.py
+++ b/backend/onyx/utils/isolated_runner.py
@@ -1,0 +1,45 @@
+"""Child entry for ``run_in_isolated_process``: read a pickled (callable, args,
+kwargs) from stdin, run it, and write the pickled result back over a private copy
+of stdout.
+
+Run as ``python -m onyx.utils.isolated_runner`` so it works even when the caller is
+a daemon (the indexing worker), which can't spawn multiprocessing children.
+"""
+
+import os
+import sys
+
+
+def main() -> None:
+    # Keep a private copy of real stdout for the result, then send stdout/stderr to
+    # devnull so imports or native-lib chatter can't corrupt the pickled result.
+    result_fd = os.dup(1)
+    devnull = os.open(os.devnull, os.O_WRONLY)
+    os.dup2(devnull, 1)
+    os.dup2(devnull, 2)
+    os.close(devnull)
+
+    import pickle
+
+    from onyx.utils.process_isolation import STATUS_EXC
+    from onyx.utils.process_isolation import STATUS_OK
+    from onyx.utils.process_isolation import STATUS_UNRELAYABLE
+
+    fn, args, kwargs = pickle.loads(sys.stdin.buffer.read())  # noqa: S301
+    try:
+        payload = (STATUS_OK, fn(*args, **kwargs))
+    except Exception as e:
+        payload = (STATUS_EXC, e)
+
+    try:
+        data = pickle.dumps(payload)
+    except Exception:
+        # Result or exception didn't pickle; relay its repr so the parent still raises.
+        data = pickle.dumps((STATUS_UNRELAYABLE, repr(payload[1])))
+
+    os.write(result_fd, data)
+    os.close(result_fd)
+
+
+if __name__ == "__main__":
+    main()
--- a/backend/onyx/utils/process_isolation.py
+++ b/backend/onyx/utils/process_isolation.py
@@ -1,28 +1,26 @@
-"""Run a callable in a throwaway child process so a native crash (SIGSEGV/SIGABRT),
-a hang, or a runaway allocation kills the child instead of the caller. The parent
-gets a typed error and decides how to recover.
+"""Run a callable in a throwaway child process so a native crash (SIGSEGV/SIGABRT)
+or a hang takes down the child, not the caller; the parent gets a typed error and
+decides how to recover.
 
-For parse paths that feed untrusted bytes to native libs (PDF/image/office).
+Uses subprocess, not multiprocessing: the indexing worker is a daemon process, and
+daemons can't have multiprocessing children.
 """
 
-import multiprocessing as mp
+import pickle
 import signal
+import subprocess
+import sys
 from collections.abc import Callable
-from multiprocessing.connection import Connection
 from typing import Any
 from typing import TypeVar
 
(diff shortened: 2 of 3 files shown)


nmgarza5:
@greptile

backend/onyx/utils/process_isolation.py
```diff
@@ -49,109 +47,45 @@ def __init__(self, timeout_seconds: float) -> None:
...
-        send_conn.close()
-
-        if not recv_conn.poll(timeout):
-            proc.terminate()
-            proc.join(_SIGTERM_GRACE_SECONDS)
-            if proc.is_alive():
-                proc.kill()
-                proc.join()
-            raise IsolatedProcessTimeout(timeout)
+        proc = subprocess.Popen(
```
evan-onyx: would be good to check if we can't do a simple subprocess.run(), Popen is really only an "if you absolutely need it" thing

backend/onyx/utils/isolated_runner.py
```diff
@@ -0,0 +1,45 @@
...
+import sys
+
+
+def main() -> None:
+    # Keep a private copy of real stdout for the result, then send stdout/stderr to
+    # devnull so imports or native-lib chatter can't corrupt the pickled result.
+    result_fd = os.dup(1)
+    devnull = os.open(os.devnull, os.O_WRONLY)
+    os.dup2(devnull, 1)
+    os.dup2(devnull, 2)
```
evan-onyx, line 19: low level python code 🫨

evan-onyx: unironically could do the processing in rust haha

evan-onyx: like make the util a "run rust in subprocess" and do the pdf parsing there. Probably not worth the time rn but it's certainly fashionable atm :D

evan-onyx (approved):

backend/onyx/utils/isolated_runner.py
nmgarza5: lowkey..... im down to start moving things to rust. will not do here though lol

backend/onyx/utils/process_isolation.py
nmgarza5: fixed. thank you 🙏

PR #12562, opened 2026-06-30
fix(google-drive): retry connection drops during retrieval

nmgarza5:
## Description

Google Drive crawls fail the whole index attempt on a transient connection drop. `_execute_single_retrieval` retries `HttpError` 5xx and timeouts but not connection-level errors, so a single `BrokenPipeError` (connection reset / SSL teardown) mid-pagination propagates up through the parallel per-user retrieval and kills the attempt.

**Fix:** add `ConnectionError` (covers `BrokenPipeError` / reset / abort) and `ssl.SSLError` to the transient-retry branch in `google_utils._execute_single_retrieval`, reusing the existing `add_retries` backoff so a dropped socket retries the page instead of failing the crawl.

Test: parametrized unit test asserting broken pipe / reset / SSL / timeout are retried, not raised.

## Additional Options

### DIFF
--- a/backend/onyx/connectors/google_utils/google_utils.py
+++ b/backend/onyx/connectors/google_utils/google_utils.py
@@ -174,10 +174,11 @@ def _execute_single_retrieval(
         else:
             logger.exception("Error executing request:")
             raise e
-    except (TimeoutError, socket.timeout, ConnectionError, ssl.SSLError) as error:
-        # Connection-level drops (broken pipe, reset, SSL teardown) hit mid-crawl
-        # on long parallel Drive paginations; retry the page rather than failing
-        # the whole attempt.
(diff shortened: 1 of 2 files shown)


nmgarza5:
@greptile

backend/onyx/connectors/google_utils/google_utils.py
```diff
@@ -173,9 +174,13 @@ def _execute_single_retrieval(
         else:
             logger.exception("Error executing request:")
             raise e
-    except (TimeoutError, socket.timeout) as error:
+    except (TimeoutError, socket.timeout, ConnectionError, ssl.SSLEOFError) as error:
+        # Connection-level drops (broken pipe, reset, TLS EOF) hit mid-crawl on
```
Danelegend, line 178: Maybe remove or shorten this comment. The caught errors are reasonably self-describing

Danelegend (approved):

PR #12564, opened 2026-06-30
feat(craft): give the agent headless browser use via a built-in `browser` skill

rohoswagger:
## Description

Gives the Craft **agent** real (headless) browser use: it drives a Chromium inside the locked-down sandbox pod via a session-pinned `browser` CLI wrapping [`vercel-labs/agent-browser`](https://github.com/vercel-labs/agent-browser). This is agent-only — there is no human-facing streamed browser; the webapp preview is unchanged.

**Pieces**
- **`browser-cli.sh` → `/usr/local/bin/browser`**: imports the egress-proxy MITM CA into Chromium's NSS db (it reads trust only from there), sets launch flags (`--no-sandbox` since the pod drops the caps Chromium's own sandbox needs; `--proxy-server` since Chromium ignores `*_PROXY`), resolves the session id from cwd, and execs `agent-browser`. Fails fast with a clear message when no egress proxy is configured.
- **Dockerfile**: `ENABLE_BROWSER` ARG (default `true`) installs `chromium` + `libnss3-tools` + `agent-browser`. The wrapper is copied unconditionally and is inert when the runtime is absent.
- **Gating**: registry entry in `built_in.py` gated per-deployment on the `ENABLE_BROWSER` runtime env via `is_available`; surfaced through the existing `_exclude_unavailable_built_ins` path, so the skill is hidden from sandbox injection / AGENTS.md when off.
- **`SKILL.md`**: vendored verbatim from agent-browser's core usage guide with a mechanical `agent-browser`→`browser` substitution, plus a short Onyx note (prefer `webfetch` for static reads; the browser is session-pinned and headless).
- **Migration**: seeds the `browser` skill row — fail-loud if a tenant's custom skill already owns the `browser` slug, idempotent upsert otherwise.

## How Has This Been Tested?

- **Live end-to-end on the `craft-dev` EKS cluster**: the agent drove the browser through the egress proxy with the CA trusted and returned accurate live Hacker News data (titles + points). A DoorDash run reached the site fine but hit Cloudflare bot-detection (site-side, not a feature failure).
- **New automated tests**: `test_browser_built_in_gated_on_enable_browser` (gating, external-dependency-unit) and `test_browser.py` (k8s integration — `browser open … && browser snapshot` in a real pod, asserts the page rendered with no cert error). CI keeps `ENABLE_BROWSER` on by default.
- **Local**: `ruff format`/`ruff check`/`ty`/`shellcheck` all green via pre-commit; `bash -n` + behavioral test of the empty-proxy guard; `alembic heads` = single head.
- **4-lens automated review** (correctness / security / tests / regressions); both "critical/high" regression findings were verified to be false positives (a naive head computation flagged a multi-head that `alembic heads` disproves; the "duplicate migration" was transient git-churn). One medium (empty-proxy → cryptic failure) was fixed.

## Known minor items (for reviewer)

- **Default image size**: the release sandbox build passes no build-args, so the default image now bakes chromium + agent-browser (~400 MB) even though the runtime gate defaults off. Fail-safe (skill stays hidden) but worth a conscious call on whether the default artifact should carry it.
- **Supply chain**: `agent-browser` is version-pinned (`0.31.1`) but not integrity/hash-pinned; consider mirroring or SHA-pinning, consistent with the repo's base-image digest pinning.
- **`--no-sandbox`**: required (the pod drops the caps Chromium's sandbox needs); the pod's seccomp/caps/iptables egress remain the real isolation boundary — keep the pod seccomp profile restrictive.
- **Test oracle**: `test_available_built_in_is_visible` derives its expectation from the same gate it checks; the dedicated gating test is what actually covers gating. Migration guard/upsert paths are untested.
- **Session-id regex** in `browser-cli.sh` matches any 36-char `[0-9a-fA-F-]` run rather than a strict UUID; the fallback handles non-matches.

## Additional Options

---
## Summary by cubic
Adds a built-in `browser` skill so the Craft agent can drive a headless Chromium in the sandbox via a session‑pinned `browser` CLI. Moves proxy CA import to sandbox startup for reliability, splits the image into Chromium and `agent-browser` layers for faster pulls, and updates docs and tests.

- **New Features**
  - `browser` CLI sets `--no-sandbox` and proxy flags, pins the session, and execs `agent-browser`; fails fast if no egress proxy is set.
  - Startup CA trust: `entrypoint.sh` imports the egress‑proxy CA into Chromium’s `NSS` DB once at pod start; only marks success after at least one cert imports and returns non‑zero when nothing imports so logs surface failures.
  - Sandbox image: `ENABLE_BROWSER` build arg (default true) installs `chromium`, `libnss3-tools`, and `agent-browser` (`0.31.1`) in separate Chromium and CLI layers; wrapper is inert if the runtime is absent.
  - Runtime gate: `ENABLE_BROWSER` env toggles visibility via the built‑in registry; config added in `features/build/configs.py`.
  - Tests/docs: gating unit test and a k8s integration test; CI keeps the browser on by default; docs include `SKILL.md` and a browser‑use spec.

- **Migration**
  - Seeds the `browser` skill and guards slug collision with any existing custom skill.
  - Uses `public_permission='VIEWER'` (post‑sharing refactor) and chains onto the latest head; normal `alembic upgrade head` is sufficient.

<sup>Written for commit d0f0fc4a2260cf075e91287fc53c0d7488c99d6c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12564?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/onyx/server/features/build/sandbox/image/browser-cli.sh
```diff
@@ -0,0 +1,46 @@
...
+if [ -f "$CA_BUNDLE" ] && [ ! -f "$NSSDB/.proxy-ca-imported" ]; then
+    mkdir -p "$NSSDB"
+    [ -f "$NSSDB/cert9.db" ] || certutil -d "sql:$NSSDB" -N --empty-password
+    SPLITDIR="$(mktemp -d)"
+    csplit -z -f "$SPLITDIR/ca-" -b "%03d.pem" "$CA_BUNDLE" "/BEGIN CERTIFICATE/" "{*}" >/dev/null 2>&1
+    for f in "$SPLITDIR"/ca-*.pem; do
+        [ -f "$f" ] && certutil -d "sql:$NSSDB" -A -t "C,," -n "proxy-$(basename "$f" .pem)" -i "$f" 2>/dev/null || true
+    done
+    rm -rf "$SPLITDIR"
+    touch "$NSSDB/.proxy-ca-imported"
```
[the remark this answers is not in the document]
rohoswagger: Good catch — fixed in 2a2102f8f0. The loop now counts successful imports and only touches `.proxy-ca-imported` when `imported > 0`; an empty/bad bundle warns and retries on the next invocation instead of latching the sentinel.

backend/onyx/server/features/build/sandbox/image/browser-cli.sh
```diff
@@ -0,0 +1,46 @@
...
+CLEAN_PROXY="$(printf '%s' "${HTTPS_PROXY:-${HTTP_PROXY:-}}" | sed -E 's#^([a-z]+://)?[^@/]*@#\1#')"
+# Fail fast: with no proxy, an empty --proxy-server= flag makes Chromium go
+# direct, which the pod's egress lockdown blocks — surfaces as cryptic connection
+# errors instead of a clear cause. Skip when AGENT_BROWSER_ARGS is overridden.
+if [ -z "${AGENT_BROWSER_ARGS:-}" ] && [ -z "$CLEAN_PROXY" ]; then
+    echo "browser: no egress proxy configured (HTTPS_PROXY/HTTP_PROXY unset)" >&2
+    exit 1
+fi
+export AGENT_BROWSER_EXECUTABLE_PATH="${AGENT_BROWSER_EXECUTABLE_PATH:-/usr/bin/chromium}"
+export AGENT_BROWSER_ARGS="${AGENT_BROWSER_ARGS:---no-sandbox,--proxy-server=${CLEAN_PROXY},--proxy-bypass-list=127.0.0.1;localhost,--disable-dev-shm-usage,--disable-gpu}"
```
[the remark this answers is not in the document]
rohoswagger: Fixed in 2a2102f8f0. The Onyx note in SKILL.md now explicitly tells the agent that `--headed`, `--provider cloud-browser`, `browser plugin add`, `browser doctor --fix`, and visible-window 2FA don't apply in the headless pod, and to drive auth via snapshot/fill/click instead.

rohoswagger:
@greptile-ai @cubic-dev-ai re-review this

rohoswagger:
@codex review this

backend/alembic/versions/989bc57562e4_seed_browser_built_in_skill.py
```diff
@@ -0,0 +1,101 @@
...
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "eebb02b43bee"
+down_revision = "99c855a8f2a1"
+branch_labels = None
+depends_on = None
+
+_BROWSER_SLUG = "browser"
```
wenxi-onyx, line 31: nit: any reason this needs to be defined twice? I think it's probably fine to just have one string since it's just a migration and not application code

backend/alembic/versions/989bc57562e4_seed_browser_built_in_skill.py
```diff
@@ -0,0 +1,101 @@
...
+    sa.column("author_user_id", postgresql.UUID(as_uuid=True)),
+    sa.column("is_public", sa.Boolean),
+    sa.column("enabled", sa.Boolean),
+)
+
+
+def upgrade() -> None:
+    bind = op.get_bind()
+
+    # Fail loud if a custom skill already owns the slug (built_in_skill_id IS
```
wenxi-onyx, line 58: Not in scope: we need to figure out how to address this after GA. If we want to publish built-ins, need a way to resolve for custom skills. Maybe could do something like temporarily re-naming it and doing some kind of alert in app

backend/onyx/server/features/build/sandbox/image/Dockerfile
```diff
@@ -47,7 +47,9 @@ FROM ${BASE_IMAGE_REGISTRY}/library/python:3.13-slim@sha256:f82c96458eedc847b233
 # the API server at session-setup, but their runtime tools must already be in
 # the image.
 ARG ENABLE_SKILLS=true
+ARG ENABLE_BROWSER=true
```
wenxi-onyx, line 50: Why does this need a switch?

backend/onyx/server/features/build/configs.py
```diff
@@ -69,6 +69,10 @@ class SandboxBackend(str, Enum):
 
 ENABLE_CRAFT = os.environ.get("ENABLE_CRAFT", "false").lower() == "true"
 
+# Gates the built-in `browser` skill. Distinct from the sandbox image's
+# build-time ENABLE_BROWSER ARG — set this runtime env to match it.
+ENABLE_BROWSER = os.environ.get("ENABLE_BROWSER", "false").lower() == "true"
```
wenxi-onyx, line 74: per above, would pref to just force enable.

wenxi-onyx (approved):

backend/onyx/server/features/build/sandbox/image/entrypoint.sh
```diff
@@ -19,6 +19,32 @@ if [ -z "${OPENCODE_SERVER_PASSWORD:-}" ]; then
     echo "[entrypoint] WARNING: OPENCODE_SERVER_PASSWORD is empty — opencode serve will run without auth"
 fi
 
+# Trust the egress-proxy MITM CA in Chromium's NSS db (Chromium reads trust only
+# from there) so browser HTTPS works. Best-effort — a failure, or no browser
+# runtime, must not block boot.
+import_proxy_ca() {
+    command -v certutil >/dev/null 2>&1 || return 0
+    local bundle="${SANDBOX_PROXY_CA_BUNDLE_DST:-/etc/ssl/sandbox/ca-bundle.crt}"
+    [ -f "$bundle" ] || return 0
```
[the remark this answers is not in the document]
rohoswagger: Fixed in d18f014a4d. `import_proxy_ca` now returns non-zero on the no-import paths (certutil missing / CA bundle not mounted), so the caller takes the `|| …` branch and logs "proxy CA not imported into NSS (browser runtime absent or CA missing)" instead of falsely claiming success. Boot is still never blocked — the function is invoked in a `&& … || …` list, so `set -e` doesn't abort — and a browser-enabled image with a broken CA mount now surfaces the misconfig in the logs.

backend/alembic/versions/989bc57562e4_seed_browser_built_in_skill.py
rohoswagger: Good call — collapsed in 19039bea57. Dropped `_BROWSER_NAME` and use the single `_BROWSER_SLUG` constant for both `slug` and `name`.

backend/alembic/versions/989bc57562e4_seed_browser_built_in_skill.py
rohoswagger: Agreed — out of scope for this PR, flagging as a post-GA follow-up. In the meantime the guard is intentionally fail-loud (raises if a custom skill already owns the `browser` slug) so a built-in can never silently clobber or shadow a tenant's custom skill. When we publish built-ins we'll need the resolution flow you describe (e.g. temporary rename + in-app alert). This is worth a tracking ticket so it isn't lost — happy to file one. Interim behavior left as-is; no code change here.

backend/onyx/server/features/build/sandbox/image/Dockerfile
rohoswagger: Keeping the switch (confirmed with @rohoswagger). The `ENABLE_BROWSER` build ARG lets dev/CI and any deployment that doesn't want the browser skip baking chromium + agent-browser (~290 MB), keeping those images lean — it's the build-time half that the runtime `ENABLE_BROWSER` env is meant to match. Default is `true`, so prod/craft-dev get it by default; only opt-out builds pass `false`. No change.

backend/onyx/server/features/build/configs.py
rohoswagger: Keeping the runtime env rather than force-on (confirmed with @rohoswagger). It has to match the image's build-time capability: a deployment whose sandbox image was built with `ENABLE_BROWSER=false` has no `browser`/agent-browser binary, so the skill must stay hidden there — forcing it on would advertise a skill to the agent whose command doesn't exist (the exact footgun). Defaults to off; deployments that bake the runtime set it on. No change.

backend/alembic/versions/989bc57562e4_seed_browser_built_in_skill.py
rohoswagger: Filed as [ENG-4264](https://linear.app/onyx-app/issue/ENG-4264/resolve-built-in-vs-custom-skill-slug-collisions-when-publishing-built) in the Craft V2 project to track the post-GA resolution flow.

PR #12565, opened 2026-06-30
fix(telemetry): emit broker-global queue metrics once per window

nmgarza5:
## Description

**Bug:** the cloud monitoring workers OOM-crash-loop. Root cause, confirmed against a week of telemetry data: `monitor_background_processes` runs once per tenant, and `_collect_queue_metrics` measures ~20 broker-global Celery queue depths and emits them with `key=None` (every run). The queues are shared across all tenants, so every tenant re-emits the same global values. With hundreds of tenants that is about 1.8 million telemetry events per day in ~300/sec bursts. Each `optional_telemetry` call spawns a thread doing a timeout-less POST, so the threads pile up (measured 2,038 on the worker) and OOM the 4 GiB pod.

**Fix:**
1. Emit the global queue metrics from at most one tenant per 5-minute window. The queue block is gated behind a cross-tenant lease in the shared redis namespace, so the first tenant of a window emits and the rest skip. This removes the per-tenant redundancy (the bulk of the volume) at the source and matches the original 5-minute monitor cadence. The metrics keep their real per-tenant context, so the EE instance_domain lookup is unaffected.
2. Add a timeout to the telemetry POST, so a slow or unreachable endpoint cannot pin a sender thread and let threads accumulate.

Connector and sync metrics are genuinely per-tenant and keyed/deduped, so they are unchanged.

Note: while tracing this I found a separate pre-existing bug where four sync metrics share one dedup key, so only `sync_run_succeeded` ever emits (duration, doc count, speed are dropped). Out of scope here, flagging for a follow-up.

## Additional Options

- [ ] This PR should be considered for cherry-pick if the release branch has been cut
- [x] Override Linear Check

### DIFF
--- a/backend/onyx/background/celery/tasks/monitoring/tasks.py
+++ b/backend/onyx/background/celery/tasks/monitoring/tasks.py
@@ -42,6 +42,7 @@ from onyx.db.models import SyncRecord
 from onyx.db.models import UserGroup
 from onyx.db.search_settings import get_active_search_settings_list
 from onyx.redis.redis_pool import get_redis_client
+from onyx.redis.redis_pool import get_shared_redis_client
 from onyx.redis.redis_pool import redis_lock_dump
 from onyx.redis.tenant_redis_client import TenantRedisClient
 from onyx.utils.platform_utils import is_running_in_container
@@ -53,6 +54,12 @@ from shared_configs.contextvars import CURRENT_TENANT_ID_CONTEXTVAR
 
 _MONITORING_SOFT_TIME_LIMIT = 60 * 5  # 5 minutes
 _MONITORING_TIME_LIMIT = _MONITORING_SOFT_TIME_LIMIT + 60  # 6 minutes
+
+# Queue lengths are broker-global, so they are emitted by at most one tenant per
+# window rather than once per tenant every cycle. Keep the lease shorter than the
+# 5-minute monitor beat so it expires before the next cycle and never blocks it.
+_GLOBAL_QUEUE_METRICS_LEASE = "monitoring_global_queue_metrics_lease"
+_GLOBAL_QUEUE_METRICS_LEASE_TTL = 60 * 4  # 4 minutes
 
 _CONNECTOR_INDEX_ATTEMPT_START_LATENCY_KEY_FMT = (
     "monitoring_connector_index_attempt_start_latency:{cc_pair_id}:{index_attempt_id}"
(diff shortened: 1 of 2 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

backend/onyx/background/celery/tasks/monitoring/tasks.py
```diff
@@ -54,6 +55,11 @@
 _MONITORING_SOFT_TIME_LIMIT = 60 * 5  # 5 minutes
 _MONITORING_TIME_LIMIT = _MONITORING_SOFT_TIME_LIMIT + 60  # 6 minutes
 
+# Queue lengths are broker-global, so they are emitted by at most one tenant per
+# this window rather than once per tenant every cycle.
+_GLOBAL_QUEUE_METRICS_LEASE = "monitoring_global_queue_metrics_lease"
+_GLOBAL_QUEUE_METRICS_LEASE_TTL = 60  # seconds
```
nmgarza5: im debating making the window longer. I'm thinking these metrics (which go to metabase) don't need to be updated so frequently

nmgarza5: gonna go with 5 minutes

nmgarza5:
@greptile

Bo-Onyx (approved):
One optional note: _GLOBAL_QUEUE_METRICS_LEASE_TTL (60 * 5) exactly equals the 5-min beat cadence, leaving zero margin. Emission normally still happens every window since tenant attempts fan out, but if they ever cluster early a window could be skipped (10-min gap). Dropping the TTL to ~240s would close that edge case. Not blocking.


eg. 

```
Say the monitoring beat fires at t=0, t=300, t=600, … (every 5 min = 300s), and in each cycle the tenant tasks happen to get picked up and attempt the lease in a tight cluster right after the cycle starts.

With TTL = 300 (current)

- Cycle 1 (t=0): first tenant acquires the lease at, say, t=5. Lease expires at t=5 + 300 = t=305.
- Cycle 2 (t=300): the tenants attempt between t=300 and t=304 (clustered early this cycle).
  - But the lease from cycle 1 is still alive until t=305.
  - Every attempt at t=300…304 → acquire() returns False.
  - No one attempts after t=305 → nobody acquires → queue metrics skipped this window.
- Result: a 10-minute gap (cycle-1 emission, then nothing in cycle 2, next emission in cycle 3).

The problem: because the lease was acquired 5s into cycle 1, it bleeds 5s into cycle 2. That 5s is a "dead zone" at the start of every cycle where the old lease is still held.

With TTL = 240

- Cycle 1 (t=0): first tenant acquires at t=5. Lease expires at t=5 + 240 = t=245.
- Cycle 2 (t=300): the lease expired at t=245 — that's 55s before cycle 2 even started.
  - So the very first tenant to attempt in cycle 2 (at t=300) → lease is free → acquires and emits. ✅
- No dead zone, no possible skip.
```

nmgarza5:
@greptile

PR #12574, opened 2026-06-30
feat(search): support time_cutoff_upper for bounded time filtering

Danelegend:
## Summary

Extends the document index filters with an **upper** bound on a document's last-updated time (`time_cutoff_upper`), so an internal search can scope to a closed `[time_cutoff, time_cutoff_upper]` range instead of only a lower cutoff. This is the index-layer plumbing only; the secondary-LLM time-filter flow that decides the window lives in a separate branch/PR.

## Changes

- **Interfaces** (`models.py`, `pipeline.py`): add `time_cutoff_upper: datetime | None` to `BaseFilters` (inherited by `IndexFilters`) and pass it through `_build_index_filters` into the final `IndexFilters`.
- **OpenSearch** (`opensearch/search.py`): thread `time_cutoff_upper` through the query builders; emit a single `range` clause with `gte` / `lte` as available. Undated documents (no `LAST_UPDATED_FIELD_NAME`) are only included for an **old, open-ended lower bound** — a bounded range excludes them, since an undated doc can't be shown to fall inside the range.
- **Vespa** (`vespa_request_builders.py`): `_build_time_filter` now builds `>=` and/or `<=` clauses joined with `and`, applying the same undated-doc rule.
- **Docs** (`FILTER_SEMANTICS.md`): document the new field and its range/undated semantics.

## Testing

- `test_vespa_query.py`: new cases for a bounded range (strict `>=` and `<=`, untimed docs excluded even when the lower bound is old) and an upper-bound-only filter.
- `test_assistant_knowledge_filter.py`: pass the new required arg.

Existing vespa unit tests pass (`pytest backend/tests/unit/onyx/utils/test_vespa_query.py`); `ty`, ruff, and ruff-format pass via pre-commit.

---
## Summary by cubic
Adds an upper time bound (`time_cutoff_upper`) to the last-updated filter so searches can use a bounded [time_cutoff, time_cutoff_upper] range or an upper-only cutoff. Undated docs are only included for an old, open-ended lower bound; bounded and upper-only ranges exclude them.

- New Features
  - Added `time_cutoff_upper` to `BaseFilters`/`IndexFilters` and threaded it through the pipeline and all OpenSearch query builders.
  - OpenSearch: emit one `range` on `last_updated` with `gte`/`lte` as provided; OR-in undated docs only for an old, open-ended lower bound; documented the new arg in `_get_search_filters` docstring.
  - Vespa: build `>=`/`<=` clauses joined with `and`, with the same undated-doc rule.
  - Docs/tests: updated `FILTER_SEMANTICS.md` to note standalone `time_cutoff_upper`; added OpenSearch and Vespa unit tests for lower-only, upper-only, and bounded ranges; updated affected tests.

<sup>Written for commit 5dd2cb3a8891034c9f17f57a647ff4a194c8b2a8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12574?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/document_index/vespa/shared_utils/vespa_request_builders.py
+++ b/backend/onyx/document_index/vespa/shared_utils/vespa_request_builders.py
@@ -125,16 +125,30 @@ def build_vespa_filters(
 
     def _build_time_filter(
         cutoff: datetime | None,
+        cutoff_upper: datetime | None = None,
         untimed_doc_cutoff: timedelta = timedelta(days=92),
     ) -> str:
-        if not cutoff:
+        if not cutoff and not cutoff_upper:
             return ""
-        include_untimed = datetime.now(timezone.utc) - untimed_doc_cutoff > cutoff
-        cutoff_secs = int(cutoff.timestamp())
 
-        if include_untimed:
-            return f"!({DOC_UPDATED_AT} < {cutoff_secs})"
-        return f"({DOC_UPDATED_AT} >= {cutoff_secs})"
+        clauses: list[str] = []
+        if cutoff:
+            # Untimed docs (no doc_updated_at) are only included for an old, open-
+            # ended lower bound. A bounded range excludes them — an undated doc
+            # cannot be shown to fall within [cutoff, cutoff_upper].
+            include_untimed = (
+                cutoff_upper is None
+                and datetime.now(timezone.utc) - untimed_doc_cutoff > cutoff
+            )
+            cutoff_secs = int(cutoff.timestamp())
+            if include_untimed:
+                clauses.append(f"!({DOC_UPDATED_AT} < {cutoff_secs})")
+            else:
+                clauses.append(f"({DOC_UPDATED_AT} >= {cutoff_secs})")
+        if cutoff_upper:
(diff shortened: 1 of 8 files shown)


backend/onyx/document_index/vespa/shared_utils/vespa_request_builders.py
```diff
@@ -125,16 +125,30 @@ def _build_kg_chunk_id_zero_only_filter(
 
     def _build_time_filter(
         cutoff: datetime | None,
+        cutoff_upper: datetime | None = None,
```
acaprau, line 128: you don't have to support vespa anymore but it's ok either way 😅

acaprau (approved):

PR #12575, opened 2026-06-30
fix(indexing): deliver repeated-error connector notifications to admins

nmgarza5:
## Description

#10207 added an admin notification when a connector enters repeated-error state, but no admin ever saw it.

**Bug:** the notification was written with `user_id=None`. `GET /notifications` only returns rows where `user_id == <requesting user>.id` (`_notification_filters`), and nothing merges global `user_id IS NULL` rows into that list — so the row was written but never read.

**Fix:**
- Fan the alert out to every active admin via `get_active_admin_users` + `batch_create_notifications` (same delivery pattern as release notes / license-expiry warnings).
- On recovery, clear all admins' rows for the connector via a new `delete_notifications_by_additional_data` helper, so a fresh alert fires if it fails again. Deletes regardless of `dismissed`, also closing the prior "re-entry after dismissal" gap.

## Additional Options

### DIFF
--- a/backend/onyx/background/celery/tasks/docprocessing/tasks.py
+++ b/backend/onyx/background/celery/tasks/docprocessing/tasks.py
@@ -94,11 +94,12 @@ from onyx.db.indexing_coordination import CoordinationStatus
 from onyx.db.indexing_coordination import IndexingCoordination
 from onyx.db.models import IndexAttempt
 from onyx.db.models import SearchSettings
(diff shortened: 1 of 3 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

acaprau (approved):

PR #12580, opened 2026-06-30
fix: scope cross-tenant permission caches per tenant/instance

nmgarza5:
## Description

Security fix found during a caching audit. Permission-resolution code cached results in process-global module state with no tenant/instance in the key, so in a shared multi-tenant worker one tenant's or instance's data served another's access check.

- **Salesforce** (`ee/onyx/external_permissions/salesforce/utils.py`): the query-time censoring client and email->user_id map are now keyed by `tenant_id`. Also deletes the unused `_get_salesforce_client_for_doc_id` and its globals (same latent defect).
- **Confluence** (`onyx/connectors/confluence/onyx_confluence.py`): the username->email, userKey->email, and userid->display-name caches are now keyed by `(instance base url, identifier)`. Server/DC usernames and userKeys are unique only within one instance.

Single-tenant and lite deployments are unaffected (default-schema / single-instance key, plain in-process dicts).

## How Has This Been Tested?

New unit tests assert the Salesforce (tenant) and Confluence (instance) caches are isolated. ruff + project-wide ty + pytest green locally.

## Additional Options

### DIFF
--- a/backend/ee/onyx/external_permissions/salesforce/utils.py
+++ b/backend/ee/onyx/external_permissions/salesforce/utils.py
@@ -2,33 +2,32 @@ from simple_salesforce import Salesforce
 from simple_salesforce.format import format_soql
 from sqlalchemy.orm import Session
 
-from onyx.db.connector_credential_pair import get_connector_credential_pair_from_id
 from onyx.db.document import get_cc_pairs_for_document
 from onyx.utils.logger import setup_logger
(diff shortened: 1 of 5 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

justin-tahara (approved):

PR #12587, opened 2026-07-01
feat(mobile): add settings layout primitives + tab nav

Subash-Mohan:
## Description

- Port web's Opal `SettingsLayouts` to the mobile app as reusable primitives: `SettingsLayout` (Root / Header / Body) plus `SettingsTabNav` (vertical section nav) and a `sliders` header icon.
- Phone-first and web-faithful: drops web's responsive width variants, the `InputSelect` dropdown, the sticky scroll-shadow, and the in-content back button (mobile back is the OS gesture / screen chrome); screen wiring is intentionally left to the consuming screens.

## How Has This Been Tested?

- Mobile typecheck, lint, and format pass; the primitives are presentational and not yet mounted in a screen.

## Additional Options

### DIFF
--- a/mobile/src/components/settings/SettingsLayout.tsx
+++ b/mobile/src/components/settings/SettingsLayout.tsx
@@ -1,0 +1,105 @@
+import type { ReactNode } from "react";
+import { ScrollView, View } from "react-native";
+
+import { cn } from "@/lib/utils";
+import { Icon } from "@/components/ui/icon";
+import { Separator } from "@/components/ui/separator";
+import { Text } from "@/components/ui/text";
+import type { IconFunctionComponent } from "@/icons/types";
+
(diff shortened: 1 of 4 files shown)


Subash-Mohan:
Addressed the `activeHref`/`Href` mismatch in `SettingsTabNav` (fc7802e83c): each tab's href is reduced to its pathname and compared against `activeHref` (a `usePathname()` string), which handles both bare-path and object-form hrefs.

I intentionally didn't take the `String(tab.href)` suggestion — `String()` on an object href yields `"[object Object]"`, so *all* object hrefs would collide and compare equal. Extracting `.pathname` avoids that. `activeHref` stays `string` since it's sourced from `usePathname()`.

nmgarza5 (approved):

PR #12613, opened 2026-07-01
feat(craft): context-window usage ring + compaction marker

rohoswagger:
## Description

Surfaces opencode's **context-window usage** and **auto-compaction** in Onyx Craft.

- **Context ring** in the composer's bottom-right: a small ring showing % of the context window used. Ambient below 70%, reveals the % (amber) at 70–90%, red near the auto-compact threshold. Tooltip shows `used / limit tokens` + `% used` (two lines, compact `K`/`M` formatting, matching Claude Code's `/context` / Codex — cost deliberately omitted). Hidden until a limit + non-zero usage are known.
- **Compaction marker** in the transcript: an understated divider (mirroring the main app's "Stopped" step — muted icon + label, not an error) shown when opencode compacts, with the summary tucked behind a "View summary" `Collapsible`. The summary message's inline text is suppressed.
- **Limit synced from opencode**: the ring's denominator is opencode's own `model.limit.context` (fetched once from `/config/providers`, cached process-wide), falling back to the model's configured `max_input_tokens`. Verified live that this yields `1,000,000` (opencode) rather than Onyx's `998,976` for `claude-opus-4-8`.

Backend adds `ContextUsagePacket` / `CompactionPacket`, translated from opencode's assistant `message.updated` token counts and the `session.compacted` event; usage + marker persist for reload. Subagent turns don't touch the main-session ring.

A plan for a follow-up user-facing `/compact` slash command is included at `docs/craft/features/compact-command.md` (not implemented in this PR).

## How Has This Been Tested?
Screenshot 2026-07-01 at 15 27 01

- **Backend unit** (`test_translate_opencode_event.py`, `test_streaming_state.py`): token-sum + zero/absent-token handling, `summary:true` suppression + id recording, `session.compacted` → marker + summary fetch, `_extract_context_limit` (list + dict catalog shapes), and the persist path (usage doesn't fragment streaming text / returns before chunk-flush, compaction persists after a flush, finalize writes exactly one usage row, no double-write on the error path). **110 pass.**
- **Frontend unit** (`formatTokens.test.ts`, `parsePacket.test.ts`): `formatTokens` boundaries incl. `999_999 → "1M"`, and `context_usage`/`compaction` parsing (snake + camel case). Existing craft suites still green. **61 pass.**
- **Live** (real Craft turn, Claude Opus 4.8): ring renders at 2% in the ambient state, tooltip correct, limit confirmed synced from opencode, and the ring survives a full page reload.
- Reviewed with a 7-lens parallel review team (correctness, security-scope, tests, regressions, performance, data-access, concurrency); confirmed findings fixed (subagent ring contamination on reload, per-turn `/config/providers` round-trip, `CompactionMarker` now uses the house `Collapsible`, `formatTokens` boundary, added persist/format/parse tests).

### Known minor items (for the reviewer)
- The `max_input_tokens` **fallback** denominator (~998,976) differs slightly from opencode's `limit.context` (1,000,000); only in effect if the `/config/providers` fetch fails. Documented, by-design.
- `session.compacted` with no `sessionID` and the summary-text delta race were both reviewed and determined non-triggerable given opencode's contract (always sends `sessionID`; sets `summary:true` at message creation) — no code change.

## Additional Options

---
## Summary by cubic
Adds a context-window usage ring in the Craft composer and a compaction marker in the transcript to surface opencode’s auto-compaction. The ring now shows percent used based on the model’s `max_input_tokens` with no provider fetch.

- **New Features**
  - Context ring (composer bottom-right): ambient <70%, amber with % at 70–90%, red near threshold; tooltip shows “N% of context used.” Hidden until usage is known.
  - Compaction marker in the transcript: muted divider with a “View summary” collapsible; inline summary text is suppressed.
  - Backend emits `context_usage` and `compaction` packets from `message.updated` and `session.compacted`. Usage persists once per turn so the ring survives reload; subagent turns are ignored.

- **Refactors**
  - Removed `/config/providers` limit fetch and shared cache; the ring computes % against the model’s `max_input_tokens` on the frontend.
  - Simplified tooltip content and dropped the unused token formatting helper.

<sup>Written for commit 496110d3dd43856e9c0d43e83a415bbf1c05e5f7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12613?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/onyx/server/features/build/sandbox/opencode/serve_client.py
```diff
@@ -53,6 +55,10 @@
 
 logger = setup_logger()
 
+# (providerID, modelID) → context window. Process-wide: clients are rebuilt per
+# turn, and models.dev limits are stable per model, so one fetch serves all turns.
+_CONTEXT_LIMIT_CACHE: dict[tuple[str, str], int] = {}
```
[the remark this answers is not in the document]
rohoswagger: Fixed in 75665d77e5 — the cache key now includes `directory`, so a per-workspace opencode config can't leak one tenant's limit to another. A session's turns share a directory, so this keeps the per-session (not per-turn) caching.

backend/onyx/server/features/build/packets.py
```diff
@@ -91,6 +91,18 @@ class ConnectAppRequestPacket(BasePacket):
     reason: str | None = None
 
 
+class ContextUsagePacket(BasePacket):
+    type: Literal["context_usage"] = "context_usage"
+    used_tokens: int
+    context_limit: int | None = None
+    cost: float | None = None
```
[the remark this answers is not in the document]
rohoswagger: Intentional. `cost` is captured/persisted on purpose for a planned session cost surface; the FE read was deliberately stripped (this indicator stays about context, not spend, matching Claude Code/Codex). It's one float inside an already-persisted packet, so no meaningful data growth. Leaving it retained rather than removing and re-adding later.

backend/onyx/server/features/build/sandbox/opencode/serve_client.py
```diff
@@ -53,6 +55,12 @@
 
 logger = setup_logger()
 
+# (directory, providerID, modelID) → context window. Directory-scoped so a
+# per-workspace opencode config can't leak one tenant's limit to another; still
```
wenxi-onyx: Misleading comment - tenant's workspaces are not stored together.

backend/onyx/server/features/build/sandbox/opencode/serve_client.py
```diff
@@ -53,6 +55,12 @@
...
+# process-wide, so one fetch serves every turn of a session (clients are rebuilt
+# per turn).
+_CONTEXT_LIMIT_CACHE: dict[tuple[str, str, str], int] = {}
```
wenxi-onyx: Should add some kind of class to type this more clearly.

wenxi-onyx: If this is truly a "model context limit" cache, there's no need to separate it by tenant/session right? If I'm using claude 4.8, the limit is the same universally?

wenxi-onyx: What exactly is the purpose of retrieving this? It looks like the context usage packet exposes limit as well? Seems to only exist/used in abort?

backend/onyx/server/features/build/sandbox/opencode/serve_client.py
rohoswagger: Fixed in ba913fad03. You're right — the tenant-leak framing was wrong. The limit is universal per (provider, model), so the misleading comment is gone and the value now lives in the shared cache (get_cache_backend) with an accurate comment.

backend/onyx/server/features/build/sandbox/opencode/serve_client.py
rohoswagger: Addressed in ba913fad03 — moot now: switched off the in-memory dict to the shared cache keyed by a namespaced string (`craft:context_limit:{provider}:{model}`), so there's no tuple key left to type.

rohoswagger: Agreed, and fixed in ba913fad03. The context limit is universal per (provider, model) for the pinned opencode version, so there's no tenant/session dimension. Dropped the directory scoping entirely — it's now a single shared-cache entry per (provider, model) with a 1h TTL.

rohoswagger: Rationale: we fetch opencode's own `model.limit.context` (from /config/providers) specifically to populate the packet's `context_limit` — the packet exposes it *because* this fetch supplies it. It matters because the alternative (the FE's fallback to Onyx's configured `max_input_tokens`) differs from what opencode actually compacts against — e.g. 998,976 vs opencode's 1,000,000 for claude-opus-4-8 — so the ring would misreport. Confirmed live. The fetch is cached (see above) so it's not per-turn.

wenxi-onyx (approved):

PR #12618, opened 2026-07-01
feat(ods): add `ods audit --actions` to scan pinned GitHub Actions

jmelahman:
## What

Adds `ods audit --actions`, auditing the GitHub Actions pinned in `.github/workflows` **and** the composite actions in `.github/actions` against OSV.dev, gating deploys on known-vulnerable actions the same way the lockfile and Dependabot sources already do. It joins the default all-sources run, so bare `ods audit` (and the CI gate + nightly SARIF) now covers actions too.

```
$ ods audit --actions
Dependency vulnerabilities:
  HIGH      GitHub Actions GHSA-mrrh-fwg8-r2c3    tj-actions/changed-files@40…  tj-actions changed-files through 45.0.7 …
  UNKNOWN   GitHub Actions GHSA-vqf5-2xx6-9wfm    github/codeql-action@ba454b…  unverified pin — GitHub PAT written to …
```

## Why a custom matcher

osv-scalibr (PR google/osv-scalibr#2032) can *extract* workflow action refs, but osv-scanner v2.4.0 **cannot match** them:

- `semantic.Parse` has no comparator for the `GitHub Actions` ecosystem, so `IsAffected` returns `false` for every action — in both online and offline modes (the code even comments this).
- GitHub Actions advisories carry only `ECOSYSTEM` version ranges (no enumerated versions, no affected commits), which OSV.dev's version-query endpoint can't evaluate. Verified against the live API: a versioned GitHub Actions query returns nothing even for a version an advisory explicitly lists as affected.

So a naive `--actions` on `DoScan` would run, extract everything, and detect nothing — a silent no-op gate. Instead this uses the extractor only for **discovery** and matches ourselves.

## How it works

1. **Extract** actions via scalibr's `github/actions` extractor (`owner/repo@ref`), across both `.github/workflows/*.yml` and the composite `.github/actions/*/action.yml` (each composite's `runs.steps` is reshaped into a synthetic one-job workflow so the same extractor parses it).
2. **Query OSV.dev** by action name only (no version), fetching all advisories + their ECOSYSTEM ranges.
3. **Resolve** each SHA pin → release tag via `gh api repos/{action}/tags` (reusing the same authed `gh` the Dependabot path uses; only for actions that *have* advisories, so a handful of calls at most). This repo pins actions by commit SHA — the recommended style — so tag comparison alone finds nothing without this step.
4. **Compare** the resolved version against the advisory ranges with scalibr's semver comparator (the same one OSV uses for npm/Go/crates), wrapped behind a small shared `internal/version` helper.

**Policy:** confirmed-affected pins report at the advisory severity and gate per `--fail-on`. A pin that can't be resolved to a comparable version (untagged commit, floating branch) is surfaced as an `unverified pin` at unknown severity — visible but non-blocking — rather than dropped or guessed. Findings flow through the existing allowlist, dedupe, and text/json/sarif reporting unchanged. Backend failures warn-and-continue during an all-sources run, so an OSV/`gh` hiccup won't break the gate.

## Testing

- 15 unit tests over the pure matcher logic (range evaluation incl. `fixed`/`last_affected`/enumerated versions, SHA→tag selection, severity, finding construction), plus 3 covering the extracted `internal/version` helpers.
- Live end-to-end:
  - Synthetic `tj-actions/changed-files@<v45.0.0 SHA>` → resolved to 45.0.0 → **HIGH `GHSA-mrrh-fwg8-r2c3`**, `fixed_in: 46.0.1`, `--fail-on high` exits 1 (gate works).
  - `actions/download-artifact` and `j178/prek-action` pins resolved and correctly dropped as patched (true negatives).
  - `github/codeql-action` correctly surfaced as `unverified` (its pin maps only to a `codeql-bundle-*` tag, a different version namespace than the advisory's `3.x`).
- `go build` / `go vet` / `gofmt` clean.

## CI gating

The audit workflow's PR `paths:` filter now includes `.github/workflows/**` and `.github/actions/**`, so a PR that bumps a pinned action to a vulnerable version trips the gate on that PR — not just the nightly run.

---
## Summary by cubic
Add `ods audit --actions` to scan pinned GitHub Actions in `.github/workflows` and composite actions in `.github/actions` against OSV.dev, failing on known‑vulnerable pins. The audit workflow now gates PRs that change workflows or composite actions, and `--actions` runs by default with `ods audit`.

- **New Features**
  - New `--actions` selector; included in the default all-sources run.
  - Scans workflows and composite actions, extracts `owner/repo@ref`, queries OSV by action, resolves SHA pins to tags via `gh`, and matches ECOSYSTEM ranges locally with `osv-scalibr`.
  - Fails on confirmed vulnerable pins per `--fail-on`; unresolvable pins are reported as “unverified” (unknown severity, non-blocking).
  - CI: PR paths now include `.github/workflows/**` and `.github/actions/**` so action bumps are gated on the PR.

- **Bug Fixes**
  - Parses `gh api --paginate` output in both merged and concatenated JSON modes to keep tag resolution reliable.
  - Range evaluation tolerates duplicate/equal-version events and evaluates correctly.
  - Returns an error when all OSV.dev advisory queries fail, preventing false “clean” results on outages.
  - Backend failures are fatal only when no lockfile scan is running; otherwise they warn-and-continue.
  - Warns on unparseable `.github/actions/*/action.yml` instead of silently skipping composite action dependencies.

<sup>Written for commit e5bdb1dea1365bfa97dc23b8113dd933e27b690e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12618?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -14,7 +14,10 @@ on:
       - "pyproject.toml"
       - "uv.lock"
       - ".github/dependabot.yml"
-      - ".github/workflows/audit.yml"
+      # Any workflow or composite-action change can bump a pinned action to a
+      # vulnerable version, so gate the PR that makes the change (not just nightly).
+      - ".github/workflows/**"
+      - ".github/actions/**"
       - "tools/ods/**"
   schedule:
     # Nightly at 09:00 UTC.
--- a/tools/ods/cmd/audit.go
+++ b/tools/ods/cmd/audit.go
@@ -14,6 +14,7 @@ type AuditOptions struct {
 	Web        bool
 	Python     bool
 	Dependabot bool
+	Actions    bool
 	Format     string
 	FailOn     string
 	IgnoreURL  string
@@ -28,10 +29,11 @@ func NewAuditCommand() *cobra.Command {
 		Short: "Audit dependencies for known vulnerabilities",
 		Long: `Audit dependencies for known vulnerabilities.
 
-Scans the JavaScript (bun.lock) and Python (uv.lock) lockfiles via osv-scanner
-and open GitHub Dependabot security alerts. With no selector flags, all sources
-are audited. Accepted advisories are suppressed via an allowlist fetched from S3
-at runtime, so releases can be unblocked without a code change.
+Scans the JavaScript (bun.lock) and Python (uv.lock) lockfiles via osv-scanner,
+open GitHub Dependabot security alerts, and the GitHub Actions pinned in
+.github/workflows against OSV.dev. With no selector flags, all sources are
+audited. Accepted advisories are suppressed via an allowlist fetched from S3 at
+runtime, so releases can be unblocked without a code change.
 
 Exits non-zero when an unignored finding at or above --fail-on remains, which is
 how it gates deploys.`,
@@ -44,6 +46,7 @@ how it gates deploys.`,
 	cmd.Flags().BoolVar(&opts.Web, "web", false, "Audit web/JS dependencies (bun.lock)")
 	cmd.Flags().BoolVar(&opts.Python, "python", false, "Audit Python dependencies (uv.lock)")
 	cmd.Flags().BoolVar(&opts.Dependabot, "dependabot", false, "Audit open Dependabot security alerts")
+	cmd.Flags().BoolVar(&opts.Actions, "actions", false, "Audit GitHub Actions pinned in .github/workflows")
 	cmd.Flags().StringVar(&opts.Format, "format", "text", "Output format: text, json, or sarif")
 	cmd.Flags().StringVar(&opts.FailOn, "fail-on", "critical", "Minimum severity that fails the audit: critical, high, moderate, or low")
 	cmd.Flags().StringVar(&opts.IgnoreURL, "ignore-url", audit.DefaultIgnoreURL, "S3 URL of the advisory allowlist")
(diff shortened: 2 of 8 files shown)


jmelahman:
Thanks for the reviews — addressed all findings:

- **P1 — `gh api --paginate` tag parsing** (greptile, cubic): `resolveActionTags` now decodes the output as a *stream* of JSON arrays (`parseTagPages`), so it works whether gh merges pages into one array (current — verified against `github/codeql-action`'s 554 tags on gh 2.93.0) or concatenates one array per page (older gh). Not an active bug on current gh, but this removes the version fragility. (`261412e`)
- **P2 — sort comparator strict-weak-ordering** (greptile, cubic): added an equality early-return so duplicate `introduced:0` events can't break `sort.SliceStable`. (`261412e`)
- **P2 — OSV outage masked as clean** (cubic): `scanActions` now returns an error when *all* advisory queries fail, instead of a no-findings result. (`261412e`)
- **P2 — `--dependabot --actions` could pass with both backends failing** (cubic): replaced `isOnlySource` with a lockfile-gate rule — a Dependabot/Actions failure is fatal whenever no lockfile scan is the primary gate. (`261412e`)
- **P2 — malformed composite `action.yml` silently skipped** (cubic): `compositeSteps` now returns an error for unparseable YAML (distinct from a valid non-composite action), so a broken file warns instead of dropping its nested `uses` from the audit. (`44c9cc2`)

Added unit tests for the pagination parser (single + concatenated), the sort tie-break, and the composite parse-error path.

justin-tahara (approved):

PR #12629, opened 2026-07-01
feat(external apps): Apply granted scopes to external apps

Danelegend:
## Description
Persist the scopes a user actually granted, per user, on the OAuth callback.

## How Has This Been Tested?
Manual + Tests

## Additional Options

Fixes ENG-4261
https://linear.app/onyx-app/issue/ENG-4261/persist-per-user-granted-hubspot-scopes-after-oauth-callback

---
## Summary by cubic
Apply per-user OAuth scope grants to external apps and persist them as `granted_scopes` on each credential for accurate permission checks. Implements ENG-4261 with HubSpot support, Slack nesting, and a shared parser for consistent scope extraction; also ensures `updated_at` is bumped on credential updates.

- **New Features**
  - On OAuth callback, extract and save the user’s granted scopes; refresh/credential-form saves omit this to preserve existing grants.
  - `upsert_external_app_user_credential` accepts `granted_scopes`: provided list overwrites; `None` clears; `UNSET` leaves unchanged; `updated_at` is explicitly set on conflict.
  - Provider extraction: default reads space/comma-delimited `scope`; Slack reads `authed_user.scope`; HubSpot fetches via the token-info endpoint (best-effort, 15s timeout). Shared parser accepts strings/lists and returns `None` when absent.

- **Migration**
  - Run Alembic migration `582269841f06`. Existing rows remain `NULL` (grant unknown).

<sup>Written for commit 9ff83a3e5635a8218176f7fdef153510245f8b76. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12629?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/db/external_app.py
+++ b/backend/onyx/db/external_app.py
@@ -5,6 +5,7 @@ from uuid import UUID
 from uuid import uuid4
 
 from sqlalchemy import delete
+from sqlalchemy import func
 from sqlalchemy import select
 from sqlalchemy.dialects.postgresql import insert as pg_insert
 from sqlalchemy.orm import selectinload
@@ -469,6 +470,7 @@ def upsert_external_app_user_credential(
     external_app_id: int,
     user_id: UUID,
     user_credentials: dict[str, Any],
+    granted_scopes: list[str] | None = None,
     resolve_masked_values: bool = False,
 ) -> ExternalAppUserCredential:
     """Create or replace the calling user's credentials for the app, and commit.
@@ -476,6 +478,11 @@ def upsert_external_app_user_credential(
     ``OnyxError(NOT_FOUND)`` if the app doesn't exist. ``resolve_masked_values``
     is for user form submissions that may echo masked display values; internal
     OAuth writers should store provider-returned values as-is.
+
+    ``granted_scopes`` is the authoritative OAuth grant, captured at connect
+    time. ``None`` means "don't touch it": an existing grant is preserved, so
+    the refresh and credential-form paths (which don't re-derive scopes) can't
+    clobber it. Pass a list only from the connect flow.
     """
     app = get_external_app_by_id(db_session, external_app_id)
     if app is None:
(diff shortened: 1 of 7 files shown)


backend/onyx/db/external_app.py
```diff
@@ -469,13 +470,20 @@ def upsert_external_app_user_credential(
     external_app_id: int,
     user_id: UUID,
     user_credentials: dict[str, Any],
+    granted_scopes: list[str] | None | UnsetType = UNSET,
```
wenxi-onyx, line 473: Seems a bit more complex than necessary, but not sure of a better alternative. As I understand it
- empty list = no granted scopes --> we have creds, but it's not allowed to do anything?
- none = could not determine --> does this mean we assume all scopes and the full endpoint catalog is available?

Can this be condensed down to just empty list = could not determine? Is it even possible to auth with 0 granted scopes?

backend/onyx/external_apps/providers/base.py
```diff
@@ -55,6 +56,26 @@ def token_response_error(http_response: requests.Response, body: Any) -> str | N
     return None
 
 
+# Handles space-delimited (RFC 6749 §3.3) and comma-delimited (GitHub, Slack)
+# scope strings with one rule; scope tokens contain neither whitespace nor commas.
+_SCOPE_DELIMITERS = re.compile(r"[,\s]+")
```
wenxi-onyx: nit: I'm personally a big regex hater. I can't look at the pattern and immediately understand it 😓 (and I don't want to learn :big_grin:).

Any opinions on just using regular string.split(" ") --> if actually split, then it parsed, otherwise try split(","). 

No biggie if you want to just use regex. It's reasonable that 1) coding agent AX matters more and they can understand regex fine 2) it's more concise

backend/onyx/external_apps/providers/base.py
```diff
@@ -55,6 +56,26 @@ def token_response_error(http_response: requests.Response, body: Any) -> str | N
...
+
+
+def parse_granted_scopes(raw: Any) -> list[str] | None:
```
wenxi-onyx: can input be more strictly typed?

backend/onyx/external_apps/providers/base.py
```diff
@@ -55,6 +56,26 @@ def token_response_error(http_response: requests.Response, body: Any) -> str | N
...
+    list. Returns ``None`` — never ``[]`` — for an absent/empty signal, so
+    callers record "grant unknown" rather than an (impossible) empty grant.
+    """
+    if isinstance(raw, str):
+        scopes = _SCOPE_DELIMITERS.split(raw.strip())
+    elif isinstance(raw, list):
+        scopes = [str(scope).strip() for scope in raw]
+    else:
+        return None
+    scopes = [scope for scope in scopes if scope]
```
wenxi-onyx: claude suggested this deduplication. feel free to ignore

```
if isinstance(raw, str):
    parts = _SCOPE_DELIMITERS.split(raw)
elif isinstance(raw, list):
    parts = [str(scope) for scope in raw]
else:
    return None
scopes = [stripped for part in parts if (stripped := part.strip())]
```

backend/onyx/external_apps/providers/hubspot.py
```diff
@@ -17,7 +18,19 @@
...
 from onyx.external_apps.providers.base import OnyxManagedExtApp
 from onyx.external_apps.providers.base import OrgCredentialField
+from onyx.external_apps.providers.base import parse_granted_scopes
 from onyx.external_apps.providers.base import token_response_error
+from onyx.utils.logger import setup_logger
+
+logger = setup_logger()
+
+# HubSpot's token response omits the granted scopes, so the actual per-user
+# grant (which diverges under `optional_scope`, ENG-4260) is read from the
```
wenxi-onyx: nit: pref no ref to linear ticket

wenxi-onyx (approved):

PR #12631, opened 2026-07-01
fix: do not mirror encrypted KV secrets to Redis in plaintext

nmgarza5:
## Description

Security fix found during a caching audit.

- **Bug:** `PgRedisKVStore.store()` wrote `json.dumps(val)` to the cache backend (Redis) regardless of the `encrypt` flag, and `load()` re-populated it with the decrypted value. Secrets encrypted at rest in Postgres sat in Redis as plaintext for up to 24h.
- **Exposure:** Google OAuth app credentials, service-account private keys, and per-credential OAuth tokens (`google_kv.py`), plus the customer UUID.
- **Fix:** cache only non-encrypted values. Skip the cache write when `encrypt=True` (and drop any stale plaintext entry), and never repopulate the cache from the encrypted column on load. Encrypted keys now always read from Postgres, which is fine since they are read during connector sync, not a per-request hot path.

## How Has This Been Tested?

Added external-dependency-unit tests asserting encrypted values are never written to the cache backend and that a load does not repopulate it. All 11 KV cache-layer tests pass against a local Postgres, ruff and project-wide ty clean.

## Additional Options

### DIFF
--- a/backend/onyx/key_value_store/store.py
+++ b/backend/onyx/key_value_store/store.py
@@ -28,16 +28,19 @@ class PgRedisKVStore(KeyValueStore):
         return self._cache
 
     def store(self, key: str, val: JSON_ro, encrypt: bool = False) -> None:
-        # Not encrypted in Cache backend (typically Redis), but encrypted in Postgres
+        # Encrypted values are protected at rest in Postgres; never mirror them to the
+        # cache backend (typically Redis) in plaintext. Cache only non-encrypted values,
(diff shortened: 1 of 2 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
Closing this. Per the discussion, the right fix is to remove the KV store's `encrypt` flag entirely and move these secrets to their own Postgres table, rather than special-casing encrypted values inside a cache-fronted store. Doing that in a follow-up PR.

PR #12646, opened 2026-07-02
feat: store Google service-account key on the credential row

nmgarza5:
## Description

Stage 1 of moving Google connector secrets off the global KV store onto the per-connector `credential` row. [Design doc](https://dev-wiki.onyx.app/app/wiki/Engineering%20Projects/Onyx%20Projects/Google%20Credential%20Model%20Refactor/Design.md).

- **Before:** the service-account key was uploaded to a global KV singleton (one per source, mirrored to Redis in plaintext), then copied onto the credential row at creation.
- **After:** the admin supplies the SA key when creating the credential, and it is written straight to the per-connector credential row (already encrypted at rest). `build_service_account_creds` takes the key as a parameter.
- Removes the global SA-key KV keys, their 6 admin endpoints, the get/upsert/delete helpers, and the one-SA-per-source wipe, so an instance can now hold multiple service-account credentials.
- **Frontend:** the admin uploads the key JSON and enters the primary-admin email in one step.
- **No data migration.** Existing service-account connectors already carry the key on their row, so nothing breaks at run time.

Browser-verified end to end on a dev environment: pasted a service-account key into the create form, submitted, confirmed the key landed encrypted on the credential row with the primary admin and uploaded auth method, and revoked it from the same UI.

## How Has This Been Tested?
for the FE changes
Screenshot 2026-07-07 at 12 00 21 PM

Backend ruff + project-wide ty + unit tests. Frontend tsgo + oxlint + oxfmt.

## Additional Options


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

evan-onyx (approved):
this is so exciting 😄

PR #12666, opened 2026-07-02
feat: store the Google OAuth app credential on the credential

nmgarza5:
## Description

Credentials are the only store for connector secrets. **Stacked on #12646.** [Design doc](https://dev-wiki.onyx.app/app/wiki/Engineering%20Projects/Onyx%20Projects/Google%20Credential%20Model%20Refactor/Design.md).

- The OAuth app credential (client_id/secret) is supplied when the credential is created and lives in its `credential_json` next to the tokens it mints. The global slot, its six admin endpoints, and the one-per-source wipe helper are removed.
- No instance default and no fallback: `get_auth_url` and the token callback read only the credential. A credential created before this change rebuilds its app cred from its own token blob (which embeds the client id/secret) and stamps it on itself at next authorize.
- Frontend: the authenticate step offers OAuth (upload the app JSON, then authenticate) and service account side by side; the separate global app-JSON upload step is gone.

## How Has This Been Tested?

Reworked unit tests (7): app cred read off the credential, token-blob reconstruction stamps the credential, hard error when neither exists, PKCE state round-trip. ruff + project-wide ty + FE types/lint/format clean.

## Additional Options

- [ ] [Optional] Please cherry-pick this PR to the latest release version.
- [x] [Optional] Override Linear Check

PR #12668, opened 2026-07-03
feat: allow multiple Google Drive and Gmail connectors

nmgarza5:
## Description

Lets an admin create and run multiple Google Drive and Gmail connectors on one instance. **Stacked on #12666.** [Design doc](https://dev-wiki.onyx.app/app/wiki/Engineering%20Projects/Onyx%20Projects/Google%20Credential%20Model%20Refactor/Design.md).

The credential-setup UI blocked a second Google credential: the auth section short-circuited to "Authentication Complete" as soon as any credential for the source existed. It now confirms only a credential created in the current session, so more can be added. Gmail is routed through the same credential-list + "Create New" modal Drive already uses, instead of a single inline flow.

Backend: drop the `cleanup_gmail_credentials` wipe from the credential-create paths (it deleted every Gmail credential on create). The helper stays for the Gmail app-credential DELETE route.

The run path, indexing, and permission sync are already per-cc_pair, so multiple connectors index independently.

## How Has This Been Tested?

Full E2E on a dev instance against the real onyx-test.com Workspace, via the admin UI end to end. Final state: five Google Drive connectors across four credential rows, all indexing independently.

- Multi-credential UX: the old any-credential-exists short-circuit is gone. Credentials created repeatedly through the modal (per-session `justCreated` panel, list reopens cleanly). Four credential rows coexisted: two rows from one service-account key, one row from a second key of the same SA, one OAuth credential.
- Two connectors on one credential, scoped to different folders: both indexed, 15 docs each, fully disjoint document sets (verified in `document_by_connector_credential_pair`, zero overlap).
- Separate credential rows (same SA, different keys) each drove their own connector to successful indexing (5 docs each from distinct fixture folders).
- OAuth path: app cred stored on the credential (no global slot) went through authorize -> Google consent -> callback; tokens landed next to the app cred and the connector indexes the consenting user's Drive (55+ docs).
- Deletion independence: deleting connectors sharing a credential left the credential and sibling connectors untouched (guards the `cleanup_gmail_credentials`-style wipe removal).
- Negative paths: a structurally-valid-but-fake SA key is rejected at connector validation (PEM parse), and a real SA without domain-wide delegation is rejected with the typed missing-scopes error. Clear toasts, not 500s.

Automated: external-dependency test confirms a second Gmail credential does not delete the first; unit suite, ruff, project-wide ty, FE types/lint all pass.

## Additional Options

### DIFF
--- a/web/src/app/admin/connectors/[connector]/AddConnectorPage.tsx
+++ b/web/src/app/admin/connectors/[connector]/AddConnectorPage.tsx
@@ -34,7 +34,6 @@ import {
 } from "@/lib/connectors/connectors";
 import { useSettings } from "@/lib/settings/hooks";
 import Modal from "@/refresh-components/Modal";
-import { GmailMain } from "@/app/admin/connectors/[connector]/pages/gmail/GmailPage";
 import {
   useGmailCredentials,
   useGoogleDriveCredentials,
@@ -56,6 +55,7 @@ import {
 import { CreateStdOAuthCredential } from "@/components/credentials/actions/CreateStdOAuthCredential";
 import { Spinner } from "@/components/Spinner";
 import { Button } from "@opal/components";
+import { Section } from "@opal/layouts";
 import { deleteConnector } from "@/lib/connector";
 import ConnectorDocsLink from "@/components/admin/connectors/ConnectorDocsLink";
 import Text from "@/refresh-components/texts/Text";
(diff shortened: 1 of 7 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

web/src/app/admin/connectors/[connector]/pages/gdrive/Credential.tsx
```diff
@@ -72,56 +37,18 @@ export const DriveAuthSection = ({
...
-                connectorAssociated,
-                existingCredential,
-                refreshCredentials
-              );
-            }}
-          >
-            Revoke Access
-          </Button>
-        </div>
+      <div className="mt-4 w-full space-y-1 rounded-sm border border-border-02 bg-background-tint-02 px-4 py-3">
```
Comment (evan-onyx) [inline web/src/app/admin/connectors/[connector]/pages/gdrive/Credential.tsx]: isn't raw div sus

web/src/app/admin/connectors/[connector]/pages/gmail/Credential.tsx
```diff
@@ -83,56 +43,18 @@ export const GmailAuthSection = ({
...
-              Revoke Access
-            </Button>
-            {buildMode && onCredentialCreated && (
-              <Button onClick={() => onCredentialCreated(existingCredential)}>
-                Continue
-              </Button>
-            )}
-          </Section>
-        </div>
+      <div className="mt-4 w-full space-y-1 rounded-sm border border-border-02 bg-background-tint-02 px-4 py-3">
```
Comment (evan-onyx) [inline web/src/app/admin/connectors/[connector]/pages/gmail/Credential.tsx]: :o

web/src/app/admin/connectors/[connector]/AddConnectorPage.tsx
```diff
@@ -534,116 +533,109 @@ export default function AddConnector({
...
+                  showIfEmpty
+                  accessType={formikProps.values.access_type}
+                  defaultedCredential={currentCredential!}
+                  credentials={credentials}
+                  editableCredentials={editableCredentials}
+                  onDeleteCredential={onDeleteCredential}
+                  onSwitch={onSwap}
+                />
+                {!createCredentialFormToggle && (
+                  <div className="mt-6 flex gap-4">
```
Comment (evan-onyx) [inline web/src/app/admin/connectors/[connector]/AddConnectorPage.tsx]: hrmhrmhrm

evan-onyx (approved):

nmgarza5:
@greptile

nmgarza5:
@greptile

web/src/app/admin/connectors/[connector]/pages/gdrive/Credential.tsx
Comment (nmgarza5) [reply web/src/app/admin/connectors/[connector]/pages/gdrive/Credential.tsx]: nah you right. I glossed over these changes becuase i figured a full refactor of these pages was coming soon but I should be better. Thanks for calling out. i will add Sections and do what i can to minimize the divs

PR #12706, opened 2026-07-06
fix(helm): v4.3 chart deploy blockers — sandbox-proxy IPv6 netpol fail-closed + duplicate server_tokens nginx crash-loop

### LINKED ISSUE #12675: Helm: nginx fails to start, `server_tokens` directive duplicated between `upstreams.conf` snippet and ingress-nginx controller
### Describe the bug

When deploying the Onyx Helm chart with the bundled ingress-nginx controller, the nginx controller pod enters a `CrashLoopBackOff` and never becomes ready. Every reload fails with:

```
nginx: [emerg] "server_tokens" directive is duplicate in /etc/nginx/custom-snippets/upstreams.conf:2
nginx: configuration file /tmp/nginx/nginx-cfg... test failed
```

The root cause is that the chart's `onyx-nginx-conf` ConfigMap defines `server_tokens off;` inside the `upstreams.conf` snippet, which is included into the `http` context via the controller's `http-snippet`. The ingress-nginx controller already emits its own `server_tokens` directive in the same `http` context by default, so nginx sees the directive twice and refuses to load the configuration.

Because the frontend proxies API calls through nginx, the web UI shows "We encountered an issue / problem loading your Onyx settings" until the controller is healthy.

### To reproduce

1. Install the Onyx Helm chart with the default bundled nginx (`nginx.enabled: true`).
2. Wait for pods to come up.
3. Observe `onyx-nginx-controller` in CrashLoopBackOff.
4. Check logs:
   ```
   kubectl logs <onyx-nginx-controller-pod> -n <namespace> --previous
   ```
   → repeated `"server_tokens" directive is duplicate` emerg errors.

### Expected behavior

The nginx controller should start cleanly with the default chart values, without a duplicate `server_tokens` directive.

### Suggested fix

Remove `server_tokens off;` from the `upstreams.conf` snippet in the `onyx-nginx-conf` ConfigMap (the `http`-context include). The controller already manages `server_tokens`, so the snippet's copy is redundant. Alternatively, move the directive into the `server` context (`server.conf`) where it does not collide with the controller default, or expose a values key to override the snippet content.

### Current workaround

Editing the `onyx-nginx-conf` ConfigMap to delete the `server_tokens off;` line, then restarting the controller deployment, resolves the issue — but the change is reverted on every `helm upgrade`.

### Environment

- Onyx chart: (fill in your chart version)
- Kubernetes: v1.35
- Bundled ingress-nginx controller image: `registry.k8s.io/ingress-nginx/controller:v1.13.3`
- Install method: `helm install` / `helm upgrade` with a custom values file

justin-tahara:
## Description

Two independent chart bugs that both crash-loop pods on a fresh v4.3.0 (chart 0.6.7) single-tenant deployment; bundled per cherry-pick convention.

**1. sandbox-proxy egress NetworkPolicy IPv6 rule fails closed (Craft + VPC CNI enforcement)**
The AWS network-policy agent intermittently fails to program the IPv6 `::/0` ipBlock into the egress eBPF map (EINVAL) — even as a separate egress rule — leaving the map empty and fail-closed; affected sandbox-proxy pods crash-loop unable to reach the api-server for the proxy CA secret. Same failure mode was previously fixed on the managed-cloud raw manifests but the chart template still shipped the rule. Fix: IPv6 rule now gated behind `sandboxProxy.egressAllowIPv6` (default `false`; single-stack IPv4 clusters have no IPv6 egress to gate). Dual-stack clusters must opt in — documented in the values comment.

**2. Duplicate `server_tokens` directive crash-loops the bundled nginx controller**

Fixes #12675.
#12401 added `server_tokens off;` to the chart's custom `upstreams.conf` http-snippet, but ingress-nginx always emits `server_tokens` in the `http{}` block itself, so `nginx -t` fails ("directive is duplicate") and new controller pods crash-loop, wedging the rollout. Untested until now because internal deployments run a hardened standalone nginx instead of the subchart path. Fix: drop the directive from the snippet and set the controller's supported `server-tokens: "false"` config knob instead (same hardening, exactly one directive); bump `onyx.app/nginx-config-version` 5→6 so nginx pods restart on upgrade. Chart bumped to 0.6.8.

## How Has This Been Tested?

- `helm template` (Craft-enabled values, kube 1.33): default render has no `::/0` rule and zero `server_tokens` occurrences in the custom snippet; the controller ConfigMap carries `server-tokens: "false"`; `--set sandboxProxy.egressAllowIPv6=true` restores the full IPv6 rule.
- Both fixes validated live on an affected NP-enforcing EKS deployment: the equivalent netpol patch recovered the crash-looping sandbox-proxy replica (2/2 Ready); the nginx crash-loop reproduces the exact `server_tokens directive is duplicate` error this removes.

## Additional Options

### DIFF
--- a/deployment/helm/charts/onyx/values.yaml
+++ b/deployment/helm/charts/onyx/values.yaml
@@ -1591,6 +1591,12 @@ sandboxPod:
 
 # -- Sandbox egress proxy.
 sandboxProxy:
+  # Allow IPv6 (::/0 minus link-local/IMDS) egress from the proxy. OFF by default: the AWS
+  # VPC CNI network-policy agent intermittently fails to program IPv6 ipBlocks into the
+  # egress eBPF map (EINVAL), leaving it empty and fail-closed (proxy pods crash-loop), and
+  # single-stack IPv4 clusters have no IPv6 egress to gate. Enable ONLY on dual-stack
+  # clusters, where an Egress policy without an IPv6 allow silently drops all IPv6 egress.
+  egressAllowIPv6: false
   replicaCount: 2
   terminationGracePeriodSeconds: 200
   progressDeadlineSeconds: 300
--- a/deployment/helm/charts/onyx/values.yaml
+++ b/deployment/helm/charts/onyx/values.yaml
@@ -297,7 +297,7 @@ nginx:
     # The ingress-nginx subchart doesn't auto-detect our custom ConfigMap changes.
     # Workaround: Helm upgrade will restart if the following annotation value changes.
     podAnnotations:
-      onyx.app/nginx-config-version: "5"
+      onyx.app/nginx-config-version: "6"
 
     # Propagate DOMAIN into nginx so server_name continues to use the same env var
     extraEnvs:
@@ -305,6 +305,11 @@ nginx:
         value: localhost
 
     config:
+      # Hide the nginx version in Server headers/error pages. Set via the
+      # controller's supported knob — do NOT put `server_tokens` in the
+      # http-snippet files: ingress-nginx already emits the directive in the
+      # http{} block, and a duplicate fails `nginx -t` (config reload/startup).
+      server-tokens: "false"
       # Expose DOMAIN to the nginx config and pull in our custom snippets
       main-snippet: |
         env DOMAIN;
--- a/deployment/helm/charts/onyx/Chart.yaml
+++ b/deployment/helm/charts/onyx/Chart.yaml
@@ -5,7 +5,7 @@ home: https://www.onyx.app/
 sources:
   - "https://github.com/onyx-dot-app/onyx"
 type: application
-version: 0.6.7
+version: 0.6.8
 appVersion: latest
 annotations:
   category: Productivity
--- a/deployment/helm/charts/onyx/templates/nginx-conf.yaml
+++ b/deployment/helm/charts/onyx/templates/nginx-conf.yaml
@@ -14,9 +14,6 @@ metadata:
   name: onyx-nginx-conf
 data:
   upstreams.conf: |
-    # Don't expose the nginx version in the Server header or on error pages
-    server_tokens off;
-
     upstream api_server {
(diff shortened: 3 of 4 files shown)


rohoswagger (approved):

deployment/helm/charts/onyx/values.yaml
```diff
@@ -297,14 +297,19 @@ nginx:
...
     extraEnvs:
       - name: DOMAIN
         value: localhost
 
     config:
+      # Hide the nginx version in Server headers/error pages. Set via the
+      # controller's supported knob — do NOT put `server_tokens` in the
+      # http-snippet files: ingress-nginx already emits the directive in the
+      # http{} block, and a duplicate fails `nginx -t` (config reload/startup).
+      server-tokens: "false"
```
jmelahman, line 312: What does this do? Can we keep this `true`? Should only need to update `deployment/helm/charts/onyx/templates/nginx-conf.yaml` and release a new version of the chart I think.

justin-tahara: It's ingress-nginx's config knob for nginx's `server_tokens` directive — it controls whether the nginx version string appears in the `Server` response header and on error pages. `"false"` = hide the version, which was the point of #12401; setting it `true` would re-expose it.

You're right that the crash-loop fix itself only needs the `nginx-conf.yaml` change — the controller has defaulted `server-tokens` to false since v1.12, so the rendered config emits `server_tokens off;` either way (verified in a full chart render with this values entry removed). I kept the explicit `"false"` for two reasons: it pins a security-relevant behavior instead of inheriting whatever the subchart default is that version (upstream already flipped this default once, at v1.12), and it gives the `do-not-put-this-in-the-snippet` warning comment a natural home so the #12401 mistake doesn't get reintroduced. Happy to drop the values line if you'd rather keep the diff minimal — behavior is identical today either way.

jmelahman (approved):

PR #12713, opened 2026-07-06
fix(voice): gate silent audio out of chunked STT to stop hallucinated transcripts

justin-tahara:
## Description

Voice recordings with an open-but-silent microphone produce transcripts full of hallucinated text — most notoriously Korean broadcast sign-offs (`MBC 뉴스 이덕영입니다`), YouTube caption disclaimers, and repeated `you` tokens. Confirmed hitting many users across multiple deployments; users end up sending these hallucinations as chat messages.

Root cause: Whisper-style STT models hallucinate training-data boilerplate on silent/near-silent audio instead of returning an empty transcript, and the chunked STT fallback path amplifies this maximally:

- The browser recorder streams raw PCM16 in ~250ms slices with no VAD gating.
- `ChunkedTranscriber` fired an independent transcription API call for every slice (`MIN_CHUNK_BYTES = 1500` ≈ 31ms of PCM), each padded to a 30s window by the model — i.e. ~99-100% silence per call, ~4 API calls per second.
- Language auto-detection re-rolls on every call, so hallucinations flip languages chunk to chunk, and the fragments are string-joined into the live transcript.
- The final full-audio pass transcribes long silent tails, which also hallucinate.

Fix (chunked PCM path only; non-PCM formats keep legacy behavior):

- Transcribe in 3-second windows instead of per-slice (also cuts STT API volume ~12x)
- Drop windows containing no chunk above a speech RMS threshold (~-46 dBFS) without calling the provider — silence can no longer reach the STT model
- Trim leading/trailing silence before the final full-audio transcription; a recording containing only silence now returns an empty transcript

## How Has This Been Tested?

- New unit tests (`test_chunked_transcriber.py`, 8 tests) covering: silent audio never reaches the provider, silent-recording flush returns empty without a provider call, 3s windowing call counts, brief-speech-in-window still transcribes, flush silence trimming bounds, non-PCM legacy threshold, and RMS/trim helpers on synthetic PCM (silence + sine tones)
- Full voice unit suite passes (71 tests)
- Verified the commit cherry-picks cleanly onto `release/v4.0` and the changed file compiles under Python 3.11

## Additional Options

---
## Summary by cubic
Stop hallucinated transcripts from silent microphones by gating silence in chunked STT for `pcm16` and trimming silence in the final pass. This removes bogus text, recovers quiet speech from low‑gain mics via a 2.5x dynamics test, and reduces STT provider calls by ~12x.

- **Bug Fixes**
  - For raw `pcm16`, transcribe in 3s windows and drop windows with no speech (RMS below threshold) without calling the provider.
  - Trim leading/trailing silence before the final pass; keep low‑gain speech when peaks are ≥2.5x above the noise floor; all‑silence returns empty; if trimming removes everything, return prior window transcripts.
  - Keep legacy byte-based chunking for non-PCM formats (e.g., `webm`); behavior unchanged.

<sup>Written for commit bb9e0a66ecbd22b620bf9a08774dc8d23ecba63f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12713?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/server/manage/voice/websocket_api.py
+++ b/backend/onyx/server/manage/voice/websocket_api.py
@@ -48,6 +48,15 @@ PCM_TRANSCRIBE_WINDOW_BYTES = PCM_BYTES_PER_SECOND * 3
 # an order of magnitude above; quiet-room mic noise well below.
 SILENCE_RMS_THRESHOLD = 150.0
 
+# Absolute floor (~-61 dBFS) for the low-gain fallback in
+# _speech_rms_threshold — audio quieter than this everywhere is silence.
+MIN_SPEECH_RMS = 30.0
+
+# Low-gain fallback: quiet audio counts as speech only when its loudest
+# frames stand at least this far above the recording's own noise floor.
+# Speech has strong dynamics (pauses between words); steady mic noise doesn't.
+SPEECH_DYNAMICS_RATIO = 4.0
+
 # Audio kept around detected speech when trimming silence off a full recording
 SILENCE_TRIM_PADDING_SECONDS = 0.5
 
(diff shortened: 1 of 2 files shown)


justin-tahara:
Addressed both review findings in 14ac76b:

**Quiet speech becomes silence** — `flush()` no longer relies solely on the absolute RMS threshold. When a recording never reaches it but its loudest frames stand well above the recording's own noise floor (speech-like dynamics, `SPEECH_DYNAMICS_RATIO`), a relative threshold is used instead, so low-gain microphones still get their speech transcribed on stop. Steady low-level noise has no such dynamics and still returns empty. The *live* window gate intentionally stays on the conservative absolute threshold — a 3s window is too small a sample for reliable relative detection, and the cost there is only that a quiet speaker's live preview lags until stop, not lost text.

**Final pass replaces good chunks** — when silence trimming leaves nothing, `flush()` now returns the accumulated window transcripts instead of `""`. For the remaining scenario (trimmed audio transcribes to a worse result than the windows): the trimmed audio is speech-bearing by construction now, which is the same situation as any normal short recording, and preferring the single full-audio pass over stitched window fragments is the pre-existing accuracy tradeoff — left unchanged.

Both cases are covered by new unit tests.

justin-tahara:
Addressed in bb9e0a66ec: lowered `SPEECH_DYNAMICS_RATIO` from 4x to 2.5x (~8dB SNR), which covers the quiet-speech-over-steady-noise-floor case from the comment (floor ~35-40 RMS with peaks ~90-100 now passes), with a unit test for exactly that scenario.

On the residual boundary: any energy-based gate has one. Below ~8dB SNR *and* below the absolute ~-46dBFS threshold, STT output is unreliable regardless, and the failure mode here is deliberately chosen — an empty transcript is visible and recoverable (re-record closer to the mic), while hallucinated text silently sent as a chat message is not. That tradeoff is now documented on the constant.

justin-tahara:
Re: the repeated "Quiet Speech Still Drops" finding — this is the same boundary objection as the previous round, restated with lower numbers (now peaks ~70 over floor ~50, i.e. ~1.4x). Not changing the threshold further, deliberately:

- At peak RMS 70 over a 50 floor, the speech energy above the noise is ~0dB SNR. STT models don't reliably transcribe that audio; they hallucinate on it — which is precisely the bug this PR fixes. Loosening the gate to admit it trades a visible, recoverable failure (empty transcript, user re-records) for an invisible, unrecoverable one (hallucinated text sent as a chat message).
- Any energy-based gate has a boundary; for every threshold T there exists "speech at T-ε". The line is set at the loosest point that still excludes steady noise (2.5x ≈ 8dB), and the tradeoff is documented on `SPEECH_DYNAMICS_RATIO`.
- In practice browsers apply auto-gain control to `getUserMedia` capture by default, which normalizes sustained real speech well above both thresholds — the remaining corner is an edge case of an edge case.

Happy to revisit with a real VAD (spectral, not energy-based) as a follow-up on main, but that's beyond backport scope.

jmelahman (approved):

PR #12714, opened 2026-07-06
fix(craft): stop idle-sandbox reaper from killing actively-used pods

rohoswagger:
## Description

Fixes two bugs where the idle-sandbox reaper (`cleanup_idle_sandboxes_task`) killed sandbox pods that were actively in use. Two distinct bugs, both fixed here:

1. **Stale heartbeat during turns.** Turns driven through the opencode-serve transport never refreshed `sandbox.last_heartbeat` — the reaper's only idleness signal — so any session doing agent work for >1h on one pod was deterministically reaped mid-turn. The fix refreshes the heartbeat unconditionally at turn start and then at most every 60s while events flow, inside the shared `yield_sandbox_events` generator. That single seam covers **both** interactive turns (`interactive_turns/executor.py`) and headless scheduled turns (`scheduled_tasks/executor.py`), and doesn't depend on the turn-budget < idle-timeout invariant. The subagent stream loop gets the same throttled refresh. The refresh is best-effort on its own short-lived DB session — a heartbeat failure can never kill the turn.

2. **TOCTOU race in the reaper.** The sweep partitioned idle/non-idle once at start, then spent minutes snapshotting before `terminate()` with no re-check — a user who resumed in that window (the originally reported incident) got their pod killed anyway. The fix re-reads the sandbox row immediately before `terminate()` and skips the reap if the heartbeat went fresh **or the status changed** (protects a concurrent re-provision from being clobbered to SLEEPING). Already-taken snapshots are harmless.

Also hardened: `yield_sandbox_events` now closes the transport generator deterministically in a `finally` (the `yield from` → `for` rewrite would otherwise have made opencode's abort-on-disconnect dependent on GC timing).

Supersedes #12708: that turn-start bump lived in `_drive_interactive_turn`, which scheduled turns never cross, and it did not address the TOCTOU race that caused the original incident.

**Known minor items** (flagged for reviewer, deliberately out of scope):
- Webapp preview HTTP traffic and the HMR websocket bridge (`webapp_proxy.py`) still never refresh the heartbeat — a user browsing their preview for >1h without chatting can still be reaped. Same bug class, different seam; needs its own ticket.
- The 6-line throttle block is duplicated between the two stream loops (intentional asymmetry: the subagent endpoint already heartbeats at stream start).

## How Has This Been Tested?

- New external-dependency-unit tests: `test_heartbeat_refresh_mid_sweep_aborts_reap` (TOCTOU: heartbeat refreshed during the snapshot phase → sandbox stays RUNNING, never terminated) and `test_turn_heartbeat_refresh.py` (turn-start refresh proven against real Postgres; per-event throttle behavior verified with the interval monkeypatched).
- Full craft external-dependency suite: 293 passed (1 pre-existing flake in `test_dispatch_uses_skip_locked_to_avoid_dupes`, fails identically on clean main — asserts a global claim count on a shared dev DB).
- Live end-to-end on a real deployment: backdated a running sandbox's heartbeat 2h, sent a message → heartbeat refreshed the moment the turn started streaming; a 100s turn with a DB monitor showed the mid-turn 60s throttle firing (`20:35:52` → `20:36:57`).

## Additional Options

---
## Summary by cubic
Stops the idle-sandbox reaper from killing active pods by refreshing heartbeats during turns and re-checking idleness right before termination. Prevents mid-turn pod kills and lost work.

- **Bug Fixes**
  - Refresh `sandbox.last_heartbeat` at turn start and then every 60s while events stream in `yield_sandbox_events`; same throttle in the subagent loop (turn-start bump comes from the `send_message` endpoint). Best-effort with a short-lived DB session so heartbeat failures never break a turn.
  - In `cleanup_idle_sandboxes_task`, re-read the sandbox row just before `terminate()` and skip the reap if the heartbeat is fresh or the status is no longer `RUNNING`.
  - Close the transport event stream deterministically to keep abort-on-disconnect behavior consistent.
  - Tests added: `test_heartbeat_refresh_mid_sweep_aborts_reap` and `test_turn_heartbeat_refresh.py`.

- **Refactors**
  - Centralized the heartbeat refresh interval in `configs.py` as `SANDBOX_HEARTBEAT_REFRESH_INTERVAL_SECONDS`.

<sup>Written for commit da2221e0416f7077a274e875d5b21012cdc13e78. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12714?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/onyx/server/features/build/session/streaming.py
```diff
@@ -814,6 +848,7 @@ def stream_subagent_turn(
         state = BuildStreamingState(turn_index=0)
 
         # Subagent runs on the parent session's model, not the child's default.
+        last_heartbeat_refresh = time.monotonic()
         for sandbox_event in sandbox_manager.send_subagent_message(
```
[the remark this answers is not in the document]
rohoswagger: Not a gap: the subagent send-message endpoint writes the turn-start heartbeat (`update_sandbox_heartbeat` in `messages.py` stream_generator) before `send_subagent_message` runs — this loop only needs the periodic refresh. Added a comment stating that in da2221e041.

backend/onyx/server/features/build/session/streaming.py
```diff
@@ -490,7 +504,10 @@ def _persist_resolved_id(new_id: str) -> None:
         # opencode session (dropping conversation history).
         _persist_opencode_session_id(db_session, session_id, new_id)
 
-    yield from sandbox_manager.send_message(
+    # The idle reaper keys off last_heartbeat; a turn can outlast the idle timeout.
+    _refresh_sandbox_heartbeat_best_effort(sandbox_id)
+    last_heartbeat_refresh = time.monotonic()
```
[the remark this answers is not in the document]
rohoswagger: Not a gap: the subagent send-message endpoint writes the turn-start heartbeat (`update_sandbox_heartbeat` in `messages.py` stream_generator) before `send_subagent_message` runs — this loop only needs the periodic refresh. Added a comment stating that in da2221e041.

backend/onyx/server/features/build/session/streaming.py
```diff
@@ -490,7 +504,10 @@ def _persist_resolved_id(new_id: str) -> None:
         # opencode session (dropping conversation history).
         _persist_opencode_session_id(db_session, session_id, new_id)
 
-    yield from sandbox_manager.send_message(
+    # The idle reaper keys off last_heartbeat; a turn can outlast the idle timeout.
+    _refresh_sandbox_heartbeat_best_effort(sandbox_id)
```
wenxi-onyx, line 508: `_drive_interactive_turn()` in `executor.py` is a cleaner place to implement IMO since a db session is already available and it's also where we do `update_session_activity()` --> no longer need the helper function that creates a db session

backend/onyx/server/features/build/session/streaming.py
```diff
@@ -500,6 +517,19 @@ def _persist_resolved_id(new_id: str) -> None:
         on_opencode_session_resolved=_persist_resolved_id,
         should_interrupt=should_interrupt,
     )
+    try:
+        for sandbox_event in event_stream:
```
wenxi-onyx, line 521: I don't think periodic refresh within the turn is necessary. The turn should not take longer than a few minutes at max.

backend/onyx/server/features/build/session/streaming.py
```diff
@@ -814,6 +844,9 @@ def stream_subagent_turn(
         state = BuildStreamingState(turn_index=0)
 
         # Subagent runs on the parent session's model, not the child's default.
+        # Turn-start heartbeat is written by the send-message endpoint; only
+        # the periodic refresh lives here.
```
wenxi-onyx, line 848: Likewise above. What does refreshing this value during the turn get us?

wenxi-onyx (approved):

backend/onyx/server/features/build/session/streaming.py
rohoswagger: Considered that, but `_drive_interactive_turn` only covers interactive turns — scheduled/headless runs drive `session_manager.yield_sandbox_events` directly from `scheduled_tasks/executor.py` and never pass through the interactive executor (that's the coverage gap that #12708 had). The generator is the one seam both paths cross, so the refresh lives there once instead of in two executors that can drift.

On the session: reusing the caller's `db_session` means the heartbeat's `commit()` also commits whatever event-persistence state is in flight on that session at that moment (the two executors have different commit cadences). The dedicated short-lived session keeps the heartbeat write isolated and best-effort — it can never flush partial turn state or poison the caller's transaction on failure. Cost is one pool checkout per minute per active turn.

backend/onyx/server/features/build/session/streaming.py
rohoswagger: Turn-start-only is safe iff turn_duration < idle_timeout, and nothing enforces that invariant:

1. `DEFAULT_INTERACTIVE_TURN_BUDGET_SECONDS` is 30min today, but it's a parameter callers can raise, and `SANDBOX_IDLE_TIMEOUT_SECONDS` is env-configurable — an operator setting the timeout below the budget silently reintroduces mid-turn reaps.
2. Subagent follow-up streams (the other refresh site) have no budget at all — they run as long as the SSE does.
3. The reaper's pre-terminate re-check can't protect a long turn: if nothing refreshes during the turn, the re-check still sees a stale heartbeat and kills the pod mid-work. Periodic refresh is the only thing that makes "actively streaming" visible to the sweep.

The second occurrence in the incident was reaped mid-multi-subagent-turn precisely because in-flight work was invisible to the reaper. Cost is one DB write per minute while events flow; happy to drop it if we'd rather enforce the budget<timeout invariant somewhere explicit instead.

backend/onyx/server/features/build/session/streaming.py
rohoswagger: Same reasoning as the main loop, but stronger here: this stream has no turn budget bounding its duration — the endpoint's start-of-stream heartbeat only protects the first `SANDBOX_IDLE_TIMEOUT_SECONDS`. A subagent stream that outlives it gets reaped mid-stream with nothing else refreshing, and the reaper's pre-terminate re-check doesn't help (it would still see a stale heartbeat).

PR #12729, opened 2026-07-07
feat: canvas perm sync

evan-onyx:
## Description

adds canvas permission syncing. frontend to follow.

## How Has This Been Tested?

added tests (unit and against real canvas instance)

## Additional Options

---
## Summary by cubic
Add enterprise Canvas permission syncing with per-item access for pages, assignments, and announcements, plus group syncing to enforce search access control.

- **New Features**
  - Per-item permissions: pages use course group; announcements honor section targeting; assignments honor student/section/group overrides and `assignment_visibility`.
  - Group sync: emit course, section, and Canvas group IDs; include `canvas-all-users` for public courses when the token can list account users.
  - Slim-doc perm sync: implement `retrieve_all_slim_docs_perm_sync`; wire `canvas_doc_sync` and `canvas_group_sync` into EE scheduling.
  - Validation: `validate_perm_sync` probes Canvas for roster email visibility and account user listing.
  - Config: add `CANVAS_PERMISSION_DOC_SYNC_FREQUENCY` and `CANVAS_PERMISSION_GROUP_SYNC_FREQUENCY` (default 30m); exposed in Docker Compose and Helm.

- **Migration**
  - Use a Canvas token that can read course users with emails; otherwise perm sync fails.
  - To include all authenticated users for public courses, use a token that can list `accounts/self/users`. Otherwise, public courses fall back to course rosters only.

<sup>Written for commit ee605094a38be73f8aaf927f4486d08ec9d09a2d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12729?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/ee/onyx/external_permissions/canvas/access.py
```diff
@@ -0,0 +1,207 @@
...
+from onyx.error_handling.exceptions import OnyxError
+from onyx.utils.logger import setup_logger
+
+logger = setup_logger()
+
+_ACTIVE_ENROLLMENT_STATES = {"active", "invited"}
+_STAFF_ENROLLMENT_TYPES = {
+    "TeacherEnrollment",
+    "TaEnrollment",
+    "DesignerEnrollment",
```
justin-tahara, line 21: For Duo?

backend/ee/onyx/external_permissions/canvas/group_sync.py
```diff
@@ -0,0 +1,131 @@
...
+    tenant_id: str,  # noqa: ARG001
+    cc_pair: ConnectorCredentialPair,
+) -> Generator[ExternalUserGroup, None, None]:
+    connector = CanvasConnector(**cc_pair.connector.connector_specific_config)
+    connector.load_credentials(_credential_json(cc_pair))
+
+    should_emit_all_users = False
+    for course in connector._list_courses():
+        context = build_course_permission_context(connector.canvas_client, course.id)
+        course_emails = set(context.user_id_to_email.values())
```
justin-tahara, line 92: I feel like this naming is a bit confusing. Emails in courses or something clearer?

backend/onyx/connectors/canvas/access.py
justin-tahara, line 1: Do you think we should have ee_* naming convention here?

backend/onyx/connectors/canvas/connector.py
```diff
@@ -38,6 +44,31 @@
 
 logger = setup_logger()
 
+_SLIM_DOC_SYNC_LABEL = "canvas_retrieve_all_slim_docs_perm_sync"
+_ACTIVE_ENROLLMENT_STATES = {"active", "invited"}
```
justin-tahara, line 48: Should we just do enrollment states and then we could potentially have a 3rd state for later on when a student is removed from a class?

justin-tahara (approved):
A lot of smaller nits

backend/ee/onyx/external_permissions/canvas/access.py
evan-onyx: yes

backend/onyx/connectors/canvas/connector.py
evan-onyx: I think if they get removed they just get cooked (i.e. lose access during perm sync) which is fine imo. no need to track them when they're gone, and if they get added back their permissions will be restored on the next perm sync

PR #12730, opened 2026-07-07
fix: scope Drive domain shares to a domain group instead of instance-public

nmgarza5:
## Description

Bug: a Drive doc shared to "everyone at \<domain\>" is stored as `is_public=True`, so every user on the instance sees it. On an instance serving more than one company, domain-internal docs leak across companies. A domain permission for any other domain granted nothing at all.

Fix:
- Map every Drive `domain` permission to a `domain:<domain>` group (`build_domain_group_id`), on both the file and folder paths. Domain shares are never instance-public.
- The Drive group sync populates that group with the domain's real Workspace roster (`admin.directory.users.list`). Membership is Google's own, not inferred from the user's email string, so a user gets the group only if Google actually places them in the Workspace. This is what makes it correct when two orgs share an email domain but live in different Workspaces.
- Cross-domain shares (a doc shared to a partner company's whole domain) grant that domain's group, populated by that domain's own connector sync (Change C connects both Workspaces).

Notes:
- "Anyone with the link" is unchanged (still public).
- The domain group is as fresh as the last group sync, not live, and it needs a service-account connector with directory access.
- Already-indexed docs keep their old permissions until a re-sync runs. After deploy, run a group sync then a permission re-sync.

## Additional Options

### DIFF
--- a/backend/ee/onyx/access/access.py
+++ b/backend/ee/onyx/access/access.py
@@ -11,11 +11,8 @@ from onyx.access.access import (
 from onyx.access.access import _get_acl_for_user as get_acl_for_user_without_groups
 from onyx.access.access import collect_user_file_access
 from onyx.access.models import DocumentAccess
-from onyx.access.utils import build_domain_group_id
-from onyx.access.utils import build_ext_group_name_for_onyx
 from onyx.access.utils import prefix_external_group
 from onyx.access.utils import prefix_user_group
-from onyx.configs.constants import DocumentSource
 from onyx.db.document import get_document_sources
 from onyx.db.document import get_documents_by_ids
 from onyx.db.models import User
@@ -210,18 +207,7 @@ def _get_acl_for_user(user: User, db_session: Session) -> set[str]:
     user_acl = set(prefixed_user_groups + prefixed_external_groups)
     user_acl.update(get_acl_for_user_without_groups(user, db_session))
 
-    # Drive "everyone at <domain>" shares are stored as derived domain groups
-    # (domain:<domain>); membership is the user's own email domain, so no
-    # directory lookup or membership rows are involved.
-    if not is_anonymous and user.email and "@" in user.email:
-        email_domain = user.email.rsplit("@", 1)[-1].strip().lower()
-        if email_domain:
-            user_acl.add(
-                prefix_external_group(
-                    build_ext_group_name_for_onyx(
-                        build_domain_group_id(email_domain), DocumentSource.GOOGLE_DRIVE
-                    )
-                )
-            )
-
+    # Drive "everyone at <domain>" shares resolve through the synced domain group
+    # in prefixed_external_groups above (real Workspace membership), not from the
+    # user's email string.
     return user_acl
--- a/backend/ee/onyx/access/access.py
+++ b/backend/ee/onyx/access/access.py
@@ -207,7 +207,4 @@ def _get_acl_for_user(user: User, db_session: Session) -> set[str]:
     user_acl = set(prefixed_user_groups + prefixed_external_groups)
     user_acl.update(get_acl_for_user_without_groups(user, db_session))
 
-    # Drive "everyone at <domain>" shares resolve through the synced domain group
-    # in prefixed_external_groups above (real Workspace membership), not from the
-    # user's email string.
     return user_acl
--- a/backend/ee/onyx/access/access.py
+++ b/backend/ee/onyx/access/access.py
@@ -11,8 +11,11 @@ from onyx.access.access import (
 from onyx.access.access import _get_acl_for_user as get_acl_for_user_without_groups
 from onyx.access.access import collect_user_file_access
 from onyx.access.models import DocumentAccess
+from onyx.access.utils import build_domain_group_id
+from onyx.access.utils import build_ext_group_name_for_onyx
 from onyx.access.utils import prefix_external_group
 from onyx.access.utils import prefix_user_group
+from onyx.configs.constants import DocumentSource
 from onyx.db.document import get_document_sources
 from onyx.db.document import get_documents_by_ids
 from onyx.db.models import User
@@ -206,5 +209,19 @@ def _get_acl_for_user(user: User, db_session: Session) -> set[str]:
 
     user_acl = set(prefixed_user_groups + prefixed_external_groups)
     user_acl.update(get_acl_for_user_without_groups(user, db_session))
+
+    # Drive "everyone at <domain>" shares are stored as derived domain groups
+    # (domain:<domain>); membership is the user's own email domain, so no
+    # directory lookup or membership rows are involved.
+    if not is_anonymous and user.email and "@" in user.email:
+        email_domain = user.email.rsplit("@", 1)[-1].strip().lower()
+        if email_domain:
+            user_acl.add(
+                prefix_external_group(
+                    build_ext_group_name_for_onyx(
+                        build_domain_group_id(email_domain), DocumentSource.GOOGLE_DRIVE
+                    )
+                )
+            )
 
     return user_acl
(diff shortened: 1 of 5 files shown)


nmgarza5:
@greptile

backend/ee/onyx/external_permissions/google_drive/doc_sync.py
```diff
@@ -187,19 +189,32 @@ def _get_permissions(
...
                 )
-        elif permission.type == PermissionType.DOMAIN and company_domain:
-            if permission.domain == company_domain:
-                public = True
+        elif permission.type == PermissionType.DOMAIN:
+            # "Everyone at <domain>" grants a derived domain group scoped to that
+            # domain's users, never instance-public. The matching user-side token
+            # is derived from the user's own email domain in ee _get_acl_for_user.
+            if permission.domain:
+                domain_groups.add(build_domain_group_id(permission.domain))
```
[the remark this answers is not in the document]
nmgarza5: Fixed in 840b722777. The domain group is now granted only when allowFileDiscovery is not False, so link-only domain shares grant nothing. This also tightens the pre-existing file-path behavior, which made link-only domain shares fully instance-public.

backend/ee/onyx/external_permissions/google_drive/doc_sync.py
```diff
@@ -285,30 +301,32 @@ def get_external_access_for_folder(
...
-            # Domain permission - check if it matches company domain
-            if permission.domain == google_domain:
-                # Only public if discoverable (allowFileDiscovery is not False)
-                # If allowFileDiscovery is False, it's "link only" access
-                is_public = permission.allow_file_discovery is not False
+            # "Everyone at <domain>" grants a derived domain group scoped to that
+            # domain's users, never instance-public. The matching user-side token
+            # is derived from the user's own email domain in ee _get_acl_for_user.
+            if permission.domain:
+                domain_groups.add(build_domain_group_id(permission.domain))
```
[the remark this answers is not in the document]
nmgarza5: Fixed in 840b722777. Same gate on the folder path, restoring the allowFileDiscovery semantics the old is_public branch had.

backend/ee/onyx/external_permissions/google_drive/doc_sync.py
```diff
@@ -187,19 +189,32 @@ def _get_permissions(
...
                     permission,
                 )
         elif permission.type == PermissionType.ANYONE:
             public = True
 
     group_ids = (
         group_emails
+        | domain_groups
         | folder_ids_to_inherit_permissions_from
         | ({drive_id} if drive_id is not None else set())
```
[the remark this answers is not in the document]
nmgarza5: The unprefixed form never persists. The permission-sync consumer prefixes every group id in upsert_document_external_perms (backend/ee/onyx/db/document.py, prefixed_external_groups via build_ext_group_name_for_onyx) before writing external_user_group_ids, which is the contract stated in this function's docstring (add_prefix=False because upsert handles prefixing). Both paths store google_drive_domain:<domain>, matching the user-side token.

nmgarza5:
@greptile review

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

backend/ee/onyx/external_permissions/google_drive/doc_sync.py
```diff
@@ -187,19 +217,18 @@ def _get_permissions(
...
-                logger.warning(
-                    "Permission is type domain but does not match company domain:\n %s",
-                    permission,
-                )
+        elif permission.type == PermissionType.DOMAIN:
+            domain_group = _domain_group_for_permission(
+                permission, company_domain, f"document {doc_id}"
+            )
+            if domain_group:
+                domain_groups.add(domain_group)
```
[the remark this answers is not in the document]
nmgarza5: Intended, and it is the fail-closed direction. We only grant a domain group from an authoritative source now. For the connector's own Workspace that is the real Directory roster. For a domain we do not administer, Google will not let us enumerate the roster, and granting on the email string is exactly the cross-org over-share this PR removes, so we fail closed instead of guessing.

The proper way to cover an unconnected partner domain is to grant domain:X when the user authenticated through X's own IdP, which is authoritative and does not trust the email string. That lands with the multi-IdP work (Change A in the [Design](https://dev-wiki.onyx.app/app/wiki/Engineering%20Projects/Onyx%20Projects/Multi-Company%20Single%20Instance%20Google%20SSO/Design.md) wiki). It unions cleanly with what is here: a user gets domain:X if they are in X's connected Drive roster or they logged in through X's IdP, so there is no gap and no email-string trust.

more often than not, companies that hit this case will have both drive connectors set up so this does not bite in practice.

backend/ee/onyx/external_permissions/google_drive/group_sync.py
```diff
@@ -279,6 +280,26 @@ def _drive_member_map_to_onyx_groups(
...
+    admin_service: AdminService,
+    google_domain: str,
+) -> list[str]:
+    """Every user Google lists in the Workspace domain. This is the real
+    membership behind an "everyone at <domain>" Drive share, so a user is only in
+    the domain group if Google actually places them in the Workspace, not because
+    their email string ends in the domain."""
+    user_emails: set[str] = set()
+    for user in execute_paginated_retrieval(
+        admin_service.users().list,  # ty: ignore[unresolved-attribute]
```
evan-onyx, line 293: do we know what scopes this requires and if it's included in what's already needed for perm sync?

backend/ee/onyx/access/access.py
```diff
@@ -207,4 +207,7 @@ def _get_acl_for_user(user: User, db_session: Session) -> set[str]:
     user_acl = set(prefixed_user_groups + prefixed_external_groups)
     user_acl.update(get_acl_for_user_without_groups(user, db_session))
 
+    # Drive "everyone at <domain>" shares resolve through the synced domain group
```
evan-onyx: not sure this is useful here

backend/ee/onyx/external_permissions/google_drive/doc_sync.py
```diff
@@ -27,6 +28,34 @@
...
+    (allow_file_discovery=False) grant nothing: Google exposes them only to
+    domain users who already hold the link, not to domain-wide search. The Drive
+    group sync populates the domain group with that domain's real Workspace
+    members."""
+    if not permission.domain:
+        logger.warning(
+            "Domain permission without a domain for %s\n %s", entity_desc, permission
+        )
+        return None
+    if permission.domain != own_domain:
```
evan-onyx, line 49: this logic is wonky; can we put this if statement after the following one? it's weird to have if + return, if + no return, if + return and then unconditional return. I don't think we care about the log if allow file discovery is false since the group id is None anyways

evan-onyx (approved):

backend/ee/onyx/access/access.py
nmgarza5: fair. can remove

backend/ee/onyx/external_permissions/google_drive/doc_sync.py
nmgarza5: this is why they pay you the big bucks. master architect and finder of dumb wonkiness that i overlook. thank you sir

backend/ee/onyx/external_permissions/google_drive/group_sync.py
nmgarza5: Yep, already covered. It's `admin.directory.user.readonly`. That's already in the Drive perm-sync scope set

nmgarza5:
@greptile

nmgarza5:
@greptile

PR #12738, opened 2026-07-07
fix(auth): restore chat access for LIMITED service-account keys

wenxi-onyx:
## Description

When the chat APIs were scoped with fine-grained permissions (#11672), LIMITED service-account keys needed a direct `write:chat` grant to keep working. #11763 added that grant — but only for newly created/updated keys (plus a lazy backfill for the Discord service key on regenerate). LIMITED keys created before that fix still have an empty `effective_permissions` column, so `is_limited_user()` blocks them at `current_user` and they fail every `require_permission` check — locking them out of the entire chat surface (`create-chat-session`, `send-message`, `get-chat-session`, …), which is the one thing a LIMITED key exists to do.

Two-part fix — repair the existing broken rows, then remove the class of bug:

- **Backfill migration**: grants `["write:chat"]` to `SERVICE_ACCOUNT` users with role `LIMITED` and empty `effective_permissions`. The anonymous user (role LIMITED but account_type ANONYMOUS) and keys with existing grants are untouched. Downgrade is a no-op: backfilled rows are indistinguishable from keys granted write:chat by the API-key code after the migration, and clearing those would break them.
- **Makes the LIMITED grant role-derived**: `role_derived_permissions` in `permissions.py` is the single place that knows LIMITED service accounts carry `write:chat`, and `recompute_user_permissions__no_commit` computes group grants ∪ role-derived permissions. Previously the grant existed only in the stored column, so every recompute path (SCIM sync, group membership edits) could silently wipe it and every writer had to remember to re-assert it.
- **Removes the now-redundant Discord regenerate-path backfill** and its legacy test — the migration covers existing keys and recompute now derives the grant.
- **`insert_api_key`/`update_api_key`/`regenerate_api_key` now just recompute** (update and rotation unconditionally), so any admin edit or rotation converges a key with stale permissions. Previously rotation only swapped the secret on the same permission-less virtual user — reproduced live with the onyx.app widget key, which kept returning 403 after rotation.
- Refreshes the stale "no permissions" docstring in `test_api_key.py`.

## How Has This Been Tested?

- Exercised the migration's `upgrade()` and `downgrade()` against a real Postgres in a rolled-back transaction with three fixture rows (pre-fix LIMITED key, anonymous user, LIMITED key with custom grants): only the pre-fix key is backfilled, downgrade inverts it exactly.
- New external-dependency-unit tests: recompute derives `write:chat` for LIMITED service accounts (alone and in a mixed batch where a standard user's stale grant is cleared); update/rotation repair a blanked legacy LIMITED key; update/rotation preserve a healthy LIMITED key's scope and a BASIC key's group-derived permissions. All pass.
- Pre-commit (ty, ruff, etc.) passes on all changed files.

## Additional Options

---
## Summary by cubic
Restores chat access for older LIMITED service-account keys by backfilling `write:chat` and deriving it from the role during permission recompute. Key edits and rotation now converge stale keys; removes the Discord-specific backfill and its legacy test.

- **Bug Fixes**
  - Alembic migration grants `["write:chat"]` to `SERVICE_ACCOUNT` users with role `LIMITED` and empty `effective_permissions`; downgrade is a no-op.
  - `recompute_user_permissions__no_commit` derives `write:chat` for LIMITED service accounts and normalizes UUIDs for consistent behavior.
  - `update_api_key`/`regenerate_api_key` recompute to re-assert `write:chat` for LIMITED keys and preserve other roles’ grants.
  - Removes the Discord regenerate-path backfill and its legacy test.
  - Tests cover recompute derivation, update/rotation repair, and role-change permission source swaps; updates a stale test docstring.

<sup>Written for commit aace9bfff5caf0d22824b162ef970de237bb2044. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12738?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/alembic/versions/20f09b642ed0_backfill_write_chat_for_limited_service_.py
+++ b/backend/alembic/versions/20f09b642ed0_backfill_write_chat_for_limited_service_.py
@@ -1,0 +1,58 @@
+"""backfill write chat for limited service accounts
+
+LIMITED service-account keys created before insert_api_key granted them
+write:chat directly (#11763) have an empty effective_permissions column,
+which locks them out of the scoped chat APIs. Backfill the grant
+(write:chat implies read:chat at read time).
+
+Revision ID: 20f09b642ed0
+Revises: 2e0b2b146de1
+Create Date: 2026-07-07 09:46:01.019413
+
+"""
+
+from collections.abc import Sequence
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "20f09b642ed0"
+down_revision = "2e0b2b146de1"
+branch_labels: str | None = None
+depends_on: str | Sequence[str] | None = None
+
+user_table = sa.table(
+    "user",
+    sa.column("role", sa.String),
+    sa.column("account_type", sa.String),
+    sa.column("effective_permissions", postgresql.JSONB),
+)
+
+WRITE_CHAT_PERMS = ["write:chat"]
+
+
+def upgrade() -> None:
+    op.execute(
+        user_table.update()
+        .where(
+            user_table.c.account_type == "SERVICE_ACCOUNT",
+            user_table.c.role == "LIMITED",
+            user_table.c.effective_permissions == sa.cast([], postgresql.JSONB),
+        )
+        .values(effective_permissions=WRITE_CHAT_PERMS)
+    )
+
+
+def downgrade() -> None:
+    op.execute(
+        user_table.update()
+        .where(
+            user_table.c.account_type == "SERVICE_ACCOUNT",
+            user_table.c.role == "LIMITED",
+            user_table.c.effective_permissions
+            == sa.cast(WRITE_CHAT_PERMS, postgresql.JSONB),
+        )
+        .values(effective_permissions=[])
+    )
--- a/backend/alembic/versions/20f09b642ed0_backfill_write_chat_for_limited_service_.py
+++ b/backend/alembic/versions/20f09b642ed0_backfill_write_chat_for_limited_service_.py
@@ -46,13 +46,7 @@ def upgrade() -> None:
 
 
 def downgrade() -> None:
-    op.execute(
-        user_table.update()
-        .where(
-            user_table.c.account_type == "SERVICE_ACCOUNT",
-            user_table.c.role == "LIMITED",
-            user_table.c.effective_permissions
-            == sa.cast(WRITE_CHAT_PERMS, postgresql.JSONB),
-        )
-        .values(effective_permissions=[])
-    )
+    # No-op: backfilled rows are indistinguishable from keys the API-key
+    # code granted write:chat after this migration, and clearing those
+    # would break them. Leaving the grant in place is harmless.
+    pass
--- a/backend/onyx/db/discord_bot.py
+++ b/backend/onyx/db/discord_bot.py
@@ -15,7 +15,6 @@ from onyx.auth.api_key import hash_api_key
 from onyx.auth.schemas import UserRole
 from onyx.configs.constants import DISCORD_SERVICE_API_KEY_NAME
 from onyx.db.api_key import insert_api_key
-from onyx.db.enums import Permission
 from onyx.db.models import ApiKey
 from onyx.db.models import DiscordBotConfig
 from onyx.db.models import DiscordChannelConfig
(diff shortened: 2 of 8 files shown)


backend/alembic/versions/20f09b642ed0_backfill_write_chat_for_limited_service_.py
```diff
@@ -0,0 +1,58 @@
...
+def downgrade() -> None:
+    op.execute(
+        user_table.update()
+        .where(
+            user_table.c.account_type == "SERVICE_ACCOUNT",
+            user_table.c.role == "LIMITED",
+            user_table.c.effective_permissions
+            == sa.cast(WRITE_CHAT_PERMS, postgresql.JSONB),
+        )
+        .values(effective_permissions=[])
```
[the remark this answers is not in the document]
wenxi-onyx: Fixed in 037aa051a4 — downgrade is now a no-op. Backfilled rows are indistinguishable from keys the API-key code granted `write:chat` after this migration ran, so reverting the grant would break legitimately created keys. Leaving it in place is harmless since it's exactly what `insert_api_key` writes.

backend/alembic/versions/20f09b642ed0_backfill_write_chat_for_limited_service_.py
```diff
@@ -0,0 +1,52 @@
...
+)
+
+WRITE_CHAT_PERMS = ["write:chat"]
+
+
+def upgrade() -> None:
+    op.execute(
+        user_table.update()
+        .where(
+            user_table.c.account_type == "SERVICE_ACCOUNT",
```
Weves, line 40: should this just be service accounts?

backend/onyx/db/discord_bot.py
```diff
@@ -107,12 +106,6 @@ def get_or_create_discord_service_api_key(
         new_api_key = generate_api_key(tenant_id)
         existing.hashed_api_key = hash_api_key(new_api_key)
         existing.api_key_display = build_displayable_api_key(new_api_key)
-        # Backfill chat scope on keys provisioned before chat APIs were scoped.
-        service_user = db_session.scalar(
-            select(User).where(User.id == existing.user_id)  # ty: ignore[invalid-argument-type]
-        )
-        if service_user is not None:
-            service_user.effective_permissions = [Permission.WRITE_CHAT.value]
         db_session.flush()
```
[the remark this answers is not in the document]
wenxi-onyx: Intentionally not restoring this — after this PR the repair can never fire in any state the normal flow produces: `insert_api_key` grants the scope at creation, the migration backfills every pre-existing LIMITED key (including this one), and the `recompute_user_permissions__no_commit` hardening closes the one real drift source at its origin. The remaining scenarios (restore from a pre-backfill snapshot, manual column edits) aren't Discord-specific — they'd break every LIMITED key equally, and the fix there is re-running the backfill, not one caller silently repairing one key on every cold start. Keeping the block would duplicate scope-granting logic and mask any future bug that clears the grant.

backend/onyx/db/permissions.py
```diff
@@ -51,13 +53,34 @@ def recompute_user_permissions__no_commit(
...
+            select(User.id).where(  # ty: ignore[no-matching-overload]
+                User.id.in_(uid_list),  # ty: ignore[unresolved-attribute]
+                User.account_type == AccountType.SERVICE_ACCOUNT,
+                User.role == UserRole.LIMITED,
+            )
+        )
+        .scalars()
+        .all()
+    }
+    uid_list = [uid for uid in uid_list if str(uid) not in limited_service_account_ids]
```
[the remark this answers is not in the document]
wenxi-onyx: Fixed in 163f4eb5f5 — the membership check now compares `str(uid).lower()`, matching Postgres's case-insensitive UUID handling on the SQL side.

wenxi-onyx:
@greptile rereview

backend/onyx/db/permissions.py
```diff
@@ -51,13 +53,34 @@ def recompute_user_permissions__no_commit(
...
     """
     if isinstance(user_ids, (UUID, str)):
         uid_list = [user_ids]
     else:
         uid_list = list(user_ids)
 
+    if not uid_list:
+        return
+
+    limited_service_account_ids = {
```
Weves: why do we skip limited users?

backend/alembic/versions/20f09b642ed0_backfill_write_chat_for_limited_service_.py
wenxi-onyx: I believe so - all other user types have had their effective permissions backfilled

backend/onyx/db/permissions.py
wenxi-onyx: As it currently stands, the APIs that limited users may access are defined in code by the `is_limited_user` escape hatch + chat APIs (via `write:chat`). The effective permissions are not defined by groups, as with all other user types, so this function would overwrite `write:chat`.

Weves: can we add a comment?

Also, do limited users get the `write:chat` permission added to them upon creation? They did try and re-create the service account, and obviously that should work (the fix shouldn't juust apply to existing limited service accounts)

Weves (approved):

PR #12758, opened 2026-07-08
feat(auth): SAML providers driven by sso_provider rows

nmgarza5:
## Description

Wires SAML into the multi-provider model from the `sso_provider` store (#12731). Stacked on that PR.

Existing SAML customers reconfigure nothing:
- **One callback, resolved by issuer.** The ACS stays `/auth/saml/callback`, the URL every IdP already posts to. It reads the response's Issuer, finds the provider row whose IdP entityId matches, then validates the signature against that row's cert. Routing on the unverified issuer is safe because validation still requires the real signature.
- **Automatic startup import.** On api_server boot (which has the `SAML_CONF_DIR` mount the migration job lacks), a single-SAML deployment's `settings.json` is seeded into a row. Idempotent, no-op unless `AUTH_TYPE=saml` with no SAML row, concurrency-safe.
- Parametric `/auth/saml/<name>/authorize` for the multi-provider login page, plus a legacy-compatible `/auth/saml/authorize`. Per-provider email-domain gate before the user is created. SP signing fields on the config model, no schema migration.
- Replaces the legacy single-config SAML router. Public-route registration + nginx `/auth/saml` proxy so the callback reaches the API.

Ships dark: no SAML rows means the routes 404. Not yet covered: live-IdP end-to-end, required before merge.

## Additional Options


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

backend/onyx/server/saml_multi.py
```diff
@@ -0,0 +1,347 @@
...
+    return await _process_saml_callback(
+        request,
+        db_session,
+        strategy,
+        user_manager,
+    )
+
+
+@router.post("/callback")
+async def saml_login_callback(
```
justin-tahara: Should we call this saml_login_callback_post to keep things consistent?

justin-tahara (approved):

nmgarza5:
@greptile

backend/onyx/server/saml_multi.py
nmgarza5: Good call, done in 01c3f27a08. Renamed it saml_login_callback_post to match saml_login_callback_get.

nmgarza5:
@greptile

nmgarza5:
@greptile

deployment/data/nginx/app.conf.template
```diff
@@ -49,6 +49,22 @@ server {
         return 200 "ok\n";
     }
 
+    # SAML login and IdP callbacks are backend routes under /auth/saml (legacy
+    # single-config and the DB-backed multi-provider router). Proxy them to the
+    # API like /scim, otherwise the IdP POST falls through to the web server.
+    location ~ ^/auth/saml(/.*)?$ {
```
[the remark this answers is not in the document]
nmgarza5: Fixed in 621aa08eb6. Added the same `^/auth/saml` location proxying to api_server in deployment/helm/charts/onyx/templates/nginx-conf.yaml (before /scim, no rewrite since SAML is at the root path), and bumped onyx.app/nginx-config-version 6 -> 7 in values.yaml so helm upgrade restarts the nginx pods to reload the ConfigMap.

nmgarza5:
@greptile

PR #12792, opened 2026-07-08
fix(chat): render full multi-model comparison in shared chats

nmgarza5:
## Description

Sharing a chat where multiple models were prompted at once only rendered one model's response — the chain walk collapses a user turn's sibling responses down to the latest child.

- Shared multi-model turns now render every response side-by-side, read-only (no select/hide, nothing persists), mirroring what the author saw.
- Provider icons resolve through the user-facing LLM provider list, so each panel shows its real vendor icon instead of the generic fallback.
- The response the flow continued through is highlighted as "Preferred Response". A turn the user never picked from stays unhighlighted, and follow-ups only show the branch that was selected at share time.
- Single-model shared chats now show the answering model in the message footer.

Frontend only — the shared payload already returns every sibling and its packets. The grouping + provider-lookup are extracted so the shared view reuses the live view's logic.

## How Has This Been Tested?
Screenshot 2026-07-08 at 2 58 49 PM

## Additional Options

### DIFF
--- a/web/src/app/app/message/MultiModelPanel.tsx
+++ b/web/src/app/app/message/MultiModelPanel.tsx
@@ -106,11 +106,13 @@ export default function MultiModelPanel({
         rightChildren={
           readOnly ? (
             isPreferred ? (
-              <span className="text-action-link-05 shrink-0 px-2">
-                <Text font="secondary-body" color="inherit" nowrap>
-                  Preferred Response
-                </Text>
-              </span>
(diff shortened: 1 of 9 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

raunakab (approved):

PR #12807, opened 2026-07-09
feat(userfile-port): new-file dual-write + reconcile swap gate

Subash-Mohan:
## Description

- Dual-write newly-indexed/edited user files into the FUTURE (secondary) index during a reindex-port, so it's complete at swap — a failure only flags the file for reconcile and never touches live status, and the metadata-sync drain supplies missing content on a 404 (so a file the backlog copy can't reach still lands).
- Re-add the swap gate on the reconcile flag, scoped to the port's COMPLETED users (mirrors the connector flag's `required_cc_pairs` scoping) — so a flag outside this switchover, or stuck on a non-COMPLETED/undrainable row, can't wedge the swap.
- Carry the "secondary pass" signal through the `IndexingBatchAdapter.post_index` interface (`index_to_secondary`) instead of a per-adapter flag, so the FUTURE write skips the once-only terminal side-effects (status/plaintext/notifications).
- Rename `UserFile.secondary_only_sync_pending` → `secondary_reconcile_pending` (a content-and-ACL FUTURE-dirty bit, broader than the `Document` ACL-only analog); additive rename migration.

## How Has This Been Tested?

External-dependency-unit tests: index-time dual-write fires only with an active target and a failure flags the file while keeping it COMPLETED; the drain's 404 fallback supplies content and clears the flag; the secondary-pass adapter skips terminal side-effects; the swap gate blocks on a set flag, releases when it clears, and ignores a flag on a non-COMPLETED file. Existing persona-sync + port-execution suites updated for the rename/restructure.

## Additional Options

---
## Summary by cubic
Dual-write newly indexed or edited user files to the `FUTURE` index during a reindex port, and hold the swap on a drainable reconcile flag so files aren’t missing at swap and swaps can’t deadlock. Adds a 404 fallback that supplies missing content to `FUTURE`, and renames the flag to `secondary_reconcile_pending`.

- New Features
  - Dual-write to `FUTURE` at index time; failures only set `secondary_reconcile_pending` and never touch live status.
  - Metadata sync now reconciles: on `SecondaryIndexDocumentMissingError`, it supplies content to `FUTURE` and clears the flag when successful.
  - Swap gate restored: waits on reconcile flags only for the port’s COMPLETED users, and ignores flags on non-COMPLETED rows.
  - Adapter API: thread `index_to_secondary` through `post_index`; the secondary pass skips terminal side-effects (status/chunk_count/plaintext/notifications).
  - Removed unused `any_user_file_reconcile_pending` helper.

- Migration
  - Rename `UserFile.secondary_only_sync_pending` → `secondary_reconcile_pending` (and its partial index). Run the Alembic migration; no data rewrite needed.

<sup>Written for commit 6f34bc882a8c9b5d366b670faa6265f6932b612a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12807?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/background/celery/tasks/user_file_processing/tasks.py
+++ b/backend/onyx/background/celery/tasks/user_file_processing/tasks.py
@@ -39,12 +39,15 @@ from onyx.connectors.models import Document
 from onyx.connectors.models import HierarchyNode
 from onyx.db.engine.sql_engine import get_session_with_current_tenant
 from onyx.db.enums import UserFileStatus
+from onyx.db.models import SearchSettings
 from onyx.db.models import UserFile
 from onyx.db.port_attempt import port_backfill_has_pending_work
 from onyx.db.port_orphan_candidate import record_port_orphan_candidates_for_user_file
+from onyx.db.search_settings import active_secondary_port_target
 from onyx.db.search_settings import get_active_search_settings
 from onyx.db.search_settings import get_active_search_settings_list
 from onyx.db.user_file import fetch_user_files_with_access_relationships
+from onyx.db.user_file import mark_user_file_reconcile_pending
 from onyx.document_index.factory import get_all_document_indices
 from onyx.document_index.interfaces_new import MetadataUpdateRequest
 from onyx.document_index.interfaces_new import SecondaryIndexDocumentMissingError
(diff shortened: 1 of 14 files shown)


backend/onyx/background/celery/tasks/user_file_processing/tasks.py
```diff
@@ -392,6 +435,131 @@ def _process_user_file_with_indexing(
...
+            adapter=adapter,
+        )
+    if (
+        result.failures
+        or result.total_docs != len(documents)
+        or result.total_chunks == 0
+    ):
+        raise RuntimeError(
+            f"secondary index write incomplete for user file {user_file_id}: {result}"
+        )
```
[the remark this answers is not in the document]
Subash-Mohan: Safe as-is — sessions use expire_on_commit=False, so the loaded scalars stay valid after detach, and both from_db_search_settings and get_all_document_indices read only scalar columns (no lazy relationship → no DetachedInstanceError).

backend/onyx/background/celery/tasks/user_file_processing/tasks.py
```diff
@@ -321,6 +324,46 @@ def _process_user_file_without_vector_db(
...
+    """Parse a user file's blob into indexable Documents (id/source stamped), plus the ids of
+    any CSVs staged for tabular sections — the caller reaps them after indexing reads them. A
+    load failure reaps its own staged files before re-raising (the caller gets no id list)."""
+    connector = LocalFileConnector(
+        file_locations=[file_id],
+        file_names=[file_name] if file_name else None,
+    )
+    connector.load_credentials({})
+
+    # User files aren't attempt-scoped, so the docfetching staging reapers don't cover them.
```
acaprau, line 342: what is docfetching staging reaper?

backend/onyx/background/celery/tasks/user_file_processing/tasks.py
```diff
@@ -392,6 +435,131 @@ def _process_user_file_with_indexing(
...
+            tenant_id=tenant_id,
+            db_session=db_session,
+        )
+        result = run_indexing_pipeline(
+            embedder=embedder,
+            document_indices=document_indices,
+            ignore_time_skip=True,
+            # skip the content_hash gate, else the PRESENT run's hash no-ops this write
+            index_to_secondary=True,
+            db_session=db_session,
```
acaprau, line 469: if this is the only way to index then it is what it is, nothing we can do right now, but isn't there a way to index without holding a db session the entire time? i swear either @nmgarza5 or @justin-tahara or @evan-onyx or some combination of the three did something about this.

acaprau (approved):

backend/onyx/background/celery/tasks/user_file_processing/tasks.py
evan-onyx: We should really change the code so that the user file indexing adapter doesn't store a db session, but this is fine for now. I'm in favor of leaving a todo

backend/onyx/background/celery/tasks/user_file_processing/tasks.py
Subash-Mohan: It's the standard docfetching cleanup that deletes attempt-staged raw files; user files have no attempt for it to key on, so we track + reap the staged CSVs ourselves.

PR #12815, opened 2026-07-09
fix(auth): stop /me from 500ing for service-account API key users

nmgarza5:
## Description

**Bug:** `cloud.onyx.app/mcp` rejects every valid Onyx API key with `401 invalid_token`.

**Why:** The MCP server authenticates by calling `/me` and treats any non-200 as auth failure. `/me` resolves a home tenant via `get_tenant_id_for_email(user.email)` for every authenticated user. In multi-tenant, API-key (service-account) users have a synthetic `@onyxapikey.ai` email with no `UserTenantMapping` row, so the lookup raises `UserNotExists`, which is unhandled and returns 500. Chat endpoints never call this path, so the same key works for chat.

**Fix:** In multi-tenant, skip the home-tenant/invitation lookup for service accounts and report the current tenant as `team_name`. An API key is tenant-bound, so this is exact. Single-tenant and CE paths are unchanged.

Confirmed in cloud logs: mcp-server `API server rejected MCP auth token with status 500` lines up with api-server tracebacks ending at `user_mapping.py get_tenant_id_for_email` for the reporting tenants.

## Additional Options

### DIFF
--- a/backend/onyx/server/manage/users.py
+++ b/backend/onyx/server/manage/users.py
@@ -1030,23 +1030,31 @@ def verify_user_logged_in(
             else oidc_expiry
         )
 
-    team_name = fetch_ee_implementation_or_noop(
-        "onyx.server.tenants.user_mapping", "get_tenant_id_for_email", None
-    )(user.email)
-
+    team_name: str | None
     new_tenant: TenantSnapshot | None = None
     tenant_invitation: TenantSnapshot | None = None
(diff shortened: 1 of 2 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

justin-tahara (approved):

PR #12820, opened 2026-07-09
refactor(logo): consolidate logo rendering into `useAppLogo` hook

raunakab:
## Description

Consolidates all sidebar logo rendering behind two hooks (`useFoldableAppLogo(folded)` and `useAppLogo()`) and makes `SidebarLayouts.Header.logo` a required `IconFunctionComponent`.

**Key changes:**
- `web/src/lib/app/components.tsx` — `Logo` moved here from `refresh-components/Logo.tsx` as a named export; handles all enterprise white-label settings (`logo_display_style`, `application_name`, `hide_onyx_branding`, powered-by toggle)
- `web/src/lib/app/hooks.tsx` — `useFoldableAppLogo(folded)` wraps `Logo` via `useMemo`; `useAppLogo()` is a folded=false shorthand; neither threads `logoUrl` through callers
- `web/lib/opal/src/layouts/sidebar/components.tsx` — `SidebarHeaderProps.logo` changed from render-prop `(folded?) => ReactNode` to required `IconFunctionComponent`; topbar split into outer spacing div and inner layout div; dead null-guards removed
- `SidebarWrapper` deleted — its single consumer (`StepSidebarWrapper`) now uses `SidebarLayouts.Root/Header/Body` directly
- `AppSidebar`, `AdminSidebar`, craft `SideBar`, `StepSidebarWrapper` — all use `useFoldableAppLogo(folded)`; `memo()` removed from `AppSidebar` (zero-prop component, no effect)
- `SvgOnyxLogoTyped` updated to accept full `IconProps`

## How Has This Been Tested?

TypeScript type check passes. Verified visually that sidebar logo renders correctly in both folded and unfolded states.

## Additional Options


web/src/lib/app/hooks.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { APP_SLOGAN } from "@/lib/constants";
+import { useSettings } from "@/lib/settings/hooks";
+import { Logo } from "@/lib/app/components";
+import type { IconFunctionComponent } from "@opal/types";
+
+export function useAppLogo(): IconFunctionComponent {
+  return useFoldableAppLogo(false);
+}
+
+export function useFoldableAppLogo(folded: boolean): IconFunctionComponent {
```
jmelahman: This seems a bit smelly fwiw. I'm not even sure why the `Logo` should know about `folded`. Feels like something the `SidebarLayouts.Header` should manage and then all of the behavior is applied via CSS (which avoids mutating the DOM)

jmelahman (approved):

web/src/lib/app/hooks.tsx
raunakab: That's a good point. Will update the logic.

jmelahman (approved):
nice

PR #12823, opened 2026-07-09
fix(web): key model-selector selection by model configuration id

jmelahman:
## Description

In the chat model selector, when two LLM providers expose a model with the same name (e.g. two openai-type providers both serving `gpt-4o`, or two Ollama instances both serving `qwen3.5:0.8b`), both entries rendered as selected when only one was, and clicking one could add/remove the other:

- `MultiModelSelector` keyed selection identity on `provider type + model name`, which collides across provider instances of the same type. `isSelected`, `isDisabled`, and the add/remove toggle all misfired on the colliding key.
- `useMultiModelChat` resolved `currentLlm` to an option by `provider + modelName` only, so the selection could snap to the wrong provider's same-named model.

### Fix

- Add `llmOptionKey`, which keys on the unique `model_configuration_id` and only falls back to `provider:modelName` for options without an id, and use it for all selection identity in `MultiModelSelector` / `useMultiModelChat` (and as the render key in `ModelSelectorContent`).
- Carry `modelConfigurationId` on `SelectedModel`.
- When resolving the current LLM descriptor to an option, prefer the candidate whose provider (instance) name matches before falling back to the first provider-type/model-name match.

## How Has This Been Tested?

- Jest unit test (`options.test.ts`) locking the invariant that same-named models from different providers get distinct keys.
- Manually against a dev deployment with two ollama_chat providers both exposing `qwen3.5:0.8b`:
  - Opening the selector from the pill shows exactly one checkmark, on the instance actually selected.
  - Selecting the other provider's same-named model moves the checkmark to that instance (and the auto-expanded group follows).
  - In multi-model mode, the same-named model from each provider can be added side by side, and toggling one off removes only its own pill (previously it removed the other provider's pill).

---
## Summary by cubic
Fixes the chat model selector showing/toggling the wrong item when different providers expose the same model name. Selections now key off unique model configuration IDs so single and multi-model modes behave correctly.

- **Bug Fixes**
  - Added `llmOptionKey` that keys by `model_configuration_id` (fallback to `provider:modelName` only when no id).
  - Switched selection, disabled, toggle logic, and render keys to use `llmOptionKey` in `MultiModelSelector` and `ModelSelectorContent`.
  - Carried `modelConfigurationId` on `SelectedModel`; `useMultiModelChat` now prefers matching provider instance name when resolving the current LLM.
  - Documented that `restoreFromModelNames` and `selectSingleModel` are name-only and resolve same-named models to the first match.
  - Added unit tests to ensure same-named models from different providers get distinct keys.

<sup>Written for commit 1dfd7d6ab001b70282953ed3d895e92620569c77. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12823?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/web/src/hooks/useMultiModelChat.ts
+++ b/web/src/hooks/useMultiModelChat.ts
@@ -7,7 +7,7 @@ import {
 } from "@/sections/model-selector/MultiModelSelector";
 import { LLMOverride } from "@/app/app/services/lib";
 import { LlmManager } from "@/lib/hooks";
-import { buildLlmOptions } from "@/lib/languageModels/options";
+import { buildLlmOptions, llmOptionKey } from "@/lib/languageModels/options";
 
 export interface UseMultiModelChatReturn {
   /** Currently selected models for multi-model comparison. */
@@ -56,16 +56,21 @@ export default function useMultiModelChat(
     if (llmOptions.length === 0) return null;
     const { currentLlm } = llmManager;
     if (!currentLlm.modelName) return null;
-    const match = llmOptions.find(
+    // Two providers can expose a model with the same name; prefer the one
+    // whose provider (instance) name matches the current descriptor.
+    const candidates = llmOptions.filter(
       (opt) =>
         opt.provider === currentLlm.provider &&
         opt.modelName === currentLlm.modelName
     );
+    const match =
+      candidates.find((opt) => opt.name === currentLlm.name) ?? candidates[0];
     if (!match) return null;
     return {
       name: match.name,
       provider: match.provider,
       modelName: match.modelName,
+      modelConfigurationId: match.modelConfigurationId ?? null,
       displayName: match.displayName,
     };
   }, [llmOptions, llmManager.currentLlm]);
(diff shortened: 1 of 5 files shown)


jmelahman:
Re: Greptile's note on `restoreFromModelNames` / `selectSingleModel` — confirmed pre-existing and inherent: chat history persists only model name strings (no `model_configuration_id`), so name-only resolution is the only option there today and same-named models across providers resolve to the first match. Documented the constraint on both functions in 1dfd7d6. Fully fixing it would require persisting the config id with chat messages (backend schema change) — worth a follow-up ticket if multi-provider same-name setups become common.

raunakab (approved):

PR #12824, opened 2026-07-09
test(auth): live multi-provider OIDC login against mock-oauth2-server

nmgarza5:
## Description

Live-IdP integration test for the merged multi-provider OIDC login (#12797). Proves the thing the feature is for: two SSO providers on one instance.

- Runs the full authorization-code flow against a real `navikt/mock-oauth2-server` (discovery fetch, code exchange, userinfo, JIT user provisioning, session issuance), no HTML scraping. Per-login claims come from the server's documented username+claims form post.
- Two providers (two issuers) log in end-to-end and land as distinct users with the right per-provider `oauth_name`. A `companyb.com` identity offered to the `companya.com`-gated provider is rejected (400) and never persisted.
- The login flow is one async test on a single event loop. A sync TestClient opens a fresh loop per request, which strands the async DB connection the callback pools.
- CI: the mock server runs only on the `auth` matrix leg, and the step fails if it never becomes ready so the live tests can't silently skip.

Unit coverage for the router already lives in `test_oidc_multi.py`; this is the live-interop layer.

## How Has This Been Tested?

Both tests pass locally against a real mock-oauth2-server, twice, and skip cleanly when it is absent.

## Additional Options

### DIFF
--- a/.github/workflows/pr-external-dependency-unit-tests.yml
+++ b/.github/workflows/pr-external-dependency-unit-tests.yml
@@ -148,6 +148,21 @@ jobs:
           alembic upgrade head
           alembic heads --verbose
 
+      # A real OIDC server for the live multi-provider login test. Auth leg only.
+      # Fail if it never becomes ready, otherwise the live tests would silently skip.
+      - name: Start mock OIDC server
+        if: matrix.test-dir == 'auth'
+        run: |
+          docker run -d --name mock-oidc -p 8086:8080 \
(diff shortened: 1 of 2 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

justin-tahara (approved):

PR #12830, opened 2026-07-09
feat(auth): multi-provider SSO login page + login-options endpoint

nmgarza5:
## Description

Multi-provider SSO login page plus the login-options endpoint (A-PR4), building on the merged DB-backed multi-provider OIDC auth (#12797).

- `/auth/type` now also returns the enabled SSO providers, labels only (name, display name, type, authorize URL). Allowed domains stay server-side so the public login page does not enumerate the participating companies' domains.
- The login page renders one button per provider above the password form. Each button fetches its per-provider authorize endpoint from the browser (so the CSRF/PKCE cookie lands on the right client) and then hands off to the IdP.
- Single-tenant only. Cloud and multi-tenant return no providers, so nothing changes for them. Legacy oidc/saml deployments keep their existing auto-redirect. Ships dark: an instance with no provider rows shows the plain password form.

## How Has This Been Tested?

Verified end-to-end in a real browser on a self-hosted (basic-auth) instance with two OIDC providers configured, and re-captured with Playwright.

Login page renders one button per provider, above the password form:

Multi-provider login page

Clicking a provider runs the full flow (authorize, IdP login, callback, JIT user, session) and lands authenticated in the app:

The backend authorization-code flow (two providers, per-provider domain gate, unverified-email rejection, no auto-link) is covered by the live-IdP test in #12824.

## Additional Options

### DIFF
--- a/backend/onyx/server/manage/get_state.py
+++ b/backend/onyx/server/manage/get_state.py
@@ -5,6 +5,7 @@ import requests
 from fastapi import APIRouter
 from fastapi import HTTPException
 from fastapi import Response
+from fastapi.concurrency import run_in_threadpool
 
 from onyx import __version__
 from onyx.auth.users import anonymous_user_enabled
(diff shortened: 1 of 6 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

justin-tahara (approved):

PR #12835, opened 2026-07-10
feat(opal): add Modal

nmgarza5:
## Description

Ports the `refresh-components` Modal (66 importers) into Opal — the library previously had no modal primitive (the biggest gap in the Opal audit: agent-wiki alone hand-rolls 11 scrims).

- Radix Dialog compound: `Modal` root, `Modal.Content` (scrim + positioning + card), `Modal.Header` / `Modal.Body` / `Modal.Footer`, plus the `BasicModalFooter` layout.
- Width/height presets, `center`/`top` positioning, `background`, `skipOverlay`, `bottomSlot`, and the two-stage accidental-close guard (typing arms it, first Escape focuses the close button).
- New `z-modal-overlay`/`z-modal` tokens in Opal's z-index.css matching the app's values, below popovers so pickers inside modals stack correctly.
- `@radix-ui/react-dialog` added to peerDependencies beside the other radix packages.

Deliberate divergences from the source: container-relative centering (`useContainerCenter`) is deferred (viewport-only), and the internal modal context is trimmed to the fields consumers read. Remaining Figma header/footer content variants (Search/Card/Panel, Checkbox/Message/Actions) are follow-ups.

## How Has This Been Tested?

Storybook stories for each variant (below), tsgo typecheck, and a live Playwright pass: open, full anatomy render, Escape close verified (dialog and overlay removed), re-verified after the handler-precedence fix.

Default (md, icon header, two-tone body, footer actions):

image

Minimal header (no icon, preview variant):

image

Danger footer:

image

Top position (command-palette placement):

image

## Additional Options

### DIFF
--- a/web/lib/opal/src/components/modal/components.tsx
+++ b/web/lib/opal/src/components/modal/components.tsx
@@ -204,14 +204,17 @@ const ModalContent = React.forwardRef<
       [preventAccidentalClose, hasAttemptedClose]
     );
 
-    const handleRef = (node: HTMLDivElement | null) => {
-      if (typeof ref === "function") {
-        ref(node);
-      } else if (ref) {
-        ref.current = node;
-      }
-      contentRef(node);
-    };
+    const handleRef = React.useCallback(
+      (node: HTMLDivElement | null) => {
+        if (typeof ref === "function") {
+          ref(node);
+        } else if (ref) {
+          ref.current = node;
+        }
+        contentRef(node);
+      },
+      [ref, contentRef]
+    );
 
     const isTop = position === "top";
 
@@ -228,7 +231,11 @@ const ModalContent = React.forwardRef<
       isTop ? "top-[72px]" : "-translate-y-1/2 top-1/2"
     );
 
+    // Internal handlers are defined after the props spread so callers cannot
+    // replace them. Each forwards to the caller's handler.
     const dialogEventHandlers = {
+      ...(!hasDescription && { "aria-describedby": undefined }),
+      ...props,
       onOpenAutoFocus: (e: Event) => {
         resetState();
         props.onOpenAutoFocus?.(e);
@@ -237,10 +244,22 @@ const ModalContent = React.forwardRef<
         resetState();
         props.onCloseAutoFocus?.(e);
       },
-      onEscapeKeyDown: handleInteractOutside,
-      onPointerDownOutside: handleInteractOutside,
-      ...(!hasDescription && { "aria-describedby": undefined }),
-      ...props,
+      onEscapeKeyDown: (e: KeyboardEvent) => {
+        handleInteractOutside(e);
+        props.onEscapeKeyDown?.(e);
+      },
+      onPointerDownOutside: (
+        e: Parameters<
+          NonNullable<
+            React.ComponentPropsWithoutRef<
+              typeof DialogPrimitive.Content
+            >["onPointerDownOutside"]
+          >
+        >[0]
+      ) => {
+        handleInteractOutside(e);
+        props.onPointerDownOutside?.(e);
+      },
     };
 
     const cardClasses = cn(
--- a/web/lib/opal/src/components/modal/components.tsx
+++ b/web/lib/opal/src/components/modal/components.tsx
@@ -1,0 +1,512 @@
+"use client";
+
+import React from "react";
+import * as DialogPrimitive from "@radix-ui/react-dialog";
+import { cn } from "@opal/utils";
+import type {
+  IconFunctionComponent,
+  RichStr,
+  WithoutStyles,
+} from "@opal/types";
+import { Button } from "@opal/components";
+import { Content, Section, type SectionProps } from "@opal/layouts";
+import { toPlainString } from "@opal/components/text/InlineMarkdown";
+import { SvgX } from "@opal/icons";
+
+// ---------------------------------------------------------------------------
+// Root + Overlay
+// ---------------------------------------------------------------------------
+
+const ModalRoot = DialogPrimitive.Root;
+
+const ModalOverlay = React.forwardRef<
+  React.ComponentRef<typeof DialogPrimitive.Overlay>,
+  WithoutStyles<React.ComponentPropsWithoutRef<typeof DialogPrimitive.Overlay>>
+>(({ ...props }, ref) => (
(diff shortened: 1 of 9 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

web/lib/opal/src/components/modal/components.tsx
```diff
@@ -0,0 +1,512 @@
...
+      },
+      onCloseAutoFocus: (e: Event) => {
+        resetState();
+        props.onCloseAutoFocus?.(e);
+      },
+      onEscapeKeyDown: handleInteractOutside,
+      onPointerDownOutside: handleInteractOutside,
+      ...(!hasDescription && { "aria-describedby": undefined }),
+      ...props,
+    };
```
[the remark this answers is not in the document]
nmgarza5: Fixed in bc6b77d0e4: `...props` now spreads first and all four internal handlers (onOpenAutoFocus, onCloseAutoFocus, onEscapeKeyDown, onPointerDownOutside) are defined after it, each forwarding to the caller's handler.

web/lib/opal/src/components/modal/components.tsx
```diff
@@ -0,0 +1,512 @@
...
+    );
+
+    const handleRef = (node: HTMLDivElement | null) => {
+      if (typeof ref === "function") {
+        ref(node);
+      } else if (ref) {
+        ref.current = node;
+      }
+      contentRef(node);
+    };
```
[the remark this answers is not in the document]
nmgarza5: Fixed in bc6b77d0e4: handleRef is wrapped in useCallback([ref, contentRef]).

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

web/lib/opal/src/components/modal/components.tsx
```diff
@@ -0,0 +1,558 @@
...
+  React.ComponentRef<typeof DialogPrimitive.Overlay>,
+  WithoutStyles<React.ComponentPropsWithoutRef<typeof DialogPrimitive.Overlay>>
+>(({ ...props }, ref) => (
+  <DialogPrimitive.Overlay
+    ref={ref}
+    className={cn(
+      "fixed inset-0 z-modal-overlay bg-mask-03 backdrop-blur-03 pointer-events-none",
+      "data-[state=open]:animate-in data-[state=closed]:animate-out",
+      "data-[state=open]:fade-in-0 data-[state=closed]:fade-out-0"
+    )}
```
raunakab: Could we throw these inside of a `styles.css`, please?

web/lib/opal/src/components/modal/components.tsx
```diff
@@ -0,0 +1,558 @@
...
+
+// ---------------------------------------------------------------------------
+// Content
+// ---------------------------------------------------------------------------
+
+interface ModalContentProps extends WithoutStyles<
+  React.ComponentPropsWithoutRef<typeof DialogPrimitive.Content>
+> {
+  width?: keyof typeof widthClasses;
+  height?: keyof typeof heightClasses;
```
raunakab: Could we possibly have these use the `SizeVariants` that is inside of `opal/src/types.ts`?

Something like:

```tsx
width?: Extract<SizeVariants, "fit" | "sm" | "lg" | "full">;
height?: Extract<SizeVariants, "fit" | "sm" | "lg" | "full">;
```

web/lib/opal/src/components/modal/components.tsx
```diff
@@ -0,0 +1,558 @@
...
+   * Content rendered below the modal card with 1rem separation. Stays inside
+   * DialogPrimitive.Content so focus management covers it.
+   */
+  bottomSlot?: React.ReactNode;
+}
+
+const ModalContent = React.forwardRef<
+  React.ComponentRef<typeof DialogPrimitive.Content>,
+  ModalContentProps
+>(
```
raunakab: Could we remove the `forwardRef` here and instead use the React17 style (just a simple `ref?: React.Ref<...>;` inside of the props is sufficient).

raunakab (approved):
Looks good! Two small things:

1. Could we move the stylings from inline Tailwind to a `modal/styles.css`?
2. React-17 allows for "bare" refs now. So instead of `React.forwardRef(..)`, we can just throw a `ref?: React.RefObject<..>` directly into the props. I think it makes the code a bit cleaner to read (one less unnecessary indent and function-wrap)!

web/lib/opal/src/components/modal/components.tsx
nmgarza5: definitely. much better

nmgarza5:
@greptile

web/lib/opal/src/components/modal/components.tsx
nmgarza5: Done in a54f509b2d — moved the overlay, positioner, and card chrome into a new `modal/styles.css` as `.opal-modal-*` classes, with the variants driven by `data-` attributes (`data-width`, `data-height`, `data-position`, `data-background`, `data-two-tone`). Only the dynamic container-center coordinates stay inline since they're runtime values.

web/lib/opal/src/components/modal/components.tsx
nmgarza5: Done in a54f509b2d. One tweak from your snippet: `width` keeps `md` and `xl` (`Extract<SizeVariants, "sm" | "md" | "lg" | "xl" | "full">`) because shipped consumers already pass `width="md"`, and `height` is `Extract<SizeVariants, "fit" | "sm" | "lg" | "full">`. Same value sets as before, now anchored to the shared vocabulary.

web/lib/opal/src/components/modal/components.tsx
nmgarza5: Done in a54f509b2d — all five (Overlay, Content, Header, Body, Footer) are now plain functions taking `ref?: React.Ref<...>` in their props, and the `displayName` assignments are gone.

nmgarza5:
@greptile

nmgarza5:
@greptile review

PR #12844, opened 2026-07-10
fix(craft): hide nextjs.log and nextjs.pid in file explorer

rohoswagger:
## What & why

In the Onyx Craft sandbox Files tab, the Next.js dev server's runtime artifacts `nextjs.log` and `nextjs.pid` were showing up as listed files in the explorer. These are internal runtime files (written to the session root by the dev-server start script) and shouldn't be surfaced to users as workspace files.

This adds `nextjs.log` and `nextjs.pid` to the existing `HIDDEN_PATTERNS` set that the file listing already uses to filter out entries like `node_modules`, `.next`, and `opencode.json`. No new mechanism is introduced — it reuses the single canonical filter in `SessionManager` (`_is_hidden_workspace_entry`). Since neither filename starts with `.`, they were not caught by the existing dotfile rule.

The same set also drives the artifact-zip walker, so these runtime files are now correctly excluded from downloaded archives as well.

## How it was verified

- Added a focused unit test (`test_hidden_entries.py`) asserting `_is_hidden_workspace_entry` hides `nextjs.log`/`nextjs.pid` while keeping regular files visible. Confirmed it fails before the fix and passes after.
- Extended the existing integration filter test (`test_list_directory_filters_hidden_entries`) to cover both filenames.
- A regression review confirmed process management (pid/log read by direct shell path) and snapshot/restore are unaffected — only the file listing and zip walker consume the filter.

## Known minor items

None.

- [x] Override Linear Check

### DIFF
--- a/backend/onyx/server/features/build/session/manager.py
+++ b/backend/onyx/server/features/build/session/manager.py
@@ -95,6 +95,8 @@ HIDDEN_PATTERNS = {
     "opencode.json",
     ".env",
     ".gitignore",
+    "nextjs.log",
+    "nextjs.pid",
 }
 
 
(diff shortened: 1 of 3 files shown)


rohoswagger:

PR #12859, opened 2026-07-10
refactor(ee): share credential_json helper across perm-sync connectors

wenxi-onyx:
## Description

Perm-sync connectors all need the same one-liner to unwrap a cc_pair's credential dict — `cc_pair.credential.credential_json.get_value(apply_mask=False)` (falling back to `{}`). Today **canvas** keeps it as a private `credential_json()` helper while **nine** other modules inline an identical copy.

This promotes it to a single shared `ee/onyx/external_permissions/utils.py::credential_json` and migrates every duplicator: canvas, github, gmail, google_drive, jira, salesforce, sharepoint, slack, teams (across their `doc_sync` / `group_sync` / `utils`). Behavior-preserving — the helper returns the exact same dict.

Extracted ahead of the Box connector work (Box reuses this helper) but stands on its own as a repo-wide de-duplication.

## How Has This Been Tested?

- `ruff` + `ty` clean across `ee/onyx/external_permissions/`.
- 88 EE permission-sync unit tests pass (jira / confluence / sharepoint).
- Import-smoke of all 15 migrated modules.

## Additional Options

- [ ] [Optional] Please cherry-pick this PR to the latest release version.
- [x] [Optional] Override Linear Check

PR #12862, opened 2026-07-10
feat(connectors): Box connector (indexing + permission sync)

wenxi-onyx:
## Description

Adds a **Box** connector.

- **Auth:** Client Credentials Grant (`box_client_id` / `box_client_secret` / `box_enterprise_id`), with optional user impersonation configured by **email** (`box_user_email`), resolved to a Box user ID internally.
- **Indexing:** BFS crawl of the folder tree — files, web links (optional), and `HierarchyNode`s for folders (parent-before-child). Content via the shared `extract_text_and_images`; stable `box-file-{id}` / `box-weblink-{id}` doc IDs; UTC `doc_updated_at` + `doc_created_at`; overlapping/duplicate entry roots de-duplicated.
- **Permission sync (EE):** document ACLs from collaborations (role ≥ previewer; `uploader` excluded) + shared links (`open` → public, `company` → synthetic enterprise-users group), inherited down the folder chain; group sync of enterprise groups/memberships. Frequencies aligned with SharePoint (doc 30m / group 5m).
- Registration, frontend config + credential forms, deployment env vars (+ Helm chart bump), and daily + unit tests.

Incremental (events-based) indexing is a **follow-up PR** stacked on this one; this PR always does a full crawl.

> Stacked on #12859 (shared `credential_json` helper). Base will retarget to `main` once that merges.

## How Has This Been Tested?

- 43 offline unit tests (traversal, checkpointing + fresh-instance resume, dedup, failure isolation, slim/full parity, permission resolution matrix incl. the uploader negative case, group-sync pagination, email→ID resolution).
- Daily suite (`test_box_basic.py`) verified end-to-end against a real Box enterprise via the CCG path (documents, hierarchy, permissions, group sync, validation). See `backend/tests/daily/connectors/box/README.md` for test-env setup.
- `ruff`, `ty`, `oxfmt`, `oxlint`, TypeScript, and env-drift all green via pre-commit.

## Additional Options

---
## Summary by cubic
Adds a Box connector with full indexing and enterprise permission sync using Client Credentials Grant, with optional user impersonation by email. Registers the source end-to-end (backend + UI), sets sane defaults, hardens ACL handling, and splits Box API usage into content vs enterprise clients.

- **New Features**
  - Auth: Client Credentials Grant via `box_client_id`, `box_client_secret`, `box_enterprise_id`; optional `box_user_email` for impersonation.
  - Indexing: BFS folders; files and optional web links; stable IDs (`box-file-{id}`, `box-weblink-{id}`); timestamps; size cap via `BOX_CONNECTOR_SIZE_THRESHOLD`; skips collaboration lookup on the Box root folder.
  - Permissions (EE): ACLs from collaborations (role ≥ previewer; excludes uploader) and shared links (open → public; company → enterprise group); inheritance; enterprise group sync; default schedules 30m (docs) and 5m (groups) via `BOX_PERMISSION_DOC_SYNC_FREQUENCY` / `BOX_PERMISSION_GROUP_SYNC_FREQUENCY`; early validation for group/user listing scopes.
  - App/UI: Source registered (`DocumentSource.BOX`) and added to auto-sync; credential + config forms (folder IDs/URLs, “include web links”); Box icon; Helm and Docker env templates expose Box env vars; adds `boxsdk` dependency.

- **Bug Fixes**
  - Corrected ACL resolution and access propagation in permission sync; cleanup from review feedback.
  - Split connector into `content_client` and `enterprise_client`; updated the daily test to use `content_client`.
  - Daily tests: mock Unstructured API key during file extraction; enable EE for the perm-sync test so external access resolvers run.

<sup>Written for commit 80cde69242f57e9477893b6f33aeaf74d403ea83. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12862?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/ee/onyx/external_permissions/box/group_sync.py
```diff
@@ -0,0 +1,80 @@
...
+    tenant_id: str,  # noqa: ARG001
+    cc_pair: ConnectorCredentialPair,
+) -> Generator[ExternalUserGroup, None, None]:
+    connector = BoxConnector(**cc_pair.connector.connector_specific_config)
+    connector.load_credentials(credential_json(cc_pair))
+    client = connector.enterprise_client
+
+    offset = 0
+    while True:
+        groups = client.groups.get_groups(limit=_PAGE_SIZE, offset=offset)
```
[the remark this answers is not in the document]
wenxi-onyx: Fixed in ec525d2b7f — added a `_MAX_OFFSET = 10_000` guard on the groups pagination loop: it stops with a warning before reaching Box's offset ceiling rather than letting the 400 abort the whole sync. (The enterprise all-users list uses marker pagination, so it has no offset ceiling.)

backend/ee/onyx/external_permissions/box/group_sync.py
```diff
@@ -0,0 +1,80 @@
...
+logger = setup_logger()
+
+_PAGE_SIZE = 1000
+
+
+def _fetch_group_member_emails(client: BoxClient, group_id: str) -> set[str]:
+    emails: set[str] = set()
+    offset = 0
+    while True:
+        memberships = client.memberships.get_group_memberships(
```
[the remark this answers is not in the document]
wenxi-onyx: Fixed in ec525d2b7f — same `_MAX_OFFSET` guard applied inside `_fetch_group_member_emails`, so an oversized group truncates with a warning instead of throwing a 400.

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,960 @@
...
+    def _resolve_user_id_from_email(self, email: str) -> str:
+        """Look up the numeric Box user ID for an email so admins configure the
+        connector with an email instead of an opaque ID. Requires the app's
+        'Manage users' scope (the enterprise-subject client)."""
+        normalized = email.strip().lower()
+        try:
+            users = self.enterprise_client.users.get_users(
+                filter_term=normalized, fields=["login"], limit=100
+            )
+        except BoxAPIError as e:
```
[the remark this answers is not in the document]
wenxi-onyx: Fixed in ec525d2b7f — `_resolve_ancestor_access` now re-raises on any non-403/404 (transient/5xx) error, so a transient failure can no longer be silently treated as 'no inherited access' and under-permission every descendant. A genuine 403/404 (the user can't see an ancestor above the configured root) is still skipped.

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,960 @@
...
+                raise ConnectorValidationError(
+                    "A configured Box folder was not found or is not visible to "
+                    "the authenticated user (HTTP 404). If using the service "
+                    "account (no impersonated user), it must be added as a "
+                    "collaborator on the folder."
+                )
+            raise UnexpectedValidationError(
+                f"Unexpected Box API error during validation (status={status}): "
+                f"{e.message}"
+            )
```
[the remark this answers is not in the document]
wenxi-onyx: Fixed in ec525d2b7f — `validate_connector_settings` now runs the identity check (`get_user_me`) in its own try/except that reports a credential error, separate from the folder-existence checks that emit the folder-not-found message.

backend/ee/onyx/external_permissions/box/group_sync.py
```diff
@@ -0,0 +1,80 @@
...
+
+def box_group_sync(
+    tenant_id: str,  # noqa: ARG001
+    cc_pair: ConnectorCredentialPair,
+) -> Generator[ExternalUserGroup, None, None]:
+    connector = BoxConnector(**cc_pair.connector.connector_specific_config)
+    connector.load_credentials(credential_json(cc_pair))
+    client = connector.enterprise_client
+
+    offset = 0
```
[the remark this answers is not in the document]
wenxi-onyx: Fixed in ec525d2b7f — `box_group_sync` strips `box_user_email` before `load_credentials`, so it uses only the enterprise client and skips the impersonation user lookup (which would also have failed the whole sync if that user was later deactivated).

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,960 @@
...
+    """Fold a Box shared link into an access context.
+
+    - "open" links are world-readable -> public (unless password protected).
+    - "company" links are readable by any logged-in enterprise user -> the
+      synthetic all-enterprise-users group (populated by group sync).
+    - "collaborators" links grant nothing beyond existing collaborations.
+    """
+    if shared_link_access is None:
+        return access
+    if shared_link_access == "open" and not is_password_enabled:
```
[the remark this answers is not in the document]
wenxi-onyx: Fixed in ec525d2b7f — added `shared_link_access_level`, which prefers `effective_access` over the raw `access`. A link the enterprise has downgraded from `open` is no longer indexed as public.

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,960 @@
...
+    "viewer",
+    "previewer",
+    "previewer uploader",
+    "viewer uploader",
+    "co-owner",
+    "owner",
+}
+
+BOX_GROUP_ID_PREFIX = "box-group"
+BOX_ALL_ENTERPRISE_USERS_GROUP_ID = "box-enterprise-all-users"
```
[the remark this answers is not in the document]
wenxi-onyx: Fixed in ec525d2b7f — `box_all_enterprise_users_group_id` now takes the enterprise id and returns a per-enterprise group id, so two Box connectors on one tenant can't cross-leak `company`-link documents through a shared synthetic group.

backend/onyx/configs/constants.py
```diff
@@ -235,6 +235,7 @@ class DocumentSource(str, Enum):
     GOOGLE_SITES = "google_sites"
     ZENDESK = "zendesk"
     LOOPIO = "loopio"
+    BOX = "box"
```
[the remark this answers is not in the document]
wenxi-onyx: Fixed in ec525d2b7f — added `DocumentSource.BOX` to `_SOURCE_IMAGE_FILENAMES` (default file icon; there's no Box slackbot asset yet), so Slack rendering of a Box citation no longer raises `KeyError`.

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,960 @@
...
+        include_permissions: bool,
+    ) -> SlimDocument:
+        external_access = None
+        if include_permissions and folder.access is not None:
+            if isinstance(item, FileFull):
+                external_access = self._resolve_file_access(
+                    item, folder.access
+                ).to_external_access()
+            else:
+                external_access = folder.access.to_external_access()
```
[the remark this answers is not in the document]
wenxi-onyx: Fixed in ec525d2b7f — `_convert_web_link` now folds the bookmark's own `shared_link` into its access via `apply_shared_link_to_access`, instead of only inheriting the parent folder ACL.

web/src/lib/connectors/credentials.ts
```diff
@@ -336,6 +343,12 @@ export const credentialTemplates: Record<ValidSources, any> = {
     loopio_client_id: "",
     loopio_client_token: "",
   } as LoopioCredentialJson,
+  box: {
+    box_client_id: "",
+    box_client_secret: "",
+    box_enterprise_id: "",
+    box_user_email: "",
```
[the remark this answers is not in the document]
wenxi-onyx: Fixed — the `box_user_email` credential template default is now `null` (and the interface type `string | null`), so the optional impersonation field is no longer treated as required by the generic validator.

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,960 @@
...
+                        parent_folder_id=entry.folder_id,
+                        path=f"{entry.path}/{child_name}",
+                        access=entry.access,
+                    )
+                )
+            elif isinstance(item, FileFull):
+                if not _in_time_window(item.modified_at, start, end):
+                    continue
+                if slim:
+                    yield self._build_slim_document(item, entry, include_permissions)
```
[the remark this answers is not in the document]
wenxi-onyx: Fixed in ec525d2b7f — extracted `_file_is_indexable` (size + type gate) and applied it in the slim path as well, so pruning no longer keeps documents alive for files the full path would skip (oversized/unsupported).

wenxi-onyx:
@greptile @codex @cubic-dev-ai review

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1026 @@
...
+    """Accepts a raw Box folder ID or a Box folder URL like
+    https://app.box.com/folder/123456789 and returns the folder ID."""
+    value = folder_id_or_url.strip()
+    if not value.startswith(("http://", "https://")):
+        return value
+
+    parsed = urlparse(value)
+    path_parts = [part for part in parsed.path.split("/") if part]
+    if len(path_parts) >= 2 and path_parts[-2] == "folder":
+        return path_parts[-1]
```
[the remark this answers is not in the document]
wenxi-onyx: Fixed in a6a86d6ff0 — `parse_box_folder_id` now accepts both `/folder/<id>` and `/folders/<id>`.

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1026 @@
...
+    collaborations with a read-capable role grant access."""
+    user_emails = set(access.user_emails)
+    group_ids = set(access.group_ids)
+    for collaboration in collaborations:
+        if not _collaboration_grants_read(collaboration):
+            continue
+        accessible_by = collaboration.accessible_by
+        if isinstance(accessible_by, UserCollaborations):
+            if accessible_by.login:
+                user_emails.add(accessible_by.login)
```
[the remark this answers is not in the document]
wenxi-onyx: Fixed in a6a86d6ff0 — added `normalize_box_login` (strip + lowercase) and applied it everywhere a Box login enters an ACL (collaborations, file/folder owners, group-sync members, enterprise users), so stored `external_user_emails` match Onyx's lowercased identities.

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1026 @@
...
+        include_permissions: bool,
+    ) -> Document | None:
+        if web_link.url is None:
+            return None
+        name = web_link.name or web_link.url
+        text = f"{name}\n{web_link.description}" if web_link.description else name
+
+        external_access = None
+        if include_permissions and folder.access is not None:
+            access = folder.access
```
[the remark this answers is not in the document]
wenxi-onyx: Fixed in a6a86d6ff0 — the slim path now resolves web-link access through the same helper as the full path (`_resolve_web_link_access` → `_apply_shared_link`), so a bookmark's own open/company shared link is applied in `_build_slim_document` too. Perm sync can no longer overwrite it with folder-only access. A regression test asserts the full and slim paths produce identical `external_access`.

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1026 @@
...
+            id=document_id,
+            external_access=external_access,
+            parent_hierarchy_raw_node_id=folder.folder_id,
+        )
+
+    def _folder_hierarchy_node(self, folder: BoxFolderFrontierEntry) -> HierarchyNode:
+        return HierarchyNode(
+            raw_node_id=folder.folder_id,
+            raw_parent_id=folder.parent_folder_id,
+            display_name=folder.display_name,
```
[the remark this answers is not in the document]
wenxi-onyx: Fixed in a6a86d6ff0 — the slim path now resolves web-link access through the same helper as the full path (`_resolve_web_link_access` → `_apply_shared_link`), so a bookmark's own open/company shared link is applied in `_build_slim_document` too. Perm sync can no longer overwrite it with folder-only access. A regression test asserts the full and slim paths produce identical `external_access`.

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1026 @@
...
+        include_permissions: bool,
+    ) -> SlimDocument:
+        external_access = None
+        if include_permissions and folder.access is not None:
+            if isinstance(item, FileFull):
+                external_access = self._resolve_file_access(
+                    item, folder.access
+                ).to_external_access()
+            else:
+                external_access = folder.access.to_external_access()
```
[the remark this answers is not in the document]
wenxi-onyx: Fixed in a6a86d6ff0 — the slim path now resolves web-link access through the same helper as the full path (`_resolve_web_link_access` → `_apply_shared_link`), so a bookmark's own open/company shared link is applied in `_build_slim_document` too. Perm sync can no longer overwrite it with folder-only access. A regression test asserts the full and slim paths produce identical `external_access`.

backend/ee/onyx/external_permissions/box/group_sync.py
```diff
@@ -0,0 +1,114 @@
...
+        )
+        entries = memberships.entries or []
+        for membership in entries:
+            if membership.user is not None and membership.user.login:
+                emails.add(membership.user.login)
+        offset += len(entries)
+        total = memberships.total_count
+        if not entries or (total is not None and offset >= total):
+            return emails
+        if offset >= _MAX_OFFSET:
```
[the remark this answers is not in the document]
wenxi-onyx: Fixed in a6a86d6ff0 — a group exceeding the offset ceiling is now treated as unpageable: `_fetch_group_member_emails` raises `BoxGroupTooLargeError` and the group is skipped entirely, so its prior membership is preserved rather than replaced by a partial set. No more silent revocation of the omitted members. The complete fix (user-centric enumeration with no ceiling) is tracked as a follow-up.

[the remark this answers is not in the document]
wenxi-onyx: Addressed in a6a86d6ff0, though differently than proposed: rather than fetch one more page, an over-ceiling group is now skipped entirely (prior membership preserved) — truncating at any boundary would still revoke access for members past it, so the exact off-by-one is moot. Full enumeration via a user-centric traversal is the tracked follow-up.

backend/ee/onyx/external_permissions/box/group_sync.py
```diff
@@ -0,0 +1,126 @@
...
+class BoxGroupTooLargeError(Exception):
+    """A group has more members than Box's offset pagination can enumerate, so
+    it can't be synced completely."""
+
+
+def _fetch_group_member_emails(client: BoxClient, group_id: str) -> set[str]:
+    emails: set[str] = set()
+    offset = 0
+    while True:
+        memberships = client.memberships.get_group_memberships(
```
evan-onyx, line 33: did you try using marker-based pagination here and it didn't work? I guess their APIs just don't support fetching groups with >10k members? :(

backend/ee/onyx/external_permissions/box/group_sync.py
```diff
@@ -0,0 +1,126 @@
...
+            # replace membership and revoke access for everyone omitted, so treat
+            # it as unpageable and let the caller preserve prior membership.
+            raise BoxGroupTooLargeError(group_id)
+
+
+def _fetch_all_enterprise_user_emails(client: BoxClient) -> set[str]:
+    # The users listing is marker-paginated, so it has no offset ceiling.
+    emails: set[str] = set()
+    marker: str | None = None
+    while True:
```
evan-onyx: while True makes me nervous, i.e. if they had a different behavior in an older API version we could infinite loop

backend/ee/onyx/external_permissions/box/group_sync.py
```diff
@@ -0,0 +1,126 @@
...
+    tenant_id: str,  # noqa: ARG001
+    cc_pair: ConnectorCredentialPair,
+) -> Generator[ExternalUserGroup, None, None]:
+    creds = credential_json(cc_pair)
+    enterprise_id = creds["box_enterprise_id"]
+    connector = BoxConnector(**cc_pair.connector.connector_specific_config)
+    # Group sync only uses the enterprise client; drop the impersonation email so
+    # a deactivated/renamed impersonation user can't fail the whole sync.
+    connector.load_credentials(
+        {k: v for k, v in creds.items() if k != "box_user_email"}
```
evan-onyx: box_user_email should probably be a top level str constant

backend/ee/onyx/external_permissions/box/group_sync.py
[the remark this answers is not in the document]
evan-onyx: I think the right way to resolve this is to have that API call be made only during the main connector flow so we can just directly load the credentials without any special-casing

backend/ee/onyx/external_permissions/box/group_sync.py
evan-onyx: (but see below, it probably shouldn't be needed here regardless as we'd prefer to load the full credentials as is done in other connectors)

backend/ee/onyx/external_permissions/box/group_sync.py
```diff
@@ -0,0 +1,126 @@
...
+                # access for members past the offset ceiling).
+                logger.warning(
+                    "Box group %s exceeds the %d-member pagination ceiling; "
+                    "skipping to preserve its existing membership.",
+                    group.id,
+                    _MAX_OFFSET,
+                )
+                continue
+            except Exception:
+                # Skip (don't yield) on failure: one bad group must not abort the
```
evan-onyx: I think we should fail loudly here

backend/ee/onyx/external_permissions/box/group_sync.py
```diff
@@ -0,0 +1,126 @@
...
+    connector = BoxConnector(**cc_pair.connector.connector_specific_config)
+    # Group sync only uses the enterprise client; drop the impersonation email so
+    # a deactivated/renamed impersonation user can't fail the whole sync.
+    connector.load_credentials(
+        {k: v for k, v in creds.items() if k != "box_user_email"}
+    )
+    client = connector.enterprise_client
+
+    offset = 0
+    while True:
```
evan-onyx: would similarly rather not while True here

evan-onyx: (common pattern is to just have a max iterations, in this case could just be _MAX_OFFSET since more iterations than that indicates something is very wrong)

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+
+
+def box_all_enterprise_users_group_id(enterprise_id: str) -> str:
+    # Scoped by enterprise so two Box connectors on one tenant can't leak
+    # "company"-link docs across enterprises via a shared group id.
+    return f"{BOX_ALL_ENTERPRISE_USERS_GROUP_PREFIX}-{enterprise_id}"
+
+
+def box_file_document_id(file_id: str) -> str:
+    return f"{_FILE_DOCUMENT_ID_PREFIX}{file_id}"
```
evan-onyx: nit: probably better to have a consistent convention for prefixes, i.e. we could make all of them not end in "-" and then put the "-" in the f-string following box_group_id or vice versa

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+def box_file_link(file_id: str) -> str:
+    return f"{BOX_APP_BASE_URL}/file/{file_id}"
+
+
+def parse_box_folder_id(folder_id_or_url: str) -> str:
+    """Accepts a raw Box folder ID or a Box folder URL like
+    https://app.box.com/folder/123456789 and returns the folder ID."""
+    value = folder_id_or_url.strip()
+    if not value.startswith(("http://", "https://")):
+        return value
```
evan-onyx: any validation we can do here? the example seems to indicate ids are just numbers, but if it's i.e. alphanumeric it would be nice to validate that here so we can error here instead of later

evan-onyx: for cleanliness you could make this a private func and have the exposed parse_box_folder_id wrap it and valid this function's output

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+        return True
+    timestamp = modified_at.timestamp()
+    if start is not None and timestamp < start:
+        return False
+    if end is not None and timestamp > end:
+        return False
+    return True
+
+
+def _collaboration_grants_read(collaboration: Collaboration) -> bool:
```
evan-onyx: brief comment explaining business logic here would be good

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+        return False
+    role_value = str(collaboration.role.value)
+    if role_value not in _READ_ACCESS_COLLABORATION_ROLES:
+        if role_value != "uploader":
+            logger.warning("Unrecognized Box collaboration role: %s", role_value)
+        return False
+    return True
+
+
+def apply_collaborations_to_access(
```
evan-onyx: this seems like an ee function to me; does CE need the access for something?

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+        self, access: BoxAccessContext, owned_by: UserMini | None
+    ) -> BoxAccessContext:
+        if owned_by is None or not owned_by.login:
+            return access
+        return access.merged_with(
+            BoxAccessContext(user_emails={normalize_box_login(owned_by.login)})
+        )
+
+    def _apply_shared_link(
+        self, access: BoxAccessContext, shared_link: Any
```
evan-onyx: Any is probably too broad here, is there a type we can get from the box SDK

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+        elif isinstance(accessible_by, GroupMini):
+            group_ids.add(box_group_id(accessible_by.id))
+    return BoxAccessContext(
+        user_emails=user_emails,
+        group_ids=group_ids,
+        is_public=access.is_public,
+    )
+
+
+def shared_link_access_level(shared_link: Any) -> str | None:
```
evan-onyx: Any hurts my soul

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+    downgrade an "open" link, and only `effective_access` reflects that (so we
+    don't treat a restricted link as public)."""
+    if shared_link is None:
+        return None
+    if shared_link.effective_access is not None:
+        return shared_link.effective_access.value
+    return shared_link.access.value if shared_link.access else None
+
+
+def apply_shared_link_to_access(
```
evan-onyx: pretty much everything with BoxAccessContext should probably be in ee/

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+    @property
+    def client(self) -> BoxClient:
+        """Client used for content reads. Uses the impersonated user when
+        box_user_email is configured, else the app's service account."""
+        if self._client is None:
+            raise ConnectorMissingCredentialError("Box")
+        return self._client
+
+    @property
+    def enterprise_client(self) -> BoxClient:
```
evan-onyx, line 253: if this is only used for perm sync then it shouldn't be in the connector

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+                client_secret=client_secret,
+                enterprise_id=enterprise_id,
+            )
+        )
+        self._enterprise_client = BoxClient(auth=auth)
+        self._enterprise_id = enterprise_id
+
+        user_email = credentials.get("box_user_email")
+        if user_email:
+            user_id = self._resolve_user_id_from_email(user_email)
```
evan-onyx: this logic can be moved to the client property in the case where _client is still None; we can store the credentials and user email as a lot of other connectors do the same

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+
+    def _all_enterprise_users_group_id(self) -> str:
+        if self._enterprise_id is None:
+            raise ConnectorMissingCredentialError("Box")
+        return box_all_enterprise_users_group_id(self._enterprise_id)
+
+    def load_credentials(self, credentials: dict[str, Any]) -> dict[str, Any] | None:
+        client_id = credentials.get("box_client_id")
+        client_secret = credentials.get("box_client_secret")
+        enterprise_id = credentials.get("box_enterprise_id")
```
evan-onyx: same as below comment (sorry for out of order) but I think this function can just be storing the relevant info in the connector and then defer creating the client until someone references the enterprise_client property

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+            self._client = self._enterprise_client
+        return None
+
+    def _resolve_user_id_from_email(self, email: str) -> str:
+        """Look up the numeric Box user ID for an email so admins configure the
+        connector with an email instead of an opaque ID. Requires the app's
+        'Manage users' scope (the enterprise-subject client)."""
+        normalized = normalize_box_login(email)
+        try:
+            users = self.enterprise_client.users.get_users(
```
evan-onyx: doesn't this only get the first 100 users? there's already a function _fetch_all_enterprise_user_emails in the ee code which we could just move to CE and then use in both places

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+
+        path_parts = [part for part in parsed.path.split("/") if part]
+        if len(path_parts) >= 2 and path_parts[-2] == "file":
+            return NormalizationResult(
+                normalized_url=box_file_document_id(path_parts[-1]),
+                use_default=False,
+            )
+        return NormalizationResult(normalized_url=None, use_default=False)
+
+    def _fetch_collaborations_access(
```
evan-onyx: seems like more ee

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+                )
+            )
+        return frontier
+
+    def _download_file(self, file: FileFull) -> bytes | None:
+        stream = self.client.downloads.download_file(file_id=file.id)
+        if stream is None:
+            return None
+        try:
+            return stream.read()
```
evan-onyx: this should follow the same convention as what we do for g drive and sharepoint; we stream the bytes but if the file size gets too large we abort and log. I see you're doing the API-based check which is a good first layer of defense but can't trust it

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+    def _build_file_sections(
+        self, file: FileFull
+    ) -> list[TextSection | ImageSection] | None:
+        """Returns None when the file should be skipped (unsupported type,
+        over the size threshold, images disabled, or empty download)."""
+        if not self._file_is_indexable(file):
+            return None
+
+        file_name = file.name or file.id
+        extension = get_file_ext(file_name)
```
evan-onyx, line 411: I believe we have a list of allowed extensions somewhere, we can throw out anything that isn't in the list since we won't be able to process it anyways (and it saves the download)

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+        link = box_file_link(file.id)
+
+        if extension in OnyxFileExtensions.IMAGE_EXTENSIONS:
```
evan-onyx, line 414: should this also be gated on allow_images?

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+        sections: list[TextSection | ImageSection] = [
+            TextSection(link=link, text=extraction.text_content)
+        ]
+        if self.allow_images:
+            for index, (image_data, image_name) in enumerate(
+                extraction.embedded_images
+            ):
+                image_section, _ = store_image_and_create_section(
+                    image_data=image_data,
+                    file_id=f"{box_file_document_id(file.id)}-img-{index}",
```
evan-onyx: probably should have a function for creating these ids like you already have for normal document ids

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+    ) -> Document | ConnectorFailure | None:
+        document_id = box_file_document_id(file.id)
+        try:
+            sections = self._build_file_sections(file)
+            if sections is None:
+                return None
+
+            external_access = None
+            if include_permissions and folder.access is not None:
+                external_access = self._resolve_file_access(
```
evan-onyx: yeah this really should not be in CE

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+        folder: BoxFolderFrontierEntry,
+        include_permissions: bool,
+    ) -> Document | None:
+        if web_link.url is None:
+            return None
+        name = web_link.name or web_link.url
+        text = f"{name}\n{web_link.description}" if web_link.description else name
+
+        external_access = None
+        if include_permissions and folder.access is not None:
```
evan-onyx, line 510: also should not be in CE (it's fine for these functions to make calls that go through fetch_versioned_implementation ofc but the implementations should def be in EE)

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+        if checkpoint.todo is None:
+            checkpoint.todo = self._seed_frontier(include_permissions)
+            checkpoint.has_more = True
+            return checkpoint
+
+        if checkpoint.current is None:
+            if not checkpoint.todo:
+                checkpoint.has_more = False
+                return checkpoint
+            entry = checkpoint.todo.pop(0)
```
evan-onyx: if you want to pop(0) this should be a queue, list is bad

evan-onyx: it can be serialized as a list, see drive connector for an example of checkpoints that serialize differently from how they're loaded

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+        items."""
+        checkpoint = deepcopy(checkpoint)
+
+        if checkpoint.todo is None:
+            checkpoint.todo = self._seed_frontier(include_permissions)
+            checkpoint.has_more = True
+            return checkpoint
+
+        if checkpoint.current is None:
+            if not checkpoint.todo:
```
evan-onyx: I think it's cleaner to put this in a helper called _pick_current_from_todos or similar, then in this if block it's just
```
yield from self._pick_current_from_todos(...)
return checkpoint
```

evan-onyx (commented):
still working through this, but here's what i have so far

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+
+    def probe_group_listing_permission(self) -> None:
+        """Group sync enumerates enterprise groups and users, which requires
+        the app's 'Manage groups' / 'Manage users' scopes."""
+        try:
+            self.enterprise_client.groups.get_groups(limit=1)
+            self.enterprise_client.users.get_users(limit=1)
+        except BoxAPIError as e:
+            if _box_api_status_code(e) == 403:
+                raise InsufficientPermissionsError(
```
evan-onyx, line 816: we should include the original error in some way so we know which place in the code errored out specifically (so maybe just a raise from)

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+            status = _box_api_status_code(e)
+            if status in (401, 404):
+                raise CredentialExpiredError(
+                    "Box credentials are invalid, or the impersonated user could "
+                    f"not be authenticated (HTTP {status}). Verify the client "
+                    "ID/secret, that the app is authorized in the Box Admin "
+                    "Console, and that the impersonated user exists."
+                )
+            if status == 403:
+                raise InsufficientPermissionsError(
```
evan-onyx, line 842: same as above

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+            raise ConnectorMissingCredentialError("Box")
+
+        # Identity check in its own block so its failure reports a credential
+        # problem, not the folder-not-found message below.
+        try:
+            self.client.users.get_user_me()
+        except BoxAPIError as e:
+            status = _box_api_status_code(e)
+            if status in (401, 404):
+                raise CredentialExpiredError(
```
evan-onyx, line 835: we should include the original error here, could have useful details for debugging

backend/onyx/connectors/box/connector.py
```diff
@@ -0,0 +1,1032 @@
...
+                f"{e.message}"
+            )
+
+        try:
+            for folder_id in self.entry_folder_ids:
+                self.client.folders.get_folder_by_id(folder_id, fields=["id"])
+        except BoxAPIError as e:
+            status = _box_api_status_code(e)
+            if status == 403:
+                raise InsufficientPermissionsError(
```
evan-onyx, line 856: okay you get the idea, we want the stack traces for all these cases (swallowing the original trace makes debugging very annoying)

evan-onyx (commented):
okay thats it

backend/ee/onyx/external_permissions/box/group_sync.py
wenxi-onyx: This is documented in the SDK/API docs (that the group membership API has this limitation). 

It is possible to use marker-based pagination + the user API to build the groups based on user data, but this adds significant complication. My gut tells me box has set this value relatively high above the number of groups real orgs have, and the limit is (hopefully) a non-issue

backend/ee/onyx/external_permissions/box/group_sync.py
wenxi-onyx: Replaced with iter to _MAX_OFFSET

backend/ee/onyx/external_permissions/box/group_sync.py
wenxi-onyx: Moved value to top level constant in the connector and now just passing in a client created from the creds at init

backend/ee/onyx/external_permissions/box/group_sync.py
wenxi-onyx: Removed skip

backend/onyx/connectors/box/connector.py
wenxi-onyx: Agreed, all prefixes have no dash and all id builders add the dash

backend/onyx/connectors/box/connector.py
wenxi-onyx: Added validation. It's not too much code and I personally dislike single-caller-helpers, so inlined as part of the parse func.

backend/onyx/connectors/box/connector.py
wenxi-onyx: added

backend/onyx/connectors/box/connector.py
wenxi-onyx: Yes to all of the ee/ce split comments

backend/onyx/connectors/box/connector.py
wenxi-onyx: tightened w/ the 3 possible types from box sdk

backend/onyx/connectors/box/connector.py
wenxi-onyx: I kept enterprise_client on the connector because it is not permission-sync-only: it is the normal content client when no impersonated user is configured, and it is required to resolve box_user_email before constructing an impersonated content client. Group sync reuses that credential-aware client boundary, avoiding duplicate Box authentication setup in EE.

backend/onyx/connectors/box/connector.py
wenxi-onyx: makes sense, implemented

backend/onyx/connectors/box/connector.py
wenxi-onyx: :ai-slop: 😓 - fixed and using the the iterate users util moved to CE

backend/onyx/connectors/box/connector.py
wenxi-onyx: implemented

backend/onyx/connectors/box/connector.py
wenxi-onyx: `_file_is_indexable` checks OnyxFileExtensions before download. Unsupported extensions are skipped in both full and slim paths.

backend/onyx/connectors/box/connector.py
wenxi-onyx: Yes

backend/onyx/connectors/box/connector.py
wenxi-onyx: Implemented

backend/onyx/connectors/box/connector.py
evan-onyx: I think it still isn't?

evan-onyx (approved):

PR #12871, opened 2026-07-10
fix(craft): hide nextjs.log and nextjs.pid in file explorer

rohoswagger:
## What & why

In the Onyx Craft sandbox **Files** tab, the dev-server runtime artifacts `nextjs.log` and `nextjs.pid` were showing up in the file explorer. These are internal Next.js dev-server files (stdout log + PID) written to the session workspace root, not user deliverables, so they should be hidden from the listing.

## Change

Add `nextjs.log` and `nextjs.pid` to the existing `HIDDEN_PATTERNS` set in `backend/onyx/server/features/build/session/manager.py`. This is the single, canonical filter (`_is_hidden_workspace_entry`) already used to hide entries like `node_modules`, `opencode.json`, and `.env` from both the directory listing and download zips. No frontend change is needed — the file explorer does no name-based filtering of its own.

## Verification

- New unit test `backend/tests/unit/onyx/server/features/build/session/test_hidden_workspace_entry.py` covers the predicate (positive: `nextjs.log`/`nextjs.pid`; negative: exact-match semantics so `nextjs.log.bak`/`mynextjs.pid` stay visible). Fails before the change, passes after.
- Extended the existing integration test `test_list_directory_filters_hidden_entries` to assert both files are filtered.
- `ruff check` and `ruff format --check` pass.

## Notes

- Filtering is by basename, consistent with the existing entries; the files remain on disk and the sandbox lifecycle scripts still read `nextjs.pid` by absolute path, so process management is unaffected.
- This declutters the listing only (it is not an access-control change), matching the pre-existing behavior of the other hidden entries.

---

- [x] Override Linear Check

### DIFF
--- a/backend/onyx/server/features/build/session/manager.py
+++ b/backend/onyx/server/features/build/session/manager.py
@@ -95,6 +95,8 @@ HIDDEN_PATTERNS = {
     "opencode.json",
     ".env",
     ".gitignore",
+    "nextjs.log",
+    "nextjs.pid",
 }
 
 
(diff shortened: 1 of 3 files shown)


rohoswagger:

PR #12883, opened 2026-07-10
fix(web): render optional credential edit fields

wenxi-onyx:
## Description

Screenshot 2026-07-10 at 3 41 11 PM
Screenshot 2026-07-10 at 3 40 49 PM
Screenshot 2026-07-10 at 3 40 43 PM

- Renders editable credential fields from the connector credential template instead of only the stored credential JSON.
- Preserves stored values only for keys in the editable template, so optional fields omitted when empty are restored without exposing unrelated stored credential keys.
- Prevents OAuth-managed credentials from opening the generic edit form, avoiding raw token fields such as refresh tokens and expiry metadata.
- Stretches both the credential selection modal and edit credential modal content to the modal body width.
- Adds focused unit coverage for optional template fields, multi-auth credential templates, and OAuth-managed credential internals.

## Additional Options


web/src/lib/credentials/components/EditCredential.tsx
raunakab, line 1: File should be updated with Opal icons. If the corresponding Opal icon does not exist, you can do the following:

1. Find the closest existing icon that *does* already exist in Opal and use that instead.
2. Implement that icon (from Figma [assuming that it exists in Figma]).
2a. If it does *not* exist in Figma, fall back to (1).

Ideally, you should replace *all* icons here; even the ones that you haven't directly touched. It will help towards the general direction of eliminating *all* 3rd party icon libraries that the FE team is pushing towards!

web/src/components/credentials/CredentialSection.tsx
```diff
@@ -34,7 +34,7 @@ import {
 } from "@/lib/connectors/oauth";
 import { Spinner } from "@/components/Spinner";
 import { CreateStdOAuthCredential } from "@/components/credentials/actions/CreateStdOAuthCredential";
-import { Card } from "../ui/card";
+import { Card } from "@/components/ui/card";
```
raunakab: Ideally, you use `@opal/components.Card` instead!

```tsx
import { Card } from "@opal/components";
```

web/src/lib/credentials/utils.test.ts
raunakab, line 1: Thanks for adding tests!

raunakab (commented):
Looks good!

A general "trend" that the FE team has been pushing towards recently is organizing all items into `lib/${feature}/` directories. For example, I'm working on auth-related items right now, and all hooks / functions / utils / API-calls / types are all under `lib/auth/`.

For instance, if anyone ever needs to reach for token-refreshing logic (an auth-related item), they can just peruse through `lib/auth/` (or tell their AI-agent to peruse through it instead).

Given that credentials are a part of the "connectors" feature, I would encourage you to create a `lib/connectors/` directory, and throw all helpers in there. This will be the direction that I'll be moving towards in the near future anyways.

The convention is this:
- Hooks go inside of `lib/${feature}/hooks.ts`.
- API functions go inside of `lib/${feature}/svc.ts` (short for service).
- Types go inside of `lib/${feature}/types.ts`.
- Utility functions go inside of `lib/${feature}/utils.ts` (assuming, of course, that these are a *feature-specific utilities*, and not *generic utilities* [think time-based utilities] which should belong at a more global scope [potentially even in Opal]).
- Helper components go inside of `lib/${feature}/components.tsx`.

For this case, I would move those "credential" components to `lib/connectors/components.tsx` and those lib functions to `lib/connectors/{hooks.ts, utils.ts, svc.ts}`.

If you think that `lib/connectors/` is too broad (which is a fair point), you could name it instead `lib/credentials/`. The general drive, nonetheless, should be to place feature-specific items inside of a corresponding `lib/${feature}/`!

raunakab:
Since the only remaining items are organizational, I trust that you'll make the right choice. I'll approve to unblock.

raunakab (approved):

PR #12899, opened 2026-07-11
feat(connectors): add LumApps connector

siriusfreak:
## Description

Adds a **LumApps** connector (LumApps is a widely-used enterprise intranet / employee-experience platform). It ingests published intranet content — pages, news, and custom content types — into Onyx.

Highlights:

- **Checkpointed, incremental indexing**: polls `content/list` sorted by `-updatedAt` and breaks early once items fall below the poll window, so recurring syncs only touch changed content. The opaque LumApps cursor is kept in the checkpoint for within-run pagination only.
- **LIVE-only ingestion, enforced on both paths**: the API-side `status` filter is not trusted on its own — each item is re-checked on the indexing *and* the pruning path, so content that leaves `LIVE` (drafts, archived, expired) gets pruned.
- **Generic metadata-label ingestion**: all LumApps metadata families are carried over as Onyx metadata (`{family_key: [labels]}`), resolved lazily with per-run caches. Whatever taxonomy an organization uses (country, department, topic, …) becomes filterable tags without connector changes.
- **Safe pruning**: `retrieve_all_slim_docs` enumerates exactly the scope indexing covers, and the client raises on any non-200, so pruning never acts on a truncated enumeration.
- **No new dependencies** — plain `requests` + `beautifulsoup4`.

Config: cell/API base URL + organization id (required); instance ids, custom content type ids, and language as advanced options. Credentials: application id, API key, and a service user email to index on behalf of.

This connector has been running in production on our self-hosted Onyx deployment.

## How Has This Been Tested?

- Unit tests: `backend/tests/unit/onyx/connectors/lumapps/` — pure helpers (language picking, template body extraction, metadata-label resolution) and connector behavior (pagination/checkpointing, LIVE filtering, document conversion) with a mocked HTTP client. All pass, along with the full `tests/unit/onyx/connectors` suite (837 passed).
- `ruff` / `ruff format` / `ty check` / lazy-import check clean; web `tsgo --noEmit`, `oxlint`, and `oxfmt` clean.
- Indexing, incremental re-sync, and pruning verified against a real LumApps tenant on our production deployment (fork of Onyx v4.2.2/v4.3.2, same connector code).

## Additional Options

---
## Summary by cubic
Adds a checkpointed LumApps connector to ingest published intranet pages, news, and custom content into Onyx with incremental sync and safe pruning. Includes UI config, credentials, and icon so teams can set it up in the app.

- **New Features**
  - Incremental indexing via `content/list` sorted by `-updatedAt`, with an early stop once items fall below the poll window; the cursor is used only within a run.
  - LIVE-only ingestion enforced per item on indexing and pruning, so drafts/archived/expired content is skipped and removed.
  - Generic metadata ingestion: all families become `{family_key: [labels]}`, with a stable fallback for non‑Latin family names (`metadata_<family_id>`); resolved via per-run caches.
  - Slim retrieval enumerates the same scope as indexing, uses the same id-or-uid fallback as indexing, and aborts on non-200 to prevent partial pruning.
  - JWT auth using application ID + API key on behalf of a service user, with token refresh and resilient retries (network errors and 5xx/429 backoff; 401 re-mints token, others reuse it). Non-https `base_url` is rejected.

- **Migration**
  - In Sources, add a new LumApps connector with `base_url` (https cell API host) and `organization_id`.
  - Add credentials: `lumapps_application_id`, `lumapps_api_key`, `lumapps_service_user`.
  - Optional: restrict to `instance_ids`, `custom_content_types`, and set `lang` (default `en`).

<sup>Written for commit 4242ce694e4397cbe3e1d496c295726e3e003ca3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12899?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/configs/constants.py
+++ b/backend/onyx/configs/constants.py
@@ -263,6 +263,7 @@ class DocumentSource(str, Enum):
     BITBUCKET = "bitbucket"
     TESTRAIL = "testrail"
     BRAINTRUST = "braintrust"
+    LUMAPPS = "lumapps"
 
     # Special case just for integration tests
     MOCK_CONNECTOR = "mock_connector"
@@ -749,4 +750,5 @@ DocumentSourceDescription: dict[DocumentSource, str] = {
     DocumentSource.IMAP: "Email messages and threads",
     DocumentSource.TESTRAIL: "Test cases and QA management",
     DocumentSource.BRAINTRUST: "LLM eval experiments, datasets, and prompts",
+    DocumentSource.LUMAPPS: "Intranet pages, news, and content",
 }
--- a/backend/onyx/connectors/lumapps/client.py
+++ b/backend/onyx/connectors/lumapps/client.py
@@ -95,12 +95,8 @@ class OnyxLumApps:
             time.monotonic() + max(expires_in, 120) - _TOKEN_REFRESH_SKEW_SECONDS
         )
 
-    def _bearer(self, force_refresh: bool = False) -> str:
-        if (
-            force_refresh
-            or not self._token
-            or time.monotonic() >= self._token_expiry_monotonic
-        ):
+    def _bearer(self) -> str:
+        if not self._token or time.monotonic() >= self._token_expiry_monotonic:
             self._mint_token()
         assert self._token is not None
         return self._token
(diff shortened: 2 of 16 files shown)


siriusfreak:
Addressed all review feedback in 4242ce694:

- **Slim/uid mismatch (greptile P1)**: slim enumeration now requests `uid` and uses the same `id or uid` fallback as indexing, so a document indexed under its `uid` can never be wrongly pruned. Covered by a new unit test.
- **http:// base URL (cubic P1)**: non-https base URLs are rejected with a `ConnectorValidationError` before the client is built (credentials travel with every token request). Covered by a unit test.
- **Network errors not retried (cubic P2)**: `requests.RequestException` (connection resets, timeouts) is now retried with the same backoff as 5xx, and exhausted retries surface as `LumAppsClientError` so validation can translate them. Covered by unit tests.
- **Unnecessary token re-mint on 429/5xx retry (cubic P2)**: dropped the force-refresh on attempt 1 — a 401 already invalidates the cached token, so rate-limit/server-error retries reuse it. Covered by unit tests (429 does not re-mint; 401 re-mints exactly once).
- **Non-Latin family names collapsing (cubic P2)**: `slugify_family_key` now takes a per-family fallback; families whose names slugify to nothing get a stable `metadata_<family_id>` key instead of all collapsing into `metadata`. Covered by a unit test.

Note on CI: all pipeline jobs currently fail at the "Log in to ECR pull-through cache" step because the `ECR_REGISTRY` repository variable isn't available to PRs from forks — the failures are unrelated to this change.

Bo-Onyx:
Please check this internal pr https://github.com/onyx-dot-app/onyx/pull/12999

PR #12900, opened 2026-07-11
fix: support IPv6-only Kubernetes clusters (Craft sandbox)

siriusfreak:
## Description

Makes Onyx deployable on **IPv6-only Kubernetes clusters** (pods get a global IPv6 address and no IPv4). Today several components bind or resolve IPv4-only and break silently there. We run Onyx in production on such a cluster; these are the fixes that were required, generalized so IPv4/dual-stack deployments see **no behavior change unless they opt in**.

> **Scope narrowed after rebase.** This PR originally also fixed the worker metrics server. That part is now obsolete: #13191 landed a dual-stack `/metrics` listener on `main` that is strictly better than what this PR proposed (it clears `IPV6_V6ONLY` explicitly rather than relying on the host's `bindv6only` sysctl). Rebasing onto `main` dropped our version in favor of it, so what remains here is the **Craft sandbox** half only.

**Craft sandbox egress proxy** (`onyx/sandbox_proxy/server.py`)
- The mitmproxy listener and the healthz server bound hardcoded `0.0.0.0`, so on IPv6-only clusters neither sandbox traffic nor the kubelet probe can reach the pod. Both now honor `SANDBOX_PROXY_LISTEN_HOST` (**default unchanged: `0.0.0.0`**). An IPv6 listen host binds `AF_INET6` with `V6ONLY` off, so dual-stack setups still serve IPv4-mapped clients.
- mitmproxy's `block` addon rejects clients connecting from "global" IPs (open-relay protection). IPv6-only clusters assign pods global-unicast addresses — there is no commonly-used IPv6 analogue of RFC1918 — so every legitimate sandbox client gets blocked. New opt-in `SANDBOX_PROXY_ALLOW_GLOBAL_CLIENTS` (**default `false`**, so the protection stays on everywhere else) for deployments whose pod ingress is already restricted by NetworkPolicy.
- The `gate` addon now unmaps IPv4-mapped peer IPs (`::ffff:a.b.c.d`) before identity lookup, so a dual-stack listener does not 403 legitimate IPv4 clients.

**Sandbox firewall bootstrap** (`image/firewall-init.sh`)
The proxy host was resolved with `getent ahostsv4` only, and the init **dies** when the proxy Service has an IPv6 ClusterIP ("could not resolve proxy host to an IPv4 address"). Resolution now tries A then AAAA, notes the address family, and installs the proxy ACCEPT rule in the matching table (`iptables`/`ip6tables`). Egress is default-DROP on **both** families regardless (with loopback + conntrack rules mirrored on both), and self-verify checks the proxy rule on the proxy's own family — so the lockdown guarantees are unchanged.

**Sandbox image** (`image/entrypoint.sh`, `image/sandbox_daemon/server.py`)
`opencode serve` and the sidecar push daemon bound `0.0.0.0`, making them unreachable from the api-server on IPv6-only clusters. Both now bind `::` (dual-stack on Linux, `net.ipv6.bindv6only=0` default); the opencode host is overridable via `OPENCODE_SERVE_HOSTNAME` for IPv4-only hosts.

## How Has This Been Tested?

- Unit tests added/updated:
  - `tests/unit/sandbox_proxy/test_healthz.py`: healthz server binds an IPv6 host and answers over IPv6 (skips when the host lacks IPv6), and the IPv4 default path is unchanged.
  - `tests/unit/sandbox_proxy/test_gate.py`: IPv4-mapped peer addresses resolve to the same identity as their plain IPv4 form.
- Full `tests/unit/sandbox_proxy` passes (206 tests) on this branch rebased onto `main`; `ruff check`, `ruff format --check`, and `shellcheck` (both touched scripts) are clean.
- The same changes have been running in production on an IPv6-only (IPv6-first) Kubernetes cluster in our Onyx fork: Craft sandboxes provision, egress through the locked-down proxy, and pass kubelet probes.

## Additional Options


siriusfreak:
Addressed review feedback in bae577ae3:

- **IPv4-mapped peers on a dual-stack listener (cubic P1)** — valid, fixed. With `SANDBOX_PROXY_LISTEN_HOST=::`, IPv4 sandbox peers arrive as `::ffff:a.b.c.d` while the identity cache stores canonical IPv4 pod IPs, so lookups would miss and every IPv4 sandbox would 403 as unidentified. `_extract_src_ip` now unmaps IPv4-mapped addresses before the lookup; covered by a parametrized unit test (plain IPv4 / IPv4-mapped / native IPv6).

- **"prometheus HTTP server does not switch its socket family" (greptile P1)** — this is not the case for the pinned `prometheus-client==0.23.1`: `start_http_server` → `start_wsgi_server` resolves the address family from the address via `getaddrinfo` (`_get_best_family`, added upstream in prometheus/client_python#870 following CPython's http.server), so `addr="::"` binds an `AF_INET6` socket. The `0.0.0.0` fallback only triggers when the `::` bind itself raises (IPv6-disabled host), where IPv4 is the only remaining option — an explicitly configured `PROMETHEUS_METRICS_HOST` is never rewritten. To make this verifiable in CI rather than argued, I added a real-socket regression test: `start_http_server(0, addr="::")` must bind `AF_INET6` and answer `GET /metrics` over `[::1]` (skips on hosts without IPv6). This setup has also been serving scrapes in production on an IPv6-only cluster for months.

backend/onyx/server/features/build/sandbox/image/sandbox_daemon/server.py
```diff
@@ -392,4 +392,8 @@ async def opencode_history_mark_restored(
 if __name__ == "__main__":
     # TODO(security): bind to 127.0.0.1 and front with an in-pod proxy, or
     # restrict the listener to the sandbox network namespace.
-    uvicorn.run(app, host="0.0.0.0", port=PUSH_DAEMON_PORT)  # noqa: S104
+    # Bind to "::" (all interfaces, IPv6) rather than "0.0.0.0" so the
+    # readiness probe and the api server reach the daemon on IPv6-only
+    # clusters. On Linux (net.ipv6.bindv6only=0, the default) a "::" socket
+    # also accepts IPv4-mapped connections, so this stays dual-stack.
+    uvicorn.run(app, host="::", port=PUSH_DAEMON_PORT)  # noqa: S104
```
wenxi-onyx, line 399: Please make IPv6 binding opt-in and retain `0.0.0.0` as the default. With the pinned Uvicorn/uvloop versions, `host="::"` creates an IPv6-only listener. IPv4 connections fail, breaking existing sandbox probes and sidecar access without any configuration change. Add a daemon listen-host environment variable. Please also retain `0.0.0.0` as the default for `OPENCODE_SERVE_HOSTNAME` in `entrypoint.sh`. Existing users should keep IPv4 unless they explicitly enable IPv6.

backend/onyx/server/features/build/sandbox/image/firewall-init.sh
```diff
@@ -59,34 +62,47 @@ step_install_ca() {
...
-    iptables -F OUTPUT
-    iptables -P OUTPUT DROP
-    iptables -P INPUT ACCEPT
-    iptables -P FORWARD DROP
 
-    iptables -A OUTPUT -o lo -j ACCEPT
-    iptables -A OUTPUT -m conntrack --ctstate ESTABLISHED,RELATED -j ACCEPT
-    iptables -A OUTPUT -p tcp -d "$PROXY_IP" --dport "$SANDBOX_PROXY_PORT" -j ACCEPT
     iptables -A OUTPUT -j REJECT --reject-with icmp-admin-prohibited
+    ip6tables -A OUTPUT -j REJECT --reject-with icmp6-adm-prohibited
```
wenxi-onyx, line 103: The IPv6 OUTPUT rules block neighbor discovery, which can prevent connections even to the explicitly allowed proxy. I reproduced this in an isolated Linux IPv6 network: self-verification passed, but the proxy connection timed out. Allowing ICMPv6 neighbor solicitations and advertisements restored connectivity. Please permit the required IPv6 control traffic and test connectivity with an empty neighbor cache.

backend/onyx/sandbox_proxy/server.py
```diff
@@ -143,7 +170,7 @@ def _factory(tenant_id: str) -> CacheBackend:
 
 def _build_mitm_options() -> Options:
     return Options(
-        listen_host="0.0.0.0",  # noqa: S104 — container scope; pod network only
+        listen_host=SANDBOX_PROXY_LISTEN_HOST,  # noqa: S104 — container scope
```
wenxi-onyx, line 173: `SANDBOX_PROXY_LISTEN_HOST=::` creates an IPv6-only mitmproxy listener, not a dual-stack listener. Pinned mitmproxy uses `asyncio.start_server`; IPv4 connections are refused before identity lookup. Only the health server explicitly clears `IPV6_V6ONLY`. An opt-in IPv6-only mode is fine, but please remove the dual-stack claim or implement listeners for both families. Add a real proxy-port connection test for the intended behavior.

PR #12901, opened 2026-07-11
feat(web-connector): URL rewrite rules for stored document links

siriusfreak:
## Description

Adds an optional **URL Rewrites** setting to the web connector: a list of `[source_prefix, target_prefix]` rules applied to the URL a document is **stored** under (document id, section link, semantic identifier). Fetching still uses the original URL — only what gets persisted is rewritten.

The admin form renders each rule as two side-by-side fields (source prefix / replacement prefix) via a new reusable `string_pair_list` option type + `StringPairListInput` component; `"source -> target"` strings are also accepted by the backend for configs written by hand against the API.

**Use case:** crawling a site through an internal mirror, a staging host, or a pre-rendering proxy while storing the canonical public links — so citations in chat point users at the real site instead of the mirror. We run this in production to index docs sites that are only reachable from inside our network through a mirror host.

Behavior details:

- First matching prefix wins; only the **leading** prefix is replaced (a URL that merely *contains* the source prefix later in its path/query is untouched).
- Slim documents get the same rewritten id as indexed documents, so pruning enumeration stays consistent with what indexing stored — no false pruning.
- Invalid rules (malformed entries, empty source/target) are skipped with a warning; the field is optional and defaults to no rewrites, so existing connectors are unaffected.

Backend: `_parse_url_rewrites` / `_rewrite_url` helpers + `url_rewrites` connector param threaded through `ScrapeSessionContext`. Frontend: new `string_pair_list` option type rendered by `StringPairListInput` (two inputs per row, add/remove), used by the web connector form (`url_rewrites` in `WebConfig`).

## How Has This Been Tested?

- New unit tests (`backend/tests/unit/onyx/connectors/web/test_url_rewrites.py`): pair and arrow-string parsing incl. invalid-entry skipping, first-match-wins ordering, leading-prefix-only replacement, no-match passthrough, and connector wiring/defaults. Full `tests/unit/onyx/connectors/web` suite passes (20).
- `ruff` / `ruff format` / `ty check` clean; web `tsgo --noEmit`, `oxlint`, `oxfmt` clean.
- The same feature has been running in production in our Onyx fork: mirrored crawls store canonical public URLs and re-syncs/pruning behave correctly.

## Additional Options

### DIFF
--- a/backend/onyx/connectors/web/connector.py
+++ b/backend/onyx/connectors/web/connector.py
@@ -103,6 +103,10 @@ class ScrapeSessionContext:
         self.to_visit = to_visit
         self.url_rewrites = url_rewrites or {}
         self.visited_links: set[str] = set()
+        # Ids already emitted this run. Rewrite rules can map two fetched URLs
+        # onto the same storage id; only the first one wins (a duplicate id
+        # would overwrite the earlier document in the index).
+        self.emitted_doc_ids: set[str] = set()
         self.content_hashes: set[int] = set()
 
         self.at_least_one_doc: bool = False
@@ -734,6 +738,15 @@ class WebConnector(LoadConnector, SlimConnector):
                         continue
 
                     if result.doc:
+                        if result.doc.id in session_ctx.emitted_doc_ids:
+                            logger.warning(
+                                "Skipping %s: rewritten document id %s was "
+                                "already emitted by another URL this run",
+                                initial_url,
+                                result.doc.id,
+                            )
+                            continue
+                        session_ctx.emitted_doc_ids.add(result.doc.id)
                         batch.append(
                             SlimDocument(id=result.doc.id) if slim else result.doc
                         )
--- a/backend/onyx/connectors/web/connector.py
+++ b/backend/onyx/connectors/web/connector.py
@@ -48,12 +48,42 @@ from shared_configs.configs import MULTI_TENANT
 logger = setup_logger()
 
 
+def _parse_url_rewrites(raw_rewrites: list[str]) -> dict[str, str]:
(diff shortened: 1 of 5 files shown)


siriusfreak:
Addressed review feedback in 8201416b9:

- **Rewrites can collapse ids (greptile P2)** — valid, fixed. Two fetched URLs can rewrite to the same storage id (e.g. a run that crawls both the mirror and the canonical site), and the later document would silently overwrite the earlier one in the index. The scrape session now tracks emitted document ids and keeps only the first, logging a warning for the skipped URL. Without rewrite rules this changes nothing (`visited_links` already dedups fetch URLs). Covered by a unit test that drives `load_from_state` with two URLs rewriting to one id and asserts a single emitted document.

cubic reported no issues (5/5).

siriusfreak:
Addressed cubic's follow-up in 52cf09a91: **duplicate-id skip retried the scrape (P2)** — valid, fixed. The guard lives inside the per-URL retry loop, so `continue` re-scraped the same URL up to `MAX_RETRIES - 1` more times (full Playwright page loads plus backoff sleeps) even though the outcome can't change. It now `break`s out of the retry loop immediately — skip-and-move-on, as suggested.

backend/onyx/connectors/web/connector.py
```diff
@@ -353,13 +405,15 @@ def __init__(
         mintlify_cleanup: bool = True,  # Mostly ok to apply to other websites as well
         batch_size: int = INDEX_BATCH_SIZE,
         scroll_before_scraping: bool = False,
+        url_rewrites: list[list[str] | str] | None = None,
```
Danelegend, line 405: Why do we need to `| str` part. If we keep deprecate the string handling, would that cause problems?

PR #12902, opened 2026-07-11
feat(gitlab): clone-based code indexing with incremental sync and glob filtering

siriusfreak:
## Description

Rewrites the GitLab connector's **code-file indexing** and makes the file filters **editable after creation**. MR/issue indexing is untouched (keeps `_gitlab_datetime_to_utc` and `doc_created_at`; `include_mrs`/`include_issues` defaults unchanged).

**Clone-based indexing instead of per-file API fetches.** The current implementation walks `repository_tree` and calls `project.files.get` for every file (N+1 API calls, easily rate-limited on large repos), and stamps every document with `doc_updated_at=datetime.now()`. This PR clones the repo once (honoring new `clone_depth` / `branch` options) and reads files from disk, with `doc_updated_at` taken from each file's real last-commit date.

**Incremental sync.** Poll windows clone with `--shallow-since` and re-index **only files changed since the previous sync** (`git diff --name-only --diff-filter=d` against the boundary commit). A window starting at/before 2005 is treated as "index from scratch" (Onyx signals a first/full index with an epoch start). If no boundary commit can be resolved, it safely falls back to a full index.

**Glob-based file selection.** `code_file_patterns` (defaults cover common code/docs extensions *plus extensionless files like `Makefile`/`Dockerfile`*, which the old extension list could never match), and `include_path_patterns` / `exclude_path_patterns` for path scoping. Legacy `code_file_extensions` configs are auto-converted to equivalent globs, and the legacy `.*` regex is rewritten to `*`, so existing connectors keep working unchanged.

**Editable filters after creation.** New allowlisted `PUT /admin/cc-pair/{id}/connector-config` endpoint (only filter keys may be edited — identity keys like `project_owner` stay immutable to avoid orphaning documents) + an "Edit indexed file types" modal on the connector status page. Saving triggers a re-index so the new filter applies repo-wide, not just to files that change later. Also adds a generic "Use defaults" button for list fields (`defaultValues` on `ListOption`) so users can reveal and edit the built-in pattern list.

**⚠️ Behavior note for maintainers:** code-file document ids change from the git **blob SHA** to a stable path-based id. Blob-SHA ids produce a *new* document on every file edit while the old one lingers (the connector has no pruning), and identical files share one id across the repo. With path-based ids a changed file updates in place. Code files indexed under old blob-SHA ids will remain until the connector is re-created — happy to add a migration note or docs wording wherever you prefer.

## How Has This Been Tested?

- New unit tests (`backend/tests/unit/onyx/connectors/gitlab/test_gitlab_incremental.py`, 11 tests): full-index vs incremental window detection, changed-path computation against a real temporary git repo (boundary commit resolution, deletion filtering, fallback-to-full when history is too new), glob matching (basename vs path patterns, extensionless files), legacy extension/`.*` migration, and candidate-file restriction.
- Full `tests/unit/onyx/connectors` suite passes (840 passed).
- `ruff` / `ruff format` / `ty check` / lazy-import check clean; web `tsgo --noEmit`, `oxlint`, `oxfmt` clean.
- This connector (same code) has been indexing large internal GitLab repos in production in our Onyx fork for months, including incremental re-syncs and post-creation filter edits.

## Additional Options

---
## Summary by cubic
Switches GitLab code-file indexing to a single repo clone with incremental sync and glob-based filters you can edit after setup. This removes N+1 API calls, uses real commit times when available, keeps documents stable with path-based IDs, and ignores symlinks that resolve outside the repo.

- **New Features**
  - Clone once and read files from disk; supports `clone_depth` and optional `branch` (empty uses the project’s default); file `doc_updated_at` uses the last commit time when history is present and stays unset on shallow full clones (content hash drives updates).
  - Incremental sync re-indexes only files changed since the previous run; windows starting at/before 2005 trigger a full index; safely falls back to full when the boundary commit can’t be resolved.
  - Glob-based selection via `code_file_patterns`, plus `include_path_patterns`/`exclude_path_patterns`; defaults cover common code, docs, and extensionless files. Legacy `code_file_extensions` are auto-converted to globs and `.*` is rewritten to `*`.
  - Filters editable post-creation via `PUT /admin/cc-pair/{id}/connector-config` and the “Edit indexed file types” modal; the modal submits only changed fields and skips the update if nothing changed; a “Use defaults” button is available for list fields. Saving schedules a repo-wide re-index.
  - Code-file document IDs move from blob SHA to path-based IDs so changes update in place; previously indexed SHA-based docs are not auto-pruned.

- **Bug Fixes**
  - Skip symlinks and any path that resolves outside the checkout for both full scans and incremental candidates to prevent leaking host files.
  - Hardened config endpoint with a typed allowlist, rejects non-GitLab connectors, and commits the config update and re-index trigger atomically.
  - “Edit indexed file types” is shown only to users who can edit the connector.

<sup>Written for commit a6016b1fe08d86b95e37e2498ea1d9c42e0cd754. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12902?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/connectors/gitlab/connector.py
+++ b/backend/onyx/connectors/gitlab/connector.py
@@ -285,9 +285,12 @@ def _convert_file_to_document(
     )
     logger.debug("Generated file ID: %s", file_id)
 
-    # Use last commit date if available, otherwise current time - ensure UTC
-    doc_updated_at = _ensure_utc_datetime(last_commit_date)
-    logger.debug("Document updated_at: %s", doc_updated_at)
+    # Leave doc_updated_at unset when no reliable commit date is available —
+    # the indexing pipeline's content hash then governs change detection. A
+    # fabricated now() timestamp would mark every doc as updated on each run.
+    doc_updated_at = (
+        _ensure_utc_datetime(last_commit_date) if last_commit_date else None
+    )
 
     # Create and return a Document object
     doc = Document(
@@ -345,25 +348,16 @@ def _should_exclude_by_glob(path: str, exclude_patterns: list[str]) -> bool:
     return _matches_any_glob(path, exclude_patterns)
 
 
-def _get_file_last_commit_date(repo: git.Repo, file_path: str) -> datetime:
-    """Get the last commit date for a specific file."""
-    logger.debug("Getting last commit date for file: %s", file_path)
+def _get_file_last_commit_date(repo: git.Repo, file_path: str) -> datetime | None:
+    """Get the last commit date for a specific file (None when unavailable)."""
     try:
         commits = list(repo.iter_commits(paths=file_path, max_count=1))
         if commits:
-            commit_date = commits[0].committed_datetime
-            # Ensure timezone info is present and convert to UTC
-            commit_date = _ensure_utc_datetime(commit_date)
-            logger.debug("Last commit date for '%s': %s", file_path, commit_date)
-            return commit_date
-        else:
-            logger.debug("No commits found for file: %s", file_path)
+            return _ensure_utc_datetime(commits[0].committed_datetime)
+        logger.debug("No commits found for file: %s", file_path)
     except Exception as e:
         logger.warning("Could not get commit date for %s: %s", file_path, e)
-
-    fallback_date = datetime.now(timezone.utc)
-    logger.debug("Using fallback date for '%s': %s", file_path, fallback_date)
-    return fallback_date
+    return None
 
 
 class GitlabConnector(LoadConnector, PollConnector):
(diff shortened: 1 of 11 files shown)


siriusfreak:
Addressed greptile's feedback in ac441ada0:

- **Hardcoded branch default (P1)** — valid, fixed. The `branch` field now defaults to empty (marked optional, with wording pointing at the project's default branch), so the backend's `self.branch or project.default_branch` fallback stays in effect for new connectors; repos whose default branch isn't `main` clone correctly.
- **Modal persisting an untouched checkbox (P2)** — valid, fixed with slightly different semantics than the suggestion: defaulting the checkbox to `true` would be just as wrong for deployments where `GITLAB_CONNECTOR_INCLUDE_CODE_FILES` is unset/false. Instead the modal now submits **only fields the user actually changed** — an untouched checkbox is never written, so a connector that omits `include_code_files` keeps following the backend env default; if nothing changed, no PUT (and no re-index) happens at all.

siriusfreak:
Addressed cubic's review in bd89e479d (greptile's two points were fixed in ac441ada0):

- **Symlink escape (P1)** — valid, fixed. Both the full-scan and incremental candidate paths now go through a guard that rejects symlinks and any path resolving outside the checkout, so a repository-controlled symlink (e.g. `leak.py -> /etc/hostname`) can't publish host files into the index. Covered by a unit test exercising both paths.
- **Generic endpoint writes GitLab-only keys anywhere (P1)** — valid, fixed. The request is now a typed schema (`GitlabFilterConfigUpdateRequest`) — the schema itself is the allowlist and validates value types — and the endpoint rejects non-GitLab connectors, since the connector factory expands `connector_specific_config` into constructors that don't accept these keys.
- **`Boolean(undefined)` initializing the modal (P1)** — the harmful half was already fixed in ac441ada0: the modal submits only fields the user actually changed, so opening + saving never persists an untouched checkbox and the backend env default stays in effect. Defaulting the checkbox to `true` instead would be wrong for deployments where `GITLAB_CONNECTOR_INCLUDE_CODE_FILES` is unset/false; the display can't know the env value, but it can no longer do damage.
- **Shallow full clones fabricate timestamps (P2)** — valid, fixed more aggressively than suggested: on a shallow full clone the per-file lookup is skipped entirely and `doc_updated_at` is left unset (the `now()` fallback is gone too), so the indexing pipeline's content hash governs change detection — same approach the web connector uses. Incremental syncs keep real commit dates (`--shallow-since` history covers the changed files).
- **Read-only users can open the modal (P2)** — valid, fixed: the button renders only when `ccPair.is_editable_for_current_user`, consistent with the credential section.
- **Non-atomic config update + reindex trigger (P2)** — valid, fixed: `update_connector_specific_config_fields` no longer commits (flush only); the REINDEX trigger's commit lands both changes in one transaction, so a trigger failure rolls the config change back too.

PR #12903, opened 2026-07-11
fix(mcp): proactively refresh OAuth tokens for SSE transport

siriusfreak:
## Description

Fixes a hard expiry wall for **SSE-transport MCP servers using per-user OAuth**: the MCP SDK's `OAuthClientProvider` (an `httpx.Auth`) normally drives token refresh transparently, but it is disabled for SSE transport (`requires_response_body` is incompatible with an open SSE stream — see the existing comment in `mcp_client.py`). The stored `Authorization: Bearer …` header is therefore frozen, and once the access token expires **every SSE tool call fails with an auth error until the user manually reconnects the server** from the MCP dropdown. With short-lived tokens (e.g. corporate IdPs issuing 1-hour tokens) that means re-authenticating every hour.

This PR refreshes the token proactively, outside the transport:

- Before each SSE tool call, if the persisted expiry (`token_expires_at`, which already bakes in `TOKEN_EXPIRY_BUFFER_SECONDS`) says the token is expired, `refresh_mcp_oauth_token_if_expired` performs a plain `grant_type=refresh_token` exchange and persists the rotated tokens/expiry/header exactly the way `OnyxTokenStorage.set_tokens` does (including preserving the previous refresh token when the provider doesn't rotate it).
- The token endpoint is resolved the same way the SDK path does: the configured endpoint for `KNOWN_PROVIDER` servers, else the discovered endpoint persisted in the OAuth metadata. The endpoint is re-validated with `validate_oauth_endpoint_url` before the POST.
- Connections without a persisted expiry are deliberately left alone — refreshing on every call would hammer the token endpoint and needlessly rotate refresh tokens; those remain a one-time manual reconnect.
- Refresh failures are non-fatal: the call proceeds with the stored token, and the existing auth-error messaging handles the rest.

Non-SSE transports are unchanged (they keep the SDK-driven refresh).

## How Has This Been Tested?

- New unit tests (`backend/tests/unit/onyx/server/features/mcp/test_mcp_oauth_refresh.py`): a correct `grant_type=refresh_token` exchange is sent and the rotated tokens + `Authorization` header are persisted; a still-valid token performs no refresh; a missing persisted expiry performs no refresh; a missing refresh token / client info / token endpoint each safely no-op; the previous refresh token is preserved when the provider response omits one.
- Full `tests/unit/onyx/server/features/mcp` (38) and `tests/unit/onyx/tools` suites pass (523 total); `ruff` / `ruff format` / `ty check` clean.
- Running in production in our Onyx fork: an SSE MCP server (GitLab search behind a corporate IdP with 1-hour tokens) survives token expiry without manual reconnects.

## Additional Options

---
## Summary by cubic
Proactively refresh per-user OAuth tokens for SSE-transport MCP servers so calls keep working after access-token expiry, removing the need to manually reconnect. Non-SSE transports are unchanged; addresses Linear DST-1273.

- **Bug Fixes**
  - Before each SSE tool call, when `token_expires_at` is past, perform `grant_type=refresh_token`, persist rotated tokens/expiry, and merge a fresh `Authorization` header into existing headers; use it for the call.
  - Resolve the token endpoint from the configured KNOWN_PROVIDER endpoint or discovered OAuth metadata and validate it; preserve the previous refresh token if the provider doesn’t rotate it.
  - Serialize concurrent refreshes with a row lock on the connection-config to avoid races when providers rotate refresh tokens.
  - No-op when expiry is missing or when refresh token/client info/endpoint is absent; on errors, fall back to the stored token. Added unit tests, including header-merge behavior.

<sup>Written for commit ee9157709e3f9baabfe5bb68cf59e6a43b7da23a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12903?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/server/features/mcp/api.py
+++ b/backend/onyx/server/features/mcp/api.py
@@ -27,6 +27,7 @@ from mcp.types import InitializeResult
 from mcp.types import Tool as MCPLibTool
 from pydantic import AnyUrl
 from pydantic import BaseModel
+from sqlalchemy import select
 from sqlalchemy.orm import Session
 
 from onyx.auth.oauth_token_manager import build_oauth_authorization_url
@@ -428,6 +429,16 @@ def refresh_mcp_oauth_token_if_expired(
     the existing stored header. All failures raise; the caller is expected to
     treat a refresh failure as non-fatal.
     """
+    # Serialize concurrent refreshes for the same connection: providers that
+    # rotate refresh tokens invalidate the old one, so two racing exchanges
+    # would leave the loser with a dead token. The row lock is held until
+    # update_connection_config commits; a blocked second caller then re-reads
+    # a fresh (non-expired) expiry below and returns without refreshing.
+    db_session.execute(
+        select(MCPConnectionConfig)
+        .where(MCPConnectionConfig.id == connection_config_id)
+        .with_for_update()
+    ).scalar_one()
     config = get_connection_config_by_id(connection_config_id, db_session)
     config_data = extract_connection_data(config)
 
@@ -484,7 +495,12 @@ def refresh_mcp_oauth_token_if_expired(
         config_data.pop(MCPOAuthKeys.TOKEN_EXPIRES_AT.value, None)
 
     header_value = f"{new_tokens.token_type} {new_tokens.access_token}"
-    config_data["headers"] = {"Authorization": header_value}
+    # Merge, don't replace: the connection may carry static headers alongside
+    # the OAuth Authorization header.
+    config_data["headers"] = {
+        **config_data.get("headers", {}),
+        "Authorization": header_value,
+    }
     update_connection_config(config.id, db_session, config_data)
 
     logger.info(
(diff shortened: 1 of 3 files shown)


siriusfreak:
Addressed both review points in ee9157709:

- **Static headers dropped on refresh (P1)** — valid, fixed: the fresh `Authorization` header is merged into the stored headers map instead of replacing it, so static headers configured alongside OAuth survive rotation. Covered by a new unit test.
- **Refresh-token rotation race (P1)** — valid, fixed: the refresh now locks the connection-config row (`SELECT … FOR UPDATE`) for the duration of the exchange; the lock is released by `update_connection_config`'s commit. A second caller racing on the same connection blocks on the lock, then re-reads the already-updated expiry (double-checked) and returns without POSTing the old, now-invalidated refresh token.

evan-onyx:
#13085 has been merged!

PR #12906, opened 2026-07-11
feat(opensearch): env-tunable kNN knobs, derived_source opt-out, lowercase index names

siriusfreak:
## Description

Three small OpenSearch operability improvements for self-hosted deployments, all **default-preserving** (no behavior change unless a deployment opts in):

**1. Env-tunable HNSW parameters.** `EF_CONSTRUCTION` / `M` / `EF_SEARCH` become overridable via `OPENSEARCH_EF_CONSTRUCTION` / `OPENSEARCH_M` / `OPENSEARCH_EF_SEARCH`, following the existing `OPENSEARCH_INDEX_NUM_SHARDS` / `NUM_REPLICAS` pattern. The recall/latency/memory trade-offs behind these knobs differ enough across corpus sizes and hardware that fixed values don't fit every deployment — we tune them per-cluster in production. Defaults are unchanged.

**2. `derived_source` opt-out.** `OPENSEARCH_KNN_DERIVED_SOURCE_ENABLED` (default `true`) lets a deployment disable derived `_source` at index-creation time. OpenSearch 3.x enables it by default for knn indices, and under sustained indexing load we hit the transient update failures the client already retries (opensearch-project/k-NN#3191, the same issue referenced in the bulk-update retry comment in `client.py`). The retry mostly covers it, but an explicit escape hatch avoids depending on retries entirely for deployments that keep hitting it. The settings payload only gains the key when the opt-out is set, so default index creation is byte-identical.

**3. Lowercase index names at the client boundary.** OpenSearch requires lowercase index names, and embedding model names — the main input to index naming (`danswer_chunk_{clean_model_name(...)}`) — can carry uppercase characters (e.g. `Qwen/Qwen3-Embedding-8B`), which otherwise fail index creation outright. Normalizing in `OpenSearchIndexClient.__init__` keeps the fix scoped to OpenSearch; Vespa index naming is untouched, and `clean_model_name` itself is unchanged to avoid re-mapping existing deployments' index names.

## How Has This Been Tested?

- New unit tests: default settings do not emit the `derived_source` key (payload unchanged); the opt-out emits `knn.derived_source.enabled: false`. Full `tests/unit/onyx/document_index` suite passes (71).
- `ruff` / `ruff format` / `ty check` clean.
- All three changes run in production in our Onyx fork (OpenSearch 3.x, uppercase embedding-model index names, tuned EF values).

## Additional Options

---
## Summary by cubic
Adds env-tunable kNN knobs, an opt-out for derived `_source`, and lowercases OpenSearch index names with a warning about normalization. Defaults stay the same.

- **New Features**
  - HNSW knobs are env-configurable: EF_CONSTRUCTION, M, EF_SEARCH via `OPENSEARCH_EF_CONSTRUCTION`, `OPENSEARCH_M`, `OPENSEARCH_EF_SEARCH`; defaults unchanged.
  - `OPENSEARCH_KNN_DERIVED_SOURCE_ENABLED` (default true) lets you disable derived `_source` at index creation; key only added when set to false.
  - Index names are lowercased at the client boundary and a warning is logged on normalization (to surface potential case-collisions); Vespa naming is unaffected.

<sup>Written for commit abf43869ff6f7bc35b92fcaffe2a88058c3b2eb8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12906?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


siriusfreak:
Addressed the shared greptile/cubic finding in 884a609f3 — **case-only index-name collision after lowercasing**. Assessment: the hazard is real in theory but requires two configured embedding models whose names differ *only* by case, which no real model registry publishes; and the status quo for the uppercase variant is a hard index-creation failure, so the change strictly grows the set of working configurations. To keep the normalization observable rather than silent, the client now logs a loud warning whenever it lowercases a name (original → normalized), and the code comment documents the collision caveat explicitly. If you'd prefer a hard rejection instead of normalize-and-warn, happy to switch — but that would keep uppercase model names unusable on OpenSearch.

PR #12928, opened 2026-07-13
fix(ttl): isolate + batch chat retention cleanup off the primary queue

justin-tahara:
## Description

TTL chat-retention cleanup (`perform_ttl_management_task`) was enqueued on the default `celery` queue — the primary worker's only lane, shared with the recurring `check_for_indexing` beat task — and deleted every expired chat session in a single unbounded task (`get_chat_sessions_older_than` had no `LIMIT`).

On an instance with a large backlog (tens of thousands of sessions), one cleanup run pinned a primary worker thread for a long time. While it ran, `check_for_indexing` (every 15s) piled up behind it on the same queue and aged past its `expires` (`BEAT_EXPIRES_DEFAULT` = 15 min), so Celery dropped them — surfacing as `Discarding revoked task: check_for_indexing[...]` and silently stalled indexing. Flushing Redis cleared the backlog temporarily, then it recurred.

Separately, the `should_perform_chat_ttl_check` guard was ineffective (nothing ever called `register_task` for this task), so the hourly beat stacked a **new** full cleanup run on top of any still-draining one, compounding the congestion on large backlogs.

This PR:

- **Dedicated queue.** Routes `perform_ttl_management_task` to a new `chat_ttl_deletion` queue consumed by the **light** worker, and enqueues it with an `expires` (previously enqueued with none). Fully off the primary/indexing lane. Added to the light worker `-Q` in `supervisord.conf` and `celery-worker-light.yaml`.
- **Batched deletes.** `get_chat_sessions_older_than` gains an optional `limit` (oldest-first) and the task drains in bounded 100-row batches, so no run loads the whole backlog into memory or holds a thread indefinitely. The loop terminates on an empty or partial batch, and stops (with an error log) if a full batch makes zero progress, so it can't spin forever.
- **Single-run guard.** Wraps the run in a per-tenant Redis lock (reacquired per batch) so overlapping hourly runs can't stack.
- **Observability.** Reports `chat_ttl_deletion` queue depth in the monitoring task.

> **Deploy note:** the light celery worker must be restarted to pick up the new `chat_ttl_deletion` queue in its `-Q` list.

## How Has This Been Tested?

- Added `test_chat_retention_batched_deletion` (enterprise integration test) that forces a small batch size and runs the real task to verify a multi-batch backlog is fully drained and the loop terminates. Existing `test_chat_retention` cases are unchanged (the `get_chat_sessions_older_than` signature is backward compatible).
- `ruff check` + `ruff format --check` pass on all changed files.
- Full integration suite not run locally (requires a live enterprise deployment); relying on CI.

## Additional Options

---
## Summary by cubic
Moved chat TTL cleanup to a dedicated `chat_ttl_deletion` queue and run it in chained 100‑session batches so it won’t block the primary queue or stall `check_for_indexing`. Added a token‑fenced Redis lock and rolling‑upgrade safety to keep a single chain per tenant.

- **Bug Fixes**
  - Dispatcher (`check_ttl_management_task`): per‑tenant `SET NX` marker claimed with a token; starts a chain only when not set; enqueues the first task to `chat_ttl_deletion` with 1h `expires` and low priority.
  - Worker (`perform_ttl_management_task`): deletes the oldest batch and, if full, enqueues the next with the same token; extends/releases the marker only if still owned; continues past per‑session errors; defaults an empty `chain_token` to safely no‑op legacy messages.
  - `get_chat_sessions_older_than`: added `limit` (oldest‑first) and float `days_old`.
  - Monitoring: added `chat_ttl_deletion` queue length metric.
  - Tests: cover full‑batch deletion and skipping a failing session.

- **Migration**
  - Restart the light Celery worker to pick up `-Q chat_ttl_deletion` (Helm chart bumped to 0.6.11 to roll the worker).

<sup>Written for commit 7e8cc3d88146ac621bb023011490466047cb272a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12928?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/ee/onyx/background/celery/tasks/ttl_management/tasks.py
+++ b/backend/ee/onyx/background/celery/tasks/ttl_management/tasks.py
@@ -54,6 +54,11 @@ def perform_ttl_management_task(
         logger.info("Chat TTL cleanup already in progress; skipping this run.")
         return
 
+    # Sessions that failed to delete are excluded from subsequent batches so a
+    # few undeletable rows can't block the rest of the (oldest-first) backlog.
+    # Every fetched session is either deleted or added here, so the eligible set
+    # shrinks each iteration and the loop always terminates.
+    failed_session_ids: set[UUID] = set()
     user_id: UUID | None = None
     session_id: UUID | None = None
     try:
@@ -61,13 +66,15 @@ def perform_ttl_management_task(
             lock.reacquire()
             with get_session_with_current_tenant() as db_session:
                 old_chat_sessions = get_chat_sessions_older_than(
-                    retention_limit_days, db_session, limit=_TTL_DELETE_BATCH_SIZE
+                    retention_limit_days,
+                    db_session,
+                    limit=_TTL_DELETE_BATCH_SIZE,
+                    exclude_session_ids=failed_session_ids,
                 )
 
             if not old_chat_sessions:
                 break
 
-            deleted_count = 0
             for user_id, session_id in old_chat_sessions:
                 lock.reacquire()
                 try:
@@ -79,27 +86,20 @@ def perform_ttl_management_task(
                             include_deleted=True,
                             hard_delete=True,
                         )
-                    deleted_count += 1
                 except Exception:
                     logger.exception(
                         "Failed to delete chat session user_id=%s session_id=%s, continuing with remaining sessions",
                         user_id,
                         session_id,
                     )
+                    failed_session_ids.add(session_id)
 
-            # If a full batch was fetched but nothing could be deleted, the same
-            # rows would be re-fetched forever. Stop and let the next scheduled
-            # check retry rather than spin.
-            if deleted_count == 0:
-                logger.error(
-                    "Chat TTL cleanup made no progress on a full batch of %d sessions; stopping.",
-                    len(old_chat_sessions),
-                )
-                break
-
-            # A partial batch means we've reached the tail of the backlog.
-            if len(old_chat_sessions) < _TTL_DELETE_BATCH_SIZE:
-                break
+        if failed_session_ids:
+            logger.error(
+                "Chat TTL cleanup finished but %d session(s) could not be deleted; "
+                "they will be retried on the next scheduled run.",
+                len(failed_session_ids),
+            )
 
     except Exception:
         logger.exception(
--- a/backend/ee/onyx/background/celery/tasks/ttl_management/tasks.py
+++ b/backend/ee/onyx/background/celery/tasks/ttl_management/tasks.py
@@ -1,13 +1,9 @@
 from celery import shared_task
 from celery import Task
-from redis.lock import Lock as RedisLock
 
 from ee.onyx.background.celery_utils import should_perform_chat_ttl_check
-from onyx.background.celery.celery_redis import celery_get_broker_client
-from onyx.background.celery.celery_redis import celery_get_queue_length
 from onyx.configs.app_configs import JOB_TIMEOUT
 from onyx.configs.constants import CELERY_CHAT_TTL_DELETE_TASK_EXPIRES
-from onyx.configs.constants import CELERY_GENERIC_BEAT_LOCK_TIMEOUT
 from onyx.configs.constants import CHAT_TTL_DELETE_BATCH_SIZE
 from onyx.configs.constants import OnyxCeleryPriority
 from onyx.configs.constants import OnyxCeleryQueues
@@ -34,7 +30,7 @@ def perform_ttl_management_task(
     self: Task,
     retention_limit_days: float,
     *,
-    tenant_id: str,  # noqa: ARG001
+    tenant_id: str,
 ) -> None:
     """Delete the oldest batch of expired chat sessions, then chain the next task.
 
@@ -42,9 +38,10 @@ def perform_ttl_management_task(
     expired sessions and exits, so it occupies a single light-worker thread for
     only one short batch and interleaves with the other light-queue work. If a
     full batch is deleted (more sessions likely remain) it enqueues the next task
-    to continue the chain. Only one task is ever in flight (``check_ttl_management_task``
-    starts a chain only when the queue is empty), so at most one light-worker
-    thread does TTL work at a time.
+    to continue the chain; otherwise the chain is done and it releases the
+    in-flight marker so ``check_ttl_management_task`` can start a fresh chain.
+    Only one chain runs per tenant at a time (guarded by the marker), so at most
+    one light-worker thread does TTL work at a time.
 
     NOTE: every run re-selects the oldest sessions, so a session that repeatedly
     fails to delete is retried on every run. If the oldest ``CHAT_TTL_DELETE_BATCH_SIZE``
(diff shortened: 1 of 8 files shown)


justin-tahara:
Good catch on the "Failed Batch Blocks Cleanup" (P1) — addressed in 88527ebe43.

The `deleted_count == 0` early-stop was replaced with a failed-id exclusion set: sessions that fail `delete_chat_session` are collected and passed to `get_chat_sessions_older_than(exclude_session_ids=...)` so subsequent oldest-first batches skip them and reach the newer expired sessions behind them. Every fetched session is now either deleted or excluded, so the eligible set strictly shrinks each iteration (loop still terminates without the old guard). Failed rows are logged and retried on the next scheduled run. Added `test_chat_retention_skips_failing_session` covering an older failing session not blocking newer deletable ones.

justin-tahara:
Re: cubic's P1 on the lock key not being tenant-scoped — this is a false positive, but I added a clarifying comment in 7374858de1 since the scoping is implicit.

`get_redis_client()` returns a `TenantRedisClient` whose `.lock(name)` prepends `"{tenant_id}:"` to the key (`_prefix_key` in `onyx/redis/tenant_redis_client.py`), so `CHAT_TTL_MANAGEMENT_LOCK` resolves to `{tenant_id}:da_lock:chat_ttl_management` — already per-tenant. The tenant comes from the current-task context (the same context this task's `get_session_with_current_tenant()` relies on). This is the identical pattern the other beat/singleton locks use (e.g. `check_for_vespa_sync_task` locks on the constant `CHECK_VESPA_SYNC_BEAT_LOCK`); manually appending `:{tenant_id}` would double-scope the key and diverge from that convention, so I kept the constant key and documented the implicit prefixing instead.

backend/onyx/configs/constants.py
```diff
@@ -435,6 +435,10 @@ class OnyxCeleryQueues:
     CONNECTOR_HIERARCHY_FETCHING = "connector_hierarchy_fetching"
     CSV_GENERATION = "csv_generation"
 
+    # Chat retention (TTL) hard-deletion queue. Kept off the primary "celery"
```
evan-onyx: cleanup comment

backend/ee/onyx/background/celery/tasks/ttl_management/tasks.py
```diff
@@ -32,30 +42,66 @@ def perform_ttl_management_task(
...
+    lock: RedisLock = r.lock(
+        OnyxRedisLocks.CHAT_TTL_MANAGEMENT_LOCK,
+        timeout=CELERY_GENERIC_BEAT_LOCK_TIMEOUT,
+    )
+    if not lock.acquire(blocking=False):
+        logger.info("Chat TTL cleanup already in progress; skipping this run.")
+        return
+
+    # Sessions that failed to delete are excluded from subsequent batches so a
+    # few undeletable rows can't block the rest of the (oldest-first) backlog.
```
evan-onyx: the comment seems to imply that future iterations of this task won't try deleting the sessions that previously failed to delete which is not true. I think the intent after discussing IRL is to have each task only delete the 100 oldest chat sessions, which seems better. We should also make note that this CAN end us up in an infinite loop (across tasks) if the oldest 100 can't be deleted for some reason.

evan-onyx (approved):
talked irl and there are some changes to make, approving to unblock u

justin-tahara:
Reworked to the design we discussed (eff740795c): **each task deletes the oldest 100 expired sessions and chains the next** — no fan-out, no per-session guards, no backpressure math.

@evan-onyx implemented your steer directly:
- Each `perform_ttl_management_task` deletes the oldest `CHAT_TTL_DELETE_BATCH_SIZE` (100) sessions and enqueues the next task only if a full batch came back. `check_ttl_management_task` starts a chain only when the queue is empty, so chains don't stack.
- Added the requested note in the docstring: because every run re-selects the oldest sessions, failed deletes are retried, and **if the oldest 100 can't be deleted the chain loops on them across tasks** — documented as an accepted trade-off.
- Fixed the misleading comment (the old one implied previously-failed sessions wouldn't be retried, which was inaccurate) and cleaned up the queue comment.

This also supersedes the two bot findings from the previous (fan-out) commit:
- **cubic** — the per-session guard timing and the queue-depth-cap concerns are moot; the guards and depth cap are gone entirely in the chain design.
- **greptile P1 ("oldest failures starve cleanup")** — this is now the *intended, documented* behavior per the design discussion, not a defect: we accept that a permanently-undeletable oldest-100 blocks the chain, rather than adding machinery to route around it.

Re-review welcome on the new commit.

justin-tahara:
Addressed both P1s from the latest re-review:

**cubic P1 (double-chain race) — fixed in ae4ad9b45e.** You're right that queue length is a leaky guard: an actively-running batch isn't in the queue, so a beat firing mid-batch saw length 0 and could start a second chain. Replaced the queue-length gate with a **per-tenant in-flight marker** (Redis key with TTL): the beat claims it via `SET NX` and only starts a chain if it isn't already held; each `perform` task refreshes it and releases it when the chain drains. The marker spans the whole chain regardless of queue state, so chains can't stack.

**greptile P1 ("Failing batches still loop") — working as intended, not a defect.** This is the exact behavior we deliberately chose with @evan-onyx: each task re-selects the oldest 100, so a permanently-undeletable oldest batch loops (and never reaches newer sessions). It's documented in the `perform_ttl_management_task` docstring as an accepted trade-off — we opted for simplicity over adding skip/offset machinery to route around poison rows, and every failure is logged. Not changing this one.

justin-tahara:
cubic P1 (lease-expiry race) — fixed in b881e2b2f4. Good find: if a batch outlived the marker's 1h TTL, the beat could start a replacement chain and the superseded task could still refresh/delete the *new* chain's marker.

Fixed with the owner-token approach you suggested: `check_ttl_management_task` claims the marker with a fresh per-chain token (`SET NX`), and that token is carried through the chained tasks. Each `perform` run does a **compare-and-refresh** (extend the lease only if the marker still equals our token, via a small Lua script) and stops if it doesn't own it; the marker is released via **compare-and-delete** (only if still owned). So a task whose lease lapsed can neither extend nor delete a later chain's lease — it just exits.

justin-tahara:
Two P2s from the latest cubic pass:

**#1 (rollout compat, tasks.py) — fixed in 7e8cc3d881.** Valid: a `perform_ttl_management_task` message enqueued by the pre-upgrade code carries no `chain_token`, so a new worker would fail it. Made `chain_token` default to `""` — an empty token never matches the fenced marker (`refresh-if-owned("")` returns 0 since the marker holds a uuid), so a legacy in-flight message harmlessly no-ops and the next beat starts a fresh chain. No task-name versioning needed.

**#2 (`_run_perform_ttl` uses plain `set()`, test) — intentional, not a defect.** The test helper deliberately claims the marker unconditionally with a known token so `perform` proceeds against a deterministic value. `SET NX` there would be *wrong* — a marker left over from a prior test would make the claim fail, the helper's token wouldn't match the stored one, and `perform` would early-return, breaking the test. The test isn't racing with any other chain, so there's nothing for NX to guard. Leaving as-is.

PR #12954, opened 2026-07-13
feat(indexing): resume indexing on graceful worker shutdown instead of failing

nmgarza5:
## Description

When a docfetching worker gets SIGTERM (rolling deploy or KEDA scale-down), its in-flight index attempt was recorded FAILED with "No heartbeat received for 1800 seconds"

**Fix:** a terminal-but-resumable `IndexingStatus.INTERRUPTED`, produced only by a `worker_shutting_down` handler that positively identifies a graceful shutdown. The watchdog marks the in-flight attempt INTERRUPTED and stops its subprocess before SIGKILL, so a fresh attempt resumes from checkpoint on the next beat (~15s).

- INTERRUPTED is excluded from the repeated-error auto-pause, and `should_reuse_checkpoint()` gates poll-window + checkpoint reuse so the resume stays checkpoint-correct
- the heartbeat validator is unchanged: a hard kill (OOM, node failure, no SIGTERM) still lands FAILED and auto-pauses a genuinely broken connector
- `terminationGracePeriodSeconds: 60` gives the watchdog time to mark and stop
- UI: neutral "Interrupted" badge (per-attempt and connector), plus a status filter
- no migration (the status column is a non-native enum)

## Additional Options

### DIFF
--- a/web/src/components/Status.tsx
+++ b/web/src/components/Status.tsx
@@ -1,5 +1,6 @@
 "use client";
 
+import { ReactNode } from "react";
 import { ValidStatuses } from "@/lib/types";
 import { Badge } from "@/components/ui/badge";
 import { timeAgo } from "@opal/time";
@@ -16,6 +17,22 @@ import {
 } from "@/app/admin/connector/[ccPairId]/types";
 import { Tooltip } from "@opal/components";
 
+// Wrap a status badge in a hover tooltip carrying its error/reason text, or
+// return the bare badge when there is no message.
+function badgeWithErrorTooltip(
+  icon: ReactNode,
+  errorMsg?: string | null
+): ReactNode {
+  if (!errorMsg) {
+    return icon;
+  }
+  return (
+    <Tooltip tooltip={errorMsg}>
+      <div className="cursor-pointer">{icon}</div>
+    </Tooltip>
+  );
+}
+
 export function IndexAttemptStatus({
   status,
   errorMsg,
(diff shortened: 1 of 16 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

web/src/components/Status.tsx
nmgarza5, line 1: just adding to this file for now. will be refactored soon im pretty sure

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
Addressed the review:
- P1 (TOCTOU in `mark_attempt_interrupted`): now re-checks `is_terminal()` after acquiring the `SELECT FOR UPDATE` lock, so a subprocess finishing SUCCESS in the race window is no longer clobbered to INTERRUPTED. Locked by `test_mark_attempt_interrupted_skips_already_terminal`.
- P2 (mid-fetch conflation): removed the beat-path reclassification entirely. INTERRUPTED is now produced only by the graceful `worker_shutting_down` handler; hard kills (OOM, no SIGTERM) stay FAILED and still auto-pause.
- UI: `CCPairStatus` now renders a neutral "Interrupted" badge instead of green "Indexed".

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

justin-tahara (approved):

PR #12956, opened 2026-07-13
feat(indexing): resume immediately on graceful worker shutdown

nmgarza5:
## Description

Base is `nikg/indexing-interrupted-status` (#12954), so this diff is only the graceful-shutdown commit.

**Why:** #12954 makes a killed docfetching worker resumable, but a hard kill is only noticed by the beat heartbeat validator up to 1800s later. This closes that gap for the graceful case.

**Fix:** on SIGTERM (rolling deploy / KEDA scale-down) a `worker_shutting_down` handler sets a process-global flag. The indexing watchdog polls it, marks its in-flight attempt INTERRUPTED (best-effort), and stops its subprocess before SIGKILL. A fresh attempt resumes from checkpoint on the next beat (~15s) instead of after ~1800s.

- new `TERMINATED_BY_WORKER_SHUTDOWN` watchdog status (set explicitly, no exit-code mapping)
- `terminationGracePeriodSeconds: 60` on the docfetching Deployment so the watchdog has time to mark and stop
- `acks_late` unchanged, so no broker redelivery. Recovery is still the beat re-create path, just faster.

## Additional Options

### DIFF
--- a/backend/onyx/background/celery/apps/docfetching.py
+++ b/backend/onyx/background/celery/apps/docfetching.py
@@ -9,8 +9,12 @@ from celery.signals import celeryd_init
 from celery.signals import worker_init
 from celery.signals import worker_ready
 from celery.signals import worker_shutdown
+from celery.signals import worker_shutting_down
 
 import onyx.background.celery.apps.app_base as app_base
+from onyx.background.celery.tasks.docfetching.worker_shutdown import (
+    signal_worker_shutting_down,
+)
 from onyx.configs.constants import POSTGRES_CELERY_WORKER_DOCFETCHING_APP_NAME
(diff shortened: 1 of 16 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
Collapsed into #12954: the two changes are one feature. Per Greptile's P2, the beat-path reclassification was removed (it conflated infra kills with genuine crashes), so INTERRUPTED is now produced only by the graceful worker_shutting_down handler. #12954 carries the full feature.

PR #12960, opened 2026-07-13
feat(craft): admin Craft section with editable workspace instructions

rohoswagger:
## Description

Gives Craft its own labeled **Craft** section in the admin sidebar and adds admin-editable workspace instructions for Craft agents.

**Admin IA**
- New sidebar section (admin-only, gated on `onyx_craft_available`) with three pages: **Access** (`/admin/craft/access`, the existing access controls), **Apps** (`/admin/craft/apps`, org app/skill management moved out of the Craft app), and **Instructions** (`/admin/craft/instructions`, new).
- Old routes redirect: `/admin/configuration/craft`, `/craft/v1/apps/manage`, `/craft/v1/apps/admin`. One-click round trip between the admin Apps page and the in-Craft user Apps page ("Manage apps" ↔ "Your apps in Craft").
- External-apps admin endpoints moved to the ungated `/build/admin` router (each already enforces `FULL_ADMIN_PANEL_ACCESS`) so admins **without personal Craft access** can still manage org apps. Final URLs are unchanged.

**Workspace instructions**
- New `craft_instructions` setting (4,000-char cap): admin-only writes via `PUT /admin/settings`, preserved when older clients omit the field, explicit `null` clears it, stripped from the `/settings` payload for non-admins.
- Injected into each session's rendered AGENTS.md as an `## Organization instructions` section (both Docker and K8s sandbox managers, including snapshot restore). Substitution runs last so admin text containing literal `{{...}}` template tokens is never expanded.
- New `GET /build/admin/base-instructions` endpoint so the editor can show the base prompt being appended to.
- Instructions editor: labeled textarea with live counter, explicit Save with dirty tracking + `beforeunload` guard, confirm-gated Reset to default, collapsible base-prompt view, "Try in Craft" shortcut.

## How Has This Been Tested?

Screenshot 2026-07-13 at 13 55 19

Screenshot 2026-07-13 at 13 55 39

- Unit: `test_agent_instructions.py` (11 tests — injection position, blank omission, token-literalness) and new `test_admin_put_settings.py` (4 tests — omit-preserves / null-clears / value-overwrites / over-length rejected).
- Integration: `GET /build/admin/base-instructions` added to the `ADMIN_ACCESS_ENDPOINTS` authz matrix.
- Live browser verification of all three pages, redirects, save/reset/dirty flows, and the apps admin page after the router split.
- Multi-lens review (correctness, security, tests, regressions, frontend style); confirmed findings fixed. Shell-escaping of the admin text through the sandbox `printf` path and tenant resolution of `load_settings()` were explicitly verified.
- `mypy`, `tsgo`, `ruff`, `oxfmt` clean.

**Known minor items** (flagged in review, intentionally left):
- Curators navigating directly to `/admin/craft/*` URLs render pages whose fetches 403 (the sidebar hides the links from them; same class of behavior as before, but a server-side gate could be added).
- `CraftManageLayout` now only wraps a redirect page and could be deleted in a follow-up cleanup.

## Additional Options

---
## Summary by cubic
Adds an admin Craft section with Access, Apps, and Instructions pages, plus workspace-wide instructions injected into every Craft agent’s prompt. This gives orgs consistent guardrails and lets admins manage apps without personal Craft access.

- **New Features**
  - Admin Craft section: `/admin/craft/access`, `/admin/craft/apps`, `/admin/craft/instructions`. Old routes redirect. Admin app endpoints live under `/build/admin` with a router‑level `FULL_ADMIN_PANEL_ACCESS` guard; one-click between the user Apps page and admin Apps (user “Manage apps” → admin; admin “Back to Craft” → user). The admin sidebar’s exit link is context-aware and returns to Craft when you’re in the Craft section.
  - Workspace instructions: new `craft_instructions` setting (max 4,000 chars) injected into AGENTS.md as “Organization instructions” (after Hard rules, before Environment). Literal `{{...}}` tokens are not expanded. Visible to authenticated users; admin-only writes. Omitted on PUT preserves value; `null` clears. `GET /build/admin/base-instructions` shows the base prompt. Editor has counter, Save with dirty guard, Reset to default, base‑prompt preview, “Try in Craft,” and error toasts.

- **Migration**
  - No API URL changes; admin apps endpoints remain under `/build/admin`. Redirects added from `/admin/configuration/craft`, `/craft/v1/apps/manage`, and `/craft/v1/apps/admin`.
  - Instructions apply to new or restored sessions only.

<sup>Written for commit 62fc67e37406ee628f240e765381be51f0f69b38. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12960?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/server/features/build/api.py
+++ b/backend/onyx/server/features/build/api.py
@@ -1,5 +1,6 @@
 from fastapi import APIRouter
 from fastapi import Depends
+from pydantic import BaseModel
 from sqlalchemy.orm import Session
 
 from onyx.auth.permissions import require_permission
@@ -10,6 +11,9 @@ from onyx.error_handling.error_codes import OnyxErrorCode
 from onyx.error_handling.exceptions import OnyxError
 from onyx.server.features.build.approvals.api import router as approvals_router
 from onyx.server.features.build.debug import router as debug_router
+from onyx.server.features.build.external_apps.api import (
+    admin_router as external_apps_admin_router,
+)
 from onyx.server.features.build.external_apps.api import router as external_apps_router
 from onyx.server.features.build.external_apps.oauth import (
     router as external_apps_oauth_router,
@@ -17,6 +21,9 @@ from onyx.server.features.build.external_apps.oauth import (
 from onyx.server.features.build.interactive_turns.api import router as turns_router
 from onyx.server.features.build.rate_limit import get_user_rate_limit_status
 from onyx.server.features.build.rate_limit import RateLimitResponse
+from onyx.server.features.build.sandbox.util.agent_instructions import (
+    AGENT_INSTRUCTIONS_TEMPLATE_PATH,
+)
 from onyx.server.features.build.scheduled_tasks.api import (
     router as scheduled_tasks_router,
 )
@@ -41,6 +48,29 @@ def require_onyx_craft_enabled(
 
 
 router = APIRouter(prefix="/build", dependencies=[Depends(require_onyx_craft_enabled)])
+
+# Admin-only Craft endpoints. Deliberately NOT behind the craft-enabled-for-user
+# gate: an admin configuring Craft may not have Craft enabled for themselves.
+admin_router = APIRouter(prefix="/build/admin")
+admin_router.include_router(external_apps_admin_router, tags=["build"])
+
+
+class BaseInstructionsResponse(BaseModel):
+    content: str
+
+
+@admin_router.get("/base-instructions")
+def get_base_instructions(
+    _: User = Depends(require_permission(Permission.FULL_ADMIN_PANEL_ACCESS)),
+) -> BaseInstructionsResponse:
+    """The base AGENTS.md template, so admins can see what their workspace
+    instructions are appended to. Dynamic sections appear as placeholders."""
+    if not AGENT_INSTRUCTIONS_TEMPLATE_PATH.exists():
+        raise OnyxError(OnyxErrorCode.NOT_FOUND, "Base instructions not found")
+    return BaseInstructionsResponse(
+        content=AGENT_INSTRUCTIONS_TEMPLATE_PATH.read_text()
+    )
+
 
 router.include_router(sessions_router, tags=["build"])
 router.include_router(messages_router, tags=["build"])
--- a/backend/onyx/server/features/build/api.py
+++ b/backend/onyx/server/features/build/api.py
@@ -51,7 +51,11 @@ router = APIRouter(prefix="/build", dependencies=[Depends(require_onyx_craft_ena
 
 # Admin-only Craft endpoints. Deliberately NOT behind the craft-enabled-for-user
 # gate: an admin configuring Craft may not have Craft enabled for themselves.
-admin_router = APIRouter(prefix="/build/admin")
+# The router-level permission guard backstops every route added here.
+admin_router = APIRouter(
+    prefix="/build/admin",
+    dependencies=[Depends(require_permission(Permission.FULL_ADMIN_PANEL_ACCESS))],
+)
 admin_router.include_router(external_apps_admin_router, tags=["build"])
 
 
--- a/backend/onyx/server/features/build/api.py
+++ b/backend/onyx/server/features/build/api.py
@@ -1,6 +1,5 @@
 from fastapi import APIRouter
 from fastapi import Depends
-from pydantic import BaseModel
 from sqlalchemy.orm import Session
 
 from onyx.auth.permissions import require_permission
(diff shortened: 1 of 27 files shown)


web/src/views/admin/CraftInstructionsPage/index.tsx
```diff
@@ -0,0 +1,239 @@
...
+    try {
+      await updateAdminSettings({
+        ...toSettings(settings),
+        craft_instructions: instructions.trim() || null,
+      });
+      await mutate(SWR_KEYS.settings);
+      setDraft(null);
+      toast.success("Craft instructions saved");
+      return true;
+    } catch (err) {
```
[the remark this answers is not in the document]
rohoswagger: Fixed in 56beb2b — the caught error is now logged via console.error before the toast, so save failures are traceable.

backend/onyx/server/features/build/api.py
```diff
@@ -42,6 +49,29 @@ def require_onyx_craft_enabled(
...
+def get_base_instructions(
+    _: User = Depends(require_permission(Permission.FULL_ADMIN_PANEL_ACCESS)),
+) -> BaseInstructionsResponse:
+    """The base AGENTS.md template, so admins can see what their workspace
+    instructions are appended to. Dynamic sections appear as placeholders."""
+    if not AGENT_INSTRUCTIONS_TEMPLATE_PATH.exists():
+        raise OnyxError(OnyxErrorCode.NOT_FOUND, "Base instructions not found")
+    return BaseInstructionsResponse(
+        content=AGENT_INSTRUCTIONS_TEMPLATE_PATH.read_text()
+    )
```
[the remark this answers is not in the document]
rohoswagger: Fixed in 56beb2b — admin_router now carries a router-level require_permission(FULL_ADMIN_PANEL_ACCESS) dependency, so future routes are guarded by default.

backend/onyx/server/features/build/api.py
```diff
@@ -42,6 +49,33 @@ def require_onyx_craft_enabled(
...
+# gate: an admin configuring Craft may not have Craft enabled for themselves.
+# The router-level permission guard backstops every route added here.
+admin_router = APIRouter(
+    prefix="/build/admin",
+    dependencies=[Depends(require_permission(Permission.FULL_ADMIN_PANEL_ACCESS))],
+)
+admin_router.include_router(external_apps_admin_router, tags=["build"])
+
+
+class BaseInstructionsResponse(BaseModel):
```
wenxi-onyx: pref moving this to models

backend/onyx/server/features/build/external_apps/api.py
```diff
@@ -65,6 +65,11 @@
 
 router = APIRouter()
 
+# Admin-only app management. Mounted under the ungated /build/admin router so
```
wenxi-onyx: nit: no need to re-iterate this since it's in the docstring already

backend/onyx/server/settings/api.py
```diff
@@ -147,6 +150,11 @@ def fetch_settings(
     )
     general_settings = apply_fn(general_settings)
 
+    # Workspace Craft instructions are admin-configured internal guidance;
+    # non-admins (and anonymous users) don't need them in the settings payload.
+    if not user or not is_user_admin(user):
```
wenxi-onyx: Any reason why normal user API shouldn't see this? I think it's fine to allow this field through/no need to filter it out. They can see it anyways in the AGENTS.md as well

wenxi-onyx (approved):

backend/onyx/server/features/build/api.py
rohoswagger: Moved to backend/onyx/server/features/build/models.py in a5ad322.

backend/onyx/server/features/build/external_apps/api.py
rohoswagger: Removed in a5ad322 (and the remaining router comment was trimmed in 4eafc97).

backend/onyx/server/settings/api.py
rohoswagger: Fair — the one reason for filtering was anonymous access: /settings serves unauthenticated visitors when anonymous mode is on, and AGENTS.md only reaches craft-enabled users, so the field would otherwise be world-readable. Split the difference in 92b0620: authenticated users (any role) now get the field; only anonymous visitors have it stripped. Happy to drop the filter entirely if you'd rather.

PR #12961, opened 2026-07-13
build(web): swap ts-jest for @swc/jest and adopt stable TypeScript 7 checker

jmelahman:
## Description

Tooling changes that move `web/` onto TypeScript 7-era infrastructure:

**1. Jest transform: `ts-jest` → `@swc/jest`**

`ts-jest` blocks TypeScript 7 adoption: its peer range caps at `<7` and it drives the classic TS compiler API, which the `typescript@7` package no longer ships. Jest was already running transpile-only (`isolatedModules: true`, types checked separately by `types:check`), so swapping to `@swc/jest` loses nothing and decouples Jest from the `typescript` package entirely.

One wrinkle: SWC's CommonJS codegen hoists all barrel reexport getters above their backing `require()` calls. Opal components that import sibling primitives from their own public barrel (e.g. `tooltip/components.tsx` importing `Text` from `@opal/components`) form require cycles, and reading a getter mid-cycle throws `ReferenceError: Cannot access '...' before initialization` — triggered in practice by `jest.requireActual(...) + spread` partial mocks, which eagerly invoke every getter. Fixed by scoping `module.lazy` to exactly the three self-referential Opal barrels (`@opal/components`, `@opal/layouts`, `@opal/core`); lazy matching is exact on the import specifier, so subpath imports and everything else keep eager semantics. A global `lazy: true` was tried and rejected — it broke 33 unrelated suites by deferring third-party import side effects.

**2. `types:check`: `@typescript/native-preview` dev builds → stable `typescript@7.0.2`**

The stable TS7 release now ships the native compiler under the regular `typescript` npm name. It's added as the `typescript-7` alias so it coexists with `typescript@5.9`, which stays because Next.js's build-time type integration calls classic compiler APIs (`createProgram` etc.) that `typescript@7` doesn't export (tsup's dts generation and Storybook's react-docgen-typescript also need the classic API). The web pre-commit hook now delegates to `bun run types:check`.

**3. `lib/shared`: `typescript@5.9` → `typescript@7`**

`lib/shared` only uses `tsc` as a CLI (emit via `build:ts`, checking via `typecheck`) with no classic-API consumers, so it takes `typescript@^7.0.2` directly under the real package name — bun nests it per-workspace since web keeps 5.9 hoisted, and `lib/shared/node_modules/.bin/tsc` resolves to 7.0.2 while web's stays 5.9. The shared pre-commit hook now delegates to `bun run typecheck`. The widget hooks are unchanged (they resolve `tsgo` from the root workspace, which keeps `@typescript/native-preview`).

## How Has This Been Tested?

- `bun run test`: 64/64 suites, 800/800 tests, verified stable across many runs including two cold-cache runs.
- Compared coverage mode (`--coverage`, what CI runs) against a ts-jest baseline by stashing the change: the occasional slow-machine flake (`CustomModal`) reproduces identically on both transforms — pre-existing, not a regression. Baseline actually collected fewer tests (775 vs 800).
- `bun run types:check` with `typescript@7.0.2`: passes clean in ~5.1s (tsgo baseline: ~5.8s, same result).
- `lib/shared`: full `bun run build` output under TS7 is **byte-identical** to the 5.9 baseline (JS, `.d.ts`, declaration maps, source maps); `typecheck` passes; `tsc --watch --preserveWatchOutput` (used by `scripts/watch.mjs`) starts and watches correctly.
- Confirmed via compiled output that `module.lazy` matches exact specifiers only, and grepped Opal for other self-referential barrels (none outside `.stories.tsx`, which Jest never loads).
- `bun run lint` passes; `web/node_modules/.bin/tsc` still resolves to `typescript@5.9` (no bin collision with the alias or the nested workspace install).

## Additional Options


raunakab (approved):

web/lib/shared/package.json
```diff
@@ -38,17 +38,17 @@
   },
   "scripts": {
     "build:tokens": "node style-dictionary.config.mjs",
-    "build:ts": "tsc -p tsconfig.build.json",
+    "build:ts": "node ./node_modules/typescript/bin/tsc -p tsconfig.build.json",
```
raunakab, line 41: Btw, is there a reason for the `bun` vs `node` discrepancy here?

raunakab (approved):

PR #7555, opened 2026-01-20
feat(filesys): creation of hierarchyfetching job

evan-onyx:
## Description

see title. ONLY the boilerplate

## How Has This Been Tested?

not yet

## Additional Options

- [x] [Optional] Override Linear Check

---
## Summary by cubic
Adds a scheduled hierarchy fetching job for connectors, including Celery beat, task discovery, queue/locks, and a DB timestamp to track last fetch. The beat checks hourly; fetching runs at most once per day.

- **New Features**
  - Added Celery beat entry: check-for-hierarchy-fetching (hourly, low priority).
  - Enabled task autodiscovery for onyx.background.celery.tasks.hierarchyfetching in primary and heavy apps.
  - Introduced CONNECTOR_HIERARCHY_FETCHING queue and CHECK_HIERARCHY_FETCHING_BEAT_LOCK.
  - Added task constants: CHECK_FOR_HIERARCHY_FETCHING and CONNECTOR_HIERARCHY_FETCHING_TASK.
  - Implemented tasks: check_for_hierarchy_fetching and connector_hierarchy_fetching_task (skeleton) with locking and per-connector scheduling.
  - Added DB helper mark_cc_pair_as_hierarchy_fetched to set last_time_hierarchy_fetch.

<sup>Written for commit 24de804e0ebe795aa66dabb543453228e07c4729. Summary will update on new commits.</sup>

PR #12964, opened 2026-07-13
fix(celery): model-server request timeouts + wire the orphaned hierarchy queue

justin-tahara:
## Description

Three fixes from a production incident investigation (customer indexing stalled for hours after a routine upgrade):

**1. Timeouts on the model-server `requests.post` calls** (embed / rerank / intent — `search_nlp_models.py`). These had none, so a model-server pod restarting mid-request left the calling worker thread blocked **forever** inside `requests.post` — silently: no exception, no log line. During an upgrade this wedged every docprocessing thread for 2.5h+ until the workers were manually restarted (same no-timeout-blocking-call class as the `document_filter` hang fixed in #12333). New `MODEL_SERVER_CONNECT_TIMEOUT` (30s) / `MODEL_SERVER_READ_TIMEOUT` (600s — generous because CPU embedding of large batches legitimately takes minutes). The existing passage-retry wrapper already retries `RequestException`, so a timeout becomes 3 bounded retries then a visible, redeliverable task failure instead of an invisible wedge.

**2. `connector_hierarchy_fetching` finally gets a consumer.** The hierarchyfetching job (#7555) shipped a producer + task implementation, but **no deployment mode ever subscribed a worker to the queue** — not the helm chart, not supervisord. #9381 later limited the producer (currently to zero, since no connector implements `HierarchyConnector` yet), but long-lived brokers still hold fossilized backlogs: one production deployment had **67,940 unconsumed tasks (71MB)** accumulated over months. The heavy worker already registers these tasks (`apps/heavy.py` imports the module) — this adds the queue to its `-Q` in both the chart and supervisord, so legacy backlogs drain safely (the task no-ops on non-hierarchy connectors) and the wiring exists when implementers land.

**3. `expires=BEAT_EXPIRES_DEFAULT` on the hierarchy fan-out `send_task`** (was unset, violating the repo's always-expire rule — this is exactly what let the backlog grow unbounded; the hourly check re-enqueues anything still due).

## How Has This Been Tested?

- Root cause verified live on the affected deployment: threads pinned in `requests.post` across a model-server pod roll; worker restart cleared it; queue contents inspected to confirm the 67.9k orphaned hierarchy tasks (headers show `expires: None`, producer `check_for_hierarchy_fetching`).
- `pre-commit` clean on all touched files (ruff/ty/lazy-imports/env-drift-baseline).
- Grepped all three deployment modes (helm templates, supervisord, compose) to confirm no other worker consumes the queue and heavy is the registering app.

## Additional Options

---
## Summary by cubic
Prevent worker wedges during indexing by adding timeouts to model-server calls, bounding the cross-batch DB lock, and wiring the orphaned hierarchy queue so backlogs drain. Ships via Helm chart `0.6.12`.

- **Bug Fixes**
  - Add connect/read timeouts to model-server `requests.post` (embed/rerank/intent): `MODEL_SERVER_CONNECT_TIMEOUT`=30s, `MODEL_SERVER_READ_TIMEOUT`=600s; timeouts become retriable failures.
  - Bound cross-batch DB lock acquire to 300s; on timeout, fail fast so the task retries against a fresh lock (prevents fleet-wide stalls on fossil locks).
  - Subscribe the heavy worker to `connector_hierarchy_fetching` in Helm and `supervisord` so legacy backlogs drain.
  - Set `expires=BEAT_EXPIRES_DEFAULT` on hierarchy fan-out tasks to enforce expiry and prevent buildup.

- **Dependencies**
  - Bump Helm chart to `0.6.12` to include the heavy worker `-Q` update.

<sup>Written for commit dfbda8219950d52372e839c3e0a6c4c74e793bce. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12964?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


justin-tahara:
Chart version bumped to 0.6.12 per the bump-per-chart-change convention.

justin-tahara:
Added a 4th fix surfaced by the same incident's second act: restarting the wedged workers killed a `cross_batch_db_lock` holder mid-critical-section, and the **unbounded `acquire()`** then wedged every replacement thread for the fossil lock's full 3h15m TTL (verified via celery SIGUSR1 thread dumps — all 6 threads in `redis/lock.py acquire` at `tasks.py:1863`). Now a bounded 300s acquire that fails the task loudly (redelivers) instead of convoying silently.

evan-onyx (approved):

PR #12966, opened 2026-07-13
fix(scim): return 409 not 500 on concurrent duplicate user create

justin-tahara:
## Description

Concurrent SCIM `POST /Users` requests for the **same email** (e.g. an IdP batch/retry sending the create twice near-simultaneously) can race on the case-sensitive `email` primary key. The create path already guards `add_user` with `except IntegrityError -> 409`, but the duplicate INSERT can be **deferred past `add_user` and surface via autoflush** inside the default-group query in `assign_user_to_default_groups__no_commit`.

At that point the broad `except Exception` in `_assign_default_groups_or_error` catches the `IntegrityError`, rolls back, **logs a full error-level traceback**, and returns a misleading **500 "Failed to assign user to default group"** — instead of the clean **409 "already exists"** the create path returns for the same race elsewhere.

This was observed in production: an IdP re-provisioning the same user produced repeated 500s + noisy `IntegrityError` tracebacks (`duplicate key value violates unique constraint "ix_user_email"`) originating from the group-assignment autoflush.

**Fix:** catch `IntegrityError` distinctly in `_assign_default_groups_or_error` and return a quiet 409. Genuine assignment failures (e.g. missing seeded default group -> `RuntimeError`) still return 500 with the traceback.

## How Has This Been Tested?

- New unit test `test_assign_default_groups_integrity_error_returns_409`: when `assign_user_to_default_groups__no_commit` raises `IntegrityError`, `create_user` returns 409, rolls back, and does not commit.
- Existing `test_promotion_default_group_failure_returns_500` (missing default group -> 500) still passes, confirming non-race errors are unaffected.
- Full SCIM user-endpoint unit suite: 73 passed.

## Additional Options

---
## Summary by cubic
Return 409 Conflict instead of 500 when concurrent SCIM `POST /Users` requests create a duplicate user during default-group assignment, matching the create path and avoiding noisy tracebacks.

- **Bug Fixes**
  - Catch `IntegrityError` in `_assign_default_groups_or_error` and return 409 only for `ix_user_email` unique violations using `is_unique_violation`.
  - Keep 500 for other failures (e.g., missing default group or non-email integrity errors) with full traceback.
  - Add `test_assign_default_groups_email_integrity_error_returns_409` and `test_assign_default_groups_other_integrity_error_returns_500`; existing 500 path test still passes.

<sup>Written for commit 3de1b26bd1f8366cadd49ccc4fb44a21e04e30bd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12966?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/ee/onyx/server/scim/api.py
+++ b/backend/ee/onyx/server/scim/api.py
@@ -252,6 +252,17 @@ def _assign_default_groups_or_error(
     """
     try:
         assign_user_to_default_groups__no_commit(db_session, user, is_admin=is_admin)
+    except IntegrityError:
+        # A concurrent SCIM provisioning request (e.g. an IdP batch retry) already
+        # created this user; the duplicate INSERT is deferred and surfaces here via
+        # autoflush rather than at ``add_user``. This is an expected provisioning
+        # race, not a server error — surface a clean 409 like the create path does.
+        dal.rollback()
+        logger.info(
+            "SCIM user %s already exists (concurrent provisioning); returning 409",
+            email,
+        )
+        return _scim_error_response(409, f"User with email {email} already exists")
     except Exception:
         dal.rollback()
         logger.exception("Failed to assign SCIM user %s to default groups", email)
--- a/backend/ee/onyx/server/scim/api.py
+++ b/backend/ee/onyx/server/scim/api.py
@@ -65,6 +65,7 @@ from onyx.db.models import UserRole
 from onyx.db.permissions import recompute_permissions_for_group__no_commit
 from onyx.db.permissions import recompute_user_permissions__no_commit
 from onyx.db.users import assign_user_to_default_groups__no_commit
+from onyx.db.utils import is_unique_violation
 from onyx.utils.logger import setup_logger
 from shared_configs.contextvars import get_current_tenant_id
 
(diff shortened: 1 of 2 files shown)


nmgarza5 (approved):

justin-tahara:
Thanks — addressed in 3de1b26:

- **Narrowed the 409 branch** (cubic P2): now catch `Exception`, and only return 409 when `is_unique_violation(e, "ix_user_email")` is true (reusing the existing `onyx.db.utils` helper). Every other failure — `RuntimeError` (missing default group) *and* any non-email integrity error (FK/other constraint) — stays on the structured 500 path, so real backend faults aren't masked as "already exists".
- **Docstring** (greptile P2): rewritten to describe both failure modes (409 email-race vs 500 everything-else).
- Added regression test `test_assign_default_groups_other_integrity_error_returns_500` for the non-email case; renamed the 409 test to `...email_integrity_error...`. Full SCIM suite: 75 passed.

evan-onyx (approved):

PR #12971, opened 2026-07-13
fix(billing): invalidate billing cache on control-plane tier push

nmgarza5:
## Description

**Bug:** on Onyx Cloud a tenant that converts trial→paid keeps trial usage caps for up to 24h. API-key calls get `RATE_LIMITED: API access is not available on trial accounts` and indexing keeps the trial chunk cap, even though billing already shows paid. A cloud tenant hit the API-access case live on 08-18.

**Cause:** the usage-limit gate reads a Redis billing cache (`cached_is_tenant_on_trial`) with a 24h TTL. The control-plane `/tenants/tier-update` push only refreshed the tier cache. It never invalidated the billing cache, so the stale `trialing` status kept enforcing trial caps until the TTL expired.

**Fix:**
- The `/tenants/tier-update` receiver drops the billing cache after a successful tier write. It reports `updated=False` when the drop fails, so the control plane does not treat a still-stale cache as refreshed. A retry is safe because the tier write is an idempotent Redis SET.
- `invalidate_billing_cache` returns `bool` to support that reporting.
- `BILLING_CACHE_TTL_SECONDS` default 24h → 1h as the backstop for a missed push.

The control plane already pushes on every tier, trial_end, or status change, including trial→active (onyx-dot-app/danswer-control-plane#185). This PR makes the data plane honor that push. Invalidation lives in the endpoint, not `update_tenant_tier`, so the data-plane lazy tier-cache refresh does not over-fire it.

An earlier version of this branch also plumbed `subscribed` through tier resolution. That part is dropped. #14046 removes the trial tier promotion entirely and owns that surface.

## How Has This Been Tested?

Unit tests. `TestUpdateTier` covers the endpoint wiring (proven failing before the fix). The billing-cache tests now assert the `bool` return contract.

## Additional Options

### DIFF
--- a/backend/ee/onyx/server/billing/api.py
+++ b/backend/ee/onyx/server/billing/api.py
@@ -264,8 +264,8 @@ def invalidate_billing_info_cache() -> None:
     """Drop the cached billing entries once the subscription is known to have moved.
 
     Best-effort. Busts the 5-min admin /billing-information entry, plus (cloud
-    only) the 24h per-tenant trial-status entry the indexing path reads via
-    ``cached_is_tenant_on_trial`` — without this a trial→paid conversion keeps
+    only) the per-tenant trial-status entry that usage limits read via
+    ``cached_is_tenant_on_trial``. Without this a trial→paid conversion keeps
     usage limits on stale trial status for up to a full TTL.
     """
     redis_client = _billing_cache_client()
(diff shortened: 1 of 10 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

justin-tahara (approved):

PR #12685, opened 2026-07-05
fix(chat): stop citation processor from corrupting code fences

markk-ns:
### Problem
`DynamicCitationProcessor` labels unlabeled code fences with `plaintext` while
streaming, but does it with a replace-all:

```python
self.curr_segment = self.curr_segment.replace("```", "```plaintext")
```

`str.replace` rewrites **every** fence in `curr_segment`, so other fences in the
same buffer get corrupted:

- ` ```bash ` → ` ```plaintextbash `
- closing ` ``` ` → ` ```plaintext ` (starts a spurious code block)

This is streamed to the client **and** persisted to `chat_message.message`, so it
shows up in the UI and survives reloads. Common for command/how-to answers that
contain more than one fence.

### Fix
Only rewrite the specific bare fence that was detected — the first ``` ``` ``` in the
segment, whose following character the code already inspected via `parts[1][0]`:

```python
self.curr_segment = parts[0] + "```plaintext" + "```".join(parts[1:])
```

Intent is preserved (a bare opening fence still gets a `plaintext` language), but
labeled fences and closing fences are left intact. The change is idempotent for
the detected fence and touches nothing else.

### Testing
- Verified on a live v4.0.9 deployment: answers containing a ` ```bash ` block plus
  a bare output block now render cleanly; no ` ```plaintextbash ` / stray
  ` ```plaintext ` in the stored message.
- Happy to add a unit test for `DynamicCitationProcessor` asserting that a
  `bash`-labeled fence adjacent to a bare fence is left unchanged, if maintainers
  want it in this PR.

Fixes #12684

---
## Summary by cubic
Fixed a bug where the streaming citation processor relabeled every three‑backtick code fence, corrupting code blocks in the UI and stored messages. It now labels only the detected bare opening fence as plaintext and leaves labeled (e.g., bash) and closing fences unchanged.

<sup>Written for commit e9763688a936cd08b62b5dcaf48a9d2d0b47bc49. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12685?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #12976, opened 2026-07-13
fix: prevent citation processor from corrupting adjacent code fences

### LINKED ISSUE #12684: Streaming citation processor corrupts code fences (injects plaintext into ```bash and closing fences)
### Summary
`DynamicCitationProcessor` in `backend/onyx/chat/citation_processor.py` tries to
add a `plaintext` language to *unlabeled* code fences while streaming the answer.
The implementation uses a **replace-all** on the streaming buffer:

```python
if piece_that_comes_after == "\n" and in_code_block(self.llm_out):
    self.curr_segment = self.curr_segment.replace("```", "```plaintext")
```

Because `str.replace` rewrites **every** ``` ``` ``` in `curr_segment`, any *other*
fence that happens to be in the same buffer is corrupted:

- a following labeled fence ` ```bash ` becomes ` ```plaintextbash `
- a closing fence ` ``` ` becomes ` ```plaintext ` (which starts a *new* spurious
  code block, so the rest of the message renders as code)

The corruption is written into the streamed answer **and persisted** to
`chat_message.message`, so it is visible in the UI and survives reloads. It is not
a frontend rendering issue.

### Reproduction
Ask any question whose answer contains a shell command block followed by a plain
command-output block, e.g. a "how do I set quotas" doc question. Typical output:

~~~
```plaintext
lfs setquota -u galb -b 500G -B 800G /space3
```plaintextbash
lfs setquota -g ns -b 78t -B 80t /space3
```
~~~

The stored message (Postgres `chat_message.message`) contains the literal
` ```plaintext ` / ` ```plaintextbash ` markers.

### Impact
Any answer that contains more than one code fence in a single streamed segment can
be mangled. Very common for how-to / command-reference answers.

### Fix
Only modify the specific bare fence that was detected (the first ``` ``` ``` in the
segment, whose following char was already inspected), instead of replace-all:

```python
self.curr_segment = parts[0] + "```plaintext" + "```".join(parts[1:])
```

This preserves the original intent (label a bare opening fence) while leaving all
other fences — labeled or closing — untouched. PR attached.

### Environment
- Onyx CE v4.0.9 (bug identical on `main` as of this writing)

shivanij1203:
## Summary

`DynamicCitationProcessor` adds a `plaintext` language tag to unlabeled code fences while streaming. It did this with a buffer-wide `curr_segment.replace("```", "```plaintext")`, but the detection logic only ever inspects the **first** fence in the segment. When a second fence lands in the same buffered segment, it gets rewritten too:

- a following labeled fence ` ```bash ` becomes ` ```plaintextbash `
- a closing fence ` ``` ` becomes ` ```plaintext `, which opens a spurious new code block so the rest of the message renders as code

Because the corruption is written into the streamed answer and persisted to `chat_message.message`, it is visible in the UI and survives reloads.

Fixes #12684
/claim #12684

## Changes

- `backend/onyx/chat/citation_processor.py`: label only the first (unlabeled) opening fence by reconstructing the segment from the already-computed `parts`, instead of a blanket `replace` that rewrites every fence in the buffer.
- `backend/tests/unit/onyx/chat/test_citation_processor.py`: add two regression tests — one for a following labeled fence, one for a closing fence sharing the same buffered segment.

## Testing

- `pytest backend/tests/unit/onyx/chat/test_citation_processor.py` — **131 passed** (129 existing + 2 new).
- Both new tests **fail on the pre-fix code** and pass after it, so they guard the actual regression rather than passing vacuously.
- No existing citation/code-fence test changed behavior — single-fence handling is byte-for-byte identical.

Before/after on the labeled-fence case (input tokens `["```python\n", "```\nplain\n```bash\n"]`):

```
before: '```python\n```plaintext\nplain\n```plaintextbash\n'   # ```bash corrupted
after:  '```python\n```plaintext\nplain\n```bash\n'            # ```bash preserved
```

## Notes

The existing tests never caught this because they feed each fence in a separate token, so two fences never shared one `curr_segment`. The fix keeps the original detection logic untouched and only narrows the mutation to the single fence that logic already selected.

---
## Summary by cubic
Fixes a streaming bug that corrupted adjacent code fences; now only the intended unlabeled opening fence is tagged. Fixes Linear #12684.

- **Bug Fixes**
  - Rebuilt labeling to reconstruct the segment and tag `plaintext` only when a fence opens a new block (decided by code‑block parity) and is followed by a newline; ` ```lang ` and closing fences are left untouched.
  - Added regression tests covering a following labeled fence, a closing fence in the same segment, and a close‑then‑open segment.

<sup>Written for commit 489f2d4a75a2e2c7b6d953226992547a417562ec. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/12976?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/chat/citation_processor.py
+++ b/backend/onyx/chat/citation_processor.py
@@ -307,23 +307,29 @@ class DynamicCitationProcessor:
         self.curr_segment += token
         self.llm_out += token
 
-        # Handle code blocks without language tags
-        # If we see ``` followed by \n, add "plaintext" language specifier
+        # Add a `plaintext` language tag to unlabeled opening code fences.
+        # Each fence's role (opening vs closing) is derived from the code-block
+        # parity established *before* this segment, so closing fences and fences
+        # that already carry a language are left untouched. A naive first-fence
+        # or replace-all approach would corrupt a closing ``` or a following
+        # ```lang that shares the same segment.
         if "`" in self.curr_segment:
             if self.curr_segment.endswith("`"):
+                # Trailing fence may be incomplete; wait for the next token.
                 pass
             elif "```" in self.curr_segment:
+                prefix = self.llm_out[: len(self.llm_out) - len(self.curr_segment)]
+                inside_code_block = prefix.count("```") % 2 != 0
                 parts = self.curr_segment.split("```")
-                if len(parts) > 1 and len(parts[1]) > 0:
-                    piece_that_comes_after = parts[1][0]
-                    if piece_that_comes_after == "\n" and in_code_block(self.llm_out):
-                        # Only label the first (unlabeled) opening fence. A
-                        # blanket replace would also rewrite any other fence in
-                        # the same segment, corrupting a following ```lang fence
-                        # or turning a closing ``` into a spurious new block.
-                        self.curr_segment = (
-                            parts[0] + "```plaintext" + "```".join(parts[1:])
-                        )
+                rebuilt = parts[0]
+                for following in parts[1:]:
+                    opens_block = not inside_code_block
+                    if opens_block and following.startswith("\n"):
+                        rebuilt += "```plaintext" + following
+                    else:
+                        rebuilt += "```" + following
+                    inside_code_block = not inside_code_block
+                self.curr_segment = rebuilt
 
         # Look for citations in current segment
         citation_matches = list(self.citation_pattern.finditer(self.curr_segment))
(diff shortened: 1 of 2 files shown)


shivanij1203:
Good catch. The parity check was being evaluated over the full output, so a segment that closed one block and then opened an unlabeled one ended up labeling the closing fence rather than the opening one.

Fixed in 489f2d4: each fence's role is now determined from the code-block parity established before the segment, so only genuine opening fences receive the `plaintext` tag. Closing fences and fences that already carry a language are left untouched. Added a regression test (`test_fence_labeled_by_parity_when_segment_closes_then_opens`) covering the close-then-open case, and the full citation processor unit suite passes (132 tests).

justin-tahara:
Closing as a duplicate — the same fix landed via #13538 (originally #12685). Thanks!

PR #12992, opened 2026-07-14
refactor(web): migrate PageLoader consumers to Opal

nmgarza5:
## Description

Migrates all 22 consumers of `PageLoader` from `@/refresh-components/PageLoader` to the Opal `PageLoader` (`@opal/layouts`), and deletes the old app `PageLoader`, `OnyxLoader`, and `onyx-loader.css`.

The Opal component is API-identical (`text` prop, same "Loading …" default) and every call site renders `<PageLoader />` bare, so each consumer is a one-line import repoint. `OnyxLoader`'s only consumer was the deleted `PageLoader` itself.

Stacked on #12917.

## Additional Options


nmgarza5:
@greptile

nmgarza5:
@greptile

PR #12999, opened 2026-07-14
feat(connectors): add LumApps connector (modified mirror of #12899)

Bo-Onyx:
Modified mirror of #12899 (LumApps connector by @siriusfreak), owned by us for review + CI.

Adds a **LumApps** connector that ingests published intranet content (pages, news, custom content types) into Onyx: checkpointed incremental indexing off `content/list` sorted by `-updatedAt` with an early-break, LIVE-only filtering enforced per-item on both the index and prune paths, generic metadata-family ingestion, and a self-contained `requests`-based client with token minting + retry.

## Changes on top of the original #12899

- **Harden `updatedAt` parsing**: `_parse_dt` now tolerates numeric epochs (seconds or milliseconds) in addition to ISO-8601, so an unexpected timestamp format can't silently disable the incremental early-break and force a full re-scan every run.
- **Preserve upstream status on exhausted retries**: the client now surfaces the real `429`/`5xx` status instead of a generic `503`, so validation/callers can translate it correctly.
- **Document the permissions model**: added a docstring note that this is a public-only connector — all LIVE content the service user can read is indexed with no per-document access controls (no `CheckpointedConnectorWithPermSync`, no EE permission-sync entry). Scope the service user / `instance_ids` accordingly.
- **Tests**: added the missing `tests/unit/onyx/connectors/lumapps/__init__.py` and coverage for the new epoch parsing and status-preservation behavior.

## Notes / open decision

- **Permission sync is intentionally not implemented** (public-only v1). This is the one item worth a conscious sign-off before merge, since LumApps is a permissioned intranet — everything indexed becomes visible to all Onyx users. Documented in the connector docstring.

## Verification

- `pytest backend/tests/unit/onyx/connectors/lumapps/` — 17 passed.
- Completeness safety-nets pass: `test_connector_factory.py`, `test_slack_icons.py`.
- `ruff` / `ruff format` / `ty` clean on the connector + tests.

Original author verified indexing, incremental re-sync, and pruning against a real LumApps tenant on their production deployment.
Screenshot 2026-07-14 at 4 31 43 PM
Screenshot 2026-07-14 at 4 31 37 PM
Screenshot 2026-07-14 at 4 31 32 PM

- [x] Override Linear Check

### DIFF
--- a/backend/onyx/configs/constants.py
+++ b/backend/onyx/configs/constants.py
@@ -263,6 +263,7 @@ class DocumentSource(str, Enum):
     BITBUCKET = "bitbucket"
     TESTRAIL = "testrail"
     BRAINTRUST = "braintrust"
+    LUMAPPS = "lumapps"
 
     # Special case just for integration tests
     MOCK_CONNECTOR = "mock_connector"
@@ -749,4 +750,5 @@ DocumentSourceDescription: dict[DocumentSource, str] = {
     DocumentSource.IMAP: "Email messages and threads",
     DocumentSource.TESTRAIL: "Test cases and QA management",
     DocumentSource.BRAINTRUST: "LLM eval experiments, datasets, and prompts",
+    DocumentSource.LUMAPPS: "Intranet pages, news, and content",
 }
--- a/backend/onyx/connectors/lumapps/client.py
+++ b/backend/onyx/connectors/lumapps/client.py
@@ -112,6 +112,8 @@ class OnyxLumApps:
     ) -> dict[str, Any]:
         url = f"{self.base_url}{path}"
         last_network_error: requests.RequestException | None = None
+        last_http_status: int | None = None
+        last_http_text: str = ""
         for attempt in range(_MAX_RETRIES):
             try:
                 response = self._session.request(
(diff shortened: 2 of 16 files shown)


Bo-Onyx:
Addressed all outstanding review comments in `d91bbe7` (client hardening) — plus earlier `8b6865d`. Mapping each thread to its fix:

**greptile P1 — malformed token response (`client.py:92`)** & **cubic P2 (`:91`)**: `_mint_token` now wraps `response.json()` / `body["access_token"]` / `int(expires_in)` in a `try/except` that raises `LumAppsClientError`, so a `200` with empty/HTML body or missing `access_token` is translatable by `validate_connector_settings()` instead of a raw `JSONDecodeError`/`KeyError`.

**greptile P1 — malformed data response (`client.py:163`)** & **cubic P2 (`:163`)**: `_request` now wraps `return response.json()` and raises `LumAppsClientError` on a non-JSON `200`.

**cubic P2 — 401 refresh gated to first attempt (`client.py:145`)**: replaced the `attempt == 0` gate with a `token_refreshed` flag, so a `401` landing after an earlier transient retry still triggers the one-time re-mint.

**cubic P2 — Retry-After HTTP-date/NaN (`client.py:26`)**: `_backoff_seconds` now uses the shared `parse_retry_after_seconds`, honoring both delay-seconds and HTTP-date forms and rejecting `nan`/`inf`.

**cubic P3 — extra backoff on final attempt (`client.py:117`)**: the network-error and 429/5xx branches skip the sleep on the last attempt (no wasted 16–60s delay before raising).

All covered by new unit tests (21 LumApps unit tests pass; `ty`/`ruff` clean). The earlier `8b6865d` also handled the epoch `updatedAt` parsing, exhausted-retry status preservation, the public-only permission-model docstring, and the missing test `__init__.py`.

evan-onyx (approved):

PR #13004, opened 2026-07-14
fix(indexing): don't invalidate saturated-but-alive index attempts

justin-tahara:
## Description

The docprocessing stall watchdog (`validate_active_indexing_attempts`) treated any **stale heartbeat with `in_flight > 0`** as *"workers crashed holding batches"* and invalidated the index attempt. But a saturated dependency — e.g. a CPU embedder pegged at 100% with thousands of batches queued — blocks every worker thread the same way while the workers themselves are perfectly alive. In that state the watchdog would discard **multi-day indexing progress** on a connector that was merely waiting on embeddings, then restart it from scratch.

This makes invalidation require corroboration, not just a stale counter:

- **Completed-batch progress is now a crash-proof liveness signal** alongside the heartbeat counter. Forward progress on completions since the last watchdog pass resets the stall window even if the heartbeat thread itself is starved. Tracked in the previously-unused `last_batches_completed_count` column — **no migration needed**.
- **Worker-liveness cross-check before the crash verdict.** When the heartbeat is stale with `in_flight > 0` and no completion progress, we ping the docprocessing workers via Celery inspect (answered by the worker's control thread, so a CPU-pegged worker still replies). Read **asymmetrically**: a positive reply → *"saturated/wedged, workers alive"*, logged loudly at ERROR and **spared**; an empty/error reply → fall through to the existing crash verdict. We only ever *spare* on a positive read, never *destroy* on a flaky empty one. The `in_flight == 0, pending == 0` (no work anywhere) case still invalidates unconditionally.
- **Hardened the heartbeat thread**: it now escalates WARNING → ERROR after sustained consecutive failures (leading hypothesis for the original stall is the heartbeat write contending for the shared worker DB pool, `pool_size = concurrency + 8`), so a starved counter surfaces in dashboards instead of being silently swallowed.

**Real crashes (no live workers) still invalidate exactly as before** — this does not soften genuine crash detection.

Related but distinct from #12964 (bounded the unbounded model-server waits); that reduced how long a single batch can block but didn't teach the watchdog to tell "all workers blocked on a saturated dependency" from "workers crashed".

## How Has This Been Tested?

Added external-dependency-unit tests (`backend/tests/external_dependency_unit/celery/test_indexing_stall_watchdog.py`):

- Stale heartbeat + `in_flight > 0` + `completed_batches` advancing → **not** invalidated.
- Stale heartbeat + no progress + workers alive → spared (grace path, window reset).
- Stale heartbeat + no progress + no live workers → invalidated exactly as today (regression guard for real crashes).
- Heartbeat resilience: killing its DB session factory mid-run escalates to ERROR rather than being swallowed forever.

All four pass locally against the branch checkout; `ty`, `ruff`, and full `pre-commit` pass on the `main` base.

## Additional Options

---
## Summary by cubic
Stop invalidating long-running indexing attempts when workers are alive but blocked on a saturated dependency. The stall watchdog now corroborates stale heartbeats, preserves progress, and caches worker-liveness checks to avoid N× timeouts; real crashes still invalidate as before.

- **Bug Fixes**
  - Treat completed-batch progress as a liveness signal alongside the heartbeat; store in `last_batches_completed_count` (no migration).
  - For stale heartbeat with `in_flight > 0`, ping docprocessing workers via Celery inspect: positive reply → spare and reset window; empty/error → fall back to existing crash verdict; `in_flight == 0 && pending == 0` still invalidates.
  - Heartbeat thread now escalates from WARNING to ERROR after sustained failures so counter stalls are surfaced; added a TODO with candidate causes to guide root-cause investigation.

- **Performance**
  - Compute the worker‑liveness ping once per watchdog pass (cached lazily) so multiple stale attempts don’t add N× `DOCPROCESSING_LIVENESS_PING_TIMEOUT` while holding the DB session.

<sup>Written for commit 3f68d0adc42b93e3c4f40793a32b59cc1c3f4332. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13004?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/background/celery/tasks/docprocessing/heartbeat.py
+++ b/backend/onyx/background/celery/tasks/docprocessing/heartbeat.py
@@ -10,12 +10,20 @@ from onyx.utils.logger import setup_logger
 
 logger = setup_logger()
 
+# After this many consecutive failed beats we escalate WARNING -> ERROR. A
+# stalled heartbeat is dangerous: the stall watchdog reads the counter as a
+# liveness signal, so a silently-wedged heartbeat (e.g. the write starving for a
+# DB connection against the shared worker pool) can get a live attempt
+# invalidated. Surfacing sustained failures loudly lets operators catch it.
+_HEARTBEAT_FAILURE_ESCALATION_THRESHOLD = 3
+
 
 def start_heartbeat(index_attempt_id: int) -> tuple[threading.Thread, threading.Event]:
     """Start a heartbeat thread for the given index attempt"""
     stop_event = threading.Event()
 
     def heartbeat_loop() -> None:
+        consecutive_failures = 0
         while not stop_event.wait(INDEXING_WORKER_HEARTBEAT_INTERVAL):
             try:
                 with get_session_with_current_tenant() as db_session:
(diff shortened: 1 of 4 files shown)


justin-tahara:
Good catch from both bots — `_docprocessing_workers_alive()` was called per stale in-flight attempt, so with workers down and N stale attempts the pass could block N × 4s while holding the DB session. Fixed in 53fa232: worker liveness is now computed at most once per watchdog pass (lazily, so we still skip the ping entirely when nothing is stale), cached in `workers_alive_this_pass`. Added a regression test asserting the ping fires exactly once across multiple stale in-flight attempts.

justin-tahara:
Per @Bo's request, added a `TODO(bo)` in `heartbeat.py` (commit 3f68d0a) flagging that this PR **mitigates** the destructive misfire but does **not** root-cause why `heartbeat_counter` can stall for the full ~30m timeout while the worker is alive. The TODO records the candidate mechanisms so a future investigation has a head start:
- heartbeat DB write starving on the shared worker pool (`pool_size = concurrency + 8`) while task threads block on a saturated embedder;
- row-lock contention on the `index_attempt` row (no `lock_timeout` set);
- CPU/GIL starvation of the docprocessing pod's daemon threads.

The new escalated ERROR logging is the diagnostic hook to tell these apart the next time it happens. Intent is to unblock now (stop discarding multi-day progress) with Bo picking up the root-cause as a follow-up.

PR #13009, opened 2026-07-14
refactor(auth): clean up auth-related env-vars

raunakab:
## Description

Auth type is a static deployment-time value — it is set once when you configure a deployment and never changes at runtime. This PR reflects that in the frontend:

- Replaces `NEXT_PUBLIC_CLOUD_ENABLED` (a derived boolean that obscured the actual auth mode) with `NEXT_PUBLIC_AUTH_TYPE`, a direct mirror of the backend's `AUTH_TYPE` env var. `NEXT_PUBLIC_CLOUD_ENABLED` is kept as a derived alias so existing call sites are unaffected.
- Removes `authType` from `AuthTypeMetadata`. Since auth type is a build-time constant, it has no business being a field on a runtime-fetched API response. All call sites now import `NEXT_PUBLIC_AUTH_TYPE` directly.
- `next.config.js` derives `NEXT_PUBLIC_AUTH_TYPE` from `AUTH_TYPE` at Next.js startup, so self-hosted deployments and dev environments that set `AUTH_TYPE` don't need a separate `NEXT_PUBLIC_AUTH_TYPE` build arg.

## How Has This Been Tested?

Type-checked with `bun run types:check` — no errors.

## Additional Options


raunakab:
Addressed all review comments in bbf06eed25:

**`svc.ts` / `svcSS.ts` — self-hosted deployments concern (greptile, cubic):** Not reverting to `data.auth_type` — auth type is intentionally a build-time constant in this PR. Self-hosted OIDC/SAML/Google OAuth deployments still work because `next.config.js` derives `NEXT_PUBLIC_AUTH_TYPE` from `AUTH_TYPE` at Next.js startup (see the `env` block added in the previous commit). This means any deployment that sets `AUTH_TYPE` at runtime picks up the correct value without needing a separate `NEXT_PUBLIC_AUTH_TYPE` build arg. Added an explanatory comment at both call sites to make this clear.

**`constants.ts` — uppercase normalization (cubic):** Added `.toLowerCase()` so `AUTH_TYPE=CLOUD` (uppercase, as the backend accepts it) doesn't silently fail the enum comparison. Fixed.

**`SettingsPage.tsx` — `showTokensSection = true` (cubic):** Removed the variable and inlined — the tokens section always renders. Fixed.

PR #13015, opened 2026-07-15
feat(auth): Classify Redis session token rejection; Redis entries now outlive logical TTL for visibility

acaprau:
## Description

**Problem:** When a session ends, neither we nor the user knows why. Redis TTL expiry is silent, so a genuine expiry is indistinguishable from Redis dropping entries (restart without persistence, eviction, flush); auth rejections log at debug level only; and the client gets the same generic 403 for every cause.

**This PR:** Session-token values now embed `issued_at`/`expires_at`. The physical Redis TTL is the logical lifetime plus a grace window (1hr), and logout overwrites the value with a tombstone instead of deleting it, so a rejected token stays classifiable: expired (embedded expiry passed), terminated (tombstone), or unrecognized (no entry: cookie outlived the grace window, or Redis dropped the key). Because `read_token` must return `None` for any invalid token (API keys/PATs/JWTs share the bearer transport and legitimately miss in Redis), the classification rides a request-scoped `ContextVar` and is raised as a reasoned `OnyxError` only where authentication conclusively fails (`double_check_user`, `/me`): the 403 body carries `SESSION_EXPIRED` / `SESSION_TERMINATED` / `SESSION_UNRECOGNIZED`, and the rejection logs at WARNING with reason, token age, and user.

Pre-upgrade values (no embedded expiry) remain valid on bare key existence, so deploying this signs nobody out.

A follow-up PR will update the logged-out modal copy based on these codes.

## How Has This Been Tested?

- [x] External-dependency-unit classification matrix against real Redis/Postgres: write/read round trip, logical expiry enforced while the key exists, tombstone on logout, absent key, legacy value accepted without rewrite, malformed value rejected, refresh extends expiry, rejection priority.
- [x] Request-level `/me` contract: expired session -> 403 `SESSION_EXPIRED`, absent session -> 403 `SESSION_UNRECOGNIZED`.

## Additional Options

---
## Summary by cubic
Adds a Redis-backed session token with embedded expiry and a 1-hour grace TTL, so session endings are explainable; logout writes a tombstone instead of deleting. `/me` and auth checks return precise 403 codes for expired, terminated, or unrecognized sessions; API keys, PATs, and JWTs are not misclassified.

- **New Features**
  - Token value stores sub, tenant_id, issued_at, expires_at; physical TTL = logical TTL + 1h; logout overwrites with a tombstone so other tabs see a sign-out.
  - Rejections classify as EXPIRED, TERMINATED, NOT_FOUND, MALFORMED; they map to 403 codes: SESSION_EXPIRED, SESSION_TERMINATED, SESSION_UNRECOGNIZED; warnings include user and token age; a stored-value-backed rejection replaces a bare miss.
  - Legacy values (no issued_at/expires_at) remain valid on key existence; `/me` reads embedded `expires_at`, or falls back to Redis TTL for legacy keys.

- **Refactors**
  - Read path parses the value and records rejections via a request-scoped ContextVar; `double_check_user` and `/me` raise a reasoned `OnyxError`. Refresh extends logical expiry while preserving `issued_at`; logout tombstones use the grace TTL; API keys/PATs/JWTs skip classification.
  - Added tests for token format, grace TTL, tombstones, classification (including replacement rules), and the `/me` 403 contract; standardized async engine `NullPool` fixture.

<sup>Written for commit e71552c55b2cf1ac98909dc114fd5967de029b8e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13015?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/auth/session_tokens.py
+++ b/backend/onyx/auth/session_tokens.py
@@ -91,8 +91,8 @@ def build_session_token_value(
     *,
     user_id: str,
     tenant_id: str | None,
-    issued_at: datetime,
-    expires_at: datetime | None,
+    issued_at: AwareDatetime,
+    expires_at: AwareDatetime | None,
 ) -> str:
     return SessionTokenValue(
         sub=user_id,
@@ -123,8 +123,8 @@ def build_session_tombstone_value(
 
 
 def compute_session_expires_at(
-    issued_at: datetime, lifetime_seconds: int | None
-) -> datetime | None:
+    issued_at: AwareDatetime, lifetime_seconds: int | None
+) -> AwareDatetime | None:
     if lifetime_seconds is None:
         return None
     return issued_at + timedelta(seconds=lifetime_seconds)
@@ -181,13 +181,15 @@ def classify_session_token_value(
 
 def record_session_rejection(rejection: SessionRejection) -> None:
     """
-    A rejection backed by a stored value outranks a bare NOT_FOUND miss from a
-    second presented credential.
+    The first rejection backed by a stored value wins; a bare NOT_FOUND miss
+    only fills a void.
     """
     existing = _SESSION_REJECTION.get()
-    if existing is not None and rejection.reason == SessionRejectionReason.NOT_FOUND:
-        return
-    _SESSION_REJECTION.set(rejection)
+    if existing is None or (
+        existing.reason == SessionRejectionReason.NOT_FOUND
+        and rejection.reason != SessionRejectionReason.NOT_FOUND
+    ):
+        _SESSION_REJECTION.set(rejection)
 
 
 def get_session_rejection() -> SessionRejection | None:
--- a/backend/onyx/auth/session_tokens.py
+++ b/backend/onyx/auth/session_tokens.py
@@ -1,0 +1,222 @@
+"""Redis session-token value format and rejection classification.
+
+Token values embed their logical expiry; the physical TTL adds a grace window
+and logout writes a tombstone instead of deleting, so a rejected token can be
+classified: EXPIRED (embedded expiry passed), TERMINATED (tombstone), NOT_FOUND
+(no entry: cookie outlived the grace window, or Redis dropped the key), LEGACY
+(pre-format value; one-time sign-out on upgrade).
+
+``read_token`` must return None rather than raise (API keys/PATs/JWTs share the
+bearer transport and legitimately miss in Redis), so the classification is
+stashed in a request-scoped ContextVar and turned into a reasoned ``OnyxError``
+only where authentication has conclusively failed for the request.
+"""
+
+from contextvars import ContextVar
+from dataclasses import dataclass
+from datetime import datetime
+from datetime import timedelta
+from datetime import timezone
+from enum import Enum
+
+from pydantic import AwareDatetime
+from pydantic import BaseModel
+from pydantic import ValidationError
+
+from onyx.auth.constants import API_KEY_PREFIX
+from onyx.auth.constants import DEPRECATED_API_KEY_PREFIX
+from onyx.auth.constants import PAT_PREFIX
+from onyx.error_handling.error_codes import OnyxErrorCode
+from onyx.error_handling.exceptions import OnyxError
+from onyx.utils.logger import setup_logger
+
+logger = setup_logger()
+
+# How long an entry outlives its logical expiry (and how long a tombstone
+# lingers); past this, rejections classify as NOT_FOUND.
+SESSION_TOKEN_GRACE_PERIOD_SECONDS = 60 * 60
+
+
+class SessionTokenValue(BaseModel):
+    """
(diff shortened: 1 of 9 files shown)


backend/onyx/auth/users.py
```diff
@@ -1455,60 +1467,87 @@ async def write_token(self, user: User) -> str:
...
-        await redis.delete(f"{self.key_prefix}{token}")
+        token_key = f"{self.key_prefix}{token}"
+        previous_raw_value = await redis.get(token_key)
+        await redis.set(
+            token_key,
+            build_session_tombstone_value(
+                previous_raw_value, fallback_user_id=str(user.id)
+            ),
+            ex=SESSION_TOKEN_GRACE_PERIOD_SECONDS,
+        )
```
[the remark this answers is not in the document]
acaprau: This race condition exists today and pre-dates this PR, a user can have 2 tabs open, refresh in one and be logged out in the other. It's an L we're just gonna have to take, it's narrow and I'm not going to do anything about it in this PR. Also, note the extra step here adds valuable information for logging/the UI, so we take that L too.

backend/onyx/auth/session_tokens.py
```diff
@@ -0,0 +1,222 @@
...
+)
+
+
+def build_session_token_value(
+    *,
+    user_id: str,
+    tenant_id: str | None,
+    issued_at: datetime,
+    expires_at: datetime | None,
+) -> str:
```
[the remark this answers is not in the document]
acaprau: Sure i can make that change but the type checker is dumb in this respect anyway, it'll pass regardless, and runtime errors will just result in logging in failing, which the most basic testing would uncover.

backend/onyx/auth/session_tokens.py
```diff
@@ -0,0 +1,222 @@
...
+
+def record_session_rejection(rejection: SessionRejection) -> None:
+    """
+    A rejection backed by a stored value outranks a bare NOT_FOUND miss from a
+    second presented credential.
+    """
+    existing = _SESSION_REJECTION.get()
+    if existing is not None and rejection.reason == SessionRejectionReason.NOT_FOUND:
+        return
+    _SESSION_REJECTION.set(rejection)
```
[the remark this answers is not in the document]
acaprau: Kind of a low IQ comment, one this scenario can never happen today with the current app and it would not make sense for it to happen in the future. Two, you're making an arbitrary judgement of order. A simpler solution might be to just make it so that NOT_FOUND does not overwrite, for whatever may be currently written.

backend/onyx/error_handling/error_codes.py
justin-tahara, line 1: Feels like we didn't necessarily need to make all of the edits in this file but odd that it did

backend/onyx/server/manage/users.py
```diff
@@ -889,43 +892,41 @@ async def get_user_role(
...
 
-        # Get the Redis client
         redis = get_raw_redis_client()
-        redis_key = REDIS_AUTH_KEY_PREFIX + token
-
-        # Get the TTL of the token
-        ttl = cast(int, redis.ttl(redis_key))
-        if ttl <= 0:
-            logger.error("Token has expired or doesn't exist in Redis")
+        raw_value = redis.get(REDIS_AUTH_KEY_PREFIX + token)
```
justin-tahara: Maybe just a small nit where we can make this set as a variable so that it's a bit more clear what this is

justin-tahara: Could reuse on line 924

backend/tests/external_dependency_unit/auth/conftest.py
justin-tahara, line 1: Do we need this file? Are you just moving it out so that we can reuse it across the tests?

justin-tahara: I guess I'm just trying to see if there is a better place for this

justin-tahara (approved):

backend/onyx/error_handling/error_codes.py
acaprau: I did this. Text should be wrapped to some consistent column, my editor happens to be set to 80, which should be the minimum. Some ppl argue 88, 100, or 120. The relevant changes in this file are lines 30-34.

backend/tests/external_dependency_unit/auth/conftest.py
acaprau: This fixture is used in 3 different test files to my knowledge. It would be worse to duplicate it in all 3.

PR #13017, opened 2026-07-15
refactor(auth): gate cloud-only behavior on MULTI_TENANT

nmgarza5:
## Description

PR 1 of the `AUTH_TYPE` removal stack (context: multi-SSO makes auth a runtime concern, so the env var's jobs move to better homes).

- The nine pure "is this Onyx Cloud" checks gate on `MULTI_TENANT` instead of `AUTH_TYPE == CLOUD`: cloud Google login setup, gmail normalization, JWT-backend validation, discord bot token/config/service-key handling, repeated-error connector pausing, `/auth/type` user-count skip, and chat rate limits.
- Two dual checks collapse (`MULTI_TENANT or CLOUD`, `not MULTI_TENANT or not CLOUD`) — every cloud deployment sets both (the EE tier resolver's multi-tenant branch requires the control plane).
- Password-model sites mixing BASIC and CLOUD are deliberately untouched; they're handled later in the stack.
- Also fixes a silently no-op test patch (`test_discord_bot_multitenant.py` patched the source module, not the consuming binding).

## Additional Options

### DIFF
--- a/backend/ee/onyx/main.py
+++ b/backend/ee/onyx/main.py
@@ -42,14 +42,12 @@ from ee.onyx.utils.encryption import test_encryption
 from onyx.auth.users import auth_backend
 from onyx.auth.users import create_onyx_oauth_router
 from onyx.auth.users import fastapi_users
-from onyx.configs.app_configs import AUTH_TYPE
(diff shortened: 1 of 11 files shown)


nmgarza5:
@greptile

justin-tahara (approved):

PR #13027, opened 2026-07-15
feat(auth): remove legacy single-provider SSO modes

nmgarza5:
## Description

Stacked on #13018. PR 3 of the AUTH_TYPE removal stack.

**Do not merge until the v4.4.0 sso_provider row migration has shipped in a release.** Legacy deployments must pick up the seed before the routes disappear.

- `AUTH_TYPE=google_oauth/oidc/saml` coerce to `BASIC` in app_configs. Provider rows serve their login. `verify_auth_setting` warns the operator at startup.
- Deletes the legacy OIDC mount block, the GOOGLE_OAUTH-only logout mount, and the legacy `/auth/saml` router (helpers stay, saml_multi consumes them). Password/mobile/refresh routers mount unconditionally. Env-Google rides on `OAUTH_ENABLED`.
- Deletes the blanket SAML/OIDC invite-whitelist bypass. Per-user `sso_managed` from #13018 replaces it.
- Seeds that migrate legacy config (alembic `1fc2904131a3` and the SAML conf-dir import) read raw `os.environ["AUTH_TYPE"]` since the coercion hides the legacy value.
- Removes the EE cloud logout mount that duplicated the now-unconditional auth router, and `get_logout_router` with it.

## Additional Options

- [ ] I have considered whether this PR needs to be cherry-picked to the latest release branch.
- [x] Override Linear Check

PR #13029, opened 2026-07-15
feat(auth): remove AUTH_TYPE and the AuthType enum, serve multi_tenant

nmgarza5:
## Description

Stacked on #13027. Final PR of the AUTH_TYPE removal stack. Merge after #13027.

**Backend**
- `app_configs.AUTH_TYPE` deleted; former consumers derive from `MULTI_TENANT` (`INSTANCE_TYPE`, invite email copy, the tag-fix migration's cloud override, which imported the deleted config and would have crashed alembic).
- `AuthType` enum deleted entirely. It carried one bit: after the legacy-mode removal, `basic` vs `cloud` is exactly `MULTI_TENANT`, so `/auth/type` serves `multi_tenant: bool`. Per-provider login type lives where it belongs, in `sso_providers[].provider_type`.
- `verify_auth_setting` is a pure deprecation warner on raw `os.environ`. The identical EE override is deleted.
- `/auth/type` stops doing per-hit DB work: `has_users` latches once a user exists, and the sso options query sits behind a 60s TTLCache invalidated by the sso admin mutations.
- Migration seeds and the SAML conf-dir seed still read the raw env value so legacy deployments migrate on upgrade. Deployment templates intentionally untouched for the same reason.

**Web**
- FE `AuthType` enum and `autoRedirect` deleted; `multiTenant: boolean` with null-safe mapping (`=== false` / `!== false`) so backend-unreachable behavior is unchanged. Dead OIDC/SAML/GOOGLE_OAUTH branches removed.
- MCP/OpenAPI OAuth pass-through option re-gated on OAuth capability (`oauthEnabled` or an OIDC/Google provider row) via a shared `useOAuthPassThroughEnabled()` hook. Also now correctly available on cloud.

**Mobile** (in dev, unshipped)
- Renders password always and Google on `oauth_enabled` (heals migrated env-Google servers). Server probe accepts old and new field shapes.

Note: external readers of `GET /auth/type` should read `multi_tenant` instead of `auth_type`.

## Additional Options

- [ ] I have considered whether this PR needs to be cherry-picked to the latest release branch.
- [x] Override Linear Check

### DIFF
--- a/backend/onyx/server/manage/models.py
+++ b/backend/onyx/server/manage/models.py
@@ -11,7 +11,6 @@ from pydantic import field_validator
 from pydantic import model_validator
 
 from onyx.auth.schemas import UserRole
-from onyx.configs.constants import AuthType
 from onyx.context.search.models import SavedSearchSettings
 from onyx.db.enums import DefaultAppMode
 from onyx.db.enums import SSOProviderType
@@ -60,8 +59,9 @@ class SSOProviderOption(BaseModel):
     authorize_url: str
 
 
-class AuthTypeResponse(BaseModel):
-    auth_type: AuthType
+class AuthConfigResponse(BaseModel):
+    # Cloud (multi-tenant) signup provisions a tenant and offers Google login.
+    multi_tenant: bool
     # specifies whether the current auth setup requires
     # users to have verified emails
     requires_verification: bool
@@ -76,7 +76,8 @@ class AuthTypeResponse(BaseModel):
     has_users: bool = True
     oauth_enabled: bool = False
     # Enabled DB-backed SSO providers, one login button each. Empty on cloud and
-    # on instances with no provider rows, so the page falls back to auth_type.
+    # on instances with no provider rows, so the page falls back to the built-in
+    # password (and Google when oauth_enabled) login.
     sso_providers: list[SSOProviderOption] = []
 
 
--- a/web/src/lib/auth/components.tsx
+++ b/web/src/lib/auth/components.tsx
@@ -8,7 +8,6 @@ import Modal from "@/refresh-components/Modal";
 import { Button, Text } from "@opal/components";
 import { SvgLogOut, SvgCheckCircle, SvgXCircle } from "@opal/icons";
 import { SvgGoogle } from "@opal/logos";
-import type { IconProps } from "@opal/types";
 import { useCaptcha } from "@/lib/hooks/useCaptcha";
 import { verifyCaptchaForOAuth } from "@/lib/auth/svc";
 import { basicLogin, basicSignup } from "@/lib/users/svc";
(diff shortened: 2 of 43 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

web/src/lib/auth/components.tsx
```diff
@@ -115,27 +114,23 @@ export function AuthenticationShell({ children }: AuthenticationShellProps) {
...
     icon = SvgGoogle;
-  } else if (authType === AuthType.OIDC) {
-    button = "Continue with OIDC SSO";
-  } else if (authType === AuthType.SAML) {
-    button = "Continue with SAML SSO";
   }
 
   if (!button) {
-    throw new Error(`Unhandled authType: ${authType}`);
+    throw new Error(`Unhandled multiTenant value: ${multiTenant}`);
```
raunakab: The `false` branch now throws `Unhandled multiTenant value: false`, but `false` isn't unhandled — it's the self-hosted case where the button simply shouldn't render. Every callsite currently guards `cloud && authUrl` before mounting, so the throw is never reached in practice, but semantically `false` is a known, valid value and a future callsite could hit it.

Suggest returning `null` for the falsy case instead of throwing.

raunakab (approved):

nmgarza5:
@greptile

nmgarza5:
@greptile

web/src/lib/auth/components.tsx
nmgarza5: Good catch. Went one step further than returning null. Every call site already gates on cloud before mounting, so the prop carried no information at all. Dropped it entirely in 684e147012. The component is just the Google button now, and the invalid state can't be expressed in the first place.

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

backend/onyx/server/manage/models.py
```diff
@@ -76,7 +76,8 @@ class AuthTypeResponse(BaseModel):
     has_users: bool = True
     oauth_enabled: bool = False
     # Enabled DB-backed SSO providers, one login button each. Empty on cloud and
-    # on instances with no provider rows, so the page falls back to auth_type.
+    # on instances with no provider rows, so the page falls back to the built-in
+    # password (and Google when oauth_enabled) login.
     sso_providers: list[SSOProviderOption] = []
```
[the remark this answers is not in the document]
nmgarza5: Intentional. The description you reviewed was stale, it described an interim design. The current description says: external readers of GET /auth/type should read multi_tenant instead of auth_type. The field carried one bit and that bit is multi_tenant.

nmgarza5:
@greptile

nmgarza5:
@greptile

justin-tahara (approved):

PR #13052, opened 2026-07-15
refactor(llm): drop hardcoded infer_vision_support, use LiteLLM cost map

jmelahman:
## Summary

`supports_image_input` was determined inconsistently across dynamic LLM providers. Most read it from the source API (Bedrock `inputModalities`, OpenRouter `input_modalities`, LM Studio `capabilities.vision`, Ollama `capabilities`). **Bifrost already switched to `litellm_thinks_model_supports_image_input`** — "*Vision support from the LiteLLM cost map, not a hardcoded list*" — leaving `infer_vision_support` (and its two hardcoded substring lists, including the Bifrost-named `BIFROST_VISION_MODEL_FAMILIES`) used by only two call sites:

- Bedrock inference-profile **fallback** (`get_bedrock_available_models`) — hit only when the base model isn't in-region
- OpenAI-compatible `/v1/models` (`get_openai_compatible_server_available_models`)

This routes both through `litellm_thinks_model_supports_image_input` and deletes `infer_vision_support`, `BEDROCK_VISION_MODELS`, and `BIFROST_VISION_MODEL_FAMILIES`, so vision support is sourced consistently.

## Changes

- `onyx/server/manage/llm/api.py` — both call sites now use `litellm_thinks_model_supports_image_input(<model_id>, <provider>)`; dropped the unused import.
- `onyx/server/manage/llm/utils.py` — deleted `infer_vision_support` and its two backing frozensets; updated the module docstring.
- `tests/unit/.../test_llm_provider_utils.py` — removed `TestInferVisionSupport`.

## Behavioral notes (LiteLLM vs. the old hardcoded lists)

Verified head-to-head against real Bedrock inference-profile IDs:

- **Same** for every common vision family — Claude 3/4, Nova Pro/Lite — and correct `False` for text-only (Nova Micro, Llama 3.2 1B, Mistral Large).
- **Better**: the old list contained no Bedrock Llama entries, so `meta.llama3-2-11b-instruct` (vision-capable) was wrongly reported as no-vision; LiteLLM gets it right.
- **Known regression (narrow)**: `mistral.pixtral-large-2502` is vision-capable and the old list caught it, but LiteLLM ships the Bedrock key with `supports_vision` **unset** (`None`), so it now reports `False`. This only surfaces in the *fallback* branch (AWS returns the Pixtral inference profile but not the base foundation model); the primary path still reads AWS's own `inputModalities` and is authoritative. It's recoverable via the DB-stored `supports_image_input` flag in the admin UI, and is a LiteLLM data gap likely to be fixed upstream.
- OpenAI-compatible: unknown model IDs LiteLLM doesn't recognize now resolve to `False` (with a debug-warning log) instead of matching the old substrings; self-hosted vision models absent from LiteLLM can be flagged via the admin UI.

## Testing

- `pytest backend/tests/unit/onyx/server/manage/llm/test_llm_provider_utils.py` — 44 passed
- `pytest backend/tests/unit/onyx/server/manage/llm/test_fetch_models_api.py` — 49 passed

### DIFF
--- a/backend/onyx/server/manage/llm/api.py
+++ b/backend/onyx/server/manage/llm/api.py
@@ -2065,7 +2065,6 @@ def get_openai_compatible_server_available_models(
                     name=model_id,
                     display_name=model_name,
                     max_input_tokens=model.get("context_length"),
-                    # Vision support from the LiteLLM cost map, not a hardcoded list
                     supports_image_input=litellm_thinks_model_supports_image_input(
                         model_id, LlmProviderNames.OPENAI_COMPATIBLE
                     ),
--- a/backend/onyx/server/manage/llm/api.py
+++ b/backend/onyx/server/manage/llm/api.py
@@ -1209,8 +1209,6 @@ def get_bedrock_available_models(
                                 if profile_name
                                 else generate_bedrock_display_name(profile_id)
                             ),
-                            # Vision support from the LiteLLM cost map, not a
-                            # hardcoded list
                             "supports_image_input": (
                                 litellm_thinks_model_supports_image_input(
                                     profile_id, LlmProviderNames.BEDROCK
--- a/backend/onyx/server/manage/llm/api.py
+++ b/backend/onyx/server/manage/llm/api.py
@@ -105,7 +105,6 @@ from onyx.server.manage.llm.provider_cache import get_cached_provider_listing
 from onyx.server.manage.llm.provider_cache import invalidate_provider_listing_cache
 from onyx.server.manage.llm.utils import generate_bedrock_display_name
 from onyx.server.manage.llm.utils import generate_ollama_display_name
-from onyx.server.manage.llm.utils import infer_vision_support
 from onyx.server.manage.llm.utils import is_embedding_model
 from onyx.server.manage.llm.utils import is_reasoning_model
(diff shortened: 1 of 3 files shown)


backend/onyx/server/manage/llm/api.py
```diff
@@ -1210,8 +1209,13 @@ def get_bedrock_available_models(
                                 if profile_name
                                 else generate_bedrock_display_name(profile_id)
                             ),
-                            # Infer vision support from known vision models
-                            "supports_image_input": infer_vision_support(profile_id),
+                            # Vision support from the LiteLLM cost map, not a
+                            # hardcoded list
```
jmelahman:
```suggestion
```

backend/onyx/server/manage/llm/api.py
```diff
@@ -2061,7 +2065,10 @@ def get_openai_compatible_server_available_models(
                     name=model_id,
                     display_name=model_name,
                     max_input_tokens=model.get("context_length"),
-                    supports_image_input=infer_vision_support(model_id),
+                    # Vision support from the LiteLLM cost map, not a hardcoded list
```
jmelahman:
```suggestion
```

justin-tahara (approved):

PR #13065, opened 2026-07-15
fix: github empty repos should not cause exception

yuhongsun96:
## Summary
- Treat GitHub’s 409 "Git Repository is empty." response as a non-fatal case during file listing
- Skip the file stage cleanly for empty repositories instead of raising an error
- Add a unit test covering the empty-repo tree response path

How Has This Been Tested?
Additional Options
 [Optional] Please cherry-pick this PR to the latest release version.
 [x] Override Linear Check


evan-onyx (approved):
testing irl is probably a good idea

yuhongsun96:
Btw I tested it with a new empty repo under onyx org

PR #13118, opened 2026-07-16
feat(skills): add per-user skill enablement

wenxi-onyx:
## Description

Replaces the tenant-wide `Skill.enabled` flag with per-user enablement for ordinary custom skills.

### Final state after this merges

- Ordinary custom skills have a preference per `(user, skill)`. A user can enable or disable any custom skill visible to them from the Skills page or the UUID-addressed API.
- Newly created custom skills are enabled for their owner. Sharing grants visibility but does not enable the skill for the recipient; each recipient must enable it explicitly.
- Native built-in skills remain permanently enabled and cannot receive user preferences.
- External-app-backed skills are not ordinary user-managed skills: they are hidden from the Skills API and UI, never receive user preferences, and are hydrated automatically when the linked app's credentials are ready for that user. Deauthentication removes them from subsequent hydration.
- External-app credential availability is independent of ordinary skill preferences. This PR temporarily removes the previous admin app enable/disable control that was backed by `Skill.enabled`; configured apps are available whenever their credential and access requirements are satisfied.
- Sandbox hydration applies visibility, validity, runtime availability, and per-user enablement independently. Pickers and explicit skill selection use the same effective access rules.
- Skill enablement changes are optimistic in the UI and immediately update the persisted preference and managed sandbox files.

The migration creates `user_skill_preference`, removes `Skill.enabled`, clears existing sharing for ordinary custom skills, and seeds enabled preferences for their owners. Native built-ins and external-app-backed skills do not receive preference rows. Existing external apps remain configured and available; historical app disabled state is not preserved.

This PR does not add duplicate skill-name support or change the existing unique runtime-name constraint.

### Phase 2b follow-up: runtime refresh

OpenCode caches its discovered skill catalog for a running directory instance. Updating managed files alone does not guarantee that an existing conversation observes the new effective skill set before its next prompt.

Phase 2b will add directory-instance invalidation and coordination so create, edit, delete, share, enablement, membership, authentication, and availability changes are reflected in active sandboxes before their next prompt without interrupting an in-flight turn.

### Phase 2c follow-up: external-app admin disablement

Phase 2c will restore the admin kill switch as `ExternalApp.enabled`, independent of the linked skill and `UserSkillPreference`. An external-app skill will then be hydrated, and its credentials made available, only when the app is admin-enabled and the user is authenticated. Disabling or deleting an app will remove its linked skill through the Phase 2b refresh mechanism.

Existing apps will default to enabled when Phase 2c lands; it will not attempt to reconstruct the historical `Skill.enabled` state removed by this migration.

## Additional Options


backend/onyx/db/external_app.py
```diff
@@ -252,15 +279,15 @@ def create_external_app(
     upstream_url_patterns: list[str],
     auth_template: dict[str, Any],
     organization_credentials: dict[str, str],
-    enabled: bool = True,
     is_public: bool = False,
```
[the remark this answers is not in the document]
wenxi-onyx: Addressed in f905c6d0fc. The integration manager and Docker E2E setup no longer pass or assert the removed external-app `enabled` field.

backend/onyx/db/skill.py
```diff
@@ -217,41 +233,11 @@ def _skill_select_for_access_policy(
...
-        stmt = stmt.where(
-            visible_to_user,
-            or_(Skill.enabled.is_(True), editable_by_user),
-        )
-        return _exclude_unavailable_built_in_skills(
-            stmt,
-            db_session=db_session,
-        )
+        stmt = stmt.where(skill_visible_to_user(user))
+        return _exclude_unavailable_built_in_skills(stmt, db_session)
```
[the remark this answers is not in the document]
wenxi-onyx: Addressed in 6103928328. External-app-backed skills are now excluded from the general skills VIEW response, so they are no longer presented as editable skills.

backend/alembic/versions/b1d060254b02_per_user_skill_preferences.py
```diff
@@ -0,0 +1,117 @@
...
+        CROSS JOIN external_app
+        JOIN skill ON skill.id = external_app.skill_id
+        """
+    )
+
+    op.drop_column("skill", "enabled")
+
+
+def downgrade() -> None:
+    # The custom-skill share reset and previous tenant-wide enabled values
```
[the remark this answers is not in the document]
wenxi-onyx: Addressed in 6103928328. External-app skills no longer use per-user skill preferences, so the user × external-app CROSS JOIN seed was removed from the migration.

web/src/lib/skills/picker.ts
```diff
@@ -39,9 +39,14 @@ export function toPickerSections(
 
   const skills: PickerSkill[] = [];
   const apps: PickerApp[] = [];
+  const enabledAppSlugs = new Set(
+    [...(skillsData?.builtins ?? []), ...(skillsData?.customs ?? [])]
+      .filter((skill) => skill.is_external_app && skill.enabled)
```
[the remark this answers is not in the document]
wenxi-onyx: Addressed in 6103928328 by removing external-app-backed skills from the skills response and picker enablement path. External apps are now listed from app metadata and gated by user authentication during hydration.

web/src/views/SkillsPage.tsx
```diff
@@ -55,6 +62,52 @@ export default function SkillsPage() {
...
+          return {
+            ...current,
+            [key]: current[key].map((skill) =>
+              skill.id === updatedSkill.id ? updatedSkill : skill
+            ),
+          };
+        },
+        { revalidate: false }
+      );
+      void refresh().catch(() => {
```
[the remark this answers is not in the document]
wenxi-onyx: Not changing this. We do not generally log frontend request failures to the browser console, and this background refresh already retains the authoritative optimistic result while showing a user-facing warning. There is no established error-tracker hook in this path to use instead.

wenxi-onyx:
@greptile-apps Follow-up on the two findings reported outside the diff:

- The implicit `ExternalApp` join dependency was removed in 6103928328. `_is_enabled_for_user` now depends only on `Skill` and `UserSkillPreference`; external-app authentication is handled separately in the USE policy.
- The misleading toggle error was fixed in 19161a91f6. It now reports that the skill cannot be enabled or disabled without asserting an incorrect cause.

wenxi-onyx:
@greptile @codex @cubic-dev-ai re-review

backend/onyx/db/skill.py
```diff
@@ -71,6 +73,12 @@ class SkillValidityUpdate:
     is_valid: bool
 
 
+@dataclass(frozen=True)
```
Danelegend, line 77: pydantic

backend/alembic/versions/b1d060254b02_per_user_skill_preferences.py
```diff
@@ -0,0 +1,104 @@
...
+    # cannot be reconstructed. Downgraded skills therefore default enabled.
+    op.add_column(
```
Danelegend, line 91: No back population?

wenxi-onyx:
@greptile @codex @cubic-dev-ai re-review

backend/onyx/server/features/skill/api.py
```diff
@@ -138,6 +141,24 @@ def list_skills_for_current_user(
     return skills_list_response_for_user(rows, user, db_session)
 
 
+@user_router.put("/{skill_id}/enabled")
+def set_skill_enabled_for_current_user(
```
Danelegend, line 145: Only enable? Do we not want users to be able to disable skills

Danelegend (approved):

backend/onyx/server/features/skill/api.py
wenxi-onyx: `enabled` not `enable` --> it accepts enabled bool

backend/alembic/versions/b1d060254b02_per_user_skill_preferences.py
wenxi-onyx: Whatever previous value that existed is not derivable from the post migration state. Instead, we just do server default true (enabled). I think this is fine

backend/onyx/db/skill.py
wenxi-onyx: Pros of Pydantic:
  - Runtime validation catches incorrect values at construction.
  - Convenient serialization with model_dump().
  - Generates schemas and integrates naturally with API boundaries.
  - Helpful when values originate from untrusted or loosely typed input.

  Cons here:
  - Adds runtime validation and object-construction overhead to DB-internal paths.
  - Suggests these are API/data-transfer models when they are not.
  - Default coercion can conceal type mistakes unless strict configuration is added.
  - Neither object needs serialization, JSON schemas, aliases, or input parsing

PR #13126, opened 2026-07-17
feat(craft): sandbox-proxy MCP credential resolver

Danelegend:
## Description

PR 3 of the Craft MCP Support plan: MCP servers an admin flags `available_in_craft` become reachable from the Craft sandbox with **zero credential duplication** — the proxy injects the sandbox owner's credentials from the same `mcp_connection_config` rows chat writes, so connecting a server on either surface authenticates both. The sandbox dials the real server URL with no auth headers; secrets only ever exist at the proxy.

**`sandbox_proxy/resolvers/mcp_server.py`** (registered in `build_resolvers()`):
- Exact URL-prefix attribution against craft-enabled servers' `server_url` — no glob/provider patterns. A request to a claimed host outside every configured prefix is **denied, never forwarded**: an MCP host must not double as a generic egress route.
- Credential resolution via the shared `resolve_mcp_credentials` / `build_headers()` (all four auth modes: NONE / API_TOKEN / OAUTH / PT_OAUTH).
- `claims()` runs on the off-catalog hot path, so craft-enabled targets are held in a 30s-TTL per-tenant cache — the one deliberate, documented exception to the "no DB in claims" rule. Admin flag flips converge within the TTL.

**`server/features/mcp/oauth_refresh.py`** — lazy OAuth refresh at the proxy. A thin wrapper over chat's shared `refresh_mcp_oauth_token_if_expired` primitive, adding only a persisted-expiry pre-check so the hot path skips refresh when the stored token is still good. The single-flight lock that serializes refreshes per config row (so a chat call and a proxy call can't race a rotating refresh token) lives on that shared primitive and is split out as #13143.

**Framework extensions** (small): the credential dispatcher returns a typed `InjectionResult` carrying agent-facing `block_detail` prose into the 403 body — missing/unrefreshable credentials fail closed with a structured 403 naming the server, which v2's connect-on-demand card will trigger off; the gate's off-catalog dispatch runs off the event loop like the ALWAYS path, since resolution can block on a token refresh.

Unreachable in practice until the opencode config emission PR turns MCP on in the sandbox — nothing is user-visible yet.

**Stacked on** #13068 (`available_in_craft` flag) and #13143 (single-flight refresh lock). The PR base is #13068's branch; #13143's commit is cherry-picked in as a dependency, so its `oauth.py` / `test_mcp_oauth_refresh.py` changes show in this diff until #13143 lands on main and the base rebases. This PR's own changes are the `sandbox_proxy/` files, `oauth_refresh.py`, and `test_mcp_server_resolver.py`.

## How Has This Been Tested?

New external-dependency-unit suite `tests/external_dependency_unit/sandbox_proxy/test_mcp_server_resolver.py` (real DB + Redis): injection per auth type, expired-token refresh through the real SDK provider machinery with persistence back to the row verified, blocked-naming-the-server on missing and on unrefreshable credentials, non-enabled servers unclaimed, out-of-prefix requests denied. Plus updated dispatcher unit tests; full `tests/unit/sandbox_proxy` suite (202) passes.

## Additional Options

---
## Summary by cubic
Adds a Craft sandbox-proxy resolver for MCP servers flagged `available_in_craft`. It injects the sandbox owner’s credentials from shared `mcp_connection_config`, refreshes OAuth when needed, and returns structured 403s with agent-facing detail.

- New Features
  - Proxy/matching: `MCPServerResolver` claims only exact scheme/host/port + normalized path-prefix URLs, blocks traversal and off-prefix requests, requires same-scheme, caches targets per-tenant for 30s, defers to already-matched external apps, and is registered before the external-app resolver.
  - Credentials/OAuth: supports NONE/API_TOKEN/OAUTH/PT_OAUTH; uses a persisted-expiry pre-check and the shared single-flight refresh to rotate tokens; access is audited.
  - Framework: the credential dispatcher returns `InjectionResult` with optional `block_detail`; `http_403(..., detail=...)` includes it; off-catalog dispatch runs via `asyncio.to_thread`.

<sup>Written for commit 0b109087bdfe13f9b7570e8e5aa3a0abecff3cc9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13126?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/sandbox_proxy/addons/gate.py
+++ b/backend/onyx/sandbox_proxy/addons/gate.py
@@ -595,8 +595,13 @@ class GateAddon:
         # ALWAYS / DENY / off-catalog terminate here; ASK falls through to the
         # approval pipeline in `request()`.
         if matched_actions is None:
-            injection = self._dispatch_injection_or_block(
-                flow, sandbox=sandbox, matched_actions=None
+            # Off-thread like the ALWAYS path: host-claiming resolvers may hit
+            # the DB or refresh an expiring OAuth token before injecting.
+            injection = await asyncio.to_thread(
+                self._dispatch_injection_or_block,
+                flow,
+                sandbox=sandbox,
+                matched_actions=None,
             )
             if injection is InjectionOutcome.BLOCKED:
                 logger.warning(
@@ -1060,16 +1065,18 @@ class GateAddon:
         matched_actions: AllMatchedActions | None,
     ) -> InjectionOutcome:
         """Runs the credential dispatcher; Fails closed with a 403 on BLOCKED."""
-        outcome = self._credential_dispatcher.apply(
+        result = self._credential_dispatcher.apply(
             flow,
             InjectionContext(
                 sandbox=sandbox,
                 matched_actions=matched_actions,
             ),
         )
-        if outcome is InjectionOutcome.BLOCKED:
-            flow.response = http_403(SandboxProxyError.CREDENTIAL_ERROR)
-        return outcome
+        if result.outcome is InjectionOutcome.BLOCKED:
+            flow.response = http_403(
+                SandboxProxyError.CREDENTIAL_ERROR, detail=result.block_detail
+            )
+        return result.outcome
 
     def _terminalize_after_unhandled_error(
         self, approval_id: UUID, tenant_id: str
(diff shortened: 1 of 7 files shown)


Danelegend:
@greptile @cubic

wenxi-onyx (approved):

PR #13019, opened 2026-07-15
feat(auth): capture SAML directory attributes for profile enrichment

nmgarza5:
## Description

Extends the opt-in IdP directory-profile enrichment (#12997) to SAML logins.

- SAML asserts the same directory data (department, title, ...) as OIDC, just as assertion attributes instead of token claims.
- Capture flattens `auth.get_attributes()` into the same Redis snapshot under a `saml_attributes` source. Resolution reuses the existing claim map, so deployments map attribute names via `IDP_PROFILE_CLAIM_MAP` exactly like OIDC.
- Hooked into the active multi-IdP SAML callback (`saml_multi.py`), best-effort and timeout-bounded like the OAuth path.
- Single-tenant only until per-tenant capture. The `MULTI_TENANT` gate still applies.

## Additional Options

- [ ] [Optional] Please cherry-pick this PR to the latest release version.
- [x] [Optional] Override Linear Check

PR #13135, opened 2026-07-17
feat(auth): per-tenant claims capture, enable IdP enrichment on multi-tenant

nmgarza5:
## Description

Stacked on #13019 (merge that first). Removes the multi-tenant restriction from IdP profile enrichment.

- The login callbacks that run claims capture are unauthenticated, so the tenant contextvar holds the default schema there. Capture now resolves the tenant from the email mapping (the same lookup `complete_login_flow` uses) and keys the snapshot with it.
- Drops the `MULTI_TENANT` force-off gate. Single-tenant behavior unchanged.
- First login of a brand-new user (no mapping yet) skips capture rather than writing under the default schema; the next login captures.

Design note for review: write-side keys by the email mapping, read-side keys by the authenticated request's contextvar. These agree under Onyx's one-active-tenant-per-email model.

## Additional Options

### DIFF
--- a/deployment/helm/charts/onyx/Chart.yaml
+++ b/deployment/helm/charts/onyx/Chart.yaml
@@ -5,7 +5,7 @@ home: https://www.onyx.app/
 sources:
   - "https://github.com/onyx-dot-app/onyx"
 type: application
-version: 0.6.15
+version: 0.6.16
 appVersion: latest
 annotations:
   category: Productivity
--- a/backend/onyx/auth/login_claims_capture.py
+++ b/backend/onyx/auth/login_claims_capture.py
@@ -31,6 +31,7 @@ from typing import Any
 
 import httpx
 import jwt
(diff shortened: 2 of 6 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

deployment/helm/charts/onyx/Chart.yaml
```diff
@@ -5,7 +5,7 @@ home: https://www.onyx.app/
 sources:
   - "https://github.com/onyx-dot-app/onyx"
 type: application
-version: 0.6.15
+version: 0.6.16
```
nmgarza5, line 8: this is because we have a ci check requiring a bump for any helm chart change (even for code comments)

justin-tahara (approved):

PR #13143, opened 2026-07-17
fix(mcp): single-flight the OAuth refresh primitive per config row

Danelegend:
## Description

`refresh_mcp_oauth_token_if_expired` (the shared MCP OAuth refresh primitive, added in #13085) has no concurrency control. Two in-process refreshes racing the same `mcp_connection_config` row can each redeem the same rotating refresh token; a provider that rotates refresh tokens on use may then reject the loser **and revoke the whole grant** as reuse detection. This is reachable on main today — two concurrent chat SSE tool calls to the same server with an expiring token, or two api-server processes — with no Craft involvement.

- Serialize the primitive with a Redis single-flight lock keyed by `(tenant, connection_config_id)`. On lock contention, hand back whatever the winning refresher has already persisted rather than refreshing again.
- Bound the token-refresh POST to an explicit 30s timeout instead of the MCP SDK client's SSE-sized default (30s connect but **300s read**), and size the lease (60s) to strictly outlast a whole refresh so the lock can't expire mid-POST and re-open the race.
- `mcp_token_expired` moves here as the shared freshness predicate (also used by the contention fallback).

Every in-process caller of this primitive is now coordinated: chat's SSE tool calls today, and the Craft sandbox proxy (#13126) once it lands.

**Known residual:** chat's STREAMABLE_HTTP path refreshes in-band through the MCP SDK's own `httpx.Auth` flow, which does not call this function and so isn't covered by this lock. Coordinating that path is the unify-oauth-refresh follow-up.

## How Has This Been Tested?

Existing MCP OAuth refresh coverage (`tests/unit/onyx/server/features/mcp/test_mcp_oauth_refresh.py`, `test_mcp_known_provider_oauth_helpers.py` — 32 tests) passes with the refactor. The single-flight lock's proxy-facing behavior (contention → persisted-header fallback) is exercised end-to-end in the Craft resolver's suite in #13126.

## Additional Options

---
## Summary by cubic
Single-flight MCP OAuth token refresh per connection-config row using a backend-agnostic `cache_shared_lock` to stop racing refreshes from burning rotating tokens. On contention, the loser waits out the refresh and returns the winner’s persisted Authorization header; the refresh POST is capped at 30s with a 60s lease.

- **Bug Fixes**
  - Serialize refresh with `cache_shared_lock` keyed by tenant and `connection_config_id`; wait up to 35s to outlast the 30s POST, then return the persisted header via `mcp_token_expired` (or None if still expired).
  - Set an explicit 30s `httpx` timeout for the refresh POST; lock lease is 60s to avoid mid-refresh expiry. Note: the STREAMABLE_HTTP path via the SDK’s `httpx.Auth` is not yet coordinated.

- **New Features**
  - Add `cache_shared_lock` context manager and `CacheLockAcquisitionError`, backed by the configured cache backend (Redis or Postgres advisory locks) for cross-deployment single-flight.

<sup>Written for commit 55b6f2257b5f7eccd5876e9ec4501e401e728156. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13143?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/cache/interface.py
+++ b/backend/onyx/cache/interface.py
@@ -28,6 +28,12 @@ class CacheLockLostError(Exception):
     it — mutual exclusion is already gone, unlike CACHE_TRANSIENT_ERRORS."""
 
 
+class CacheLockAcquisitionError(Exception):
+    """A shared cache lock could not be acquired within the allotted wait — a
+    concurrent holder still owns it. Distinct from CacheLockLostError, which is
+    about losing a lock already held."""
+
+
 class CacheLock(abc.ABC):
     """Abstract distributed lock returned by CacheBackend.lock()."""
 
--- a/backend/onyx/cache/locks.py
+++ b/backend/onyx/cache/locks.py
@@ -1,0 +1,61 @@
+import time
+from collections.abc import Generator
+from contextlib import contextmanager
+from logging import Logger
+from logging import LoggerAdapter
+
+from onyx.cache.factory import get_shared_cache_backend
+from onyx.cache.interface import CacheLockAcquisitionError
+
+
+@contextmanager
+def cache_shared_lock(
+    lock_name: str,
+    max_time_lock_held_s: float,
+    wait_for_lock_s: float,
+    logger: Logger | LoggerAdapter,
+) -> Generator[None, None, None]:
+    """Acquire a system-wide (cross-tenant) distributed lock via the configured
+    cache backend.
+
+    Unlike ``redis_shared_lock``, this works on every deployment: full ones back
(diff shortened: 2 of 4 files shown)


Danelegend:
@cubic @greptile

PR #13145, opened 2026-07-17
feat(cache): add cache_shared_lock backend-agnostic distributed lock

Danelegend:
## What

Adds `cache_shared_lock`, a system-wide (cross-tenant) distributed lock context manager acquired through the configured `CacheBackend`, plus `CacheLockAcquisitionError` for the contended case.

## Why

Unlike `redis_shared_lock`, this works on **every** deployment:
- Full deployments back the lock with **Redis**.
- Lite deployments (`CACHE_BACKEND=postgres`) back it with a **PostgreSQL advisory lock**.

Callers get the same single-flight guarantee without assuming Redis is available — the engineering rules require code to work on both full and Lite deployments.

This is the base of a small stack: the first consumer is the MCP OAuth single-flight refresh (#follow-up PR), which is why this lands on its own so the primitive can be reviewed independently.

## Notes

- No caller on this branch yet — the consumer arrives in the stacked PR.
- Semantics mirror the existing `redis_shared_lock` (blocking acquire with wait timeout, auto-release on exit), but delegate to `CacheBackend.lock()`.

## Additional Options

- [ ] [Optional] Please cherry-pick this PR to the latest release version.
- [x] [Optional] Override Linear Check

---
## Summary by cubic
Adds a backend-agnostic distributed lock context manager `cache_shared_lock` for cross-tenant single-flight via the configured `CacheBackend`. Introduces `CacheLockAcquisitionError` on acquisition timeouts and logs a warning if the lease expires mid-work.

- **New Features**
  - Cross-tenant `cache_shared_lock(...)` via `get_shared_cache_backend().lock(...)`, backed by Redis (Full) or PostgreSQL advisory locks (Lite).
  - Blocking acquire with `wait_for_lock_s` and lease `max_time_lock_held_s`; auto-release on exit with a debug log; raises `CacheLockAcquisitionError` on timeout; warns if the lease is exceeded (lease TTL enforced on Redis; Postgres advisory locks have no TTL).

<sup>Written for commit a952f604fd97233ab5881dea77dde2f19ecc4641. Summary will update on new commits.</sup>

[...]

PR #13146, opened 2026-07-17
fix(mcp): single-flight the OAuth refresh primitive per config row

Danelegend:
## What

Serializes SSE-transport MCP OAuth token refreshes **per connection-config row** so two racing callers can't redeem — and burn — the same rotating refresh token. On contention, the loser returns whatever the winner has already persisted rather than refreshing again.

## Why

`refresh_mcp_oauth_token_if_expired` is hit concurrently by chat's SSE tool calls (`MCPTool.run`) and the Craft sandbox proxy. Without coordination, two callers can each POST the same refresh token; providers that rotate refresh tokens invalidate the first, breaking the other caller.

## How

Coordinates via the shared cache backend (`cache_shared_lock`) rather than a Redis-only lock, so single-flight holds on **both** full (Redis) and Lite (Postgres advisory lock) deployments without assuming Redis is available.

## Stacked on

- Base: **#13145** (`feat(cache): add cache_shared_lock ...`). This PR targets `dane/cache-shared-lock` so the diff shows only the MCP-side changes (`oauth.py` + tests). Rebase/retarget to `main` once #13145 merges.

## Tests

`backend/tests/unit/onyx/server/features/mcp/test_mcp_oauth_refresh.py` — pure unit tests (cache lock stubbed, no live Redis/Postgres), covering refresh, auth-method branching, the no-refresh cases, and the lock-contention fallback.

## Additional Options

---
## Summary by cubic
Single-flights MCP OAuth token refresh per connection-config row (tenant-scoped) to stop two callers from redeeming the same rotating refresh token. On contention, we wait for the in-flight refresh and reuse the winner’s persisted Authorization header; if the stored token is still expired, return None.

- **Bug Fixes**
  - Coordinate refresh via `cache_shared_lock` so single-flight works in both Redis and Lite (Postgres advisory lock); key is `mcp_token_refresh:{tenant_id}:{config_id}`.
  - Use `httpx` with a 30s timeout for the refresh POST; set lock wait/lease to 35s/60s so the loser wakes after the refresh and sees the fresh header.
  - Add helpers to detect token expiry and read persisted headers; on lock acquisition failure, return the stored header only if still valid. Unit tests cover refresh, no-refresh, and contention (header or None) with no network calls.

<sup>Written for commit a59d35a5e4bfcb3fe938acc888bace6567a9bbf7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13146?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/server/features/mcp/oauth.py
+++ b/backend/onyx/server/features/mcp/oauth.py
@@ -12,6 +12,7 @@ import time
 from typing import Any
 from urllib.parse import urlparse
 
+import httpx
 from mcp.client.auth import OAuthClientProvider
 from mcp.client.auth import TokenStorage
 from mcp.client.auth.oauth2 import OAuthContext
@@ -23,6 +24,8 @@ from pydantic import AnyUrl
 from pydantic import BaseModel
 from sqlalchemy.orm import Session
 
+from onyx.cache.interface import CacheLockAcquisitionError
+from onyx.cache.locks import cache_shared_lock
 from onyx.configs.app_configs import WEB_DOMAIN
 from onyx.db.engine.sql_engine import get_session_with_current_tenant
 from onyx.db.enums import MCPOAuthProviderMode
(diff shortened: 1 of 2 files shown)


Danelegend:
@greptile @cubic

evan-onyx (approved):

PR #13158, opened 2026-07-18
feat(mcp): emit craft-enabled MCP servers into opencode config

Danelegend:
## What

Makes MCP servers usable by the Craft agent by emitting an opencode `mcp` block for every craft-enabled server and threading it through the sandbox provision path. This is the "config emission" phase of Craft MCP Support (stacked on the proxy resolver PR).

## How

- `build_multi_provider_opencode_config` gains a keyword-only `mcp_servers` param and emits `config["mcp"]` = `{key: {type: "remote", url, enabled}}` — **no auth headers** (the sandbox proxy injects credentials per request, so the static config never changes when a user connects a server).
- `resolve_craft_mcp_servers` (new `sandbox/util/mcp_config.py`) builds that input from `get_craft_enabled_mcp_servers` + each server's `Tool.enabled` curation, and maps enabled tools → `allow` / disabled → `deny` in the permission block (proxy is the sole gate, so opencode must not double-prompt).
- Threaded through `provision()` in both the k8s and docker managers alongside `all_llm_configs`, resolved in `provision_sandbox`. Pre-registered regardless of per-user connection state (mirrors the LLM-provider pre-load); regenerate-on-wake is free via re-provisioning.
- Backwards compatible: `mcp_servers` defaults to `None`, so existing provision callers are unaffected, and no `mcp` key is emitted when there are none.

## Validation

Verified live against opencode **1.15.7** with a real Braintrust MCP server (always-allow, i.e. approval gating not yet wired):

- New pod config contains `braintrust-mcp-1` (`type:remote`, url, no headers) and `<serverKey>_<tool>: allow` permission entries.
- opencode: `mcp key=braintrust-mcp-1 transport=StreamableHTTP connected`, `toolCount=7 create() successfully created client`.
- Sandbox proxy: every `api.braintrust.dev` request `credential_outcome=headers_injected` — opencode's remote-MCP client honors the proxy env (a previously-unverified risk).
- opencode evaluated `braintrust-mcp-1_<tool>` permissions to `allow` and the agent called the tools without an in-sandbox prompt.

## Notes / follow-ups

- **Approval gating is a separate PR.** Until it lands, MCP is effectively always-allow in the sandbox (off-catalog credential injection). Config emission must not ship to users ahead of the gating PR.
- Per-tool **disable** behavior is implemented (permission `deny`) but not yet exercised live (the test server had all tools enabled).
- AGENTS.md per-server connection-state hint is not included here (small follow-up).

## Tests

- Unit: `mcp` block emission, no-`mcp`-key-when-empty, and enabled→allow / disabled→deny permission mapping in `test_opencode_config.py`.
- External-dependency-unit: `resolve_craft_mcp_servers` against a seeded craft-enabled server + tools (`test_mcp_config_resolution.py`).

- [x] Override Linear Check

---
## Summary by cubic
Adds Craft MCP support by emitting an `opencode` `mcp` block and wiring it through sandbox provisioning so the agent can call remote MCP tools without extra prompts. Servers are scoped to each user’s access; the proxy enforces access and injects credentials per request, and static config never includes auth.

- **New Features**
  - `build_multi_provider_opencode_config` accepts `mcp_servers` (Sequence, default `()`) and emits `mcp` entries with `type: "remote"`, real `url`, `enabled: true`; permissions add `<serverKey>_*: allow` plus per-tool `deny` from admin curation (now built in `_build_permissions`).
  - Added `CraftMCPServerConfig`; `resolve_craft_mcp_servers` builds stable keys (`slug-id`), bulk-fetches tool curation via `get_mcp_tools_for_servers`, and threads into Docker/Kubernetes `provision` and `provision_sandbox` with an always-present `mcp_servers` flow.

- **Bug Fixes**
  - Filter craft-enabled servers to public/shared/owned for the user in config emission and the `/servers/craft` listing, and require access in the proxy before credential injection; host-claim matching stays unfiltered (no user yet) and auth remains per-request via the proxy.

<sup>Written for commit 8dfcfbc0ca771a015108c32f8a4db702800d25f8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13158?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/server/features/build/sandbox/util/mcp_config.py
+++ b/backend/onyx/server/features/build/sandbox/util/mcp_config.py
@@ -27,10 +27,9 @@ def resolve_craft_mcp_servers(
     """Craft-enabled MCP servers ``user`` may use, as opencode config input.
     Two queries: the servers, then a bulk fetch of their tools.
 
-    Access (public / shared / owned) is filtered here; authentication is not —
-    auth state changes mid-session (connect, token expiry) while this config is
-    baked at provision, so the proxy enforces credentials per request and
-    surfaces a "connect this server" detail for unauthenticated ones."""
+    Access is filtered here; authentication is not — auth state changes
+    mid-session while this config is baked at provision, so the proxy enforces
+    credentials per request."""
     servers = get_craft_enabled_mcp_servers(db_session, user)
     disabled_by_server: dict[int, list[str]] = defaultdict(list)
     for tool in get_mcp_tools_for_servers([s.id for s in servers], db_session):
--- a/backend/onyx/server/features/build/sandbox/util/mcp_config.py
+++ b/backend/onyx/server/features/build/sandbox/util/mcp_config.py
@@ -8,7 +8,7 @@ from collections import defaultdict
 from sqlalchemy.orm import Session
 
 from onyx.db.mcp import get_craft_enabled_mcp_servers, get_mcp_tools_for_servers
-from onyx.db.models import MCPServer
+from onyx.db.models import MCPServer, User
 from onyx.server.features.build.sandbox.models import CraftMCPServerConfig
 
 _NON_IDENTIFIER = re.compile(r"[^a-z0-9]+")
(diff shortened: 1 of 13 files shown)


Danelegend:
@greptile @cubic

backend/onyx/server/features/build/sandbox/util/mcp_config.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def _server_key(server: MCPServer) -> str:
+    """Identifier-safe opencode server id; the id suffix keeps it unique across
+    servers that slugify to the same name."""
+    slug = _NON_IDENTIFIER.sub("-", server.name.lower()).strip("-") or "mcp"
+    return f"{slug}-{server.id}"
+
+
+def resolve_craft_mcp_servers(db_session: Session) -> list[CraftMCPServerConfig]:
```
wenxi-onyx: Should this resolve what MCPs the user actually has access to + are authenticated?

backend/onyx/server/features/build/sandbox/util/opencode_config.py
```diff
@@ -138,13 +166,17 @@ def build_multi_provider_opencode_config(
     disabled_tools: list[str] | None = None,
     dev_mode: bool = False,
     plugins: list[str] | None = None,
+    mcp_servers: list[CraftMCPServerConfig] | None = None,
```
wenxi-onyx: nit: i think empty list for none makes this more clear instead of allowing optional none. Could then possibly wrap mcp perms into _build_permissions so it all happens in one place

wenxi-onyx (approved):
lgtm - note there is some orchestration when a user has a live sandbox+session and they authenticate or remove an MCP server --> the open code session has to be `disposed` so it picks up the new configs. You can see how to do this in my recent commits on main for Skills.

PR #13163, opened 2026-07-19
feat(perms): group-manager assignment & membership scoping

Subash-Mohan:
## Description

PR5 of the group-manager scoped-permissions series (stacked on PR4). Completes the backend: lets admins and in-group managers create managers, scopes group membership edits, and exposes the manager flag to clients.

- Add a manager-assignment endpoint (`PUT /admin/user-group/{group_id}/manager`) plus `make_group_manager`/`revoke_group_manager` — admins assign anywhere, a scoped manager only within a group they manage; non-member targets are rejected.
- Scope group membership edits (patch, add-users, rename) so a manager can only edit a group they manage.
- Close the junction-rewrite escalation vector: a newly-attached connector must be private and within the manager's managed scope (authorized against `is_current` junction rows only, so a removed-but-not-yet-swept pair can't slip past).
- Scope the admin group list so a manager sees only the groups they manage (admins keep the full list).
- Expose `is_manager` + `managed_group_ids` on `/me/permissions` so the frontend can reveal manager nav.
- Group create, delete, and set-permissions stay admin-only.

## How Has This Been Tested?

Added an integration escalation suite (`test_group_manager_membership.py`, 22 tests) covering the assignment endpoint, membership/rename gates, the cc_pair re-attach gate (incl. a stale-junction-row regression), the scoped group list, `/me/permissions` flags, and the admin-only carve-outs. Verified with ruff + ty + the `has_permission` unit suite + `pytest --collect-only`; the integration suite needs a live EE deployment to run (pending, same as PR3/PR4).

## Additional Options

### DIFF
--- a/backend/ee/onyx/db/user_group.py
+++ b/backend/ee/onyx/db/user_group.py
@@ -473,6 +473,22 @@ def _mark_user_group__cc_pair_relationships_outdated__no_commit(
         user_group__cc_pair_relationship.is_current = False
 
 
+def _current_cc_pair_ids(db_user_group: UserGroup) -> list[int]:
+    """The cc_pairs currently attached to the group — is_current junction rows only.
+
+    A removed cc_pair keeps a stale ``is_current=False`` row until the Vespa sync
+    deletes it, and the plain ``cc_pairs`` relationship has no is_current filter, so
+    it would still surface the removed pair. Reading it as "current" lets a removed
+    (possibly public / out-of-scope) pair be re-attached without re-clearing the
+    scope gate, so always derive the current set from the live relationships.
+    """
+    return [
+        relationship.cc_pair_id
+        for relationship in db_user_group.cc_pair_relationships
+        if relationship.is_current
+    ]
+
+
 def add_users_to_user_group(
     db_session: Session,
     user: User,
@@ -504,7 +520,7 @@ def add_users_to_user_group(
 
     user_group_update = UserGroupUpdate(
         user_ids=current_user_ids + new_user_ids,
-        cc_pair_ids=[cc_pair.id for cc_pair in db_user_group.cc_pairs],
+        cc_pair_ids=_current_cc_pair_ids(db_user_group),
     )
 
     return update_user_group(
@@ -577,7 +593,7 @@ def update_user_group(
 
     _check_user_group_is_modifiable(db_user_group)
 
-    current_cc_pair_ids = set(cc_pair.id for cc_pair in db_user_group.cc_pairs)
+    current_cc_pair_ids = set(_current_cc_pair_ids(db_user_group))
     requested_cc_pair_ids = set(user_group_update.cc_pair_ids)
     _assert_group_update_within_scope(
         db_session,
--- a/backend/ee/onyx/db/user_group.py
+++ b/backend/ee/onyx/db/user_group.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 from collections.abc import Sequence
 from operator import and_
 from uuid import UUID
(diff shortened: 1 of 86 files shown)


backend/onyx/auth/permissions.py
```diff
@@ -255,6 +255,16 @@ def resolve_effective_permissions(granted: set[str]) -> set[str]:
     return effective
 
 
+# The bundle plus everything it implies. A scoped manager holds the bundle's
+# write tokens AND their implied reads (e.g. MANAGE_USER_GROUPS ⇒ READ_USER_GROUPS)
+# — scoped to their managed groups. has_permission classifies against this so a
+# manager resolves SCOPED for those reads; every READ + allow_scope endpoint must
+# still apply its own GATE 2 scope filter (they'd otherwise return all rows).
+SCOPED_MANAGER_PERMISSIONS_EXPANDED: frozenset[str] = frozenset(
```
nmgarza5, line 265: `has_permission` now classifies against the expanded set but `get_scoped_groups` still guards on the unexpanded `SCOPED_MANAGER_PERMISSIONS`, so a handler that resolves SCOPED on `READ_USER_GROUPS` and then asks `get_scoped_groups` for that same token gets an empty set and silently returns nothing. `list_user_groups` avoids it by passing `MANAGE_USER_GROUPS`, but that is an easy one to trip over next time.

backend/ee/onyx/db/user_group.py
```diff
@@ -518,6 +600,15 @@ def update_user_group(
 
     _check_user_group_is_modifiable(db_user_group)
 
+    current_cc_pair_ids = set(_current_cc_pair_ids(db_user_group))
+    requested_cc_pair_ids = set(user_group_update.cc_pair_ids)
+    _assert_group_update_within_scope(
+        db_session,
+        user,
+        user_group_id,
+        added_cc_pair_ids=requested_cc_pair_ids - current_cc_pair_ids,
```
nmgarza5, line 609: Only additions go through the scope check, so removals are ungated. A scoped manager can drop a public cc_pair that an admin attached to their group, and nothing stops them re-shaping the group's document access that way.

backend/ee/onyx/server/user_group/api.py
```diff
@@ -290,3 +316,27 @@ def update_group_agents(
...
+    request: SetGroupManagerRequest,
+    user: User = Depends(
+        require_permission(Permission.MANAGE_USER_GROUPS, allow_scope=True)
+    ),
+    db_session: Session = Depends(get_session),
+) -> None:
+    # GATE 2: an admin / global MANAGE_USER_GROUPS holder may assign in any group;
+    # a scoped manager may only (de)assign managers within a group they manage — so
+    # a manager can delegate within their own group but not beyond it.
+    assert_manages_group(user, db_session, group_id=user_group_id)
```
nmgarza5, line 354: This lets a scoped manager mint new managers with no admin in the loop, and `is_group_manager` is a global flag that hands the promoted user the whole scoped bundle in that group (connectors, doc sets, agents, skills, actions, token limits). Was manager-creates-manager the intent, or should promotion stay global-only?

backend/onyx/server/manage/users.py
```diff
@@ -863,8 +865,13 @@ def _get_token_created_at(
 @router.get("/me/permissions", tags=PUBLIC_API_TAGS)
 def get_current_user_permissions(
     user: User = Depends(require_permission(Permission.BASIC_ACCESS)),
-) -> list[str]:
-    return sorted(p.value for p in get_effective_permissions(user))
+    db_session: Session = Depends(get_session),
+) -> UserPermissionsResponse:
```
nmgarza5, line 869: This flips a `PUBLIC_API_TAGS` endpoint from `list[str]` to an object, which breaks anyone calling it outside the app. Worth adding the new fields alongside the list, or versioning it, rather than replacing the shape.

nmgarza5 (approved):

backend/onyx/auth/permissions.py
Subash-Mohan: fixed get_scoped_groups to gate on expanded bundle.

backend/ee/onyx/db/user_group.py
Subash-Mohan: Intended — managers control resources within their group.

backend/onyx/server/manage/users.py
Subash-Mohan: I don't think so people using it now, because the implementation for permissions are still in feature branch.

PR #13173, opened 2026-07-20
fix(connectors): preserve BookStack timestamp filters

### LINKED ISSUE #11913: BookStack connector silently drops documents updated on the current day (date-only updated_at filter)

shubhamsinnh:
## Description

This PR fixes BookStack connector polling so documents updated during the current day are not missed by incremental indexing.

The root cause was that the connector sent day-level update boundaries to BookStack. BookStack interprets those boundaries at midnight, which can exclude documents updated later on the same day. This change keeps full timestamp precision for polling windows, formats filter bounds as explicit UTC timestamps, and preserves epoch-start windows instead of treating them as absent.

Closes #11913.

## How Has This Been Tested?

- `git diff --check origin/main...HEAD`
- `.venv\Scripts\python.exe -m py_compile backend/onyx/connectors/bookstack/connector.py backend/tests/unit/onyx/connectors/bookstack/test_bookstack_connector.py`
- `.venv\Scripts\python.exe -m pytest backend/tests/unit/onyx/connectors/bookstack/test_bookstack_connector.py` (`2 passed` on Python 3.13.14)
- `.venv\Scripts\python.exe -m ruff check backend/onyx/connectors/bookstack/connector.py backend/tests/unit/onyx/connectors/bookstack/test_bookstack_connector.py`
- `.venv\Scripts\python.exe -m ruff format --check backend/onyx/connectors/bookstack/connector.py backend/tests/unit/onyx/connectors/bookstack/test_bookstack_connector.py`
- `.venv\Scripts\python.exe -m ty check backend/onyx/connectors/bookstack/connector.py backend/tests/unit/onyx/connectors/bookstack/test_bookstack_connector.py`

## Additional Options

### DIFF
--- a/backend/onyx/connectors/bookstack/connector.py
+++ b/backend/onyx/connectors/bookstack/connector.py
@@ -48,6 +48,14 @@ class BookstackConnector(LoadConnector, PollConnector):
         return None
 
     @staticmethod
+    def _format_bookstack_datetime(timestamp: SecondsSinceUnixEpoch) -> str:
+        return (
+            datetime.fromtimestamp(timestamp, tz=timezone.utc)
+            .isoformat(timespec="seconds")
+            .replace("+00:00", "Z")
+        )
+
+    @staticmethod
     def _get_doc_batch(
         batch_size: int,
         bookstack_client: BookStackApiClient,
@@ -63,15 +71,15 @@ class BookstackConnector(LoadConnector, PollConnector):
             "sort": "+id",
         }
 
-        if start:
-            params["filter[updated_at:gte]"] = datetime.fromtimestamp(
-                start, tz=timezone.utc
-            ).strftime("%Y-%m-%d")
+        if start is not None:
+            params["filter[updated_at:gte]"] = (
+                BookstackConnector._format_bookstack_datetime(start)
+            )
 
-        if end:
-            params["filter[updated_at:lte]"] = datetime.fromtimestamp(
-                end, tz=timezone.utc
-            ).strftime("%Y-%m-%d")
+        if end is not None:
+            params["filter[updated_at:lte]"] = (
+                BookstackConnector._format_bookstack_datetime(end)
+            )
 
         batch = bookstack_client.get(endpoint, params=params).get("data", [])
         doc_batch: list[Document | HierarchyNode] = [
--- a/backend/tests/unit/onyx/connectors/bookstack/test_bookstack_connector.py
+++ b/backend/tests/unit/onyx/connectors/bookstack/test_bookstack_connector.py
@@ -1,0 +1,60 @@
+from typing import Any
+from typing import cast
+from unittest.mock import MagicMock
+
+from onyx.connectors.bookstack.client import BookStackApiClient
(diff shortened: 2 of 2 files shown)


shubhamsinnh:
@wenxi-onyx @evan-onyx could one of you help route this connector fix to the right reviewer? It fixes #11913, Cubic passed with no issues, and the change is focused on preserving timestamp precision for BookStack incremental update filters.

PR #13176, opened 2026-07-20
fix(deployment): avoid MCP nginx enable flag

### LINKED ISSUE #11612: MCP_SERVER_ENABLED environment variable missing for nginx in prod deployment
so when you want to use onyx as an MCP server you need to set `MCP_SERVER_ENABLED=true` for a couple of services. One of which is nginx. The current issue is that nginx uses its own .env file `.env.nginx` when you use the prod compose file (https://github.com/onyx-dot-app/onyx/blob/1e989763c8230c00b8aa9303040f5afc6c17172b/deployment/docker_compose/docker-compose.prod.yml#L322) and this one does not contain `MCP_SERVER_ENABLED`

You we should either
- add this variable to the compose file
- add this variable to the .env.nginx
- make nginx use the general .env file like the docker-compose.yml does

shubhamsinnh:
## Description

Fixes #11612.

## Objective

Avoid requiring production Docker Compose nginx services to receive `MCP_SERVER_ENABLED` while still allowing `/mcp` traffic to route correctly when the MCP server is enabled.

## Summary

- Remove nginx's conditional `MCP_SERVER_ENABLED` handling.
- Always generate the MCP nginx location include.
- Remove the separate MCP upstream include and use Docker DNS lazy resolution for the MCP proxy target so nginx can start even when the optional `mcp_server` service is not running.
- Leave production compose files as configuration-light as they were before this PR.

## How Has This Been Tested?

- `git diff --check`
- With temporary dummy production secret values and a temporary ignored `.env.nginx` generated from `env.nginx.template`:
  - `docker compose -f docker-compose.prod.yml config --quiet`
  - `docker compose -f docker-compose.prod-no-letsencrypt.yml config --quiet`
  - `docker compose -f docker-compose.prod-cloud.yml config --quiet`

Note: I attempted to run `nginx -t` in the nginx container, but Docker Desktop is not running in this local environment, so containerized nginx syntax validation could not be executed here.

## Additional Options

### DIFF
--- a/deployment/data/nginx/app.conf.template
+++ b/deployment/data/nginx/app.conf.template
@@ -23,9 +23,6 @@ upstream web_server {
     server ${ONYX_WEB_SERVER_HOST}:3000 fail_timeout=0;
 }
 
-# Conditionally include MCP upstream configuration
-include /etc/nginx/conf.d/mcp_upstream.conf.inc;
-
 # WebSocket support: only set Connection "upgrade" for actual upgrade requests
 map $http_upgrade $connection_upgrade {
     default upgrade;
@@ -39,7 +36,6 @@ server {
 
     access_log /var/log/nginx/access.log custom_main;
 
-    # Conditionally include MCP location configuration
     include /etc/nginx/conf.d/mcp.conf.inc;
 
     # First-party health probe served directly by nginx (no upstreams),
--- a/deployment/data/nginx/app.conf.template.no-letsencrypt
+++ b/deployment/data/nginx/app.conf.template.no-letsencrypt
@@ -23,9 +23,6 @@ upstream web_server {
     server web_server:3000 fail_timeout=0;
 }
 
-# Conditionally include MCP upstream configuration
-include /etc/nginx/conf.d/mcp_upstream.conf.inc;
-
 # WebSocket support: only set Connection "upgrade" for actual upgrade requests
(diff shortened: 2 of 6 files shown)


shubhamsinnh:
@justin-tahara @jmelahman could one of you help route this deployment PR to the right reviewer? It fixes #11612, Cubic passed with no issues, and the change is limited to passing MCP_SERVER_ENABLED into the prod nginx compose services while preserving the default disabled behavior.

jmelahman:
Thanks for the contribution.

I see all three of these services already support a 
```
    env_file:
      - .env.nginx
```
so this issue shouldn't be blocked. Moreover, these docker-compose files have been overloaded with flags and we're trying to reduce their role as configuration files. Additionally, I am not sure why Nginx needs to know if the MCP server is enabled at all, so I would like to look into removing this flag entirely rather than plumb more flags through here.

shubhamsinnh:
Thanks for the direction, @jmelahman. I updated the PR to avoid plumbing MCP_SERVER_ENABLED through the prod compose files. The new version removes nginx's MCP enable-flag branch, always generates the MCP location include, and uses Docker DNS lazy resolution for the MCP proxy target so nginx can still start when the optional mcp_server service is absent. Compose rendering passes for all three prod variants with dummy required prod env values. I noted in the PR body that I could not run containerized nginx -t locally because Docker Desktop is not running here.

PR #13191, opened 2026-07-20
fix(metrics): serve worker /metrics on a dual-stack socket

justin-tahara:
## Description

**Supersedes #13179 — that PR's commit is included here as-is, authored by @alex000kim.** Merging this one lands both changes together; #13179 can be closed rather than merged.

@alex000kim correctly identified that the worker `/metrics` server is unreachable on IPv6-only clusters — Prometheus gets `connection refused`, so every KEDA scaler on `onyx_queue_depth` sits in fallback.

`start_http_server()` defaults to `addr="0.0.0.0"`, which is v4-only. Switching that default to `"::"` fixes IPv6-only clusters, but it makes IPv4 reachability a property of the host rather than of our code:

`prometheus_client` picks an address family via `getaddrinfo` and hands off to `wsgiref.make_server`. It never sets `IPV6_V6ONLY`. So whether an `AF_INET6` listener also answers IPv4 scrapes is inherited from `net.ipv6.bindv6only`. Two consequences, both raised by the review bots on #13180:

- **Where `bindv6only=1`, IPv4 scrapes break.** The v4-mapped behavior is a default, not a guarantee. Verified directly — with `IPV6_V6ONLY=1` an IPv4 connect to a `::`-bound socket is refused; with it cleared, it succeeds.
- **Where the kernel has no `AF_INET6` at all** (built without IPv6, or booted `ipv6.disable=1`), creating the socket fails and metrics stop with no indication that the bind address was the cause. Note this is narrower than it first appears: with `net.ipv6.conf.all.disable_ipv6=1` binding `::` still succeeds *and* still serves IPv4, verified in a container — so the IPv4 candidate is insurance for a rarer case than the sysctl.

The second commit keeps the `::` default and closes both holes:

- Bind `::` with `IPV6_V6ONLY` **explicitly cleared**, so one socket serves both families independent of host sysctl.

  **Why this can't just be `start_http_server(port, addr="::")`:** upstream already does almost everything here — `prometheus_client` 0.23.1 resolves the address family with the same `getaddrinfo(..., AI_PASSIVE)` call, and its `ThreadingWSGIServer` already sets `daemon_threads`, a silent request handler, and a daemon `serve_forever` thread. The *one* thing it does not do is clear `IPV6_V6ONLY`, and `start_wsgi_server` accepts no `server_class` argument, so there is no hook to add it. That single `setsockopt` is the entire reason for the hand-rolled server; everything else is reproducing upstream behavior to get at it. Only public `prometheus_client` API is used (`make_wsgi_app`).

  Ordering matters and is asserted in tests: `setsockopt(IPV6_V6ONLY)` on an already-bound socket fails with `EINVAL`, so it must precede `server_bind`.
- **Fall back to `0.0.0.0` with a warning** when the IPv6 bind fails, so IPv6-disabled hosts keep serving metrics and the reason is visible in logs.
- **`PROMETHEUS_METRICS_BIND_ADDR`** pins the bind address and skips the fallback, as an operator escape hatch (e.g. loopback-only, or forcing IPv4).

Net effect: IPv6-only clusters are fixed, and no currently-working deployment can regress — the dual-stack guarantee no longer depends on the host, and hosts that cannot do IPv6 at all degrade to today's exact behavior.

## How Has This Been Tested?

Unit tests (`backend/tests/unit/server/metrics/`, 167 passed), including new coverage that exercises a **real socket** rather than asserting on call arguments:

- `test_server_bind_clears_v6only_explicitly` — asserts the `setsockopt` against a stub socket, so it holds on hosts whose sysctl already defaults to 0 and would otherwise mask a missing call. Confirmed via negative control: reverting to a plain `::` bind fails this test while the other 12 still pass, i.e. this is the assertion that carries the guarantee.
- `test_serves_both_ipv4_and_ipv6_scrapers` — binds a real dual-stack listener, asserts `IPV6_V6ONLY == 0`, and scrapes `/metrics` over both `127.0.0.1` and `[::1]`.
- `test_falls_back_to_ipv4_when_ipv6_unavailable` / `test_explicit_bind_addr_is_pinned` — cover the fallback and the pin.

Live end-to-end against a real worker-style server:

```
INFO  Prometheus metrics server started on [::]:19098 for primary
scrape  127.0.0.1: HTTP 200
scrape      [::1]: HTTP 200
V6ONLY: 0

# with the IPv6 bind forced to fail:
WARN  Metrics server for primary could not bind [::]:19099 (Address family not
      supported). Falling back to 0.0.0.0:19099.
INFO  Prometheus metrics server started on 0.0.0.0:19099 for primary
fallback family: 2 (AF_INET)
```

`pre-commit` clean (ruff, ruff format, `ty`, env drift baseline).

## Additional Options


justin-tahara:
@greptile can you review the PR?

backend/onyx/server/metrics/metrics_server.py
```diff
@@ -32,8 +40,94 @@
...
+
+    class _Server(_DualStackWSGIServer):
+        address_family = family
+
+    # Bind the sockaddr getaddrinfo produced rather than just its host part, so a
+    # scoped IPv6 address (fe80::1%eth0) keeps the scope id it needs to bind.
+    # This is make_server() inlined; it accepts only a (host, port) pair.
+    # The cast drops getaddrinfo's link-layer sockaddr form, which SOCK_STREAM
+    # over an IP address cannot return.
+    address = cast("tuple[str, int] | tuple[str, int, int, int]", sockaddr)
```
jmelahman, line 118: lmao

jmelahman (approved):

PR #13193, opened 2026-07-20
fix(scripts): make tenant cleanup toolkit safe to run concurrently

justin-tahara:
## Description

The no-bastion tenant cleanup scripts are unsafe at `--concurrency > 1`, which is what
`QUICK_START_NO_BASTION.md` told you to use. Found while running a 100-tenant cleanup batch.

Each per-tenant SQL statement was written to a single fixed path on one pod
(`/tmp/control_plane_query.py`) and then executed as a second step. With multiple threads, one
thread's statement overwrites another's in the gap between write and exec, so a tenant executes
someone else's SQL. Measured against a real control plane at `--concurrency 16`, **15 of 16
status lookups returned a different tenant's row**. These same code paths issue the `DELETE`
statements that remove control plane records.

The blast radius is bounded by the input CSV, since only statements for tenants in that CSV are
ever in flight — but within a batch, deletes land on the wrong tenant and are reported as success.

Fixed by piping the script over stdin rather than staging it on disk. That removes the shared
state by construction and drops a round trip per query. The connector deletion script gets a
unique remote path for the same reason.

Three separate defects made these failures invisible, each of which caused a failed cleanup to be
reported as a successful one:

- **`cleanup_tenant_schema.py` deleted from an unqualified `user_tenant_mapping`.**
  `schema_translate_map` only rewrites SQLAlchemy `Table` constructs, not raw `text()`, so the name
  resolved against whatever `search_path` the pooled backend happened to carry. It intermittently
  failed *after* the schema was already dropped, orphaning the mapping row. Hit ~16% of tenants in
  a real run. Now qualified as `public.user_tenant_mapping`.
- **`cleanup_control_plane` discarded `execute_control_plane_delete`'s return value**, so tenants
  whose control plane rows survived were still written to `cleaned_tenants.csv`. Now propagated.
- **`run_connector_deletion` never captured output**, so its `RuntimeError` carried an empty
  message; and the on-pod script reports failures in its JSON payload while still exiting 0. Now
  captures output and checks the payload.

`cleaned_tenants.csv` opened `"w"`, truncating the record of previously deleted tenants on every
run. It is the only record of what was deleted — and the only way to sweep orphaned search index
chunks later — so it now appends.

Docs updated: the `--concurrency 16` guidance, plus notes that pods are resolved once per run,
that contexts need a namespace set, that results should be verified against the database rather
than the run summary, and that `tenant_data_*.json` contains real user chat message text.

### Not fixed here, flagged for a follow-up

- **Search indices are never cleaned up.** `get_tenant_index_name` is defined and copied to the pod
  but never called. In multi-tenant deployments this is less severe than it looks: tenants share
  indices and are separated by a `tenant_id` field, so dropped tenants leave orphaned chunks that
  can be swept later by selecting on `tenant_id` — provided `cleaned_tenants.csv` survives, which
  is now the case. Documented in QUICK_START.
- **`check_documents_deleted` uses `cc_count == 0 or doc_count <= 5`** while its own error message
  says "and". Tightening it to `and` would make the mark-connectors step strictly mandatory, so
  it's left alone pending a decision.

## How Has This Been Tested?

Against a live multi-tenant deployment.

- **Concurrency regression, before:** 16 concurrent status lookups on the old code path, each
  asking for its own tenant id — 1 correct, 15 returned another tenant's row.
- **Concurrency regression, after:** same probe against the fixed code path — 16 of 16 correct.
- **Read path:** verified the stdin-piped query returns correct results for a plain aggregate.
- **End-to-end:** cleaned a 100-tenant batch and reconciled directly against `pg_namespace`,
  `public.user_tenant_mapping`, and the control plane `tenant` table. Final state 100/100 fully
  removed from both planes, with exactly 0 tenants outside the batch affected.
- The `public.` qualification fix addresses the 17 orphaned mapping rows observed in that batch,
  which had to be swept manually.
- `pre-commit run` passes (`ty`, `ruff`, `ruff format`).

## Additional Options

---
## Summary by cubic
Make the no-bastion tenant cleanup toolkit safe to run with concurrency and stop false “success” reports. Adds `--data-plane-pod` and `--control-plane-pod` to spread parallel batches across pods, and updates docs to stress DB writer limits (watch DiskQueueDepth; pin to spread load, not to go faster; ~30 tenants/min ceiling).

- **Bug Fixes**
  - Remove shared on-pod state: pipe control-plane SQL over stdin; use a unique remote path for connector deletion and delete it after execution.
  - Correctly report failures: propagate control-plane delete failures; capture output and parse the JSON payload from connector deletion (multi-line), raising on reported errors even when exit code is 0.
  - Fix schema cleanup: qualify `public.user_tenant_mapping` to avoid `search_path` issues.
  - Preserve cleanup history safely: append to `cleaned_tenants.csv`, write the header once under an exclusive lock, and skip any literal "tenant_id" row when reading.
  - Validate tenant IDs before interpolating into SQL, rejecting values outside `[A-Za-z0-9_-]`.

<sup>Written for commit 736bb5eb0cd08429e3fa9d50f54a5945cf2cd9fa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13193?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/scripts/tenant_cleanup/QUICK_START_NO_BASTION.md
+++ b/backend/scripts/tenant_cleanup/QUICK_START_NO_BASTION.md
@@ -54,14 +54,28 @@ kubectl get po | grep celery-worker-user-file-processing | grep Running
 ## Important Notes
 
 1. **Step 2 triggers background deletion** - the actual document deletion happens asynchronously via Celery workers
-2. **You MUST wait** between Step 2 and Step 3 for deletion to complete (can take 6+ hours)
+2. **You MUST wait** between Step 2 and Step 3 for deletion to complete (can take 6+ hours).
+   Tenants with no documents drain within minutes, since there is nothing to delete.
 3. **Monitor deletion progress** with: `kubectl logs -f <celery-worker-pod>`
 4. **All scripts verify tenant status** - they'll refuse to process active (non-GATED_ACCESS) tenants
+5. **Pods are resolved once per run.** If that pod restarts mid-run, every remaining tenant fails.
+   Prefer batches of ~1000 over one enormous run.
+6. **Set the namespace on your kubectl contexts.** Pod discovery runs `kubectl get po` with no
+   `-n`, so it only sees the context's default namespace.
+7. **Verify against the database afterwards, not against the summary.** A tenant counted as
+   successful can still leave rows behind if a later step failed. Check `pg_namespace`,
+   `public.user_tenant_mapping`, and the control plane `tenant` table.
+8. **Search indices are not cleaned up.** In multi-tenant deployments all tenants share indices
+   and are separated by a `tenant_id` field, so dropping a schema leaves that tenant's chunks
+   behind. They can be swept later by selecting on `tenant_id` - keep `cleaned_tenants.csv`.
 
 ## Files Generated
 
+- `tenant_data_YYYYMMDD_HHMMSS.json` - Raw per-tenant data. **Contains real user chat message
+  text** (`last_query_text`); keep it out of the repo and off shared storage.
 - `gated_tenants_no_query_3mo_YYYYMMDD_HHMMSS.csv` - List of tenants to clean
-- `cleaned_tenants.csv` - Successfully cleaned tenants with timestamps
+- `cleaned_tenants.csv` - Successfully cleaned tenants with timestamps. Appended across runs, and
+  the only record of what was deleted - needed for any later search-index sweep.
 
 ## Safety First
 
--- a/backend/scripts/tenant_cleanup/QUICK_START_NO_BASTION.md
+++ b/backend/scripts/tenant_cleanup/QUICK_START_NO_BASTION.md
@@ -68,25 +68,41 @@ kubectl get po | grep celery-worker-user-file-processing | grep Running
 8. **Search indices are not cleaned up.** In multi-tenant deployments all tenants share indices
    and are separated by a `tenant_id` field, so dropping a schema leaves that tenant's chunks
    behind. They can be swept later by selecting on `tenant_id` - keep `cleaned_tenants.csv`.
-9. **To go faster, spread across pods - do not just raise `--concurrency`.** Every operation is a
-   `kubectl exec` against the one pod the run picked, so that pod is the bottleneck. Running two
-   batches at once without pinning gains nothing: pod selection is random over a small set, so
-   both runs usually land on the same pod, and it ends up saturated while its replicas idle.
-   Pin distinct pods per run instead:
+9. **The database writer is the real limit. Do not tune for throughput.**
+   Dropping a tenant schema deletes on the order of 500 relations, so cleanup is a sustained
+   burst of catalog writes against the cluster writer. On a live cluster this is the constraint
+   that matters, well before pod CPU is.
 
-   ```bash
-   # terminal 1
-   ... --csv batch_a.csv --concurrency 8 \
-       --data-plane-pod <worker-pod-1> --control-plane-pod <control-pod-1>
-   # terminal 2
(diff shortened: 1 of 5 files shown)


nmgarza5 (approved):

justin-tahara:
Added one more commit (024e56f): `--data-plane-pod` / `--control-plane-pod` to pin the pods a run
uses.

This came out of running the cleanup rather than from reading the code. Every operation is a
`kubectl exec` against the single pod a run selects, so that pod — not the database, and not the
thread count — is the throughput limit. Running two batches concurrently to go faster does nothing,
because pod selection is random over a small set and both runs land on the same pod. Observed
directly: throughput *dropped* from ~34 to ~29 tenants/min with two runs, and one worker sat at
1993m CPU while its replica idled at 23m.

With the pods pinned to different pairs:

| | 1 run, `--concurrency 12` | 2 pinned runs, `--concurrency 8` |
|---|---|---|
| throughput | ~34 tenants/min | **~67 tenants/min** |
| worker A | 1993m CPU | 914m |
| worker B | 23m (idle) | 1244m |
| errors | 0 | 0 |

Roughly double the throughput at a *lower* peak load on any single pod, which matters because these
are live workers also serving user file processing — the previous behaviour was concentrating a
2-core load onto one tenant-facing pod for no benefit.

Docs updated with the flags and the "spread across pods, don't just raise `--concurrency`" guidance,
since the intuitive move is the one that doesn't work.

For context on how much of this is exercised: the three commits here have now cleaned ~1000 tenants
across two planes with zero failures, versus 17 orphaned rows out of the first 100 on the unfixed
toolkit.

justin-tahara:
Correction to the previous comment, and a caveat on merging.

Running the configuration I recommended two comments ago paged an on-call engineer. Writer
`DiskQueueDepth` went from <1 to a sustained 50-64 and tripped a >25 alarm, and the run started
failing on infrastructure timeouts (39 failures: `TLS handshake timeout`, `EOF`, failed execs). It
recovered as soon as the runs were stopped — during: avg 34.9 / max 64.7; after: avg 8.3 / max 26.0.

The mistake was in how I validated it. I measured throughput and pod CPU, saw ~2x with lower
per-pod load, and wrote that up as guidance without ever checking what the load did to the
database. Dropping a tenant schema deletes on the order of 500 relations, so this is a sustained
burst of catalog writes against the cluster writer — the writer is the binding constraint, and pod
CPU was never the interesting number.

| rate | writer `DiskQueueDepth` | failures |
|---|---|---|
| ~34 tenants/min | low, occasional spikes | 0 in 500 tenants |
| ~67 tenants/min | 50-64 sustained, alarm tripped | 39 |

`WriteLatency` held at 2-5 ms the whole time, so it is not a useful early warning — `DiskQueueDepth`
is the signal.

736bb5e reframes pod pinning as a way to avoid saturating a single live worker rather than a way to
go faster, documents the signal to watch, and gives a rate ceiling.

**On merging:** commits 1-2 (the concurrency race, the silent-failure fixes, the review feedback)
are validated by ~1,500 tenants cleaned across both planes and I have no reservations about them.
The failure propagation fix proved itself here — those 39 failures were surfaced instead of being
recorded as successes, which is exactly the old behaviour this PR set out to kill. But I would not
have merged this at 024e56f, because it shipped a runbook that reproduces the incident. Worth a
closer look at commit 3's guidance specifically.

PR #13228, opened 2026-07-21
fix(onboarding): make flow scrollable and responsive at small viewports

nmgarza5:
## Description

At small viewports the onboarding flow overflowed the centered app grid: the welcome header clipped at the top, the LLM provider grid clipped mid-card behind the input bar, and there was no way to scroll.

Fix:
- During onboarding the grid rows become `minmax(min-content, 1fr) minmax(0, max-content) minmax(0, 1fr)`: the welcome header can no longer be squeezed below its content height, the form row compresses and scrolls, and the layout stays vertically centered when the form is short.
- Removed the nested `max-h-[40vh]` scroll on the LLM provider grid; the form scrolls as one unit inside a `ShadowDiv`.
- New `mask` prop on Opal `ShadowDiv` fades the scrolled content itself via `mask-image` instead of painting gradient overlays, so the radial page background shows through (default overlay behavior unchanged for existing consumers).
- Welcome → onboarding → input gaps equalized at 1rem.
- The welcome-row model selector is hidden until an LLM provider is configured (design request).

## How Has This Been Tested?

- Manually on dev1 across window sizes: welcome header stays visible, form scrolls with faded edges, input bar stays pinned, Welcome step stays centered.
- `tsgo --noEmit` clean; `useShowOnboarding` jest suite 16/16; oxlint/oxfmt via pre-commit.

## Additional Options

### DIFF
--- a/web/lib/opal/src/components/shadow-div/components.tsx
+++ b/web/lib/opal/src/components/shadow-div/components.tsx
@@ -30,6 +30,12 @@ interface ShadowDivProps extends React.HTMLAttributes<HTMLDivElement> {
    * Show only top shadow
    */
   topOnly?: boolean;
+
+  /**
+   * Fade the content itself via mask-image instead of painting gradient
+   * overlays. Use over non-flat backgrounds the gradients can't match.
+   */
+  mask?: boolean;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


nmgarza5:
@greptile

Subash-Mohan (approved):

PR #13236, opened 2026-07-21
fix(sidebar): scroll-aware edge fades, hide native scrollbar

nmgarza5:
## Description

Design feedback on the sidebar scroll area: the top was a straight cutoff with no fade, the bottom fade was a static painted gradient that sat above the actual scroll boundary (a 2rem tail spacer pushed content past it), and the native scrollbar rendered with visual artifacts.

Fix, all inside `SidebarLayouts.Body`:
- Compose `ShadowDiv mask` (added in #13228): the list fades under the sticky header only once scrolled, and fades at the true bottom edge only while more content remains. At full scroll no fade renders.
- Delete the always-on painted bottom gradient.
- Hide the native scrollbar on the sidebar scroller (`scrollbar-width: none` + webkit fallback); wheel/trackpad/keyboard scrolling unchanged.
- `ShadowDiv` gains a `containerClassName` prop so consumers can size the wrapper without an extra div.

Applies to all three consumers: app sidebar, admin sidebar, connector-step sidebar.

## How Has This Been Tested?

- Manually on dev1: admin + app sidebars mid-scroll (both fades), at full scroll (no fade, clean gap above the separator), scrollbar gone.
- `tsgo --noEmit` clean; full web jest suite 849/849 after rebasing onto main; oxlint/oxfmt via pre-commit.

## Additional Options

### DIFF
--- a/web/lib/opal/src/components/shadow-div/components.tsx
+++ b/web/lib/opal/src/components/shadow-div/components.tsx
@@ -36,6 +36,11 @@ interface ShadowDivProps extends React.HTMLAttributes<HTMLDivElement> {
    * overlays. Use over non-flat backgrounds the gradients can't match.
    */
   mask?: boolean;
+
+  /**
+   * Class for the outer wrapper (e.g. flex sizing within a parent column).
+   */
+  containerClassName?: string;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


nmgarza5:
@greptile

raunakab (approved):

raunakab (approved):

PR #13248, opened 2026-07-21
feat: OpenAI-compatible Onyx LLM gateway endpoint

rohoswagger:
## Description

Adds an OpenAI-compatible LLM gateway endpoint (`POST /llm-gateway/v1/chat/completions`) to the API server. It resolves admin-configured LLM providers per user (`{provider_id}/{model_name}` model IDs), streams via SSE, and reuses the shared LLM layer (litellm translation, prompt caching, tracing spans via a new `LLMFlow.CRAFT_LLM_GENERATION`).

Access is gated by a new `USE_LLM_GATEWAY` permission that is only reachable via token scopes (today: implied by `CRAFT_SANDBOX`) — sessions and API keys are rejected, and the scope is not user-selectable on PATs. Upstream failures are sanitized (generic in-band SSE error frames with distinguishable `rate_limit_error` / `timeout_error` / `upstream_error` types, always followed by `[DONE]`; non-streaming errors map to OnyxError).

First PR of a stack; the follow-ups route Craft sandboxes through this gateway and update the Craft UI.

## How Has This Been Tested?

- 33 unit tests in `test_onyx_llm_gateway.py` (auth scope gating, model resolution ACLs, SSE payload shapes, tool-call delta merging, error sanitization, stream terminator behavior)
- New `test_processor.py` covering `process_with_prompt_cache(with_metadata=...)`
- Permission-expansion tests in `test_permissions.py`; full `tests/unit/onyx/llm` suite green (343)
- Separate correctness/security/regression reviews: no unauthorized provider/model reachable; permission change strictly narrowing

## Additional Options

---
## Summary by cubic
Adds an OpenAI-compatible LLM gateway at `/gateway/v1/chat/completions` that proxies `{provider_id}/{model_name}` with SSE streaming, prompt-cache skipping, and tracing. Access requires a token scoped with `USE_LLM_GATEWAY` (implied by `CRAFT_SANDBOX`); upstream errors are sanitized and streams always end with `[DONE]`.

- **New Features**
  - Endpoint: `POST /gateway/v1/chat/completions` returns OpenAI-shaped responses, streams via SSE (tool-call deltas forwarded, role sent once), supports structured outputs, records accumulated streamed `reasoning_content` on spans, and includes `usage.prompt_tokens_details.cached_tokens`; skips prompt-hash work via `process_with_prompt_cache(..., with_metadata=False)`.
  - Routing and auth: only visible models via `{provider_id}/{model_name}`; provider resolution uses `fetch_accessible_llm_provider_by_id` with ACL checks (persona-restricted excluded). Requires `USE_LLM_GATEWAY` (sessions/API keys rejected) and `READ_SEARCH`. `tool_choice` supports `auto`/`required`/`none`; `reasoning_effort` defaults to `AUTO`. Streaming emits `rate_limit_error`, `timeout_error`, or `upstream_error` frames, then `[DONE]`; non-streaming errors map to Onyx codes.

- **Refactors**
  - OpenAI wire payloads are typed as frozen pydantic models with explicit `to_wire` JSON serialization to enforce field-presence semantics and avoid nulls for absent fields.

<sup>Written for commit 210960e52762c3023c39b101fe8d78bb0cec06b7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13248?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/onyx/server/features/build/llm_gateway.py
```diff
@@ -0,0 +1,514 @@
...
+    llm_generation_span,
+    record_llm_response,
+    record_llm_span_output,
+)
+from onyx.utils.logger import setup_logger
+from onyx.utils.threadpool_concurrency import start_thread_with_context
+
+logger = setup_logger()
+
+router = APIRouter(prefix=ONYX_GATEWAY_PATH_PREFIX)
```
wenxi-onyx: I think we should not have a prefix and match the spec --> make sure to cross reference implementation with the openai compatible endpoint spec

backend/onyx/server/features/build/llm_gateway.py
```diff
@@ -0,0 +1,514 @@
...
+
+
+def _resolve_model(
+    db_session: Session, user: User, requested_model: str
+) -> tuple[LLMProviderView, ModelConfigurationView]:
+    provider_id_text, separator, model_name = requested_model.partition("/")
+    try:
+        provider_id = int(provider_id_text)
+    except ValueError:
+        provider_id = -1
```
evan-onyx, line 76: ? this seems jank to me and should at least be logged

backend/onyx/server/features/build/llm_gateway.py
```diff
@@ -0,0 +1,514 @@
...
+        raise OnyxError(
+            OnyxErrorCode.INVALID_INPUT, f"Invalid messages: {e.error_count()} errors"
+        ) from e
+
+
+def _parse_reasoning_effort(
+    raw: str | None, supports_reasoning: bool
+) -> ReasoningEffort:
+    if raw is None:
+        return ReasoningEffort.HIGH if supports_reasoning else ReasoningEffort.AUTO
```
evan-onyx: probably should make the default configurable but low prio

backend/onyx/server/features/build/llm_gateway.py
wenxi-onyx, line 1: I think files should be organized like the rest of the codebase to server/gateway/api.py and models.py + a db file if necessary

backend/onyx/server/features/build/llm_gateway.py
```diff
@@ -0,0 +1,514 @@
...
+        return ReasoningEffort.AUTO
+
+
+def _prepare_messages(
+    llm: LLM, messages: list[ChatCompletionMessage]
+) -> list[ChatCompletionMessage]:
+    if not messages:
+        raise OnyxError(OnyxErrorCode.INVALID_INPUT, "messages must not be empty")
+    cacheable_prefix = messages[:-1] or None
+    processed_messages, _ = process_with_prompt_cache(
```
evan-onyx, line 122: btw we turned anthropic caching off a while ago, should turn it back on if it isn't already

backend/onyx/server/features/build/llm_gateway.py
```diff
@@ -0,0 +1,514 @@
...
+
+
+@router.post("/v1/chat/completions")
+def gateway_chat_completions(
+    request: ChatCompletionRequest,
+    http_request: Request,
+    user: User = Depends(require_permission(Permission.READ_SEARCH)),
+    db_session: Session = Depends(get_session),
+) -> Any:
+    _require_sandbox_gateway_access(http_request, user)
```
wenxi-onyx: I think this resolution should go inside `require_permissions`? Also, why is the `require_permissions` value `READ_SEARCH`?

backend/onyx/server/features/build/llm_gateway.py
wenxi-onyx: Do we want this? HIGH can cause very long latency/sessions with certain models.

backend/onyx/server/features/build/llm_gateway.py
rohoswagger: yea not in love with this, reworking

evan-onyx (approved):
here are my thoughts, broadly lgtm

backend/onyx/llm/prompt_cache/processor.py
```diff
@@ -20,6 +20,7 @@ def process_with_prompt_cache(
     cacheable_prefix: LanguageModelInput | None,
     suffix: LanguageModelInput,
     continuation: bool = False,
+    with_metadata: bool = True,
```
wenxi-onyx, line 23: OOC why do we want/need this?

rohoswagger: the metadata path SHA256-hashes the entire cacheable prefix per request to build a cache key the gateway never uses (implicit provider caching only); the flag skips pure CPU waste on agent-sized prompts.

rohoswagger:

backend/onyx/server/features/build/db/build_session.py
```diff
@@ -645,3 +645,29 @@ def fetch_all_supported_build_llm_providers(
             p, user_group_ids, persona=None, is_admin=is_admin
         )
     ]
+
+
+def fetch_accessible_build_llm_provider_by_id(
```
wenxi-onyx: This doesn't seem craft specific? There's probably an existing func that does this in the llm db file. If not, it should be moved there as fetch provider by id.

backend/onyx/server/gateway/__init__.py
wenxi-onyx, line 1: nit: probably don't need an init file?

backend/onyx/server/features/build/db/build_session.py
```diff
@@ -645,3 +645,29 @@ def fetch_all_supported_build_llm_providers(
...
+    db_session: Session, user: User, provider_id: int
+) -> LLMProviderView | None:
+    provider_model = db_session.scalar(
+        select(LLMProviderModel)
+        .where(LLMProviderModel.id == provider_id)
```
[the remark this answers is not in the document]
rohoswagger: valid, but im removing it in a PR merging shortly after this

backend/onyx/server/features/build/llm_gateway.py
rohoswagger: base url is just a /v1/chat/completions endpoint at whatever prefix, thats what the openai compatible spec requires

backend/onyx/server/gateway/api.py
```diff
@@ -0,0 +1,500 @@
...
+logger = setup_logger()
+
+router = APIRouter(prefix=GATEWAY_PATH_PREFIX)
+
+_MESSAGES_ADAPTER: TypeAdapter[list[ChatCompletionMessage]] = TypeAdapter(
+    list[ChatCompletionMessage]
+)
+
+
+def resolve_gateway_model(
```
wenxi-onyx, line 62: no existing utility for this?

backend/onyx/server/gateway/api.py
```diff
@@ -0,0 +1,500 @@
...
+            if model.is_visible and model.name == model_name
+        ),
+        None,
+    )
+    if model is None:
+        raise not_found_error
+    return provider, model
+
+
+def _parse_messages(raw: list[dict[str, Any]]) -> list[ChatCompletionMessage]:
```
wenxi-onyx: This should probably just be wrapped into `_prepare_messages`

backend/onyx/server/features/build/db/build_session.py
rohoswagger: Agreed — moved. It's now `fetch_accessible_llm_provider_by_id` in `onyx/db/llm.py` (72bfdf0), reusing the existing `fetch_existing_llm_provider_by_id` for the query and layering `can_user_access_llm_provider` on top (there was no existing accessible-by-id-for-user variant there). `build_session.py` no longer defines it; the craft gateway consumer imports it from `onyx.db.llm`.

backend/onyx/server/gateway/__init__.py
rohoswagger: Kept it deliberately: this repo uses regular packages, not namespace packages — every package under `onyx/server/` ships an `__init__.py` (`find onyx/server -type d` has full coverage). Omitting just this one would make it the only implicit namespace package in the tree.

backend/onyx/server/gateway/api.py
```diff
@@ -0,0 +1,500 @@
...
+            record_llm_response(span, response)
+
+    return _completion_payload(response, request.model)
+
+
+@router.post("/v1/chat/completions")
+def gateway_chat_completions(
+    request: ChatCompletionRequest,
+    http_request: Request,
+    user: User = Depends(require_permission(Permission.READ_SEARCH)),
```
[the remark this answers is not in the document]
rohoswagger: Not valid in practice, and the proposed fix would break the only real caller:

1. A token scoped directly to `USE_LLM_GATEWAY` cannot exist — `USE_LLM_GATEWAY` is not in `SELECTABLE_PAT_SCOPES` (`onyx/server/pat/models.py`; only `READ_SEARCH`/`READ_CHAT`/`WRITE_CHAT` are user-mintable), and the only system-minted gateway-capable token is the sandbox PAT scoped `[CRAFT_SANDBOX]`, which expands to include `READ_SEARCH` and passes this dependency.
2. Gating on `require_permission(USE_LLM_GATEWAY)` would 403 the sandbox PAT itself: `require_permission` requires `permitted_by_user AND permitted_by_token`, and the PAT authenticates as its owner, whose role permissions (BASIC_ACCESS expansion) do not include `use:llm_gateway`.

So `READ_SEARCH` here is baseline authentication (the weakest permission every gateway credential is guaranteed to carry); the actual gateway gate is the explicit token-scope check in the consumer's `matches`.

backend/onyx/server/features/build/db/build_session.py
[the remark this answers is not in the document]
rohoswagger: Intentional, not a gap. Two parts:

1. The security boundary that matters — per-user provider accessibility (`can_user_access_llm_provider` with the user's groups) — **is** enforced by this lookup (now `fetch_accessible_llm_provider_by_id` in `onyx/db/llm.py`). A sandbox can only reach providers its owning user can already use in normal Onyx chat; the decrypted config never leaves the server (it feeds `llm_from_provider`).
2. `BUILD_MODE_ALLOWED_PROVIDER_TYPES` is a temporary product restriction (the three provider types opencode speaks natively), not an ACL — and the next PR in this stack (#13249) deletes it entirely: the whole point of the gateway is to front *every* accessible provider (bedrock, azure, vertex, ...) through one OpenAI-compatible endpoint. Adding the predicate here would be reverted one PR up.

backend/onyx/server/gateway/api.py
```diff
@@ -0,0 +1,500 @@
...
+                close = getattr(upstream, "close", None)
+                if callable(close):
+                    close()
+            except Exception:
+                logger.exception(
+                    "LLM gateway stream cleanup failed for model %s", model
+                )
+            try:
+                if span is not None:
+                    record_llm_span_output(
```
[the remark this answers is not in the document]
rohoswagger: Valid — fixed in a0b44bc. `_stream_worker` now accumulates `chunk.choice.delta.reasoning_content` and passes `reasoning="".join(...) or None` to `record_llm_span_output`, so streamed `CRAFT_LLM_GENERATION` spans carry the model's reasoning like the non-streaming path does. Covered by `test_stream_records_accumulated_reasoning_on_span`.

backend/onyx/server/gateway/api.py
rohoswagger: No — the `<provider_id>/<model_name>` composite identifier is invented by the gateway itself (it's what we register as the model id in the sandbox's opencode config, since OpenAI-compat model names are a single string and model names alone collide across providers). Nothing existing parses that format. The accessibility half of this function now reuses the shared `onyx/db/llm.fetch_accessible_llm_provider_by_id` (moved per the other thread), so the only bespoke logic left here is the id split + visible-model match.

backend/onyx/server/gateway/api.py
rohoswagger: Done in a0d3ab4 — `_parse_messages` is gone; `_prepare_messages(llm, raw_messages)` now validates and prepares in one step.

backend/onyx/server/features/build/craft_gateway.py
```diff
@@ -0,0 +1,19 @@
+from fastapi import Request
+
+from onyx.auth.permissions import resolve_effective_permissions
+from onyx.db.enums import Permission
+from onyx.db.models import User
+from onyx.server.features.build.utils import is_craft_enabled_for_user
+
+
+def is_craft_gateway_request(request: Request, user: User) -> bool:
```
wenxi-onyx, line 9: If just one func, maybe wrap into caller?

backend/onyx/server/gateway/api.py
wenxi-onyx: I'm not confident this is true - I swear I've touched code that processes `provider/model` strings before, but maybe that has gone away

backend/onyx/server/gateway/api.py
```diff
@@ -0,0 +1,510 @@
...
+_MESSAGES_ADAPTER: TypeAdapter[list[ChatCompletionMessage]] = TypeAdapter(
+    list[ChatCompletionMessage]
+)
+
+
+def resolve_gateway_model(
+    db_session: Session,
+    user: User,
+    requested_model: str,
+    fetch_provider: ProviderFetcher,
```
wenxi-onyx: Why pass a callable here? i.e. what is the case where we would pass a different func than "get provider and model from user if accessible"

backend/onyx/server/gateway/api.py
```diff
@@ -0,0 +1,510 @@
...
+            if model.is_visible and model.name == model_name
+        ),
+        None,
+    )
+    if model is None:
+        raise not_found_error
+    return provider, model
+
+
+def _parse_reasoning_effort(raw: str | None) -> ReasoningEffort:
```
wenxi-onyx, line 101: nit: could be good to publish this func with the reasoning effort enum for other code that touches these values, but not in scope

backend/onyx/server/gateway/api.py
```diff
@@ -0,0 +1,510 @@
...
+
+def _created_epoch(created: str) -> int:
+    try:
+        return int(float(created))
+    except (TypeError, ValueError):
+        return 0
+
+
+def _usage_payload(usage: Usage) -> dict[str, Any]:
+    return {
```
wenxi-onyx: `model_dump`?

backend/onyx/server/gateway/api.py
```diff
@@ -0,0 +1,510 @@
...
+        "prompt_tokens": usage.prompt_tokens,
+        "completion_tokens": usage.completion_tokens,
+        "total_tokens": usage.total_tokens,
+        "prompt_tokens_details": {"cached_tokens": usage.cache_read_input_tokens},
+    }
+
+
+def _completion_payload(response: ModelResponse, model: str) -> dict[str, Any]:
```
wenxi-onyx: I'd be surprised if this shouldn't just be a `model_dump`

backend/onyx/server/gateway/api.py
```diff
@@ -0,0 +1,510 @@
...
+                "finish_reason": response.choice.finish_reason,
+            }
+        ],
+    }
+    if response.usage is not None:
+        payload["usage"] = _usage_payload(response.usage)
+    return payload
+
+
+def _chunk_payload(
```
wenxi-onyx: Same comment as above

backend/onyx/server/gateway/api.py
wenxi-onyx: This would also fix the return type I think? I.e. there should be some BaseModel return type rather than just dict{str Any} --> BaseModel serialize automatically I believe

wenxi-onyx (changes requested):

backend/onyx/server/gateway/api.py
rohoswagger: You're remembering right — `onyx/llm/model_name_parser.py` parses `provider/model` strings. But those are litellm-style provider-**type** prefixes (`bedrock/anthropic.claude...`) used for capability/cost lookups, and its helpers strip or extract the type segment. The gateway's identifier is `<provider_row_id>/<model_name>` — a numeric DB row id for routing — so running `"23/anthropic/claude-3.5-sonnet"` through those helpers would treat `23` as a provider type and misparse. Same-looking string, different semantics; sharing the parser would couple two unrelated formats.

backend/onyx/server/features/build/craft_gateway.py
rohoswagger: Kept it separate deliberately: the check's dependencies (`is_craft_enabled_for_user`) are craft feature code, and `gateway/api.py` is the consumer-agnostic module — inlining it there moves craft policy into the generic file rather than removing anything (the import just shifts from `build.craft_gateway` to `build.utils`). `build/craft_gateway.py` is where craft's gateway-facing policy lives; it grows if craft ever needs more (e.g. a craft-specific default when the effort picker lands). Happy to inline if you feel strongly — it's 20 lines either way.

backend/onyx/server/gateway/api.py
rohoswagger: Agreed — there is no such case anymore (the parameterization was a leftover from an earlier design with per-consumer provider scoping). Removed in 7ebf5af: `resolve_gateway_model` now calls `fetch_accessible_llm_provider_by_id` directly; tests patch it at the module level.

backend/onyx/server/gateway/api.py
rohoswagger: Agreed, and agreed it's out of scope here. Note the fallback semantics are gateway-specific (unrecognized/absent → `AUTO` because the body is untrusted client input), so a generalized version would live next to `ReasoningEffort` in `onyx/llm/models.py` with the caller choosing the fallback — good follow-up when a second consumer of these string values appears.

backend/onyx/server/gateway/api.py
rohoswagger: Deliberately not — `Usage.model_dump()` emits our internal shape (`cache_creation_input_tokens` / `cache_read_input_tokens` as flat fields), while the OpenAI wire shape needs the `prompt_tokens_details: {cached_tokens: ...}` nesting and omits the anthropic-specific fields. This function is a reshape between the two schemas, not a serialization.

backend/onyx/server/gateway/api.py
rohoswagger: Considered it, but the OpenAI wire format has field-*presence* semantics that a single `model_dump` policy can't express: `content` must be present even when `null` (tool-call-only messages), while `reasoning_content` / `tool_calls` / `usage` must be *absent* when empty. `exclude_none=True` drops the required `content: null`; `exclude_none=False` emits `reasoning_content: null` etc., which diverges from what OpenAI-compat clients are built against. Typed response models would need per-field serializer overrides to reproduce this — more machinery than the explicit dicts, which are the honest encoding of the wire contract. There's also no FastAPI-serialization win to collect: the streaming path `json.dumps`es these into SSE frames itself, and the repo convention is no `response_model` on endpoints.

(Same rationale for the chunk payload thread below.)

backend/onyx/server/gateway/api.py
rohoswagger: Same rationale as the `_completion_payload` thread above: OpenAI chunk deltas have presence semantics (role only on the first chunk, keys absent-when-empty, `finish_reason` present-even-null) that `model_dump` can't reproduce without per-field overrides, and chunks are `json.dumps`ed into SSE frames rather than FastAPI-serialized.

backend/onyx/server/features/build/craft_gateway.py
wenxi-onyx: I don't think this reasoning is valid, but w/e, can leave as is

wenxi-onyx (approved):

PR #13249, opened 2026-07-21
feat: route Craft sandboxes through the Onyx LLM gateway

rohoswagger:
## Description

Routes Craft sandboxes through the Onyx LLM gateway introduced in the previous PR, replacing per-provider API-key injection:

- opencode gets a single "onyx" provider (`@ai-sdk/openai-compatible`) whose catalog lists every admin-configured provider/model the user can access; per-turn `reconcile_session_llm_config` regenerates the config so provider changes apply without sandbox recreation
- Removes provider API keys from sandboxes entirely (deletes `LLMProviderKeyResolver` and the not-configured-key sentinel; the egress proxy now only injects the sandbox PAT)
- `SANDBOX_API_SERVER_URL` becomes a full base URL (path prefix included) and is probed at provision time — a missing `/api` prefix now fails fast with the corrected URL in the error
- Deployment updates (compose, env template, helm) for the one-URL contract

## How Has This Been Tested?

- Unit suites for build/sandbox/session/proxy: 1185+ passed (only known environmental failures: event-bus per-directory DB-engine tests, helm-dependent pod-spec tests)
- Tree-equivalence check against the original combined branch (`litellm-craft`) — this stack reproduces it exactly modulo the `/llm-gateway` path rename
- Live-tested end-to-end on k3d and the craft-dev EKS cluster on the combined branch (streaming, tool calls, reasoning effort, model switching)

## Additional Options

---

**Stack:**
1. [craft-llm-gateway #13248](https://github.com/onyx-dot-app/onyx/pull/13248)

---
## Summary by cubic
Routes all Craft sandbox LLM calls through the Onyx LLM gateway as a single `onyx` provider with a per-user model catalog. Improves per-turn config reconciliation and removes provider keys from sandboxes and the proxy.

- **New Features**
  - Single `onyx` provider via `@ai-sdk/openai-compatible`; per-message overrides use `provider_id` + `model` (legacy `provider` + `model` still works). Typed `AgentSelection`; access scoped by `fetch_all_accessible_llm_providers`.
  - Default model now comes from `recommended-models.json` (provider defaults), not a hardcoded list.
  - Security/proxy: removed provider API keys and the resolver; egress proxy injects only the sandbox PAT; gateway access policy is inlined.
  - Config/hot-reload: per-session `opencode.json` overlays the gateway catalog + MCP servers and hot-reloads on dispose; set the dispose-pending marker before writes and retry missed disposes; defensively regenerate on transient read errors; render a byte-stable, sorted catalog; fail a turn if the session is deleted mid-run; per-model output limits derive from the litellm map; skip provider-key decrypt/audit when building the catalog; deduped display-name/provider-label helpers and reuse `_session_llm_config` in reconcile.
  - API base detection: probe `SANDBOX_API_SERVER_URL` once per process and latch on fail-open paths. Applied to Docker `ONYX_SERVER_URL`, per-pod K8s override, and the gateway `api_base`. Compose adds an `onyx-craft-api` alias.
  - Provider config simplified: renamed to `CraftLLMProviderConfig` and the provider block always uses `@ai-sdk/openai-compatible` (removed native-provider branch and unused options).

- **Migration**
  - Set `SANDBOX_API_SERVER_URL` to the full API base used by sandboxes (include `/api` when routing through a public proxy). The resolver auto-corrects when safe and caches probes per process.
  - Remove sandbox/provider LLM key secrets; only the sandbox PAT is used.
  - If you override Docker/Helm envs, include `ONYX_API_PREFIX`.
  - Prefer per-message `provider_id` + `model`; legacy `provider` + `model` remains accepted.
  - Session create no longer supports `llm_provider_type`/`llm_model_name`; selection comes from the gateway default and per-message overrides.

<sup>Written for commit fd0f643d47987802f7289be14065ae68c3e21bfc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13249?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/onyx/server/features/build/session/llm_config.py
```diff
@@ -1,127 +1,182 @@
...
+
+    api_root = SANDBOX_API_SERVER_URL.rstrip("/")
+    api_base = f"{api_root}{ONYX_GATEWAY_PATH_PREFIX}/v1"
+    default_provider_id, default_model_name = default_selection
+    return LLMProviderConfig(
+        provider=ONYX_GATEWAY_PROVIDER_ID,
+        model_name=f"{default_provider_id}/{default_model_name}",
+        # The egress proxy overwrites auth headers for the API host with the
+        # sandbox PAT; the key here is never used.
+        api_key=SANDBOX_PROXY_INJECTED_PLACEHOLDER,
```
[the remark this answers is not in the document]
rohoswagger: Not a reachable state: there is no proxyless posture. `DockerSandboxManager._initialize` raises when `SANDBOX_PROXY_HOST` is unset (`docker_sandbox_manager.py:649-654`, "the proxy is mandatory whenever craft is enabled"), and `KubernetesSandboxManager` likewise fails provisioning without it (`kubernetes_sandbox_manager.py:1037-1039`). Since the manager singleton cannot be constructed proxyless, provisioning never reaches the gateway provider config with the placeholder unswapped. The `else onyx_pat` branch in `provision()` predates this PR and is a leftover from before the proxy became mandatory — it is dead code, not evidence of a supported proxyless mode.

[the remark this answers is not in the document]
rohoswagger: The proxyless Docker posture this assumes does not exist: `DockerSandboxManager._initialize` raises `RuntimeError` when `SANDBOX_PROXY_HOST` is unset (`docker_sandbox_manager.py:649-654`), so no sandbox can be provisioned — and no `opencode.json` written — without the egress proxy. The K8s manager has the same guard (`kubernetes_sandbox_manager.py:1037`). The conditional `ONYX_PAT` assignment you reference is pre-existing dead-else code from before the proxy became mandatory (mirroring #11604), not a supported configuration; the placeholder `api_key` is therefore always swapped by the proxy in every reachable deployment.

backend/onyx/server/features/build/sandbox/util/api_url_check.py
```diff
@@ -0,0 +1,78 @@
...
+
+def _probe_health(base_url: str) -> int | None:
+    """Status of GET {base_url}/health, or None when unreachable."""
+    try:
+        return httpx.get(
+            f"{base_url}/health",
+            timeout=_PROBE_TIMEOUT_SECONDS,
+            follow_redirects=True,
+        ).status_code
+    except httpx.HTTPError:
```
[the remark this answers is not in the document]
rohoswagger: Fixed in 6766458. The probe no longer trusts the status code alone: `_is_health_response` now requires the API health payload (`200` + JSON with `success: true`, matching `StatusResponse` from `/health`). A redirect to a login/web page that answers 200 is now classified like a 404 — the check probes `{base}/api/health` and raises with the corrected URL when that serves the real payload. `follow_redirects=True` is kept deliberately so a working deployment behind an http→https redirect still validates; ambiguous responses (401/403/5xx, or non-health 200 on both paths) still pass, preserving the never-block-a-working-deployment contract. Covered by `test_redirected_web_page_does_not_mask_missing_prefix` and `test_spa_catch_all_on_both_paths_warns_without_raising` in `test_api_url_check.py`.

backend/onyx/server/features/build/sandbox/util/api_url_check.py
```diff
@@ -0,0 +1,78 @@
...
+_validated_lock = threading.Lock()
+
+
+def _probe_health(base_url: str) -> int | None:
+    """Status of GET {base_url}/health, or None when unreachable."""
+    try:
+        return httpx.get(
+            f"{base_url}/health",
+            timeout=_PROBE_TIMEOUT_SECONDS,
+            follow_redirects=True,
```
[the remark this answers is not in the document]
rohoswagger: Fixed in 6766458, though not with `follow_redirects=False` — that would introduce a false positive: a working deployment reached over http behind an http→https redirect would 3xx on both the bare and prefixed probes, and since the prefixed probe "responds", the old raise condition would block a correct URL. Instead the probe now verifies the response body: `_is_health_response` requires `200` + JSON `success: true` (the `StatusResponse` payload of `/health`). A redirect landing on a login/web page fails that check and is treated like a 404, so the `/api`-prefixed probe runs and raises with the corrected URL only when it serves the real health payload. New tests: `test_redirected_web_page_does_not_mask_missing_prefix`, `test_spa_catch_all_on_both_paths_warns_without_raising`.

backend/onyx/server/features/build/session/llm_config.py
```diff
@@ -1,127 +1,182 @@
...
+                    for provider in providers
+                    if provider.provider == requested_provider_type
+                ),
+                None,
+            )
+        )
+        if requested_provider and any(
+            model.is_visible and model.name == requested_model_name
+            for model in requested_provider.model_configurations
+        ):
```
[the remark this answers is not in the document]
rohoswagger: This is intentional graceful degradation rather than a silent divergence. The fallback is not hidden: `reconcile_session_llm_config` persists the resolved selection back to `session.agent_provider`/`session.agent_model` (`manager.py:284-289, 307-309`), so the session state the client reads reflects the model actually used from that turn on, and the fallback path logs a warning. The alternative — rejecting the turn — would hard-break every existing session whose model an admin hides or removes, with no self-service recovery for the user, which is worse than continuing on the admin-sanctioned recommended/visible model. Admin hides a model in the window between client selection and turn reconciliation is exactly the case where deferring to the admin's new visibility policy is the intended behavior.

backend/onyx/server/features/build/db/build_session.py
```diff
@@ -625,17 +624,15 @@ def clear_nextjs_ports_for_user(db_session: Session, user_id: UUID) -> int:
     return result
 
 
-def fetch_all_supported_build_llm_providers(
+def fetch_all_accessible_build_llm_providers(
```
wenxi-onyx: If this doesn't already exist, maybe should move into llm db file and renamed?

backend/onyx/server/features/build/sandbox/util/opencode_config.py
```diff
@@ -21,6 +24,12 @@
     {"claude-opus-4-7", "claude-opus-4-8", "claude-sonnet-4-6"}
 )
 
+# Model configurations don't track max output tokens, so gateway model entries
+# advertise this fixed budget to opencode. 128k matches the recommended Craft
+# models (Claude Fable/Opus 4.8, GPT-5.6) per models.dev; providers enforce
+# their own real caps on models with smaller limits.
+_GATEWAY_DEFAULT_MAX_OUTPUT_TOKENS = 128_000
```
wenxi-onyx, line 25: Might be good to default to the litellm map's value? At some point, we did this in the main app. If the user has set a hard output limit on custom provider/models, maybe we should honor that?

backend/onyx/server/features/build/sandbox/util/opencode_config.py
```diff
@@ -183,55 +181,104 @@ def _build_provider_block(
     return block
 
 
-def build_multi_provider_opencode_config(
-    providers: list[LLMProviderConfig],
-    default_provider: str,
-    default_model: str,
+def _build_custom_provider_block(
+    provider_config: LLMProviderConfig,
+) -> dict[str, Any]:
```
wenxi-onyx, line 155: At some point, we might want to type the opencode config blocks with dataclasses/datamodels. Maybe now?

backend/onyx/server/features/build/sandbox/util/opencode_config.py
```diff
@@ -183,55 +181,104 @@ def _build_provider_block(
...
+        config["plugin"] = list(plugins)
+    return config
 
-    Craft MCP servers are NOT emitted here — they live in per-session config
-    (``build_session_mcp_config``) so they can hot-reload without a pod restart.
 
-    Raises:
-        ValueError: If ``providers`` is empty or ``default_provider`` is
-            not in ``providers``.
+def build_provider_opencode_config(
```
wenxi-onyx, line 202: Note that @Danelegend is using a different definition of `provider` and we should probably separate two concepts. LLM provider vs. credentials provider

backend/onyx/server/features/build/sandbox/util/opencode_config.py
```diff
@@ -183,55 +181,104 @@ def _build_provider_block(
...
     return config
+
+
+def build_session_opencode_config(
+    provider_config: LLMProviderConfig,
+    disabled_tools: list[str],
+    mcp_servers: Sequence[CraftMCPServerConfig] = (),
+    session_id: str | None = None,
+) -> str:
+    return json.dumps(
```
wenxi-onyx: nit: pref in-lining when all this does is json.dumps another func

backend/onyx/server/features/build/session/llm_config.py
```diff
@@ -1,127 +1,178 @@
...
-) -> LLMProviderConfig:
-    return LLMProviderConfig(
-        provider=provider.provider,
-        model_name=model_name,
-        api_key=provider.api_key,
-        api_base=provider.api_base,
+def normalize_agent_selection(
+    provider_id: int | None, model_name: str
+) -> tuple[str, str]:
+    if provider_id is None:
```
wenxi-onyx: If `provider_id` none is not valid, ideally, the arg shouldn't allow `| None`

backend/onyx/server/features/build/session/manager.py
```diff
@@ -249,6 +257,118 @@ def _prewarm_opencode_session(
...
+        sandbox: Sandbox,
+        session: BuildSession,
+        user: User,
+    ) -> None:
+        requested_provider_id: int | None = None
+        requested_provider_type: str | None = None
+        requested_model_name: str | None = None
+        if session.agent_model:
+            if session.agent_provider == ONYX_GATEWAY_PROVIDER_ID:
+                parsed = parse_gateway_model_id(session.agent_model)
```
wenxi-onyx: Should double check how/what we are storing for this value. It seems like we do some dance of parsing, extracting provider id, then recompiling it into a single string that is unnecessary? See `normalize_agent_selection`

rohoswagger:

backend/onyx/server/features/build/db/build_session.py
```diff
@@ -625,17 +624,15 @@ def clear_nextjs_ports_for_user(db_session: Session, user_id: UUID) -> int:
...
-    """Every provider of a Craft-supported type (anthropic, openai, openrouter)
-    that the ``user`` can access. Respects is_public / group restrictions so a
-    user never gets a sandbox keyed with a provider they can't use. Providers
-    are ordered by ID so provisioning and proxy credential selection agree on
-    the first provider of each type."""
+    """Every provider the ``user`` can access (is_public / group rules).
+    persona=None below: Craft has no persona context, so a provider restricted
+    to specific personas is intentionally excluded even when otherwise
+    public."""
     provider_models = db_session.scalars(
```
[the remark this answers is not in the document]
rohoswagger: Valid for #13249 in isolation, but resolved by the stacked frontend PR #13250 (same feature; #13250 is based on this branch). The legacy 3-provider gate is gone there: `hasSupportedCraftProvider` (web/src/app/craft/onboarding/constants.ts:65-73) now returns true when *any* configured provider exposes a visible model — no anthropic/openai/openrouter restriction. `BuildWelcome` no longer gates on the legacy list; it gates on `hasAnyProvider` (BuildWelcome.tsx:45,92,108), which is that same any-visible-model check. `COMMON_PROVIDERS` is retained only as display ordering ("Craft routes every provider through the gateway, so this is display ordering only, not a gate", constants.ts:35-40). So a Bedrock/Azure/custom-only org with a visible model gets the picker + prompt enabled. No change needed on this PR — the frontend belongs to #13250.

backend/onyx/server/features/build/session/messages.py
```diff
@@ -146,7 +147,11 @@ def send_message(
         check_token_rate_limits(user)
 
         turn_index = count_user_messages(session_id, db_session)
-        if request.provider and request.model:
+        if request.provider_id is not None and request.model:
+            session.agent_provider, session.agent_model = normalize_agent_selection(
+                request.provider_id, request.model
+            )
```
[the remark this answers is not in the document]
rohoswagger: Valid, and resolved by the stacked frontend PR #13250: the client now sends the selected provider row id — `provider_id: model.providerId` alongside `provider`/`model` (web/src/app/craft/services/apiServices.ts:382-384) — so `normalize_agent_selection` takes the provider_id path and disambiguates same-type rows instead of falling back to first-match-by-type. Worth noting the type-based "first matching row" resolution is pre-existing Craft behavior; this PR *adds* provider_id as the disambiguation fix and #13250 wires the client to use it, so #13249 introduces no regression on its own. No change needed on this PR.

backend/onyx/server/features/build/session/llm_config.py
```diff
@@ -1,127 +1,176 @@
...
+        "gpt-5.5",
+        "claude-fable-5",
+        "claude-opus-4-8",
+        "moonshotai/kimi-k3",
+        "z-ai/glm-5.2",
+    }
+)
+
+
+def _visible_models_by_name(provider: LLMProviderView) -> list[str]:
```
wenxi-onyx, line 25: Looks like `build_onyx_gateway_config` already sorts?

backend/onyx/server/features/build/session/manager.py
```diff
@@ -249,6 +257,117 @@ def _prewarm_opencode_session(
...
+            # Transient exec/API failure while merely checking; assume stale
+            # and regenerate defensively rather than failing the turn.
+            logger.warning(
+                "Could not read opencode.json for session %s; regenerating",
+                session.id,
+            )
+            current = None
+
+        cache = get_cache_backend()
+        dispose_pending_key = f"craft:llm_config_dispose_pending:{session.id}"
```
wenxi-onyx, line 299: Is this necessary?

deployment/docker_compose/env.template
```diff
@@ -11,6 +11,9 @@
 ## sandbox image for normal deployments.
 IMAGE_TAG=latest
 
+# HOST_PORT=5000
```
wenxi-onyx: Is this intended? Why 5000?

backend/onyx/server/features/build/db/build_session.py
rohoswagger: Done in df044e62ce. Moved `fetch_all_accessible_build_llm_providers` to `onyx/db/llm.py` as `fetch_all_accessible_llm_providers` — it is a generic access-scoped LLM-provider query (is_public/group rules, persona=None) and now sits with the sibling access helpers (`fetch_first_accessible_llm_provider_by_type`, `fetch_accessible_llm_provider_by_id`); dropped the build-specific name. No new imports needed there (it already had `can_user_access_llm_provider`, `fetch_user_group_ids`, etc.). Updated the caller (`manager.build_llm_configs`) and both tests (`test_build_llm_provider_access`, `test_session_gateway_config` patch target); ruff pruned the now-unused imports in build_session.py. ty green.

backend/onyx/server/features/build/session/llm_config.py
rohoswagger: Done in df044e62ce. `normalize_agent_selection(provider_id: int, model_name: str)` — dropped the `| None` and the internal `raise`. The sole caller (messages.py) already guards `if request.provider_id is not None and request.model:` before calling, so the None branch was unreachable; the type now reflects the real contract and ty narrows the caller correctly. Removed the obsolete `test_normalize_agent_selection_requires_provider_id_for_gateway` (it asserted the removed raise).

deployment/docker_compose/env.template
rohoswagger: Good catch — not intended for this PR. The `HOST_IP`/`HOST_PORT` compose binding (and the `# HOST_PORT=5000` example) are unrelated to the LLM gateway; they rode along in the gateway commit. Reverted both `env.template` and `docker-compose.yml` to main in c03fb1d9af, so this PR no longer touches host-port binding.

backend/onyx/server/features/build/sandbox/util/opencode_config.py
rohoswagger: Done in 875a9c0516. Gateway model entries now derive `output` from the litellm map via `get_llm_max_output_tokens(get_model_map(), model.name, provider.provider)` (same helper the main app uses), computed once per catalog build and carried on `GatewayModelConfig.max_output_tokens`; `_build_custom_provider_block` uses it. The 128k constant stays only as the fallback for models the map has no entry for. On the "honor a user-set hard output limit" half: there is no such field — neither `ModelConfigurationView` nor the DB `ModelConfiguration` stores a max_output_tokens (only max_input), and the main app doesn't have per-model user output caps either — so there was nothing to honor without a schema addition, which didn't seem warranted. `get_model_map` is `@lru_cache(maxsize=1)`, so the per-turn reconcile cost is a dict lookup, not a deepcopy.

backend/onyx/server/features/build/session/manager.py
rohoswagger: Refactored in 875a9c0516 — though keeping the parse, which is load-bearing: it feeds the per-turn re-validation of the stored pick against currently-accessible providers (a removed/hidden provider must fall back, see `_select_gateway_default` + `test_gateway_config_falls_back_when_requested_selection_is_stale`), so the stored id/model is never trusted as-is. What was genuinely awkward — and what this fixes — is that the overloaded storage was leaking out as three loose `requested_provider_id/_type/_model_name` params threaded through `build_llm_configs` → `build_onyx_gateway_config` → `_select_gateway_default`, decoded inline in *two* places (`_session_llm_config` and `reconcile_session_llm_config`). Now the columns decode once into a typed `AgentSelection` (`GatewaySelection | LegacySelection`) via `parse_agent_selection`; that value flows through selection/reconcile; and the `provider_id/model_name` wire id is rendered only at the gateway-config boundary (`GatewaySelection.to_columns`/`wire_id`). No schema change — the 2-column storage stays. Prevents invalid provider_type+provider_id combos and removes the duplicated decode.

backend/onyx/server/features/build/sandbox/util/opencode_config.py
rohoswagger: this is used in like 5 places so i think its worth keeping

backend/onyx/server/features/build/sandbox/util/opencode_config.py
rohoswagger: i think not in scope for this pr but yea can do it at some point

backend/onyx/server/features/build/sandbox/util/opencode_config.py
wenxi-onyx: Since the function is effectively 1 line, there's no savings over just doing the json.dumps in-line in the 5 places. Alternatively, `build_provider_opencode_config` should just return a sring?

backend/onyx/server/features/build/session/llm_config.py
rohoswagger: Not redundant — the two sorts feed different outputs. `_visible_models_by_name` (→ list[str]) makes `_select_gateway_default`'s *default pick* deterministic (recommended model, then first-visible fallback); `build_onyx_gateway_config`'s own sort (→ (provider, model) tuples) makes the *rendered catalog* byte-stable for the per-turn byte-compare. Dropping the former would make the fallback default non-deterministic across DB reads. Leaving both.

backend/onyx/server/features/build/session/manager.py
rohoswagger: Kept, but fixed a real ordering gap it had (bfe6296b8e). The marker is load-bearing: after a write-succeeded-but-dispose-failed crash, the file matches on the next reconcile, so without the marker every later reconcile skips the dispose and the running opencode instance serves stale config until pod death. The bug: the marker was set *after* regenerate_session_config, leaving that exact write→dispose window uncovered. Moved `cache.set(dispose_pending_key, ...)` to *before* the write so the match path retries the missed dispose.

backend/onyx/db/llm.py
```diff
@@ -564,6 +564,33 @@ def fetch_first_accessible_llm_provider_by_type(
...
+        .options(
+            selectinload(LLMProviderModel.model_configurations),
+            selectinload(LLMProviderModel.groups),
+            selectinload(LLMProviderModel.personas),
+        )
+    )
+    user_group_ids = fetch_user_group_ids(db_session, user)
+    is_admin = user.role == UserRole.ADMIN
+    return [
+        LLMProviderView.from_model(p)
```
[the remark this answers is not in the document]
rohoswagger: Valid — fixed in fca22f3f9c. `LLMProviderView.from_model` was decrypting the key and emitting a credential-access audit per provider, and this catalog runs every reconcile (per turn) but never uses the key (the gateway config sets the proxy placeholder; the proxy injects the real key for the selected model). Added `include_api_key` (default True, so all other callers are unchanged) and pass `include_api_key=False` from `fetch_all_accessible_llm_providers`, so the per-turn path skips the decrypt + audit for every accessible provider.

backend/onyx/server/features/build/sandbox/util/opencode_config.py
rohoswagger: Renamed the builder param `provider_config` → `llm_provider_config` in opencode_config.py (da564c9c4f) so it clearly reads as the LLM provider (the type is already `LLMProviderConfig`). The broader LLM-provider-vs-credentials-provider vocabulary alignment with @Danelegend spans other work, so I kept this PR to the local clarity rename.

backend/onyx/server/features/build/sandbox/util/opencode_config.py
```diff
@@ -130,121 +141,162 @@ def _build_permissions(
...
+        "provider": {
+            llm_provider_config.provider: _build_provider_block(llm_provider_config)
+        },
+        "enabled_providers": [llm_provider_config.provider],
+        "permission": _build_permissions(disabled_tools, dev_mode, mcp_servers),
     }
     if plugins:
         config["plugin"] = list(plugins)
+    if mcp_servers:
+        assert session_id is not None  # guarded above
```
wenxi-onyx: No asserts in app code. Should just check this explicitly and log explicitly as well

backend/onyx/server/features/build/sandbox/models.py
```diff
@@ -11,16 +11,22 @@
...
+    id: str
+    display_name: str
+    supports_reasoning: bool = False
+    max_input_tokens: int | None = None
+    max_output_tokens: int | None = None
 
-    Passed to SandboxManager.provision() to configure the LLM.
-    """
 
+class LLMProviderConfig(BaseModel):
```
wenxi-onyx: sorry for the churn here, but we might want to make this very clear that it's `CraftLLMProviderConfig` and not possibly the model for the main app.

backend/onyx/server/features/build/sandbox/models.py
```diff
@@ -11,16 +11,22 @@
...
     provider: str
     model_name: str
     api_key: str | None
     api_base: str | None
+    npm_package: str | None = None
```
wenxi-onyx: Kind of weird that we need this as a param on the model

backend/onyx/server/features/build/session/llm_config.py
```diff
@@ -1,127 +1,203 @@
...
 from onyx.utils.logger import setup_logger
 
 logger = setup_logger()
 
+CRAFT_RECOMMENDED_MODEL_NAMES = frozenset(
+    {
+        "gpt-5.6-sol",
+        "gpt-5.5",
+        "claude-fable-5",
+        "claude-opus-4-8",
```
wenxi-onyx: Opus 5?

backend/onyx/server/features/build/session/llm_config.py
```diff
@@ -1,127 +1,203 @@
...
+        for _, model in visible_models
+    ]
+    display_name_counts = Counter(clean_display_names)
+    # Model configs don't track max output tokens; derive it from the litellm
+    # map (as the main app does) so opencode's per-model limit is accurate.
+    model_map = get_model_map()
+    models: list[GatewayModelConfig] = []
+    for provider, model in visible_models:
+        display_name = model.custom_display_name or model.display_name or model.name
+        if display_name_counts[display_name] > 1:
```
wenxi-onyx, line 172: len?

backend/onyx/server/features/build/sandbox/util/opencode_config.py
rohoswagger: Done in addc125 — removed `build_session_opencode_config` and inlined `json.dumps(build_provider_opencode_config(...))` at the 5 call sites. Kept the dict-returning `build_provider_opencode_config` since the unit tests assert on its structure and folding the `json.dumps` into it would mix construction with serialization.

wenxi-onyx (approved):

backend/onyx/server/features/build/sandbox/util/opencode_config.py
rohoswagger: Done in `addc125` — removed `build_session_opencode_config` and inlined `json.dumps(build_provider_opencode_config(...))` at the 5 call sites. Kept `build_provider_opencode_config` returning a dict since the unit tests assert on its structure and folding the dumps in would mix construction with serialization.

backend/onyx/server/features/build/sandbox/util/opencode_config.py
rohoswagger: Fixed in `fd0f643`. Dropped the `assert` (and the now-redundant standalone guard); `build_provider_opencode_config` now checks `session_id` explicitly inside the `if mcp_servers:` block and raises `ValueError` at the point of use. I didn't add a separate log line: this is a caller-contract violation, so raising with a descriptive message (logged with a traceback by the global handler) is the right signal — log-and-continue would let a broken config through.

backend/onyx/server/features/build/sandbox/models.py
rohoswagger: Done in `fd0f643` — renamed `LLMProviderConfig` → `CraftLLMProviderConfig` across the module (models, builders, managers, tests) to disambiguate it from the main-app LLM provider models.

backend/onyx/server/features/build/sandbox/models.py
rohoswagger: Agreed — collapsed it in `fd0f643`. `build_onyx_gateway_config` is the only producer and the gateway is always openai-compatible, so `npm_package` was effectively a constant. Removed the field, hardcoded the npm package in `_build_provider_block`, and dropped the now-dead native-provider branch plus its unused `_model_options`/adaptive-thinking helpers (thinking translation lives in the core LLM layer, not here). ~40 fewer lines.

backend/onyx/server/features/build/session/llm_config.py
rohoswagger: Outdated — resolved by `0c91b62`. The hardcoded `CRAFT_RECOMMENDED_MODEL_NAMES` set (which pinned `claude-opus-4-8`) is gone; the default now derives from each provider's `default_model` in `recommended-models.json`, where anthropic's default is `claude-opus-5`. It tracks the json (regenerated by the update-recommended-models workflow) instead of a hand-maintained list.

backend/onyx/server/features/build/session/llm_config.py
rohoswagger: The check needs per-name multiplicity — whether *this* display name appears for more than one model across providers — so we can append the provider label only to the colliding ones. `len()` gives cardinality, not per-key counts; the alternatives are `Counter` (current: O(n) build + O(1) lookup) or `list.count()` in the loop (O(n²)). Happy to switch to `.count()` if you'd prefer fewer moving parts, but `Counter` read as the cleaner primitive. Let me know if you meant something else.

PR #13256, opened 2026-07-21
fix(auth): harden SSO callbacks, invite-only toggle, and error surfacing

nmgarza5:
## Description

Four gaps found auditing the multi-IdP SSO surface (OIDC + SAML + permission sync). No auth bypass; these are session-integrity, fail-open, and error-UX fixes. H1's real unblock (per-provider account linking) is a separate follow-up.

- **Deactivated SAML login:** SAML minted a session and logged a SUCCESS audit for a deactivated user. Now gated before login, matching OIDC.
- **Raw-JSON errors:** non-legacy OIDC and all SAML callbacks hit FastAPI directly, so rejections (invite-only, domain, already-exists) rendered as raw JSON. A shared decorator now redirects browsers to `/auth/error`; mobile and API callers still get JSON via content negotiation.
- **Invite-only silently disabled:** a partial settings PUT reset `invite_only_enabled` to `False`, and a settings-read error made the invite gate fail open. Now preserved through partial PUTs and fails closed on read error.
- **Cryptic codes:** `/auth/error` humanizes `OAUTH_USER_ALREADY_EXISTS` and `LOGIN_BAD_CREDENTIALS`.

Note: branches off current main, which predates #13237 (sso_managed removal). Overlaps are non-conflicting; whichever merges second rebases trivially.

## Additional Options

### DIFF
--- a/backend/onyx/server/settings/api.py
+++ b/backend/onyx/server/settings/api.py
@@ -86,12 +86,15 @@ def admin_put_settings(
         current_tier = Tier.COMMUNITY
     existing = load_settings()
 
-    # craft_default_enabled is access control: a PUT that omits it (e.g. an
-    # older client) must not silently reset it to the pydantic default.
+    # These fields are access control: a PUT that omits one (e.g. an older
+    # client, or a partial write) must not silently reset it to the pydantic
+    # default. invite_only_enabled off-by-default would let uninvited users in.
     if "craft_default_enabled" not in settings.model_fields_set:
         settings.craft_default_enabled = existing.craft_default_enabled
     if "craft_instructions" not in settings.model_fields_set:
         settings.craft_instructions = existing.craft_instructions
+    if "invite_only_enabled" not in settings.model_fields_set:
+        settings.invite_only_enabled = existing.invite_only_enabled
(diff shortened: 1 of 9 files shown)


backend/onyx/server/settings/api.py
```diff
@@ -86,12 +86,15 @@ def admin_put_settings(
         current_tier = Tier.COMMUNITY
     existing = load_settings()
 
-    # craft_default_enabled is access control: a PUT that omits it (e.g. an
-    # older client) must not silently reset it to the pydantic default.
+    # These fields are access control: a PUT that omits one (e.g. an older
```
acaprau, line 89: Think this kind of thing is what PATCH is for, we should try to use it more in our codebase in the future. Fine for now.

acaprau (approved):

backend/onyx/server/settings/api.py
nmgarza5: agreed. but didnt wanna make the change bigger than it needed to be. can follow up on this if you want

acaprau: Nah it's fine just whenever someone is touching something and they have the time to fix it we should keep PUT -> PATCH in mind.

PR #13257, opened 2026-07-21
feat(auth): per-provider account linking for SSO, and verify Google emails

nmgarza5:
## Description

Follow-up to #13256. Unblocks the real SSO-only lockout: an existing account (created by password, SAML, or another provider) could not sign in through a new OIDC/Google provider because auto-link was hardcoded off, and there was no admin or self-service way to link. This blocks IdP migrations.

- **Per-provider `allow_email_link` flag:** when on, a login may claim an existing same-email account instead of being rejected. Gated on the provider having a non-empty `allowed_email_domains`, so it can only adopt accounts in domains it is authoritative for. Create/update refuse the flag without a domain allowlist.
- **Verified email required (fixes M11):** linking is only safe for an IdP-verified email. OIDC already verified via `VerifiedEmailOpenID`; Google now does too via `VerifiedEmailGoogleOAuth2`, which also stops the People API raising raw `KeyError`/`StopIteration` on a malformed body.
- Column + migration, API + admin-modal toggle, unit tests.

Off current main, so overlaps #13256 and the sso_managed removal in `oidc_multi.py` at different lines. Whichever merges last rebases.

## Additional Options

### DIFF
--- a/backend/alembic/versions/48589afda2aa_add_allow_email_link_to_sso_provider.py
+++ b/backend/alembic/versions/48589afda2aa_add_allow_email_link_to_sso_provider.py
@@ -1,0 +1,33 @@
+"""add allow_email_link to sso_provider
+
+Revision ID: 48589afda2aa
+Revises: e2875ce6454b
+Create Date: 2026-07-21 12:50:50.134166
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "48589afda2aa"
+down_revision = "e2875ce6454b"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "sso_provider",
+        sa.Column(
+            "allow_email_link",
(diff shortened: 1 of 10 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
Closing for now, may reopen. On review the per-provider allow_email_link flag re-opens an account-takeover surface: it lets a provider claim existing accounts in its allowed domains by verified email, so if a provider is a lower-assurance IdP or shares a domain with another provider, it can adopt accounts created elsewhere. The domain allowlist only mitigates this under strict admin discipline (one authoritative IdP per domain). The customer's actual issue (permission-synced users logging in) is already handled by #13256 plus the merged placeholder-promotion fixes and does not need this flag. If IdP migration becomes a real requirement, a targeted admin "link this identity to this user" action is safer than a blanket per-provider domain-trust flag. The M11 Google email_verified hardening from this PR is worth salvaging into a separate PR if we keep pursuing this area.

PR #13266, opened 2026-07-21
refactor(settings): make /admin/settings a PATCH with partial-merge

nmgarza5:
## Description

Follow-up to @acaprau's note on #13256: the admin settings write is a `PUT` that behaves like a partial update, worked around with per-field `model_fields_set` guards.

- Convert the endpoint to `PATCH` and merge only the fields the caller sent onto the stored settings (`model_copy(update=...)`).
- This drops the per-field guards and gives every field correct partial-update semantics: an omitted field keeps its value instead of resetting to the pydantic default.
- Frontend writers (`updateAdminSettings`, ChatPreferences) and the integration `SettingsManager` switch to `PATCH`.

Full-object callers (all current FE writers) are behavior-neutral. Partial callers now preserve omitted fields.

Note: this reworks the same block as #13256's invite-only guard, so it should merge after #13256 (the merge subsumes that guard).

## Additional Options


nmgarza5:
@greptile

acaprau (approved):

justin-tahara (approved):

PR #13168, opened 2026-07-20
feat(mcp): proxy-authoritative per-tool approvals for craft MCP

Danelegend:
## What

Proxy-authoritative per-tool MCP approvals for Craft. The sandbox proxy becomes the sole gate for MCP tool calls: it recognizes each MCP request, resolves the admin's per-tool policy, and either forwards, denies, or parks it for user approval — the same pipeline external apps already use.

Stacked on **#13158** (PR 5, `dane/mcp-craft-sandbox-config`). Review/merge that first.

## Schema: one polymorphic `gated_app` identity table

The approval pipeline (per-action policy, approval rows, scheduled-task pre-approvals) previously attributed a gated request to an `external_app` row. This PR generalizes the target to *either* an external app *or* an MCP server via a single polymorphic identity table, rather than sprinkling `(external_app_id?, mcp_server_id?)` companion columns + CHECK across every consumer table.

- **`GatedApp`** (`gated_app`) — one row per gated target: `kind` + nullable `external_app_id`/`mcp_server_id` (each unique, `ON DELETE CASCADE`) + `num_nonnulls(...) = 1` CHECK. The single-target invariant lives here alone.
- **`ExternalAppPolicy` → `GatedActionPolicy`** — keyed by `gated_app_id`.
- **`action_approval`** / **`scheduled_task_pre_approved_app`** — each drop their app-specific FKs for one `gated_app_id` (`SET NULL` / `CASCADE` respectively).
- **`onyx/db/gated_app.py`** — centralizes all `(kind, id) ↔ gated_app_id` mapping (race-safe get-or-create via `ON CONFLICT`) and the shared policy read/write. Adding a third gated-target catalog is one nullable column on `gated_app` and nothing on the consumers.

The migration creates + backfills `gated_app`, repoints the three consumers, and renames the policy table. Validated with a full base→head upgrade **and** downgrade round-trip on a real DB.

## Runtime

- `McpRequestEvaluator` recognizes MCP JSON-RPC `tools/call` requests, resolves the effective per-tool policy (admin override else default `ASK`), and feeds the shared approval/park/grant pipeline.
- MCP tools default to `ASK` (no self-declared `readOnlyHint` auto-grant); admins relax via per-tool policy.
- Proxy remains the security boundary — opencode keeps MCP tools at `allow` (from PR 5), so enforcement is default-ASK at the proxy.

## Tests

[...]

PR #13293, opened 2026-07-22
feat(mcp): hot-reload craft MCP servers into running sandboxes

Danelegend:
## What

Hot-reloads a user's craft MCP set into their **running** sandbox — enabling a server, toggling its tools, or connecting credentials now reaches a live Craft session on its next turn, with no pod re-provision.

> **Base:** this stacks on `dane/mcp-craft-approvals-fe` (the craft MCP frontend), which in turn stacks on the approvals backend (#13168). Review after those.

## Why

The craft `mcp` block was baked into the pod-global `opencode.json` (delivered once via `OPENCODE_CONFIG_CONTENT`), and opencode-serve loads that config once at process start. So a RUNNING+healthy sandbox is reused as-is and never sees an MCP config change until it's destroyed and re-provisioned.

## How

- **Per-session config.** The craft `mcp` block + its per-tool permission gates move out of the pod-global config into a per-session `opencode.json` (`build_session_mcp_config`), written by the sandbox managers (`write_session_opencode_config`) and re-read when the session's opencode instance is disposed. opencode merges it with the pod-global config, so the server set can change without restarting opencode-serve.
- **Rides the existing skills-reload pipeline.** The craft MCP set (server ids + urls, disabled-tool sets, per-user auth) is folded into the managed-runtime hash (`compute_skill_runtime_hash`), so an MCP change marks the session stale via the existing `skills_are_stale` / `reload_session_skills` path — no new staleness plumbing. `push_skills_for_users` is triggered on craft-enable (admin) and per-user credential connect (`affected_user_ids_for_mcp_server`).
- **Session attribution for MCP approvals.** opencode's in-process MCP client uses the untagged base proxy env (the shell-env proxy tag only rides shell-subprocess egress), so ASK-policy MCP tool calls previously couldn't be attributed to a session (`no_active_session`). Each per-session MCP server is now stamped with an `X-Onyx-Mcp-Session` header carrying the BuildSession id; the egress proxy reads it in `_extract_session_tag` and strips it before the origin sees it.

Approval **policy** enforcement is unchanged and still fully live (proxy `effective_mcp_tool_policy`) — only the server set / tool set / discovery state hot-reloads.

## Tests

- Unit: pod-global config never carries `mcp`; per-session builder emits the server set + permission gates + session-tag header; runtime-hash fingerprint is order-independent and reacts to server set / url / tools / per-user auth.
- External-dependency-unit: `resolve_craft_mcp_servers` populates `server_id`/`authenticated`; connecting credentials flips `authenticated` and changes the fingerprint; skills-push self-heal still holds with MCP folded into the hash.

## Additional Options

---
## Summary by cubic
Hot-reload Craft MCP servers and tool gates into running sandboxes via per-session `opencode.json`. Sessions pick up server, tool, and credential changes on their next turn without re-provisioning; MCP staleness is tracked separately from skills.

- **New Features**
  - Per-session MCP config: write `sessions/<session>/opencode.json` on create/restore and before reload; wildcard allow with per-tool denies; pod-global `opencode.json` no longer carries `mcp`.
  - Session attribution header: each MCP server includes `X-Onyx-Mcp-Session=<session_id>`; the proxy prefers it, attributes approvals, and strips it (and `Proxy-Authorization`) before forwarding.
  - Separate MCP staleness: add `mcp_config_hash` to `sandbox` and `build_session`; sessions reload when either `skills_hash` or `mcp_config_hash` changes. `refresh_mcp_config_hashes_for_users` restamps RUNNING sandboxes after MCP edits and credential connects (incl. OAuth), includes admins, and unions pre/post ACL recipients so removed users drop servers.

- **Bug Fixes**
  - Hot-reload now also triggers on MCP server delete and user credential disconnect: snapshot recipients before delete and restamp after commit so running sessions drop the server/credentials next turn.
  - OAuth auto-discovery fix: restamp the caller’s running sandbox after the token-exchange commit so discovery state is reflected immediately.

<sup>Written for commit 72a8d63fc63b99be2667502caa3ada476d1a471d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13293?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/server/features/build/configs.py
+++ b/backend/onyx/server/features/build/configs.py
@@ -1,5 +1,9 @@
+import hashlib
+import hmac
 import os
 from enum import Enum
+
+from onyx.configs.app_configs import ENCRYPTION_KEY_SECRET
 
 
 class SandboxBackend(str, Enum):
@@ -144,6 +148,33 @@ SANDBOX_PROXY_INJECTED_PLACEHOLDER = "replaced_by_egress_proxy"
 # The egress proxy reads it to attribute the tool call to a session for approval,
 # then strips it so it never reaches the MCP origin.
 MCP_SESSION_TAG_HEADER = "X-Onyx-Mcp-Session"
+
+
+def sign_session_tag(session_id: str) -> str:
+    """Value for ``MCP_SESSION_TAG_HEADER``: ``<session_id>.<hmac>``. The session
+    id is not secret (it leaks via deep links / notifications), so the sandbox
+    could otherwise forge a tag for a session it doesn't own. HMAC with the
+    shared ``ENCRYPTION_KEY_SECRET`` (which the sandbox never sees) makes a tag
+    unforgeable without the key."""
+    mac = hmac.new(
+        ENCRYPTION_KEY_SECRET.encode(), session_id.encode(), hashlib.sha256
+    ).hexdigest()
+    return f"{session_id}.{mac}"
+
+
+def verify_session_tag(value: str) -> str | None:
+    """The session id from a signed tag, or ``None`` if the signature is absent
+    or invalid — callers fail closed."""
+    session_id, _, mac = value.partition(".")
+    if not mac:
+        return None
+    expected = hmac.new(
+        ENCRYPTION_KEY_SECRET.encode(), session_id.encode(), hashlib.sha256
+    ).hexdigest()
+    if not hmac.compare_digest(mac, expected):
+        return None
+    return session_id
+
 
 # ==============================================================================
 # Docker sandbox (SANDBOX_BACKEND=docker, self-hosted docker-compose)
--- a/backend/onyx/server/features/build/configs.py
+++ b/backend/onyx/server/features/build/configs.py
@@ -137,6 +137,14 @@ SANDBOX_PROXY_CA_VOLUME_NAME = "sandbox_proxy_ca"
 # never see the raw values.
 SANDBOX_PROXY_INJECTED_PLACEHOLDER = "replaced_by_egress_proxy"
 
+# Header carrying the originating BuildSession id on opencode's in-process MCP
+# client requests. opencode-serve is one process for many sessions and uses the
+# untagged base proxy env, so the shell-env proxy tag can't ride MCP egress;
+# instead the per-session opencode.json stamps this header on each MCP server.
+# The egress proxy reads it to attribute the tool call to a session for approval,
+# then strips it so it never reaches the MCP origin.
+MCP_SESSION_TAG_HEADER = "X-Onyx-Mcp-Session"
+
 # ==============================================================================
 # Docker sandbox (SANDBOX_BACKEND=docker, self-hosted docker-compose)
 # ==============================================================================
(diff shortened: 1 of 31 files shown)


backend/onyx/server/features/build/session/sandbox_lifecycle.py
```diff
@@ -176,7 +179,12 @@ def hydrate_managed_content(
 
     skills_hydrated = False
     try:
-        skills_hash = compute_skill_runtime_hash(skills_files, connectable_apps_section)
+        mcp_fingerprint = craft_mcp_fingerprint(
+            resolve_craft_mcp_servers(db_session, user)
+        )
+        skills_hash = compute_skill_runtime_hash(
```
wenxi-onyx: MCPs are riding along skills here? Maybe should rename this path to reflect that MCPs are part of the payload hash + migrate the column name to something appropriate.

I think it's also justifiable to store the MCP config hash separate from the skill hash so we aren't re-purposing the skill push code just to update MCPs.

backend/onyx/server/features/build/configs.py
```diff
@@ -137,6 +141,41 @@ class SandboxBackend(str, Enum):
...
+# Header carrying the originating BuildSession id on opencode's in-process MCP
+# client requests. opencode-serve is one process for many sessions and uses the
+# untagged base proxy env, so the shell-env proxy tag can't ride MCP egress;
+# instead the per-session opencode.json stamps this header on each MCP server.
+# The egress proxy reads it to attribute the tool call to a session for approval,
+# then strips it so it never reaches the MCP origin.
+MCP_SESSION_TAG_HEADER = "X-Onyx-Mcp-Session"
+
+
+def sign_session_tag(session_id: str) -> str:
```
wenxi-onyx: I don't think this is providing any additional security - the encoded value is stored in the sandbox as plaintext, so any malicious actor in the sandbox can just read this value and append it to their request. At the API server side, this request would look identical to a real one from the MCP client

backend/onyx/server/features/mcp/api.py
```diff
@@ -358,6 +359,19 @@ def _build_oauth_admin_config_data_for_update(
...
+    """Push the current craft MCP config to affected users' running sandboxes so
+    a live Craft session picks it up on its next turn (via the skills-reload
+    pipeline) without a pod re-provision. Best-effort — the push pipeline logs
+    and swallows per-sandbox failures. Imported lazily to avoid a build-layer
+    import cycle at module load."""
+    if not user_ids:
+        return
+    from onyx.skills.push import push_skills_for_users
+
+    push_skills_for_users(user_ids, db_session)
```
wenxi-onyx: See above comment - I think we should rename this path OR give MCPs their own path to dispose sessions

wenxi-onyx (approved):

Danelegend:
@greptile @cubic

PR #13294, opened 2026-07-22
Bound search preprocessing LLM calls

### LINKED ISSUE #13178: Search hangs indefinitely: query rephrase / keyword expansion / source-scope LLM calls in /api/search are unbounded (same class as #12333)
## Summary

The synchronous `/api/search` path runs three query-preprocessing "secondary LLM flows" before retrieval, and all three call `llm.invoke()` **without a timeout**:

- `semantic_query_rephrase` (`backend/onyx/secondary_llm_flows/query_expansion.py`)
- `keyword_query_expansion` (`backend/onyx/secondary_llm_flows/query_expansion.py`)
- `decide_search_scope` (`backend/onyx/secondary_llm_flows/source_filter.py`)

If the LLM provider connection stalls, the call never returns. These run in a threadpool via `run_functions_tuples_in_parallel` inside the sync `/api/search` endpoint, so the worker hangs indefinitely and search capacity is exhausted until the pod is restarted.

This is the **same failure mode** that #12333 ("bound document_filter LLM calls with a timeout") fixed, but on the pre-retrieval path #12333 did not cover. `document_filter` is now bounded by `SECONDARY_LLM_FLOW_TIMEOUT_S`; these three flows are not.

## How it manifests

Observed in production behind the MCP gateway: once a worker hangs, every search returns "search failed" and the gateway hits `httpx.ReadTimeout` on `POST /api/search`. A py-spy dump of the api-server showed workers stuck in the litellm/httpx path for `semantic_query_rephrase` / `keyword_query_expansion` / `decide_search_scope`, blocked acquiring a connection from litellm's shared pool (deepest frame `httpcore/_synchronization.py __enter__`). Restarting the api-server clears it; it recurs.

`skip_query_expansion=true` does **not** avoid it: it only skips the two `query_expansion` calls, while `decide_search_scope` still runs.

Still present on `main` (checked through v4.4.0-beta.1): all three `invoke()` calls pass no `timeout_override`, and `_expand_queries_and_decide_scope` calls the runner with no wall-clock timeout. An upgrade does not fix it.

## Proposed fix

Mirror #12333: pass `timeout_override=SECONDARY_LLM_FLOW_TIMEOUT_S` to all three `invoke()` calls so they fail fast, with graceful fallback (source_filter already degrades to all-sources; semantic rephrase falls back to the raw query; keyword expansion falls back to no expansion).

I have this change ready (with unit tests) and am happy to open a PR once the issue is triaged. cc @yuhongsun96 per CONTRIBUTING.

## Environment

- Onyx v4.3.6, self-hosted, Anthropic provider. Also verified against current `main`.

shubhamsinnh:
## Summary

Fixes #13178.

The synchronous `/api/search` path runs secondary LLM preprocessing before retrieval. Three of those calls were invoking the configured LLM without a timeout:

- semantic query rephrase
- keyword query expansion
- source-scope detection

If the provider connection stalled, the worker could wait indefinitely before retrieval. This mirrors the failure mode previously addressed for document filtering.

## Changes

- Reuse `SECONDARY_LLM_FLOW_TIMEOUT_S` for the three reported preprocessing calls.
- Pass `timeout_override` into `semantic_query_rephrase`, `keyword_query_expansion`, and `decide_search_scope`.
- Add focused unit coverage asserting the timeout is forwarded to `LLM.invoke`.

## Validation

- `.venv\Scripts\python.exe -m pytest -q backend/tests/unit/onyx/secondary_llm_flows/test_query_expansion.py backend/tests/unit/onyx/secondary_llm_flows/test_source_filter.py backend/tests/unit/onyx/secondary_llm_flows/test_document_filter.py`
- `.venv\Scripts\python.exe -m ruff check backend/onyx/secondary_llm_flows/query_expansion.py backend/onyx/secondary_llm_flows/source_filter.py backend/tests/unit/onyx/secondary_llm_flows/test_query_expansion.py backend/tests/unit/onyx/secondary_llm_flows/test_source_filter.py`

---
## Summary by cubic
Prevents hangs in synchronous `/api/search` by bounding all secondary LLM work before and after retrieval, with timeouts and safe fallbacks if a provider stalls.

- **Bug Fixes**
  - Pre-retrieval: enforce `SECONDARY_LLM_FLOW_TIMEOUT_S` on semantic query rephrase, keyword expansion, and source-scope; tests assert `timeout_override` is forwarded.
  - Post-retrieval: add wall-clock caps for section selection and expansion (select up to 10, limit workers), with timeouts falling back to top retrieved sections or the original section; add `DISABLE_LLM_SEARCH_SECTION_PROCESSING`, `SEARCH_SECTION_SELECTION_TIMEOUT_S`, `SEARCH_SECTION_EXPANSION_TIMEOUT_S`, `SEARCH_SECTION_EXPANSION_MAX_WORKERS`; tests cover fallbacks and bounded runner; share `noop_span`/`make_llm` helpers.

<sup>Written for commit 21e344ea777d226e52e931c3ebcba5347db16e89. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13294?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/configs/chat_configs.py
+++ b/backend/onyx/configs/chat_configs.py
@@ -48,6 +48,35 @@ DR_REPORT_LLM_TIMEOUT_S = int(os.environ.get("DR_REPORT_LLM_TIMEOUT_S") or "60")
 SECONDARY_LLM_FLOW_TIMEOUT_S = int(
     os.environ.get("SECONDARY_LLM_FLOW_TIMEOUT_S") or "60"
 )
+# Emergency escape hatch for the post-retrieval LLM section-selection and
+# context-expansion phase. Retrieval still runs, but results are passed through
+# without secondary LLM filtering/expansion.
+DISABLE_LLM_SEARCH_SECTION_PROCESSING = (
+    os.environ.get("DISABLE_LLM_SEARCH_SECTION_PROCESSING", "").lower() == "true"
+)
+# Section expansion issues one LLM call per selected section. Keep the default
+# serialized to avoid stressing shared sync provider connection pools.
+SEARCH_SECTION_EXPANSION_MAX_WORKERS = max(
+    1, int(os.environ.get("SEARCH_SECTION_EXPANSION_MAX_WORKERS") or "1")
+)
+# Wall-clock cap for the post-retrieval section-selection LLM call. This is
+# separate from the provider timeout because lock acquisition may not observe it.
+SEARCH_SECTION_SELECTION_TIMEOUT_S = max(
+    1,
+    int(
+        os.environ.get("SEARCH_SECTION_SELECTION_TIMEOUT_S")
+        or str(SECONDARY_LLM_FLOW_TIMEOUT_S)
+    ),
+)
+# Wall-clock cap for the whole section-expansion phase. Timed-out work falls
+# back to the original selected section while the stuck worker is abandoned.
+SEARCH_SECTION_EXPANSION_TIMEOUT_S = max(
+    1,
+    int(
+        os.environ.get("SEARCH_SECTION_EXPANSION_TIMEOUT_S")
+        or str(SECONDARY_LLM_FLOW_TIMEOUT_S)
+    ),
+)
 # Live buffer TTL. Refreshed per write.
 CHAT_STREAM_BUFFER_TTL_S = int(os.environ.get("CHAT_STREAM_BUFFER_TTL_S") or 3600)
 # Retention after the run is done.
--- a/backend/onyx/secondary_llm_flows/query_expansion.py
+++ b/backend/onyx/secondary_llm_flows/query_expansion.py
@@ -1,3 +1,4 @@
+from onyx.configs.chat_configs import SECONDARY_LLM_FLOW_TIMEOUT_S
 from onyx.configs.constants import MessageType
 from onyx.llm.interfaces import LLM
 from onyx.llm.models import (
@@ -138,7 +139,11 @@ def semantic_query_rephrase(
     with llm_generation_span(
         llm=llm, flow=LLMFlow.SEMANTIC_QUERY_REPHRASE, input_messages=messages
     ) as span_generation:
-        response = llm.invoke(prompt=messages, reasoning_effort=ReasoningEffort.OFF)
+        response = llm.invoke(
+            prompt=messages,
+            reasoning_effort=ReasoningEffort.OFF,
+            timeout_override=SECONDARY_LLM_FLOW_TIMEOUT_S,
+        )
         record_llm_response(span_generation, response)
         final_query = response.choice.message.content
 
@@ -218,7 +223,11 @@ def keyword_query_expansion(
     with llm_generation_span(
         llm=llm, flow=LLMFlow.KEYWORD_QUERY_EXPANSION, input_messages=messages
     ) as span_generation:
-        response = llm.invoke(prompt=messages, reasoning_effort=ReasoningEffort.OFF)
+        response = llm.invoke(
+            prompt=messages,
+            reasoning_effort=ReasoningEffort.OFF,
+            timeout_override=SECONDARY_LLM_FLOW_TIMEOUT_S,
+        )
         record_llm_response(span_generation, response)
         content = response.choice.message.content
 
--- a/backend/onyx/secondary_llm_flows/source_filter.py
+++ b/backend/onyx/secondary_llm_flows/source_filter.py
@@ -2,6 +2,7 @@ import json
 
 from pydantic import BaseModel
 
+from onyx.configs.chat_configs import SECONDARY_LLM_FLOW_TIMEOUT_S
 from onyx.configs.constants import DocumentSource, MessageType
 from onyx.llm.interfaces import LLM
 from onyx.llm.models import ChatCompletionMessage, ReasoningEffort, UserMessage
@@ -114,7 +115,11 @@ def decide_search_scope(
             flow=LLMFlow.SOURCE_FILTER_EXTRACTION,
             input_messages=messages,
         ) as span_generation:
-            response = llm.invoke(prompt=messages, reasoning_effort=ReasoningEffort.OFF)
+            response = llm.invoke(
+                prompt=messages,
+                reasoning_effort=ReasoningEffort.OFF,
+                timeout_override=SECONDARY_LLM_FLOW_TIMEOUT_S,
+            )
             record_llm_response(span_generation, response)
             content = response.choice.message.content
     except Exception:
--- a/backend/onyx/tools/tool_implementations/search/search_tool.py
+++ b/backend/onyx/tools/tool_implementations/search/search_tool.py
@@ -42,7 +42,13 @@ from pydantic import BaseModel
 from sqlalchemy.orm import Session
 
 from onyx.chat.emitter import Emitter
-from onyx.configs.chat_configs import MAX_CHUNKS_FED_TO_CHAT
+from onyx.configs.chat_configs import (
+    DISABLE_LLM_SEARCH_SECTION_PROCESSING,
+    MAX_CHUNKS_FED_TO_CHAT,
+    SEARCH_SECTION_EXPANSION_MAX_WORKERS,
+    SEARCH_SECTION_EXPANSION_TIMEOUT_S,
+    SEARCH_SECTION_SELECTION_TIMEOUT_S,
+)
 from onyx.configs.constants import DocumentSource, FederatedConnectorSource
 from onyx.context.search.federated.slack_search import slack_retrieval
 from onyx.context.search.models import (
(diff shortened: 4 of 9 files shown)


shubhamsinnh:
Hey  @jmelahman @justin-tahara, would appreciate if you could take a look.

This draft fixes #13178 by mirroring the timeout pattern from #12333 for the three reported `/api/search` preprocessing LLM calls:
- `semantic_query_rephrase`
- `keyword_query_expansion`
- `decide_search_scope`

I checked for overlap and didn’t find another open PR linked to the issue.

cc @jaceksan since they opened the issue and mentioned having a patch ready. Also cc @yuhongsun96, who was already tagged in the issue per CONTRIBUTING.

jaceksan:
Two findings from a production hang that I think fall outside this PR's scope.

**1. Disabling the LLM search flows is not possible via config.** 
On v4.3.6 and current `main`, `SearchTool.run` calls `select_sections_for_expansion` -> `expand_section_with_context` -> `classify_section_relevance` unconditionally.
There is no `evaluation_type`, env var, persona field, or search-request flag to skip the LLM relevance / section-expansion step, so operators have no way to turn it off as a mitigation.

**2. Per-call timeouts likely won't fix the hang, because the failure is a connection-pool lock deadlock, not a slow response.** 
py-spy on a hung api-server
showed ~13 threads on the same frame and zero threads doing socket I/O:
```
    __enter__ (httpcore/_synchronization.py:268)          # sync pool lock
    handle_request (httpcore/_sync/connection_pool.py:218)
    send (httpx/_client.py:914) -> litellm anthropic make_sync_call
    classify_section_relevance (secondary_llm_flows/document_filter.py)
    expand_section_with_context / expand_section_safe (search_tool.py)
```
One thread was stuck in the release path (`connection_pool.py:416 close -> __enter__`) while the others blocked acquiring the same lock; egress to the provider was healthy (TLS ~25ms). 
So a read/flow timeout has nothing to interrupt: you cannot cancel a `threading.Lock` acquisition with an HTTP timeout, and this is the post-retrieval section-expansion flow, not the query-preprocessing calls this PR bounds.

Root cause is the thread model: section expansion fans out ~10 concurrent *sync* litellm calls via a ThreadPoolExecutor, all sharing litellm's single global sync httpcore connection pool, and they deadlock on that pool's internal lock under connection reuse/close timing. Timeouts bound the wrong thing.

A durable fix probably needs a concurrency redesign rather than more timeouts, e.g.:
- cap or serialize the section-expansion fan-out (small / single-worker pool), or
- use a per-call / per-thread httpx client so threads don't share one sync pool, or
- run these secondary flows on the async LLM path (single event loop + async pool) instead of thread-pool + sync httpx.

An httpx `Timeout(pool=...)` would at least turn the hang into a raised error, but does not remove the deadlock itself.

shubhamsinnh:
@jaceksan thanks for looking into this.

I pushed an additional mitigation commit (`d65c64a`) to this PR branch. It keeps the original preprocessing timeout fix and also bounds the later section-processing phase:

- Added `DISABLE_LLM_SEARCH_SECTION_PROCESSING` as an emergency escape hatch. When enabled, retrieval still runs, but LLM section selection/context expansion are skipped and the retrieved sections are returned directly.
- Added `SEARCH_SECTION_SELECTION_TIMEOUT_S` around the post-retrieval LLM section-selection call, with fallback to the top retrieved sections.
- Added `SEARCH_SECTION_EXPANSION_MAX_WORKERS` (default `1`) and `SEARCH_SECTION_EXPANSION_TIMEOUT_S` around per-section context expansion, with fallback to the original selected section on timeout.
- Added unit coverage for the bounded runner config, disable flag, section-selection timeout fallback, and section-expansion timeout fallback.

Local validation completed:

- `python -m ruff check backend/onyx/configs/chat_configs.py backend/onyx/tools/tool_implementations/search/search_tool.py backend/tests/unit/onyx/tools/tool_implementations/search/test_search_tool_run.py`
- `python -m pytest -q backend/tests/unit/onyx/tools/tool_implementations/search backend/tests/unit/onyx/secondary_llm_flows` -> 52 passed
- `python -m pytest -q backend/tests/unit/onyx/tools/test_tool_runner.py backend/tests/unit/onyx/chat/test_context_files.py` -> 43 passed

I also tried broader local suites:

- `backend/tests/unit/onyx/tools` with UTF-8 mode got to 460 passed, then hit an unrelated timing-sensitive failure in `open_url/test_onyx_web_crawler.py::TestParallelExecution::test_multiple_urls_fetched_concurrently`.
- `backend/tests/unit/onyx/chat` got to 458 passed, then hit an unrelated `test_save_chat.py::test_csv_file_type` assertion around generated CSV file typing.

Live smoke challenge:

Live smoke note: I could not complete a local end-to-end search smoke because the local Docker dependency stack was unavailable, so Postgres/Redis/OpenSearch/MinIO were not running. Targeted backend unit coverage and lint passed; a runtime smoke on a working Onyx dev stack is still recommended.

Expectation for runtime behavior:

Search should no longer hang indefinitely in the post-retrieval LLM section-processing path. If section selection or expansion gets stuck past the wall-clock cap, the request should degrade to retrieved/original sections and continue. If operators need an immediate mitigation, `DISABLE_LLM_SEARCH_SECTION_PROCESSING=true` should bypass that LLM section-processing phase entirely while preserving retrieval results.

shubhamsinnh:
Updated the branch against current `upstream/main` and force-pushed the rebased PR branch.

Follow-up on the Cubic comments:

- `P1` / disabled-path elapsed timing: this path was already guarded by the `document_selection_elapsed = 0.0` and `document_expansion_elapsed = 0.0` initialization at the top of `SearchTool.run`; the existing disable-path unit test exercises a non-empty search through final prompt construction.
- `P2` / timeout env values: addressed by clamping `SEARCH_SECTION_SELECTION_TIMEOUT_S` and `SEARCH_SECTION_EXPANSION_TIMEOUT_S` to a positive minimum at config parsing.
- `P3` / test helper callback typing: addressed by narrowing the fake timeout callback guards with `callable(timeout_callback)` before invocation.

Re-ran after the rebase:

- `python -m ruff check backend/onyx/configs/chat_configs.py backend/onyx/tools/tool_implementations/search/search_tool.py backend/tests/unit/onyx/tools/tool_implementations/search/test_search_tool_run.py` -> passed (Ruff cache write warnings only)
- `python -m pytest -q backend/tests/unit/onyx/tools/tool_implementations/search backend/tests/unit/onyx/secondary_llm_flows` -> 52 passed

jaceksan:
@shubhamsinnh Thanks! However, I am wandering if anyone will review/merge this. I analyzed Git history and there is almost zero chance anyone will look into this ;-(
Let's hope.
As a dirty nasty workaround, I deployed a thin CronJob restarting the affected Onyx POD every 24 hours. It lowers the probability of the issue significantly but not 100%.

shubhamsinnh:
Following up because the production hang reported in #13178 is still affecting at least one deployment, whose current workaround is scheduled pod restarts. This PR bounds both preprocessing and post-retrieval section processing and includes an emergency bypass, with the reported review findings addressed. Since the PR was closed without feedback, could a maintainer confirm whether the mitigation direction was declined or whether a smaller replacement PR would be welcome?

jaceksan:
Closing the loop for anyone landing here from #13178.

This was fixed upstream by #13295 (commit afaa3552d, "isolate Anthropic sync calls
from litellm's shared HTTP pool"), merged 2026-07-24, about 30 minutes after this PR
was closed. So the direction wasn't declined, it was superseded by a maintainer's own fix.

It fixes the pool-lock deadlock rather than adding timeouts: `LitellmLLM._uses_isolated_client()`
now gives Anthropic a fresh per-call `HTTPHandler` in both the invoke and streaming paths,
as true-OpenAI already did. That matches the py-spy trace above (threads blocked on
`httpcore/_synchronization.py:268 __enter__`, none doing socket I/O); the commit cites
encode/httpcore#996.

Version gotcha for other self-hosters: the fix is **not** in v4.4.x. v4.4.7 was cut
2026-07-28, after the fix, but from a release branch that diverged before it, and both
call sites there still test `is_true_openai_model(...)` directly. The earliest published
tag carrying it is **v4.5.0-beta.0**.

We upgraded v4.3.6 -> v4.5.0-beta.0 on 2026-08-03: 34 migrations applied cleanly, no
reindex, search healthy. Whether the deadlock is fully gone needs longer observation (our
previous wedges took hours to appear under real traffic), so we've kept the nightly
api-server restart CronJob until we have a couple of clean weeks.

Thanks @shubhamsinnh for the work here.

PR #13297, opened 2026-07-22
feat(auth): add OIDC_TRUST_UNVERIFIED_EMAIL option to bypass strict verification

Oht8wooWi8yait9n:
### Summary
This PR introduces a new environment variable `OIDC_TRUST_UNVERIFIED_EMAIL` to allow OIDC sign-ins and bypass the strict `"email_verified"` check when it is missing or false.
### Why is this needed?
In Onyx `v4.4.0`, a strict check was introduced requiring identity providers to explicitly return `"email_verified": true` in the OpenID user information payload.
However, **Microsoft Entra ID (Azure AD)** — the single most common enterprise identity provider in the world — **does not emit the standard OIDC `email_verified` claim** by default for workforce or managed tenant accounts. This causes all stock `v4.4.0` Onyx OIDC deployments with Entra ID to fail out-of-the-box with:
`OnyxError VALIDATION_ERROR: Could not retrieve a verified identity from the SSO provider`
Setting `OIDC_TRUST_UNVERIFIED_EMAIL=true` provides a standard way to bypass this strict check and trust the email, resolving this blocker for all Azure AD/Entra ID enterprise users.
### Proposed Changes
- **`backend/onyx/configs/app_configs.py`**: Added the `OIDC_TRUST_UNVERIFIED_EMAIL` environment variable configuration (defaults to `False`).
- **`backend/onyx/auth/oidc_client.py`**: Updated `VerifiedEmailOpenID.get_id_email()` to bypass the strict `"email_verified"` claim check if `OIDC_TRUST_UNVERIFIED_EMAIL` is set to `True`.
### Verification Plan
1. Deploy Onyx behind Microsoft Entra ID.
2. Set `OIDC_TRUST_UNVERIFIED_EMAIL=true` in the environment.
3. Attempt OIDC SSO login. It should succeed without any validation errors.

---
## Summary by cubic
Adds `OIDC_TRUST_UNVERIFIED_EMAIL` to allow OIDC sign-ins when `email_verified` is missing or false. This keeps the strict default while enabling Microsoft Entra ID setups to work.

- New Features
  - Introduced `OIDC_TRUST_UNVERIFIED_EMAIL` env var (default: false).
  - When true, `VerifiedEmailOpenID.get_id_email()` skips the `email_verified` check and trusts the `email` claim.

- Migration
  - For Microsoft Entra ID, set `OIDC_TRUST_UNVERIFIED_EMAIL=true` to allow sign-in.

[...]

PR #13300, opened 2026-07-22
fix(auth): tolerate OIDC IdPs that omit email_verified, strict as per-provider opt-in

nmgarza5:
## Description

**Bug:** Microsoft Entra ID does not emit the optional OIDC `email_verified` claim, so every v4.4.0 OIDC + Entra login fails with "Could not retrieve a verified identity from the SSO provider". Reported in #13297 (thanks @Oht8wooWi8yait9n).

**Fix:** an absent `email_verified` claim is now tolerated by default, so Entra works out of the box. A new per-provider `require_verified_email` switch on the OIDC provider row opts back into requiring the claim, instead of #13297's global env var. Per-provider matches the v4.4 model (one strict setting must not be forced by another IdP on a multi-IdP instance) and follows Dex/oauth2-proxy per-connector precedent.

**Semantics:** any *present* `email_verified` value other than `true` (explicit `false`, JSON `null`, stringly-typed values) is rejected in both modes, so the guard against the nOAuth takeover vector stays intact. IdPs that send the claim (Google, Okta, Auth0, Keycloak, Cognito) keep full strictness automatically. The default only changes behavior for deployments that are currently hard-broken.

- `OIDCProviderConfig.require_verified_email: bool = False` (no migration, config is a validated JSON blob; `extra="forbid"` keeps it OIDC-only)
- `VerifiedEmailOpenID` constructor kwarg consulted in `get_id_email`
- Admin modal: "Require Verified Email Claim" switch via a new `switch` field kind rendered by the existing Formik `SwitchField`
- Unit tests pin the full semantics matrix (54 pass)

Closes #13297.

## How Has This Been Tested?
Screenshot 2026-07-22 at 10 24 33 AM

## Additional Options

### DIFF
--- a/backend/onyx/auth/oidc_client.py
+++ b/backend/onyx/auth/oidc_client.py
@@ -69,11 +69,13 @@ def validate_issuer_owns_config_url(
 
 
 class VerifiedEmailOpenID(OpenID):
-    """OpenID client that refuses to hand back an email the IdP has not
-    verified. An absent email_verified claim counts as unverified, since a
-    mutable, unverified email claim is exactly the nOAuth account-takeover
-    vector. A response with no email at all is passed through unchanged and
-    rejected by the login flow's existing no-email handling."""
+    """OpenID client that guards the email claim. Any present email_verified
+    value other than true is rejected, since a mutable, unverified email
+    claim is exactly the nOAuth account-takeover vector. An absent claim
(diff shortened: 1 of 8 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

acaprau (approved):

Issue #11420, opened 2026-05-27
Azure Government / Azure OpenAI Reasoning Models (gpt-5.1/o1/o3) Fail with '400 Bad Request' due to /responses API Routing


Oht8wooWi8yait9n:
## Description
When attempting to use OpenAI reasoning models (specifically `gpt-5.1` or `o1`/`o3` family models) deployed via **Azure OpenAI** (especially in restricted or regulated environments like **Azure Government**), chat queries fail with a `400 Bad Request` or `404 Not Found` error.

This happens because Onyx classifies all standard OpenAI reasoning models as "true" OpenAI models and routes them through LiteLLM's `responses` bridge (appending `/responses` to the model name, e.g. `azure/responses/gpt-5.1`). However, Azure Government and many enterprise Azure OpenAI deployments do not support the unified `/responses` API and reject these requests. They expect standard chat completion routing via `/chat/completions` with the `reasoning_effort` parameter passed at the root of the request payload.

---

## Steps to Reproduce
1. In the Onyx LLM administration page, add an **Azure OpenAI** provider pointing to an Azure Government resource.
2. Configure a reasoning model deployment (such as `gpt-5.1` or `o1-mini`).
3. Select this model in the chat interface and submit a query.
4. The system returns an empty response, and the backend logs show a `Client error '400 Bad Request'` from the Azure endpoint when trying to POST to the `/responses` path.

---

## Traceback
```python
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/litellm/llms/custom_httpx/llm_http_handler.py", line 2254, in response_api_handler
    response = sync_httpx_client.post(
               ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/litellm/llms/custom_httpx/http_handler.py", line 1085, in post
    _raise_masked_sync_error(e, stream)
  File "/usr/local/lib/python3.11/site-packages/litellm/llms/custom_httpx/http_handler.py", line 352, in _raise_masked_sync_error
    raise MaskedHTTPStatusError(e, message=_body, text=_body) from None
litellm.llms.custom_httpx.http_handler.MaskedHTTPStatusError: Client error '400 Bad Request' for url 'https://<your-azure-resource>.openai.azure.us/openai/responses?api-version=2025-03-01-preview'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #13334, opened 2026-07-22
fix(llm): don't send reasoning_effort to o1-mini / o1-preview

jmelahman:
## Description

Related to #11420 — fixes one of the two failure modes reported there (the other, `/responses`-routing on Azure Government, is pending validation from the reporter: https://github.com/onyx-dot-app/onyx/issues/11420#issuecomment-5052075774).

`o1-mini` and `o1-preview` reject the reasoning-effort parameter on **every** API surface (chat completions and responses alike) — only their default effort works. Onyx sent an effort value to any model flagged `supports_reasoning`, so these models failed with:

```
400 - Unsupported parameter: 'reasoning.effort' is not supported with this model.
```

on OpenAI and Azure both.

## Changes

- New capability predicate `openai_model_rejects_reasoning_effort()` in `onyx/llm/model_capabilities.py`, alongside its companion `model_is_reasoning_model()`. It uses an explicit deny-list (`o1-mini`, `o1-preview`, dated variants included via prefix match) rather than a registry lookup — LiteLLM's Azure config wrongly claims `reasoning_effort` support for these models, and its correct OpenAI answer is an accident of the models missing from the registry entirely.
- `LitellmLLM._completion` gates the reasoning-params block on it, so neither the nested responses-API `reasoning` param nor the root-level `reasoning_effort` is sent to these models, regardless of routing.

## Testing

- New unit tests: parametrized over `o1-mini` / `o1-preview` / `o1-mini-2024-09-12` on both routing paths (responses bridge and chat completions) asserting no reasoning params are sent, plus an `o1` counter-test guarding against over-matching.
- `pytest backend/tests/unit/onyx/llm` — 348 passed.
- Full `pytest backend/tests/unit` — one failure in `test_retrieve_all_slim_docs_perm_sync` that reproduces identically on a clean tree (pre-existing order-dependent flake, also noted in #12890).

---
## Summary by cubic
Stop sending `reasoning_effort` to OpenAI `o1-mini` and `o1-preview` (including dated variants) to prevent 400 errors on OpenAI and Azure. Addresses one failure mode in #11420.

- **Bug Fixes**
  - Added `openai_model_rejects_reasoning_effort()` deny-list for `o1-mini` and `o1-preview` (prefix match). Name-only to avoid 400s behind OpenAI-compatible gateways; Azure can misreport support.
  - Updated `LitellmLLM._completion` to skip both `reasoning` and `reasoning_effort` for these models on all routes (chat and `/responses`).
  - Added unit tests for both routes and a guard that `o1` still receives `reasoning_effort`.

<sup>Written for commit 64cd9c9dedbf68def06fa2c55dc729fc76dde480. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13334?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/llm/model_capabilities.py
+++ b/backend/onyx/llm/model_capabilities.py
@@ -419,6 +419,12 @@ _OPENAI_MODELS_REJECTING_REASONING_EFFORT = ("o1-mini", "o1-preview")
 
 
 def openai_model_rejects_reasoning_effort(model_name: str) -> bool:
+    """Deliberately name-only — no provider guard. These names are OpenAI
+    models wherever they're hosted (Azure, LiteLLM proxy, OpenAI-compatible
+    gateways), and a provider guard would reintroduce the 400 behind
+    gateways. The asymmetry favors matching broadly: a false positive only
+    omits an optional parameter, a false negative is a hard request failure.
+    """
     base_model_name = model_name.lower().split("/")[-1]
     return base_model_name.startswith(_OPENAI_MODELS_REJECTING_REASONING_EFFORT)
 
--- a/backend/onyx/llm/model_capabilities.py
+++ b/backend/onyx/llm/model_capabilities.py
@@ -410,6 +410,19 @@ def model_is_reasoning_model(model_name: str, model_provider: str) -> bool:
         return False
 
 
+# OpenAI models that reject the reasoning-effort parameter on every API surface
+# (chat completions and responses alike) — only their default effort works.
+# Explicit list rather than a registry lookup: LiteLLM's Azure config wrongly
+# claims support for these, and its OpenAI answer is an accident of the models
+# missing from the registry.
+_OPENAI_MODELS_REJECTING_REASONING_EFFORT = ("o1-mini", "o1-preview")
+
+
+def openai_model_rejects_reasoning_effort(model_name: str) -> bool:
+    base_model_name = model_name.lower().split("/")[-1]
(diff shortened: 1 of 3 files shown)


jmelahman:
Empirical verification against the live OpenAI API (2026-07-22):

| Call | Result |
|---|---|
| `o1-mini` + `reasoning_effort` (chat completions) | `404 model_not_found` — retired on OpenAI |
| `o1-preview` + `reasoning_effort` | `404 model_not_found` — retired on OpenAI |
| `o1-mini-2024-09-12` + `reasoning_effort` | `404 model_not_found` |
| `gpt-5-mini` + `reasoning_effort: "low"` | `200 OK` |

Two takeaways:

1. The `gpt-5-mini` success confirms the deny-list doesn't over-match — modern models keep receiving the effort param.
2. These legacy models are no longer callable on OpenAI itself, so the practical scope of this fix is **Azure deployments**, where retirement lags (Azure Government especially) and where #11420's traceback shows a live deployment returning exactly the 400 this change prevents. The rejection behavior there is evidenced by that traceback rather than reproducible from our side (no Azure Gov access).

justin-tahara (approved):

PR #13331, opened 2026-07-22
Fix form-encoded MCP OAuth refresh responses

jack-watroba:
## What

Fix MCP OAuth refresh handling for providers that return form-encoded token responses.

Key changes:

- Parse `application/x-www-form-urlencoded` OAuth token responses in addition to JSON.
- Persist rotated refresh tokens through the existing `OnyxTokenStorage` path.
- Add regression coverage for GitHub-style refresh responses.
- Preserve existing refresh locking, error handling, and secret-free logging.

## Why

GitHub returned a successful form-encoded refresh response, but Onyx rejected it because the refresh handler only supported JSON. The rotated refresh token was not persisted, causing later requests to reuse an invalid refresh token and fail with `bad_refresh_token`.

## How to Review

Review the response parsing and persistence flow in:

- `backend/onyx/server/features/mcp/oauth.py`

Review the regression coverage in:

- `backend/tests/unit/onyx/server/features/mcp/test_mcp_oauth_refresh.py`

Pay particular attention to:

- JSON and form-encoded response compatibility.
- Persistence of replacement refresh tokens.
- Expiry and authorization-header updates.
- Existing single-flight refresh behavior.
- Secret-free logging.

## Testing

Passed:

- 36 focused and related MCP OAuth tests.
- Ruff checks for the changed files.
- Ty type checks for the changed files.
- `git diff --check`.

The behavior was also manually verified against a local GitHub MCP connection by forcing the stored token expiry and confirming a successful refresh with a rotated refresh token.

## Risks / Impact

- No API, schema, migration, or dependency changes.
- JSON-based OAuth providers are unaffected.
- Providers returning form-encoded token responses now refresh correctly.
- Users whose refresh tokens were already invalidated by the previous behavior may need to reauthenticate once.
- Existing refresh locking and failure behavior remain unchanged.

## Related
- [GitHub MCP OAuth refresh fails with rotating refresh tokens](https://github.com/onyx-dot-app/onyx/issues/13330)

---
## Summary by cubic
Fixes MCP OAuth refresh for providers that return form-encoded token responses, including cases where the `Content-Type` is missing or mislabeled. Rotated refresh tokens are now persisted, and refresh works reliably with GitHub-style flows.

- **Bug Fixes**
  - Parse token responses as JSON or form-encoded based on body, not headers; handle mislabeled or absent `Content-Type`.
  - Persist rotated `refresh_token` via `OnyxTokenStorage`, update `Authorization` header and token expiry, and log a safe “persisted” event.
[...]

PR #13336, opened 2026-07-22
fix: form-encoded MCP OAuth refresh responses [edit of #13331]

evan-onyx:
Based off #13331 for edits and CI

- [x] Override Linear Check

The original will be closed after this one is merged

---
## Summary by cubic
Fixes MCP OAuth refresh to accept form-encoded and mislabeled responses, and adds safe, structured telemetry around each refresh attempt. Improves token rotation reliability and avoids leaking secrets in logs. Addresses issue #13331.

- **Bug Fixes**
  - Parse both JSON and `application/x-www-form-urlencoded` token responses; fall back to form when JSON validation fails.
  - Extract and log only safe OAuth error codes; clear tokens on failure so callers re-auth instead of retrying a dead token.
  - Handle incorrect or missing content-type headers and log body format and status.

- **New Features**
  - Replaces the SDK provider with `OnyxOAuthClientProvider`, emitting `mcp_oauth.refresh.started/failed/invalid_response/succeeded/persisted` with per-attempt correlation IDs.
  - Logs context (server ID/name, transport, connection config, request ID) and token endpoint hostname without tokens or client secrets.
  - Persists rotated tokens/expiry and records whether the refresh token changed; links persistence logs to the same attempt ID.

<sup>Written for commit 63d9acae69eaf3fb55e748dd34af87e2b3566294. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13336?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/server/features/mcp/oauth.py
+++ b/backend/onyx/server/features/mcp/oauth.py
@@ -9,8 +9,10 @@ Routes and route-only flow helpers stay in `api.py`.
 import asyncio
 import json
 import time
-from typing import Any
-from urllib.parse import urlparse
+from collections.abc import Awaitable, Callable
+from typing import Any, TypedDict
+from urllib.parse import parse_qs, urlparse
+from uuid import uuid4
 
 import httpx
 from mcp.client.auth import OAuthClientProvider, TokenStorage
@@ -21,7 +23,7 @@ from mcp.shared.auth import (
     OAuthMetadata,
     OAuthToken,
 )
-from pydantic import AnyUrl, BaseModel
+from pydantic import AnyUrl, BaseModel, ValidationError
 from sqlalchemy.orm import Session
 
 from onyx.cache.interface import CacheLockAcquisitionError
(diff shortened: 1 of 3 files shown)


backend/onyx/server/features/mcp/oauth.py
```diff
@@ -399,6 +501,110 @@ async def set_client_info(  # ty: ignore[invalid-method-override]
...
+            http_status=response.status_code,
+            response_content_type=content_type,
+            response_body_format=body_format,
+            oauth_error=oauth_error,
+            token_endpoint_hostname=token_endpoint_hostname,
+        )
+
+        if response.status_code != 200:
+            logger.warning("mcp_oauth.refresh.failed", extra=response_fields)
+            self.context.clear_tokens()
```
Bo-Onyx, line 592: **`storage.refresh_attempt_id` is never cleared, so a failed refresh can stamp its id on a later, unrelated `set_tokens` call.**

`set_tokens` logs `refresh.persisted` whenever the field is non-`None`, but `set_tokens` is also called by the full re-auth flow. In the streamable-HTTP path, a failed refresh leaves the id set, the SDK falls back to full re-auth, and the re-auth's token write then logs `refresh.persisted` under the failed attempt's id — so the logs read `started → failed → persisted` for the same attempt.

Suggested fix — clear it in both failure branches of `_handle_refresh_response`:

```python
if isinstance(self.context.storage, OnyxTokenStorage):
    self.context.storage.refresh_attempt_id = None
```

Bo-Onyx (approved):

PR #13338, opened 2026-07-22
fix(db): make external app downgrade resilient

wenxi-onyx:
## Description

Make the latest external-app Alembic revision downgrade cleanly when an app has zero or multiple associated skills.

Apps without a skill receive an organization-viewable, provider-named custom placeholder. Its synthetic bundle metadata satisfies older schema constraints, while `is_valid = false` prevents it from representing a real FileStore bundle. This also keeps the placeholder out of the ancestor migration's built-in cleanup so multi-step rollbacks preserve the external-app row.

Apps with multiple skills retain every skill and FileStore bundle reference while the legacy schema selects the first association ordered by skill UUID. This replaces the previous hard failure and preserves existing skill permissions and bundle data.

## Additional Options

### DIFF
--- a/backend/alembic/versions/ea9771dd828c_associate_external_apps_with_skills.py
+++ b/backend/alembic/versions/ea9771dd828c_associate_external_apps_with_skills.py
@@ -19,6 +19,11 @@ down_revision = "f0ff4d3e69ac"
 branch_labels = None
 depends_on = None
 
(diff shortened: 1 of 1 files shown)


Danelegend (approved):

wenxi-onyx:
@greptile re-review

PR #13346, opened 2026-07-23
fix(auth): refresh OAuth tokens with the SSO provider row's credentials

nmgarza5:
## Description

First PR of the SSO env-var retirement stack (target: all SSO configuration lives on provider rows by `v4.5`).

**Bug:** token refresh only knew the account names `google`/`openid` and always POSTed the env `OAUTH_CLIENT_ID`/`OAUTH_CLIENT_SECRET` (endpoint for `openid` from env `OPENID_CONFIG_URL`). Accounts created through new multi-IdP provider rows (any other name, e.g. `okta`) had no refresh path at all, and the env vars were unremovable.

**Fix:** `_resolve_refresh_context` resolves the SSO provider row matching the account's `oauth_name`. Google rows use the static Google token endpoint, OIDC rows resolve discovery from the row's `openid_config_url`, credentials come from the row's encrypted config. No matching row falls back to the exact legacy env behavior.

Hardening that came out of review:

- A failed row lookup rolls back the shared request session and skips the refresh (POST-then-fail-persist would lose the rotated refresh token on IdPs like Entra).
- Config decryption failure disables refresh for that row with an actionable error, never falls back to env credentials.
- Discovery failures are negative-cached (45s) and locks are per discovery URL, so a hard-down IdP costs one 10s attempt per window instead of one per request and cannot head-of-line block other IdPs.
- Misconfiguration logs at error with `has_endpoint`/`has_client_id`/`has_client_secret` booleans, never values.

## Additional Options

### DIFF
--- a/backend/onyx/auth/oauth_refresher.py
+++ b/backend/onyx/auth/oauth_refresher.py
@@ -2,6 +2,7 @@ import asyncio
 import os
 import time
 import uuid
+from dataclasses import dataclass, field
 from datetime import datetime, timezone
 from typing import Any, Dict, List, Optional, cast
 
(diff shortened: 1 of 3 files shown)


nmgarza5:
@greptile

justin-tahara (approved):

nmgarza5:
@greptile

PR #5444, opened 2025-09-17
fix: sharepoint memory via excel parsing

evan-onyx:
## Description

Fixes https://linear.app/danswer/issue/DAN-2574/excrel-file-size-blowup
This branch was originally used to add memory logging to sharepoint connector to find OOMs. I left the utility function in (commented out), but for convenience this PR now includes the fix for our recent sharepoint OOMs.

There is a known issue with MarkitDown: https://github.com/microsoft/markitdown/issues/183
that causes excel file sizes to explode during parsing. Our previous code handled this, and at the moment the best way to fix this is to revert to the previous approach. We should check back with the issue in a few months to see if we can return to the old approach, since it would be ideal to have a unified approach for office files as we tried to do before.

## How Has This Been Tested?

tested in UI

## Backporting (check the box to trigger backport action)

Note: You have to check that the action passes, otherwise resolve the conflicts manually and tag the patches.

- [ ] This PR should be backported (make sure to check that the backport attempt succeeds)
- [ ] [Optional] Override Linear Check

    

---

## Summary by cubic
Adds detailed memory usage logging across the SharePoint connector and docprocessing pipeline to diagnose OOM issues. Introduces a memory_logger utility to record RSS/VMS and deep object sizes at key points.

- **New Features**
  - Added onyx.utils.memory_logger.log_memory_usage to log RSS/VMS and optional deep object size.
  - Instrumented SharePoint connector: HTTP/SDK downloads, text/image extraction, drive/page discovery and pagination, document conversion, and checkpoint processing.
  - Instrumented docprocessing tasks around batch load and indexing pipeline execution.
  - Throttled logs in large loops (e.g., every N items) to limit noise. No functional behavior changes.

- **Dependencies**
  - Added pympler for deep object size measurement.

PR #5115, opened 2025-07-28
Feat/office parsing markitdown

evan-onyx:
## Description

Parsing Office files using Markitdown for a consistent approach across files and a currently used and maintained library.

## How Has This Been Tested?

integration test for parsing docx files with images

## Backporting (check the box to trigger backport action)

Note: You have to check that the action passes, otherwise resolve the conflicts manually and tag the patches.

- [ ] This PR should be backported (make sure to check that the backport attempt succeeds)
- [ ] [Optional] Override Linear Check

PR #13354, opened 2026-07-23
fix: extract cached calculated values from xlsx, not formula strings

### LINKED ISSUE #13353: xlsx extraction indexes formula strings instead of calculated values (load_workbook missing data_only=True)
## Summary

`xlsx_sheet_extraction` loads workbooks with `openpyxl.load_workbook(file, read_only=True)` (`_load_readonly_workbook`, `backend/onyx/file_processing/extract_file_text.py:681` on current main). openpyxl's default is `data_only=False`, which returns the **formula string** for any formula cell. So every computed cell — SUM totals, margins, KPI aggregations — is indexed as `=SUM(C57:E57)` rather than the number Excel displays. The calculated figure never enters the index, and questions about computed values in spreadsheets are unanswerable end-to-end (search returns the formula text; the LLM correctly says it cannot see the value).

Real spreadsheets are dominated by this: in our production deployment (v4.2.0, SharePoint connector, business/financial workbooks), **63% of sampled xlsx chunks contained formula strings, across 18 of 24 indexed xlsx documents**.

## Reproduction

1. In Excel, create a workbook: `A1=2`, `A2=3`, `A3=SUM(A1:A2)`, save.
2. Index it via the File connector (no Unstructured API key configured).
3. Inspect the chunk / search result: it contains `=SUM(A1:A2)`; the value `5` appears nowhere.

## History — this is a silent regression from values

- Between #5115 (2025-08, xlsx via MarkItDown) and #5444 (2025-09, revert), Onyx actually extracted calculated values: MarkItDown uses `pandas.read_excel`, and pandas hardcodes `data_only=True` in its openpyxl reader. The revert to raw openpyxl (done for the legitimate memory reasons in microsoft/markitdown#183) silently reintroduced formula-string extraction; formulas/values are not mentioned in that PR.
- `data_only` has never appeared anywhere in the extractor's history.

## Ecosystem comparison

Every mainstream ingestion path emits values, not formulas: pandas (hardcoded `data_only=True`), and therefore unstructured, MarkItDown, LlamaIndex, LangChain, Haystack; Apache POI/Tika (cached results by default); calamine and SheetJS (formula text structurally not emitted). Two projects shipped this exact bug and fixed it with `data_only=True`: docling (docling-project/docling#2520) and RAGFlow (infiniflow/ragflow#6613).

## Fix and one design decision

Minimal fix: add `data_only=True` in `_load_readonly_workbook`. Caveat: `data_only=True` returns the *cached* value from the file; workbooks written programmatically (openpyxl/pandas, no calc engine) carry no cached values, so their formula cells extract as empty — this is the ecosystem-standard trade-off (see the docling PR, which had to regenerate its own programmatic test fixture for exactly this reason; the same may apply to fixtures in this repo).

A stronger variant, which we run in production: load the workbook twice (`data_only=True` + default) and merge per cell — cached value preferred, formula-string fallback when no cached value exists. Same output as the minimal fix for Excel-authored files, and no information loss for programmatically-generated ones. Happy to contribute either — minimal PR attached.

swissmike-zh:
Fixes #13353.

openpyxl defaults to `data_only=False`, so every formula cell extracts as its formula string (`=SUM(C57:E57)`) instead of the calculated number. Spreadsheet figures — totals, margins, KPIs — therefore never reach the index and are unanswerable in search/chat. In our production index, 63% of sampled xlsx chunks contained formula strings.

`data_only=True` returns the value cached by the application that last saved the file, which is what every mainstream ingestion path emits (pandas hardcodes it; docling and RAGFlow fixed this same bug: docling-project/docling#2520, infiniflow/ragflow#6613). Onyx itself extracted values during the MarkItDown period (#5115) and silently regressed when #5444 reverted to raw openpyxl for memory reasons.

Trade-off (ecosystem-standard): workbooks written programmatically without a calc engine store no cached values, so their formula cells extract as empty instead of the formula string. If any test fixture xlsx is generated by openpyxl and contains formulas, its expected output will need regenerating (docling hit exactly this in their fix). A per-cell fallback (cached value if present, else formula string, via a second `load_workbook` pass) avoids that loss entirely — we run that variant in production and can follow up with it if preferred.

---
## Summary by cubic
Read cached calculated values from .xlsx files instead of formula strings by loading workbooks with data_only=True. This makes totals and other formula-derived numbers indexable and answerable in search/chat.

- **Bug Fixes**
  - Use `openpyxl.load_workbook(file, read_only=True, data_only=True)` so formula cells return cached numbers rather than "=SUM(...)" strings.

<sup>Written for commit 6bd7f0949d80a673e8a368652caafb79288a6546. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13354?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/file_processing/extract_file_text.py
+++ b/backend/onyx/file_processing/extract_file_text.py
@@ -678,7 +678,7 @@ def _load_readonly_workbook(file: IO[Any], file_name: str) -> openpyxl.Workbook 
     / known-openpyxl-bug cases the xlsx indexers treat as skip-and-continue rather
     than a hard failure."""
     try:
-        return openpyxl.load_workbook(file, read_only=True)
+        return openpyxl.load_workbook(file, read_only=True, data_only=True)
     except BadZipFile as e:
         error_str = f"Failed to extract text from {file_name or 'xlsx file'}: {e}"
         if file_name.startswith("~"):


justin-tahara:
Thanks @swissmike-zh — your fix landed via #13536 (adopted as a maintainer PR so CI could run, with credit). Closing this original.

PR #13368, opened 2026-07-23
feat(db): tenant -> database routing seam for multi-shard multitenancy

justin-tahara:
## Description

Adds the seam that lets a tenant's schema live in a database other than the default one. Onyx already addresses a tenant by *schema* via `schema_translate_map`; this adds the orthogonal axis of *which database* that schema is in. It is the prerequisite for incrementally migrating tenants off a single overloaded Postgres instance.

**This is dormant by default and changes no behavior for existing deployments.** With no `ONYX_DB_SHARDS` configured there is exactly one shard, routing short-circuits before any catalog read, and `public.tenant_shard` ships empty. Nothing in this PR ever writes that table — placement is a follow-up.

What's here:

- **`shard_registry.py`** — shard table parsed from `ONYX_DB_SHARDS`, lazily-built engines for non-default shards. The default shard's engine stays owned by `SqlEngine`, so the ~20 existing `init_engine` call sites are untouched. Pool budget is **divided** across shards rather than multiplied, so total connection load on the database/pooler stays bounded.
- **`shard_routing.py`** — `tenant_id -> shard` resolution: static override, then in-process TTL cache, then `public.tenant_shard`, then default. A tenant with no row is on the default shard, which is what lets the table stay empty and avoids a backfill.
- **`shard_version.py`** — cross-process cache invalidation via a Redis version counter. Without it, a map flip leaves every process routing to the database the tenant just moved off — sending *writes* to an abandoned copy — until its TTL expires.
- **DDL routing** — the schema-management primitives and the Alembic URL now follow the same map as data sessions. Previously `setup_tenant` migrated the default database and then seeded through a routed session.
- **`tenant_shard` catalog table** + nullable `available_tenant.shard_name`, via the `schema_private` tree. No backfill.

Two deliberate calls worth reviewing:

- The version counter is read from the Redis **primary, not a replica**, so replica lag stays out of the correctness argument for a flip.
- A Redis failure **does not** invalidate — the local cache is left intact and the TTL is the backstop, so an outage degrades to prior behavior instead of flushing. The consequence is that Redis is a *precondition* for a migration flip, not an optimization, so `bump_shard_map_version()` raises rather than silently succeeding.

## How Has This Been Tested?

17 external-dependency-unit tests in `backend/tests/external_dependency_unit/db/test_shard_routing.py`, run against **two real Postgres databases** rather than mocks — the property that matters is that a session for tenant A physically talks to a different database than one for tenant B, which a mock cannot demonstrate.

Covered: resolution order and fallback, physical database separation, cross-database invisibility of rows, pool-divisor bounding, malformed-config rejection, cross-process invalidation without a local invalidate call (standing in for a migrator in another pod), poll throttling, Redis-failure tolerance, and that DDL + the Alembic URL land on the tenant's shard.

Each behavior was **mutation-tested** — the fix was reverted and the suite confirmed to fail, so the tests are known to be load-bearing rather than merely green. Every mutation was caught by exactly one test, and the expected one.

Single-shard behavior is asserted to be byte-identical: the generated Alembic URL for the default shard equals the historical `build_connection_string()` output.

## Additional Options

---
## Summary by cubic
Adds tenant-to-database routing so each tenant schema can live on a different Postgres database. Keeps single-database deployments unchanged by default.

- **New Features**
  - Shard registry with lazy engines; default shard stays owned by `SqlEngine`. Build-and-store is lock-protected with fast-path reads. Pool budget is split via `divide_pool_budget`. IAM tokens are minted per shard.
  - Async path routes by shard with per-shard async engines. `get_async_session` uses the tenant’s shard, `reset_sqlalchemy_async_engine` disposes all shard engines, and shard resolution is offloaded from the event loop to avoid stalls.
  - Tenant routing order: static overrides → in-process TTL cache (generation-protected) → `public.tenant_shard` → default. Routing fails closed on catalog errors or unknown shards; only a missing `tenant_shard` table falls back to default. Single-shard deployments skip the catalog (`is_sharded()`).
  - Cross-process cache invalidation via a Redis version counter. Redis outages fall back to the TTL; propagation timing is bounded by the TTL with a safety margin.
  - DDL and Alembic follow the tenant’s shard. `alembic/env.py` reads `ALEMBIC_TARGET_URL_ATTRIBUTE` (not `sqlalchemy.url`) and mints IAM tokens from that URL. New `get_catalog_session` pins shared `public` tables to the catalog DB; shared tables moved to `PublicBase` (access via `get_catalog_session` — tenant-scoped sessions are fine single-shard but wrong once sharded).
  - Schema/config: added `public.tenant_shard` and `public.available_tenant.shard_name`. New env vars `ONYX_DB_SHARDS`, `ONYX_DB_DEFAULT_SHARD`, `ONYX_DB_CATALOG_SHARD`, `ONYX_DB_SHARD_OVERRIDES`, `ONYX_DB_SHARD_MAP_TTL_SECONDS`, `ONYX_DB_SHARD_MAP_VERSION_POLL_SECONDS`. Validates catalog shard and timing; rejects redefining the default shard. Removed `POSTGRES_HOSTS`.
  - Tests: added end-to-end routing tests across two real databases; updated async engine fixture to clear per-shard cache to ensure `NullPool` rebuilds.

- **Migration**
  - Apply migrations to create `public.tenant_shard`.
  - To enable sharding: set `ONYX_DB_SHARDS` (and optionally `ONYX_DB_CATALOG_SHARD`), then insert a row in `public.tenant_shard` for a tenant.
  - After changing a tenant’s shard, call `bump_shard_map_version()` and wait `shard_map_propagation_seconds()` before unfreezing writes.

<sup>Written for commit 3f7f8166ce592479ba152e56191b75f7da9e786c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13368?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


justin-tahara:
### Review follow-ups

Thanks — most of these were real. Summary of what changed and what I deliberately left out.

**Fixed (correctness of the seam):**

- **Alembic ignored the configured URL.** `env.py` builds its own engine from `build_connection_string()`, so per-tenant migrations still ran against the default database. The earlier test asserted the URL *builder* returned the right string but never that Alembic *used* it — it passed while the behavior was broken. Now covered end-to-end by intercepting engine creation inside the real `run_alembic_migrations -> alembic -> env.py` path.
- **Routing now fails closed.** A catalog error was indistinguishable from "no mapping", so a transient failure resolved a migrated tenant to the default shard *and cached it* for the full TTL — sending writes to the database it had moved off. Catalog failures and mappings naming an unknown shard now raise. The one surviving fallback is a missing `tenant_shard` table, which is provably safe: if the table does not exist, nothing can be mapped.
- **Freeze window is bounded by the TTL, not the poll interval.** A successful version bump only proves the *migrator* reached Redis. A serving process partitioned from Redis never observes the flip and expires on TTL alone, so the freeze has to outlast the TTL for the guarantee to hold.
- **Cache generation guard** so a lookup in flight during a flip cannot install its stale answer afterwards.
- **Per-shard IAM tokens** in both the registry and the Alembic path — RDS scopes a token to a host/port/user.
- Reject `ONYX_DB_SHARDS` entries redefining the default shard; URL-encode explicit shard passwords; preserve an explicit `max_overflow=0` through the pool divisor; validate the catalog shard and timing settings at startup.

The three P0/P1 fixes are mutation-tested — reverting each one fails the suite.

**Deferred, deliberately:**

- **Async sessions still bind to the default engine**, and **all-tenant Alembic runs enumerate from the catalog database only**. Both are real, both are latent today (nothing is on a non-default shard, and nothing writes `tenant_shard`), and both are *enumeration* problems rather than routing ones. They belong with the shard-aware `get_all_tenant_ids()` work, where they can be tested together, rather than growing this PR further. Flagging here so it is a recorded decision rather than an oversight — placement must not be enabled before that lands, or a tenant on a second shard becomes invisible to beat scheduling.

**Not reproduced:** the claim that `uq_user_tenant` would be dropped from metadata — that constraint is not present in `models.py` on either side of this diff, so nothing was removed.

backend/alembic/env.py
```diff
@@ -55,6 +56,24 @@
...
+
+
+def connection_url() -> str:
+    """Database URL for this migration run.
+
+    Defaults to the process-wide POSTGRES_* settings. A caller that has already
+    decided which database to target — notably per-tenant migrations, which must
+    follow the tenant's shard — passes it via `TARGET_URL_ATTRIBUTE`.
+    """
+    return config.attributes.get(TARGET_URL_ATTRIBUTE) or build_connection_string()
```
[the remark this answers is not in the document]
justin-tahara: Confirmed on both halves, and deferring deliberately — details so it's a recorded decision rather than an oversight.

**The claim holds.** `get_all_tenant_ids()` (`tenant_utils.py:129`) goes through `get_session_with_shared_schema` → the catalog database, and scans *that one database's* `information_schema.schemata`, so tenants on any other shard are invisible. The CLI then spawns `alembic -x schemas=a,b,c` as **subprocesses** (`run_multitenant_migrations.py:66`) — and since `onyx_target_url` is an in-process Config *attribute*, it cannot cross that boundary, so those runs fall through to the default database.

**The fix is three parts:**

1. `env.py` — resolve a shard passed over the `-x` channel (`get_x_argument()` already parses arbitrary `key=value`), keeping `onyx_target_url` as the in-process path and the current default as the fallback.
2. `tenant_utils` — add `get_tenant_ids_by_shard()` scanning each shard, with `get_all_tenant_ids()` retained as the flattened form so existing callers are unaffected. Mid-migration a schema exists on both the old and new shard, so a tenant seen twice must be attributed via `get_shard_for_tenant` — the map is the authority, not whichever scan saw it first.
3. The CLI — group tenants by shard and pass `-x shard=<name>` per batch; batching, `--jobs` and `--continue` semantics unchanged.

**Why not in this PR.** This is the enumeration half of the seam, and the CLI is not the only caller with the blind spot — the same `get_all_tenant_ids()` feeds beat scheduling, monitoring, and several ops scripts. Fixing one caller here would leave the rest half-done while growing an already-large PR, so it lands as a unit in the enumeration follow-up.

It is also latent on this branch: nothing writes `tenant_shard`, so every tenant resolves to the default shard and the CLI is correct as written. It becomes reachable the moment shard-aware placement lands — which is the same change that makes enumeration necessary, so the two ship together. Placement must not be enabled before then, or a tenant on a second shard silently stops being scheduled.

backend/onyx/db/engine/shard_registry.py
```diff
@@ -0,0 +1,333 @@
...
+    """Connection coordinates for one physical database."""
+
+    name: str
+    host: str
+    port: str
+    db: str
+    user: str
+    password: str
+
+    def describe(self) -> str:
```
acaprau: i think when you `print(shard_spec)` it should just print the class name and a memory address so this probably doesn't matter but regardless, if the point is to ensure you obscure passwords when printing this is tedious bc you have to remember to do this on every callsite (see lines 76 and 167 below where this isn't done). overwriting `def __str__(self) -> str:` might be better for this.

backend/onyx/db/engine/shard_registry.py
```diff
@@ -0,0 +1,333 @@
...
+    return len(get_shard_specs()) > 1
+
+
+def divide_pool_budget(pool_size: int, max_overflow: int) -> tuple[int, int]:
+    """Split a connection budget across shards, keeping the total roughly constant.
+
+    Multiplying it per shard would put N times the connection load on the
+    database/pooler, which is how this deployment has hurt itself before. An explicit
+    zero-overflow budget (celery beat) stays zero rather than being floored to 1.
+    """
```
acaprau, line 211: i found it confusing what the second return is for this function, a docstring explaining what it returns would be helpful.

backend/onyx/db/engine/shard_registry.py
```diff
@@ -0,0 +1,333 @@
...
+
+    @classmethod
+    def _build_engine(cls, spec: ShardSpec) -> Engine:
+        """Build a shard engine mirroring the default engine's pool configuration."""
+        from onyx.db.engine.iam_auth import make_provide_iam_token
+        from onyx.db.engine.sql_engine import (
+            SYNC_DB_API,
+            SqlEngine,
+            build_connection_string,
+        )
```
acaprau, line 291: i don't think any of the dynamic imports in this file are necessary but there have been debates on dynamic imports in the past so i guess it's up for debate.

backend/onyx/db/engine/shard_registry.py
```diff
@@ -0,0 +1,333 @@
...
+        profile = SqlEngine.get_engine_profile()
+
+        connection_string = build_connection_string(
+            db_api=SYNC_DB_API,
+            user=spec.user,
+            password=spec.password,
+            host=spec.host,
+            port=spec.port,
+            db=spec.db,
+            app_name=f"{SqlEngine.get_app_name()}_sync_{spec.name}",
```
acaprau: nit, magic string. would be nice if we had a free function to generate this, especially if there are other places that construct it. if this string is just cosmetic e.g. only used in prints then it doesn't rly matter.

backend/onyx/db/engine/shard_version.py
```diff
@@ -0,0 +1,187 @@
...
+    _lock = threading.Lock()
+    _last_polled_at: float | None = None
+    _last_seen_version: str | None = None
+    # None = never polled. Distinct from False so the *first* failure still logs;
+    # starting at False would silently swallow an outage present from boot.
+    _redis_healthy: bool | None = None
+
+    @classmethod
+    def _read_version(cls) -> str | None:
+        # Imported lazily: this module sits under the request hot path and the
```
acaprau: this doesn't make sense to me. the fact that it's on the hot path and lazily imported means the first request on a fresh process is going to suck, vs just it takes a little longer to start the app. i lose hair over lazy imports, the only world where this helps is if there's a deployment that doesn't use redis at all, in which case i'd be surprised if this module is imported at all, or if it is, really it should be gated more intelligently like `if I_USE_REDIS: import redis_nonsense` at the top of a file in a good spot in our code, although ofc that's not in scope for this pr.

backend/onyx/db/engine/sql_engine.py
```diff
@@ -257,8 +278,13 @@ def init_engine(
...
                     del final_engine_kwargs["max_overflow"]
             else:
-                final_engine_kwargs["pool_size"] = pool_size
-                final_engine_kwargs["max_overflow"] = max_overflow
+                # Split the connection budget across shards rather than letting each
+                # shard add a full pool. With one shard the divisor is 1, i.e. sizing
+                # is unchanged from single-database deployments.
+                (
+                    final_engine_kwargs["pool_size"],
+                    final_engine_kwargs["max_overflow"],
```
acaprau, line 286: i see, `max_overflow` is just a connection pool concept. still, a docstring explaining this on `divide_pool_budget` would be nice.

backend/onyx/db/engine/sql_engine.py
```diff
@@ -417,13 +462,41 @@ def get_session_with_current_tenant_if_none(
...
-def get_session_with_shared_schema() -> Generator[Session, None, None]:
+def get_catalog_session() -> Generator[Session, None, None]:
+    """Session against the shared `public` catalog tables.
+
+    Always bound to the catalog database, never to the current tenant's shard —
+    `user_tenant_mapping`, `tenant_shard` and friends exist in exactly one place.
+    """
+    engine = get_catalog_engine()
+
     token = CURRENT_TENANT_ID_CONTEXTVAR.set(POSTGRES_DEFAULT_SCHEMA)
```
acaprau, line 474: sorry i know your pr doesn't touch this at all but why do we do this? this assumes that the thread calling `get_catalog_session` never deals with the concept of a tenant during the lifetime of the db session, which would be a really nice property if our code adhered to this but i doubt it does because there is nothing enforcing this as far as i am aware. setting this contextvar to public to me just seems like we're exposing ourselves to annoying bugs, as the comment below addresses. in fact, i would even split that `finally:` thing up into its own bug fix pr.

backend/onyx/db/engine/sql_engine.py
```diff
@@ -417,13 +461,41 @@ def get_session_with_current_tenant_if_none(
...
+            finally:
+                _safe_close_session(session)
+    finally:
+        # Must be in a finally: an exception inside the block previously leaked the
+        # token, leaving the contextvar pinned to `public` for the rest of the task.
+        CURRENT_TENANT_ID_CONTEXTVAR.reset(token)
+
+
+# Deprecated alias. Prefer `get_catalog_session`, which names what it actually does.
+get_session_with_shared_schema = get_catalog_session
```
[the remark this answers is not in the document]
acaprau: huh?

backend/onyx/db/models.py
```diff
@@ -5294,33 +5187,33 @@ class UserFile(Base):
...
+    database. They are deliberately kept off `Base` so that:
+
+      - the per-tenant Alembic tree (whose target_metadata is `Base.metadata`) never
+        considers creating them inside a tenant schema, and
+      - the `alembic_tenants` tree (target_metadata `PublicBase.metadata`) can actually
+        autogenerate against them.
+
+    Access these only through `get_catalog_session()`. Reading them from a
+    tenant-scoped session resolves against whichever database that tenant lives on,
+    which is the wrong database once tenants are sharded.
```
acaprau: for my own understanding, it's not necessarily the wrong db is it? in a deployment with no sharding the public schema is on the same db as all other schemas right? are we assuming that will not be the case in the future?

deployment/docker_compose/env.template
```diff
@@ -114,6 +114,31 @@ POSTGRES_PASSWORD=password
 # POSTGRES_TCP_KEEPALIVES_IDLE=
 # POSTGRES_TCP_KEEPALIVES_INTERVAL=
 # POSTGRES_TCP_KEEPALIVES_COUNT=
+# --- Tenant sharding across multiple databases (multi-tenant cloud only) ---
```
acaprau, line 117: eh not sure how helpful this is, don't think anyone running a docker compose instance should be thinking about this, really i don't think anyone other than onyx devs should have to think about tenant db sharding, but i don't feel too strongly abt this, can keep if you think it's helpful.

backend/ee/onyx/server/tenants/schema_management.py
```diff
@@ -24,7 +42,11 @@ def run_alembic_migrations(schema_name: str) -> None:
 
         # Configure Alembic
         alembic_cfg = Config(alembic_ini_path)
-        alembic_cfg.set_main_option("sqlalchemy.url", build_connection_string())
+        # Pin the run to the tenant's shard. Uses env.py's dedicated attribute rather
+        # than `sqlalchemy.url`, which env.py ignores by design.
+        alembic_cfg.attributes["onyx_target_url"] = _tenant_connection_string(
```
acaprau: magic string. can we use `TARGET_URL_ATTRIBUTE = "onyx_target_url"` here?

backend/onyx/db/engine/async_sql_engine.py
```diff
@@ -29,80 +37,131 @@
...
-            db_api=ASYNC_DB_API,
-            use_iam_auth=USE_IAM_AUTH,
+# One async engine per shard, created lazily. With no sharding configured this holds
+# exactly one entry and behaves as the previous process-global singleton did.
+_ASYNC_ENGINES: dict[str, AsyncEngine] = {}
+_ASYNC_ENGINES_LOCK = threading.Lock()
+
+
+def _build_async_engine(spec: ShardSpec) -> AsyncEngine:
+    app_name = SqlEngine.get_app_name() + "_async"
```
acaprau: magic string. same comment as the other place, if this string is load-bearing in several places and not just cosmetic it would be nice to have a function to generate it.

backend/onyx/db/engine/shard_routing.py
```diff
@@ -0,0 +1,250 @@
...
+    return _OVERRIDES
+
+
+class _ShardCache:
+    """tenant_id -> (shard_name, expires_at monotonic seconds).
+
+    Entries carry the generation they were resolved under. A catalog read that began
+    before an invalidation must not install its now-stale answer afterwards, which is
+    otherwise a live race during a migration flip: the reader wins, and the tenant
+    stays routable to its old database for a full TTL past the freeze.
```
acaprau: nit this paragraph kinda reads like ai slop, i think i've understood the entire pr by now and i still barely understand what this is trying to say.

backend/onyx/db/engine/shard_version.py
```diff
@@ -0,0 +1,187 @@
...
+Two deliberate choices:
+
+- **Read from the primary, not a replica.** This is one GET per process per poll
+  interval, so the cost is negligible, and it keeps replica lag out of the
+  correctness argument for a flip.
+- **Redis failure does not invalidate.** If the version cannot be read the local
+  cache is left alone and the TTL remains the backstop, i.e. behavior degrades to
+  what it would have been without this module. The consequence is that the
+  migrator must treat Redis as a *precondition* for a flip, not an optimization —
+  see ``shard_map_propagation_seconds``.
```
acaprau: i imagine these questions would be answered in followup prs but just from reading this pr i don't immediately know the answers. what does the migration look like? how do we guarantee we don't do stale reads or writes during the ttl window? does the migration check redis is up each time before doing any operation? i imagine it would have to. if you have a plan.md or something i can read that would work too.

backend/onyx/db/engine/shard_version.py
```diff
@@ -0,0 +1,187 @@
...
+
+    _lock = threading.Lock()
+    _last_polled_at: float | None = None
+    _last_seen_version: str | None = None
+    # None = never polled. Distinct from False so the *first* failure still logs;
+    # starting at False would silently swallow an outage present from boot.
+    _redis_healthy: bool | None = None
+
+    @classmethod
+    def _read_version(cls) -> str | None:
```
acaprau: what does a None return mean for this function?

backend/onyx/db/engine/shard_version.py
```diff
@@ -0,0 +1,187 @@
...
+def reset_shard_map_version_poller() -> None:
+    """Forget the observed version. For tests and forked children."""
+    _VersionPoller.reset()
+
+
+def bump_shard_map_version() -> int:
+    """Signal that the tenant -> shard map changed.
+
+    Must be called by the migrator *after* committing a ``tenant_shard`` flip and
+    *before* unfreezing the tenant. Raises if Redis is unreachable — a flip whose
```
acaprau, line 155: ok nice i think this partially answers my question how do we guarantee that we don't permanently write data to a stale db, a tenant is frozen and not unfrozen until all data is migrated. how does the freezing work?

backend/onyx/db/engine/shard_version.py
```diff
@@ -0,0 +1,187 @@
...
+    logger.info("Bumped shard map version to %s", version)
+    return version
+
+
+def shard_map_propagation_seconds() -> float:
+    """How long to wait after a bump before every process is guaranteed current.
+
+    This is the minimum a tenant must stay frozen after its map entry is flipped.
+
+    It is bounded by the **TTL**, not the poll interval. A successful
```
acaprau, line 169: nice

backend/tests/external_dependency_unit/db/test_shard_routing.py
```diff
@@ -0,0 +1,798 @@
...
+def conn_with_schema(engine: Engine, tenant_id: str) -> Any:
+    return engine.connect().execution_options(schema_translate_map={None: tenant_id})
+
+
+def _current_database(session: Any) -> str:
+    return str(session.execute(text("SELECT current_database()")).scalar())
+
+
+def test_tenants_resolve_to_their_configured_shards(two_shards: dict[str, Any]) -> None:
+    assert get_shard_for_tenant(two_shards["tenant_a"]) == "default"
```
acaprau: i think there's a constant for this string, like DEFAULT_SCHEMA or something like that idk, but we should use that here.

backend/tests/external_dependency_unit/db/test_shard_routing.py
```diff
@@ -0,0 +1,798 @@
...
+
+
+def test_unmapped_tenant_falls_back_to_default_shard(
+    two_shards: dict[str, Any],  # noqa: ARG001
+) -> None:
+    """A tenant with no `tenant_shard` row must land on the default shard.
+
+    This is what lets the mapping table stay empty until tenants are migrated.
+    """
+    assert get_shard_for_tenant(f"tenant_{uuid4()}") == "default"
```
acaprau: same as above. all throughout.

backend/tests/external_dependency_unit/db/test_shard_routing.py
```diff
@@ -0,0 +1,798 @@
...
+        with get_session_with_tenant(tenant_id=two_shards[tenant_key]) as session:
+            markers = session.execute(select(SHARD_PROBE.c.marker)).scalars().all()
+        # Reading only its own marker proves both the schema and the database are right.
+        assert markers == [expected]
+
+
+def test_catalog_session_ignores_the_current_tenant(two_shards: dict[str, Any]) -> None:
+    """The catalog must be reachable from a tenant on any shard, at the same place."""
+    with get_session_with_tenant(tenant_id=two_shards["tenant_b"]):
+        with get_catalog_session() as catalog:
```
acaprau, line 240: see this is exactly what i mean with my contextvar comment above (again ik ur pr has nothing to do with that) if there's any logic in here that is tenant-related and tries to read that contextvar ur in a bad spot.

acaprau (approved):
just nits and questions about the overall intended design, the key logic in this pr checks out to me.

justin-tahara:
Thanks — pushed `be0b3fd` addressing these. Grouped by what changed vs. what I answered.

### Changed

- **`ShardSpec.describe()` → `__repr__`/`__str__`.** You're right that this was tedious and easy to miss, and worse than I'd framed it: the dataclass-generated `__repr__` *did* include the password, so any traceback or stray `%s` would have leaked it. Overriding both is the safe form.
- **`onyx_target_url` magic string.** Now `ALEMBIC_TARGET_URL_ATTRIBUTE`, shared by both call sites. It lives in `shard_registry` rather than `alembic/env.py` because `env.py` runs migrations at import time.
- **`application_name` magic strings.** Added `shard_app_name()`; the sync and async engines were building it inconsistently. It is load-bearing — it's what attributes connections per shard in `pg_stat_activity`.
- **`divide_pool_budget` docstring.** Now says what both args are and what the tuple is.
- **Lazy redis import.** Removed — you were right. I checked for an import cycle and there isn't one, and the pre-commit lazy-import hook is happy. It was only moving the cost to the first request.
- **`_read_version` returning `None`.** It never did; the annotation was wrong. Now `-> str`, and it raises on a Redis failure so callers can't confuse "unreachable" with "never bumped".
- **Test `"default"` literals.** Now a `DEFAULT_SHARD` constant.
- **The `_ShardCache` paragraph.** Fair — rewritten, and I trimmed several other docstrings that had grown past their usefulness.

### Answered, not changed

- **`env.template`** — I agree nobody running docker-compose should think about sharding, and I tried removing it. The repo's `env drift baseline` pre-commit gate fails if an env var isn't documented there, so it has to stay.
- **`models.py` "wrong db"** — correct, it isn't wrong today. With no sharding, `public` is on the same database as every tenant schema, so a tenant-scoped session reaches it fine. The comment is about what happens once a tenant is on a second shard: a tenant-scoped session then lands on *that* database, where `public.user_tenant_mapping` doesn't exist. I've reworded it to be about the sharded case specifically.
- **`get_catalog_session` setting the contextvar to public** — agreed, and agreed it's out of scope here. Nothing enforces that the caller doesn't touch tenant state for the session's lifetime, so it is exactly the footgun you describe. Worth its own PR; I'd rather not fold a behavioural fix to a shared session helper into this one.
- **Dynamic imports elsewhere in `shard_registry`** — the remaining ones break a genuine `sql_engine` ↔ `shard_registry` cycle (sql_engine imports the registry for pool sizing, the registry needs `SqlEngine` for the default engine and profile). Happy to revisit if you'd rather restructure the two modules.

### On the migration questions

Reasonable to ask, and the answers aren't in this PR because the machinery isn't here yet. Briefly: a flip is freeze → copy → verify → update `tenant_shard` → `bump_shard_map_version()` → wait `shard_map_propagation_seconds()` → unfreeze. Redis is a hard precondition — `bump_shard_map_version()` raises rather than silently succeeding, because a flip whose invalidation can't be published isn't safe to proceed from. The freeze window is bounded by the **TTL**, not the poll interval, since a Redis-partitioned pod never sees the bump and only recovers when its entry expires.

Freezing itself is the part that doesn't exist yet — it lands with the migrator, keyed separately from `GATED_ACCESS` so it can't collide with the tenant-cleanup tooling. I'll write the plan up properly in that PR rather than leave it implied here.

backend/onyx/db/engine/shard_routing.py
```diff
@@ -0,0 +1,242 @@
...
+            f"ONYX_DB_SHARD_OVERRIDES is not valid JSON: {e}"
+        ) from e
+    if not isinstance(raw, dict):
+        raise ShardConfigurationError(
+            "ONYX_DB_SHARD_OVERRIDES must be a JSON object of tenant_id -> shard name"
+        )
+    return {str(k): str(v) for k, v in raw.items()}
+
+
+_OVERRIDES: dict[str, str] | None = None
```
evan-onyx, line 71: i think python dicts are not threadsafe by default, probably just need to use ThreadSafeDict?

evan-onyx (approved):

justin-tahara:
Good instinct to check — but I dug into it and I don't think the swap is right here, for a reason worth writing down.

CPython dict operations are individually atomic, so a single `.get()` / `[k] = v` / `.pop()` on `str` keys doesn't tear. What isn't atomic is **compound** access, and `ThreadSafeDict` doesn't help there — it locks each operation separately, so a `get`-then-build-then-`set` is still wide open between the halves.

That's exactly the shape of the lazy engine caches, so I measured it. 32 threads released from a barrier onto the same not-yet-built shard:

| | engines built | distinct engines |
|---|---|---|
| double-checked lock (as written) | **1** | 1 |
| `ThreadSafeDict`, per-op locking only | **32** | 32 |

31 orphaned connection pools. Making this a `ThreadSafeDict` would look safer and be strictly worse, because the property that matters is build-and-store being atomic *as a unit*, which only the explicit lock gives.

For the specific line you flagged, `_OVERRIDES`: it's built once under `_OVERRIDES_LOCK` and then never mutated in place — only rebound wholesale, under the same lock. Readers are hitting a dict nobody writes to.

The other two are deliberate as well:
- `_ShardCache._entries` — `get()` reads unlocked on purpose (it's the request hot path). `invalidate()` **rebinds** `_entries` rather than emptying it, so a concurrent reader sees the whole old dict or the whole new one, never one mid-clear. Nothing iterates it. The `put()` generation check is a genuine check-then-act, which is why that one *is* under the lock.
- `ShardRegistry._engines` / `_ASYNC_ENGINES` — double-checked as above.

**What I did change:** none of that is obvious from reading it, and you're not going to be the last person to wonder. Pushed `3f7f816` adding a one-line note at each unlocked read stating the invariant, plus a stronger note on the two double-checked locks explaining why the lock has to span build-and-store. Mostly so nobody later "simplifies" it into the 32-engine version.

PR #13373, opened 2026-07-23
refactor(craft): make external app skills first-class

wenxi-onyx:
## Description

Final result video:

https://github.com/user-attachments/assets/86c59079-d4ab-4ceb-a79e-194c714930ee

Make external-app-associated custom skills first-class skill records while keeping built-in provider skills hidden and automatic.

This separates skill management authorization from runtime selection, requires custom skills to be selected and their associated app to be ready before sandbox hydration, preserves normal same-name conflict handling, and removes stale preferences when users disconnect or admins delete an app. It also exposes per-user app dependency state through the skill APIs for the stacked frontend PR.

This is the backend half of a two-PR stack split only for reviewability. It is intended to ship together with the frontend PR.

## Additional Options

---
## Summary by cubic
Make external‑app–associated custom skills first‑class while keeping provider‑owned built‑ins hidden and automatic. Separates management from runtime, surfaces per‑user app readiness in skill APIs, and refreshes affected users on app update, disconnect, or deletion.

- **New Features**
  - Associated custom skills are visible and manageable; provider‑owned built‑ins stay hidden and auto‑managed.
  - Runtime requires user selection and app readiness; provider built‑ins follow readiness without a preference.
  - Skill list/fetch/preview include external app state (external_app_id, name, enabled, ready).
  - New endpoint: DELETE `/apps/{external_app_id}/credentials` disconnects a user, clears related preferences, and refreshes their sandbox.
  - Deleting an app detaches associated custom skills (left disabled), removes provider‑owned skills, clears related preferences, and refreshes affected users.

- **Refactors**
  - Replace `SkillAccessPolicy` with `SkillManagementPolicy`; add `list_runtime_skills_for_user` for sandbox hydration.
  - Add `get_skill_external_app_dependencies`; include dependency state in responses and gate `can_toggle` on readiness.
  - Token refresh terminal failure uses `disconnect_external_app_for_user`, triggers `push_skills_for_users`, and commits.
  - Enforce associated skills must keep org visibility at `VIEWER`.
  - External‑app associations no longer reserve names; per‑user enablement can replace a same‑name skill when selecting an associated custom skill.
  - Use `Skill.is_custom` for clearer built‑in/custom checks.

<sup>Written for commit bd34e6f8ed45fe21cb9f9cddff49ac58b5603e36. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13373?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/db/external_app.py
+++ b/backend/onyx/db/external_app.py
@@ -497,7 +497,7 @@ def delete_external_app(
         )
 
     skills = get_skills_for_external_app(db_session, app.id)
-    custom_skill_ids = [skill.id for skill in skills if skill.built_in_skill_id is None]
+    custom_skill_ids = [skill.id for skill in skills if skill.is_custom]
     if custom_skill_ids:
         db_session.execute(
             delete(UserSkillPreference).where(
@@ -507,7 +507,7 @@ def delete_external_app(
 
     db_session.delete(app)
     for skill in skills:
-        if skill.built_in_skill_id is not None:
+        if not skill.is_custom:
             db_session.delete(skill)
     db_session.flush()
 
--- a/backend/onyx/db/external_app.py
+++ b/backend/onyx/db/external_app.py
@@ -1,4 +1,6 @@
 import re
+from collections.abc import Iterable
+from dataclasses import dataclass
 from typing import Any, cast
 from uuid import UUID
 
@@ -22,6 +24,7 @@ from onyx.db.models import (
     ExternalAppUserCredential,
     Skill,
     User,
+    UserSkillPreference,
 )
 from onyx.db.utils import UNSET, UnsetType, is_set
 from onyx.error_handling.error_codes import OnyxErrorCode
@@ -34,6 +37,14 @@ from onyx.utils.sensitive import SensitiveValue
 logger = setup_logger()
 
 _PLACEHOLDER_RE = re.compile(r"\{([a-zA-Z_][a-zA-Z0-9_]*)\}")
+
+
+@dataclass(frozen=True)
+class SkillExternalAppDependencyState:
+    external_app_id: int
+    name: str
+    enabled: bool
+    ready: bool
 
 
 def _placeholders_in_template(auth_template: dict[str, Any]) -> set[str]:
@@ -175,20 +186,6 @@ def get_skills_for_external_app(
     )
 
 
-def get_first_skill_for_external_app(
-    db_session: Session,
-    external_app_id: int,
-) -> Skill:
-    """Return the first associated skill for transitional bundle operations."""
-    skills = get_skills_for_external_app(db_session, external_app_id)
-    if not skills:
-        raise OnyxError(
-            OnyxErrorCode.NOT_FOUND,
-            f"External app {external_app_id} has no associated skills.",
-        )
-    return skills[0]
-
-
 def get_connectable_apps_for_user(
     db_session: Session,
     user: User,
@@ -206,12 +203,13 @@ def get_connectable_apps_for_user(
     ]
 
 
-def available_external_app_skill_ids_for_user(
+def get_skill_external_app_dependencies(
     db_session: Session,
     user: User,
-) -> list[UUID]:
-    """Return app-backed skill IDs whose credentials are ready for this user."""
-    rows = db_session.execute(
+    skill_ids: Iterable[UUID] | None = None,
+) -> dict[UUID, SkillExternalAppDependencyState]:
+    """Return dependency state for all or selected associated skills."""
+    stmt = (
         select(ExternalApp__Skill.skill_id, ExternalApp, ExternalAppUserCredential)
         .join(
             ExternalApp,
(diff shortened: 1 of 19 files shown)


backend/onyx/db/skill.py
```diff
@@ -360,29 +372,7 @@ def fetch_skill(
...
-            "another skill" if is_external_app_backing else "an external app"
-        )
-        raise OnyxError(
-            OnyxErrorCode.DUPLICATE_RESOURCE,
-            f"The skill name '{skill.name}' is already used by {conflicting_resource}.",
-        )
-
     db_session.add(skill)
     db_session.flush()
     return skill
```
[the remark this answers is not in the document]
wenxi-onyx: This state cannot be created through any supported custom-skill path. `ingest_skill_bundle()` rejects names present in `BUILT_IN_SKILLS`, the editor path enforces the same invariant in `rewrite_custom_bundle_skill_md()`, and `validate_stored_custom_skill()` marks any legacy or out-of-band custom row using a reserved built-in name invalid before runtime fileset assembly. External-app built-in skill IDs come from that same registry. Therefore a custom skill and built-in external-app skill with the same name cannot both reach runtime selection; adding precedence here would duplicate and weaken the reserved-name invariant rather than fix a reachable collision. No code change is needed.

backend/onyx/db/external_app.py
```diff
@@ -36,6 +39,14 @@
 _PLACEHOLDER_RE = re.compile(r"\{([a-zA-Z_][a-zA-Z0-9_]*)\}")
 
 
+@dataclass(frozen=True)
```
wenxi-onyx, line 42: A frozen standard-library dataclass is intentional here. This is a small internal value object assembled from already-typed, trusted SQLAlchemy query results; it is not an API, parsing, or serialization boundary. `BaseModel` would add runtime validation and serialization semantics that no caller needs, while `frozen=True` expresses the actual requirement: immutable query state. This also matches the neighboring internal DB state containers.

wenxi-onyx:
@greptile re-review

backend/onyx/db/external_app.py
```diff
@@ -491,12 +497,19 @@ def delete_external_app(
...
+            delete(UserSkillPreference).where(
+                UserSkillPreference.skill_id.in_(custom_skill_ids)
+            )
+        )
+
     db_session.delete(app)
     for skill in skills:
-        db_session.delete(skill)
+        if skill.built_in_skill_id is not None:
+            db_session.delete(skill)
```
Danelegend, line 511: What about the bundles (if those are still a thing)?

backend/onyx/db/external_app.py
```diff
@@ -225,13 +223,23 @@ def available_external_app_skill_ids_for_user(
             ),
             isouter=True,
         )
-        .where(ExternalApp.enabled.is_(True))
-    ).all()
-    return [
-        skill_id
+    )
+    if skill_ids is not None:
```
Danelegend, line 227: Could skill_ids ever be empty?

backend/onyx/db/external_app.py
```diff
@@ -491,12 +497,19 @@ def delete_external_app(
...
+        db_session.execute(
+            delete(UserSkillPreference).where(
+                UserSkillPreference.skill_id.in_(custom_skill_ids)
+            )
+        )
+
     db_session.delete(app)
     for skill in skills:
-        db_session.delete(skill)
+        if skill.built_in_skill_id is not None:
```
Danelegend: There's quite a few instances of this `skill.built_in_skill_id is / is not None`.

Thoughts on adding an `is_custom() -> bool` property to skill?

backend/onyx/db/skill.py
```diff
@@ -338,15 +315,50 @@ def list_skills(
...
+    Management visibility, user selection, bundle validity, built-in
+    availability, and external-app readiness remain independent inputs.
+    """
+    external_app_dependencies = get_skill_external_app_dependencies(db_session, user)
+    ready_external_app_skill_ids = [
+        skill_id
+        for skill_id, dependency in external_app_dependencies.items()
+        if dependency.ready
+    ]
+    has_external_app_dependency = _has_external_app_dependency()
```
Danelegend: Does this need to be a var

backend/onyx/db/skill.py
```diff
@@ -214,66 +218,39 @@ def _skill_select_with_eager_load(*, order_by_name: bool) -> Select[tuple[Skill]
     return stmt
 
 
-def _skill_select_for_access_policy(
+def _has_external_app_dependency() -> ColumnElement[bool]:
+    return exists().where(ExternalApp__Skill.skill_id == Skill.id)
```
Danelegend: Does this need to be a function?

Danelegend (approved):

backend/onyx/db/external_app.py
wenxi-onyx: When normal skills are deleted, bundles are also cleaned up (this is db func only).

This line specific is for skills associated with built-in apps. The "bundle" is materialized from files in the image, not stored in filestore.

backend/onyx/db/external_app.py
wenxi-onyx: If the arg skill_ids is None, this func will return all skills with external app dependencies for the user. If the arg is specified, it will only return the dependencies for those skills.

backend/onyx/db/external_app.py
wenxi-onyx: Will do

backend/onyx/db/skill.py
wenxi-onyx: No, it just makes the sql statement below slightly easier to read. Can flatten, it's not necessary

backend/onyx/db/skill.py
wenxi-onyx: No, it just makes the sql statement callers slightly easier to read. Can flatten, it's not necessary

PR #13378, opened 2026-07-23
feat(llm): per-session reasoning effort control and provider-rejection fallback

nmgarza5:
## Description

Backend for user-controlled reasoning levels, stacked on #13377.

- `ReasoningEffort` gains user-selectable levels (off, low, medium, high, xhigh) with provider maps: OpenAI `reasoning_effort` (incl. xhigh), Anthropic adaptive `output_config.effort` for Claude >= 4.7 or legacy `budget_tokens`, generic litellm clamped to high.
- PUT `/api/chat/update-chat-session-reasoning` validates and stores the override. NULL clears it. Turn setup threads it through the llm loop and deep research.
- Temperature control defaults on (`temperature_override_enabled`).
- Unit tests: parse validation, endpoint contract, retry ladder behavior, effort-map parity.

### Retry ladder

`_completion` retries a provider rejection at most twice, re-sending the identical request with fewer tuning kwargs each time:

1. Attempt 1 sends the full request.
2. Attempt 2 drops the reasoning kwargs. `thinking`, `output_config`, `reasoning`, and `reasoning_effort` are the per-provider spellings of the one reasoning setting, and a request only carries its provider's spelling.
3. Attempt 3 drops `temperature` as well.

A retry only happens when both gates pass:

- Error type. Only litellm `BadRequestError` (provider 400s) enters the ladder. Timeouts, rate limits, connection errors, and 5xx propagate immediately with zero retries.
- Error content. The 400's message must name a kwarg a later rung would drop (`_KWARG_ERROR_ALIASES` covers provider spellings like `budget_tokens` and `effort`). Unrelated 400s such as context length fail on attempt 1.

Rungs only exist when they remove something, so a request with no reasoning kwargs has at most one fallback, and a request with no tuning kwargs is a single attempt. Semantics-changing kwargs (`messages`, `tools`, `tool_choice`, `response_format`, `parallel_tool_calls`) are never stripped, and neither is `stream_options`.

### Why not litellm's support registry

litellm publishes `get_supported_openai_params` and `get_model_info`, but the registry is wrong exactly where this matters. It still advertises `temperature` as supported for adaptive-thinking Claude models (for example claude-opus-4-8), which Anthropic rejects when thinking is on (BerriAI/litellm#26444). `drop_params` relies on the same data, so it would silently drop nothing and let the 400 through. The known Anthropic case is instead handled up front by a version gate that omits `temperature` for Claude >= 4.7, and the ladder exists as a safety net for registry drift and unknown models.

## Additional Options

### DIFF
--- a/backend/onyx/deep_research/dr_loop.py
+++ b/backend/onyx/deep_research/dr_loop.py
@@ -38,7 +38,7 @@ from onyx.deep_research.utils import (
 )
 from onyx.llm.interfaces import LLM, LLMUserIdentity
 from onyx.llm.model_capabilities import model_is_reasoning_model
-from onyx.llm.models import ToolChoiceOptions
+from onyx.llm.models import ReasoningEffort, ToolChoiceOptions
 from onyx.prompts.deep_research.orchestration_layer import (
     CLARIFICATION_PROMPT,
     FINAL_REPORT_PROMPT,
@@ -111,6 +111,7 @@ def generate_final_report(
     turn_index: int,
     citation_mapping: CitationMapping,
     user_identity: LLMUserIdentity | None,
+    reasoning_effort: ReasoningEffort = ReasoningEffort.AUTO,
     saved_reasoning: str | None = None,
     pre_answer_processing_time: float | None = None,
     all_injected_file_metadata: dict[str, FileToolMetadata] | None = None,
@@ -159,6 +160,7 @@ def generate_final_report(
             tool_definitions=[],
             tool_choice=ToolChoiceOptions.NONE,
             llm=llm,
+            reasoning_effort=reasoning_effort,
             placement=Placement(turn_index=turn_index),
             citation_processor=citation_processor,
             state_container=state_container,
@@ -204,6 +206,7 @@ def run_deep_research_llm_loop(
     custom_agent_prompt: str | None,  # noqa: ARG001
     llm: LLM,
     token_counter: Callable[[str], int],
+    reasoning_effort: ReasoningEffort = ReasoningEffort.AUTO,
     skip_clarification: bool = False,
     user_identity: LLMUserIdentity | None = None,
     chat_session_id: str | None = None,
@@ -283,6 +286,7 @@ def run_deep_research_llm_loop(
                     tool_definitions=get_clarification_tool_definitions(),
                     tool_choice=ToolChoiceOptions.AUTO,
                     llm=llm,
+                    reasoning_effort=reasoning_effort,
                     placement=Placement(turn_index=0),
                     # No citations in this step, it should just pass through all
                     # tokens directly so initialized as an empty citation processor
@@ -343,6 +347,7 @@ def run_deep_research_llm_loop(
                 tool_definitions=[],
                 tool_choice=ToolChoiceOptions.NONE,
                 llm=llm,
+                reasoning_effort=reasoning_effort,
                 placement=Placement(turn_index=0),
                 citation_processor=None,
                 state_container=state_container,
@@ -459,6 +464,7 @@ def run_deep_research_llm_loop(
                         turn_index=report_turn_index,
                         citation_mapping=citation_mapping,
                         user_identity=user_identity,
+                        reasoning_effort=reasoning_effort,
                         pre_answer_processing_time=elapsed_seconds,
                         all_injected_file_metadata=all_injected_file_metadata,
                     )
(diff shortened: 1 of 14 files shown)


backend/onyx/deep_research/dr_loop.py
```diff
@@ -204,6 +206,7 @@ def run_deep_research_llm_loop(
     custom_agent_prompt: str | None,  # noqa: ARG001
     llm: LLM,
     token_counter: Callable[[str], int],
+    reasoning_effort: ReasoningEffort = ReasoningEffort.AUTO,
```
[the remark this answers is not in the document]
nmgarza5: Fixed. reasoning_effort now threads from the orchestration loop through run_research_agent_calls, run_research_agent_call, and generate_intermediate_report to both step calls that hardcoded LOW. An explicit session override applies to sub-agents and intermediate reports, AUTO keeps the tuned LOW default.

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

backend/onyx/llm/models.py
```diff
@@ -31,6 +31,31 @@ class ReasoningEffort(str, Enum):
...
+    # Supported by OpenAI and Anthropic adaptive-thinking models (Claude >= 4.7).
+    # Other provider mappings clamp it to their highest supported effort.
+    XHIGH = "xhigh"
+
+
+# Reasoning-effort values a user may pin per chat session. AUTO is excluded
+# because a cleared override (NULL) already resolves to AUTO.
+USER_SELECTABLE_REASONING_EFFORTS: frozenset[ReasoningEffort] = frozenset(
+    {
+        ReasoningEffort.OFF,
```
justin-tahara, line 43: Should we call this off? 

If there is no reasoning maybe we call it quick or fast or something?

backend/onyx/llm/models.py
```diff
@@ -50,6 +76,7 @@ class ReasoningEffort(str, Enum):
     ReasoningEffort.LOW: 1024,
     ReasoningEffort.MEDIUM: 2048,
     ReasoningEffort.HIGH: 4096,
+    ReasoningEffort.XHIGH: 4096,
```
justin-tahara, line 79: Should this be the same as high?

backend/onyx/llm/multi_llm.py
```diff
@@ -699,6 +729,10 @@ def _completion(
                     ReasoningEffort.HIGH,
                 ]:
                     optional_kwargs["reasoning_effort"] = reasoning_effort.value
+                elif reasoning_effort is ReasoningEffort.XHIGH:
+                    # Provider mappings behind litellm's reasoning_effort are
+                    # uneven (Gemini raises on xhigh), clamp to high.
+                    optional_kwargs["reasoning_effort"] = ReasoningEffort.HIGH.value
```
justin-tahara, line 735: Is this true and the way we want to do this?

backend/onyx/llm/models.py
nmgarza5: I'd keep the stored value as off since it maps straight to OpenAI's reasoning_effort none. For the UI label, the slider is titled Reasoning Level so Off reads as the zero of that scale, but if we want something friendlier it's a one-line label change in the frontend PR.

backend/onyx/llm/models.py
nmgarza5: Yes, intentional. Legacy budget_tokens Claude has no xhigh tier, so xhigh clamps to the highest supported budget like everywhere else, and the UI greys the xhigh stop out for these models so this mapping only hits when a stored xhigh lands on an older model.

backend/onyx/llm/multi_llm.py
nmgarza5: It's true, verified against our pinned litellm: the Gemini mapper raises ValueError on xhigh (vertex_and_google_ai_studio_gemini.py:865) even though the completion signature advertises it. Clamping the known case up front keeps the retry ladder as a safety net for unknown drift, same split as the Anthropic temperature version gate.

backend/onyx/llm/models.py
nmgarza5: i think we can adjust this later if we need to

justin-tahara (approved):

PR #13382, opened 2026-07-23
feat(terraform): t-shirt sizing (small/medium/large) for the AWS stack

justin-tahara:
## Description

Adds a `size` input (`small` | `medium` | `large`, default `medium`) to the `deployment/terraform/modules/aws/onyx` composition that sets coherent defaults for every compute/data-plane knob, each individually overridable:

| Setting | small | medium | large |
|---|---|---|---|
| Main EKS node group | m7i.4xlarge ×1–3 | m7i.4xlarge ×1–5 | m7i.4xlarge ×2–8 |
| Document-index node | m6i.xlarge, 100 GB | m6i.2xlarge, 100 GB | r6i.4xlarge, 512 GB |
| RDS Postgres | db.t4g.large, 64→256 GB | db.t4g.large, 128→512 GB | db.m7g.xlarge, 256→1024 GB |
| ElastiCache Redis | cache.m6g.large | cache.m6g.xlarge | cache.m6g.2xlarge |
| OpenSearch data | r7g.large ×1, 256 GB | r8g.xlarge ×1, 512 GB | r8g.2xlarge ×1, 1 TB (12k IOPS) |
| OpenSearch masters | 3× m7g.medium | 3× m7g.medium | 3× m7g.medium |

Tiers are calibrated from usage across the production fleet we operate (node/pod utilization, RDS CloudWatch peaks, and the OpenSearch domain shapes deployments actually converge on). Notable findings baked in: memory (not CPU) is the binding dimension on EKS; burstable `db.t4g.large` holds until roughly the medium tier (largest deployments peak 48–72% CPU); and every production OpenSearch domain runs 1 data node + 3 m7g.medium masters, not the module's old 3×r8g.large multi-AZ-standby default that nothing used.

Supporting changes:
- `eks` module: expose `main_node_min_size` / `main_node_max_size` and `vespa_node_disk_size_gb` (null keeps prior behavior).
- `postgres` module: `max_storage_gb` (RDS storage autoscaling; 0/null disables) and `storage_type` (gp3 default — old 20 GB gp2 instances had only ~100 baseline IOPS).
- Composition now passes OpenSearch `zone_awareness_enabled` + `ebs_iops` through, and sizes the domain subnet count to the AZ spread (1 subnet for single-node domains).
- README: sizing tables, tier-selection guidance, and migration notes for existing stacks.

## How Has This Been Tested?

- `terraform validate` on root modules instantiating all three tiers, including per-setting override combinations (validated with terraform 1.12 / aws provider ~>5.100).
- Provider-free `terraform plan` evaluation of the eks node-group merge logic verifying overrides land and null overrides preserve the previous node-group defaults.
- Cross-checked tier values against the live fleet so that `size = "medium"` reproduces today's EKS/Redis defaults exactly (no node-group churn on upgrade).

## Additional Options

---
## Summary by cubic
Add a `size` input (`small`/`medium`/`large`, default `medium`) to the AWS `onyx` Terraform module to set fleet‑calibrated defaults for EKS, RDS, ElastiCache, and OpenSearch, with per‑setting overrides and safer create-time behavior. The `small` tier now uses `m7i.2xlarge` for main nodes and skips the index node group by default to align with Helm chart ≥0.8.0 sizing.

- New Features
  - `onyx`: `size` tiers set defaults for EKS nodes (main/index), RDS (`gp3` + storage autoscaling), Redis, and OpenSearch (data/masters, gp3 IOPS/throughput); subnet count now follows zone awareness. Paired with chart ≥0.8.0 sizing; small uses `m7i.2xlarge` and omits the index node group unless `vespa_node_enabled = true`.
  - `eks` module: new `main_node_min_size`, `main_node_max_size`, `vespa_node_enabled`, and `vespa_node_disk_size_gb` (null keeps prior defaults). On create, `desired_size` is set to the resolved `min_size`. Disk override preserves other device mappings.
  - `postgres` module: new `max_storage_gb` (0 disables) and `storage_type` (`gp3` default) with RDS 10% headroom validation; in `onyx`, the autoscaling ceiling is floored at 2× the initial size to avoid validation failures.

- Migration
  - Default `size = "medium"` keeps today’s EKS/Redis shapes; RDS converts to `gp3` and may grow storage (cannot shrink).
  - If you enabled OpenSearch on the old 3× `r8g.large.search` zone-aware default, moving to the new single-node default replaces the domain and loses index data. Pin the old shape (e.g. `opensearch_instance_count = 3`, `opensearch_zone_awareness_enabled = true`, `opensearch_multi_az_with_standby_enabled = true`) or snapshot and plan a restore before changing.
  - The `small` tier no longer creates the index node group; set `vespa_node_enabled = true` to keep it, or leave it disabled when using a managed OpenSearch domain.
  - To disable RDS storage autoscaling, set `postgres_max_storage_gb = 0`.

<sup>Written for commit ef36759bcb5422e649984a765cc13da0a3772e45. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13382?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/deployment/terraform/modules/aws/README.md
+++ b/deployment/terraform/modules/aws/README.md
@@ -137,14 +137,19 @@ its tier.
 were `db.t4g.large` with 20 GB gp2 and no storage autoscaling, `cache.m6g.xlarge`, and a
 3×r8g.large multi-AZ OpenSearch domain. The default `medium` tier keeps the same EKS node
 groups and Redis node type, and grows Postgres storage online (gp2→gp3 conversion is also
-online; storage can never shrink). If you enabled OpenSearch and relied on the old
-defaults, `medium` would reconfigure the domain to a single-node shape — an in-place AWS
-blue/green change that preserves index data but shrinks capacity and drops multi-AZ
-redundancy, so pin the old values explicitly (`opensearch_instance_count = 3`,
+online; storage can never shrink).
+
+⚠️ If you enabled OpenSearch and relied on the old defaults, applying `medium` **replaces
+the domain and loses its index data**: a single-AZ domain must live in exactly one subnet,
+and the Terraform AWS provider marks `vpc_options` as ForceNew, so the 3-subnet → 1-subnet
+change destroys and recreates the domain (capacity-only changes that don't touch subnets —
+instance type/count, masters, EBS — are in-place blue/green updates). To keep the old
+topology, pin it explicitly: `opensearch_instance_count = 3`,
 `opensearch_zone_awareness_enabled = true`, `opensearch_multi_az_with_standby_enabled =
 true`, `opensearch_dedicated_master_type = "m7g.large.search"`,
-`opensearch_instance_type = "r8g.large.search"`) if you want to keep the old topology.
-Review the plan before applying either way.
+`opensearch_instance_type = "r8g.large.search"`. To adopt the new shape on an existing
+domain, take a manual snapshot first and plan a restore. Either way, check `terraform
+plan` for `-/+ destroy and then create replacement` on the domain before applying.
 
 ### Using an existing VPC
 If you already have a VPC and subnets, disable VPC creation and provide IDs, CIDR, and the ID of the existing S3 gateway endpoint in that VPC:
--- a/deployment/terraform/modules/aws/README.md
+++ b/deployment/terraform/modules/aws/README.md
@@ -107,23 +107,32 @@ What each tier provisions:
 
 | Setting | small | medium | large |
 |---|---|---|---|
-| Main EKS node group | m7i.4xlarge ×1–3 | m7i.4xlarge ×1–5 | m7i.4xlarge ×2–8 |
-| Document-index node¹ | m6i.xlarge, 100 GB | m6i.2xlarge, 100 GB | r6i.4xlarge, 512 GB |
+| Main EKS node group | m7i.2xlarge ×1–3³ | m7i.4xlarge ×1–5 | m7i.4xlarge ×2–8 |
+| Document-index node¹ | none³ | m6i.2xlarge, 100 GB | r6i.4xlarge, 512 GB |
 | RDS Postgres | db.t4g.large, 64→256 GB | db.t4g.large, 128→512 GB | db.m7g.xlarge, 256→1024 GB |
 | ElastiCache Redis | cache.m6g.large | cache.m6g.xlarge | cache.m6g.2xlarge |
 | OpenSearch data² | r7g.large.search ×1, 256 GB | r8g.xlarge.search ×1, 512 GB | r8g.2xlarge.search ×1, 1 TB (12k IOPS) |
 | OpenSearch masters² | 3× m7g.medium.search | 3× m7g.medium.search | 3× m7g.medium.search |
 
+Pair each tier with the matching sizing snippets from the Helm chart's
+`deployment/helm/charts/onyx/SIZING.md` (chart ≥ 0.8.0) — the tiers here size the
+infrastructure, the chart snippets size the workloads on it.
+
 ¹ The dedicated index node group only matters when running the document index in-cluster
-(the Helm chart's bundled OpenSearch StatefulSet, which is pinned to this node group). If
-your deployment instead points the chart at a managed OpenSearch domain (provision one
-with `enable_opensearch = true` *and* disable the chart's bundled OpenSearch / set the
-managed endpoint in its values), the index node group sits idle — override
-`vespa_node_instance_types` to a small instance in that case.
+(the Helm chart's bundled OpenSearch StatefulSet). It is created tainted
+(`vespa-dedicated=true`), so the StatefulSet must carry the toleration *and* nodeSelector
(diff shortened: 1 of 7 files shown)


justin-tahara:
Triaged all six review findings; fixes pushed in 03df153f5a. Summary:

**Fixed (real):**
- **`desired_size` vs `min_size` (cubic P1)** — correct. The upstream EKS module defaults `desired_size` to 1, so a fresh `size = "large"` create would send min=2/desired=1 and be rejected by the EKS API. Now set `desired_size` to the resolved min (explicit map values still win). Post-create desired changes remain ignored by the upstream module, so existing node groups see no diff.
- **Postgres storage override vs tier ceiling (greptile #2)** — correct. The resolved autoscaling ceiling is now floored at 2× the resolved initial storage, so overriding `postgres_storage_gb` alone can't trip validation.
- **RDS 10% headroom (cubic)** — correct; module validation now enforces `max >= ceil(1.1 × storage)`.
- **`xvda` merge robustness (cubic P2)** — the disk override now merges into existing `block_device_mappings` and tolerates a missing `xvda` entry.
- **README nits (cubic P2 ×2)** — table now uses exact `.search` instance strings, and the idle-index-node note spells out that the chart's bundled OpenSearch must also be disabled/pointed at the managed domain.

**Not fixed (claims don't hold):**
- **"gp3 throughput 1024 MiB/s exceeds OpenSearch's 1000 max" (greptile #1 + cubic P1)** — the 1000 MiB/s figure is the EC2 EBS gp3 limit, not OpenSearch Service's. Amazon OpenSearch allows higher provisioned throughput on larger instances, and we have production domains running exactly the large-tier combination today: `r8g.2xlarge.search` / gp3 / 12000 IOPS / **1024 MiB/s** (verified via `describe-domain` this morning). Keeping 1024.
- **"Default upgrade replaces the domain and loses index data" (greptile #3)** — topology/subnet changes on `aws_opensearch_domain` are in-place updates executed as AWS blue/green deployments, not Terraform resource replacement; index data is preserved. It *is* a capacity/redundancy reduction for anyone who relied on the old default shape, which is why the README migration note tells those users to pin the old values — wording tightened to say exactly that (and for what it's worth, zero deployments we operate use the old default shape).

justin-tahara:
Follow-up on the re-review's remaining point — **greptile is right on the subnet-change consequence, and my earlier reply was wrong on that specific claim.** Correction pushed in 5df08d5319.

I verified against the provider source: `aws_opensearch_domain` marks the entire `vpc_options` block `ForceNew: true` ([domain.go](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/opensearch/domain.go), schema at the `vpc_options` definition — the `HasChange("vpc_options")` branch in the update path is unreachable because any change forces replacement first). Since a single-AZ domain must live in exactly one subnet, migrating an old-default (3-subnet zone-aware) domain to the new single-node default necessarily changes the subnet set → **destroy + recreate, index data lost** unless snapshotted. Capacity-only changes that don't touch subnets (instance type/count, masters, EBS) remain in-place blue/green, which is what I over-generalized from.

README migration note now says exactly that: pin the old values to keep the domain, snapshot + plan a restore if adopting the new shape, and check the plan for `-/+ replace` on the domain either way. Also added a warning comment on the subnet-selection local.

The throughput rebuttal stands as-is (1024 MiB/s verified live on an r8g.2xlarge.search domain).

justin-tahara:
Updated for the chart-sizing companion PR #13392 (ef36759bcb) — **#13392 should merge first**; this PR's small tier is sized against chart ≥ 0.8.0:

- **Small tier main nodes: m7i.4xlarge → m7i.2xlarge ×1–3.** With the chart's retuned request defaults plus the SIZING.md small snippets, the full stack (including in-cluster OpenSearch) fits one m7i.2xlarge when Postgres/Redis/S3 are external; with plain chart defaults the cluster-autoscaler settles at two nodes. No failure mode either way — only cost (~$300/mo less at rest than the old small).
- **New `vespa_node_enabled` knob, tier-driven** (small = false, medium/large = true, overridable). At small the dedicated index node group is pointless: the group is created tainted, the small chart config doesn't tolerate it, and the index fits on the main nodes — it would sit 100% idle. Managed-OpenSearch deployments at any tier can set it false to drop the node entirely.
- README now pairs each tier with the chart's SIZING.md snippets and spells out the placement requirements (toleration + nodeSelector) for the in-cluster index to use the dedicated node group at medium/large.

Verified: `terraform validate` across all tiers including `vespa_node_enabled` override combinations, plus a provider-free plan asserting the node-group map contains `[main]` vs `[main, vespa]` as the knob toggles. `size = "medium"` (the default) still reproduces today's node groups exactly.

jmelahman (approved):

PR #13397, opened 2026-07-24
fix(notion): add slim retrieval so pruning streams IDs instead of full content

justin-tahara:
## Description

Fixes ENG-4308

Pruning a Notion connector currently falls back to running the full connector — downloading and materializing every page and block of the workspace just to enumerate document IDs — because `NotionConnector` has no `SlimConnector` implementation. On large workspaces memory grows with workspace size (~500MiB/hr observed in prod) until `celery-worker-heavy` OOMKills; the prune retries and loops every cycle.

**Approach (revised per review):** `is_slim` is routed through the *same* traversal, GitHub/Drive-style — no parallel slim code paths. The enumeration loop, dedup sets, hierarchy-node conditions, and emit/skip predicate are shared by construction:

- `_read_blocks(is_slim)` / `_read_pages_from_database(is_slim)`: identical traversal and API calls; in slim mode text is still *computed* (per-block, transient) but never *retained* — a `has_content` boolean carries the content signal instead of block lists.
- `_read_pages(is_slim)`: at the document-build spot, slim mode yields a content-less `Document` under the **same emit/skip predicate** as the full build (`has_content or title`), so blank/untitled pages prune exactly as they do today.
- `load_from_state` and `retrieve_all_slim_docs` are thin wrappers over one shared driver; the slim wrapper converts `Document` → `SlimDocument` at the edge (same pattern as the GitHub connector).
- The one shared-behavior change: discovered child pages are pushed as **ID batches fetched lazily at pop** instead of eagerly materializing full `NotionPage` objects for the whole workspace. Order-preserving; bounds memory for indexing too.

API-call count is unchanged by design — the block walk is still required for coverage (search misses nested pages; that's why `recursive_index_enabled` exists). The win is that memory is bounded by the traversal frontier instead of workspace size.

## How Has This Been Tested?

- `test_notion_slim_retrieval.py`: mocked workspace incl. pages that search does *not* return (child page, database rows) and a blank+untitled page. Headline test runs **both** `load_from_state` and `retrieve_all_slim_docs` against the same workspace and asserts slim IDs == full-load IDs (both directions — the blank page must be absent from both) and slim hierarchy nodes ⊇ full. Also covers parent/created-at propagation, lazy child fetching, root-page mode, non-recursive mode, and synced-block cycles in slim mode.
- Existing 35 Notion unit tests pass unchanged (incl. the `_read_blocks` order-preservation test — full-mode traversal is byte-identical).
- Smoke-tested the real pruning entrypoint (`extract_ids_from_runnable_connector`): routes through the slim branch, returns expected IDs/parents/nodes/created-at, excludes the blank page.

## Additional Options

### DIFF
--- a/backend/onyx/connectors/notion/connector.py
+++ b/backend/onyx/connectors/notion/connector.py
@@ -101,15 +101,9 @@ class BlockReadOutput(BaseModel):
     blocks: list[NotionBlock]
     child_page_ids: list[str]
     hierarchy_nodes: list[HierarchyNode]
-
-
-class SlimBlockReadOutput(BaseModel):
-    """Output from a slim (ID-only) traversal of a page's block tree."""
-
-    child_page_ids: list[str]
-    hierarchy_nodes: list[HierarchyNode]
-    # True if any child_database block was seen, even if its node was deduped
-    found_child_database: bool = False
+    # Whether any text was found. In slim mode `blocks` is not retained, so this
+    # carries the "would the page produce content" signal for the emit predicate.
+    has_content: bool = False
 
 
 class _ProcessBlock(BaseModel):
@@ -598,7 +592,7 @@ class NotionConnector(LoadConnector, PollConnector, SlimConnector):
         database_id: str,
         database_parent_raw_id: str | None = None,
         database_name: str | None = None,
-        collect_text: bool = True,
+        is_slim: bool = False,
     ) -> BlockReadOutput:
         """Returns blocks, page IDs, and hierarchy nodes from a database.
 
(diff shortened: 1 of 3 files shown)


justin-tahara:
Restructured per review feedback — the parallel slim traversal (`_read_pages_slim` / `_read_blocks_slim`) is gone. `is_slim` now routes through the existing `_read_pages` / `_read_blocks` / `_read_pages_from_database` (same pattern as the GitHub connector: shared loop, slim yields content-less Documents, converted to SlimDocuments at the edge). The only mode-specific behavior left is *retention*: slim computes text transiently for the content signal but never accumulates it.

The three bot findings were all real drift bugs in the parallel-path version and are resolved structurally by the rewrite:

1. **Blank/untitled pages protected from pruning** — slim now uses the exact same emit/skip predicate as the full build (`has_content or title`), so pages that became empty prune exactly as they do with today's full-crawl fallback. Covered by the parity test (blank page asserted absent from both enumerations).
2. **`_data_source_to_database_map` skipped in non-recursive slim** — the early-return that skipped the data-source loop is removed entirely; slim runs the identical loop, so the mapping is always populated.
3. **LIFO seed ordering breaking parent mapping** — `_read_pages_slim` no longer exists; the shared `_read_pages` processes seeds in input order in both modes.

One shared-behavior change worth reviewer attention: `_read_pages` now pushes discovered children as ID batches fetched lazily at pop, instead of eagerly fetching full `NotionPage` objects for the whole subtree. Order-preserving (existing order-preservation test passes), and it bounds indexing memory as well — the eager prefetch was one of the accumulators behind the prod OOMs.

evan-onyx (approved):

PR #13401, opened 2026-07-24
feat(reindex): old-index reclaim beat task (state machine)

Subash-Mohan:
## Description

- Adds the `check-for-old-index-reclaim` beat task that drives each old index `PENDING → SOAKING → DELETING → RECLAIMED` — soak window, new-index health gate, count-verified deletion, and failure → `BLOCKED` with alerting.
- Deletes only the OpenSearch index; the DB `SearchSettings` row is **kept** (marked `RECLAIMED`) as the durable record.
- Deletes the consented not-ported connectors (INVALID always; PAUSED under Active-Only) once the port drains, re-validated at fire time so a reactivated connector is spared.
- Beat fans out per-row tasks to a dedicated `index_reclaim` light-worker queue (heavy deletion off the primary beat); runs on gated tenants too, to free storage.

## How Has This Been Tested?

- External-dependency-unit tests (real Postgres + OpenSearch): full state-machine walk, consent re-validation, failure→BLOCKED, multi-tenant vs single-tenant deletion, and beat fan-out plumbing.

## Additional Options

---
## Summary by cubic
Adds a beat-driven old-index reclamation pipeline that deletes past OpenSearch indices after reindex. `OLD_INDEX_RECLAIM_ENABLED` now defaults to on but is read at startup, so toggling it takes effect only when workers restart. Heavy work runs on the `index_reclaim` queue via a crash-resumable PENDING → SOAKING → DELETING → RECLAIMED state machine with per-row locks and a bounded delete budget.

**New Features**
- Added `check-for-old-index-reclaim` beat (30m) that enqueues per-row `RUN_OLD_INDEX_RECLAIM` tasks; runs on gated tenants.
- Introduced `index_reclaim` queue; wired into the light worker, Helm, and supervisord; exposed `index_reclaim_queue_length` metric.
- Implemented soak window (`OLD_INDEX_RETENTION_HOURS`), new-index health gate, and time-budgeted deletes; deletes only the OpenSearch index and keeps the `SearchSettings` row marked `RECLAIMED`.
- Enabling has no effect yet: no caller sets reclaim intent on a search settings row, so the beat finds nothing to reclaim.

**Bug Fixes**
- Task body re-checks `OLD_INDEX_RECLAIM_ENABLED` so queued tasks no-op if the flag is turned off.
- Closed reactivation race: consented not-ported connectors move to `DELETING` via a single conditional update that re-validates status (`INVALID`/`PAUSED`) and returns only the ids transitioned.

<sup>Written for commit cb4fc8d4805b2f427d6bc53bdabea3616ead1cfc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13401?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


backend/onyx/background/celery/tasks/beat_schedule.py
```diff
@@ -89,6 +89,21 @@
             "work_gated": True,
         },
     },
+    {
+        "name": "check-for-old-index-reclaim",
```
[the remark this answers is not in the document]
Subash-Mohan: Intentional — OLD_INDEX_RECLAIM_ENABLED is an emergency kill switch that stays on in normal operation, so the disabled no-op fan-out is a rare exception rather than steady state; not worth conditionally scheduling the template.

acaprau:
@greptile @cubic

backend/onyx/db/search_settings.py
```diff
@@ -368,6 +368,18 @@ def advance_to_deleting__no_commit(search_settings: SearchSettings) -> bool:
...
 
 
+def advance_to_reclaimed__no_commit(search_settings: SearchSettings) -> bool:
+    """DELETING -> RECLAIMED: the old index's data is gone. Terminal success — the PAST
+    row is KEPT as the durable record (we only delete the OpenSearch index, not the row).
+    No-op returning False unless currently DELETING. Caller commits."""
+    if search_settings.reclaim_status != IndexReclaimStatus.DELETING:
+        return False
+    search_settings.reclaim_status = IndexReclaimStatus.RECLAIMED
+    search_settings.reclaim_attempts = 0
```
acaprau, line 378: what can increment this number above 0? it looks like we just reclaim successfully and we're done no?

backend/onyx/background/celery/tasks/index_reclaim/tasks.py
```diff
@@ -0,0 +1,344 @@
...
+from onyx.document_index.opensearch.client import OpenSearchIndexClient
+from onyx.document_index.opensearch.index_reclaim import (
+    ReclaimOutcome,
+    reclaim_index_data,
+)
+from onyx.redis.redis_pool import get_redis_client
+from shared_configs.configs import MULTI_TENANT
+from shared_configs.contextvars import get_current_tenant_id
+
+_RECLAIM_BEAT_SOFT_TIME_LIMIT = 60 * 5
```
acaprau: please indicate units to this variable like you did for `_DELETE_TIME_BUDGET_S`, where the `_S` implies seconds. same for the other constants in this file.

acaprau (approved):
i agree with greptile and cubic's comments on `OLD_INDEX_RECLAIM_ENABLED`, that should be a more dynamic thing. pls address comments before merging, also i'd recommend explicitly invoking greptile and cubic again after your fixes just to be sure they don't complain about anything else, for some reason on our repo they don't seem to consistently auto-run anymore.

backend/onyx/background/celery/tasks/index_reclaim/tasks.py
```diff
@@ -0,0 +1,344 @@
...
+
+
+def execute_old_index_reclaim(
+    celery_app: Celery, tenant_id: str, search_settings_id: int
+) -> None:
+    """Body of run_old_index_reclaim_task (testable): under the per-row lock, drive one
+    reclaim step for this PAST index. The lock (TTL) makes a duplicate/concurrent
+    dispatch for the same row a no-op and self-heals if a worker dies mid-run."""
+    if not OLD_INDEX_RECLAIM_ENABLED:
+        return
```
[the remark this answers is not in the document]
Subash-Mohan: Not a real staleness bug — os.environ is fixed at process exec, so reading the flag at call time returns the same value; disabling it requires a worker restart either way, and the guard here exists precisely so already-queued tasks no-op once workers come back with the flag off (consistent with USAGE_LIMITS_ENABLED and other deploy-time flags).

backend/onyx/db/search_settings.py
Subash-Mohan: record_failure__no_commit bumps it from the except in run_old_index_reclaim (tasks.py:236), so a transient failure mid-DELETING (e.g. OpenSearch blip) leaves attempts/last_error set, and a later tick can then succeed straight into RECLAIMED. Clearing them here keeps the kept PAST row from permanently showing a stale error.

Subash-Mohan:
@greptile

PR #13405, opened 2026-07-24
feat(reindex): pre-flight consent modal for won't-port connectors

Subash-Mohan:
## Description

- Before a reindex starts, show a confirmation modal listing the connectors whose data won't be carried into the new index (INVALID always; PAUSED only under Active-Only switchover) and will be deleted, with recovery guidance (re-index from the beginning)
- Send the acknowledged connector set to the backend so it can enforce consent (the server recomputes the authoritative set and rejects if it drifted)
- Gate "Apply & Re-index" until connector statuses have loaded so an unresolved/failed fetch can't skip the modal, and surface the server's error detail on failure
- Stacked on #13403. No behavior change until reclaim is enabled; this PR only captures consent in the UI

## How Has This Been Tested?

- Type-checked + oxfmt/oxlint clean; consent-set logic mirrors the backend rule already covered by the reclaim unit suite

## Additional Options

---
## Summary by cubic
Adds a pre-flight consent modal before reindexing. It lists the connectors that won't be ported (INVALID always; PAUSED only for Active-Only) and that reindexing deletes from Onyx along with their indexed documents, and requires confirmation before Apply. The acknowledged set is frozen when the modal opens and sent to the server, which enforces consent and rejects if the set drifted.

- New Features
  - Submit is labeled "Delete & Re-index"; recovery means adding the connectors back after the reindex finishes.
  - Mirrors the backend rule to compute the won't-port set, drops federated entries, and submits it as `acknowledgedWontPortCcPairIds` via `setNewSearchSettings`.
  - Gates "Apply & Re-index" until the first status fetch settles, not just `isLoading` clearing; `useConnectorIndexingStatusWithPagination` now exposes `isValidating`.
  - Keeps the modal open when the request fails and surfaces the server's `detail`; it closes only on success.

- Dependencies
  - Add `index_reclaim` to local worker queues in `.vscode/launch.json`.

<sup>Written for commit 4d0543ecbf903a9f15ee61c532d5ab687a4ccfb1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13405?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


acaprau:
@greptile @cubic

Subash-Mohan:
@greptile

acaprau (approved):

PR #13413, opened 2026-07-24
fix(mcp): stop emitting unauthenticated MCP servers into craft sessions

Danelegend:
## Description

`resolve_craft_mcp_servers` filtered craft-enabled MCP servers by **access** only, so every server a user could see was written into their session's `opencode.json` with `enabled: true` — including servers the sandbox proxy cannot authenticate them against. opencode dials each one, credential injection answers `403 credential_error` on both transport legs (streamable-HTTP `POST`, then the SSE fallback `GET`), and the server lands in opencode as `status: "failed"` with zero tools.

Three things this fixes:

1. **Dead MCP clients.** Unusable servers are no longer emitted. Nothing is lost by dropping them: they surface no tools either way, and `connect_app` is keyed by `external_app_id` against the external-apps catalog, so the agent can neither see such a server nor prompt the user to connect it.
2. **The `authenticated` flag was unusable as a predicate.** It came from `get_user_authenticated_server_ids`, which matches only `MCPConnectionConfig` rows keyed to the user's own email — but admin-managed credentials are stored with `user_email=""`. It therefore read `False` for every ADMIN-performer, `PT_OAUTH`, and `auth_type=NONE` server, all of which the proxy authenticates fine. Filtering on it would have broken working integrations. The new `can_resolve_mcp_credentials` mirrors `MCPServerResolver._resolve_for_server`'s terminal condition (`requires_auth and not headers → block`) by delegating to `resolve_mcp_credentials`, so emission and injection cannot drift.
3. **Admin-managed servers never hot-reloaded.** That flag fed the per-session runtime fingerprint, and since it could never flip for ADMIN-performer servers, an admin connecting one did not restamp any user's running sandbox — the hot-reload from #13293 only ever worked for PER_USER servers. Credential state now rides in **set membership**, which the fingerprint already covers, so connecting/disconnecting reloads the session for every auth performer.

Also emits `oauth: false` per server. Credentials come from the proxy, so opencode's OAuth auto-detection only generates discovery / dynamic-client-registration requests against paths the proxy blocks, plus `needs_auth` reports for servers that work.

`get_user_authenticated_server_ids` is deleted — it had no remaining callers.

### Scope note: this is not the 60s MCP-init fix

This was investigated as the cause of a 60s MCP initialization stall. It is **not** that, and the ticket should not be closed on this PR. Measured against opencode's client:

| Origin behavior | Cost |
| --- | --- |
| Proxy 403 on both legs, reachable host | 0.17s |
| Origin holds the SSE `GET` open | 30s |
| Origin never responds | 60s (2 × 30s `withTimeout`, StreamableHTTP then SSE) |

Servers connect with unbounded concurrency, so the stall is set by the slowest single server, not the sum. A dead server on a *reachable* host costs ~0.17s.

Two related findings, both deliberately out of scope here:

- **`mcp.timeout` is a trap.** opencode passes it into `client.callTool`, not just connect, and craft's approval gate holds a gated `tools/call` for up to `SANDBOX_APPROVAL_WAIT_TIMEOUT_SECONDS` (180). A value short enough to cap a hanging connect would kill approval-gated tool calls. Separately: with it unset, tool calls inherit the MCP SDK's 60s default, so an MCP call the user takes >60s to approve already fails client-side while the gate still waits.
- **`connection_strategy` is `eager`.** `_build_mitm_options()` doesn't override mitmproxy's default, so the proxy establishes the upstream TCP+TLS connection *before* the request hook can block. A blocked request to an unreachable host therefore still burns the full connect timeout and never returns its 403. Verified locally with a 403-everything addon: eager → reachable 0.14s, blackholed hangs >25s; `lazy` → reachable 0.11s, blackholed **0.01s**. `lazy` looks like the real fix for that class of stall, but it changes upstream cert / ALPN mirroring for all sandbox egress, so it belongs in its own PR with a pass over the gate's TLS assumptions.

## How Has This Been Tested?

- `backend/tests/external_dependency_unit/build/test_mcp_config_resolution.py` — 5 passed. Rewrote the credential test as absence → presence → absence with fingerprint assertions across each transition, and added the two cases the old flag would have regressed: an ADMIN-performer server with an admin credential is emitted for a user with no per-user row, and an `auth_type=NONE` server is emitted with nothing stored. Its fixture now uses `NONE` auth so the access/tool-curation tests stay about access and curation.
- `backend/tests/external_dependency_unit/mcp/test_craft_admin_api.py` — 6 passed, after giving the restamp fixture a credential; without one its server is now absent both before and after the delete, which would have made the assertion vacuous.
- `backend/tests/unit/onyx/server/features/build/sandbox/test_opencode_config.py` — 20 passed, updated for the `oauth` key and the dropped model field.
- `pre-commit` clean on all touched files (`ty`, `ruff`, `ruff format`).

`test_session_lifecycle.py` fails locally on a missing `external_app__skill` table, which reproduces identically with these changes stashed — a local schema lag, not this diff.

## Additional Options

---
## Summary by cubic
Stop emitting unauthenticated MCP servers into craft sessions to prevent permanently failed clients. Also batch-load credential state so session resolution avoids N+1 queries.

- **Bug Fixes**
  - Filter craft MCP servers with an auth check that mirrors injection; skip servers the proxy cannot authenticate.
  - Remove the unusable `authenticated` flag; fingerprint now tracks server set + disabled tools only.
  - Ensure admin-managed, `PT_OAUTH`, and no-auth servers are emitted correctly.
  - Emit `oauth: false` per server to disable opencode’s OAuth auto-detection.
  - Delete the old per-user auth lookup and add logging for skipped servers.
  - Update tests and fixtures, including fingerprinting and a query-count check.

- **Refactors**
  - Inline the emission filter into an explicit loop for clarity and skip-logging.
  - Batch-load per-user connection configs and eager-load admin configs to eliminate N+1; resolution now uses a constant number of DB queries across server counts.
  - Extend `resolve_mcp_credentials` with `user_configs` and add `get_user_connection_configs` for bulk loads.

<sup>Written for commit 328b67db334e54798d5f833d1c6bf090ee3fffdd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13413?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/db/mcp.py
+++ b/backend/onyx/db/mcp.py
@@ -1,10 +1,11 @@
 import datetime
+from collections.abc import Mapping
 from typing import cast
 from uuid import UUID
 
 from pydantic import BaseModel, ConfigDict
 from sqlalchemy import Select, and_, delete, select
-from sqlalchemy.orm import Session, aliased
+from sqlalchemy.orm import Session, aliased, selectinload
 from sqlalchemy.orm.attributes import flag_modified
 
 from onyx.db.constants import UNSET, UnsetType
@@ -65,8 +66,15 @@ def get_craft_enabled_mcp_servers(
 ) -> list[MCPServer]:
     """MCP servers an admin has made available to the Craft agent, filtered to
     those ``user`` may use (public / shared / owned). ``None`` skips the access
-    filter — only for host matching before a user is known (proxy claim path)."""
-    stmt = select(MCPServer).where(MCPServer.available_in_craft.is_(True))
+    filter — only for host matching before a user is known (proxy claim path).
+
+    Eager-loads ``admin_connection_config`` so credential resolution across the
+    returned set doesn't lazy-load one row per admin-managed server."""
+    stmt = (
+        select(MCPServer)
+        .where(MCPServer.available_in_craft.is_(True))
+        .options(selectinload(MCPServer.admin_connection_config))
+    )
     if user is not None:
         stmt = _add_mcp_server_access_filter(stmt, user)
     return list(db_session.scalars(stmt).all())
@@ -452,6 +460,8 @@ def resolve_mcp_credentials(
     mcp_server: MCPServer,
     user: User,
     db_session: Session,
+    *,
+    user_configs: Mapping[int, MCPConnectionConfig] | None = None,
 ) -> ResolvedMCPCredentials:
     """Resolve which stored credentials authenticate `user` against `mcp_server`.
 
@@ -461,6 +471,11 @@ def resolve_mcp_credentials(
     - API_TOKEN / OAUTH: the user's own `mcp_connection_config` row for
       PER_USER servers, the admin config row otherwise.
     - NONE: no credentials.
+
+    `user_configs` lets a caller resolving many servers pre-load the per-user
+    rows in one query (see `get_user_connection_configs`) instead of one per
+    server. It must cover every server the caller resolves — a miss reads as no
+    stored credential, not as unknown.
 
     Raises MCPCredentialsError for PT_OAUTH with the anonymous user, who has no
     login OAuth token.
@@ -482,8 +497,10 @@ def resolve_mcp_credentials(
         MCPAuthenticationType.OAUTH,
     ):
         if mcp_server.auth_performer == MCPAuthenticationPerformer.PER_USER:
-            connection_config = get_user_connection_config(
-                mcp_server.id, user.email, db_session
+            connection_config = (
+                user_configs.get(mcp_server.id)
+                if user_configs is not None
+                else get_user_connection_config(mcp_server.id, user.email, db_session)
             )
         else:
             connection_config = mcp_server.admin_connection_config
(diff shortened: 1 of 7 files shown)


backend/onyx/db/mcp.py
```diff
@@ -512,6 +494,29 @@ def resolve_mcp_credentials(
     return ResolvedMCPCredentials(connection_config=None, user_oauth_token=None)
 
 
+def can_resolve_mcp_credentials(
```
wenxi-onyx, line 514: Should this be in db file?

wenxi-onyx (approved):

PR #13425, opened 2026-07-24
fix(license): auto-deliver renewals and warn admins only when action exists

nmgarza5:
## Description

Bug: a renewed license can never reach a self-hosted instance on its own. The proxy license-refresh route rejects expired licenses, and renewal is exactly the expired case. Nothing instance-side ever re-fetches, so every renewal ends in a manual `.lic` upload. Admins with a live subscription are also routed into checkout, which the control plane's duplicate-subscription guard 409s.

Fix:
- Proxy `/license/{tenant_id}` and `/billing-information` accept a signature-valid expired license regardless of age (ownership checks unchanged). These are the routes a lapsed customer needs to renew, open the billing portal, or buy again, so an age cutoff there would leave someone who is still being billed with no self-serve way back in
- Point-of-use reclaim: any request seeing a license expired or expiring within 7 days enqueues one debounced re-claim before gating, so a lapsed instance heals on first use. A 6-hourly beat task backstops idle instances, and deployments without Redis or Celery run the same reclaim from the API server's periodic poller
- Billing page: with a live subscription, an active-subscription card (renewal date, Sync License button) replaces the payment CTA
- A license rejected by the control plane is blocked from re-claim for 24h, keyed to the rejected blob, so a dead credential is not re-sent on every request
- Concurrency: license stores serialize on an advisory lock, the metadata cache is rebuilt from the stored row under a cache write lock, entries expire at the next status boundary, and a deleted license cannot be resurrected by an in-flight publish
- Enforcement middleware re-reads the row before gating and runs its Redis and DB work off the event loop
- Trials are described as trials. A license that runs only to the end of a trial carries `trial_end`, and the expiry banner, the email, and the billing card all say the trial is ending and billing is about to start rather than warning that access is about to be cut off. A trial also stays quiet until its final 3 days, since the paid warning ladder assumes a lead-up longer than a whole trial
- The billing card reads `Trial ends in N days. Payment required on <date>` while a trial is live, and is unchanged otherwise
- Seat counts render from the license rather than the billing snapshot, so the page cannot show a count the instance would refuse to honor
- Publishing the metadata cache no longer defers to a cached entry that claims a later `issued_at` than the stored row. That guard predated deriving the entry from the row under the cache lock, and it was the only way the cache could disagree with the row indefinitely
- A seat update clears the license-sync cooldown, so the follow-up claim that carries the reissued license is not turned away by a sync the admin ran moments earlier

## Background

How this state came about:
- Renewal was designed as key delivery over email plus manual upload. Automatic refresh was never built, so the system's steady state after a subscription renewal was a working payment and a dying license
- The billing UI gated on subscription state, not license state. An expired license with a live subscription fell into the new-purchase flow, which the control plane correctly refuses

What was missed:
- The expired case is the renewal case. Rejecting expired licenses on the refresh route made self-recovery impossible by construction
- No monitoring existed on license expiry or failed refreshes, so the first signal was a customer report

Learnings:
- A credential that expires must be renewable while expired, within an explicit bound, or every renewal becomes a support ticket
- Cached entitlement state must be derived from the persisted row and bounded by its own status boundaries, never trusted past them
- Recovery paths need to be exercised by the product itself (point-of-use reclaim) rather than depending on an admin noticing

## How Has This Been Tested?

- Full backend unit suite green (6.5k tests). The license area carries ~120 new unit tests covering the reclaim scheduler, the store lock discipline, cache publish/delete interleavings, the lite-mode poller registration, and the middleware recheck
- Race guards are proven load-bearing: each one (row re-read under the cache lock, lease fence, store-lock ordering, delete-side lock, lite gate) was reverted and its named test failed, then restored
- Full web suite green (jest 943 tests, tsc, oxlint). The active-subscription card and Sync License flow were verified live in the browser against a dev deployment
- Production license blobs were signature-verified locally against the production public key to confirm the verify-before-persist path
- Concurrency interleavings (older-publisher, delete-vs-publish, lease expiry, clock-sample skew) were independently traced and in one case reproduced by an external review agent before being fixed and tested

### End-to-end, against live services

The reclaim round trip was run for real rather than mocked. A self-hosted instance, a multi-tenant data plane, and a control plane were brought up together against real Postgres, Redis, and OpenSearch 3.6.0 (the version pinned in `docker-compose.prod.yml`), with the instance reaching the control plane over HTTP through the actual proxy router.

- An instance holding a license that expired thirty days ago healed itself: reclaim ran, the proxy forwarded, the control plane reissued, and the new license verified and persisted
- Grace-window boundaries driven end to end at one, eighty-nine, ninety-one, and two hundred days past expiry: accepted inside ninety days, refused outside it
- A license belonging to another tenant was refused at the ownership check rather than served
- Rejection is sticky: once refused, the instance blocks and later attempts short-circuit instead of re-asking
- Ten concurrent stores against live Postgres and Redis. Eight stale writes were discarded, the newest won, and the published cache entry matched the persisted row
- Verified in all three deployment shapes, including the lite overlay that runs neither Redis nor Celery, where reclaim falls to the in-process poller

The rig also caught three control-plane defects on the other side of the wire, fixed in onyx-control-plane#197.

This is now a standing suite rather than a one-off, run by `ods test multitenant`.

## Screenshots

Every state below was captured live against a real Stripe subscription on a self-hosted instance talking to a multi-tenant data plane and control plane.

The trial card and the sales-issued card are the pair worth reading together. A Stripe-billed license renews itself, so it stays quiet until its final days and never asks the admin to do anything. A sales-issued one warns across its whole lead-up, because a person has to go get the replacement.

|  |  |
| --- | --- |
| **Trial billing card**<br><sub>`Trial ends in 2 days. Payment required on Aug 2, 2026`, both halves read from the license.</sub><br>01-trial-billing-card | **Trial banner**<br><sub>Fires inside the final 3 days, framed as billing starting rather than access ending.</sub><br>02-trial-banner |
| **Sales-issued license**<br><sub>No Stripe subscription, so no Sync, no Manage Plan, no Update Seats. Its whole lead-up still warns, because a human has to issue the replacement.</sub><br>03-sales-license-managed-through-sales | **Control plane unreachable**<br><sub>A lapsed instance can still reach billing, and is told why Stripe is unavailable.</sub><br>04-control-plane-unreachable |
| **Seat reduction below usage**<br><sub>Refused against live usage before any Stripe call.</sub><br>05-seat-reduction-below-usage | **Gated instance**<br><sub>`A license is required to use Onyx`, no longer claiming Enterprise specifically.</sub><br>06-gated-access-restricted |
| **Gated, billing still reachable**<br><sub>A customer can always pay their way back in.</sub><br>07-gated-billing-still-reachable | **Plan configurator**<br><sub>Seat minimum derived from current usage.</sub><br>08-plan-configurator |
| **Duplicate subscription guard**<br><sub>Checkout refused for an instance that already subscribes.</sub><br>09-duplicate-subscription-guard | **Recovered**<br><sub>Back to active after the license is restored.</sub><br>10-recovered-after-license-restore |
| **Sync cooldown**<br><sub>Rapid clicks throttled. Action errors are toasts now, not text under the button.</sub><br>11-sync-cooldown-toast | **Seat limit exceeded**<br><sub>Product gated at 842 users against 10 seats.</sub><br>12-seat-limit-exceeded-gate |

## Additional Options


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

justin-tahara (approved):

PR #13427, opened 2026-07-24
feat(skills): import skills from GitHub

wenxi-onyx:
## Description

https://github.com/user-attachments/assets/56c04d48-6d2b-431e-bcc3-7b571afbcb91

Add GitHub repository imports to the Skills page with public and connected-private repository support. Imports are pinned to the previewed revision, package complete skill bundles, auto-enable non-conflicting skills, preserve conflicting imports as disabled, and report partial outcomes clearly. External-app flows direct users to import on the Skills page before associating skills.

## Additional Options

---
## Summary by cubic
Adds GitHub import to the Skills page with repository preview, selective import, and clear results. Works with public repos and connected private repos, pins imports to a commit, and auto-enables unless a conflicting name is already enabled.

- **New Features**
  - Backend: `/api/skills/github/preview` and `/api/skills/github/import` accept owner/repo or `https://github.com/...` (supports `tree/<ref>/subpath>`), resolve to an immutable commit and optional subpath, and build bundles only for selected paths.
  - Private repos: uses the connected GitHub app (token refresh, SSO/permission checks), keeps public requests anonymous, never forwards private tokens to `codeload.github.com`, and returns clear errors for auth expiry, rate limits, and not-found.
  - Safe ingest: 25 MiB archive cap, per-file/total size limits, max 10k entries, no symlinks, duplicate-path checks; finds `SKILL.md`, normalizes bundles, caps at 500 skills, and blocks built-in name collisions.
  - Import behavior: creates skills from the pinned commit, enables when no enabled name conflict, keeps conflicts disabled with reasons, and pushes updates only when something was enabled.
  - UI: new “Import from GitHub” modal on the Skills page (search, select, import, outcome summary). The App editor’s create menu directs GitHub imports to the Skills page.
  - Shared GitHub ingestion: centralized, SSRF-safe archive download/parsing via `codeload.github.com` tarballs with strict commit SHA validation; the coding agent now uses it (size/timeouts enforced, SSH URLs supported).

- **Dependencies**
  - Helm: bump chart to 0.8.4 with Artifact Hub notes for GitHub skill imports.

<sup>Written for commit dac8c67c56c94c7e67e25b2a3b9d124d8f73d757. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13427?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


wenxi-onyx:
@greptile @codex @cubic-dev-ai re-review

backend/onyx/skills/ingest_from_github.py
```diff
@@ -0,0 +1,285 @@
...
+    download_github_archive,
+    parse_github_source,
+    resolve_github_revision,
+)
+
+_ARCHIVE_MAX_BYTES: Final[int] = 25 * 1024 * 1024
+_ARCHIVE_MAX_MEMBERS: Final[int] = 10_000
+
+
+def fetch_github_skill_bundles(
```
Danelegend, line 37: Considering both preview and import use this, and I assume the flow is someone previews then imports, some sort of caching would be nice (5min TTL?)

backend/onyx/skills/ingest_from_github.py
```diff
@@ -0,0 +1,285 @@
...
+    )
+    archive_bytes = download_github_archive(
+        github_source,
+        resolved_source.revision,
+        authorization_header,
+        max_size_bytes=_ARCHIVE_MAX_BYTES,
+    )
+
+    try:
+        archive = tarfile.open(
```
Danelegend: Resource leakage potential if we exit between here and the `with archive`

backend/onyx/skills/ingest_from_github.py
```diff
@@ -0,0 +1,285 @@
...
+        raise OnyxError(
+            OnyxErrorCode.BAD_GATEWAY,
+            "GitHub returned an unreadable repository download. Try again.",
+        ) from exc
+
+    files: dict[str, bytes] = {}
+    total_size = 0
+    try:
+        with archive:
+            for member_index, member in enumerate(archive, start=1):
```
Danelegend, line 69: Why starting at 1? Kind of non-intuitive?

backend/onyx/skills/ingest_from_github.py
```diff
@@ -0,0 +1,285 @@
...
+        github_source,
+        resolved_source.revision,
+        authorization_header,
+        max_size_bytes=_ARCHIVE_MAX_BYTES,
+    )
+
+    try:
+        archive = tarfile.open(
+            fileobj=io.BytesIO(archive_bytes),
+            mode="r:gz",
```
Danelegend: `r:` is seekable so TarInfo is cached for each archive.members. It doesn't look like we need random access to maybe `r|` instead for a streaming approach?

backend/onyx/skills/ingest_from_github.py
```diff
@@ -0,0 +1,285 @@
...
+                        OnyxErrorCode.PAYLOAD_TOO_LARGE,
+                        f"Repository archive contains more than {_ARCHIVE_MAX_MEMBERS:,} entries. Use a smaller repository.",
+                    )
+                member_path = member.name.rstrip("/") if member.isdir() else member.name
+                path_parts = member_path.split("/")
+                if (
+                    not member_path
+                    or member_path.startswith("/")
+                    or "\\" in member_path
+                    or any(part in {"", ".", ".."} for part in path_parts)
```
Danelegend, line 81: `{"", ".", ".."}` use a list instead (faster <255 elems)

backend/onyx/skills/ingest_from_github.py
```diff
@@ -0,0 +1,285 @@
...
+    resolve_github_revision,
+)
+
+_ARCHIVE_MAX_BYTES: Final[int] = 25 * 1024 * 1024
+_ARCHIVE_MAX_MEMBERS: Final[int] = 10_000
+
+
+def fetch_github_skill_bundles(
+    source: str,
+    authorization_header: str | None = None,
```
Danelegend, line 39: This function gets a bit long. Is it possible to break it into some smaller helper functions (ie. the inside of the for loop)?

backend/onyx/skills/ingest_from_github.py
```diff
@@ -0,0 +1,285 @@
...
+                        f"Repository contains duplicate path '{member.name}'.",
+                    )
+                files[member.name] = extracted.read(DEFAULT_PER_FILE_MAX_BYTES + 1)
+    except (tarfile.TarError, EOFError, OSError) as exc:
+        raise OnyxError(
+            OnyxErrorCode.BAD_GATEWAY,
+            "GitHub returned an unreadable repository download. Try again.",
+        ) from exc
+
+    root_directories = {path.split("/", maxsplit=1)[0] for path in files}
```
Danelegend, line 123: I have a feeling lines 129-139 can be done in the for loop? Following the reasoning after adds a bit of unclearness.

backend/onyx/utils/github.py
```diff
@@ -1,87 +1,342 @@
...
+    if not value:
+        raise OnyxError(
+            OnyxErrorCode.INVALID_INPUT,
+            "Enter a GitHub repository URL or owner/repository.",
+        )
 
-    ssh_match = re.match(r"^git@github\.com:([^/]+)/([^/\s]+?)(?:\.git)?$", repo)
+    tree_tail: tuple[str, ...] = ()
+    ssh_match = (
+        re.fullmatch(r"git@github\.com:([^/]+)/([^/\s]+?)(?:\.git)?$", value)
```
Danelegend: Define regex as constant

backend/onyx/utils/github.py
```diff
@@ -1,87 +1,342 @@
...
-GITHUB_DOWNLOAD_CONNECT_TIMEOUT_SECONDS = 30
-GITHUB_DOWNLOAD_READ_TIMEOUT_SECONDS = 300
-GITHUB_DOWNLOAD_CHUNK_SIZE_BYTES = 1024 * 1024  # 1 MiB
-DEFAULT_MAX_TARBALL_SIZE_BYTES = 500 * 1024 * 1024  # 500 MiB
+_GITHUB_FETCH_TIMEOUT_SECONDS: Final[int] = 30
 
 
-def parse_github_repo(repo: str) -> tuple[str, str]:
-    """Parse a GitHub repo identifier into (owner, name).
+class GitHubSource(NamedTuple):
```
Danelegend: Would prefer pydantic model for consistency

backend/onyx/utils/github.py
```diff
@@ -1,87 +1,342 @@
...
+                        "Enter a GitHub repository URL without a branch, file, or subpage."
+                        if not allow_tree_path
+                        else "Enter a repository URL or a URL for a folder within a repository."
+                    ),
+                )
 
-    short_match = re.match(r"^([^/\s]+)/([^/\s]+?)(?:\.git)?$", repo)
-    if short_match:
-        return short_match.group(1), short_match.group(2)
+    if not all(re.fullmatch(r"[A-Za-z0-9_.-]+", part) for part in (owner, repo)):
```
Danelegend: Define as constant - Seems like quite a general one, so maybe util?

backend/onyx/utils/github.py
```diff
@@ -1,87 +1,342 @@
...
+    authorization_header: str | None = None,
+    *,
+    revision: str | None = None,
+    subpath: str | None = None,
+) -> GitHubRevision:
+    """Resolve a repository source to an immutable commit and optional subpath."""
+    normalized_subpath = subpath.rstrip("/") if subpath else None
+    if normalized_subpath and (
+        "\\" in normalized_subpath
+        or any(part in {"", ".", ".."} for part in normalized_subpath.split("/"))
```
Danelegend, line 157: list

backend/onyx/utils/github.py
```diff
@@ -1,87 +1,342 @@
...
+    authorization_header: str | None = None,
+    *,
+    max_size_bytes: int,
+    timeout: float | tuple[float, float] = _GITHUB_FETCH_TIMEOUT_SECONDS,
+) -> bytes:
+    """Download a bounded tarball without forwarding credentials to archive hosts."""
+    encoded_owner = quote(source.owner, safe="")
+    encoded_repo = quote(source.repo, safe="")
+    encoded_revision = quote(revision, safe="")
+    archive_url = (
```
Danelegend: Constant?

backend/onyx/utils/github.py
```diff
@@ -1,87 +1,342 @@
...
+    if response.status_code != 200 and authorization_header is not None:
+        response.close()
+        redirect_response = _github_get(
+            f"https://api.github.com/repos/{encoded_owner}/{encoded_repo}"
+            f"/tarball/{encoded_revision}",
+            authorization=authorization_header,
+            follow_redirects=False,
+        )
+        with redirect_response:
+            if redirect_response.status_code == 401:
```
Danelegend, line 277: switch statement instead?

backend/onyx/utils/github.py
```diff
@@ -1,87 +1,342 @@
 import re
```
Danelegend, line 1: The functions in this file seem a bit large. Let's see if we can break them up into smaller ones as well

deployment/helm/charts/onyx/Chart.yaml
```diff
@@ -5,7 +5,7 @@ home: https://www.onyx.app/
 sources:
```
Danelegend, line 5: Do we stil need this? No values.yaml change¿

Danelegend (approved):
Can't see much functionally wrong, just a bunch of code nits / improvements

backend/onyx/skills/ingest_from_github.py
wenxi-onyx: This is at max 2 calls per github import, so I don't think we gain a lot from caching. Adding caching also means we have to manage a tenant-, perhaps even user-, specific cache in memory (i.e. no cross-tenant leakage, if a private repo, only what the fetching user can access), which I don't think is worth the complexity when the trade off is just 1 extra call.

backend/onyx/skills/ingest_from_github.py
wenxi-onyx: We're counting the number of items. If there is 1 item and we 0-index, it would report "there are 0 items".

backend/onyx/skills/ingest_from_github.py
wenxi-onyx: Cache is bounded by 10k member limit and Codex reports that `r|` also caches TarInfo. While quite low risk, will just add `stream=True` to tarfile.open to achieve the desired streaming.

backend/onyx/skills/ingest_from_github.py
wenxi-onyx: I disagree quite strongly with single-use helper functions. It just obfuscates the actual logic with a semantic function name that could lie/not accurately reflect the actual logic. Additionally, forces you to cmd click the function and lose your place in the code to see what is actually happening.

backend/onyx/skills/ingest_from_github.py
wenxi-onyx:
```
 In Python 3.13:

  - The set literal is compiled once as a constant frozenset; it is not recreated per iteration.
  - The list literal is optimized into a tuple and requires a linear scan.
  - Benchmarking this exact three-string check showed the set faster for first-item hits, last-item hits, and misses.

  The performance difference is immaterial here anyway, but the existing set is both semantically clear and faster in this case. I would not change it.
```

backend/onyx/skills/ingest_from_github.py
wenxi-onyx:
```
Not valid. It is possible to combine this with the extraction loop, but it would make the loop more stateful:

  - Remember the first file’s archive root.
  - Verify every subsequent file has the same root.
  - Strip that root before inserting each file.
  - Handle files without a root component.

  The current post-pass keeps the responsibilities clearer:

  1. Safely extract and validate archive members.
  2. Confirm GitHub returned exactly one archive root.
  3. Remove that GitHub-generated root from every path.
  ```

File: backend/onyx/utils/github.py (hunk 9 above)
Comment (wenxi-onyx) [reply backend/onyx/utils/github.py]:
```
These are internal validated value objects, not API payloads, so BaseModel would add unnecessary Pydantic validation and serialization behavior. A frozen, slotted dataclass is a better fit than both NamedTuple and BaseModel: named construction, immutable fields, no tuple semantics, and minimal overhead.
  ```

backend/onyx/utils/github.py
wenxi-onyx: I think we historically have not used switch statements in codebase. Nothing against them, but not going to churn here for no real benefit

backend/onyx/utils/github.py
wenxi-onyx: Not changing, see above

backend/onyx/utils/github.py
wenxi-onyx: Not changing, see above comment about single use helpers, obfuscation, and redirection

deployment/helm/charts/onyx/Chart.yaml
wenxi-onyx: :ai-slop: fixing

PR #13446, opened 2026-07-27
fix(chat): explain empty attachments to the model instead of silent emptiness

justin-tahara:
## Description

When a user attaches an image-only PDF (e.g. a scanned document) in chat, text extraction yields nothing and the vision captions produced by the user-file worker either don't exist yet (indexing still running) or never will (no vision-capable LLM configured). The model was handed a literally empty file block:

```
File: scan.pdf

End of File
```

with no explanation — and predictably invented workarounds: trying the code interpreter, web-searching for the document by its metadata, or guessing contents. Reported by a customer testing an image-based patent PDF (both Sonnet 4.6 and GPT 5.4 behaved this way).

Two changes:

1. **`build_file_context` explains empty files.** Empty/whitespace-only text files now inject an explicit notice: "still being processed — tell the user to ask again in a moment" while the user-file worker is running (`UserFile.status` PROCESSING/INDEXING), otherwise "no machine-readable text could be extracted... do not guess".
2. **The chat path no longer pre-empts the worker's plaintext cache.** `load_chat_file` extracted inline on cache miss and stored the result under the same key the user-file worker writes its canonical plaintext to (which can include image captions inline extraction can't produce). While the file is still processing, we now extract for the current turn without storing.

Not in scope: actually passing PDF page images to vision-capable chat models (feature gap — the chat LLM only ever sees text for `DOC` files).

## How Has This Been Tested?

- New unit tests: `backend/tests/unit/onyx/chat/test_build_file_context.py`
- `uv run pytest backend/tests/unit/onyx/chat` — 464 passed

## Additional Options

---
## Summary by cubic
Explain empty file attachments in chat so the model stops guessing; instead, it tells users when files are still processing and only reads canonical plaintext once ready. Also prevents caching inline extractions under the worker’s key until processing finishes.

- **Bug Fixes**
  - `build_file_context`: Adds notices for empty/whitespace-only text files — “still processing, ask again soon” while processing; otherwise “no machine-readable text, do not guess.”
  - `load_chat_file`: Checks `UserFile` status, avoids storing inline-extracted plaintext under the worker’s cache key while processing, and threads `content_pending` into the model context.

- **Refactors**
  - Add `get_user_file_by_id` and reuse it across duplicate query sites (`get_file_id_by_user_file_id`, `load_user_file`, `get_user_files`); default `user_file_id` lookup to `""` in chat.

<sup>Written for commit a006efd43a1b97acdcbcbd39ac3bae2f977ccb00. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13446?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Linear: https://linear.app/onyx-app/issue/CS-32

### DIFF
--- a/backend/onyx/chat/chat_utils.py
+++ b/backend/onyx/chat/chat_utils.py
@@ -38,6 +38,7 @@ from onyx.db.models import ChatMessage, ChatSession, Persona, User, UserFile
 from onyx.db.models import SearchDoc as DbSearchDoc
 from onyx.db.persona import user_can_access_persona
 from onyx.db.projects import check_project_ownership
+from onyx.db.user_file import get_user_file_by_id
 from onyx.file_processing.extract_file_text import extract_file_text
 from onyx.file_store.file_store import get_default_file_store
 from onyx.file_store.models import ChatFileType, FileDescriptor
@@ -466,10 +467,7 @@ def load_chat_file(
     user_file: UserFile | None = None
     if user_file_id_str:
         try:
-            user_file_id = UUID(user_file_id_str)
-            user_file = (
-                db_session.query(UserFile).filter(UserFile.id == user_file_id).first()
-            )
+            user_file = get_user_file_by_id(UUID(user_file_id_str), db_session)
         except (ValueError, TypeError) as e:
             logger.warning("Failed to look up user file for %s: %s", file_id, e)
     token_count = user_file.token_count if user_file and user_file.token_count else 0
--- a/backend/onyx/chat/chat_utils.py
+++ b/backend/onyx/chat/chat_utils.py
@@ -463,7 +463,7 @@ def load_chat_file(
 
     # Look up the UserFile row first (when one exists) — it supplies the token
     # count and tells us whether the user-file worker is still processing.
-    user_file_id_str = file_descriptor.get("user_file_id")
+    user_file_id_str = file_descriptor.get("user_file_id", "")
     user_file: UserFile | None = None
     if user_file_id_str:
         try:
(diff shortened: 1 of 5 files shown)


backend/onyx/chat/chat_utils.py
```diff
@@ -426,6 +461,21 @@ def load_chat_file(
     file_type = ChatFileType(file_descriptor["type"])
     filename = file_descriptor.get("name")
 
+    # Look up the UserFile row first (when one exists) — it supplies the token
+    # count and tells us whether the user-file worker is still processing.
+    user_file_id_str = file_descriptor.get("user_file_id")
```
Danelegend: `file_descriptor.get("user_file_id", "")` for stricter  type / none value checking

backend/onyx/db/user_file.py
```diff
@@ -104,6 +104,10 @@ def update_last_accessed_at_for_user_files(
     db_session.commit()
 
 
+def get_user_file_by_id(user_file_id: UUID, db_session: Session) -> UserFile | None:
```
Danelegend: Use this in `db/user_file.py::get_file_id_by_user_file_id` and `file_store/utils.py::get_user_files`

Danelegend (approved):

backend/onyx/chat/chat_utils.py
justin-tahara: Done in a006efd43a — descriptor get now defaults to "".

backend/onyx/db/user_file.py
justin-tahara: Done in a006efd43a — helper reused in get_file_id_by_user_file_id, get_user_files, and load_user_file (same inline query in the same file).

PR #13451, opened 2026-07-27
fix(pruning): short-term — trim enumeration memory + reap Chromium zombies

justin-tahara:
## Description

Short-term fix for the recurring `OnyxContainerMemoryNearLimit` pages on cloud heavy workers ([ENG-4309](https://linear.app/danswer/issue/ENG-4309)). Two small, independent commits; the durable process-isolation fix is split out into #13453 per review feedback that the bundled version was too large.

**1. Periodic memory release during prune enumeration (the OOM fix).** Long enumerations (10k+ page recursive web crawl) churn transient allocations for hours inside the long-lived heavy worker; glibc retains the freed memory, ratcheting RSS ~0.2–0.4Mi/page to the 5Gi limit → OOMKill → re-dispatch → repeat. `gc.collect()` + `malloc_trim(0)` every 30s at batch boundaries keeps it bounded — measured in the deployed `v4.4.0-cloud.6` image on the same crawl: **flat ~233Mi from ~100 pages onward vs an unbounded linear ramp without it**. (Capping `MALLOC_ARENA_MAX` alone also curbs the ramp but converges slower and higher, ~250Mi after ~350 pages; the trim alone is sufficient, so no deployment env change is required.)

**2. Zombie reaping in spawned connector children (the pid-leak fix).** Every Playwright browser teardown orphans ~2 Chromium helpers that re-parent to the container's PID 1 (celery, which never `wait()`s): ~8k unreaped pids on the alerting heavy pod, 2–3k on every docfetching pod. `SimpleJobClient` children now set `PR_SET_CHILD_SUBREAPER` and drain orphans with `waitpid(WNOHANG)` on exit. Validated in-image: 0 zombies leak through the harness (previously 2 per browser cycle, forever).

## How Has This Been Tested?

- Memory: side-by-side crawl runs in the `v4.4.0-cloud.6` image — unmitigated RSS ramps linearly with pages crawled; with the 30s gc+trim it holds flat at ~233Mi through 240 pages; `MALLOC_ARENA_MAX=2` control run plateaus higher (~250Mi at 480 pages).
- Zombies: `SimpleJobClient` child running 3 Playwright browser cycles leaves 0 zombies in the container (6 accumulate without the fix).
- `pre-commit` (ruff, ty) green.

## Additional Options

### DIFF
--- a/backend/onyx/utils/os_reaper.py
+++ b/backend/onyx/utils/os_reaper.py
@@ -1,0 +1,66 @@
+"""Reap orphaned child processes in spawned connector workers (Linux).
+
+Chromium helpers orphaned by Playwright browser teardown re-parent to the
+container's PID 1 (the celery worker, which never wait()s) and accumulate as
+zombies. Spawned connector children mark themselves as the subreaper instead
+and drain the orphans before exiting.
+
+Only call reap_exited_children from a process whose subprocess usage is
+sequential and fully owned — a blanket waitpid(-1) steals exit statuses from
+concurrent waiters.
+"""
+
+import os
+import sys
+
+from onyx.utils.logger import setup_logger
+
+logger = setup_logger()
+
+_PR_SET_CHILD_SUBREAPER = 36
+
+
+def become_child_subreaper() -> bool:
+    """Re-parent orphaned descendants to this process instead of PID 1."""
+    if sys.platform != "linux":
+        return False
+
+    try:
+        import ctypes
+
+        libc = ctypes.CDLL(None, use_errno=True)
+        if libc.prctl(_PR_SET_CHILD_SUBREAPER, 1, 0, 0, 0) != 0:
+            logger.warning(
+                "become_child_subreaper: prctl failed errno=%s", ctypes.get_errno()
+            )
+            return False
+    except Exception:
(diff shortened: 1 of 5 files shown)


backend/onyx/utils/os_reaper.py
```diff
@@ -0,0 +1,66 @@
...
+from onyx.utils.logger import setup_logger
+
+logger = setup_logger()
+
+_PR_SET_CHILD_SUBREAPER = 36
+
+
+def become_child_subreaper() -> bool:
+    """Re-parent orphaned descendants to this process instead of PID 1."""
+    if sys.platform != "linux":
```
evan-onyx, line 25: this seems odd to me

backend/onyx/utils/os_reaper.py
```diff
@@ -0,0 +1,66 @@
...
+            logger.warning("reap_exited_children: waitpid failed", exc_info=True)
+            break
+
+        if pid == 0:
+            # children exist but none have exited
+            break
+
+        reaped += 1
+
+    return reaped
```
evan-onyx, line 66: am concerned about no tests, if you can add those I think it's good. tbf this one is probably hard to meaningfully test

evan-onyx (commented):
lets get Jamison's thoughts here too

PR #13490, opened 2026-07-28
feat(craft): pre-pull the sandbox image onto the sandbox node pool

Danelegend:
## Problem

The first sandbox pod on a node that lacks the image pays the pull. Measured by evicting `onyxdotapp/sandbox:latest` from a kind node and timing `crictl pull` (1070 MB compressed, **34 MB/s effective**):

| | pull | create → Ready |
|---|---|---|
| cold pull | **31.4 s** | 32.6 s |
| blobs cached, image record dropped | 1.4 s | 3.3 s |

Roughly **5x worse** than the 3–6 s previously assumed from "150–300 MB/s AZ-local bandwidth" — that figure assumed a same-region registry, and self-hosted installs pulling from Docker Hub over their own link are the realistic case.

It isn't only latency. That cost lands inside the 90 s budget `_wait_for_pod_ip` shares with the opencode-history restore, so on a slower link the download alone blows the deadline and provisioning **fails** with `Timeout waiting for sandbox pod ... to be assigned an IP` — or succeeds and starves the restore.

Nodes lack the image whenever they're new, whenever no sandbox has landed on them yet, after a tag bump, or after kubelet image GC reclaims it. Previously the first user onto each cold node did the pre-pull, unpaid, on their own latency budget.

## Additional Options

## Change

A DaemonSet holding the sandbox image on every node matching `sandboxPod.nodeSelector`, so the pull happens at deploy time instead of on a user's first request. On by default when `ENABLE_CRAFT` is set; `sandboxImagePrepull.enabled: false` to opt out.

Load-bearing details, each of which fails silently if wrong:

- **DaemonSet, not a one-shot pre-pull Job.** The kubelet only exempts images referenced by a *running* pod from image GC. A Job pulls and exits, leaving the layers evictable with nothing to signal it. It also covers nodes that join after install.
- **Image ref shared with the PodTemplate** via new `onyx.sandboxImage` / `onyx.sandboxImagePullPolicy` helpers. A drifted tag pins layers nobody uses while every sandbox still cold-pulls — and it looks healthy. A test asserts the two agree; the PodTemplate refactor is a verified byte-identical no-op.
- **Negative priority** (`-10`, `preemptionPolicy: Never`) so a pending sandbox preempts the prepuller rather than failing to schedule and triggering a needless scale-up. Safe because a sandbox running on the node pins the image itself — the prepuller is redundant exactly when it would compete.
- **Portable idle loop**, not `sleep infinity` (a GNU coreutils extension that dies on busybox/alpine; hit this for real while testing).
- **Labelled `component: sandbox-image-prepuller`**, not `component: sandbox` — the latter is the selector for the sandbox NetworkPolicies.
- **Defaults resolved field by field, not via sprig `merge`**, which treats `false`/`0` as empty and was silently overwriting `priorityClass.create: false`.

## Verification

Ran real image GC on `kind-onyx-dev` by lowering `imageGCHighThresholdPercent` to 30 — kind ships **100**, i.e. image GC is off, which is why this behaviour is invisible on a dev cluster:

```
17:37:39  images=61  sandbox_present=1
17:41:40  images=63  sandbox_present=1
17:43:11  images=25  sandbox_present=1   <- 38 images / 33.5 GB reclaimed
17:44:41  images=25  sandbox_present=1
```

Every in-use image survived; the 38 that went were unreferenced. The sandbox image had been resident for days, so it was age-eligible and spared on the in-use rule.

Also verified live: `priority=-10`, ephemeral-storage request present, 0 restarts on the portable command, and real sandbox pods at `priority=0` (strictly above the prepuller, so preemption works as intended).

**Not fully isolated:** on that node the sandbox image was held by the prepuller *and* a live sandbox pod, so "prepuller alone suffices" rests on kubelet's in-use rule rather than a single-holder demonstration. A clean control was set up but never resolved (both images too recently pulled for that GC pass).

## Tradeoffs

- **~3.3 GB becomes unreclaimable per sandbox node**, on the same disk as the sandbox workspaces. Disk pressure that image GC used to absorb now resolves by evicting pods instead. Documented in `values.yaml` with the two cases where you should disable it.
- Eviction ordering is *not* a lever: kubelet ranks "exceeds ephemeral-storage request" ahead of priority, and evicting the prepuller frees ~0 bytes anyway.
- **Does not fix autoscale-up.** Node boot (60–120 s) dominates the pull there, and the DaemonSet lands on the new node simultaneously with the sandbox that triggered it. That needs pre-booted spare capacity, node-image baking, or lazy pull (GKE Image Streaming / SOCI).

## Follow-ups, deliberately not in this PR

1. **`docs/craft/sandbox/image-and-spinup.md` still documents the opposite decision.** Its "Cold pulls vs. image warming" section says *"Current state: we accept cold image pulls"* and lists this exact DaemonSet under *"What we considered and rejected."* Merging this makes that section wrong. A doc update is written and ready to land separately.
2. **`test_pod_spec.py` skips silently.** It renders the chart without a push keypair, so all 24 sandbox-pod security-invariant tests skip everywhere including CI (only the full-cluster lane supplies a key). One-line fix, separate PR.
3. **`provision()` has no retry**, and `_wait_for_pod_ip`'s 90 s deadline must cover autoscaler scan + node boot + pull. That looks like it fails outright on a cold scale-up. Grepping prod logs for `Timeout waiting for sandbox pod` against node-creation events would confirm.
4. `test_event_bus_per_directory.py` has 9 pre-existing failures (`Engine not initialized`) unrelated to this change — confirmed identical at `HEAD`.

### DIFF
--- a/backend/tests/unit/onyx/server/features/build/sandbox/test_sandbox_image_prepuller.py
+++ b/backend/tests/unit/onyx/server/features/build/sandbox/test_sandbox_image_prepuller.py
@@ -1,0 +1,309 @@
+"""Invariants for the sandbox image-prepull DaemonSet.
+
+The DaemonSet exists to keep the sandbox image resident on every node in the
+sandbox pool so a user's first sandbox there doesn't pay the pull. Three
+properties make it work, and each fails silently if broken:
+
+  - It must reference the *same* image as the sandbox pods. A drifted tag pins
+    layers nobody uses while every sandbox still cold-pulls.
+  - It must land on the same nodes (nodeSelector + tolerations), or it warms
+    the wrong pool.
+  - It must resolve the same pull credentials, or a private registry leaves it
+    in ImagePullBackOff while it appears deployed.
+  - It must stay running, because the kubelet only skips image GC for images a
+    running pod references. Hence the long-lived, shell-portable command.
+
+Skips if the ``helm`` binary or chart deps are unavailable.
+"""
+
+from __future__ import annotations
+
+import re
+import shutil
+import subprocess
+from pathlib import Path
+
+import pytest
+import yaml
+
+# backend/tests/unit/onyx/server/features/build/sandbox/ -> repo root
+_REPO_ROOT = Path(__file__).resolve().parents[8]
+_CHART_DIR = _REPO_ROOT / "deployment" / "helm" / "charts" / "onyx"
+_PREPULLER_TEMPLATE = "templates/sandbox-image-prepuller.yaml"
+_PODTEMPLATE_TEMPLATE = "templates/sandbox-podtemplate.yaml"
+
+# The chart refuses to render without a push keypair; any valid base64 seed works.
+_FAKE_PUSH_KEY = "ZmFrZWtleWZha2VrZXlmYWtla2V5ZmFrZWtleWZha2VrZXk="
+
+
+def _run_helm(
+    template: str, extra_args: list[str] | None = None
+) -> subprocess.CompletedProcess[str]:
+    helm = shutil.which("helm")
+    if helm is None:
+        pytest.skip("helm binary not available")
+    cmd = [
+        helm,
+        "template",
+        "onyx",
+        str(_CHART_DIR),
+        "-n",
+        "onyx",
+        "-f",
+        str(_CHART_DIR / "values-ci.yaml"),
+        "--kube-version",
+        "1.33.0",
+        "--set",
+        f"auth.sandboxPushSecret.values.private_key={_FAKE_PUSH_KEY}",
+        *(extra_args or []),
+        "--show-only",
+        template,
+    ]
+    return subprocess.run(cmd, capture_output=True, text=True)
+
+
+def _render(template: str, extra_args: list[str] | None = None) -> str:
+    result = _run_helm(template, extra_args)
+    if result.returncode != 0:
+        pytest.skip(f"helm template failed (chart deps?): {result.stderr.strip()}")
+    return result.stdout
+
+
+def _renders_nothing(extra_args: list[str]) -> bool:
+    """True when the template produced no manifest at all.
+
+    ``--show-only`` on a template that rendered empty is a helm *error*
+    ("could not find template"), not empty stdout — so a plain string check on
+    stdout would pass for a genuinely broken chart too.
+    """
+    result = _run_helm(_PREPULLER_TEMPLATE, extra_args)
+    if result.returncode == 0:
+        return not result.stdout.strip().strip("-")
+    if "could not find template" in result.stderr:
+        return True
+    pytest.skip(f"helm template failed (chart deps?): {result.stderr.strip()}")
+
+
+def _prepuller(extra_args: list[str] | None = None) -> dict:
+    """The DaemonSet. The template also emits a PriorityClass, so select by kind
+    rather than assuming a single document."""
+    docs = yaml.safe_load_all(_render(_PREPULLER_TEMPLATE, extra_args))
+    return next(d for d in docs if d and d["kind"] == "DaemonSet")
+
+
+def _prepuller_pod_spec(extra_args: list[str] | None = None) -> dict:
+    return _prepuller(extra_args)["spec"]["template"]["spec"]
+
+
+def test_prepuller_image_matches_the_sandbox_pod_image() -> None:
+    """The whole mechanism is a no-op unless both resolve to the same ref, and
+    a mismatch is invisible at runtime."""
+    args = [
+        "--set",
+        "global.version=v9.8.7",
+        "--set-string",
+        "configMap.SANDBOX_CONTAINER_IMAGE=",
+    ]
+
+    prepuller_image = _prepuller_pod_spec(args)["containers"][0]["image"]
+    pod_template = yaml.safe_load(_render(_PODTEMPLATE_TEMPLATE, args))
+    sandbox_images = {
+        c["image"]
+        for c in pod_template["template"]["spec"]["containers"]
+        + pod_template["template"]["spec"]["initContainers"]
+    }
+
+    assert sandbox_images == {"onyxdotapp/sandbox:v9.8.7"}
+    assert prepuller_image == "onyxdotapp/sandbox:v9.8.7"
+
+
+def test_prepuller_image_follows_explicit_override() -> None:
+    spec = _prepuller_pod_spec(
+        ["--set-string", "configMap.SANDBOX_CONTAINER_IMAGE=onyxdotapp/sandbox:dev"]
+    )
+    assert spec["containers"][0]["image"] == "onyxdotapp/sandbox:dev"
+
+
+def test_prepuller_scheduling_matches_the_sandbox_pod() -> None:
+    """Warming a node the sandboxes can't be scheduled onto buys nothing."""
+    prepuller = _prepuller_pod_spec()
+    sandbox = yaml.safe_load(_render(_PODTEMPLATE_TEMPLATE))["template"]["spec"]
+
+    assert prepuller["nodeSelector"] == sandbox["nodeSelector"]
+    assert prepuller["tolerations"] == sandbox["tolerations"]
+
+
+def test_prepuller_stays_running_to_pin_layers() -> None:
+    """The kubelet only exempts images referenced by a *running* pod from image
+    GC, so a one-shot pull would leave the layers evictable.
+
+    The command must also be portable: `sleep infinity` is a GNU coreutils
+    extension that dies on busybox/alpine, and a CrashLooping prepuller unpins
+    the image silently.
+    """
+    container = _prepuller_pod_spec()["containers"][0]
+    assert container["command"] == ["sh", "-c", "while :; do sleep 86400; done"]
+
+
+def test_prepuller_pull_policy_is_if_not_present() -> None:
+    """`Always` would re-pull on every pod restart, defeating the point."""
+    container = _prepuller_pod_spec(
+        ["--set", "configMap.SANDBOX_IMAGE_PULL_POLICY=Always"]
+    )["containers"][0]
+    assert container["imagePullPolicy"] == "IfNotPresent"
+
+
+def test_prepuller_requests_ephemeral_storage() -> None:
+    """A container with no ephemeral-storage request is invisible to the
+    scheduler's disk accounting — which this repo treats as a bug (see
+    test_pod_spec.test_all_containers_set_ephemeral_storage_requests). The
+    prepuller writes nothing, but it must still be accounted for."""
+    resources = _prepuller_pod_spec()["containers"][0]["resources"]
+    assert "ephemeral-storage" in resources["requests"]
+    assert "ephemeral-storage" in resources["limits"]
+    assert resources["requests"] == resources["limits"]
+
+
+def test_prepuller_yields_to_pending_sandbox_pods() -> None:
+    """A do-nothing pod must never be the reason a sandbox fails to schedule and
+    triggers a scale-up. Negative priority lets a pending sandbox preempt it;
+    preemptionPolicy Never means the prepuller evicts nothing itself."""
+    docs = list(yaml.safe_load_all(_render(_PREPULLER_TEMPLATE)))
+    pc = next(d for d in docs if d and d["kind"] == "PriorityClass")
+    ds = next(d for d in docs if d and d["kind"] == "DaemonSet")
+
+    assert pc["value"] < 0
+    assert pc["globalDefault"] is False
+    assert pc["preemptionPolicy"] == "Never"
+    assert ds["spec"]["template"]["spec"]["priorityClassName"] == pc["metadata"]["name"]
+
+
+def test_existing_priority_class_skips_creation() -> None:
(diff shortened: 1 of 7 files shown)


Danelegend:
Thanks — issues 1 and 2 are valid and fixed in `cf4aa2b`. Issue 3 doesn't apply. Details, including where the premises were slightly off:

### Issue 1 — private image credentials: **valid, fixed**

Fixed by giving the prepuller both credential routes: `.Values.imagePullSecrets` and `serviceAccountName: {{ SANDBOX_SERVICE_ACCOUNT_NAME }}` (the same SA the sandbox pods use), with `automountServiceAccountToken: false` retained since this pod makes no API calls.

Two corrections to the premise, which change the impact rather than the fix:

- **The sandbox SA carries no `imagePullSecrets` today** — `sandbox-rbac.yaml` creates it with only `automountServiceAccountToken: false`. So the prepuller wasn't dropping credentials the sandbox pods had; neither had any.
- **The sandbox PodTemplate doesn't set `imagePullSecrets` either.** So a private sandbox image is currently unsupported chart-wide for Craft, not something the prepuller uniquely broke.

The fix is still worth making, and it's worth *more* than the report suggests: because pulls are `IfNotPresent`, a prepuller that can authenticate warms the node, and sandbox pods then never need to pull at all. This effectively makes private registries work for Craft sandboxes. It's load-bearing rather than belt-and-braces, so I've locked it with a test asserting the prepuller's `serviceAccountName` equals the PodTemplate's and that the chart-wide secret propagates.

Adding `imagePullSecrets` to the sandbox PodTemplate as well is the complete fix, but it's a pre-existing gap in a file this PR only refactors, so I've left it out.

### Issue 2 — PriorityClass name collision: **valid, fixed**

Correct, and the chart already had the convention for it: `sandbox-rbac.yaml` builds its names from `printf "%s-%s" (include "onyx.fullname" .) .Release.Namespace` with a `trunc 38` + `sha256sum | trunc 8` fallback, precisely because those objects live in a shared namespace. I mirrored that shape exactly rather than inventing a scheme:

```
ns=onyx          -> onyx-onyx-sandbox-image-prepuller
ns=onyx-staging  -> onyx-onyx-staging-sandbox-image-prepuller
```

Two tests: one asserting the names differ across namespaces, one asserting a 60-char namespace still hashes down to a valid DNS-subdomain name within the 253-char limit.

### Issue 3 — documented idle command: **doesn't apply**

This PR doesn't touch `docs/craft/sandbox/image-and-spinup.md` — the doc changes were pulled out to keep it code-only, so there's no line 130 in the diff. The `sleep infinity` text you're pointing at is main's **"What we considered and rejected"** section, describing a hypothetical `sandbox-image-warmer` that was turned down; it isn't documentation of this DaemonSet.

The underlying concern is real for the follow-up doc PR, though, so I checked the staged doc patch: it says *"a portable idle loop, not `sleep infinity`: that's a GNU coreutils extension which dies on busybox/alpine"* — it already documents the portable command and explains why, so your suggested edit is covered.

Worth flagging the bigger version of this: once this merges, that "we considered and rejected" section actively contradicts the shipped code, since it rejects this exact DaemonSet. That's follow-up #1 in the PR description and the doc change is written and ready.

### Verification

18 tests pass, `helm lint` clean. Re-verified live on `kind-onyx-dev`: `phase=Running sa=sandbox priority=-10 class=onyx-onyx-sandbox-image-prepuller restarts=0`.

Danelegend:
This one is already fixed — the comment is anchored to an older commit. Current head `c59e6a9` has it at lines 126–131 of the template:

```yaml
      {{- with .Values.imagePullSecrets }}
      imagePullSecrets:
        {{- toYaml . | nindent 8 }}
      {{- end }}
      serviceAccountName: {{ $saName }}
      automountServiceAccountToken: false
```

Rendered proof with a private registry and a non-default SA name:

```
$ helm template ... --set 'imagePullSecrets[0].name=regcred' \
    --set-string configMap.SANDBOX_SERVICE_ACCOUNT_NAME=custom-sa

      imagePullSecrets:
        - name: regcred
      serviceAccountName: custom-sa
      automountServiceAccountToken: false
```

Line 93 on current head is inside the `updateStrategy` comment, unrelated to pull credentials.

Two corrections to the premise, repeated from [my earlier reply](https://github.com/onyx-dot-app/onyx/pull/13490#issuecomment-5098807118) since they change the impact rather than the fix:

- **The sandbox SA carries no `imagePullSecrets` today.** `sandbox-rbac.yaml` creates it with only `automountServiceAccountToken: false`, so the prepuller was not dropping credentials that sandbox pods had — neither had any.
- **The sandbox PodTemplate does not set `imagePullSecrets` either**, so a private sandbox image is currently unsupported chart-wide for Craft. Adding it there too is the complete fix, but it is a pre-existing gap in a file this PR only refactors, so it is deliberately out of scope.

Worth noting the fix is more valuable than "belt and braces": because pulls are `IfNotPresent`, a prepuller that can authenticate warms the node and sandbox pods then never pull at all — which effectively makes private registries work for Craft sandboxes. Two tests lock it (`test_prepuller_pull_credentials_match_the_sandbox_pod`, `test_prepuller_service_account_follows_the_configured_name`), the first asserting the prepuller's `serviceAccountName` equals the PodTemplate's so the two cannot drift.

Danelegend:
@greptile @cubic

Danelegend:
### CI: the two `craft-k8s` failures are pre-existing, not from this PR

`craft-k8s (snapshot_restore)` and `craft-k8s (skill_push)` fail here, and `craft-k8s-tests-required` reflects them. Evidence they aren't caused by this change:

**The lane is fully red on main.** All four runs below execute the same 11-job matrix:

| run | branch | craft-k8s jobs | failed |
|---|---|---|---|
| 30320435316 | this PR | 11 | **2** |
| 30249054810 | main (07-27) | 11 | **11** |
| 30193543287 | main (07-26) | 11 | **11** |
| 30149771040 | main (07-25) | 11 | **11** |

Both `snapshot_restore` and `skill_push` are in main's failing set in all three daily runs. This PR fails a strict *subset* — it passes 9 of the 11 that main fails. Last green run on main was 2026-07-21.

(Other PRs' green `craft-k8s` runs are not a usable baseline: they're path-filtered to 1 job with 3 skipped, because they don't touch craft paths. This PR touches `deployment/helm/charts/onyx/**`, so it runs the full matrix.)

**Provisioning demonstrably works with the prepuller deployed.** In the failing job, 7 tests passed and the api-server log shows three successful provisions:

```
01:40:20 provisioning 63807386 -> opencode-serve ready 01:40:30  (10s)
01:40:45 provisioning 63807386 -> ready 01:40:56              (11s)
01:41:01 provisioning 80dc96c1 -> ready 01:41:17              (16s)
01:41:50 ERROR Session creation failed: timed out   [API:zec4yIoh]
```

The failing request never reached provisioning — there is no `Starting Kubernetes sandbox provisioning` line for `zec4yIoh`, so it timed out upstream of the sandbox manager. It was the 8th session in the job; had the DaemonSet broken image pulls or the ServiceAccount, the *first* would have failed.

**Ruled out the mechanisms this PR could plausibly cause:**
- No `ImagePullBackOff` anywhere; the prepuller pod ran (its log was collected in the diagnostics bundle).
- No `DiskPressure`, no evictions, no `Insufficient <resource>` — so the unreclaimable-disk tradeoff documented in `values.yaml` is not what fired here.
- No PriorityClass or RBAC errors during `helm install`.

I can't prove a negative from CI alone, and main being 11/11 red means there's no clean baseline to compare against either. If a reviewer wants certainty, re-running these two jobs on a branch that only bumps `Chart.yaml` would isolate it.

deployment/helm/charts/onyx/Chart.yaml
```diff
@@ -5,7 +5,7 @@ home: https://www.onyx.app/
 sources:
   - "https://github.com/onyx-dot-app/onyx"
 type: application
-version: 0.8.5
+version: 0.8.6
```
wenxi-onyx, line 8: desc. with version bump

deployment/helm/charts/onyx/values.yaml
```diff
@@ -1654,6 +1654,41 @@ sandboxPod:
       value: sandbox
       effect: NoSchedule
 
+# -- Pre-pull the sandbox image onto the sandbox node pool (DaemonSet), so a
+# user's first sandbox on a node doesn't pay the image pull (~31s measured; see
+# docs/craft/sandbox/image-and-spinup.md). Image and pull policy are shared with
```
wenxi-onyx: nit: pref no reference to docs because those often go stale and/or get deleted

deployment/helm/charts/onyx/values.yaml
```diff
@@ -1654,6 +1654,41 @@ sandboxPod:
...
+# the sandbox-pod PodTemplate; scheduling follows sandboxPod above.
+#
+# The cost is disk, not CPU: holding the image makes ~3.3 GB UNRECLAIMABLE on
```
wenxi-onyx: I dont think comments below line 1662 are necessary

deployment/helm/charts/onyx/templates/sandbox-image-prepuller.yaml
```diff
@@ -0,0 +1,169 @@
...
+{{- $pcCreate := true }}
+{{- if hasKey $pc "create" }}{{- $pcCreate = $pc.create }}{{- end }}
+{{- $pcValue := -10 }}
+{{- if hasKey $pc "value" }}{{- $pcValue = $pc.value }}{{- end }}
+{{- if and (eq (include "onyx.craftEnabled" .) "true") $prepull.enabled }}
+{{- /*
+Pre-pulls the sandbox image onto every node in the sandbox pool so a user's
+first sandbox there doesn't pay the pull.
+
+Why a DaemonSet running `sleep infinity` and not a one-shot pre-pull Job: the
```
wenxi-onyx: pref no design decision logs like this in comments. Will clutter actually important code files relatively quickly if we leave these in. 

If it's important for anyone else to know this, I would just update the docs/ file that is referenced below in a separate comment

deployment/helm/charts/onyx/templates/sandbox-image-prepuller.yaml
```diff
@@ -0,0 +1,169 @@
...
+      {{- /* Pull credentials must match the sandbox pods' exactly, or the
+      prepuller ImagePullBackOffs against a private registry and silently warms
+      nothing. Both routes are covered: the chart-wide secrets, and the sandbox
+      ServiceAccount, which an operator may have attached secrets to. The token
+      stays unmounted — this pod makes no API calls; it is here only so image
+      pull resolves the same credentials a sandbox pod would. */}}
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
```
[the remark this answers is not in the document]
wenxi-onyx: I'm pretty sure we will never run into this, but could be valid? 

Would face this if someone builds sandbox images and hosts them in their own registry as opposed to using our dockerhub. Might want to check if we know of anyone that currently does this

deployment/helm/charts/onyx/values.yaml
```diff
@@ -1654,6 +1654,41 @@ sandboxPod:
...
+# every sandbox node, on the same disk as the sandbox workspaces. Turn this off
+# when either applies:
+#   - Single-node / fixed tiny deployments: the image stays resident anyway and
+#     nothing evicts it, so the DaemonSet is pure overhead.
+#   - Nodes without headroom for 3.3 GB on top of the workspace ephemeral-storage
+#     limits. Disk pressure that image GC used to absorb would otherwise resolve
+#     by evicting pods instead.
+sandboxImagePrepull:
+  enabled: true
```
wenxi-onyx, line 1667: Should this depend on ENABLE_CRAFT instead of default enable?

backend/tests/unit/onyx/server/features/build/sandbox/test_sandbox_image_prepuller.py
```diff
@@ -0,0 +1,309 @@
+"""Invariants for the sandbox image-prepull DaemonSet.
+
+The DaemonSet exists to keep the sandbox image resident on every node in the
+sandbox pool so a user's first sandbox there doesn't pay the pull. Three
+properties make it work, and each fails silently if broken:
```
wenxi-onyx: Why does it fail silently? Should it?

wenxi-onyx (approved):

deployment/helm/charts/onyx/values.yaml
Danelegend: `{{- if and (eq (include "onyx.craftEnabled" .) "true") $prepull.enabled }}`

deployment/helm/charts/onyx/templates/sandbox-image-prepuller.yaml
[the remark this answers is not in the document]
Danelegend: tbh I'm not exactly sure what this means, nor what the action item is .-.

Danelegend:
@greptile

Danelegend:
@greptile

Danelegend:
@greptile

PR #13497, opened 2026-07-28
feat(perms): agents + doc sets — server-driven capability projection

Subash-Mohan:
## Description

- Extends the capability model to **agents (personas) and document sets** — the worst mis-gated surface. The server now reports which actions a user may take on each agent/doc-set, and the client renders that instead of guessing from ownership.
- **AgentCard**: edit / share / stats / feature now gate on the server's map, not `checkUserOwnsAgent`. A manager or editor who can edit an agent but doesn't *own* it is no longer hidden — the actual bug behind most of the agent gaps.
- **Doc-set admin page**: dropped the client-side double-fetch + set-diff; each row carries its own edit capability.
- Backend guards are **unchanged**; read-side helpers mirror them and are pinned to enforcement by a contract test that drives the real guards.

## How Has This Been Tested?

- External-dependency-unit contract suite drives the **real** persona guard (`_assert_persona_update_within_managed_scope`), the stats gate (`user_can_view_assistant_stats`), the delete handler's ownership gate, and doc-set `within_scope` — 18/18. Frontend `tsgo` + oxlint/oxfmt clean.

## Additional Options

---

Follow-up (not in this PR): the admin agents table (`AgentRowActions`) per-row affordances aren't yet cut over to the map — they still over-show for a manager on out-of-scope rows (backend 403s the action).

---
## Summary by cubic
Adds server-driven per-item permissions across agents, document sets, actions (`ToolSnapshot`), MCP servers, skills, and user groups. The UI reads stamped affordance maps to show only allowed actions; persona publish is owner-or-admin, delete follows `ADD_AGENTS`, and `ADD_AGENTS` no longer implies `READ_AGENTS`.

- **New Features**
  - Actions/MCP: UI gates `toggle`, `authenticate`, and `manage_status` from server maps; owner-or-admin management only; `ToolItem` respects `canToggle`.
  - Skills: publish is admin-only; `SkillSharePicker` disables org-wide publish when not permitted; `UploadSkillModal` defaults to private unless global `MANAGE_SKILLS`.
  - Agents: list and details stamp `permissions` (incl. `share`); publish is owner-or-admin; `AgentRowActions` and `AgentCard` gate every action on the server map.
  - Groups/Doc sets: list endpoints stamp per-item `permissions` (manage/delete/edit_permissions/edit_token_limits); document sets stamp editability in one fetch; admin pages read the maps.

- **Bug Fixes**
  - Agents: delete returns 403 when unauthorized; scoped `ADD_AGENTS` owners can delete; publish restricted to owner-or-admin; tests added.
  - Permissions: `ADD_AGENTS` no longer implies `READ_AGENTS`; tests updated.
  - Document sets: removed duplicate refresh call; list-stamped permissions avoid flicker.

<sup>Written for commit fd2702ba955d7c5942291f3118896ea43bbf52e8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13497?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/auth/permission_projection.py
+++ b/backend/onyx/auth/permission_projection.py
@@ -26,3 +26,54 @@ def cc_pair_permissions(
         "delete": is_connectors_admin,
         "publish": is_connectors_admin,
     }
+
+
+PERSONA_ACTIONS: frozenset[str] = frozenset(
+    {"edit", "share", "view_stats", "delete", "publish", "feature", "list", "reorder"}
+)
+
+
+def persona_permissions(
+    *,
+    can_edit: bool,
+    can_view_stats: bool,
+    can_delete: bool,
+    is_manage_agents_admin: bool,
+    is_full_admin: bool,
+) -> dict[str, bool]:
+    """Agent (persona) affordance map. ``edit``/``share`` are the scoped
+    editable-and-managed-scope decision (share shares edit's gate). ``view_stats`` is
+    owner-or-full-admin. ``delete``/``publish`` are owner-or-admin (the handler's own
+    ownership check, not the route token). ``feature``/``list`` need global MANAGE_AGENTS
+    and ``reorder`` needs full admin — so a scoped manager may edit a managed agent but
+    not delete, publish, feature, list, or reorder it."""
+    return {
+        "edit": can_edit,
+        "share": can_edit,
+        "view_stats": can_view_stats,
+        "delete": can_delete,
+        "publish": can_delete,
+        "feature": is_manage_agents_admin,
+        "list": is_manage_agents_admin,
+        "reorder": is_full_admin,
+    }
+
+
+DOCUMENT_SET_ACTIONS: frozenset[str] = frozenset(
+    {"edit", "manage_access", "delete", "publish"}
+)
+
+
+def document_set_permissions(
+    *, is_editable: bool, is_document_sets_admin: bool
+) -> dict[str, bool]:
+    """Document set affordance map. ``edit`` and ``manage_access`` are the managed-scope
+    editable decision the write guard enforces (a doc set has no editor-share arm, so
+    editable membership is that decision). ``delete`` and ``publish`` (make org-wide
+    public) are global MANAGE_DOCUMENT_SETS only."""
+    return {
+        "edit": is_editable,
+        "manage_access": is_editable,
+        "delete": is_document_sets_admin,
+        "publish": is_document_sets_admin,
+    }
--- a/backend/onyx/auth/permission_projection.py
+++ b/backend/onyx/auth/permission_projection.py
@@ -46,17 +46,18 @@ def persona_permissions(
     """Agent (persona) affordance map. ``edit``/``share`` gate on editable alone (their routes are
     BASIC_ACCESS) — an editor-shared user without ADD_AGENTS may still edit and manage sharing.
     ``edit`` is the editable-AND-managed-scope decision the update guard enforces; ``share`` is the
-    broader editable decision the share guard enforces (get_editable, no scope AND). ``delete``/
-    ``publish`` AND in ``holds_add_agents`` — the ADD_AGENTS authority their routes gate on at
-    GATE 1 (allow_scope), so an owner lacking it is 403'd there. ``view_stats`` is
-    owner-or-full-admin; ``delete``/``publish`` are owner-or-admin. ``feature``/``list`` need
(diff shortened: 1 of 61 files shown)


backend/onyx/auth/permission_projection.py
```diff
@@ -26,3 +26,54 @@ def cc_pair_permissions(
...
+    editable-and-managed-scope decision (share shares edit's gate). ``view_stats`` is
+    owner-or-full-admin. ``delete``/``publish`` are owner-or-admin (the handler's own
+    ownership check, not the route token). ``feature``/``list`` need global MANAGE_AGENTS
+    and ``reorder`` needs full admin — so a scoped manager may edit a managed agent but
+    not delete, publish, feature, list, or reorder it."""
+    return {
+        "edit": can_edit,
+        "share": can_edit,
+        "view_stats": can_view_stats,
+        "delete": can_delete,
```
nmgarza5: `can_delete_persona` is owner-or-admin, but `DELETE /persona/{persona_id}` is still `require_permission(ADD_AGENTS)` with no `allow_scope`, so a group manager who owns an agent gets `delete: true` and then a 403. The contract test pins this against `get_persona_by_id` rather than the route dependency, which is why it passes.

backend/onyx/server/features/persona/api.py
```diff
@@ -597,6 +603,22 @@ def get_persona(
         snapshot.user_permission = get_persona_access_level(
             persona, user, user_group_ids
         )
+        snapshot.permissions = persona_permissions(
```
nmgarza5, line 623: Only the single-persona GET stamps this. `PersonaSnapshot` has no `permissions` field, so `can(agent, ...)` on anything that came from a list endpoint is permanently false, and `AgentCard` works around it by reading the per-agent fetch instead.

web/src/sections/agents/AgentCard.tsx
```diff
@@ -158,7 +153,7 @@ export default function AgentCard({ agent }: AgentCardProps) {
                       className="hidden group-hover/AgentCard:flex"
                     />
                   )}
-                  {isOwnedByUser && (
+                  {can(fullAgent, "edit") && (
```
nmgarza5: `edit` comes from the get_editable filter, which includes the org-wide arm (`is_public` plus `public_permission = EDITOR`), so every user now gets a pencil on any public editable agent where before only the owner did. It also depends on the per-card `useAgent` fetch resolving, so the icons pop in a beat after the card renders.

nmgarza5 (approved):

backend/onyx/server/features/oauth_config/api.py
```diff
@@ -59,13 +62,31 @@ def _oauth_config_to_snapshot(
...
+def _assert_can_manage_oauth_config(
+    oauth_config: OAuthConfig, user: User, db_session: Session
+) -> None:
+    """Owner-or-admin gate: a global actions-admin manages any OAuth config; a scoped manager
+    manages one only when they own every action referencing it — editing shared credentials
+    would otherwise affect the other creators."""
+    if has_global_permission(user, Permission.MANAGE_ACTIONS):
+        return
+    tools = get_tools_by_oauth_config(oauth_config.id, db_session)
+    if tools and all(tool.user_id == user.id for tool in tools):
```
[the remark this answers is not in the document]
Subash-Mohan: Not a credential-exposure path: the API never returns client_id/secret (OAuthConfigSnapshot is has_client_credentials only) and scoped managers can't list or read configs they don't own — the residual is just an insider being able to tamper/delete a config during a create-race, which is low-risk enough to accept; will revisit with an owner column if we ever expose config reuse.

backend/onyx/server/features/tool/api.py
```diff
@@ -168,9 +132,11 @@ def update_custom_tool(
 def delete_custom_tool(
     tool_id: int,
     db_session: Session = Depends(get_session),
-    user: User = Depends(require_permission(Permission.MANAGE_ACTIONS)),
+    user: User = Depends(
+        require_permission(Permission.MANAGE_ACTIONS, allow_scope=True)
```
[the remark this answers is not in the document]
Subash-Mohan: Intentional: actions are owner-or-admin, not group-scoped.

PR #13500, opened 2026-07-28
feat(perms): PR-D groups + token limits — close out the capability matrix

Subash-Mohan:
## Description

- Closes out the server-driven capability model with the final scoped-manager surface — **user groups** (+ token rate limits). Groups previously had *no per-group manager signal at all*, so every group control rendered unconditionally; the server now stamps a per-group affordance map and the client renders only what the backend allows.
- **GroupsPage**: opening/renaming a group and editing its members/agents now gate on per-group `manage`; deleting a group gates on global `MANAGE_USER_GROUPS`; the permissions + token-limit sections gate on `FULL_ADMIN`. "New Group" is hidden from scoped managers (create is global-only).
- **Gate classification** verified against the real routes (and correcting the design doc): `delete` is global `MANAGE_USER_GROUPS` (not FULL_ADMIN), and "New Group" gates on the global token — not `useCapabilities`, which would show it to scoped managers who'd then 403.
- Backend guards are **unchanged**; the read side reuses A1's `manages_group` and is pinned to enforcement by the contract test. The per-group stamp resolves the manager's scope once, so the list has no N+1. DTO map defaults empty (fail-closed).

## How Has This Been Tested?

- External-dependency-unit contract suite extended to user groups — drives the **real** `assert_manages_group` guard across in-scope/out-of-scope/admin actors, and asserts the scoped token-limit create (`assert_within_scope`) equals `manage`: 13/13 total. Frontend `tsgo` + oxlint/oxfmt clean.

## Additional Options

---

Notes / pre-existing gaps (out of scope): PR5's "Make/Revoke Manager" toggle UI is not in the tree — the route/DTO/type exist but no UI consumes them; when built, gate it on `can(group, "manage")`. Federated connectors have no per-affordance gate (route-gated only).

---
## Summary by cubic
Closes the server-driven capability model for user groups and lets group managers fully manage their groups’ token limits. The UI now shows only allowed actions, avoiding 403s and extra queries.

- **New Features**
  - Backend stamps per-group `permissions` on `UserGroup` via `user_group_permissions` (`manage`, `delete`, `edit_permissions`, `edit_token_limits`), with `edit_token_limits` keyed to `manage`; manager scope/global authority are resolved once (no N+1); DTO defaults fail-closed.
  - Group token limits admit scope: GET allows_scope + `assert_manages_group`; group-scoped PUT/DELETE at `/user-group/{group_id}/rate-limit/{id}` fence by `group_id` via `_authorize_group_token_rate_limit_write` and `get_token_rate_limit_scope_and_group`.
  - Frontend gates with `can(group, ...)`: manage (rename/members/save + Make/Revoke Manager via `setGroupManager`), delete card, permissions section, and token limit section (fetch/save only when allowed); save uses group-scoped endpoints; “New Group” shows only for global `MANAGE_USER_GROUPS`; list filters to manageable groups; `GroupCard` open/rename require manage; `AdminListHeader` action is optional.
  - Contract tests pin the projection to real guards and routes; token-limit read GATE1 ∧ `assert_manages_group`; group write guard is enforced; action-key coverage added. Integration tests cover manager GET/POST/PUT/DELETE for group limits.

- **Bug Fixes**
  - Token-limit read was global-only; scoped managers can now read and CRUD limits for groups they manage, while group paths cannot reach global/user limits.
  - Unknown `rate_limit_id` on group writes now returns a structured 404 (not a generic 400), consistent with wrong-group IDs and without leaking existence.
  - Cleared `any_rate_limit_exists` cache on token-limit updates to apply changes immediately (shared by-id and group-scoped routes).
  - Gate classification matches routes: delete requires global `MANAGE_USER_GROUPS`; scoped managers no longer see create or admin-only controls that would 403.

<sup>Written for commit 5557e513ee1b5412eafcc253063164215a8c8a5f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13500?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/auth/permission_projection.py
+++ b/backend/onyx/auth/permission_projection.py
@@ -134,3 +134,25 @@ def custom_skill_permissions(
         "delete": is_full_admin,
         "publish": is_skills_admin,
     }
+
+
+USER_GROUP_ACTIONS: frozenset[str] = frozenset(
+    {"manage", "delete", "edit_permissions", "edit_token_limits"}
+)
+
+
+def user_group_permissions(
+    *, can_manage: bool, is_user_groups_admin: bool, is_full_admin: bool
+) -> dict[str, bool]:
+    """User group affordance map. ``manage`` (rename, membership, assign agents, set
+    manager) is the per-group ``manages_group`` decision — group in the manager's managed
+    set, or global MANAGE_USER_GROUPS. ``delete`` needs global MANAGE_USER_GROUPS (its
+    route has no ``allow_scope``). ``edit_permissions`` and ``edit_token_limits`` are
+    FULL_ADMIN (the permission-toggle route and the token-limit PUT/DELETE routes require
+    it); the scoped token-limit create folds into ``manage``."""
+    return {
+        "manage": can_manage,
+        "delete": is_user_groups_admin,
+        "edit_permissions": is_full_admin,
+        "edit_token_limits": is_full_admin,
+    }
--- a/backend/onyx/auth/permission_projection.py
+++ b/backend/onyx/auth/permission_projection.py
@@ -145,14 +145,14 @@ def user_group_permissions(
     *, can_manage: bool, is_user_groups_admin: bool, is_full_admin: bool
 ) -> dict[str, bool]:
     """User group affordance map. ``manage`` (rename, membership, assign agents, set
-    manager) is the per-group ``manages_group`` decision — group in the manager's managed
-    set, or global MANAGE_USER_GROUPS. ``delete`` needs global MANAGE_USER_GROUPS (its
-    route has no ``allow_scope``). ``edit_permissions`` and ``edit_token_limits`` are
-    FULL_ADMIN (the permission-toggle route and the token-limit PUT/DELETE routes require
-    it); the scoped token-limit create folds into ``manage``."""
+    manager) and ``edit_token_limits`` are the per-group ``manages_group`` decision — group
+    in the manager's managed set, or global MANAGE_USER_GROUPS. A scoped manager gets full
+    token-limit CRUD for groups they manage: every token route (read/create/update/delete)
+    now admits scope. ``delete`` needs global MANAGE_USER_GROUPS (its route has no
+    ``allow_scope``). ``edit_permissions`` is FULL_ADMIN (the permission-toggle route)."""
     return {
         "manage": can_manage,
         "delete": is_user_groups_admin,
         "edit_permissions": is_full_admin,
-        "edit_token_limits": is_full_admin,
+        "edit_token_limits": can_manage,
     }
--- a/backend/onyx/db/token_limit.py
+++ b/backend/onyx/db/token_limit.py
@@ -78,6 +78,24 @@ def insert_global_token_rate_limit(
     return token_limit
 
 
+def get_token_rate_limit_scope_and_group(
+    db_session: Session, token_rate_limit_id: int
+) -> tuple[TokenRateLimitScope, int | None]:
+    """Scope + owning group id for a rate limit (group is None unless USER_GROUP). Raises
+    ValueError if the id is unknown."""
+    token_limit = db_session.get(TokenRateLimit, token_rate_limit_id)
+    if token_limit is None:
+        raise ValueError(f"TokenRateLimit with id '{token_rate_limit_id}' not found")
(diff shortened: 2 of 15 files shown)


backend/onyx/auth/permission_projection.py
```diff
@@ -134,3 +134,25 @@ def custom_skill_permissions(
...
+    manager) is the per-group ``manages_group`` decision — group in the manager's managed
+    set, or global MANAGE_USER_GROUPS. ``delete`` needs global MANAGE_USER_GROUPS (its
+    route has no ``allow_scope``). ``edit_permissions`` and ``edit_token_limits`` are
+    FULL_ADMIN (the permission-toggle route and the token-limit PUT/DELETE routes require
+    it); the scoped token-limit create folds into ``manage``."""
+    return {
+        "manage": can_manage,
+        "delete": is_user_groups_admin,
+        "edit_permissions": is_full_admin,
+        "edit_token_limits": is_full_admin,
```
nmgarza5: There are no PUT or DELETE token-limit routes in `ee/onyx/server/token_rate_limits/api.py`, only GETs and POSTs, and the per-group GET is global `MANAGE_USER_GROUPS` rather than FULL_ADMIN. So a non-admin holding global `MANAGE_USER_GROUPS` gets `edit_token_limits: false` and loses a section they can actually read and write.

backend/tests/external_dependency_unit/auth/test_permission_projection_contract.py
```diff
@@ -651,3 +653,78 @@ def test_actions_and_skill_key_coverage() -> None:
...
+            actor, Permission.MANAGE_USER_GROUPS
+        )
+        is_full_admin = has_global_permission(actor, Permission.FULL_ADMIN_PANEL_ACCESS)
+        tags = user_group_permissions(
+            can_manage=can_manage,
+            is_user_groups_admin=is_user_groups_admin,
+            is_full_admin=is_full_admin,
+        )
+        assert tags["manage"] == manage_enforced
+        assert tags["delete"] == is_user_groups_admin  # A — global MANAGE_USER_GROUPS
```
nmgarza5: These three compare the map back to the same booleans that built it, so only `manage` is actually pinned to a guard. That's why the token-limit mistake above sails through, the assertion the file needs is against the route's `require_permission` dependency.

web/src/refresh-pages/admin/GroupsPage/GroupCard.tsx
```diff
@@ -62,13 +67,17 @@ function GroupCard({ group }: GroupCardProps) {
...
               </Text>
             </div>
-            <Button
-              icon={SvgChevronRight}
-              prominence="tertiary"
-              tooltip="View group"
-              aria-label="View group"
-              onClick={() => router.push(`/admin/groups/${group.id}` as Route)}
-            />
+            {canManage && (
```
nmgarza5, line 69: `manage` is `manages_group`, which needs global `MANAGE_USER_GROUPS` or a managed group, but the list itself only needs `READ_USER_GROUPS`. Someone with `MANAGE_LLMS` or `MANAGE_DOCUMENT_SETS` picks up the implied read, still sees every group, and now can't open any of them.

nmgarza5 (approved):

backend/onyx/db/token_limit.py
```diff
@@ -78,6 +78,24 @@ def insert_global_token_rate_limit(
...
+    token_limit = db_session.get(TokenRateLimit, token_rate_limit_id)
+    if token_limit is None:
+        raise ValueError(f"TokenRateLimit with id '{token_rate_limit_id}' not found")
+    if token_limit.scope != TokenRateLimitScope.USER_GROUP:
+        return token_limit.scope, None
+    group_id = db_session.scalar(
+        select(TokenRateLimit__UserGroup.user_group_id).where(
+            TokenRateLimit__UserGroup.rate_limit_id == token_rate_limit_id
+        )
+    )
```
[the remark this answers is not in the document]
Subash-Mohan: Its not valid

PR #13501, opened 2026-07-28
fix(connectors): keep Google Drive checkpoint frontier monotonic across shortcut resolution

justin-tahara:
## Description

Google Drive initial indexing can loop forever and never complete when any user's My Drive contains a shortcut pointing at a file with an older `modifiedTime`.

Drive listings are ordered by `modifiedTime` and resume from the per-user checkpoint frontier (`StageCompletion.completed_until`). When retrieval resolves a shortcut, the target's metadata replaces the shortcut's — so the frontier was being overwritten with the **target's** `modifiedTime`, which can sit far before the shortcut's actual position in the listing:

1. Frontier snaps backward months/years.
2. Next checkpoint cycle rebuilds the listing query with `modifiedTime >= <polluted frontier>`.
3. The saved page token no longer matches the query → Google rejects it → `google_utils` "Invalid page token ... retrying from start of request" silently restarts the listing from the polluted timestamp.
4. Retrieval re-walks the same window (all dedup-skipped), hits the same shortcut, and repeats — a deterministic infinite loop. `new_docs_indexed` flatlines while the attempt heartbeat stays alive.

Observed live on a cloud tenant: 3 of 125 users' My Drives cycled with ~8-minute periods (same `completed_until` + page-token pair recurring), pinning the connector in `INITIAL_INDEXING` for days. A shortcut to a *newer* file has the mirror failure: the frontier jumps forward and silently skips every file in between.

Fixes:
- `_resolve_file_or_shortcut` replaces the resolved target's `modifiedTime` with the shortcut's own, preserving the invariant that a retrieved item's `modifiedTime` lies within the requested time range (checkpoint frontier and re-poll windows both rely on it).
- `_checkpointed_retrieval` never moves the frontier backward within the same stage + drive/folder (legitimate per-drive/folder resets are unaffected).
- Resume starts are clamped to the configured range start (`_resume_start`), so checkpoints already corrupted by this bug recover on upgrade instead of widening the requested time range (frontiers polluted to pre-range timestamps were observed live).

## How Has This Been Tested?

- New unit tests (`test_drive_checkpoint_frontier.py`) replay the failure: a resolved shortcut with an older target mid-listing keeps the frontier at the listing position; an unexplained backward timestamp is clamped; per-drive resets still work; `_resume_start` clamps to range start.
- New unit test in `test_drive_shortcut_following.py` verifies the resolver stamps the listing timestamp while preserving the target's `modifiedTime`.
- Full `backend/tests/unit/onyx/connectors/` + `backend/tests/unit/ee/onyx/external_permissions/google_drive/` suites: 955 passed, 22 skipped, 0 failures.

## Additional Options

---
## Summary by cubic
Fixes an infinite loop and skipped-file risk in Google Drive indexing by keeping the checkpoint frontier monotonic and using the shortcut’s modifiedTime on resolved targets. Resumes are clamped to the configured start across My Drive, Shared Drives, and OAuth.

- **Bug Fixes**
  - On resolved shortcuts, replace the target’s `modifiedTime` with the shortcut’s value so the item stays within the listing range and checkpoint progress stays monotonic.
  - Never move `StageCompletion.completed_until` backward within the same stage and parent (drive/folder) to avoid invalid page tokens and retry loops.
  - Clamp resume starts via `_resume_start` to the configured range start for both service-account and OAuth paths so corrupted checkpoints recover without widening the range.

<sup>Written for commit b225f21bfa8e03bfbefdee093c5853f32ed8917d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13501?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/connectors/google_drive/connector.py
+++ b/backend/onyx/connectors/google_drive/connector.py
@@ -37,6 +37,7 @@ from onyx.connectors.google_drive.doc_conversion import (
     onyx_document_id_from_drive_file,
 )
 from onyx.connectors.google_drive.file_retrieval import (
+    LISTING_MODIFIED_TIME_KEY,
     DriveFileFieldType,
     crawl_folders_for_files,
     get_all_files_for_oauth,
@@ -215,6 +216,15 @@ def _is_shared_drive_root(folder: GoogleDriveFileType) -> bool:
 
     # For shared drive content, the root has id == driveId
     return bool(drive_id and folder_id == drive_id)
+
+
+def _resume_start(
+    completed_until: SecondsSinceUnixEpoch,
+    start: SecondsSinceUnixEpoch | None,
+) -> SecondsSinceUnixEpoch:
+    """Resume from the checkpointed frontier, but never before the configured
+    range start (a corrupted frontier must not widen the requested time range)."""
+    return max(completed_until, start) if start is not None else completed_until
 
 
 def _public_access() -> ExternalAccess:
@@ -946,7 +956,11 @@ class GoogleDriveConnector(
                         field_type=field_type,
                         include_shared_with_me=self.include_files_shared_with_me,
                         max_num_pages=MY_DRIVE_PAGES_PER_CHECKPOINT,
-                        start=curr_stage.completed_until if resuming else start,
+                        start=(
+                            _resume_start(curr_stage.completed_until, start)
+                            if resuming
+                            else start
+                        ),
                         end=end,
                         cache_folders=not bool(curr_stage.completed_until),
                         page_token=curr_stage.next_page_token,
@@ -995,7 +1009,7 @@ class GoogleDriveConnector(
             if resuming:
                 drive_id = curr_stage.current_folder_or_drive_id
                 if drive_id:
-                    resume_start = curr_stage.completed_until
+                    resume_start = _resume_start(curr_stage.completed_until, start)
                     for file_or_token in _yield_from_drive(drive_id, resume_start):
                         if isinstance(file_or_token, str):
                             checkpoint.completion_map[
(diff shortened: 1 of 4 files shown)


justin-tahara:
Addressed in 9bd265a — the flagged OAuth shared-drive resume now goes through `_resume_start`, and the same clamp is applied to the OAuth folder-crawl resume just below it (same pattern, not flagged). All `completed_until` resume sites (service-account my-drive/shared-drive/folder, OAuth all-files/drives/folders) now clamp to the configured range start.

backend/onyx/connectors/google_drive/connector.py
```diff
@@ -217,6 +218,15 @@ def _is_shared_drive_root(folder: GoogleDriveFileType) -> bool:
     return bool(drive_id and folder_id == drive_id)
 
 
+def _resume_start(
```
evan-onyx, line 220: I'd prefer to replace the modifiedtime of the retrieved document with the shortcut's modified time to preserve the invariant of "the modifiedTime of the item we fetched is in the range of modified times we requested". If that's impossible for some reason, then this approach is fine.

evan-onyx (approved):
see comment

backend/onyx/connectors/google_drive/connector.py
justin-tahara: Done in b225f21 — `_resolve_file_or_shortcut` now replaces the target's modifiedTime with the shortcut's, so the retrieved item's modifiedTime always lies within the requested range. The side-channel key is gone and the checkpoint funnel reads plain modifiedTime again; the monotonic clamp stays as defense-in-depth against any other out-of-order source.

PR #13505, opened 2026-07-28
fix(craft): resolve aliased model token limits

wenxi-onyx:
## Description

Fix Craft sessions entering repeated OpenCode compaction when a configured model alias is absent from LiteLLM's capability map.

The gateway catalog now resolves an unrecognized alias through the provider's underlying deployment name, preserving the provider's actual input and output limits. Persisted administrator overrides remain distinguishable from inferred fallback values.

Because the Onyx LLM gateway is a custom OpenCode provider, OpenCode cannot recover limits from models.dev. The generated model configuration therefore includes explicit `context`, `input`, and `output` limits. Supplying the separate input budget is important for OpenCode 1.15.x: without it, OpenCode subtracts output from context during compaction accounting, which can collapse the usable budget to zero. Unknown models retain the repository's existing token fallbacks rather than relying on OpenCode's custom-provider defaults, while partial limit metadata is omitted at the adapter boundary.

This preserves Azure GPT-5's full 128k output capability metadata, applies across providers, and introduces no configuration variables. It builds on the shared gateway model capability catalog merged in #13502.

## Additional Options

### DIFF
--- a/backend/onyx/server/features/build/sandbox/util/opencode_config.py
+++ b/backend/onyx/server/features/build/sandbox/util/opencode_config.py
@@ -18,12 +18,6 @@ from onyx.server.features.build.sandbox.models import (
     CraftMCPServerConfig,
 )
 
-# Fallback output budget for gateway models the litellm map has no entry for
-# (build_onyx_gateway_config derives the real per-model value). 128k matches the
-# recommended Craft models (Claude Fable/Opus 4.8, GPT-5.6) per models.dev;
-# providers enforce their own real caps regardless.
-_GATEWAY_DEFAULT_MAX_OUTPUT_TOKENS = 128_000
-
 # The gateway is an OpenAI-compatible endpoint, wired via opencode's
 # openai-compatible SDK package.
 _OPENAI_COMPATIBLE_NPM = "@ai-sdk/openai-compatible"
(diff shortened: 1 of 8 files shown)


backend/onyx/server/features/build/sandbox/util/opencode_config.py
```diff
@@ -181,11 +175,14 @@ def _build_provider_block(
...
-            # opencode's schema requires both keys when "limit" is present.
+        if model.max_input_tokens is not None and model.max_output_tokens is not None:
+            # OpenCode 1.15.x subtracts output from context when input is absent.
+            # LiteLLM reports input and output budgets independently, so include
+            # input explicitly to keep compaction from collapsing to zero.
             entry["limit"] = {
                 "context": model.max_input_tokens,
-                "output": model.max_output_tokens or _GATEWAY_DEFAULT_MAX_OUTPUT_TOKENS,
+                "input": model.max_input_tokens,
+                "output": model.max_output_tokens,
```
rohoswagger, line 185: can we leave _GATEWAY_DEFAULT_MAX_OUTPUT_TOKENS as a fallback? seems reasonable to leave it in given we rely on litellm for this number

rohoswagger: actually maybe just worth checking what we do in the main app chat and mirroring that here

wenxi-onyx: We use the existing app fallback `GEN_AI_MAX_TOKENS=32_000` in `llm_max_input_tokens` for both input and output fallbacks, which stops the auto-compaction loop bug as well.

rohoswagger (approved):

PR #13506, opened 2026-07-28
feat(cli): attach release binaries to a GitHub release

jmelahman:
## Description

Adds a `github-release` job to the CLI release workflow (`cli/v*.*.*` tags) that cross-compiles the CLI for linux/darwin/windows × amd64/arm64, packages tar.gz archives (zip for Windows) plus a sha256 checksums file, and attaches them to a GitHub release — alongside the existing PyPI wheel and Docker image jobs.

Design notes:

- **Plain `go build` loop, no GoReleaser.** GoReleaser was evaluated first (earlier commits on this branch), but its flagship features can't be used here: OSS GoReleaser only supports plain semver tags (monorepo tag prefixes and the `release.tag` override are Pro-only), and letting it publish would mint a bare `vX.Y.Z` tag upstream — colliding with the main product's tag namespace and triggering `deployment.yml`, which fires on all tags. With release publishing and changelog disabled, GoReleaser was left doing only a build loop + archiving + checksums, so the workflow does that directly with the same GOOS/GOARCH loop idiom as the wheel builds.
- The release is created with `gh release create` against the real `cli/v*` tag. Release notes are generated from commits touching `cli/` since the previous release, where "previous" is resolved from the tag's **ancestry** (`git describe --match 'cli/v*' TAG^`) rather than version sort, so backport releases get a correct changelog range. Creation and asset upload are **separate steps** (`gh release create` without assets, then `gh release upload --clobber`), so a rerun after a partial upload heals the release instead of failing on "release already exists". Existing releases are resolved via the **list endpoint** rather than tag lookup — a pre-existing draft for the tag is invisible to `gh release view` and would otherwise get a duplicate release created over it (the same issue deployment.yml fixed in #12123); a draft is published instead. Prerelease-style tags (e.g. `cli/v1.2.3-rc1`) get `--prerelease`, and `--latest=false` keeps CLI releases from displacing main product releases as the repo's "latest".
- Ldflags follow the `hatch_build.py` contract: `main.version` keeps the `v` prefix, `main.commit` gets the full SHA. Unlike the wheel builds (which keep the symbol table so CI govulncheck can scan them, #12787), the standalone archives are fully stripped (`-s -w`).
- `cli/dist/` added to `.gitignore` (CI and local `uv build` both write there; it was never ignored).
- The job needs only `contents: write` (no AWS/Docker secrets) and runs in parallel with the pypi/docker jobs.

## How Has This Been Tested?

- Ran the exact build loop locally (Go 1.26.4) with a simulated `cli/v9.9.9` tag: produces `onyx-cli_9.9.9_<os>_<arch>.tar.gz`/`.zip` for all six targets plus `onyx-cli_9.9.9_checksums.txt`; the binary runs, has `v9.9.9` and the commit SHA stamped via ldflags, and is fully stripped (`go tool nm`: no symbol section); the Windows zip contains `onyx-cli.exe`.
- Tested the previous-tag resolution against real tag history for the normal case (`cli/v1.2.3` → `cli/v1.2.2`), an out-of-order backport (`cli/v1.2.0` → `cli/v1.1.1`, where version sort would wrongly pick `cli/v1.2.3` and produce an empty range), and confirmed the first-release fallback path.
- Pre-commit (zizmor, actionlint, YAML checks) passes on all changed files.

## Additional Options


rohoswagger (approved):

jmelahman:
@greptile please review

PR #13292, opened 2026-07-22
chore(deployment): stop advertising env-based SSO setup in templates and installers

nmgarza5:
## Description

Since `v4.4.0`, SSO is configured from the admin panel and the legacy env vars no longer enable it. The deployment templates still presented them as the setup path. A fresh-install user set `AUTH_TYPE=google_oauth` + `OAUTH_CLIENT_*` and silently got no Google login (reported on Discord).

- `env.prod.template`: drop the "supported flows" list and active `AUTH_TYPE=google_oauth`. State the current model (auth always on, SSO via admin panel) and keep the legacy vars commented as an upgrade-import group.
- `env.template`: same treatment, drop active `AUTH_TYPE=basic` (inert since v4.4).
- `install.sh` / `install.ps1`: remove the dead "enable basic auth" steps, the orphaned commented-out auth prompt, and the "customize for OAuth/SAML" hints.
- `docker-compose.yml`: fix the stale "select an authentication method" comment.
- Helm: correct the `AUTH_TYPE` and `auth.oauth` comments (import + legacy login/token-refresh only, removal planned in v4.5). Chart 0.7.1 → 0.7.2.

Deliberately unchanged: compose `${AUTH_TYPE:-oidc}` forwarding (pre-v4.4 upgraders relying on the compose default need it to reach the one-time import) and ECS templates.

## Additional Options

- [ ] [Optional] Please cherry-pick this PR to the latest release version.
- [x] [Optional] Override Linear Check

PR #13510, opened 2026-07-28
chore(auth): drop the AUTH_TYPE startup warning

nmgarza5:
## Description

`AUTH_TYPE` is on its way out of deployments (#13292 removed it from the templates and installers). Startup still reads it and logs a warning on every boot when it's set to `disabled` or a retired single-provider mode, telling operators to remove it.

That warning is pointless in both directions. If the variable is gone it never fires, and if it's still there the message is telling people to do something the cleanup is already doing for them.

The mode itself never came from `AUTH_TYPE` — it comes from `MULTI_TENANT`. So this drops the env read entirely and leaves the function doing the one thing it still does, recording whether the deployment is cloud or basic. Renamed `log_auth_mode()` to match.

No auth behaviour changes. The removed code only logged.

## How Has This Been Tested?

`test_verify_auth_setting.py` → `test_auth_startup_checks.py`, since the old name referenced a function that no longer exists and the file always covered `verify_user_auth_secret` too. Two tests: the mode notice over `(multi_tenant, expected_mode)`, and no-warning over the four values that used to warn. 10/10 pass.

`ruff format --check` clean, `ruff check` clean, `ty check` "All checks passed!".

Confirmed the warning fires on every boot in st-dev's running api-server before this change:

```
WARNING users.py 195: AUTH_TYPE='google_oauth' single-provider mode was removed and
Onyx is running as 'basic'. ... Remove AUTH_TYPE and the legacy SSO env vars.
```

## Additional Options


nmgarza5:
closing for now. will completely remove this loggin in a release or two

PR #13518, opened 2026-07-28
feat(db): enumerate tenants across every shard

justin-tahara:
## Description

Follow-up to #13368, which added the tenant → database routing seam. That PR routed
*sessions*. This one fixes *enumeration*, which was still single-database.

`get_all_tenant_ids()` listed schemas from `information_schema.schemata` on the catalog
database. That is right only while every tenant lives there. Once a tenant is moved to
another shard, its schema is gone from the database being asked, so it drops out of the
list — no error, it just stops being handed work. The callers that matter:

- `celery/apps/beat.py` — the `DynamicTenantScheduler`. A dropped tenant stops being
  scheduled for indexing, pruning and sync.
- `monitoring/tasks.py`, `ee/.../cloud/tasks.py`
- `onyxbot/slack/listener.py`, `onyxbot/discord/cache.py`
- `alembic/env.py`, `alembic/run_multitenant_migrations.py`
- four scripts under `backend/scripts/`

Enumeration now unions every configured shard, and `get_tenant_ids_by_shard()` exposes
the grouping. Both are keyed on where a schema physically lives rather than on the
`tenant_shard` map, because the consumer that needs the grouping is migrations, which
must reach the database actually holding the schema. A tenant mid-copy shows up under
both shards and gets migrated in both — the safe way round — while
`get_all_tenant_ids()` dedupes so beat still schedules it once.

Fixing the primitive is what makes the 11 call sites above correct; none of them
needed changing.

Migrations then follow the same axis:

- `-x shard=<name>` pins an alembic run to one database.
- `upgrade_all_tenants` builds one engine per shard instead of running every tenant
  against the default.
- `get_schemas_needing_migration()` takes the shard to read. It reads `alembic_version`
  *inside each schema*, so asking the wrong database reports an already-migrated tenant
  as unmigrated, forever.
- The tenant range filter is applied before partitioning by shard. Filtering per shard
  would select the range once per database.
- Offline mode refuses to run across shards — it emits one SQL script, which can only
  be applied to one database.

Also binds the alembic IAM listener to the URL its engine connects to. It derived
host/port/user from process-wide settings, which mints a token for the wrong host as
soon as a shard is on a different instance.

**Unchanged with one shard**, which is every deployment today: the union degenerates to
the same single query, and every call site keeps its current behaviour.

Resolves the bulk-migration finding raised on #13368 and the all-tenant enumeration gap
deferred there.

## How Has This Been Tested?

New `backend/tests/external_dependency_unit/db/test_shard_enumeration.py` (7 tests)
runs against two real databases:

- a tenant on the second database is still enumerated
- grouping reports where each schema physically lives
- a tenant present on both shards mid-copy is enumerated once
- non-tenant schemas excluded on every shard
- single-shard enumeration unchanged
- `get_schemas_needing_migration` consults the named shard — a tenant at head on
  another shard is reported as needing migration when the default is asked
- `-x shard=<name>` actually changes the database alembic connects to, asserted by
  intercepting `create_async_engine` inside the real alembic → `env.py` path rather
  than by inspecting a generated URL

Mutation-tested — each of the three fixes fails the suite when reverted:

| mutation | result |
|---|---|
| enumerate only the default shard | caught |
| ignore `shard_name` in the migration check | caught |
| drop the shard from the alembic URL | caught |

Also run: the 4 `test_run_multitenant_migrations.py` integration tests (real subprocess
against a live deployment), 466 db/onyxbot unit tests, and the full
`external_dependency_unit/db/` directory — its 17 failures are identical on `main` and
on this branch (pre-existing, unrelated files).

## Additional Options

---
## Summary by cubic
Enumerates tenants across all shards and makes `alembic` and the migration runner shard-aware so scheduling and migrations stay correct after tenants move. Single‑shard deployments behave the same.

- **New Features**
  - `get_all_tenant_ids()` now unions tenants across every shard; `get_tenant_ids_by_shard()` groups them by the database that physically holds each schema. Mid-copy tenants appear on both shards; the union dedupes.
  - Alembic is shard-aware: `-x shard=<name>` targets a specific DB (validated); `upgrade_all_tenants` migrates per shard; `get_schemas_needing_migration(…, shard)` reads the right DB; offline runs require one `shard`. IAM auth binds tokens to the engine’s URL.
  - The multi-tenant migration runner groups tenants by shard and passes `-x shard=<name>` to each batch; batches never span shards.

- **Refactors**
  - Introduced `make_provide_iam_token_async` and reused it in `alembic` and async engines to remove duplicate IAM listeners.
  - Added `validate_shard_name()` to fail fast on bad `-x shard` inputs.

<sup>Written for commit cef0aee07ff4104c0873b7316072381ae9adbff5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13518?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/alembic/env.py
+++ b/backend/alembic/env.py
@@ -1,3 +1,4 @@
+from collections.abc import Callable
 from typing import Any
 from onyx.db.engine.iam_auth import get_iam_auth_token
 from onyx.db.engine.pg_ssl import create_pg_ssl_context
@@ -7,8 +8,10 @@ from onyx.configs.app_configs import POSTGRES_PORT
 from onyx.configs.app_configs import POSTGRES_USER
 from onyx.configs.app_configs import AWS_REGION_NAME
 from onyx.db.engine.shard_registry import ALEMBIC_TARGET_URL_ATTRIBUTE
+from onyx.db.engine.shard_registry import get_shard_spec
+from onyx.db.engine.shard_registry import is_sharded
 from onyx.db.engine.sql_engine import build_connection_string
-from onyx.db.engine.tenant_utils import get_all_tenant_ids
+from onyx.db.engine.tenant_utils import get_tenant_ids_by_shard
 from sqlalchemy import event
 from sqlalchemy import pool
 from sqlalchemy import text
@@ -21,6 +24,7 @@ import logging
 from logging.config import fileConfig
 
 from alembic import context
+from sqlalchemy.ext.asyncio import AsyncEngine
 from sqlalchemy.ext.asyncio import create_async_engine
 from onyx.configs.constants import SSL_CERT_FILE
 from shared_configs.configs import (
@@ -58,15 +62,27 @@ target_metadata = [Base.metadata, ResultModelBase.metadata]
 logger = logging.getLogger(__name__)
 
 
-def connection_url() -> str:
+def connection_url(shard_name: str | None = None) -> str:
     """Database URL for this migration run.
 
-    Defaults to the process-wide POSTGRES_* settings. A caller that has already
-    decided which database to target — notably per-tenant migrations, which must
-    follow the tenant's shard — passes it via `ALEMBIC_TARGET_URL_ATTRIBUTE`.
+    In precedence order: a caller that has already decided which database to target
+    (per-tenant migrations, via `ALEMBIC_TARGET_URL_ATTRIBUTE`), then an explicit
+    shard, then the process-wide POSTGRES_* settings.
     """
-    return (
-        config.attributes.get(ALEMBIC_TARGET_URL_ATTRIBUTE) or build_connection_string()
+    configured = config.attributes.get(ALEMBIC_TARGET_URL_ATTRIBUTE)
+    if configured:
+        return configured
+
+    if shard_name is None:
+        return build_connection_string()
+
+    spec = get_shard_spec(shard_name)
+    return build_connection_string(
+        user=spec.user,
+        password=spec.password,
+        host=spec.host,
+        port=spec.port,
+        db=spec.db,
     )
 
 
@@ -127,7 +143,7 @@ def filter_tenants_by_range(
 
 
 def get_schema_options() -> tuple[
-    bool, bool, bool, int | None, int | None, list[str] | None
+    bool, bool, bool, int | None, int | None, list[str] | None, str | None
 ]:
     x_args_raw = context.get_x_argument()
     x_args = {}
@@ -212,6 +228,13 @@ def get_schema_options() -> tuple[
             "or provide schemas. Cannot run default migration."
         )
 
+    # Pins the run to one physical database. Omitted means the default shard for a
+    # named-schema run, and every shard for upgrade_all_tenants.
+    shard = x_args.get("shard") or None
+    if shard is not None:
+        # Fails here rather than after connecting to the wrong database.
+        get_shard_spec(shard)
+
     return (
         create_schema,
         upgrade_all_tenants,
@@ -219,6 +242,7 @@ def get_schema_options() -> tuple[
         tenant_range_start,
         tenant_range_end,
         schemas,
+        shard,
     )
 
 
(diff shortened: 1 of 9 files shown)


backend/alembic/env.py
```diff
@@ -212,13 +228,21 @@ def get_schema_options() -> tuple[
             "or provide schemas. Cannot run default migration."
         )
 
+    # Pins the run to one physical database. Omitted means the default shard for a
+    # named-schema run, and every shard for upgrade_all_tenants.
+    shard = x_args.get("shard") or None
+    if shard is not None:
+        # Fails here rather than after connecting to the wrong database.
+        get_shard_spec(shard)
```
acaprau: just to be clear it is intended to not use the result of `get_shard_spec` and just use this function as a fail check?

backend/alembic/env.py
```diff
@@ -251,30 +275,113 @@ def do_run_migrations(
...
+def make_provide_iam_token_for_alembic(
+    target_url: str,
+) -> Callable[[Any, Any, Any, Any], None]:
+    """Build a `do_connect` listener bound to one database's coordinates.
+
+    An RDS IAM token is only valid for the host/port/user it was minted for, so the
+    listener has to close over the URL its engine actually connects to rather than
+    re-deriving it from process-wide settings.
+    """
+    url = make_url(target_url)
```
acaprau: this logic feels familiar, didn't you alr implement something like this in your previous pr?

backend/alembic/env.py
```diff
@@ -251,30 +275,113 @@ def do_run_migrations(
...
+        tenants_by_shard = {shard: tenants_by_shard.get(shard, [])}
+
+    if tenant_range_start is None and tenant_range_end is None:
+        return tenants_by_shard
+
+    # The range is positional over *all* tenants, so it has to be applied before
+    # partitioning — filtering per shard would select the range N times over.
+    all_tenants = sorted({t for ts in tenants_by_shard.values() for t in ts})
+    selected = set(
+        filter_tenants_by_range(all_tenants, tenant_range_start, tenant_range_end)
```
acaprau, line 352: just checking that we don't blow up if either the start or end are None or weird values.

backend/alembic/env.py
```diff
@@ -251,30 +275,113 @@ def do_run_migrations(
...
+    tenants_by_shard = get_tenant_ids_by_shard()
+    if shard is not None:
+        tenants_by_shard = {shard: tenants_by_shard.get(shard, [])}
+
+    if tenant_range_start is None and tenant_range_end is None:
+        return tenants_by_shard
+
+    # The range is positional over *all* tenants, so it has to be applied before
+    # partitioning — filtering per shard would select the range N times over.
+    all_tenants = sorted({t for ts in tenants_by_shard.values() for t in ts})
```
acaprau: i hate this.

backend/alembic/run_multitenant_migrations.py
```diff
@@ -252,30 +269,39 @@ def main() -> int:
         return 1
 
     with SqlEngine.scoped_engine(pool_size=5, max_overflow=2):
-        tenant_ids = get_all_tenant_ids()
-        tenant_schemas = [tid for tid in tenant_ids if tid.startswith(TENANT_ID_PREFIX)]
+        tenants_by_shard = {
+            shard_name: [tid for tid in tenant_ids if tid.startswith(TENANT_ID_PREFIX)]
```
acaprau: i'm just calling stuff out here i could be wrong feel free to ignore me but there might be some ancient tenants that do not have this prefix, i feel like i remember seeing that.

backend/alembic/run_multitenant_migrations.py
```diff
@@ -252,30 +269,39 @@ def main() -> int:
...
 
-        schemas_to_migrate = get_schemas_needing_migration(tenant_schemas, head_rev)
-
-    if not schemas_to_migrate:
-        print(
-            f"All {len(tenant_schemas)} tenants are already at head revision ({head_rev})."
-        )
+        # Per shard: alembic_version lives in the schema, so each shard has to be
+        # asked about its own tenants.
+        schemas_by_shard = {
```
acaprau: this is crazy.

acaprau (approved):

PR #13520, opened 2026-07-28
fix(slack): resolve workspace metadata lazily instead of on credential load

Danelegend:
## Problem

Creating a Slack connector can time out with **"Operation timed out after 10 seconds. Check your configuration for errors?"** while another Slack connector is indexing on the same credential — with no error logged anywhere, and nothing actually wrong with the configuration.

`SlackConnector.set_credentials_provider` eagerly called `auth_test()` on `self.client` to populate the workspace URL and Enterprise Grid topology. That client is `OnyxSlackWebClient`, which before every request:

1. acquires a per-credential Redis lock (`connector:slack:credential_<id>:delay_lock`), and
2. sleeps off the shared rate-limit delay (`…:delay`) **while holding that lock**.

Connector validation goes `associate_credential_to_connector` → `validate_ccpair_for_user` → `instantiate_connector` → `set_credentials_provider`, so a user-facing request with a 10s budget ended up waiting on a lock whose TTL is 1800s, behind indexing jobs backing off on Slack 429s.

Observed on a customer cluster with four Slack connectors on one bot token: the delay key never dropped below ~12s and the lock was refreshed continuously, making connector creation fail 100% of the time. The frontend races creation against a 10s timer (`AddConnectorPage.tsx`) and **deletes the connector** when it loses, which is why nothing surfaced as an error — the backend never failed, it was just slow.

## Fix

Resolve the metadata lazily, on first use from the indexing entry points, rather than when credentials are set.

Validation never reads the workspace URL or `grid_*` values, and `validate_connector_settings` already uses the uncoordinated `fast_client` — so after this change the entire creation path is free of the Redis lock and bounded by `FAST_TIMEOUT`.

Indexing behaviour is unchanged: it still uses the coordinated client with full retries, just resolved at the start of the run instead of at credential load.

### Why not just use `fast_client` for that `auth_test`

It's the obvious one-line fix, but the metadata is load-bearing for indexing — `workspace_url` builds document permalinks and `is_grid` gates all Grid team resolution. `fast_client` has `timeout=1` and no 429 coordination, so it would fail *precisely* when the workspace is rate-limited, and the failure is a `logger.warning`. That trades a loud timeout for silently degraded documents.

## Notes

- State collapses into a single frozen `_WorkspaceMetadata` model, assigned in one operation, so readers observe either `None` or fully-formed metadata rather than a partially-populated object.
- The exposed properties are plain accessors — they do **not** resolve on read. Any new entry point that reads them must call `_ensure_workspace_metadata()` first; this is documented on that method.
- Failure semantics preserved: a failed `auth_test` caches empty metadata, so it costs one attempt per credential load rather than retrying per document.

## Tests

New `test_lazy_workspace_metadata.py` pins the guarantee that neither `set_credentials_provider` nor `validate_connector_settings` touches the coordinated client, plus caching, failure-not-retried, credential-swap invalidation, and Grid resolution.

`test_enterprise_grid.py` updated to construct the single value object instead of setting four private fields.

```
943 passed, 22 skipped   # backend/tests/unit/onyx/connectors
```

## Not addressed here

Deliberately scoped to the Slack connector. Still worth separate fixes:

- `OnyxSlackWebClient._perform_urllib_http_request` can spin unbounded — the `if self._redis.exists(self._delay_lock): continue` branch skips the `ONYX_SLACK_LOCK_TOTAL_BLOCKING_TIMEOUT` check.
- Validation has no server-side timeout, so any other slow path still yields the same unhelpful message. Wrapping it in `run_with_timeout` would turn it into an actionable error.
- The frontend deleting a connector the backend is still working on.

---
## Summary by cubic
Resolve Slack workspace metadata lazily to remove creation-time timeouts, and fail closed on resolution errors to avoid widening Slack ACLs. Validation stays off the coordinated client and now verifies Grid users scopes up front.

- **Bug Fixes**
  - Moved `auth_test` and Grid topology into `_ensure_workspace_metadata()` used only by indexing; removed eager resolution from `set_credentials_provider`.
  - Failure handling: let `auth_test`, `list_grid_team_ids`, and `fetch_team_user_emails` raise; cache nothing on failure; only per-team URL lookups swallow errors. Metadata is a frozen `_WorkspaceMetadata`, invalidated on credential change; properties are read-only accessors.
  - Validation: uses the uncoordinated `fast_client`; on Grid, probes `users_list(team_id=…)` after `auth_teams_list` to surface missing `users:read`/`users:read.email` at creation time, within `FAST_TIMEOUT`.
  - Tests: added `test_lazy_workspace_metadata.py`; updated `test_enterprise_grid.py`. Indexing behavior is otherwise unchanged.

<sup>Written for commit f9ce3a9b95674c2425825955e444f0356cbc124c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13520?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/connectors/slack/connector.py
+++ b/backend/onyx/connectors/slack/connector.py
@@ -1061,35 +1061,24 @@ class SlackConnector(
             self._workspace_metadata = self._resolve_workspace_metadata(self.client)
 
     def _resolve_workspace_metadata(self, client: WebClient) -> _WorkspaceMetadata:
-        try:
-            auth_response = client.auth_test()
-        except Exception as e:
-            logger.warning("Failed to get workspace URL from auth_test: %s", e)
-            return _WorkspaceMetadata()
+        """Fails closed: nothing is cached and the index attempt surfaces the error.
+
+        Grid public-channel ACLs derive from ``team_id_to_user_emails``, and
+        ``get_channel_access`` falls back to org-wide ``is_public=True`` when it
+        is empty. Swallowing a failure here would silently widen
+        workspace-scoped permissions, so only cosmetic lookups may degrade.
+        """
+        auth_response = client.auth_test()
 
         url = auth_response.get("url")
         if not auth_response.get("enterprise_id"):
             return _WorkspaceMetadata(url=url)
 
-        try:
-            team_ids = list_grid_team_ids(client)
-        except SlackApiError as e:
-            logger.warning(
-                "auth.teams.list failed on Grid org: %s", e.response.get("error", "")
-            )
-            team_ids = []
-
+        team_ids = list_grid_team_ids(client)
         team_id_to_url = self._fetch_team_urls(client, team_ids) if team_ids else {}
-        team_id_to_user_emails: dict[str, set[str]] = {}
-        if team_ids:
-            try:
-                team_id_to_user_emails = fetch_team_user_emails(client, team_ids)
-            except SlackApiError as e:
-                # Grid public-channel access stays org-wide instead of per-workspace.
-                logger.warning(
-                    "users.list per-team failed on Grid org: %s",
-                    e.response.get("error", ""),
-                )
+        team_id_to_user_emails = (
+            fetch_team_user_emails(client, team_ids) if team_ids else {}
+        )
 
         logger.info(
             "Slack Enterprise Grid detected: teams=%s urls_resolved=%s users_scoped=%s",
@@ -1513,9 +1502,13 @@ class SlackConnector(
                     f"Slack API returned a failure: {error_msg}"
                 )
 
-            # 3) Grid: verify team:read by calling auth.teams.list
+            # 3) Grid: verify team:read, and the users scopes that public-channel
+            # ACLs depend on, so a missing scope fails here instead of at index time.
             if auth_response.get("enterprise_id"):
-                self.fast_client.auth_teams_list(limit=1)
+                teams_response = self.fast_client.auth_teams_list(limit=1)
+                teams = teams_response.get("teams", [])
+                if teams:
+                    self.fast_client.users_list(team_id=teams[0]["id"], limit=1)
 
             # 4) If channels are specified and regex is not enabled, verify each is accessible
             # NOTE: removed this for now since it may be too slow for large workspaces which may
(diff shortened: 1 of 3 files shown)


backend/onyx/connectors/slack/connector.py
```diff
@@ -1016,67 +1037,92 @@ def set_credentials_provider(
...
+    def _resolve_workspace_metadata(self, client: WebClient) -> _WorkspaceMetadata:
         try:
-            auth_response = self.client.auth_test()
-            self._workspace_url = auth_response.get("url")
-            is_grid = bool(auth_response.get("enterprise_id"))
+            auth_response = client.auth_test()
         except Exception as e:
             logger.warning("Failed to get workspace URL from auth_test: %s", e)
-            self._workspace_url = None
+            return _WorkspaceMetadata()
```
nmgarza5: Grid permissions fail open here. A swallowed `auth_test` failure caches empty metadata for the whole run, and public Grid channels then fall back to `is_public=True` in `ee/onyx/external_permissions/slack/channel_access.py`. A transient failure widens workspace-scoped ACLs to org-wide. The swallow predates this PR, but `test_failed_resolution_is_not_retried` now pins it as intended.

Can we fail closed in this PR instead: let this raise, let `list_grid_team_ids` and `fetch_team_user_emails` below propagate too (keep only the per-team URL swallow, links are cosmetic), and flip the new test to expect the raise. The index attempt then fails loudly and the scheduler retries. This doesn't re-introduce the creation timeout. The coordinated client still waits out 429s and only raises once its retry budget is exhausted, at which point the run couldn't have proceeded anyway.

backend/onyx/connectors/slack/connector.py
```diff
@@ -1016,67 +1037,92 @@ def set_credentials_provider(
...
-        self._team_id_to_url = {}
-        self._team_id_to_user_emails = {}
-        if self._is_grid and self.client is not None:
+        url = auth_response.get("url")
+        if not auth_response.get("enterprise_id"):
+            return _WorkspaceMetadata(url=url)
+
+        try:
+            team_ids = list_grid_team_ids(client)
+        except SlackApiError as e:
```
nmgarza5: Related: failures here used to raise out of `set_credentials_provider` and fail connector creation loudly. Now the connector gets created and they surface later as failed scheduled index attempts. I'd rather broken Grid config never creates the connector. `validate_connector_settings` already checks `team:read` via `auth_teams_list(limit=1)`. Can we also probe the per-team users scope there, `users_list(team_id=<first team>, limit=1)` on the fast client, so missing `users:read` / `users:read.email` fails at creation instead of at index time.

nmgarza5 (approved):

PR #13524, opened 2026-07-28
feat(ods): sync onyx-cli's embedded deployment files via generate-compose

jmelahman:
## Description

Split out of #13522 (the `onyx-cli deploy` installer) so the tooling half can land — and `onyx-devtools` can cut a release — ahead of the installer itself.

onyx-cli's upcoming `deploy` commands ship the guided-install deployment files (the generated `docker-compose.yml`, the lite/craft overlays, `env.template`, the nginx config, and the install-root README) inside the binary via `go:embed`. Since `go:embed` cannot reference files outside the cli module, byte-identical copies are checked in under `cli/internal/deploy/deployfiles/embedded/` with a manifest (`deployfiles.All`: embed path, repo path for pinned-tag fetches, install destination, file mode — `run-nginx.sh` keeps 0755, which `embed.FS` drops).

Rather than a separate sync command, the copies are refreshed by **`ods generate-compose --write` itself**, right after it renders the compose variants — the two are one pipeline ("materialize every derived deployment artifact") and must always run together, since the embedded `docker-compose.yml` is a copy of the generated one. Check mode reports stale embedded copies alongside stale variants.

Keeping the sync inside `generate-compose` also simplifies rollout: the existing `docker-compose-sync` pre-commit hook already runs that command from the released `onyx-devtools` package, so no new hook entry is needed — this PR just widens its trigger paths (harmless no-ops until the next ods release ships the folded behavior). Until then, staleness is still gated in CI by `TestEmbeddedFilesMatchSource` in the cli module's own `go test ./...`, which byte-compares every embedded copy against its source under `deployment/`.

Also included:
- `.gitattributes` marks the embedded dir `linguist-generated` so the copies collapse in PR diffs.
- A contributor note in `deployment/docker_compose/README.md`'s "Maintaining the compose files" section.

After this merges: cut an `ods/v*` release so the pre-commit hook picks up the embedded sync, and #13522 rebases down to just the installer commits.

## How Has This Been Tested?

- `go test ./...` in both modules: new `deployfilessync` suite (tempdir fixtures: missing copies are stale, write→check turns clean, source edits re-flag, missing sources error) and the cli drift-gate suite (embedded-vs-source byte compare, manifest↔embed-FS coverage, mode invariants).
- `ods generate-compose --write` run end-to-end (renders variants + syncs copies); check mode verified clean afterward.
- `golangci-lint run` clean in both modules; `pre-commit run` on all touched files — including the `docker-compose-sync` hook itself executing the released ods against the widened triggers as a no-op.

## Additional Options

---
## Summary by cubic
Make `ods generate-compose` also sync the deployment files that `onyx-cli` embeds, so the binary always ships the latest `docker-compose.yml`, overlays, `env.template`, nginx config, and README. Adds drift checks and keeps the copies lightweight in diffs.

- New Features
  - `ods generate-compose --write` now refreshes embedded copies under `cli/internal/deploy/deployfiles/embedded/` after rendering; check mode flags drift for both generated files and embedded copies with compact diffs that show the source path.
  - New `tools/ods/internal/deployfilessync`; CLI tests byte-compare embedded files to sources, verify manifest coverage, and enforce `run-nginx.sh` stays 0755.
  - Pre-commit `docker-compose-sync` widens trigger paths; `.gitattributes` marks the embedded dir `linguist-generated`; docs note the workflow.

- Migration
  - Publish a new `ods` in `onyx-devtools` so the pre-commit hook picks up the embedded sync; until then, the widened triggers are no-ops and `go test` in `cli` gates drift.

<sup>Written for commit 387f699900d4cb2ce5926687b87dbb5f98806fdf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13524?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


cli/internal/deploy/deployfiles/embedded/data/nginx/app.conf.template
```diff
@@ -0,0 +1,136 @@
...
+    # used by the container healthcheck.
+    location = /nginx-health {
+        access_log off;
+        return 200 "ok\n";
+    }
+
+    # SAML login and IdP callbacks are backend routes under /auth/saml (legacy
+    # single-config and the DB-backed multi-provider router). Proxy them to the
+    # API like /scim, otherwise the IdP POST falls through to the web server.
+    location ~ ^/auth/saml(/.*)?$ {
```
justin-tahara, line 55: Not trying to flag you or anything but the amount of custom location things that we do really is not great. I wish we didn't have this pattern since it's so easy for a customer to miss these and some feature just completely doesn't work. Maybe something to consider while we clean this up since this is super annoying to upkeep and maintain

cli/internal/deploy/deployfiles/embedded/docker_compose/docker-compose.craft.yml
justin-tahara, line 1: I feel like we don't need this one since craft will most likely just be part of the product right?

cli/internal/deploy/deployfiles/embedded/docker_compose/env.template
justin-tahara, line 1: One day this thing will be much smaller

justin-tahara (approved):

cli/internal/deploy/deployfiles/embedded/data/nginx/app.conf.template
jmelahman: I think we're on the same page. I would say, the main intent of this change is to make it easier to update these in the future. I haven't looked into this stuff too much, but it would be great to get out of the business of needing a reverse-proxy at all for simple deployments.

cli/internal/deploy/deployfiles/embedded/docker_compose/docker-compose.craft.yml
jmelahman: Maybe. This might change by the final revision circa https://github.com/onyx-dot-app/onyx/pull/13522 . Initial goal is to keep parity with the `install.sh` script which carries these.

PR #13526, opened 2026-07-28
fix(craft): stop the apps page shifting when switching tabs

Danelegend:
## Problem

On `/craft/v1/apps`, switching between the **Apps** and **MCP servers** tabs moved the whole page, and connecting an app moved its card out from under the cursor.

Three causes, all from the two tabs sharing a layout without sharing a shape:

1. **~7px sideways on tab switch.** The panels differ in height, so on a viewport where one tab overflows and the other doesn't, the settings scroll container gains/loses its scrollbar. Where scrollbars take up space (Windows/Linux, or macOS with "always show scrollbars") that narrows the centering box by 15px and re-centers the entire page — header, tabs and cards — by half that. Invisible on default macOS, where scrollbars are overlays.
2. **16–24px vertically on tab switch.** Below ~1150px the MCP blurb wraps to 2–3 lines where the Apps one is 1, so each tab's list started at a different height.
3. **A card jumping on connect.** Connected items rendered as full-width rows in a "Connected" section, unconnected ones as tiles in a "Browse" grid, so connecting an app teleported its card up the page and shoved everything below it. The two variants also disagreed on height (118px vs 114px), leaving the grid ragged.

Screenshot 2026-07-28 at 3 36 22 PM
Screenshot 2026-07-28 at 3 36 24 PM

## Fix

**Stacked slots (commit 1).** Every kind renders through one `KindSlot`, which stacks all kinds in a single CSS grid cell with only the selected one visible (`invisible` + `aria-hidden`). Each slot holds the tallest kind's height, so the page's geometry — and with it the scrollbar — can't change when the tab changes. Used twice: once for the blurb, once for the panels. The selected kind is still a real `Tabs.Content`, so there is exactly one `role="tabpanel"` and Radix's trigger/panel wiring is untouched; an unselected kind renders a stand-in mirroring `Tabs.Content`'s own `pt-4`. `connectParam` only reaches the visible kind, so the `?connect=` deep link can't focus a hidden card.

**One card shape (commit 2).** The row/tile variants collapse into a single `ContentAction` card whose connected state only swaps the status line and the action beside it, in one uniform 2-column grid. Titles and status lines are clamped to one line via `titleMaxLines`/`descriptionMaxLines`, so the card is a fixed height *by construction* rather than by hand-rolled sizing — the inner flex column, `mt-auto`, `min-h-8`, `h-full` and `*:h-full` are gone, and a long app name or wordy server description ellipsizes instead of growing the card. Connected items sort first, then by name. The green "ready" tick next to Disconnect is dropped, since the status line already says Connected.

That last change made some state redundant: with no tick, "skills all enabled" and "skills not loaded" render identically, so the per-app skill counts collapse to a `Set` of ids needing setup, which retires the `skillsLoaded` prop and its guard. `page.tsx` ends up **504 lines, down from 556** on `main`, despite gaining `KindSlot`.

No shared or Opal components were changed — the whole fix is contained to the page.

## Verification

Sideways shift, in a real Chrome launched with `--disable-features=OverlayScrollbar` (i.e. behaving like a space-taking-scrollbar machine), 24 apps vs 1 server so the Apps tab overflows and the MCP tab doesn't:

| | gutter (Apps → MCP) | tab-bar x | shift |
|---|---|---|---|
| before | 15 → 0 | 333 → 340 | **7px** |
| after | 15 → 15 | 333 → 333 | **0px** |

Pixel diff of the header/tab region between the two tabs: **5006** differing pixels before, **0** after.

- Overlay-scrollbar Chromium, 1000–1440px: `scrollHeight` identical per tab at every width, no positional difference in the tab list, panel origin or first section.
- Card geometry: every card on both tabs measures exactly **70px**, including a connected pass-through-OAuth server with no user action at all, a 75-character app name and a 185-character server description (both ellipsize).
- `bun run jest src/app/craft/v1/apps` — 39 passed. Three existing tests asserted the unselected kind was *unmounted*, which is no longer true; they're re-scoped to the active `role="tabpanel"` via an `activeTab()` helper. New tests cover the reserved-space mechanism and the connected-first ordering.

## Trade-offs

- The shorter tab scrolls into blank space, bounded by the taller tab's height. That's the cost of holding geometry page-side rather than reserving a scrollbar gutter in the shared settings layout.
- Connecting an item now reorders it into the connected group. Cards no longer move on their own, but a connect/disconnect does move one — inherent to grouping, and it's a deliberate action.
- Two user-visible copy changes: the skill-setup line is now `Connected · not all associated skills are enabled` with the full sentence in a tooltip on the warning glyph, and the unavailable state reads `Not available for your account — ask an admin` in the status line rather than a wrapped paragraph in the card body.

## Additional Options

### DIFF
--- a/web/src/app/craft/v1/apps/page.tsx
+++ b/web/src/app/craft/v1/apps/page.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useEffect, useMemo, useRef, useState } from "react";
+import { ReactNode, useEffect, useMemo, useRef, useState } from "react";
 import { useSearchParams } from "next/navigation";
 import useSWR from "swr";
 import { errorHandlingFetcher } from "@/lib/fetcher";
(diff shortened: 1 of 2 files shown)


web/src/app/craft/v1/apps/page.tsx
```diff
@@ -221,31 +215,80 @@ function AppConnections({ query }: AppConnectionsProps) {
...
+}
+
+/**
+ * Renders one piece of the page for every kind, stacked in a single grid cell
+ * with only the active kind visible. Apps and MCP servers share the layout but
+ * not their content lengths, so every slot reserves the tallest kind's height:
+ * the page's geometry — and with it the scrollbar, which would otherwise
+ * re-center the whole page sideways — stays put when the tab changes.
+ */
+function KindSlot({ tab, panel, children }: KindSlotProps) {
```
wenxi-onyx, line 257: I feel like this is not necessary to render the two tab lists, but probably fine

wenxi-onyx (approved):

PR #13535, opened 2026-07-28
fix(chat): auto-name chat sessions with the session's model, not the global default

### LINKED ISSUE #8330: Performance Issue: Chat Naming Always Uses Default LLM Instead of Session Model
Analyzed with warp.dev
----

## Description
The automatic chat session naming feature always uses the default LLM provider/model (`get_default_llm()`), even when the user is actively chatting with a different model. This causes unnecessary model loading/unloading on custom GPU servers, significantly impacting performance for self-hosted setups.

## Impact
- **Performance**: Wastes GPU memory and time loading/unloading models
- **User Experience**: Delays in chat responses while default model loads
- **Resource Usage**: Doubles LLM API costs (one for chat, one for naming)
- **Latency**: Especially problematic for large models (20B+) that take time to load

### Code Location
**File**: `backend/onyx/server/query_and_chat/chat_backend.py`  
**Function**: `rename_chat_session()`  
**Lines**: ~1050-1055

```python
@router.put("/rename-chat-session")
def rename_chat_session(
    rename_req: ChatRenameRequest,
    request: Request,
    user: User | None = Depends(current_user),
) -> RenameChatSessionResponse:
    # ... validation code ...
    
    # BUG: Always uses default LLM
    llm = get_default_llm(
        additional_headers=extract_headers(
            request.headers, LITELLM_PASS_THROUGH_HEADERS
        )
    )
    
    # ... rest of naming logic ...
    new_name = generate_chat_session_name(chat_history=simple_chat_history, llm=llm)
```

### What Happens
1. User starts chat with Model A (e.g., `bw-ai1/nvidia/nemotron-3-nano`)
2. Model A loads on GPU
3. User sends message
4. Model A generates response
5. **Chat naming triggers** → Loads Default Model (e.g., `gpt-oss-20b-bw-failover`)
6. Default Model unloads Model A from GPU
7. Default Model generates chat name
8. User sends another message → Model A must reload

For self-hosted GPU servers with limited VRAM, this causes constant model swapping.

### Real-World Example
```
[User Chat] Model: bw-ai1/nvidia/nemotron-3-nano
→ GPU loads nemotron-3-nano (7GB VRAM)
→ User: "What is machine learning?"
→ nemotron-3-nano: [generates response]
→ [CHAT NAMING TRIGGERED]
→ GPU unloads nemotron-3-nano
→ GPU loads gpt-oss-20b-bw-failover (20GB VRAM) 
→ gpt-oss-20b: "Machine Learning Basics" [chat name]
→ User: "Tell me more"
→ GPU unloads gpt-oss-20b
→ GPU loads nemotron-3-nano again (7GB VRAM)
→ nemotron-3-nano: [generates response]
```

# Expected Behavior
The chat naming should **reuse the same LLM** that the user is currently chatting with, avoiding unnecessary model switching.

### Option 1: Use Chat Session's Selected Model (Recommended)
Pass the chat session's active LLM to the naming function instead of always using the default.

**Changes needed**:
```python
@router.put("/rename-chat-session")
def rename_chat_session(
    rename_req: ChatRenameRequest,
    request: Request,
    user: User | None = Depends(current_user),
) -> RenameChatSessionResponse:
    # ... existing code ...
    
    # Get the chat session to retrieve its LLM configuration
    chat_session = get_chat_session_by_id(
        chat_session_id=chat_session_id,
        user_id=user_id,
        db_session=db_session
    )
    
    # Use the same LLM as the chat session
    # Falls back to default if chat session has no specific LLM
    if chat_session.persona and chat_session.persona.llm_model_provider_override:
        llm = get_llm_for_persona(
            persona=chat_session.persona,
            user=user,
            additional_headers=extract_headers(
                request.headers, LITELLM_PASS_THROUGH_HEADERS
            )
        )
    else:
        llm = get_default_llm(
            additional_headers=extract_headers(
                request.headers, LITELLM_PASS_THROUGH_HEADERS
            )
        )
    
    # ... rest of the function ...
```

### Option 2: Configuration to Disable Auto-Naming
Add an environment variable or admin setting to disable automatic chat naming for performance-sensitive deployments.

**Example**:
```python
DISABLE_AUTO_CHAT_NAMING = os.getenv("DISABLE_AUTO_CHAT_NAMING", "false").lower() == "true"

if not DISABLE_AUTO_CHAT_NAMING:
    new_name = generate_chat_session_name(chat_history=simple_chat_history, llm=llm)
```

### Option 3: Async/Background Chat Naming
Move chat naming to a background task so it doesn't block the chat flow or trigger immediate model loading.

## Environment
- **Onyx Version**: `latest` (nightly, as of 2026-02-08)
- **Deployment**: Docker Compose (self-hosted)
- **GPU Setup**: Custom inference servers with limited VRAM
- **Model Size**: 7B-20B+ parameters
- **Provider**: OpenAI-compatible (custom endpoints)

### Chat Naming Prompt
The prompt used for naming (from `/app/onyx/prompts/chat_prompts.py`):
```python
CHAT_NAMING_SYSTEM_PROMPT = """
Given the conversation history, provide a SHORT name for the conversation. Focus the name on the important keywords to convey the topic of the conversation. 
Make sure the name is in the same language as the user's language.

IMPORTANT: DO NOT OUTPUT ANYTHING ASIDE FROM THE NAME. MAKE IT AS CONCISE AS POSSIBLE. NEVER USE MORE THAN 5 WORDS, LESS IS FINE.
"""
```

This is a simple task that doesn't require the default/largest model - the currently active model can handle it just fine.

### Performance Impact Metrics
For a typical chat session with 10 messages:
- **Without fix**: 10+ model load/unload cycles
- **With fix**: 1 model load (stays loaded for entire session)
- **Time saved**: 30-120 seconds per chat session (depending on model size)
- **VRAM churn**: Reduced by 90%+

### Temporary Solutions (until fixed):
1. **Disable auto-naming**: Manually name chats (if UI allows)
2. **Use fast default model**: Set a tiny model as default (e.g., GPT-3.5) for naming only
3. **Use same model everywhere**: Configure all personas to use the same model as default

### Not Recommended:
- Setting `DISABLE_LLM_*` env vars disables other useful features

## Related Issues
- Related to LLM provider/model selection logic
- May affect other "secondary LLM flows" that use `get_default_llm()`

## Reproduction Rate
100% - occurs after every first message in a new chat session

## Labels
- `performance`
- `llm`
- `backend`
- `enhancement`

justin-tahara:
## Description

The `PUT /rename-chat-session` auto-naming path always called `get_default_llm()`, ignoring the session's model override and the persona's configured model. On deployments where the global default provider is gated or unavailable to some users, chat naming silently failed or billed to the wrong model.

Auto-naming now resolves the LLM exactly the way send-message does (`get_llm_for_persona` with `chat_session.llm_override`): session override → persona default → global default. Fetching the chat session up front (with `user_id`) also validates ownership before any LLM call is made.

Fixes #8330.

Follow-up: #13557 adds an optional dedicated naming model (env-configured) for single-stream local-model deployments.

## How Has This Been Tested?

- `pre-commit` (ruff, ruff format, ty) green on the touched file
- Resolution mirrors the existing `process_message` path (`new_msg_req.llm_override or chat_session.llm_override` → `get_llm_for_persona`), which is covered by the chat streaming integration tests

## Additional Options

### DIFF
--- a/backend/onyx/server/query_and_chat/chat_backend.py
+++ b/backend/onyx/server/query_and_chat/chat_backend.py
@@ -70,7 +70,7 @@ from onyx.error_handling.error_codes import OnyxErrorCode
 from onyx.error_handling.exceptions import OnyxError
 from onyx.file_store.file_store import get_default_file_store
 from onyx.llm.constants import LlmProviderNames
-from onyx.llm.factory import get_default_llm, get_llm_for_persona, get_llm_token_counter
+from onyx.llm.factory import get_llm_for_persona, get_llm_token_counter
 from onyx.llm.models import (
     USER_SELECTABLE_REASONING_EFFORTS,
     ReasoningEffort,
@@ -482,18 +482,29 @@ def rename_chat_session(
     # send-message. Manual renames return above and stay free.
     check_token_rate_limits(user)
 
-    llm = get_default_llm(
-        additional_headers=extract_headers(
-            request.headers, LITELLM_PASS_THROUGH_HEADERS
-        )
-    )
-
     # Read-phase short session: usage check + history fetch. Closed before the
     # LLM call so the underlying pool connection is fully released for the
     # 2-10s generation window. (db_session.close() alone is insufficient in
     # multi-tenant mode where the session is bound to an explicit Connection
     # held by get_session_with_tenant's outer with-block.)
     with get_session_with_current_tenant() as db_session:
+        chat_session = get_chat_session_by_id(
+            chat_session_id=chat_session_id,
+            user_id=user_id,
+            db_session=db_session,
+            eager_load_persona=True,
+        )
+        # Name with the model the session actually uses (session override →
+        # persona default → global default) — the global default may be a
+        # provider this user's session deliberately avoids.
+        llm = get_llm_for_persona(
+            persona=chat_session.persona,
+            user=user,
+            llm_override=chat_session.llm_override,
+            additional_headers=extract_headers(
+                request.headers, LITELLM_PASS_THROUGH_HEADERS
+            ),
+        )
         check_llm_cost_limit_for_provider(
             db_session=db_session,
             tenant_id=get_current_tenant_id(),


nmgarza5 (approved):

santomet:
@justin-tahara 
Why don't you add an explicit selection for the chat naming model? That would make much more sense.
Because this is actually a significant downgrade for users who use local models without parallelization - the model will just be stuck.

Until now, a viable workaround was setting some cheap model as the global default so that the naming is fast and doesn't block the model.

justin-tahara:
@santomet Fair point — the single-stream local model case is real, and the old workaround only worked by accident of the global default (and broke naming entirely for deployments whose default provider is gated, which is what this PR fixes).

Rather than reverting to default-model naming, the latest #13557 adds an explicit version of your workaround: set `CHAT_SESSION_NAMING_MODEL_PROVIDER` + `CHAT_SESSION_NAMING_MODEL` and naming is pinned to that model regardless of what the session uses. Unset, naming follows the session's model. Documented in `env.template` and the Helm values.

A full admin-UI model picker for naming overlaps with the previously-removed fast-model feature, so keeping this as deployment config for now — if there's broader demand for per-flow model selection in the UI, that's worth its own issue.

PR #13540, opened 2026-07-28
fix(sharepoint): retry empty/non-JSON Graph responses instead of aborting indexing

### LINKED ISSUE #12554: SharePoint connector: unhandled JSONDecodeError in _graph_api_get_json aborts indexing of large libraries
### Overview 
Indexing a SharePoint connector against a document library with a large number of files (>2K) intermittently fails the entire run with a `requests.exceptions.JSONDecodeError`. The crash originates in `_graph_api_get_json`, where `response.json()` is called without guarding against an empty or non-JSON response body.

`_graph_api_get_json`'s retry loop only catches `requests.ConnectionError` and `requests.Timeout`. When Microsoft Graph returns a 2xx response with an empty/non-JSON body (which happens under sustained load — see "Root cause" below), `_log_and_raise_for_status` passes (status is OK) and `response.json()` raises `JSONDecodeError`, which escapes the loop uncaught and aborts the whole indexing job.

### Traceback
```
File "/app/onyx/connectors/sharepoint/connector.py", line ~2358, in _load_from_checkpoint
    for driveitem in driveitems:
File "/app/onyx/connectors/sharepoint/connector.py", in _iter_drive_items_paged
    data = self._graph_api_get_json(page_url, params)
File "/app/onyx/connectors/sharepoint/connector.py", in _graph_api_get_json
    return response.json()
File ".../requests/models.py", line 982, in json
    raise RequestsJSONDecodeError(e.msg, e.doc, e.pos)
requests.exceptions.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```
(`Expecting value: line 1 column 1 (char 0)` = the body was empty / 0 chars.)

### To reproduce
1. Configure a SharePoint connector pointing at a library with >2,000 files.
2. Run indexing.
3. Intermittently (probabilistically — see below) the run fails with the traceback above.

### Expected behavior
A single transient empty/non-JSON Graph response should be retried, not crash the entire connector run. All documents should index.

### Root cause
`_graph_api_get_json` (current `main`):
```python
            _log_and_raise_for_status(response)
            return response.json()                      # unguarded
        except (requests.ConnectionError, requests.Timeout):   # too narrow
            ...
```
Two gaps:
1. **`response.json()` is unguarded.** Microsoft Graph intermittently returns an empty/non-JSON body with a non-error status under load — throttling shed at the gateway, backend [list-view-threshold](https://learn.microsoft.com/troubleshoot/sharepoint/lists-and-libraries/items-exceeds-list-view-threshold) timeouts on large libraries, or a connection dropped mid-response. Graph's own [best-practices doc](https://learn.microsoft.com/graph/best-practices-concept#handle-responses-effectively) notes 503s under load and advises retrying over a *new* connection — i.e. the body can be absent. `response.json()` then raises `JSONDecodeError`.
2. **The `except` is too narrow.** `ChunkedEncodingError` / `ContentDecodingError` are siblings of `ConnectionError` under `RequestException` (not subclasses), so mid-stream drops here also escape uncaught. (Note: `TRANSIENT_TRANSPORT_EXCEPTIONS` already exists in this module for exactly this reason, but `_graph_api_get_json` doesn't use it.)

**Why it scales with file count:** each page (`$top=200`) is one Graph call with some small probability `p` of a bad response. A 2K-file library = 10+ pages/drive across many drives → hundreds of calls. `1 − (1−p)ⁿ` approaches certainty as `n` grows, so the failure is near-deterministic at scale and rare for small libraries. One bad response kills the whole job because there's no per-call recovery for a non-JSON body.

### Proposed fix
Guard `response.json()` and treat an empty/non-JSON body as a transient, retryable failure; also widen the transport `except` to `TRANSIENT_TRANSPORT_EXCEPTIONS`:

```python
            _log_and_raise_for_status(response)
            try:
                return response.json()
            except (ValueError, requests.exceptions.JSONDecodeError):
                # Graph intermittently returns an empty/non-JSON 2xx body under
                # load; treat it as transient and retry.
                if attempt < GRAPH_API_MAX_RETRIES:
                    wait = _backoff_seconds(attempt, response.headers.get("Retry-After"))
                    logger.warning(
                        "Graph API non-JSON/empty body (status %s, len=%s) on "
                        "attempt %s, retrying in %.1fs: %s",
                        response.status_code, len(response.content),
                        attempt + 1, wait, url,
                    )
                    time.sleep(wait)
                    access_token = self._get_graph_access_token()
                    headers = {"Authorization": f"Bearer {access_token}"}
                    continue
                raise
        except TRANSIENT_TRANSPORT_EXCEPTIONS:   # was (ConnectionError, Timeout)
            if attempt < GRAPH_API_MAX_RETRIES:
                wait = _backoff_seconds(attempt, retry_after=None)
                ...
                continue
            raise
```

### Environment
- onyx version: v4
- Connector: SharePoint
- Affected file: `backend/onyx/connectors/sharepoint/connector.py` — `_graph_api_get_json`

justin-tahara:
## Description

Microsoft Graph intermittently returns an empty or non-JSON 2xx body under load (gateway-shed throttling, backend list-view timeouts on large libraries, mid-response drops). The bare `response.json()` in `_graph_api_get_json` raised an unhandled `JSONDecodeError`, aborting the entire index attempt — repeatedly, on large SharePoint libraries.

Changes:
- Treat empty/non-JSON bodies as transient: retry with the same backoff (shared `_backoff_seconds` helper honoring `Retry-After`) and token re-acquisition as retryable statuses
- Widen the transport-exception handler from `(ConnectionError, Timeout)` to the file's existing `TRANSIENT_TRANSPORT_EXCEPTIONS` set
- After exhausting retries, the error propagates as before

Credit: adopted from #12556 by @CE11-Kishan — re-opened as a maintainer PR so CI can run (fork PRs can't access the required CI secrets).

Fixes #12554.

## How Has This Been Tested?

- New unit test file `test_graph_api_get_json_retry.py` (included in the adopted change) covering non-JSON-body retry, retry exhaustion, and transport-exception retry; full SharePoint unit suite: 144 passed
- `pre-commit` green on touched files

## Additional Options

---
## Summary by cubic
Retry empty or non-JSON 2xx Microsoft Graph responses so SharePoint indexing doesn’t abort under load, especially on large libraries. Improves reliability without changing success/error semantics.

- **Bug Fixes**
  - Treat empty/non-JSON 2xx bodies as transient; retry via `_backoff_seconds`, honoring `Retry-After`, and re-acquire the token on each attempt.
  - Broaden transport exception handling to `TRANSIENT_TRANSPORT_EXCEPTIONS` (e.g., mid-stream drops).
  - After max retries, re-raise the original error.
  - Added unit tests for non-JSON retry, retry exhaustion, and transport-error retry.

- **Refactors**
  - Fold JSON decode failures into the transient-exception path (`ValueError`/`JSONDecodeError`) and remove nested try/except.

<sup>Written for commit 58b1528e8bf660826f5b39821443f2375d1fcc4c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13540?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/connectors/sharepoint/connector.py
+++ b/backend/onyx/connectors/sharepoint/connector.py
@@ -1976,34 +1976,49 @@ class SharepointConnector(
                 )
                 if response.status_code in GRAPH_API_RETRYABLE_STATUSES:
                     if attempt < GRAPH_API_MAX_RETRIES:
-                        parsed_retry_after = parse_retry_after_seconds(
-                            response.headers.get("Retry-After")
+                        wait = _backoff_seconds(
+                            attempt, response.headers.get("Retry-After")
                         )
-                        retry_after = (
-                            parsed_retry_after
-                            if parsed_retry_after is not None
-                            else float(2**attempt)
-                        )
-                        wait = min(retry_after, 60)
                         logger.warning(
-                            "Graph API %s on attempt %s, retrying in %ss: %s",
+                            "Graph API %s on attempt %s, retrying in %.1fs: %s",
                             response.status_code,
                             attempt + 1,
                             wait,
                             url,
                         )
                         time.sleep(wait)
-                        # Re-acquire token in case it expired during a long traversal
                         access_token = self._get_graph_access_token()
                         headers = {"Authorization": f"Bearer {access_token}"}
                         continue
                 _log_and_raise_for_status(response)
-                return response.json()
-            except (requests.ConnectionError, requests.Timeout):
+                try:
+                    return response.json()
+                except (ValueError, requests.exceptions.JSONDecodeError):
+                    # Graph intermittently returns an empty/non-JSON 2xx body
+                    # under load; treat it as transient and retry.
+                    if attempt < GRAPH_API_MAX_RETRIES:
+                        wait = _backoff_seconds(
+                            attempt, response.headers.get("Retry-After")
+                        )
+                        logger.warning(
+                            "Graph API non-JSON/empty body (status %s, len=%s) on "
+                            "attempt %s, retrying in %.1fs: %s",
+                            response.status_code,
+                            len(response.content),
+                            attempt + 1,
+                            wait,
+                            url,
+                        )
+                        time.sleep(wait)
+                        access_token = self._get_graph_access_token()
+                        headers = {"Authorization": f"Bearer {access_token}"}
+                        continue
+                    raise
+            except TRANSIENT_TRANSPORT_EXCEPTIONS:
                 if attempt < GRAPH_API_MAX_RETRIES:
-                    wait = min(2**attempt, 60)
+                    wait = _backoff_seconds(attempt, retry_after=None)
                     logger.warning(
-                        "Graph API connection error on attempt %s, retrying in %ss: %s",
+                        "Graph API transport error on attempt %s, retrying in %.1fs: %s",
                         attempt + 1,
                         wait,
                         url,
--- a/backend/onyx/connectors/sharepoint/connector.py
+++ b/backend/onyx/connectors/sharepoint/connector.py
@@ -1963,10 +1963,10 @@ class SharepointConnector(
         params: dict[str, str] | None = None,
     ) -> dict[str, Any]:
         """Make an authenticated GET request to the Graph API with retry."""
-        access_token = self._get_graph_access_token()
-        headers = {"Authorization": f"Bearer {access_token}"}
-
         for attempt in range(GRAPH_API_MAX_RETRIES + 1):
+            # Tokens can expire during long traversals — re-acquire per attempt.
+            access_token = self._get_graph_access_token()
+            headers = {"Authorization": f"Bearer {access_token}"}
             try:
                 response = requests.get(
                     url,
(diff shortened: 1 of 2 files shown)


backend/onyx/connectors/sharepoint/connector.py
```diff
@@ -1976,34 +1976,49 @@ def _graph_api_get_json(
...
+                            "Graph API non-JSON/empty body (status %s, len=%s) on "
+                            "attempt %s, retrying in %.1fs: %s",
+                            response.status_code,
+                            len(response.content),
+                            attempt + 1,
+                            wait,
+                            url,
+                        )
+                        time.sleep(wait)
+                        access_token = self._get_graph_access_token()
```
evan-onyx: if there's an explicit reason this tokenneeds to be refreshed, let's just put it inside the for loop above. If there isn't a good reason, just remove it

backend/onyx/connectors/sharepoint/connector.py
```diff
@@ -1976,34 +1976,49 @@ def _graph_api_get_json(
...
+                        headers = {"Authorization": f"Bearer {access_token}"}
+                        continue
+                    raise
+            except TRANSIENT_TRANSPORT_EXCEPTIONS:
```
evan-onyx: the original except case already handled the retry, I'd say make a slightly larger category of exceptions here that includes ValueError and JsonDecode so you don't need the above block. if you want some different logging feel free to case on the excepted error, but imo just logging the actual error is fine

backend/onyx/connectors/sharepoint/connector.py
```diff
@@ -1976,34 +1976,49 @@ def _graph_api_get_json(
...
                         time.sleep(wait)
-                        # Re-acquire token in case it expired during a long traversal
                         access_token = self._get_graph_access_token()
                         headers = {"Authorization": f"Bearer {access_token}"}
                         continue
                 _log_and_raise_for_status(response)
-                return response.json()
-            except (requests.ConnectionError, requests.Timeout):
+                try:
+                    return response.json()
```
evan-onyx: see comment below, we shouldn't need another layer of try-except here

evan-onyx (approved):
seems like a valid thing to address, just have reservations on style (we have an existing retry mechanism which should be reused)

backend/onyx/connectors/sharepoint/connector.py
justin-tahara: There is a reason (tokens can expire mid-way through long traversals — that comment predates this PR), so moved acquisition to the top of the retry loop as suggested; the duplicated re-acquire blocks are gone.

backend/onyx/connectors/sharepoint/connector.py
justin-tahara: Done — dropped the inner try/except and widened the existing handler to `TRANSIENT_TRANSPORT_EXCEPTIONS + (ValueError,)` (requests' JSONDecodeError subclasses ValueError), logging the actual error in the warning.

backend/onyx/connectors/sharepoint/connector.py
justin-tahara: Removed — single try/except now, see reply below.

PR #13545, opened 2026-07-29
fix(craft): honor the admin-configured default model

### LINKED ISSUE #13499: Allow admins to configure the default Craft model. Scheduled task won't work because of Claude 5 adaptive-thinking bug
Craft currently selects the provider’s recommended model instead of respecting the models enabled or selected by the administrator. For example, Claude 5 is selected automatically even when it is not enabled as an Anthropic model.
Expected behavior: Add a separate “Default Craft model” setting, or make Craft honor the configured default and visible model list.

The Claude 5 adaptive-thinking bug may already be fixed in v4.5.0-beta.0 in this [commit](https://github.com/onyx-dot-app/onyx/commit/424d3d8a75dc7162f03c3365ed71fc5b6664bd4c), but the stable version still has this issue, and administrators still need control over which model Craft uses by default.

rohoswagger:
## Description

Fixes #13499.

Craft resolved its default model from the built-in recommendation list (`recommended-models.json`), falling back to the alphabetically-first visible model. It never consulted the org's **configured default model** — the `LLMModelFlowType.CHAT` default an admin sets via `POST /admin/llm/default`, which regular chat already honors through `get_default_llm`. So an admin who had deliberately picked a default model still got Craft running on whatever the recommendation list named, with no way to steer it.

Concretely, on the current stable line an org that had not enabled `claude-opus-5` could still see Craft select it, because the pre-gateway `_recommended_model` returned the recommendation for the provider *type* without checking visibility at all. The visibility hole was already closed by the Onyx-gateway rework (#13249), which added an `in visible` guard. What remained — and what this PR fixes — is that admin intent was still ignored entirely.

**Resolution order is now:** the session's stored pick → the configured default → the built-in recommendation → first visible model.

The configured default is validated against the calling user's accessible + visible providers using the same matching logic as a stored pick (extracted into `_matching_visible_model`), so it can never select a model the admin has not enabled or a provider the user cannot access. `fetch_default_llm_model` is a global, unscoped read; re-validating it against `fetch_all_accessible_llm_providers` output is what keeps it safe.

Scope notes:

- Orgs with no configured default get byte-identical behavior to before.
- Existing sessions are unaffected: they already carry a stored `agent_provider`/`agent_model`, which still resolves first. The configured default applies to new sessions (and to sessions whose stored pick has since become inaccessible or hidden).
- All Craft entry points — interactive turns, session create, and scheduled tasks — share `SessionManager.build_llm_configs`, so they all pick this up.
- The Craft model picker mirrors the same order client-side, using the `default_text` that `useLLMProviders()` already returns, so the picker keeps showing the model a new session will actually use rather than disagreeing with the backend.

A dedicated "Default Craft model" setting — a new `LLMModelFlowType` plus admin UI — remains a possible follow-up. This PR intentionally reuses the existing shared chat default instead, since that needs no migration, no new endpoint, and no new admin surface, and it resolves the reported loss of admin control on its own.

## How Has This Been Tested?

Automated:

- New backend unit tests in `test_session_gateway_config.py`: the configured default wins over a competing recommendation, and a configured default whose model is not visible correctly falls through to the recommendation.
- New frontend unit tests in `constants.test.ts` covering the same two cases for the picker mirror.
- `uv run pytest backend/tests/unit/onyx/server/features/craft backend/tests/unit/onyx/server/gateway` — 697 passed.
- `bun run test src/app/craft` — 222 passed across 30 suites.
- `pre-commit run --files <changed>` — clean (ty, ruff, ruff format, oxfmt, oxlint, TypeScript type check).

Manually verified while tracing the change:

- `fetch_all_accessible_llm_providers` filters via `can_user_access_llm_provider`, and it builds views with `include_api_key=False`, so an out-of-reach configured default can neither be selected nor leak a key.
- Existing sessions keep their model, confirmed by tracing `session_llm_config` → `parse_agent_selection` → the stored-selection branch.

## Known minor items (left for the reviewer)

Surfaced by review, deliberately not changed here:

- The frontend test "ignores a configured default that is not visible" also passes against pre-change code, because the old single-argument signature simply dropped the second argument. It is kept because it still pins the visibility guard against a *future* regression (someone trusting `configuredDefault.model_name` without re-checking `is_visible`), which is the behavior worth protecting.
- `build_llm_configs` now issues one extra single-row query (`fetch_default_llm_model`) per call. It runs once per turn, never in a loop, and its result is always re-validated, so the added load is negligible.
- A legacy session row with both `agent_provider` and `agent_model` null would adopt the configured default on its next reconcile. That is a pre-existing nullable-column edge case rather than something this change introduces, and the new behavior is the better of the two outcomes.

## Additional Options

### DIFF
--- a/backend/onyx/server/features/build/session/llm_config.py
+++ b/backend/onyx/server/features/build/session/llm_config.py
@@ -81,26 +81,43 @@ def parse_agent_selection(
     return None
 
 
+def _matching_visible_model(
+    providers: list[LLMProviderView], selection: AgentSelection
+) -> tuple[int, str] | None:
+    for provider in providers:
+        matches_request = (
+            provider.id == selection.provider_id
+            if isinstance(selection, GatewaySelection)
+            else provider.provider == selection.provider_type
+        )
+        if matches_request and any(
+            model.is_visible and model.name == selection.model_name
+            for model in provider.model_configurations
+        ):
+            return provider.id, selection.model_name
+    return None
+
+
 def _select_gateway_default(
     providers: list[LLMProviderView],
     selection: AgentSelection | None,
+    configured_default: GatewaySelection | None = None,
 ) -> tuple[int, str] | None:
     if selection is not None:
-        for provider in providers:
-            matches_request = (
-                provider.id == selection.provider_id
-                if isinstance(selection, GatewaySelection)
-                else provider.provider == selection.provider_type
-            )
-            if matches_request and any(
-                model.is_visible and model.name == selection.model_name
-                for model in provider.model_configurations
-            ):
-                return provider.id, selection.model_name
+        resolved = _matching_visible_model(providers, selection)
+        if resolved is not None:
+            return resolved
         logger.warning(
             "Requested Craft gateway provider/model is not accessible or visible; "
             "falling back"
         )
+
+    # The admin's configured default model (shared with chat) outranks the
+    # built-in recommendation.
+    if configured_default is not None:
+        resolved = _matching_visible_model(providers, configured_default)
+        if resolved is not None:
+            return resolved
 
     # Auto-pick a default: prefer each provider's recommended default model
     # (recommended-models.json, kept current by the update-recommended-models
@@ -123,12 +140,15 @@ def _select_gateway_default(
 def build_onyx_gateway_config(
     gateway_providers: list[LLMProviderView],
     selection: AgentSelection | None = None,
+    configured_default: GatewaySelection | None = None,
 ) -> CraftLLMProviderConfig | None:
     if not ONYX_SERVER_URL:
         return None
 
     models = build_gateway_model_catalog(gateway_providers)
-    default_selection = _select_gateway_default(gateway_providers, selection)
+    default_selection = _select_gateway_default(
+        gateway_providers, selection, configured_default
+    )
     if not models or default_selection is None:
         return None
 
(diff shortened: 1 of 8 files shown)


backend/onyx/server/features/build/session/llm_config.py
```diff
@@ -81,27 +81,44 @@ def parse_agent_selection(
...
+    providers: list[LLMProviderView], selection: AgentSelection
+) -> tuple[int, str] | None:
+    for provider in providers:
+        matches_request = (
+            provider.id == selection.provider_id
+            if isinstance(selection, GatewaySelection)
+            else provider.provider == selection.provider_type
+        )
+        if matches_request and any(
+            model.is_visible and model.name == selection.model_name
```
Danelegend, line 94: Are UserGroups and permissions for who can see what model considered?

Danelegend (approved):

backend/onyx/server/features/build/session/llm_config.py
rohoswagger: No code change is needed. UserGroup and permission access is already enforced before this matcher: SessionManager.build_llm_configs calls fetch_all_accessible_llm_providers(db_session, user), which applies can_user_access_llm_provider for is_public, UserGroup membership, MANAGE_LLMS, and persona restrictions. _matching_visible_model then accepts a configured default only when its provider is in that filtered list and the model is_visible. The gateway also repeats fetch_accessible_llm_provider_by_id plus the visible-model check for every request. Focused verification passed: backend 23 tests and frontend 18 tests.

PR #13551, opened 2026-07-29
feat(connector-stability): slack capability checks

acaprau:
## Description

Reworks the Slack capability checks, reviewed here pre-gateway, onto the merged SlackSourceOperations gateway.

- 14 named checks composing typed gateway operations: 9 INDEXING (OSS registry) and 5 DOC_PERMISSION_SYNC (EE hook). No EXTERNAL_GROUP_SYNC checks: Slack group sync is unregistered by design, so the capability stays NOT_APPLICABLE.
- Fixes the review P1: a new required private-history check reads one message page and one thread from a private channel, so a token with channels:history but not groups:history no longer passes while private channels are in scope.
- Private-channel listing is warning-level (required=False) under BOTH capabilities: indexing and doc sync each silently fall back to public-only without groups:read, so the missing scope degrades the badge instead of failing it. The new slack_perm_sync_private_channel_listing check surfaces the perm-sync consequence: private-channel access lists stop being created or updated, and previously indexed private documents keep stale ones. The required private probes (history, members) skip only on that missing scope; every other listing error maps normally, so a rate-limited listing is INDETERMINATE rather than a false pass.
- Net-new probes beyond the original 11 checks: conversations.replies (public and private threads), conversations.info (checkpoint resume), users.info (author resolution and doc sync's external-user fallback), and a perm-sync channel enumeration check.
- The runner constructs the registered gateway (build_db_credentials_provider plus the supplied config) and hands it to checks via context.source_operations.
- check_auth exposes granted_scopes parsed from the X-OAuth-Scopes response header; the join-scope check reads it, keeping conversations.join side-effect-free and untested.
- Drops the nine temporary untested stamps; only the four permanent exemptions remain (join_channel, fetch_team_info, both usergroups ops). The auto-discovering coverage harness now enforces every non-exempt (operation, variant, capability) unit for Slack.
- fetch_user_info sheds its EXTERNAL_GROUP_SYNC tag: its only group-sync caller is the dormant, unregistered group_sync.py path.
- Carries the two comment-only review items: the regex config-shape boundary note and the PATCH-path validation TODO.
- Behavior note: slack_configured_channels_visible is required and config-dependent, so a config-less credential-time run aggregates INDEXING to SKIPPED until connector binding re-runs checks with a config.

Required permissions (for docs):
- channels:read -> INDEXING (list public channels, read channel metadata), DOC_PERMISSION_SYNC (enumerate channels for ACLs)
- groups:read -> INDEXING (list private channels), DOC_PERMISSION_SYNC (private member ACLs); both silently fall back to public-only without it, and previously indexed private documents keep stale access lists
- channels:history / groups:history -> INDEXING (message and thread reads; private channels fail mid-indexing without groups:history)
- channels:join -> INDEXING (self-join public channels; warning-level, manual invites work)
- team:read -> INDEXING and DOC_PERMISSION_SYNC on Enterprise Grid (workspace enumeration)
- users:read -> INDEXING (author names; degrades to raw ids), DOC_PERMISSION_SYNC (member resolution)
- users:read.email -> DOC_PERMISSION_SYNC (member-to-email mapping)

## How Has This Been Tested?

47-test unit suite driving each check against an autospecced gateway; updated gateway pins (untested inventory, tags, scope parsing); runner gateway-construction tests; EE hook registration tests.

## Additional Options

---
## Summary by cubic
Adds Slack-specific capability checks for INDEXING and DOC_PERMISSION_SYNC and wires them into the registries, so capability reports run real Slack API probes through the gateway without changing production indexing or sync behavior.

- Registers named INDEXING checks: bot token validity (`auth.test` with `xoxb-` guard), public/private channel listing, message history and thread reads, `channels:join` via granted scopes (no side effects), Enterprise Grid workspace listing, user profile read, and a config-gated “configured channels visible” check with a longer timeout for full enumeration.
- Registers named DOC_PERMISSION_SYNC checks: channel listing (required), private-channel listing (warning; public-only fallback), user email visibility (`users:read.email`, content-based), private-channel member listing plus profile resolve for externals, and Grid public-channel scoping (warning).
- Runner now constructs registered `SlackSourceOperations` with a DB credentials provider and passes it to checks; unmigrated sources receive no gateway. `auth.test` parses `X-OAuth-Scopes` to expose `granted_scopes` for the join-scope check.
- Slack EXTERNAL_GROUP_SYNC remains not applicable; removes that tag from `fetch_user_info`. Tightens operation coverage by dropping temporary “untested” stamps; only permanent exemptions remain. Adds tests for registry wiring, gateway construction in context, scope parsing, and end-to-end verdicts.

<sup>Written for commit 02531b07288ddaef27d209f94e65c4e13c3be4e4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/onyx-dot-app/onyx/pull/13551?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/onyx/connectors/slack/capability_checks.py
+++ b/backend/onyx/connectors/slack/capability_checks.py
@@ -1,0 +1,883 @@
+"""Capability checks for the Slack connector.
+
+Each check probes one permission assumption the connector makes at indexing or
+doc-permission-sync time by composing the same gateway operations production
+code calls (``limit=1`` where possible). Checks need no connector instance: the
+runner constructs the registered ``SlackSourceOperations`` gateway from the
+credential and hands it to every check via the context, so checks run
+config-less at credential-creation time (except the configured-channels check,
+which declares its config requirement and is skipped until one exists).
+
+Slack registers no EXTERNAL_GROUP_SYNC checks: group sync is unregistered for
+Slack by design (``sync_params.py``), so that capability stays NOT_APPLICABLE.
+
+Scope-to-capability mapping (see also the OAuth scope list in
+``ee/onyx/server/oauth/slack.py``):
+
+- ``channels:read``    -> INDEXING (public channel listing and metadata),
+  DOC_PERMISSION_SYNC (channel enumeration for ACLs)
+- ``groups:read``      -> INDEXING (private channels), DOC_PERMISSION_SYNC
+  (private member ACLs); both silently fall back to public-only without it
+- ``channels:history`` / ``groups:history`` -> INDEXING (message and thread
+  reads)
+- ``channels:join``    -> INDEXING (self-join public channels)
+- ``team:read``        -> INDEXING and DOC_PERMISSION_SYNC on Enterprise Grid
+  (workspace enumeration)
+- ``users:read``       -> INDEXING (author names; degrades to raw ids),
+  DOC_PERMISSION_SYNC (member-to-profile resolution)
+- ``users:read.email`` -> DOC_PERMISSION_SYNC (member-to-email resolution)
+"""
+
+from typing import Any, NoReturn
+
+from onyx.connectors.capability_checks.models import (
+    CapabilityCheck,
+    CapabilityCheckContext,
+    CredentialCapability,
+)
+from onyx.connectors.exceptions import (
+    ConnectorValidationError,
+    CredentialExpiredError,
+    CredentialInvalidError,
+    InsufficientPermissionsError,
+    UnexpectedValidationError,
+)
+from onyx.connectors.slack.connector import (
+    get_channels,
+    get_channels_across_teams,
+    list_grid_team_ids,
+)
+from onyx.connectors.slack.source_operations import (
+    SlackApiError,
+    SlackAuthTestResponse,
+    SlackChannelVariant,
+    SlackSourceOperations,
+)
+
+_SLACK_DOCS_LINK = (
+    "https://docs.onyx.app/admins/connectors/official/slack/slack_indexed"
+)
+
+_ADD_SCOPE_REMEDIATION = (
+    "Add the `{scope}` bot scope under OAuth & Permissions in the Slack app "
+    "settings and reinstall the app to the workspace."
+)
+
+# Page size when scanning for a channel to probe message history against; a
+# single page is plenty since any public channel works for the probe.
+_HISTORY_PROBE_CHANNEL_PAGE_SIZE = 20
+
+# Page size when sampling users for email visibility. Workspaces commonly have
+# many bot/deactivated users, so sample enough to nearly always include a human
+# member.
+_EMAIL_PROBE_PAGE_SIZE = 100
+
+# Hang guard for the full-workspace channel enumeration, the one multi-page
+# probe: triple the 600s ``CAPABILITY_CHECK_TIMEOUT_SECONDS`` default, since
+# the gateway's redis-coordinated client also serves any concurrently running
+# indexing job's rate-limit backoff.
+_CHANNEL_ENUMERATION_TIMEOUT_SECONDS = 1800.0
+
+
+def _slack_client(context: CapabilityCheckContext) -> SlackSourceOperations:
+    """Returns the gateway the runner constructed for this run.
+
+    A registered gateway that was not constructed is a programmer error in the
+    runner, not a check outcome.
+    """
+    assert isinstance(context.source_operations, SlackSourceOperations), (
+        "The runner constructs the registered gateway for migrated sources."
+    )
+    return context.source_operations
+
+
+def _raise_for_slack_api_error(
+    e: SlackApiError, missing_scope_message: str
+) -> NoReturn:
+    """Maps a Slack API error onto the validation-exception family.
+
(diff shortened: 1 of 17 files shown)


backend/onyx/connectors/slack/connector.py
```diff
@@ -1427,6 +1427,11 @@ def validate_connector_settings(self) -> None:
         Channel existence (for non-regex includes) is validated during indexing
         via filter_channels, not here.
         """
+        # Config-shape validation, load-bearing at creation time: unlike the
```
evan-onyx, line 1404: turn ai slop into a human readable and token efficient comment pls :D

backend/onyx/server/documents/connector.py
```diff
@@ -1491,6 +1491,12 @@ def update_connector_from_model(
     except ValueError as e:
         raise HTTPException(status_code=400, detail=str(e))
 
+    # TODO(andrei, evan): Unlike the creation flows, this update path never
```
evan-onyx: pls make this more concise, rlly all that needs to be communicated is "we should validate here too" and sure can include the funcs we might call

backend/onyx/connectors/slack/capability_checks.py
```diff
@@ -0,0 +1,547 @@
...
+# single page is plenty since any public channel works for the probe.
+_HISTORY_PROBE_CHANNEL_PAGE_SIZE = 20
+
+# Page size when sampling users for email visibility. Workspaces commonly have
+# many bot/deactivated users, so sample enough to nearly always include a human
+# member.
+_EMAIL_PROBE_PAGE_SIZE = 100
+
+
+def _build_client(context: CapabilityCheckContext) -> WebClient:
```
evan-onyx: this should be unified with the non-redis way the slack connector builds its client. probably good to have i.e. the connectionerror retryhandler in there so if that breaks the validation also fails

backend/onyx/connectors/slack/capability_checks.py
```diff
@@ -0,0 +1,547 @@
...
+            "Slack credential is missing the `slack_bot_token` field."
+        )
+    # Deliberately a plain client with the SDK's default timeout: capability
+    # checks have no run budget, so no FAST_TIMEOUT-style short timeouts here.
+    return WebClient(token=token)
+
+
+def _raise_for_slack_api_error(
+    e: SlackApiError, missing_scope_message: str
+) -> NoReturn:
```
evan-onyx, line 96: neat, I haven't seen this return type before

backend/onyx/connectors/slack/capability_checks.py
```diff
@@ -0,0 +1,547 @@
...
+            f"Slack bot token has been revoked ({error})."
+        ) from e
+    if error == "ratelimited":
+        raise UnexpectedValidationError(
+            "Slack rate limited the check; re-run the checks in a minute."
+        ) from e
+    raise UnexpectedValidationError(f"Unexpected Slack error `{error}`.") from e
+
+
+def _auth_test(client: WebClient) -> Any:
```
evan-onyx: dict[str, Any] probably since it seems thats how you're using the response

backend/onyx/connectors/slack/capability_checks.py
```diff
@@ -0,0 +1,547 @@
...
+    """First workspace id of an Enterprise Grid org, or None off-Grid.
+
+    On Grid org installs ``users.list`` requires a ``team_id``, so user-facing
+    checks call this to mirror the production call shape.
+    """
+    auth_response = _auth_test(client)
+    if not auth_response.get("enterprise_id"):
+        return None
+    try:
+        teams_response = client.auth_teams_list(limit=1)
```
evan-onyx: I'm definitely worried about drift between the connector file and this one. I can very easily see edits being made and i.e. functions getting added to the connector that are not tested here, or functions getting removed because they end up unnecessary, but they stick around here. Basically we now need to maintain two sets of slack calls...

I'm in favor of some mechanism for easily collecting stuff to be tested to keep these files synced. I haven't thought about it too much and I'm sure you can think of a better way, but one option would be:

factor out all client calls into wrapper functions, each of the wrappers gets an @validate_capability decorator that registers the func as something that should be tested in a dict or something. Then This file pulls from that dict for each validation function. The registering mechanism could keep track of all the wrappers that were called, and the validation caller could have some baked in mechanism for ensuring that all registered functions are called at least once during the validation process.

^^ sorry this is a little deranged. Also in practice there are different conditions under which different client calls should be tested which map pretty neatly to the capabilities you've defined already, i.e. "function A needs to be called to verify credentials, B for indexing, C for permission sync" etc

evan-onyx (approved):
It can go in but we should probably yap in person about drift between connector and validation files

acaprau:
@greptile @cubic

